Issue #3494, opened 2019-08-13
Build fails with resource errors


rihadavid:
### Steps to Reproduce

1. Create the sample single view android project
2. Clean the project
3. Build the project
4. Open content_main.axml and change anything in the file - for example, change "Hello world!" to "Hello Xamarin!" in the TextView
5. Save all files
6. Edit the file again
7. Build without saving
8. The build fails with those errors and any ruther attempts to build fail again. You need to clean and rebuild.
```
1>  C:\Users\David\source\repos\App15\App15\Resources\values\styles.xml(2): error APT0000: Error retrieving parent for item: No resource found that matches the given name 'Theme.AppCompat.Light.DarkActionBar'.
1>  C:\Users\David\source\repos\App15\App15\Resources\values\styles.xml(2): error APT0000: No resource found that matches the given name: attr 'colorAccent'.
1>  C:\Users\David\source\repos\App15\App15\Resources\values\styles.xml(2): error APT0000: No resource found that matches the given name: attr 'colorPrimary'.
1>  C:\Users\David\source\repos\App15\App15\Resources\values\styles.xml(2): error APT0000: No resource found that matches the given name: attr 'colorPrimaryDark'.
1>  C:\Users\David\source\repos\App15\App15\Resources\values\styles.xml(2): error APT0000: No resource found that matches the given name: attr 'windowActionBar'.
1>  C:\Users\David\source\repos\App15\App15\Resources\values\styles.xml(2): error APT0000: No resource found that matches the given name: attr 'windowNoTitle'.
1>  C:\Users\David\source\repos\App15\App15\Resources\values\styles.xml(2): error APT0000: Error retrieving parent for item: No resource found that matches the given name 'ThemeOverlay.AppCompat.Dark.ActionBar'.
1>  C:\Users\David\source\repos\App15\App15\Resources\values\styles.xml(2): error APT0000: Error retrieving parent for item: No resource found that matches the given name 'ThemeOverlay.AppCompat.Light'.
```

Seems similar to #2427 

### Version Information

Microsoft Visual Studio Community 2019
Version 16.2.1
VisualStudio.16.Release/16.2.1+29201.188
Microsoft .NET Framework
Version 4.7.03190

Installed Version: Community

Azure App Service Tools v3.0.0   16.2.290.48383
Azure App Service Tools v3.0.0

C# Tools   3.2.0-beta4-19359-03+15b43b33901c88f68ef43f8314b5a2457716780d
[...]

PR #3501, opened 2019-08-14
[Xamarin.Android.Build.Tasks] XA5302 warning for parallel builds

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/issues/3494#issuecomment-521256360

In the past, we have seen some cases where the IDE can run MSBuild on
the same project in parallel. This causes inexplicable failures
throughout Xamarin.Android's MSBuild targets.

If we generate a build warning indicating this is happening, we can
have insight release over release if this happens in the wild. We will
also have a visible warning recorded in customer build logs.

Basically, we should:

* Write a zero-byte "lock file" at the beginning of a build
    * Warn if this file already exists
* Delete the file at the end of the current build
    * Even if the build failed, was cancelled, etc.

This is simple and should work cross-platform, where a named `Mutex`
would only work on Windows.

To achieve this we can perform some MSBuild wizardry:

* Register a new MSBuild target with `InitialTargets`. This target
  will run before *anything* no matter what the top-level target was.
* Pass an `IDisposable` to `RegisterTaskObject` with a lifetime of the
  current build. MSBuild will `Dispose` it automatically at the end of
  the build.

I added an assertion in one test to make sure the lock file does not
exist after a build. I did some manually testing also, and things seem
to be working as expected.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/WriteLockFile.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/WriteLockFile.cs
@@ -1,0 +1,63 @@
+using System;
+using System.IO;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Xamarin.Android.Tasks
+{
+	/// <summary>
+	/// Generates a warning if LockFile exists, registers the file to be deleted at end of build
+	/// </summary>
+	public class WriteLockFile : Task
+	{
+		[Required]
+		public string LockFile { get; set; }
+
+		public override bool Execute ()
+		{
+			try {
+				var path = Path.GetFullPath (LockFile);
+				var key = new Tuple<string, string> (nameof (WriteLockFile), path); // Use the full path as part of the key
+
+				// Check if already registered, for sanity
+				var existing = BuildEngine4.GetRegisteredTaskObject (key, RegisteredTaskObjectLifetime.Build) as DeleteFileAfterBuild;
+				if (existing == null) {
+					if (File.Exists (path)) {
+						Log.LogCodedWarning ("XA5302", "Two processes may be building this project at once. Lock file exists at path: {0}", path);
+					} else {
+						Directory.CreateDirectory (Path.GetDirectoryName (path));
+						File.WriteAllText (path, "");
+					}
+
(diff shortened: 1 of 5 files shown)


jonathanpeppers:
When the build finishes here, I will need to manually test the VSIX in the IDE.

src/Xamarin.Android.Build.Tasks/Tasks/WriteLockFile.cs
```diff
@@ -0,0 +1,51 @@
...
+		public override bool Execute ()
+		{
+			var path = Path.GetFullPath (LockFile);
+			var key = new Tuple<string, string> (nameof (WriteLockFile), path); // Use the full path as part of the key
+
+			// Check if already registered, for sanity
+			var existing = BuildEngine4.GetRegisteredTaskObject (key, RegisteredTaskObjectLifetime.Build) as DeleteFileAfterBuild;
+			if (existing == null) {
+				if (File.Exists (path)) {
+					Log.LogCodedWarning ("XA5302", "Two processes may be building this project at once. Lock file exists at path: {0}", path);
```
jonpryor: Suggestion?  Should the lock file contain the pid of the first process to create the file?  That way the message could print which process id started the build and what the current process id is.

However, as we believe that this is largely for intra-IDE scenarios, that might not actually be incredibly useful...

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -399,6 +400,10 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt2.targets" />
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.SkipCases.projitems" />
 
+<Target Name="_WriteLockFile">
+    <WriteLockFile LockFile="$(IntermediateOutputPath).__lock" />
```
dellis1972: Should we be using `CondinueOnError` here? Just in case something horrible goes wrong... If this task errors when the entire build will stop.

src/Xamarin.Android.Build.Tasks/Tasks/WriteLockFile.cs
jonathanpeppers: I don't think writing the pid helps us that much. If I was actively diagnosing this warning, I think I could figure out where the rogue MSBuild was coming from.

Issue #3477, opened 2019-08-10
error MSB4018: The "ConvertCustomView" task failed unexpectedly.


harshamg8:
### Steps to Reproduce

1. Build Xamarin.Android project
2. 
3. 

### Expected Behavior
Build should succeed

### Actual Behavior
Getting failed with the error "error MSB4018: The "ConvertCustomView" task failed unexpectedly."

### Version Information

```
=== Visual Studio Community 2019 for Mac ===

Version 8.2.2 (build 21)
Installation UUID: b8b23875-142b-439b-9f0c-174101613dcc
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 5.14.0.85 (d16-2-fix-mac-updates / 86973979)

	Package version: 600000319

=== Mono Framework MDK ===

Runtime:
	Mono 6.0.0.319 (2019-02/52203786470) (64-bit)
	Package version: 600000319

=== NuGet ===

Version: 5.1.0.6013

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.1.12
	2.1.11
	2.1.9
	2.1.8
	2.1.7
	2.1.2
	2.1.1
	2.0.5
	2.0.0
SDK: /usr/local/share/dotnet/sdk/2.1.701/Sdks
SDK Versions:
	2.1.701
	2.1.700
	2.1.505
	2.1.504
	2.1.503
	2.1.302
	2.1.301
	2.1.4
	2.0.0
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/6.0.0/lib/mono/msbuild/Current/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.11
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Updater ===

Version: 11

=== Xamarin.Android ===

Version: 9.4.1.0 (Visual Studio Community)
Commit: xamarin-android/d16-2/ebd65c9
Android SDK: /Users/harshamg/Library/Developer/Xamarin/android-sdk-macosx
	Supported Android versions:
		7.1 (API level 25)
		8.0 (API level 26)

SDK Tools Version: 26.1.1
SDK Platform Tools Version: 29.0.2
SDK Build Tools Version: 26.0.0

Build Information: 
Mono: mono/mono/2019-02@e6f5369c2d2
Java.Interop: xamarin/java.interop/d16-2@d64ada5
[...]

PR #3502, opened 2019-08-14
[Xamarin.Android.Build.Tasks] fix NRE in <ConvertCustomView/>

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/3477

An issue was reported of an error such as:

    /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(2315,3): error MSB4018: The "ConvertCustomView" task failed unexpectedly.
    System.NullReferenceException: Object reference not set to an instance of an object
        at Xamarin.Android.Tasks.ConvertCustomView+<>c__DisplayClass20_1.<Execute>b__0 (System.Diagnostics.TraceLevel level, System.String message) [0x0005b] in <35be9ecd0f44486bbcde5cb44c4d94c6>:0
        at Xamarin.Android.Tasks.ConvertCustomView.TryFixCustomView (System.Xml.Linq.XElement elem, System.Collections.Generic.Dictionary`2[TKey,TValue] acwMap, System.Action`2[T1,T2] logMessage) [0x0006f] in <35be9ecd0f44486bbcde5cb44c4d94c6>:0
        at Xamarin.Android.Tasks.ConvertCustomView.Execute () [0x00110] in <35be9ecd0f44486bbcde5cb44c4d94c6>:0
        at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute () [0x00029] in <9793e3774951417ca4f48a7c60193b35>:0
        at Microsoft.Build.BackEnd.TaskBuilder.ExecuteInstantiatedTask (Microsoft.Build.BackEnd.ITaskExecutionHost taskExecutionHost, Microsoft.Build.BackEnd.Logging.TaskLoggingContext taskLoggingContext, Microsoft.Build.BackEnd.TaskHost taskHost, Microsoft.Build.BackEnd.ItemBucket bucket, Microsoft.Build.BackEnd.TaskExecutionMode howToExecuteTask) [0x002a9] in <9793e3774951417ca4f48a7c60193b35>:0

In the `<ConvertCustomView/>` MSBuild task, there is a code path in
which `resdir` can be `null` and we directly call `resdir.ItemSpec`.
It looks like we can just use `resdir?.ItemSpec`, and
`FixupResourceFilenameAndLogCodedError` safely handles `null` passed
in.

I was only able to reproduce the issue by uppercasing the directory
name -- I am not sure how this would happen in a real project.

In either case, I added the null check and used
`StringComparison.OrdinalIgnoreCase` where appropriate.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ConvertCustomView.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ConvertCustomView.cs
@@ -49,13 +49,13 @@ namespace Xamarin.Android.Tasks {
 						bool update = false;
 						foreach (var elem in AndroidResource.GetElements (e).Prepend (e)) {
 							update |= TryFixCustomView (elem, acw_map, (level, message) => {
-								ITaskItem resdir = ResourceDirectories?.FirstOrDefault (x => file.StartsWith (x.ItemSpec)) ?? null;
+								ITaskItem resdir = ResourceDirectories?.FirstOrDefault (x => file.StartsWith (x.ItemSpec, StringComparison.OrdinalIgnoreCase)) ?? null;
 								switch (level) {
 								case TraceLevel.Error:
-									Log.FixupResourceFilenameAndLogCodedError ("XA1002", message, file, resdir.ItemSpec, resource_name_case_map);
+									Log.FixupResourceFilenameAndLogCodedError ("XA1002", message, file, resdir?.ItemSpec, resource_name_case_map);
 									break;
 								case TraceLevel.Warning:
-									Log.FixupResourceFilenameAndLogCodedError ("XA1001", message, file, resdir.ItemSpec, resource_name_case_map);
+									Log.FixupResourceFilenameAndLogCodedError ("XA1001", message, file, resdir?.ItemSpec, resource_name_case_map);
 									break;
 								default:
 									Log.LogDebugMessage (message);
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ConvertResourcesCasesTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ConvertResourcesCasesTests.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using NUnit.Framework;
 using Xamarin.ProjectTools;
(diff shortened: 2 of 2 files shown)


jonathanpeppers:
@dellis1972 any ideas on how to make a better test here? How do you think the developer hit this issue?

src/Xamarin.Android.Build.Tasks/Tasks/ConvertCustomView.cs
```diff
@@ -52,10 +52,10 @@ public override bool Execute ()
 								ITaskItem resdir = ResourceDirectories?.FirstOrDefault (x => file.StartsWith (x.ItemSpec)) ?? null;
```
jonpryor: This is why we need nullable reference types. ;-)

dellis1972:
@jonathanpeppers the only way I can see this getting a null `resdir` is if the file `ItemSpec` does not exist in ANY of the provided `ResourceDirectories`. I'm wondering if this is related to `LogicalName` metadata. It might be the file is coming from some other place but is having its `LogicalName` fixed up. that might result in the actual source file being outside one of the provided `ResourceDirectories` values. I've seen a few third party products (Uno being one) which is making use of the `LogicalName` stuff for `AndroidResources` and having the actual files in a folder called `Assets`, which in this case won't appear. 

So it might be a combination of using `LogicalName` AND having a custom view in the file.

jonathanpeppers:
This change did not reproduce the problem either:
```diff
@@ -152,7 +152,8 @@ namespace Xamarin.Android.Build.Tests {
                        proj.Sources.Add (new BuildItem.Source ("CustomView.cs") {
                                TextContent = () => "namespace ClassLibrary1 { public class CustomView : Android.Views.View { public CustomView (Android.Content.Context context) : base (context) { } } }"
                        });
-                       proj.AndroidResources.Add (new AndroidItem.AndroidResource ("Resources\\layout\\CustomView.xml") {
+                       proj.AndroidResources.Add (new AndroidItem.AndroidResource ("CustomView.xml") {
+                               MetadataValues = "LogicalName=Resources\\layout\\CustomView2.xml",
```

I'm leaning toward just putting the fix in--we just won't have a test 😞 Unless there are other ideas?

jonathanpeppers:
@dellis1972 we also have a binlog, but I don't see anything out of the ordinary: https://github.com/xamarin/xamarin-android/issues/3477#issuecomment-521331864

It doesn't look like `%(LogicalName)` was used.

PR #3518, opened 2019-08-18
Fix mingw dependency version constraint

SamuelDebruyn:
When running `make prepare` with `AutoProvision` set to true, the script installs mingw 6.0.0_2. When it detects that mingw is not installed, it will install version 6.0.0_2 and continue. The next time, it will detect 6.0.0_2 and complain about the version, unable to proceed.


grendello:
build

PR #3519, opened 2019-08-19
[tests/ci] Fix Azure Pipeline hosted detection

pjcollins:
The hosted pools we're using on Azure Pipelines no longer appear to have
'hosted' in the name. Updates the running on hosted check to look for
the new hosted pool agent naming schema, and adds a new variable check
to help make this more robust. Predefined Azure Pipeline variables can be
found here: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/TestEnvironment.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/TestEnvironment.cs
@@ -56,9 +56,9 @@ namespace Xamarin.ProjectTools
 		public static bool IsRunningOnHostedAzureAgent {
 			get {
 				string agentNameValue = Environment.GetEnvironmentVariable ("AGENT_NAME");
-				bool hasHostedAgentName = !string.IsNullOrEmpty (agentNameValue) && agentNameValue.ToUpper ().Contains ("AZURE PIPELINES");
+				bool hasHostedAgentName = !string.IsNullOrEmpty (agentNameValue) && agentNameValue.ToUpperInvariant ().Contains ("AZURE PIPELINES");
 				string serverTypeValue = Environment.GetEnvironmentVariable ("SYSTEM_SERVERTYPE");
-				bool isHostedServerType = !string.IsNullOrEmpty (serverTypeValue) && serverTypeValue.ToUpper ().Contains ("HOSTED");
+				bool isHostedServerType = !string.IsNullOrEmpty (serverTypeValue) && serverTypeValue.ToUpperInvariant ().Contains ("HOSTED");
 				return hasHostedAgentName || isHostedServerType;
 			}
 		}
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/TestEnvironment.cs
```diff
@@ -56,7 +56,10 @@ static bool IsDarwin ()
 		public static bool IsRunningOnHostedAzureAgent {
 			get {
 				string agentNameValue = Environment.GetEnvironmentVariable ("AGENT_NAME");
-				return !string.IsNullOrEmpty(agentNameValue) && agentNameValue.ToUpper ().Contains ("HOSTED");
+				bool hasHostedAgentName = !string.IsNullOrEmpty (agentNameValue) && agentNameValue.ToUpper ().Contains ("AZURE PIPELINES");
+				string serverTypeValue = Environment.GetEnvironmentVariable ("SYSTEM_SERVERTYPE");
+				bool isHostedServerType = !string.IsNullOrEmpty (serverTypeValue) && serverTypeValue.ToUpper ().Contains ("HOSTED");
```
jonathanpeppers: Do they have these env vars documented? We could link to?

pjcollins: There is _some_ predefined variable documentation here - https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml, but unfortunately it doesn't list the server type variable and it doesn't specify any hosted naming conventions.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/TestEnvironment.cs
```diff
@@ -56,7 +56,10 @@ static bool IsDarwin ()
 		public static bool IsRunningOnHostedAzureAgent {
 			get {
 				string agentNameValue = Environment.GetEnvironmentVariable ("AGENT_NAME");
-				return !string.IsNullOrEmpty(agentNameValue) && agentNameValue.ToUpper ().Contains ("HOSTED");
+				bool hasHostedAgentName = !string.IsNullOrEmpty (agentNameValue) && agentNameValue.ToUpper ().Contains ("AZURE PIPELINES");
```
jonpryor: This should be `.ToUpperInvariant()`, to avoid the [Turkish 'I' problem](https://docs.microsoft.com/en-us/previous-versions/dotnet/articles/ms994325(v=msdn.10)#the-turkish-example) (and I suspect `.ToUpperInvariant()` was added *after* that article was written...).

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/TestEnvironment.cs
jonpryor: Ditto use of [`.ToUpperInvariant()`](https://docs.microsoft.com/en-us/dotnet/api/system.string.toupperinvariant?view=netframework-4.8) here.

PR #3522, opened 2019-08-20
[Xamarin.Android.Build.Tasks] Add EnvVar and File support for Signing APKs

### LINKED ISSUE #3513: Provide way to sign package/bundle with custom keyfile without writing passphrase to log
When I run `msbuild` with `/t:SignAndroidPackage` and provide a custom keystore, the keystore and key passwords end up being written out to the build log because it logs the call made to `jarsigner` / `apksigner`.

I wanted to prevent the passwords being written to the log, so I have ended up doing the signing as a separate step. In this step I make use of the `:env` the suffix in `jarsigner` or the `env:` prefix in `apksigner` before I switched to bundles.

Below is what my build looks like now. `AndroidSigning_StorePass` is an environment variable that I populate in a pre-build step with the password:

```
    commands:
       - msbuild MyProj/MyProj.csproj /restore
         /p:Configuration=Release
         /p:AndroidSdkDirectory=/android/sdk
         /t:Package
       - jarsigner -sigalg SHA256withRSA -digestalg SHA-256
         -keystore ../my.keystore
         -keypass:env AndroidSigning_StorePass
         -storepass:env AndroidSigning_StorePass
         -signedjar ./MyProj/bin/Release/com.me.myapp-Signed.aab
         ./MyProj/obj/Release/android/bin/com.me.myapp.aab $AndroidSigning_KeyAlias
```

The above seems to work fine, but it took a lot of effort to get there. It would be better if I could use the standard build target but still keep the sensitive data out of the log. Then I wouldn't need to do the custom signing and could benefit from any future tweaks that you put in the standard build target.

dellis1972:
Fixes #3513

Both `jarsigner` and `apksigner` provide a way to use both
files and environment variables for the store and key passwords.

For `jarsigner` you have to suffix the parameter switch with either
`:env` or `:file` to use those options. For `apksigner` you have
to prefix the value with either `:env`, `:file` or `:pass`.

We currently only support raw passwords.

This commit adds support for using both `env:` and `file:` for
signing. When providing values for the MSBuild properties
such as `AndroidSigningStorePass` and `AndroidSigningKeyPass` all
they need to do is prefix the value with `env:` or `file:` to
use the alternative parameters.

	/p:AndroidSigningKeyPass=env:MyPasswordEnvVar
	/p:AndroidSigningKeyPass=file:PathToPasswordFile

This will stop passwords appearing in build logs etc.

### DIFF
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -1073,6 +1073,15 @@ server, the following MSBuild properties can be used:
     the value entered when `keytool` asks **Enter key password for
     $(AndroidSigningKeyAlias)**.
 
+    You can use the raw password here, however if you want to hide your password in logs
+    you can use a prefix of env: or file: to point it to an Environment variable or 
+    a file. For example
+
+    ```
+        env:<PasswordEnvironentVariable>
+        file:<PasswordFile> 
+    ```
+
 -   **AndroidSigningKeyStore** &ndash; Specifies the filename of the
     keystore file created by `keytool`. This corresponds to the value
     provided to the **keytool -keystore** option.
@@ -1081,6 +1090,15 @@ server, the following MSBuild properties can be used:
     `$(AndroidSigningKeyStore)`. This is the value provided to
     `keytool` when creating the keystore file and asked **Enter
     keystore password:**.
+
+    You can use the raw password here, however if you want to hide your password in logs
+    you can use a prefix of env: or file: to point it to an Environment variable or 
+    a file. For example
+
+    ```
+        env:<PasswordEnvironentVariable>
+        file:<PasswordFile> 
+    ```
 
 For example, consider the following `keytool` invocation:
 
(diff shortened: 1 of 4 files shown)


dellis1972:
@jonathanpeppers was just doing that exact thing :D

Documentation/guides/BuildProcess.md
```diff
@@ -1073,6 +1073,15 @@ server, the following MSBuild properties can be used:
     the value entered when `keytool` asks **Enter key password for
     $(AndroidSigningKeyAlias)**.
 
+    You can use the raw password here, however if you want to hide your password in logs
+    you can use a preview of env: or file: to point it to an Environment variable or
```
mungojam: I think this is meant to say `prefix` rather than `preview`

dellis1972: Fixed, thanks for the review :)

PR #3523, opened 2019-08-20
[Xamarin.Android.Build.Tasks] tests for MSBuild.Sdk.Extras

jonathanpeppers:
Context: https://github.com/onovotny/MSBuildSdkExtras
Context: https://www.nuget.org/packages/MSBuild.Sdk.Extras/

Today, you can create an SDK-style project such as:

      <PropertyGroup>
        <TargetFramework>netstandard2.0</TargetFramework>
      </PropertyGroup>
    </Project>

If you want to "multi-target", you can take advantage of
MSBuild.Sdk.Extras, such as:

    <Project Sdk="MSBuild.Sdk.Extras/2.0.41">
      <PropertyGroup>
        <TargetFrameworks>netstandard2.0;MonoAndroid90</TargetFrameworks>
      </PropertyGroup>
    </Project>

This imports `Xamarin.Android.CSharp.targets` on top of the SDK-style
MSBuild targets. `MSBuild.Sdk.Extras` is a community-led project.

During a build both will be produced:

* `bin\Debug\netstandard2.0\foo.dll`
* `bin\Debug\monoandroid90\90\foo.dll` - *NOTE* this seems wrong

Unfortunately, our `$(OutputPath)` is not correct, this is problem
number one. We also have no tests verifying that MSBuild.Sdk.Extras
continues to work! Lastly, we are missing some support for the NuGet
"SDK Resolver" in `xabuild.exe`.

## xabuild ##

SDK-style projects have a concept of a "SDK resolver", which locates
the appropriate assemblies, MSBuild targets, etc.
`Microsoft.Build.NuGetSdkResolver.dll` enables SDKs to come from
NuGet, such as `MSBuild.Sdk.Extras`.

Using `xabuild.exe` as-is was producing errors such as:

    Project Name=UnnamedProject.csproj id:4 File=UnnamedProject.csproj
        Resolving SDK 'MSBuild.Sdk.Extras/2.0.41'...
        MSB4236: The SDK 'MSBuild.Sdk.Extras/2.0.41' specified could not be found.

For `Microsoft.Build.NuGetSdkResolver.dll` to be *found*:

* Relative to `%MSBUILD_EXE_PATH%`, a
  `SdkResolvers\Microsoft.Build.SdkResolver\Microsoft.Build.SdkResolver.xml`
  file must exist.
* `Microsoft.Build.SdkResolver.xml` contains a path to
  `Microsoft.Build.NuGetSdkResolver.dll`

An example of this XML file would be:

    <SdkResolver>
      <Path>C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\CommonExtensions\Microsoft\NuGet\Microsoft.Build.NuGetSdkResolver.dll</Path>
    </SdkResolver>

To get `xabuild.exe` working as needed, I had to:

* `%MSBUILD_EXE_PATH%` is created in a subdirectory of `%TEMP%`
* This subdirectory also contains
  `SdkResolvers\Microsoft.Build.SdkResolver\Microsoft.Build.SdkResolver.xml`
* `xabuild.exe` now generates the extra config file
* On exit, `xabuild.exe` deletes the entire temp directory

## Xamarin.Android.DefaultOutputPaths.targets ##

I fixed up the weirdness when MSBuild.Sdk.Extras is used:

* If `$(UsingMicrosoftNETSdk)` is `True`, we should not mess with
  `$(OutputPath)` or `$(IntermediateOutputPath)`.
* Proper paths are used already by default, such as:
  `obj\Debug\monoandroid90` and `bin\Debug\monoandroid90`

Other refactoring:

* `$(BaseIntermediateOutputPath)` should be `obj\` by default`
* I made use of `[MSBuild]::EnsureTrailingSlash`
* Removed a duplicate line setting
  `$(AppendTargetFrameworkToIntermediateOutputPath)`

## tests ##

In `Xamarin.ProjectTools`, I created a `MSBuildSdkExtrasProject` class
and added tests for a couple scenarios:

* A regular class library
* A class library with no `@(AndroidResource)`
* A multi-targeted library using `#if __ANDROID__`

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/MSBuildSdkExtrasTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/MSBuildSdkExtrasTests.cs
@@ -1,0 +1,92 @@
+using System.IO;
+using System.Linq;
+using Mono.Cecil;
+using NUnit.Framework;
+using Xamarin.ProjectTools;
+
+namespace Xamarin.Android.Build.Tests
+{
+	[TestFixture]
+	[NonParallelizable] // On MacOS, parallel /restore causes issues
+	public class MSBuildSdkExtrasTests : BaseTest
+	{
+		[Test]
+		public void ClassLibrary ()
+		{
+			var proj = new MSBuildSdkExtrasProject ();
+			using (var b = CreateDllBuilder (Path.Combine ("temp", TestName))) {
+				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
+				var output = Path.Combine (Root, b.ProjectDirectory, proj.OutputPath, proj.ProjectName + ".dll");
+				FileAssert.Exists (output);
+				AssertContainsClass (output, "Class1", contains: true);
+			}
+		}
+
+		[Test]
+		public void ClassLibraryNoResources ()
+		{
+			var proj = new MSBuildSdkExtrasProject ();
+			proj.Sources.Remove (proj.Sources.First (s => s.BuildAction == "AndroidResource"));
+			using (var b = CreateDllBuilder (Path.Combine ("temp", TestName))) {
+				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
+				var output = Path.Combine (Root, b.ProjectDirectory, proj.OutputPath, proj.ProjectName + ".dll");
+				FileAssert.Exists (output);
+				AssertContainsClass (output, "Class1", contains: true);
+			}
+		}
+
+		[Test]
+		public void BindingProject ()
+		{
+			var proj = new MSBuildSdkExtrasProject {
+				IsBindingProject = true,
+			};
+			proj.OtherBuildItems.Add (new AndroidItem.EmbeddedJar ("Jars\\svg-android.jar") {
+				WebContent = "https://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/svg-android/svg-android.jar"
+			});
+			using (var b = CreateDllBuilder (Path.Combine ("temp", TestName))) {
+				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
+				var output = Path.Combine (Root, b.ProjectDirectory, proj.OutputPath, proj.ProjectName + ".dll");
+				FileAssert.Exists (output);
+				AssertContainsClass (output, "Com.Larvalabs.Svgandroid.SVG", contains: true);
+			}
+		}
+
+		[Test]
+		public void MultiTargeting ()
+		{
+			var proj = new MSBuildSdkExtrasProject ();
+			proj.TargetFrameworks += ";netstandard2.0";
+			proj.Sources.Add (new BuildItem.Source ("MyView.cs") {
+				TextContent = () =>
+@"#if __ANDROID__
+class MyView : Android.Views.View
+{
+	public MyView (Android.Content.Context c) : base (c) { }
+}
+#endif",
+			});
+			using (var b = CreateDllBuilder (Path.Combine ("temp", TestName))) {
+				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
+				var output = Path.Combine (Root, b.ProjectDirectory, proj.OutputPath, proj.ProjectName + ".dll");
+				FileAssert.Exists (output);
+				AssertContainsClass (output, "MyView", contains: true);
+				output = Path.Combine (Root, b.ProjectDirectory, proj.OutputPath, "..", "netstandard2.0", proj.ProjectName + ".dll");
+				FileAssert.Exists (output);
+				AssertContainsClass (output, "MyView", contains: false);
+			}
+		}
+
+		void AssertContainsClass (string assemblyFile, string className, bool contains)
+		{
+			using (var assembly = AssemblyDefinition.ReadAssembly(assemblyFile)) {
+				bool result = assembly.MainModule.Types.Select (t => t.FullName).Contains (className);
+				if (contains) {
+					Assert.IsTrue (result, $"{assemblyFile} should contain {className}!");
+				} else {
+					Assert.IsFalse (result, $"{assemblyFile} should *not* contain {className}!");
+				}
+			}
+		}
+	}
+}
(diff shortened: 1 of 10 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/MSBuildSdkExtrasProject.cs
```diff
@@ -0,0 +1,49 @@
...
+		{
+			Sdk = "MSBuild.Sdk.Extras/2.0.41";
+			Sources.Add (new BuildItem.Source ("Class1.cs") {
+				TextContent = () => "public class Class1 { }"
+			});
+			Sources.Add (new AndroidItem.AndroidResource ("Resources\\values\\Strings.xml") {
+				TextContent = () => StringsXml.Replace ("${PROJECT_NAME}", ProjectName)
+			});
+			StringsXml = XamarinAndroidLibraryProject.default_strings_xml;
+			TargetFrameworks = "MonoAndroid10.0";
```
jonathanpeppers: Note that I had to use `MonoAndroid10.0`, since `MonoAndroid100` set the TFV to `v1.0.0`...

However, `MonoAndroid90` works property: TFV is `v9.0`.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/MSBuildSdkExtrasTests.cs
```diff
@@ -0,0 +1,57 @@
...
+class MyView : Android.Views.View
+{
+	public MyView (Android.Content.Context c) : base (c) { }
+}
+#endif",
+			});
+			using (var b = CreateDllBuilder (Path.Combine ("temp", TestName))) {
+				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
+				var output = Path.Combine (Root, b.ProjectDirectory, proj.OutputPath, proj.ProjectName + ".dll");
+				FileAssert.Exists (output);
```
jonpryor: Should we (somehow?) assert that the generated file contains `MyView`, ensuring that `__ANDROID__` is defined as part of the MSBuild.Sdk.Extras build?

jonpryor: Under mono, `monodis --typedef ASSEMBLY | grep MyView` would work...

jonathanpeppers: I think there is a test in here that uses Mono.Cecil to do something like this, I’ll check.

tools/xabuild/XABuildPaths.cs
```diff
@@ -180,15 +197,19 @@ public XABuildPaths ()
...
 				}
 			}
 
 			FrameworksDirectory       = Path.Combine (prefix, "xbuild-frameworks");
 			MSBuildExtensionsPath     = Path.Combine (prefix, "xbuild");
 			MonoAndroidToolsDirectory = Path.Combine (prefix, "xbuild", "Xamarin", "Android");
-			MSBuildExeTempPath        = Path.GetTempFileName ();
+			MSBuildTempPath           = Path.Combine (Path.GetTempPath (), Path.GetRandomFileName ());
+			MSBuildExeTempPath        = Path.Combine (MSBuildTempPath, Path.GetRandomFileName ());
+			SdkResolverConfigPath     = Path.Combine (MSBuildTempPath, "SdkResolvers", "Microsoft.Build.NuGetSdkResolver", "Microsoft.Build.NuGetSdkResolver.xml");
```
jonpryor: I would suggest two things:

 1. If not already done -- I didn't check -- we should ensure that the files we create in `MSBuildTempPath` are deleted on process exit, if possible, and
 2. For the "multiple `xabuild.exe` invocations in parallel" case -- and for general sanity -- we probably don't want to create `%TEMP%\SdkResolvers\...`, but instead `%TEMP%\xabuild.(SomeRandomStuff)\SdkResolvers\...`.

jonathanpeppers: This PR should be doing that already, I think.

Before:
* `%TEMP%\GetRandomFileName`
* `%TEMP%\GetRandomFileName.config`

Now:
* `%TEMP%\GetRandomFileName\GetRandomFileName`
* `%TEMP%\GetRandomFileName\GetRandomFileName.config`
* `%TEMP%\GetRandomFileName\SdkResolvers\Microsoft.Build.NuGetSdkResolvers\Microsoft.Build.NuGetSdkResolvers.xml`

Now we can just delete the entire `%TEMP%\GetRandomFileName` directory when `xabuild.exe` exits.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/MSBuildSdkExtrasProject.cs
dellis1972: Is that going to cause us problems? Or will this just affect those using MSBuild.Sdk.Extras?

jonathanpeppers: I have a feeling that NuGet authors will have to use `monoandroid10.0` for their directory names... where they have been used to using `monoandroid81`, `monoandroid90`, etc.

jonathanpeppers:
The only test failure I see here is:
```
(_ResolveXamarinAndroidTools target) -> 
    error XA5300:  Error finding SDK version
    error XA5300: System.ArgumentNullException: Value cannot be null.
    error XA5300: Parameter name: path1
    error XA5300:   at System.IO.Path.Combine (System.String path1, System.String path2) [0x00003] in <8446a011a55940e69e19c2e3544f66ff>:0
    error XA5300:   at Xamarin.AndroidTools.MonoDroidSdkBase+<GetVersionFileLocations>d__43.MoveNext () [0x00056] in <cdec3287ec264f8c9f40d2bb445cc0a0>:0
    error XA5300:   at Xamarin.AndroidTools.MonoDroidSdkBase.FindVersionFile () [0x0004e] in <cdec3287ec264f8c9f40d2bb445cc0a0>:0
    error XA5300:   at Xamarin.AndroidTools.MonoDroidSdk.LoadVersionInfo () [0x00006] in <cdec3287ec264f8c9f40d2bb445cc0a0>:0
```
Which we've seen randomly... 

https://github.com/xamarin/monodroid/pull/1020 might help.

Issue #3498, opened 2019-08-13
Writing to System.IO.Stream doesn't work for SAF Google Drive URI


mrjfalk:
### Steps to Reproduce

1. Use Storage Access Framework API to create a new file in Google Drive. Intent I use:
```csharp
Intent intent = new Intent(Intent.ActionCreateDocument);
intent.AddCategory(Intent.CategoryOpenable);
intent.SetType("*/*");
intent.PutExtra(Intent.ExtraTitle, "MyFile.ext");
etc...
```

2. Get the URI for the created file, e.g. content://com.google.android.apps.docs.storage/document/&lt;documentIdentifier&gt;

3. Get an output stream to this file:
```csharp
using(Stream stream = Application.Context.ContentResolver.OpenOutputStream(Android.Net.Uri.Parse(Path)))
{
…
}
```

4. Write to the stream:
```csharp
byte[] data = Encoding.UTF8.GetBytes("Test string...");
stream.Write(data, 0, data.Length);
```

### Expected Behavior
The file should contain the string "Test string...".

### Actual Behavior
The file in Google Drive is empty (0 KB). No exception is thrown either.

### Workaround
As mentioned in the answer here: https://stackoverflow.com/questions/48876846/store-file-with-the-storage-access-framework-on-google-drive writing to the stream as a Java `BufferedOutputStream` actually works:
```csharp
using(Stream stream = Application.Context.ContentResolver.OpenOutputStream(Android.Net.Uri.Parse(Path)))
{
    using (var javaStream = new global::Java.IO.BufferedOutputStream(stream))
    {
        byte[] data = Encoding.UTF8.GetBytes("Test string...");
        javaStream.Write(data, 0, data.Length);
        javaStream.Close();
     }
}
```

The `javaStream.Close()` is required, otherwise nothing is written here either.
                    

### Version Information
Xamarin.Android SDK 9.4.1.0

### Log File
Could not find anything interesting at all in Logcat when executing the writing using System.IO.Stream.

mrjfalk:
If anyone else has the same issue and wants to still use System.IO.Stream (e.g. in shared code), a class such as this seems to work fine:

```csharp
internal class AndroidOutputStream : Stream
    {
        public override bool CanRead => false;

        public override bool CanSeek => false;

        public override bool CanWrite => true;

        public override long Length => throw new NotImplementedException();

        public override long Position { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }

[...]

PR #3524, opened 2019-08-20
[Mono.Android] Close BaseOutputStream when OutputStreamInvoker.Close is called

jpobst:
From #3498:

When writing to Google Drive using `Application.Context.ContentResolver.OpenOutputStream` the content is not actually written.  This method returns an `Android.Runtime.OutputStreamInvoker` that wraps the `Java.IO.OutputStream`.  However the Invoker does not override `Close ()` so the base stream does not get closed.

This does not seem to be an issue when writing a file locally but causes 0 bytes to get written to Google Drive.  (Likely due to buffering.)

The `KitKat` project in `monodroid-samples` can be used as a repro:
https://github.com/xamarin/monodroid-samples/tree/master/KitKat

* Run project
* Click Storage Access Framework
* Click Create new document
* Choose Google Drive as the target
* Hit Save

Without this change the file contains 0 bytes.  With this change the file contains 13 bytes.

### DIFF
--- a/src/Mono.Android/Android.Runtime/OutputStreamInvoker.cs
+++ b/src/Mono.Android/Android.Runtime/OutputStreamInvoker.cs
@@ -12,6 +12,11 @@ namespace Android.Runtime
 			this.BaseOutputStream = stream;
 		}
 
+		public override void Close ()
+		{
+			BaseOutputStream.Close ();
+		}
+
 		protected override void Dispose (bool disposing)
 		{
 			if (disposing && BaseOutputStream != null) {


src/Mono.Android/Android.Runtime/OutputStreamInvoker.cs
```diff
@@ -12,6 +12,11 @@ public OutputStreamInvoker (Java.IO.OutputStream stream)
...
 		}
 
+		public override void Close ()
+		{
+			BaseOutputStream.Close ();
+		}
+
 		protected override void Dispose (bool disposing)
 		{
 			if (disposing && BaseOutputStream != null) {
```
jonpryor: Should we *also* call `.Close()` as part of `Dispose()` on the following line?

jpobst: Good question, I just tried this and verified that `Stream.Dispose()` calls `.Close()` for us:
```
  at Android.Runtime.OutputStreamInvoker.Close () [0x00000] in <b56c0a2d3d9045ab919839b6b74ef47d>:0 
  at System.IO.Stream.Dispose () [0x00000] in <590c33eecca549339b1b4b3b25a9dba2>:0 
  at KitKat.StorageActivity.WriteToSaf (Android.Net.Uri uri) [0x00000] in <99537faebd544453953fb4c279932f04>:0 
  at KitKat.StorageActivity.OnActivityResult (System.Int32 requestCode, Android.App.Result resultCode, Android.Content.Intent data) [0x00000] in <99537faebd544453953fb4c279932f04>:0 
  at Android.App.Activity.n_OnActivityResult_IILandroid_content_Intent_ (System.IntPtr jnienv, System.IntPtr native__this, System.Int32 requestCode, System.Int32 native_resultCode, System.IntPtr native_data) [0x00000] in <b56c0a2d3d9045ab919839b6b74ef47d>:0 
  at Android.Runtime.DynamicMethodNameCounter.7 (System.IntPtr , System.IntPtr , System.Int32 , System.Int32 , System.IntPtr ) [0x00000] in <b56c0a2d3d9045ab919839b6b74ef47d>:0 
```

jonpryor:
```
Fixes: https://github.com/xamarin/xamarin-android/issues/3498

When writing to a file using
`Application.Context.ContentResolver.OpenOutputStream()`:

	using(Stream stream = Application.Context.ContentResolver.OpenOutputStream(Android.Net.Uri.Parse(Path))) {
	    byte[] data = Encoding.UTF8.GetBytes("Test string...");
	    stream.Write(data, 0, data.Length);
	}

the content is not actually written!

What happens is as follows:

 1. App calls `ContentResolver.OpenOutputStream()`.
 2. Android creates and returns an `java.io.OutputStream` instance.
 3. The `OutputStream` is wrapped by an
    `Android.Runtime.OutputStreamInvoker`
 4. `stream.Write()` calls `OutputStreamInvoker.Write()` which calls
    `OutputStream.write()`.
 5. Implicit `stream.Dispose()` call calls `Java.Lang.Object.Dispose()`
    which calls `OutputStreamInvoker.Dispose()`, which calls
    `OutputStream.Dispose()`, which frees the JNI Global Reference.

    `OutputStream.Dispose()` does *not* cause an `OutputStream.flush()`
    or `OutputStream.close()` invocation to be performed.

The result is that the Java GC heap has an `OutputStream` instance
which is unreferenced and eventually finalized, with nothing actually
writing the contents to disk, in part because `OutputStream.finalize()`
either isn't overridden or doesn't write contents to disk *either*.

As it happens, `Stream.Dispose()` will call `Stream.Close()`, so if
`OutputStreamInvoker` overrides `Stream.Close()`, then we can "insert"
a `BaseOutputStream.Close()` invocation as part of `Dispose()`:

	at Android.Runtime.OutputStreamInvoker.Close ()
	at System.IO.Stream.Dispose ()
	...	

This ensures that `using(stream)` will write the contents to disk.

The [`KitKat` project in `monodroid-samples`][0] can be used as a repro:

 1. Run project
 2. Click Storage Access Framework
 3. Click Create new document
 4. Choose Google Drive as the target
 5. Hit Save

Without this change the file contains 0 bytes.
With this change the file contains 13 bytes.

[0]: https://github.com/xamarin/monodroid-samples/tree/master/KitKat
```

Issue #3263, opened 2019-06-19
Link assemblies causes app crashes if you have an EditText in VS2019 Preview 2


FANMixco:
### Steps to Reproduce

1. Add an EditText
2. Choose TextChanged. (Example:  `txtSearch.TextChanged += TxtSearch_TextChanged;`)
3. Choose Linking with either "Sdk Assemblies Only" or "Sdk and User Assemblies".
4. Execute the program.

### Expected Behavior

Be able to execute the TextChanged on Linking with either "Sdk Assemblies Only" or "Sdk and User Assemblies".

### Actual Behavior

App crashes because it cannot load ITextWatcherInvoker.

### Version Information

Windows 10, May Update.
Visual Studio 2019 Preview 2.

### Log File

> System.TypeLoadException: 'Could not load type 'Android.Text.ITextWatcherInvoker' from assembly 'Mono.Android, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.'

![Annotation 2019-06-20 003519](https://user-images.githubusercontent.com/1666156/59806029-54449600-92f3-11e9-9df5-b3c453facd90.png)

**P.S.**

It's working in Visual Studio for Mac.

> VS bug [#947890](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/947890)

pjcollins:
Thanks for the report! I assume you're referring to VS 2019 16.2 Preview 2? Could you paste the information displayed in `Help`  -> `About Microsoft Visual Studio`? I was unable to reproduce this with the upcoming VS 2019 16.2 Preview 3 release, though my repro may not have been sufficient. Could you also attach a solution file that demonstrates this crash?

FANMixco:
> Thanks for the report! I assume you're referring to VS 2019 16.2 Preview 2? Could you paste the information displayed in `Help` -> `About Microsoft Visual Studio`? I was unable to reproduce this with the upcoming VS 2019 16.2 Preview 3 release, though my repro may not have been sufficient. Could you also attach a solution file that demonstrates this crash?

Yes, it's in the Preview 2 of VS 2019 for Windows only, I could try maybe sharing a OneDrive link because the solution is somehow huge more than a 100Mb.

gmck:
I have exactly the same error however I'm using 16.2.0 Preview 3. I'm going to try an reproduce in a test project

pjcollins:
[...]

PR #3525, opened 2019-08-20
[linker] Improve *Invoker types linking

radekdoulik:
Improves fix of https://github.com/xamarin/xamarin-android/issues/3263

The
https://github.com/xamarin/xamarin-android/commit/4d8c28f624b50a86b564b0bc6bdefb85ea822b69
fix increased the apk size more than needed. It marked all the
*Invoker types, beause MarkJavaObjects substep iterates over all
Java.Object derived types, before all the linking.

Turned out that the original regression came from the linker
itself. The new linker optimization was introduced in
https://github.com/mono/linker/commit/13ea1582e518f5f64cb93057aca962983e58e337
and so the ITextWatcher interface was not marked anymore.

Before the new optimization, the linker preserved these
ifaces (like `ITextWatcher`) when marking the types based on them
(like `AbsListView` or `ITextWatcherImplementor` in this case). That also
resulted in keeping the invoker interfaces in the assembly. The above
mentioned optimization changed that.

Now they were linked out, because the interfaces implementation
are only accessed (during app runtime) after the invoker type is loaded by
`Type.GetType(string,bool)` in the native members registration
process. The linker cannot know about that without XA help during the
linking. The result of `ITextWatcher` being linked away is that the
`ITeextWatcherInvoker` type was linked away too.

It is now fixed by overriding the `ShouldMarkInterfaceImplementation`
method in `MarkStep`, which returns true for registered interfaces.
These interfaces are special for XA in a way, that it can be,
 in some cases, reached only from java side and thus linker is unable
to track the interface usage. It is similar to how linker handles the COM
interfaces.

The crash in Topeka sample:

    I MonoDroid: System.TypeLoadException: Could not load type 'Android.Text.ITextWatcherInvoker' from assembly 'Mono.Android, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
    I MonoDroid:   at (wrapper managed-to-native) System.RuntimeTypeHandle.internal_from_name(string,System.Threading.StackCrawlMark&,System.Reflection.Assembly,bool,bool,bool)
    I MonoDroid:   at System.RuntimeTypeHandle.GetTypeByName (System.String typeName, System.Boolean throwOnError, System.Boolean ignoreCase, System.Boolean reflectionOnly, System.Threading.StackCrawlMark& stackMark, System.Boolean loadTypeFromPartialName) [0x0008d] in <ab232a84cff64754aaec9d455f50a794>:0
    I MonoDroid:   at System.RuntimeType.GetType (System.String typeName, System.Boolean throwOnError, System.Boolean ignoreCase, System.Boolean reflectionOnly, System.Threading.StackCrawlMark& stackMark) [0x0000e] in <ab232a84cff64754aaec9d455f50a794>:0
    I MonoDroid:   at System.Type.GetType (System.String typeName, System.Boolean throwOnError) [0x00002] in <ab232a84cff64754aaec9d455f50a794>:0
    I MonoDroid:   at Android.Runtime.AndroidTypeManager.RegisterNativeMembers (Java.Interop.JniType jniType, System.Type type, System.String methods) [0x00131] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at Android.Runtime.JNIEnv.RegisterJniNatives (System.IntPtr typeName_ptr, System.Int32 typeName_len, System.IntPtr jniClass, System.IntPtr methods_ptr, System.Int32 methods_len) [0x00128] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at (wrapper managed-to-native) Java.Interop.NativeMethods.java_interop_jnienv_call_static_object_method_a(intptr,intptr&,intptr,intptr,Java.Interop.JniArgumentValue*)
    I MonoDroid:   at Java.Interop.JniEnvironment+StaticMethods.CallStaticObjectMethod (Java.Interop.JniObjectReference type, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args) [0x00073] in <fbb1482d0c6b44009a68f9f7e8c6389a>:0
    I MonoDroid:   at Android.Runtime.JNIEnv.CallStaticObjectMethod (System.IntPtr jclass, System.IntPtr jmethod, Android.Runtime.JValue* parms) [0x0000f] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at Android.Runtime.JNIEnv.CallStaticObjectMethod (System.IntPtr jclass, System.IntPtr jmethod, Android.Runtime.JValue[] parms) [0x0001a] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at Android.Runtime.JNIEnv.FindClass (System.String classname) [0x0003f] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at Android.Runtime.JNIEnv.AllocObject (System.String jniClassName) [0x00001] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at Android.Runtime.JNIEnv.StartCreateInstance (System.String jniClassName, System.String jniCtorSignature, Android.Runtime.JValue* constructorParameters) [0x0000a] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at Android.Runtime.JNIEnv.StartCreateInstance (System.String jniClassName, System.String jniCtorSignature, Android.Runtime.JValue[] constructorParameters) [0x00019] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at Android.Text.TextWatcherImplementor..ctor (System.Object inst, System.EventHandler`1[TEventArgs] changed_handler, System.EventHandler`1[TEventArgs] before_handler, System.EventHandler`1[TEventArgs] after_handler) [0x00010] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at Android.Widget.TextView.add_TextChanged (System.EventHandler`1[TEventArgs] value) [0x0001f] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at Topeka.Fragments.SignInFragment.InitContentViews (Android.Views.View view) [0x00011] in <201b9318435b446d821a19dfd1fa7bfb>:0
    I MonoDroid:   at Topeka.Fragments.SignInFragment.OnViewCreated (Android.Views.View view, Android.OS.Bundle savedInstanceState) [0x0003e] in <201b9318435b446d821a19dfd1fa7bfb>:0
    I MonoDroid:   at Android.App.Fragment.n_OnViewCreated_Landroid_view_View_Landroid_os_Bundle_ (System.IntPtr jnienv, System.IntPtr native__this, System.IntPtr native_view, System.IntPtr native_savedInstanceState) [0x0001a] in <4bc2304bf5d8408baea9dd059d7ea365>:0
    I MonoDroid:   at (wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.9(intptr,intptr,intptr,intptr)

The comparison of apk sizes with original and this fix (Topeka sample,
which was also used as repro):

| Version        | Size in bytes |
|----------------|--------------:|
| Before @4d8c28 |      21434941 |
| After  @4d8c28 |      28242405 |
| This fix       |      21638678 |

So the previous fix increased the apk size by around 6.5Mbytes, this
fix adds less than 199kbytes instead.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MonoDroidMarkStep.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MonoDroidMarkStep.cs
@@ -436,6 +436,14 @@ namespace MonoDroid.Tuner
 			PreserveRegisteredMethod (method.DeclaringType, member);
 		}
 
+		protected override bool ShouldMarkInterfaceImplementation (TypeDefinition type, InterfaceImplementation iface, TypeDefinition resolvedInterfaceType)
+		{
+			if (base.ShouldMarkInterfaceImplementation (type, iface, resolvedInterfaceType))
+				return true;
+
+			return resolvedInterfaceType.TryGetRegisterAdapter (out _);
+		}
+
 		protected override void DoAdditionalTypeProcessing (TypeDefinition type)
 		{
 			// If we are preserving a Mono.Android interface,
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/Extensions.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/Extensions.cs
@@ -120,14 +120,14 @@ namespace MonoDroid.Tuner {
 			return true;
 		}
 
-		public static bool TryGetRegisterAttribute (this MethodDefinition method, out CustomAttribute register)
+		static bool TryGetRegisterAttribute (ICustomAttributeProvider provider, out CustomAttribute register)
 		{
 			register = null;
 
-			if (!method.HasCustomAttributes)
+			if (!provider.HasCustomAttributes)
 				return false;
 
-			foreach (CustomAttribute attribute in method.CustomAttributes) {
+			foreach (CustomAttribute attribute in provider.CustomAttributes) {
 				if (!IsRegisterAttribute (attribute))
 					continue;
 
@@ -136,6 +136,25 @@ namespace MonoDroid.Tuner {
 			}
 
 			return false;
+		}
+
+		public static bool TryGetRegisterAdapter (this TypeDefinition td, out string adapter)
+		{
+		       CustomAttribute register;
+		       adapter = null;
+
+		       if (!TryGetRegisterAttribute (td, out register))
+			       return false;
+
+		       if (register.ConstructorArguments.Count != 3)
+			       return false;
+
+		       adapter = (string) register.ConstructorArguments [2].Value;
+
+		       if (string.IsNullOrEmpty (adapter))
+			       return false;
+
+		       return true;
 		}
 
 		public static bool TryGetRegisterMember (this MethodDefinition md, out string method)
@@ -150,7 +169,7 @@ namespace MonoDroid.Tuner {
 			nativeMethod = null;
 			signature = null;
 
-			if (!md.TryGetRegisterAttribute (out register))
+			if (!TryGetRegisterAttribute (md, out register))
 				return false;
 
 			if (register.ConstructorArguments.Count != 3)
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MarkJavaObjects.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MarkJavaObjects.cs
@@ -192,8 +192,6 @@ namespace MonoDroid.Tuner {
 			if (invoker == null)
 				return;
 
-			Annotations.Mark (invoker);
-
 			PreserveIntPtrConstructor (invoker);
 			PreserveInterfaceMethods (type, invoker);
 		}


jonpryor:
Addendum/clarification for the commit message:

> so the ITextWatcher interface was not marked anymore - as it is
> only accessed by reflection and instantiated during the runtime

I believe that this is incorrect.  I believe what was happening is that the referenced linker optimization meant that given:

```csharp
partial class TextWatcherImplementor : Java.Lang.Object, ITextWatcher {
}

// ...elsewhere...
var implementor = new TextWatcherImplementor ();
```

Assuming that the `new TextWatcherImplementor()` instruction was preserved, then the `new TextWatcherImplementor` type was preserved.

This was observed when investigating Issue #3263:

```
$ monodis --typedef obj/Release/android/assets/shrunk/Mono.Android.dll | grep TextW
133: Android.Text.TextWatcherImplementor (flist=934, mlist=1402, flags=0x100100, extends=0x4e0)
```

The problem was that after linking, `TextWatcherImplementor` no longer implemented `ITextWatcher`.  *This* change was what the linker optimization introduced: *Just Because* a type was preserved -- as was the case with `TextWatcherImplementor` -- with mono/linker@13ea158 that *did not* mean that all interfaces that the type implemented were preserved.

This is why, after linking, `TextWatcherImplementor` no longer implements `ITextWatcher`, and why `ITextWatcher` is not defined in the post-linked `Mono.Andrroid.dll`.

"Reflection" and "instantiation at runtime" have nothing to do with it.

radekdoulik:
To clarify the analysis of the issue, the problem wasn't only with the `TextWatcherImplementor` type, but also `AbsListViewer` which is based on `ITextWatcher` iface as well, and the iface wasn't preserved. That's what I observed in the linker. And that led me to the mentioned linker optimization.

> I believe that this is incorrect.

I believe it is correct.

What happened is this: because the linker didn't see any usage of the `ITextWatcher` methods implementation in the instantiated types, it removed the iface base from the relevant types (like `AbsListView` or `TextWatcherImplementor`) and didn't mark the iface type.

The reflection and instantiation is indeed relevant to this. The iface and its implementation (`ITextWatcherInvoker`) is only used during the runtime from XA with the help of the reflection. In the registration process, where it finds the `ITextWatcherInvoker` type name in the registration attribute. (as can be seen in the stacktrace in the commit message). And because we use it only through the reflection, we need to provide a way to preserve it, which is what this fix does.

jonpryor:
> That's what I observed in the linker.

Agreed, but none of that is *on-device app-runtime Reflection*.  The linker is run at packaging time, so packaging time is what determines which types are preserved & removed.

> The iface and its implementation (`ITextWatcherInvoker`) is only used during the runtime from XA with the help of the reflection.

Yes, which is why the linker looks for and explicitly preserves the `Invoker` types:

https://github.com/xamarin/xamarin-android/blob/5d865311166985eb867dce4074275dbd24b2ef91/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MarkJavaObjects.cs#L189

Note that this code looks for an `Invoker` *suffix*, and does not rely on the 3rd parameter to the `[Register]` attribute.

> In the registration process, where it finds the `ITextWatcherInvoker` type name in the registration attribute. (as can be seen in the stacktrace in the commit message)

I wanted to comment on this too, as well as on `TryGetRegisterAdapter()` and it's requirement of a 3rd parameter to `RegisterAttribute`.

Here's the thing: ~nothing (afaict) actually uses the `[Register]` attribute third parameter to determine the `*Invoker` type name.  *Instead*, we append `Invoker` to interface names and abstract class names in:

* [`MarkJavaObject.PreserveInvoker()`](https://github.com/xamarin/xamarin-android/blob/5d865311166985eb867dce4074275dbd24b2ef91/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MarkJavaObjects.cs#L191)
* [`JavaObjectExtensions.CastClass()`](https://github.com/xamarin/xamarin-android/blob/5d865311166985eb867dce4074275dbd24b2ef91/src/Mono.Android/Java.Interop/JavaObjectExtensions.cs#L89)

Furthermore, you mention the method registration process and the `TypeLoadException` mentioned in Issue #3263.  The `TypeLoadException` does *not* involve the `[Register]` attribute on *types*, not *directly*, afaict.  Instead, the `TypeLoadException` comes from the `[Register]` attribute *on interface methods*, which specifies the type to pass to `Type.GetType()`:

* https://gist.github.com/radekdoulik/ff047b822bb8b34569ed2f74ef34d45c
* https://github.com/xamarin/xamarin-android/blob/5d865311166985eb867dce4074275dbd24b2ef91/src/Mono.Android/Android.Runtime/AndroidRuntime.cs#L348

e.g.

```csharp
[Register ("android/text/TextWatcher", "", "***I do not believe that this value is actually used***", ApiSince = 1)]
partial interface ITextWatcher {
	[Register ("afterTextChanged", "(Landroid/text/Editable;)V",
	  "GetAfterTextChanged_Landroid_text_Editable_Handler:Android.Text.ITextWatcherInvoker, Mono.Android, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null")]
	void AfterTextChanged (Android.Text.IEditable s);
}
```

`[Register]` on interface types has little to do with it, other than as a way to tell the linker that this is a Java type (and I wonder if we should instead use `IJavaPeerable`/`IJavaObject` as such a marker?).

---

I *agree* with your analysis that the cause of the problem was the linker optimization within mono/linker@13ea158, because after that commit interfaces of preserved types were no longer preserved: we preserved `TextWatcherImplementor`, but after linking `TextWatcherImplementor` no longer implement `ITextWatcher`, and there was no longer an `ITextWatcher` interface at all.

None of that is "on-device app runtime", that's all linker time.

The *use* and lookup of the `ITextWatcherInvoker` interface *is* runtime, but that type would already be preserved *so long as* the `ITextWatcher` type is preserved, based on *naming convention* (`Invoker` suffix), *not* via `[Register]` attribute.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MonoDroidMarkStep.cs
```diff
@@ -436,6 +436,14 @@ protected override void DoAdditionalMethodProcessing (MethodDefinition method)
 			PreserveRegisteredMethod (method.DeclaringType, member);
 		}
 
+		protected override bool ShouldMarkInterfaceImplementation (TypeDefinition type, InterfaceImplementation iface, TypeDefinition resolvedInterfaceType)
+		{
+			if (base.ShouldMarkInterfaceImplementation (type, iface, resolvedInterfaceType))
+				return true;
+
+			return resolvedInterfaceType.TryGetRegisterAdapter (out _);
```
jonpryor: Based on [these musings](https://github.com/xamarin/xamarin-android/pull/3525#issuecomment-523270275), I don't think that this method should look for `[Register]` on an interface to determine if the interface type should be preserved.  The 3rd constructor argument is meaningless, and ~anybody can apply `[Register]` on their type.  (Why they would *do* so, ¯\_(ツ)_/¯)

```csharp
[Register("terrible/Example", "", "I mean really terrible, no?")]
public interface IBadExample {
}
```

Instead, I think a *better* indicator would be if `resolvedInterfaceType` implements the `IJavaPeerable` or `IJavaObject` interfaces.  Yes, this can also be done by anyone, but it's also *semantically meaningful*: if the interface implements `IJavaPeerable`/`IJavaObject`, then it can be passed into Java code and gains lots of special abilities (via `Java.Lang.Object` and JCW generation and...):

```csharp
[Register("better/Example", "", "I still don't think that this value is meaningfully used")]
public interface IBetterExample : IJavaObject, IJavaPeerable {
}
```

jonpryor:
Suggested commit message for merging (please discuss if incomplete!)

```
Fixes: https://github.com/xamarin/xamarin-android/issues/3454

Context: https://github.com/xamarin/xamarin-android/issues/3263

Commit 4d8c28f had an undesirable side effect that it *significantly*
increased `.apk` size, far more than necessary, because it marked
*all* `Invoker` types for preservation, because `MarkJavaObjects`
iterates over all `Java.Lang.Object` subclasses before linking.
For example, an "Android single view app" (VS template) grew from
~11MiB with Xamarin.Android 9.3.0.23 to ~18MiB in XA 9.4.1.0.

Further investigation suggests that analysis in 4d8c28f was
incomplete: the *actual* cause for Issue #3263 -- in which
`Android.Text.TextWatcherImplementor` would be preserved while the
`Android.Text.ITextWatcher` and `Android.Text.ITextWatcherInvoker`
types were removed by the linker -- was due to [linker@13ea1582][0]:

> Interfaces on a type will no longer be marked simply because the
> type is marked. Instead, marking of interfaces on types is deferred
> until that interface type is marked for some other reason.

Until 4d8c28f, there was no such "other reason", which is why the
linker *preserved* `TextWatcherImplementor` while *removing* the
`ITextWatcher` interface it implemented.  (Then, because
`ITextWatcher` wasn't preserved, `ITextWatcherInvoker` wasn't either.)

The fix is to override the "new"
`MarkStep.ShouldMarkInterfaceImplementation()` method, which allows
us to tell the linker that bound Java interfaces are "special", in
a way similar to COM interfaces and WindowsRuntime interfaces, and
thus they should *always* be preserved if any type implementing them
is also preserved.

[0]: https://github.com/mono/linker/commit/13ea1582e518f5f64cb93057aca962983e58e337
```

radekdoulik:
>>  That's what I observed in the linker.

> Agreed, but none of that is on-device app-runtime Reflection. 

Sure, here I wanted to give you more context. I didn't say that the runtime reflection is involved in this part. It comes into play later during the app runtime, where the iface invoker is linked away and the registration code tries to load it thru the `Type.GetType(string,bool)`.

>>The iface and its implementation (ITextWatcherInvoker) is only used during the runtime from XA with the help of the reflection.

> Yes, which is why the linker looks for and explicitly preserves the Invoker types:

> xamarin-android/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MarkJavaObjects.cs

> Line 189 in 5d86531
>  void PreserveInvoker (TypeDefinition type) 

> Note that this code looks for an Invoker suffix, and does not rely on the 3rd parameter to the [Register] attribute.

Note that `PreserveInvoker` only adds invoker methods to the preserve list, but doesn't mark the type. So that doesn't mean the invoker will survive the linking. Otherwise we would have not this crash.

The 3rd parameter was still used in the linker until https://github.com/xamarin/xamarin-android/commit/5b945abb2e1f0fe1a9ab6589993c970869c4a752 and the fix is reviving a bit of it.

This fix is not even using the 3rd parameter value. It just checks that the iface type has register attribute and that the parameter is not empty.

> Furthermore, you mention the method registration process and the TypeLoadException mentioned in Issue #3263. The TypeLoadException does not involve the [Register] attribute on types, not directly, afaict. Instead, the TypeLoadException comes from the [Register] attribute on interface methods, which specifies the type to pass to Type.GetType():

I think this is incorrect, the type name comes from java side, where the java class contains static `__md_methods` string, and that's where the `ITextWatcherInvoker` name comes from during the registration.

> I agree with your analysis that the cause of the problem was the linker optimization within mono/linker@13ea158, because after that commit interfaces of preserved types were no longer preserved: we preserved TextWatcherImplementor, but after linking TextWatcherImplementor no longer implement ITextWatcher, and there was no longer an ITextWatcher interface at all.

> None of that is "on-device app runtime", that's all linker time.

I didn't say the linker part of it happens during app runtime, so I am not sure why do you insist on this. The crash part is happening at runtime and is result of what happened during linking.

> The use and lookup of the ITextWatcherInvoker interface is runtime, but that type would already be preserved so long as the ITextWatcher type is preserved, based on naming convention (Invoker suffix), not via [Register] attribute.

Sure, that is how the fix works. It makes sure that the `ITextWatcher` stays in the assembly.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MonoDroidMarkStep.cs
radekdoulik: That would work too. I wanted to make it close to what linker did before (https://github.com/xamarin/xamarin-android/commit/5b945abb2e1f0fe1a9ab6589993c970869c4a752), thus I reused the code, which used and checked the Register attribute on the iface.

Wonder about the Register attribute on types though. Why do we still generate it with the 3rd argument then? Was the linker only user of it? Do we have documentation for the Register attributes our generator produces?

radekdoulik:
> Suggested commit message for merging (please discuss if incomplete!)

I like my original commit message more, it contains more information about the crash and the apk sizes.

I can extend it with more information about why the `ITextWatcher` didn't survive linking if needed.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MonoDroidMarkStep.cs
jonpryor: > Why do we still generate it with the 3rd argument then?

I think it was mostly a "this makes sense" decision, rather than anything concrete, as it (kinda) "mirrored" the connector value for `[Register]` on methods, which *is* used.

I can't find any evidence of the 3rd parameter ever being used on *types*, *certainly* not to find the `Invoker` for the interface (which is determined by appending `Invoker` to the type name, as linked elsewhere).

Thus, I think we should *stop* emitting that 3rd argument within `generator`.

The only reason I can think of to keep it is to use that value preferentially over appending `Invoker` to the type name.  We should measure this, but I doubt that a `Type.GetCustomAttributes(typeof(RegisterAttribute))` invocation will be faster than the existing "grab the type name and append `Invoker`" approach.

(Additionally, with the introduction of C#8 default interface methods, I'm highly tempted to just require that the `Invoker` type be a nested type, which could be done consistently for *everything*, and would instead involve a `Type.GetNestedType()` invocation.  However, this would probably constitute an ABI break, as newer assemblies couldn't work with previous releases...)

> Do we have documentation for the Register attributes our generator produces

Somewhat: https://docs.microsoft.com/en-us/xamarin/android/platform/java-integration/working-with-jni#binding-interfaces

I don't find any uses of `[Register]` *on types* within that document in which more than one constructor parameter is provided.

jonpryor:
> > The TypeLoadException does not involve the [Register] attribute on types, not directly, afaict. Instead, the TypeLoadException comes from the [Register] attribute on interface methods, which specifies the type to pass to Type.GetType():

> I think this is incorrect, the type name comes from java side, where the java class contains static `__md_methods` string, and that's where the `ITextWatcherInvoker` name comes from during the registration.

Yes, the string provided to `AndroidTypeManager.RegisterNativeMembers()` comes from `__md_methods`, and `__md_methods` contains `ITextWatcherInvoker`, as per private chat awhile ago:

```
                __md_methods = 
                        "n_afterTextChanged:(Landroid/text/Editable;)V:GetAfterTextChanged_Landroid_text_Editable_Handler:Android.Text.ITextWatcherInvoker, Mono.Android, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null\n" +
                        "n_beforeTextChanged:(Ljava/lang/CharSequence;III)V:GetBeforeTextChanged_Ljava_lang_CharSequence_IIIHandler:Android.Text.ITextWatcherInvoker, Mono.Android, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null\n" +
                        "n_onTextChanged:(Ljava/lang/CharSequence;III)V:GetOnTextChanged_Ljava_lang_CharSequence_IIIHandler:Android.Text.ITextWatcherInvoker, Mono.Android, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null\n" +
                        "";
```

But where does `__md_methods` *come from*?  *That* comes from `JavaCallableWrapperGenerator`, which in turn uses the `[Register]` values specified on the interface methods:

```csharp
	[Register ("android/text/TextWatcher", "", "Android.Text.ITextWatcherInvoker", ApiSince = 1)]
	public partial interface ITextWatcher : Android.Text.INoCopySpan {

		[Register ("afterTextChanged", "(Landroid/text/Editable;)V", "GetAfterTextChanged_Landroid_text_Editable_Handler:Android.Text.ITextWatcherInvoker, Mono.Android, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null")]
		void AfterTextChanged (Android.Text.IEditable s);
	}
```

Thus, I re-assert that *nowhere* in that chain is the 3rd constructor parameter specified on the interface type *used*.  Rather, the chain is:

 1. `class-parse` emits XML declaration
 2. `generator` consumes (1), generates C# binding declarations, binding assembly produced.
 3. `jcw-gen.exe`/`<GenerateJavaStubs/>` task takes assembly from (2) -- possibly after assembly is linked -- and generates Java Callable Wrappers
4. (3) uses the `[Register]` attributes on types to determine Java-side base class names & implemented interface names, which involves the 1st `[Register]` parameter, and the `[Register]` on methods to generate method declarations and `__md_methods` contents, which uses all `[Register]` parameters.

Nowhere in this process, afaict, is the 3rd `[Register]` parameter specified on types *used*.

PR #3528, opened 2019-08-20
[Xamarin.Android.Build.Tasks] improve incremental builds + designer

jonathanpeppers:
I was reviewing MSBuild logs inside the IDE when using the Android
designer and noticed something odd:

    Building target "_CompileToDalvikWithDx" completely.
    Input file "obj\Debug\android\bin\classes.zip" is newer than output file "obj\Debug\stamp\_CompileToDalvik.stamp".

It appears that editing an Android layout was causing
`_CompileToDalvikWithDx` to run again?

Then I looked at the build log from `SetupDependenciesForDesigner`:

    Building target "SetupDependenciesForDesigner" completely.
    Output file "obj\Debug\android\bin\classes\mono\MonoPackageManager.class" does not exist.

Ok, we no longer generate `MonoPackageManager.class`. It is included
as part of `java-runtime.jar`.

Looking closer, the `Inputs` and `Outputs` did not seem right for
`SetupDependenciesForDesigner`:

    Inputs="$(IntermediateOutputPath)android\src\mono\MonoPackageManager.java;$(_AndroidTypeMappingJavaToManaged);$(_AndroidTypeMappingManagedToJava)"
    Outputs="$(IntermediateOutputPath)android\bin\classes\mono\MonoPackageManager.class"

We no longer generate `$(_AndroidTypeMappingJavaToManaged)` or
`$(_AndroidTypeMappingManagedToJava)`, so we would also hit:

    Building target "SetupDependenciesForDesigner" completely.
    Input file "obj\Debug\android\typemap.jm" does not exist.

In the current state of affairs, `SetupDependenciesForDesigner` will
always run -- and trigger `_CompileToDalvikWithDx` on incremental
*regular* builds.

Two goals here would make things a lot better:

1. `SetupDependenciesForDesigner` should be skippable.
2. `SetupDependenciesForDesigner` should not trigger anything later if
   you do a regular build.

For no. 1 we can more appropriately write the `Inputs` and `Outputs` as:

    Inputs="$(MSBuildAllProjects);$(MonoPlatformJarPath);@(_JavaStubFiles);@(AndroidJavaSource)"
    Outputs="$(_AndroidStampDirectory)SetupDependenciesForDesigner.stamp"

This matches what we do for `_CompileJava`, and
`SetupDependenciesForDesigner` can manage its own stamp file.

For goal no. 2, however, we need an option for the `<Javac/>` MSBuild
task to *not* generate `classes.zip`. The designer does not use this
file, so we don't need to create it! We also don't want a regular
build later to see the timestamp on `classes.zip` change.

## Xamarin.ProjectTools ##

In trying to write some tests, I actually found a bug in
Xamarin.ProjectTools:

* Call `Touch()` on a file
* The timestamp on the `.csproj` is updated -- when it did not change

To account for this, I reworked the `XamarinProject` class to save a
copy of the `ProjectResource` instance between builds.

These changes allowed me to add the needed tests in `DesignerTests`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -2552,6 +2552,7 @@ because xbuild doesn't support framework reference assemblies.
   <Javac
     JavaPlatformJarPath="$(JavaPlatformJarPath)"
     ClassesOutputDirectory="$(IntermediateOutputPath)android\bin\classes"
+    ClassesZip="$(IntermediateOutputPath)android\bin\classes.zip"
     TargetFrameworkDirectory="$(TargetFrameworkDirectory)"
     StubSourceDirectory="$(IntermediateOutputPath)android\src"
     JavaSourceFiles="@(AndroidJavaSource)"
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -1382,7 +1382,7 @@ because xbuild doesn't support framework reference assemblies.
 
 <Target Name="_ResolveLibraryProjectImports"
 		DependsOnTargets="$(CoreResolveReferencesDependsOn)"
-		Inputs="$(ProjectAssetsFile);$(MSBuildProjectFullPath);@(_MonoAndroidReferencePath);$(_AndroidBuildPropertiesCache)"
+		Inputs="$(ProjectAssetsFile);$(MSBuildProjectFullPath);@(_MonoAndroidReferencePath);@(AndroidAarLibrary);$(_AndroidBuildPropertiesCache)"
 		Outputs="$(_AndroidStampDirectory)_ResolveLibraryProjectImports.stamp">
 	<ResolveLibraryProjectImports
 		ContinueOnError="$(DesignTimeBuild)"
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Designer.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Designer.targets
@@ -135,8 +135,8 @@ Copyright (C) 2016 Xamarin. All rights reserved.
 
 <Target Name="SetupDependenciesForDesigner"
     DependsOnTargets="UpdateAndroidResources;_AdjustJavacVersionArguments;_GeneratePackageManagerJavaForDesigner;_GetMonoPlatformJarPath;_DetermineJavaLibrariesToCompile"
-    Inputs="$(IntermediateOutputPath)android\src\mono\MonoPackageManager.java;$(_AndroidTypeMappingJavaToManaged);$(_AndroidTypeMappingManagedToJava)"
-    Outputs="$(IntermediateOutputPath)android\bin\classes\mono\MonoPackageManager.class">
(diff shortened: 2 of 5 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2584,6 +2584,7 @@ because xbuild doesn't support framework reference assemblies.
   <Javac
     JavaPlatformJarPath="$(JavaPlatformJarPath)"
     ClassesOutputDirectory="$(IntermediateOutputPath)android\bin\classes"
+    ClassesZip="$(IntermediateOutputPath)android\bin\classes.zip"
```
jonathanpeppers: @garuma I wanted to verify that the designer does not in fact use this file -- and it uses the `.class` files directly? `SetupDependenciesForDesigner` will no longer generate `classes.zip`.

If `classes.zip` is used, I can revisit this and do something else.

dellis1972: @jonathanpeppers ideally we should use the classes.zip. We switched to that to get around various command line and path length issues.

jonathanpeppers: This won't change a regular build, we will still use `classes.zip` as input to dx/d8.

With this change, the designer won't generate `classes.zip` (if it doesn't use it), so that prevents us from re-running the dx/d8 step on an incremental build.

What I was seeing:
* Edit layout in designer
* Save, designer runs some targets, `classes.zip` changes
* On the next build we run dx/d8 again, even though nothing on the Java side changed.

jonathanpeppers:
There are some test failures I need to look into:
```
Xamarin.Android.Build.Tests.BuildTest.AarContentExtraction(False)
    _ResolveLibraryProjectImports should have run.
    Expected: False
    But was: True
```
Likely was caused by the new Xamarin.ProjectTools behavior

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
dellis1972: my concern here is that the designer build will fail because of path length issues...

jonathanpeppers: The designer build doesn't run dx/d8, right? Unless I missed something?

I think the designer just needs aapt/javac to be able to load.

It seems like `classes.zip` isn't used by the designer at all, because the integration tests passed.

garuma: > I think the designer just needs aapt/javac to be able to load.

Correct

jonathanpeppers:
Ahhhhhh, more bugs!
```
Skipping target "_ResolveLibraryProjectImports" because all output files are up-to-date with respect to the input files.
Input files: C:\src\xamarin-android\bin\Debug\lib\xamarin.android\xbuild-frameworks\MonoAndroid\v1.0\Java.Interop.dll;C:\src\xamarin-android\bin\Debug\lib\xamarin.android\xbuild-frameworks\MonoAndroid\v10.0\Mono.Android.dll;C:\src\xamarin-android\bin\TestDebug\temp\AarContentExtractionFalse\UnnamedProject.csproj;obj\Debug\build.props
Output files: obj\Debug\stamp\_ResolveLibraryProjectImports.stamp
```
This was a test that does a `touch` on an `@(AndroidAarLibrary)`:

https://github.com/xamarin/xamarin-android/blob/d762aa993491e93cc006e3e6207ad40097159fa9/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L2299-L2302

The fixes in Xamarin.ProjectTools brought this to the surface...

PR #3531, opened 2019-08-21
[Xamarin.Android.Build.Task] Handle res-auto headerLayout

dellis1972:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/969915

Certain android resource items can make use of the following
namespace

	xmlns:app=""http://schemas.android.com/apk/res-auto""

this provides additional extension attributes which need
to be fixed up. Some examples are `actionLayout`, `rectLayout`
and `roundLayout`. The values for these attributes should
be lowercased as they refer to a `@layout` item.

It seems however that this list is always expanding. A new
item `headerLayout` seems to have appeared. Rather than handcoding
each of these items, the code here changes to check for `Layout`
in the attrbiute and will then lowercase the value. This should
protect us from future "additions", and help our customers since
they won't have to work around casing problems for attributes we
do not support.


dellis1972:
BCL Test Failures are not related.

PR #1939, opened 2018-07-09
default interface methods support

atsushieno:
1) You have to manually provide CscToolPath MSBuild property,
   unless you install xamarin.android.csc.dim package.
2) You have to explicitly specify AndroidEnableDefaultInterfaceMethods
   MSBuild property as True.
3) You have to explicitly specify LangVersion MSBuild property as "latest".

When C# 8.0 becomes stable, 1) and 3) will become unnecessary.
Then we should make changes to 2) as default.

Remaining tasks:

- [x] add support for DIMs in Xamarin.Android.Bindings.targets so that DIM bindings can be generated.
- [x] add apk tests that assures DIMs work as expected at run time.
- [x] add DIM-backed methods in Mono.Android.dll (for now the new ones are disabled; those methods that "became default" are generated as non-default)
- [x] ~~(optional) update xamarin-android Xamarin.Android.Bindings.targets to customize csc tool task invocation, if no CscToolPath is provided.~~ -> there is no csc-dim on Windows, so skip that.

PR #3533, opened 2019-08-21
[Xamarin.Android.Build.Tasks] Add MSBuild support for DIM.

jpobst:
Adds MSBuild support for DIM to Android Bindings projects.

To enable, add the following to a binding project:
```
<LangVersion>preview</LangVersion>
<_EnableInterfaceMembers>True</_EnableInterfaceMembers>
```

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets
@@ -494,6 +494,8 @@ Copyright (C) 2012 Xamarin Inc. All rights reserved.
       ToolPath="$(MonoAndroidToolsDirectory)"
       ToolExe="$(BindingsGeneratorToolExe)"
       UseShortFileNames="$(UseShortGeneratorFileNames)"
+      LangVersion="$(LangVersion)"
+      EnableInterfaceMembersPreview="$(_EnableInterfaceMembers)"
     />
 
     <!-- Write a flag so we won't redo this target if nothing changed -->
--- a/Documentation/release-notes/master.md
+++ b/Documentation/release-notes/master.md
@@ -17,8 +17,35 @@ Description of new feature
 
   * Description of known issue in the new feature, if applicable
 
+### Static/Default Interface Methods (Preview)
+
+With the new support for default interface methods in C# 8.0 we can now produce bindings that better match Java libraries 
+that use these features.  This includes:
+
+* Default interface methods
+* Static interface methods
+* Interface constants
+
+To enable this preview in your bindings project, add the following properties to your `.csproj`:
+
+```
+<LangVersion>preview</LangVersion>
+<_EnableInterfaceMembers>True</_EnableInterfaceMembers>
+```
+
+Note that enabling this only adds new members, it does not remove the existing alternatives previously used to expose
+these methods/fields.
+
+
 ### Build and deployment performance
 
+  * Bindings projects should now build considerably faster:
+    * [GitHub PR 440](https://github.com/xamarin/java.interop/pull/440)
+    * [GitHub PR 441](https://github.com/xamarin/java.interop/pull/441)
+    * [GitHub PR 442](https://github.com/xamarin/java.interop/pull/442)
+    * [GitHub PR 448](https://github.com/xamarin/java.interop/pull/448)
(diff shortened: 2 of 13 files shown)


tests/CodeGen-Binding/Xamarin.Android.DefaultInterfaceMethods-Tests/DimBindingTests.cs
```diff
@@ -0,0 +1,39 @@
...
+			Assert.AreEqual (2, iface.Bar);
+			Assert.DoesNotThrow (() => iface.Bar = 5);
+
+			Assert.AreEqual (0, iface.InvokeFoo ());
+
+			Assert.Throws<UnsupportedOperationException> (() => iface.ToImplement ());
+		}
+
+		[Test]
+		public void TestOverriddenDefaultInterfaceMethods ()
```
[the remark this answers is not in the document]
jpobst: Added, which revealed a new issue that required fixing. (https://github.com/xamarin/java.interop/pull/486)

jpobst:
I'm not aware of any benefit to `Xamarin.Android.DefaultInterfaceMethods-Lib`. I merely ported it from https://github.com/xamarin/xamarin-android/pull/1939.  I suspect it was needed at the time because it required a custom compiler with DIM support.  Now that DIM is built into shipping compilers and runtimes that isn't needed.  Moved tests to `McwGen` and `JcwGen`.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets
```diff
@@ -494,6 +494,8 @@ Copyright (C) 2012 Xamarin Inc. All rights reserved.
       ToolPath="$(MonoAndroidToolsDirectory)"
       ToolExe="$(BindingsGeneratorToolExe)"
       UseShortFileNames="$(UseShortGeneratorFileNames)"
+      LangVersion="$(LangVersion)"
+      EnableInterfaceMembersPreview="$(EnableInterfaceMembersPreview)"
```
jonathanpeppers: Do we need to document this yet? https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/BuildProcess.md

dellis1972: Good catch, I forgot about documentation.

jonpryor: @jpobst: please update this section: https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/BuildProcess.md#binding-project-build-properties

To document the new `$(EnableInterfaceMembersPreview)` property.

Additionally, convention is that public properties begin with `Android`.

The `Preview` suffix is...disturbing?  What's the longevity for this property?

Should this instead be a `$(_EnableInterfaceMembers)` property?  A `_` prefix is "private"/"subject to change without notice" designator, so maybe that's what we should use?

jpobst: The intention is to explicitly denote that this field is temporary and not something we are going to support for the rest of time.  I am fine with a "private" property if that conveys the same idea.  :)

jonpryor:
[`Xamarin.Android.Bcl_Tests-NUnit` crashed](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3026697&view=ms.vss-test-web.build-test-results-tab&runId=8351948&paneView=debug), but the associated logical doesn't really mention a "crash":

```
09-06 22:32:49.779  3242  5090 I NUnit   : Passed: 22173, Failed: 0, Skipped: 419, Inconclusive: 1, Total: 0, Filtered: 0
09-06 22:32:49.788  1954  1973 I ActivityManager: Force stopping Xamarin.Android.Bcl_Tests appid=10103 user=0: finished inst
09-06 22:32:49.789  1954  1973 I ActivityManager: Killing 3242:Xamarin.Android.Bcl_Tests/u0a103 (adj 0): stop Xamarin.Android.Bcl_Tests
...
09-06 22:32:49.881  1651  1651 I Zygote  : Process 3242 exited due to signal 9 (Killed)
```

It was force-stopped, for some reason, which was detected as a crash.

Ignoring.

PR #3535, opened 2019-08-22
[Xamarin.Android.Build.Tasks] fully incremental linker for Debug builds

jonathanpeppers:
If I time an incremental build of a "Hello World" Xamarin.Forms app,
one of the last remaining areas to improve is:

    340 ms  LinkAssemblies                             1 calls

In this build, I made a one-line change in a XAML file in a
NetStandard project with the optimal settings such as
`$(ProduceReferenceAssembly)`.

On a slower PC I have with Windows defender enabled, this step can
take even longer:

    767 ms  LinkAssemblies                             1 calls

Part of the reason for the time taken is the linker needs to load
*every* assembly, even if not every assembly is actually changed or
saved.

A `Debug` build (or `$(AndroidLinkMode)` = `None`) only needs to run a
single linker step:

    var pipeline = new Pipeline ();
    if (options.LinkNone) {
        pipeline.AppendStep (new FixAbstractMethodsStep ());
        pipeline.AppendStep (new OutputStepWithTimestamps ());
        return pipeline;
    }

`FixAbstractMethodsStep` also only applies to "Xamarin.Android"
assemblies which would reference `Mono.Android.dll`. The
`OutputStepWithTimestamps` step is also responsible for getting all
assemblies to their final destination in
`$(IntermediateOutputPath)android\assets\`.

In an ideal world, in an incremental build:

* A `Foo.dll` NetStandard assembly changes
    * We know it is *not* a "Xamarin.Android" assembly
* The linker step loads 0 assemblies
    * We merely copy `Foo.dll` to `android\assets\`

We can achieve this with a new `<LinkAssembliesNoShrink/>` MSBuild
task that doesn't use the normal code path for the full linker. We can
split out `FixAbstractMethodsStep` so we can call it on a per-assembly
basis as needed.

## Partial Builds ##

To achieve our ideal scenario, we need to leverage a feature of
MSBuild that enables an MSBuild target to build *partially*:

    Building target "Foo" partially, because some output files are out of date with respect to their input files.

For example, we had a target such as:

    <Target Name="_GenerateDocumentation"
        Inputs="@(_JavaFiles)"
        Outputs="@(_JavaFiles->'$(DocsDirectory)%(Filename).md')">
      <GenerateDocumentation
          SourceFiles="@(_JavaFiles)"
          DestinationFiles="@(_JavaFiles->'$(DocsDirectory)%(Filename).md')"
      />
    </Target>

MSBuild can compare the `Inputs` and `Outputs` and only pass in the
`@(_JavaFiles)` that have changed to `<GenerateDocumentation/>`.

The trick, is that `<LinkAssembliesNoShrink/>` needs to process a
subset of the assemblies that changed, while still *knowing* the full
list of assemblies. An assembly could change that references another
assembly that Mono.Cecil needs to be able to load.

So we can record an `@(_AllResolvedAssemblies)` item group such as:

    <Target Name="_LinkAssembliesNoShrinkInputs">
      <ItemGroup>
        <_AllResolvedAssemblies Include="@(ResolvedAssemblies)" />
      </ItemGroup>
    </Target>

    <Target Name="_LinkAssembliesNoShrink"
        DependsOnTargets="_LinkAssembliesNoShrinkInputs"
        Inputs="@(ResolvedAssemblies);$(_AndroidBuildPropertiesCache)"
        Outputs="@(ResolvedAssemblies->'$(MonoAndroidIntermediateAssemblyDir)%(Filename)%(Extension)')">
      <LinkAssembliesNoShrink
          ResolvedAssemblies="@(_AllResolvedAssemblies)"
          SourceFiles="@(ResolvedAssemblies)"
          DestinationFiles="@(ResolvedAssemblies->'$(MonoAndroidIntermediateAssemblyDir)%(Filename)%(Extension)')"
      />
    ...
    </Target>

This way `SourceFiles` and `DestinationFiles` files have a subset of
assemblies, while `ResolvedAssemblies` can still have a known list of
all assemblies.

## Results ##

Incremental builds are improved dramatically for a one-line XAML
change:

    Before:
    340 ms  LinkAssemblies                             1 calls
    After:
      9 ms  LinkAssembliesNoShrink                     1 calls

The only I/O operation is a `MonoAndroidHelper.CopyIfChanged` on a
single assembly.

## Other Refactoring ##

Since `<LinkAssemblies/>` is no longer used for `Debug` builds, we can
remove a few options that are not currently used for `Release` builds:

* `OptionalDestinationDirectory`
* `LinkOnlyNewerThan`

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
@@ -1,0 +1,98 @@
+using Java.Interop.Tools.Cecil;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using Mono.Cecil;
+using Mono.Cecil.Cil;
+using System;
+using System.IO;
+
+namespace Xamarin.Android.Tasks
+{
+	/// <summary>
+	/// This task is for Debug builds where LinkMode=None, LinkAssemblies is for Release builds
+	/// </summary>
+	public class LinkAssembliesNoShrink : Task
+	{
+		/// <summary>
+		/// These are used so we have the full list of SearchDirectories
+		/// </summary>
+		[Required]
+		public ITaskItem [] ResolvedAssemblies { get; set; }
+
+		[Required]
+		public ITaskItem [] SourceFiles { get; set; }
+
+		[Required]
+		public ITaskItem [] DestinationFiles { get; set; }
+
+		public override bool Execute ()
+		{
+			if (SourceFiles.Length != DestinationFiles.Length)
+				throw new ArgumentException ("source and destination count mismatch");
+
+			var readerParameters = new ReaderParameters {
+				ReadSymbols = true,
+			};
+			var writerParameters = new WriterParameters {
+				DeterministicMvid = true,
+			};
+
+			using (var resolver = new DirectoryAssemblyResolver (this.CreateTaskLogger (), loadDebugSymbols: true, loadReaderParameters: readerParameters)) {
+				// Add SearchDirectories with ResolvedAssemblies
+				foreach (var assembly in ResolvedAssemblies) {
+					var path = Path.GetFullPath (Path.GetDirectoryName (assembly.ItemSpec));
+					if (!resolver.SearchDirectories.Contains (path))
+						resolver.SearchDirectories.Add (path);
+				}
+
+				// Run the FixAbstractMethodsStep
+				var step = new FixAbstractMethodsStep (resolver, Log);
+				for (int i = 0; i < SourceFiles.Length; i++) {
+					var source = SourceFiles [i];
+					var destination = DestinationFiles [i];
+
+					// Only run the step on "MonoAndroid" assemblies
+					if (MonoAndroidHelper.IsMonoAndroidAssembly (source) && !MonoAndroidHelper.IsSharedRuntimeAssembly (source.ItemSpec)) {
+						var assemblyDefinition = resolver.GetAssembly (source.ItemSpec);
+						if (step.FixAbstractMethods (assemblyDefinition)) {
+							Log.LogDebugMessage ($"Saving modified assembly: {destination.ItemSpec}");
+							writerParameters.WriteSymbols = assemblyDefinition.MainModule.HasSymbols;
+							assemblyDefinition.Write (destination.ItemSpec, writerParameters);
+							continue;
+						}
+					}
+
+					if (MonoAndroidHelper.CopyAssemblyAndSymbols (source.ItemSpec, destination.ItemSpec)) {
+						Log.LogDebugMessage ($"Copied: {destination.ItemSpec}");
+					} else {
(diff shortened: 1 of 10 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
```diff
@@ -0,0 +1,98 @@
...
+				}
+
+				// Run the FixAbstractMethodsStep
+				var step = new FixAbstractMethodsStep (resolver, Log);
+				for (int i = 0; i < SourceFiles.Length; i++) {
+					var source = SourceFiles [i];
+					var destination = DestinationFiles [i];
+
+					// Only run the step on "MonoAndroid" assemblies
+					if (!MonoAndroidHelper.IsSharedRuntimeAssembly (source.ItemSpec) && MonoAndroidHelper.IsMonoAndroidAssembly (source)) {
```
dellis1972: should we care about the order of the `&&` here. C# will have an early out option in `if` statements, so the quickest check should be first in the list. We should probably try to remember that when reviewing code from here on :)

My guess is the BinarySearch will be faster than the `assembly.GetMetadata` call, correct?

Documentation/guides/MSBuildIncrementalBuilds.md
```diff
@@ -0,0 +1,149 @@
...
+performant for incremental builds to build *partially*. When one
+`.java` file changes, we only have to generate one documentation file.
+
+Building partially also works, if you need to invalidate on additional
+files, such as:
+
+```xml
+<Target Name="_GenerateDocumentation"
+    Inputs="@(_JavaFiles)"
+    Outputs="@(_JavaFiles->'$(DocsDirectory)%(Filename).md');$(MSBuildAllProjects)">
```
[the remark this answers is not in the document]
jonathanpeppers: Whoops 👀

src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
```diff
@@ -0,0 +1,98 @@
...
+using Mono.Cecil.Cil;
+using System;
+using System.IO;
+
+namespace Xamarin.Android.Tasks
+{
+	/// <summary>
+	/// This task is for Debug builds where LinkMode=None, LinkAssemblies is for Release builds
+	/// </summary>
+	public class LinkAssembliesNoShrink : Task
```
radekdoulik: Wouldn't `FixAbstractMethods` be a more descriptive task name here?

Otherwise LGTM.

radekdoulik:
I like the new building *partially* feature of msbuild.

I think `_GenerateJniMarshalMethods` target might be another candidate for this. Do I understand it correctly that it would need to have the destination files in the outputs instead of the stamp file. Or would it need more changes there?

src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
dellis1972: I would agree, however we might add another step later on, so its probably best to keep it as a `Link` named task.

PR #3536, opened 2019-08-22
Add notarization to the Xamarin Android pipeline

somu84:


build-tools/automation/azure-pipelines.yaml
```diff
@@ -165,6 +167,14 @@ stages:
...
       parameters:
         UnsignedPkgPath: $(XA.Unsigned.Pkg)
+    
+    - script: |
+        cd $(Build.SourcesDirectory)/..
+        git clone -b $(ReleaseScriptsBranch) https://$(GitHub.Token):x-oauth-basic@github.com/xamarin/release-scripts
+        cd release-scripts
+        ruby notarize.rb $(XA.Unsigned.Pkg) $(XamarinIdentifier) $(XamarinUserId) $(XamarinPassword) $(TeamID)
+      displayName: Notarize PKG
+      condition: and(succeeded(), or(startsWith(variables['Build.SourceBranch'], 'refs/heads/master' ), startsWith(variables['Build.SourceBranch'], 'refs/heads/release' )))
```
pjcollins: Let's update this to use the same "Is not pull request" condition used elsewhere, as our release branches historically don't start with `release`:
```yaml
condition: and(eq(variables['XA.Commercial.Build'], 'true'), ne(variables['Build.Reason'], 'PullRequest'))
```

somu84: Yes the script handles stapling. I don't think we should be creating a template for just one task.

somu84: Pushed the change for the condition

pjcollins: Fair enough 👍

PR #3539, opened 2019-08-23
[ci] Fix vsix sign job and other conditions

pjcollins:
Context: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#dependencies

We shouldn't run the vsix sign job if any failures occurred in the build
before it. We also shouldn't run notarization and upload to storage steps
if any failures occur in our build.


kvpt:
There is a typo line 120.
> **eq**(succeeded(), eq(variables['XA.Commercial.Build'], 'true'))

Must be

> **and**(succeeded(), eq(variables['XA.Commercial.Build'], 'true'))

PR #3552, opened 2019-08-26
[ci] Move Windows build and test to VS2019

pjcollins:
We've been encountering a handful of issues recently with the Windows
machine pool we're using for building and testing. There have been quite
a few failures caused by files being locked, or dependencies being out
of date or missing. Moving to the VSEng VS2019 Windows pool will
hopefully resolve some of these issues.

Unfortunately the size of our build dependencies presents an issue when
attempting to use the Microsoft hosted Windows VS2019 agents, as our Win
job ultimately runs out of space when executing on those VMS. Using the
VSEng VS2019 pool will allow us to migrate to VS2019 while avoiding disk
space limitations.

As of commit 8cdda11, `$(JavaSdkDirectory)` will have a default value
of the path to the Corretto JDK we install on both Windows and macOS.
We should try to use this JDK path for all Android/Java tool invocations
which require `$(JAVA_HOME)` to be set.

Also adds a [process monitor/killer][0] to our Windows build jobs to attempt
to better monitor, or hopefully prevent, cases of our source directory being
in use by an external process when a job starts.

[0]: https://github.com/xamarin/azdev-extension/tree/4444d45d51e15ad8c84f0b8ce72ee710a71352f9/lingering-process-task

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs
@@ -17,7 +17,6 @@ namespace Xamarin.ProjectTools
 	{
 		const string SigSegvError = "Got a SIGSEGV while executing native code";
 		const string ConsoleLoggerError = "[ERROR] FATAL UNHANDLED EXCEPTION: System.ArgumentException: is negative";
-		static string frameworkSDKRoot = null;
 
 		string root;
 		string buildLogFullPath;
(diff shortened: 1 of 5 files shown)


pjcollins:
@dellis1972 @jonathanpeppers This should hopefully be ready to go now.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs
```diff
@@ -119,24 +119,25 @@ public class Builder : IDisposable
...
+					}
+					var x86 = Environment.GetFolderPath (Environment.SpecialFolder.ProgramFilesX86);
+					path = Path.Combine (x86, "MSBuild", "Sdks", "Microsoft.NET.Sdk");
 					if (File.Exists (Path.Combine (path, "Sdk", "Sdk.props")))
 						return path;
 				}
-				var x86 = Environment.GetFolderPath (Environment.SpecialFolder.ProgramFilesX86);
-				path = Path.Combine (x86, "MSBuild", "Sdks", "Microsoft.NET.Sdk");
-				if (File.Exists (Path.Combine (path, "Sdk", "Sdk.props")))
-					return path;
```
jonathanpeppers: Hey, I was looking at this is only used in one place: https://github.com/xamarin/xamarin-android/blob/9e7b246f7e8cb46248027260e90068ff2e99efd6/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs#L612-L613

Can we just remove that property and this assertion? I don't see how this is testing anything...

pjcollins: Heh I didn't actually look at usage, I agree we should be able to remove this and just end up failing later if we don't have the right SDK installed.

Issue #2957, opened 2019-04-10
Android builds are not fully deterministic with <Deterministic>true</Deterministic>.


plynkus:
### Steps to Reproduce

1. Obtain and run the latest VS2019 (currently 16.0.1 with Xamarin.Android SDK 9.2.0.5).
2. Create a blank solution.
3. Add a Xamarin.Android class library project.
4. (Add a Xamarin.iOS class library project for comparison, if you wish.)
5. (Add a UWP class library project for comparison, if you wish.)
6. To each added project, add <Deterministic>true</Deterministic> to the first existing <PropertyGroup>.
7. For each project, rebuild repeatedly and perform an md5sum on the output DLL.

### Expected Behavior

Stable hashes for each project type after each and every rebuild.

### Actual Behavior

For the Xamarin.Android project, hashes vary with each rebuild.  Rebuilds of the other project types result in stable hashes.

### Version Information
Microsoft Visual Studio Community 2019
Version 16.0.1
VisualStudio.16.Release/16.0.1+28803.156
Microsoft .NET Framework
Version 4.7.03190

Installed Version: Community

Visual C++ 2019   00435-60000-00000-AA963
Microsoft Visual C++ 2019

Application Insights Tools for Visual Studio Package   9.0.20307.1
Application Insights Tools for Visual Studio

ASP.NET and Web Tools 2019   16.0.12311.10635
ASP.NET and Web Tools 2019

ASP.NET Web Frameworks and Tools 2019   16.0.12311.10635
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   16.0.12311.10635
Azure App Service Tools v3.0.0

Azure Functions and Web Jobs Tools   16.0.12311.10635
Azure Functions and Web Jobs Tools

C# Tools   3.0.0-beta4-19170-01+1deafee3682da88bf07d1c18521a99f47446cee8
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.1.77 (master@24013d5)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

Microsoft Azure Tools   2.9
Microsoft Azure Tools for Microsoft Visual Studio 0x10 - v2.9.0.0

Microsoft Continuous Delivery Tools for Visual Studio   0.4
[...]

PR #3554, opened 2019-08-27
[Xamarin.Android.Build.Tasks] support Deterministic builds

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/2957
Context: https://blog.paranoidcoding.com/2016/04/05/deterministic-builds-in-roslyn.html

Since VS 2019, `$(Deterministic)` is set in the .NET framework class
library project template:

    <PropertyGroup>
      ...
      <Deterministic>True</Deterministic>
    </PropertyGroup>

It is also the default for SDK-style projects.

What this setting does is enable a stable MVID for the produced .NET
assembly. So if you do `touch Foo.cs && msbuild`, the produced
assembly will have its contents unchanged. However, the timestamp of
the file *will* change. There may be some narrow performance gains we
can get there.

The only drawback of `$(Determinitic)` is the following will produce a
build error:

    [assembly: AssemblyVersion ("1.0.*")]

Since `1.0.*` implies that the assembly changes on incremental builds,
`Deterministic=True` doesn't support it. So we can't just make it
default without breaking someone.

## What should we do? ##

We should start building Xamarin.Android using `Deterministic=True`
and add a few tests for it. We had some usage of `AssemblyVersion`
wildcards that seemed unimportant/not needed.

When `Deterministic=True`, any places using Mono.Cecil, need to use
the appropriate `DeterministicMvid` setting when writing assemblies.
The linker has an equivalent `DeterministicOutput` setting.

After this change makes it in, we should consider putting
`<Deterministic>True</Deterministic>` in the Xamarin templates. We
would also need to remove some comments in `AssemblyInfo.cs` that
suggests to use `[assembly: AssemblyVersion ("1.0.*")]`.

## Other changes ##

I added a few general fixes to Xamarin.ProjectTools:

* Made properties so `AndroidUseSharedRuntime` and
  `EmbedAssembliesIntoApk` are strongly typed.
* Added a `Deterministic` property
* In `BaseTest`, `CreateApkBuilder` and `CreateDllBuilder` have the
  `directory` parameter optional

### DIFF
--- a/Configuration.props
+++ b/Configuration.props
@@ -33,6 +33,7 @@
     <AndroidFrameworkVersion Condition=" '$(AndroidFrameworkVersion)' == '' And '$(_IsRunningNuGetRestore)' != 'True' ">$(AndroidLatestStableFrameworkVersion)</AndroidFrameworkVersion>
     <AndroidUseLatestPlatformSdk Condition=" '$(AndroidFrameworkVersion)' == '' And '$(_IsRunningNuGetRestore)' == 'True' ">True</AndroidUseLatestPlatformSdk>
     <DebugType Condition=" '$(DebugType)' == '' ">portable</DebugType>
+    <Deterministic Condition=" '$(Deterministic)' == '' ">True</Deterministic>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(HostOS)' == '' ">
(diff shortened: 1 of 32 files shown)


jonathanpeppers:
I suspect that AOT is somehow broken here in combination with `Deterministic=True`, will look into this.

Issue #3547, opened 2019-08-24
CI build under systemprofile account fails on signing


cail:
### Steps to Reproduce

1. build any android apk with signature key
2. for this use CI automation which runs under system account (think this is the key to the problem)
3. see signing failed because of _CreateAndroidDebugSigningKey task

### Expected Behavior

build success

### Actual Behavior

First, its not clear why build system does _CreateAndroidDebugSigningKey even if I build **Release** configuration.
Next, the path for the system account:  
```C:\Windows\system32\config\systemprofile\AppData\Local\Xamarin\Mono for Android```
does not exists.
Next, even if I create this path, the build still fails second time with
```Key pair not generated, alias <androiddebugkey> already exists``` because debug key already generated and keytool will not overwrite it.

### Version Information

VS2019 Community edition, Xamarin

### Log File

```
[00:16:42][Execute app] _CreateAndroidDebugSigningKey:
[00:16:42][Execute app]   C:\Program Files\Android\Jdk\microsoft_dist_openjdk_1.8.0.25\bin\keytool.exe -genkeypair -alias androiddebugkey -storepass android -keypass android -keystore "C:\Windows\system32\config\systemprofile\AppData\Local\Xamarin\Mono for Android\debug.keystore" -dname "CN=Android Debug,O=Android,C=US" -keyalg RSA -validity 10950 
[00:16:42][Execute app] C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets(3205,2): error ANDKT0000: keytool error: java.lang.Exception: Key pair not generated, alias <androiddebugkey> already exists [C:\BuildAgent\work\445be90eac994802\app-android\app-android.csproj]
```

**Comment by dellis1972** [comment]
I just looked at the code if '_ApkDebugKeyStore' is not set then we will try to create a key https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L3119.

To work around this issue you can provide your own `AppSettingsDirectory` which will specify where the key will be created , for example.

```
<AppSettingsDirectory>$(MSBuildThisFileDirectory)</AppSettingsDirectory>
```

Checkin or download a `debug.keystore` in that directory which will then be picked up by our build system.

**Comment by cail** [comment]
> Does your Release build contain keystore information?

Yes it does:
```
[...]

PR #3556, opened 2019-08-28
[Xamarin.Android.Build.Tasks] Dont create debug.keystore if we don't need to.

dellis1972:
Fixes https://github.com/xamarin/xamarin-android/issues/3547

When building in release mode if the user has `AndroidKeyStore` set
we should not try to create the debug.keystore. This is causing
problems on CI systems since sometimes the CI job is running
as a `system` account. Mostly around writting to the system
profile directory `C:\Windows\system32\config\systemprofile\AppData`.


dellis1972:
@jonathanpeppers we do https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L1731

Issue #3562, opened 2019-08-29
Fast deploy of wearable project has regressed between d16-3 Preview 2 and Preview 3


pjcollins:
Our fastdev support has regressed for wearable project types between d16-3 Preview 2 and Preview 3.

### Steps to Reproduce

1. Download and restore NuGets for: https://github.com/xamarin/monodroid-samples/tree/master/wear/DelayedConfirmation.
2. msbuild /t:Install DelayedConfirmationApp\DelayedConfirmationApp.csproj
3. msbuild /t:Install DelayedConfirmationApp\DelayedConfirmationApp.csproj

### Expected Behavior
Second deploy with no code changes should be fast.

### Actual Behavior
Second deploy with no code changes results in the .apk being rebuilt.

The issue appears to have been introduced somewhere in this diff: http://github.com/xamarin/xamarin-android/compare/21a650c170483a1ac99e1c4bf80d7ef3158ab281...489f2580373b0ddb2e1268a159e3eb935b2871a5

### Version Information
d16-3 Preview 3

### Log File
[Binary logs for good and bad incrementalinstall attempts](https://github.com/xamarin/xamarin-android/files/3556349/binlogs.zip)

brendanzagaeski:
*Release status update*

A new Preview version has now been published that includes the fix for this item. The fix is not yet included in a Release version. I will update this item again when a Release version is available that includes the fix.

The fix is not yet available on macOS. I will update this item again when a Preview version with the fix is available on macOS.

Fix included in Xamarin.Android 10.0.99.100.

Fix included on Windows in Visual Studio 2019 version 16.4 Preview 1. To try the Preview version that includes the fix, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) in [Visual Studio Preview](https://visualstudio.microsoft.com/vs/preview/).

Fix not yet available on macOS.

brendanzagaeski:
*Release status update*

A new Preview version has now been published for macOS that includes the fix for this item.

Fix included in Xamarin.Android 10.1.0.1.

Fix included on macOS in Visual Studio 2019 for Mac version 8.4 Preview 1. To try the Preview version that includes the fix, check for the latest updates on the **Preview** [updater channel](https://docs.microsoft.com/visualstudio/mac/update).

brendanzagaeski:
*Release status update*

[...]

PR #3569, opened 2019-08-30
[Xamarin.Android.Build.Tasks] fix Android Wear incremental builds

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/3562

Our QA tests caught a regression where building a Xamarin.Android app
that includes an embedded Android Wear app would always rebuild.

This brought up a few things:

1. We have a `XamarinAndroidWearApplicationProject` class, but no
   tests using it?
2. What does our MSBuild targets even do for Wear projects?

## Tests ##

In adding some tests, I had to fix a few other things:

* We should use the 1.0.0 Xamarin.Android.Wearable NuGet instead of
  `preview7`.
* `Xamarin.ProjectTools` was generating `csproj` files of an encoding
  that did not match:

```
   Xamarin.Android.Common.targets(423,2): error MSB4018: The "ParseAndroidWearProjectAndManifest" task failed unexpectedly.
        System.Xml.XmlException: There is no Unicode byte order mark. Cannot switch to Unicode.
        at System.Xml.XmlTextReaderImpl.Throw(Exception e)
        at System.Xml.XmlTextReaderImpl.ThrowWithoutLineInfo(String res)
        at System.Xml.XmlTextReaderImpl.CheckEncoding(String newEncodingName)
        at System.Xml.XmlTextReaderImpl.ParseXmlDeclaration(Boolean isTextDecl)
        at System.Xml.XmlTextReaderImpl.Read()
        at System.Xml.Linq.XDocument.Load(XmlReader reader, LoadOptions options)
        at System.Xml.Linq.XDocument.Load(String uri, LoadOptions options)
        at System.Xml.Linq.XDocument.Load(String uri)
        at Xamarin.Android.Tasks.ParseAndroidWearProjectAndManifest.Execute()
        at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
        at Microsoft.Build.BackEnd.TaskBuilder.<ExecuteInstantiatedTask>d__26.MoveNext() 
```

The file was actually `utf-8` but we were generating:

    <?xml version="1.0" encoding="utf-16"?>

I reworked `Xamarin.ProjectTools` to fix this.

I was able to add a simple test that just build a wear project, and
then added a second test that reproduces the problem.

## The Problem ##

Generally embedded Android Wear solutions:

* A main app references the Wear project.
* The main app grabs a few files from the Wear project: the APK and an
  xml configuration file and copies them into `obj\Debug\res`.

What was even weirder, since we put them in `@(AndroidResource)` the
`_GenerateAndroidResourceDir` MSBuild target would attempt to copy
these files to themselves. The same files paths were in
`@(_AndroidResolvedResources)` and `@(_AndroidResourceDest)`...

    <CopyIfChanged
        SourceFiles="@(_AndroidResolvedResources)"
        DestinationFiles="@(_AndroidResourceDest)"

The problem was we basically were copying these files and never adding
them to `@(_ModifiedResources)`. This meant that
`$(_AndroidResFlagFile)` was not `<Touch/>`'d in some cases:

    <Touch
        Files="$(_AndroidResFlagFile)"
        AlwaysCreate="True"
        Condition=" !Exists ('$(_AndroidResFlagFile)') Or '@(_ModifiedResources)' != '' Or '@(_AndroidResourceDestRemovedFiles)' != '' "
    />

Two fixes seemed to solve the issue:

* The `<PrepareWearApplicationFiles/>` task needs to return a
  `ModifiedFiles` output we can add to `@(_ModifiedResources)`.
* We should define the Wear files in `@(_AndroidResourceDest)` after
  the `<CopyIfChanged/>` call such as:

    <_AndroidResourceDest Include="@(_WearableApplicationDescriptionFile);@(_BundledWearApplicationApkResourceFile)" />

This way they are never part of `@(AndroidResource)`, since they are
auto-generated.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetXamarinProject.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetXamarinProject.cs
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Xml.Linq;
 using Microsoft.Build.Construction;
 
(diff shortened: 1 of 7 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetXamarinProject.cs
```diff
@@ -113,30 +114,33 @@ public virtual ProjectRootElement Construct ()
...
-						XNamespace ns = "http://schemas.microsoft.com/developer/msbuild/2003";
-						var e = XElement.Parse ($"<PackageReference Include=\"{pr.Id}\" Version=\"{pr.Version}\"/>");
-						e.Name = ns + e.Name.LocalName;
-						referenceGroup.Add (e);
+			XDocument document;
+			using (var stream = new MemoryStream ())
+			using (var sw = new StreamWriter (stream, encoding, 8 * 1024, leaveOpen: true)) {
+				root.Save (sw);
+				stream.Position = 0;
+				document = XDocument.Load (stream);
```
dellis1972: any reason why we could not use the XDeclaration.Encoding property? https://docs.microsoft.com/en-us/dotnet/api/system.xml.linq.xdeclaration.encoding?view=netframework-4.8

jonathanpeppers: I found a way to set it:
```csharp
document.Declaration.Encoding = encoding.HeaderName;
```
The value was already `utf-8`, but I think it's probably better to do this anyway.

jonathanpeppers:
Something is still going wrong on Mac here:
```
`_UpdateAndroidResgen` in wear build should be skipped!
Expected: True
But was: False
```
This was the new test -- looking into it.

jonathanpeppers:
Test is fixed now.

jonathanpeppers:
@pjcollins when you're back, can we re-run those QA tests against master? Or show me how 😄 ?

pjcollins:
@jonathanpeppers the majority of them should hopefully be integrated into our pipeline through https://github.com/xamarin/xamarin-android/pull/3595.

PR #3570, opened 2019-08-30
[tests] add [NonParallelizable] to ConvertCustomView

jonathanpeppers:
Context: http://build.azdo.io/2989026

We have seen test failures such as:

    Xamarin.Android.Build.Tests.IncrementalBuildTest.ConvertCustomView(True)
    Xamarin.Android.Build.Tests.IncrementalBuildTest.ConvertCustomView(False)

They both get a build error such as:

    /Library/Frameworks/Mono.framework/Versions/6.0.0/lib/mono/msbuild/Current/bin/NuGet.targets(121,5):
        error : The file '/Users/vsts/.nuget/packages/netstandard.library/2.0.3/.nupkg.metadata' already exists.

I think this a general bug when using `/t:Restore` or `/restore` with
an SDK-style project on Mac in parallel. I saw a similar issue when
writing the `MSBuildSdkExtrasTests`.

For now `[NonParallelizable]` should solve the issue.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
@@ -520,6 +520,7 @@ namespace Lib2
 		}
 
 		[Test]
+		[NonParallelizable] // /restore can fail on Mac in parallel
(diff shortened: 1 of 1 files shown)


dellis1972:
This is ok for now, but as we move towards more SDK style projects we need to come up with a workaround that will still allow tests to run in parallel. So sort of locking within the calls to `/restore` within the ProjectTools I suspect.

PR #3574, opened 2019-09-02
More C++ warnings cleanup + C++17 changes

grendello:
A small cleanup which deals with a handful of remaining warnings after enable
more extensive diagnostics in our compilers.

Deal with `read(2)` definition differences between Android+Unix compilers and
MinGW (on *nix the third argument, `count`, is defined to be of type `size_t`
while on MinGW it's `unsigned int` which caused an integer conversion warning
when building for Windows)

Take advantage of the C++17 `if constexpr ()` construct (works similar to the
preprocessor `#if{n}def` but is part of the language standard and is type-safe)
as well as the more compact (and less of an eyesore) namespace declaration,
instead of:

```cpp
namespace xamarin { namespace android { namespace internal
```
we can now use

```cpp
namespace xamarin::android::internal
```

C++17 allows one to declare and define certain static `constexpr` members
directly in the header file, without the need to add the definition of such
member in a source file somewhere. Take advantage of that.

### DIFF
--- a/src/monodroid/jni/android-system.cc
+++ b/src/monodroid/jni/android-system.cc
@@ -44,7 +44,6 @@ using namespace xamarin::android::internal;
 
 #if defined (DEBUG) || !defined (ANDROID)
 BundledProperty *AndroidSystem::bundled_properties = nullptr;
-constexpr char AndroidSystem::OVERRIDE_ENVIRONMENT_FILE_NAME[];
 #endif // DEBUG || !ANDROID
 
 #if defined (WINDOWS)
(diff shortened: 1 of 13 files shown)


src/monodroid/jni/android-system.cc
```diff
@@ -600,8 +607,8 @@ AndroidSystem::setup_environment_from_override_file (const char *path)
 	size_t   nread = 0;
 	ssize_t  r;
 	do {
-		size_t i = nread;
-		r = read (fd, buf + i, file_size - i);
+		read_count_type read_count = static_cast<read_count_type>(file_size - nread);
```
jonpryor: should we start using `auto` for instances like this?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -997,28 +997,30 @@ create_domain (JNIEnv *env, jclass runtimeClass, jstring_array_wrapper &runtimeA
...
+			// tell the IDE that the project likely need to be recompiled.
+			char* corlib_error_message = const_cast<char*>(mono_check_corlib_version ());
+			if (corlib_error_message == nullptr) {
+				if (!androidSystem.monodroid_get_system_property ("xamarin.studio.fakefaultycorliberrormessage", &corlib_error_message)) {
+					free (corlib_error_message);
+					corlib_error_message = nullptr;
+				}
+			}
+			if (corlib_error_message != nullptr) {
+				jclass ex_klass = env->FindClass ("mono/android/MonoRuntimeException");
```
jonpryor: If things are *really* FUBAR, this type won't exist.  Should we instead throw `java.lang.Error`, which should (!) *always* exist?

jonpryor: ....but that would be a semantic break from existing code.  Never mind.

PR #3577, opened 2019-09-03
[Xamarin.Android.Build.Tasks] Parse AAPT errors so every build error isn't APT0000

jpobst:
Error messages taken from https://gist.github.com/grendello/72c1ce5f6e8cdd3aefd393b983576d9a.

Created by grepping the `aapt` source for `error`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt.cs
@@ -412,7 +412,7 @@ namespace Xamarin.Android.Tasks
 					return;
 				}
 				if (level.Contains ("warning")) {
-					LogCodedWarning ("APT0000", singleLine);
+					LogCodedWarning (GetErrorCode (singleLine), singleLine);
 					return;
 				}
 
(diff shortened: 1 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/Aapt.cs
```diff
@@ -428,16 +428,177 @@ protected void LogEventsFromTextOutput (string singleLine, MessageImportance mes
...
+		{
+			foreach (var tuple in error_codes)
+				if (message.IndexOf (tuple.Item1, StringComparison.OrdinalIgnoreCase) >= 0)
+					return tuple.Item2;
+
+			return "APT0000";
+		}
+
+		static readonly List<Tuple<string, string>> error_codes = new List<Tuple<string, string>> () {
+			Tuple.Create ("AndroidManifest.xml is corrupt", "APT1100"),
```
jonpryor: For ease of human parsing, could the error number be *first* in the tuple, instead of second?  That would allow all of the error messages and codes to line up vertically and be easier (for me) to read:

```csharp
Tuple.Create ("APT1100", "AndroidManifest.xml is corrupt"),
Tuple.Create ("APT1001", "can't use '-u' with add"),
Tuple.Create ("APT1002", "dump failed because assets could not be loaded"),
// ...
```

jonpryor: Commit message/PR description should also provide where/how you got these strings.  Grepping the `aapt` source (with what `grep` command)?  Something else?

jpobst: Updated.

jonpryor:
Unit test expected output needs to be updated so that unit tests still pass.

Consider: [`Xamarin.Android.Build.Tests.ManifestTest.Bug12935(False)`](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3017500&view=ms.vss-test-web.build-test-results-tab&runId=8288478&resultId=100493&paneView=attachments):

```
String did not contain 'APT0000: '!
```

That's because the *new* output is:

```
 .../Bug12935_False/obj/Release/android/manifest/AndroidManifest.xml(6): error APT1134: String types not allowed (at 'screenOrientation' with value 'sensorPortait')
```

dellis1972:
ok another slight problem... currently we document all of our error codes [here](https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/messages/README.md). Do we need to auto generate a bunch of documents for these new error codes? I believe they use our docs to generate the docs on the VS site which the IDE links to if you want more help on an error.

PR #3590, opened 2019-09-05
[docs] Add some more information about profiling

radekdoulik:


Documentation/guides/profiling.md
```diff
@@ -38,12 +38,11 @@ You can also run `adb logcat -c` to clear the log at any point.
 
 Before launching the app, run the `adb` command:
 
-    > adb shell setprop debug.mono.log default,timing
+    > adb shell setprop debug.mono.log timing=bare
```
jonpryor: Won't `timing=bare` *prevent* creation of `counters.txt`?  `counters.txt` is mentioned and used later in the document.

AFAICT `timing=bare` *will* prevent creation of `counters.txt`.  I'm not sure we want to do that in this doc.

radekdoulik: We use a different setting later in the document. For runtime timing measurements it is better to use `timing=bare` as the JIT related processing affects the measurements.

jonathanpeppers: You might adjust the first sentence: https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/profiling.md#profiling-the-jit

> Before launching the app, you need the adb command as in Profiling the Runtime:

If they will be different, just change it to:

> Before launching the app, run the adb command:

Issue #3593, opened 2019-09-06
Add a test for apk size regression for CI builds


radekdoulik:
Recently I introduced a fix, which caused apk size regression. https://github.com/xamarin/xamarin-android/pull/3364

To avoid increases of apk sizes going unnoticed in future, I would like to enhance our apk tests by checking the apk tests sizes by comparing to previous known sizes, which we record for the size plots.

The test should have a threshold, say 100kB, and when the sizes differ more, the test run should fail.

PR #3594, opened 2019-09-06
[tests] Check apk tests sizes to avoid regressions

radekdoulik:
Implements: https://github.com/xamarin/xamarin-android/issues/3593

Added new `ProcessApkSizes` task based on `ProcessPlotInput`, which
takes additional argument with apk sizes reference file. The file is
used to check the current apk sizes and fails when the size difference
is over a given threshold, we use 100KiB now. There's one additional
threshold to warn about smalled differences, with current threshold
50KiB.

The reference files can be updated with

    > make update-apk-sizes-reference CONFIGURATION=<Debug or Release>

which copies the current apk sizes files to
`tests/apk-sizes-reference` directory, these are then used by
`ProcessApkSizes` task.

The apk sizes files were split by configuration and contain
`$(Configuration)` in the name. The reason for that is an easier
update of the references after failed PR build. (after the size
difference is checked to be valid and desired) This way it is possible
to just download the results artifacts file and copy the sizes
recorded by CI build into the reference directory.


radekdoulik:
build

PR #3595, opened 2019-09-06
[ci] Add integrated regression test jobs

pjcollins:
Context: https://devdiv.visualstudio.com/DevDiv/_release?_a=releases&view=mine&definitionId=476

Adds two new "integrated regression test" jobs to Azure Pipelines to run
additional external tests which were previously only ran and reviewed in
an ad-hoc manner.

Permission issues with PR builds from forks are currently preventing us
from running these jobs against such builds, so we'll only run against
manually triggered or regular CI builds instead.


pjcollins:
Permissions surrounding PR builds from forks will prevent us from running this automatically on every PR, however these changes will run on every CI build of master or our release branches. The latest test run to validate these changes has been triggered here: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3059408

PR #3600, opened 2019-09-06
[Xamarin.Android.BuildTasks] fix for _LinkAssembliesNoShrink

jonathanpeppers:
In 5320c25a, I implemented a way for `_LinkAssembliesNoShrink` to
build partially.

However, I found a bug when testing out the new feature on a project
with existing `bin` and `obj` directories:

* `build.props` was updated because `$(XamarinAndroidVersion)` changed
* `_LinkAssembliesNoShrink` ran, but many of the files were not
  actually copied, such as `FormsViewGroup.dll` from the Xamarin.Forms
  NuGet package.
* Any subsequent build would always run `_LinkAssembliesNoShrink`
  fully! Because `FormsViewGroup.dll` would eternally have an older
  timestamp than `build.props`.

I was able to reproduce the problem with a new test.

The fix is to set the timestamp on copied assemblies, even if they
weren't actually copied. This way the `Inputs` and `Outputs` of
`_LinkAssembliesNoShrink` will always be correct.

Building partially is a tricky operation... If applied in the future
to other targets, we'll have to specifically test scenarios like this!

### DIFF
--- a/Documentation/guides/MSBuildBestPractices.md
+++ b/Documentation/guides/MSBuildBestPractices.md
@@ -251,6 +251,11 @@ will re-run the target completely. `$(MSBuildAllProjects)` is a list
 of every MSBuild file imported during a build, MSBuild automatically
 evaluates `$(MSBuildAllProjects)` since [MSBuild 16.0][allprojects].
 
+One pitfall, is this `_GenerateDocumentation` example *must* touch the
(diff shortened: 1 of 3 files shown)


dellis1972:
This looks ok. I seem to remember ( I think)  we used to set the Timestamp on those files, we removed it at some point though to fix our incremental build. Funny how things work out lol :D

Issue #3086, opened 2019-05-09
macOS: build and package android-api-docs


jonpryor:
The Android API docs have been open-sourced!

  * https://devblogs.microsoft.com/xamarin/xamarin-api-docs-open-source-available-now/
  * https://github.com/xamarin/android-api-docs

Instead of building these in the monodroid repo for inclusion in the macOS `.pkg` installer, we can now move that into the OSS xamarin-android repo.

PR #3601, opened 2019-09-08
[build] Migrate assemble-docs target from monodroid

brendanzagaeski:
Fixes: https://github.com/xamarin/xamarin-android/issues/3086

Now that the [Xamarin.Android API docs][0] have been open sourced,
migrate the `assemble-docs` target and its dependencies from the
`docs.make` file in monodroid to a set of MSBuild targets in
xamarin-android.

Move `@(_FrameworkDirs)` and `@(_FrameworkDirsThatExist)` into their own
helper target so that the new targets can use them.

Add a trailing slash on the value of `$(FrameworkSrcDir)` so that the
new targets can follow the convention that the value of a directory path
property always includes a trailing slash.

Add the new targets that replace the old `docs.make` targets.  These
targets run `mdoc` to generate "msxdoc" Microsoft XML Documentation
format files that are used in the `.vsix` installer for Windows as well
as a monodoc format `MonoAndroid-lib.zip` archive and an accompanying
`.tree` file that are used in the `.pkg` installer for macOS.

Bring in android-api-docs as a submodule, set at the same commit as the
corresponding submodule in monodroid.

There are a few intentional differences compared to the original
`docs.make` targets:

  * The Microsoft XML Documentation files are output to the directory of
    the latest *stable* framework version rather than the latest *known*
    framework version.  This is useful because it means the MSBuild
    targets don't need access to the `$(FRAMEWORKS)` `make` variable.

  * The OpenTK-1.0 docs are only generated for the `mdoc assemble` step
    and not for the `mdoc export-msxdoc` step.  The `export-msxdoc`
    output isn't needed at the moment because the `.vsix` installer
    instead uses the `.xml` file that's generated when
    `Mono.Android.csproj` builds `OpenTK-1.0.csproj` for
    `$(TargetFrameworkVersion)` `v4.4`.

  * The OpenTK docs are omitted in favor of the OpenTK-1.0 docs because
    OpenTK has been deprecated and removed from Xamarin.Android.

  * The `pimpup-docs` target is omitted.  I checked the build logs for a
    recent commercial build and saw that while the build does compile
    `doc-relcontent-merger.exe` and `recipe-relcontent-generator.exe`,
    it never runs either one.  Starting in [monodroid@c83b93e2][1], the
    `$(PIMP_STAMP)` target that used to run `doc-relcontent-merger.exe`
    was no longer needed and was disabled.

  * The `docs/netdocs.zip` and `docs/mono-libs.zip` targets are also
    intentionally omitted.  Although commercial builds do still run
    those targets, the outputs are not used for anything.  For example,
    they aren't included in the `.pkg` installer for macOS or the
    `.vsix` installer for Windows.

    Based on the contents of the `push-docs` target in `docs.make` and
    the fact that that target was last edited in 2012, I believe the
    `netdocs.zip` and `mono-libs.zip` archives were for an old online
    documentation system that was retired long ago.

I verified that the `Mono.Android.xml` Microsoft XML Documentation file
generated by these new targets was identical to the version generated by
the `docs.make` targets.  Similarly, I verified that the new monodoc
`MonoAndroid-lib.zip` archive generated by these targets contained the
same list of `xml.summary.*` files as the `.zip` archive generated by
the `docs.make` targets.

[0]: https://github.com/xamarin/android-api-docs
[1]: https://github.com/xamarin/monodroid/commit/c83b93e20b0e6f80af7f8e19a6f1dc73f187e4f0

### DIFF
--- a/build-tools/installers/create-installers.targets
+++ b/build-tools/installers/create-installers.targets
@@ -6,7 +6,11 @@
   <UsingTask AssemblyFile="..\..\bin\Build$(Configuration)\xa-prep-tasks.dll" TaskName="Xamarin.Android.BuildTools.PrepTasks.ReplaceFileContents" />
   <PropertyGroup>
     <RootBuildDir>$(XamarinAndroidSourcePath)\bin\$(Configuration)\</RootBuildDir>
-    <FrameworkSrcDir>$(RootBuildDir)\lib\xamarin.android\xbuild-frameworks\MonoAndroid</FrameworkSrcDir>
+    <FrameworkSrcDir>$(RootBuildDir)lib\xamarin.android\xbuild-frameworks\MonoAndroid\</FrameworkSrcDir>
+    <AndroidApiDocsPath Condition=" '$(AndroidApiDocsPath)' == '' ">$(XamarinAndroidSourcePath)external\android-api-docs\</AndroidApiDocsPath>
+    <_AndroidApiDocsPath>$([MSBuild]::EnsureTrailingSlash($(AndroidApiDocsPath)))</_AndroidApiDocsPath>
+    <_LatestStableFrameworkDir>$(FrameworkSrcDir)$(AndroidLatestStableFrameworkVersion)\</_LatestStableFrameworkDir>
+    <_MonoDocOutputPath>$(RootBuildDir)lib\monodoc\</_MonoDocOutputPath>
     <MSBuildSrcDir>$(RootBuildDir)\lib\xamarin.android\xbuild\Xamarin\Android</MSBuildSrcDir>
     <MSBuildTargetsSrcDir>$(XamarinAndroidSourcePath)\src\Xamarin.Android.Build.Tasks\MSBuild\Xamarin\Android</MSBuildTargetsSrcDir>
     <DefaultRuntimeEntitlementsPath>$(MSBuildThisFileDirectory)\..\create-pkg\runtime-entitlements.plist</DefaultRuntimeEntitlementsPath>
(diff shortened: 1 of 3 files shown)


build-tools/installers/create-installers.targets
```diff
@@ -20,6 +24,60 @@
...
+  <Target Name="_CopyMonoAndroidDocsSource"
+      Inputs="$(_AndroidApiDocsPath)MonoAndroid-docs.source"
+      Outputs="$(_MonodocOutputPath)MonoAndroid-docs.source">
+    <Copy
+        SourceFiles="$(_AndroidApiDocsPath)MonoAndroid-docs.source"
+        DestinationFolder="$(_MonoDocOutputPath)" />
+  </Target>
+  <Target
+      Name="AssembleApiDocs"
+      DependsOnTargets="_GenerateMsxDocXmls;_GenerateMsxDocXmlRedirects;_GenerateMonoAndroidLibArchive;_CopyMonoAndroidDocsSource"
```
brendanzagaeski: One option I considered was to move the `_GenerateMsxDocXmls` and `_GenerateMsxDocXmlRedirects` targets into the `create-vsix.targets` file and the `_GenerateMonoAndroidLibArchive` and `_CopyMonoAndroidDocsSource` targets into the `create-pkg.targets` file. That could make it easy to see which targets are used for which installer. On the other hand, the current approach has the advantage that both invocations of `mdoc` are right next to each other. I'd be happy to split them up into the individual installer targets files if that seems better though.

jonpryor:
Please review `$(_HasCommercialFiles )` within [`build-tools/installers/create-installers.targets`](https://github.com/xamarin/xamarin-android/blob/bc1efe13d241fc742c01408ae84648bfac0cebb6/build-tools/installers/create-installers.targets).

For example, this `Condition` should be removed so that the OSS `.vsix` files *also* contain `Mono.Android.xml`:

https://github.com/xamarin/xamarin-android/blob/e27f818267e07d03383293253a4a6547d41a1ba4/build-tools/installers/create-installers.targets#L298

`MonoAndroid-lib.tree` and `MonoAndroid-lib.zip` are also currently listed as a "Commercial-Only" file, and that shouldn't need to be the case here.

Both OSS installers should contain classlib docs after this change.  Compare to the [Jenkins PR Build#2253](https://jenkins.mono-project.com/job/xamarin-android-pr-pipeline-release/2253/Azure/):

```
# Windows .vsix:
$ unzip -l ~/Downloads/Xamarin.Android.Sdk-OSS-10.0.99.104_pr-3601-merge_a731479.vsix | grep Mono.Android.xml
# no matches

# macOS .pkg
$ xar -xf ~/Downloads/Xamarin.Android.Sdk-OSS-10.0.99.104_pr-3601-merge_a731479.pkg  xamarin.android.pkg/Payload
$ gunzip -c xamarin.android.pkg/Payload | cpio -i -t | grep MonoAndroid-doc
./Library/Frameworks/Mono.framework/External/monodoc/MonoAndroid-docs.source
# No .tree or .zip file
```

brendanzagaeski:

Remove the `$(_HasCommercialFiles)` conditions from the Android API
documentation items because the files are now part of the open source
build.

I locally verified that the `Mono.Android.xml` file is now included in
the `.vsix` installer and that the `MonoAndroid-*` files are now
included in the `.pkg` installer for an open source build.  I also
compared the file lists from those installers against the previous PR
build and verified that there weren't any other unintended changes to
the installer contents.

brendanzagaeski:
**Changes in the latest commit**:

Remove the `$(HostOS)` `Condition` from `AssembleApiDocs` to enable
building the API docs on Windows.

Update the `<Exec/>` task invocations to work on both macOS and Windows
by using `$(ManagedRuntime)` to invoke the version of `mdoc.exe` that's
in the build tree.

Include an extra `\` at the end of `%(SourceDir)` in the `<Exec/>`
tasks.  Without this, `mdoc.exe` would hit an `Illegal characters in
path` exception in `System.IO.Path.LegacyNormalizePath()` on Windows
because [`CommandLineToArgvW()`][0] treats `\` as an escape character
when it precedes `"`.  Adding an additional `\` is a little easier than
using property functions to remove the original `\`.

`mdoc.exe` depends on `ICSharpCode.SharpZipLib.dll` for the `assemble`
subcommand, so add `ICSharpCode.SharpZipLib.dll` to the
`MonoUtilityFile` list in `xaprepare`.  This means that
`ICSharpCode.SharpZipLib.dll` will be added to the bundle during `make
prepare` on macOS and extracted alongside `mdoc.exe` when an existing
bundle is used on macOS or Windows.

Now that `ICSharpCode.SharpZipLib.dll` is part of the bundle, it is
available for inclusion in the open source installer packages, so move
the corresponding `<_MSBuildFiles>` item out from the
`$(_HasCommercialFiles)` condition.

Add a trailing slash on the value of `$(MSBuildSrcDir)` so that the new
targets can follow the convention that the value of a directory path
property always includes a trailing slash.  One reason for this
convention is that if the value of a path item does not have a trailing
slash, then the last part of the path is assigned to the `%(Filename)`
metadata rather than being part of the `%(Directory)` metadata.

Rename `SourceFolder` to `SourceDir` because "Dir" is more common than
"Folder" both in this file and in the built-in MSBuild tasks like
`<MakeDir/>` and `<RemoveDir/>`.

[0]: https://docs.microsoft.com/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw

Issue #3598, opened 2019-09-06
 Android App Bundle - does not automatically compress .WAV files


jgold6:
### Steps to Reproduce

1. Open the attached test project.
2. Run the app to device or simulator (tested with API 28 simulator)

[TestAABWav.zip](https://github.com/xamarin/xamarin-android/files/3585401/TestAABWav.zip)

### Expected Behavior
App will run and an audio file of a drum beat will play.

### Actual Behavior
App crashes with exception: `[MonoDroid] Java.IO.IOException: This file can not be opened as a file descriptor; it is probably compressed`

### Notes/Workaround
Add `.wav;` to the **Uncompressed Resources extensions** field in the Android Build project properties (VSforMac) or to the **Leave the following resource extensions uncompressed** field in the Android Options page (VS 2019 windows), then the file descriptor can successfully be opened and the WAV file played. 

During some research, supposedly the following file types are supposed to be uncompressed by default as per [this article](https://ponystyle.com/blog/2010/03/26/dealing-with-asset-compression-in-android-apps/):
/* these formats are already compressed, or don't compress well */
static const char* kNoCompressExt[] = {
    ".jpg", ".jpeg", ".png", ".gif",
    ".wav", ".mp2", ".mp3", ".ogg", ".aac",
    ".mpg", ".mpeg", ".mid", ".midi", ".smf", ".jet",
    ".rtttl", ".imy", ".xmf", ".mp4", ".m4a",
    ".m4v", ".3gp", ".3gpp", ".3g2", ".3gpp2",
    ".amr", ".awb", ".wma", ".wmv"
};

This would seem to be true when building an APK but not when building an AAB. 

### Version Information

<details>
<summary>
=== Visual Studio Enterprise 2019 for Mac ===
</summary>

Version 8.2.5 (build 5)
Installation UUID: f86726f2-bd5d-4610-867e-44e82f306ca2
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 5.14.0.85 (d16-2-fix-mac-updates / 86973979)

	Package version: 600000327

=== Mono Framework MDK ===

Runtime:
	Mono 6.0.0.327 (2019-02/f8ea05bddcb) (64-bit)
	Package version: 600000327

=== NuGet ===

Version: 5.1.0.6013

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	3.0.0-preview8-28405-07
	2.1.12
	2.1.11
[...]

PR #3604, opened 2019-09-09
[Xamarin.Android.Build.Tasks] .aab default uncompressed file extensions

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/3598

By default, Android leaves "already compressed" files such as
`*.wav` uncompressed in APKs.

This behavior is implemented in `aapt` & `aapt2`:

* `aapt`: https://github.com/aosp-mirror/platform_frameworks_base/blob/036ea39d96babb6c8a20464319903aea130e1a5f/tools/aapt/Package.cpp#L30-L37
* `aapt2`: https://github.com/aosp-mirror/platform_frameworks_base/blob/e80b45506501815061b079dcb10bf87443bd385d/tools/aapt2/cmd/Link.cpp#L2339-L2343

`bundletool` has a similar list of file extensions:

https://github.com/google/bundletool/blob/5ac94cb61e949f135c50f6ce52bbb5f00e8e959f/src/main/java/com/android/tools/build/bundletool/io/ApkSerializerHelper.java#L111-L115

However, if you specify `BundleConfig.json` (the `--config` switch),
`bundletool` does not use the `NO_COMPRESSION_EXTENSIONS`
list at all:

https://github.com/google/bundletool/blob/5ac94cb61e949f135c50f6ce52bbb5f00e8e959f/src/main/java/com/android/tools/build/bundletool/io/ApkSerializerHelper.java#L323-L330

    For bundle versions starting by 0.7.3 the no-compression is fully
    configured through the bundle config file.

Since .NET assemblies need to be uncompressed, we now need to manage
our own list of file extensions... I grabbed the list from
`bundletool`, since it will only be used in the `<BuildAppBundle/>`
MSBuild task.

I added some tests around this scenario.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
@@ -14,6 +14,46 @@ namespace Xamarin.Android.Tasks
 	/// </summary>
 	public class BuildAppBundle : BundleTool
 	{
+		static readonly string [] UncompressedByDefault = new [] {
+			// Xamarin.Android specific files
+			"typemap.mj",
+			"typemap.jm",
+			"assemblies/**",
+			// Android specific files, listed here:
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
I can rebase and bring this out of draft when this is merged: https://github.com/xamarin/xamarin-android/pull/3603

PR #3606, opened 2019-09-09
New bundled assembly mapping code

grendello:
`Xamarin.Android` stores uncompressed assemblies inside the APK archive in order
to be able to map them to the device's memory and hand the pointer to Mono,
without having to unpack the files from the APK. We've been using an embedded
copy of the `unzip` utility in order to access the archive. Not only the code
was old, it also did much more than was necessary for our purposes. This
incurred runtime overhead which this commit aims to fix by doing the following:

  * Remove the old C code and replace with a simple implementation tailored
    specifically to our needs

  * Delay mapping of assemblies, debug information files and config files until
    such time when Mono asks us to load a required assembly - this should
	improve application startup time.

  * Store information about the bundled assemblies in a dynamically allocated
    array without constantly reallocating it (and thus copying data) at the
	expense of a little memory wasted (up to 16k currently)

In addition to the above performance changes, the following C++ features are
added here:

  * Automatic memory management (via RAII) for memory allocated from the heap
    but used only within a single function (the `simple_pointer_guard` class)

  * Implement portions of the C++ standard header `type_traits` (which is
    shipped with the NDK but for some reason not found by `clang++`)

Performance impact with our Xamarin.Forms integration test, Android 10 on Pixel
3 XL. Measurements are in milliseconds, Xamarin.Android total init time only:

|  **Assembly preload** |  **master** | **this commit** | **Architecture** |
|-----------------------|------------ |-----------------|------------------|
| **enabled**           | 271.373     | 216.931         | armeabi-v7a      |
| **disabled**          | 238.404     | 181.080         | armeabi-v7a      |
| **enabled**           | 255.984     | 201.498         | arm64-v8a        |
| **disabled**          | 225.933     | 169.507         | arm64-v8a        |


jonpryor:
Let's try to back up some of the assertions in the commit message: *does* it help with startup time?  If so, by how much?

jonpryor:
This PR can also remove `src/monodroid.tpnitems`, and should similarly update the `ThirdPartyNotice.txt` generation code to no longer mention `minizip`, as we're no longer using it.

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -36,12 +37,39 @@ namespace xamarin { namespace android { namespace internal {
...
+		size_t      mmap_size;
+		void       *mmap_area;
+		void       *mmap_file_data;
 	};
-}}}
+
+	// Amount of RAM that is allowed to be left unused ("wasted") in the `bundled_assemblies`
+	// array after all the APK assemblies are loaded. This is to avoid unnecessary `realloc`
+	// calls in order to save time.
+	static constexpr size_t BUNDLED_ASSEMBLIES_MEMORY_EXCESS_LIMIT = 16384;
```
jonpryor: Units?  I assume bytes.

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -52,13 +80,78 @@ const char *EmbeddedAssemblies::suffixes[] = {
...
+
+		// We know the entry has one of the debug info extensions, so we can take this shortcut to
+		// avoid having to allocate memory and use string comparison in order to find a match for
+		// the `aname` assembly
+		if (strncmp (entry.name, aname, aname_len) != 0)
+			continue;
+
+		size_t ext_len = strlen (entry.name) - aname_len;
+		if (ext_len != 4 || // Assembly.pdb
+		    ext_len != 8) { // Assembly.{exe,dll}.mdb
```
jonpryor: This should be two tab stops from the previous `if`, not 4 spaces.

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -256,312 +350,580 @@ EmbeddedAssemblies::add_type_mapping (TypeMappingInfo **info, const char *source
...
+	}
 
-	if ((file = unzOpen2 (nullptr, &funcs)) != nullptr) {
-		do {
-			unz_file_info info;
-			uLong offset;
-			unsigned int *psize;
-			char cur_entry_name [256];
-			MonoBundledAssembly *cur;
+	simple_pointer_guard<uint8_t>  buf (new uint8_t[cd_size]);
```
jonpryor: My C++ is assuredly rusty, but...  Shouldn't this be `simple_pointer_guard<uint8_t[]>`?  Otherwise, I imagine that the destructor would use`std::is_array_v<uint8_t>`, which would be *false*, which would cause the `new[]` to correspond with `delete`, *not* `delete[]`.

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -174,18 +276,6 @@ EmbeddedAssemblies::typemap_managed_to_java (const char *managed)
 	return find_entry_in_type_map (managed, mj_typemap, mj_typemap_header);
 }
 
-MONO_API const char *
-monodroid_typemap_java_to_managed (const char *java)
```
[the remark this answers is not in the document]
jonpryor: If you're going to move them *again*, why move them now?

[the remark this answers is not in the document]
grendello: Because I dislike how the C wrappers are sprinkled all around the code in the files and while making major changes to this particular file I ran across these two and adding a handful of lines to the diff didn't seem like much trouble, for the sake of a cleaner diff later on at some point. Anyhow, reverted this change.

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -256,308 +358,569 @@ EmbeddedAssemblies::add_type_mapping (TypeMappingInfo **info, const char *source
...
+		return false;
+	}
 
-			size_t alloc_size = MULTIPLY_WITH_OVERFLOW_CHECK (size_t, sizeof(void*), bundled_assemblies_count + 1);
-			bundled_assemblies = reinterpret_cast<MonoBundledAssembly**> (utils.xrealloc (bundled_assemblies, alloc_size));
-			cur = bundled_assemblies [bundled_assemblies_count] = reinterpret_cast<MonoBundledAssembly*> (utils.xcalloc (1, sizeof (MonoBundledAssembly)));
-			++bundled_assemblies_count;
+	// We scan from the end to save time
+	bool found = false;
+	for (size_t i = alloc_size - (ZIP_EOCD_LEN + 2); i >= 0; i--) {
```
jonpryor: This loop *may* be why many of our unit tests are timing out, and is *also* a perfect example of how [unsigned integers are error prone](https://github.com/xamarin/xamarin-android/pull/3606/files#diff-fcb50a6533eb26323059bc5c8bda1dd7R687).

Consider!

```c
#include <stdio.h>

int main ()
{
	size_t len = 5;
	for (size_t i = len; i >= 0; i--) {
		printf ("len=%zu\n", i);
	}
	return 0;
}
```

This is (more or less) the same loop construct: an (unsigned) `size_t` iterator type, counting downward.

The result?  Infinite loop:

```
$  cc loop.c
$ ./a.out > o.txt
^C
$ head -8 o.txt
len=5
len=4
len=3
len=2
len=1
len=0
len=18446744073709551615
len=18446744073709551614
...
```

I don't know *for certain* that this loop is causing the hangs, but it certainly *could* be causing the hangs.

grendello: `i` should have been a `ssize_t`, good catch. This code, however, is for ZIP archives with comments, so unless the APK has one, it will never run.

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -52,13 +77,80 @@ const char *EmbeddedAssemblies::suffixes[] = {
...
+			continue;
+
+		// We know the entry has one of the debug info extensions, so we can take this shortcut to
+		// avoid having to allocate memory and use string comparison in order to find a match for
+		// the `aname` assembly
+		if (strncmp (entry.name, aname, aname_len) != 0)
+			continue;
+
+		size_t ext_len = strlen (entry.name) - aname_len;
+		if (ext_len != 4 || ext_len != 8) { // 'Assembly.pdb' || 'Assembly.{exe,dll}.mdb'
```
jonpryor: I believe that this logic is wrong.

I applied this patch:

```diff
diff --git a/src/monodroid/jni/embedded-assemblies.cc b/src/monodroid/jni/embedded-assemblies.cc
index 0577db22..3025efec 100644
--- a/src/monodroid/jni/embedded-assemblies.cc
+++ b/src/monodroid/jni/embedded-assemblies.cc
@@ -100,6 +100,7 @@ EmbeddedAssemblies::get_mmap_file_data (XamarinBundledAssembly& xba)
 inline void
 EmbeddedAssemblies::load_assembly_debug_info_from_bundles (const char *aname)
 {
+       log_warn (LOG_DEFAULT, "# jonp: load_assembly_debug_info_from_bundles: aname=%s", aname);
        if (!register_debug_symbols || !bundled_assemblies_have_debug_info || aname == nullptr)
                return;
 
@@ -112,11 +113,13 @@ EmbeddedAssemblies::load_assembly_debug_info_from_bundles (const char *aname)
                // We know the entry has one of the debug info extensions, so we can take this shortcut to
                // avoid having to allocate memory and use string comparison in order to find a match for
                // the `aname` assembly
+               log_warn (LOG_DEFAULT, "# jonp: load_assembly_debug_info_from_bundles: strncmp(%s, %s, %i)", entry.name, aname, (int) aname_len);
                if (strncmp (entry.name, aname, aname_len) != 0)
                        continue;
 
                size_t ext_len = strlen (entry.name) - aname_len;
                if (ext_len != 4 || ext_len != 8) { // 'Assembly.pdb' || 'Assembly.{exe,dll}.mdb'
+                       log_warn (LOG_DEFAULT, "# jonp: load_assembly_debug_info_from_bundles: extension mismatch");
                        continue;
                }
 
```

`adb logcat` output contains:

```
# jonp: load_assembly_debug_info_from_bundles: aname=UnnamedProject
# jonp: load_assembly_debug_info_from_bundles: strncmp(UnnamedProject.pdb, UnnamedProject, 14)
# jonp: load_assembly_debug_info_from_bundles: extension mismatch
# jonp: load_assembly_debug_info_from_bundles: strncmp(FormsViewGroup.pdb, UnnamedProject, 14)
# jonp: load_assembly_debug_info_from_bundles: strncmp(Xamarin.Forms.Core.pdb, UnnamedProject, 14)
# jonp: load_assembly_debug_info_from_bundles: strncmp(Xamarin.Forms.Platform.Android.pdb, UnnamedProject, 14)
# jonp: load_assembly_debug_info_from_bundles: strncmp(Xamarin.Forms.Xaml.pdb, UnnamedProject, 14)
# jonp: load_assembly_debug_info_from_bundles: strncmp(Mono.Android.pdb, UnnamedProject, 14)
```

In order for `mono_register_symfile_for_assembly()` to be called, we must *not* print `extension mismatch`, yet we do.

One #facepalm later, and it's because of the `||`!  `ext_len` is 4, but it *isn't* 8, and the result of `ext_len != 4 || ext_len != 8` is True, because `ext_len != 8`!

This should be `&&`!

grendello: Well spotted! The debug files are now registered properly, but the same tests still fail :(

grendello:
This PR is now rebased on #3660

PR #3492, opened 2019-08-13
[Xamarin.Android.Build.Tasks] improve MAX_PATH errors

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/2848
Context: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation
Context: https://blogs.msdn.microsoft.com/jeremykuhne/2016/07/30/net-4-6-2-and-long-paths-on-windows-10/

A #deletebinobj problem has surfaced related to `MAX_PATH`. We have a
discrepancy of path length between:

* `obj\Debug\android\src` - where we generate `.java` source code
* `obj\Debug\android\bin\classes` - `javac` compiles `.class` files

If you have a project on the edge, where `android\src` does *not* have
a long path, but `android\bin\classes` *does*, we get the issue here:

    error MSB3231: Unable to remove directory "obj\Debug\90\android\bin\classes".
        Could not find a part of the path 'AccessibilityManagerCompat_AccessibilityStateChangeListenerImplementor.class'.

`javac` was able to create this file that is over `MAX_PATH`, and
unfortunately, .NET is throwing a `DirectoryNotFoundException` here...

An example:

    var dir = @"C:\src\xamarin-android\bin\TestDebug\temp\OnTheEdgeNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNNN\obj\Debug\android\bin\classes";
    Directory.Delete(dir, recursive: true);

This code throws this when there is a file inside the directory over
`MAX_PATH`:

    System.IO.DirectoryNotFoundException: 'Could not find a part of the path 'AccessibilityManagerCompat_AccessibilityStateChangeListenerImplementor.class'.'

A solution, is to prefix `\\?\` in the `Directory.Delete` call as long
as you are already using a full path:

    Directory.Delete(@"\\?\" + dir, recursive: true);

As long as you are running on .NET 4.6.2 or higher, .NET will pass the
path directly to the Windows APIs without modification. This works
even without the `Local Group Policy` setting to enable long paths.

However, we should use `\\?\` sparingly:

* Not all `System.IO` and .NET APIs support `\\?\`-prefixed paths
* You have to already know that the path you are using is a full path
  -- `Path.GetFullPath()` will not expand a long path!

So I think we can safely prefix `\\?\` in some cases, such as:

* We hit a `DirectoryNotFoundException` or `PathTooLongException`
[...]

PR #3609, opened 2019-09-09
[Xamarin.Android.Build.Tasks] Ensure our tasks never leak an unhandled exception, which causes an MSB4018.

jpobst:
Errors and warning codes for errors and warnings reported within Visual Studio from developers participating in the Visual Studio Customer Experience Improvement Program are reported "telemetry" events, but only the error and warning codes are reported. The warning and error message contents are not sent.

Unfortunately all unhandled exceptions not caught by our MSBuild tasks are reported as `MSB4018`, which reduces the utility of telemetry, making it more difficult to better understand where our build system could be improved.

This PR creates new base tasks for us to inherit: `AndroidTask`, `AndroidAsyncTask`, and `AndroidToolTask`.  These tasks add catch-all exception handlers that capture and log unhandled exceptions instead of allowing them to leak out to MSBuild.

### Implementation

Each error begins with `XA` to denote it is from Xamarin Android so our codes do not collide with other components.

This is followed by a 3 character prefix given to each Task so we can see which Task is causing the error.  For example, the AOT task's prefix is `AOT` and the Linker is `LNK`.

Next is a `7` to denote these are unhandled exceptions.

Finally, common Exception types are encoded as integers to denote the type of Exception being thrown.  For example, `NullReferenceException` is `001`. (`000` denotes an Exception type we aren't specifically looking for.)

The exception table is here:
https://github.com/xamarin/xamarin-android/blob/b4e3c009a2f46ff38b4fba435efbd3fcbf60466b/src/Xamarin.Android.Build.Tasks/Utilities/UnhandledExceptionLogger.cs

Putting it all together:
* A `NullReferenceException` in `LinkAssemblies` would be `XALNK7001`
* A `StackOverflowException` in `AOT` would be `XAAOT7016`

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
@@ -17,6 +17,8 @@ namespace Xamarin.Android.Tasks {
 	
 	//aapt2 link -o resources.apk.bk --manifest Foo.xml --java . --custom-package com.infinitespace_studios.blankforms -R foo2 -v --auto-add-overlay
 	public class Aapt2Link : Aapt2 {
+		public override string TaskPrefix => "A2L";
+
 		[Required]
 		public ITaskItem [] ManifestFiles { get; set; }
 
@@ -75,7 +77,7 @@ namespace Xamarin.Android.Tasks {
 		AssemblyIdentityMap assemblyMap = new AssemblyIdentityMap ();
 		List<string> tempFiles = new List<string> ();
 
-		public override bool Execute ()
+		public override bool RunTask ()
 		{
 			Yield ();
 			try {
@@ -83,7 +85,7 @@ namespace Xamarin.Android.Tasks {
 
 				task.ContinueWith (Complete);
 
-				base.Execute ();
+				base.RunTask ();
 			} finally {
 				Reacquire ();
 			}
--- a/src/Xamarin.Android.Build.Tasks/Tasks/AndroidTask.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/AndroidTask.cs
@@ -1,0 +1,61 @@
+using System;
+using Microsoft.Build.Utilities;
+using Xamarin.Build;
+
+namespace Xamarin.Android.Tasks
+{
+	// We use this task to ensure that no unhandled exceptions
+	// escape our tasks which would cause an MSB4018
+	public abstract class AndroidTask : Task
+	{
+		public abstract string TaskPrefix { get; }
+
+		public override bool Execute ()
+		{
+			try {
+				return RunTask ();
+			} catch (Exception ex) {
+				Log.LogUnhandledException (TaskPrefix, ex);
+				return false;
+			}
+		}
+
+		public abstract bool RunTask ();
+	}
+
+	public abstract class AndroidAsyncTask : AsyncTask
+	{
+		public abstract string TaskPrefix { get; }
+
+		public override bool Execute ()
+		{
+			try {
+				return RunTask ();
+			} catch (Exception ex) {
+				Log.LogUnhandledException (TaskPrefix, ex);
+				return false;
+			}
+		}
+
+		public virtual bool RunTask () => base.Execute ();
+	}
+
+	public abstract class AndroidToolTask : ToolTask
+	{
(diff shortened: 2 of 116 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
```diff
@@ -75,15 +77,15 @@ public class Aapt2Link : Aapt2 {
...
 			try {
 				var task = this.RunTask (DoExecute);
 
 				task.ContinueWith (Complete);
 
-				base.Execute ();
+				base.RunTask ();
 			} finally {
 				Reacquire ();
 			}
```
jonathanpeppers: Since we have `AndroidAsyncTask` now, can we put this yield-try-finally-reacquire shenanigans in there, too?

I think the same pattern is in here a few times -- would be nice to override `RunTask` and that's it.

src/Xamarin.Android.Build.Tasks/Utilities/UnhandledExceptionLogger.cs
```diff
@@ -0,0 +1,61 @@
...
+			else if (ex is TimeoutException)
+				log.LogCodedError (prefix + "7017", ex.ToString ());
+			else if (ex is TypeInitializationException)
+				log.LogCodedError (prefix + "7018", ex.ToString ());
+			else if (ex is UnauthorizedAccessException)
+				log.LogCodedError (prefix + "7019", ex.ToString ());
+			else if (ex is ApplicationException)
+				log.LogCodedError (prefix + "7020", ex.ToString ());
+			else if (ex is KeyNotFoundException)
+				log.LogCodedError (prefix + "7021", ex.ToString ());
```
jonathanpeppers: This needs `PathTooLongException`, and probably `DirectoryNotFoundException`, too?

Related to: https://github.com/xamarin/xamarin-android/pull/3492

jonpryor:
**Xamarin.Android (Mac Build)** [failed to build](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3033664&view=logs&s=1afc3bfe-122c-538b-e9ad-2a86c2efcfef&j=96fd57f5-f69e-53c7-3d47-f67e6cf9b93e):

```
  Tasks/Adb.cs(33,29): error CS0115: 'Adb.DefaultErrorCode': no suitable method found to override [/Users/vsts/agent/2.155.1/work/1/s/external/monodroid/tools/msbuild/Xamarin.Android.Build.Debugging.Tasks.csproj]
  Tasks/Adb.cs(18,15): error CS0534: 'Adb' does not implement inherited abstract member 'AndroidToolTask.TaskPrefix.get' [/Users/vsts/agent/2.155.1/work/1/s/external/monodroid/tools/msbuild/Xamarin.Android.Build.Debugging.Tasks.csproj]
```

Looks like the `monodroid` repo needs updating?

src/Xamarin.Android.Build.Tasks/Utilities/UnhandledExceptionLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+{
+	static class UnhandledExceptionLogger
+	{
+		public static void LogUnhandledException (this TaskLoggingHelper log, string prefix, Exception ex)
+		{
+			prefix = "XA" + prefix;
+
+			// Some ordering is necessary here to ensure exceptions are before their base exceptions
+			if (ex is NullReferenceException)
+				log.LogCodedError (prefix + "7001", ex.ToString ());
```
dellis1972: not sure what our preference is here..

```
prefix + "XXXX"
```

or 

```
$"{prefix}XXXX"
```

@jonpryor ?

src/Xamarin.Android.Build.Tasks/Tasks/AndroidTask.cs
```diff
@@ -0,0 +1,61 @@
+using System;
+using Microsoft.Build.Utilities;
+using Xamarin.Build;
+
+namespace Xamarin.Android.Tasks
+{
+	// We use this task to ensure that no unhandled exceptions
+	// escape our tasks which would cause an MSB4018
+	public abstract class AndroidTask : Task
```
dellis1972: I'm torn between wanting each of these in their own file, and them all being in one place to make it easy to change...

PR #3611, opened 2019-09-10
Add Workspace for VSCode

dellis1972:
These changes let us open the root XA folder in VS Code. We can them build the Xamarin.Android.Build.Tasks and Tests, and then run them. Unit tests can then be run and debugged if the [nxunit-test-adapter](https://marketplace.visualstudio.com/items?itemName=wghats.vscode-nxunit-test-adapter) is installed.


jonpryor:
This should also add a [`.vscode/extensions.json`](https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions) file so that the `nxunit-test-adapter` package is recommended.

PR #3615, opened 2019-09-11
Removing HearingAid from enum-conversion file and manually adding it to map.csv

gugavaro:
We should prevent HearingAid to be automatically generated as enum item because it will change related fields from integer type to the enum type and this will cause breaks!
Instead we manually add the item to the enum, leavign all fields with its original type (integer)


src/Mono.Android/map.csv
```diff
@@ -64,6 +64,8 @@
 0,,,android/view/Surface.FX_SURFACE_MASK,98304
 0,,,android/view/Surface.FX_SURFACE_NORMAL,0
 
+// Manually added because cannot be auto generated.
```
jpobst: I would probably change this to something like `// Manually added to preserve backwards compatibility (leave the original const as int)` so we better understand why we're doing this.

gugavaro: Sure! fixed

PR #3617, opened 2019-09-11
Bump to bundletool 0.10.2

jonathanpeppers:
Context: https://github.com/google/bundletool/releases/tag/0.10.2
Changes: https://github.com/google/bundletool/compare/0.10.0...0.10.2

Bumping `bundletool` as I see some fixes that look important.

I then needed to fix the scenario where the `<DownloadUri/>` task was
skipping the download completely. By keeping the version in the
filename, it will force a download if `$(XABundleToolVersion)`
changes.

I also added `SkipUnchangedFiles=True` to speed up incremental builds
of `Xamarin.Android.sln`.


jonathanpeppers:
Only test failure seems unrelated:
```
TimeZone should have been Indian/Reunion
Expected: True
But was: False
```
`api-compatibility` also failed on Jenkins, which I think is known.

PR #3618, opened 2019-09-11
[Xamarin.Android.Build.Tasks] disambiguate $(MonoAndroidIntermediateAssetsDir)

jonathanpeppers:
Refactoring our MSBuild targets, I noticed two MSBuild properties:

* `MonoAndroidIntermediateAssetsDir`
* `MonoAndroidIntermediateAssemblyDir`

One appears to be favored for `Debug` builds and the other `Release`
builds. But wait, what are these set to?

Then I looked at their starting value:

    <MonoAndroidIntermediateAssetsDir>$(IntermediateOutputPath)android\assets\</MonoAndroidIntermediateAssetsDir>
    <MonoAndroidIntermediateAssemblyDir>$(IntermediateOutputPath)android\assets\</MonoAndroidIntermediateAssemblyDir>

I think we should only use `$(MonoAndroidIntermediateAssemblyDir)`
throughout our MSBuild targets, but keep
`$(MonoAndroidIntermediateAssetsDir)` since it is a public property. I
found other usage of `$(MonoAndroidIntermediateAssetsDir)` in other
Xamarin GitHub repositories.

This makes things more clear, I think. It also allowed me to remove a
duplicate `<MakeDir/>` call!

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -1189,8 +1189,9 @@ because xbuild doesn't support framework reference assemblies.
 <!-- Resource build properties -->
 <PropertyGroup>
 	<MonoAndroidResDirIntermediate>$(IntermediateOutputPath)res\</MonoAndroidResDirIntermediate>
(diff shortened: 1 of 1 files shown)


dellis1972:
I would have gone the other way with this as the `MonoAndroidIntermediateAssemblyDir` is called `assets`... but I get why you want to use `MonoAndroidIntermediateAssemblyDir` cos its where we put the assemblies.

PR #3623, opened 2019-09-12
[Xamarin.Android.Build.Tasks] property refactoring/cleanup

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/compare/master...jonathanpeppers:fully-incremental

I am working on a revamp of the `<GenerateJavaStubs/>` MSBuild task to
make it "fully incremental".

The following parts of the build would happen on a per-assembly basis:

* Java sources generated
* Java class files compiled via `javac`
* `classes.zip` or `*.jar` files

This will be hidden behind a feature flag such as
`/p:_AndroidFullyIncremental=True` for its first implementation.

The problem I am facing is that our build throws random Java sources
into `$(IntermediateOutputPath)android\src` throughout the build. This
includes files like `R.java`, `MonoRuntimeProvider.java`, etc.

To make it where the location of these files is modifiable, we can
create a few new private properties:

* `$(_AndroidIntermediateJavaSourceDirectory)` which is `$(IntermediateOutputPath)android\src\`
* `$(_AndroidIntermediateJavaClassDirectory)` which is `$(IntermediateOutputPath)android\bin\classes\`
* `$(_AndroidIntermediateClassesZip)` which is `$(IntermediateOutputPath)android\bin\classes.zip`

This way I can modify the location when the `_AndroidFullyIncremental`
feature flag is set.

To refactor things a bit further, I also removed usage of
`$(MonoAndroidIntermediate)` and just used `$(IntermediateOutputPath)`
instead. `$(IntermediateOutputPath)` should be generally more
well-known for MSBuild, and I kept the original property around
because I found usage in other repos.


jonathanpeppers:
It looks like there is a change needed on the monodroid side, let me look into that.

PR #3625, opened 2019-09-12
[xaprepare] Improve external git deps processing

pjcollins:
The `PrepareExternalGitDependencies` xaprepare step has been updated to
generate the required `MonoGitHash.props` file when processing the mono
entry in the `.external` file, rather than cloning or updating a full
checkout of the repository. We don't necessarily want or need to clone
and update every repo listed in the `.external` file by default.

The msbuild target `CheckoutExternalGitSources` has also been removed as
its functionality has been moved into `xaprepare`.


jonpryor:
```markdown
As part of the commercial build, `azure-pipelines.yaml` executes
[`make prepare-external-git-dependencies`][0] in order to obtain the
private `monodroid` repo, to create commercial builds.

An unintentional side effect of this is that we *also* clone the
`mono/mono` submodule, as it is also specified within `.external`:

	Updating external repositories
	  * monodroid
	    -> cloning from xamarin/monodroid
	...
        * mono
          -> cloning from mono/mono
	...
	Elapsed: 00:01:25.3891298

Cloning `mono/mono` should be unnecessary, and adds ~85 seconds to
the build.

Update the `xaprepare.exe -s:PrepareExternalGitDependencies` scenario
so that the `Step_PrepareExternalGitDependencies` step generates
`bin/Build$(Configuration)/MonoGitHash.props` and *skips* the mono/mono
"submodule" clone.

Additionally, remove the `msbuild /t:CheckoutExternalGitSources`
target, as its functionality has been moved into `xaprepare`.

[0]: https://github.com/xamarin/xamarin-android/blob/master/build-tools/automation/azure-pipelines.yaml#L121
```

PR #3631, opened 2019-09-13
[Xamarin.Android.Build.Tasks] async helpers in AndroidAsyncTask

jonathanpeppers:
Right now there is a lot of boilerplate when using `AsyncTask`:

    {
        Yield ();
        try {
            var task = this.RunTask (DoExecute);

            task.ContinueWith (Complete);

            base.RunTask ();
        } finally {
            Reacquire ();
        }

        return !Log.HasLoggedErrors;
    }

We don't want to remember all that... or have to consistently get it
right.

Then if you want to use `Parallel.ForEach` it gets even more complex:

    this.RunTask (
        () => this.ParallelForEach (SomeArray, DoExecute)
    ).ContinueWith (Complete);

In d4c8f077, we created a base `AndroidAsyncTask` for error handling.
Let's put this weird `Yield-try-finally-Reacquire` logic in there,
too!

So instead, you would override:

    public override System.Threading.Tasks.Task RunTaskAsync ()

This method is:

* Already on a background thread.
* You can use `async` / `await` nicely.

I also added helper methods, so that instead of:

    this.ParallelForEach (ManifestFiles, ProcessManifest);

You can instead do:

    await this.WhenAll (ManifestFiles, ProcessManifest);

This plays well with `async` / `await`.

In refactoring this, I found some places where `<ResolveAssemblies/>`
was accessing `Log.Log*` on a background thread... I fixed that.

### DIFF
--- a/Documentation/guides/MSBuildBestPractices.md
+++ b/Documentation/guides/MSBuildBestPractices.md
@@ -448,6 +448,68 @@ specific to Xamarin.Android, and this will prevent them from
 conflicting with MSBuild properties from other products. All MSBuild
 properties are effectively "global variables"...
 
+## Xamarin.Android MSBuild Task base classes
+
+We have a few base classes to simplify error handling, `async` /
+`await` usage, etc.
+
+`AndroidTask` is a plain `Task`, override `RunTask` and use it as you
+would `Task.Execute()`:
+
+```csharp
+public class MyTask : AndroidTask
+{
+    public override bool RunTask ()
+    {
+        // Implementation
+        return !Log.HasLoggedErrors;
+    }
+}
+```
+
+The benefit here is that if an unhandled exception is thrown, `MyTask`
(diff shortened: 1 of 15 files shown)


Documentation/guides/MSBuildBestPractices.md
```diff
@@ -448,6 +448,68 @@ specific to Xamarin.Android, and this will prevent them from
...
+## Xamarin.Android MSBuild Task base classes
+
+We have a few base classes to simplify error handling, `async` /
+`await` usage, etc.
+
+`AndroidTask` is a plain `Task`, override `RunTask` and use it as you
+would `Task.Execute()`:
+
+```csharp
+public class MyTask : AndroidTask
```
jonpryor: This also needs to override `AndroidTask.TaskPrefix`, as it is `abstract`.

PR #3643, opened 2019-09-17
[ci] Add Windows MSBuild test job

pjcollins:
We'd like to run most if not all of our tests against a complete .vsix
or .pkg installer in order to better replicate our users development
environments. All of our 'Test' jobs on Azure Pipelines are currently
configured to test against our installers, however these jobs are only
running on macOS environments.

Adds one new jobs to run the msbuild tests which don't require an
emulator on a Windows test environment against our .vsix installer.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -197,7 +197,7 @@ stages:
       parameters:
         artifactName: mac-build-results
 
-  # Check - "Xamarin.Android (Mac Queue Vsix Signing)"
+  # Check - "Xamarin.Android (Build Queue Vsix Signing)"
   # Actually runs on a Windows host, but the work is done in a Jenkins job. Does not run on PR builds.
   - job: queue_vsix_signing
     displayName: Queue Vsix Signing
@@ -224,14 +224,10 @@ stages:
           GITHUB_CONTEXT=$(GitHub.Artifacts.Context)
           ENABLE_JAR_SIGNING=true
 
-# This stage ensures Windows specific build steps continue to work, and runs unit tests.
-# Check - "Xamarin.Android (Windows Build and Test)"
-- stage: win_build_test
-  displayName: Windows
-  dependsOn: prepare
-  jobs:
+  # This stage ensures Windows specific build steps continue to work, and runs a limited set of unit tests.
+  # Check - "Xamarin.Android (Build Windows)"
   - job: win_build_test
-    displayName: Build and Test
+    displayName: Windows
     pool: $(VSEngWinVS2019)
     timeoutInMinutes: 360
     cancelTimeoutInMinutes: 5
@@ -285,14 +281,6 @@ stages:
         arguments: Xamarin.Android-Tests.sln /p:Configuration=$(XA.Build.Configuration) /p:XAIntegratedTests=False /bl:$(System.DefaultWorkingDirectory)\bin\Test$(XA.Build.Configuration)\msbuild-build-tests.binlog $(AndroidTargetAbiArgs)
 
     - task: MSBuild@1
-      displayName: nunit Xamarin.Android.Build.Tests
-      inputs:
-        solution: Xamarin.Android.sln
-        configuration: $(XA.Build.Configuration)
-        msbuildArguments: /t:RunNUnitTests /bl:$(System.DefaultWorkingDirectory)\bin\Test$(XA.Build.Configuration)\msbuild-run-nunit-tests.binlog
-      timeoutInMinutes: 240
-
-    - task: MSBuild@1
       displayName: nunit Java.Interop Tests
       inputs:
         solution: Xamarin.Android.sln
@@ -314,6 +302,8 @@ stages:
     - template: yaml-templates\upload-results.yaml
       parameters:
         artifactName: win-build-test-results
+
+  # TODO Check - "Xamarin.Android (Build Linux)"
 
 - stage: test
   displayName: Test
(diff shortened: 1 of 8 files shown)


jonpryor:
Keeping the current **Windows** > **Build and Test** is probably a good idea, because any "Test the `.vsix` on Windows" job will be testing the *macOS-built* `.vsix` on Windows -- because that's what we actually ship, and thus that's the most meaningful thing -- which in no way means that xamarin-android *built on **Windows*** works and passes unit tests.

I thus think **Windows** > **Build and Test** should be left unchanged.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -571,7 +561,56 @@ stages:
       parameters:
         artifactName: mac-msbuild-test-results
 
- # Check - "Xamarin.Android (Test MSBuild With Emulator Mac)"
+  # Check - "Xamarin.Android (Test MSBuild Win)"
+  - job: win_msbuild_tests
+    displayName: MSBuild Win
```
jonpryor: Because of [this](https://github.com/xamarin/xamarin-android/pull/3643#issuecomment-532689933), should this `displayName` make it obvious that this is testing *macOS*-built packages on Windows, and not Windows-built packages on Windows?

Perhaps we want: `MSBuild Win (macOS .vsix)`?  Or perhaps we don't need to worry that much?

pjcollins: Maybe instead of having jobs/checks named `Test $foo Mac` and `Test $foo Win` we could just specify the installer extension, which also makes it clear the commercial artifacts are being tested. For example `Test MSBuild With Emulator - .vsix`?

I agree with keeping the Windows stage around, and in doing so we can perhaps safely assume that every job in the `Test` stage is running against macOS produced installers.

pjcollins: Putting the installer extension in a the job looked a bit funny to me, I went with `Test $foo - Windows` and `Test $foo - macOS` for now.

pjcollins:
Emulator testing on hosted Windows machines is proving to be a bit tricky, I've removed that component for now. We can revisit it in a future PR.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -552,16 +552,16 @@ stages:
       parameters:
         artifactName: mac-bcl-test-results
 
-  # Check - "Xamarin.Android (Test MSBuild Mac)"
+  # Check - "Xamarin.Android (Test MSBuild - macOS)"
```
jonathanpeppers: The only thing I see to bring up is what the checks look like:

image

Since the end of the names get chopped off, is there a way to drop the name `Xamarin.Android` on the front? Is it just based on the `displayName`?

pjcollins: I think we'd need to submit a feature request of sorts to the Azure Pipelines GitHub App team for this, unfortunately I'm not aware of a way to change the display names here. They seem to be constructed automatically based on pipeline name, stage name, and job name, and I haven't found any way to override that. Things are a _bit_ better on the Azure Pipelines website itself, and in the Checks summary on the main PR page though:

Screen Shot 2019-09-23 at 4 49 34 PM

Screen Shot 2019-09-23 at 4 51 04 PM

pjcollins:
@dellis1972 @jonathanpeppers The new Window stage is now green, and I've made a few more test related changes since you've reviewed if you wouldn't mind taking another quick look.

The one test failure on the `Test MSBuild - macOS` stage looks to be another potential intermittent (perhaps parallel execution related) failure is likely not related to these changes.

jonpryor:
Suggested **Squash and merge** commit message:

```
We'd like to run most if not all of our tests against a complete
`.vsix` or `.pkg` installer in order to better replicate our users
development environments.  All of our 'Test' jobs on Azure Pipelines
are currently configured to test against our installers, however
these jobs are only running on macOS environments.

We'll start to extend our test coverage on Windows by copying the
MSBuild test steps which are in the 'Win Build and Test' stage into
a new **Test** > **MSBuild - Windows** job.

Additionally, perform a bit of refactoring to
`Builder.FrameworkLibDirectory` and `Builder.AndroidMSBuildDirectory`
to support testing against local build trees and system installations
on both macOS and Windows.
```

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs
```diff
@@ -97,22 +87,28 @@ public class Builder : IDisposable
...
+			get {
+				var msbuildDir = Path.Combine (BuildOutputDirectory, "lib", "xamarin.android", "xbuild", "Xamarin", "Android");
+				if (Directory.Exists (msbuildDir) && File.Exists (Path.Combine (msbuildDir, "lib", "armeabi-v7a", "libmono-android.release.so")))
+					return msbuildDir;
+
+				return TestEnvironment.MonoAndroidToolsDirectory;
+			}
+		}
+
+		public string FrameworkLibDirectory {
```
jonathanpeppers: Should we put a `///<summary>` comment on each of these properties to explain what path they point to? I will always be confused looking at these.

Something like:
```
FrameworkLibDirectory is the home of v1.0`, v10.0, etc., the path is:
Build tree - ./bin/Debug/lib/xamarin.android/...
madOS - /Library/Frameworks/Xamarin.Android.framework/...
Windows - C:\Program Files (x86)\Microsoft Visual Studio\...
```
Same for the other two properties here.

pjcollins: 👍

jonpryor:
Mono aborted during the **Prepare** stage.

Trying again: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3085263&view=results

jonpryor:
Latest & Greatest build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3088370&view=results

PR #3645, opened 2019-09-18
[aprofutil] Add the aprofutil to the installation

radekdoulik:
Let the xaprep know about the `aprofutil` and `Mono.Profiler.Log`
assembly.

Add documentation about *Profiling the AOT*.

Add script to call the aprofutil on MacOS.


radekdoulik:
the azp test succeeded and then failed with: ##[error]Artifact mac-apk-test-results already exists for build 3062620. (https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3062620&view=logs - see attempt 2)

thus it looks OK to merge

radekdoulik:
@dellis1972 That's a good idea.

I plan to discuss that with IDE people to see whether they prefer to handle that through our targets or doing it as part of IDE (the socket iface to aot profiler is pretty easy to use).

The motivation for current cmd line approach is to let people try it even before the IDE support is finished. There the new target would indeed help I think.

I plan to write a blog post about recording AOT profile from command line and I am going to try it with SmartHotel360 app, so I might prepare the new target during that.

PR #3648, opened 2019-09-18
[tests] fix DebuggingTest hang on failure

jonathanpeppers:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3062430&view=logs&s=859b8d9a-8fd6-5a5c-6f5e-f84f1990894e&j=4348ab7c-72f0-52b0-44fd-6f17a16a558f

We had a test hang for the MSBuild device tests.

Reviewing the `while` loop in `ApplicationRunsWithDebuggerAndBreaks`:

    TimeSpan timeout = TimeSpan.FromSeconds (60);
    while (session.IsConnected && breakcountHitCount < 3) {
        Thread.Sleep (10);
        timeout = timeout.Subtract (TimeSpan.FromMilliseconds (10));
    }

This would be an infinite loop if the test failed. I think it was
intended to check:

    while (session.IsConnected && breakcountHitCount < 4 && timeout >= TimeSpan.Zero) {
        ...

### DIFF
--- a/tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
+++ b/tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
@@ -179,14 +179,14 @@ namespace Xamarin.Android.Build.Tests
 				// we need to wait here for a while to allow the breakpoints to hit
 				// but we need to timeout
 				TimeSpan timeout = TimeSpan.FromSeconds (60);
-				while (session.IsConnected && breakcountHitCount < 3) {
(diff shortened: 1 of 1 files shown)


dellis1972:
This looks ok, one failure is a BCL test the other is 
```
1) Failed : Xamarin.Android.Build.Tests.ResolveSdksTaskTests.ResolveSdkTiming
    Task should not take more than 2 seconds to run.
    Expected: less than or equal to 00:00:02
    But was:  00:00:02.5955699
```

I think we can live with that since its not related.

Issue #1580, opened 2018-04-19
Xamarin Android app build fails on Windows 10 with GetAdditionalResourcesFromAssemblies  throwing an exception (GenerateBuildSessionId in case of iOS)


srikcgaa2:
### Steps to Reproduce

1.   On a Windows 10 machine, ensure the group policy "System cryptography: Use FIPS compliant algorithms for encryption, hashing, and signing" is set to "Enabled"
2.   Create a new Xamarin Android application in Visual Studio 2015 Professional Update 3.  Sample project attached with this issue.
3.   Build the project 

### Expected Behavior
Build should be successful

### Actual Behavior
Build fails due to a crash with GetAdditionalResourcesFromAssemblies   step throwing an InvalidOperationException.   If we create a new Xamarin iOS Application instead of Android, the failure occurs from GenerateBuildSessionId step

### Version Information

Please see the attached file for Visual Studio version information as well as Windows 10 system information along with the Security Group Policy screenshot.  Pasted below as well for quick reference

Microsoft Visual Studio Professional 2015
Version 14.0.25431.01 Update 3
Microsoft .NET Framework
Version 4.7.02558

Installed Version: Professional

Visual Basic 2015   00322-50052-91952-AA572
Microsoft Visual Basic 2015

Visual C# 2015   00322-50052-91952-AA572
Microsoft Visual C# 2015

Visual C++ 2015   00322-50052-91952-AA572
Microsoft Visual C++ 2015

Application Insights Tools for Visual Studio Package   7.0.20622.1
Application Insights Tools for Visual Studio

ASP.NET and Web Tools 2015.1 (Beta8)   14.1.11107.0
ASP.NET and Web Tools 2015.1 (Beta8)

ASP.NET Web Frameworks and Tools 2012.2   4.1.41102.0
For additional information, visit http://go.microsoft.com/fwlink/?LinkID=309563

ASP.NET Web Frameworks and Tools 2013   5.2.40314.0
For additional information, visit http://www.asp.net/

Common Azure Tools   1.8
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

ILSpy.AddIn   1.0
Integration of the ILSpy Decompiler into Visual Studio.

JavaScript Language Service   2.0
JavaScript Language Service

JavaScript Project System   2.0
JavaScript Project System

JetBrains ReSharper Ultimate 2016.1.2    Build 105.0.20160522.80219
[...]

PR #3649, opened 2019-09-18
[Xamarin.Android.Build.Tasks] use CRC64 over SHA1 and MD5

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/issues/1580

Bump to xamarin/java.interop/master@397013ed

Changes: https://github.com/xamarin/java.interop/compare/be6048ed...397013ed

Use of MD5 results in an inability to use Xamarin.Android on
FIPS-enabled machines, because when the **System cryptography:
[Use FIPS compliant algorithms][0] for encryption,
hashing, and signing** setting is enabled on Windows:

!['Use Fips compliant algorithms' setting][1]

then the `System.Security.Cryptography.MD5.Create()` method will
throw `InvalidOperationException`.

In order to support FIPS-enabled machines, use a 64-bit
[Cyclic redundancy check][2] algorithm (CRC64) to instead generate
the Java package names, e.g. `crc64aae7d955a89b38db.Name` and
`crc64217c8705f00a5054.Name`.  A simple implementation using a
lookup table will suffice for our purposes, ported to C# from:

    https://github.com/gityf/crc/blob/8045f50ba6e4193d4ee5d2539025fef26e613c9f/crc/crc64.c

The use of CRC64 avoids the use of MD5, thus permitting execution on
FIPS-enabled Windows machines, and also results in *shorter*
directory names -- as the Java package name is a directory name --
which helps reduce Windows `MAX_PATH` issues.

## Changes so far ##

All usage of `MD5.Create()` or `new SHA1Managed()` have been either:

* Switched to use `new Crc64()`
* Used `Files.HashString` if more appropriate

Calls to `jarsigner` now use:

    -sigalg SHA256withRSA -digestalg SHA-256

The previous default can be restored via
`$(AndroidApkSigningAlgorithm)` and `$(AndroidApkDigestAlgorithm)`
MSBuild properties.

[0]: https://docs.microsoft.com/en-us/windows/security/threat-protection/security-policy-settings/system-cryptography-use-fips-compliant-algorithms-for-encryption-hashing-and-signing
[1]: https://user-images.githubusercontent.com/24926263/38981221-7e0bf7c4-43dc-11e8-8f16-9b6d284a55fb.PNG
[2]: https://en.wikipedia.org/wiki/Cyclic_redundancy_check


src/Xamarin.Android.Build.Tasks/Tasks/GetAdditionalResourcesFromAssemblies.cs
```diff
@@ -203,14 +199,14 @@ bool IsValidDownload(string file, string sha1)
 			if (string.IsNullOrEmpty (sha1))
 				return true;
 
-			var hashFile = file + ".sha1";
+			var hashFile = file + ".hash";
```
dellis1972: This isn't going to work. Take a look at var sha1sum = attributeValue.NamedArguments.FirstOrDefault (p => p.Name == "Sha1sum");.

the `Sha1sum` value is embedded into the attribute in the assembly. unless we alter all the old components/Nuget Package to use the new hash algorithm this is going to fail.

dellis1972:
We are doing to have to invalidate builds for this upgrade, we somehow need to delete any old files still using the MD5 hash (especially generated Java stuff).

tests/MSBuildDeviceIntegration/Tests/InstallTests.cs
```diff
@@ -323,5 +323,35 @@ public void BlankAdbTarget ()
...
+			if (!HasDevices)
+				Assert.Ignore ("Skipping Test. No devices available.");
+
+			var proj = new XamarinAndroidApplicationProject ();
+			proj.SetProperty ("AndroidPackageNamingPolicy", packageNamingPolicy);
+			proj.SetProperty (KnownProperties.AndroidSupportedAbis, "armeabi-v7a;x86");
+			using (var b = CreateApkBuilder ()) {
+				Assert.IsTrue (b.Install (proj), "Install should have succeeded.");
+				ClearAdbLogcat ();
+				AdbStartActivity ($"{proj.PackageName}/{activityPackageName}.MainActivity");
```
jonpryor: This is why we should put `[Register]` on `MainActivity`, so that the name doesn't change based on `$(AndroidPackageNamingPolicy)`.

jonathanpeppers:
Still working on how to fix the designer:
```
[2019-09-20 16:25:40.7] Renderer >> INFO: Trying to load class md5ac80b49d1943cc97166c48330708421e.AvatarView
```
This would be a crc64 hash now.

src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -239,6 +239,11 @@ internal static unsafe void Initialize (JnienvInitializeArgs* args)
 					Java.Lang.Thread.DefaultUncaughtExceptionHandler = defaultUncaughtExceptionHandler;
 			}
 
+			var packageNamingPolicy = Environment.GetEnvironmentVariable ("__XA_PACKAGE_NAMING_POLICY__");
+			if (Enum.TryParse (packageNamingPolicy, out PackageNamingPolicy pnp)) {
+				JavaNativeTypeManager.PackageNamingPolicy = pnp;
+			}
```
jonathanpeppers: @garuma if the designer can get the value of the `$(AndroidPackageNamingPolicy)` MSBuild property and pass it as an environment variable named `__XA_PACKAGE_NAMING_POLICY__`, I think that will get things working properly.

jonathanpeppers:
This is ready to review (I got all the tests to pass), but we still will need to wait on the designer changes before we can merge.

There is only one BCL test failure with the GC that seems unrelated.

jonpryor:
@garuma: Please let us know when we can merge this PR.  I assume that would be once https://github.com/xamarin/designer/pull/2394/files is merged.

src/Mono.Android/Android.Runtime/JNIEnv.cs
jonathanpeppers: Thanks @garuma we're green!

PR #3654, opened 2019-09-19
[Xamarin.Android.Build.Tasks] deprecate <GetAdditionalResourcesFromAssemblies/>

jonathanpeppers:
The `<GetAdditionalResourcesFromAssemblies/>` MSBuild task is the
precursor to Xamarin.Build.Download. It is only used by very old
versions of the Xamarin support library components.

We want to deprecate its behavior in an effort to remove complexity
and improve build performance. We have to scan each assembly for these
assembly-level attributes.

I added a new `XA0121` warning if a library is encountered during a
build that uses these attributes:

* `IncludeAndroidResourcesFromAttribute`
* `NativeLibraryReferenceAttribute`
* `JavaLibraryReferenceAttribute`

I added `[Obsolete]` on these as well.

### DIFF
--- a/src/Mono.Android/Android/IncludeAndroidResourcesFromAttribute.cs
+++ b/src/Mono.Android/Android/IncludeAndroidResourcesFromAttribute.cs
@@ -3,6 +3,7 @@ using System;
 namespace Android {
 
 	[AttributeUsage (AttributeTargets.Assembly, AllowMultiple = true)]
+	[Obsolete ("This attribute is deprecated and will be removed in a future release.")]
 	public class IncludeAndroidResourcesFromAttribute : ReferenceFilesAttribute
 	{
 		public IncludeAndroidResourcesFromAttribute (string path)
--- a/Documentation/guides/messages/xa0121.md
+++ b/Documentation/guides/messages/xa0121.md
@@ -1,0 +1,34 @@
+---
+title: Xamarin.Android warning XA0121
+description: XA0121 warning code
+ms.date: 09/19/2019
+---
+# Xamarin.Android warning XA0121
+
+## Issue
+
+The behavior implemented in the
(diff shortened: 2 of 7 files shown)


src/Mono.Android/Android/IncludeAndroidResourcesFromAttribute.cs
```diff
@@ -3,6 +3,7 @@
 namespace Android {
 
 	[AttributeUsage (AttributeTargets.Assembly, AllowMultiple = true)]
+	[Obsolete ("This attribute is deprecated and will be removed in a future release.")]
```
jonpryor: Should this include `, error: true` so that it cannot be referenced from C# code?  (I assume it can still be access via Reflection/MSBuild, but it would prevent C#-side references and more strongly signify that it's going away.)

jonpryor: ...and ditto all the other `[Obsolete]`s in this PR.

dellis1972: Would library creations be able to ignore this error if they get it? Generally Deprecation should not result in build breakage, at least not right away.

jonathanpeppers: Doing some web searching, it appears there is not a way to get around the error at all if `IsError=True`:

https://stackoverflow.com/a/10608567/132442

They suggest using System.Reflection to workaround, which might not be possible with custom attributes. You would have to use Mono.Cecil or something.

jonathanpeppers: Discussed on Slack: we won't add `IsError=True` until we actually remove `<GetAdditional*/>`.

PR #3656, opened 2019-09-20
[Xamarin.Android.Build.Tasks] Clean did not delete App Bundles

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/3635

Running a `Clean`, would leave the app bundle in the `bin` directory:

    > msbuild Foo.Android.csproj /p:Configuration=Release /p:AndroidPackageFormat=aab /t:SignAndroidPackage
    > msbuild Foo.Android.csproj /p:Configuration=Release /p:AndroidPackageFormat=aab /t:Clean
    > ls -n bin\Release
    com.mycompany.foo-Signed.aab

The `_CollectMonoAndroidOutputs` MSBuild target has a series of
`<CreateItem/>` calls that add apk-related files to `@(FileWrites)`.

We need to do the same for all app-bundle-related files.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -524,12 +524,13 @@ namespace UnamedProject
 		}
 
 		[Test]
-		public void BuildApplicationAndClean ([Values (false, true)] bool isRelease)
+		public void BuildApplicationAndClean ([Values (false, true)] bool isRelease, [Values ("apk", "aab")] string packageFormat)
 		{
 			var proj = new XamarinFormsAndroidApplicationProject {
 				IsRelease = isRelease,
 			};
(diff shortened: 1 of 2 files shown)


dellis1972:
@jonathanpeppers this is weird. We have a test to make sure a clean actually does a clean. I wonder why this test didn't fail?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -524,12 +524,13 @@ public void CheckTimestamps ([Values (true, false)] bool isRelease)
 		}
 
 		[Test]
-		public void BuildApplicationAndClean ([Values (false, true)] bool isRelease)
+		public void BuildApplicationAndClean ([Values (false, true)] bool isRelease, [Values ("apk", "aab")] string packageFormat)
```
dellis1972: Ah, its because we only build apks... right got it.

PR #3660, opened 2019-09-20
New ZIP routines for embedded assembly handling

grendello:
Context: https://github.com/xamarin/xamarin-android/pull/3606
    
Xamarin.Android stores uncompressed assemblies inside the APK archive in order
to be able to map them to the device's memory and hand the pointer to Mono,
without having to unpack the files from the APK. We've been using an embedded
copy of the `unzip` utility in order to access the archive. Not only the code
was old, it also did much more than was necessary for our purposes.
    
PR #3606 implemented both the ZIP handling portion as well as had changes to
delay mmapping of the files embedded in APK till the moment they were actually
needed by the Mono runtime. However, some of the tests using the debugger would
either hang or fail to work properly with #3606 and so we decided to split it up
into two parts - zip handling and mmap code. This commit implements the former.
    
ZIP handling code added here makes a number of assumptions about the state of
the archive:
    
  * APK is a valid ZIP archive
    Rationale: Android verifies APK/AAB archives and if they weren't valid our
    code wouldn't be executed at all.
  * APK is not an empty ZIP archive
    Rationale: if it were empty, we would have a rather hard time running our code from it
  * APK does not use ZIP-64 extensions
    Rationale: for the moment Google use ZIP archives without the 64-bit extensions and 
    there's no sign of that changing. If ever the change is made, we can easily adjust our code.
    
Additionally, this commit adds a simple C++ guard class useful in managing
dynamically allocated memory within a function, ensuring that it is
automatically freed on function exit.
    
Performance impact with our Xamarin.Forms integration test, Android 10 on Pixel
3 XL. Measurements are in milliseconds. Despite omitting the lazy mmaping code,
this commit is nearly as fast as #3606, however the cost of mmaping is not zero
and it will scale up with the number of assemblies in the APK.
    
Figures represent Xamarin.Android total init time only:
    
|  **Assembly preload** |  **master** | **this commit** | **Architecture** |
|-----------------------|------------ |-----------------|------------------|
| **enabled**           | 271.373     | 215.785         | armeabi-v7a      |
| **disabled**          | 238.404     | 182.055         | armeabi-v7a      |
| **enabled**           | 255.984     | 197.586         | arm64-v8a        |
| **disabled**          | 225.933     | 169.131         | arm64-v8a        |


src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,443 @@
...
+
+bool
+EmbeddedAssemblies::zip_adjust_data_offset (int fd, size_t local_header_offset, uint32_t &data_start_offset)
+{
+	static constexpr size_t LH_FILE_NAME_LENGTH_OFFSET   = 26;
+	static constexpr size_t LH_EXTRA_LENGTH_OFFSET       = 28;
+
+	off_t result = ::lseek (fd, static_cast<off_t>(local_header_offset), SEEK_SET);
+	if (result < 0) {
+		log_fatal (LOG_DEFAULT, "Failed to seek to archive entry local header at offset %u", local_header_offset);
```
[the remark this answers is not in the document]
grendello: Yeah, I guess it should. I'll update.

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,443 @@
...
+	return true;
+}
+
+bool
+EmbeddedAssemblies::zip_read_field (uint8_t* buf, size_t buf_len, size_t index, size_t count, char*& characters)
+{
+	if (!zip_ensure_valid_params (buf, buf_len, index, count))
+		return false;
+
+	characters = new char[count + 1];
```
jonpryor: Should this use `ADD_WITH_OVERFLOW_CHECK()` in case `index` is some "unreasonable" value so that we don't overflow?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,443 @@
...
+
+	index = buf_offset + CD_LOCAL_HEADER_POS_OFFSET;
+	if (!zip_read_field (buf, buf_len, index, local_header_offset)) {
+		log_fatal (LOG_DEFAULT, "Failed to read Central Directory entry 'relative offset of local header' field");
+		return false;
+	}
+	index += sizeof(local_header_offset);
+
+	if (file_name_length == 0) {
+		file_name = new char[1];
```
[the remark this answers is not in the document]
jonpryor: Or should we just `assert()` that this doesn't happen?  What's it *mean* to encounter a filename with no length?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,443 @@
...
+	return true;
+}
+
+bool
+EmbeddedAssemblies::zip_ensure_valid_params (uint8_t* buf, size_t buf_len, size_t index, size_t to_read)
+{
+	assert (buf != nullptr);
+	if (index + to_read > buf_len) {
+		log_fatal (LOG_DEFAULT, "Buffer too short to read %u bytes of data", to_read);
+		return false;
```
jonpryor: I feel like this should be a "must pass" method, and if this check fails, the process should be exited.  As such, `zip_ensure_valid_params()` could return `void`: it either works, or the process dies.

If done, this should _not_ be done via assert, as it's too easy to disable asserts (are they even enabled on our release builds?), but should instead exit with a `FATAL_EXIT_*` value.

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,443 @@
...
+	static constexpr size_t CD_LOCAL_HEADER_POS_OFFSET   = 42;
+	static constexpr size_t CD_COMMENT_LENGTH_OFFSET     = 32;
+
+	size_t index = buf_offset;
+	if (!zip_ensure_valid_params (buf, buf_len, index, ZIP_CENTRAL_LEN))
+		return false;
+
+	uint8_t signature[4];
+	if (!zip_read_field (buf, buf_len, index, signature)) {
+		log_fatal (LOG_DEFAULT, "Failed to read Central Directory entry signature");
```
jonpryor: Ditto all of these.  If *any* `log_fatal()` call is hit because something went wrong, we should just *die immediately*, not...muddle through at runtime with all of these return value checks.

src/monodroid/jni/embedded-assemblies-zip.cc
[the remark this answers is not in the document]
grendello: We pass `file_name` to a number of other functions which may or may not accept a `nullptr`. Using an empty string for a name in a very unlikely situation like here makes for simpler and shorter code later on - instead of checking for `nullptr` we simply pass the name to other functions and let them do their thing.

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: Making every `log_fatal` exit immediately without returning a boolean would require passing extra parameters to some of those functions (apk name, central directory index for instance - they aren't relevant for e.g. `zip_read_field`) which have nothing to do with what the functions do. I will modify the code to use `log_error` instead of `log_fatal` in the "inner", return a boolean and do a "final" `log_fatal` + `exit` instead.

src/monodroid/jni/embedded-assemblies-zip.cc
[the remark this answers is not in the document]
grendello: An empty file name is valid in ZIP - it means the data came from [standard input](https://github.com/nih-at/libzip/blob/master/docs/appnote.txt#L874)

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: I initially though it was a good idea, but then I realized that `zip_read_field` is a "generic" method to read data and if it fails with a fatal error message + exit, then all we'd see in the logcat would be a generic error message with no useful information to act on (e.g. what field wasn't read, from which part of ZIP was it read from etc) and while it could be solved by passing extra parameters to `zip_read_field` to print a more specific error message, that would require us using variable arguments and those would create a problem with the `log_*` functions which have no overload accepting `va_list` (and we do use multiple parameters when printing error messages here). All in all, I don't think it's worth it and that it would make for much better code.

src/monodroid/jni/android-system.cc
```diff
@@ -266,7 +265,7 @@ AndroidSystem::monodroid_read_file_into_memory (const char *path, char **value)
 			r = ADD_WITH_OVERFLOW_CHECK (size_t, static_cast<size_t>(fileStat.st_size), 1);
 			if (value && (*value = new char[r])) {
 				size_t nread = fread (*value, 1, static_cast<size_t>(fileStat.st_size), fp);
-				if (nread == 0 || nread != r) {
+				if (nread == 0 || nread != r - 1) {
```
jonpryor: Instead of using `r-1`, which could slowly "drift" over time if any additional lines are added between the `r=ADD_WITH_OVERFLOW_CHECK()` and here, why not instead do `nread != fileStat.st_size`?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,449 @@
...
+	uint16_t              compression_method;
+	uint32_t              local_header_offset;
+	uint32_t              data_offset;
+	uint32_t              file_size;
+	char                 *entry_name;
+	char                 *file_name;
+
+	ssize_t nread = read (fd, buf.get (), cd_size);
+	if (static_cast<size_t>(nread) != cd_size) {
+		log_fatal (LOG_DEFAULT, "Failed to read Central Directory from the APK archive %s. %s (nread: %d; errno: %d)", apk_name, std::strerror (errno), nread, errno);
```
jonpryor: Given that all of these are exiting with `FATAL_EXIT_NO_ASSEMBLIES`, would it make more sense for all of the `log_*()` invocations to use `LOG_ASSEMBLY`?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,449 @@
...
+
+		bool entry_is_overridden = !should_register (strrchr (file_name, '/') + 1);
+
+		if ((utils.ends_with (file_name, ".mdb") || utils.ends_with (file_name, ".pdb")) &&
+				register_debug_symbols &&
+				!entry_is_overridden &&
+				bundled_assemblies != nullptr) {
+			md_mmap_info map_info = md_mmap_apk_file(fd, data_offset, file_size, file_name, apk_name);
+			if (register_debug_symbols_for_assembly (file_name, (bundled_assemblies) [bundled_assemblies_count - 1],
+						(const mono_byte*)map_info.area,
```
jonpryor: Indentation here doesn't match context; this line should be *un*-indented one tab stop, to align with the following line.  Or the following line should be indented a tab stop to align with this line.

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,449 @@
...
+				        static_cast<int>(file_size)))
+				continue;
+		}
+
+		if (utils.ends_with (file_name, ".config") && bundled_assemblies != nullptr) {
+			char *assembly_name = strdup (basename (file_name));
```
jonpryor: Should `basename_r()` be used instead?  Or do this ourselves by using `strcchr(file_name, '/')`?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,449 @@
...
+
+		if (!(utils.ends_with (file_name, ".dll") || utils.ends_with (file_name, ".exe")))
+			continue;
+
+		if (entry_is_overridden)
+			continue;
+
+		size_t alloc_size = MULTIPLY_WITH_OVERFLOW_CHECK (size_t, sizeof(void*), bundled_assemblies_count + 1);
+		bundled_assemblies = reinterpret_cast<MonoBundledAssembly**> (utils.xrealloc (bundled_assemblies, alloc_size));
+		MonoBundledAssembly *cur = bundled_assemblies [bundled_assemblies_count] = reinterpret_cast<MonoBundledAssembly*> (utils.xcalloc (1, sizeof (MonoBundledAssembly)));
```
jonpryor: Why `xcalloc()` and not `new`?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,449 @@
...
+		++bundled_assemblies_count;
+
+		md_mmap_info map_info = md_mmap_apk_file (fd, data_offset, file_size, file_name, apk_name);
+		cur->name = utils.monodroid_strdup_printf ("%s", strstr (file_name, prefix) + prefix_len);
+		log_warn (LOG_DEFAULT, "cur->name == %s", cur->name);
```
jonpryor: Should this be surrounded by some `#if 0`-fu or something?  I doubt that we want this *always* printed for *every* app...

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -0,0 +1,449 @@
...
+		return false;
+	}
+
+	if (memcmp (signature, ZIP_EOCD_MAGIC, sizeof(signature)) == 0) {
+		return zip_extract_cd_info (eocd, ZIP_EOCD_LEN, cd_offset, cd_size, cd_entries);
+	}
+
+	// Most probably a ZIP with comment
+	size_t alloc_size = 65535 + ZIP_EOCD_LEN; // 64k is the biggest comment size allowed
+	ret = lseek (fd, static_cast<off_t>(-alloc_size), SEEK_END);
```
jonpryor: Should this be `::lseek()`, as done elsewhere?  Or should elsewhere use `lseek()`?  If `::` *shouldn't* be used, the `::read()` calls should also be fixed up.

src/monodroid/jni/embedded-assemblies-zip.cc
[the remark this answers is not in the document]
jonpryor: That's good to know.  TIL!

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: `basename_r` is deprecated, `basename` should be used instead.

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: 1. because of `xrealloc` above (`realloc`-like in C++ is trickier - you have to take into account constructors, make sure the type is POD etc - I *DO* plan to modify this code but in a separate PR. I deliberately left this code as it is in `master` to avoid introducing, for some reason, the same debugging problem PR #3606 has)

grendello:
It appears Azure had some connectivity issues

Issue #3662, opened 2019-09-23
VS2019 16.3.0 Preview 4.0 Compiler warnings about no debug symbols when converting app to use androidX.


gmck:
When building a release build I get the following bunch of warnings. The apk runs fine.
 

Task "Exec"
Task Parameter:Command="C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\mono-symbolicate" store-symbols "bin\Release\com.companyname.navxapp.apk.mSYM" "obj\Release\90\android/assets"
"C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\mono-symbolicate" store-symbols "bin\Release\com.companyname.navxapp.apk.mSYM" "obj\Release\90\android/assets"
Directory obj\Release\90\android/assets contains Java.Interop.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains mscorlib.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains System.Core.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains System.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains System.Runtime.Serialization.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains Xamarin.AndroidX.Annotation.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains Xamarin.AndroidX.AppCompat.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains Xamarin.AndroidX.Arch.Core.Common.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains Xamarin.AndroidX.Arch.Core.Runtime.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains Xamarin.AndroidX.AsyncLayoutInflater.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains Xamarin.AndroidX.Browser.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains Xamarin.AndroidX.CardView.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains Xamarin.AndroidX.Collection.dll but no debug symbols file was found.
Directory obj\Release\90\android/assets contains Xamarin.AndroidX.ConstraintLayout.dll but no debug symbols file was found.
[...]

PR #3624, opened 2019-09-12
[installer] Include symbols and add whitelist files

pjcollins:
Context: https://github.com/xamarin/monodroid/pull/941
Context: https://github.com/xamarin/xamarin-android/issues/3662

Adds support for symbol archive creation and publishing to the "inverted"
xamarin-android -> monodroid build system by including all required
symbol files in our installers. Adds two new whitelist files for assemblies
that don't currently have symbols, or have symbols which are incompatible
with the archiving process.

The [VS insertion release](https://devdiv.visualstudio.com/DevDiv/_releaseProgress?_a=release-environment-logs&releaseId=425862&environmentId=2054847) has also been updated to upload and archive
copies of all relevant executable and symbol files directly from the `.vsix`
installer, rather than using a separate `symbols.zip` file.

Publishing symbol files for the executable files we ship will make it
easier to investigate any potential hangs/crashes reported through
telemetry and other monitoring services.

PR #3671, opened 2019-09-24
[installer] Include DebugSymbols in the installer packages.

dellis1972:
Context #3662

We currently do not include any sybmols for our
Framework assemblies. As a result when users try to
use `mono-symbolicate` they get allot of warnings such
as

	Directory obj\Release\90\android/assets contains System.dll but no debug symbols file was found.

this can easily be fixed by including sybmol files in our
installers.


pjcollins:
After speaking with @dellis1972 on Slack, I've attempted to include this diff in https://github.com/xamarin/xamarin-android/pull/3624. I think we can continue to track this work over there.

dellis1972:
Closing since https://github.com/xamarin/xamarin-android/pull/3624 was merged

PR #3673, opened 2019-09-24
[tests] DeviceTest captures & records screenshots

jonathanpeppers:
Context: https://dev.azure.com/DevDiv/DevDiv/_build/results?buildId=3075928&view=ms.vss-test-web.test-result-details

I was originally using this in #3649, but it seemed better to do this
in a separate PR.

If a `DeviceTest` fails, it can take a screenshot, add an attachment
to the failed test, and archive it as an artifact. This will help
troubleshoot if things like a button press fails, and if we add more
tests like this down the road.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
@@ -1,23 +1,33 @@
 using NUnit.Framework;
+using NUnit.Framework.Interfaces;
 using System;
-using System.Collections;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
-using System.Security.Cryptography;
-using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
-using System.Xml;
 using System.Xml.Linq;
 using System.Xml.XPath;
-using Xamarin.ProjectTools;
 
 namespace Xamarin.Android.Build.Tests
 {
 	public class DeviceTest: BaseTest
 	{
+		[TearDown]
+		protected override void CleanupTest ()
+		{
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
```diff
@@ -1,23 +1,33 @@
...
 
 namespace Xamarin.Android.Build.Tests
 {
 	public class DeviceTest: BaseTest
 	{
+		[TearDown]
+		protected override void CleanupTest ()
+		{
+			if (HasDevices && TestContext.CurrentContext.Result.Outcome.Status == TestStatus.Failed &&
+				TestOutputDirectories.TryGetValue (TestContext.CurrentContext.Test.ID, out string outputDir)) {
```
dellis1972: formatting seems a bit off, I think these needs to be indented one more tab.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
```diff
@@ -1,23 +1,33 @@
...
 {
 	public class DeviceTest: BaseTest
 	{
+		[TearDown]
+		protected override void CleanupTest ()
+		{
+			if (HasDevices && TestContext.CurrentContext.Result.Outcome.Status == TestStatus.Failed &&
+					TestOutputDirectories.TryGetValue (TestContext.CurrentContext.Test.ID, out string outputDir)) {
+				string local = Path.Combine (outputDir, "screenshot.png");
+				string remote = "/sdcard/screenshot.png";
```
jonpryor: I'm not certain on the "reliability" of `/sdcard`; iirc the path of the SD Card mount point has changed over time, and have little reason to doubt that it won't change in the future.

Should this instead use `/data/local/tmp`?  We've been using that directory for *years* as part of Fast Deployment, so it *hasn't* changed.

PR #3674, opened 2019-09-24
[Xamarin.Android.Build.Tasks] MSBuild target does not properly build incrementally

### LINKED ISSUE #2413: [_ManagedUpdateAndroidResgen] MSBuild target does not properly build incrementally
### Steps to Reproduce

1. Run a design-time build.
2. Future design-time builds seems to always run `_ManagedUpdateAndroidResgen`.

Seems be caused due to a few issues, such as:
```
Building target "_ManagedUpdateAndroidResgen" completely.
Input file "obj\Debug\designtime\resourcepaths.cache" does not exist.
```

It also seems one of the `Inputs` is picked up as `build.props` instead of `designtime\build.props`. It looks like this is happening because `$(_ManagedUpdateAndroidResgenInputs)` is set at evaluation time, and a target changes the MSBuild property from `build.props` to `designtime\build.props`.

### What should we do?

We should refactor and add tests to verify the `_ManagedUpdateAndroidResgen` target builds incrementally.

It will be a bit tricky to get right--which is why I haven't done it yet.

dellis1972:
Fixes #2413

We should calculate the values for `_ManagedUpdateAndroidResgenInputs`
in a target rather than at evaluation time. This is so
we can pick up the correct `_AndroidBuildPropertiesCache`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -1227,23 +1227,28 @@ because xbuild doesn't support framework reference assemblies.
 	</ItemGroup>
 </Target>
 
-<PropertyGroup>
-	<_ManagedUpdateAndroidResgenInputs>
-		$(MSBuildAllProjects);
-		@(AndroidResource);
-		@(AndroidBoundLayout);
-		@(_MonoAndroidReferencePath);
-		@(_LibraryResourceDirectoryStamps);
-		$(_AndroidBuildPropertiesCache);
-		$(ProjectAssetsFile);
-	</_ManagedUpdateAndroidResgenInputs>
-</PropertyGroup>
+<Target Name="_BeforeManagedUpdateAndroidResgen">
+  <PropertyGroup>
+    <_ManagedUpdateAndroidResgenInputs>
+      $(MSBuildAllProjects);
+      @(AndroidResource);
+      @(AndroidBoundLayout);
+      @(_MonoAndroidReferencePath);
+      @(_LibraryResourceDirectoryStamps);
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1223,24 +1223,29 @@ because xbuild doesn't support framework reference assemblies.
...
-</PropertyGroup>
+<Target Name="_BeforeManagedUpdateAndroidResgen">
+  <PropertyGroup>
+    <_ManagedUpdateAndroidResgenInputs>
+      $(MSBuildAllProjects);
+      @(AndroidResource);
+      @(AndroidBoundLayout);
+      @(_ReferencePath);
+      @(_LibraryResourceDirectoryStamps);
+      @(_AdditonalAndroidResourceCachePaths->'%(Identity)\cache.stamp');
```
jonathanpeppers: This might conflict with: https://github.com/xamarin/xamarin-android/pull/3908

You can probably go ahead and remove this line:
```suggestion
```

dellis1972:
Last Failure was 

```
1) Failed : Xamarin.Android.Build.Tests.PerformanceTest.Build_JLO_Change
Exceeded expected time of 8150ms, actual 8200ms
  at Xamarin.Android.Build.Tests.PerformanceTest.Profile (Xamarin.ProjectTools.ProjectBuilder builder, System.Action`1[T] action, System.String caller) [0x00000] in <9ce1f7fb36544fb4b1f305a1c5a353e0>:0 
  at Xamarin.Android.Build.Tests.PerformanceTest.Build_JLO_Change () [0x00000] in <9ce1f7fb36544fb4b1f305a1c5a353e0>:0 
```

My guess is this is good to go.

Issue #3360, opened 2019-07-17
Analysis output not useful without option to not minimize Android resources


SamuelDebruyn:
I've [reported](https://github.com/xamarin/xamarin-android/issues?utf8=%E2%9C%93&q=is%3Aissue+author%3ASamuelDebruyn+is%3Aclosed+lint) a few issues before so that I can use the linter with my builds and they were all fixed, thank you!

Now that I have some meaningful output in my build log, it's still hard to find the mentioned elements in my XML because the files in obj/config/res are minimized to a single line.

**Feature request:**

- Either add an MSBuild property to disable the minification
- Either don't minimze the files when the linter is enabled

### Version Information

```
=== Visual Studio Enterprise 2019 for Mac (Preview) ===

Version 8.2 Preview (8.2 build 1076)
Installation UUID: e1324239-09f8-4ef0-841d-818a176fdfbd
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 5.14.0.83 (d16-2 / f08d4e3e)

	Package version: 600000296

=== Mono Framework MDK ===

Runtime:
	Mono 6.0.0.296 (2019-02/1e9f3bfbc5a) (64-bit)
	Package version: 600000296

=== NuGet ===

Version: 5.1.0.6013

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.2.5
	2.2.0
	2.1.9
	2.1.2
	2.1.1
	2.1.0
	2.0.7
	2.0.5
SDK: /usr/local/share/dotnet/sdk/2.2.300/Sdks
SDK Versions:
	2.2.300
	2.2.100
	2.1.505
	2.1.302
	2.1.301
	2.1.300
	2.1.200
	2.1.4
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/6.0.0/lib/mono/msbuild/Current/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.10
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Updater ===

Version: 11

=== Apple Developer Tools ===

Xcode 11.0 (14883.7)
Build 11M362v

=== Xamarin.Android ===

Version: 9.3.0.23 (Visual Studio Enterprise)
Commit: HEAD/d0b48056f
Android SDK: /Users/sam/Library/Android/sdk
	Supported Android versions:
		4.4 (API level 19)
[...]

PR #3688, opened 2019-09-25
Implement lazy mmapping of embedded assemblies

grendello:
Context: https://github.com/xamarin/xamarin-android/pull/3606
Context: https://github.com/xamarin/xamarin-android/pull/3660
Context: https://github.com/xamarin/xamarin-android/commit/3634b0f7dce6cfef64eb662a00c2cd321a4f3e5e

This commit implements lazy `mmap()`ing of assemblies, debug symbols and config
files found in the APK. Entries are mapped on-demand, when the Mono runtime asks
us to load an assembly.

This is the second part of the split mentioned in PR #3360

| **Assembly preload**  | **master**  | **this commit** | **Architecture** |
|-----------------------|-------------|-----------------|------------------|
| **enabled**           | 215.785     | 217.039         | armeabi-v7a      |
| **disabled**          | 182.055     | 182.784         | armeabi-v7a      |
| **enabled**           | 197.586     | 204.494         | arm64-v8a        |
| **disabled**          | 169.131     | 170.164         | arm64-v8a        |

### DIFF
--- a/ThirdPartyNotices.txt
+++ b/ThirdPartyNotices.txt
@@ -12,7 +12,7 @@ bring it to our attention. Post an issue or email us:
 The attached notices are provided for information only.
 
 1. bazelbuild/bazel (https://github.com/bazelbuild/bazel/)
-2. force-net/crc32.net (https://github.com/xamarin/proguard/)
+2. force-net/crc32.net (https://github.com/force-net/Crc32.NET)
 3. google/desugar (https://android.googlesource.com/platform/external/desugar/+/master/)
 4. nunit/nunitlite (https://github.com/nunit/nunitlite/)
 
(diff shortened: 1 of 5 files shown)


jonpryor:
`ApplicationRunsWithDebuggerAndBreaks()` fails, as with #3606. :-(

At least this narrows down the set of changes which causes the failure.

grendello:
> `ApplicationRunsWithDebuggerAndBreaks()` fails, as with #3606. :-(
> 
> At least this narrows down the set of changes which causes the failure.

Yeah, and I can see the app fully load and run locally, which is also an improvement. However, I'm still puzzled by what might be causing the behavior change. Will keep digging.

grendello:
Turns out the changes here make the startup *slower* - as surprising as it is, there's no point in merging this PR then. Closing!

dellis1972:
wow, really. The was unexpected :(

PR #3693, opened 2019-09-26
[MSBuildDeviceIntegration] Harden checks for the debugger.

dellis1972:
We hit a problem where the text for the log message
for the debugger was changed. It went from

	monodroid-debug: Trying to initialize the debugger with options:

to

	monodroid-debug: mono_runtime_init: Trying to initialize the debugger with options:

Which stopped our unit test matching the line.
So lets make this use just the text

	Trying to initialize the debugger with options:

and make sure we comment the line to if it changes people
also go an update the unit test.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
@@ -5,6 +5,7 @@ using System.Diagnostics;
 using System.IO;
 using System.Text.RegularExpressions;
 using System.Threading;
+using System.Threading.Tasks;
 using System.Xml.Linq;
 using System.Xml.XPath;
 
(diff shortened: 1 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
```diff
@@ -74,22 +89,29 @@ protected static bool MonitorAdbLogcat (Func<string, bool> action, string logcat
...
 					proc.BeginOutputReadLine ();
-					proc.WaitForExit (timeout * 1000);
+					TimeSpan time = TimeSpan.FromSeconds (timeout);
+					while (!stdout_done.IsSet && !didActionSucceed && time.TotalMilliseconds > 0) {
+						proc.WaitForExit (100);
+						time -= TimeSpan.FromMilliseconds (100);
+					}
+					proc.Kill ();
+					stdout_done.Wait ();
+					proc.WaitForExit ();
```
grendello: This line needs to be swapped with the previous one - buffers must be flushed after the process exits.

dellis1972: Fixed

PR #3696, opened 2019-09-26
[build] remove usage of ant

jonathanpeppers:
Context: http://build.azdo.io/3085600

We have been hitting a build error of:

    ==> Downloading https://www-us.apache.org/dist/commons/bcel/binaries/bcel-6.3.1-bin.tar.gz
    stderr | curl: (22) The requested URL returned error: 404 Not Found
    stderr | Error: An exception occurred within a child process:
    stderr |   DownloadError: Failed to download resource "ant--bcel"
    stderr | Download failed: https://www-us.apache.org/dist/commons/bcel/binaries/bcel-6.3.1-bin.tar.gz
    Error: Installation of ant failed

Which is just coming from `brew install ant`.

Looking into it, we only have usage of `<Ant/>` to build `proguard`.
It looks like there are instructions to build it with `maven`, `ant`,
`gradle`, `javac`, etc.

I reworked proguard to be built with `gradle` the same way we are
building `r8.jar`. This allows us to not need `ant` anymore.


jonathanpeppers:
👀 I keep finding more and more things that say `ant` in here... I'll try to get them all.

jonpryor:
More context: https://github.com/Homebrew/homebrew-core/issues/44608

PR #3697, opened 2019-09-26
[tests] add an MSBuild test for AndroidX migration

jonathanpeppers:
Context: https://www.nuget.org/packages/Xamarin.AndroidX.Migration

Xamarin.AndroidX.Migration has a lot of integration into the
Xamarin.Android MSBuild targets. We should include a test for
AndroidX, so we have better coverage of our MSBuild extension points.

We don't want to break it!

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -140,6 +140,38 @@ class MemTest {
 		}
 
 		[Test]
+		[NonParallelizable]
+		public void AndroidXMigration ([Values (true, false)] bool isRelease)
+		{
+			var proj = new XamarinFormsAndroidApplicationProject {
+				IsRelease = isRelease,
+			};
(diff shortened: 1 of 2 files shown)


jonpryor:
> We don't want to break it!

Uhh....?

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3087116&view=ms.vss-test-web.build-test-results-tab&runId=8715804&resultId=100141&paneView=attachments

```
/Library/Frameworks/Mono.framework/Versions/6.0.0/lib/mono/msbuild/Current/bin/NuGet.targets(121,5): error : A local file header is corrupt. [/Users/vsts/agent/2.158.0/work/1/s/bin/TestRelease/temp/AndroidXMigrationTrue/UnnamedProject.csproj]
  System.IO.InvalidDataException: A local file header is corrupt.
    at System.IO.Compression.ZipArchiveEntry.ThrowIfNotOpenable (System.Boolean needToUncompress, System.Boolean needToLoadIntoMemory) [0x00012] in <ae99261aa6fe4b3b969b577c81057d45>:0 
```

Not entirely sure what to make of that, other than NuGet is complaining about *something*.

jonathanpeppers:
Hmm, so this test brings in a huge number of new NuGet packages, I think it needs `[NonParallelizable]` to work on Mac. 😞

PR #3695, opened 2019-09-26
[tests] include all of bin/Test$(Configuration)/temp in xa-tests-results.zip

jonathanpeppers:
There have been some cases where we wanted to get the entire project
directory from the CI system after a test run has completed.

We can zip up all of:

    bin\Test$(Configuration)\temp\**\*

However, there are a few MSBuild-related tests that don't drop their
files in this directory:

* `CodeBehind`
* `CodeGen-MkBundle`
* `EmbeddedDSOs`

I changed these tests to output to `temp`. This should be OK even if
there is a naming collision, since they don't run at the same time as
the other MSBuild tests.

PR #3707, opened 2019-09-30
[Xamarin.Android.Build.Tasks] dispose GetManifestResourceStream

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/3695
Context: https://dev.azure.com/DevDiv/DevDiv/_build/results?buildId=3095849&view=ms.vss-test-web.test-result-details

We have been hitting a random MSBuild test failure on macOS:

    Unable to open 'obj/Debug/android/bin/UnnamedProject.UnnamedProject.apk' as zip archive

This appears to be coming from `zipalign` directly. In #3695 I was
able to get the APK file in question.

Using 7-zip on Windows, it reports the file is invalid:

    > 7z t "C:\Users\jopepper\Downloads\CheckResourceDesignerIsCreatedFalseCSharp\obj\Debug\android\bin\UnnamedProject.UnnamedProject.apk"                                                                                                                                                                                                                   7-Zip 19.00 (x64) : Copyright (c) 1999-2018 Igor Pavlov : 2019-02-21
    Scanning the drive for archives:
    1 file, 5578397 bytes (5448 KiB)
    Testing archive: C:\Users\jopepper\Downloads\CheckResourceDesignerIsCreatedFalseCSharp\obj\Debug\android\bin\UnnamedProject.UnnamedProject.apk
    --
    Path = C:\Users\jopepper\Downloads\CheckResourceDesignerIsCreatedFalseCSharp\obj\Debug\android\bin\UnnamedProject.UnnamedProject.apk
    Type = zip
    Physical Size = 5578397
    ERROR: Headers Error : NOTICE
    ERROR: Headers Error : classes.dex
    ERROR: Headers Error : lib\arm64-v8a\libmonodroid.so
    Sub items Errors: 3
    Archives with Errors: 1
    Sub items Errors: 3

Reviewing the code in `<BuildApk/>`, we aren't disposing what is
returned from:

    Assembly.GetExecutingAssembly ().GetManifestResourceStream ("NOTICE.txt")

An `UnmanagedMemoryStream` is returned here, so it is possible that
this could be the source of the issue? Disposing is good hygiene here
anyway. It is good to make this change even if it doesn't fix the
issue.

I also removed a completely unused `Close()` method in `ZipArchiveEx`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -115,10 +115,10 @@ namespace Xamarin.Android.Tasks
 			ArchiveFileList files = new ArchiveFileList ();
 			if (apkInputPath != null)
 				File.Copy (apkInputPath, temp, overwrite: true);
+			using (var notice = Assembly.GetExecutingAssembly ().GetManifestResourceStream ("NOTICE.txt"))
 			using (var apk = new ZipArchiveEx (temp, apkInputPath != null ? FileMode.Open : FileMode.Create )) {
 				apk.FixupWindowsPathSeparators ((a, b) => Log.LogDebugMessage ($"Fixing up malformed entry `{a}` -> `{b}`"));
-				apk.Archive.AddEntry (RootPath + "NOTICE",
-						Assembly.GetExecutingAssembly ().GetManifestResourceStream ("NOTICE.txt"));
+				apk.Archive.AddEntry (RootPath + "NOTICE", notice);
 
 				// Add classes.dx
 				foreach (var dex in DalvikClasses) {
--- a/src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
@@ -129,13 +129,6 @@ namespace Xamarin.Android.Tasks
 			}
 		}
 
-		public void Close ()
-		{
-			if (zip != null) {
-				zip.Close ();
-			}
-		}
-
 		public void Dispose ()
 		{
 			Dispose(true);


src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -115,10 +115,10 @@ void ExecuteWithAbi (string [] supportedAbis, string apkInputPath, string apkOut
 			ArchiveFileList files = new ArchiveFileList ();
 			if (apkInputPath != null)
 				File.Copy (apkInputPath, temp, overwrite: true);
+			using (var notice = Assembly.GetExecutingAssembly ().GetManifestResourceStream ("NOTICE.txt"))
```
grendello: It would be good measure to add a `.Flush()` call before the `using` block is exited as `Dispose()` doesn't flush buffers for some streams (neither does `Close()`)

jonathanpeppers: In this case we have our own `Flush` method (that isn't the real .NET `Flush`) that just does this: https://github.com/xamarin/xamarin-android/blob/936459c00e4a2da08e20cdbd83e5e23823529fbe/src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs#L30-L38

`ZipArchiveEx` does the same thing in `Dispose`: https://github.com/xamarin/xamarin-android/blob/936459c00e4a2da08e20cdbd83e5e23823529fbe/src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs#L145-L153

dellis1972:
I'm not entirely sure this is the problem. LibZipSharp keeps track of all `Stream` types it is given [1] via the `sources` HashSet.
These are then disposed of when the zip is Disposed [2]. The reason it is done this way is because libzip does not so its writing until the zip file is closed. So all the passed in streams must be valid at the time the zip file is closed.  In this case the `notice` should have been disposed of during the first call to `ZipArchiveEx.Flush`. 

That said if this cleans things up a bit (and does not cause any `Already Disposed` errors) then its a good change. 

[1] https://github.com/xamarin/LibZipSharp/blob/ffffd815b6ca2fb316d925c3242d4a7e580e630f/ZipArchive.cs#L323
[2] https://github.com/xamarin/LibZipSharp/blob/ffffd815b6ca2fb316d925c3242d4a7e580e630f/ZipArchive.cs#L820

jonpryor:
Unit test failures [are reported](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3097323&view=ms.vss-test-web.build-test-results-tab&runId=8761624&resultId=100474&paneView=debug), but they appear to be the same failures as [on master](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3097147&view=ms.vss-test-web.build-test-results-tab).

Merging...

PR #3711, opened 2019-10-01
[C++ification] Convert monodroid-glue to a class

grendello:
monodroid-glue is one of the last pieces of our native code that, so far, hasn't
been converted to "proper" C++ by putting the data and functions in a class,
while migrating some of them to more appropriate classes.

This commit implements just that:

  * move all the data members to a new `MonodroidRuntime` class
  * move a number of profiler and debugging functions to the `Debug` class
  * move most `MONO_API` functions to a separate file

Additionally, bits and pieces of code are modified to make them more C++ than
C (including some Windows-specific code to work around the ancient mingw version
we used at some point).

The simple pointer guard class is extended to work with C-allocated
pointers (the class still needs some work to make it more generally useful, but
that's beyond the scope of this commit).

### DIFF
--- a/src/monodroid/jni/debug.hh
+++ b/src/monodroid/jni/debug.hh
@@ -2,8 +2,9 @@
 #ifndef __MONODROID_DEBUG_H__
 #define __MONODROID_DEBUG_H__
 
-#include <stdint.h>
+#include <cstdint>
 #include <pthread.h>
+#include <sys/time.h>
 
 #ifdef __cplusplus
 namespace xamarin::android
@@ -15,6 +16,9 @@ namespace xamarin::android
 		{
 			int64_t timeout_time;
 		};
+
+	private:
+		static constexpr char INITIALIZER_NAME[] = "mono_profiler_init";
 
 	public:
 		/* Android property containing connection information, set by XS */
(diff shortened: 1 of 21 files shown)


src/monodroid/jni/debug.hh
```diff
@@ -2,7 +2,7 @@
 #ifndef __MONODROID_DEBUG_H__
 #define __MONODROID_DEBUG_H__
 
-#include <stdint.h>
+#include <cstdint>
 #include <pthread.h>
```
radekdoulik: Are you missing a `<sys/time.h>` include here?

```
12:15:41  "/Users/builder/jenkins/workspace/xamarin-android-pr-pipeline-release/xamarin-android/src/monodroid/monodroid.csproj" (default target) (33) ->
12:15:41  (_BuildHostRuntimes target) -> 
12:15:41    ../../../jni/debug.hh(75,3): error GF2D19E48: unknown type name 'timeval' [/Users/builder/jenkins/workspace/xamarin-android-pr-pipeline-release/xamarin-android/src/monodroid/monodroid.csproj]```

grendello: Huh, interesting. I didn't see that locally.

PR #3713, opened 2019-10-01
[profiled aot] Added 2 new properties for our targets

radekdoulik:
Add `AndroidUseDefaultAotProfile`. It can be used to suppress usage of
the default AOT profiles.

Add `AndroidAotProfiles` property to specify AOT profiles from command
line.

### DIFF
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -692,7 +692,7 @@ when packaging Release applications.
 -   **AndroidUseDefaultAotProfile** &ndash; A bool property which allows
     the developer to suppress usage of the default AOT profiles.
 
-    Added in Xamarin.Android 10.0.
+    Added in Xamarin.Android 10.1.
 
 -   **AndroidUseLegacyVersionCode** &ndash; A boolean property will allows
     the developer to revert the versionCode calculation back to its old pre
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -219,6 +219,12 @@ used by the `Install` and `SignAndroidPackage` targets.
 The [Signing Properties](#Signing_Properties) are also relevant
 when packaging Release applications.
 
+-   **AndroidAotProfiles** &ndash; A string property which allows the
+    developer to add AOT profiles from the command line. It is
+    semicolon or comma separated list of absolute paths.
+
+    Added in Xamarin.Android 10.1.
+
 -   **AndroidApkDigestAlgorithm** &ndash; A string value which specifies
     the digest algorithm to use with `jarsigner -digestalg`.
 
@@ -342,9 +348,6 @@ when packaging Release applications.
     The profiles are listed in `AndroidAotProfile` item group. This
     ItemGroup contains default profile(s). It can be overriden by
     removing the existing one(s) and adding your own AOT profiles.
-
-    The profiles can also be added with `AndroidAotProfiles` property,
-    which should contain colon separated list of absolute paths.
 
     Support for this property was added in Xamarin.Android 9.4.
 
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -343,6 +343,9 @@ when packaging Release applications.
     ItemGroup contains default profile(s). It can be overriden by
     removing the existing one(s) and adding your own AOT profiles.
 
+    The profiles can also be added with `AndroidAotProfiles` property,
+    which should contain colon separated list of absolute paths.
+
     Support for this property was added in Xamarin.Android 9.4.
 
     This property is `False` by default.
@@ -685,6 +688,11 @@ when packaging Release applications.
     use the to the `apksigner` tool rather than the `jarsigner`.
 
     Added in Xamarin.Android 8.2.
+
+-   **AndroidUseDefaultAotProfile** &ndash; A bool property which allows
+    the developer to suppress usage of the default AOT profiles.
+
+    Added in Xamarin.Android 10.0.
 
 -   **AndroidUseLegacyVersionCode** &ndash; A boolean property will allows
     the developer to revert the versionCode calculation back to its old pre
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -46,7 +46,7 @@ namespace Xamarin.Android.Build.Tests
 			};
 			using (var b = CreateApkBuilder (Path.Combine ("temp", TestName))) {
 				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
-				StringAssertEx.ContainsRegex (@"\[aot-compiler stdout\] Using profile data file.*xbuild.Xamarin.Android.startup\.aotprofile", b.LastBuildOutput, "Should use default AOT profile");
+				StringAssertEx.ContainsRegex (@"\[aot-compiler stdout\] Using profile data file.*build.Xamarin.Android.startup\.aotprofile", b.LastBuildOutput, "Should use default AOT profile", RegexOptions.IgnoreCase);
 			}
 		}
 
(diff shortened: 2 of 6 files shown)


Documentation/guides/BuildProcess.md
```diff
@@ -686,6 +689,11 @@ when packaging Release applications.
 
     Added in Xamarin.Android 8.2.
 
+-   **AndroidUseDefaultAotProfile** &ndash; A bool property which allows
+    the developer to suppress usage of the default AOT profiles.
+
+    Added in Xamarin.Android 10.0.
```
jonpryor: At this point, it will be added in *10.1*, assuming we add this to d16-4...

dellis1972:
This looks ok. Do we need unit tests though?

radekdoulik:
> This looks ok. Do we need unit tests though?

Not sure. What do you usually test when adding properties like these?

dellis1972:
@radekdoulik we would normally check that the value of `_AotProfiles` is correct. Or that the app at least builds correctly when the properties are set. 

Now that I think about it we can simplify this and remove the need for `AndroidUseDefaultAotProfile`

```
  <ItemGroup Condition="'$(AndroidEnableProfiledAot)' == 'True'">
    <_AotProfiles Include="$(AndroidAotProfiles.Split(':'))" />
    <AndroidAotProfile Condition=" '$(_AotProfiles)' == '' " Include="$(MSBuildThisFileDirectory)$(_StartupAotProfile)" />
  </ItemGroup>
```

So if the user does provide `AndroidAotProfiles` we will default to the `_StartupAotProfile` one. 
Do we need to use `:` as well? This is an itemGroup after all `;` should be fine. We can then remove the `Split` call.

radekdoulik:
Thanks for suggestions for the test, I will create some.

The `AndroidUseDefaultAotProfile` is not only for cases where `AndroidAotProfiles` is used from command line, but also for parties like XForms, where they intend to put own aot profile in the nuget package. In XF case they will want to disable the default profile.

For the `AndroidAotProfiles`, I tried to use semicolon (before using the `Split` call), but it didn't work for me and the item in `AndroidAotProfile` contained the whole string.

dellis1972:
that is weird.. using the `;` and quoting the entire property used to work 🤷‍♂

Documentation/guides/BuildProcess.md
```diff
@@ -343,6 +343,9 @@ when packaging Release applications.
     ItemGroup contains default profile(s). It can be overriden by
     removing the existing one(s) and adding your own AOT profiles.
 
+    The profiles can also be added with `AndroidAotProfiles` property,
```
jonpryor: `$(AndroidAotProfiles)` should be listed as a separate property, not grouped with `AndroidEnableProfiledAot`, in part because `AndroidEnableProfiledAot` was added to XA 9.4 while `$(AndroidAotProfiles)` will be added in XA 10.1/10.2 (10.1 if we CP this to d16-4).

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2829,11 +2829,12 @@ because xbuild doesn't support framework reference assemblies.
     <_StartupAotProfile Condition="'%(ResolvedAssemblies.Filename)' == 'Xamarin.Forms.Platform.Android'">startup-xf.aotprofile</_StartupAotProfile>
     <_StartupAotProfile Condition="'$(_StartupAotProfile)' == ''">startup.aotprofile</_StartupAotProfile>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(AndroidUseDefaultAotProfile)' != 'False'">
     <AndroidAotProfile Include="$(MSBuildThisFileDirectory)$(_StartupAotProfile)" />
   </ItemGroup>
   <ItemGroup Condition="'$(AndroidEnableProfiledAot)' == 'True'">
     <_AotProfiles Include="@(AndroidAotProfile)" />
+    <_AotProfiles Include="$(AndroidAotProfiles.Split(':'))" />
```
jonpryor: Splitting on `:` is going to be *really* problematic on Windows, as `:` is the *drive separator*.

Is there an actual benefit/use case to having an `$(AndroidAotProfiles)` MSBuild property, instead of using the `@(AndroidAotProfile)` build action?

If we want to keep `$(AndroidAotProfiles)`, then it should probably be `;`-separated, which in turn is *highly* problematic (because it can't be used on the command-line).

radekdoulik: The benefit is to be able to specify it from command line. One cannot specify items on command line when calling `msbuild` AFAIK.

I used `:` as we use it for ABIs. Now I noticed it is only used in our build system and it would be indeed problematic on Windows. The `;` can be used on command line, but needs to be escaped. So it might be still better choice.

Finally, I have found out we now have `SplitProperty` task, which looks like the best choice. It uses `;` and `,` as separators and is already used for `AndroidSupportedAbis`.

Documentation/guides/BuildProcess.md
radekdoulik: Which will place it far away from the `AndroidEnableProfiledAot` entry :-( It is property though, so I will make a it a separate entry.

I might write some separate piece of doc on how to use the profiled AOT from command line I guess.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -37,6 +37,33 @@ public void BuildBasicApplicationRelease ()
...
+
+		[Test]
+		public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
+		{
+			var proj = new XamarinAndroidApplicationProject () {
+				IsRelease = true,
+				AndroidEnableProfiledAot = true,
+			};
+			proj.SetProperty (proj.ActiveConfigurationProperties, "AndroidUseDefaultAotProfile", "false");
+			using (var b = CreateApkBuilder ("temp/BuildBasicApplicationRelease")) {
```
dellis1972: Can you replace this with `Path.Combine ("temp", TestName);` other wise the two tests will collide with each other.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -37,6 +37,33 @@ public void BuildBasicApplicationRelease ()
...
 		}
 
+		[Test]
+		public void BuildBasicApplicationReleaseProfiledAot ()
+		{
+			var proj = new XamarinAndroidApplicationProject () {
+				IsRelease = true,
+				AndroidEnableProfiledAot = true,
+			};
+			using (var b = CreateApkBuilder ("temp/BuildBasicApplicationRelease")) {
```
dellis1972: Can you replace this with Path.Combine ("temp", TestName); other wise the two tests will collide with each other.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -44,7 +44,7 @@ public void BuildBasicApplicationReleaseProfiledAot ()
 				IsRelease = true,
 				AndroidEnableProfiledAot = true,
 			};
-			using (var b = CreateApkBuilder ("temp/BuildBasicApplicationRelease")) {
+			using (var b = CreateApkBuilder ("temp/BuildBasicApplicationReleaseProfiledAot")) {
```
dellis1972: @radekdoulik There is no need to hard code these. You can use the `TestName` property. It will match the method name (and include/fix up arguments). That way if we add arguments or rename the test you only need to do it in one place. 

So use the following instead. 

```
using (var b = CreateApkBuilder (Path.Combine ("temp", TestName))) {
}
```

dellis1972: Here is a sample https://github.com/xamarin/xamarin-android/blob/a360cb47023121dbb3ecbb9fab01cb345c5921fd/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L228

radekdoulik: I already fixed that yesterday :-)

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -37,6 +37,33 @@ public void BuildBasicApplicationRelease ()
 			}
 		}
 
+		[Test]
+		public void BuildBasicApplicationReleaseProfiledAot ()
```
dellis1972: @radekdoulik I just realised these tests need a device right? 
So they are in the wrong project. The should be in the `MSBuildDeviceIntegration` project in `tests/MSBuildDeviceIntegration` where we have all the device based tests. 

It should also make use of the `HasDevices` property and `Ignore` the test if we don't have any devices attached.
For example https://github.com/xamarin/xamarin-android/blob/master/tests/MSBuildDeviceIntegration/Tests/InstallTests.cs#L233

Issue #3348, opened 2019-07-12
Crash while using kotlinx-couroutine 


Maddy79:
### Steps to Reproduce

We have created a binding from a AAR file that has kotlinx-couroutine as one of its dependency. While creating the binding the dependency is added as below.
    <EmbeddedReferenceJar Include="Jars\kotlinx-coroutines-android-1.1.1.jar" />

The app that uses the binding builds fine and on launch it crashes with the below exception.

[AndroidRuntime] FATAL EXCEPTION: main
[AndroidRuntime] Process: com.sample.xamarinsampleapp, PID: 20262
[AndroidRuntime] java.lang.IllegalStateException: Module with the Main dispatcher is missing. Add dependency providing the Main dispatcher, e.g. 'kotlinx-coroutines-android'
[AndroidRuntime] 	at kotlinx.coroutines.internal.MissingMainCoroutineDispatcher.missing(MainDispatchers.kt:73)
[AndroidRuntime] 	at kotlinx.coroutines.internal.MissingMainCoroutineDispatcher.isDispatchNeeded(MainDispatchers.kt:54)
[AndroidRuntime] 	at kotlinx.coroutines.DispatchedKt.resumeCancellable(Dispatched.kt:373)
[AndroidRuntime] 	at kotlinx.coroutines.intrinsics.CancellableKt.startCoroutineCancellable(Cancellable.kt:25)
[AndroidRuntime] 	at kotlinx.coroutines.CoroutineStart.invoke(CoroutineStart.kt:109)
[AndroidRuntime] 	at kotlinx.coroutines.AbstractCoroutine.start(AbstractCoroutine.kt:160)
[AndroidRuntime] 	at kotlinx.coroutines.BuildersKt__Builders_commonKt.launch(Builders.common.kt:54)
[AndroidRuntime] 	at kotlinx.coroutines.BuildersKt.launch(Unknown Source:1)
[AndroidRuntime] 	at kotlinx.coroutines.BuildersKt__Builders_commonKt.launch$default(Builders.common.kt:47)
[AndroidRuntime] 	at kotlinx.coroutines.BuildersKt.launch$default(Unknown Source:1)
[AndroidRuntime] 	at com.sample.core.compliance.ComplianceExecutor$Companion.runTaskParallel(ComplianceExecutor.kt:30)
[AndroidRuntime] 	at com.sample.core.compliance.ComplianceManager.executeTask(ComplianceManager.kt:61)
[...]

PR #3716, opened 2019-10-01
[Xamarin.Android.Build.Tasks] support for META-INF/services inclusion in APKs

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/3348

Usage of a Kotlin library in a Xamarin.Android application could fail
with:

    [AndroidRuntime] FATAL EXCEPTION: main
    [AndroidRuntime] Process: com.sample.xamarinsampleapp, PID: 20262
    [AndroidRuntime] java.lang.IllegalStateException: Module with the Main dispatcher is missing. Add dependency providing the Main dispatcher, e.g. 'kotlinx-coroutines-android'

Reviewing the results of an APK from Android Studio, these files were
missing in the equivalent Xamarin.Android apk:

    META-INF/services/kotlinx.coroutines.internal.MainDispatcherFactory
    META-INF/services/kotlinx.coroutines.CoroutineExceptionHandler

I was able to piece together what the behavior is from Google's
Android Gradle plugin.

The top level gradle task:

https://android.googlesource.com/platform/tools/base/+/gradle_2.0.0/build-system/gradle-core/src/main/groovy/com/android/build/gradle/internal/tasks/ExtractJavaResourcesTask.java?autodive=0%2F%2F%2F%2F%2F%2F%2F%2F%2F#199

Ignores `java.util.jar.JarFile.MANIFEST_NAME`:

http://hg.openjdk.java.net/jdk8u/jdk8u-dev/jdk/file/0fc878b99541/src/share/classes/java/util/jar/JarFile.java#l91

Then makes use of `PackagingUtils.java` for the remaining checks:

https://android.googlesource.com/platform/tools/base/+/d41d662dbf89f9b60ca6256415a059c0107749b8/sdk-common/src/main/java/com/android/ide/common/packaging/PackagingUtils.java

I have ported the code to C#, so our behavior should be identical to
Android Studio now.

I added a few tests:

* MSBuild tests that use `kotlinx-coroutines-android-1.3.2.jar` in
  both APKs and app bundles.
* Unit tests checking the general `PackagingUtils.cs` class works.

### DIFF
--- a/ThirdPartyNotices.txt
+++ b/ThirdPartyNotices.txt
@@ -11,10 +11,219 @@ bring it to our attention. Post an issue or email us:
 
 The attached notices are provided for information only.
 
-1. bazelbuild/bazel (https://github.com/bazelbuild/bazel/)
-2. force-net/crc32.net (https://github.com/xamarin/proguard/)
-3. google/desugar (https://android.googlesource.com/platform/external/desugar/+/master/)
-4. nunit/nunitlite (https://github.com/nunit/nunitlite/)
+1. android/platform/tools/base (https://android.googlesource.com/platform/tools/base/+/d41d662dbf89f9b60ca6256415a059c0107749b8/sdk-common/NOTICE)
+2. bazelbuild/bazel (https://github.com/bazelbuild/bazel/)
+3. force-net/crc32.net (https://github.com/force-net/Crc32.NET)
+4. google/desugar (https://android.googlesource.com/platform/external/desugar/+/master/)
+5. nunit/nunitlite (https://github.com/nunit/nunitlite/)
+
+%% android/platform/tools/base NOTICES AND INFORMATION BEGIN HERE
+=================================================================
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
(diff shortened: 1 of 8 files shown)


src/Xamarin.Android.Build.Tasks/Utilities/PackagingUtils.cs
```diff
@@ -0,0 +1,104 @@
...
+		/// <summary>
+		/// Checks whether a folder and its content is valid for packaging into the .apk as standard Java resource.
+		/// </summary>
+		/// <param name="folderName">the name of the folder.</param>
+		/// <returns>true if the folder is valid for packaging.</returns>
+		static bool CheckFolderForPackaging (string folderName)
+		{
+			return !EqualsIgnoreCase (folderName, "CVS") &&
+				!EqualsIgnoreCase (folderName, ".svn") &&
+				!EqualsIgnoreCase (folderName, "SCCS") &&
```
dellis1972: I know its not in the source code, but should we include `.git` there? or any other well known source control hidden files?

src/Xamarin.Android.Build.Tasks/Utilities/PackagingUtils.cs
```diff
@@ -0,0 +1,104 @@
...
+		/// <summary>
+		/// Checks a file to make sure it should be packaged as standard resources.
+		/// </summary>
+		/// <param name="fileName">the name of the file (including extension)</param>
+		/// <param name="extension">the extension of the file (including '.')</param>
+		/// <returns>true if the file should be packaged as standard java resources.</returns>
+		static bool CheckFileForPackaging (string fileName, string extension)
+		{
+			// ignore hidden files and backup files
+			return !(fileName [0] == '.' || fileName [fileName.Length - 1] == '~') &&
```
dellis1972: actually, this should skip any hidden file.

src/Xamarin.Android.Build.Tasks/Utilities/PackagingUtils.cs
jonathanpeppers: I think they did `.svn` this way, because Subversion stores stuff in directories the whole way down. It's not just at the top level.

jonathanpeppers:
I was testing this and there is still another bug here:
```
> 7z l obj\Debug\100\lp\0\jl\__reference__kotlinx-coroutines-core-1.3.1.jar | grep services
```
But the original jar:
```
> 7z l Jars\kotlinx-coroutines-android-1.3.1.jar | grep service                                                                                                                                                                                                                                2019-09-04 15:13:12 D....            0            2  META-INF\services
2019-09-04 15:13:12 .....           53           49  META-INF\services\kotlinx.coroutines.CoroutineExceptionHandler
2019-09-04 15:13:12 .....           51           48  META-INF\services\kotlinx.coroutines.internal.MainDispatcherFactory
```
So there is still a problem if using `@(EmbeddedJar)` in a class library.

ThirdPartyNotices.txt
```diff
@@ -12,9 +12,10 @@ bring it to our attention. Post an issue or email us:
 The attached notices are provided for information only.
 
 1. bazelbuild/bazel (https://github.com/bazelbuild/bazel/)
-2. force-net/crc32.net (https://github.com/xamarin/proguard/)
-3. google/desugar (https://android.googlesource.com/platform/external/desugar/+/master/)
-4. nunit/nunitlite (https://github.com/nunit/nunitlite/)
+2. force-net/crc32.net (https://github.com/force-net/Crc32.NET)
+3. google/aosp (https://source.android.com/setup/start/licenses)
```
jonpryor: Instead of `google/aosp`, this should be `android/platform/tools/base`

jonpryor: Additionally, the URL shouldn't be the URL to the license -- `/setup/start/licenses` -- but instead to *where the source code was obtained*, in this case (presumably) https://android.googlesource.com/platform/tools/base/+/d41d662dbf89f9b60ca6256415a059c0107749b8/sdk-common/src/main/java/com/android/ide/common/packaging/PackagingUtils.java or some "higher-level" URL like https://android.googlesource.com/platform/tools/base.

src/Xamarin.Android.Build.Tasks/Utilities/PackagingUtils.cs
```diff
@@ -0,0 +1,104 @@
...
+	/// <summary>
+	/// Utility class for packaging.
+	/// </summary>
+	internal class PackagingUtils
+	{
+		/// <summary>
+		/// The JAR manifest file name.
+		/// http://hg.openjdk.java.net/jdk8u/jdk8u-dev/jdk/file/0fc878b99541/src/share/classes/java/util/jar/JarFile.java#l91
+		/// </summary>
+		const string MANIFEST_NAME = "META-INF/MANIFEST.MF";
```
jonpryor: This should be `ManifestFileName` or something more .NET-"normal", as there doesn't appear to be a `MANIFEST_NAME` field in the Java code to be consistent with.

jonpryor: Or you could go one further, and do:

```csharp
static readonly string[] InvalidEntryPaths = new[]{
    "META-INF/MANIFEST.MF",
};
```

Then in `CheckEntryForPackaging()`:

```csharp
if (InvalidEntryPaths.Contains (entryName))
    return false;
```

src/Xamarin.Android.Build.Tasks/Utilities/PackagingUtils.cs
```diff
@@ -0,0 +1,104 @@
...
+
+		/// <summary>
+		/// Checks if a zip entry is valid for packaging into the .apk as standard Java resource.
+		/// </summary>
+		/// <param name="entryName">the name of the zip entry from Xamarin.Tools.Zip.ZipEntry.FullName.</param>
+		/// <returns>true if the entry is valid for packaging.</returns>
+		public static bool CheckEntryForPackaging (string entryName)
+		{
+			if (entryName == MANIFEST_NAME)
+				return false;
```
jonpryor: ...though I'm not sure why this code exists at all, considering that the Java code doesn't have this check.

jonathanpeppers:
The issue I found in https://github.com/xamarin/xamarin-android/pull/3716#issuecomment-537130858

Was actually just the repro in https://github.com/xamarin/xamarin-android/issues/3348, I changed the build action from:
```
<LibraryProjectZip Include="Jars\kotlinx-coroutines-android-1.3.1.jar" />
```

To this and it worked:
```xml
<EmbeddedReferenceJar Include="Jars\kotlinx-coroutines-android-1.3.1.jar" />
```
Which I think was actually incorrect, `aar` files are what `@(LibraryProjectZip)` is for.

I'll address the other comments here.

src/Xamarin.Android.Build.Tasks/Utilities/PackagingUtils.cs
jonathanpeppers: They check it at the gradle task level: https://android.googlesource.com/platform/tools/base/+/gradle_2.0.0/build-system/gradle-core/src/main/groovy/com/android/build/gradle/internal/tasks/ExtractJavaResourcesTask.java?autodive=0%2F%2F%2F%2F%2F%2F%2F%2F%2F#201

jonathanpeppers:
The only failure was the random `zipalign` issue we've been seeing: 
```
error MSB6006: "zipalign" exited with code 1.
```

Maddy79:
@jonathanpeppers , thanks for helping to get this closed. 
@dellis1972 , do we have any timeline on when this will be available in the Preview channel?

PR #3720, opened 2019-10-01
[Xamarin.Android.Build.Tasks] <FilterAssemblies/> shouldn't throw FNFE

jonathanpeppers:
A few places during a Xamarin.Android build, we have the following
pattern:

    if (DesignTimeBuild && !File.Exists (assemblyPath)) {
        Log.LogDebugMessage ($"Skipping non-existent dependency '{assemblyPath}' during a design-time build.");
        continue;
    }

In the case of a `<ProjectReference/>` not building, there starts to
be a problem here.

What happens during a regular build, is we end up throwing a
`FileNotFoundException` in `<FilterAssemblies/>`.

If we remove the `if (DesignTimeBuild)` check, we would instead see
the behavior you would get for a .NET Console app referencing a .NET
Framework library project:

    (ResolveProjectReferences target) ->
        C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(1875,5):
        warning : The referenced project '..\MyLibrary\MyLibrary.csproj' does not exist.
    (CoreCompile target) ->
        Foo.cs(1,20,1,23):
        error CS0246: The type or namespace name 'Bar' could not be found (are you missing a using directive or an assembly reference?)
        1 Warning(s)
        1 Error(s)

We get a warning about a missing `<ProjectReference/>` and any C#
compiler errors that would be encountered.

I think this makes more sense and developers will less likely point
fingers at `<FilterAssemblies/>`?

I added a test around this scenario, too.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
@@ -972,5 +972,54 @@ namespace Lib2
 				Assert.IsTrue (DexUtils.ContainsClass (className, dexFile, AndroidSdkPath), $"`{dexFile}` should include `{className}`!");
 			}
 		}
+
+		[Test]
+		public void MissingProjectReference ()
+		{
+			var path = Path.Combine ("temp", TestName);
+
+			var bar = "public class Bar { }";
+			var lib = new XamarinAndroidLibraryProject {
+				ProjectName = "MyLibrary",
+				Sources = {
+					new BuildItem.Source ("Bar.cs") {
+						TextContent = () => bar
+					},
+				}
+			};
+			var app = new XamarinAndroidApplicationProject {
+				ProjectName = "MyApp",
+				Sources = {
+					new BuildItem.Source ("Foo.cs") {
+						TextContent = () => "public class Foo : Bar { }"
+					},
+				}
+			};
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
```diff
@@ -972,5 +972,54 @@ public void ChangePackageNamingPolicy ()
...
+
+			using (var libBuilder = CreateDllBuilder (Path.Combine (path, lib.ProjectName)))
+			using (var appBuilder = CreateApkBuilder (Path.Combine (path, app.ProjectName))) {
+				libBuilder.ThrowOnBuildFailure =
+					appBuilder.ThrowOnBuildFailure = false;
+
+				// Build app before library is built
+				Assert.IsFalse (appBuilder.Build (app), "app build should have failed.");
+				Assert.IsTrue (StringAssertEx.ContainsText (appBuilder.LastBuildOutput, $"The referenced project '{reference}' does not exist."));
+				Assert.IsTrue (StringAssertEx.ContainsText (appBuilder.LastBuildOutput, " 1 Warning(s)"), "Should receive 1 Warning");
```
jonathanpeppers: Hmm, on macOS we get 0 warnings:
```
    0 Warning(s)
```
Will look into this when I return from vacation.

PR #3729, opened 2019-10-02
JNIEnv.Initialize optimization

grendello:
`JNIEnv.Initialize` is part of the Xamarin.Android runtime startup sequence and
it contributes a sizable chunk of time to it. Some code executed during startup
sequence wasn't as optimized as it could be and this commit aims to pick all the
low-hanging fruit in that area:

  * Don't use the `Logger` class.
    This removes one class instantiation which wasn't absolutely necessary.
    Calls to `Logger` are replaced with calls to a new native function
    `monodroid_log`
  * Use native functions for timing.
    It won't affect the user's code (unless it runs with timing enabled, of
    course) but will allow our timing to be more precise and have less impact on
    code performance. Implemented with two new native functions
    `monodroid_timing_start` and `monodroid_timing_stop`
  * Remove calls to `Environment.GetEnvironmentVariable`
    The calls took a non-trivial amount of time (around 4ms each) and were
    replaced by putting the environment variables in `libxamarin-app.so`, in the
    application environment configuration block.
  * Do not fully initialize `UncaughtExceptionHandler` during init
    The handler's initialization is quite heavy while the handler itself is
    going to be used only when there's an unhandled exception thrown. Delay full
    initialization to the point in time when the handler is actually needed.
  * Avoid parsing of the `PackageNamingPolicy` enum values on runtime.
    Instead, the desired value is placed in the application's environment
    configuration block and simply cast from `int` to the enum.
  * Initialize `Logger` lazily.
    The logging category is requested by `Logger` lazily instead of being set
    from `JNIEnv.Initialize`. This allows the `Logger` class not to be
    instantiated during startup.
  * Avoid static initialization of the `Java.Interop.Runtime` class.
  * Avoid heavy static initialization of the `Java.Interop.TypeManager` class.
    This is done by initializing the type lookup dictionaries lazily, via
    another "proxy" class.
  * Make Java class name lookup faster.
    This is done by moving the lookup (and processing) to native code and
    p/invoking the new `monodroid_TypeManager_get_java_class_name` function.

Other changes:

  * Remove dead code from `JNIEnv`
    The `JAVA_INTEROP` symbol is always "defined" now, remove code used when it
    was not defined.
  * Introduce fast performance measurement for managed code.
    The timing code used by the native runtime is now exposed to the managed
    land, thus allowing us fast and accurate (to the nanosecond level) timing
    measurement.

Impact of the above changes on performance is summarized in the table below.

Device name: **Pixel 3 XL**
Device architecture: **arm64-v8a**
Number of test runs: **10**

|                 | **Runtime init** | **Displayed** | **Notes**                      |
|-----------------|------------------|---------------|--------------------------------|
| **this commit** | 187.907          | 828.90        |  preload enabled; 32-bit build |
| **master**      | 218.804          | 852.70        |                                |
| **this commit** | 154.718          | 803.50        | preload disabled; 32-bit build |
| **master**      | 185.240          | 827.40        |                                |
| **this commit** | 174.211          | 754.50        |  preload enabled; 64-bit build |
| **master**      | 201.779          | 783.20        |                                |
| **this commit** | 144.158          | 741.80        | preload disabled; 64-bit build |
| **master**      | 173.322          | 759.40        |                                |


src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -29,7 +29,9 @@ struct JnienvInitializeArgs {
 		public int             localRefsAreIndirect;
 		public int             grefGcThreshold;
 		public IntPtr          grefIGCUserPeer;
-		public int             isRunningOnDesktop;
```
jonpryor: I wonder if this could *possibly* break the Designer...  Just a rando fear.

src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -29,7 +29,9 @@ struct JnienvInitializeArgs {
...
+		public byte            isRunningOnDesktop;
+		public byte            brokenExceptionTransitions;
+		public PackageNamingPolicy packageNamingPolicy;
```
jonpryor: As this struct is shared with C++/`libmonodroid.so`, I think this field should properly be an `int`, and then we cast in `JNIEnv.Initialize()`.  The cast should be optimized away to ~nothing.

src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -115,53 +117,42 @@ public static void CheckHandle (IntPtr jnienv)
...
-			if (Logger.LogTiming) {
-				var __end = DateTime.UtcNow;
-				Logger.Log (LogLevel.Info,
-						"monodroid-timing",
-						"JNIEnv.RegisterJniNatives total time: " + (__end - new DateTime (1970, 1, 1)).TotalMilliseconds + " [elapsed: " + (__end - __start).TotalMilliseconds + " ms]");
-			}
 		}
 
+		[DllImport ("__Internal")]
+		extern static void monodroid_log (LogLevel level, LogCategories category, string message);
```
jonpryor: At *this* point in time, it's "weird" to have this and the following methods declared when they're not actually used from C#.  I imagine this will change in the future, but for *now*, this is weird.

src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -193,60 +177,34 @@ internal static unsafe void Initialize (JnienvInitializeArgs* args)
 
 			Mono.SystemDependencyProvider.Initialize ();
 
-#if JAVA_INTEROP
+			// TODO: env, javaVm, grefLoader, Loader_loadClass should all be obtained by AndroidRuntime constructor with p/invokes
```
jonpryor: Related: P/Invokes may be "fast", but let's not do one P/Invoke per value.  Bundle them!  Minimize P/Invoke transitions!

src/Mono.Android/Android.Runtime/Logger.cs
```diff
@@ -58,18 +61,28 @@ internal static class Logger {
...
-		Unknown,
-		Default,
-		Verbose,
-		Debug,
-		Info,
-		Warn,
-		Error,
-		Fatal,
-		Silent
+		Unknown = 0x00,
```
jonpryor: Why the "noise"?

For that matter, why *hex* for values which *aren't* flags/bitfields?

src/Mono.Android/Java.Interop/TypeManager.cs
```diff
@@ -21,26 +21,6 @@ public static partial class TypeManager {
 
 		internal static IntPtr id_Class_getName;
 
-		static TypeManager ()
```
jonpryor: *Relatedly*, *in this file*, we want to *remove* the allocations on lines 19  & 20, as that will require JIT'ing Dictionary instantiations which we'd rather avoid if at all possible, especially *on-device*, when we shouldn't really need those fields *at all*.  (Not needing them will require additional changes, but they shouldn't be needed...)

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -141,11 +145,16 @@ void AddEnvironment ()
...
 			bool usesMonoAOT = false;
 			bool usesAssemblyPreload = EnablePreloadAssembliesDefault;
+			bool brokenExceptionTransitions = false;
 			uint monoAOTMode = 0;
 			string androidPackageName = null;
 			var environmentVariables = new Dictionary<string, string> (StringComparer.Ordinal);
 			var systemProperties = new Dictionary<string, string> (StringComparer.Ordinal);
 
+			if (!Enum.TryParse (PackageNamingPolicy, out PackageNamingPolicy pnp)) {
+				pnp = PackageNamingPolicyEnum.LowercaseHash;
```
jonpryor: As of 88a1d6cd88e68bda208a6eb05d99d802c7800d2e, this should probably default to `PackageNamingPolicy.LowercaseCrc64`.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -120,11 +120,11 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 <!--
 *******************************************
   Extensibility hook that allows VS to
-  provide extra behavior without modifying 
+  provide extra behavior without modifying
```
jonpryor: Is there any way to *avoid* all of these whitespace changes?

Sure, in theory, they improve the code base.  But it's making this PR *really* noisy.

src/Mono.Android/Android.Runtime/JNIEnv.cs
grendello: Designer tests *did* fail in this build, although it doesn't appear to be related (but might be?):

```
1) Failed : Xamarin.AndroidDesigner.Tests.Moyeu.CanLoadLayoutsInProject
  /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/ADT-0/RT-Moyeu-1/Moyeu/Resources/layout/InfoPaneLayout.axml
  Expected: no item equal to "MockView"
  But was:  < "MockView" >
  First non-matching item at index [0]:  "MockView"
  at Xamarin.AndroidDesigner.Tests.RegressionTest.CanLoadLayoutsInProject () [0x00000] in <055988ff4d7c4b6f95c58bd23af8596c>:0 
  at System.Runtime.CompilerServices.AsyncMethodBuilderCore+MoveNextRunner.InvokeMoveNext (System.Object stateMachine) [0x00000] in <208e3d8480c243b7804b22b6a5eb72fc>:0
```

src/Mono.Android/Android.Runtime/JNIEnv.cs
grendello: The other two are used, this one was used instead of passing a string to `monodroid_timing_{start,end}` but it will still come useful IMO (I think it might be a good idea to replace `Logger.Log` calls with this in the code we ship)

src/Mono.Android/Android.Runtime/JNIEnv.cs
grendello: That's the plan :)

src/Mono.Android/Android.Runtime/Logger.cs
grendello: The "noise" is to keep it explicit since the enum is "shared" with `C++` - it's better to have the values stated clearly IMO. And why hex? I like the way hex numbers look, no other reason :)

src/Mono.Android/Java.Interop/TypeManager.cs
grendello: Yep, that's the next step here.

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
grendello: it was a copy-and-paste, oops. I'll change the default in the original location too, then (the default should be really specified somewhere else...)

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
grendello: The `*.targets` files are often badly formatted and my editor removes trailing whitespace (as I want it to). I'll hand-edit the changes.

src/Mono.Android/Android.Runtime/JNIEnv.cs
grendello: And they failed again in the same way after reverting this change, so this is not the cause

src/Mono.Android/Android.Runtime/Logger.cs
```diff
@@ -58,18 +61,28 @@ internal static class Logger {
 					System.Console.WriteLine ("[{0}] {1}: {2}", level, appname, line);
 			}
 		}
+
+		[DllImport ("__Internal")]
+		extern static uint monodroid_get_log_categories ();
+
+		static Logger ()
+		{
+			lock (initLock) {
```
jonpryor: Is this strictly necessary?  Static constructors will only ever be executed once, so this seems redundant.

src/Mono.Android/Android.Runtime/Logger.cs
```diff
@@ -58,18 +60,26 @@ internal static class Logger {
 					System.Console.WriteLine ("[{0}] {1}: {2}", level, appname, line);
 			}
 		}
+
+		[DllImport ("__Internal")]
```
jonpryor: Shouldn't this specify `CallingConvention=CallingConvention.Cdecl`?

src/Mono.Android/Android.Runtime/Logger.cs
```diff
@@ -58,18 +60,26 @@ internal static class Logger {
...
+		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
+		extern static uint monodroid_get_log_categories ();
+
+		static Logger ()
+		{
+			logCategories = (LogCategories) monodroid_get_log_categories ();
+		}
 	}
 
 	internal enum LogLevel {
```
jonpryor: If these need to be kept in-sync with `libmonodroid`, we should add an appropriate comment here, as was done with `LogCategories`.

src/Mono.Android/Java.Interop/TypeManager.cs
```diff
@@ -9,48 +9,54 @@
...
 			}
 		}
+	}
+
+	public static partial class TypeManager {
+		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
+		extern static IntPtr monodroid_TypeManager_get_java_class_name (IntPtr klass);
+
+		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
+		extern static void monodroid_free (IntPtr ptr);
```
jonpryor: `AndroidEnvironment` already has this P/Invoke declaration.  We should make the `AndroidEnvironment` method internal, or move it to some shared location, instead of having two separate `monodroid_free()` declarations.

src/monodroid/jni/external-api.cc
```diff
@@ -174,6 +180,40 @@ MONO_API int monodroid_embedded_assemblies_set_assemblies_prefix (const char *pr
...
+
+	if (sequence == nullptr)
+		return;
+
+	sequence->period.mark_end ();
+	Timing::info (sequence->period, message == nullptr ? DEFAULT_MESSAGE : message);
+	timing->release_sequence (sequence);
+}
+
+MONO_API __stdcall const char*
```
jonpryor: I believe that this line should _not_ have `__stdcall` anymore.

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -906,8 +907,28 @@ MonodroidRuntime::lookup_bridge_info (MonoDomain *domain, MonoImage *image, cons
...
+
+	log_warn (LOG_GC, "GREF GC Threshold: %i", init.grefGcThreshold);
+
+	init.grefClass = utils.get_class_from_runtime_field (env, runtimeClass, "java_lang_Class", true);
+	Class_getName  = env->GetMethodID (init.grefClass, "getName", "()Ljava/lang/String;");
+	init.Class_forName = env->GetStaticMethodID (init.grefClass, "forName", "(Ljava/lang/String;ZLjava/lang/ClassLoader;)Ljava/lang/Class;");
+
 	MonoAssembly *assm = utils.monodroid_load_assembly (domain, "Mono.Android");
-	MonoImage *image   = mono_assembly_get_image (assm);
+	MonoImage *image = mono_assembly_get_image  (assm);
```
jonpryor: The whitespace changes here are weird.  _Two_ spaces before `(`?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1759,6 +1759,29 @@ MonodroidRuntime::Java_mono_android_Runtime_destroyContexts (JNIEnv *env, jclass
 	log_info (LOG_DEFAULT, "All domain cleaned up");
 }
 
+const char*
```
jonpryor: This should have a `char*` return type, not `const char*`, because the returned string should be free'd by the caller.

src/monodroid/jni/timing.hh
```diff
@@ -0,0 +1,161 @@
...
+		~Timing () noexcept
+		{
+			delete[] sequence_pool;
+		}
+
+		static void debug (timing_period const &period, const char *message) noexcept
+		{
+			timing_diff diff (period);
+
+			log_debug_nocheck (LOG_TIMING, MESSAGE_FORMAT, message == nullptr ? "" : message, diff.sec, diff.ms, diff.ns);
```
jonpryor: Why the `_nocheck` variant here?

For that matter, why's it _exist_?  I can't see any mention of `Timing::debug` in this PR.  I *do* see `Timing::info`...

grendello: `_nocheck` is used because we always hide timing stuff behind a check if the timing category is enabled, so there's no point in re-checking it. And why it exists? Because I used it in test code and forgot to remove.

src/monodroid/jni/timing.hh
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		static void info (timing_period const &period, const char *message) noexcept
+		{
+			timing_diff diff (period);
+
+			log_info_nocheck (LOG_TIMING, MESSAGE_FORMAT, message == nullptr ? "" : message, diff.sec, diff.ms, diff.ns);
```
jonpryor: ~Same question as with `Timing::debug()`: why the `_nocheck` variant?  *Currently* all callers appear to check the log level before calling `Timing::info` before hand, but will that remain the case in the future?

src/monodroid/jni/timing.hh
```diff
@@ -0,0 +1,161 @@
...
+			return ret;
+		}
+
+		void release_sequence (managed_timing_sequence *sequence)
+		{
+			if (sequence == nullptr)
+				return;
+
+			std::lock_guard<std::mutex> lock (sequence_lock);
+			if (sequence->dynamic) {
```
jonpryor: Shouldn't this be done *before* `lock` is constructed?  We don't need the mutex when `sequence->dynamic` is true.

src/monodroid/jni/timing.hh
grendello: Yes, this is the intention - that it will remain so in the future.

src/monodroid/jni/timing.hh
grendello: If you do it before the lock you introduce a race condition

PR #3365, opened 2019-07-17
[monodroid] Allow loading assemblies from memory

garuma:
This PR adds the necessary code and entry-point to be able to supply assembly data directly from memory.

The motivation is to allow the designer more flexibility in how it acquire those assemblies to begin with and avoid a complicated dance of file paths juggling and file copying that are detrimental to performance on Windows and cause file locking problems.

This is achieved by introducing a new class `InMemoryAssemblies` that, much like `EmbeddedAssemblies`, can be queried when a new .dll needs to be loaded to fetch it from a store that's keyed per-assembly name and per-appdomain. If not found in this new store it falls back to the existing logic that can load assemblies directly from disk.

Since we only care about this capability on desktop, the support is only compiled for host platforms (both Mac and Windows).

PR #3734, opened 2019-10-03
[monodroid] Allow sideloading real paths in addition to in-memory assemblies

garuma:
This PR extends the previous work from https://github.com/xamarin/xamarin-android/pull/3365 that allowed in-memory data to be given for an assembly content to also permit supplying a real path instead.

The reasoning is that for some assemblies it's safe and preferable to allow them to be loaded from a given location. A good example are assemblies coming from the NuGet global package cache which are considered read-only and thus won't cause Windows file-locking issues when loaded up (as opposed to when they are loading from the project obj directory).

By selectively allowing those type of assemblies to be loaded from disk, we avoid putting unnecessary memory pressure on the system while retaining the flexibility to do so if we choose.


src/java-runtime/java/mono/android/Runtime.java
```diff
@@ -16,7 +16,7 @@ private Runtime ()
 	public static native void register (String managedType, java.lang.Class nativeClass, String methods);
 	public static native void notifyTimeZoneChanged ();
 	public static native int createNewContext (String[] runtimeApks, String[] assemblies, ClassLoader loader);
-	public static native int createNewContextWithData (String[] runtimeApks, String[] assemblies, byte[][] assembliesBytes, ClassLoader loader, boolean forcePreloadAssemblies);
+	public static native int createNewContextWithData (String[] runtimeApks, String[] assemblies, byte[][] assembliesBytes, String[] assembliesPaths, ClassLoader loader, boolean forcePreloadAssemblies);
```
jonpryor: As with [last time](https://github.com/xamarin/xamarin-android/pull/3365#discussion_r305070762), ABI breakage like this will probably break the Designer.

I also think it's time that we started introducing aspects of https://github.com/xamarin/designer/issues/1597

Does `createNewContextWithData()` *need* to be "complete"?  Or can it "mutate" an existing context?

```java
public class DesignerRuntime {
    private DesignerRuntime() {
    }

    public static native int createNewContext (String[] runtimeApks, String[] assemblies, ClassLoader loader);
    public static void loadAssemblies(int contextID, String[] assembliesNames, byte[][] assembliesBytes, boolean forcePreloadAssemblies);
    public static void loadAssemblies(int contextID, String[] assembliesNames, String[] assembliesPaths, boolean forcePreloadAssemblies);

    public static native void switchToContext (int contextID);

    public static native void destroyContexts (int[] contextIDs);
}
```

src/monodroid/CMakeLists.txt
```diff
@@ -60,7 +60,7 @@ if(NOT ANDROID)
   endforeach()
   set(MONODROID_SOURCES
     ${MONODROID_SOURCES}
-    ${SOURCES_DIR}/inmemory-assemblies.cc
+    ${SOURCES_DIR}/sideloaded-assemblies.cc
```
jonpryor: Should we just call it `designer-assemblies.cc` and be done with it? :-)

src/monodroid/jni/jni-wrappers.hh
```diff
@@ -123,7 +130,8 @@ namespace xamarin::android
 		explicit jstring_array_wrapper (JNIEnv *env) noexcept
```
jonpryor: @grendello: should we just *remove* this constructor?  I don't see it used anywhere, and if it *had* been used, we'd have had a runtime crash because `wrappers` would have been garbage.

src/java-runtime/java/mono/android/Runtime.java
garuma: Fortunately it won't! The designer change that uses the new API is still in PR mode so it's fine to continue iterating on that specific method (the current master version continues using `createNewContext`).

That method needs to be complete because we still require loading all the assemblies up-front when the context is created (adding that last `forcePreloadAssemblies` parameter is a way to disable that potentially in a future when we can do it)

src/monodroid/CMakeLists.txt
garuma: Up to you, if you think it makes it clearer what the purpose is

src/monodroid/jni/jni-wrappers.hh
garuma: My PR [makes uses of it](https://github.com/xamarin/xamarin-android/pull/3734/files#diff-09282c2637d665875eb937b5222b94f7R2077) now to essentially have a `nullptr` array

src/monodroid/CMakeLists.txt
garuma: @jonpryor do you still want that rename?

garuma: renamed

PR #3735, opened 2019-10-03
[Xamarin.Android.Build.Tests] Add some additional helper methods.

dellis1972:
This commit adds some helper methods which will pull
out Task and Target timings from the PerformanceSummary
of a build log.

It also picks up the data from `adb logcat` which tells
us how long it took for the app to start.

These methods will be used in future tests.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/BuildOutput.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/BuildOutput.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Xamarin.Tools.Zip;
 
 namespace Xamarin.ProjectTools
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/BuildOutput.cs
```diff
@@ -74,6 +75,24 @@ public bool IsTargetPartiallyBuilt (string target)
...
 		}
 
+		/// <summary>
+		/// Extracts the timing (in milliseconds) from the Performance Summary.
+		/// </summary>
+		/// <param name="targetOrTask">The Target or Task to get the timing for.</param>
+		/// <returns>The time it took as a TimeSpan. Or Zero if the data could not be found.</returns>
+		public TimeSpan GetTargetOrTaskTime (string targetOrTask)
+		{
+			var regex = new Regex (@"\s+(?<time>\d+)\s+(ms)\s+(" + targetOrTask + @")\s+(?<calls>\d+)\scalls", RegexOptions.Compiled | RegexOptions.ExplicitCapture);
```
jonathanpeppers: I think you should make this a static Regex instance, just for performance.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/BuildOutput.cs
```diff
@@ -1,12 +1,15 @@
...
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using Xamarin.Tools.Zip;
 
 namespace Xamarin.ProjectTools
 {
 	public class BuildOutput
 	{
+		Regex regex = new Regex (@"\s+(?<time>\d+)\s+(ms)\s+(" + targetOrTask + @")\s+(?<calls>\d+)\scalls", RegexOptions.Compiled | RegexOptions.ExplicitCapture);
```
jonathanpeppers: `targetOrTask` here is a compiler error.

This `GetTargetOrTaskTime` seems useful, though!

dellis1972: Yup I moved this RegEx back to the `GetTargetOrTaskTime` method.

PR #3747, opened 2019-10-07
Don't use RTLD_NODELETE flag with dlopen

grendello:
Fixes: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/989404
Context: https://developercommunity.visualstudio.com/content/problem/749635/xamarin-stop-on-load.html

Xamarin.Android runtime used the `RTLD_NODELETE` flag with `dlopen(3)` in order
to make subsequent opens of the same shared library slightly faster (the flag
causes `dlclose(3)` not to remove the mapped DSO from memory and not to
reinitialize its static variables etc on the subsequent call to `dlopen`). The
flag is defined in the NDK without any API level (platform) version restrictions
but, apparently, it fails to work on some API 22 Samsung devices with the
following error:

    F/debug-app-helper( 9171): Failed to dlopen Mono runtime from /data/data/com.companyname.xamarinlearning/files/.__override__/links/libmonosgen-2.0.so: dlopen failed: invalid flags to dlopen: 1003

Remove the flag in order to make our runtime work on the affected devices.

### DIFF
--- a/src/monodroid/jni/debug-app-helper.cc
+++ b/src/monodroid/jni/debug-app-helper.cc
@@ -88,7 +88,7 @@ Java_mono_android_DebugRuntime_init (JNIEnv *env, jclass klass, jobjectArray run
 	}
 
 	char *monosgen_path = get_libmonosgen_path ();
-	void *monosgen = dlopen (monosgen_path, RTLD_LAZY | RTLD_GLOBAL | RTLD_NODELETE);
+	void *monosgen = dlopen (monosgen_path, RTLD_LAZY | RTLD_GLOBAL);
 	if (monosgen == nullptr) {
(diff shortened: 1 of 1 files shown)


pjcollins:
This also occurred in recent [longer running regression test runs against master](https://dev.azure.com/devdiv/DevDiv/_releaseProgress?_a=release-environment-extension&releaseId=469145&environmentId=2323126&extensionId=ms.vss-test-web.test-result-in-release-environment-editor-tab&runId=8875182&resultId=100032&paneView=debug), on an old v4.4.2 x86 Toshiba device: http://xqa.blob.core.windows.net/gist/report-39d30d3c74e44926816c104a112c3035.txt. The failure is only occurring during debug deployments.

dellis1972:
one of the failures is zipalign, so not related.

PR #3748, opened 2019-10-07
[tests] Import regression test for bug 29730

pjcollins:
Context: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3107238&view=ms.vss-test-web.test-result-details
Context: https://bugzilla.xamarin.com/show_bug.cgi?id=29730

This test has been failing with some frequency since it was imported in
the new Integrated Regression test Azure Pipeline jobs. After some
investigation, it appears that the failure is due to test environment
limitations. Some of the Android devices attached to the machine pools
being used for this test job have a lock screen enabled, and in some
cases this is preventing `ViewTreeObserver` events from executing.

Fixes the intermittent failure by moving this test into the
`MSBuildDeviceIntegration` suite which runs on an x86 emulator in CI.
This emulator is never locked and should always display activites for
applications we're installing and launching.

### DIFF
--- a/tests/MSBuildDeviceIntegration/Tests/BugzillaTests.cs
+++ b/tests/MSBuildDeviceIntegration/Tests/BugzillaTests.cs
@@ -18,10 +18,11 @@ namespace Xamarin.Android.Build.Tests
 		[TearDown]
 		public void Teardown ()
 		{
-			if (HasDevices)
+			if (HasDevices && proj != null)
 				RunAdbCommand ($"uninstall {proj.PackageName}");
 
-			if (TestContext.CurrentContext.Result.Outcome.Status == NUnit.Framework.Interfaces.TestStatus.Passed && Directory.Exists (builder.ProjectDirectory))
+			if (TestContext.CurrentContext.Result.Outcome.Status == NUnit.Framework.Interfaces.TestStatus.Passed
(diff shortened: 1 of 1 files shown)


pjcollins:
The test addition failed in Debug mode it seems, but was working for me locally. I'll need to take another look.

pjcollins:
@dellis1972 The test failures on the latest run appear to be unrelated, this should be ready to merge if the latest changes look good to you?

tests/MSBuildDeviceIntegration/Tests/BugzillaTests.cs
```diff
@@ -0,0 +1,61 @@
...
+	[SingleThreaded]
+	public class BugzillaTests : DeviceTest
+	{
+		static ProjectBuilder builder;
+		static XamarinAndroidApplicationProject proj;
+
+		[TearDown]
+		public void Teardown ()
+		{
+			if (HasDevices)
```
dellis1972: @pjcollins this method is going to need the changes from https://github.com/xamarin/xamarin-android/pull/3746/files#diff-05fcc732a1dcabddef8dd1b3ff6bb693R60 as well. 
If we skip the test because no devices are present, both `builder` and `pro` will be `null`

pjcollins: Good call, updated.

PR #3749, opened 2019-10-07
[Xamarin.Android.Build.Tasks] AutoSDK doesn't install Tools or Build Tools

dellis1972:
Fixes https://devdiv.visualstudio.com/DevDiv/_workitems/edit/994496

When we call `GetAndroidDependencies` we expect to get
a list of TaskItems back describing the tooling the
project needs.

However if certain parts of the Android SDK have NOT
been installed we get errors such as

	XA5205: Cannot find aapt.exe. Please install the Android SDK Build-tools package with the C:\Program Files (x86)\Android\android-sdk\tools\android.bat program.

This is because `ResolveAndroidTooling` throws an error. This
is expected behaviour when building an app. However when calling
`GetAndroidDependencies` we dont want it to throw. So lets
add a property which will control the `ContinueOnError` property
of the task. Doing a normal build the error will be reported.
However when calling via the `GetAndroidDependencies` target,
the error will be converted into a warning so that the later
tasks can still complete.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -3279,6 +3279,8 @@ AAMMAAABzYW1wbGUvSGVsbG8uY2xhc3NQSwUGAAAAAAMAAwC9AAAA1gEAAAAA") });
 			var androidSdkPath = Path.Combine (path, "android-sdk");
 			if (createSdkDirectory)
 				Directory.CreateDirectory (androidSdkPath);
+			else if (Directory.Exists (androidSdkPath))
+				Directory.Delete (androidSdkPath, recursive: true);
 			var referencesPath = CreateFauxReferencesDirectory (Path.Combine (path, "xbuild-frameworks"), apis);
 			var proj = new XamarinAndroidApplicationProject () {
 				IsRelease = true,
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -3271,13 +3271,14 @@ AAMMAAABzYW1wbGUvSGVsbG8uY2xhc3NQSwUGAAAAAAMAAwC9AAAA1gEAAAAA") });
 		}
 
 		[Test]
-		public void GetDependencyWhenSDKIsMissingTest ()
+		public void GetDependencyWhenSDKIsMissingTest ([Values (true, false)] bool createSdkDirectory)
 		{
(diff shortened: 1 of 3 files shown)


dellis1972:
Test failure is `DesignTime build should be less than 10000 milliseconds.`. 

So its not related to this PR.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -3241,6 +3241,65 @@ public void IfAndroidJarDoesNotExistThrowXA5207 ()
...
+			}
+		}
+
+		[Test]
+		public void GetDependencyWhenSDKIsMissingTest ()
+		{
+			var apis = new ApiInfo [] {
+			};
+			var path = Path.Combine ("temp", TestName);
+			var androidSdkPath = Path.Combine (path, "android-sdk");
```
jonpryor: What happens if this directory doesn't even exist?  Will it still work?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -834,6 +834,9 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 </Target>
 
 <Target Name="_ResolveSdks" DependsOnTargets="_GetReferenceAssemblyPaths">
+  <PropertyGroup>
+      <_AndroidAllowMissingSdkTooling Condition=" '$(_AndroidAllowMissingSdkTooling)' == '' ">False</_AndroidAllowMissingSdkTooling>
```
jonpryor: 2 space indents.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -3438,7 +3442,13 @@ because xbuild doesn't support framework reference assemblies.
 
 
 <!-- SDK Management Targets -->
-<Target Name="GetAndroidDependencies" DependsOnTargets="_SetLatestTargetFrameworkVersion;$(GetAndroidDependenciesDependsOn)" Returns="@(AndroidDependency)">
+<Target Name="_BeforeGetAndroidDependencies">
+    <PropertyGroup>
```
jonpryor: 2 space element indents.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
dellis1972: I'll add a flag so we don't create the directory and see what happens

dellis1972: It works :D

dellis1972:
Failure is zip align. So Merging.

PR #3759, opened 2019-10-09
[C++ification] Continue converting code to "proper" C++

grendello:
Next commit in the series of making our native code use "proper" C++ constructs,
take advantage of the language's features etc. Changes:
    
  * Fix more compilation warnings arising from differences between platforms
  * Compare pointers against `nullptr` instead of the traditional `!p` C-style check
  * Use the `simple_pointer_guard` class in more places
  * Use references to pointers instead of pointers to pointers in a handful of
    places to decrease the number of null checks
  * Switch nested namespace declarations to C++17 style
  * Make `Util::ends_with` slightly faster when used with literal strings,
    taking advantage of C++ templates
  * Move variable declarations closer to places where they are actually used
  * Move a number of extern "C" functions to `external-api.cc`

### DIFF
--- a/src/monodroid/jni/debug.cc
+++ b/src/monodroid/jni/debug.cc
@@ -57,53 +57,48 @@ using namespace xamarin::android;
 //
 //
 
-// monodroid-glue.c
-extern int process_cmd (int fd, char *cmd);
-namespace xamarin { namespace android
+namespace xamarin::android
 {
 	void* conn_thread (void *arg);
-}}
+}
 
 void
 Debug::monodroid_profiler_load (const char *libmono_path, const char *desc, const char *logfile)
 {
 	const char* col = strchr (desc, ':');
-	char *mname;
+	char *mname_ptr;
 
 	if (col != nullptr) {
 		size_t name_len = static_cast<size_t>(col - desc);
 		size_t alloc_size = ADD_WITH_OVERFLOW_CHECK (size_t, name_len, 1);
-		mname = new char [alloc_size];
-		strncpy (mname, desc, name_len);
-		mname [name_len] = 0;
+		mname_ptr = new char [alloc_size];
+		strncpy (mname_ptr, desc, name_len);
+		mname_ptr [name_len] = 0;
 	} else {
-		mname = utils.strdup_new (desc);
+		mname_ptr = utils.strdup_new (desc);
 	}
+	simple_pointer_guard<char[]> mname (mname_ptr);
 
 	int dlopen_flags = RTLD_LAZY;
-	simple_pointer_guard<char[]> libname (utils.string_concat ("libmono-profiler-", mname, ".so"));
+	simple_pointer_guard<char[]> libname (utils.string_concat ("libmono-profiler-", mname.get (), ".so"));
 	bool found = false;
 	void *handle = androidSystem.load_dso_from_any_directories (libname, dlopen_flags);
 	found = load_profiler_from_handle (handle, desc, mname);
 
 	if (!found && libmono_path != nullptr) {
-		char *full_path = utils.path_combine (libmono_path, libname);
+		simple_pointer_guard<char[]> full_path (utils.path_combine (libmono_path, libname));
 		handle = androidSystem.load_dso (full_path, dlopen_flags, FALSE);
-		delete[] full_path;
 		found = load_profiler_from_handle (handle, desc, mname);
 	}
 
 	if (found && logfile != nullptr)
 		utils.set_world_accessable (logfile);
 
-	if (!found) {
+	if (!found)
 		log_warn (LOG_DEFAULT,
 				"The '%s' profiler wasn't found in the main executable nor could it be loaded from '%s'.",
-				mname,
-				libname.get ());
-	}
-
-	delete[] mname;
+		        mname.get (),
+		        libname.get ());
 }
 
 /* Profiler support cribbed from mono/metadata/profiler.c */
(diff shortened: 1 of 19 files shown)


jonpryor:
[Mono.Android_TestsAppBundle](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3132187&view=ms.vss-test-web.build-test-results-tab) crashed:

```
Unhandled Exception:
System.DllNotFoundException: __Internal assembly:<unknown assembly> type:<unknown type> member:(null)
  at (wrapper managed-to-native) Android.Runtime.JNIEnv.monodroid_timing_start(string)
  at Android.Runtime.JNIEnv.Initialize (Android.Runtime.JnienvInitializeArgs* args) [0x0001c] in <1260048817b5414c97481e16c94f6621>:0 
```

grendello:
> [Mono.Android_TestsAppBundle](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3132187&view=ms.vss-test-web.build-test-results-tab) crashed:
> 
> ```
> Unhandled Exception:
> System.DllNotFoundException: __Internal assembly:<unknown assembly> type:<unknown type> member:(null)
>   at (wrapper managed-to-native) Android.Runtime.JNIEnv.monodroid_timing_start(string)
>   at Android.Runtime.JNIEnv.Initialize (Android.Runtime.JnienvInitializeArgs* args) [0x0001c] in <1260048817b5414c97481e16c94f6621>:0 
> ```

Huh... This is exactly the same code as in master, don't see a reason why it would fail. We'll see if re-run helps.

grendello:
The weird `DllNotFoundException` failure is gone on rebuild. There's one BCL failure:

```
10-11 09:07:48.995  3228  5104 I NUnit   : 1) MonoTests.System.GCTest.TestGetBytesAllocatedForCurrentThread (monodroid_corlib_test)
10-11 09:07:48.995  3228  5104 I NUnit   :   Expected: 219949840
10-11 09:07:48.995  3228  5104 I NUnit   :   But was:  -4075017456
10-11 09:07:48.995  3228  5104 I NUnit   : 
10-11 09:07:48.995  3228  5104 I NUnit   : 
10-11 09:07:48.995  3228  5104 I NUnit   :   at MonoTests.System.GCTest.TestGetBytesAllocatedForCurrentThread () [0x00107] in <5db7dfc9077540ba9da416882d254e81>:0 
10-11 09:07:48.995  3228  5104 I NUnit   :   at (wrapper managed-to-native) System.Reflection.RuntimeMethodInfo.InternalInvoke(System.Reflection.RuntimeMethodInfo,object,object[],System.Exception&)
10-11 09:07:48.995  3228  5104 I NUnit   :   at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <af3ab1ad8108411d85c71d327db0ec0f>:0 
```

and one our recent favorite failure:

```
  /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(3151,2): error MSB6006: "zipalign" exited with code 1. [/Users/vsts/agent/2.158.0/work/1/s/bin/TestRelease/temp/CheckSequencePointGenerationTrueTrueFalseTrueFullFalserelease/UnnamedProject.csproj]

    1 Warning(s)
    1 Error(s)
```

Neither of those appear to be related to this PR.

src/monodroid/jni/debug.cc
```diff
@@ -409,55 +400,68 @@ Debug::handle_server_connection (void)
...
+	if (strcmp (cmd, "exit process") == 0) {
+		log_info (LOG_DEFAULT, "Debugger requested an exit, will exit immediately.\n");
 		fflush (stdout);
 		fflush (stderr);
 		exit (0);
-	} else if (!strncmp (cmd, "start debugger: ", 16)) {
+	}
+
+	bool use_fd = false;
+	if (strncmp (cmd, "start debugger: ", 16) == 0) {
```
jonpryor: Question: *stylistically* would it be better if we *removed* the hardcoded length and instead went with string constants & `sizeof()`?

```cpp
static constexpr char COMMAND_START_DEBUGGER[] = "start debugger: ";
if (strncmp (cmd, COMMAND_START_DEBUGGER, sizeof (COMMAND_START_DEBUGGER)-1) == 0) {
	...
}
```

grendello: Yeah, it would be better (even better with `constexpr size_t COMMAND_START_DEBUGGER_LEN = sizeof(COMMAND_START_DEBUGGER)-1` but it would blow out the code size...)

grendello: but if you don't mind the visual bloat, I'm ok with it :)

jonpryor: I think the "visual bloat" would be a good idea, considering that all of these lines are changing *anyway*..,.

jaykrell:
@grendello 
"• Compare pointers against `nullptr` instead of the traditional `!p` C-style check"

Why?

I think `if (p)` and `if (!p)` are both idiomatic C and even better idiomatic C++.
Because `p` could be a `Win32Handle` and it'll compare against aboth `nullptr` and `INVALID_HANDLE_VALUE` in its `explicit operator bool` (C++11) or its C++98 workaround.

or `p` would be `File` (perhaps from `open`) and `if (p)` would compare against `-1`.
etc.

It means "if valid", not "if not 0".

Issue #3702, opened 2019-09-30
Resource.Id [placeholder title]


brendanzagaeski:
### Background context

*rsgrey commented on Sep 26* <https://github.com/xamarin/xamarin-android/issues/3494#issuecomment-535468674>

> Hello!
> The problem is actual!
> Xamarin.Android SDK 10.0.0.43
> Visual Studio 2019 version 16.3.1

* * *

*brendanzagaeski commented on Sep 26* <https://github.com/xamarin/xamarin-android/issues/3494#issuecomment-535596240>

> rsgrey, note that errors like this can also occur if the Xamarin.Android.Support NuGet packages have not yet been restored successfully into a project. If you are stuck on similar errors, then when you get a chance, you can report a [new issue](https://github.com/xamarin/xamarin-android/issues/new), if possible attaching either a zipped up example project that demonstrates the problem or the [diagnostic MSBuild output](https://docs.microsoft.com/visualstudio/ide/how-to-view-save-and-configure-build-log-files?view=vs-2019#to-change-the-amount-of-information-included-in-the-build-log) from your project, and the team can take a look. Thanks in advance!

* * *

*rsgrey commented on Sep 27* <https://github.com/xamarin/xamarin-android/issues/3494#issuecomment-536010726>

> [log.txt](https://github.com/xamarin/xamarin-android/files/3663743/log.txt)
> 
> Hello.
> I attached a log file

* * *

### Steps to reproduce

(Not yet described.)

### Actual behavior

The build fails due to a mismatch between the resource IDs that are requested in the C# files and the resource IDs that are available in the generated *Resource.designer.cs* file:

```
BaseAbstract.cs(60,43,60,54): error CS0117: 'Resource.Id' does not contain a definition for 'contentView'
BaseAbstract.cs(61,43,61,52): error CS0117: 'Resource.Id' does not contain a definition for 'emptyView'
2View.cs(52,99,52,110): error CS0117: 'Resource.Id' does not contain a definition for 'contentView'
2View.cs(53,91,53,104): error CS0117: 'Resource.Id' does not contain a definition for 'btn_container'
2View.cs(73,99,73,110): error CS0117: 'Resource.Id' does not contain a definition for 'contentView'
2View.cs(74,91,74,104): error CS0117: 'Resource.Id' does not contain a definition for 'btn_container'
2View.cs(94,74,94,82): error CS0117: 'Resource.Id' does not contain a definition for 'btn_next'
[...]

PR #3767, opened 2019-10-11
[Xamarin.Android.Build.Tasks] Add resource file conflict error

brendanzagaeski:
Fixes: https://github.com/xamarin/xamarin-android/issues/3702

A tricky behavior happens when a project references two Android layout
resource files that only differ by their file extensions, like:

  * `Resources\layout\layout1.axml`
  * `Resources\layout\layout1.xml`

To give projects the flexibility to use either extension, the
`<AndroidComputeResPaths/>` intentionally replaces `.axml` with `.xml`
when computing the intermediate output location.  But this means that
both files end up with the same intermediate output location:

	obj\Debug\100\res\layout\layout1.xml

Consequently, the`<CopyIfChanged/>` task in the
`_GenerateAndroidResourceDir` target copies both files to that single
output location, one after the other, overwriting the contents of the
first file with the contents of the second file.

As an example of what a user might see, if the first file defines
resource IDs that don't exist in the second file, this can result in
build errors similar to:

	'Resource.Id' does not contain a definition for 'textView1'

The same kind of conflict can also arise on a case-sensitive file system
if two resource files only differ by capitalization.

To improve the visibility of this complication, adjust
`<AndroidComputeResPaths/>` to emit an error if two input files have the
same computed intermediate output path.

Add a corresponding test as well.  Refactor the existing test in
`AndroidComputeResPathsTests` a little along the way so the tests can
share the setup code.

The performance cost of the new check in `<AndroidComputeResPaths/>`
looks acceptably small.  I timed the task locally across three trials
before and then three trials after the changes, using a scenario with
10,000 input resource file names.

Before: 390 ms, 375 ms, 372 ms. Average: 379 ms
After:  382 ms, 412 ms, 391 ms. Average: 395 ms


src/Xamarin.Android.Build.Tasks/Tasks/AndroidUpdateResDir.cs
```diff
@@ -61,9 +61,9 @@ public class AndroidComputeResPaths : AndroidTask
 
 		public override bool RunTask ()
 		{
-			var intermediateFiles = new List<ITaskItem> ();
-			var resolvedFiles = new List<ITaskItem> ();
-			
+			var intermediateFiles = new Dictionary<string, ITaskItem> (ResourceFiles.Length, StringComparer.Ordinal);
```
jonpryor: Filename comparisons generally use `StringComparer.OrdinalIgnoreCase`, because macOS and Windows are both case-insensitive (by default).

brendanzagaeski:

Add errors and warnings for additional conflict scenarios. Along the way, set up more advanced handling for file system case sensitivity, and take care of the `DuplicateValuesInResourceCaseMap` test failure.

Add an empty commit with a draft of a new combined commit message.

**Questions:**

Does the location of the `DirectoryIsCaseSensitive()` make sense in `Utilities/Files.cs`? Should it be in `FileUtil.cs` or `MonoAndroidHelper.cs` instead? Or maybe it should be a private method in `AndroidComputeResPaths` itself?

src/Xamarin.Android.Build.Tasks/Tasks/AndroidUpdateResDir.cs
```diff
@@ -61,9 +62,10 @@ public class AndroidComputeResPaths : AndroidTask
 
 		public override bool RunTask ()
 		{
-			var intermediateFiles = new List<ITaskItem> ();
-			var resolvedFiles = new List<ITaskItem> ();
-			
+			bool caseSensitive = LowercaseFilenames || (Directory.Exists (IntermediateDir) ? Files.DirectoryIsCaseSensitive (IntermediateDir) : true);
```
brendanzagaeski: I'm inclined to remove `LowercaseFilenames ||` from this line for simplicity.

I originally included it because technically if `LowercaseFilenames` is `true`, then the destination files will all be lowercase, so it doesn't matter if the dictionary uses `StringComparer.Ordinal` or `StringComparer.OrdinalIgnoreCase`, and I believe `StringComparer.Ordinal` has a performance advantage.

But it looks like the performance of `StringComparer.OrdinalIgnoreCase` is fairly comparable in this case, so I think it might be OK to remove the `LowercaseFilenames ||` expression.

Example timings for 10,000 resource names with `StringComparer.Ordinal`:

    385 ms, 376 ms, 379 ms. Average: 380 ms

Example timings for 10,000 resource names `StringComparer.OrdinalIgnoreCase`:

    387 ms, 392 ms, 400 ms. Average: 393 ms

src/Xamarin.Android.Build.Tasks/Utilities/Files.cs
```diff
@@ -179,6 +179,35 @@ public static bool CopyIfZipChanged (string source, string destination)
 			return false;
 		}
 
+		public static bool DirectoryIsCaseSensitive (string fullPath)
```
jonathanpeppers: This method is creating a new directory that is left behind. But that is the best way to test for a case-sensitive file system, I guess?

I don't think we really need to worry about that, can we just assume it is `true`? It would be fine to give the duplicate file name warning on Linux?

I think it would be bad to have duplicate file names on any platform and the developer should fix the project.

dellis1972: I agree I think we should just always check for duplicates

brendanzagaeski: This method should at least in theory remove the directory if it creates one when it runs: https://github.com/xamarin/xamarin-android/blob/45c01ff6aa7810a94b6f778e623080588d260e68/src/Xamarin.Android.Build.Tasks/Utilities/Files.cs#L205-L206

But I agree that if we're OK with showing a warning to users who are using asset file names that are technically valid on a case-sensitive file system (when they only differ by capitalization), then I'm in favor of removing `DirectoryIsCaseSensitive()` for the sake of simplicity.

I'll work on that change.

* * *

I also have to take a look at why several of the build tests are failing with:

```
error APT2000: The system cannot find the file specified.
```

brendanzagaeski:

- As discussed in the last comments on this PR, remove special handling for case-sensitive file systems. This will technically result in extra messages on case-sensitive file systems, but the messages are still likely useful to prevent unintentional duplicate files that only differ by capitalization.

- Rebase.

  - Update all of the error code numbers.
  - Place the error messages in the _.resx_ file.
  - Update the tests to generate empty placeholder files to satisfy the `File.Exists()` check that was added to `AndroidComputeResPaths` for XA2001.

- Reword the message guide pages and expand the release note.

- Update the commit message, including the timing numbers. Note that all the times are longer now, presumably due to the `File.Exists()` check that now runs on all 10,000 input files.

src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
```diff
@@ -133,16 +133,52 @@ public override bool RunTask ()
...
+					string conflictPath = Path.GetFullPath (conflictItemSpec);
+					string path = Path.GetFullPath (item.ItemSpec);
+					if (string.Compare (conflictPath, path, StringComparison.Ordinal) == 0) {
+						Log.LogCodedWarning (
+							"XA1029",
+							file: conflictItemSpec,
+							lineNumber: 0,
+							message: Properties.Resources.XA1029,
+							conflictItemSpec,
+							conflict.GetMetadata ("LogicalName"),
```
jonathanpeppers, line 147: Is `%(LogicalName)` ever blank here? I think we would have to set it--it doesn't seem like MSBuild would fill this in by default.

src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
```diff
@@ -133,16 +133,52 @@ public override bool RunTask ()
...
 				newItem.SetMetadata ("_FlatFile", Monodroid.AndroidResource.CalculateAapt2FlatArchiveFileName (dest));
 				newItem.SetMetadata ("_ArchiveDirectory", AndroidLibraryFlatFilesDirectory);
 				item.CopyMetadataTo (newItem);
-				intermediateFiles.Add (newItem);
+				intermediateFiles.Add (dest, newItem);
 				resolvedFiles.Add (item);
 			}
 
-			IntermediateFiles = intermediateFiles.ToArray ();
+			IntermediateFiles = intermediateFiles.Values.ToArray ();
```
jonathanpeppers, line 181: @dellis1972 would the ordering of `@(AndroidResource)` files matter? Before the ordering was preserved from the ordering in the `.csproj` file, and this will change them to a random-ish order.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/AndroidComputeResPathsTests.cs
```diff
@@ -99,5 +111,212 @@ public void AndroidComputeResPathsTest (string identity, string logicalName, str
...
				/* link2 */     "Resources\\layout-xhdpi\\Layout1.xml",
				/* resources */ true,
			},
+			new object [] {
				/* identity1 */ Path.Combine ("Assets", "asset1.txt"),
				/* identity2 */ Path.Combine ("Assets", "asset3.txt"),
				/* link1 */     "Assets\\asset2.txt",
				/* link2 */     "Assets\\Asset3.txt",
				/* resources */ false,
			},
```
jonathanpeppers, line 239: I see some weird character here:

image

Maybe that is on accident?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/AndroidComputeResPathsTests.cs
```diff
@@ -99,5 +111,212 @@ public void AndroidComputeResPathsTest (string identity, string logicalName, str
...
+			var task = new AndroidComputeResPaths {
+				BuildEngine = engine,
+				ResourceFiles = items.ToArray (),
+				IntermediateDir = intermediateDir,
+				LowercaseFilenames = true
+			};
+
+			var stopwatch = System.Diagnostics.Stopwatch.StartNew ();
+			task.Execute ();
+			Assert.Fail (stopwatch.Elapsed.TotalMilliseconds.ToString ());
```
jonathanpeppers, line 319: This test looks like it would always fail, was this line temporary?

brendanzagaeski: Yes indeed. Somehow I pushed the wrong local testing commit to this branch. Let me fix that.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/AndroidComputeResPathsTests.cs
brendanzagaeski: Ha. That'll show me to try to get the newline character correct in a search and replace in Visual Studio on Windows. Let me fix that too.

brendanzagaeski:

Correct typos.

**TODO:**

At a minimum, look into and remember why I used the `LogicalName` metadata in `conflict.GetMetadata ("LogicalName")` in the XA1029 message when I first set up that message last year. It seems to work as desired based on the tests, but I don't trust that that's a stable feature of the `LogicalName` metadata that should be trusted not to change in the future.

src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
brendanzagaeski: Ohhh. Now I remember why I did this. It's because this warning will always happen on an item that has already been through the full loop, so the following code has necessarily run previously for the `conflict` object:

https://github.com/xamarin/xamarin-android/blob/f1a6e22ffa72655b96766da7ee1d1ce0ae2f5008/src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs#L137

So `LogicalName` will always contain the value of `rel` for each item, where `rel` has the carefully set up non-empty value as prepared at the beginning of the method.

brendanzagaeski:
> **TODO:**
>
> At a minimum, look into and remember why I used the `LogicalName` metadata in `conflict.GetMetadata ("LogicalName")` in the XA1029 message when I first set up that message last year. It seems to work as desired based on the tests, but I don't trust that that's a stable feature of the `LogicalName` metadata that should be trusted not to change in the future. 

Now I remember why I did this. It's because this warning will always happen on an item that has already been through the full loop, so the following code has necessarily run previously for the `conflict` object:

https://github.com/xamarin/xamarin-android/blob/f1a6e22ffa72655b96766da7ee1d1ce0ae2f5008/src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs#L137

So `LogicalName` will always contain the value of `rel` for each item, where `rel` has the carefully set up non-empty value as prepared at the beginning of the method.

src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
dellis1972: It shouldn't matter for those within the app project. The ordering only really matters when you do the final "link" stage. That is where the order of app resources vs library resources really matters.

dellis1972 (approved):
This looks ok. Would be nice to have it green though, it might need rebasing.

brendanzagaeski:
Oh right, there was a fix (8d0ca1cc0769e82505427bd265beed88f2e9ed28) to address the build issue the macOS build was hitting. I'll rebase this PR and see if that lets the macOS build pass!

brendanzagaeski:
Mildly interesting side note: Restarting the macOS job without rebasing didn't work because the _TestConditions_ step of the macOS job was still looking for the old merge commit from the previous build attempt. That might be a useful result anyway because it avoids potential confusing results where retried steps of a PR build could be run again a different merge commit than the other steps of the build.

brendanzagaeski:
Ah ha. Now that the tests ran on macOS, they revealed that I forgot to adjust the path separators in the tests to use `Path.Combine()`. I'll fix that.

Maybe the change to make ProGuard _mapping.txt_ opt-in will also land before I push the next commit so that those tests will pass on the next run too.

brendanzagaeski:

Use `Path.Combine()` for all of the `Link` metadata in the tests to make it relatively straightforward to have the tests pass on both macOS and Windows.

PR #3780, opened 2019-10-14
A handful of timing changes

grendello:
This commit introduces a new public class, `Android.Runtime.Timing` which can be
used by the managed code to efficiently and accurately (to the nanosecond level)
measure code performance. At the moment the class is used only by our
Xamarin.Forms integration test.

Additionally, this commit removes support for namespaced system properties (that
is properties suffixed by the package name) which were added for the benefit of
the IDEs but never seen any use. Namespaced properties require two system
property retrieval calls and so they cost extra time (not much, around 25k
nanoseconds but taken together they can contribute to >1ms of startup time with
no real purpose).

### DIFF
--- a/src/Mono.Android/Android.Runtime/JNIEnv.cs
+++ b/src/Mono.Android/Android.Runtime/JNIEnv.cs
@@ -61,10 +61,10 @@ namespace Android.Runtime {
 		extern static void monodroid_log (LogLevel level, LogCategories category, string message);
 
 		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
-		extern static IntPtr monodroid_timing_start (string message);
+		internal extern static IntPtr monodroid_timing_start (string message);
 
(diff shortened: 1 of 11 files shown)


jonpryor:
Please mention how much time is saved by *not* doing the namespace lookup.

src/Mono.Android/Android.Runtime/Timing.cs
```diff
@@ -0,0 +1,59 @@
+using System;
+
+namespace Android.Runtime
+{
+	public class Timing : IDisposable
```
jonpryor: Should we have doc-comments on this?

I'm also not thrilled with `Timing`, because the class *itself* doesn't provide timing info.  Instead, the timing info is written to `adb logcat`.

I can't think of a better name either. `LoggingTimer`, perhaps?

grendello:
> Please mention how much time is saved by _not_ doing the namespace lookup.

The commit message says that :) It's around 25000ns

Issue #3751, opened 2019-10-07
ILifecycleObserver issue with d8 and multi-dex


tcoxMonkeyCoder:
### Steps to Reproduce

1. Add a class that implements ILifecycleObserver
2. Add some methods and add the Lifecycle.Event and Export attributes
3. Add the class to a lifecycle owner
4. Enable Multi-Dex
5. Set Dex Compiler to d8

[lifecycleowner.zip](https://github.com/xamarin/xamarin-android/files/3699921/lifecycleowner.zip)

### Expected Behavior
The methods with the Lifecycle.Event attributes are called
### Actual Behavior
The methods with the Lifecycle.Event attributes are not called.
It's not a problem on older Xamarin.Android versions.
### Version Information

Visual Studio Community 2019 for Mac
Version 8.3.1 (build 18)
Installation UUID: fab306c2-c3e7-43bb-89f8-ff320f51bd8b
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 5.16.1.24 (d16-3 / 08809f5b)

	Package version: 604000198

Mono Framework MDK
Runtime:
	Mono 6.4.0.198 (2019-06/fe64a4765e6) (64-bit)
	Package version: 604000198

NuGet
Version: 5.3.0.6192

.NET Core SDK
SDK: /usr/local/share/dotnet/sdk/3.0.100/Sdks
SDK Versions:
	3.0.100
	2.1.701
	2.1.700
	2.1.302
	2.1.4
	2.0.3
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/6.4.0/lib/mono/msbuild/Current/bin/Sdks

.NET Core Runtime
Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	3.0.0
	2.1.13
	2.1.12
	2.1.11
	2.1.2
	2.0.5
	2.0.3

Xamarin.Profiler
'/Applications/Xamarin Profiler.app' not found

Updater
Version: 11

Apple Developer Tools
Xcode 10.1 (14460.46)
Build 10B61

Xamarin.Mac
Xamarin.Mac not installed. Can't find /Library/Frameworks/Xamarin.Mac.framework/Versions/Current/Version.

Xamarin.iOS
Version: 13.2.0.42 (Visual Studio Community)
Hash: 5e8a208b
Branch: d16-3
Build date: 2019-09-17 14:19:30-0400

Xamarin Designer
Version: 16.3.0.244
[...]

PR #3787, opened 2019-10-14
[Xamarin.Android.Build.Tasks] r8 should not optimize for multi-dex only

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/3751

When enabling multi-dex, but *not* code shrinking, such as:

    <AndroidDexTool>d8</AndroidDexTool>
    <AndroidLinkTool></AndroidLinkTool>
    <AndroidEnableMultiDex>True</AndroidEnableMultiDex>

In this configuration, we run r8 instead of d8, because it fully
supports multi-dex. r8 automatically calculates what needs to go in
the main dex file, which d8 doesn't do.

We pass these flags to r8 in this mode:

    --no-tree-shaking --no-minification

However, I noticed a few problems in the dex file I analyzed with:

    dexdump -d obj\Debug\android\bin\classes.dex > classes.txt

* The dex code appears to be optimized. The code within methods is
  shortened/scrambled.
* Annotations are missing???

`dexdump` does not appear to list annotations, but I could see them
missing from viewing the `classes.dex` file in Android Studio and the
file size was smaller.

I discovered adding this to the r8 command solved the issue:

    --pg-conf "C:\Program Files (x86)\Android\android-sdk\tools\proguard\proguard-android.txt"

Debugging further, specifically these rules solve the issue:

    -dontoptimize
    -dontpreverify
    -keepattributes *Annotation*

The dropped annotations specifically broke #3751, but I suspect other
*weird* issues would occur from the optimized code? If no code
shrinker is used, we don't want to optimize anything!

I am unsure how to add a test here, as `dexdump` (or `DexUtils`)
doesn't give us a way see that the annotations got stripped or the dex
code is optimized? *Open to ideas*

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/R8.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/R8.cs
@@ -35,16 +35,16 @@ namespace Xamarin.Android.Tasks
 
 		protected override string MainClass => "com.android.tools.r8.R8";
 
-		string temp;
+		readonly List<string> tempFiles = new List<string> ();
 
 		public override bool RunTask ()
 		{
 			try {
-				temp = Path.GetTempFileName ();
 				return base.RunTask ();
 			} finally {
-				if (!string.IsNullOrEmpty (temp))
+				foreach (var temp in tempFiles) {
 					File.Delete (temp);
+				}
 			}
 		}
 
@@ -61,6 +61,8 @@ namespace Xamarin.Android.Tasks
 					Log.LogCodedWarning ("XA4305", $"MultiDex is enabled, but '{nameof (MultiDexMainDexListFile)}' was not specified.");
 				} else {
 					var content = new List<string> ();
+					var temp = Path.GetTempFileName ();
+					tempFiles.Add (temp);
 					if (CustomMainDexListFiles != null) {
 						foreach (var file in CustomMainDexListFiles) {
 							if (File.Exists (file.ItemSpec)) {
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/R8.cs
```diff
@@ -117,6 +119,15 @@ protected override CommandLineBuilder GetCommandLineBuilder ()
 				//NOTE: we may be calling r8 *only* for multi-dex, and all shrinking is disabled
 				cmd.AppendSwitch ("--no-tree-shaking");
 				cmd.AppendSwitch ("--no-minification");
+				// Rules to turn off optimizations
+				var temp = Path.GetTempFileName ();
+				File.WriteAllLines (temp, new [] {
+					"-dontoptimize",
```
dellis1972: can I assume that `r8` doesn't care if you have duplicate entries? What does it do about conflicting ones? Could we get in to a situation where a User ha a custom config, which turns on optimisations, but we then turn it off again?

jonathanpeppers: In this case (multi-dex only), we don't use any of the user's proguard rules at all.

But there might be an issue when you are using a "Code Shrinker".

In the case `AndroidLinkTool=r8`, we do pass:

    --pg-conf " C:\Program Files (x86)\Android\android-sdk\tools\proguard\proguard-android.txt"

And this file is probably going to trump anything the user supplies. But we have the same code for r8 & proguard:

https://github.com/xamarin/xamarin-android/blob/431329d38137ce47c2d92f76509b84ef8567c33d/src/Xamarin.Android.Build.Tasks/Tasks/Proguard.cs#L127-L135

https://github.com/xamarin/xamarin-android/blob/431329d38137ce47c2d92f76509b84ef8567c33d/src/Xamarin.Android.Build.Tasks/Tasks/R8.cs#L100-L108

In both cases this is first:

https://github.com/xamarin/xamarin-android/blob/431329d38137ce47c2d92f76509b84ef8567c33d/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L319

Another issue is the header of this proguard file:

```
# This is a configuration file for ProGuard.
# http://proguard.sourceforge.net/index.html#manual/usage.html
#
# This file is no longer maintained and is not used by new (2.2+) versions of the
# Android plugin for Gradle. Instead, the Android plugin for Gradle generates the
# default rules at build time and stores them in the build directory.
```

So I should file an issue to look into this at some point.

jonathanpeppers:
Only failures are `zipalign` related, so merging.

PR #3790, opened 2019-10-15
[build] on Windows, kill processes before zipping test results

jonathanpeppers:
Context: https://build.azdo.io/3142932

Builds have been occasionally failing with:

    build-tools\Xamarin.Android.Tools.BootstrapTasks\result-packaging.targets(79,5): Error MSB4018: The "Zip" task failed unexpectedly.
    System.IO.IOException: The process cannot access the file 'E:\A\_work\214\a\xa-test-results-10.1.99.28_62cd0ab-Windows-Release\temp\XA5213IsRaisedWhenOutOfMemoryErrorIsThrown\build.log' because it is being used by another process.

We are using a "kill lingering process" task before the build, but I
think we should also use it at the end of the build in
`upload-results.yaml`.


pjcollins:
@jonathanpeppers Unfortunately I don't think this will work as expected. The extension as written hooks into "before" and "after" build events (running before and after all other user defined steps in the YAML). Process info is collected at the start of the job, and then re-evaluated at the end to determine which processes to attempt to kill after all YAML steps run. These pre and post build actions execute in this manner regardless of where the task is "called" from YAML.

The source can be found here: 
https://github.com/xamarin/azdev-extension/blob/master/lingering-process-task/task.json#L24
and 
https://github.com/xamarin/azdev-extension/blob/master/lingering-process-task/postcollectlingeringprocs.ts#L17

We may need to write our own process kill script, or otherwise build it into `result-packaging.targets`.

jonathanpeppers:
After looking at the other builds more, it seems like it happens if the tests timeout -- which I guess makes sense?

PR #3798, opened 2019-10-16
saa

Pikachu443:


jonpryor:
Why the whitespace changes?  Is there a rationale?

PR #3800, opened 2019-10-16
[Xamarin.Android.Build.Tasks] Produce a better ZipAlign Error.

dellis1972:
The standard behaviour of the `ToolTask` is to write
any `stderr` output out as a message. It then logs
a generic "Tool existed with exitcode X" error. This
is not great as the end user then has to dig through
the logs to find the actual problem.

The `ToolTask` does have a property `LogStandardErrorAsError`
which defaults to `false`. When set to `true` any `stderr`
output will be logged as an error. However there is no way
in the `ToolTask` to provide an `ErrorCode` for those errors.
You just get a blank `ErrorCode`. So while that might seem
like a good solution, its not what we are after.

So in this case we want an `ANDZA0000` errode code. Looking
at the source code for `ToolTask` [1] the best way to do
this is to intercept the message logging in the
`LogEventsFromTextOutput`. Fortunately the `ExitCode` for the
app is set by the time we enter the `LogEventsFromTextOutput`
method. So we can just check the `ExitCode` and then log
any line we get as an Error with our customer `ErrorCode`.

[1] https://github.com/Microsoft/msbuild/blob/master/src/Utilities/ToolTask.cs#L1057

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/AndroidZipAlign.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/AndroidZipAlign.cs
@@ -1,21 +1,12 @@
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Linq;
-using System.Text;
-
 using Microsoft.Build.Framework;
-using Microsoft.Build.Utilities;
 
 namespace Xamarin.Android.Tasks
 {
 	public class AndroidZipAlign : AndroidRunToolTask
 	{
 		public override string TaskPrefix => "AZA";
-
-		public AndroidZipAlign ()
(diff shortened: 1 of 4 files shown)


jonpryor:
Re-ran: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3150201&view=results

jonpryor:
[First run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3147832&view=results) didn't have any relevant failures.

Re-ran the build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3150208&view=results

PR #3801, opened 2019-10-16
[Xamarin.Android.Build.Tasks] Add _Run Target for OSS.

dellis1972:
Add a helper target for users of the open source
version of Xamarin.Android. With this target users
will be able to start the android application
from msbuild.

	msbuild foo.csproj /t:_Run

By default the target will look up the application
to run from the `AndroidManifest.xml`. However if
you want to start a different app you can override
the `_LaunchAcivity` MSBuild property.

	msbuild foo.csproh /t:_Run /p:_LaunchAcivity=something.app

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Device.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Device.targets
@@ -1,0 +1,27 @@
+<!--
+***********************************************************************************************
+Xamarin.Android.Device.targets
+
+WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and have
+  created a backup copy.  Incorrect changes to this file will make it
+  impossible to load or build your projects from the command-line or the IDE.
+
+This file imports the version- and platform-specific targets for the project importing
+this file. This file also defines targets to produce an error if the specified targets
+file does not exist, but the project is built anyway (command-line or IDE build).
+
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Device.targets
```diff
@@ -0,0 +1,27 @@
+<!--
+***********************************************************************************************
+Xamarin.Android.Device.targets
```
jonathanpeppers: We should add this to the installer but when `'$(_HasCommercialFiles)' != 'True'`

https://github.com/xamarin/xamarin-android/blob/008ba983fa38305284f55be9e8ea0d520d99e992/build-tools/installers/create-installers.targets#L341

So you would get the file in OSS vsix or pkgs, but not the commercial one.

jonpryor: @jonathanpeppers: *why* do we want to make this file conditional on OSS vs. commercial?  I think we'll want to migrate e.g. GDB/lldb support here in the future as well, and that would be a *move*, not a "copy", and we'd *certainly* need to have this new file included in *all* installers at that time.

So why not just include it now?

jonathanpeppers: As long as `_Run` is replaced and the Debugging targets are imported last, that would be ok.

Issue #3740, opened 2019-10-04
Generate default AOT startup profile(s) as part of CI builds


radekdoulik:
And put them in the build artifacts, so that they are easily available for optional update before the release.

PR #3805, opened 2019-10-16
[Xamarin.Android.Build.Tasks] Add Run target

radekdoulik:
Add new `Run` target to start the app activity on the device/emulator

Add new `GetAndroidActivityName` task to retrieve the activity name
from the manifest file

Add new `Xamarin.Android.Application.targets` file for application
specific targets and import it conditionaly into the
`Xamarin.Android.Common.targets`.

### DIFF
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -100,6 +100,10 @@ The following build targets are defined for Xamarin.Android projects:
     (`.apk`). Use with `/p:Configuration=Release` to generate
     self-contained "Release" packages.
 
+-   **StartAndroidActivity** &ndash; Starts *launch* activity on the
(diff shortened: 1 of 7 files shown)


radekdoulik:
Didn't add test for it yet. I plan to use it in the https://github.com/xamarin/xamarin-android/issues/3740 implementation, which will run on CI and should serve as a good test for that.

jonathanpeppers:
Is this the same as: https://github.com/xamarin/xamarin-android/pull/3801

You both happened to implement the same thing?

dellis1972:
🤣

radekdoulik:
the only failing build has the upload error => ready to go

(##[error]Artifact Test Results - MSBuild - macOS already exists for build 3235725.)

PR #3806, opened 2019-10-16
[tests] MSBuild performance regression tests

jonathanpeppers:
A new set of tests to catch regressions in build performance. They
would likely catch a bug where we accidentally caused MSBuild targets
to run when they shouldn't.

In a previous build, these tests passed while using `[Repeat(10)]`:

https://build.azdo.io/3193441

Hopefully, this means they will be stable (not flaky) going forward.

Cases we are now testing:

- Build with no changes
- Build with C# change
- Build with AndroidResource change
- Build with Android designer change (call designer's MSBuild targets)
- Build with XAML change & *no* reference assemblies
- Build with XAML change & reference assemblies
- Build with new `Java.Lang.Object`
- Build with `csproj` change
- Install with C# change
- Install with XAML change & reference assemblies

These run on on the MSBuild Device tests on Mac. I gave about 250ms of
leeway, but I was seeing 2 or 3 times better numbers locally.

### DIFF
--- a/tests/MSBuildDeviceIntegration/Tests/PerformanceTest.cs
+++ b/tests/MSBuildDeviceIntegration/Tests/PerformanceTest.cs
@@ -10,20 +10,23 @@ namespace Xamarin.Android.Build.Tests
 	[TestFixture, NonParallelizable]
 	public class PerformanceTest : DeviceTest
 	{
-		readonly Dictionary<string, int> csv_values = new Dictionary<string, int> ();
+		static readonly Dictionary<string, int> csv_values = new Dictionary<string, int> ();
 
-		[SetUp]
-		public void Setup ()
+		[OneTimeSetUp]
+		public static void Setup ()
 		{
 			var csv = Path.Combine (XABuildPaths.TopDirectory, "tests", "msbuild-times-reference", "MSBuildDeviceIntegration.csv");
 			using (var reader = File.OpenText (csv)) {
-				var keys = reader.ReadLine ().Split (',');
-				var values = reader.ReadLine ().Split (',');
-				Assert.AreEqual (keys.Length, values.Length, $"{csv} is not a valid CSV file.");
-				for (int i = 0; i < values.Length; i++) {
-					string text = values [i];
+				while (!reader.EndOfStream) {
+					var line = reader.ReadLine ();
+					if (line.StartsWith ("#") || string.IsNullOrWhiteSpace (line)) {
+						continue;
+					}
+					var split = line.Split (',');
+					Assert.AreEqual (2, split.Length, $"{csv} should have two entries per line.");
+					string text = split [1];
 					if (int.TryParse (text, out int value)) {
-						csv_values [keys [i]] = value;
+						csv_values [split [0]] = value;
 					} else {
 						Assert.Fail ($"'{text}' is not a valid integer!");
 					}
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
Hmm...
```
Build_No_Changes
Exceeded expected time of 1000ms, actual 3070ms
Build_CSharp_Change
Exceeded expected time of 2500ms, actual 5270ms
Install_CSharp_Change
Exceeded expected time of 3000ms, actual 5710ms
```
It looks like the `[Repeat(10)]` wasn't even used. I guess I bump the numbers and see what happens?

jonathanpeppers:
The latest numbers:
```diff
Build_No_Changes,Build_CSharp_Change,Install_CSharp_Change
-1000,2500,3000 
+3250,5500,6000
```
Only 1 failed:
```
Install_CSharp_Change
Exceeded expected time of 6000ms, actual 6060ms
```
So I think this idea will work, time to write more tests.

jonathanpeppers:
The CSV is to match what radek did for similar tests for APK sizes: https://github.com/xamarin/xamarin-android/commit/e8b9ee21d2fb93fdfa0457da4283efdb7f2a9c84

jonp thought we should be able to adjust the times without rebuilding anything.

tests/MSBuildDeviceIntegration/Tests/PerformanceTest.cs
```diff
@@ -0,0 +1,273 @@
...
+		[SetUp]
+		public void Setup ()
+		{
+			var csv = Path.Combine (XABuildPaths.TopDirectory, "tests", "msbuild-times-reference", "MSBuildDeviceIntegration.csv");
+			using (var reader = File.OpenText (csv)) {
+				var keys = reader.ReadLine ().Split (',');
+				var values = reader.ReadLine ().Split (',');
+				Assert.AreEqual (keys.Length, values.Length, $"{csv} is not a valid CSV file.");
+				for (int i = 0; i < values.Length; i++) {
+					int.TryParse (values [i], out int value);
```
jonpryor: Should this error out loudly on non-integral values?

tests/msbuild-times-reference/MSBuildDeviceIntegration.csv
```diff
@@ -0,0 +1,2 @@
+Build_No_Changes,Build_CSharp_Change,Build_AndroidResource_Change,Build_Designer_Change,Build_JLO_Change,Build_CSProj_Change,Build_XAML_Change,Build_XAML_Change_RefAssembly,Install_CSharp_Change,Install_XAML_Change
+3250,4000,4750,4250,9000,9500,9500,6250,6000,7500
```
jonpryor: I realize this would introduce a "file format incompatibility" with `tests/apk-sizes-reference`, but this approach will result in *really long lines*, and makes it difficult to understand what time goes with which test.

Could we instead change the file format to:

```
# Comments and blank lines are ignored
# First non-comment row is human description of columns
Test Name,Test Time in ms (int)
# Data
Build_No_Changes,3250
Build_CSharp_Change,4000
...
```

`Setup()` gets nominally more complicated, but not *significantly* so:

```
class CSVParser {
	public static Dictionary<string, int> Parse (string file)
	{
		var contents = new Dictionary<string, int>();
		using (var reader = File.OpenText (file)) {
			string line;
			bool inData = false;
			while ((line = reader.ReadLine ()) != null) {
				line = line.Trim ();
				if (string.IsNullOrEmpty (line) || line.StartsWith ("#"))
					continue;
				// Special-case first line
				if (!inData) {
					var names = line.Split (',');
					if (names.Length != 2) {
						throw new ArgumentException ($"Found unexpected header `{line}`.  Expected 2 values, found {names.Length}!");
					}
					inData = true;
					continue;
				}
				var entry = line.Split (new[]{','}, 2);
				contents [entry [0]] = int.Parse (entry [1]);
			}
		}
		return contents;
	}
}
```

jonpryor: Or we might want to use `=` on each line, instead of `,`...

Issue #3802, opened 2019-10-16
Error: 'tid_type gettid()' was first declared as ‘extern’ and then declared as 'static’


TheForgetTime:
```
 ../../../jni/osbridge.cc:78:17: 错误：‘tid_type gettid()’先被声明为‘extern’后又被声明为‘static’ [-fpermissive]
     78 | static tid_type gettid ()
        |                 ^~~~~~
  In file included from /usr/include/unistd.h:1170,
                   from ../../../jni/util.hh:26,
                   from ../../../jni/globals.hh:5,
                   from ../../../jni/osbridge.cc:22:
  /usr/include/bits/unistd_ext.h:34:16: 附注：previous declaration of ‘__pid_t gettid()’
     34 | extern __pid_t gettid (void) __THROW;
        |                ^~~~~~
```

---
OS: Linux
Version: Arch rolling
GCC Version: 9.2.0
Mono Version: 6.0.0

TheForgetTime:
The contents of the file  "/usr/include/bits/unistd_ext.h"
---
```c
/* System-specific extensions of <unistd.h>, Linux version.
   Copyright (C) 2019 Free Software Foundation, Inc.
   This file is part of the GNU C Library.

   The GNU C Library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
   version 2.1 of the License, or (at your option) any later version.

   The GNU C Library is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
   Lesser General Public License for more details.

   You should have received a copy of the GNU Lesser General Public
   License along with the GNU C Library; if not, see
   <http://www.gnu.org/licenses/>.  */

#ifndef _UNISTD_H
# error "Never include <bits/unistd_ext.h> directly; use <unistd.h> instead."
#endif

#ifdef __USE_GNU

/* Return the kernel thread ID (TID) of the current thread.  The
   returned value is not subject to caching.  Most Linux system calls
   accept a TID in place of a PID.  Using the TID to change properties
   of a thread that has been created using pthread_create can lead to
   undefined behavior (comparable to manipulating file descriptors
   directly that have not been created explicitly).  Note that a TID
   uniquely identifies a thread only while this thread is running; a
   TID can be reused once a thread has exited, even if the thread is
[...]

PR #3809, opened 2019-10-17
Try to detect the `gettid` function on Linux hosts

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/3802

Most Linux distributions don't have a wrapper for the `gettid` syscall in the
standard C library. It appears that distributions with newer glibc define it in
`unistd_ext.h` header included by `unistd.h`. Add a symbol check to avoid build
errors on those distributions which do define `gettid` in the system headers.


grendello:
Despite the errors I'm going to merge this commit. The change in it affects only macOS and Linux *libmonodroid* builds, and that part worked fine.

PR #3816, opened 2019-10-17
[build] Enable building Mono archive from sources

pjcollins:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3153963

Introduces a new `CloneMonoIfNeeded` step to xaprepare to ensure mono
sources are pulled down and updated before attempting to build. This
will only occur when a Mono archive is not available.


pjcollins:
Current build is failing due to missing Android Windows BCL components. The mono team builds this archive on Windows. Some options to move this forward:

1) Introduce a new conditional Windows prepare job that runs before Mac prepare and produces these components.
2) Ignore these components in our build and installers when building Mono ourselves, though this wouldn't produce shippable artifacts...

jonpryor:
If we want to "support" this, we need to limit the scope as much as possible.  Without mono-side changes -- which don't appear to be coming, or even possible? -- it is not possible to get a Windows BCL from a macOS/Linux mono source build.  Thus, if we want to support using a mono submodule, we'd have to:

 1. Make the Windows BCL *optional*, along with any other outputs which the mono sdks build doesn't produce, and
 2. Figure out "what kind" of installer this is.

This is where having a "NuGet package *graph*" would be handy: such an environment could produce the "common/core" bits and *skip* the Windows-specific bits.  (We still are not sure if this is possible, but it sounds good on paper!)

I don't think we want it to be possible to produce a `.vsix` with "missing" contents, and thus `create-vsix` *should* fail.

Additionally, I think the `Step_CloneMonoIfNeeded` step would need to go further, and not only clone mono, but *build it as well*, otherwise we'll still fail, as e.g. `libmonosgen-2.0.so` won't exist for use.  We'd *have* to build the mono archive!

pjcollins:
Agreed. I was hoping to restore usage of https://github.com/xamarin/xamarin-android/blob/master/build-tools/xaprepare/xaprepare/Steps/Step_BuildMonoRuntimes.Unix.cs to at least improve the local build scenario on macOS in this PR. Once that work is complete we can potentially bring this in as a piece-wise improvement, even though CI builds and commercial artifact creation would continue to fail when a mono archive is missing.

pjcollins:
I'm going to close this, the value prospect is no longer very high given the reduced frequency in mono changes coming in.

PR #3839, opened 2019-10-24
Disable archive verification in xaprepare

grendello:
In order to make builds faster, we're going to remove downloaded archive
verification step. The way the verification works is basically unpack the
archive without writing its contents to disk, merely calculating the checksum.
In effect, the time it takes to verify is just slightly less than it takes to
actually unpack the archive. Removing the verification step appears to give us
~6 minutes back on Azure builds for the **prepare** target and just above 1 minute 
for the **prepare -> bundle** stage.

### DIFF
--- a/build-tools/xaprepare/xaprepare/ToolRunners/SevenZipRunner.cs
+++ b/build-tools/xaprepare/xaprepare/ToolRunners/SevenZipRunner.cs
@@ -48,30 +48,6 @@ namespace Xamarin.Android.Prepare
 			}
 		}
 
-		public async Task<bool> VerifyArchive (string archivePath)
-		{
-			if (String.IsNullOrEmpty (archivePath))
-				throw new ArgumentException ("must not be null or empty", nameof (archivePath));
-
-			ProcessRunner runner = CreateProcessRunner ("t");
-			AddStandardArguments (runner);
-			runner.AddQuotedArgument (archivePath);
-
-			try {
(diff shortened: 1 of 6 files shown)


jonpryor:
The time savings may be somewhat overstated:

New: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3178956&view=results
Master: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3177710&view=results

Don't look at the **Prepare** time; look at the **Prepare** > **Bundle** time.  It drops from 12min 51s on master to 11min 44s on this PR, so slightly more than a minute.

I had thought/hoped it would be more.

build-tools/xaprepare/xaprepare/ToolRunners/SevenZipRunner.cs
```diff
@@ -50,26 +50,28 @@ public async Task<bool> Extract (string archivePath, string outputDirectory, Lis
...
-							return runner.Run ();
-						}
-					}
-				);
-			} finally {
-				StopTwiddler ();
-			}
+			return true;
+
+			// if (String.IsNullOrEmpty (archivePath))
```
jonpryor: Should we be commenting this out, or should we instead use `#if false` around the code?  Using the `#if` would result in less "noise" in the PR, but might be undesirable otherwise...

PR #3842, opened 2019-10-25
[xaprepare] Remove extra bundle creation

pjcollins:
Our bundle now only contains content coming from the Mono Archive, so
there is no longer a need to archive and host this content ourselves.
Removes the majority of Bundle related logic from xaprepare, and updates
the core scenario to ensure that Mac and Windows Mono Archive download
and extraction also occurs on Windows.


build-tools/xaprepare/xaprepare/Steps/Step_DownloadMonoArchive.cs
```diff
@@ -36,19 +36,6 @@ async Task<bool> DownloadMonoArchive (Context context)
 				return false;
 			}
 
-			Log.StatusLine ("Checking if all runtime files are present");
-			var allRuntimes = new Runtimes ();
-			if (MonoRuntimesHelpers.AllBundleItemsPresent (allRuntimes)) {
```
grendello: We might actually want to keep this check, just rename it. It would check if all the **archive** files are present to decide whether we need to download or reinstall it.

pjcollins: I've updated this check and also added it to the new InstallMonoRuntimes step.

grendello:
It seems the failed tests are not related to this PR, I think this is ready to merge.

grendello:
> I tried this locally, and I am not able to use AOT on Windows after this.
> 
> I think we _have_ the files, can we just not `strip` them? @grendello what do you think?

I seem to recall there were problems with stripped Windows `.exe` and `.dll` files. They mustn't have their symbols stripped, just debug info.

Issue #3849, opened 2019-10-26
Hard to debug errors in resolving ssembly


kant2002:
### Steps to Reproduce

1. git clone https://github.com/kant2002/XamringGenericHostNetCoreIssue/
2. build in VS

### Expected Behavior

I want that build log show information about offending assembly

### Actual Behavior

```
warning XA0107: Ignoring  as it is a Reference Assembly
```
Note that no assembly name is printed.

### Version Information

```
Xamarin.Android Reference Assemblies and MSBuild support.
    Mono: mono/mono/2019-06@5608fe0abb3
    Java.Interop: xamarin/java.interop/d16-3@5836f58
    LibZipSharp: grendello/LibZipSharp/d16-3@71f4a94
    LibZip: nih-at/libzip/rel-1-5-1@b95cf3fd
    ProGuard: xamarin/proguard/master@905836d
    SQLite: xamarin/sqlite/3.27.1@8212a2d
    Xamarin.Android Tools: xamarin/xamarin-android-tools/d16-3@cb41333
```

Look like this is happens on following lines: https://github.com/xamarin/xamarin-android/blob/fea32f64c58e6942785d75dddfc28c51c3f524a3/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAssemblies.cs#L92-L93

when `resolved_assembly == null` condition is met. Maybe it would be practical to at least print `assembly.ItemSpec` ?

brendanzagaeski:
Thanks for this issue report and the fix!

*Release status update*

A new Preview version has now been published on Windows that includes the fix for this item. The fix is not yet included in a Release version. I will update this item again when a Release version is available that includes the fix.

The fix is not yet available on macOS. I will update this item again when a Preview version with the fix is available on macOS.

Fix included in Xamarin.Android 10.2.0.16.

Fix included on Windows in Visual Studio 2019 version 16.5 Preview 1. To try the Preview version that includes the fix, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) in [Visual Studio Preview](https://visualstudio.microsoft.com/vs/preview/).

Fix not yet available on macOS.

brendanzagaeski:
*Release status update*

A new Preview version has now been published on macOS that includes the fix for this item.

Fix included in Xamarin.Android 10.2.0.84.

[...]

PR #3850, opened 2019-10-26
Print assembly which was not properly resolved

kant2002:
See #3849

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAssemblies.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAssemblies.cs
@@ -90,7 +90,7 @@ namespace Xamarin.Android.Tasks
 						if (lockFile != null)
 							resolved_assembly = ResolveRuntimeAssemblyForReferenceAssembly (lockFile, assembly.ItemSpec);
 						if (lockFile == null || resolved_assembly == null) {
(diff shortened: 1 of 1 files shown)


dellis1972:
Thanks for the contribution. I think this looks like a reasonable thing to do in this case.

src/Xamarin.Android.Build.Tasks/Tasks/ResolveAssemblies.cs
```diff
@@ -90,7 +90,7 @@ void Execute (MetadataResolver resolver)
 						if (lockFile != null)
 							resolved_assembly = ResolveRuntimeAssemblyForReferenceAssembly (lockFile, assembly.ItemSpec);
 						if (lockFile == null || resolved_assembly == null) {
-							LogCodedWarning ("XA0107", resolved_assembly, 0, "Ignoring {0} as it is a Reference Assembly", resolved_assembly);
+							LogCodedWarning ("XA0107", resolved_assembly, 0, "Ignoring {0} as it is a Reference Assembly", resolved_assembly ?? assembly.ItemSpec);
```
jonpryor: Close, but not quite; if `resolved_assembly` is `null`, which is what we're assuming will cause the assembly name to be empty, then that means that the `file` parameter provide to `LogCodedWarning()` will *also* be null/empty, which presumably is not quite useful within the IDE.

Issue #3841, opened 2019-10-25
IOException-from-Stream control


jonpryor:
Related: https://github.com/xamarin/xamarin-android/issues/3646

We have decided that for .NET 5 we will fix our `System.IO.Stream` wrappers so that they throw `System.IO.IOException` instead of `Java.IO.IOException`, which we know will potentially break existing apps, but will allow other apps to work.  (Plus, .NET 5 will likely break semantics of other things *anyway*, so might as well!)

*In the meantime*, we want to introduce a new MSBuild property to control/override semantics:

**`$(AndroidBoundStreamIOExceptionType)`** will be either `System.IO.Stream` or `Java.IO.Exception`.  If it is set to `Java.IO.IOException`, the existing behavior will continue.  If it's `System.IO.Stream`, then e.g. `Android.Runtime.OutputStreamInvoker` will wrap all thrown `Java.IO.IOException` instances with `System.IO.IOException`.

We will need to audit the rest of our `Android.Runtime` namespace API sfor similar "leaky abstraction" behavior.

Additionally, we will investigate a "lower-level" fix of updating the Mono runtime so that when it's JIT'ing a catch block for `System.IO.IOException`/`Java.IO.IOException`, it will also handle the other exception type.

Issue #3646, opened 2019-09-18
BluetoothSocket.InputStream should throw a System.IO.IOException


kwintAl:
[BluetoothSocket.InputStream](https://docs.microsoft.com/en-us/dotnet/api/android.bluetooth.bluetoothsocket.inputstream?view=xamarin-android-sdk-9) is a Stream and [Stream.Read](https://docs.microsoft.com/de-de/dotnet/api/system.io.stream.read?view=netstandard-2.0) should throw a System.IO.IOException

### Steps to Reproduce

1. Create SPP RFCOMM Connection
  ```C#
_adapter = BluetoothAdapter.DefaultAdapter;
// ...
var device = _adapter.GetRemoteDevice(address);
var socket = device.CreateRfcommSocketToServiceRecord(UUID.FromString("00001101-0000-1000-8000-00805F9B34FB"));
await socket.ConnectAsync();
var stream = socket.InputStream;
  ```

2. start read operation
```C#
try {
     var buffer = new byte[128];
     var read = stream.Read(buffer, 0, buffer.Length);
    // ...
}
catch(IOException e) {
    throw;
}
catch (Exception e)
{
    // Java.IO.IOException...
    throw;
}
```
3. turn remote device off and wait few seconds

### Expected Behavior
System.IO.IOException

### Actual Behavior
Java.IO.IOException

gugavaro:
@grendello have we deal with similar issue in the past?

grendello:
@gugavaro yes, but for our own code, not the platform code. 

@kwintAl  In this case I don't believe we should wrap the native exception in a .NET one. We merely provide a thin bridge between the .NET land and the Java land and the type is inherently Java/Android. Therefore it should throw exceptions documented in Android docs (in this case [here](https://developer.android.com/reference/kotlin/java/io/InputStream.html?hl=en)) to stay as close as possible to the platform. Whenever one uses a type from the `Android*` namespaces, it should be expected that such classes will throw Java exceptions in some situations, not the .NET BCL ones. Changing the behavior here would require to change the behavior everywhere else for consistency - and that's not a small task, not to mention breaking backward compatibility.

kwintAl:
> We merely provide a thin bridge between the .NET land and the Java land and the type is > inherently Java/Android

@grendello the return type ist **not** Java.IO.InputStream like in native Android/Java it's a System.IO.Stream see links above. 

```C#
[...]

PR #3855, opened 2019-10-29
Optionally wrap Java exceptions in BCL exceptions

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/3841
Context: https://github.com/xamarin/xamarin-android/issues/3646
    
`Xamarin.Android` has been wrapping `Java.IO.{Input,Output}Stream` in the bound
Java IO classes in the BCL `System.IO.Stream` class since forever as the
choice appeared to make the most sense for .NET developers at that time.
However, instead of wrapping `Java.IO.IOException` exception in
`System.IO.IOException` that could be thrown by thus wrapped streams, we've been
throwing the original exception as thrown by Java.

For code backward compatibility we cannot simply fix this discrepancy by
starting to throw `System.IO.IOException`, as that would break an unknown number
of applications and libraries out there. However, we can introduce a way for the
developers to opt-in to throwing the BCL exception which would simply include
the original Java one in its `InnerException` property.

This commit implements the above approach by adding an MSBuild property called
`AndroidBoundExceptionType` which accepts one of the two values:

  * `Java` (**default**)
    Continue current behavior of throwing the original Java exception
  * `System`
    Wrap all the Java exceptions in certain context in the corresponding
    BCL exception class (depending on context).

The `Android.Runtime` namespace has been audited for all occurences of Java API
calls that may throw exceptions and the documented exceptions are now optionally
wrapped in a corresponding .NET BCL exception.


gugavaro:
A couple of suggestions

Instead of bound the exception against Stream, couldn't we actually have it generic enough that we could handle any variation needed in the future?
Instead of create a StreamInvoker class, leave it the way it is now and create an extension method that would do what those methods are doing and make it generic, so you can specify the type of the exception you want to fire.

i.e.:
` this.WrapJavaCall<IOException>(code);`

this would allow we use this method from any class and specify the exception on the .net side we would like to replace with.

You can also add a condition on your catch so you don't need to re-throw the original exception, if the condition is not match the exception does not even get trapped.

i.e.:

catch (IOException e) when (AndroidRuntime.IOExceptionType == BoundStreamIOExceptionType.System)


```
protected void WrapJavaCall<T> (Action code)
 {
 	if (code == null)
 		throw new ArgumentNullException (nameof (code));

 	try {
 		code ();
 	} catch (Java.IO.IOException ex) when (AndroidRuntime.IOExceptionType == BoundStreamIOExceptionType.System)
 		throw new T (ExceptionInterceptMessage, ex);
 	}
 }
```

Also do we really need an enum, or should just be a Boolean, Native vs System ? if is just a boolean it can avoid we introduce a new enum BoundStreamIOExceptionType

this 
`static BoundStreamIOExceptionType ioExceptionType;`
could be replace by
`static bool useNativeExceptionType;`

grendello:
@gugavaro These are all good suggestions in general, but in this particular case we don't need to make the code generic. This is to solve one particular and very specific case and I don't believe we have any other issues of this kind. Regarding the enum choice - it's on purpose because the values used must be "shared" between MSBuild, Xamarin.Android native runtime and the managed code and it's good to use something that's explicit, informative and easy to match between the three locations.

grendello:
build

grendello:
There are two test failures here:

 * ABI broken: this is due to the fact that this PR adds a new public type to `Mono.Android.dll` (`StreamInvoker`)
 * AppBundle test: the old friend `System.DllNotFoundException: __Internal assembly:<unknown assembly> type:<unknown type> member:(null)`

The former is expected, the latter is not caused by this PR.

jonpryor:
While this conversation *started* with `InputStreamInvoker.cs` and the throwing of `System.IO.IOException`, we need to expand the scope of this to *all* .NET subclasses and interfaces.

Fortunately there aren't many of these; a `git grep 'class.*:'` within `src/Mono.Android/Android.Runtime` shows the relevant classes that need investigation:

```
AndroidEnvironment.cs:          class _Proxy : IWebProxy {
InputStreamInvoker.cs:  public class InputStreamInvoker : Stream
JavaCollection.cs:      public class JavaCollection : Java.Lang.Object, System.Collections.ICollection {
JavaCollection.cs:      public sealed class JavaCollection<T> : JavaCollection, ICollection<T> {
JavaDictionary.cs:      public class JavaDictionary : Java.Lang.Object, System.Collections.IDictionary {
JavaDictionary.cs:              class DictionaryEnumerator : IDictionaryEnumerator {
JavaDictionary.cs:      public class JavaDictionary<K, V> : JavaDictionary, IDictionary<K, V> {
JavaList.cs:    public partial class JavaList : Java.Lang.Object, System.Collections.IList {
JavaList.cs:    public class JavaList<T> : JavaList, IList<T> {
JavaSet.cs:     public class JavaSet : Java.Lang.Object, ICollection {
JavaSet.cs:     public class JavaSet<T> : JavaSet, ICollection<T> {
OutputStreamInvoker.cs: public class OutputStreamInvoker : Stream
XmlPullParserReader.cs: public class XmlResourceParserReader : XmlPullParserReader
XmlPullParserReader.cs: public class XmlPullParserReader : XmlReader, IXmlLineInfo
XmlReaderPullParser.cs: public class XmlReaderResourceParser : XmlReaderPullParser, IXmlResourceParser
XmlReaderPullParser.cs: public class XmlReaderPullParser : Java.Lang.Object, IXmlPullParser
```

`Xamarin.Android.Net/AndroidClientHandler.cs` may also warrant review.

For example, consider [`System.Collections.IDictionary`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.idictionary?view=netframework-4.8), implemented by [`Android.Runtime.JavaDictionary`](https://github.com/xamarin/xamarin-android/blob/995d73978048ae4e57fe88a498829a553b886396/src/Mono.Android/Android.Runtime/JavaDictionary.cs).  The [`Dictionary.Item(Object)`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.idictionary.item?view=netframework-4.8) indexer property only declares two exceptions: `ArgumentNullException` and `NotSupportedException`.

Can we say with certainty that [`JavaDictionary.Get(object)`](https://github.com/xamarin/xamarin-android/blob/995d73978048ae4e57fe88a498829a553b886396/src/Mono.Android/Android.Runtime/JavaDictionary.cs#L61-L69), called by the indexer, will only throw those two exception types?

The answer is *No*, we can't say that with certainty, because `JavaConvert.FromJniHandle()` calls `Java.Lang.Object.GetObject()`, which could throw lots of things (e.g. `MissingMethodException`, when a constructor can't be found, because e.g. an instance was prematurely collected).

Thus, `JavaDictionary.Get()` should be likewise updated to:

```csharp
	partial class JavaDictionary {
		internal object Get (object key)
		{
			if (key == null) throw new ArgumentNullException (nameof (key));
			try {
				if (id_get == IntPtr.Zero)
					id_get = JNIEnv.GetMethodID (map_class, "get", "(Ljava/lang/Object;)Ljava/lang/Object;");
				return JavaConvert.FromJniHandle (
						JavaConvert.WithLocalJniHandle (key,
							lref => JNIEnv.CallObjectMethod (Handle, id_get, new JValue (lref))),
						JniHandleOwnership.TransferLocalRef);
			} catch (Exception e) when (AndroidRuntime.IOExceptionType == BoundStreamIOExceptionType.System) {
				throw new NotSupportedException (e);
			}
		}
	}
```

Does all this need to be in *this* PR?  Arguably not.

However, the *naming* of things needs to take this process into consideration.

Thus, an `$(AndroidBoundStreamIOExceptionType)` property is too specific (my bad!).

Perhaps we should use `$(AndroidWrapNativeExceptionsForFrameworkCompatibility)` (yes, a terrible name; suggestions welcome!)?  *Something* that will allow things to be equally applicable for *both* `Stream` exceptions handling (this PR) and exception handling for other types (now or "soon")?

src/Mono.Android/Android.Runtime/StreamInvoker.cs
```diff
@@ -0,0 +1,40 @@
...
+		const string ExceptionInterceptMessage = "Java exception intercepted";
+
+		protected void WrapJavaCall (Action code)
+		{
+			if (code == null)
+				throw new ArgumentNullException (nameof (code));
+
+			try {
+				code ();
+			} catch (Java.IO.IOException ex) {
```
jonpryor: I believe that this should use a [conditional `catch` block](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/when#when-in-a-catch-statement):

```csharp
catch (Java.IO.IOException ex) when (AndroidRuntime.IOExceptionType == BoundStreamIOExceptionType.System) {
    throw new IOException (ExceptionInterceptMessage, ex);
}
```

This will avoid the need for the `throw;` (line 21), and make for an overall "better" "first chance exception" experience, as the full "original" stack trace will be present on the first chance exception, instead of a calls sack with this method as the "top" method.

src/Mono.Android/Android.Runtime/OutputStreamInvoker.cs
```diff
@@ -9,25 +9,26 @@ public class OutputStreamInvoker : Stream
...
+			if (stream == null)
+				throw new ArgumentNullException (nameof (stream));
+
 			this.BaseOutputStream = stream;
 		}
 
 		public override void Close ()
 		{
-			BaseOutputStream.Close ();
+			JavaCallWrapper.Call (() => BaseOutputStream.Close (), JavaCallWrapper.MakeIOException);
```
jonpryor: Philosophical question is, do we really want to be creating all these delegates, or should we instead "manually inline" the try/catch blocks at each call site?

This is certainly *nicer to look at* -- though "nicer to *reason* about" is debatable -- but this "niceness" requires wrapping `BaseOutputStream.Close()` with a delegate instance (GC heap allocation!) and virtual method call (more  overhead).

How much overhead is that?  I don't know.  Presumably not *lots*, but it's not *zero* either.

Thus I *am* rather suggesting that we do:

```csharp
try {
    BaseOutputStream.Close ();
}
catch (Java.Lang.Throwable t) when (AndroidRuntime.BoundExceptionType == BoundExceptionType.System) {
    return JavaCallWrapper.MakeIOException (t.Message, t);
}
```

src/Mono.Android/Android.Runtime/OutputStreamInvoker.cs
```diff
@@ -9,25 +9,26 @@ public class OutputStreamInvoker : Stream
...
 		}
 
 		protected override void Dispose (bool disposing)
 		{
-			if (disposing && BaseOutputStream != null) {
```
jonpryor: This change is unsafe.  If `disposing` is false, meaning the GC is collecting this instance, then `BaseOutputStream` may have already been collected, and we *should not* call `Dispose()` upon it.

src/Mono.Android/Android.Runtime/OutputStreamInvoker.cs
grendello: I'm torn on this one. Thought about it too, and I chose the DRY approach since I prefer it this way, but yes, there's some overhead. OTOH, it's an exception so the overhead doesn't really matter that much, it's slow as molasses anyway. But maybe you're right and we should be more "explicit" in handling all of this...

src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
```diff
@@ -254,8 +294,12 @@ static string GetDefaultTimeZone ()
 		static SynchronizationContext GetDefaultSyncContext ()
 		{
 			var looper = Android.OS.Looper.MainLooper;
+			try {
 			if (Android.OS.Looper.MyLooper() == looper)
 				return Android.App.Application.SynchronizationContext;
+			} catch (Java.Lang.Throwable ex) when (JNIEnv.ShouldWrapJavaException (ex)) {
```
jonpryor: I suspect we should remove the `when` here, and just catch `System.Exception`.  There is likely *no sane scenario* in which `SynchronizationContext.Current` should throw an exception.

src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
```diff
@@ -254,8 +294,12 @@ static string GetDefaultTimeZone ()
 		static SynchronizationContext GetDefaultSyncContext ()
 		{
 			var looper = Android.OS.Looper.MainLooper;
+			try {
 			if (Android.OS.Looper.MyLooper() == looper)
```
jonpryor: I think we're going to want to reindent this/similar code so that it actually looks "correct".  We'll just have to ignore whitespace changes while reviewing this PR so that it doesn't look ridiculously huge.

src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -75,6 +79,11 @@ public static partial class JNIEnv {
 			}
 		}
 
+		internal static BoundExceptionType BoundExceptionType {
```
jonpryor: It's internal; just leave it as a field, so we don't have to rely on the inliner?  This shouldn't even be used from outside this class, anyway...

src/monodroid/jni/xamarin-app.h
```diff
@@ -21,6 +21,7 @@ struct ApplicationConfig
 	bool uses_assembly_preload;
 	bool is_a_bundled_app;
 	bool broken_exception_transitions;
+	uint8_t bound_exception_type;
```
jonpryor: Should we care about struct packing & padding?  (Probably should have asked this earlier...)

grendello: I don't think we need to

PR #3856, opened 2019-10-29
[Xamarin.Android.Build.Tasks] fix aapt2 incremental build for GPS

jonathanpeppers:
Context: https://github.com/xamarin/GooglePlayServicesComponents/pull/271

In an application with Google Play Services, a json configuration file
is used, such as:

    <GoogleServicesJson Include="google-services.json" />

There are MSBuild targets in the NuGet package that process this file,
and add to the `@(LibraryResourceDirectories)` item group:

https://github.com/xamarin/GooglePlayServicesComponents/blob/fc057c754e04d3e719d8c111d03d60eb2467b9ce/source/com.google.android.gms/play-services-basement/merge.targets#L44-L80

Unfortunately, this causes MSBuild targets to run every time if
`$(AndroidUseAapt2)` is `True`:

    Building target "_ConvertLibraryResourcesCases" completely.
    Input file "obj\Debug.stamp" does not exist.

Such a project will run `aapt2` every time, even with no changes.

The `<CollectNonEmptyDirectories/>` MSBuild task has some logic that
doesn't take custom `@(LibraryResourceDirectories)` items into
account:

    { "StampFile", Path.GetFullPath (Path.Combine (directory.ItemSpec, "..", "..")) + ".stamp" },

This code assumes that the file is in `obj\Debug\lp\1\res`, and so
using `..\..` is completely fine. If not, we get a weird
`obj\Debug.stamp` file!

The fix would be:

* Check if the file is in the `lp` directory, if so, use the existing
  behavior.
* Create a new stamp file in `obj\Debug\stamp` that is a hash of path.
* Create any `%(StampFile)` files in `_CompileAndroidLibraryResources`
  that do not exist.

This should get things working for the current released GPS NuGet.

When #GooglePlayServicesComponents/271 lands in a future update of the
GPS NuGet package, it will supply a proper `%(StampFile)`.

/cc @Redth

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Aapt2.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Aapt2.targets
@@ -49,7 +49,10 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 <Target Name="_CollectLibraryResourceDirectories"
     Condition=" '$(_AndroidUseAapt2)' == 'True' "
   >
-  <CollectNonEmptyDirectories Directories="@(LibraryResourceDirectories);@(_AdditonalAndroidResourceCachePaths->'%(Identity)\res')">
+  <CollectNonEmptyDirectories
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
Only failures are network-related or zipalign, so looks good to merge.

PR #3857, opened 2019-10-29
[Xamarin.Android.Build.Tasks] remove the code that XA0121 deprecates

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/3654
Context: https://docs.microsoft.com/en-us/xamarin/android/release-notes/10/10.1#xa0121-deprecation-warning-for-old-xamarinandroidsupport-library-versions

In VS 16.4 / Xamarin.Android 10.1, we deprecated
`<GetAdditionalResourcesFromAssemblies/>`.

This PR removes the behavior completely for 16.5, since it uses SHA1.
We aim to remove all SHA1 and MD5 usage for 16.5.

NuGet packages I've seen that no longer work:

* Android Support Libraries, 25.x and older
* Google Play Services older than 44.x

I have still been able to use an old Xamarin.Forms (2.3.4) in
combination with newer support libraries.

## XA0121 is now an error ##

We give a build error now in `<FilterAssemblies/>` if we encounter the
unsupported attributes. This target runs very early in the build, so
we have to do it at this point.

## JcwGen-Tests ##

These tests were using these attributes such as:

    [assembly: Android.IncludeAndroidResourcesFromAttribute ("./",
        SourceUrl="file:///JavaLib.zip")]
    [assembly: Java.Interop.JavaLibraryReference ("classes.jar",
        SourceUrl="file:///JavaLib.zip")]
    [assembly: Android.NativeLibraryReference ("arm64-v8a/libsimple.so",
        SourceUrl="file:///NativeLib.zip", Version="native-lib-1")]

Instead, I switched to use `@(LibraryProjectZip)` for the `.aar` file
and `@(EmbeddedNativeLibrary)` for the `.so` files.

## Designer Tests ##

The designer tests build the MyDriving app:

https://github.com/Azure-Samples/MyDriving

Which uses old NuGet packages that now fail, and is now in read-only
mode on Github. I suspect we may just need to remove that test?

    error XA0121: Assembly 'Xamarin.Android.Support.CustomTabs' is using
        '[assembly: Java.Interop.JavaLibraryReferenceAttribute]', which is no longer supported.
        Use a newer version of this NuGet package or notify the library author.


jonathanpeppers:
@garuma with the changes here, the MyDriving app won't build anymore unless the NuGet packages are updated. Is that worth it, or do you think it could be dropped from the Designer's tests?

Their repo is readonly now: https://github.com/Azure-Samples/MyDriving

jonathanpeppers:
I started a build on a public branch: https://build.azdo.io/3211313

To see how our regression tests fare, I suspect I will need to update lots of projects 😨

garuma:
@jonathanpeppers hmmm, ideally I would want to keep the project (doing the updates that are needed) as it usually tends to uncover a lot of good regression with those style of older projects, should we try to fork the repo and work and making it compliant with this change?

jonathanpeppers:
@garuma do you have a fork already with some changes?

You have this change somewhere for aapt2: https://github.com/Azure-Samples/MyDriving/pull/607

There might be other changes, too, I vaguely remember some.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -1534,25 +1534,6 @@ public void BuildApplicationCheckThatAddStaticResourcesTargetDoesNotRerun ()
 			}
 		}
 
-		[Test]
-		[Ignore ("Re enable when MergeResources work is complete")]
```
dellis1972: Future work. Let's remove all of the tests labeled with this. As well as the code for the managed resource merger. I think we are just going to use the google tooling (aapt2) going forward.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
```diff
@@ -444,26 +430,6 @@ protected string GetPathToAapt ()
 			return GetPathToLatestBuildTools (exe);
 		}
 
-		[OneTimeSetUp]
-		public void FixtureSetup ()
-		{
-			// Clean the Resource Cache.
-			if (!TestEnvironment.IsRunningOnCI)
-				return;
-			if (Directory.Exists (CachePath)) {
```
dellis1972: This brings up an interesting thought. Should we be deleting this Cache for the user? or issue a warning that they can now remove this directory? It might be taking up valuable space. 

Its probably something for the release notes I think.

jonathanpeppers: Yes, we should probably document on the next release notes for 10.2 that will expand upon: https://docs.microsoft.com/en-us/xamarin/android/release-notes/10/10.1#xa0121-deprecation-warning-for-old-xamarinandroidsupport-library-versions

I don't know about automatically deleting, though. You can install VS side-by-side on Windows, and they could have 2017, etc.

PR #3859, opened 2019-10-30
[tests] Build, run, and clean up missing test suites

pjcollins:
Context: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3196159&view=artifacts

I recently noticed the `win-build-test-results` artifact that we were
uploading was enormous, even when all tests had passed. This was due to
the fact that we weren't cleaning up after successful tests from the
`CodeBehind`, `CodeGen-MkBundle`, and `EmbeddedDSOs` suites. I've also
updated our `result-packaging` globbing to exclude NuGet packages from
our test result zips to further reduce unnecessary uploading.

This also identified an issue in our test execution. The three suites
mentioned above are not running in any test stages against an installer.
To address this our build pipeline has been updated to ensure these
assemblies are built and uploaded during the `Mac Build` job, and then
downloadedd and ran in the `Test MSBuild - macOS` job.

### DIFF
--- a/tests/CodeBehind/UnitTests/BuildTests.cs
+++ b/tests/CodeBehind/UnitTests/BuildTests.cs
@@ -553,8 +553,15 @@ namespace CodeBehindUnitTests
 				}
 			} catch {
 				CopyLogs (testInfo, false);
+				foreach (var file in Directory.GetFiles (testInfo.OutputDirectory, "*.log", SearchOption.AllDirectories)) {
+					TestContext.AddTestAttachment (file);
+				}
 				throw;
 			}
+
+			// Clean up successful tests
+			FileSystemUtils.SetDirectoryWriteable (testInfo.OutputDirectory);
+			Directory.Delete (testInfo.OutputDirectory, recursive: true);
 		}
 
 		bool WasParsedInParallel (TestProjectInfo testInfo)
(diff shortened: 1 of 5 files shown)


tests/CodeBehind/UnitTests/BuildTests.cs
```diff
@@ -553,8 +553,15 @@ void RunTest (string testName, bool many, bool dtb, Action<TestProjectInfo, bool
 				}
 			} catch {
 				CopyLogs (testInfo, false);
+				foreach (var file in Directory.GetFiles (testInfo.OutputDirectory, "*.log", SearchOption.AllDirectories)) {
```
jonathanpeppers: There are probably `.binlog` files, too. Maybe could be:
```suggestion
				foreach (var file in Directory.GetFiles (testInfo.OutputDirectory, "*.*log", SearchOption.AllDirectories)) {
```

jonathanpeppers: Ok, yeah we are only using `*.log` files as attachments, we can wait on this.

pjcollins:
Latest force push was clean up, the same changes previously worked as expected against windows - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3201273&view=results. The Win job failure can be ignored.

PR #3860, opened 2019-10-30
Add unofficial XA project SDK and support for `dotnet build`

pjcollins:
Context: https://github.com/xamarin/xamarin-android/projects/13

Enables creation of an unofficial package which introduces a Xamarin.Android [MSBuild Project SDK](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk).
In some ways this iteration is a standalone implementation / extension of MSBuild.Sdk.Extras for Xamarin.Android.

The `TargetFramework` property is the only required `.csproj` file element when using this SDK:

```xml
<Project Sdk="Xamarin.Android.Sdk/0.0.1">
  <PropertyGroup>
    <TargetFramework>MonoAndroid10.0</TargetFramework>
  </PropertyGroup>
</Project>
```

Default Android property behavior is defined in `Xamarin.Android.Sdk.props`.

Default Android related file globbing behavior is defined in `Xamarin.Android.Sdk.DefaultItems.props`.

This unofficial project SDK should only be used for internal testing purposes at this time. It supports building and packaging of Xamarin Android projects which follow the format above when using both `msbuild` and `dotnet build` on macOS and Windows.

--- 

The default `AndroidSupportedTargetJitAbis` property has been updated to include all supported ABIs, as all of our Azure Pipelines based builds are already doing this. This property can still be overridden locally to reduce overall build time when needed.

A couple of `StreamWriter` instances have been wrapped  in a using block to fix an issue which would occur only when building with `dotnet build`. The `_CompileResources` target was failing during the `Aapt2Compile` task because we were attempting to use a file that we were writing to before the underlying `StreamWriter` was flushed. For some reason this was not causing an issue when running on .NET Framework.

The `System.ServiceModel` reference, as well as its single `PipeException` type usage have been removed for the purposes of compatibility with `dotnet build`. The `XA7029` error code was only added in `d16-4`, and as a result we don't yet have any hits in the wild. However, since this was only recently released, I believe it should be relatively harmless to remove.

We continue to run into test timeouts when running the entire
Xamarin.Android.Build.Tests suite against a build tree on Windows. In
some cases our test execution slows down dramatically. We'll attempt to
address this by reducing the amount of work we do in parallel.
Our Windows agents seemingly [have 8 cores available][0]. NUnit-Console
will attempt to use all cores when running tests in parallel by default.
Setting the `--workers` argument to 4 should cut our parallelization in
half, and potentially prevent the machine from running out of resources.

[0]: https://dev.azure.com/devdiv/DevDiv/_settings/agentqueues?agentId=688797&queueId=819&view=capabilities


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -3209,7 +3209,7 @@ because xbuild doesn't support framework reference assemblies.
     _Sign;
   </SignAndroidPackageDependsOn>
 </PropertyGroup>
-<Target Name="SignAndroidPackage" DependsOnTargets="$(SignAndroidPackageDependsOn)">
+<Target Name="SignAndroidPackage" DependsOnTargets="$(SignAndroidPackageDependsOn)" BeforeTargets="Publish">
```
jonathanpeppers: Yeah this could use `$(PublishDependsOn)`: https://github.com/microsoft/msbuild/blob/0c507a29b6d59340ef5221f66df5d953c1049ba1/src/Tasks/Microsoft.Common.CurrentVersion.targets#L5181

src/Xamarin.Android.Sdk/targets/Xamarin.Android.Sdk.props
```diff
@@ -0,0 +1,69 @@
...
+  <!-- User-facing configuration-agnostic defaults -->
+  <PropertyGroup>
+    <AndroidManifest Condition=" '$(AndroidManifest)' == '' ">Properties\AndroidManifest.xml</AndroidManifest>
+    <AndroidResgenFile Condition=" '$(AndroidResgenFile)' == '' ">Resources\Resource.designer.cs</AndroidResgenFile>
+    <AndroidResgenClass Condition=" '$(AndroidResgenClass)' == '' ">Resource</AndroidResgenClass>
+    <MonoAndroidResourcePrefix Condition=" '$(MonoAndroidResourcePrefix)' == '' ">Resources</MonoAndroidResourcePrefix>
+    <MonoAndroidAssetsPrefix Condition=" '$(MonoAndroidAssetsPrefix)' == '' ">Assets</MonoAndroidAssetsPrefix>
+    <AndroidApplication Condition=" '$(AndroidApplication)' == '' And !Exists('$(AndroidManifest)') ">false</AndroidApplication>
+    <AndroidApplication Condition=" '$(AndroidApplication)' == '' And Exists('$(AndroidManifest)') ">true</AndroidApplication>
+    <TargetFrameworkVersion Condition="'$(TargetFrameworkVersion)' == ''">v4.4</TargetFrameworkVersion>
```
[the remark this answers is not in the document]
pjcollins: I've updated things to require the following element in the proj file (similar to MSBuildSdkExtras)
```xml
<TargetFramework>MonoAndroid$(ShortVersion)</TargetFramework>
```
and we implicitly resolve the identifier and version from that.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
pjcollins: I've decided to cut `dotnet publish` considerations for this iteration.

src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
```diff
@@ -20,10 +20,13 @@ public static bool UpdateXmlResource (string res, string filename, Dictionary<st
...
-					using (var xw = new LinePreservedXmlWriter (new StreamWriter (stream)))
-						xw.WriteNode (doc.CreateNavigator (), false);
-
+				using (var stream = File.OpenWrite (tmpfile)) {
+					using (var sw = new StreamWriter (stream)) {
+						using (var xw = new LinePreservedXmlWriter (sw)) {
+							xw.WriteNode (doc.CreateNavigator (), false);
+						}
+					}
+				}
```
jonathanpeppers: This is a nitpick, but sometimes I stack multiple `using` in a row like this:
```
using (var stream = File.OpenWrite (tmpfile))
using (var sw = new StreamWriter (stream))
using (var xw = new LinePreservedXmlWriter (sw)) {
		xw.WriteNode (doc.CreateNavigator (), false);
}
```

src/Xamarin.Android.Build.Tasks/Utilities/XDocumentExtensions.cs
```diff
@@ -58,11 +58,14 @@ public static string ToFullString (this XElement element)
...
-				xw.Flush ();
-				return MonoAndroidHelper.CopyIfStreamChanged (stream, fileName);
+			using (var stream = new MemoryStream ()) {
+				using (var sw = new StreamWriter (stream)) {
+					using (var xw = new Monodroid.LinePreservedXmlWriter (sw)) {
+						xw.WriteNode (document.CreateNavigator (), false);
+						xw.Flush ();
+						return MonoAndroidHelper.CopyIfStreamChanged (stream, fileName);
+					}
+				}
```
jonathanpeppers: Same idea here with `using`.

src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
dellis1972: I'm beginning to like the stack `using` as well.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/XamarinProject.cs
```diff
@@ -354,5 +358,50 @@ public virtual string ProcessSourceTemplate (string source)
...
+
+		public void CopyNuGetConfig (string relativeDirectory)
+		{
+			// Copy our solution's NuGet.config
+			var nuget_config = Path.Combine (XABuildPaths.TopDirectory, "NuGet.config");
+			if (File.Exists (nuget_config)) {
+				string projNugetConfig = Path.Combine (Root, relativeDirectory, "NuGet.config");
+				File.Copy (nuget_config, projNugetConfig, overwrite: true);
+				// Write additional sources to NuGet.config if needed
+				if (ExtraNuGetConfigSources != null) {
```
dellis1972: nice :) I could have used this in #3938

dellis1972:
@pjcollins very nice work 😃 have a 🦄

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/XamarinProject.cs
pjcollins: Feel free to steal it for that PR, I can rebase this if that PR ends up landing first or vice versa.

pjcollins:
The last three Windows builds have hit a pretty dramatic slow down during test execution, though it doesn't appear to be unique to this PR. In any event, I think we'll want to try to diagnose that issue further before merging.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -140,6 +141,20 @@ stages:
         artifactName: $(InstallerArtifactName)
         targetPath: bin/Build$(XA.Build.Configuration)/$(InstallerArtifactName)
 
+    - task: DotNetCoreCLI@2
+      displayName: pack Xamarin.Android.Sdk
+      inputs:
+        command: pack
+        packagesToPack: $(System.DefaultWorkingDirectory)/build-tools/create-packs/Packager.csproj
+        configuration: $(XA.Build.Configuration)
+        packDirectory: $(System.DefaultWorkingDirectory)/bin/packs
```
jonpryor: Presumably `packDirectory` is the output directory.  Shouldn't this be e.g. `bin/Build$(XA.Build.Configuration)` as is done with the `.vsix` & `.pkg` files?

build-tools/automation/azure-pipelines.yaml
```diff
@@ -240,6 +263,7 @@ stages:
         testRunTitle: Xamarin.Android.Build.Tests - Windows Build Tree
         testAssembly: $(System.DefaultWorkingDirectory)\bin\Test$(XA.Build.Configuration)\Xamarin.Android.Build.Tests.dll
         testResultsFile: TestResult-MSBuildTests-WinBuildTree-$(XA.Build.Configuration).xml
+        nunitConsoleExtraArgs: --workers=4
```
jonpryor: Should this use `$(NUMBER_OF_PROCESSORS)` (assuming that you can access the `%NUMBER_OF_PROCESSORS%` environment variable...)?

src/Xamarin.Android.Build.Tasks/Utilities/UnhandledExceptionLogger.cs
```diff
@@ -67,8 +67,6 @@ public static void LogUnhandledException (this TaskLoggingHelper log, string pre
 				log.LogCodedError (prefix + "7027", ex.ToString ());
 			else if (ex is FileNotFoundException)		// IOException
 				log.LogCodedError (prefix + "7028", ex.ToString ());
-			else if (ex is PipeException)			// IOException
```
jonpryor: I guess this means nobody is hitting it yet?

src/Xamarin.Android.Sdk/Sdk/Sdk.targets
```diff
@@ -0,0 +1,3 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildThisFileDirectory)..\targets\Xamarin.Android.Sdk.targets"  />
```
jonpryor: Where does this `targets` directory/convention come from?  Is it an artifact of SDK Short-form projects?

build-tools/automation/azure-pipelines.yaml
pjcollins: NUnitConsole will use all by default, but I wanted to limit this as I was seeing a consistent slowdown here. I had a hunch that we may be trying to do too much in parallel on these Windows machines, see https://github.com/xamarin/xamarin-android/pull/3860/commits/c5e291f709cd41901d58f1a1dc00f5f53059fec2.
Previous builds (such as https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3297546) hit this slowdown on the Windows build tree tests, and when this was happening we weren't able to remote into the machine at all...

src/Xamarin.Android.Build.Tasks/Utilities/UnhandledExceptionLogger.cs
pjcollins: As mentioned above:
 > The XA7029 error code was only added in d16-4, and as a result we don't yet have any hits in the wild. However, since this was only recently released, I believe it should be relatively harmless to remove.

We can circle back once we have some more telemetry data from d16-4, or go with the proposed reflection based change now if that's preferable?

src/Xamarin.Android.Sdk/Sdk/Sdk.targets
pjcollins: This was done to follow `Microsoft.Build.Sdk` behavior:
https://github.com/dotnet/sdk/tree/master/src/Tasks/Microsoft.NET.Build.Tasks/targets

build-tools/automation/azure-pipelines.yaml
pjcollins: I made this configuration agnostic so that we could easily use the same path in a test projects `NuGet.Config`: https://github.com/xamarin/xamarin-android/pull/3860/files#diff-6557e277ae88a841e6f9afa8423f9359R34

We could rework this at this time to use an azure devops feed instead however: https://docs.microsoft.com/en-us/azure/devops/pipelines/artifacts/nuget?view=azure-devops&tabs=yaml#publish-your-packages

jonpryor: A concern of mine with a "configuration agnostic" output directory is that if you locally build a Debug config + .nupkg, then build a Release config + .nupkg, the second build will overwrite the first.  (I assume; I can't easily determine  what the`.nupkg` filename would be based on `Packager.csproj`.

As such, I *think* I'd prefer to leave the Configuration within the output path, unless it would otherwise be within the `.nupkg` filename (though I'm not sure that would actually be better in practice).

build-tools/automation/azure-pipelines.yaml
jonpryor: i.e. `%NUMBER_OF_PROCESSORS%` may in fact be too high, and unless YAML supports arithmetic -- which it doesn't appear to -- then hardcoding to a "known" lower value is the best we can do.

Right.  Thanks!

jonpryor: Presumably YAML has a comment syntax; could we comment what we're trying to do here?

src/Xamarin.Android.Build.Tasks/Utilities/UnhandledExceptionLogger.cs
jonpryor: Fair enough.  d16-4 only hit stable ~3 days ago...

src/Xamarin.Android.Sdk/Sdk/Sdk.targets
jonpryor: Yeah, but....

This is where my brain starts exploding.

As far as I can tell, `Microsoft.NET.Build.Tasks/targets` works because they use a wildcard within their `.csproj`:

https://github.com/dotnet/sdk/blob/c60cea4d72e90efac857a86a2215a590e77964b0/src/Tasks/Microsoft.NET.Build.Tasks/Microsoft.NET.Build.Tasks.csproj#L77

```xml
  <ItemGroup>
    <None Include="targets\**\*" PackagePath="targets\" />
  </ItemGroup>
```

Fortunately (?), this did lead me to the `%(PackagePath)` metadata -- which is presumably documented somewhere?  are there any good NuGet packaging docs? -- which is used in `Packager.csproj`:

https://github.com/pjcollins/xamarin-android/blob/c5e291f709cd41901d58f1a1dc00f5f53059fec2/build-tools/create-packs/Packager.csproj#L55-L69

...though this still leaves me partially confused (need more docs & context!): is `targets` a "conventional" directory in NuGet packages?  (Clearly I haven't unzipped enough of them.)  As the above fragment *appears* to place `src\Xamarin.Android.Sdk\targets\**\*` into a `PackagePath` of `/targets/**/*`...

jonpryor:
For commit message purposes, it would be nice to know what the *structure* of the generated `.nupkg` file is.  What directories & files does it contain?

src/Xamarin.Android.Sdk/Sdk/Sdk.targets
pjcollins: For conventional NuGet packages, I think the `build` folder would be the expected/preferred location for targets/props folders and any custom build tasks, as is documented here:

https://docs.microsoft.com/en-us/nuget/create-packages/creating-a-package#from-a-convention-based-working-directory

However, this package represents an MSBuild Project SDK, the directory hierarchy of which doesn't have any good documentation that I could find. I've been _roughly_ following how the `Microsoft.NET.Sdk` is laid out, for instance on macOS:

```
/usr/local/share/dotnet/sdk/3.1.100/Sdks/Microsoft.NET.Sdk/tools/netcoreapp2.1/*
/usr/local/share/dotnet/sdk/3.1.100/Sdks/Microsoft.NET.Sdk/tools/net472/*
/usr/local/share/dotnet/sdk/3.1.100/Sdks/Microsoft.NET.Sdk/targets/*
/usr/local/share/dotnet/sdk/3.1.100/Sdks/Microsoft.NET.Sdk/Sdk/Sdk.targets
/usr/local/share/dotnet/sdk/3.1.100/Sdks/Microsoft.NET.Sdk/Sdk/Sdk.props
...
````

The `targets/*` path above "works" because the files in here are explicitly imported through the main `Sdk/Sdk.props`  and `Sdk/Sdk.targets` entry points: 
https://github.com/dotnet/sdk/blob/c60cea4d72e90efac857a86a2215a590e77964b0/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.props#L41
https://github.com/dotnet/sdk/blob/c60cea4d72e90efac857a86a2215a590e77964b0/src/Tasks/Microsoft.NET.Build.Tasks/sdk/Sdk.targets#L19

Below is the current structure of the Xamarin.Android.Sdk.nupkg which contains both macOS and Windows components. The folder hierarchy is a bit arbitrary at the moment, because we will be looking to break this down further into an SDK (the current `Sdk`, `targets` and `tools` folders), and targeting/framework/runtime packs (the current `lib` folder) in the ~near future:
 
https://gist.github.com/pjcollins/e3ce7616bcea6d343aff6d607e07d4f7

File: build-tools/automation/azure-pipelines.yaml (hunk 6 above)
Comment (pjcollins) [reply build-tools/automation/azure-pipelines.yaml]: Fair point, I'll look at publishing to an internal feed instead. That way we don't need to explicitly download the .nupkg for our test jobs.

File: build-tools/automation/azure-pipelines.yaml (hunk 7 above)
Comment (pjcollins) [reply build-tools/automation/azure-pipelines.yaml]: Comment added in https://github.com/xamarin/xamarin-android/pull/3860/commits/7b38640bfb8895dc0a96ba5c28a34aae33e6b362.

**Comment by jonpryor** [comment]
@pjcollins: Thank you for the directory structure list at https://gist.github.com/pjcollins/e3ce7616bcea6d343aff6d607e07d4f7

Looking at `tools/bcl/Facades/...`, I *really* want to get a `Windows` directory in there, updating `$(MonoAndroidBinDirectory)` appropriately.  (Not now...)

How easy would it be to produce *multiple* NuGet packages with OS-specific dependencies?  (Not for now, but for future exploration.)  It seems somewhat "odd" that the NuGet package will contain *both* Windows & macOS-specific binary files, while leaving out Linux altogether.  Would it be possible to place `tools/bcl`/`tools/Darwin`/etc. into OS-specific NuGet packages?  Ditto NDK tools, etc.

File: build-tools/automation/azure-pipelines.yaml (hunk 6 above)
Comment (pjcollins) [reply build-tools/automation/azure-pipelines.yaml]: After further consideration/investigation, I think we should hold off on feed distribution for the moment as it is dependent on unique versioning. Considering that the next phase of investigation will be breaking up this NuGet package, I don't think it makes sense to set up a robust versioning schema at this time. I've updated output and pack directories to instead contain `$(Configuration)` information for now.

File: src/Xamarin.Android.Sdk/Sdk/Sdk.targets (hunk 9 above)
Comment (Nirmal4G) [reply src/Xamarin.Android.Sdk/Sdk/Sdk.targets:2]: There are a few ways you can import an SDK

global.json
```json
{
    "msbuild-sdks":
    {
         "Xamarin.Sdk": "1.0.0"
    }
}
```
or installed through `$(MSBuildExtensionsPath)/Sdks` or use your own `MSBuildSdksPath` env var, which will lose access to the former path. If you don't provide the version, the resolver (`MSBuildSdkResolver`) will search through the Sdks paths defined earlier.
```xml
<Project Sdk="Xamarin.Sdk" />

<Sdk Name="Xamarin.Sdk" />

<Import Sdk="Xamarin.Sdk" Project="Sdk.props">
<Import Sdk="Xamarin.Sdk" Project="Sdk.targets">
<Import Sdk="Xamarin.Sdk" Project="Sdk.Custom.file">
<Import Sdk="Xamarin.Sdk" Project="..\targets\Xamarin.Sdk.props">
```
--- OR ---

providing explicit version: (will search through NuGet through `NuGetSdkResolver`)
```xml
<Project Sdk="Xamarin.Sdk/1.0.0" />
<Sdk Name="Xamarin.Sdk/1.0.0">
<Import Sdk="Xamarin.Sdk/1.0.0" Project="...">
```

pinging @rainersigwald to verify...

jonathanpeppers: @Nirmal4G we’ve progressed quite beyond this PR at this point, thanks!

You might learn more about what we are doing in this doc, we are adding to it as we go: https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/DotNet5.md

If you review our current PRs with the title [.NET 5] those would also be of interest.

Nirmal4G: Sure, Will do!

PR #3862, opened 2019-10-31
[Xamarin.Android.Build.Tasks] Fix Library Resource ordering.

dellis1972:
Context http://work.azdo.io//1009593

Both aapt and aapt2 will use overlay semantics when deciding
which final resource to use. The order in which the resources
are passed to the tooling is important.

For `aapt` the it used the FIRST item it came across, this is
why we always passed the Application resource directory in first.
For `aapt2` the LAST item is used... this is why we always pass
the Application resoruce directory last.

However we still use the same ordering for the references as
we do for `aapt`. As a result if you have duplicate resources
in two library projects `aapt` and `aapt2` will use different
files.

This PR reverses the order of the libraries we use for `aapt2`.
This ensures that the behaviour we see in `aapt2` is consistent
with `aapt`.

Future work we need to find a way to allow users to change the
ordering of these referenes. In AndroidStudio the order of the
files in the gradle config is used. So users can change the order.
However in MSBuild the order is controlled by MSBuild itself,
and there is no way to override it.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
@@ -137,8 +137,8 @@ namespace Xamarin.Android.Tasks {
 				cmd.AppendSwitchIfNotNull ("--custom-package ", PackageName.ToLowerInvariant ());
 			
 			if (AdditionalResourceArchives != null) {
-				foreach (var item in AdditionalResourceArchives) {
-					var flata = Path.Combine (WorkingDirectory, item.ItemSpec);
+				for (int i = AdditionalResourceArchives.Length - 1; i >= 0; i--) {
+					var flata = Path.Combine (WorkingDirectory, AdditionalResourceArchives [i].ItemSpec);
 					if (File.Exists (flata)) {
 						cmd.AppendSwitchIfNotNull ("-R ", flata);
 					} else {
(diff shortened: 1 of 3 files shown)


brendanzagaeski:
*Release status update*

A new Preview version has now been published that includes the fix from this item. The fix is not yet included in a Release version. I will update this item again when a Release version is available that includes the fix.

brendanzagaeski:
*Release status update*

A new Release version has now been published on Windows that includes the fix from this item. The fix is not yet published in a Release version on macOS. I will update this item again when a Release version is available on macOS that includes the fix.

Fix included in Xamarin.Android 10.2.0.100.

Fix included on Windows in Visual Studio 2019 version 16.5. To get the new version that includes the fix, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) or install the latest version from <https://visualstudio.microsoft.com/downloads/>.

(Fix also included on macOS in Visual Studio 2019 for Mac version 8.5 Preview 1 and higher. To try the Preview version that includes the fix, check for the latest updates on the **Preview** [updater channel](https://docs.microsoft.com/visualstudio/mac/update).)

brendanzagaeski:
*Release status update*

A new Release version has now been published on macOS that includes the fix from this item.

Fix included in Xamarin.Android 10.2.0.100.

Fix included on macOS in Visual Studio 2019 for Mac version 8.5. To get the new version that includes the fix, check for the latest updates on the **Stable** [updater channel](https://docs.microsoft.com/visualstudio/mac/update).

(Fix also included on Windows in Visual Studio 2019 version 16.5 and higher. To get the new version that includes the fix, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) or install the latest version from <https://visualstudio.microsoft.com/downloads/>.)

PR #3867, opened 2019-11-01
[build] include *.apk files from the bin\Test$(Configuration) directory

jonathanpeppers:
Expanding upon 4d2ac4ae, we should archive `.apk` files as well.

This will be useful for the current mono/2019-10 bump.


jonathanpeppers:
The download was ~139MB, which I think is OK. I will change this to just include the `*-Signed.apk` and it should get down to ~78MB.

Issue #2230, opened 2018-09-26
[Proposal] Remove MDB and non-portable PDB support


jonathanpeppers:
Currently, let's say you have a "Hello World" Xamarin.Android application project on Windows. 

You _might_ have the following properties in your `HelloWorld.csproj`:
```
<PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
    <DebugSymbols>True</DebugSymbols>
    <DebugType>full</DebugType>
</PropertyGroup>
```
_NOTE: the latest template puts `DebugType=portable` here_

When built, what will currently happen:
* `HelloWorld.dll` is compiled
* `HelloWorld.dll.mdb` is the generated symbol file
* We run a `ConvertDebuggingFiles` MSBuild task (runs `pdb2mdb`) that is taking ~50ms for "Hello World". (I have seen it take up to 4 seconds, in cases where `Mono.Android.dll.mdb` existed, see #2215)
* We look for any `*.mdb` files in `_CollectMdbFiles` and `_CopyMdbFiles` targets

As you can see, it is a lot of work to support this file format. Making sure all our of project templates have `DebugType=portable` is the first step.

### What happens in VS for Mac?

The VS for Mac templates are not yet shared with the VS Windows templates, and they are specifying `DebugType=full` by defaut.

However, Mono is building portable `*.pdb` files with this setting. `ConvertDebuggingFiles` and `pdb2mdb` aren't running at all.

## What should we change?

The first thought, is that `Xamarin.Android.Common.targets` should do this by default:
```
<DebugType Condition=" '$(DebugType)' = '' OR '$(DebugType)' == 'full' OR '$(DebugType)' = 'pdbonly'>portable</DebugType>
```

This would make Xamarin.Android projects always use portable `*.pdb` files by default. `DebugType=portable` is set by SDK-style NetStandard projects [by default](https://github.com/dotnet/sdk/blob/e44c2520e36dc171688d8e9cadc38c65ca309add/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props#L78). It works for both `Debug` and `Release` builds.

We would also remove:
- Anything looking for `*.mdb` files
- `ConvertDebuggingFiles` and `pdb2mdb`

If we encounter a "non-portable" pdb file, it will also be skipped and a new warning emitted.

## Benefits

These changes cleanup our MSBuild targets and remove code (reduce technical debt).

[...]

PR #3868, opened 2019-11-01
[Xamarin.Android.Build.Tasks] ignore mdbs and non-portable pdbs

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/2230

We currently spend some time during builds converting symbol files if
`DebugType=Full` or `DebugType=PdbOnly` is used.

These are older symbol formats, and we have already done the work to
use `DebugType=Portable` everywhere in the Xamarin templates.

My thought is to just remove support for these symbols by default
unless `$(AndroidLegacySymbols)` is set to `True`. This way if a
problem comes up, developers can opt-in to the old behavior.

Eventually we may phase out support completely, but it may not matter
to leave it in.

I updated various tests here to make sure we continue to test both
`Full` and `Portable` in `DebuggingTest.cs`.

## XA0122 ##

I added a new warning, for cases where a PCL `<ProjectReference/>`
have `DebugType=Full` set. This seems like the only likely cause a
user could hit an issue. I added a test for this scenario.

## Other changes ##

* I found the `_CollectMdbFiles` MSBuild target is completely unused.
  It should have been removed in 2ea31e6f, but we can remove it now.
* I found a place `$(AllowedReferenceRelatedFileExtensions)` was
  adding duplicate values. We should have removed this in e39070277.

### DIFF
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -498,6 +498,14 @@ when packaging Release applications.
 
     Added in Xamarin.Android 6.1.
 
+-   **AndroidLegacySymbols** &ndash; If set to `True`, opts back into
+    supporting `DebugType=Full` and `DebugType=PdbOnly`. These are
+    Windows-specific symbol formats and `DebugType=Portable` is
+    generally preferred. See the [msbuild documentation][msbuild_common]
+    for details.
+
+    [msbuild_common]: https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties
+
 -   **AndroidLinkMode** &ndash; Specifies which type of
     [linking](~/android/deploy-test/linker.md) should be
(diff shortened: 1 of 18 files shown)


jonathanpeppers:
The only problem with this, we discussed on Slack: https://xamarinhq.slack.com/archives/C03CEGRUW/p1572625192284800

If you had a `<ProjectReference/>` to a PCL library with `DebugType=Full`, you wouldn't be able to debug anymore. This isn't as likely with a NetStandard library, because it defaults to `Portable`.

I need to figure out how we can warn for any `<ProjectRerference/>` using `DebugType=Full`.

jonathanpeppers:
We are doing this in .NET 6 (see: https://github.com/xamarin/xamarin-android/commit/aab69af77dc6a5a90cb63c9289b3dbca5ae1bc8f), so this isn't needed.

PR #3393, opened 2019-07-26
[Mono.Android-Tests] Run Java.Interop tests as well.

jonpryor:
Context: https://github.com/xamarin/monodroid/commit/e318861e
Context: https://github.com/xamarin/xamarin-android/commit/7d32ef3e6ef88eba004f1d2a9f9eb0f42cbf887e
Context: https://github.com/xamarin/xamarin-android/commit/1a2eb953af76b76b2443b744894523ddfe239ecf

When use of `Java.Interop.dll` was originally added in
xamarin/monodroid@e318861e, it didn't implement the "full"
`Java.Interop.JniRuntime` abstraction, which meant that until
commit 1a2eb953, trying to instantiate a `JavaInt32Array` instance
would result in a `NotImplementedException`, because of
`Android.Runtime.AndroidValueManager.AddPeer()`. 

Commit 1a2eb953 removed the exception but didn't *implement*
`JniRuntime.JniValueManager.AddPeer()`, meaning even though a
`JavaInt32Array` could be *instantiated*, it wouldn't *work*
"properly", e.g. `JniRuntime.JniValueManager.PeekPeer()` wouldn't
*find* the `JavaInt32Array` instance when given the same handle.

This in turn "happened" because `Mono.Android-Tests.dll` never
*executed* the unit tests for `Java.Interop.dll`, which it *couldn't*
do, because it didn't fully support the semantics.

Add the `Java.Interop.dll` unit tests to `Mono.Android-Tests.dll`, and
fix `Mono.Android.dll` and company so that they can actually *pass*:

  * Add a new `Java.Interop-Tests.csproj` project to "host" the
    Java.Interop unit tests in a MonoAndroid-profile project.

  * Add `java-interop.jar` to the `javac` command line generated by
    `Jar.targets`, so that e.g. `com.xamarin.java_interop.ManagedPeer`
    can be used.

  * Update `Mono.Android.dll` so that the Java.Interop unit tests
    work.  This includes "migrating" the (jobject => instance) mapping
    previously accessible via `Java.Lang.Object.GetObject<T>()` into
    `Android.Runtime.AndroidValueManager`.

  * `AndroidRuntime.GetExceptionForThrowable()` will now "unwrap" a
    `Java.Interop.JavaProxyThrowable` instance.

  * `AndroidTypeManager.GetSimpleReferences()` can't use
    `JNIEnv.GetJniName()` on-device because that always returns
    `java/lang/Object`, even if a type can't be found, which breaks
[...]

PR #3877, opened 2019-11-04
[build] improve Xamarin.Android.sln usability in VS Windows

jonathanpeppers:
Xamarin.Android.sln has two issues that make `Xamarin.Android.sln` not
work too well in VS Windows:

* It builds MSBuild tasks and uses them within the same solution, such
  as `xa-prep-tasks.csproj` and
  `Xamarin.Android.Tools.BootstrapTasks.csproj`.
* We have conditional `@(ProjectReference)`

To improve things, I completely removed the `$(XAIntegratedTests)`
MSBuild property and any `@(ProjectReference)` that were conditional.

I moved the two bootstrap `.csproj` files to a new
`Xamarin.Android.BootstrapTasks.sln` which gets built during `make
prepare` or `msbuild Xamarin.Android.sln /t:Prepare`.

VS Windows was also complaining about the
`CreateManifestResourceNames` target missing for projects that merely
import `Microsoft.Common.targets` such as `r8.csproj`. I added an
empty target in `Configuration.props` to fix all cases of this.

Unfortunately, things are still not perfect as you can hit errors such
as:

    Unable to copy file "obj\Debug\Xamarin.Android.Build.Tasks.dll" to "C:\src\xamarin-android\bin\Debug\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Build.Tasks.dll".
    The process cannot access the file 'C:\src\xamarin-android\bin\Debug\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Build.Tasks.dll' because it is being used by another process.
    Could not copy "obj\Debug\Xamarin.Android.Build.Tasks.dll" to "C:\src\xamarin-android\bin\Debug\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Build.Tasks.dll".
    Exceeded retry count of 10. Failed. The file is locked by: "Microsoft Visual Studio 2019 (26660), MSBuild.exe (18800), MSBuild.exe (17416)"	Xamarin.Android.Build.Tasks

This occurs if you modify `Xamarin.Android.Build.Tasks.csproj`.

I am not yet sure if we can get around this remaining problem, but we
can revisit in the future.


jonpryor:
Do we have an explanation for the file sharing issue on `Xamarin.Android.Build.Tasks.dll`?  Why is it being held open?

jonathanpeppers:
VS Windows keeps MSBuild nodes alive, and if they have previously loaded `Xamarin.Android.Build.Tasks.dll` the assembly is locked.

At least under .NET framework, it seems like they are just using `Assembly.LoadFrom` in MSBuild: https://github.com/microsoft/msbuild/search?q=assembly.loadfrom

Makefile
```diff
@@ -198,6 +202,7 @@ prepare-build: prepare-build-init
 .PHONY: prepare
 prepare: prepare-build
 	mono --debug $(PREPARE_EXE) $(_PREPARE_ARGS)
+	msbuild $(BOOTSTRAP_MSBUILD_FLAGS) $(BOOTSTRAP_SOLUTION)
```
jonpryor: Why do we need this change on *Windows*?

Xamarin.Android.sln
```diff
@@ -71,10 +69,6 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "sqlite-xamarin", "src\sqlit
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "OpenTK", "src\OpenTK-1.0\OpenTK.csproj", "{5EB9E888-E357-417E-9F39-DDEC195CE47F}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "create-bundle", "build-tools\create-bundle\create-bundle.csproj", "{1640725C-4DB8-4D8D-BC96-74E688A06EEF}"
```
jonpryor: I think we could also *remove* `build-tools/create-bundle`, though perhaps that should be a separate PR?  If it should be in a separate PR, I don't see why we'd need to remove `create-bundle` here.

Xamarin.Android.sln
```diff
@@ -137,6 +131,7 @@ Global
 	GlobalSection(SharedMSBuildProjectFiles) = preSolution
 		src\Xamarin.Android.NamingCustomAttributes\Xamarin.Android.NamingCustomAttributes.projitems*{3f1f2f50-af1a-4a5a-bedb-193372f068d7}*SharedItemsImports = 4
 		src\Xamarin.Android.Build.Tasks\Tests\Xamarin.Android.Build.Tests\Xamarin.Android.Build.Tests.Shared.projitems*{53e4abf0-1085-45f9-b964-dcae4b819998}*SharedItemsImports = 4
+		src\Xamarin.Android.NamingCustomAttributes\Xamarin.Android.NamingCustomAttributes.projitems*{66cf299a-ce95-4131-bcd8-db66e30c4bf7}*SharedItemsImports = 4
```
jonpryor: I'm no expert on .sln syntax but...  isn't this line nearly duplicative of the following line, with just the GUID changed?  Do we really need both?

jonpryor:
I think we need to do the "same" thing in Java.Interop -- not that I'm asking @jonathanpeppers to do so -- as that may help the VSMac experience.  (I think it'll also be needed to land PR #3393...)

Xamarin.Android.sln
jonathanpeppers: I usually let VS Windows do what it wants with SLN files, but I removed this because I don't know what it is...

Makefile
jonathanpeppers: This PR removes the "bootstrap" MSBuild tasks from the SLN, so we need `make prepare` to build them now.

I think it makes sense to build them at that step? or do you think `make all` should do that?

PR #3880, opened 2019-11-05
Bump LibZipSharp to 1.0.6

dellis1972:
Fixes an issue where the Linux native binaries were inverted i.e 32bit in x64 and 64bit in x86.


dellis1972:
Commercial PR https://github.com/xamarin/monodroid/pull/1047

jonpryor:
Merged https://github.com/xamarin/monodroid/pull/1047.

Please update the monodroid reference within this PR.

dellis1972:
> If we are going to bump libZipSharp a lot, would it make sense to put the `@(Reference)` in a file you import like this one: https://github.com/xamarin/xamarin-android/blob/master/build-tools/scripts/MSBuildReferences.projitems
> So then going forward you should only have to change the `packages.config` and one MSBuild file?

Long term we are gonna move to PackageReference and get rid of `packages.config`, that would be a better option IMHO. That way we can centralise it.

PR #3881, opened 2019-11-05
Macos mojave builds

cadsit:
Changes:

* Keychain management on build bots (adding before build + removing after)
* Enforces using the proper xcode for notarizing
* Forces referencing /usr/local/bin/git for provisioning
* [tests] Allow suites to build and run on dif hosts
    We saw a handful of test failures on the `msbuild - macOS` job due
    to the fact that our build and test machines no longer have the same
    user name. Updates the `CodeGen-MkBundle` and `EmbeddedDSO` test suites
    to look up Android SDK and Android NDK paths at runtime rather than
    compile time, to allow us to build and run them on different hosts.
* Only VSIX signs on trusted pipeline


cadsit:
Beautifying the branch, please hold

build-tools/automation/azure-pipelines.yaml
```diff
@@ -17,14 +17,23 @@ resources:
...
 variables:
-  InstallerArtifactName: installers
-  TestAssembliesArtifactName: test-assemblies
-  NUnitConsoleVersion: 3.9.0
-  DotNetCoreVersion: 2.1.701
-  HostedMacMojave: Hosted Mac Internal Mojave
-  HostedMac: Hosted Mac Internal
-  HostedWinVS2019: Hosted Windows 2019 with VS2019
-  VSEngWinVS2019: VSEng-MicroBuildVS2019
+  - name: InstallerArtifactName
```
jonpryor: Is this *really* "beautifying"?  There's a certain elegance and simplicity to what you're replacing...

build-tools/automation/azure-pipelines.yaml
```diff
@@ -51,14 +66,20 @@ stages:
   # Check - "Xamarin.Android (Mac Build)"
   - job: mac_build_create_installers
     displayName: Build
-    pool: $(HostedMacMojave)
+    pool: $(OnPremMacMojave)
```
jonpryor: Is `$(OnPremMacMojave)` using VMs, or *real* hardware?  See also PR #3875, in which the use of Mojave VMs appears to cause `ANDZA0000` errors while building.

(We still don't understand *why*.)

build-tools/gradle/gradlew
```diff
@@ -65,6 +65,9 @@ case "`uname`" in
 esac
 
 CLASSPATH=$APP_HOME/gradle/wrapper/gradle-wrapper.jar
+if [ -z ${JAVA_HOME+x} ] && [ -d "~/Library/Android/jdk" ]; then
+  JAVA_HOME="$HOME/Library/Android/jdk"
+fi
```
pjcollins: Let's try to remove this diff entirely, I think it should be fixed by https://github.com/xamarin/xamarin-android/commit/9fd888d02c1743035f6dc39da7c96a507259cd6b

build-tools/automation/azure-pipelines.yaml
cadsit: Hah, by beautifying, I mean the commit history :P

I had to change the syntax to this in order to reference the variable group at the end for the keychain stuff.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -322,7 +359,7 @@ stages:
   # Check - "Xamarin.Android (Test API Compatibility)"
   - job: mac_api_compat
     displayName: API Compatibility
-    pool: $(HostedMacMojave)
+    pool: $(HostedMac)
```
pjcollins: Could we restore the `HostedMacMojave` variable? We'll still want to use it for certain test stages, rather than reverting to the generic High Sierra pool.

build-tools/automation/azure-pipelines.yaml
pjcollins: This syntax change is needed in order to use the following variable group:
```yaml
 - group: Xamarin-Build-Secrets
```

However, we could add this variable group to the build pipeline web UI, to avoid having to change the syntax for all other global variables. The majority of our secrets are coming from variable groups configured in the [build pipeline UI](https://devdiv.visualstudio.com/DevDiv/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=11410&view=Tab_Variables), I'd rather continue to maintain global variable groups there.

cadsit:
I made the proposed changes and am waiting on https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3216057 for verification

build-tools/automation/azure-pipelines.yaml
cadsit: OnPrem is physical hardware. I hope they're not affected by this D:

build-tools/automation/azure-pipelines.yaml
cadsit: Resolved

build-tools/gradle/gradlew
cadsit: Resolved

build-tools/automation/azure-pipelines.yaml
cadsit: Resolved

build-tools/automation/azure-pipelines.yaml
```diff
@@ -159,6 +177,12 @@ stages:
     steps:
     - checkout: self
       submodules: recursive
+    ### PLEASE REMOVE THIS BLOCK IF IN THE DIFF
+    - script: |
+        git checkout master
+        git pull --rebase origin macos-mojave-builds
+        git submodule update --init --recursive
+    ### PLEASE REMOVE THIS BLOCK IF IN THE DIFF
```
pjcollins: I think all the custom rebasing logic in the YAML file can be removed now? This should happen automatically when building the PR.

pjcollins:
We've got a handful of msbuild test failures on the latest run due the fact that we're now building test assemblies on one machine and downloading and running them on another with a different user name. Ideally we can update these suites to not rely as much on various generated build files. We should likely remove the generation of the following .cs files[0] and instead use the Android tool lookup code in Xamarin.ProjectTools[1] if possible.

[0] https://github.com/xamarin/xamarin-android/blob/master/tests/EmbeddedDSOs/EmbeddedDSO-UnitTests/Config.cs.in
[0] https://github.com/xamarin/xamarin-android/blob/master/tests/CodeGen-MkBundle/Xamarin.Android.MakeBundle-UnitTests/Config.cs.in
[1]https://github.com/xamarin/xamarin-android/blob/f4f538e560a40b732d9be69eb0765f1723701b4e/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs#L33

---

Finally, I think https://github.com/xamarin/xamarin-android/blob/master/tests/CodeBehind/BuildTests/CodeBehindBuildTests.csproj needs to import `Configuration.props`.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -34,12 +32,18 @@ stages:
   # Check - "Xamarin.Android (Code Analysis CredScan)"
   - job: run_static_analysis
     displayName: CredScan
-    pool: $(HostedWinVS2019)
+    pool: $(XA.Build.Win.Pool)
```
pjcollins: I'd prefer to keep as much of the pipeline configuration in YAML as possible - could we restore the `HostedMacMojave` and `HostedMac` variables and their usages? Ideally for new machine pool usage we can just introduce one new variable set by the pipeline, something like `$(OnPremMacMojave)`?

cadsit: I think the existing variable names are falling out of touch with what they mean as we start teasing apart trusted/untrusted builds. To encourage YAML reuse between the trusted and untrusted pipelines, we need this abstraction.

There is no need to conditionally set these variables IMO because there is no circumstance that an untrusted build should run on the untrusted pool (and likewise for the trusted pool). Writing in the condition only increases the chance of having the build get assigned to the wrong pool.

cadsit: I'm thinking more about this -- would it make sense to keep the declarations of the pools in the YAML and then have the XA.Build.*.Pool point to the right variable (E.g. HostedWinVS2019)?

cadsit: That way you can keep the bot pool configuration in YAML, which I think is what you want?

pjcollins: > I think the existing variable names are falling out of touch with what they mean as we start teasing apart trusted/untrusted builds. To encourage YAML reuse between the trusted and untrusted pipelines, we need this abstraction.

The existing hosted pool variable definitions in YAML are simply a convenience that allows us to avoid having to type the name of a machine pool more than once. I am not entirely sure how the existing hosted machine pool usage is relevant to the trusted/untrusted conversation. I don't see a need to change anything surrounding the jobs which won't be using the new on-prem pools. Am I missing something there?

My thinking was that we would introduce one new variable for the on-prem machines (named `$(OnPremMacMojave)` or `$(OnPremMacMojavePoolName)`, the value of which would be set (and differ) based on the [variable configuration](https://dev.azure.com/devdiv/DevDiv/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=11410&view=Tab_Variables) of the pipeline that picked up the job. The "mac build" and "notarize + upload installers" jobs could then be updated to use this new variable.

jonpryor:
PR message mentions:

> Ensures JAVA_HOME is set for gradle

but i don't see how that happens?  Searching for `JAVA_HOME` or `gradle` doesn't show anything relevant within the PR.

cadsit:
@jonpryor that was part of the build, but it is no longer. Will update the description

build-tools/automation/azure-pipelines.yaml
cadsit: Ah, gotcha. I think I was a little more focused on doing more work in the renaming now because we will likely also want to swap over the test hosts to on-prem machines eventually. If you would prefer we limit the change to as few places as possible for now and deal with the rest as it comes up, I can swap the names back.

I did see the XA.Build.*.Pool variables and didn't realize they were there for legacy reasons, so it makes sense to not tinker around with those for now.

pjcollins: Yeah, let's wait to introduce pool naming changes for those other jobs in a future PR.

cadsit:
Latest PR validation is here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3237385

Issue #1118, opened 2017-12-15
`api-compatibility` needs to check `[Register]` attribute values.


jonpryor:
Related: PR #930  
Related: https://bugzilla.xamarin.com/show_bug.cgi?id=60069

We need an API compatibility check that not only ensure that we don't break API -- which `tests/api-compatibility` does -- but *also* that we don't break *semantics*. The problem with [Bug #60069](https://bugzilla.xamarin.com/show_bug.cgi?id=60069) was that the *semantics* of `Build.Serial` changed from a Java field read to a Java method invocation, and since the `Build.getSerial()` method that was invoked only existed in API-26, the result was that the formerly working expression `Build.Serial` now threw everywhere.

We need to check for and prevent this in the future.

My current idea is that we improve `mono-api-html` to check the values of custom attributes, and warn if they change. For example, with `Build.Serial` the `[Register]` attribute contents changed:

```csharp
		// API < 26, field use:
		[Register ("SERIAL", ApiSince = 9)]
		public static string Serial {
			get {
				const string __id = "SERIAL.Ljava/lang/String;";

				var __v = _members.StaticFields.GetObjectValue (__id);
				return JNIEnv.GetString (__v.Handle, JniHandleOwnership.TransferLocalRef);
			}
		}

		// API >= 26, method invocation:
		[Register ("getSerial", "()Ljava/lang/String;", "", ApiSince = 26)]
		public static string Serial {
			get {
				const string __id = "getSerial.()Ljava/lang/String;";

				try {
					var __rm = _members.StaticMethods.InvokeObjectMethod (__id, null);
					return JNIEnv.GetString (__rm.Handle, JniHandleOwnership.TransferLocalRef);
				} finally {
				}
			}
		}
```

jonpryor:
Given the various issues we've had around `mono-api-html`-based API checking -- see also https://github.com/xamarin/xamarin-android-api-compatibility/commit/a773887d455eae6bf594e3d9075e0be1c88b1310 -- I think the "fix" here is to start using `Microsoft.DotNet.GenAPI` to generate "reference sources" for `Mono.Android.dll`/etc. and use those "reference sources" for API breakage checking.

The `api-snapshot` repo contains prebuilt binaries: https://github.com/xamarin/xamarin-android-api-compatibility/commit/a773887d455eae6bf594e3d9075e0be1c88b1310

[...]

Issue #3313, opened 2019-07-01
Missing AudioRouting interface for AudioRecord, etc, preventing solution to problem


gkarabin:
### Steps to Reproduce

In API level 24, Android changed the java [AudioRecord](https://developer.android.com/reference/android/media/AudioRecord) class to extend [AudioRouting](https://developer.android.com/reference/android/media/AudioRouting).  This is backwards compatible on the Java side, but not for the C# binding.

I noticed that [someone made notes on the problem](https://github.com/xamarin/xamarin-android/commit/576fb9814662f796cb5b0d466a7629b68c908fa9#diff-14048fd4cb6595cd0ad8cf1b41175ef5) in the Xamarin.Android repo.

As a result, if you look at the docs for Xamarin.Android's [AudioRecord](https://docs.microsoft.com/en-us/dotnet/api/android.media.audiorecord?view=xamarin-android-sdk-9) it does not implement the new interface from AudioRouting. A developer cannot access the new SDK level 24 methods:
- [addOnRoutingChangedListener](https://developer.android.com/reference/android/media/AudioRouting.html#addOnRoutingChangedListener(android.media.AudioRouting.OnRoutingChangedListener,%2520android.os.Handler))
- [removeOnRoutingChangedListener ](https://developer.android.com/reference/android/media/AudioRouting.html#removeOnRoutingChangedListener(android.media.AudioRouting.OnRoutingChangedListener))

I'm trying to resolve a customer issue, and want to provide them a solution based on non-obsolete APIs, on the belief/hope that more Android audio routing bugs are likely to be fixed on newer SDK levels.

### Expected Behavior

I can access the new SDK 24 AudioRouting APIS.

### Actual Behavior

I can't access the new SDK 24 AudioRouting APIS.

### Version Information

Visual Studio Enterprise 2019 for Mac
Version 8.1.3 (build 19)
Installation UUID: 77317fbd-d6e3-451c-b93f-33774ca9e0cf
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 5.6.0.25 (d16-0 / 50f75273)

	Package version: 518010028

Mono Framework MDK
Runtime:
	Mono 5.18.1.28 (2018-08/223ea7ef92e) (64-bit)
	Package version: 518010028

NuGet
Version: 5.0.2.5988

.NET Core
Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.2.4
	2.1.9
[...]

PR #3884, opened 2019-11-06
Adding new api break workflow that uses Microsoft.DotNet.ApiCompat.dll

gugavaro:
Our current api check workflow is not able to detect some api breaks or is flagging api breaks that actually should not be raised.

examples are:
If we change Attributes current check is unable to validate that.
Also, if we add abstract methods to an abstract class, it also does not flag as break

Some instances of false positives are:
When a base class change. 
from: ClassA : Object, ClassB: Object, ClassC: ClassA
to: ClassA : Object, ClassB: ClassA, ClassC: ClassB

this should not be considered a break, but our current check flags this as a break.

Also, using Microsoft.DotNet.ApiCompat.dll in our workflow makes sure we are following the same workflow .net team and azure team are using.

Another change is related to make the check as part of the project build step, so we can catch breakages much earlier and without any extra step.

original issue: https://github.com/xamarin/xamarin-android/issues/1118

here is an example of how the build would report the breakage:

AfterBuild:
  CheckApiCompatibility for ApiLevel: v10.0
  Compat issues with assembly Mono.Android:
  CannotRemoveBaseTypeOrInterface : Type 'Java.Util.Concurrent.Atomic.LongAdder' does not inherit from base type 'Java.Util.Concurrent.Atomic.Striped64' in the implementation but it does in the contract.
  Total Issues: 1
/home/gugavaro/Work/microsoft/Xamarin/xamarin-android-1/src/Mono.Android/Mono.Android.csproj(384,5): error : CheckApiCompatibility found nonacceptable Api breakages for ApiLevel: v10.0.
Done Building Project "/home/gugavaro/Work/microsoft/Xamarin/xamarin-android-1/src/Mono.Android/Mono.Android.csproj" (default targets) -- FAILED.

Build FAILED.

"/home/gugavaro/Work/microsoft/Xamarin/xamarin-android-1/src/Mono.Android/Mono.Android.csproj" (default target) (1) ->
(AfterBuild target) -> 
  /home/gugavaro/Work/microsoft/Xamarin/xamarin-android-1/src/Mono.Android/Mono.Android.csproj(384,5): error : CheckApiCompatibility found nonacceptable Api breakages for ApiLevel: v10.0.

    141 Warning(s)
    1 Error(s)


build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -0,0 +1,217 @@
...
+		// Validates Api compatibility between contract (previous version) and implmentation (current version)
+		// We do that by using Microsoft.DotNet.ApiCompat.dll
+		void ValidateApiCompat (string contractPath) {
+
+			var apiCompat = Path.Combine (ApiCompatPath, "Microsoft.DotNet.ApiCompat.exe");
+
+			string content;
+			using (var genApiProcess = new Process()) {
+
+				genApiProcess.StartInfo.FileName = "mono";
```
pjcollins: We likely need an OS check here to allow this task to run on Windows, or we'll need to make the target which runs this conditional.

src/Mono.Android/Mono.Android.csproj
```diff
@@ -372,4 +373,13 @@
       <ReferenceOutputAssembly>False</ReferenceOutputAssembly>
     </ProjectReference>
   </ItemGroup>
+  <Target Name="AfterBuild">
```
jonathanpeppers: Should we add `Inputs/Outputs` here? https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/MSBuildBestPractices.md#incremental-builds

The reason would be if this slows our build time, does this target run on every build?:
```
> msbuild src\Mono.Android\Mono.Android.csproj
> msbuild src\Mono.Android\Mono.Android.csproj /clp:performancesummary
```
Is this target taking up time on the second one?

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
gugavaro: @pjcollins we don't install mono on windows?

src/Mono.Android/Mono.Android.csproj
gugavaro: We should run on every build, if we are building Mono.Android we should check if we are breaking the API

gugavaro: the first one:

jonathanpeppers: Right so we only should check if `$(TargetPath)` changes, and is newer than a stamp file or something.

Something like (_untested_):

```xml
<Target Name="AfterBuild"
    Inputs="$(TargetPath)"
    Outputs="$(IntermediateOutputPath)\CheckApiCompatibility.stamp">
<!-- ... -->
  <Touch Files="$(IntermediateOutputPath)\CheckApiCompatibility.stamp" AlwaysCreate="True" />
  <ItemGroup>
    <FileWrites Include="$(IntermediateOutputPath)\CheckApiCompatibility.stamp" />
  </ItemGroup>
</Target>
```

This should make the second call skip `AfterBuild` if the assembly didn't change.

When I build `Xamarin.Android.sln`, I rarely change `Mono.Android.csproj` when working on MSBuild-related stuff. This should save ~3 seconds from my dev-loop.

gugavaro: got it, I will add it. thanks for the tip!

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
pjcollins: Not as far as I know. We'd want to invoke the .exe file directly on Windows I believe.

src/Mono.Android/Mono.Android.csproj
gugavaro: @jonathanpeppers Awesome Hint!

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
gugavaro: Added a check if windows vs use mono

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -0,0 +1,227 @@
...
+				genApiProcess.WaitForExit ();
+			}
+
+			ValidateIssues(content);
+		}
+
+		// Validates there is no issue or issues found are acceptable
+		void ValidateIssues(string content) {
+
+			Console.ForegroundColor = ConsoleColor.Red;
```
jonathanpeppers: One other thing I see here, is we probably shouldn't use `System.Console` directly from inside an MSBuild task.

They have their own logging methods, so I think you could probably:
* `Log.LogMessage` for general messages
* `Log.LogError` for each error -- and you don't need to turn the console red

You probably won't have to log the total errors either, as MSBuild will print its own summary.

jonathanpeppers: The benefit of using MSBuild's logging methods is that things like `.binlog` and custom MSBuild loggers will work. While `System.Console` would only appear in stdout.

gugavaro: I tried to use LogError here, but the way Msbuild counts the errors is not the way I want to represent it. It will count one error per line I print, in this case my string has lots of new lines, and I don't want that every single line on my string represent an error.The content of this string contains things that are not errors.

jonathanpeppers: You might have to do `LogMessage` for each line, and `LogError` to give the total number of errors if it fails.

But then we'd get those log messages in `binlog` files, too.

gugavaro: I update it to use log.Message, but that's very unfortunate because it will not give the same experience I was looking for with Console.WriteLine in Red, but I understand it is important to have those lines on Msbuild log. thanks for point that out.

pjcollins:
@gugavaro to fix the NuGet restore issue on the test jobs I think we'll need to update this [YAML template](https://github.com/xamarin/xamarin-android/blob/master/build-tools/automation/yaml-templates/setup-test-environment.yaml#L36) to include `feedsToUse` and `nugetConfigPath` params:

```yaml
- task: NuGetCommand@2
  displayName: nuget restore Xamarin.Android solutions
  inputs:
    restoreSolution: '**/Xamarin.Android*.sln'
    feedsToUse: config
    nugetConfigPath:  $(System.DefaultWorkingDirectory)/NuGet.config
```

Reference: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/nuget?view=azure-devops#yaml-snippet

gugavaro:
Thanks @pjcollins, I have updated the yaml file.

NuGet.config
```diff
@@ -4,5 +4,6 @@
     <clear />
     <!-- ensure only the sources defined below are used -->
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="dotnet internal feed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" protocolVersion="3" />
```
jonpryor: ...it never occurred to me that the keys could contain spaces...  TIL.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -0,0 +1,225 @@
...
+			var apiCompat = Path.Combine (ApiCompatPath, "Microsoft.DotNet.ApiCompat.exe");
+
+			string content;
+			using (var genApiProcess = new Process()) {
+
+				if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows)){
+  					genApiProcess.StartInfo.FileName = apiCompat;
+					genApiProcess.StartInfo.Arguments = $"{contractPath} -i {TargetImplementationPath}";
+				} else {
+					genApiProcess.StartInfo.FileName = $"mono";
```
jonpryor: I'm not sure this is needed.  I *thought* Mono's `Process` would check `FileName` and *implicitly* use `mono` if it's a managed app.

jonpryor:
Given a previously stated intention of "merging" the xamarin-android and xamarin-android-api-compatibility repos -- because the new "reference `.cs` file"  will be *much* smaller than the `mono-api-info`-generated XML file, making it *sane* to have everything within this repo -- why are the new "inter-API diff" files going into xamarin-android-api-compatibility ?

Why not place them into `tests\api-compatibility`?

src/Mono.Android/Mono.Android.csproj
```diff
@@ -372,4 +373,24 @@
...
+  <PropertyGroup>
+    <ApiCompatibilityDir>../../external/xamarin-android-api-compatibility</ApiCompatibilityDir>
+  </PropertyGroup>
+  <ItemGroup>
+    <ApiCompatibilityFiles Include="$(ApiCompatibilityDir)/v*.txt" />
+    <ApiCompatibilityFiles Include="$(ApiCompatibilityDir)/reference/*.dll" />
+  </ItemGroup>
+  <Target Name="AfterBuild" Inputs="$(TargetPath);@(ApiCompatibilityFiles)" Outputs="$(IntermediateOutputPath)CheckApiCompatibility.stamp">
+    <CheckApiCompatibility
+      ApiCompatPath="../../packages/Microsoft.DotNet.ApiCompat.5.0.0-beta.19554.1/tools/net472/"
```
jonpryor: Coding convention in XML:

1. Use spaces (which you've done)
2. Two-space indents.
3. *Attributes* are "indented an extra indent".

(3) is so that if the element has an output, you can trivially distinguish the attributes from the child elements:

```xml
<FooTask
    Value1="Foo"
    Value2="Bar">
  <Output TaskParameter="Something" PropertyName="Else" />
</FooTask>
```

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -0,0 +1,225 @@
...
+				genApiProcess.Start ();
+				content = genApiProcess.StandardOutput.ReadToEnd();
+				genApiProcess.WaitForExit ();
+			}
+
+			ValidateIssues(content);
+		}
+
+		// Validates there is no issue or issues found are acceptable
+		void ValidateIssues(string content) {
```
jonpryor: Code convention: space before `(`, `{` on a new line *for **methods***.

jonpryor:
We should review the App Bundle crash: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3223480&view=ms.vss-test-web.build-test-results-tab&runId=9669612&resultId=100000&paneView=debug

See if it's fixed by PR #3888, which fixes:

```
11-04 12:27:55.722  3339  3339 E mono    : Unhandled Exception:
11-04 12:27:55.722  3339  3339 E mono    : System.DllNotFoundException: __Internal assembly:<unknown assembly> type:<unknown type> member:(null)
11-04 12:27:55.722  3339  3339 E mono    :   at (wrapper managed-to-native) Android.Runtime.JNIEnv.monodroid_timing_start(string)
```

jonpryor:
The PR message mentions an error that doesn't appear to be fixed in the PR itself, and also doesn't error on a PR build:

```
AfterBuild:
CheckApiCompatibility for ApiLevel: v10.0
Compat issues with assembly Mono.Android:
CannotRemoveBaseTypeOrInterface : Type 'Java.Util.Concurrent.Atomic.LongAdder' does not inherit from base type 'Java.Util.Concurrent.Atomic.Striped64' in the implementation but it does in the contract.
Total Issues: 1
/home/gugavaro/Work/microsoft/Xamarin/xamarin-android-1/src/Mono.Android/Mono.Android.csproj(384,5): error : CheckApiCompatibility found nonacceptable Api breakages for ApiLevel: v10.0.
Done Building Project "/home/gugavaro/Work/microsoft/Xamarin/xamarin-android-1/src/Mono.Android/Mono.Android.csproj" (default targets) -- FAILED.

```

Am I misunderstanding something?

NuGet.config
gugavaro: yes, they can :)

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
gugavaro: I tired with out it and it errors out. so this check seems to be needed.

src/Mono.Android/Mono.Android.csproj
gugavaro: Fixed.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
gugavaro: Fixed.

gugavaro:
> Given a previously stated intention of "merging" the xamarin-android and xamarin-android-api-compatibility repos -- because the new "reference `.cs` file" will be _much_ smaller than the `mono-api-info`-generated XML file, making it _sane_ to have everything within this repo -- why are the new "inter-API diff" files going into xamarin-android-api-compatibility ?
> 
> Why not place them into `tests\api-compatibility`?


DotNet.ApiCompat does not use .cs, we will not be able to get rid of xamarin-android-api-compatibility repos yet.
I took the decision of using DotNet.ApiCompat instead of GenApi after I talked with the .net team and they don't recommend we use the .cs (Microsoft.DotNet.GenApi) approach we initially had, instead they asked us to use Microsoft.DotNet.ApiCompat

Microsoft.DotNet.ApiCompat is the tool .net and azure uses and it does a lot of further checks that is not possible to do with a simple text comparison.

gugavaro:
> The PR message mentions an error that doesn't appear to be fixed in the PR itself, and also doesn't error on a PR build:
> 
> ```
> AfterBuild:
> CheckApiCompatibility for ApiLevel: v10.0
> Compat issues with assembly Mono.Android:
> CannotRemoveBaseTypeOrInterface : Type 'Java.Util.Concurrent.Atomic.LongAdder' does not inherit from base type 'Java.Util.Concurrent.Atomic.Striped64' in the implementation but it does in the contract.
> Total Issues: 1
> /home/gugavaro/Work/microsoft/Xamarin/xamarin-android-1/src/Mono.Android/Mono.Android.csproj(384,5): error : CheckApiCompatibility found nonacceptable Api breakages for ApiLevel: v10.0.
> Done Building Project "/home/gugavaro/Work/microsoft/Xamarin/xamarin-android-1/src/Mono.Android/Mono.Android.csproj" (default targets) -- FAILED.
> ```
> 
> Am I misunderstanding something?

I am not sure I understand the question. This message is an example of what will see if there is a breakage. The check happens when we build Mono.Android, and if a break happens the build will fail with similar message.

The error does not happen on this PR because that error is an acceptable one based on the rules added on xamarin-android-api-compatibility repo
https://github.com/xamarin/xamarin-android-api-compatibility/blob/gugavaro_apicompat/v10.0.txt

jonpryor:
> The error does not happen on this PR because that error is an acceptable one based on the rules added on xamarin-android-api-compatibility repo

Ah.  Which specific rule disables the warnings that were mentioned?

jonpryor:
> we will not be able to get rid of xamarin-android-api-compatibility repos yet.

We may not be able to remove them *yet*, but if the *intention* is to *eventually* remove them, why *increase* our use of them, via these new `v*.txt` files?

Additionally, `v*.txt` is not a good semantic name; what are they *for*?  They're "for" `Microsoft.DotNet.ApiCompat`, but that's not at all obvious from the name.

Thus, let's look at *usage*:

```
var acceptableIssuesFile = Path.Combine (ApiCompatContractPath, $"{ApiLevel}.txt");
```

In context, these appear to be "contracts".  Thus, instead of `v*.txt`, `contract-v*.txt` would be more *meaningful*.

Though looking at the file *contents*:

https://github.com/xamarin/xamarin-android-api-compatibility/compare/a61271e046db989b4c9ec9f4d00c4358e07a2930...a9247ed847e039ec30467be9768a033288bd3ac2#diff-3cc026f88bf794495bd1a9ddb20a6643R1

`contract-issues-v10.txt` or `contract-rules-v10.txt` may be "better" names.

Thus, instead of e.g. `external/xamarin-android-api-compatibility/v10.txt`, I would instead suggest that these new files be placed in in `tests/api-compatibility/contract-rules-v10.txt`.  This is a more semantically meaningful name, and won't increase our dependency on the xamarin-android-api-compatibility repo.

gugavaro:
> > The error does not happen on this PR because that error is an acceptable one based on the rules added on xamarin-android-api-compatibility repo
> 
> Ah. Which specific rule disables the warnings that were mentioned?

for this specific error, this line will allow that specific breakage
https://github.com/xamarin/xamarin-android-api-compatibility/blob/a9247ed847e039ec30467be9768a033288bd3ac2/v10.0.txt#L14

gugavaro:
> > we will not be able to get rid of xamarin-android-api-compatibility repos yet.
> 
> We may not be able to remove them _yet_, but if the _intention_ is to _eventually_ remove them, why _increase_ our use of them, via these new `v*.txt` files?
> 
> Additionally, `v*.txt` is not a good semantic name; what are they _for_? They're "for" `Microsoft.DotNet.ApiCompat`, but that's not at all obvious from the name.
> 
> Thus, let's look at _usage_:
> 
> ```
> var acceptableIssuesFile = Path.Combine (ApiCompatContractPath, $"{ApiLevel}.txt");
> ```
> 
> In context, these appear to be "contracts". Thus, instead of `v*.txt`, `contract-v*.txt` would be more _meaningful_.
> 
> Though looking at the file _contents_:
> 
> [xamarin/xamarin-android-api-compatibility@a61271e...a9247eddiff-3cc026f88bf794495bd1a9ddb20a6643R1](https://github.com/xamarin/xamarin-android-api-compatibility/compare/a61271e046db989b4c9ec9f4d00c4358e07a2930...a9247ed847e039ec30467be9768a033288bd3ac2diff-3cc026f88bf794495bd1a9ddb20a6643R1)
> 
> `contract-issues-v10.txt` or `contract-rules-v10.txt` may be "better" names.
> 
> Thus, instead of e.g. `external/xamarin-android-api-compatibility/v10.txt`, I would instead suggest that these new files be placed in in `tests/api-compatibility/contract-rules-v10.txt`. This is a more semantically meaningful name, and won't increase our dependency on the xamarin-android-api-compatibility repo.

I agree with all your points and will fix all of them. Thanks!!!

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -0,0 +1,240 @@
...
+				Log.LogError ($"CheckApiCompatibility found nonacceptable Api breakages for ApiLevel: {ApiLevel}.");
+			}
+		}
+
+		// Converts list of issue into a dictinary
+		Dictionary<string, HashSet<string>> LoadIssues (string content)
+		{
+			var issues = new Dictionary <string, HashSet<string>> ();
+			HashSet<string> currentSet = null;
+			var lines = content.Split ('\n');
```
jonpryor: This should use `StreamReader.ReadLine()`.

build-tools/automation/yaml-templates/setup-test-environment.yaml
```diff
@@ -37,6 +37,8 @@ steps:
   displayName: nuget restore Xamarin.Android solutions
   inputs:
     restoreSolution: '**/Xamarin.Android*.sln'
+    feedsToUse: config
+    nugetConfigPath:  $(System.DefaultWorkingDirectory)/NuGet.config
```
jonpryor: @brendanzagaeski : this may be of interest to you wrt PR #3900.

tests/api-compatibility/README.md
```diff
@@ -0,0 +1,126 @@
+# Check-Api-Compatibility
+
+**Check-Api-Compatibility** is a process to check that changes are not causing
+APIs breakages on Mono.Android.dll
+The process uses [Microsoft.DotNet.ApiCompat][mdac] tool in order to
+test API compatibility between a two .NET assemblies
+When testing, the tool will compare a contract (old assembly) agaisnt an implementation (new assembly).
```
jonpryor: GitHub needs spell check.  "against".

src/Mono.Android/Mono.Android.csproj
```diff
@@ -373,4 +374,23 @@
...
     </ProjectReference>
   </ItemGroup>
+  <PropertyGroup>
+    <ApiCompatibilityDir>../../tests/api-compatibility</ApiCompatibilityDir>
+  </PropertyGroup>
+  <ItemGroup>
+    <ApiCompatibilityFiles Include="$(ApiCompatibilityDir)/acceptable-breakages-v*.txt" />
+    <ApiCompatibilityFiles Include="$(ApiCompatibilityDir)/reference/*.*" />
+  </ItemGroup>
+  <Target Name="AfterBuild" Inputs="$(TargetPath);@(ApiCompatibilityFiles)" Outputs="$(IntermediateOutputPath)CheckApiCompatibility.stamp">
```
jonpryor: Targets should go into `Mono.Android.targets`, because some IDEs (*cough* Visual Studio for Mac *cough*) will *rewrite and re-format* `.csproj` files in certain circumstances, e.g. adding a new project to the `.sln`.  This is annoying and makes for extra work (e.g. reverting the change to the `.csproj`).

jonpryor: `Inputs` and `Outputs` should also go on separate lines.

tests/api-compatibility/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+`LastStableApiLevel` - The last stable api level.
+
+`TargetImplementationPath` - Path to the implementation assembly. Based on this path we will generate the contract path by replace current version with previous version
+
+`ApiCompatibilityPath` - Path to the Api Compatibility folder. On this folder we should find the acceptable-breakages files and the reference folder that contians the current Mono.Android.dll assembly.
+
+The *task* is aware of the Api levels and what is the previous Api level for agiven version, so When building api version 2, the Task knows if previous version is the version 1.5 or 1.0 for instance.
+
+## Usage
```
jonpryor: Usage is both too much information and not enough information.

This isn't something for 3rd party use; it's only for *our* use.  As such, we don't need the `<Target/>` example, as that is "already" present in `src/Mono.Android/Mono.Android.csproj`/`src/Mono.Android/Mono.Android.targets`, meaning this comment will be invalidated/out-of-date as soon as we make changes to that call-site.

In some circumstances, "less is more."

On the other hand, there's not enough info here.  We can (should be!) "opinionated" about how things should work.  You mention `ApiCompatibilityPath`, but you don't mention *where that is within the tree*.  *That* part should be explicit.

tests/api-compatibility/README.md
```diff
@@ -0,0 +1,126 @@
...
+
+`ApiCompatibilityFiles` - Are the files we would like to monitor for changes, this is doen to prevent the task to run if it is not required. (there were no changes)
+
+`Touch Files` - It is our timestamp file to control if task should run.
+
+[bst]: https://github.com/xamarin/xamarin-android/tree/master/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks
+
+## ApiCompatibilityPath
+
+On this path we will place the files that contain all acceptable breakages for a given Api version.
```
jonpryor: These files need to follow some form of "file format".  Are there docs for that file format?  What's a line look like?

An example is provided below, but we don't know what are the available messages, nor -- more importantly! -- what the *workflow* is for updating these files.

*Presumably* some tool will print out these "rules" for us.  What is that tool?  How do we obtain the output?  Where does the output *go* so that new errors are ignored (after review)?

tests/api-compatibility/README.md
```diff
@@ -0,0 +1,126 @@
...
+```
+acceptable-breakages-v7.0.txt
+acceptable-breakages-v7.1.txt
+acceptable-breakages-v8.0.txt
+acceptable-breakages-v9.0.txt
+```
+
+`acceptable-breakages-v7.0.txt` - Contains any rules that are accept to be broken in relation to the previous Api level of v7.0 (in this case v6.0)
+
+Api Level v6 used to have thefollow piece of code:
```
jonpryor: We need to be careful about "api level" and "target framework version".   Android v6.0, which is `$(TargetFrameworkVersion)`=v6.0, is API level 23: https://source.android.com/setup/start/build-numbers

"Api level v6" is thus easily confusable: does it mean API-6, which is for Android 2.0.1 (2009!), or does it mean TargetFrameworkVersion/Android v6.0, released 2015?

*In general*, "API-" should be for API levels, e.g. API-23, while TargetFrameworkVersion should be used for Android versions.

...which will be completely meaningless come .NET 5, when our TargetFrameworkVersion scheme potentially changes, and we'll need to use some other terminology.  "Android v11.0", perhaps?

tests/api-compatibility/README.md
```diff
@@ -0,0 +1,126 @@
...
+Mono CIL Stripper
+
+26747392  Mono.Android.dll
+19300864  Mono.Android-out.dll
+```
+
+Rename Mono.Android-out.dll to Mono.Android.dll, compress (zip Mono.Android.dll.zip Mono.Android.dll)
+
+```
+4145708  Mono.Android.dll.zip
```
jonpryor: We should *probably* provide an MSBuild target to do this, e.g.

```
msbuild src/Mono.Android/Mono.Android.csproj /t:UpdateReferenceAssembly
```

Additionally, this doc should mention *where* these strip'd-and-compressed files should *go*.

Based on `.gitignore`, presumably the files should go into `tests/api-compatibility/reference/*.dll.zip`?

jonpryor: Where is `Mono.Android.dll.zip` *used from*?  I don't see anything within this PR to *unzip* the file so that it can be used by the tooling.

jonpryor:
Should this PR *also* add the handling of `Java.Interop.dll`, `Mono.Android.Export.dll`, `Xamarin.Android.NUnitLite.dll`, etc.?  Or should those be handled in separate PRs?

jonpryor:
What error is produced when the *value* of a custom attribute *changes*?

Is Issue #1118 fixed here, in that the *value* of the `[Register]` attribute will be checked for changes?

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
gugavaro: Is there a specific line ending issues you foresee? I am not seeing yet the benefit of using StreamReader since I am reading the output of the console execution of Microsoft.DotNet.ApiCompat.exeand I need to read the entire content anyways.
Using the string gives me some other advantages like not worry about dispose the stream once I am done.

I have added the string.Trim()

tests/api-compatibility/README.md
gugavaro: Fixed!

src/Mono.Android/Mono.Android.csproj
gugavaro: I wasn't aware of this hiccup of VSMac, thanks for pointing it out.
All Fixed!

tests/api-compatibility/README.md
gugavaro: Fixed!

tests/api-compatibility/README.md
gugavaro: The files are the exact content of the output of Microsoft.DotNet.ApiCompat.exe.
The best workflow is to build Mono.Android, and if any breakage happen, see the error printed on the screen or log, and decided if it is an acceptable breakage, just add that exact line to the acceptable-*.txt file

tests/api-compatibility/README.md
gugavaro: Updated to use target framework version for now.

tests/api-compatibility/README.md
gugavaro: The compressed files are unzipped during the check process. I believe you missed, but on CheckApiCompatibility.cs (line 93) I have this code:
```
// Before validate, check that zip files were decompressed.
var zipFiles = Directory.GetFiles(referenceContractPath, "*.zip");
foreach (var zipFile in zipFiles) {
	using (var zip = ZipArchive.Open (zipFile, FileMode.Open)) {
		zip.ExtractAll (referenceContractPath);
	}
}
```

gugavaro:
> Should this PR _also_ add the handling of `Java.Interop.dll`, `Mono.Android.Export.dll`, `Xamarin.Android.NUnitLite.dll`, etc.? Or should those be handled in separate PRs?

this PR should address Mono.Android, we should validate the new process is good enough, fix any issue and then we can expand to other assemblies, if this is okay with you?

gugavaro:
> What error is produced when the _value_ of a custom attribute _changes_?
> 
> Is Issue #1118 fixed here, in that the _value_ of the `[Register]` attribute will be checked for changes?

here is an example of an attribute change

```
CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'Android.Database.Sqlite.SQLiteDatabase.Path.get()' changed from '[RegisterAttribute("getPath", "()Ljava/lang/String;", "")]' in the contract to '[RegisterAttribute("getPath", "()Ljava/lang/String;", "GetGetPathHandler")]' in the implementation.
```

tests/api-compatibility/acceptable-breakages-v10.0.txt
```diff
@@ -0,0 +1,40 @@
+Compat issues with assembly Mono.Android:
```
jonpryor: Instead of `Compat issues with assembly ...`, could we do: `## Assembly: ...`?

e.g.

```
## Assembly: Mono.Android
```

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
gugavaro: Fixed!

src/Xamarin.Android.Build.Tasks/packages.config
```diff
@@ -7,6 +7,8 @@
   <package id="Microsoft.Build.Framework" version="14.3.0" targetFramework="net472" />
   <package id="Microsoft.Build.Tasks.Core" version="14.3.0" targetFramework="net472" />
   <package id="Microsoft.Build.Utilities.Core" version="14.3.0" targetFramework="net472" />
+  <package id="Microsoft.DotNet.ApiCompat" version="5.0.0-beta.19602.1" targetFramework="net472" />
```
jonpryor: Why add these to `src/Xamarin.Android.Build.Tasks` when it's `build-tools/Xamarin.Android.Tools.BootstrapTasks` which actually uses these utilities?

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -0,0 +1,294 @@
...
+					}
+
+					File.Copy (contractAssembly, Path.Combine (contractPathDirectory, assemblyToValidate), true);
+					File.Copy (implementationAssembly, Path.Combine (targetImplementationPathDirectory, assemblyToValidate), true);
+				}
+
+				using (var genApiProcess = new Process ()) {
+
+					genApiProcess.StartInfo.FileName = apiCompat;
+					genApiProcess.StartInfo.Arguments = $"{contractPathDirectory} -i {targetImplementationPathDirectory} ";
```
jonpryor: For "sanity", directories & path names should be quoted with `"` in case they contain spaces.

tests/api-compatibility/acceptable-breakages-v10.0.txt
```diff
@@ -0,0 +1,40 @@
...
+MembersMustExist : Member 'Android.Media.AudioRecord.add_RoutingChanged(System.EventHandler<Android.Media.AudioRecord.RoutingChangedEventArgs>)' does not exist in the implementation but it does exist in the contract.
```
jonpryor: ...and I find this both horrifying and confusing: this sounds like a breaking change, yet it's not listed in https://github.com/xamarin/xamarin-android-api-compatibility/blob/master/inter-api-extra-v9.0-v10.0.txt nor in https://github.com/xamarin/xamarin-android-api-compatibility/commit/336eedaa51883d90100e30c95a007d8922ab19f0 (which must be viewed locally, as it's too big for GitHub to display), and I don't remember it coming up in the context of breaking changes previously.

```
$ git show 336eedaa51883d90100e30c95a007d8922ab19f0 | grep -i RoutingChanged
# no output
```

Yet, when I diff, it's there:

```diff
$ diff -u src/Mono.Android/obj/Debug/android-28/mcw/Android.Media.AudioRecord.cs src/Mono.Android/obj/Debug/android-29/mcw/Android.Media.AudioRecord.cs
-               public event EventHandler<Android.Media.AudioRecord.RoutingChangedEventArgs> RoutingChanged {
+               public event EventHandler<Android.Media.AudioRoutingOnRoutingChangedEventArgs> RoutingChanged {
```

Why didn't our previous tooling find this?

For that matter, this finds nothing within the xamarin-android-api-compat repo:

```
$ grep '<event.*RoutingChanged' reference/Mono.Android.xml
# no matches
```

(Feeling of horror increases...)

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
gugavaro: Sure, Fixed!

src/Xamarin.Android.Build.Tasks/packages.config
gugavaro: I could not do it before because of a bug on the build prepare. Fixed by https://github.com/xamarin/xamarin-android/pull/3997

Moved the packages to Xamarin.Android.Tools.BootstrapTasks

tests/api-compatibility/acceptable-breakages-v10.0.txt
gugavaro: It took me quite a bit to remember this, but that's expected!
We were wrongly not generating that event, so that's why previous tool never complained about it.

see this: https://github.com/xamarin/xamarin-android/commit/bcae9415f188ae0f602bab89c57db1c265fb36d0#diff-14048fd4cb6595cd0ad8cf1b41175ef5L1301

After my fix, we started generating that but only until api 28

gugavaro:
This PR now can only be merged after PR: https://github.com/xamarin/xamarin-android/pull/3997 be merged.

tests/api-compatibility/acceptable-breakages-v10.0.txt
jonpryor: > We were wrongly not generating that event

OK, so it didn't previously exist

> After my fix, we started generating that but only until api 28

...and this is where we made a mistake.  If we weren't previously emitting it, we should have *continued* not emitting it, so that when we "introduced" it in API-29, it wouldn't be incompatible with previous versions.

😞

gugavaro: We discussed the issue together. 
The bigger issue was that the handlers/delegate were generated, but not the event. So the api was "half" there.
There was a bug: https://github.com/xamarin/xamarin-android/issues/3313
were someone was trying to use it on API 24, if we did not fix it, then that person could not have his project working for API 24

gugavaro:
Update the changes with mono.android.dll from Release 10.1.1

jonpryor:
Commit message for squash-and-merge:

```
Fixes: https://github.com/xamarin/xamarin-android/issues/1118

Context: https://bugzilla.xamarin.com/show_bug.cgi?id=60069
Context: https://github.com/xamarin/xamarin-android/pull/930

Our current API check workflow is not able to detect some API breakage
or is flagging non-breaking API changes as breaking.

Examples of breakages not previously reported:

  * Changing the values of `[Register]` attributes may be a breaking
    change but is not flagged by `mono-html-html`.

  * Adding `abstract` methods to an `abstract` class

Examples of acceptable changes which elicit warnings:

  * Changing a base class in a compatible manner.

Instead of using `mono-api-info` + `mono-api-html`, use
[Microsoft.DotNet.ApiCompat][0] to perform API compatibility
validation, which is what the .NET and Azure teams use.

Microsoft.DotNet.ApiCompat reports `[Register]` changes:

	CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'Android.Database.Sqlite.SQLiteDatabase.Path.get()' changed from '[RegisterAttribute("getPath", "()Ljava/lang/String;", "")]' in the contract to '[RegisterAttribute("getPath", "()Ljava/lang/String;", "GetGetPathHandler")]' in the implementation.

and thus would have caught the `Build.SERIAL` breakage in #930.

To override the warnings produced by Microsoft.DotNet.ApiCompat,
the message can be copied into an appropriate
`tests/api-compatibility/acceptable-breakages-*.txt` file.

See `tests/api-compatibility/README.md` for details.

[0]: https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.ApiCompat
```

PR #3886, opened 2019-11-06
[build] archive APKs/AABs in yaml

jonathanpeppers:
In 22de1178, it did not actually archive *all* `.apk` files.

Throughout our build pipeline, we build projects with different
MSBuild properties as input such as `/p:AotAssemblies=True`. This
causes the `.apk` files to be overwritten, and we lose them...

Instead, we should copy each APK to a directory inside
`$(Build.ArtifactStagingDirectory)` with a path identifying.

And so I propose modifying the template such as:

    - template: yaml-templates/apk-instrumentation.yaml
      parameters:
        configuration: $(ApkTestConfiguration)
        testName: Mono.Android_Tests-Aot
        project: src/Mono.Android/Test/Mono.Android-Tests.csproj
        testResultsFiles: TestResult-Mono.Android_Tests-$(ApkTestConfiguration)-Aot.xml
        extraBuildArgs: /p:AotAssemblies=True /p:EnableLlvm=True
        artifactName: Mono.Android_Tests-Signed.apk
        artifactFolder: Aot

Would yield an artifact at `Aot/Mono.Android_Tests-Signed.apk` that we
could download independently.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -381,7 +381,7 @@ stages:
         testResultsFiles: TestResult-Mono.Android_Tests-$(ApkTestConfiguration)-Aot.xml
         extraBuildArgs: /p:AotAssemblies=True /p:EnableLlvm=True
         artifactName: Mono.Android_Tests-Signed.apk
-        artifactFolder: Aot
+        artifactFolder: AotLlvm
 
     - template: yaml-templates/apk-instrumentation.yaml
       parameters:
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -370,6 +370,8 @@ stages:
         testName: Mono.Android_Tests
         project: src/Mono.Android/Test/Mono.Android-Tests.csproj
         testResultsFiles: TestResult-Mono.Android_Tests-$(ApkTestConfiguration).xml
+        artifactName: Mono.Android_Tests-Signed.apk
+        artifactFolder: Default
 
     - template: yaml-templates/apk-instrumentation.yaml
       parameters:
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
The last build before I did some squashing & cleanup looked like what we need: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3220679&view=artifacts

image

build-tools/automation/azure-pipelines.yaml
```diff
@@ -378,6 +380,8 @@ stages:
         project: src/Mono.Android/Test/Mono.Android-Tests.csproj
         testResultsFiles: TestResult-Mono.Android_Tests-$(ApkTestConfiguration)-Aot.xml
         extraBuildArgs: /p:AotAssemblies=True /p:EnableLlvm=True
+        artifactName: Mono.Android_Tests-Signed.apk
+        artifactFolder: Aot
```
jonpryor: Should this `artifactFolder` be `Aot+LLVM`?  Or do we not need to distinguish when LLVM was used?

build-tools/automation/yaml-templates/apk-instrumentation.yaml
```diff
@@ -19,10 +21,18 @@ steps:
       ${{ parameters.extraBuildArgs }}
   condition: succeededOrFailed()
 
+- script: |
+    SOURCE=$(System.DefaultWorkingDirectory)/bin/Test${{ parameters.configuration }}/${{ parameters.artifactName }}
+    DEST=$(Build.ArtifactStagingDirectory)/${{ parameters.artifactFolder }}/
+    mkdir $DEST
```
jonpryor: Should this quote `"$DEST"`?

build-tools/automation/yaml-templates/apk-instrumentation.yaml
```diff
@@ -19,10 +21,18 @@ steps:
...
+    cp $SOURCE $DEST
```
jonpryor: Ditto here: generally we should quote shell commands.

PR #3887, opened 2019-11-06
[tests] MSBuild performance raise expected times by 100ms

jonathanpeppers:
Context: https://build.azdo.io/3218699
Context: https://build.azdo.io/3219042

Raising some of the expected times, as we have seen some random
failures that do not appear to be performance regressions.

The assertion failures were:

    Build_No_Changes: Exceeded expected time of 3250ms, actual 3270ms
    Build_AndroidResource_Change: Exceeded expected time of 4750ms, actual 4770ms
    Build_XAML_Change(True,False): Exceeded expected time of 6250ms, actual 6270ms
    Build_XAML_Change(False,False): Exceeded expected time of 9500ms, actual 9560ms

### DIFF
--- a/tests/msbuild-times-reference/MSBuildDeviceIntegration.csv
+++ b/tests/msbuild-times-reference/MSBuildDeviceIntegration.csv
@@ -2,13 +2,13 @@
 # First non-comment row is human description of columns
 Test Name,Time in ms (int)
 # Data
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
It looks like the new times are working, I think we can go ahead and merge this.

PR #3890, opened 2019-11-08
Build only Debug configuration of libxamarin-debug-app-helper.so

grendello:
`libxamarin-debug-app-helper.so` is a shared library used only for Debug
Xamarin.Android applications to set up the debugging environment, amongst others
to copy the fast-deployed shared libraries from the override directories to an
internal location where we're allowed to use `dlopen` to load the DSOs.

The library used to be built twice (like our runtime), but unlike the runtime it
was copied to the same location and, with Release being built second, the
release version replaced the debug one resulting in a library that does not
include the necessary code to set up the scenario described above.

Fix the issue by building the library only in Debug configuration.

### DIFF
--- a/src/monodroid/CMakeLists.txt
+++ b/src/monodroid/CMakeLists.txt
@@ -270,7 +270,7 @@ endif()
 set(XAMARIN_APP_STUB_SOURCES ${SOURCES_DIR}/application_dso_stub.cc)
 add_library(xamarin-app SHARED ${XAMARIN_APP_STUB_SOURCES})
 
-if(NOT WIN32 AND NOT MINGW)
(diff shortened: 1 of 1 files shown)


grendello:
There are two failing tests:

  * `CodeBehindUnitTests.BuildTests_CodeBehindBuildTests.SuccessfulBuildMany` because of:
```
error ANDKT0000: keytool error: java.lang.Exception: Key pair not generated, alias <androiddebugkey> already exists [/Users/runner/runners/2.160.0/work/1/s/bin/TestRelease/temp/CodeBehind/SuccessfulBuildMany/Release/project/CodeBehindBuildTests.csproj]
```

 * `Mono.Android_TestsAppBundle` because of the not yet committed fix from https://github.com/xamarin/xamarin-android/pull/3888

None of them are caused by this PR, therefore I'm going to merge.

PR #3891, opened 2019-11-08
[Xamarin.Android.Build.Tasks] DTBs and _ResolveLibraryProjectImports

jonathanpeppers:
I spent a little time looking at build logs of design-time builds
inside VS Windows. To trigger these, I was editing an `.axml` file and
saving it.

I found some perf issues...

The first thing I noticed was that `_ResolveLibraryProjectImports`
runs on every DTB:

    Building target "_ResolveLibraryProjectImports" completely.
    Input file "obj\Debug\100\designtime\build.props" is newer than output file "obj\Debug\100\stamp\_ResolveLibraryProjectImports.stamp".

Then I looked at how `build.props` was actually changing between
different targets such as:

* `UpdateGeneratedFiles` - the general DTB target
* `SetupDependenciesForDesigner` - the Android designer runs this
  target

Comparing the files, I saw this difference:

```diff
- AndroidSdkPath = C:\Program Files (x86)\Android\android-sdk
+ AndroidSdkPath = C:\Program Files (x86)\Android\android-sdk\
- AndroidNdkPath = C:\Program Files (x86)\Android\android-sdk\ndk-bundle
+ AndroidNdkPath = C:\Program Files (x86)\Android\android-sdk\ndk-bundle\
- JavaSdkPath = C:\Program Files\Android\Jdk\microsoft_dist_openjdk_1.8.0.25
+ JavaSdkPath = C:\Program Files\Android\Jdk\microsoft_dist_openjdk_1.8.0.25\
```

So in some cases `_ResolveMonoAndroidSdks` is running to normalize the
paths and other times not... `_CreatePropertiesCache` should depend on
`_ResolveMonoAndroidSdks` to fix this issue.

After that I found the *next* issue:

```diff
- DesignTimeBuild = True
+ DesignTimeBuild = true
```

What??? It appears that the casing differs depending on which part of
the IDE is triggering the build. I suspect the designer is using one
casing while VS proper uses another.

I think we should call `ToLowerInvariant()` on every line in the
`build.props` file to work around MSBuild's case insensitivity.

This change saves ~350ms on every design-time build. However, if
`build.props` changes all the time it may do even better, since
`build.props` triggers almost all MSBuild targets in Xamarin.Android.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -1172,7 +1172,7 @@ because xbuild doesn't support framework reference assemblies.
 	<_AndroidManagedResourceDesignerFile>$(_AndroidIntermediateDesignTimeBuildDirectory)$(_AndroidResourceDesigner)</_AndroidManagedResourceDesignerFile>
 </PropertyGroup>
 
-<Target Name="_CreatePropertiesCache" DependsOnTargets="_SetupDesignTimeBuildForBuild;_SetLatestTargetFrameworkVersion">
+<Target Name="_CreatePropertiesCache" DependsOnTargets="_SetupDesignTimeBuildForBuild;_SetLatestTargetFrameworkVersion;_ResolveMonoAndroidSdks">
 	<ItemGroup>
 		<!--- List of items we want to trigger a build if changed -->
 		<_PropertyCacheItems Include="BundleAssemblies=$(BundleAssemblies)" />
@@ -1206,7 +1206,7 @@ because xbuild doesn't support framework reference assemblies.
 	<MakeDir Directories="$(_AndroidStampDirectory)" Condition="!Exists('$(_AndroidStampDirectory)')" />
 	<WriteLinesToFile
 			File="$(_AndroidBuildPropertiesCache)"
-			Lines="@(_PropertyCacheItems)"
+			Lines="@(_PropertyCacheItems->ToLowerInvariant())"
 			Overwrite="true"
 			WriteOnlyWhenDifferent="true"
 	/>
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1291,13 +1292,13 @@ because xbuild doesn't support framework reference assemblies.
...
 		<_PropertyCacheItems Include="TargetFrameworkVersion=$(TargetFrameworkVersion)" />
-		<_PropertyCacheItems Include="AndroidCreatePackagePerAbi=$(AndroidCreatePackagePerAbi)" />
-		<_PropertyCacheItems Include="AndroidGenerateJniMarshalMethods=$(AndroidGenerateJniMarshalMethods)" />
+		<_PropertyCacheItems Include="AndroidCreatePackagePerAbi=$(AndroidCreatePackagePerAbi.ToLowerInvariant())" />
+		<_PropertyCacheItems Include="AndroidGenerateJniMarshalMethods=$(AndroidGenerateJniMarshalMethods.ToLowerInvariant())" />
 		<_PropertyCacheItems Include="OS=$(OS)" />
-		<_PropertyCacheItems Include="DesignTimeBuild=$(DesignTimeBuild)" />
-		<_PropertyCacheItems Include="AndroidIncludeDebugSymbols=$(AndroidIncludeDebugSymbols)" />
+		<_PropertyCacheItems Include="DesignTimeBuild=$(DesignTimeBuild.ToLowerInvariant())" />
+		<_PropertyCacheItems Include="AndroidIncludeDebugSymbols=$(AndroidIncludeDebugSymbols.ToLowerInvariant())" />
```
jonathanpeppers: Should we consider just doing `.ToLowerInvariant()` on every line of the `build.props` file?

dellis1972: We could. I have no problem with trying to see if it works and what impact it has :)

PR #3895, opened 2019-11-11
[Xamarin.Android.Build.Tasks] remove the old desugar support

jonathanpeppers:
Context: https://source.android.com/setup/build/jack
Context: https://android-developers.googleblog.com/2017/03/future-of-java-8-language-feature.html

Before d8/r8 was implemented in Xamarin.Android, we were distributing
`desugar_deploy.jar` to support the `$(AndroidEnableDesugar)` option.
This was the initial implementation in the "Jack" toolchain to enable
Java 8 language features on Android.

"Jack" was deprecated in March 2017, and we now have support for d8/r8
in Xamarin.Android. d8/r8 has desugaring built-in, and desugars by
default. We should remove `desugar_deploy.jar` and the `<Desugar/>`
MSBuild task.

However, this also introduces a new default, if you have:

    <AndroidDexTool>dx</AndroidDexTool>
    <AndroidEnableDesugar>True</AndroidEnableDesugar>

This will implicitly turn on `AndroidDexTool=d8`, since we no longer
have desugaring support with dx.

Another case is:

    <AndroidDexTool>d8</AndroidDexTool>
    <AndroidLinkTool>proguard</AndroidLinkTool>
    <AndroidEnableDesugar>True</AndroidEnableDesugar>

ProGuard can't handle Java 8 language features as input, so this
configuration will opt into r8 implicitly with `AndroidLinkTool=r8`.

Most users will not hit these cases, however:

* `AndroidDexTool=d8` is the current default in d16-5/master.
* It is *possible* to use ProGuard with d8, but you should just use r8
  instead.
* Setting `AndroidEnableDesugar=True` is not available the UI in VS
  Project Options, so it should not be common.

I only found these were needed with the `BuildTest.Desugar` test that
runs a matrix of these options.


dellis1972:
@jonathanpeppers the debugging targets has a reference to `AndroidEnableDesugar` when calling the `D8 task. We should clean up the side as well in this PR.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -3503,11 +3503,10 @@ public void RunXABuildInParallel ()
 		}
 
 		[Test]
-		public void ProguardBOMError ([Values ("dx")] string dexTool, [Values ("proguard")] string linkTool)
+		public void ProguardBOMError ([Values ("dx", "d8")] string dexTool, [Values ("proguard")] string linkTool)
 		{
 			var proj = new XamarinAndroidApplicationProject () {
 				IsRelease = true,
-				EnableDesugar = true, //It is certain this test would fail without desugar
```
jonpryor: Should we just *remove* this property as well?

jonpryor:
The `ThirdPartyNotice.txt` info should also be removed:

https://github.com/xamarin/xamarin-android/blob/f82cd9d696fe9c76b1adc71433a67db9852beb9e/build-tools/xaprepare/xaprepare/ThirdPartyNotices/Xamarin.Android.Build.Tasks.cs#L36-L45

This will also cause the top-level `ThirdPartyNotices.txt` to be updated, removing:

  * https://github.com/xamarin/xamarin-android/blob/f82cd9d696fe9c76b1adc71433a67db9852beb9e/ThirdPartyNotices.txt#L17
  * https://github.com/xamarin/xamarin-android/blob/f82cd9d696fe9c76b1adc71433a67db9852beb9e/ThirdPartyNotices.txt#L464-L670

jonathanpeppers:
@dellis1972 

> the debugging targets has a reference to AndroidEnableDesugar when calling the `D8 task. We should clean up the side as well in this PR.

It is actually still valid to set `AndroidEnableDesugar=False`, which would pass this flag to D8: https://github.com/xamarin/xamarin-android/blob/f82cd9d696fe9c76b1adc71433a67db9852beb9e/src/Xamarin.Android.Build.Tasks/Tasks/D8.cs#L79-L80

The code in monodroid, seems like it is still OK.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
jonathanpeppers: There is still one test that will use it:
https://github.com/xamarin/xamarin-android/blob/f82cd9d696fe9c76b1adc71433a67db9852beb9e/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L3542-L3547

This test would fail otherwise, because it has a library with Java 8 features.

PR #3897, opened 2019-11-11
[build] use "clean: resources" instead of "clean: all"

jonathanpeppers:
Context: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#workspace
Context: https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/pipeline-options-for-git?view=azure-devops

We are commonly hitting this error on Windows:

    The process cannot access the file 'E:\A\_work\237\s' because it is being used by another process.

It appears that using `all` will make Azure DevOps delete the entire
directory, where (perhaps?) we can make it only run commands such as
`git clean -dxff` instead?

The only side effect is that `$(Build.BinariesDirectory)` will never be
deleted, but it appears that our build never uses that directory at
all.

NOTE: I have no idea if this will actually fix anything.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -55,7 +55,7 @@ stages:
     timeoutInMinutes: 240
     cancelTimeoutInMinutes: 5
     workspace:
-      clean: all
+      clean: resources
     steps:
     - checkout: self
       submodules: recursive
@@ -155,7 +155,7 @@ stages:
     timeoutInMinutes: 360
     cancelTimeoutInMinutes: 5
     workspace:
-      clean: all
+      clean: resources
     steps:
     - checkout: self
       submodules: recursive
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
This setting did this:
```
##[debug]Primary repository: xamarin/xamarin-android. repository type: GitHub
Prepare build directory.
##[debug]Delete existing artifacts directory: 'E:\A\_work\105\a'
##[debug]Deleting artifacts directory: 'E:\A\_work\105\a'
##[debug]Creating artifacts directory: 'E:\A\_work\105\a'
##[debug]Delete existing test results directory: 'E:\A\_work\105\TestResults'
##[debug]Deleting test results directory: 'E:\A\_work\105\TestResults'
##[debug]Creating test results directory: 'E:\A\_work\105\TestResults'
##[debug]Delete existing source directory: 'E:\A\_work\105\s'
##[debug]Deleting source directory: 'E:\A\_work\105\s'
##[debug]Creating source directory: 'E:\A\_work\105\s'
```
The build machine, `DDMBLDW193`, appears to not be having this issue right now.

jonathanpeppers:
After doing more research, I don't think this actually helps anything.

Issue #3636, opened 2019-09-16
Aapt and Aapt2 MSBuild tasks do not yet use -G and --proguard options to help generate rules when ProGuard or R8 is enabled


brendanzagaeski:
Previous related reports:

  * <https://github.com/xamarin/Xamarin.Forms/issues/2709>
  * <https://developercommunity.visualstudio.com/content/problem/365357/problem-with-android-linking-in-forms-app.html>
  * <https://github.com/xamarin/AndroidSupportComponents/issues/141>
  * [VS bug <span>#</span>825420](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/825420)
  * <https://developercommunity.visualstudio.com/content/problem/618060/xmarineandroid-execution-error.html>

### Steps to reproduce

 1. Create a new Single View Android app.
 2. Set **Code shrinker** to **ProGuard** or **r8** in the **Android Options** section of the project properties for the Release configuration.
 3. Set the **Compile using Android version: (Target Framework)** to **Android 8.1 (Oreo)** as a way to avoid the [missing *proguard.txt* files in the *monoandroid90* target framework directories in the Xamarin.Android.Support NuGet packages](https://developercommunity.visualstudio.com/content/problem/679674/r8-linker-breaks-checkboxpreference-animation.html).
 4. Build and run the app in the Release configuration on an emulator or device.

Example project: [SingleViewAndroidApp1.zip](https://github.com/xamarin/xamarin-android/files/3614846/SingleViewAndroidApp1.zip)

### Expected behaviors

In a Gradle build in Android Studio for a matching project, some important types are preserved by *app\\build\\intermediates\\proguard-rules\\release\\aapt\_rules.txt*. Based on those results, I believe the expected behaviors are:

  * The `Aapt2` task passes the `--proguard aapt_rules.txt` option to `aapt2.exe` and then passes the resulting *aapt_rules.txt* file to ProGuard or R8.

    (Or for `aapt`, the `Aapt` task passes the `-G aapt_rules.txt` option to `aapt.exe` and then passes the resulting *aapt_rules.txt* file to ProGuard or R8.)

  * The app runs without error.

### Actual behaviors

  * The `Aapt2` task does not pass the `--proguard` option to `aapt2.exe`, and the `Aapt` task does not pass the `-G` option to `aapt.exe`.

  * The app exits during startup because some required types are missing:

    ```
    UNHANDLED EXCEPTION:
[...]

PR #3899, opened 2019-11-11
[Xamarin.Android.Build.Tasks] use aapt2 option to generate proguard rules

jonathanpeppers:
Context: https://developer.android.com/studio/build/shrink-code#configuration-files
Fixes: http://work.azdo.io/825420
Fixes: https://github.com/xamarin/xamarin-android/issues/3636

Let's say you used the following code fragment in an Android layout:

    <android.support.v7.widget.Toolbar />

Currently, you would need to add a ProGuard rule regardless if you
were using ProGuard or R8.

Fortunately, `aapt2` has a command-line switch:

    > aapt2 link -h
    ...
    --proguard arg    Output file for generated Proguard rules.

According to the Google docs for R8:

    AAPT2 generates keep rules based on references to classes in your
    app’s manifest, layouts, and other app resources. For example, AAPT2
    includes a keep rule for each Activity that you register in your app’s
    manifest as an entry point.

I added a new `ProguardRuleOutput` property for `<Aapt2Link/>` and set
it in cases where `$(AndroidLinkTool)` is not blank. I added this file
to `@(ProguardConfiguration)`, so it will get picked up by ProGuard or
R8.

I updated an existing test for this scenario, checking that a type is
generated in the rules and exists in the final `classes.dex` file.


dellis1972:
Weird error on windows?
```
/Library/Frameworks/Mono.framework/Versions/6.4.0/lib/mono/msbuild/Current/bin/NuGet.targets(123,5): error : The given key '(Xamarin.Android.Support.Constraint.Layout.Solver, 1.0.2.2)' was not present in the dictionary. [/Users/runner/runners/2.160.0/work/1/s/bin/TestRelease/temp/CheckLintErrorsAndWarnings/UnnamedProject.csproj]

```

jonathanpeppers:
The test failure is probably a random issue with NuGet on Mac. If it comes up again, we might look into adding `[NonParallelizable]`...

PR #3907, opened 2019-11-12
[Xamarin.Android.Build.Tasks] use SRM for <GenerateResourceDesigner/>

jonathanpeppers:
I noticed the `<GenerateResourceDesigner/>` MSBuild task had a couple
things we could change to speed it up:

* Use System.Reflection.Metadata instead of Mono.Cecil
* Only look at "MonoAndroid" assemblies, so we don't have to load the
  BCL or NetStandard assemblies.
* Remove System.Linq usage in favor of `foreach`

The SRM port was straightforward, as the code only needs to look at
type and field names.

I could also simply use `@(_MonoAndroidReferencePath)` over
`@(_ReferencePath)`.

However, I had to split up `@(_MonoAndroidReferencePath)` and include
`@(_ReferenceDependencyPaths)` in a new item group:
`@(_MonoAndroidReferenceDependencyPaths)`.

Otherwise, this test case failed:

    App -> LibraryA -> LibraryB

The app's `Resource.designer.cs` was attempting to set values in
LibraryB, even though it did not reference it. LibraryB was in
`@(_ReferenceDependencyPaths)`.

These changes will particularly help design-time (or incremental)
builds while editing `AndroidResource` files.
## Notes about F# ##

When refactoring the code, I found what appeared to be a dead code path
that used to be for F# support:

https://github.com/xamarin/xamarin-android/blob/e2e1a846ca6136f0a3bac5ffb48277938ed5c204/src/Xamarin.Android.Build.Tasks/Utilities/ResourceDesignerImportGenerator.cs#L54-L58

Current F# projects make use of the
Xamarin.Android.FSharp.ResourceProvider NuGet Package:

https://github.com/xamarin/Xamarin.Android.FSharp.ResourceProvider/blob/9d3661bd5f6db515377499b7285423d765069ce0/ResourceTypeProvider.fs

The way it works:

* `Resource.designer.cs` is generated as C# code
* Xamarin.Android.FSharp.ResourceProvider compiles this file

You reference the C# class from F# such as:

    type Resources = MyApp.Android.Resource
    [<assembly: Android.Runtime.ResourceDesigner("MyApp.Android.Resources", IsApplication=true)>]

Thus we can remove the old code path, since both languages use the
default C# `Resource.designer.cs`.

## Results ##

Using the Xamarin.Forms integration project in this repo:

    Before:
    465 ms  GenerateResourceDesigner                   1 calls
    After:
    371 ms  GenerateResourceDesigner                   1 calls

This seems like it will save ~100ms on design-time builds and
incremental builds.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/ResourceDesignerImportGenerator.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/ResourceDesignerImportGenerator.cs
@@ -1,8 +1,10 @@
 using System;
 using System.CodeDom;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
-using Mono.Cecil;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
 using Microsoft.Build.Utilities;
 
 namespace Xamarin.Android.Tasks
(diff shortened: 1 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Utilities/ResourceDesignerImportGenerator.cs
```diff
@@ -34,52 +36,71 @@ public ResourceDesignerImportGenerator (string ns, CodeTypeDeclaration applicati
...
-				if (att == null)
-					continue;
-				if ((bool) att.Properties.First (p => p.Name == "IsApplication").Argument.Value == true)
-					continue; // application resource IDs are constants, cannot merge.
-				var td = assembly.Modules.SelectMany (m => m.Types).FirstOrDefault (
-					t => t.FullName == (string)att.ConstructorArguments.First ().Value);
-
-				// F# has no nested types, so we need special care.
-				if (td.NestedTypes.Any ())
-					CreateImportFor (true, td.NestedTypes, method);
```
jonathanpeppers: TODO, I need to figure out the exact state of F# here. Do F# projects even call this or do they use this?

https://www.nuget.org/packages/Xamarin.Android.FSharp.ResourceProvider/

jonathanpeppers:
Looks like there are some test failures I need to look into.

jonathanpeppers:
The builds look good after re-run, just a failure on a duplicate build artifact name 👍

PR #3909, opened 2019-11-13
[docs] Add release note for PR 3899

brendanzagaeski:
Context: https://github.com/xamarin/xamarin-android/pull/3899

This is a first example of a new thing I'm trying to help the team add release notes along the way for all PRs that change user-facing behavior. There isn't any terribly special reason I picked this item as the first example. It was just an item where I saw a notification when the PR was merged, so it reminded me that I could add a release note.

### DIFF
--- a/Documentation/release-notes/3899.md
+++ b/Documentation/release-notes/3899.md
@@ -1,0 +1,4 @@
+### Issues fixed
+
+  * [GitHub 3636](https://github.com/xamarin/xamarin-android/issues/3636):
+    *Error inflating class android.support.v7.widget.FitWindowsLinearLayout ---> Java.Lang.ClassNotFoundException: Didn't find class "android.support.v7.widget.FitWindowsLinearLayout"* and similar errors could prevent apps built with **Code shrinker** set to **ProGuard** or **r8** from running successfully unless custom ProGuard rules were added to the project by hand. In Xamarin.Android 10.x, projects configured to use AAPT2 now pass the `--proguard` option to AAPT2 to generate default ProGuard rules that more closely match the defaults for Android Studio projects, removing the need to add custom rules for common scenarios like the `FitWindowsLinearLayout` type from the Android Support Library.


brendanzagaeski:
Something occurred to me after setting up this PR. Using a single file for the release notes across all PRs will lead to lots of unhelpful merge conflicts.

So I think I should revise my suggestion for how to add release notes going forward. I think it would be better to use a separate file for each PR.

This will prevent merge conflicts while also keeping the following features:

  * PRs can include release notes alongside corresponding code changes.
  * Any contributor can propose updates to the release notes for a PR after the PR is merged, and the PR author can review the proposed updates.
  * Release notes can include images.
  * Release notes can use all of the normal Markdown syntax.

I'm thinking that each file can be named with the PR number. This will require adding the notes after the PR is first opened, but I think that's worth the advantage of having a simple way to avoid file name conflicts.

Example for PR 3899:

```patch
diff --git a/Documentation/release-notes/3899.md b/Documentation/release-notes/3899.md
new file mode 100644
index 00000000..aa290b57
--- /dev/null
+++ b/Documentation/release-notes/3899.md
@@ -0,0 +1,4 @@
+### Issues fixed
+
+  * [GitHub 3636](https://github.com/xamarin/xamarin-android/issues/3636):
+    *Error inflating class android.support.v7.widget.FitWindowsLinearLayout ---> Java.Lang.ClassNotFoundException: Didn't find class "android.support.v7.widget.FitWindowsLinearLayout"* and similar errors could prevent apps built with **Code shrinker** set to **ProGuard** or **r8** from running successfully unless custom ProGuard rules were added to the project by hand. In Xamarin.Android 10.x, projects configured to use AAPT2 now pass the `--proguard` option to AAPT2 to generate default ProGuard rules that more closely match the defaults for Android Studio projects, removing the need to add custom rules for common scenarios like the `FitWindowsLinearLayout` type from the Android Support Library.
```

jonpryor:
> I'm thinking that each file can be named with the PR number

*An* issue with this approach is that it requires knowing the PR number before the file can be created, which means a "work locally, create PR, watch it build" workflow would no longer be "sufficient" when release notes need updating, because you won't know what the PR number is until after you create the branch & push.

Is the number *required*?  Or would *any* "unique" string be sufficient, e.g. the branch name used, or `jonp-1.md`, or....something?

jonpryor:
@brendanzagaeski: we may want to take a 'poll' of other projects (VSCode, corefx, others?) to see if they do anything special here.  Maybe there are some good ideas we should follow?

brendanzagaeski:
> we may want to take a 'poll' of other projects (VSCode, corefx, others?) to see if they do anything special here. Maybe there are some good ideas we should follow?

Good thought. I looked at those other projects before I made my initial suggestion to use a single file for the release notes ([reference](https://dev.azure.com/devdiv/DevDiv/_wiki/wikis/DevDiv.wiki/3137/2019-07-01-Weekly-Minutes)), but it's a good time to check back, see if anything has changed, and record a few points for future reference about what some other projects do.

### Summary

Visual Studio Code's process is a fairly close match for what we're currently trying. One important difference is that the draft release notes are submitted directly to the documentation repository instead of the source code repo. This avoids the main concern with merge conflicts because source code PRs don't need to worry about merge conflicts in the release notes.

Xamarin.Android could achieve the same result by keeping release notes commits separate from source code PRs. The downside would be that source code PRs wouldn't be able to modify the release notes. This general topic is also related to the xamarin-engineering-docs-pr repo.

### microsoft/vscode

*Previous status*: When I last looked, the approach for Visual Studio Code was that draft release notes were added directly to the [separate documentation repo](https://github.com/microsoft/vscode-docs/tree/master/release-notes). For example, a [source code commit](https://github.com/microsoft/vscode/commit/146115b78cc7298caa88b8cf67e0c2b58e055d4b) had a [separate documentation commit](https://github.com/microsoft/vscode-docs/commit/505fe7915ac79a6d146264477d8c7499dc7cff66) by the same author.

*Current status*: The approach remains the same. For example, a [recent source code commit](https://github.com/microsoft/vscode/commit/844335b389f8e3a2d4410700777b23e5db3fe459) from the October 2019 release has a [separate documentation commit](https://github.com/microsoft/vscode-docs/commit/aab2b3964fea7ac201a05c102b13b31c335d7d33) by the same author.

### dotnet/corefx

It looks like the process might start with collecting items outside of GitHub for a blog post. For example, for `MetadataLoadContext`:

  * December 4: https://devblogs.microsoft.com/dotnet/announcing-net-core-3-preview-1-and-open-sourcing-windows-desktop-frameworks/
  * December 5: https://github.com/dotnet/docs/commit/1e4dc480bccce0d74b3928152f695d16e98eb1e4

Based on searching dotnet/corefx and dotnet/core, I believe these notes first appeared in the blog post.

At a glance, it looks like [Roslyn](https://github.com/dotnet/docs/blob/master/docs/csharp/whats-new/csharp-8.md), [ASP.NET Core](https://github.com/aspnet/AspNetCore.Docs/blob/master/aspnetcore/release-notes/aspnetcore-3.0.md), and [Entity Framework Core](https://github.com/aspnet/EntityFramework.Docs/blob/master/entity-framework/core/what-is-new/ef-core-3.0/index.md) use a similar combination of blog posts and "what's new" articles.

### microsoft/TypeScript

I searched for a few phrases from the [TypeScript 3.7 release notes](https://github.com/microsoft/TypeScript-Handbook/blob/master/pages/release%20notes/TypeScript%203.7.md) to see if they might have been gathered from smaller earlier commits. I didn't find any results.

### Possible next steps

If we'd be interested to know more about how .NET Core, Roslyn, ASP.NET Core, EF Core, or TypeScript gathers release notes outside of GitHub, I can ask around.

brendanzagaeski:
> Is the \[PR\] number _required_? Or would _any_ "unique" string be sufficient, e.g. the branch name used, or `jonp-1.md`, or....something?

Any unique string would work. I think `System.Guid.NewGuid()` would work too, for example.

jonpryor:
I'm not fond of using a separate documentation repo any more than we currently are, and would thus endorse using a "unique" string for the filename.

PR #3913, opened 2019-11-13
[Xamarin.Android.Build.Tasks] AndroidResource changes do not affect _GenerateJavaStubs

jonathanpeppers:
When editing `AndroidResource` files and saving in the IDE, I noticed
that `_GenerateJavaStubs` always runs.

It doesn't seem like we actually need to do that?

The value for `Inputs` makes sense when you look at how the task is
called:

    <Target Name="_GenerateJavaStubs"
        Inputs="$(MSBuildAllProjects);@(_ResolvedUserMonoAndroidAssemblies);$(_AndroidManifestAbs);$(_AndroidBuildPropertiesCache);@(_AndroidResourceDest)">
      <GenerateJavaStubs
        ...
        ResourceDirectory="$(MonoAndroidResDirIntermediate)"
      />
    </Target>

But looking closer, the `ResourceDirectory` is not even used by
`<GenerateJavaStubs/>`. It seems like we can just remove the
`@(_AndroidResourceDest)` item group from the list of `Inputs`.

Testing an `AndroidResource`-change in the Toggl app:

https://github.com/toggl/mobileapp/tree/develop/Toggl.Droid

Before:

    415 ms  GenerateJavaStubs                          1 calls

This task no longer runs. This should be a ~400ms improvement on
design-time builds or incremental builds with `AndroidResource`
changes.

I also added a test checking `AndroidResource` changes and which
targets are skipped during incremental builds.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Designer.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Designer.targets
@@ -80,7 +80,6 @@ Copyright (C) 2016 Xamarin. All rights reserved.
       MergedAndroidManifestOutput="$(IntermediateOutputPath)android\AndroidManifest.xml"
       UseSharedRuntime="$(AndroidUseSharedRuntime)"
       EmbedAssemblies="$(EmbedAssembliesIntoApk)"
-      ResourceDirectory="$(MonoAndroidResDirIntermediate)"
       BundledWearApplicationName="$(BundledWearApplicationPackageName)"
       PackageNamingPolicy="$(AndroidPackageNamingPolicy)"
       ApplicationJavaClass="$(AndroidApplicationJavaClass)"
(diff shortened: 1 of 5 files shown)


dellis1972:
How will this effect situations where `@string`  values are used within the manifest? Logic suggests that it won't matter since the apk will be rebuilt and the new values from resources will be pulled in by `aapt/aapt2`.

dellis1972:
I'm trying to remember why we did that...

jonathanpeppers:
> I'm trying to remember why we did that...

I can find a commit in ancient history circa 2012 where `ResourceDirectory` was introduced: https://github.com/xamarin/monodroid/commit/7538f80217e0e68f785cd9fcdc155779d49efb83

Back in the day there was some kind of splash screen for a trial of Xamarin? It might make sense in that case because it was adding resource files to make that work.

dellis1972:
@jonathanpeppers Yup , looks like the purpose was to allow us to write style information about the splash screen to the resource directory. That is no longer needed so this should be ok to do.

jonathanpeppers:
There is a network-related failure and:
```
(_CreateAndroidDebugSigningKey target) -> 
  /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(2911,2): error ANDKT0000: keytool error: java.lang.Exception: Key pair not generated, alias <androiddebugkey> already exists [/Users/runner/runners/2.160.0/work/1/s/bin/TestRelease/temp/CodeBehind/SuccessfulBuildMany/Release/project/CodeBehindBuildTests.csproj]

    38 Warning(s)
    1 Error(s)
```
I have seen this happen on other PRs, should be unrelated.

Issue #3813, opened 2019-10-17
Android App Bundle: disable types of configuration APKs


uwe-neuronation:
It seems to be not possible to configure the Android App Bundle.

Native Android has an option to **enableSplit** in the build.gradle. Thats missing on Xamarin.

```
android {
    // When building Android App Bundles, the splits block is ignored.
    splits {...}

    // Instead, use the bundle block to control which types of configuration APKs
    // you want your app bundle to support.
    bundle {
        language {
            // Specifies that the app bundle should not support
            // configuration APKs for language resources. These
            // resources are instead packaged with each base and
            // dynamic feature APK.
            enableSplit = false
        }
        density {
            // This property is set to true by default.
            enableSplit = true
        }
        abi {
            // This property is set to true by default.
            enableSplit = true
        }
    }
}
```

https://developer.android.com/studio/projects/dynamic-delivery/configure-base#disable_config_apks

brendanzagaeski:
*Release status update*

A new Preview version has now been published on Windows that includes new functionality for this item. The functionality is not yet included in a Release version. I will update this item again when a Release version is available that includes the functionality.

The change is not yet available on macOS. I will update this item again when a Preview version with the change is available on macOS.

New functionality included in Xamarin.Android 10.2.100.7

New functionality included on Windows in Visual Studio 2019 version 16.6 Preview 1. To try the Preview version that includes the change, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) in [Visual Studio Preview](https://visualstudio.microsoft.com/vs/preview/).

New functionality not yet available on macOS.

* * *

The new functionality allows a [*BundleConfig.json*](https://developer.android.com/studio/build/building-cmdline#bundleconfig) file to be specified for the Xamarin.Android project by using the `AndroidBundleConfigurationFile` MSBuild property:

```xml
<PropertyGroup>
  <AndroidBundleConfigurationFile>BundleConfig.json</AndroidBundleConfigurationFile>
</PropertyGroup>
```

See the [release notes](https://github.com/xamarin/xamarin-android/releases/tag/v10.2.100.7#support-for-android-app-bundle-configuration-files-1021007) for additional information.

brendanzagaeski:
*Release status update*

A new Preview version has now been published on macOS that includes new functionality for this item. The functionality is not yet included in a Release version. I will update this item again when a Release version is available that includes the functionality.

[...]

PR #3915, opened 2019-11-14
Support for custom App Bundle configuration file

mathieubourgeois:
Fixes: https://github.com/xamarin/xamarin-android/issues/3813

Currently, `bundletool` is not configurable in any significant way with Xamarin.Android. However, there are useful properties that can be defined in the configuration file for `bundletool`, especially related to the dimensions on which to split the resulting apks.

Therefore, added a way to set a custom configuration file with `AndroidBundleConfigurationFile`. Xamarin still has final control over the resulting configuration file, and will still set the uncompressed extensions no matter what is set in the configuration file. Knowledge of the underlying configuration is required however, which may be problematic. 

Tested on d16-3 and with it we're able to control the split and the extraction of the native libs.

This was enough to fix our underlying issues for my team, however I'm still open to suggestions on how to adapt it so that it's useful for everyone. I was thinking somebody might want properties for the most common options, which could then be exposed in the IDE, but I was unsure of the best way to proceed from there and still think that having the configuration available to somebody who understands the underlying format is still a useful idea.

### Example `BuildConfig.json`

```
{
    "optimizations": {
        "splits_config": {
            "split_dimension": [
                {
                    "value": "ABI",
                    "negate": false
                },
                {
                    "value": "SCREEN_DENSITY",
                    "negate": true
                },
                {
                    "value": "LANGUAGE",
                    "negate": true
                }
            ]
        },
        "uncompress_native_libraries":
        {
            "enabled":false
        }
    }
}
```


src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
```diff
@@ -78,13 +81,34 @@ public override bool RunTask ()
...
+					try
+					{
+						using (StreamReader file = File.OpenText(CustomBuildConfigFile))
+						using (JsonTextReader reader = new JsonTextReader(file))
+						{
+							json = (JObject)JToken.ReadFrom(reader);
+						}
+					}
+					catch (Exception e)
+					{ } // Do nothing for now
```
dellis1972: If there is an exception we should at least log a warning via `Log.LogCodedWarning` with the details of the exception.

dellis1972:
Thanks for the contribution. @jonathanpeppers is the main developer on the app bundle tooling, so I'll wait for his feedback. But this does seem like a good addition. 

I would ask if we can add some documentation about this new property in https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/BuildProcess.md#packaging-properties.

mathieubourgeois:
Ok great. I'll wait for @jonathanpeppers ' feedback and I will do the requested changes afterwards then.

src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
jonathanpeppers: You might just let this exception `throw` as we generate a reasonable error code by default now: https://github.com/xamarin/xamarin-android/blob/fbe4b44c7b40dcaacdecaed834c631b15490cc52/src/Xamarin.Android.Build.Tasks/Tasks/AndroidTask.cs#L18-L21

mathieubourgeois: Fixed in 478cf6d

src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
```diff
@@ -78,13 +81,34 @@ public override bool RunTask ()
 						uncompressed.Add ("**/*" + extension);
 					}
 				}
-				var json = JsonConvert.SerializeObject (new {
+
+				var json = JObject.FromObject(new { });
+				if(!string.IsNullOrWhiteSpace(CustomBuildConfigFile) && File.Exists(CustomBuildConfigFile))
```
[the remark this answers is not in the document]
mathieubourgeois: Adjusted (and the rest of that source file) in 478cf6d

mathieubourgeois:
Documentation and test added.

For the test, I modified the existing test, as you mentioned. I used the build configuration to stop the aab from splitting along language (i.e. base-en.apk shouldn't appear anymore) and adjusted the test accordingly.

mathieubourgeois:
The build error doesn't seem related to the pull request (seems like it tried to access a resource at the same time as another process)

dellis1972:
Thanks for the contribution :)

brendanzagaeski:
Thanks again for the contribution!

*Release status update*

A new Release version of Xamarin.Android has now been published that includes the new functionality from this item.

New functionality included in Xamarin.Android 10.3.1.0.

New functionality included on Windows in Visual Studio 2019 version 16.6. To get the new version that includes the fix, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) or install the latest version from <https://visualstudio.microsoft.com/downloads/>.

New functionality included on macOS in Visual Studio 2019 for Mac version 8.6. To get the new version that includes the fix, check for the latest updates on the **Stable** [updater channel](https://docs.microsoft.com/visualstudio/mac/update).

PR #3917, opened 2019-11-15
[Xamarin.Android.Build.Tasks] Add Ability to include additional Modules

dellis1972:
AppBundles can support multiple modules. This PR adds
the ability to specify additional modules to the
`bundle-tool`. These options might be used in the
future to help support Dynamic features.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
@@ -5,6 +5,7 @@ using Newtonsoft.Json.Linq;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 
 namespace Xamarin.Android.Tasks
 {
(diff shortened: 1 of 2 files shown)


dellis1972:
@jonathanpeppers this is mostly in preparation for future work.

src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
```diff
@@ -119,7 +122,11 @@ protected override CommandLineBuilder GetCommandLineBuilder ()
 		{
 			var cmd = base.GetCommandLineBuilder ();
 			cmd.AppendSwitch ("build-bundle");
-			cmd.AppendSwitchIfNotNull ("--modules ", BaseZip);
+			var modules = new List<string> ();
+			modules.Add (BaseZip);
+			if (Modules != null && Modules.Any ())
+				modules.AddRange (Modules);
+			cmd.AppendSwitchIfNotNull ("--modules ", string.Join (",", modules));
```
dellis1972: We might hit command line argument length issues in the future with this if we are not careful.

PR #3918, opened 2019-11-15
[Xamarin.Android.Build.Tasks] Use Google Manifest Merger.

dellis1972:
Lets use the Google Manifest Merger tooling to merger
library Manifest files. This shoud allow developers to
override/remove some unwanted permissions which are
currently automatically imported from support libraries.

Documentation on how to use the new manifest merger tooling can be found [https://developer.android.com/studio/build/manifest-merge](https://developer.android.com/studio/build/manifest-merge)

Because of the limitations in Command line argument lengths under windows
we have to build our own entry point for the manifest merger tool. 
Our entry point takes a single "response" file as an input. 
Each argument needs to be on a separate line for example.

```
--main
AndroidManifest.xml
--libs
obj\Debug\lp\0\ji\AndroidManifest.xml:obj\Debug\lp\2\ji\AndroidManifest.xml
--out
android\AndroidManifest.xml
```

We have to put each argument on its own line because of the way the built in
Manifest Merger Argument parser works. But this does mean we can handle 
very long command line arguments.


dellis1972:
We still need to do a bunch of work within `GenerateJavaStubs` on the Manifest (adding custom providers etc). This is because we need to do things that depend on code/information which `GenerateJavaStubs` generates.

However what we DONT need to do is handle the merging there. We can use the google tool for that.

dellis1972:
There are a number of unit tests that need fixing up because the order of the elements in the manifest files are changing.

build-tools/installers/create-installers.targets
```diff
@@ -186,6 +186,14 @@
...
     <_MSBuildFiles Include="$(MSBuildSrcDir)\java_runtime.dex" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\java_runtime_fastdev.dex" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\manifestmerger\common.jar" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\manifestmerger\guava.jar" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\manifestmerger\kotlin-stdlib-jdk8.jar" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\manifestmerger\kotlin-stdlib.jar" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\manifestmerger\kxml2.jar" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\manifestmerger\manifest-merger.jar" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\manifestmerger\sdk-common.jar" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\manifestmerger\sdklib.jar" />
```
jonathanpeppers: You should use the trick I used with r8.jar: https://github.com/xamarin/xamarin-android/blob/fbe4b44c7b40dcaacdecaed834c631b15490cc52/src/r8/build.gradle#L19-L28

I made a gradle project that brings in the packages, and this `zipTree` code combines them all into a self-contained jar file.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2083,29 +2084,37 @@ because xbuild doesn't support framework reference assemblies.
...
 
+  <ManifestMerger
+    ToolPath="$(JavaToolPath)"
+    JavaOptions="$(JavaOptions)"
+    ManifestMergerJarPath="$(MSBuildThisFileDirectory)manifestmerger"
+    AndroidManifest="$(IntermediateOutputPath)android\AndroidManifest.xml"
+    OutputManifestFile="$(IntermediateOutputPath)android\AndroidManifest.xml"
+    LibraryManifestFiles="@(ExtractedManifestDocuments)"
+  />
+
```
jonathanpeppers: I suspect this will make the `PerformanceTests` I added fail, we might have to wait on this one to get past that: https://github.com/xamarin/xamarin-android/pull/3913

dellis1972:
To work around the windows path length issues I have added a Java class which processes a very simple "response" file. https://github.com/xamarin/xamarin-android/pull/3918/files#diff-5dff12ff96ddcb2413445b9172b7e967.

This Java file is compiled and added to the final `manifestmerger.jar`.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2116,35 +2123,53 @@ because xbuild doesn't support framework reference assemblies.
...
   <ItemGroup>
     <FileWrites Include="@(_TypeMapAssemblySource)" />
+    <FileWrites Include="$(IntermediateOutputPath)AndroidManifest.xml" />
   </ItemGroup>
 
   <Touch Files="$(_AndroidStampDirectory)_GenerateJavaStubs.stamp" AlwaysCreate="True" />
 </Target>
 
+<Target Name="_ManifestMerger"
+    Inputs="$(IntermediateOutputPath)AndroidManifest.xml;@(ExtractedManifestDocuments);$(_AndroidBuildPropertiesCache)"
```
jonathanpeppers: Maybe this needs `$(MSBuildAllProjects)` to match `_GenerateJavaStubs`:
```suggestion
    Inputs="$(IntermediateOutputPath)AndroidManifest.xml;@(ExtractedManifestDocuments);$(_AndroidBuildPropertiesCache);$(MSBuildAllProjects)"
```
I could see a `@(Reference)` or `@(PackageReference)` being removed and it wouldn't retrigger this target. You could end up with a removed library's manifest included.

jonathanpeppers: It would be OK to bump this time, if that is the case: https://github.com/xamarin/xamarin-android/blob/7ef398af1e23e280070300d430c8610ad25593b7/tests/msbuild-times-reference/MSBuildDeviceIntegration.csv#L10

dellis1972: In which case the `ExtractedManifestDocuments` ItemGroup should be updated.

dellis1972: But now I think about it, you make a good point. The Documents themselves won't be... 🤔

dellis1972:
@jonathanpeppers one of the performance tests failed 
```
Exceeded expected time of 9500ms, actual 9660ms
```
this is even with the separate target. It works locally. Should I bump the timeout?

jonathanpeppers:
@dellis1972 yeah `Build_CSProj_Change` is the one I thought would regress: https://github.com/xamarin/xamarin-android/pull/3918#discussion_r353786634

If that is the only one, I think that is definitely worth this feature. Changing the `csproj` isn't really an devloop build in my opinion. Not the same bar as just changing source code like C# or XAML.

Issue #3792, opened 2019-10-15
New targets for easier recording of AOT profiles


radekdoulik:
I plan to create few new target to make it easier for our users to record custom AOT profiles. 

Currently one needs to make multiple steps described in https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/profiling.md#profiling-the-aot

PR #3924, opened 2019-11-18
[Xamarin.Android.Build.Tasks] Add StopAndroidPackage target

radekdoulik:
Add new target to stop the app package on the device/emulator.

Add documentation for the new target

This is one of the steps needed for [recording AOT profiles][1]. It is
also useful for scripts to measure the app performance on a
device or an emulator.

[1]: https://github.com/xamarin/xamarin-android/issues/3792

### DIFF
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -104,6 +104,9 @@ The following build targets are defined for Xamarin.Android projects:
     device or the running emulator. The launch activity can be
     overriden by `AndroidLaunchActivity` property.
 
+-   **StopAndroidPackage** &ndash; Completely stops the application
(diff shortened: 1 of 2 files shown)


Documentation/guides/BuildProcess.md
```diff
@@ -104,6 +104,9 @@ The following build targets are defined for Xamarin.Android projects:
     device or the running emulator. The launch activity can be
     overriden by `AndroidLaunchActivity` property.
 
+-   **StopAndroidPackage** &ndash; Completely stops the application
+    package on the device or the running emulator.
```
jonathanpeppers: It might be nice to comment what the equivalent `adb` command would be.

PR #3927, opened 2019-11-18
Fix fastdev library side loading

grendello:
Override directories must be set up both in the debug app helper and the
monodroid runtime as both of them are loaded separately and they do not depend
on each other. This fixes loading of the side loaded shared libraries from the
internal location where they are copied from the override directories.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -433,6 +433,17 @@ stages:
 
     - template: yaml-templates/apk-instrumentation.yaml
       parameters:
+        condition: and(succeeded(), eq(variables['XA.Commercial.Build'], 'true'))
+        configuration: $(ApkTestConfiguration)
+        testName: Xamarin.Android.JcwGen_Tests_FastDev
+        project: tests/CodeGen-Binding/Xamarin.Android.JcwGen-Tests/Xamarin.Android.JcwGen-Tests.csproj
+        testResultsFiles: TestResult-Xamarin.Android.JcwGen_Tests-$(ApkTestConfiguration).xml
+        artifactName: Xamarin.Android.JcwGen_Tests-Signed.apk
+        artifactFolder: FastDev
+        extraBuildArgs: /p:AndroidFastDeploymentType=Assemblies:Dexes
+
+    - template: yaml-templates/apk-instrumentation.yaml
+      parameters:
         configuration: $(ApkTestConfiguration)
         testName: Xamarin.Android.Locale_Tests
         project: tests/locales/Xamarin.Android.Locale-Tests/Xamarin.Android.Locale-Tests.csproj
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -439,7 +439,7 @@ stages:
         project: tests/CodeGen-Binding/Xamarin.Android.JcwGen-Tests/Xamarin.Android.JcwGen-Tests.csproj
         testResultsFiles: TestResult-Xamarin.Android.JcwGen_Tests-$(ApkTestConfiguration).xml
         artifactName: Xamarin.Android.JcwGen_Tests-Signed.apk
-        artifactFolder: FastDev
+        artifactFolder: FastDev_Assemblies_Dexes
         extraBuildArgs: /p:AndroidFastDeploymentType=Assemblies:Dexes
 
     - template: yaml-templates/apk-instrumentation.yaml
--- a/build-tools/automation/yaml-templates/apk-instrumentation.yaml
+++ b/build-tools/automation/yaml-templates/apk-instrumentation.yaml
@@ -1,13 +1,14 @@
 parameters:
-  configuration: []
-  testName: []
-  project: []
+  configuration: ""
+  testName: ""
+  project: ""
   testResultsFiles: ""
   extraBuildArgs: ""
   testResultsFormat: NUnit
   packageType: Apk
-  artifactName: []
-  artifactFolder: []
+  artifactName: ""
+  artifactFolder: ""
+  condition: succeededOrFailed()
 
 steps:
 - task: MSBuild@1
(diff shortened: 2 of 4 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -431,6 +431,17 @@ stages:
         artifactName: Xamarin.Android.JcwGen_Tests-Signed.apk
         artifactFolder: Default
 
+    - template: yaml-templates/apk-instrumentation.yaml
+      condition: and(succeeded(), eq(variables['XA.Commercial.Build'], 'true'))
```
jonathanpeppers: It seems like the `condition` here has to go under `parameters`? Weird.

image

My [VS Code extension](https://marketplace.visualstudio.com/items?itemName=ms-azure-devops.azure-pipelines) complains about this, too.

build-tools/automation/yaml-templates/apk-instrumentation.yaml
```diff
@@ -1,13 +1,14 @@
 parameters:
-  configuration: []
-  testName: []
-  project: []
+  configuration: ""
+  testName: ""
+  project: ""
```
jonathanpeppers: These first three parameters were set to `[]` by default so you get an error if you forget to set them when using this tempalte. I think that is correct, @pjcollins?

build-tools/automation/yaml-templates/apk-instrumentation.yaml
```diff
@@ -1,13 +1,14 @@
...
   testResultsFiles: ""
   extraBuildArgs: ""
   testResultsFormat: NUnit
   packageType: Apk
-  artifactName: []
-  artifactFolder: []
+  artifactName: ""
+  artifactFolder: ""
```
jonathanpeppers: Same here, `artifactName` and `artifactFolder` are required.

build-tools/automation/yaml-templates/apk-instrumentation.yaml
pjcollins: I think when I originally authored this the usage of `[]` was accidental. As far as I can tell, `[]` is intended to represent an empty array, yet these parameters are strings. I believe we need this change, but I'm not sure why the `[]` syntax didn't produce an error earlier. Azure Pipelines must cache our YAML in some way and the AZP syntax checking must have changed in the somewhat recent past.

There doesn't seem to be a clean way to specify that a template parameter is required. The documentation suggests using a [somewhat convoluted solution](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#required-parameters) for this.

build-tools/automation/yaml-templates/apk-instrumentation.yaml
```diff
@@ -19,20 +20,20 @@ steps:
...
+  condition: ${{ parameters.condition }}
 
 - task: PublishTestResults@2
   displayName: publish ${{ parameters.testName }} results
   inputs:
     testResultsFormat: ${{ parameters.testResultsFormat }}
     testResultsFiles: ${{ parameters.testResultsFiles }}
     testRunTitle: ${{ parameters.testName }}
-  condition: and(succeededOrFailed(), ne('${{ parameters.testResultsFiles }}', ''))
+  condition: ${{ parameters.condition }}
```
pjcollins: The latest attempt had a few errors in places where we aren't passing this parameter:
> `Error: Input required: testResultsFiles`.

Looks like we'll need to bring this check back. Perhaps the following will work?

```yaml
and(${{ parameters.condition }}, ne('${{ parameters.testResultsFiles }}', ''))
```

src/monodroid/jni/debug-app-helper.cc
```diff
@@ -97,7 +109,7 @@ Java_mono_android_DebugRuntime_init (JNIEnv *env, jclass klass, jobjectArray run
 	}
 }
 
-#if defined (ANDROID) && !defined (RELEASE)
```
jonpryor: I'm not sure I'm comfortable with defining `copy_file_to_internal_location()` in RELEASE runtimes.  This feels like a potential security hole for Play Store apps.

grendello: This helper is never [built](https://github.com/xamarin/xamarin-android/blob/master/src/monodroid/CMakeLists.txt#L273-L287) or [used](https://github.com/xamarin/xamarin-android/blob/master/src/java-runtime/java/mono/android/MonoPackageManager.java#L73-L78) in [RELEASE](https://github.com/xamarin/xamarin-android/blob/master/src/java-runtime/java/mono/android/release/BuildConfig.java#L4) builds.

dellis1972:
So the Mac error is 
`error ANDKT0000: keytool error: java.io.EOFException`

interestingly there is a warning further up 

```
message ANDKT0000: Warning: The JKS keystore uses a proprietary format. It is recommended to migrate to PKCS12 which is an industry standard format
```

Which is new. I wonder if this Microsoft version of java has been hardened for FIPS like we had too, or has had something else done to it.

jonpryor:
I've been seeing that ANDKT0000 error *a lot* of late.  I'd love to know why.

src/monodroid/jni/debug-app-helper.cc
```diff
@@ -83,6 +81,20 @@ Java_mono_android_DebugRuntime_init (JNIEnv *env, jclass klass, jobjectArray run
...
+	jstring_wrapper jstr (env);
+	jstr = env->GetObjectArrayElement (externalStorageDirs, 0);
+	androidSystem.set_override_dir (1, utils.strdup_new (jstr.get_cstr ()));
+
+	jstr = env->GetObjectArrayElement (externalStorageDirs, 1);
+	androidSystem.set_override_dir (2, utils.strdup_new (jstr.get_cstr ()));
+
+	for (uint32_t i = 0; i < BasicAndroidSystem::MAX_OVERRIDES; ++i) {
+		const char *p = androidSystem.get_override_dir (i);
+		if (!utils.directory_exists (p))
```
jonpryor: This is slightly concerning, in that it's *imaginable* that `externalStorageDirs` may be null at some point -- especially when we're looking into shared runtime removal/etc. -- in which case it's *imaginable* that `p` will be `nullptr` here, which would (eventually) result in `stat(nullptr, &buf)`, which is....probably "fine"?  But not desirable?

PR #3928, opened 2019-11-19
[build] Update Bootstrap task projects to short-form style

pjcollins:
Updates `Xamarin.Android.BootstrapTasks` and `xa-prep-tasks` to short
form style projects targeting `netstandard2.0`.

A new top level `Directory.Build.props` file has been added to contain
common properties that will be automatically imported everywhere. In the
longer term, we can look to migrate content from `Configuration.props`
to this file.

New `PrepTasksAssembly` and `BootstrapTasksAssembly` properties have
been added so that fewer changes will be required when updating the
output path of these assemblies in the future. All files which reference
these assemblies have been updated to use these new properties.

`RunTests.targets` and `RunApkTests.targets` have been updated to
explicitly import `Directory.Build.props` as we pass these files directly to
`msbuild` invocations in certain cases. `Directory.Build.props` was  not
being implicitly imported in those cases, I believe that implicit import
only occurs when building a `.sln` or `.csproj`.

Finally, `CopyLocalLockFileAssemblies=true` has been moved to the new
`Directory.Build.props` file so that all of our short-form style projects will
inherit this behavior.


Xamarin.Android.Build.Tasks.sln
```diff
@@ -3,12 +3,14 @@ Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.27703.2026
 MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "xa-prep-tasks", "build-tools\xa-prep-tasks\xa-prep-tasks.csproj", "{7CE69551-BD73-4726-ACAA-AAF89C84BAF8}"
```
jonpryor: I think we don't want to add these projects; see also 91467bbdd695c3fb92d2aed3327a3d2c64b01c1e.

Configuration.props
```diff
@@ -12,11 +12,16 @@
       Condition=" Exists('$(MSBuildThisFileDirectory)Configuration.OperatingSystem.props') And '$(DoNotLoadOSProperties)' != 'True' "
   />
   <PropertyGroup>
+    <ConfigurationPropsImported>true</ConfigurationPropsImported>
```
jonpryor: Question: can we rename `Configuration.props` to `Directory.Build.props`? https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019

pjcollins: I started _roughly_ looking at this  in a separate branch in my fork - https://github.com/pjcollins/xamarin-android/tree/dirprops. I'd rather do this separately in the future if we do choose to migrate.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -265,7 +265,7 @@ stages:
       displayName: xabuild Xamarin.Android-Tests
       inputs:
         filename: bin\$(XA.Build.Configuration)\bin\xabuild.exe
-        arguments: Xamarin.Android-Tests.sln /p:Configuration=$(XA.Build.Configuration) /bl:$(System.DefaultWorkingDirectory)\bin\Test$(XA.Build.Configuration)\msbuild-build-tests.binlog
+        arguments: Xamarin.Android-Tests.sln /t:Restore,Build /p:Configuration=$(XA.Build.Configuration) /bl:$(System.DefaultWorkingDirectory)\bin\Test$(XA.Build.Configuration)\msbuild-build-tests.binlog
```
akoeplinger: calling `/t:Restore,Build` is usually a bad idea, better to pass `/restore` instead

pjcollins: Good to know, updated.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Properties/AssemblyInfo.cs
```diff
@@ -4,15 +4,10 @@
...
 
-[assembly: AssemblyTitle ("Xamarin.Android.Tools.BootstrapTasks")]
 [assembly: AssemblyDescription ("")]
-[assembly: AssemblyConfiguration ("")]
-[assembly: AssemblyCompany ("Xamarin Inc.")]
-[assembly: AssemblyProduct ("")]
 [assembly: AssemblyCopyright ("Xamarin Inc.")]
 [assembly: AssemblyTrademark ("Xamarin")]
 [assembly: AssemblyCulture ("")]
-[assembly: AssemblyVersion ("1.0.0.0")]
```
jonathanpeppers: For assemblies like this that we don't ship. Should we just delete the entire `Properties/AssemblyInfo.cs` file?

That is generally what I've don't when porting to SDK-style.

build-tools/scripts/MSBuildReferences.projitems
```diff
@@ -1,13 +1,13 @@
...
   <ItemGroup>
     <!-- Compile against Microsoft.Build* NuGet refs, but do not copy to OutputDir. -->
     <PackageReference Include="Microsoft.Build" Version="15.9.20" IncludeAssets="compile" />
     <PackageReference Include="Microsoft.Build.Framework" Version="15.9.20" IncludeAssets="compile" />
     <PackageReference Include="Microsoft.Build.Tasks.Core" Version="15.9.20" IncludeAssets="compile" />
     <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.9.20" IncludeAssets="compile"/>
-    <PackageReference Include="Mono.Posix.NETStandard" Version="1.0.0" />
-    <PackageReference Include="Xamarin.Build.AsyncTask" Version="0.3.4" />
+    <PackageReference Include="Mono.Posix.NETStandard" Version="1.0.0" GeneratePathProperty="true" />
+    <PackageReference Include="Xamarin.Build.AsyncTask" Version="0.3.4" GeneratePathProperty="true" />
```
jonathanpeppers: What does `%(PackageReference.GeneratePathProperty)` do?

build-tools/scripts/MSBuildReferences.projitems
```diff
@@ -20,4 +20,8 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
   </ItemGroup>
+  <!-- Copy PackageReference content (as specified above) to OutputDir. This no longer happens by default in short-form projects. -->
+  <PropertyGroup>
+      <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
```
jonathanpeppers: Should we put this in `Configuration.props` for all projects at the moment? Less things might break when keeping the old behavior.

Configuration.props
```diff
@@ -12,7 +12,12 @@
...
     <ProductVersion>10.2.99</ProductVersion>
+    <!-- TF and output paths for MSBuild task projects -->
+    <TargetFrameworkNETStandard>netstandard2.0</TargetFrameworkNETStandard>
+    <SupportingBuildTasksOutputRoot>$(MSBuildThisFileDirectory)bin\Build$(Configuration)\</SupportingBuildTasksOutputRoot>
+    <SupportingBuildTasksDirectory>$(SupportingBuildTasksOutputRoot)$(TargetFrameworkNETStandard)\</SupportingBuildTasksDirectory>
```
jonathanpeppers: So this makes our bootstap tasks output to `bin\BuildDebug\netstandard2.0\`?

Would it have been easier to define `<AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>`?

If that is in `Configuration.props`, you probably wouldn't have to change as much in every project.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
```diff
@@ -1,53 +1,23 @@
...
-    <AssemblyName>Xamarin.Android.Build.Tasks</AssemblyName>
-    <FileAlignment>512</FileAlignment>
-    <SchemaVersion>2.0</SchemaVersion>
-    <LibZipSharpBundleAllNativeLibraries>true</LibZipSharpBundleAllNativeLibraries>
-    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <DebugSymbols>True</DebugSymbols>
-  </PropertyGroup>
+<Project>
+
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
```
jonathanpeppers: Did you need to workaround something by importing in a specific order? Or could this have `<Project Sdk="Microsoft.NET.Sdk">`?

build-tools/scripts/MSBuildReferences.projitems
```diff
@@ -20,4 +20,8 @@
...
+  </PropertyGroup>
 </Project>
```
dellis1972: @pjcollins So I looked at the issue with the tests not running in VSForMac and I had to apply this patch https://gist.github.com/dellis1972/e9961b958e78c3ee45ed608617cb1335. Not sure if this is something you want to include in the PR or if I should do a separate one. 

Oddly the tests work fine in VSCode on Mac without this change.

build-tools/scripts/MSBuildReferences.projitems
pjcollins: This sets an msbuild property to the package path in the `nuget.g.props` file that is imported by the project. It is used later in  https://github.com/xamarin/xamarin-android/pull/3928/files#diff-e388143943a5a1fdab55ecb6f09169e7R340 to copy "unexpected" NuGet package content to the `OutputDirectory` so that these files can be picked up by our installers.

build-tools/scripts/MSBuildReferences.projitems
pjcollins: We could, though I am a bit worried about introducing some more potentially unexpected behavior. I could go either way here.

Configuration.props
pjcollins: I'd rather not set `AppendTargetFrameworkToOutputPath` in `Configuration.props` just in case we introduce any multi targeting projects or don't want this behavior in projects which import those props for other reasons. We could apply this to the supporting build task projects to simplify this PR, though I'd like to continue using a property which points to the location of `xa-prep-tasks.dll` and `Xamarin.Android.Bootstrap.Tasks.dll` so that we don't have to worry about mass path updates in the future.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
pjcollins: This was done to work around the issue described in https://github.com/microsoft/msbuild/pull/4922

build-tools/scripts/MSBuildReferences.projitems
pjcollins: I'd rather keep that patch in a separate PR. As things stand now the scope of change is probably already too large, and I am considering splitting up the changes in this PR further.

pjcollins:
Sorry the review request here was still a bit premature. There's a size increase in our test APKs being reported that needs further evaluation, and there's somewhat frequent hang in some of the x.a.b.test execution. I'm working to break more pieces out of this to help cut the scope of change/review.

pjcollins:
Moving this into a handful of smaller PRs, and will keep this one around just to introduce changes to the "Supplemental" build task projects (xa-prep-tasks and Bootstrap tasks).

Xamarin.Android.Build.Tasks - https://github.com/xamarin/xamarin-android/pull/4132
Xamarin.Android.Build.Tests - https://github.com/xamarin/xamarin-android/pull/4140
Other desktop unit tests - https://github.com/xamarin/xamarin-android/pull/4143

build-tools/scripts/MSBuildReferences.projitems
pjcollins: I moved this into `src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj` in https://github.com/xamarin/xamarin-android/pull/4132

build-tools/Xamarin.Android.Tools.BootstrapTasks/Properties/AssemblyInfo.cs
pjcollins: I've removed both this and the `AssemblyInfo.cs` file associated with `xa-prep-tasks`.

Directory.Build.props
```diff
@@ -0,0 +1,12 @@
+<Project>
+
+  <PropertyGroup>
+    <TargetFrameworkNETStandard>netstandard2.0</TargetFrameworkNETStandard>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <BuildOutputDirectory>$(MSBuildThisFileDirectory)bin\$(Configuration)\</BuildOutputDirectory>
+    <SupplementalBuildOutputDirectory>$(MSBuildThisFileDirectory)bin\Build$(Configuration)\</SupplementalBuildOutputDirectory>
+    <TestOutputDirectory>$(MSBuildThisFileDirectory)bin\Test$(Configuration)\</TestOutputDirectory>
+    <SupportingBuildTasksDirectory>$(SupplementalBuildOutputDirectory)$(TargetFrameworkNETStandard)\</SupportingBuildTasksDirectory>
```
jonathanpeppers: Comment on naming: looks like two different adjectives are used here: Supplemental vs Supporting

Could these be `$(BootstrapOutputDirectory)` and `$(BootstrapTasksDirectory)` to match what we already have for `*.BootstrapTasks.csproj`?

It might also make sense to have more properties for the two assemblies:
```xml
<PrepTasksAssembly>$(BootstrapTasksDirectory)xa-prep-tasks.dll</PrepTasksAssembly>
<BootstrapTasksAssembly>$(BootstrapTasksDirectory)Xamarin.Android.Tools.BootstrapTasks.dll</BootstrapTasksAssembly>
```

pjcollins: Good suggestion, I definitely like this more. Seems best to surface properties for the two task assemblies rather than their directory. I may keep `SupplementalBuildOutputDirectory` as named and just remove the `SupportingBuildTasksDirectory` property outright in favor of the two assembly props mentioned above if that sounds good?

jonathanpeppers: Sounds good 👍

Directory.Build.props
```diff
@@ -0,0 +1,15 @@
...
+    <TargetFrameworkNETStandard>netstandard2.0</TargetFrameworkNETStandard>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <BuildOutputDirectory>$(MSBuildThisFileDirectory)bin\$(Configuration)\</BuildOutputDirectory>
+    <BootstrapOutputDirectory>$(MSBuildThisFileDirectory)bin\Build$(Configuration)\</BootstrapOutputDirectory>
+    <TestOutputDirectory>$(MSBuildThisFileDirectory)bin\Test$(Configuration)\</TestOutputDirectory>
+    <BootstrapTasksAssembly>$(BootstrapOutputDirectory)$(TargetFrameworkNETStandard)\Xamarin.Android.Tools.BootstrapTasks.dll</BootstrapTasksAssembly>
+    <PrepTasksAssembly>$(BootstrapOutputDirectory)$(TargetFrameworkNETStandard)\xa-prep-tasks.dll</PrepTasksAssembly>
+    <!-- Copy PackageReference content to OutputDir for our build tasks, tests, and installer creation logic. This no longer happens by default in short-form projects. -->
+    <CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>
+  </PropertyGroup>
```
jonathanpeppers: Maybe TODO on another PR. Can we kill `Configuration.props` in favor of the new hotness: `Directory.Build.props`?

pjcollins: Yeah I definitely think that's a good longer term goal. We may want to do it in a couple of stages though, I imagine the diff is going to be pretty ugly (not to mention the changes required in external sources...).

PR #3936, opened 2019-11-19
Set the default $(AndroidSupportedTargetAotAbis) value

radekdoulik:
Recently I run into missing AOT cross compilers after successful
build.

Turned out the prepare step (`make prepare`) uses
`$(AndroidSupportedTargetAotAbis` value and it wasn't set on my
system.

Not sure why it was working before for me (maybe I had
`Configuration.Override.props` in place?), but now when the AOT is
non-enterprise feature, we should set that.

This is how part of my prepare step looked like:

    =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
    Configured targets
    =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=

    Enabled Mono Android runtime ABIs:
      * armeabi-v7a
      * arm64-v8a
      * x86

    Enabled Mono host runtimes:
      * Darwin

    Enabled Mono cross compilers:
      NONE

    Enabled LLVM cross compilers:
      NONE


radekdoulik:
the failing runs are ok, they just fail to upload the results => good to go

PR #3940, opened 2019-11-20
Fix wrong AOT mode symbol being set

mathieubourgeois:
Since https://github.com/xamarin/xamarin-android/commit/decfbccf3e99449c9a8d6afef113c9339f1bc85b, the runtime code seems to expect that `mono_aot_mode_name` is a linked symbol that's written by the build system. It also expects it to be a string that's written, and the code tries to understand in what AOT mode it's supposed to be based on the first character (`n`ormal, `h`ybrid or `f`ull). However, we currently convert the AotMode enum to a `uint` before converting it to a string, meaning instead of writing `normal`, we write `1`. This causes issues with people receiving the following line in their logs :

`W/monodroid(7449): Unknown Mono AOT mode: 1`

This can be confirmed by analyzing the `environment.{env}.s` files generated in the intermediary folder :

```
 .section .rodata..L.str.2,"aMS",@progbits,1
 .type .L.str.2, @object
.L.str.2:
 .asciz "1"
 .size .L.str.2, 2
 .section .data.mono_aot_mode_name,"aw",@progbits
 .global mono_aot_mode_name
mono_aot_mode_name:
 .xword .L.str.2
```

Fix by making sure that we convert the enum to a string instead of the uint. Based on a cursory analysis of the code, this would not affect normal AOT mode, since the call to get_aot_mode seems to only be used for something if aot is not disabled and isn't "normal" (i.e. hybrid or full)


mathieubourgeois:
Just as a small addendum, I have also noticed that since https://github.com/xamarin/xamarin-android/commit/6be4bdbcfd9f2fbe6267aceb934394bcdd0f13b4#diff-85c855eb9ec814df87fe0f2bf21b9311, it looks like the warning log has been inverted to `if (aotMode == MonoAotMode::MONO_AOT_MODE_LAST)`, which means the log will probably complain that the AOT mode is wrong even if it's entirely correct (and will say it's right when it says it's wrong. I don't think that's intended.

brendanzagaeski:
*Release status update*

A new Preview version has now been published that includes the fix from this item. The fix is not yet included in a Release version. I will update this item again when a Release version is available that includes the fix.

brendanzagaeski:
*Release status update*

A new Release version has now been published on Windows that includes the fix from this item. The fix is not yet published in a Release version on macOS. I will update this item again when a Release version is available on macOS that includes the fix.

Fix included in Xamarin.Android 10.2.0.100.

Fix included on Windows in Visual Studio 2019 version 16.5. To get the new version that includes the fix, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) or install the latest version from <https://visualstudio.microsoft.com/downloads/>.

(Fix also included on macOS in Visual Studio 2019 for Mac version 8.5 Preview 1 and higher. To try the Preview version that includes the fix, check for the latest updates on the **Preview** [updater channel](https://docs.microsoft.com/visualstudio/mac/update).)

brendanzagaeski:
*Release status update*

A new Release version has now been published on macOS that includes the fix from this item.

Fix included in Xamarin.Android 10.2.0.100.

Fix included on macOS in Visual Studio 2019 for Mac version 8.5. To get the new version that includes the fix, check for the latest updates on the **Stable** [updater channel](https://docs.microsoft.com/visualstudio/mac/update).

(Fix also included on Windows in Visual Studio 2019 version 16.5 and higher. To get the new version that includes the fix, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) or install the latest version from <https://visualstudio.microsoft.com/downloads/>.)

PR #3941, opened 2019-11-20
[monodroid] don't start the debugger if not configured

jonathanpeppers:
In a Debug build of `samples/HelloWorld`, startup was taking way longer
than a Release build:

    Debug build:
    ActivityTaskManager: Displayed com.xamarin.android.helloworld/example.MainActivity: +2s689ms
    Release build:
    ActivityTaskManager: Displayed com.xamarin.android.helloworld/example.MainActivity: +678ms

In this case I was launching the app with no debugger attached. Why is
it *so* much slower?

Reading through `debug.cc` and the `start_connection` method, the
return values and comments did not quite match up--or make sense.

It seemed that even though `conn_port` is 0, it was trying to connect
the debugger? I believe this check needed to actually return 2:

    if (!conn_port)
        return 0;

I changed this line, and reworked things to use an `enum` instead of
an `int`. Everything seems clearer now.

After this change we are now down to:

    ActivityTaskManager: Displayed com.xamarin.android.helloworld/example.MainActivity: +725ms

This is a ~2 second startup improvement for debug builds when no
debugger is attached.

### DIFF
--- a/src/monodroid/jni/debug.cc
+++ b/src/monodroid/jni/debug.cc
@@ -172,12 +172,8 @@ Debug::parse_options (char *options, ConnOptions *opts)
  * start_connection:
  *
  *   Handle the communication with XS on startup. Call process_cmd () for each command received from XS.
- * Returns:
- * - 1 on success
- * - 0 on error
- * - 2 if no connection is neccessary
  */
-int
+DebuggerConnectionStatus
 Debug::start_connection (char *options)
 {
 	int res;
@@ -192,19 +188,21 @@ Debug::start_connection (char *options)
 
 	if (opts.timeout_time && cur_time > opts.timeout_time) {
 		log_warn (LOG_DEBUGGER, "Not connecting to IDE as the timeout value has been reached; current-time: %lli  timeout: %lli", cur_time, opts.timeout_time);
-		return 2;
+		return DebuggerConnectionStatus::Unconnected;
 	}
 
-	if (!conn_port)
-		return 0;
+	if (!conn_port) {
+		// If the port is 0, we should not connect the debugger
+		return DebuggerConnectionStatus::Unconnected;
+	}
 
 	res = pthread_create (&conn_thread_id, nullptr, xamarin::android::conn_thread, this);
 	if (res) {
 		log_error (LOG_DEFAULT, "Failed to create connection thread: %s", strerror (errno));
-		return 1;
+		return DebuggerConnectionStatus::Error;
 	}
 
-	return 0;
+	return DebuggerConnectionStatus::Connected;
 }
 
 void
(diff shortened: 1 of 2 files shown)


src/monodroid/jni/debug.cc
```diff
@@ -214,13 +212,11 @@ Debug::start_debugging_and_profiling ()
 	androidSystem.monodroid_get_system_property (Debug::DEBUG_MONO_CONNECT_PROPERTY, &connect_args);
 
 	if (connect_args != nullptr) {
```
jonathanpeppers: Just another thought, could `connect_args` sometimes be an empty string here, too?

grendello: In theory it could, since we might read its value from the environment file

src/monodroid/jni/debug.hh
```diff
@@ -9,6 +9,14 @@
 #ifdef __cplusplus
 namespace xamarin::android
 {
+	enum class DebuggerResult : unsigned int
+	{
+		Error = 0,
+		Success = 1,
```
jonpryor: I would prefer `Connected` over `Success`.

src/monodroid/jni/debug.hh
```diff
@@ -9,6 +9,14 @@
 #ifdef __cplusplus
 namespace xamarin::android
 {
+	enum class DebuggerResult : unsigned int
```
jonpryor: This isn't a "result" from a debugger (which would be...what? result of "breakpoint reached"?  result of "read this address"?).

This is the "status" of *connecting* the debugger.

I think `DebuggerConnectionStatus` might make more sense in-context.

Furthermore, considering that we have separate "domains" of errors and "success" codes -- granted only one error so far, but that's not necessarily going to remain true going forward -- we might want to make the "domains" more explicit:

  * Error domain: negative values
  * Unconnected, not an error: 0
  * Connected: positive values.

This in turn means changing the underlying type to a signed type.

Thus:

```cpp
enum class DebuggerConnectionStatus : int
{
    Connected = 1,
    Unconnected = 0,
    Error = -1,
    Error_CouldNotCreateThread = -2,
};
```

...though perhaps `Error_CouldNotCreateThread` is a bridge too far?

Assuming we go with that, we could write a function to wrap error checking:

```cpp
inline bool is_error (DebuggerConnectionStatus status)
{
    return status < 0; // ...assuming this can even compile?
}
```

...might be a bridge too far.

Even if we stick with { Error, Unconnected, Connected }, I think -1, 0, and 1 are "saner" values than 0, 1, 2.

jonathanpeppers: I think if there was more than one `Error` case it might be worth making the multiple enum values & helper method.

Since, it's just the once case I went with renaming everything, which I think looks good. 👍

src/monodroid/jni/debug.cc
```diff
@@ -213,14 +211,12 @@ Debug::start_debugging_and_profiling ()
...
 
-	if (connect_args != nullptr) {
-		int res = start_connection (connect_args);
-		if (res != 2) {
-			if (res) {
-				log_fatal (LOG_DEBUGGER, "Could not start a connection to the debugger with connection args '%s'.", connect_args);
-				exit (FATAL_EXIT_DEBUGGER_CONNECT);
-			}
-
+	if (connect_args != nullptr && strcmp (connect_args, "") != 0) {
```
jonpryor: This introduces a memory leak, as we'll skip the `delete[] connect_args` (line 227) if `connect_args` is the empty string.

PR #3942, opened 2019-11-20
[monodroid] add timing message for the debugger

jonathanpeppers:
This adds a useful timing message for Debug builds:

    11-20 16:05:55.683 10984 10984 I monodroid-timing: start_debugging_and_profiling; elapsed: 2s:0::94210

If it really takes 2 seconds like this, it should be easier for us to spot going forward.

### DIFF
--- a/src/monodroid/jni/debug.cc
+++ b/src/monodroid/jni/debug.cc
@@ -237,9 +237,7 @@ Debug::start_debugging_and_profiling ()
 
 	if (XA_UNLIKELY (utils.should_log (LOG_TIMING))) {
 		total_time.mark_end ();
-
-		timing_diff diff (total_time);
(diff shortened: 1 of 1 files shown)


src/monodroid/jni/debug.cc
```diff
@@ -230,6 +234,13 @@ Debug::start_debugging_and_profiling ()
 		}
 		delete[] connect_args;
 	}
+
+	if (XA_UNLIKELY (utils.should_log (LOG_TIMING))) {
+		total_time.mark_end ();
+
+		timing_diff diff (total_time);
+		log_info (LOG_TIMING, "start_debugging_and_profiling; elapsed: %lis:%lu::%lu", diff.sec, diff.ms, diff.ns);
```
jonpryor: This should take a page out of `monodroid-glue.cc` and use the `Timing::info()` member function:

```cpp
Timing::info (total_time, "Debug::start_debugging_and_profiling: end");
```

PR #3952, opened 2019-11-22
[monodroid] check return value of monodroid_get_system_property

jonathanpeppers:
In a26e0dbf, I fixed a 2 second delay on startup, which was somewhat
perplexing... It helped by 2 seconds regardless if a debugger was
present or not.

It turned out that a `""` returned from `monodroid_get_system_property`
was the actual root cause of the issue. Luckily, this issue was only
present on d16-5 and master.

We think checking the return value of `monodroid_get_system_property`
is most appropriate. I changed this call to look the same as the rest
of the calls in `monodroid`.

### DIFF
--- a/src/monodroid/jni/debug.cc
+++ b/src/monodroid/jni/debug.cc
@@ -212,7 +212,7 @@ Debug::start_debugging_and_profiling ()
 	if (XA_UNLIKELY (utils.should_log (LOG_TIMING)))
 		total_time.mark_start ();
 
-	char *connect_args;
+	char *connect_args = nullptr;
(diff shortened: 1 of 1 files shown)


src/monodroid/jni/debug.cc
```diff
@@ -213,13 +213,8 @@ Debug::start_debugging_and_profiling ()
 		total_time.mark_start ();
 
 	char *connect_args;
```
grendello: For absolute correctness, please set `connect_args` to `nullptr` here should the call to `androidSystem.monodroid_get_system_property` ever fail to set the pointer to a meaningful value (it's OK to pass a `nullptr` to `delete`, not so much a pointer with any other value that points to a random location)

brendanzagaeski:
*Release status update*

A new Preview version has now been published that includes the fix from this item.

PR #3937, opened 2019-11-20
[Xamarin.Android.Build.Tasks] Use PKCS12 Key Store format

dellis1972:
We have recently started to see the following error
```
rror ANDKT0000: keytool error: java.lang.Exception: Key pair not generated, alias <androiddebugkey> already exists
```

Looking at the logs, we see the following warning issued further up.

```
message ANDKT0000: androiddebugkey, Nov 25, 2019, PrivateKeyEntry, [/Users/runner/runners/2.160.1/work/1/s/bin/TestRelease/temp/CodeBehind/SuccessfulBuildFew/Release/CommonSampleLibrary/CommonSampleLibrary.csproj] (TaskId:476)
message ANDKT0000: Certificate fingerprint (SHA1): 8F:DE:A1:AF:26:8B:E7:02:4D:D0:57:86:F4:81:4F:B7:4A:80:60:AA [/Users/runner/runners/2.160.1/work/1/s/bin/TestRelease/temp/CodeBehind/SuccessfulBuildFew/Release/CommonSampleLibrary/CommonSampleLibrary.csproj] (TaskId:476)
message ANDKT0000: Warning: The JKS keystore uses a proprietary format. It is recommended to migrate to PKCS12 which is an industry standard format using "keytool -importkeystore -srckeystore /Users/runner/.local/share/Xamarin/Mono for Android/debug.keystore -destkeystore /Users/runner/.local/share/Xamarin/Mono for Android/debug.keystore -deststoretype pkcs12". [/Users/runner/runners/2.160.1/work/1/s/bin/TestRelease/temp/CodeBehind/SuccessfulBuildFew/Release/CommonSampleLibrary/CommonSampleLibrary.csproj] (TaskId:476)
```

While this is not an error we should take note of the warning and update our code. 
So we should make use of the `-storetype` argument to specify which format the keystore should be saved in. The recommendation is `pkcs12`. This only applies to the `debug.keystore` as it

PR #3957, opened 2019-11-25
[tests] workaround ANDKT0000 error

jonathanpeppers:
We are randomly getting the failure:

    Xamarin.Android.Common.targets(2944,2): error ANDKT0000: keytool error: java.lang.Exception: Key pair not generated, alias <androiddebugkey> already exists

We run a lot of tests in parallel and it seems like two builds are
running `_CreateAndroidDebugSigningKey` at once.

There is an `Exists` check in the MSBuild target:

    <Target Name="_CreateAndroidDebugSigningKey"
            Condition="!Exists ('$(_ApkDebugKeyStore)') And '$(AndroidKeyStore)' != 'True' "

To verify this is what might be happening, I added a `LogWarning` call
*inside* the `<AndroidCreateDebugKey/>` MSBuild task. Maybe we could
just return early if we hit this?

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/AndroidCreateDebugKey.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/AndroidCreateDebugKey.cs
@@ -21,6 +21,17 @@ namespace Xamarin.Android.Tasks
 			Validity = 30 * 365; // 30 years
 		}
 
+		public override bool Execute ()
+		{
+			//TODO: this should actually return early and not run base.Execute()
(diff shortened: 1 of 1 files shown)


jonpryor:
Related: PR #3937, which is trying to fix the ANDKT0000 error.

jonathanpeppers:
I did not see the warning mentioned in https://github.com/xamarin/xamarin-android/pull/3937.

That might actually be a better fix to address this? After 3 runs, I couldn't actually get the error here.

PR #3959, opened 2019-11-25
[xabuild] Find Microsoft.Portable.CSharp.targets

jonpryor:
Context: https://github.com/xamarin/xamarin-android/pull/3393/files

Builds for PR #3393 were failing on Windows because
`Microsoft.Portable.CSharp.targets` couldn't be found:

	> bin\Debug\bin\xabuild /bl Xamarin.Android-Tests.sln
	...
	...\xamarin-android\external\Java.Interop\lib\mono.linq.expressions\Mono.Linq.Expressions.csproj(108,3):
	error MSB4019: The imported project "...\xamarin-android\bin\Debug\lib\xamarin.android\xbuild\Microsoft\Portable\v4.0\Microsoft.Portable.CSharp.targets"
	was not found. Confirm that the path in the <Import> declaration is
	correct, and that the file exists on disk.

Turns Out that there are *multiple* `Microsoft` directories of
consequence, and we need *all* of them to build, including:

  * `%ProgramFiles(x86)%\Microsoft Visual Studio\2019\*\MSBuild\Microsoft`
  * `%ProgramFiles(x86)%\MSBuild\Microsoft\Portable`

`xabuild.exe` would symlink the first one into
`bin\$(Configuration)\lib\xamarin.android\xbuild`, but not the latter,
and lack of the `Microsoft\Portable` directory prevented
`Microsoft.Portable.CSharp.targets` from being found.

Update `xabuild.exe` to know about
`%ProgramFiles(x86)%\MSBuild\Microsoft`.

Furthermore, because we now have two `Microsoft` directories,
`Microsoft` must now become a "normal" directory, which contains
symlinks to *all* the known child directories.  To support this,
"invert" the `symbolicLinks` dictionary so that the keys are the
in-tree paths, making it easier to detect when we need to do such
merges.

Finally, add code to detect if an existing "conflicting" directory
(such as `Microsoft`) is a symlink, and remove it if necessary.
This should allow `git pull` + rebuild scenarios to work.

### DIFF
--- a/tools/xabuild/SymbolicLink.cs
+++ b/tools/xabuild/SymbolicLink.cs
@@ -10,6 +10,7 @@ namespace Xamarin.Android.Build
 		public static bool Create (string source, string target)
 		{
 			if (!Directory.Exists (source)) {
+				Directory.CreateDirectory (Path.GetDirectoryName (source));
 				if (Environment.OSVersion.Platform == PlatformID.Win32NT) {
 					return CreateWindowsSymLink (source, target);
 				} else {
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
Just a thought, should we convert `Mono.Linq.Expressions` to an SDK-style project? target netstandard2.0?

https://github.com/jonpryor/mono.linq.expressions/blob/e3372fbb49c59d9616b22333fb22045acad110c1/Mono.Linq.Expressions.csproj#L108

`Microsoft.Portable.CSharp.targets` is working in some fashion, I guess it's only failing for `Profile92`? Which appears to be Windows Phone 8.1? https://portablelibraryprofiles.stephencleary.com/

jonpryor:
Locally, I didn’t see it find `Microsoft.Portable.CSharp.targets` at all...but I didn’t look for every possible instance either.  Maybe it occasionally worked?

Converting `Mono.Linq.Expressions` to an SDK-style project is an interesting idea.  However, that lib hasn’t seen any changes since 2016...  which isn’t an argument *not* to do so...

Issue #3963, opened 2019-11-26
Add new task to watch logcat output and wait for activity displayed


radekdoulik:
The task would wait for activity manager reporting that the launcher activity was displayed.

It will also have a timeout as a parameter as the activity displayed message is not reliable.

PR #3964, opened 2019-11-26
Add new targets for AOT profiling

radekdoulik:
Implements https://github.com/xamarin/xamarin-android/issues/3792

Add 2 targets to start and finish the profiling. The start target also
builds the project with embedded aot profiler and sets its
options. The finish one collects data from the device/emulator and
writes them to custom.aprof file with aprofutil tool.

Add documentation.

Add test.

Add `AndroidEmbedProfilers` to cached properties, to trigger
rebuild when changed.

Originally I wanted to have also single step process target, which
would use `duration=n` option of the aot profiler. After discussion with
@dellis1972, it looks like it might not be usable for IDEs as it might block
the UI thread. So I didn't implement that one.

Instead I might do it differently by introducing new task to watch for
activity displayed message in the adb logcat output, combined with
timeout. https://github.com/xamarin/xamarin-android/issues/3963

### DIFF
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -120,23 +120,25 @@ The following build targets are defined for Xamarin.Android projects:
     IDE when new resources are added to the project.
 
 -   **BuildAndStartAotProfiling** &ndash; Builds the package with
-    embedded aot profiler, sets aot profiler socket port to
-    $(AndroidAotProfilerPort) and starts the *launch* activity.
+    embedded AOT profiler, sets the AOT profiler socket port to
+    `$(AndroidAotProfilerPort)` and starts the *launch* activity.
 
     Added in Xamarin.Android v10.3.
 
--   **FinishAotProfiling** &ndash; Collects the aot profiler data from
+-   **FinishAotProfiling** &ndash; Collects the AOT profiler data from
     the device or the emulator through sockets port
-    $(AndroidAotProfilerPort) and writes them to
-    $(AndroidAotCustomProfilePath).
+    `$(AndroidAotProfilerPort)` and writes them to
+    `$(AndroidAotCustomProfilePath)`.
 
     The default values for port and custom profile are `9999` and
     `custom.aprof`.
 
-    This is equivalent to `aprofutil $(AProfUtilExtraOptions) -s -v -f -p $(AndroidAotProfilerPort) -o "$(AndroidAotCustomProfilePath)"`
-
-    The `aprofutil` call might be extended with
+    The `aprofutil` call may be extended with
     `$(AProfUtilExtraOptions)`, to pass additional options.
+
+    This is equivalent to:
+    
+        aprofutil $(AProfUtilExtraOptions) -s -v -f -p $(AndroidAotProfilerPort) -o "$(AndroidAotCustomProfilePath)"
 
     Added in Xamarin.Android v10.3.
 
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -119,6 +119,27 @@ The following build targets are defined for Xamarin.Android projects:
     `Resource.designer.cs` file. This target is usually called by the
     IDE when new resources are added to the project.
 
+-   **BuildAndStartAotProfiling** &ndash; Builds the package with
+    embedded aot profiler, sets aot profiler socket port to
+    $(AndroidAotProfilerPort) and starts the *launch* activity.
+
+    Added in Xamarin.Android v10.3.
+
+-   **FinishAotProfiling** &ndash; Collects the aot profiler data from
+    the device or the emulator through sockets port
+    $(AndroidAotProfilerPort) and writes them to
+    $(AndroidAotCustomProfilePath).
+
+    The default values for port and custom profile are `9999` and
+    `custom.aprof`.
+
+    This is equivalent to `aprofutil $(AProfUtilExtraOptions) -s -v -f -p $(AndroidAotProfilerPort) -o "$(AndroidAotCustomProfilePath)"`
+
+    The `aprofutil` call might be extended with
+    `$(AProfUtilExtraOptions)`, to pass additional options.
+
+    Added in Xamarin.Android v10.3.
+
 <a name="Build_Extension_Points" />
 
 ## Build Extension Points
--- a/Documentation/release-notes/3964.md
+++ b/Documentation/release-notes/3964.md
@@ -1,0 +1,18 @@
+### Add new AOT profiling targets
+
+The new `BuildAndStartAotProfiling` target builds the package with
+embedded aot profiler, sets aot profiler socket port to
+$(AndroidAotProfilerPort) and starts the *launch* activity.
+
+The new `FinishAotProfiling target` collects the aot profiler data
+from the device or the emulator through sockets port
+$(AndroidAotProfilerPort) and writes them to
+$(AndroidAotCustomProfilePath).
+
+The default values for port and custom profile are `9999` and
+`custom.aprof`.
+
+This is equivalent to `aprofutil $(AProfUtilExtraOptions) -s -v -f -p $(AndroidAotProfilerPort) -o "$(AndroidAotCustomProfilePath)"`
+
+The `aprofutil` call might be extended with
+`$(AProfUtilExtraOptions)`, to pass additional options.
--- a/Documentation/release-notes/3964.md
+++ b/Documentation/release-notes/3964.md
@@ -1,18 +1,21 @@
 ### Add new AOT profiling targets
 
 The new `BuildAndStartAotProfiling` target builds the package with
-embedded aot profiler, sets aot profiler socket port to
-$(AndroidAotProfilerPort) and starts the *launch* activity.
+embedded AOT profiler, sets AOT profiler socket port to
+`$(AndroidAotProfilerPort)` and starts the *launch* activity.
 
-The new `FinishAotProfiling target` collects the aot profiler data
+The new `FinishAotProfiling` target collects the AOT profiler data
 from the device or the emulator through sockets port
-$(AndroidAotProfilerPort) and writes them to
-$(AndroidAotCustomProfilePath).
+`$(AndroidAotProfilerPort)` and writes them to
+`$(AndroidAotCustomProfilePath)`.
 
 The default values for port and custom profile are `9999` and
 `custom.aprof`.
 
-This is equivalent to `aprofutil $(AProfUtilExtraOptions) -s -v -f -p $(AndroidAotProfilerPort) -o "$(AndroidAotCustomProfilePath)"`
-
-The `aprofutil` call might be extended with
+The `aprofutil` call may be extended with
 `$(AProfUtilExtraOptions)`, to pass additional options.
+
+This is equivalent to:
+
+    aprofutil $(AProfUtilExtraOptions) -s -v -f -p $(AndroidAotProfilerPort) -o "$(AndroidAotCustomProfilePath)"
+
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -66,6 +66,23 @@ namespace Xamarin.Android.Build.Tests
 			}
 		}
 
+		[Test, Category ("UsesDevice")]
+		public void BuildBasicApplicationAndAotProfileIt ()
+		{
+			if (!HasDevices)
+				Assert.Ignore ("Skipping test. No devices available.");
+
+			var proj = new XamarinAndroidApplicationProject ();
+			var projDirectory = Path.Combine ("temp", TestName);
+			using (var b = CreateApkBuilder (projDirectory)) {
+				Assert.IsTrue (b.RunTarget (proj, "BuildAndStartAotProfiling"), "Run of BuildAndStartAotProfiling should have succeeded.");
+				System.Threading.Thread.Sleep (5000);
+				Assert.IsTrue (b.RunTarget (proj, "FinishAotProfiling"), "Run of FinishAotProfiling should have succeeded.");
+				var customProfile = Path.Combine (Root, projDirectory, "custom.aprof");
+				FileAssert.Exists (customProfile);
+			}
+		}
+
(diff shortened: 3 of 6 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Application.targets
```diff
@@ -35,4 +35,29 @@ Copyright (C) 2019 Microsoft Corporation. All rights reserved.
...
+  <Target Name="_SetAotProfilingPropsOnDevice">
+    <Exec Command="&quot;$(AdbToolPath)adb&quot; $(AdbTarget) shell setprop debug.mono.profile aot:port=$(AndroidAotProfilerPort)" />
+  </Target>
+  <Target Name="BuildAndStartAotProfiling"
+      DependsOnTargets="$(_BeginAotProfilingDependsOnTargets)">
+  </Target>
+  <Target Name="FinishAotProfiling"
+      DependsOnTargets="_ResolveSdks">
+    <Exec Command="&quot;$(MonoAndroidBinDirectory)aprofutil&quot; -s -v -f -p $(AndroidAotProfilerPort) -o &quot;$(AndroidAotCustomProfile)&quot;" />
+  </Target>
```
jonathanpeppers: Does the new target need a `$(AProfUtilExtraArgs)` property? Are there some flags people might want to add in the future like `--verbose`, etc.?

Documentation/guides/BuildProcess.md
```diff
@@ -119,6 +119,24 @@ The following build targets are defined for Xamarin.Android projects:
     `Resource.designer.cs` file. This target is usually called by the
     IDE when new resources are added to the project.
 
+-   **BuildAndStartAotProfiling** &ndash; Builds the package with
+    embedded aot profiler, sets aot profiler socket port to
+    $(AndroidAotProfilerPort) and starts the *launch* activity.
+
+    Added in Xamarin.Android v10.4.
```
jonpryor: This is actually going into v10.3...

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -66,6 +66,24 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
...
+		[Test, Category ("UsesDevice")]
+		public void BuildBasicApplicationAndAotProfileIt ()
+		{
+			if (!HasDevices)
+				Assert.Ignore ("Skipping test. No devices available.");
+
+			var proj = new XamarinAndroidApplicationProject ();
+			var projDirectory = Path.Combine ("temp", TestName);
+			using (var b = CreateApkBuilder (projDirectory)) {
+				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
```
jonpryor: Based on the documentation, `b.Build()` shouldn't be necessary, as `BuildAndStartAotProfiling` will implicitly perform a Build anyway.  Should this line be removed?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Application.targets
```diff
@@ -35,4 +35,29 @@ Copyright (C) 2019 Microsoft Corporation. All rights reserved.
     </GetAndroidPackageName>
     <Exec Command="&quot;$(AdbToolPath)adb&quot; $(AdbTarget) shell am force-stop &quot;$(_AndroidPackage)&quot;" />
   </Target>
+  <PropertyGroup>
+    <AndroidAotCustomProfile Condition=" '$(AndroidAotCustomProfile)' == '' ">custom.aprof</AndroidAotCustomProfile>
```
jonpryor: This property should also be separately documented, in the Properties section of `BuildProcess.md`.

`$(AndroidAotCustomProfilePath)` (appending `Path`) might make it clearer that this needs to contain a filename.  Or perhaps `AndroidAotCustomProfileOutputPath`?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Application.targets
```diff
@@ -35,4 +35,29 @@ Copyright (C) 2019 Microsoft Corporation. All rights reserved.
...
+    <AndroidAotProfilerPort Condition=" '$(AndroidAotProfilerPort)' == '' ">9999</AndroidAotProfilerPort>
```
jonpryor: This likewise should be documented in the Properties section of `BuildProcess.md`.

jonpryor:
This PR should also add a `Documentation/release-notes/3964.md` file which contains the release notes for this new functionality.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Application.targets
radekdoulik: Good idea, I named it `$(AProfUtilExtraOptions)` to be similar to `$(AndroidExtraAotOptions)`.

Documentation/guides/BuildProcess.md
radekdoulik: Fixed.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Application.targets
radekdoulik: It is really specific to the aot profiling targets, so I am not sure it makes sense to add it to properties sections. It is same case as $(AndroidLaunchActivity), which is also only documented in the build targets section.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Application.targets
radekdoulik: Changed to `AndroidAotCustomProfilePath`

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -66,6 +66,23 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
...
+		[Test, Category ("UsesDevice")]
+		public void BuildBasicApplicationAndAotProfileIt ()
+		{
+			if (!HasDevices)
+				Assert.Ignore ("Skipping test. No devices available.");
+
+			var proj = new XamarinAndroidApplicationProject ();
+			var projDirectory = Path.Combine ("temp", TestName);
+			using (var b = CreateApkBuilder (projDirectory)) {
+				Assert.IsTrue (b.RunTarget (proj, "BuildAndStartAotProfiling"), "Run of BuildAndStartAotProfiling should have succeeded.");
```
jonpryor: Looks like something in this test method [is failing](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3276974&view=ms.vss-test-web.build-test-results-tab&runId=9973166&resultId=100000&paneView=attachments):

```
/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.Debugging.targets(612,2): error : Parameter name: source [/Users/runner/runners/2.160.1/work/1/s/bin/TestRelease/temp/BuildBasicApplicationAndAotProfileIt/UnnamedProject.csproj]
/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.Debugging.targets(612,2): error :   at System.Linq.Enumerable.Select[TSource,TResult] (System.Collections.Generic.IEnumerable`1[T] source, System.Func`2[T,TResult] selector) [0x0000d] in <549f46d1789f4a51aa5fa1e7d715fe0d>:0  [/Users/runner/runners/2.160.1/work/1/s/bin/TestRelease/temp/BuildBasicApplicationAndAotProfileIt/UnnamedProject.csproj]
/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.Debugging.targets(612,2): error :   at Xamarin.Android.Tasks.InstallPackageAssemblies.RunTask () [0x003d0] in <26f94f6d10aa4e62aabb7af2f165e6d5>:0  [/Users/runner/runners/2.160.1/work/1/s/bin/TestRelease/temp/BuildBasicApplicationAndAotProfileIt/UnnamedProject.csproj]
```

jonpryor:
Note that I've added some extra commits to your branch.  Please don't lose them. :-)

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
radekdoulik: Yay, so it runs on CI :-)

I wasn't able to reproduce this one locally, but noticed another problem introduced by removing this line from the test:

    Assert.IsTrue (b.Build (proj), "Build should have succeeded.");

And failing locally with:

    error MSB3073: The command ""/Users/rodo/android-toolchain/sdk/platform-tools/adb"  shell am start -n "UnnamedProject.UnnamedProject/"" exited with code 255.

I was able to overcome that with adding `Build` target to `BuildAndStartAotProfiling` dependencies. I expected the `Install` target to be enough to trigger the `Build` target as well, but looks like it wasn't the case and as a result we didn't have launcher activity set in the manifest? @dellis1972 @jonathanpeppers is that expected behavior?

Thanks to that I also noticed we need to have `AndroidEmbedProfilers` in cached properties to trigger rebuild.

radekdoulik:
> Note that I've added some extra commits to your branch. Please don't lose them. :-)

Also note that I have updated the commit message (PR top level comment), so please use that instead of the list of commit messages when doing `Squash and merge`.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
dellis1972: The `Install` target does NOT always run a `Build`, it depends on certain conditions like if you are running in an IDE. All of our using tests pass `BuildingInsideVisualStudio=true` so in this case `Install` will NOT call `Build`.

Documentation/release-notes/3964.md
```diff
@@ -0,0 +1,21 @@
...
+The new `FinishAotProfiling` target collects the AOT profiler data
+from the device or the emulator through sockets port
+`$(AndroidAotProfilerPort)` and writes them to
+`$(AndroidAotCustomProfilePath)`.
+
+The default values for port and custom profile are `9999` and
+`custom.aprof`.
+
+The `aprofutil` call may be extended with
+`$(AProfUtilExtraOptions)`, to pass additional options.
```
dellis1972: should we be consistent about our property naming. We have `AndroidAotProfilerPort` and `AndroidAotCustomProfilePath` which match all of our other (new) properties. But we have `AProfUtilExtraOptions`.
Should it be `AndroidAProfUtilExtraOptions` or `AndroidProfUtilExtraOptions` ?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Application.targets
```diff
@@ -35,4 +35,30 @@ Copyright (C) 2019 Microsoft Corporation. All rights reserved.
     </GetAndroidPackageName>
     <Exec Command="&quot;$(AdbToolPath)adb&quot; $(AdbTarget) shell am force-stop &quot;$(_AndroidPackage)&quot;" />
   </Target>
+  <PropertyGroup>
+    <AndroidAotCustomProfilePath Condition=" '$(AndroidAotCustomProfilePath)' == '' ">custom.aprof</AndroidAotCustomProfilePath>
+    <AndroidAotProfilerPort Condition=" '$(AndroidAotProfilerPort)' == '' ">9999</AndroidAotProfilerPort>
+    <_BeginAotProfilingDependsOnTargets>
+      _SetupAotProfiling;
+      Build;
```
dellis1972: @jonathanpeppers should this be conditional based on if we are in the IDE or not like we do for the install target?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
dellis1972: The `Install` target only calls build if we are running from the command line. In the IDE is does NOT call `Build`. Our tests run as if they are in the IDE, so a call to Build will be needed.

Documentation/release-notes/3964.md
radekdoulik: `AndroidAProfUtilExtraOptions` seemed too long and as it was already suggested and thumbed up by @jonpryor without the Android prefix, I went ahead with that :-)

Note that the `A` in `AProf` is for AOT.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Application.targets
jonathanpeppers: I think what is here seems OK, because it is a specific target not in the devloop.

If it's the condition here: https://github.com/xamarin/xamarin-android/blob/fd64aaff618f5f818831bdef844876e138fec7a0/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L3097-L3110

This was just a performance thing, we shouldn't need to worry about this custom target, then?

jonpryor:
Execution of the new `BuildBasicApplicationAndAotProfileIt` test is failing: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3278268&view=ms.vss-test-web.build-test-results-tab&runId=9979214&resultId=100000&paneView=attachments

```
  EXEC : error : aotprofile-tool: Unable to read profile through local port: 9999. [/Users/runner/runners/2.160.1/work/1/s/bin/TestRelease/temp/BuildBasicApplicationAndAotProfileIt/UnnamedProject.csproj]
  /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Application.targets(62,5): error MSB3075: The command ""/Library/Frameworks/Xamarin.Android.framework/Libraries/xbuild/Xamarin/Android/Darwin/aprofutil"  -s -v -f -p 9999 -o "custom.aprof"" exited with code 5. Please verify that you have sufficient rights to run this command. [/Users/runner/runners/2.160.1/work/1/s/bin/TestRelease/temp/BuildBasicApplicationAndAotProfileIt/UnnamedProject.csproj]

```

PR #3967, opened 2019-11-26
[Xamarin.Android.Build.Tasks] remove usage of <CreateItem/>

jonathanpeppers:
Context: https://docs.microsoft.com/visualstudio/msbuild/createitem-task

    CreateItem task
    This task is deprecated. Starting with .NET Framework 3.5, item groups may be placed within Target elements.

The `<CreateItem/>` MSBuild task is deprecated, this removes all usage
of it in xamarin-android's targets we ship. I tried to make changes to
simplify `Condition`, and hopefully no actual behavior will be changed
here.

There may also be some light performance impact, in a build with no
changes in `samples\HelloWorld`:

    20 ms  CreateItem                                39 calls
    Time Elapsed 00:00:01.16

Now there are 0 calls. I suspect some of this time is merely moved
elsewhere, however...

MSBuild likely has some performance improvements to using
`<ItemGroup/>` over `<CreateItem/>`. I was also able to simplify some
general logic and repeated `Condition` attributes that won't be
evaluated so many times now.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets
@@ -221,14 +221,10 @@ Copyright (C) 2012 Xamarin Inc. All rights reserved.
        Condition="'$(TargetFrameworkMoniker)' != ''"
     />
     </CreateProperty>
-    <CreateItem Include="$(TargetFrameworkMonikerAssemblyAttributesPath)">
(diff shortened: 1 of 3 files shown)


dellis1972:
So both `CreateItem` and `ItemGroup` execute when a target is skipped, so that is one concern I had with this. That `CreateItem` did not run and this would change behaviour, but it seems it does so we should be good.

jonathanpeppers:
The failure is uploading a dup artifact on re-run, looks green 👍

Issue #3912, opened 2019-11-13
Warn about app domains removal in linker


radekdoulik:
We plan to add a warning for our users in the linker. When app domains are used, we should warn that it will be removed in newer API.

radekdoulik:
Implemented by https://github.com/xamarin/xamarin-android/pull/3974

Issue #3845, opened 2019-10-25
Check for to-be-removed APIs


jonpryor:
We know that in the .NET 5 timeframe, certain APIs such as `System.AppDomain.CreateDomain()` will no longer work.

We should updat the `<LinkAssemblies/>` task/etc. to check for these to-be-altered APIs and start warning about them, so that the community has time to migrate away from them before .NET 5 ships.

radekdoulik:
Implemented by https://github.com/xamarin/xamarin-android/pull/3974

PR #3974, opened 2019-11-29
Warn about AppDomain.CreateDomain usage

radekdoulik:
Context:
https://github.com/xamarin/xamarin-android/issues/3912
https://github.com/xamarin/xamarin-android/issues/3845

Issue warning when non-product/non-SDK assembly uses
`System.AppDomain::CreateDomain` as app domains will be removed in
.NET5

The check is added to `FixAbstractMethodsStep`, so this way it is done
in both configurations, `Debug` and `Release`.

Add tests for both configurations.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
@@ -22,9 +22,13 @@ namespace MonoDroid.Tuner
 			if (!Annotations.HasAction (assembly))
 				Annotations.SetAction (assembly, AssemblyAction.Skip);
 
-			bool changed = FixAbstractMethods (assembly);
 
-			if (changed) {
+			if (IsProductOrSdkAssembly (assembly))
+				return;
+
+			CheckAppDomainUsageUnconditional (assembly, (string msg) => Context.Logger.LogMessage (MessageImportance.High, msg));
+
+			if (FixAbstractMethodsUnconditional (assembly)) {
 				Context.SafeReadSymbols (assembly);
 				AssemblyAction action = Annotations.HasAction (assembly) ? Annotations.GetAction (assembly) : AssemblyAction.Skip;
 				if (action == AssemblyAction.Skip || action == AssemblyAction.Copy || action == AssemblyAction.Delete)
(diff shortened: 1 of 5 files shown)


dellis1972:
> Just a thought, would it possible to just put the `[Obsolete]` attribute on the type/methods in the BCL in Mono and that's it?
> 
> Then we wouldn't have any extra Mono.Cecil code running at build time.

I like this idea :D

dellis1972:
Also what is the impact of the build of this?

radekdoulik:
> Just a thought, would it possible to just put the `[Obsolete]` attribute on the type/methods in the BCL in Mono and that's it?
> 
> Then we wouldn't have any extra Mono.Cecil code running at build time.

That would not warn about code in assemblies which are not compiled in the project.

radekdoulik:
I tried to measure the build times for XA template.

LinkAssembliesNoShrink call

| Build | Before (ms) | After (ms) |
|----------------|-----|-----|
| The initial build | 747 | 877 | 
| Touch `MainActivity.cs` and rebuild | 141 | 144 |

src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
```diff
@@ -53,10 +53,12 @@ public override bool RunTask ()
 				for (int i = 0; i < SourceFiles.Length; i++) {
 					var source = SourceFiles [i];
 					var destination = DestinationFiles [i];
+					var assemblyDefinition = resolver.GetAssembly (source.ItemSpec);
```
jonathanpeppers: It is possible to check `Profile.IsSdkAssembly (source.ItemSpec) || Profile.IsProductAssembly (source.ItemSpec)` and not load the assembly at all here? That might help perf slightly.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -22,9 +22,13 @@ protected override void ProcessAssembly (AssemblyDefinition assembly)
 			if (!Annotations.HasAction (assembly))
 				Annotations.SetAction (assembly, AssemblyAction.Skip);
 
-			bool changed = FixAbstractMethods (assembly);
 
-			if (changed) {
+			if (IsProductOrSdkAssembly (assembly))
+				return;
+
+			CheckAppDomainUsageUnconditional (assembly, (string msg) => Context.Logger.LogMessage (MessageImportance.High, msg));
```
jonathanpeppers: This needs a warning code, and then we usually write a small doc to put in: https://github.com/xamarin/xamarin-android/tree/master/Documentation/guides/messages

jonathanpeppers: Whoops the warning is actually this line: https://github.com/xamarin/xamarin-android/pull/3974/files#diff-fa265e57634387510f2adaed81eb2fd3R58

src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
```diff
@@ -53,10 +53,12 @@ public override bool RunTask ()
...
+
+					step.CheckAppDomainUsage (assemblyDefinition, (string msg) => Log.LogWarning (msg));
```
jonpryor: As @jonathanpeppers mentioned earlier, this should be a `LogCodedWarning()`, and please:

1. Create a new warning code
2. Add docs for the new code in https://github.com/xamarin/xamarin-android/tree/master/Documentation/guides/messages

Perhaps `XA2000 member semantics will change in a future release` would work?

src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
```diff
@@ -53,10 +55,18 @@ public override bool RunTask ()
...
 					var destination = DestinationFiles [i];
+					AssemblyDefinition assemblyDefinition = null;
+
+					if (!MTProfile.IsSdkAssembly (source.ItemSpec) && !MTProfile.IsProductAssembly (source.ItemSpec)) {
+						assemblyDefinition = resolver.GetAssembly (source.ItemSpec);
+						step.CheckAppDomainUsage (assemblyDefinition, (string msg) => Log.LogCodedWarning ("XA2000", msg));
+					}
 
 					// Only run the step on "MonoAndroid" assemblies
 					if (MonoAndroidHelper.IsMonoAndroidAssembly (source) && !MonoAndroidHelper.IsSharedRuntimeAssembly (source.ItemSpec)) {
```
jonathanpeppers: At some point I should figure out what the difference is between `Profile.IsSdkAssembly`, `Profile.IsProductAssembly`, and `MonoAndroidHelper.IsSharedRuntimeAssembly`. Maybe one is duplicate?

That is a different PR, though.

PR #3977, opened 2019-12-02
[build] clean directories ourselves on AzDO

jonathanpeppers:
Context: https://build.azdo.io/3281436

Expanding upon e608464d, let's disable the clean behavior provided by
Azure DevOps and clean ourselves. We have an opportunity to kill
processes this way.

Otherwise, we seem to be hitting this on every build now:

    ##[error]The process cannot access the file 'F:\A\xamarin-v000001-1\_work\2\s' because it is being used by another process.

I am also interested in what `Get-Process` will print here.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -159,13 +159,13 @@ stages:
     pool: $(VSEngWinVS2019)
     timeoutInMinutes: 360
     cancelTimeoutInMinutes: 5
-    workspace:
-      clean: all
     steps:
     - checkout: self
       submodules: recursive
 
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
> I am also interested in what `Get-Process` will print here.

It was `adb` locking the `F:\A\xamarin-v000001-1\_work\2\s` directory.

mattleibow:
I notice that locally when adb locks a directory. Is there anything we can do about this?

jonathanpeppers:
> I notice that locally when adb locks a directory. Is there anything we can do about this?

If you just run `adb shell echo test` and there is no `adb` running at all, the newly spawned daemon process locks the current directory.

I have no idea why `adb` does this? If you do some Google searching, it seems people complain about this for Unity & Cordova, too.

PR #3981, opened 2019-12-03
Bump to mono:2019-12

directhex:

PR #3987, opened 2019-12-03
[ci] Run Windows build tree tests with powershell

pjcollins:
We've encountered numerous timeouts in our Windows build job when
executing the `RunNUnitTests` msbuild target. This timeout can result in
various file locking issues, as well as poor partial test result
reporting. To help address these two problems, we'll change the order of
test execution so that our short running test suites execute and publish
test results first. We'll also use the `run-nunit-tests` YAML template
to run the xamarin-android test suites. This template will run each test
assembly in separate powershell invocations as opposed to a batched
MSBuild target, which may result in fewer hangs. Finally, an unused
(and undefined) parameter has been removed from the `run-nunit-tests`
template.

### DIFF
--- a/build-tools/automation/yaml-templates/run-nunit-tests.yaml
+++ b/build-tools/automation/yaml-templates/run-nunit-tests.yaml
@@ -10,9 +10,9 @@ parameters:
 steps:
 - powershell: |
     if ([Environment]::OSVersion.Platform -eq "Unix") {
-        mono ${{ parameters.nunitConsole }} ${{ parameters.testAssembly }} ${{ parameters.whereClause }} --result ${{ parameters.testResultsFile }} ${{ parameters.nunitConsoleExtraArgs }}
+        mono ${{ parameters.nunitConsole }} ${{ parameters.testAssembly }} --result ${{ parameters.testResultsFile }} ${{ parameters.nunitConsoleExtraArgs }}
     } else {
-        ${{ parameters.nunitConsole }} ${{ parameters.testAssembly }} ${{ parameters.whereClause }} --result ${{ parameters.testResultsFile }}  ${{ parameters.nunitConsoleExtraArgs }}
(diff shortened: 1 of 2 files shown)


BrzVlad:
@pjcollins Does this fix the timeout from the mono integration PR ? (https://github.com/xamarin/xamarin-android/pull/3981)

build-tools/automation/yaml-templates/run-nunit-tests.yaml
```diff
@@ -10,9 +10,9 @@ parameters:
 steps:
 - powershell: |
     if ([Environment]::OSVersion.Platform -eq "Unix") {
-        mono ${{ parameters.nunitConsole }} ${{ parameters.testAssembly }} ${{ parameters.whereClause }} --result ${{ parameters.testResultsFile }} ${{ parameters.nunitConsoleExtraArgs }}
+        mono ${{ parameters.nunitConsole }} ${{ parameters.testAssembly }} --result ${{ parameters.testResultsFile }} ${{ parameters.nunitConsoleExtraArgs }}
     } else {
-        ${{ parameters.nunitConsole }} ${{ parameters.testAssembly }} ${{ parameters.whereClause }} --result ${{ parameters.testResultsFile }}  ${{ parameters.nunitConsoleExtraArgs }}
+        ${{ parameters.nunitConsole }} ${{ parameters.testAssembly }} --result ${{ parameters.testResultsFile }}  ${{ parameters.nunitConsoleExtraArgs }}
     }
```
jonathanpeppers: Would there be any benefit to using the VSTest task instead?

https://docs.microsoft.com/azure/devops/pipelines/tasks/test/vstest

pjcollins:
@BrzVlad it will hopefully help with that type of timeout.

@jonathanpeppers does VSTest work on macOS? Maybe we can look to change this in a separate PR for even better GitHub reporting?

PR #3988, opened 2019-12-04
[docs] Add release-notes README

brendanzagaeski:
Remove the old `master.md` release notes file in favor of per-PR files
to follow the new style discussed in [a7f5cb26][0].

Add a new `README.md` file that describes the format and process for the
per-PR files.

[0]: https://github.com/xamarin/xamarin-android/commit/a7f5cb26f7493028345534b97a8ac8c80daae8a7


brendanzagaeski:
I was thinking maybe now that I'm removing the `master.md` I should add a `README.md` in the `Documentation/release-notes` directory to explain how the directory is used.

Does a README in that directory sound like a good way to help surface that info?

jonpryor:
> Does a README in that directory sound like a good way to help surface that info?

Yes.

jonpryor:
Related point: how do we *cleanup* the `Documentation/release-notes` directory?

Under current guidelines (a7f5cb26f7493028345534b97a8ac8c80daae8a7), every "feature" or "significant" commit will add a new file into that directory.  What is the mechanism for *removing* those files?  Delete when a new release branch is created?

Documentation/release-notes/3855.md
```diff
@@ -0,0 +1,12 @@
+### Java exception wrapping
```
jonpryor: ...for example, PR #3855 is part of d16-5.  As such, does this file need to exist in master, which will become d16-6?  I don't think so...?

jonpryor:
The fact that some things are cherry-picked into release branches *after* merging to master certainly doesn't help with explaining things...

brendanzagaeski:
The initial thought I had for the cleanup strategy was to delete any remaining outdated files as part of the work on each release publishing day.

That approach is compatible with clearing the directory when branching, but it would mean we wouldn't have to worry if by chance on some occasion the directory didn't get cleared before branching. There would be another chance to clean up items at publishing time.

One feature of this strategy is that we don't have to worry about deleting notes for individual cherry-picked items. And if we occasionally have notes (or supplemental notes) in separate PRs from the source code commits, potentially after branching, we could optionally skip cherry-picking those to the appropriate branch.

One non-feature is that it misses out on the opportunity to have the `Documentation/release-notes` directory in each branch contain only items that were *first* published in that branch.

brendanzagaeski:

Remove the changes to the `xamarin-android/README.md`. I have put those in a separate PR instead (<https://github.com/xamarin/xamarin-android/pull/4065>) so this PR can be used to iterate a bit on the new `Documentation/release-notes/README.md` file.

brendanzagaeski:
**Changes in latest force-pushes:**

  * Add a *Templates* section with examples of release notes for various kinds of changes.
  * Remove the *Removing old notes* section for now, and add a placeholder *Publication workflow* section header in its place.

In the interest of making the release drafting process available now, I'm thinking it might be OK to fill out the *Publication workflow* section in a separate PR? I think there might be a few details in that section that will take a little extra time to discuss and settle on, so it could be handy to have the other sections already available in the mean time.

Documentation/release-notes/README.md
```diff
@@ -0,0 +1,177 @@
...
+### Other issue fix or small improvement
+
+  * Use past tense verbs when describing the previous problematic behavior.
+  * List all public GitHub or Developer Community items that were fixed by the
+    change.  If no public items were fixed, or if the scope of the fix was
+    larger than any of the individual items, list the PR instead.
+  * Where possible and appropriate, include the error text that users would have
+    seen before the fix.
+  * If the fix is for a regression and you know when the problem started,
+    include that version info.
```
brendanzagaeski: Another possible point to include:

```
  * For items not based on end-user reports, still aim to include information
    about how end-user projects could have in theory encountered the previous
    behavior.
```

Documentation/release-notes/README.md
```diff
@@ -0,0 +1,177 @@
...
+updated from 3.27.1 to [3.28.0](https://sqlite.org/releaselog/3_28_0.html),
+bringing in several improvements and bug fixes.
+```
+
+### Other issue fix or small improvement
+
+  * Use past tense verbs when describing the previous problematic behavior.
+  * List all public GitHub or Developer Community items that were fixed by the
+    change.  If no public items were fixed, or if the scope of the fix was
+    larger than any of the individual items, list the PR instead.
```
brendanzagaeski: On second thought, this probably needs to be something like:

```diff
-larger than any of the individual items, list the PR instead.
+larger than any of the individual items, skip adding any items, or if you
+already have a PR number, list that instead.
```

There won't be a PR number yet in cases where the draft release note is added before the PR is even started.

Documentation/release-notes/README.md
```diff
@@ -0,0 +1,177 @@
...
+    Xamarin.Android users, so Xamarin.Android 10.1 now reports them as
+    informational messages rather than warnings.
+```
+
+### Submodule or `.external` bump
+
+  * Use a single file for all of the notes for the bump.
+  * Use section headings to help organize the notes.
+  * For items that do not have a public issue or PR on GitHub or Developer
+    Community, use the xamarin-android bump PR as the item link instead.
```
brendanzagaeski: Like in the *Other issue fix or small improvement* section,

```suggestion
    Community, you can skip adding any links to the item.
```

There won't be a PR number yet in cases where the draft release notes are committed to the author's branch before a PR is even opened.

brendanzagaeski:
**Changes in latest commit:**

Add a note about using the most specific template that fits the PR. Adjust the order of the templates slightly so the external tool template (more specific) comes before the new feature template (less specific).

Add a template section about changes in shipping code that should not (yet) be mentioned in the release notes.

Add information about what to put in the heading for each of the templates.

Correct the information about linking to PRs to clarify that including a link to a PR is optional and only relevant if the PR already exists.

Correct "comparsion" and "mutiple" typos.

PR #3995, opened 2019-12-05
[build] Sanitize Microsoft.Build projects

pjcollins:
Context: https://github.com/xamarin/xamarin-android/pull/3928
Context: https://github.com/xamarin/monodroid/pull/1048

Attempts to clean up and modernize various projects to help break up the
scope of change required to convert them to .NET Standard libraries. 

Updates the shared `MSBuildReferences.projitems` file to instead include
NuGet references to the required Microsoft.Build.* assemblies. The
Mono.Posix.NETStandard, Xamarin.Build.AsyncTask, and Xamarin.LibZipSharp
NuGet references have also been moved into this file to make it easier
to update versions in one place as needed. We're only using the compile
time assets from the MSBuild NuGet packages, which ensures that their
content is not copied to the output directory. An additional `ItemGroup`
has been added to this file to copy the required assemblies from the system
to the output directory so that `Xamarin.Android.Build.Tests` can run against
them. These changes are required because the MSBuild assemblies coming
from NuGet do not yet fully support running on Mono.

All projects which reference `MSBuildReferences.projitems` have been
converted from the old `packages.config` to the new `PackageReference`
NuGet reference style.

The Azure Pipeline build job has been updated to completely build
`Xamarin.Android-Tests.sln` with xabuild, rather than only building select
projects. This was done to mirror our Jenkins build and to ensure that
`make all-tests` continues to work moving forward.

.NET Core and `NuGetCommand` tasks have been bumped to bring
in a newer version of NuGet with better PackageReference support.

NuGet restore will be skipped if any issues occur when calculating the
dependency graph for the solution or project that it is operating on.
A lot of our sources attempt to import props/targets/projitems files
that are generated by the build, and these should be conditionally
imported so that NuGet restore attempts can succeed before these files
are generated. This should also help when loading these projects in
VS or VS Mac.

Example NuGet restore error:

    src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets(8,3): error MSB4019: The imported project "bin/BuildRelease/ProfileAssemblies.projitems" was not found. Confirm that the expression in the Import declaration "../../bin/BuildRelease/ProfileAssemblies.projitems" is correct, and that the file exists on disk.
    /var/folders/pp/x88nqyvd0h18ttl996vv2c3w0000gn/T/NuGetScratch/ok90j1ok.3t3.nugetrestore.targets(266,5): warning : Skipping restore for project 'src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj'. The project file may be invalid or missing targets required for restore.

I've also moved the `NuGet.exe` that is copied and used by tests into it's own
folder to avoid a crash on macOS which would occur when `NuGet.exe` is
invoked from the same folder as `Xunit.NetCore.Extensions.dll`:

    mono bin/TestRelease/NuGet.exe

    VTable setup of type Xunit.NetCore.Extensions.SkippedTestCase failed
    System.TypeLoadException: VTable setup of type Xunit.NetCore.Extensions.SkippedTestCase failed
      at (wrapper managed-to-native) System.RuntimeType.GetPropertiesByName_native(System.RuntimeType,intptr,System.Reflection.BindingFlags,System.RuntimeType/MemberListType)
      at System.RuntimeType.GetPropertiesByName (System.String name, System.Reflection.BindingFlags bindingAttr, System.RuntimeType+MemberListType listType, System.RuntimeType reflectedType) [0x0001b] in <7d0bc631cadd4a07bf74717847e90cfa>:0
      at System.RuntimeType.GetPropertyCandidates (System.String name, System.Reflection.BindingFlags bindingAttr, System.Type[] types, System.Boolean allowPrefixLookup) [0x00010] in <7d0bc631cadd4a07bf74717847e90cfa>:0
      at System.RuntimeType.GetProperties (System.Reflection.BindingFlags bindingAttr) [0x00000] in <7d0bc631cadd4a07bf74717847e90cfa>:0
      at System.ComponentModel.Composition.AttributedModel.AttributedPartCreationInfo+<GetExportMembers>d__32.MoveNext () [0x000eb] in <5e20d4aa8c6c44a89bf1aeb3c06f166c>:0
      at System.Linq.Enumerable.Any[TSource] (System.Collections.Generic.IEnumerable1[T] source) [0x00015] in <e22351e395f44d7b8c4860fca2dcee71>:0
      ...


pjcollins:
There's a macOS specific issue in Xamarin.Android.Build.Tests that I'm looking at now.

build-tools/Xamarin.Android.Tools.BootstrapTasks/packages.config
```diff
@@ -1,7 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<packages>
-  <package id="Microsoft.DotNet.ApiCompat" version="5.0.0-beta.19606.1" targetFramework="net472" />
```
jonpryor: This removal is an odd one; see also:

  * https://github.com/xamarin/xamarin-android/pull/3884#discussion_r354643291
  * https://github.com/xamarin/xamarin-android/pull/3997

It's the `<CheckApiCompatibility/>` in `build-tools/Xamarin.Android.Tools.BootstrapTasks` task which actually uses the `Microsoft.DotNet.ApiCompat` package outputs, so it makes sense *to me* for `build-tools/Xamarin.Android.Tools.BootstrapTasks` to be responsible for restoring this package.

The alternative is that *every project* which uses `<CheckApiCompatibility/>` task would also need to have this `@(PackageReference)`.  (Which certainly could be done -- you do so here! -- but it just "feels weird" to me.)

I would *prefer* that `Microsoft.DotNet.ApiCompat` be added as a `@(PackageReference)` in `Xamarin.Android.Tools.BootstrapTasks.csproj`.

pjcollins: I've moved this reference back and added a `globalPackagesFolder` path to our `NuGet.config` so that we can continue to use local paths for the tooling.

pjcollins: The scope of change needed to introduce a new `globalPackagesFolder` is a bit larger than anticipated, we should do that in a separate PR. For this PR I've updated the path to the ApiCompat tooling in `Mono.Android.targets` to use the default user based folder.

PR #3999, opened 2019-12-06
[Xamarin.Android.Build.Tasks] Allow BuildApk to run incrementally.

dellis1972:
The `BuildApk` task currently does not work incrementally.
When it is called it throws away the previously built apk
in favour of using the `base` on that `aapt/aapt2` just built.
With incremental builds it is highly likely that only a
few files have actually changed so we are wasting a ton of time.

During my tests on a simple app `BuildApk` generally takes
about 4 seconds to run. This is the same for both clean
and incremental builds.
With these changes, this time could now go down to 600 ms
for an incremental build. Which is a decent improvement.

The new system does not throw away all the work done in
the previous build. Instead it uses the previous output
apk as a base on which to produce the new one.
The output from `aapt/aapt2` is instead `merged` into the
current apk. We do this by checking the `CRC` values of
each entry and only merging in files which have actually
changed. We need to use the `CRC` here because `aapt/aapt2`
is NEVER setting the `ModifiedDate` on the zip items :(.
It seems to always be 01/01/1980.

Once those changes are out of the way , we can now move on
to our custom content. This time we can use the `ModifiedDate`
of both the `ZipEntry` AND the source file. This allows us
to, again, only update the items which have accutally changed.

One final part is we need to keep track of what files are in the
`apk` in the first place and which ones get added to it. This is
what the `existingEntries` List is used for. This allows us to
detect when files are REMOVED from the system. Otherwise we end up
with stale files in the apk from previous builds.

On my tests when changing a single AndroidResource file I got
the following results

	3664 ms BuildApk                  1 calls (Before)
	669 ms BuildApk                  1 calls (After)

Which is a decent improvement.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -13,7 +13,7 @@ using Microsoft.Build.Framework;
 
 using Java.Interop.Tools.Cecil;
 
-using ArchiveFileList = System.Collections.Generic.List<System.Tuple<string, string>>;
+using ArchiveFileList = System.Collections.Generic.List<(string filePath, string archivePath)>;
 using Mono.Cecil;
 using Xamarin.Android.Tools;
 using Xamarin.Tools.Zip;
@@ -109,22 +109,61 @@ namespace Xamarin.Android.Tasks
 
 		protected virtual void FixupArchive (ZipArchiveEx zip) { }
 
+		List<string> existingEntries = new List<string> ();
+
 		void ExecuteWithAbi (string [] supportedAbis, string apkInputPath, string apkOutputPath)
 		{
-			var temp = apkOutputPath + "new";
 			ArchiveFileList files = new ArchiveFileList ();
-			if (apkInputPath != null)
-				File.Copy (apkInputPath, temp, overwrite: true);
+			bool refresh = true;
+			if (apkInputPath != null && File.Exists (apkInputPath) && !File.Exists (apkOutputPath)) {
+				Log.LogDebugMessage ($"Copying {apkInputPath} to {apkInputPath}");
+				File.Copy (apkInputPath, apkOutputPath, overwrite: true);
+				refresh = false;
+			}
 			using (var notice = Assembly.GetExecutingAssembly ().GetManifestResourceStream ("NOTICE.txt"))
-			using (var apk = new ZipArchiveEx (temp, apkInputPath != null ? FileMode.Open : FileMode.Create )) {
+			using (var apk = new ZipArchiveEx (apkOutputPath, File.Exists (apkOutputPath) ? FileMode.Open : FileMode.Create )) {
+				if (refresh) {
+					for (long i = 0; i < apk.Archive.EntryCount; i++) {
+						ZipEntry e = apk.Archive.ReadEntry ((ulong) i);
+						Log.LogDebugMessage ($"Registering item {e.FullName}");
+						existingEntries.Add (e.FullName);
+					}
+				}
+				if (apkInputPath != null && File.Exists (apkInputPath) && refresh) {
+					var lastWriteOutput = File.Exists (apkOutputPath) ? File.GetLastWriteTimeUtc (apkOutputPath) : DateTime.MinValue;
+					var lastWriteInput = File.GetLastWriteTimeUtc (apkInputPath);
+					using (var packaged = new ZipArchiveEx (apkInputPath, FileMode.Open)) {
+						foreach (var entry in packaged.Archive) {
+							Log.LogDebugMessage ($"Deregistering item {entry.FullName}");
+							existingEntries.Remove (entry.FullName);
+							if (lastWriteInput <= lastWriteOutput)
+								continue;
+							if (apk.Archive.ContainsEntry (entry.FullName)) {
+								ZipEntry e = apk.Archive.ReadEntry (entry.FullName);
+								// check the CRC values as the ModifiedDate is always 01/01/1980 in the aapt generated file.
+								if (entry.CRC == e.CRC) {
+									Log.LogDebugMessage ($"Skipping {entry.FullName} from {apkInputPath} as its up to date.");
+									continue;
+								}
+							}
+							var ms = new MemoryStream ();
+							entry.Extract (ms);
+							Log.LogDebugMessage ($"Refreshing {entry.FullName} from {apkInputPath}");
+							apk.Archive.AddStream (ms, entry.FullName, compressionMethod: entry.CompressionMethod);
+						}
+					}
+				}
 				apk.FixupWindowsPathSeparators ((a, b) => Log.LogDebugMessage ($"Fixing up malformed entry `{a}` -> `{b}`"));
-				apk.Archive.AddEntry (RootPath + "NOTICE", notice);
+				string noticeName = RootPath + "NOTICE";
+				existingEntries.Remove (noticeName);
+				if (!apk.Archive.ContainsEntry (noticeName))
+					apk.Archive.AddEntry (noticeName, notice);
 
 				// Add classes.dx
 				foreach (var dex in DalvikClasses) {
 					string apkName = dex.GetMetadata ("ApkName");
 					string dexPath = string.IsNullOrWhiteSpace (apkName) ? Path.GetFileName (dex.ItemSpec) : apkName;
-					apk.Archive.AddFile (dex.ItemSpec, DalvikPath + dexPath);
+					AddFileToArchiveIfNewer (apk, dex.ItemSpec, DalvikPath + dexPath);
 				}
 
 				if (EmbedAssemblies && !BundleAssemblies)
@@ -133,25 +172,25 @@ namespace Xamarin.Android.Tasks
 				AddRuntimeLibraries (apk, supportedAbis);
 				apk.Flush();
 				AddNativeLibraries (files, supportedAbis);
-				apk.Flush();
 				AddAdditionalNativeLibraries (files, supportedAbis);
-				apk.Flush();
 
 				if (TypeMappings != null) {
 					foreach (ITaskItem typemap in TypeMappings) {
-						apk.Archive.AddFile (typemap.ItemSpec, RootPath + Path.GetFileName(typemap.ItemSpec), compressionMethod: UncompressedMethod);
+						AddFileToArchiveIfNewer (apk, typemap.ItemSpec, RootPath + Path.GetFileName(typemap.ItemSpec), compressionMethod: UncompressedMethod);
 					}
 				}
 
 				int count = 0;
 				foreach (var file in files) {
-					var item = Path.Combine (file.Item2, Path.GetFileName (file.Item1))
+					var item = Path.Combine (file.archivePath, Path.GetFileName (file.filePath))
 						.Replace (Path.DirectorySeparatorChar, '/');
-					if (apk.Archive.ContainsEntry (item)) {
-						Log.LogCodedWarning ("XA4301", file.Item1, 0, "Apk already contains the item {0}; ignoring.", item);
+					existingEntries.Remove (item);
+					if (apk.SkipExistingFile (file.filePath, item)) {
+						Log.LogDebugMessage ($"Skipping {file.filePath} as the archive file is up to date.");
 						continue;
 					}
-					apk.Archive.AddFile (file.Item1, item, compressionMethod: GetCompressionMethod (file.Item1));
+					Log.LogDebugMessage ("\tAdding {0}", file.filePath);
+					apk.Archive.AddFile (file.filePath, item, compressionMethod: GetCompressionMethod (file.filePath));
 					count++;
 					if (count == ZipArchiveEx.ZipFlushLimit) {
 						apk.Flush();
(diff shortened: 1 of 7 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -133,25 +169,28 @@ void ExecuteWithAbi (string [] supportedAbis, string apkInputPath, string apkOut
...
-					var item = Path.Combine (file.Item2, Path.GetFileName (file.Item1))
+					var item = Path.Combine (file.archivePath, Path.GetFileName (file.filePath))
 						.Replace (Path.DirectorySeparatorChar, '/');
+					existingEntries.Remove (file.filePath);
 					if (apk.Archive.ContainsEntry (item)) {
-						Log.LogCodedWarning ("XA4301", file.Item1, 0, "Apk already contains the item {0}; ignoring.", item);
-						continue;
+						DateTime lastWrite = File.GetLastWriteTimeUtc (file.filePath);
+						if (lastWrite <= apk.Archive.ReadEntry (item).ModificationTime || !refresh) {
+							Log.LogCodedWarning ("XA4301", file.filePath, 0, "Apk already contains the item {0}; ignoring.", item);
```
dellis1972: I need to figure out what to do with this Warning now.. its causing a few tests to break.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -133,25 +169,28 @@ void ExecuteWithAbi (string [] supportedAbis, string apkInputPath, string apkOut
 				AddRuntimeLibraries (apk, supportedAbis);
 				apk.Flush();
 				AddNativeLibraries (files, supportedAbis);
-				apk.Flush();
 				AddAdditionalNativeLibraries (files, supportedAbis);
-				apk.Flush();
```
jonathanpeppers: I think these two `Flush()` calls were removed on accident?

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -380,6 +453,14 @@ CompressionMethod GetCompressionMethod (string fileName)
...
 		{
 			string archivePath = $"lib/{abi}/{inArchiveFileName}";
+			existingEntries.Remove (archivePath);
+			if (apk.Archive.ContainsEntry (archivePath)) {
+				DateTime lastWrite = File.GetLastWriteTimeUtc (filesystemPath);
+				if (lastWrite <= apk.Archive.ReadEntry (archivePath).ModificationTime) {
+					Log.LogDebugMessage ($"Skipping {filesystemPath} (APK path: {archivePath}) as it is up to date.");
+					return;
+				}
+			}
```
jonathanpeppers: There are a few places code exactly like this. I think we should put it on `ZipArchiveEx` so it isn't duplicated?

We could also add some unit tests for `ZipArchiveEx` to be sure it works as expected.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -109,22 +109,58 @@ public class BuildApk : AndroidTask
...
+							foreach (var entry in packaged.Archive) {
+								existingEntries.Remove (entry.FullName);
+								if (apk.Archive.ContainsEntry (entry.FullName)) {
+									ZipEntry e = apk.Archive.ReadEntry (entry.FullName);
+									// check the CRC values as the ModifiedDate is always 01/01/1980 in the aapt generated file.
+									if (entry.CRC == e.CRC) {
+										Log.LogDebugMessage ($"Skipping {entry.FullName} from {apkInputPath} as its up to date.");
+										continue;
+									}
+								}
```
jonathanpeppers: I think we need a test where we test Add/Remove/Update of `AndroidResource` files to make sure the APK is updated in the right way.

jonathanpeppers:
@dellis1972 if you happen to stop by, I'm getting something weird on Windows:

image

In this unit test, the zip entry has a `ModificationTime` older than the file on disk?

Could the code producing these timestamps have a problem?

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
dellis1972: They are not needed as `AddAdditionalNativeLibraries` and the call above do not write to the apk. That is done further down when we look through the `files` list.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
dellis1972: Good idea.

dellis1972:
> @dellis1972 if you happen to stop by, I'm getting something weird on Windows:
> 
> image
> 
> In this unit test, the zip entry has a `ModificationTime` older than the file on disk?
> 
> Could the code producing these timestamps have a problem?

The timestamps in the zip are from the File afaik, and libzip handles all of that.

jonathanpeppers:
> In this unit test, the zip entry has a ModificationTime older than the file on disk?

The issue was that the zip file doesn't include milliseconds, but on Windows you get millisecond precision from file timestamps.

dellis1972:
Getting weird test results with the `SkipExistingFile_Exist` and `SkipExistingFile_TimeStamp` tests.

The `A.txt` file is written to disk before the zip is created... 

BUT when we examine the date times later.. the file on disk has a NEWER time???? By almost a second. Nothing touches this file between it being created, added to the zip and then being read to get the TimeStamp.

```
14/01/2020 15:49:59 (disk 01/14/2020 15:49:59:688) <= 14/01/2020 15:49:58 (zip 01/14/2020 15:49:58:000) 
```

dellis1972:
We have a weird problem where the Modified Date of the entry in a zip is somehow OLDER than the source file it is created from. Its consistently 1 second older, but its not 100% of the time. It seems like some sort of rounding error 🤷‍♂

It turns out this is a limitation of the Zip file format. If we read the Structure document [1] we come across this sentence.

The FAT filesystem of DOS has a timestamp resolution of only two seconds; ZIP file records mimic this. As a result, the built-in timestamp resolution of files in a ZIP archive is only two seconds, though extra fields can be used to store more precise timestamps.
So a modified date in a zip file can be one of TWO seconds. We can confirm this issue by looking at the time encoding functions used within lib zip itself. The Encoding code makes used of bit shifting to encode the unix timestamp into the required Zip format. The Decoding code reverses this operation. So lets run through a few examples.

So given a time of 12:46 and 29 seconds if we run through the code we get a value of 26062 to store in the zip file local header. If we use 12:46 and 30 seconds we get 26063. So far so go. Now let's plug these encoded values into our decoding code (x << 1) & 62.

(26062 << 1) & 62 = 28
(26063 << 1) & 62 = 30

As you can see while the 30 second timestamp decoded correctly, the 29 second one did not. So when we convert these values back to DateTime values we can end up with a timestamp that is 1 second in the past as it always rounds down. This is not great when we want to try to compare timestamps with files created on disk. Say a file was created at 29 seconds and we create the zip at the same time (within a second), according to the zip it would have been created at 28 seconds. So if we use the timestamps to compare it will thing the file is NEWER. Because this only happens if the zip and the file are created within the same 1 second timespan, this is not always a problem.

The work around in this case is to detect when we get these collisions and bump UP to the next second when we are creating the zip file.

But to be honest, I'm not sure there will ever be a good workaround because of the format of the zip file.

[1] https://en.wikipedia.org/wiki/Zip_(file_format)#Structure

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ZipArchiveExTests.cs
```diff
@@ -1,48 +1,75 @@
...
+		ConcurrentDictionary<string, string> paths = new ConcurrentDictionary<string, string> ();
+		string root;
 
 		[SetUp]
 		public void SetUp ()
 		{
-			string root = Path.GetDirectoryName (GetType ().Assembly.Location);
-			temp = Path.Combine (root, "temp", TestContext.CurrentContext.Test.Name);
+			root = Path.GetDirectoryName (GetType ().Assembly.Location);
+			string temp = Path.Combine (root, "temp", TestContext.CurrentContext.Test.Name);
```
dellis1972: Tests were running in Parallel, and because some used the same `root` folder they tripped up over each other. So I had to rework this a bit so it would work in a parallel environment.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -199,14 +241,13 @@ void ExecuteWithAbi (string [] supportedAbis, string apkInputPath, string apkOut
 						count = 0;
 					}
 				}
+				// Clean up Removed files. 
+				foreach (var entry in existingEntries) {
+					Log.LogDebugMessage ($"Removing {entry} as it is not longer required.");
+					apk.Archive.DeleteEntry (entry);
```
jonathanpeppers: If we removed files, does this need a `Flush()`?

I vaguely remember having problems here when I didn't `Flush()`: https://github.com/xamarin/xamarin-android/blob/0153c2ceeaaa80b0dacb2f5f840ee1bf5f541f32/src/Xamarin.Android.Build.Tasks/Tasks/BuildBaseAppBundle.cs#L33-L43

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
```diff
@@ -1084,20 +1088,38 @@ public class Foo {
...
 					string text = File.ReadAllText (Path.Combine (Root, path, appProj.ProjectName, "Resources", "Resource.designer.cs"));
 					Assert.IsTrue (text.Contains ("theme_devicedefault_background2"), "Resource.designer.cs was not updated.");
 					Assert.IsFalse (appBuilder.Output.IsTargetSkipped ("_UpdateAndroidResgen"), "_UpdateAndroidResgen target should NOT be skipped.");
+					Assert.IsFalse (appBuilder.Output.IsTargetSkipped ("_CreateBaseApk"), "_CreateBaseApk target should NOT be skipped.");
+					Assert.IsFalse (appBuilder.Output.IsTargetSkipped ("_BuildApkEmbed"), "_BuildApkEmbed target should NOT be skipped.");
+					byte[] rawContentBuildTwo = ZipHelper.ReadFileFromZip (Path.Combine (Root, appBuilder.ProjectDirectory, appProj.OutputPath, appProj.PackageName + "-Signed.apk"),
+						"res/raw/test.txt");
+					Assert.IsNotNull (rawContentBuildTwo, "res/raw/test.txt should have been in the apk ");
+					txt = Encoding.UTF8.GetString (rawContentBuildTwo ?? new byte[0]);
+					StringAssert.Contains ("Test Build 2 Now", txt, $"res/raw/test.txt should have been 'Test Build 2' not {txt}");
```
jonathanpeppers: Do we need a test here checking when a file is removed? To make sure it is gone from the apk.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ZipArchiveExTests.cs
```diff
@@ -102,5 +142,50 @@ public void AddDirectoryCurrentDirectory ()
 				Directory.SetCurrentDirectory (cwd);
 			}
 		}
+
+		[Test]
+		//[Retry (50)]
```
jonathanpeppers: Comment could be removed here?

Otherwise this looks good to me, except the builds are hitting some issue on DevOps: `fatal: reference is not a tree: 2d6332c056206000a4751081c8cd7f462d9d9715`

dellis1972: I'll remove the comment, and hopefully the next build will be good.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/XamarinProject.cs
```diff
@@ -272,7 +272,7 @@ public virtual void UpdateProjectFiles (string directory, IEnumerable<ProjectRes
 					continue;
 				}
 				string filedir = directory;
-				if (path.Contains (Path.DirectorySeparatorChar)) {
+				if (path.Contains ($"{Path.DirectorySeparatorChar}")) {
```
dellis1972: @jonathanpeppers do we prefer this or `.ToString` ?

jonathanpeppers: I think both *look* fine, so I looked at the IL produced:
```msil
// if (text.Contains($"{Path.DirectorySeparatorChar}"))
IL_00bb: ldloc.2
IL_00bc: ldstr "{0}"
IL_00c1: ldsfld char [netstandard]System.IO.Path::DirectorySeparatorChar
IL_00c6: box [netstandard]System.Char
IL_00cb: call string [netstandard]System.String::Format(string, object)
IL_00d0: callvirt instance bool [netstandard]System.String::Contains(string)
```
So I think it only matters if perf is a concern, `ToString()` would be better and avoid a `string.Format("{0}", Path.DirectorySeparatorChar)`.

This PR is green and it's a test, so I wouldn't change it.

dellis1972: Its good to know that `ToString` is more performant :D

PR #4011, opened 2019-12-09
[xaprepare] Update Android 10 name

brendanzagaeski:
Context: https://www.blog.google/products/android/evolving-android-brand/

Starting with Android 10, Google is no longer adding nicknames to the
Android releases.

### DIFF
--- a/Documentation/release-notes/4011.md
+++ b/Documentation/release-notes/4011.md
@@ -1,0 +1,6 @@
+### Issues fixed
+
+  * [GitHub PR 4011](https://github.com/xamarin/xamarin-android/pull/4011):
+    The old *Q* nickname from the Android 10 preview versions was still being
+    shown for Android 10 in the Visual Studio project property pages and the
+    Android SDK Manager.
--- a/build-tools/xaprepare/xaprepare/ConfigAndData/BuildAndroidPlatforms.cs
+++ b/build-tools/xaprepare/xaprepare/ConfigAndData/BuildAndroidPlatforms.cs
@@ -37,7 +37,7 @@ namespace Xamarin.Android.Prepare
 			new AndroidPlatform (apiName: "Oreo",                   apiLevel: 26, platformID: "26",  include: "v8.0",    framework: "v8.0"),
 			new AndroidPlatform (apiName: "Oreo",                   apiLevel: 27, platformID: "27",  include: "v8.1",    framework: "v8.1"),
 			new AndroidPlatform (apiName: "Pie",                    apiLevel: 28, platformID: "28",  include: "v9.0",    framework: "v9.0"),
-			new AndroidPlatform (apiName: "Q",                      apiLevel: 29, platformID: "29",  include: "v10.0",   framework: "v10.0"),
+			new AndroidPlatform (apiName: "Android 10",             apiLevel: 29, platformID: "29",  include: "v10.0",   framework: "v10.0"),
 		};
 
 		public static readonly Dictionary<string, uint> NdkMinimumAPI = new Dictionary<string, uint> {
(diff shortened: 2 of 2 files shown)


brendanzagaeski:
This might or might not be the ideal way to name this Android release. In particular, the current suggestion in this PR results in an entry that appears in the Visual Studio project property pages as:

```
Android 10.0 (Android 10)
```

... and in the Android SDK manager as:

```
Android 10.0 – Android 10
```

brendanzagaeski:
In case it's handy to know, with the current way the logic and string resources are set up, if the `<Name/>` element is left empty or omitted from `AndroidApiInfo.xml` altogether, then the entry in the Visual Studio project property pages looks like:

```
Android 10.0 ()
```

... and in the Android SDK Manager it looks like:

```
Android 10.0 –
```

JonDouglas:
This is a tough one. I don't honestly know what's the best way forward. I believe your proposal makes the most sense although it is a bit duplicative in nature.

I'm wondering if we can add a conditional for >= 10.0 and drop the codename entirely?

It's interesting because Android Studio is still using `Q` everywhere(Project settings & SDK Manager):

image

It'll be interesting to see what is done for Android 11.

jonpryor:
@JonDouglas, @joj: What should we do here then?  Update the VS IDE so that "code names" are dropped for >= API-29/Android 10?  Update the code name to be `Android 10`, resulting in a property page value of `Android 10.0 (Android 10)`?  Something else?

jonpryor:
@JonDouglas, @joj: Do we even need to care about/present the "code name"/"mnemonic" name *at all*?  Or would we be better served to just have Android version numbers in the drop-down, and *no* code names?

JonDouglas:
I think moving forward, we should drop the codenames. That was the primary motivation for Google to drop dessert names in the first place! @joj thoughts?

joj:
I will follow what you guys decide is best. The main point here is if people are still using the names in everyday conversation. In that case, having the name will make it faster to understand what version we're talking about here. It is true that having 3 different pieces of data that represent the same thing doesn't make a lot of sense.

JonDouglas:
> First, we’re changing the way we name our releases. Our engineering team has always used internal code names for each version, based off of tasty treats, or desserts, in alphabetical order. This naming tradition has become a fun part of the release each year externally, too. But we’ve heard feedback over the years that the names weren’t always understood by everyone in the global community. 

> For example, L and R are not distinguishable when spoken in some languages. So when some people heard us say Android Lollipop out loud, it wasn’t intuitively clear that it referred to the version after KitKat. It’s even harder for new Android users, who are unfamiliar with the naming convention, to understand if their phone is running the latest version. We also know that pies are not a dessert in some places, and that marshmallows, while delicious, are not a popular treat in many parts of the world. 

> As a global operating system, it’s important that these names are clear and relatable for everyone in the world. So, this **next release of Android will simply use the version number and be called Android 10**. We think this change helps make release names simpler and more intuitive for our global community. And while there were many tempting “Q” desserts out there, we think that at version 10 and 2.5 billion active devices, it was time to make this change. 

https://www.blog.google/products/android/evolving-android-brand/

Let's move forward with removing internal code names.

brendanzagaeski:
Just to be extra careful, one clarifying question might be whether we want to stop showing the additional names only for Android 10 and higher or for all the versions from Android 9 Pie and lower too.

I don't have a strong opinion myself.

JonDouglas:
I was personally for >= Android 10. Prior to that, some dessert names are more meaningful than the version number for most developers (i.e. KitKat / Lollipop / Marshmallow / Oreo). I think it makes sense from Android 10 onwards?

PR #4012, opened 2019-12-10
[Xamarin.Android.Build.Tools] Bump Android tools versions

brendanzagaeski:
Context: https://github.com/xamarin/xamarin-android/commit/c5231075e11176db03c338ec370bf18b3e70a8b2
Context: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1032105

As described in [c5231075][0], we should be using the latest versions of
the Android tooling if possible, matching the major version with the max
API level we support.  Xamarin.Android now supports Android 10 (API
level 29), so we should be using the 29.x.x versions of the tooling.

Update `$(AndroidSdkBuildToolsVersion)` to 29.0.2.
Update `$(AndroidSdkPlatformToolsVersion)` to 29.0.5.

[0]: https://github.com/xamarin/xamarin-android/commit/c5231075e11176db03c338ec370bf18b3e70a8b2

### DIFF
--- a/Configuration.props
+++ b/Configuration.props
@@ -97,9 +97,9 @@
     <AllSupported32BitTargetAndroidAbis>armeabi-v7a;x86</AllSupported32BitTargetAndroidAbis>
     <AllSupported64BitTargetAndroidAbis>arm64-v8a;x86_64</AllSupported64BitTargetAndroidAbis>
     <AllSupportedTargetAndroidAbis>$(AllSupported32BitTargetAndroidAbis);$(AllSupported64BitTargetAndroidAbis)</AllSupportedTargetAndroidAbis>
-    <XABuildToolsVersion>29</XABuildToolsVersion>
-    <XABuildToolsFolder Condition="'$(XABuildToolsFolder)' == ''">29.0.0</XABuildToolsFolder>
-    <XAPlatformToolsVersion>29.0.1</XAPlatformToolsVersion>
+    <XABuildToolsVersion>29.0.2</XABuildToolsVersion>
+    <XABuildToolsFolder Condition="'$(XABuildToolsFolder)' == ''">29.0.2</XABuildToolsFolder>
+    <XAPlatformToolsVersion>29.0.5</XAPlatformToolsVersion>
     <XAIncludeProprietaryBits Condition="'$(XAIncludeProprietaryBits)' == ''">False</XAIncludeProprietaryBits>
     <XABundleToolVersion Condition="'$(XABundleToolVersion)' == ''">0.10.2</XABundleToolVersion>
     <XAManifestMergerToolVersion Condition="'$(XAManifestMergerToolVersion)' == ''">26.5.0</XAManifestMergerToolVersion>
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -3308,7 +3308,7 @@ AAMMAAABzYW1wbGUvSGVsbG8uY2xhc3NQSwUGAAAAAAMAAwC9AAAA1gEAAAAA") });
 				Assert.True (builder.Build (proj, parameters: parameters),
 					string.Format ("First Build should have succeeded"));
 				StringAssertEx.Contains ("platforms/android-26", builder.LastBuildOutput, "platforms/android-26 should be a dependency.");
-				StringAssertEx.Contains ("build-tools/28.0.3", builder.LastBuildOutput, "build-tools/28.0.3 should be a dependency.");
+				StringAssertEx.Contains ("build-tools/29.0.2", builder.LastBuildOutput, "build-tools/29.0.2 should be a dependency.");
 				StringAssertEx.Contains ("platform-tools", builder.LastBuildOutput, "platform-tools should be a dependency.");
 			}
 		}
@@ -3341,7 +3341,7 @@ AAMMAAABzYW1wbGUvSGVsbG8uY2xhc3NQSwUGAAAAAAMAAwC9AAAA1gEAAAAA") });
 				Assert.True (builder.Build (proj, parameters: parameters),
 					string.Format ("First Build should have succeeded"));
 				StringAssertEx.Contains ("platforms/android-26", builder.LastBuildOutput, "platforms/android-26 should be a dependency.");
-				StringAssertEx.Contains ("build-tools/28.0.3", builder.LastBuildOutput, "build-tools/28.0.3 should be a dependency.");
+				StringAssertEx.Contains ("build-tools/29.0.2", builder.LastBuildOutput, "build-tools/29.0.2 should be a dependency.");
 				StringAssertEx.Contains ("platform-tools", builder.LastBuildOutput, "platform-tools should be a dependency.");
 			}
 		}
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props.in
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props.in
@@ -14,8 +14,8 @@
 		<AndroidUseAapt2 Condition=" '$(AndroidUseAapt2)' == '' ">True</AndroidUseAapt2>
 		<AndroidPackageNamingPolicy Condition=" '$(AndroidPackageNamingPolicy)' == '' ">LowercaseCrc64</AndroidPackageNamingPolicy>
 		<AndroidUseManagedDesignTimeResourceGenerator Condition=" '$(AndroidUseManagedDesignTimeResourceGenerator)' == '' And '$(OS)' != 'Windows_NT' ">False</AndroidUseManagedDesignTimeResourceGenerator>
-		<AndroidSdkBuildToolsVersion Condition="'$(AndroidSdkBuildToolsVersion)' == ''">28.0.3</AndroidSdkBuildToolsVersion>
-		<AndroidSdkPlatformToolsVersion Condition="'$(AndroidSdkPlatformToolsVersion)' == ''">28.0.0</AndroidSdkPlatformToolsVersion>
+		<AndroidSdkBuildToolsVersion Condition="'$(AndroidSdkBuildToolsVersion)' == ''">29.0.2</AndroidSdkBuildToolsVersion>
+		<AndroidSdkPlatformToolsVersion Condition="'$(AndroidSdkPlatformToolsVersion)' == ''">29.0.5</AndroidSdkPlatformToolsVersion>
 		<AndroidSdkToolsVersion Condition="'$(AndroidSdkToolsVersion)' == ''">26.1.1</AndroidSdkToolsVersion>
 		<AndroidNdkVersion Condition="'$(AndroidNdkVersion)' == ''">16.1</AndroidNdkVersion>
 		<BundleToolVersion Condition="'$(BundleToolVersion)' == ''">@BUNDLETOOL_VERSION@</BundleToolVersion>


brendanzagaeski:
Sounds good. I've added a commit for that. It looks like there are some tests that might also need to be updated to expect these versions. I'll take a look at those now. One consideration I'll keep in mind with those is that the template projects in Visual Studio are for the moment still defaulting to Android 9 Pie (API level 28), so it makes sense to have some tests using those versions.

brendanzagaeski:
The latest commit failed to build because Android SDK Platform-Tools 29.0.5 contains changes to the `api-versions.xml` file compared to Android SDK Platform-Tools 29.0.1.

`generator.exe` uses `api-versions.xml` as an input when generating the C# bindings, so this change causes the API compatibility check in `MonoAndroid.csproj` to fail. Excerpt of the errors:

```
AfterBuild:
  CheckApiCompatibility for ApiLevel: v10.0
  CompatApi command: ../../packages/Microsoft.DotNet.ApiCompat.5.0.0-beta.19606.1/tools/net472/Microsoft.DotNet.ApiCompat.exe "../../tests/api-compatibility\reference\ApiCompatTemp" -i "F:\A\xamarin-v000006-1\_work\2\s\bin\Release\lib\xamarin.android\xbuild-frameworks\MonoAndroid\v10.0\ApiCompatTemp" --exclude-attributes ../../tests/api-compatibility\api-compat-exclude-attributes.txt 
  Compat issues with assembly Mono.Android:
  CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'System.String Android.Manifest.Permission.AccessBackgroundLocation' changed from '[RegisterAttribute("ACCESS_BACKGROUND_LOCATION")]' in the contract to '[RegisterAttribute("ACCESS_BACKGROUND_LOCATION", ApiSince=29)]' in the implementation.
  CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'System.String Android.Manifest.Permission.AccessMediaLocation' changed from '[RegisterAttribute("ACCESS_MEDIA_LOCATION")]' in the contract to '[RegisterAttribute("ACCESS_MEDIA_LOCATION", ApiSince=29)]' in the implementation.
  CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'System.String Android.Manifest.Permission.ActivityRecognition' changed from '[RegisterAttribute("ACTIVITY_RECOGNITION")]' in the contract to '[RegisterAttribute("ACTIVITY_RECOGNITION", ApiSince=29)]' in the implementation.
```

Excerpt of the diff of `api-versions.xml` showing where these changes came from:

```diff
--- "a/platform-tools-29.0.1\\api\\api-versions.xml"
+++ "b/platform-tools-29.0.5\\api\\api-versions.xml"
@@ -10,0 +11 @@
+               <field name="ACCESS_BACKGROUND_LOCATION" since="29"/>
@@ -14,0 +16 @@
+               <field name="ACCESS_MEDIA_LOCATION" since="29"/>
@@ -20,0 +23 @@
+               <field name="ACTIVITY_RECOGNITION" since="29"/>
```

### Questions

I think having these `ApiSince` properties on the `[Register]` attributes might be a desirable change. Is it ok to bring in that change?

If so, I think I can resolve the build breakage by creating a new `MonoAndroid.dll.zip` as describe in <https://github.com/xamarin/xamarin-android/blob/master/tests/api-compatibility/README.md#assembly-reference> and including in this PR.

jonathanpeppers:
@jonpryor or @gugavaro is it fine for @brendanzagaeski to add this attributes here?

https://github.com/xamarin/xamarin-android-api-compatibility/blob/master/reference/Mono.Android.xml

brendanzagaeski:

For the sake of being ready to accept the `ApiSince` API changes if those are approved, I worked through the steps to update the API compatibility files

To address the `ApiSince` changes, update the reference `tests/api-compatibility/reference/Mono.Android.dll.zip` file with a new version built using the `api-versions.xml` from Android SDK Platform-Tools 29.0.5.  This new file was built locally on Windows and then stripped and zipped up following the steps on [the API compatibility README][0].

One subtlety with this change is that previously the `Mono.Android.dll.zip` committed in the repo would return some expected API breaks when comparing it to a new `Mono.Android.dll`.  In contrast, the updated `Mono.Android.dll.zip` now returns *no* API breaks, so `acceptable-breakages-vReference.txt` needs to be empty.  Otherwise, the API compatibility check would fail with an error caused by the mismatch:

    There is an invalid assembly listed on the acceptable breakages file: Compat issues with assembly Mono.Android:

[0]: https://github.com/xamarin/xamarin-android/blob/master/tests/api-compatibility/README.md#assembly-reference

**Questions:**

I'm not sure if I've correctly understood how the `acceptable-breakages-vReference.txt` file works. I might not have created the new `Mono.Android.dll.zip` correctly. Why were there entries in that file for the version of `Mono.Android.dll.zip` that is currently in master?

It looks like the reason might be that the `Mono.Android.dll.zip` currently in master was originally built before https://github.com/xamarin/xamarin-android/commit/1a61fa811e370b68b38a7f318b7d4059dc70779e but then committed after it. So it needed entries in `acceptable-breakages-vReference.txt` to account for the changes to the `[Obsolete]` attributes.

Maybe I've misunderstood when we want to update `Mono.Android.dll.zip` versus just updating `acceptable-breakages-vReference.txt`. Maybe I should have just updated `acceptable-breakages-vReference.txt` in this case. That would definitely cut down on the total repository download size since it would mean fewer changes to the `Mono.Android.dll.zip` binary artifact.

**To do:**

  * I think I need to rebuild the new `Mono.Android.dll.zip` in the Release configuration or just modify `acceptable-breakages-vReference.txt` instead. The build for the latest commit still failed the API check because the new `Mono.Android.dll.zip` had extra `[DebuggerStepThrough]` attributes.
  * I haven't yet addressed the original test breakages from the first commit. I'll take a look at those.
  * The error message I saw for the out-of-date `acceptable-breakages-vReference.txt` was a little confusing to me because the error said "invalid assembly listed," but then the problematic assembly name was "Compat issues with assembly Mono.Android" rather than just an assembly name. Additionally, in the case where this error appears because the new API compatibility check returns *no* errors, the current error message doesn't quite make sense. I'll submit a separate issue or PR about that.

brendanzagaeski:

Revert the previous commit that changed `Mono.Android.dll.zip`, and instead update just `acceptable-breakages-vReference.txt` to account for the `ApiSince` changes to see if that does indeed allow the CI build to succeed like my local build.

brendanzagaeski:

Fix the test failures. It turned out the failing tests were for the automatic SDK download feature. Those tests have hardcoded strings that are compared against the default build-tools version, so those strings needed to be updated.

That resolves the question I had [earlier](https://github.com/xamarin/xamarin-android/pull/4012#issuecomment-564315405) about making sure the updated tests would still cover the Android 9 Pie (API level 29) target framework version. Since the test fixes only touch the *build-tools* version, the coverage of *target framework* versions is unchanged.

## Remaining questions

Do we want to take the `ApiSince` API changes that are currently included in this PR?

If so, is the current approach of updating `acceptable-breakages-vReference.txt` correct, or does `Mono.Android.dll.zip` need to be updated instead?

The answers might need to wait until January to make sure any team members who are away for the end of the year can weigh in.

brendanzagaeski:
@gugavaro, when you have a moment, I think a couple questions you'll be interested in are:

  * Should the missing `ApiSince` info be added in this particular case, and if so, should it be added to the reference `Mono.Android.dll.zip` or the acceptable breakages list?
  * Was this missing `ApiSince` info in the original version of the bindings for API level 29 unavoidable because the `api-versions.xml` in the platform-tools was incomplete? Or does `generator.exe` also have a way to determine the appropriate `ApiSince` without `api-versions.xml` that could help avoid similar issues in the future?

Thanks a bunch!

gugavaro:
> The latest commit failed to build because Android SDK Platform-Tools 29.0.5 contains changes to the `api-versions.xml` file compared to Android SDK Platform-Tools 29.0.1.
> 
> `generator.exe` uses `api-versions.xml` as an input when generating the C# bindings, so this change causes the API compatibility check in `Mono.Android.csproj` to fail. Excerpt of the errors:
> 
> ```
> AfterBuild:
>   CheckApiCompatibility for ApiLevel: v10.0
>   CompatApi command: ../../packages/Microsoft.DotNet.ApiCompat.5.0.0-beta.19606.1/tools/net472/Microsoft.DotNet.ApiCompat.exe "../../tests/api-compatibility\reference\ApiCompatTemp" -i "F:\A\xamarin-v000006-1\_work\2\s\bin\Release\lib\xamarin.android\xbuild-frameworks\MonoAndroid\v10.0\ApiCompatTemp" --exclude-attributes ../../tests/api-compatibility\api-compat-exclude-attributes.txt 
>   Compat issues with assembly Mono.Android:
>   CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'System.String Android.Manifest.Permission.AccessBackgroundLocation' changed from '[RegisterAttribute("ACCESS_BACKGROUND_LOCATION")]' in the contract to '[RegisterAttribute("ACCESS_BACKGROUND_LOCATION", ApiSince=29)]' in the implementation.
>   CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'System.String Android.Manifest.Permission.AccessMediaLocation' changed from '[RegisterAttribute("ACCESS_MEDIA_LOCATION")]' in the contract to '[RegisterAttribute("ACCESS_MEDIA_LOCATION", ApiSince=29)]' in the implementation.
>   CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'System.String Android.Manifest.Permission.ActivityRecognition' changed from '[RegisterAttribute("ACTIVITY_RECOGNITION")]' in the contract to '[RegisterAttribute("ACTIVITY_RECOGNITION", ApiSince=29)]' in the implementation.
> ```
> 
> Excerpt of the diff of `api-versions.xml` showing where these changes came from:
> 
> ```diff
> --- "a/platform-tools-29.0.1\\api\\api-versions.xml"
> +++ "b/platform-tools-29.0.5\\api\\api-versions.xml"
> @@ -10,0 +11 @@
> +               <field name="ACCESS_BACKGROUND_LOCATION" since="29"/>
> @@ -14,0 +16 @@
> +               <field name="ACCESS_MEDIA_LOCATION" since="29"/>
> @@ -20,0 +23 @@
> +               <field name="ACTIVITY_RECOGNITION" since="29"/>
> ```
> 
> ### Questions
> I think having these `ApiSince` properties on the `[Register]` attributes might be a desirable change. Is it ok to bring in that change?
> 
> If so, I think I can resolve the build breakage by creating a new `Mono.Android.dll.zip` as describe in https://github.com/xamarin/xamarin-android/blob/master/tests/api-compatibility/README.md#assembly-reference and including in this PR.

Those are acceptable changes, we just need to update the acceptable files to allow those changes pass by.

gugavaro:
> **Changes in the latest commit:**
> 
> For the sake of being ready to accept the `ApiSince` API changes if those are approved, I worked through the steps to update the API compatibility files
> 
> To address the `ApiSince` changes, update the reference `tests/api-compatibility/reference/Mono.Android.dll.zip` file with a new version built using the `api-versions.xml` from Android SDK Platform-Tools 29.0.5. This new file was built locally on Windows and then stripped and zipped up following the steps on [the API compatibility README](https://github.com/xamarin/xamarin-android/blob/master/tests/api-compatibility/README.md#assembly-reference).
> 
> One subtlety with this change is that previously the `Mono.Android.dll.zip` committed in the repo would return some expected API breaks when comparing it to a new `Mono.Android.dll`. In contrast, the updated `Mono.Android.dll.zip` now returns _no_ API breaks, so `acceptable-breakages-vReference.txt` needs to be empty. Otherwise, the API compatibility check would fail with an error caused by the mismatch:
> 
> ```
> There is an invalid assembly listed on the acceptable breakages file: Compat issues with assembly Mono.Android:
> ```
> 
> **Questions:**
> 
> I'm not sure if I've correctly understood how the `acceptable-breakages-vReference.txt` file works. I might not have created the new `Mono.Android.dll.zip` correctly. Why were there entries in that file for the version of `Mono.Android.dll.zip` that is currently in master?
> 
> It looks like the reason might be that the `Mono.Android.dll.zip` currently in master was originally built before [1a61fa8](https://github.com/xamarin/xamarin-android/commit/1a61fa811e370b68b38a7f318b7d4059dc70779e) but then committed after it. So it needed entries in `acceptable-breakages-vReference.txt` to account for the changes to the `[Obsolete]` attributes.
> 
> Maybe I've misunderstood when we want to update `Mono.Android.dll.zip` versus just updating `acceptable-breakages-vReference.txt`. Maybe I should have just updated `acceptable-breakages-vReference.txt` in this case. That would definitely cut down on the total repository download size since it would mean fewer changes to the `Mono.Android.dll.zip` binary artifact.
> 
> **To do:**
> 
> * I think I need to rebuild the new `Mono.Android.dll.zip` in the Release configuration or just modify `acceptable-breakages-vReference.txt` instead. The build for the latest commit still failed the API check because the new `Mono.Android.dll.zip` had extra `[DebuggerStepThrough]` attributes.
> * I haven't yet addressed the original test breakages from the first commit. I'll take a look at those.
> * The error message I saw for the out-of-date `acceptable-breakages-vReference.txt` was a little confusing to me because the error said "invalid assembly listed," but then the problematic assembly name was "Compat issues with assembly Mono.Android" rather than just an assembly name. Additionally, in the case where this error appears because the new API compatibility check returns _no_ errors, the current error message doesn't quite make sense. I'll submit a separate issue or PR about that.

Where are you getting the Mono.Android.dll from? You should not use the one built locally, instead we use a released one.

If there is no breakage, we can remove the acceptable-breakages-vReference.txt file.

gugavaro:
> @gugavaro, when you have a moment, I think a couple questions you'll be interested in are:
> 
> * Should the missing `ApiSince` info be added in this particular case, and if so, should it be added to the reference `Mono.Android.dll.zip` or the acceptable breakages list?
> * Was this missing `ApiSince` info in the original version of the bindings for API level 29 unavoidable because the `api-versions.xml` in the platform-tools was incomplete? Or does `generator.exe` also have a way to determine the appropriate `ApiSince` without `api-versions.xml` that could help avoid similar issues in the future?
> 
> Thanks a bunch!

@brendanzagaeski those are good questions:
I believe the right thing to do is to add the ApiSince as acceptable break changes like you did. The tool considers it a break since we are changing Attribute.
I don't know how the generator determines the ApiSince. @jpobst can you jump here? Do you know what the generate  uses to determine ApiSince?

thanks!

jpobst:
You can see in a build log what's happening.

First we `class-parse` each `.jar` file in the platform sdks:
```
  Creating directory "c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api".
    c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\Debug\lib\xamarin.android\xbuild\Xamarin\Android\class-parse.exe C:\Users\jopobst\android-toolchain\sdk\platforms\android-10\android.jar -platform=10 -parameter-names="c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\src\Mono.Android\Profiles\api-10.params.txt" -o="c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-10.xml.class-parse"
_ClassParse:
    c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\Debug\lib\xamarin.android\xbuild\Xamarin\Android\class-parse.exe C:\Users\jopobst\android-toolchain\sdk\platforms\android-15\android.jar -platform=15 -parameter-names="c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\src\Mono.Android\Profiles\api-15.params.txt" -o="c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-15.xml.class-parse"
_ClassParse:
    c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\Debug\lib\xamarin.android\xbuild\Xamarin\Android\class-parse.exe C:\Users\jopobst\android-toolchain\sdk\platforms\android-16\android.jar -platform=16 -parameter-names="c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\src\Mono.Android\Profiles\api-16.params.txt" -o="c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-16.xml.class-parse"
_ClassParse:
    c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\Debug\lib\xamarin.android\xbuild\Xamarin\Android\class-parse.exe C:\Users\jopobst\android-toolchain\sdk\platforms\android-17\android.jar -platform=17 -parameter-names="c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\src\Mono.Android\Profiles\api-17.params.txt" -o="c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-17.xml.class-parse"
```

Then we `api-xml-adjuster` each level.  I think this updates the parameter names from values scraped from JavaDoc.
```
_AdjustApiXml:
    c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api-xml-adjuster.exe c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-10.xml.class-parse c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-10.xml.in
_AdjustApiXml:
    c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api-xml-adjuster.exe c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-15.xml.class-parse c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-15.xml.in
_AdjustApiXml:
    c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api-xml-adjuster.exe c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-16.xml.class-parse c:\code\xamarin-android-backport\build-tools\api-xml-adjuster\..\..\bin\BuildDebug\api\api-16.xml.in
```

Finally we `api-merge` them all into a single file, which adds the `ApiSince` attribute based on which file it used:
```
    ..\..\bin\BuildDebug\api-merge.exe -o "obj\Debug\android-29\mcw\api.xml" -s '..\..\bin\BuildDebug\api\api-*.xml.in' ..\..\bin\BuildDebug\api\api-10.xml.in ..\..\bin\BuildDebug\api\api-15.xml.in ..\..\bin\BuildDebug\api\api-16.xml.in ..\..\bin\BuildDebug\api\api-17.xml.in ..\..\bin\BuildDebug\api\api-18.xml.in ..\..\bin\BuildDebug\api\api-19.xml.in ..\..\bin\BuildDebug\api\api-20.xml.in ..\..\bin\BuildDebug\api\api-21.xml.in ..\..\bin\BuildDebug\api\api-22.xml.in ..\..\bin\BuildDebug\api\api-23.xml.in ..\..\bin\BuildDebug\api\api-24.xml.in ..\..\bin\BuildDebug\api\api-25.xml.in ..\..\bin\BuildDebug\api\api-26.xml.in ..\..\bin\BuildDebug\api\api-27.xml.in ..\..\bin\BuildDebug\api\api-28.xml.in ..\..\bin\BuildDebug\api\api-29.xml.in --last-description=..\..\bin\BuildDebug\api\api-29.xml.in
```

jonpryor:
...but that just raises a slightly different question:

Consider `Android.Manifest.Permission.AccessBackgroundLocation`:

```
CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'System.String Android.Manifest.Permission.AccessBackgroundLocation' changed from '[RegisterAttribute("ACCESS_BACKGROUND_LOCATION")]' in the contract to '[RegisterAttribute("ACCESS_BACKGROUND_LOCATION", ApiSince=29)]' in the implementation.
```

[`android.Manifest.permission.ACCESS_BACKGROUND_LOCATION`](https://developer.android.com/reference/android/Manifest.permission#ACCESS_BACKGROUND_LOCATION) was added in *API-29*.

Why, then, are we binding it with *no* `RegisterAttribute.ApiSince` value?  My current build tree has:

```csharp
			[Register ("ACCESS_BACKGROUND_LOCATION")]
			public const string AccessBackgroundLocation = (string) "android.permission.ACCESS_BACKGROUND_LOCATION";
```

`obj/Debug/android-29/mcw/api.xml`, meanwhile, contains:

```xml
<field deprecated="not deprecated"
    final="true"
    name="ACCESS_BACKGROUND_LOCATION"
    jni-signature="Ljava/lang/String;"
    static="true"
    transient="false"
    type="java.lang.String"
    type-generic-aware="java.lang.String"
    value="&quot;android.permission.ACCESS_BACKGROUND_LOCATION&quot;"
    visibility="public"
    volatile="false"
    merge.SourceFile="../../bin/BuildDebug/api/api-29.xml.in"
/>
```

(Newlines added for readability.)

Note that `//field/@merge.SourceFile` *is* set, and thus, `ApiVersionSupport.ApiAvailableSince` *should* be set (...right?), and thus we *should* be including `RegisterAttribute.ApiSince`!

https://github.com/xamarin/java.interop/blob/4565369743a58a1d43e5a79c406c27d46b81bb08/tools/generator/ApiVersionsSupport.cs#L12
https://github.com/xamarin/java.interop/blob/4565369743a58a1d43e5a79c406c27d46b81bb08/tools/generator/Java.Interop.Tools.Generator.CodeGeneration/CodeGenerator.cs#L439

...yet the fact that we're *not* emitting `RegisterAttribute.ApiSince` implies that *something is missing*.  What's missing?  Why aren't we *always* emitting `ApiSince` *on **every** member*?

jpobst:
It looks like we actually pull our `ApiSince` versions from `C:\Users\jopobst\android-toolchain\sdk\platform-tools\api\api-versions.xml` which ships with the Android SDK.  

(`https://android.googlesource.com/platform/development/+/refs/heads/master/sdk/api-versions.xml`).

Therefore I suspect that this file was updated in this Android SDK version.

I have `28.0.2` installed and it does not contain an entry for `ACCESS_BACKGROUND_LOCATION`.

https://github.com/xamarin/java.interop/blob/master/tools/generator/CodeGenerator.cs#L158

jpobst:
Not very hard, though I would probably prefer to change `api-merge` to add an `@api-since` attribute that we read instead.  I'd like any new `generator` change we make to be as generic as possible and not `Mono.Android.dll` specific.

jonpryor:
With the merging of #4186, this PR (hopefully) will no longer trigger API compat breakage.  Let's find out!

brendanzagaeski:
I think I can now I can remove the `acceptable-breakages-vReference.txt` changes from this PR. I'll take a look at that now.

brendanzagaeski:
Ah ha. https://github.com/xamarin/xamarin-android/commit/4cd20605c01d92acfec74d328c251e1729039d9f contains identical changes to `acceptable-breakages-vReference.txt`, so a squashed merge would handle that with no complaints. But I'll pre-squash it on this branch so the commit history is a little easier to review before the final merge.

brendanzagaeski:

Rebased on master and squashed the commits, removing the API compatibility adjustments along the way since those are now already in master.

dellis1972:
@brendanzagaeski how is this PR going? Looks like we might need to get it into d16-5?

brendanzagaeski:
I think this PR is all set to merge to master. The CI test failures are unrelated to this PR specifically. So for the question of merging to master, I think this is ready to go.

**Thoughts on backporting:**

On the one hand, backporting this PR would address some oddness in Visual Studio 2019 version 16.5 Preview where the automatic SDK installation feature currently installs Android SDK Build-Tools version 29.0.2 even though it isn't used. And Build-Tools versions usually have good backwards compatibility, so using version 29.0.2 would in theory be fine for all of the current Android 9 Pie (API level 28) templates and users' existing projects.

On the other hand, as long as users don't manually adjust the installed Build-Tools versions after accepting the auto SDK prompt, the experience is fairly smooth. And there could be some risk in changing the default Build-Tools at this point in the Xamarin.Android 10.2 schedule.

If we do decide to backport, the Visual Studio Installer should probably also be adjusted to install Build-Tools 29.0.2 instead of 28.0.3.

Separate from whatever we decide on this PR, for the *next* anticipated Build-Tools update (30.0), I think it might be worth updating the default `$(AndroidSdkBuildToolsVersion)` version earlier, as part of the general availability release of the corresponding Android 11 (API level 30) bindings, even though the templates will still be on Android 10 (API level 29).

**Background info:**

In Visual Studio 2019 version 16.5 Preview 2 (and also the latest internal preview):

 1. The Visual Studio Installer installs only Android SDK Build-Tools version 28.0.3.

 2. The Android templates target Android 9.0 Pie (API level 28). The switch to Android 10 (API level 29) is planned for the d16-6 branch ([internal work item](https://dev.azure.com/devdiv/DevDiv/_workitems/edit/825086)).

 3. The first time an Android project is loaded (for example, when creating a new project from a template), the automatic SDK installation (**Auto Install Android SDKs**) feature requests permission to install Android SDK Build-Tools version 29.0.2.

 4. If the user cancels the installation at step 3, then the auto SDK installation feature will block building within Visual Studio:

    ```
    Project 'AndroidApp1' requires the following components installed on your machine: 
    Android Emulator, Android SDK Build-Tools 29.0.2
    ```

    (The behavior that aborts the build comes from `AndroidSetup.cs` lines 83-118 in the proprietary Visual Studio Tools for Xamarin source code. It's not part of the MSBuild targets themselves.)

 5. If the user accepts the installation at step 3 or builds on the command line, they will be able to build the project successfully. The build will use the default `$(AndroidSdkBuildToolsVersion)` of 28.0.3 (what this PR is changing).

 6. If the user installs Build-Tools version 29.0.2 and also *uninstalls* Build-Tools version 28.0.3, then the auto SDK installation feature will again block building in the IDE, this time requiring Build-Tools version 28.0.3 because of the default `$(AndroidSdkBuildToolsVersion)` (what this PR is changing).
 
    The workaround is to set `$(AndroidSdkBuildToolsVersion)` in the *.csproj* file:

    ```xml
    <PropertyGroup>
      <AndroidSdkBuildToolsVersion>29.0.2</AndroidSdkBuildToolsVersion>
    </PropertyGroup>
    ```

    (In contrast, for command line builds, even if `$(AndroidSdkBuildToolsVersion)` is left at the default of `28.0.3`, the build will automatically use Build-Tools version 29.0.2 if it can't find the preferred version 28.0.3.)

Overall, the current first-run user experience does work, but it's a little odd. In particular, it's odd that auto SDK asks to install Build-Tools version 29.0.2 when that version won't be used.

I haven't studied the Visual Studio Tools for Xamarin `AndroidToolsBuildService.cs` file enough to understand where the Build-Tools version 29.0.2 dependency is coming from, but I have a hunch it might come from the Android SDK Manager manifest, which now offers Build-Tools version 29.0.2 to satisfy [a request I helped submit](https://developercommunity.visualstudio.com/content/problem/844306/the-android-sdk-manager-does-not-yet-offer-downloa-1.html).

(On a side note, with all of the various tools Xamarin.Android now has in its own installer, it's interesting how little of Android SDK Build-Tools Xamarin.Android uses for projects that are configured to use AAPT2, D8, and the Android App Bundle publishing format. I think it might only use `zipalign` and `aidl`. For APKs, it also uses `apksigner.jar`.)

PR #4016, opened 2019-12-10
[Xamarin.Android.Build.Tasks] fix $(AndroidExplicitCrunch) warning

jonathanpeppers:
This warning message actually prints:

    XA0110: Disabling true as it is not supported by `aapt2`. If you wish to use true please set True to false.

In the long term, I think we should remove `$(AndroidExplicitCrunch)`.
I have found some bugs using it that cause targets to re-run on
incremental builds. It is also not supported when using `aapt2`.

It's worth fixing the warning message for now, though.


jonathanpeppers:
The re-run looks green, just the publish artifact step & a networking BCL test.

PR #4020, opened 2019-12-11
[FixAbstractMethodsStep] Speedup the step

radekdoulik:
By checking whether the assembly has a type reference to
`Java.Lang.Object` first. `ModuleDefinition::HasTypeReference` call is
relatively fast in cecil and by using it we save a lot of further
processing.

Measurements of the `_LinkAssembliesNoShrink` target (XA template,
rebuild after touching MainActivity.cs, 5 runs average):

Before: 160ms
After:   52ms


radekdoulik:
The designer test is failing because `error : There is not enough space on the disk.` Reading the slack discussion, it looks like it is happening in other PR's as well, so I will ignore it here.

The MSBuild - Windows tests was rerun OK, just the usual upload of artifacts is failing.

Thus LGTM.

PR #4022, opened 2019-12-11
[Xamarin.Android.Build.Tasks] add feature flag for Google Manifest Merger

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/3918

In 2c6f5cd1, we added support for the official Google/Android Manifest
Merger. After further QA testing, it seems that a reasonable number of
existing apps require code changes to *appease* the new manifest
merger.

This is a similar situation we were in compared to `aapt2`, where the
suggested changes are good (and usually problems). I think we should
add a feature flag, so we can gradually turn this on. We could add the
feature off by default for one release, and then turn it on by default
for the next.

This PR adds:

* `$(AndroidManifestMerger)` with enum-style values of `legacy` and
`manifestmerger.jar`.
* Updated a few tests to leverage both code paths.


jonpryor:
> Name suggestions welcome

My "concern" with `managed` and `android` is that in *other* contexts, "managed" is the *good* version, e.g. the `ManagedResourceParser`, which we have delusions of taking over everything at some point.

In this context, "managed" would be the *bad* version.

"managed" also precludes us from doing a different implementation from C# in the future which follows the Android Studio semantics.

I thus propose the alternate names:

  * `legacy` for existing Xamarin.Android v10.1 and earlier behavior
  * `manifestmerger.jar ` for the new behavior.

This "nicely" has a clear "old and busted" value, and if (when?) Google comes out with a *different* native tool, we'll be able to reasonably distinguish between them.

brendanzagaeski:
Using `legacy` as the name for the old approach sounds good to me. That would match up roughly with the naming scheme used for values of `$(AndroidTlsProvider)`, where one of the values is likewise `legacy`. And `$(AndroidUseLegacyVersionCode)` also uses the term "legacy" in a similar way.

JonDouglas:
👍 to `legacy`. 👎 to `managed`.

I like the idea that we would in the future use whatever the new tooling name is as the replacement to  `manifestmerger.jar`.

Add a 👍 to the current proposal from me.

jonathanpeppers:
I think the failures here are ignorable, this one has started popping up again randomly:
```
Xamarin.Android.Common.targets(2916,2): error ANDKT0000: keytool error: java.io.EOFException 
```

PR #4030, opened 2019-12-12
[ci] Use d16-5 designer source on d16-5

pjcollins:
Attempts to clone the `designer` source branch that has the same name
as the target branch of the PR being built, or the branch of the commit
being built if we aren't building a PR.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -780,7 +780,7 @@ stages:
       EnableRegressionTest: true
     steps:
     - script: |
-        git clone -q https://$(GitHub.Token)@github.com/xamarin/designer.git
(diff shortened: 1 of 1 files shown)


jonpryor:
I like the idea, but... ;-)

Could/should we introduce a `DesignerBranchName` YAML variable and use that, defaulting to `master` on master?  Then we could document that as part of the branching documentation so that it'll be updated to the correct branch name by @MSylvia.

pjcollins:
@jonpryor I updated to this to dynamically set the source branch to the relevant branch name of Xamarin.Android so we don't need to update things in the future. This should continue to work as long as our release branches stay in sync across repos.

PR for master: https://github.com/xamarin/xamarin-android/pull/4034

pjcollins:
Replaced by https://github.com/xamarin/xamarin-android/pull/4034.

PR #4034, opened 2019-12-12
Dynamically determine designer source branch

pjcollins:
Attempts to clone the `designer` source branch that has the same name
as the target branch of the PR being built, or the branch of the commit
being built if we aren't building a PR.


pjcollins:
I kicked off a build manually for extra testing, I think all corner cases should be addressed now:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3318864

Issue #3993, opened 2019-12-05
You uploaded an APK or Android App Bundle with invalid or missing signing information


cschwarz:
### Steps to Reproduce

1. Update Visual Studio to 16.4
2. Build and package the app as Android App Bundle
3. Upload the app to Google Play Store

### Expected Behavior

Should upload and validate the package successfully.

### Actual Behavior

Upload Fail

You uploaded an APK or Android App Bundle with invalid or missing signing information for some of its files. You need to create a valid signed APK or Android App Bundle.

The same package works fine if build with Visual Studio 16.3.x.

If you need the package for analysis, I can share it privately.

### Version Information

Microsoft Visual Studio Professional 2019
Version 16.4.0
VisualStudio.16.Release/16.4.0+29519.181
Microsoft .NET Framework
Version 4.8.03761

Installed Version: Professional

Visual C++ 2019   00435-60000-00000-AA853
Microsoft Visual C++ 2019

ADL Tools Service Provider   1.0
This package contains services used by Data Lake tools

ASP.NET and Web Tools 2019   16.4.457.38025
ASP.NET and Web Tools 2019

ASP.NET Web Frameworks and Tools 2019   16.4.457.38025
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   16.4.457.38025
Azure App Service Tools v3.0.0

Azure Data Lake Node   1.0
This package contains the Data Lake integration nodes for Server Explorer.

Azure Data Lake Tools for Visual Studio   2.4.1000.0
Microsoft Azure Data Lake Tools for Visual Studio

Azure Functions and Web Jobs Tools   16.4.457.38025
Azure Functions and Web Jobs Tools

Azure Stream Analytics Tools for Visual Studio   2.4.1000.0
Microsoft Azure Stream Analytics Tools for Visual Studio

C# Tools   3.4.0-beta4-19569-03+82f2e2541478e239dc4b04f231e90dc2b3dcb422
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.2.0 (d16-2@8b56e20)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

Fabric.DiagnosticEvents   1.0
Fabric Diagnostic Events

IntelliCode Extension   1.0
[...]

PR #4038, opened 2019-12-13
[Xamarin.Android.Build.Tasks] fix signing for App Bundles with multi-dex

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/3993

After 16.4 shipped, we started getting reports that App Bundle signing
was broken. Google Play was not accepting the file:

    You uploaded an APK or Android App Bundle with invalid or missing signing information for some of its files.
    You need to create a valid signed APK or Android App Bundle.

Both @pjcollins and myself tested this, but it worked for us?

Finally we discovered the issue only occurred when multi-dex was
enabled! I was able to reproduce the issue and do a diff of the file
contents of good and bad `.aab` files.

These two files were present in the broken App Bundle:

    +1969-12-31 18:00:00 .....         1480          716  base\root\META-INF\MSFTSIG.SF
    +1969-12-31 18:00:00 .....         8546         5469  base\root\META-INF\MSFTSIG.RSA

These definitely look signing related!

In c4f032a3, I ported some Java code from Android Studio to C# that
allowed Kotlin libraries to function.

Specifically files such as:

    META-INF/services/kotlinx.coroutines.internal.MainDispatcherFactory
    META-INF/services/kotlinx.coroutines.CoroutineExceptionHandler

Needed to be present in the APK. This inadvertently added files like
`META-INF\MSFTSIG.SF` also!

I checked the latest source from Android Studio:

https://android.googlesource.com/platform/tools/base/+/refs/heads/master/build-system/gradle-core/src/main/groovy/com/android/build/gradle/internal/tasks/ExtractJavaResourcesTask.java#199
https://android.googlesource.com/platform/tools/base/+/refs/heads/master/sdk-common/src/main/java/com/android/ide/common/packaging/PackagingUtils.java#24

I see nothing accounting for these files...

For now I think we should ignore:

* For files under `META-INF`
* Ignore `.SF`, `.MF`, and `.RSA` files

I updated a test to validate these changes.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/PackagingUtilsTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/PackagingUtilsTests.cs
@@ -17,9 +17,18 @@ namespace Xamarin.Android.Build.Tests
 		}
 
 		[Test]
-		public void JarManifest ()
+		public void SigningFiles ()
 		{
 			AssertIsNotValid ("META-INF/MANIFEST.MF");
+			AssertIsNotValid ("META-INF/MANIFEST.SF");
+			AssertIsNotValid ("META-INF/MSFTSIG.SF");
+			AssertIsNotValid ("META-INF/MSFTSIG.RSA");
+			AssertIsNotValid ("META-INF/GOOG.RSA");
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
There is a known issue on the designer build they are looking into:
```
NuGet.targets(123,5): error : There is not enough space on the disk.
```

PR #4045, opened 2019-12-17
Backport manifest merger

jonathanpeppers:
A collection of commits that brings the manifest merger feature to `d16-5`.


jonathanpeppers:
@pjcollins @jonpryor it looks like some performance tests failed here, should I add this one, too?

https://github.com/xamarin/xamarin-android/commit/566633a16c0b6898a04c2af302ba6e07a269643c

I think these perf tests are just getting better times on the other pool? I'm not seeing a perf-improvement PR that was missed.

jonathanpeppers:
Test results look OK now.

The one error is the random issue: 
```
Xamarin.Android.Common.targets(2913,2): error ANDKT0000: keytool error: java.lang.Exception: Key pair not generated, alias <androiddebugkey> already exists
```

The designer test was already failing on [d16-5](https://github.com/xamarin/xamarin-android/runs/351487540), so I think that is unchanged.

PR #4052, opened 2019-12-18
[Xamarin.Android.Build.Tasks] Use ILRepack CLI

pjcollins:
Removes the previous dependency on `ILRepack.MSBuild.Task` and replaces
it with a direct invocation of the latest version of `ILRepack.exe`. I
also spent some time looking at https://github.com/dotnet/ILMerge, but
this tool still only supports running on Windows. Moving to the latest
version of ILRepack includes fixes for .NET Standard 2.0 assemblies.

Repack logic has been moved to run before `$(CopyFilesToOutputDirectory)`
and operate out of `$(IntermediateOutputPath)` to support incremental
building of `Xamarin.Android.Build.Tasks.dll`.


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
```diff
@@ -290,4 +290,45 @@
...
+    <InputAssemblies Include="$(OutputPath)\NuGet.Protocol.dll" />
+    <InputAssemblies Include="$(OutputPath)\NuGet.Versioning.dll" />
+    <InputAssemblies Include="$(OutputPath)\System.Collections.Immutable.dll" />
+    <InputAssemblies Include="$(OutputPath)\System.Reflection.Metadata.dll" />
+  </ItemGroup>
+
+  <Target Name="ILRepacker"
+      AfterTargets="CopyFilesToOutputDirectory"
+      Inputs="$(OutputPath)\$(AssemblyName).dll;@(InputAssemblies)"
+      Outputs="$(OutputPath)\$(AssemblyName).dll;$(IntermediateOutputPath)ILRepacker.stamp"
```
jonathanpeppers: Putting the same `$(OutputPath)\$(AssemblyName).dll` file in `Inputs` and `Outputs` is probably not needed. You should be able to just rely on the stamp file.

pjcollins: Updated

pjcollins:
NuGet restore is failing on `Xamarin.Android-Tests.sln`:

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Xamarin.Android.Sdk.props(39,3): error MSB4062: The "Xamarin.Android.Tasks.SetVsMonoAndroidRegistryKey" task could not be loaded from the assembly C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\Xamarin.Android.Build.Tasks.dll. Could not load file or assembly 'System.Reflection.Metadata, Version=1.4.3.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [F:\A\xamarin-v000001-1\_work\2\s\src\Mono.Android\Test\Mono.Android-Tests.csproj]
...
WARNING: Error reading msbuild project information, ensure that your input solution or project file is valid. NETCore and UAP projects will be skipped, only packages.config files will be restored.
```

I believe this was caused by a test run of #3928. I think we need to consider reverting the Xamarin.Android version that is installed in the instance of VS that we are building on to avoid potentially breaking changes introduced by building against a .vsix from a PR build.

PR #4054, opened 2019-12-18
[JNIEnv] Add support for Kotlin's unsigned types.

jpobst:
Companion to https://github.com/xamarin/java.interop/pull/539.

- There were changes need to `JNIEnv` to support marshaling unsigned types for arrays.
- Add `protobuf-net.dll` to our installers, as it is required to read the metadata Kotlin puts in jar files.
- Add on-device tests for calling Kotlin code that uses unsigned types.
  - We don't currently have a way of acquiring the Kotlin compiler for builds, so the test .jar and the Kotlin stdlib .jar are included.  The source of the test .jar is provided for future use.


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
```diff
@@ -77,6 +77,7 @@
     <PackageReference Include="NuGet.ProjectModel" Version="4.6.0" />
     <PackageReference Include="NuGet.Protocol" Version="4.6.0" />
     <PackageReference Include="NuGet.Versioning" Version="4.6.0" />
+    <PackageReference Include="protobuf-net" Version="2.4.4" />
```
jonathanpeppers: Is this needed in this project? Would some csproj in java.interop get this assembly to the output folder?

jpobst: Nah, it wasn't the fix I was looking for.  It should get the assembly because it is referenced by `Xamarin.Android.Tools.Bytecode`.

jonpryor:
While thinking through the eventual commit message for https://github.com/xamarin/java.interop/commit/71afce55812fc2c5efd6002580b539d8394f7bca , it occurred to me that we *might* also need `JNIEnv.NewArray()` method overloads as well to take `uint[]`/etc. types.

Has a complete on-device end-to-end test been performed with Kotlin unsigned types, in particular array types?

jpobst:
This PR includes on-device tests that roundtrip the various unsigned array types:
https://github.com/xamarin/xamarin-android/pull/4054/files#diff-e5a7b7515df9a54919e687e8c196a0b1R52

Is that what you are looking for?

PR #4078, opened 2020-01-02
[xaprepare] Install NUnit.ConsoleRunner locally

pjcollins:
Our azure pipeline YAML and other sources expect certain tools to be
installed into a local `packages` directory. However, with most of our
NuGet references migrating to `PackageReference`s, we are giving up the
default behavior of installing NuGet packages into this local folder.

To address this, a new step has been added to `xaprepare` to install
`NUnit.ConsoleRunner.3.9.0` into a local `packages` folder.

This step runs as part of the default `Standard` scenario, and will be
skipped when other shorter scenarios are explicitly ran.


pjcollins:
This is currently being tested in https://github.com/xamarin/xamarin-android/pull/4077, which was previously failing due to us no longer having an `nunit3-console.exe` in our local `packages` folder.

akoeplinger:
If you know that package restore ran then you could also use a special msbuild variable that nuget automatically injects into the build which points to the local path for the PackageReference:

```
  <Target Name="RunTests">
    <Exec Command="mono $(PkgNUnit_ConsoleRunner)/tools/nunit3-console.exe MyTest.dll" />
  </Target>
```

pjcollins:
To get the `$(PkgFoo)` property to show up we'll need to set `GeneratePathProperty="true"` on the `PackageReference`. The downside with this approach is that the property is only available when there is a direct or indirect reference to the .csproj that contains the `PackageReference` defining the path property, as it's written to `obj/Foo.csproj.nuget.g.props`. We'd need to add an `NUnit.ConsoleRunner` package ref to every desktop unit test project to go with this approach, but even then we'd only be fixing the MSBuild scenarios. We also invoke `nunit3-console` directly in YAML and in some other scripts in a handful of places.

pjcollins:
It looks like #4077 is still having issues that appear to be due to lower casing changes in the executable path, so we may need to update all `nunit-console` invocations after all.

pjcollins:
As suggested by @jpobst, I've updated this to just use the `install` command which follows the same casing and folder structure that we're expecting in various places. We'll have to bump the version in xaprepare when we update the NUnit.ConsoleRunner package reference though.

pjcollins:
Latest results here and on 4077 look good: https://github.com/xamarin/xamarin-android/pull/4077/checks?check_run_id=372702501

build-tools/xaprepare/xaprepare/Application/Scenario.cs
```diff
@@ -58,6 +58,7 @@ protected virtual void AddStartSteps (Context context)
 		{
 			// These are steps that have to be executed by all the scenarios
 			Steps.Add (new Step_DownloadNuGet ());
+			Steps.Add (new Step_InstallPackagesLocally ());
```
jonpryor: Does this need to be in the *default* `Scenario.AddStartSteps()` implementation?

Wouldn't `Scenario_Required` make more sense, or `Scenario_AndroidToolchain`?

grendello: When building locally for a daily development cycle you don't usually run scenarios other than the default one, I think this is the right place to put this.

pjcollins: This is a good point, I think it may make more sense to move this into the default (`Standard`) scenario, rather than the base class for all scenarios.

jonpryor:
I think that the better solution, instead of effectively duplicating the NUnit.ConsoleRunner version info in two places (`Step_InstallPackagesLocally.cs` and `src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/packages.config`), we should instead take a page from Java.Interop and tell NuGet to *continue* using the local `packages` directory:

https://github.com/xamarin/java.interop/commit/dec2e3902514cafad004203ca783ca87ee38ed49#diff-548a262632fd747c93a4db50b2d785f1

This can be done by adding a `NuGet.config` file which sets the `globalPackagesFolder` config property:

```xml
 <!--
      Our testing infrastructure depends on a known location for:
      - NUnit.Console
      - csc.exe
 -->
 <configuration>  
   <config>
     <add key="globalPackagesFolder" value="$\..\packages" />
   </config>
 </configuration>
```

pjcollins:
If we go the `globalPackagesFolder`  route, we'll still need to update every `nunit3-console.exe` invocation as the path will change to a `PackageReference` style path:

`packages/NUnit.ConsoleRunner.3.9.0/tools/nunit3-console.exe`
versus
`packages/nunit.consolerunner/3.9.0/tools/nunit3-console.exe`

pjcollins:
Replacing this with https://github.com/xamarin/xamarin-android/pull/4105.

Issue #3852, opened 2019-10-28
App builds hang when executing task LinkAssemblies if linking enabled


frankbuckley:
### Steps to Reproduce

1. Build app with AndroidLinkMode=Full or AndroidLinkMode=SdkOnly

### Expected Behavior

App builds successfully.

### Actual Behavior

Build hangs.

It always hangs at the same point - while executing `Task "LinkAssemblies"` after logging:

    Type System.Reflection.Emit.PointerType has no fields to preserve

Example log excerpt: https://gist.github.com/frankbuckley/8b729ccfa89b0d01970557d5e83dd604

I can reproduce with multiple projects building on Windows and on OSX.

An apparently identical report is on Stackoverflow: https://stackoverflow.com/questions/58552286/xamarin-android-build-just-hangs-when-linker-is-enabled

The issue seems to have been introduced since mid-September.

### Version Information

	Microsoft Visual Studio Enterprise 2019
	Version 16.3.6
	VisualStudio.16.Release/16.3.6+29418.71
	Microsoft .NET Framework
	Version 4.8.03752

	Installed Version: Enterprise

	Visual C++ 2019   00435-60000-00000-AA197
	Microsoft Visual C++ 2019

	ADL Tools Service Provider   1.0
	This package contains services used by Data Lake tools

	ASP.NET and Web Tools 2019   16.3.286.43615
	ASP.NET and Web Tools 2019

	ASP.NET Web Frameworks and Tools 2019   16.3.286.43615
	For additional information, visit https://www.asp.net/

	Azure App Service Tools v3.0.0   16.3.286.43615
	Azure App Service Tools v3.0.0

	Azure Data Lake Node   1.0
	This package contains the Data Lake integration nodes for Server Explorer.

	Azure Data Lake Tools for Visual Studio   2.4.2000.0
	Microsoft Azure Data Lake Tools for Visual Studio

	Azure Functions and Web Jobs Tools   16.3.286.43615
	Azure Functions and Web Jobs Tools

	Azure Logic Apps Tools for Visual Studio   1.0
	Add-in for the Azure Resource Group project to support the Logic App Designer and template creation.

	Azure Stream Analytics Tools for Visual Studio   2.4.2000.0
	Microsoft Azure Stream Analytics Tools for Visual Studio

	C# Tools   3.3.1-beta3-19461-02+2fd12c210e22f7d6245805c60340f6a34af6875b
[...]

PR #4079, opened 2020-01-03
[linker] Update ShouldMarkInterfaceImplementation logic

radekdoulik:
We should return `false` when the iface was already marked, to avoid
unnecessary processing, which can cause infinite loop in the linker.

Fixes https://github.com/xamarin/xamarin-android/issues/3852

What was happening: we were returning `true` from
`ShouldMarkInterfaceImplementation` for ifaces of types that
implements `IJavaObject` or `IJavaPeerable`. In case of #3852, 3 of
these had also `System.Runtime.CompilerServices.NullableAttribute`
custom attributes. That led to enqueuing the custom attribute
constructor every pass through the primary queue in the linker and so
the queue was never emptied and that resulted in an infinite loop.

IMO linker should check for iface being already marked outside of
`ShouldMarkInterfaceImplementation`. I will add PR to the linker for
that, meanwhile we should check it here.


radekdoulik:
I have an idea how to test that so I will try to implement a simpler test for that, without requiring the VCL nuget.

DevEddy:
In which Xamarin.Android version will this fix be in?

PR #4085, opened 2020-01-06
[ci] Use multi-repo checkout feature

pjcollins:
Context: https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/multi-repo-checkout?view=azure-devops
Context: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3360513

Some of our build and test jobs started failing recently due to what
appear to be authorization issues when attepting to clone private
sources:

    stderr | remote: Repository not found.

In order to avoid this, we'll try to move away from using custom scripts
and tokens for source checkout and instead use the new "multi-repo"
checkout feature in Azure Pipelines that @akoeplinger suggested.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -27,6 +27,21 @@ resources:
     type: github
     name: xamarin/designer
     endpoint: xamarin
+  - repository: qa
+    type: github
+    name: xamarin/QualityAssurance
+    ref: refs/heads/master
+    endpoint: xamarin
+  - repository: samples
+    type: github
+    name: xamarin/monodroid-samples
+    ref: refs/heads/master
+    endpoint: xamarin
+  - repository: xfsamples
+    type: github
+    name: xamarin/xamarin-forms-samples
+    ref: refs/heads/master
+    endpoint: xamarin
 
 # Global variables
 variables:
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -6,13 +6,26 @@ trigger:
   - master
   - d16-*
 
-# External yaml template files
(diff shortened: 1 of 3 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -71,31 +98,41 @@ stages:
 
     # Prepare and build everything
     - script: make prepare-update-mono V=1 CONFIGURATION=$(XA.Build.Configuration) PREPARE_CI=1 PREPARE_AUTOPROVISION=1
+      workingDirectory: $(System.DefaultWorkingDirectory)/xamarin-android
```
jonpryor: This part confuses me; how/why would `$(System.DefaultWorkingDirectory)/xamarin-android` exist? We have files such as `/Users/vsts/agent/2.155.1/work/1/s/bin/TestRelease/temp/SetupDependenciesForDesigner/UnnamedProject/foo.targets` (see e.g. cc21afa7ad222db77bcb4ce9f7484be40809aa08), i.e. `/Users/vsts/agent/2.155.1/work/1/s` is the clone of the xamarin-android directory.  Why would there be a `xamarin-android` directory *anywhere*?

jonpryor: ...or is one of the changes in this PR a change to that semantic?  In which case what *is* the `/Users/vsts/agent/2.155.1/work/1/s` directory?

pjcollins: `$(System.DefaultWorkingDirectory)` is the clone of `xamarin-android` only if one `checkout` is specified. When multiple checkouts are specified, this default behavior changes, though we have some control over it. https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/multi-repo-checkout?view=azure-devops#checkout-path has some more info in that regard.

PR #4104, opened 2020-01-09
[Xamarin.Android.Build.Tasks] improve <RemoveDirFixed/> behavior

jonathanpeppers:
We got report of a failure such as:

    XARDF7004: System.ArgumentException: Illegal characters in path.
        at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
        at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)
        at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
        at Xamarin.Android.Tasks.RemoveDirFixed.RunTask()
    XARDF7000: System.ArgumentException: Illegal characters in path.
        at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
        at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)
        at System.IO.Directory.Delete(String fullPath, String userPath, Boolean recursive, Boolean checkHost)
        at Xamarin.Android.Tasks.RemoveDirFixed.RunTask()

We were reviewing the `<RemoveDirFixed/>` MSBuild task, and saw some
error cases that were inconsistent with `<RemoveDir/>`:

https://github.com/microsoft/msbuild/blob/8d37fbb97003e96516a2054e51dd5fabfd720ed6/src/Tasks/RemoveDir.cs

1. `<RemoveDir/>` always continues on to deleting all directories if
   one fails.
2. `<RemoveDir/>` returns `false` on failure.

I think we should emulate the same behavior as MSBuild.

Additionally:

* We should be calling `Log.LogUnhandledException (TaskPrefix, ex)`
  for proper error codes.
* We should log both errors if we attempt a retry operation, such as
  in the cases of `UnauthorizedAccessException` and
  `DirectoryNotFoundException`.

I do not think this will fix the error encountered, however. It will
merely emit better error messages and abort the build at this failure.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/RemoveDirFixed.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/RemoveDirFixed.cs
@@ -29,7 +29,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using Microsoft.Build.Framework;
-using Microsoft.Build.Utilities;
 using Xamarin.Android.Tools;
 
 namespace Xamarin.Android.Tasks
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/RemoveDirFixed.cs
```diff
@@ -61,11 +60,15 @@ public override bool RunTask ()
...
+							fullPath = Files.ToLongPath (fullPath);
+							Log.LogDebugMessage ("Trying long path: " + fullPath);
+							Directory.Delete (fullPath, true);
+							temporaryRemovedDirectories.Add (directory);
+							continue;
+						} catch (Exception inner) {
+							Log.LogErrorFromException (inner);
+						}
 					}
 					Log.LogErrorFromException (ex);
```
jpobst: This will cause both errors to get logged, which may be desired?

src/Xamarin.Android.Build.Tasks/Tasks/RemoveDirFixed.cs
```diff
@@ -76,7 +79,7 @@ public override bool RunTask ()
 			RemovedDirectories = temporaryRemovedDirectories.ToArray ();
 			Log.LogDebugTaskItems ("  RemovedDirectories: ", RemovedDirectories);
 
-			return true;
+			return !Log.HasLoggedErrors;
```
jpobst: I think this technically would return `false` if a previous task logged an error, but let the build keep going.  Not sure if that matters.

jpobst:
I think we have the same issue on the other `catch` as well:

https://github.com/xamarin/xamarin-android/pull/4104/files#diff-7328014ed9b356ba2131086a455c21d1R58

src/Xamarin.Android.Build.Tasks/Tasks/RemoveDirFixed.cs
jonathanpeppers: I think we want both logged. Is two errors fine, or should I use `AggregateException`?

All these calls should also be doing:
```csharp
Log.LogUnhandledException (TaskPrefix, ex);
````
I will change that.

src/Xamarin.Android.Build.Tasks/Tasks/RemoveDirFixed.cs
jonathanpeppers: Almost every task we return `!Log.HasLoggedErrors`, I think it is a good pattern.

Issue #4057, opened 2019-12-19
[linker] Consider removing attributes from non-linked assemblies


radekdoulik:
When working on https://github.com/xamarin/xamarin-android/issues/3911 I noticed that we remove the attributes from assemblies with action set to `Link` only.

That means that we usually remove them from Sdk assemblies (SdkOnly linker mode is the default and most common). Thus the application and 3rd party referenced assemblies (including nuget assemblies) are left untouched.

Removing attributes in `RemoveAttributes` should be relatively safe, so it might help to remove them in all assemblies.

Needs more investigation, what the size saving would be and what would be build time penalty (usually only in Release configuration, so shouldn't be that important)

PR #4107, opened 2020-01-10
[linker] Remove known attributes in all assemblies

radekdoulik:
Context: https://github.com/xamarin/xamarin-android/issues/4057

It reduces the size of the apk slightly, the performance is nearly the
same, so I think it still makes sense to do.

SmartHotel360:

apk size difference 32768 bytes:

old 50991603
new 50958835

build time:

old    16191 ms  LinkAssemblies                             1 calls
new    16207 ms  LinkAssemblies                             1 calls

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveAttributes.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveAttributes.cs
@@ -15,8 +15,52 @@ namespace MonoDroid.Tuner {
 
 	public class RemoveAttributes : RemoveAttributesBase {
 
+		bool changed;
+
 		protected virtual bool DebugBuild {
 			get { return context.LinkSymbols; }
+		}
+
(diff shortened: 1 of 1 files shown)


jonpryor:
This change breaks the [`MonoTests.System.Reflection.FieldInfoTest.GetCustomAttributes` test](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3375051&view=ms.vss-test-web.build-test-results-tab&runId=10544760&resultId=117613&paneView=debug):

```
Expected: 1
 But was: 0
```

Probably because of this assert:

https://github.com/mono/mono/blob/2019-10/mcs/class/corlib/Test/System.Reflection/FieldInfoTest.cs#L201-L202

```csharp
attrs = fi.GetCustomAttributes (typeof (ObsoleteAttribute), false);
Assert.AreEqual (1, attrs.Length, "#B7");
```

which asserts that `ObsoleteAttribute` is found...which is a type removed in this PR.

We might need to run this idea by the #mono team, and possibly update the unit tests to assert that some *other* attribute type exists.

PR #4113, opened 2020-01-11
Use XA version for unofficial SDK if it is found

kzu:
Introduces a "lite" SDK that redirects to an existing Xamarin.Android
installation on disk. Please note that this redirect will not work with 
`dotnet` commands.

Adds the Xamarin.Android product version to the early msbuild project
SDK prototype, which makes it easier to work with this package in local
environments. Adding a versioning schema will also allow us to push
these packages to a feed on demand.

This feed can be accessed by adding the following package source:

    <add key="xamarin-impl" value="https://pkgs.dev.azure.com/azure-public/vside/_packaging/xamarin-impl/nuget/v3/index.json" />

For now, we will only push packages to this feed when builds are 
manually triggered with the `PushXAPackages` environment variable set to
`true`. The value of this variable will default to false.


pjcollins:
Unfortunately any package version changes will need to be synced with a test update -- https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs#L14

We'll likely be revisiting this alongside other broader package changes in the coming weeks/months.

kzu:
well, not really. I can fix that too :)

pjcollins:
I wasn't able to specify a version range or other wildcard when working with a NuGet provided msbuild project SDK, do you know of a better way that would avoid having to hard code the full version in a project which uses the package?

kzu:
The easiest way I've found is to use compile-time generated assembly-level attributes :). This is, of course, way simpler in SDK-style projects, but it's still somewhat easy to achieve in long-form too, as shown here. It could be generalized and placed in a top-level .targets, if deemed useful (but I think it would just be better to switch to SDK style for all projects, like xamarin-macios is doing).

kzu:
Aaaand, the SDK project tests now passed :)

build-tools/create-packs/Packager.csproj
```diff
@@ -4,7 +4,8 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <IncludeBuildOutput>false</IncludeBuildOutput>
     <PackageId>Xamarin.Android.Sdk</PackageId>
-    <PackageVersion>0.0.1</PackageVersion>
+    <PackageVersion Condition="Exists('..\..\bin\$(Configuration)\Version')">$([System.IO.File]::ReadAllText('..\..\bin\$(Configuration)\Version'))</PackageVersion>
+    <PackageVersion Condition="$(PackageVersion) == ''">0.0.1</PackageVersion>
```
jonpryor: Shouldn't this use quotes, `'$(PackageVersion)' == ''`?

pjcollins:
If we want to push this to a feed, we may need to consider some git hash based versioning to ensure PR builds have unique versions? Without using a 4th "revision" we're effectively back in the same boat of using `v0.0.1`:

>  ##[error]The nuget command failed with exit code(1) and error(System.AggregateException: One or more errors occurred. (Response status code does not indicate success: 409 (Conflict - The feed already contains 'Xamarin.Android.Sdk 10.2.99'.

kzu:
I think we can force push, no big deal. On the other hand, I though the XA Version already did that?

build-tools/automation/azure-pipelines.yaml
```diff
@@ -180,6 +180,15 @@ stages:
...
         packDirectory: $(System.DefaultWorkingDirectory)/xamarin-android/bin/Build$(XA.Build.Configuration)/$(NuGetArtifactName)
 
+    - task: NuGetCommand@2
+      displayName: push Xamarin.Android.Sdk
+      continueOnError: true
+      inputs:
+        command: push
+        packagesToPush: '$(System.DefaultWorkingDirectory)/xamarin-android/bin/Build$(XA.Build.Configuration)/$(NuGetArtifactName)/*.nupkg'
+        nuGetFeedType: external
+        publishFeedCredentials: 'xamarin-impl public feed'
```
dellis1972: @pjcollins @kzu are we really pushing this to Nuget? Or is this to an internal pipelines feed?

kzu: We are pushing this via nuget to an undisclosed feed used for CI/temp/internal stuff that's not really critical or important to keep in a private secure feed (much like the VSSDK stuff is all available that way in another feed just like this). It's a new feature DevOps added for this very purpose, and was first enabled for the VSSDK and we asked to have the same for our stuff.

kzu:
@pjcollins can we merge this so we have something at least better than the existing one which has *no versioning* at all? Otherwise, it's extremely hard to iterate on this. 

For now I'm consuming the PR-produced package, but I don't see why we wouldn't want to merge this which is already an improvement :)

pjcollins:
@kzu sorry, I thought we still had msbuild test failures here and that this was still in progress. If this latest run looks good I'm happy to merge it.

kzu:
Ok, checked the build errors, they don't seem to be related to the test, AFAICS. They seem like unrelated plain build errors :/

pjcollins:
I re-ran this yet again, we had an unfortunately timed build issue on that previous run that has since been resolved. The job is here - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3422872&view=results

pjcollins:
@kzu the couple of tests we had which consume this SDK are failing with these changes - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3422872&view=ms.vss-test-web.build-test-results-tab

```
C:\A\xamarin-v00000K-1\_work\1\s\bin\TestRelease\temp\DotNetPackageXASdkProjectFalse\UnnamedProject.csproj : error : Unable to find package Xamarin.Android.Sdk with version (= 10.2.99)
C:\A\xamarin-v00000K-1\_work\1\s\bin\TestRelease\temp\DotNetPackageXASdkProjectFalse\UnnamedProject.csproj : error :   - Found 1 version(s) in testsource1 [ Nearest version: 10.2.99.185 ]
```

src/Xamarin.Android.Sdk/Xamarin.Android.Sdk.Lite.proj
```diff
@@ -0,0 +1,24 @@
...
+        <!-- The Lite SDK replaces the full Sdk.props/Sdk.targets -->
+        <_PackageFiles Include="targets\Xamarin.Android.Sdk.Lite.props" PackagePath="targets\Xamarin.Android.Sdk.props" />
+        <_PackageFiles Include="targets\Xamarin.Android.Sdk.Lite.targets" PackagePath="targets\Xamarin.Android.Sdk.targets" />
+        <!-- But it still imports the full SDK props by renaming it -->
+        <_PackageFiles Include="targets\Xamarin.Android.Sdk.props" PackagePath="targets\Xamarin.Android.Sdk.Default.props" />
+
+        <_PackageFiles Include="Sdk\*.*" PackagePath="Sdk" />
+        <_PackageFiles Include="README.md" PackagePath="." />
+    </ItemGroup> 
+</Project>
```
pjcollins: To follow the existing convention of using `build-tools` for packaging, maybe we should move this to `build-tools/create-packs/Xamarin.Android.Sdk.Lite.proj`?

src/Xamarin.Android.Build.Tasks/Tests/Directory.Build.targets
```diff
@@ -0,0 +1,35 @@
...
+      <Compile Remove="$(GeneratedSdkVersionFile)" />
+      <_SdkVersionInfo Include="System.Reflection.AssemblyMetadata">
+        <_Parameter1>SdkVersion</_Parameter1>
+        <_Parameter2>$(_SdkVersion)</_Parameter2>
+      </_SdkVersionInfo>
+    </ItemGroup>
+
+    <WriteCodeFragment AssemblyAttributes="@(_SdkVersionInfo)" Language="$(Language)" OutputFile="$(GeneratedSdkVersionFile)" >
+      <Output TaskParameter="OutputFile" ItemName="Compile" />
+      <Output TaskParameter="OutputFile" ItemName="FileWrites" />
```
jonathanpeppers: If `GenerateSdkVersionInfo` is skipped due to the `Inputs` / `Outputs` being up to date.

Will `$(GeneratedSdkVersionFile)` get added to `@(Compile)` and `@(FileWrites)`?

Should this be at the bottom of the target instead of using `<Output/>`?
```xml
<ItemGroup>
  <Compile Include="$(GeneratedSdkVersionFile)" />
  <FileWrites Include="$(GeneratedSdkVersionFile)" />
</ItemGroup>
```

kzu: Yes, absolutely!

pjcollins:
Previous build with successful push - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3571171&view=logs&jobId=58fa601a-973f-52de-d6ce-8b812c2da347&j=96fd57f5-f69e-53c7-3d47-f67e6cf9b93e&t=20f66bcf-9f97-5624-1289-1f1a0c1fb0fe

PR #4120, opened 2020-01-13
Implicit operator for Color to to enable int assignment

mister-giga:
it enables:
Color c = -65536; //Red
Simple conversion and intuitive since Color can be assigned to int parameter/field, vice versa shall be available.

### DIFF
--- a/src/Mono.Android/Android.Graphics/Color.cs
+++ b/src/Mono.Android/Android.Graphics/Color.cs
@@ -176,6 +176,11 @@ namespace Android.Graphics
 		{
 			return c.color;
 		}
+		
+		public static implicit operator Color(int argb)
+		{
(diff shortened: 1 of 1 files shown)


jonpryor:
There is a problem with this PR.  Consider this code, in which `.ToAndroid()` returns an `Android.Graphics.Color`:

```
SetScrimColor(isShowingSplit ? Color.Transparent.ToAndroid() : (int)DefaultScrimColor);
```

This code compiles Just Fine with d16-5, but with this change it results in a CS0172 error:

```
Type of conditional expression cannot be determined because 'Color' and 'int' implicitly convert to one another.
```

Oops.

I fear we'll need to revert, as otherwise this would be an source compatibility change.

PR #4121, opened 2020-01-14
[linker][docs] Adjust AppDomain.CreateDomain warning wording

brendanzagaeski:
Putting up this PR just in case any of these changes might be handy. Feel free to take any, all, or none of them.

* * *

Remove the extra `Warning: ` prefix to match up with warnings from the
other MSBuild tasks.

Switch to the C# `AppDomain.CreateDomain()` syntax instead of the CIL
`AppDomain::CreateDomain` syntax.

Adjust the second sentence of the warning to align with the information
from the [System.AppDomain API documentation][0] about how there is only
one `AppDomain` in .NET Core.

Tidy up a few other little items in the docs page.

[0]: https://docs.microsoft.com/dotnet/api/system.appdomain?view=netcore-3.1

### DIFF
--- a/Documentation/guides/messages/xa2000.md
+++ b/Documentation/guides/messages/xa2000.md
@@ -1,18 +1,18 @@
 ---
-title: Xamarin.Android error XA2000
+title: Xamarin.Android warning XA2000
 description: XA2000 warning code
-ms.date: 12/3/2019
+ms.date: 1/13/2019
 ---
 # Xamarin.Android warning XA2000
-
-Member semantics will change in a future release
 
 ## Example messages
 
 ```
-Warning: Use of AppDomain::CreateDomain detected in assembly: {assembly}. The AppDomain's will not be part of .NET 5 and therefore will be missing in newer Xamarin.Android releases.
+Use of AppDomain.CreateDomain() detected in assembly: {assembly}. .NET 5 will only support a single AppDomain, so this API will no longer be available in Xamarin.Android once .NET 5 is released.
 ```
 
 ## Solution
 
-Stop using `AppDomain`s and replace it with a different API. Like `AssemblyLoadContext`, see https://docs.microsoft.com/en-us/dotnet/standard/assembly/unloadability for example.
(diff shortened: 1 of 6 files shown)


brendanzagaeski:

Move the `warning XA2000:` message prefix to be part of the message string itself, and use [`TaskLoggingHelper.LogMessageFromText()`][0] to parse that prefix back out of the string and set the message severity level and code.  The reason to do this is that `<LinkAssemblies/>` calls `CheckAppDomainUsage()` indirectly via `Linker.Process()`, so it doesn't have a way to pass in a custom logger.  In that scenario, the only logging method available is `LinkAssemblies.LogMessage()`, which means that that one method needs to handle not only messages but also warnings and errors.

Unlike `<LinkAssemblies/>`, `<LinkAssembliesNoShrink/>` calls `CheckAppDomainUsage()` *directly*, so it *can* pass in a custom logger, but because the message prefix is now part of the message string itself, `<LinkAssembliesNoShrink/>` must now also use `LogMessageFromText()`.

This new use of `LogMessageFromText()` is similar to how `AndroidToolTask` uses `MSBuildExtensions.LogFromStandardError()` to log messages from external tools.  In a way, the linker is also an external tool because the `Linker` class comes from the Mono dependency rather than being a direct part of the Xamarin.Android build tasks.

Other changes:

Update the unit tests to check that MSBuild actually treats the message as a warning.

[0]: https://docs.microsoft.com/dotnet/api/microsoft.build.utilities.tasklogginghelper.logmessagefromtext?view=netframework-4.8

* * *

Before these changes, Release builds for apps using `AppDomain.CreateDomain()` produced MSBuild *messages* rather than an MSBuild *warnings*. That problem can be seen in Xamarin.Android 10.2.0.84 in Visual Studio 2019 version 16.5 Preview 2.

An unrelated side topic for possible future consideration is whether the custom `LogFromStandardError()` from the Xamarin.Android build tasks does anything special that `LogMessageFromText()` doesn't do. If not, then perhaps that custom method can be deprecated in favor of `LogMessageFromText()`.

PR #4124, opened 2020-01-15
[tests] use string.Contains() in Xamarin.ProjectTools

jonathanpeppers:
Context: https://build.azdo.io/3384618

We are seeing various PRs fail with:

    System.MissingMethodException : Method not found: 'Boolean System.String.Contains(Char)'.
      at Xamarin.ProjectTools.XamarinProject.UpdateProjectFiles(String directory, IEnumerable`1 projectFiles, Boolean doNotCleanup)
      at Xamarin.ProjectTools.ProjectBuilder.Save(XamarinProject project, Boolean doNotCleanupOnUpdate, Boolean saveProject) in /Users/builder/azdo/_work/3/s/xamarin-android/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/ProjectBuilder.cs:line 47
      at Xamarin.ProjectTools.ProjectBuilder.Build(XamarinProject project, Boolean doNotCleanupOnUpdate, String[] parameters, Boolean saveProject, Dictionary`2 environmentVariables) in /Users/builder/azdo/_work/3/s/xamarin-android/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/ProjectBuilder.cs:line 68
      at Xamarin.Android.Build.Tests.Aapt2Tests.Aapt2Disabled() in /Users/builder/azdo/_work/3/s/xamarin-android/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/Aapt2Tests.cs:line 171

This doesn't quite make sense, the failing line is:

    if (path.Contains (Path.DirectorySeparatorChar)) {

Which is actually using:

    namespace System.Linq {
        public static class Enumerable {
            public static bool Contains<TSource>(this IEnumerable<TSource> source, TSource value)

I changed this to use `Path.DirectorySeparatorChar.ToString()`, so
`string.Compare` will be used instead.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/XamarinProject.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/XamarinProject.cs
@@ -272,7 +272,7 @@ namespace Xamarin.ProjectTools
 					continue;
 				}
 				string filedir = directory;
-				if (path.Contains (Path.DirectorySeparatorChar)) {
+				if (path.Contains (Path.DirectorySeparatorChar.ToString ())) {
 					filedir = Path.GetDirectoryName (path);
 					if (!Directory.Exists (filedir) && (p.Content != null || p.BinaryContent != null))
 						Directory.CreateDirectory (filedir);


jonpryor:
The explanation here is partially incomplete.  The problem isn't *per-se* that `path.Contains(Path.DirectorySeparatorChar)` is using `Enumerable.Contains()`.

The problem is more subtle: the system mono `mscorlib.dll` assembly has a `System.String.Contains(Char)` method:

```
$ monodis --method /Library/Frameworks/Mono.framework//Versions/6.6.0/lib/mono/4.5/mscorlib.dll
...
########## System.String
...
3674: instance default bool Contains (string 'value')  (param: 5966 impl_flags: cil managed )
3675: instance default bool Contains (string 'value', valuetype System.StringComparison comparisonType)  (param: 5967 impl_flags: cil managed )
3676: instance default bool Contains (char 'value')  (param: 5969 impl_flags: cil managed )
3677: instance default bool Contains (char 'value', valuetype System.StringComparison comparisonType)  (param: 5970 impl_flags: cil managed )
```

Here we see a `String.Contains(Char)` overload.

Compare to the "4.7 reference assembly":

```
$ monodis --method /Library/Frameworks/Mono.framework//Versions/6.6.0/lib/mono/4.7-api/mscorlib.dll
...
########## System.String
...
2464: instance default bool Contains (string 'value')  (param: 3550 impl_flags: cil managed )
```

The reference assembly for .NET 4.7's `mscorlib.dll` contains *only* a `String.Contains(String)` method.

The problem is that, for a currently unknown reason, when we build `Xamarin.ProjectTools.csproj` with `xabuild` on macOS, it uses the "real" `mscorlib.dll`, not the "4.7 reference assembly" `mscorlib.dll`, and thus the compiler finds -- and *uses* -- the `String.Contains(Char)` method.  A method which doesn't exist on .NET.

Thus, when we attempt to execute this method on Windows + .NET 4.7, we get the `MissingMethodException`.

jonpryor:
...and now that `string.Contains()` is "fixed", now we hit `string.Split()`! https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3387110&view=ms.vss-test-web.build-test-results-tab&runId=10634888&resultId=100283&paneView=debug

```
System.MissingMethodException : Method not found: 'System.String[] System.String.Split(Char, System.StringSplitOptions)'.
```

"Same but different" explanation [as above](https://github.com/xamarin/xamarin-android/pull/4124#issuecomment-574839164): `Xamarin.ProjectTools.csproj` is built against the "wrong" `mscorlib.dll`, and the "wrong" `mscorlib.dll` contains a `string.Split(Char, StringSplitOptions)` method, which, uh, "does not exist": https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=netframework-4.8

jonpryor:
There are "only" (?) two paths forward:

 1. Continue playing whack-a-mole with all these runtime errors until we get C# source code which explicitly specifies the methods to use so that we are actually runnable on .NET 4.7, or
 2. Fix the `Xamarin.ProjectTools.csproj` build so that we compile against `.../4.7-api/mscorlib.dll`, *not* `.../4.5/mscorlib.dll`.

jonpryor:
PR #4125 is basically a form of (2) "Fix `Xamarin.ProejctTools.csproj`...".

As such I think PR #4125 supersedes this PR.

PR #4131, opened 2020-01-16
Bump to xamarin/monodroid/d16-5@dab75e88

jonathanpeppers:
Changes: https://github.com/xamarin/monodroid/compare/2c9ce23d...dab75e88

Fixes: https://work.azdo.io/1041849
Fixes: https://developercommunity.visualstudio.com/content/problem/858509/adb0030-monoandroidtoolsrequiresuninstallexception-2.html

Bump to xamarin/androidtools/d16-5@62844ab

Bump to xamarin/android-sdk-installer/d16-5@5adcbfa

[tools/msbuild] prevent InstallPackageAssemblies from reporting two errors

### DIFF
--- a/.external
+++ b/.external
@@ -1,2 +1,2 @@
-xamarin/monodroid:d16-5@2c9ce23d2288cc245729d32c2eabf6deebdf554b
+xamarin/monodroid:d16-5@dab75e88055c1bfaf53f810b0c3a9218fc53c60a
 mono/mono:2019-10@df42020fe6f7a0a1b5fa59b481fb603d88d72348


jonathanpeppers:
I can reproduce the failure here in a unit test in `android-sdk-installer`:
```
Failed to load local addon manifest: System.NullReferenceException: Object reference not set to an instance of an object.
```
Some other commit has caused an NRE with `/t:InstallAndroidDependencies`.

jonathanpeppers:
The failures seem like they will be fixed by: https://github.com/xamarin/xamarin-android/commit/d3281ac3f76f9bc7bbc56efa864d25ec0e2a06d5

There is one other test failure, `Xamarin.Android.Build.Tests.BuildTest.BuildWithResolveAssembliesFailure(False)`, but it seems like it failed to NuGet restore. I'll look and see if there is something we can do to fix the test.

PR #4132, opened 2020-01-16
[Xamarin.Android.Build.Tasks] Target netstandard2.0

pjcollins:
Updates `Xamarin.Android.Build.Tasks.csproj` to a short-form style
project which targets `netstandard2.0` to help ensure future compliance
with the dotnet version of MSBuild.

A few PackageReference related changes were needed to work with our
existing installer packaging logic, as NuGet assets are no longer copied
to the OutputDirectory by default for short-form style projects. To work
around this, we're setting `CopyLocalLockFileAssemblies` to true to copy
over all relevant NuGet package content. This property doesn't work for
debug symbols and certain NuGet folder hierarchies however. The new
`_CopyExtraPackageContent` target has been added to copy over these
unexpected files. This target relies on the `GeneratePathProperty`
metadata that is set on the `PackageReference` element.

Additionally, the behavior of the `netstandard2.0` version of
`System.CodeDom` seems to have changed slightly. The Runtime Version
information that was previously added to the `<auto-generated>` comment
header is now only added when running on mono. The test cases in
`ManagedResourceParserTests` have been updated to handle this. This
assembly has also been added to our ILRepack target as it is not available
by default in `netstandard2.0`.

Finally, we are explicitly importing `Microsoft.NET.Sdk.props` and
`Microsoft.NET.Sdk.targets` in `Xamarin.Android.Build.Tasks.csproj` to
work around an issue that prevents us from overriding `<BuildDependsOn>`
in `Xamarin.Android.Build.Tasks.targets`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
@@ -327,9 +327,9 @@ int xml myxml 0x7f140000
 			var content1 = File.ReadAllText (file1);
 			var content2 = File.ReadAllText (file2);
 			// This string is only generated when running on mono, replace with a new line that will be stripped when comparing.
-			string versionInfo = "//     Runtime Version:4.0.30319.42000";
-			content1 = content1.Replace (versionInfo, Environment.NewLine);
-			content2 = content2.Replace (versionInfo, Environment.NewLine);
+			var runtimeVersionRegex = new Regex (@"//\s*Runtime Version:.*");
+			content1 = runtimeVersionRegex.Replace (content1, Environment.NewLine);
+			content2 = runtimeVersionRegex.Replace (content2, Environment.NewLine);
 
 			using (var s1 = new MemoryStream (Encoding.UTF8.GetBytes (content1)))
 			using (var s2 = new MemoryStream (Encoding.UTF8.GetBytes (content2))) {
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
@@ -320,6 +320,27 @@ int xml myxml 0x7f140000
 			}
 		}
 
+		void CompareFilesIgnoreRuntimeInfoString (string file1, string file2)
+		{
+			FileAssert.Exists (file1);
+			FileAssert.Exists (file2);
+			var content1 = File.ReadAllText (file1);
+			var content2 = File.ReadAllText (file2);
+			// This string is only generated when running on mono, replace with a new line that will be stripped when comparing.
+			string versionInfo = "//     Runtime Version:4.0.30319.42000";
+			content1 = content1.Replace (versionInfo, Environment.NewLine);
+			content2 = content2.Replace (versionInfo, Environment.NewLine);
+
+			using (var s1 = new MemoryStream (Encoding.UTF8.GetBytes (content1)))
+			using (var s2 = new MemoryStream (Encoding.UTF8.GetBytes (content2))) {
+				if (!StreamCompare (s1, s2)) {
+					TestContext.AddTestAttachment (file1, Path.GetFileName (file1));
+					TestContext.AddTestAttachment (file2, Path.GetFileName (file2));
+					Assert.Fail ($"{file1} and {file2} do not match.");
+				}
+			}
+		}
+
 		[Test]
 		public void GenerateDesignerFileWithÜmläüts ()
 		{
@@ -348,7 +369,7 @@ int xml myxml 0x7f140000
 			Assert.IsTrue (task.Execute (), "Task should have executed successfully.");
 			Assert.IsTrue (File.Exists (task.NetResgenOutputFile), $"{task.NetResgenOutputFile} should have been created.");
 			var expected = Path.Combine (Root, "Expected", "GenerateDesignerFileExpected.cs");
-			AssertFileContentsMatch (task.NetResgenOutputFile, expected);
+			CompareFilesIgnoreRuntimeInfoString (task.NetResgenOutputFile, expected);
 			Directory.Delete (Path.Combine (Root, path), recursive: true);
 		}
 
@@ -388,7 +409,7 @@ int xml myxml 0x7f140000
 			Assert.IsTrue (task.Execute (), "Task should have executed successfully.");
 			Assert.IsTrue (File.Exists (task.NetResgenOutputFile), $"{task.NetResgenOutputFile} should have been created.");
 			var expected = Path.Combine (Root, "Expected", withLibraryReference ? "GenerateDesignerFileWithLibraryReferenceExpected.cs" : "GenerateDesignerFileExpected.cs");
-			AssertFileContentsMatch (task.NetResgenOutputFile, expected);
+			CompareFilesIgnoreRuntimeInfoString (task.NetResgenOutputFile, expected);
 			Directory.Delete (Path.Combine (Root, path), recursive: true);
 		}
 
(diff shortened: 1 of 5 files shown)


pjcollins:
I've finally narrowed down the hang in `Xamarin.Android.Build.Tests` on macOS, and it appears appears to be caused by a crash in the F# flavors of this test:
https://github.com/xamarin/xamarin-android/blob/38a10a17eaf16b052e3c5ffbee502565acc688f5/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs#L517

```
FSharp.Compiler.CodeDom.FSharpCodeGenerator doesn't implement interface System.CodeDom.Compiler.ICodeGenerator
```
```


	Native Crash Reporting
Got a SIGABRT while executing native code. This usually indicates
used by your application.

	Native stacktrace:
	0x10174b8db - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : mono_dump_native_crash_info
	0x1016e3f15 - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : mono_handle_native_crash
	0x10174ae9b - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : sigabrt_signal_handler
	0x7fff647edb5d - /usr/lib/system/libsystem_platform.dylib : _sigtramp
	0x1019a32ff - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : monoeg_g_timer_elapsed
	0x7fff646a76a6 - /usr/lib/system/libsystem_c.dylib : abort
	0x101947dc7 - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : monoeg_assert_abort
	0x101928cff - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : mono_log_write_logfile
	0x10194814e - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : monoeg_g_logv_nofree
	0x101948202 - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : monoeg_g_log
	0x1016e61ae - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : mini_resolve_imt_method
	0x1016e66ca - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : common_call_trampoline
	0x1016e7a2d - /Library/Frameworks/Mono.framework/Versions/6.8.0/bin/mono : mono_vcall_trampoline
	0x101b5e513 - Unknown
	0x106d0874c - Unknown
	0x106be95ab - Unknown
	0x105deef57 - Unknown
	0x105dd6930 - Unknown
	0x105dd050b - Unknown
	0x105dcf7e3 - Unknown
	0x105dcf65b - Unknown

	Telemetry Dumper:
Pkilling 0x70000938a000 from 0x700009790000
Pkilling 0x700009144000 from 0x700009790000
Pkilling 0x700009187000 from 0x700009790000
Pkilling 0x700008f41000 from 0x700009790000
Pkilling 0x106ea05c0 from 0x700009790000
Pkilling 0x700008d3e000 from 0x700009790000
Pkilling 0x700008a35000 from 0x700009790000
Pkilling 0x700009d99000 from 0x700009790000
Pkilling 0x700009b96000 from 0x700009790000
Pkilling 0x700009993000 from 0x700009790000
Pkilling 0x70000958d000 from 0x700009790000
Entering thread summarizer pause from 0x700009790000
Finished thread summarizer pause from 0x700009790000.

Waiting for dumping threads to resume

	Basic Fault Address Reporting
Memory around native instruction pointer (0x7fff6473d2c6):0x7fff6473d2b6  ff ff c3 90 90 90 b8 48 01 00 02 49 89 ca 0f 05  .......H...I....
0x7fff6473d2c6  73 08 48 89 c7 e9 87 a1 ff ff c3 55 48 89 e5 41  s.H........UH..A
0x7fff6473d2d6  57 41 56 41 55 41 54 53 48 81 ec 88 00 00 00 4d  WAVAUATSH......M
0x7fff6473d2e6  89 cf 49 89 f4 48 8d 05 76 52 5b 36 48 8b 00 48  ..I..H..vR[6H..H

	Managed Stacktrace:
	  at <unknown> <0xffffffff>
	  at Xamarin.Android.Tasks.GenerateResourceDesigner:WriteFile <0x00b1b>
	  at Xamarin.Android.Tasks.GenerateResourceDesigner:RunTask <0x0148a>
	  at Xamarin.Android.Tasks.AndroidTask:Execute <0x00046>
```

This test is skipped on Windows, which is why those jobs are green.

pjcollins:
Manual triggered full test run can be found here - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3422053&view=results.

pjcollins:
I think this should be good to go. We had one timeout during artifact download in  `Regression Tests TimeZone - Mac-1`. The failure on `Regression Tests Windows-3` has been fixed by https://github.com/xamarin/QualityAssurance/commit/b1ffb3956d3502e6e8a3e4670a90acb4e76aa3e7

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
```diff
@@ -320,6 +320,27 @@ void BuildLibraryWithResources (string path)
...
 		}
 
+		void CompareFilesIgnoreRuntimeInfoString (string file1, string file2)
+		{
+			FileAssert.Exists (file1);
+			FileAssert.Exists (file2);
+			var content1 = File.ReadAllText (file1);
+			var content2 = File.ReadAllText (file2);
+			// This string is only generated when running on mono, replace with a new line that will be stripped when comparing.
+			string versionInfo = "//     Runtime Version:4.0.30319.42000";
```
jonathanpeppers: Should we make this a `Regex` just in case the version number changes? If this just removed an entire line that matches: `^\/\/\s*Runtime Version:`

jonpryor:
> Finally, we are explicitly importing `Microsoft.NET.Sdk.props` and
> `Microsoft.NET.Sdk.targets` in `Xamarin.Android.Build.Tasks.csproj` to
> work around an issue that prevents us from overriding `<BuildDependsOn>`
> in `Xamarin.Android.Build.Tasks.targets`.

As per https://github.com/dotnet/sdk/pull/579#issuecomment-270515692 perhaps the "proper" way is to use `AfterTargets="Build"` instead of overriding `$(BuildDependsOn)`?

(Since, apparently, `$(*DependsOn)` is not a pattern that will necessarily be supported long-term?)

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
jonpryor: Not sure this requires a *regex*.  Removing any line starting with `//     Runtime Version:` would likely be sufficient.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
```diff
@@ -83,112 +56,13 @@
...
-    <Compile Include="Tasks\AndroidSignPackage.cs" />
-    <Compile Include="Tasks\RemoveDirFixed.cs" />
-    <Compile Include="Tasks\ResolveJdkJvmPath.cs" />
-    <Compile Include="Tasks\ResolveSdksTask.cs" />
-    <Compile Include="Tasks\CompileToDalvik.cs" />
-    <Compile Include="Tasks\AndroidToolTask.cs" />
-    <Compile Include="Tasks\PrepareAbiItems.cs" />
-    <Compile Include="Properties\AssemblyInfo.cs" />
-    <Compile Include="Properties\Resources.Designer.cs">
-      <AutoGen>True</AutoGen>
```
jonpryor: Do we no longer need this metadata?  (I'm not sure what the metadata is for or who uses it, which makes it doubly difficult to reason about...)

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
pjcollins: Updated to use a regex, the regex replace behavior seems like it'll be easier to read than the alternative.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
pjcollins: It seems that this metadata is not needed by short-form style projects, based on this comment https://github.com/dotnet/sdk/issues/1199#issuecomment-332052321

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
```diff
@@ -690,41 +411,25 @@
...
       <ReferenceOutputAssembly>False</ReferenceOutputAssembly>
     </ProjectReference>
+    <ProjectReference Include="..\..\external\xamarin-android-tools\src\Xamarin.Android.Tools.AndroidSdk\Xamarin.Android.Tools.AndroidSdk.csproj" />
+    <ProjectReference Include="..\..\external\Java.Interop\src\Java.Interop.Export\Java.Interop.Export.csproj" />
+    <ProjectReference Include="..\..\external\Java.Interop\src\Java.Interop\Java.Interop.csproj" />
   </ItemGroup>
+
+  <!-- Import Microsoft.NET.Sdk targets before our targets so we can override behavior -->
+  <!-- See https://github.com/microsoft/msbuild/pull/4922 -->
+  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
```
dellis1972: On previous projects where I have had to do this I only needed the `Sdk.targets` import. so the `Sdk.props` one *might* be redundant.

PR #4134, opened 2020-01-16
[build-tools] Remove tools replaced by xaprepare

pjcollins:
Removes project/targets files are no longer used. Their functionality
has been replaced by the `xaprepare` tool.


pjcollins:
Looks like the only remaining failures here are unrelated to these changes.

PR #4138, opened 2020-01-17
[Xamarin.Android.Build.Tasks] prototype of a Java "daemon"

jonathanpeppers:
This creates a new mode for `JavaToolTask`, where if `UseDaemon=True`:

* Instead of just running a new `java.exe`, a new `JavaDaemonClient`
  class will start a java process that runs `java-daemon.jar`.
* The daemon reads stdin to run a jar file's `Main` method, and
  replies to stdout with the result of running that jar file.
* If `JavaDaemonClient` is disposed (and when the MSBuild node shuts
  down), the java process will exit.

Some of the *gotchas* here were:

* We cannot change the working directory of `java` after the process
  has started:

https://stackoverflow.com/a/840229

To resolve this, we have to create an instance of the daemon for each
working directory. This should only happen if a solution has multiple
Xamarin.Android head projects.

* The `-classpath` of the daemon must include all possible jar files
  that might be used by the build.

I used the `@(_JavaDaemonClassPath)` for this.

* There is not currently a way to set java properties with the daemon,
  such as `-Dfile.encoding=UTF8`.

There were two usages of this in `<CreateMultiDexMainDexClassList/>
and `<CompileToDalvik/>`. I will need to add something to pass these
through to the daemon if they are required.

* We need low memory usage for the daemon after a request has
  completed, when it is idle.

Originally `java.exe` was using ~531MB of RAM on my PC that has a
total of 64GB of RAM. For comparison `MSBuild.exe` is using ~62MB.

I added a call to `System.gc()` and used the `-XX:+UseG1GC` switch to
get `java.exe` idling at ~90MB.

https://www.javacodegeeks.com/2017/11/minimize-java-memory-usage-right-garbage-collector.html

~~ Results ~~

I built the Xamarin.Forms project in this repo, using
`%MSBUILDNOINPROCNODE%` = 1.

The command was:

    git clean -dxf tests
    xabuild tests\Xamarin.Forms-Performance-Integration\Droid\Xamarin.Forms.Performance.Integration.Droid.csproj /r /t:SignAndroidPackage /bl /clp:performancesummary /p:AndroidUseJavaDaemon=True /p:AndroidManifestMerger=manifestmerger.jar

Not using daemon:

     303 ms  ManifestMerger                             1 calls
     806 ms  AndroidApkSigner                           1 calls
    3804 ms  D8                                         1 calls
    Time Elapsed 00:00:15.24

Using daemon, after daemon is running:

     151 ms  ManifestMerger                             1 calls
     437 ms  AndroidApkSigner                           1 calls
    3506 ms  D8                                         1 calls
    Time Elapsed 00:00:14.55

This seems to save ~1 second of build time for these three tasks that
run java processes.

In past iterations, before I changed java GC settings. I was seeing ~7
seconds saving for this app. I need to do more investigation there, as
maybe we can do something to only free up the memory after the entire
`Build`.


src/java-daemon/src/main/java/xamarin/android/JavaDaemon.java
```diff
@@ -0,0 +1,173 @@
...
+    private DocumentBuilder builder;
+    private Transformer transformer;
+
+    /**
+     * Main entry point
+     * @param args
+     */
+    public static void main (String[] args) {
+        try {
+            JavaDaemon daemon = new JavaDaemon();
```
jonpryor: Should we check for a `--help` arg?

jonathanpeppers:
Two problems here, that I need to work through:
1. My PC runs 16 `MSBuild.exe`'s inside VS Windows. That means we would have potentially 16 daemons in this case. This makes me think the communication over stdin/stdout isn't going to work. The daemon needs to support multiple processes talking to it.
2. How to use this inside the AndroidX.Migration NuGet package? Since AndroidX.Migration needs to work with current versions of Xamarin.Android, there would need to be a shared assembly somehow: `Xamarin.JavaDaemon.dll` or something.

PR #4142, opened 2020-01-17
[CI] Run integrated regression tests in parallel

jpobst:
Breaks up the `Integrated Regression` tests to run across 3 nodes, reducing the time taken to run from `~150 minutes` to `3 x ~50 minutes`.

Moves the `Integrated Regression` tests to a separate Stage to improve readability.  This also allows us to move the `condition` to the Stage level instead of each Job.

image

### DIFF
--- a/build-tools/automation/yaml-templates/run-integrated-regression-tests.yaml
+++ b/build-tools/automation/yaml-templates/run-integrated-regression-tests.yaml
@@ -54,15 +54,15 @@ steps:
 
 - template: run-installer.yaml
 
-- ${{ if eq(parameters.node_id, 1) }}:
-  - powershell: |   
-      if ([Environment]::OSVersion.Platform -eq "Unix") {
-         mono $(System.DefaultWorkingDirectory)/NUnit.ConsoleRunner/tools/nunit3-console.exe $(System.DefaultWorkingDirectory)/XQA.Android/XQA.Android.dll --where "cat == XAUnitTestSetup || cat == XATestPrep || cat == EnvironmentInfo"
-      } else {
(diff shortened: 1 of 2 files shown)


build-tools/automation/yaml-templates/run-integrated-regression-tests.yaml
```diff
@@ -51,160 +54,163 @@ steps:
...
-    testRunTitle: Test Environment Cleanup
-    nunitConsoleExtraArgs: --where "cat == XATestCleanUp"
+- ${{ if eq(parameters.node_id, 1) }}:
+  - powershell: |   
+      if ([Environment]::OSVersion.Platform -eq "Unix") {
+         mono $(System.DefaultWorkingDirectory)/NUnit.ConsoleRunner/tools/nunit3-console.exe $(System.DefaultWorkingDirectory)/XQA.Android/XQA.Android.dll --where "cat == XAUnitTestSetup || cat == XATestPrep || cat == EnvironmentInfo"
+      } else {
+         $(System.DefaultWorkingDirectory)\NUnit.ConsoleRunner\tools\nunit3-console.exe $(System.DefaultWorkingDirectory)\XQA.Android\XQA.Android.dll --where "cat == XAUnitTestSetup || cat == XATestPrep || cat == EnvironmentInfo"
+      }
+    displayName: Test Environment Setup
```
pjcollins: We'll want to move this step out and make sure it runs on every node before any other test categories. It's needed to make sure any attached devices/emulators are ready, and to prepare some test sources.

jpobst: Good call, I originally debated on this one for a while.  Moved it up to all nodes.

jpobst:
This is the first of several planned PRs that will spread our tests across multiple machines, so I'm happy to change it in future ones.  Do you have a better noun for `node`?

pjcollins:
I can't think of one of the fly, maybe removing it all together would be a bit more succinct? e.g. `Regression Tests Windows - 1`

PR #4143, opened 2020-01-17
[tests] Convert unit tests to short-form style projects

pjcollins:
Updates our desktop unit test projects to short-form style projects
targeting `net472`. Any existing `AssemblyInfo.cs` files have also been
removed from these projects. The unused `Test.cs` file which was part of
CodeBehind unit tests has also been removed.


pjcollins:
Re-running against latest xamarin-android/master, though the failures here are unrelated to the projects being changed.

tests/CodeGen-MkBundle/Xamarin.Android.MakeBundle-UnitTests/Xamarin.Android.MakeBundle-UnitTests.csproj
```diff
@@ -1,48 +1,21 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>{FA8EEC88-CA3C-4D69-B206-54B392570DC6}</ProjectGuid>
-    <OutputType>Library</OutputType>
+    <TargetFramework>net472</TargetFramework>
```
jonathanpeppers: Is there any value to making all these test projects `netstandard2.0`? Like maybe it would enable us to run `dotnet test` easily in the future?

pjcollins: Unfortunately NUnit.ConsoleRunner is not compatible with netstandard, so we'd need to rewrite all of our runner logic to make that change.

dellis1972:
@pjcollins can we not use the `--framework=4.7.1` command line argument for the nunit3-console.exe ? 
```
--framework=FRAMEWORK  FRAMEWORK type/version to use for tests.
                               Examples: mono, net-3.5, v4.0, 2.0, mono-4.0.
                               If not specified, tests will run under the
                               framework they are compiled with.
```

in theory that should allow us to use netstandard2.0...

dellis1972:
> @pjcollins can we not use the `--framework=4.7.1` command line argument for the nunit3-console.exe ?
> 
> ```
> --framework=FRAMEWORK  FRAMEWORK type/version to use for tests.
>                                Examples: mono, net-3.5, v4.0, 2.0, mono-4.0.
>                                If not specified, tests will run under the
>                                framework they are compiled with.
> ```
> 
> in theory that should allow us to use netstandard2.0...

Mind you that might not help with the tests in the IDE. Worth taking a look though :)

pjcollins:
I hadn't tried that CLI parameter -- that's definitely something consider if/when we look to updated these libraries to netstandard in the future.

PR #4146, opened 2020-01-20
Use the Framework MSBuild on Mac/Linux

dellis1972:
Commit dc49cf92 switched over our build system to use the MSBuild Nuget Packages. 
However when we use that and run the Unit Tests from within VSForMac they all fail with a
`PlatformNotSupported` exception. This is despite the fact that we never actually copy the files over to the output path, but use the Framework ones. 

It turns out the MSBuild Nuget File are loaded by the IDE and the NUnit Test Addin so it tries to use those when running the tests. This then results in the error. This issue does not occur on the command line or in VSCode since they do not load the MSBuild Nuget assemblies into the process.


pjcollins:
I believe this will cause issues when our build tasks project is updated to target `netstandard2.0` via https://github.com/xamarin/xamarin-android/pull/4132. I'll test this tomorrow, let's try to hold off on merging for now if possible.

Issue #4068, opened 2019-12-21
VS 16.5.0  P1.0 How to avoid - The "CopyGeneratedJavaResourceClasses" task was not given a value for the required parameter "SourceTopDirectory".


gmck:
Virtually any code change results in the above message when next trying to build. Cleaning before building fixes it. The app is an Android 10 - AndroidX app if that makes any difference.

moljac:
@gmck 

OK. I can reproduce this error with my AndroidX test samples.

Workaround for now: `rebuild` and then `run`. 

Sole `run` will cause this error. 

Tests continued.

moljac:
Binlog 

`Run` after code change.

[SampleApp.Android.csproj_Evaluation_2019-12-24T10_34_55.7709154+01_00.binlog.zip](https://github.com/xamarin/xamarin-android/files/3997878/SampleApp.Android.csproj_Evaluation_2019-12-24T10_34_55.7709154%2B01_00.binlog.zip)

moljac:
Repro sample

https://github.com/moljac/MovesenseDotNet/tree/master_based_improvements

https://github.com/moljac/MovesenseDotNet/tree/master_based_improvements/Samples/GraphPlotSample

moljac:
After getting error pressing `run` (`install`/`deploy`) again will deploy the new app (with code change).

lafritay:
I'm hitting this in 16.4.2.  Rebuild / clean and then redeploy does not fix it.  Any ideas on what else to try in order to troubleshoot?

dellis1972:
@moljac I am having trouble reproducing this issue on the latest preview version of VS 2019 with your sample app. 
Also would you be able to capture the binlog files again? The file in the zip that you linked did not seen to include any information at all.

jonathanpeppers:
@mattleibow this project is using AndroidX.Migration, and I am guessing that this target is not running for some reason on some incremental builds:

https://github.com/xamarin/xamarin-android/blob/afb389cbb3ea59d4254038e0f739052e4d374c40/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L2230-L2237

I suspect AndroidX.Migration is affecting the tree of MSBuild target dependencies? Maybe the `$(_CreateBaseApkDependsOnTargets)` property is overwritten?

I'm not able to build the linked sample, the NuGets do not restore for me.

lafritay:
[...]

PR #4151, opened 2020-01-21
[Xamarin.Android.Build.Tasks] reorder two targets to fix AndroidX.Migration

jonathanpeppers:
Context: https://github.com/xamarin/XamarinAndroidXMigration
Fixes: https://github.com/xamarin/xamarin-android/issues/4068

Developers have been reporting an error when using AndroidX.Migration:

    The "CopyGeneratedJavaResourceClasses" task was not given a value for the required parameter "SourceTopDirectory".

It is reported another build fixes the issue, and it "randomly"
happens on incremental builds.

The error seems to indicate either `$(AaptTemporaryDirectory)` or
`$(ResgenTemporaryDirectory)` are blank? But it only happens when
using the AndroidX.Migration NuGet package.

From a user's build log:

    Skipping target "_PrepareCreateBaseApk" because all output files are up-to-date with respect to the input files.

Which would indicate in the case of AndroidX.Migration:

* `_PrepareCreateBaseApk` was skipped.
* `_AndroidXJetifyManifest` was injected, it came along and modified
  `$(IntermediateOutputPath)android\AndroidManifest.xml`.
* `_CreateBaseApk` ran (with updated inputs), and
  `$(AaptTemporaryDirectory)` is blank!

I could reproduce the problem in a test:

* Build a project with AndroidX.Migration
* Change C# code (a non-Java.Lang.Object class)
* Build the project again

Reviewing the MSBuild targets: `_PrepareCreateBaseApk` and
`_CreateBaseApk`. They define a property that is passed between two
targets. The following sets of targets have nearly identical
`Input`/`Output`:

* `_PrepareCreateBaseApk` and `_CreateBaseApk`
* `_PrepareUpdateAndroidResgen` and `_UpdateAndroidResgen`

This was done to workaround a long-lived bug in MSBuild:

https://github.com/Microsoft/msbuild/issues/1006

A fix that doesn't seem break anything is to just run
`_PrepareCreateBaseApk` *after* `$(AfterGenerateAndroidManifest)`. So
if something triggers `_CreateBaseApk`, `_PrepareCreateBaseApk` will
always run as well.

I wrote a test for this scenario, and updated the AndroidX NuGets in
our tests to use the RC now.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/KnownPackages.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/KnownPackages.cs
@@ -223,51 +223,61 @@ namespace Xamarin.ProjectTools
 		};
 		public static Package AndroidXMigration = new Package {
 			Id = "Xamarin.AndroidX.Migration",
-			Version = "1.0.0-preview04",
+			Version = "1.0.0-rc1",
 			TargetFramework = "MonoAndroid10",
 		};
 		public static Package AndroidXAppCompat = new Package {
 			Id = "Xamarin.AndroidX.AppCompat",
-			Version = "1.0.2-preview02",
+			Version = "1.1.0-rc1",
 			TargetFramework = "MonoAndroid10",
 		};
 		public static Package AndroidXBrowser = new Package {
 			Id = "Xamarin.AndroidX.Browser",
-			Version = "1.0.0-preview02",
+			Version = "1.0.0-rc1",
 			TargetFramework = "MonoAndroid90",
 			References =  {
 				new BuildItem.Reference ("Xamarin.AndroidX.Browser") {
-					MetadataValues = "HintPath=..\\packages\\Xamarin.AndroidX.Browser.1.0.0-preview02\\lib\\MonoAndroid90\\Xamarin.AndroidX.Browser.dll"
+					MetadataValues = "HintPath=..\\packages\\Xamarin.AndroidX.Browser.1.0.0-rc1\\lib\\MonoAndroid90\\Xamarin.AndroidX.Browser.dll"
 				},
 			}
 		};
 		public static Package AndroidXMediaRouter = new Package {
 			Id = "Xamarin.AndroidX.MediaRouter",
-			Version = "1.0.0-preview02",
+			Version = "1.1.0-rc1",
 			TargetFramework = "MonoAndroid90",
 			References =  {
 				new BuildItem.Reference ("Xamarin.AndroidX.MediaRouter") {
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1562,7 +1540,24 @@ because xbuild doesn't support framework reference assemblies.
...
+  <!-- Create a dummy manifest file for aapt to work with -->
+  <CreateResgenManifest
+      ManifestOutputFile="$(ResgenTemporaryDirectory)\AndroidManifest.xml"
+      PackageName="$(_AndroidPackage)"
+  />
+
+  <GetExtraPackages
+      IntermediateOutputPath="$(IntermediateOutputPath)"
+      LibraryProjectImportsDirectoryName="$(_LibraryProjectImportsDirectoryName)">
+    <Output TaskParameter="ExtraPackages" PropertyName="AaptExtraPackages" />
```
dellis1972: I seem to remember these we split out because of the `CallTarget` calls. They cannot make use of properties or ItemGroups which are created within the calling target. I seem to remember they needed to be calculated outside of the target which called `CallTarget`.

jonathanpeppers: Yeah this broke 🤦‍♂ 

It is weird that AndroidX projects were building somehow...

jonathanpeppers: Ok, I was able to fix it by just reordering the targets.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/KnownPackages.cs
```diff
@@ -253,17 +253,27 @@ public static class KnownPackages
 		};
 		public static Package AndroidXLegacySupportV4 = new Package {
 			Id = "Xamarin.AndroidX.Legacy.Support.V4",
-			Version = "1.0.0-preview02",
+			Version = "1.0.0-rc1",
 			TargetFramework = "MonoAndroid90",
 			References =  {
 				new BuildItem.Reference ("Xamarin.AndroidX.Legacy.Support.V4") {
 					MetadataValues = "HintPath=..\\packages\\Xamarin.AndroidX.Legacy.Support.V4.1.0.0-preview02\\lib\\MonoAndroid90\\Xamarin.AndroidX.Legacy.Support.V4.dll"
```
mattleibow: Does it matter if this path is different in the version number?

jonathanpeppers: I should fix it, thanks.

Those will only be used if a test was doing `packages.config`, and none are doing that with AndroidX.

mattleibow:
Sorry about not reviewing. I though I hit the submit button. Looked good to me.

PR #238, opened 2016-09-27
[F#] Use C# Resource.designer.cs file for F# [C10 MERGE ONLY]

nosami:
Removes the use of F# CodeDOM for generating the resource
designer file for F#.

Instead, a [type provider](https://github.com/xamarin/Xamarin.Android.FSharp.ResourceProvider/blob/master/ResourceTypeProvider.fs)
is used at compile time (both xbuild and the IDE) that compiles the
C# code using CSharpCodeProvider and makes the types available to
the consuming F# project. The type provider is available as a NuGet
package and is included in the Android templates.

This is similar to #159, but uses an F# type provider to compile the C# instead of an msbuild task. The advantage of this method is that intellisense works in the IDE without having to change the IDE code.

PR #159, opened 2016-08-12
[Xamarin.Android.Build.Tasks] Fixed F# Resource Designer compilation issues [DONT MERGE]

dellis1972:
The latest release of F# removed support for static fields. As a result
the Resource.Designer.fs file that wes being generated by our build
system no longer compiled.

This commit reworks the Designer.cs code to emit a C# assembly for
F# projects rather than just the code. This has the benifit of
us not having to produce F# code in the first place.

The new system works as follows.

1) GenerateResourceDesigner Task generates a Resources.Designer.xx file as normal
2) If the target project is F# that file is then compiled into a Resource.Designer.dll
3) The new .dll is then included in the @ReferencePaths ItemGroup
4) The Resource.Designer.xx is removed from the @Complie ItemGroup

If the project is C# or VS the Designer file is just included as normal.

PR #4168, opened 2020-01-23
[F#] Use C# Resource.designer.cs file for F#

pjcollins:
Context: https://github.com/xamarin/xamarin-android/pull/238
Context: https://github.com/xamarin/xamarin-android/pull/4132

Removes the use of F# CodeDOM for generating the resource
designer file for F#.

Instead, a type provider
is used at compile time (both xbuild and the IDE) that compiles the
C# code using CSharpCodeProvider and makes the types available to
the consuming F# project. The type provider is available as a NuGet
package and is included in the Android templates.

This is similar to #159, but uses an F# type provider to compile the C# instead of an msbuild task. The advantage of this method is that intellisense works in the IDE without having to change the IDE code.

### DIFF
--- a/build-tools/xaprepare/xaprepare/ThirdPartyNotices/Xamarin.Android.Build.Tasks.cs
+++ b/build-tools/xaprepare/xaprepare/ThirdPartyNotices/Xamarin.Android.Build.Tasks.cs
@@ -40,7 +40,6 @@ namespace Xamarin.Android.Prepare
 
 		public override List<ThirdPartyNotice> Notices => new List <ThirdPartyNotice> {
 			new XamarinAndroidBuildTasks_fsharp_fsharp_TPN (),
-			new XamarinAndroidBuildTasks_fsprojects_FSharpCompilerCodeDom_TPN (),
 			new XamarinAndroidBuildTasks_IronyProject_Irony_TPN (),
 			new XamarinAndroidBuildTasks_JamesNK_NewtonsoftJson_TPN (),
 			new XamarinAndroidBuildTasks_NuGet_NuGetClient_TPN (),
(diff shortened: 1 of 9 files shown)


build-tools/xaprepare/xaprepare/ThirdPartyNotices/Xamarin.Android.Build.Tasks.cs
```diff
@@ -58,42 +57,6 @@ class XamarinAndroidBuildTasks_fsharp_fsharp_TPN : ThirdPartyNotice
...
-        MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
-        IN NO EVENT SHALL THE AUTHORS BE LIABLE FOR ANY CLAIM, DAMAGES OR
-        OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
-        ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
-        OTHER DEALINGS IN THE SOFTWARE.
-
-        For more information, please refer to <http://unlicense.org>
-";
-	}
-
```
jonathanpeppers: If you build this branch locally, there should be a `ThirdPartyNotices.txt` we commit, too. Sometimes we forget to update it.

dellis1972:
@pjcollins I don't see where we add the Nuget for the TypeProvider for our unit tests?

pjcollins:
>  I don't see where we add the Nuget for the TypeProvider for our unit tests?

This still needs a bit of work. It seems we have at least one test failing for the reason you've mentioned above. We also may want to consider introducing a new error message for much older F# projects which aren't using this NuGet, though I'm not sure if this will be worth the effort or not.

build-tools/xaprepare/xaprepare/ThirdPartyNotices/Xamarin.Android.Build.Tasks.cs
pjcollins: @jonathanpeppers is this in https://github.com/xamarin/xamarin-android/blob/master/ThirdPartyNotices.txt? It looks like this file currently doesn't have any mention of `FSharp.Compiler.CodeDom`...

My local build didn't show any diffs in this file, maybe I missed a step? Or maybe this content was never included? (cc @grendello).

jonathanpeppers: Maybe some of these notices aren't making it to the file on disk? 👀

pjcollins: I added an issue to track this https://github.com/xamarin/xamarin-android/issues/4179. I am not sure if we need to address that audit as part of this PR.

nosami:
> We also may want to consider introducing a new error message for much older F# projects which aren't using this NuGet, though I'm not sure if this will be worth the effort or not. I believe our templates have been using this NuGet for 3+ years now.

I don't think you need to worry about this. Most F# Android projects right now are using the Fabulous XF abstraction which is using the type provider already.

PR #4172, opened 2020-01-24
[ci] Add build success dependency to regression tests

pjcollins:
Updates the regression test stage introduced in commit 47e00d7c to only
run when the `mac_build` stage completes successfully, as it depends
on artifacts produced by that stage.


jpobst:
I noticed this problem earlier.  Since I am changing the `condition` anyways in this PR #4161, how about I add it there?

https://github.com/xamarin/xamarin-android/pull/4161/files#diff-9268c11f675853b241d743a346444d86L987

pjcollins:
Ahh yep sounds good 👍 I'll close this.

PR #4175, opened 2020-01-25
[Xamarin.Android.Build.Tasks] Move Generator.cs error strings into .resx file

brendanzagaeski:
Context: [0342fe56][0]

Move the error strings for `Generator.cs` into the `.resx` file.

Other changes:

Change "JavaTypeInfo" to "Java type" in the messages because there is no
such class as `JavaTypeInfo`, so preserving that name across the
translations would not be meaningful.

Correct a "The the" typo in the comment for one of the existing
resources in the `.resx` file.


Documentation/guides/messages/README.md
```diff
@@ -100,7 +100,7 @@ ms.date: 08/23/2019
 + [XA5205](xa5205.md): Cannot find `{ToolName}` in the Android SDK.
 + [XA5207](xa5207.md): Could not find android.jar for API Level `{compileSdk}`.
 + [XA5300](xa5300.md): The Android/Java SDK Directory could not be found.
-+ [XA5301](xa5301.md): Failed to create JavaTypeInfo for class: {t.FullName} due to MAX_PATH: {ex}
++ [XA5301](xa5301.md): Failed to create Java type for class: {t.FullName} due to MAX_PATH: {ex}
```
jonpryor: I wonder if "obtain" would be a better verb than "create" here.

brendanzagaeski:
Yuck. I didn't regenerate the `.xlf` files after the wording change. Force-pushing a corrected commit in a moment.

brendanzagaeski:

Rebase on master.

Based on a quick chat with Jon on the comment about wording, change "create" to "generate" in the messages to match the name of the failing method and hopefully reduce potential confusion with *instantiating* Java types. Other wording recommendations are welcome if for example "obtain" or another word looks like it would work better in context.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -135,6 +135,15 @@
   <data name="XA3005" xml:space="preserve">
     <value>Could not link native shared library: {0}</value>
   </data>
+  <data name="XA4209" xml:space="preserve">
+    <value>Failed to generate Java type for class: {0} due to {1}</value>
+    <comment>{0} - The managed type name
+{1} - The exception message of the associated exception</comment>
+  </data>
+  <data name="XA4210" xml:space="preserve">
+    <value>You need to add a reference to Mono.Android.Export.dll when you use ExportAttribute or ExportFieldAttribute.</value>
```
jonpryor: Should we also make these messages more polite while we're at it?  "You need to add..." sounds kinda rude; perhaps this should be "Please add..."?  Additionally, "you" should probably be removed.  Thus:

```xml
    <value>Please add a reference to Mono.Android.Export.dll when using ExportAttribute or ExportFieldAttribute.</value>
```

brendanzagaeski: Yeah, that sounds good. I'll add that.

I made a similar change to remove an exclamation point from another message to make it gentler. (That'll be coming up in another PR.)

PR #4176, opened 2020-01-25
[Xamarin.Android.Build.Tasks] Split up XA3001 errors and move them into .resx file

brendanzagaeski:
Context: [0342fe56][0]

Split up the different uses of XA3001 by giving each error condition its
own code.  This makes it possible to gather telemetry for each condition
separately, and it also makes it easier for users to search for the
particular code they are seeing.

Move XA3001 and the other new error codes to the `.resx` file so that
they are localizable.

Remove the `try-catch` from `CilStrip.RunTask()` because after
[d4c8f077][1], the `AndroidTask` base class now provides that
functionality.

[1]: https://github.com/xamarin/xamarin-android/commit/d4c8f077faefdfc6174355848a9d8c74ecaa8f56

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/CilStrip.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/CilStrip.cs
@@ -31,12 +31,7 @@ namespace Xamarin.Android.Tasks
 
 		public override bool RunTask ()
 		{
-			try {
-				return DoExecute ();
-			} catch (Exception e) {
-				Log.LogCodedError ("XA3001", "{0}", e);
-				return false;
-			}
+			return DoExecute ();
 		}
 
 		bool DoExecute () {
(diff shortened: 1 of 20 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/CilStrip.cs
```diff
@@ -31,12 +31,7 @@ public CilStrip ()
 
 		public override bool RunTask ()
 		{
-			try {
-				return DoExecute ();
-			} catch (Exception e) {
-				Log.LogCodedError ("XA3001", "{0}", e);
-				return false;
-			}
```
dellis1972: Should we remove this? What happens if we get an error which we are not expecting? This would then bubble up to MSBuild and give a horrible MSxxxx error. Should we keep a generic error here which asked users to report the issue ?

jonathanpeppers: Thanks to Pobst, it should go there here now: https://github.com/xamarin/xamarin-android/blob/1bee4ad176863bdfc0de258542a2c3685fc60f2d/src/Xamarin.Android.Build.Tasks/Tasks/AndroidTask.cs#L16-L21

I think using the default "let it throw" behavior is better now as long as this subclasses `AndroidTask`.

PR #4185, opened 2020-01-28
[xaprepare] Update Android 10 name (#4011)

jonpryor:
Context: https://www.blog.google/products/android/evolving-android-brand/

The `AndroidApiInfo.xml` files produced by `xaprepare` during build
preparation includes a *Name* value:

        <AndroidApiInfo>
          <Id>29</Id>
          <Level>29</Level>
          <Name>Q</Name>
          <Version>v10.0</Version>
          <Stable>True</Stable>
        </AndroidApiInfo>

Both the version number and name are shown within various parts of
[the IDEs][0], e.g.

VS Minimum Android Version

        Android 9.0 (Pie)
        Android 8.1 (Oreo)

The problem is that starting with Android 10, Google is no longer
providing "nicknames" for Android releases.

What should Visual Studio do?

*Currently*, we still emit a Name of "Q" for API-29, thus VS shows:

        Android 10.0 (Q)

We could "clear out" the Name value, but unfortunately VS currently
doesn't special-case this in any way, and thus would show:

        Android 10.0 ()

Which looks "wrong".

Fixing this will require two changes:

 1. We (xamarin-android) will begin clearing out the Name for Android
    releases which lack a nickname, and

 2. The VS team will modify their Target Framework name generation
    logic such that the parenthesis and Name are only used when Name
    is a non-empty value.  This will allow API-29 to have a displayed
    name of `Android 10.0`.

Begin the process: begin emitting an empty Name field for API-29,
allowing the IDE team to begin implementing their side.

[0]: https://stackoverflow.com/questions/55945365/how-can-i-target-android-4-1-in-visual-studio-2019

### DIFF
--- a/Documentation/release-notes/4011.md
+++ b/Documentation/release-notes/4011.md
@@ -1,0 +1,6 @@
+### Issues fixed
+
+  * [GitHub PR 4011](https://github.com/xamarin/xamarin-android/pull/4011):
+    The old *Q* nickname from the Android 10 preview versions was still being
+    shown for Android 10 in the Visual Studio project property pages and the
+    Android SDK Manager.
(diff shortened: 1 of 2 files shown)


Ghostbird:
Maybe this PR isn't very relevant any more and can be abandoned? The non-nicknaming of Android 10 turned out to be a one-off. Subsequent Android versions have all had nicknames (Red Velvet Cake, Snow Cone, Tiramisu, Upside-down Cake)

jonpryor:
Closing as "moot" as this is dealt with in the IDEs.

PR #4186, opened 2020-01-28
Adding api-versions.xml based on the AndroidApiLevel

gugavaro:
We currently use api-versions.xml based out of platform-tools\api directory, however it seems this file is not the latest one and also does not contain the ApiLevelSince attribute.
We are now favoring the api-versions.xml out of platforms\android-$(AndroidApiLevel)\data and in case this file does not exist we fall back for the previous location.
On those changes also updating Microsoft.DotNet.ApiCompat to the latest version
Making some small changes on CheckApiCompatibility to better retry and output logs
Updating all acceptable-breakages files


build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks.csproj
```diff
@@ -62,8 +62,8 @@
     <Compile Include="Xamarin.Android.Tools.BootstrapTasks\JdkInfo.cs" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.DotNet.ApiCompat" Version="5.0.0-beta.19606.1" IncludeAssets="none" />
-    <PackageReference Include="Microsoft.DotNet.GenAPI" Version="5.0.0-beta.19606.1" />
+    <PackageReference Include="Microsoft.DotNet.ApiCompat" Version="5.0.0-beta.20078.1" IncludeAssets="none" />
+    <PackageReference Include="Microsoft.DotNet.GenAPI" Version="5.0.0-beta.20078.1" />
```
jonpryor: Any specific reason why to bump the version?

gugavaro: I talked to Santi the person responsible and he said they are making some perf improvements, so I decide to bump the version.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -142,44 +142,41 @@ void ValidateApiCompat (string contractPath, bool validateAgainstReference)
 					File.Copy (implementationAssembly, Path.Combine (targetImplementationPathDirectory, assemblyToValidate), true);
 				}
 
-				using (var genApiProcess = new Process ()) {
+				for (int i = 0; i < 3; i++) {
```
jonpryor: PR #4174 added support to re-execute `Microsoft.DotNet.ApiCompat.exe` if/when mono crashes.  What are these more recent changes for?  I don't see a rationale for these changes within the PR message.

gugavaro: I wanted to created a new Process object to re-try instead of using the one already created. For some reason I was getting some odd behavior when re-running the process from the same Process object after the crash. When I was reading online, the recommendation was to dispose the object and use a new one.

PR #4189, opened 2020-01-29
[Xamarin.Android.Build.Tasks] Move XA4301 and XA0107 message text into .resx file

brendanzagaeski:
Context: [0342fe56][0]

Move the text for the XA4301 and XA0107 messages into the `.resx` file.

Other changes:

Standardize the capitalization of "APK" and "ABI" in the messages.

Replace an exclamation point with a period in one of the XA0107 messages
to make the message a little gentler.

Replace a comma with a period in one of the XA4301 messages.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -117,6 +117,13 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="XA0107" xml:space="preserve">
+    <value>{0} is a Reference Assembly.</value>
+  </data>
+  <data name="XA0107_Ignoring" xml:space="preserve">
+    <value>Ignoring Reference Assembly `{0}`.</value>
+    <comment>{0} - File name of the assembly.</comment>
+  </data>
   <data name="XA1005" xml:space="preserve">
     <value>Attempting naive type name fixup for element with ID '{0}' and type '{1}'</value>
     <comment>{0} - The Android resource ID name</comment>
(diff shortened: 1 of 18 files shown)


src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -117,6 +117,13 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="XA0107" xml:space="preserve">
+    <value>{0} is a Reference Assembly.</value>
+  </data>
+  <data name="XA0107_Ignoring" xml:space="preserve">
+    <value>Ignoring Reference Assembly `{0}`.</value>
+    <comment>{0} - File name of the assembly.</comment>
+  </data>
```
brendanzagaeski: I debated on whether to split XA0107 into two different codes for the two different messages. For the moment, I left them both under one code. In contrast to XA3001, which I split into separate codes as part of https://github.com/xamarin/xamarin-android/commit/1524e6e24780de610faa299065285ef63b790fef, the two messages for XA0107 have the same underlying cause. The reason there are two messages for XA0107 is instead that the messages are emitted from different places. The meaning of the warning differs slightly in those different places.

If telemetry some day indicates that XA0107 is appearing frequently in user projects and the team would like to know which of the code paths is most frequently causing the warning, new message codes can be added at that time. (Note that `XA0107_Ignoring` is currently used in two different places, so having separate codes for each of the two messages wouldn't actually be enough to distinguish all of the different code paths anyway.)

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -161,4 +168,18 @@
...
+  <data name="XA4301_ABI" xml:space="preserve">
+    <value>Cannot determine ABI of native library {0}.</value>
+    <comment>The abbreviation "ABI" should not be translated.</comment>
+  </data>
+  <data name="XA4301_ABI_Ignoring" xml:space="preserve">
+    <value>Could not determine ABI of some native libraries. Ignoring those: {0}</value>
+    <comment>The abbreviation "ABI" should not be translated.
+{0} - Comma-separated list of the ignored library file names.</comment>
+  </data>
```
brendanzagaeski: Similar to the XA0107 messages, I debated on splitting these XA4301 into different error codes, and for similar reasons, I left them all under one code for now.

Although the wording of the first `XA4301` message doesn't directly mention it, that message is about native libraries too, like the other messages.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
jonpryor: Further note that the two "different yet related" messages exist so that things "look reasonable" within Visual Studio, as multi-line messages get rendered in the Warning/Error panel "weirdly".

brendanzagaeski:
> should we be consistent with google docs regarding Apk vs APK and Abi vs abi vs ABI ?

Yeah, I got in the habit of using the capitalizations from developer.android.com for the release notes and when syncing some other docs.

For the MSBuild messages, another option would be standardize on something different, as long each individual term was consistent across messages to keep things sensible for the translators. But I figure we might as well use the same capitalizations from developer.android.com in the MSBuild messages too.

brendanzagaeski:

Rebase on master.

brendanzagaeski:

Rebase on master one more time to fix new conflicts.

brendanzagaeski:
**Changes in latest commit:**

Add back the changes to `BuildTest.cs` that I accidentally lost when rebasing.

PR #4190, opened 2020-01-29
[Xamarin.Android.Build.Tasks] Run `aapt2 compile` incrementally.

dellis1972:
The `aapt2 compile` command runs in two modes. The one we
currently use is the `archive` mode. We calling `aapt2 compile`
with the `--dir` argument you end up generating one `flata`
archive for all the files. The side effect of this is that even
if you only change one file, it will need to regenerate the
entire `flata` archive.

But it has a second mode. Rather than using `--dir` you just send
in a single file. This then writes a single `.flat` file to the
output directory. While this does mean you have to call `aapt2 compile`
for EVERY file, it does mean we can leverage MSbuilds support for
partial targets. This means MSbuild will detect ONLY the files which
changed and allow us to call `aapt2 compile` with just THOSE files.

While this may impact on initial build times, the goal is to make
incremental builds quicker. This is especially true for users to
use ALLOT of `AndroidResource` items.

A note regarding the `aapt2 daemon` mode. In order to write
accented characters we need to set the `StandardInput`
encoding to UTF8. This is not possible directly
in netstandard 2.0. So we have to use `Console.InputEncoding`
instead. Also not that we MUST not include a BOM when writting
the commands. This is because `aapt2` will try to parse the BOM
as command characters.

Also the `aapt2 link` command sometimes reports it is "Done" before
it has even written the archive for the file. So we can get into a
position where we think we are done but the file is not on disk.
So we have had to include a nasty wait which will poll for the
existence of the expected output file and only return when it
exists. The good news is we know at this point if the command
failed or not, so we can bypass the check on failure.

| Test         | Latest Stable | Aapt2 Daemon |
|--------------|---------------|--------------|
| From Clean   | 00:00:27.92   | 00:00:27.41  |
| No Changes   | 00:00:03.14   | 00:00:03.72  |
| Alter Layout | 00:00:16.37   | 00:00:05.53  |

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Aapt2.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Aapt2.targets
@@ -17,6 +17,30 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+<PropertyGroup>
+  <Aapt2DaemonMaxInstanceCount Condition=" '$(Aapt2DaemonMaxInstanceCount)' == '' " >0</Aapt2DaemonMaxInstanceCount>
+  <_Aapt2DaemonKeepInDomain Condition=" '$(_Aapt2DaemonKeepInDomain)' == '' ">false</_Aapt2DaemonKeepInDomain>
+</PropertyGroup>
+
+
+<Target Name="_InjectAapt2Dependencies" AfterTargets="_ResolveSdks" Condition=" '$(_AndroidUseAapt2)' == 'True' ">
+  <PropertyGroup>
+    <_SetLatestTargetFrameworkVersionDependsOnTargets>
+      $(_SetLatestTargetFrameworkVersionDependsOnTargets);
+      _CreateAapt2VersionCache;
+    </_SetLatestTargetFrameworkVersionDependsOnTargets>
+    <_PrepareUpdateAndroidResgenDependsOnTargets>
+      _CompileResources;
+      _Aapt2UpdateAndroidResgenInputs;
+      $(_PrepareUpdateAndroidResgenDependsOnTargets);
+    </_PrepareUpdateAndroidResgenDependsOnTargets>
+    <_AfterConvertCustomView>
+      $(_AfterConvertCustomView);
+      _FixupCustomViewsForAapt2;
+    </_AfterConvertCustomView>
+  </PropertyGroup>
+</Target>
+
 <Target Name="_ReadAapt2VersionCache">
   <ReadLinesFromFile File="$(_AndroidAapt2VersionFile)"
       Condition="Exists('$(_AndroidAapt2VersionFile)')">
@@ -37,7 +61,9 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
   />
   <ItemGroup Condition="'$(_Aapt2Version)' != '@(_Aapt2VersionCache)'">
     <_CompiledFlataArchive Include="$(_AndroidLibrayProjectIntermediatePath)**\*.flata" />
-    <_CompiledFlataArchive Include="$(IntermediateOutputPath)\*.flata" />
+    <_CompiledFlataArchive Include="$(_AndroidLibrayProjectIntermediatePath)**\*.flat" />
+    <_CompiledFlataArchive Include="$(_AndroidLibraryFlatFilesDirectory)*.flat" />
+    <_CompiledFlataArchive Include="$(_AndroidLibraryFlatArchivesDirectory)\*.flata" />
     <_CompiledFlataStamp Include="$(_AndroidLibrayProjectIntermediatePath)**\compiled.stamp" />
   </ItemGroup>
   <Delete
@@ -54,68 +80,19 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
       LibraryProjectIntermediatePath="$(_AndroidLibrayProjectIntermediatePath)"
       StampDirectory="$(_AndroidStampDirectory)">
     <Output TaskParameter="Output" ItemName="_LibraryResourceDirectories" />
+    <Output TaskParameter="LibraryResourceFiles" ItemName="_LibraryResourceFiles" />
+    <Output TaskParameter="LibraryResourceFiles" ItemName="_CompileResourcesInputs" />
   </CollectNonEmptyDirectories>
-  <ComputeHash Source="@(_LibraryResourceDirectories)"  >
-    <Output TaskParameter="Output" ItemName="_LibraryResourceHashDirectories" />
-  </ComputeHash>
-</Target>
-
-<Target Name="_ConvertLibraryResourcesCases"
-    Condition=" '$(_AndroidUseAapt2)' == 'True' "
-    DependsOnTargets="_CollectLibraryResourceDirectories"
-    Inputs="@(_LibraryResourceHashDirectories->'%(StampFile)')"
-    Outputs="$(_AndroidStampDirectory)_ConvertLibraryResourcesCases.stamp">
-  <ConvertResourcesCases
-      Condition=" '@(_LibraryResourceDirectories)' != '' "
-      ContinueOnError="$(DesignTimeBuild)"
-      AcwMapFile="$(_AcwMapFile)"
-      AndroidConversionFlagFile="$(_AndroidStampDirectory)_ConvertLibraryResourcesCases.stamp"
-      CustomViewMapFile="$(_CustomViewMapFile)"
-      ResourceDirectories="@(_LibraryResourceDirectories)"
-      ResourceNameCaseMap="$(_AndroidResourceNameCaseMap)"
-  />
-  <Touch Files="$(_AndroidStampDirectory)_ConvertLibraryResourcesCases.stamp" AlwaysCreate="True" />
-</Target>
-
-<Target Name="_CompileAndroidLibraryResources"
-    Condition=" '$(_AndroidUseAapt2)' == 'True' "
-    DependsOnTargets="_ConvertLibraryResourcesCases"
-    Inputs="@(_LibraryResourceHashDirectories->'%(StampFile)')"
-    Outputs="@(_LibraryResourceHashDirectories->'$(_AndroidLibraryFlatArchivesDirectory)%(Hash).stamp')"
-  >
-  <MakeDir Directories="$(_AndroidLibraryFlatArchivesDirectory)" Condition="!Exists('$(_AndroidLibraryFlatArchivesDirectory)')" />
-  <Aapt2Compile
-      Condition=" '@(_LibraryResourceHashDirectories)' != '' "
-      ContinueOnError="$(DesignTimeBuild)"
-      ExtraArgs="$(AndroidAapt2CompileExtraArgs)"
-      FlatArchivesDirectory="$(_AndroidLibraryFlatArchivesDirectory)"
-      ResourceDirectories="@(_LibraryResourceHashDirectories)"
-      ToolPath="$(Aapt2ToolPath)"
-      ToolExe="$(Aapt2ToolExe)"
-  />
-  <ItemGroup>
-    <_MissingStampFiles Include="@(_LibraryResourceHashDirectories->'%(StampFile)')" Condition="!Exists('%(StampFile)')" />
-    <_HashStampFiles Include="@(_LibraryResourceHashDirectories->'$(_AndroidLibraryFlatArchivesDirectory)%(Hash).stamp')" />
-    <_HashFlataFiles Include="@(_LibraryResourceHashDirectories->'$(_AndroidLibraryFlatArchivesDirectory)%(Hash).flata')" />
-  </ItemGroup>
-  <Touch
-      Files="@(_MissingStampFiles);@(_HashStampFiles)"
-      AlwaysCreate="True"
-  />
-  <ItemGroup>
-    <FileWrites Include="@(_MissingStampFiles)" />
-    <FileWrites Include="@(_HashStampFiles)" />
-    <FileWrites Include="@(_HashFlataFiles)" />
-  </ItemGroup>
 </Target>
 
 <Target Name="_ConvertResourcesCases"
     Condition=" '$(_AndroidUseAapt2)' == 'True' "
-    Inputs="$(MSBuildAllProjects);$(_AndroidBuildPropertiesCache);@(AndroidResource)"
+    Inputs="$(MSBuildAllProjects);$(_AndroidBuildPropertiesCache);@(AndroidResource);@(_LibraryResourceDirectories->'%(StampFile)')"
     Outputs="$(_AndroidStampDirectory)_ConvertResourcesCases.stamp"
-    DependsOnTargets="$(_BeforeConvertResourcesCases)"
+    DependsOnTargets="_CollectLibraryResourceDirectories;$(_BeforeConvertResourcesCases)"
   >
   <MakeDir Directories="$(_AndroidLibraryFlatArchivesDirectory)" Condition="!Exists('$(_AndroidLibraryFlatArchivesDirectory)')" />
+  <MakeDir Directories="$(_AndroidLibraryFlatFilesDirectory)" Condition="!Exists('$(_AndroidLibraryFlatFilesDirectory)')" />
   <!-- Change cases so we support mixed case resource names -->
   <ConvertResourcesCases
       ContinueOnError="$(DesignTimeBuild)"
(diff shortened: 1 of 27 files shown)


dellis1972:
@jonathanpeppers should we make this feature optional? So users can turn it on/off. Also should we restrict this only to app projects?

dellis1972:
> I think it would also be good to write a test checking if the target builds partially:

Yup, this is till a work in progress so I'll get to that too :)

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
```diff
@@ -148,8 +148,16 @@ string GenerateCommandLineCommands (string ManifestFile, string currentAbi, stri
 			}
 
 			if (CompiledResourceFlatArchive != null) {
-				var flata = Path.Combine (WorkingDirectory, CompiledResourceFlatArchive.ItemSpec);
-				if (File.Exists (flata)) {
+				var flata = CompiledResourceFlatArchive.ItemSpec;//Path.Combine (WorkingDirectory, CompiledResourceFlatArchive.ItemSpec);
```
jonathanpeppers: The space issue is tough. What if `$(Configuration)` has a space in it? Like `Configuration=Google Play`. I've seen devs do it before. 🤦‍♂

dellis1972: Indeed those tests are failing :) So I'm working on something else :P

dellis1972:
Windows tests are failing with 
```
 C:\A\xamarin-v00000G-1\_work\1\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Aapt2.targets(151,3): warning APT2000: aapt2.exe F 02-06 17:05:23  7556  7648 Command.cpp:41] Check failed: android::base::UTF8PathToWindowsLongPath(arg.data(), &path16) Failed to convert file path to UTF-16: file path C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\obj\Release\flat\ [C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\UnnamedProject.csproj]
  C:\A\xamarin-v00000G-1\_work\1\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Aapt2.targets(151,3): warning APT2000: aapt2.exe F 02-06 17:05:23  1388  9012 Command.cpp:41] Check failed: android::base::UTF8PathToWindowsLongPath(arg.data(), &path16) Failed to convert file path to UTF-16: file path C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\obj\Release\flat\ [C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\UnnamedProject.csproj]
  C:\A\xamarin-v00000G-1\_work\1\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Aapt2.targets(151,3): warning APT2000: aapt2.exe F 02-06 17:05:24  7328 10176 Command.cpp:41] Check failed: android::base::UTF8PathToWindowsLongPath(arg.data(), &path16) Failed to convert file path to UTF-16: file path C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\obj\Release\flat\ [C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\UnnamedProject.csproj]
  C:\A\xamarin-v00000G-1\_work\1\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Aapt2.targets(151,3): warning APT2000: aapt2.exe F 02-06 17:05:24  1272   504 Command.cpp:41] Check failed: android::base::UTF8PathToWindowsLongPath(arg.data(), &path16) Failed to convert file path to UTF-16: file path C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\obj\Release\flat\ [C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\UnnamedProject.csproj]
  C:\A\xamarin-v00000G-1\_work\1\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Aapt2.targets(151,3): warning APT2000: aapt2.exe F 02-06 17:05:25   912  5228 Command.cpp:41] Check failed: android::base::UTF8PathToWindowsLongPath(arg.data(), &path16) Failed to convert file path to UTF-16: file path C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\obj\Release\flat\ [C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\UnnamedProject.csproj]
  C:\A\xamarin-v00000G-1\_work\1\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Aapt2.targets(151,3): warning APT2000: aapt2.exe F 02-06 17:05:25  9752  5668 Command.cpp:41] Check failed: android::base::UTF8PathToWindowsLongPath(arg.data(), &path16) Failed to convert file path to UTF-16: file path C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\obj\Release\flat\ [C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\UnnamedProject.csproj]
  C:\A\xamarin-v00000G-1\_work\1\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\Xamarin.Android.Aapt2.targets(151,3): warning APT2000: aapt2.exe F 02-06 17:05:26  6912  8696 Command.cpp:41] Check failed: android::base::UTF8PathToWindowsLongPath(arg.data(), &path16) Failed to convert file path to UTF-16: file path C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\obj\Release\flat\ [C:\A\xamarin-v00000G-1\_work\1\s\bin\TestRelease\temp\BuildReleaseAppWithA InItAndÜmläüts(Truedxr8)\UnnamedProject.csproj]
```

so it looks like the daemon mode does not work with long paths on windows 🤦‍♂

src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
```diff
@@ -0,0 +1,146 @@
...
+using System.IO;
+using System.Text;
+using System.Threading;
+using Microsoft.Build.Framework;
+
+namespace Xamarin.Android.Tasks {
+	class Aapt2Daemon : IDisposable {
+		struct DaemonProcess {
+			public Process Process;
+			public object LockObject;
```
jonathanpeppers: I would make this:
```suggestion
			public readonly object LockObject = new object ();
```
So you don't have to worry about assigning it or overwriting it.

src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
```diff
@@ -0,0 +1,146 @@
...
+				UseShellExecute = false,
+				StandardErrorEncoding = Encoding.UTF8,
+				StandardOutputEncoding = Encoding.UTF8,
+				//StandardInputEncoding = Encoding.UTF8,
+			};
+			aapt2.LockObject = new object ();
+			aapt2.Process = new Process ();
+			aapt2.Process.StartInfo = info;
+			aapt2.Process.Start();
+			if (daemons.TryAdd (id, aapt2)) {
```
jonathanpeppers: Should this use [GetOrAdd](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=netframework-4.8)?

If this finds a key that is already present, it wouldn't even call aapt2. Is that correct or not?

src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
```diff
@@ -0,0 +1,146 @@
...
+		bool Run (DaemonProcess aapt2, IList<OutputLine> output, params string[] args) {
+			lock (aapt2.LockObject) {
+				StreamWriter writer = aapt2.Process.StandardInput;
+				Console.WriteLine ($"Writer = {writer.Encoding}");
+				foreach (var arg in args) {
+					writer.WriteLine (arg.Trim ());
+				}
+				writer.WriteLine ();
+				writer.Flush ();
+			}
```
jonathanpeppers: I think this lock might need to extend to the part reading from stderr, too?

Otherwise one thread might read another's standard error?

src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
dellis1972: The `DaemonProcess` is a struct, so we can't use that. I could make it a class. However its such a small object I'm not sure its worth it.

src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Aapt2.targets
```diff
@@ -130,37 +124,74 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
...
+      @(_CompiledFlatFiles);
+      @(_LibraryResourceDirectoryStamps);
+    </_CreateBaseApkInputs>
+  </PropertyGroup>
 </Target>
 
 <Target Name="_UpdateAndroidResgenAapt2"
-    Condition="'$(_AndroidUseAapt2)' == 'True'">
+    Condition="'$(_AndroidUseAapt2)' == 'True'"
+  >
```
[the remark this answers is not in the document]
dellis1972: I assume this is missing the spaces before and after the condition. Cos that is what I fixed :P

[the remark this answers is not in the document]
jonathanpeppers: I just thought the `>` on a new line with no other changes was odd.

dellis1972:
@jonathanpeppers  Performance Tests are failing :(

```
     1009 ms  IncrementalClean                           1 calls
     1847 ms  _CalculateResourceFileName                 1 calls
     2694 ms  _CleanGetCurrentAndPriorFileWrites         1 calls
...
...
     1131 ms  Aapt2CalculateCompiledResourceFileName     1 calls
     1306 ms  FindUnderPath                              5 calls
     1325 ms  RemoveDuplicates                           2 calls
```

So the tasks which calculate these files are taking allot longer on the build machines than they are locally (I/O bound ?). Its adding the 5 seconds to the build which is pushing it over the limit.

src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Aapt2.targets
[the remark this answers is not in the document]
dellis1972: @jonathanpeppers jonp usually likes the `>` on the new line so that is doesn't mess up diffs when we add properties later.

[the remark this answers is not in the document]
dellis1972: but I can change it if needed.

jonathanpeppers:
@dellis1972 the one that is most important is `Build_XAML_Change(True,False)`, it builds a Xamarin.Forms app with a XAML-change and uses `$(ProduceReferenceAssembly)`.

In that case it looks like this new target:
```
     1218 ms  Aapt2CalculateCompiledResourceFileName     1 calls
```
Is this running on every build and it shouldn't be?

The other thing I'm wondering, have we added a massive new number of files to `@(FileWrites)`?
```
     1223 ms  FindUnderPath                              5 calls
     1278 ms  RemoveDuplicates                           2 calls
```
Maybe it's OK to not add some of them to `@(FileWrites)`? In some cases I've found sub-directories are OK we just have to delete the files ourself:

1. If you put files in `obj\Debug\foo\**\*`
2. You just need to make sure you have a manual call to delete `obj\Debug\foo` on `Clean`.

I've found `IncrementalClean` only deletes things automatically in `obj\Debug\*` and doesn't search recursively.

dellis1972:
@jonathanpeppers well I would expect a massive number of new files because we are generating a `.flat` file for every single resource. This includes ALL the support libraries. They are all in various `flat` folders like `obj/Debug/flat` and `obj/Debug/lp/0/flat`.

`Aapt2CalculateCompiledResourceFileName` is a problem. We need it to calculate the filenames which we expect `aapt2` to generate (e.g layout_main_xml.flat`). This way we can build `_CompileResources` incrementally. I'll have to think of a way to cache that data between builds so that we don't have to calculate it every build.

dellis1972:
still failing on the Xaml performances tests :(


```
643 ms  _CalculateResourceFileName                 1 calls
396 ms  CollectNonEmptyDirectories                 1 calls
```

these two calls make up the roughly 1 second additional time we are spending in the build.
`Aapt2CalculateCompiledResourceFileName` task was removed and the work was placed in other tasks which were already dealing with the files in question. That got us closer, but we are still 1 second off. 

I need to figure out a way to speed this stuff up.  Enumerating all the files in the `lp` directories I suspect is causing much of the delay in `CollectNonEmptyDirectories`. However why it is taking 600+ ms to create a couple of `ItemGroup` items in `_CalculateResourceFileName` is concerning. 
These do contain a large number of files. I might try to NOT use a custom item group instead see if that helps.

jonathanpeppers:
@dellis1972 after looking at the latest build log:

image

But it's weird:
1. `CollectNonEmptyDirectories` everything looked like it was cached?
2. `_CalculateResourceFileName` took time and it's just item groups.
3. `_CompileResources` was skipped!

It makes me think we have to rework things, maybe if we:
1. Use the current system we have today for referenced libraries, NuGet packages, aar files, etc.
2. Only use the new `.flat` / incremental build features of `aapt2` for files _inside_ the current project.

When I look at the work these targets are doing, most of the files are inside `obj\Debug\lp`. So if we didn't do the incremental stuff for files inside there, would we be better off?

dellis1972:
The only problem with using `flata` for references projects is the case when an android library project is updated. If it has allot of resources we need to process all of them again, that is going to hurt build times. Its fine for Nuget /aar since they don't change very often. The issue then is do we mix and match... that will make things more complicated.

dellis1972:
@jonathanpeppers ok we are down to 1 failure

```
EXEC : error : aprofutil: Unable to read profile through local port: 9013.
```

its the profile AOT one.

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
```diff
@@ -212,33 +302,10 @@ string ExpandString (string s)
...
-					}
-					File.Delete (tmpfile);
-				}
-				// Delete the archive on failure
-				if (!aaptResult && File.Exists (currentResourceOutputFile)) {
-					File.Delete (currentResourceOutputFile);
-				}
-			}
-			return aaptResult;
+			apks.Add (currentResourceOutputFile, RunAapt (cmd, currentResourceOutputFile));
```
jonathanpeppers: Just like `tempFiles` does this need a lock around it? If you used the apk-per-abi feature?

src/Xamarin.Android.Build.Tasks/Tasks/CopyIfChanged.cs
```diff
@@ -24,11 +24,18 @@ public class CopyIfChanged : AndroidTask
 		[Required]
 		public ITaskItem[] DestinationFiles { get; set; }
 
+		public bool CompareFileLengths { get; set; }
```
jonathanpeppers:
```suggestion
		public bool CompareFileLengths { get; set; } = true;
```
Then you could remove that new ctor.

src/Xamarin.Android.Build.Tasks/Tasks/ConvertCustomView.cs
```diff
@@ -79,24 +79,23 @@ public override bool RunTask ()
 					}
 				}
 			}
-			var output = new List<ITaskItem> ();
+			var output = new Dictionary<string, ITaskItem> ();
 			foreach (var file in processed) {
 				ITaskItem resdir = ResourceDirectories?.FirstOrDefault (x => file.StartsWith (x.ItemSpec)) ?? null;
-				if (resdir == null) {
+				if (output.ContainsKey (file))
```
jonathanpeppers: If `processed` is a `HashSet<string>`:

https://github.com/xamarin/xamarin-android/blob/802f0d3af0a17c952673127b12f6058218262edd/src/Xamarin.Android.Build.Tasks/Tasks/ConvertCustomView.cs#L35

Do we need this `ContainsKey` check?

src/Xamarin.Android.Build.Tasks/Tasks/ConvertCustomView.cs
```diff
@@ -79,24 +79,23 @@ public override bool RunTask ()
 					}
 				}
 			}
-			var output = new List<ITaskItem> ();
+			var output = new Dictionary<string, ITaskItem> ();
```
jonathanpeppers: Minor, but could maybe be:
```suggestion
			var output = new Dictionary<string, ITaskItem> (processed.Count);
```

src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
```diff
@@ -0,0 +1,259 @@
...
+						while (stdError.Count > 0) {
+							line = stdError.Dequeue ();
+							job.Output.Add (new OutputLine (line, stdError: !IsAapt2Warning (line), errored: errored, jobId: job.JobId));
+						}
+						// wait for the file we expect to be created. There can be a delay between
+						// the daemon saying "Done" and the file finally being written to disk.
+						if (!string.IsNullOrEmpty (job.OutputFile) && !errored) {
+							while (!File.Exists (job.OutputFile)) {
+								Thread.Sleep (10);
+							}
```
jonathanpeppers: I wish I could think of a better thing to do here. 🤔

src/Xamarin.Android.Build.Tasks/Tasks/CopyIfChanged.cs
dellis1972: I tried that but it didn't work for some reason...

src/Xamarin.Android.Build.Tasks/Tasks/ConvertCustomView.cs
dellis1972: I think we do, because we want to skip the code underneath if we already added a file of the same name.

jonathanpeppers: I don't think `processed` can't have a duplicate if it is a `HashSet`.

If you do: `Add ("Foo")` and `Add ("Foo")` it will have one item.

dellis1972: yes, but there are a number of lines of code where we are calculating stuff before we add. If we do have a duplicate, lets at least NOT do those calculations again.

dellis1972: actually ignore me, I just realised what you were saying 🤦‍♂

brendanzagaeski:
*Draft release notes*

Here is the note I'm thinking of including for this pull request:

```
### Build and deployment performance

  * [GitHub PR 4190](https://github.com/xamarin/xamarin-android/pull/4190):
    Switch to running AAPT2 in single-file mode rather than `--dir` mode.  This
    reduced the total incremental build time from about 16.4 seconds to about 5.5
    seconds for a test project when a single Android resource file was changed
    between builds.
```

To suggest something different, feel free to reply in this PR or add a `Documentation/release-notes/4190.md` file with the new suggestion. Thanks in advance!

acuntex:
Is it possible to disable this new feature?

We suspect this feature to break the functionality of ResizetizerNT and since it's apparently impossible to downgrade Xamarin.Android after a Visual Studio for Mac upgrade we are not able to build the Android app.

dellis1972:
@acuntex its is not possible to disable this feature, and I can't see how it would effect ResizetizerNT. I would recommend you open an Issue in this repo https://github.com/xamarin/xamarin-android/issues/new/choose . Provide as much detail as possible including diagnostic build logs which show the error you are getting.

PR #4193, opened 2020-01-30
[Xamarin.Android.Build.Tasks] Move XA1006, XA1007, XA1008 message text into .resx file

brendanzagaeski:
Context: [0342fe56][0]

Move the text for the XA1006, XA1007, and XA1008 messages into the
`.resx` file.

Other changes:

Remove "you" and "your" from the messages.

Clarify that the number shown for `$(TargetFrameworkVersion)` is the
corresponding Android API level, not the `$(TargetFrameworkVersion)`
itself.

Make the messages for XA1006 and XA1008 more similar to each other.

### DIFF
--- a/Documentation/guides/messages/README.md
+++ b/Documentation/guides/messages/README.md
@@ -72,9 +72,9 @@ ms.date: 01/24/2020
 + [XA1003](xa1003.md): '{zip}' does not exist. Please rebuild the project.
 + [XA1004](xa1004.md): There was an error opening {filename}. The file is probably corrupt. Try deleting it and building again.
 + [XA1005](xa1005.md): Attempting naive type name fixup for element with ID '{id}' and type '{managedType}'
-+ [XA1006](xa1006.md): Your application is running on a version of Android ({compileSdk}) that is more recent than your targetSdkVersion specifies ({targetSdk}). Set your targetSdkVersion to the highest version of Android available to match your TargetFrameworkVersion ({compileSdk}).
++ [XA1006](xa1006.md): The TargetFrameworkVersion (Android API level {compileSdk}) is higher than the targetSdkVersion ({targetSdk}).
 + [XA1007](xa1007.md): The minSdkVersion ({minSdk}) is greater than targetSdkVersion. Please change the value such that minSdkVersion is less than or equal to targetSdkVersion ({targetSdk}).
-+ [XA1008](xa1008.md): The TargetFrameworkVersion ({compileSdk}) should not be lower than targetSdkVersion ({targetSdk})
++ [XA1008](xa1008.md): The TargetFrameworkVersion (Android API level {compileSdk}) is lower than the targetSdkVersion ({targetSdk}).
 + [XA1009](xa1009.md): The {assembly} is Obsolete. Please upgrade to {assembly} {version}
 
 ## XA2xxx: Linker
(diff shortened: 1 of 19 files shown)


Documentation/guides/messages/README.md
```diff
@@ -72,9 +72,9 @@ ms.date: 01/24/2020
 + [XA1003](xa1003.md): '{zip}' does not exist. Please rebuild the project.
 + [XA1004](xa1004.md): There was an error opening {filename}. The file is probably corrupt. Try deleting it and building again.
 + [XA1005](xa1005.md): Attempting naive type name fixup for element with ID '{id}' and type '{managedType}'
-+ [XA1006](xa1006.md): Your application is running on a version of Android ({compileSdk}) that is more recent than your targetSdkVersion specifies ({targetSdk}). Set your targetSdkVersion to the highest version of Android available to match your TargetFrameworkVersion ({compileSdk}).
++ [XA1006](xa1006.md): The TargetFrameworkVersion (Android API level {compileSdk}) is higher than the targetSdkVersion ({targetSdk}).
 + [XA1007](xa1007.md): The minSdkVersion ({minSdk}) is greater than targetSdkVersion. Please change the value such that minSdkVersion is less than or equal to targetSdkVersion ({targetSdk}).
-+ [XA1008](xa1008.md): The TargetFrameworkVersion ({compileSdk}) should not be lower than targetSdkVersion ({targetSdk})
++ [XA1008](xa1008.md): The TargetFrameworkVersion (Android API level {compileSdk}) is lower than the targetSdkVersion ({targetSdk}).
```
brendanzagaeski: If these wording changes look debatable, I can undo them. Part of the motivation was to make the translations easier by making it possible to reuse pieces of the messages.

jonpryor: The problem with the updates is that messages need to be *actionable*, and the update *removes the action*:

This is an action, something that the user can do to remove the message:

```
Set your targetSdkVersion to the highest version of Android available to match your TargetFrameworkVersion ({compileSdk})
```

This action is removed.  In its place is an *implicit inference*:

```
The TargetFrameworkVersion (Android API level {compileSdk}) is higher than the targetSdkVersion ({targetSdk}).
```

A more concrete example:

```
The TargetFrameworkVersion (Android API level 29) is higher than the targetSdkVersion (28).
```

How do I fix this message?  There are (at least?) two options:

 1. Lower `$(TargetFrameworkVersion)` -- and thus `compileSdk` -- so that it matches `targetSdkVersion`, or
 2. Increase `targetSdkVersion` to match `$(TargetFrameworkVersion)`.

These happen in entirely different files, neither of which is even mentioned in the message!

*Is This Message Actionable*?  I would say not.

If we're going to update the message, then we should *increase* it's utility.

brendanzagaeski: Whoops. In hindsight, I realize I should have highlighted different lines for this comment. The full new messages in the `.resx` file each include an extra sentence with an instruction that doesn't appear here in the README:

```
The TargetFrameworkVersion (Android API level {0}) is higher than the
targetSdkVersion ({1}). Please increase the `android:targetSdkVersion` in the
`AndroidManifest.xml` so that the API levels match.
```

And:

```
The TargetFrameworkVersion (Android API level {0}) is lower than the
targetSdkVersion ({1}). Please increase the `$(TargetFrameworkVersion)` or
decrease the `android:targetSdkVersion` in the `AndroidManifest.xml` so that
the API levels match.
```

When I was adding this comment, I forgot that I had left out the "Please increase" sentences from the README for brevity after noticing that the original version of XA1008 did that (which seemed potentially nice for tidiness of the README).

Happy to take another crack at wordings for the messages too, maybe to combine the two sentences for more compactness, if that would be useful.

PR #4197, opened 2020-01-30
InputStreamInvoker: Seeking for FileInputStreams using FileChannels

catlan:
Hi Xamarin Team,

`InputStreamInvoker` lack of seeking support seems like a oversight to me: 

When working with files across apps, `ContentResolver.openInputStream()` returns `ParcelFileDescriptor.AutoCloseInputStream` which is wrapped to `InputStreamInvoker`.

Requiring to copy this stream to a `MemoryStream` just to get seeking support is **bad** for the **memory usage** of any android app, but especially for older and entry level devices.  

This pull request implements seeking for all `FileInputStream` subclasses using `FileChannel` while keeping the old behaviour for other types of streams.

Review Notes:
* Given that `FileChannel` is connected to the `FileInputStream`, it seems good to me to `null` it before closing the stream.
* The channel doesn't required an explicit closed:  
    > ParcelFileDescriptor.AutoCloseInputStream close: If this stream has an associated channel then the channel is closed as well.

I tried to get into your [coding-guidelines](https://www.mono-project.com/community/contributing/coding-guidelines/, please excuse if I overlooked something and let me know.

Cheers,
Christopher

### DIFF
--- a/src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
+++ b/src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
@@ -6,7 +6,7 @@ namespace Android.Runtime
 	public class InputStreamInvoker : Stream
 	{
 		public Java.IO.InputStream BaseInputStream {get; private set;}
-		public Java.Nio.Channels.FileChannel BaseFileChannel {get; private set;}
+		protected Java.Nio.Channels.FileChannel BaseFileChannel {get; set;}
 
 		public InputStreamInvoker (Java.IO.InputStream stream)
 		{
@@ -16,7 +16,7 @@ namespace Android.Runtime
 			BaseInputStream = stream;
 
 			Java.IO.FileInputStream fileStream = stream as Java.IO.FileInputStream;
-			if (fileStream is object)
+			if (fileStream != null)
 				BaseFileChannel = fileStream.Channel;
 		}
 
@@ -98,22 +98,21 @@ namespace Android.Runtime
 
 		public override long Seek (long offset, SeekOrigin origin)
 		{
-			if (BaseFileChannel is object) {
-				switch (origin) {
-				case SeekOrigin.Begin:
-					BaseFileChannel.Position (offset);
-					break;
-				case SeekOrigin.Current:
-					BaseFileChannel.Position (BaseFileChannel.Position() + offset);
-					break;
-				case SeekOrigin.End:
-					BaseFileChannel.Position (BaseFileChannel.Size() + offset);
-					break;
-				}
-				return BaseFileChannel.Position ();
-			} else {
-				throw new NotSupportedException ();
+			if (BaseFileChannel == null)
+				throw new NotSupportedException();
+
+			switch (origin) {
+			case SeekOrigin.Begin:
+				BaseFileChannel.Position (offset);
+				break;
+			case SeekOrigin.Current:
+				BaseFileChannel.Position (BaseFileChannel.Position() + offset);
+				break;
+			case SeekOrigin.End:
+				BaseFileChannel.Position (BaseFileChannel.Size() + offset);
+				break;
 			}
+			return BaseFileChannel.Position ();
 		}
 
 		public override void SetLength (long value)
@@ -127,12 +126,12 @@ namespace Android.Runtime
 		}
 
 		public override bool CanRead { get { return true; } }
-		public override bool CanSeek { get { return (BaseFileChannel is object); } }
+		public override bool CanSeek { get { return (BaseFileChannel != null); } }
 		public override bool CanWrite { get { return false; } }
 
 		public override long Length {
 			get {
-				if (BaseFileChannel is object)
+				if (BaseFileChannel != null)
 					return BaseFileChannel.Size ();
 				else
 					throw new NotSupportedException ();
@@ -141,13 +140,13 @@ namespace Android.Runtime
 
 		public override long Position {
 			get {
-				if (BaseFileChannel is object)
+				if (BaseFileChannel != null)
 					return BaseFileChannel.Position ();
 				else
 					throw new NotSupportedException ();
 			}
 			set {
-				if (BaseFileChannel is object)
+				if (BaseFileChannel != null)
 					BaseFileChannel.Position (value);
 				else
 					throw new NotSupportedException ();
--- a/src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
+++ b/src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
@@ -99,7 +99,7 @@ namespace Android.Runtime
 		public override long Seek (long offset, SeekOrigin origin)
 		{
 			if (BaseFileChannel == null)
-				throw new NotSupportedException();
+				throw new NotSupportedException ();
 
 			switch (origin) {
 			case SeekOrigin.Begin:
--- a/src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
+++ b/src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
@@ -6,7 +6,8 @@ namespace Android.Runtime
 	public class InputStreamInvoker : Stream
 	{
 		public Java.IO.InputStream BaseInputStream {get; private set;}
-		protected Java.Nio.Channels.FileChannel BaseFileChannel {get; set;}
+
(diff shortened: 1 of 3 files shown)


src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
```diff
@@ -6,13 +6,18 @@ namespace Android.Runtime
 	public class InputStreamInvoker : Stream
 	{
 		public Java.IO.InputStream BaseInputStream {get; private set;}
+		public Java.Nio.Channels.FileChannel BaseFileChannel {get; private set;}
```
jonpryor: *Should* this be public?  I'm not sure I see immediate benefit to making this a public API.

src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
```diff
@@ -6,13 +6,18 @@ namespace Android.Runtime
...
 		public InputStreamInvoker (Java.IO.InputStream stream)
 		{
 			if (stream == null)
 				throw new ArgumentNullException (nameof (stream));
 
-			this.BaseInputStream = stream;
+			BaseInputStream = stream;
+
+			Java.IO.FileInputStream fileStream = stream as Java.IO.FileInputStream;
+			if (fileStream is object)
```
jonpryor: Why not `fileStream != null`?

src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
```diff
@@ -92,7 +98,22 @@ public override int Read (byte[] buffer, int offset, int count)
...
+				case SeekOrigin.Current:
+					BaseFileChannel.Position (BaseFileChannel.Position() + offset);
+					break;
+				case SeekOrigin.End:
+					BaseFileChannel.Position (BaseFileChannel.Size() + offset);
+					break;
+				}
+				return BaseFileChannel.Position ();
+			} else {
+				throw new NotSupportedException ();
```
jonpryor: Style convention: short blocks should go first, as that can take advantage of early exits and reduce indentation levels:

```csharp
if (BaseFileChannel == null)
    throw new NotSupportedException ();
switch (origin) {
    ...
}
```

jonpryor:
Would it be possible to add a unit test to `src/Mono.Android/Test`?

src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
catlan: Just a preference (https://twitter.com/jaredpar/status/1115019017297596416), if you like me to change it no problem either.

src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
catlan: You're right, will change to protected.

jonpryor: Not sure I see much benefit to `protected` either.  All it does is prevent an `BaesInputStream as FileChannel` expression within the subclass, which doesn't feel like much benefit to me...

catlan:
Added unit test.

src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
catlan: I think it should be direct accessible from a subclass. For example in my project I was able to fix this issue by subclassing InputStreamInvoker and having access to the BaseInputStream. While it is true that even if it is marked as private, any subclass could just access it via ``Java.IO.FileInputStream fileStream = stream as Java.IO.FileInputStream; if (fileStream is object)``, I don't see the benefit in private here.
I don't understand your point about "All it does is prevent an BaesInputStream as FileChannel expression".
In the end, please let me know how you want it, it's your code base and I wanna finish this.

src/Mono.Android/Android.Runtime/InputStreamInvoker.cs
```diff
@@ -92,7 +98,21 @@ public override int Read (byte[] buffer, int offset, int count)
...
+
+			switch (origin) {
+			case SeekOrigin.Begin:
+				BaseFileChannel.Position (offset);
+				break;
+			case SeekOrigin.Current:
+				BaseFileChannel.Position (BaseFileChannel.Position() + offset);
+				break;
+			case SeekOrigin.End:
+				BaseFileChannel.Position (BaseFileChannel.Size() + offset);
```
jonpryor: Shouldn't this be `BaseFileChannel.Size () - offset`?  (Note **`-`** subtraction.)

catlan: If you have a look at https://docs.microsoft.com/en-us/dotnet/api/system.io.seekorigin?view=netframework-4.8 you will see that a negative offset parameter is used. Other reference implementations: https://github.com/dotnet/aspnetcore/blob/58a107b27e7a73cf24f68b7ad38818a59b5289b9/src/Http/Http/src/Internal/ReferenceReadStream.cs#L88-L91 and https://github.com/dotnet/aspnetcore/blob/86728ecc17c339f699d21c8abd5ee93037b51a1e/src/Http/WebUtilities/src/MultipartReaderStream.cs#L113-L116

jonpryor: Ah, silly me.  Callers would provide a negative `offset` value with `SeekOrigin.End`.

jonpryor:
Unfortunately, the new `Android.RuntimeTests.InputStreamInvokerTest.InputStreamTest()` unit test is failing:

```
Expected string length 24 but was 1. Strings differ at index 1.
 Expected: "Your're so good looking!"
 But was: "Y"

---
  at Android.RuntimeTests.InputStreamInvokerTest.InputStreamTest () [0x00256] in <8d3a4772726b422188fc09fdd6bb401f>:0 
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <0e3c71155e0949349dcc08f05811134e>:0 
```

src/Mono.Android/Test/Android.Runtime/InputStreamInvokerTest.cs
```diff
@@ -0,0 +1,63 @@
...
+                                Assert.IsTrue (stream.CanRead);
+                                Assert.IsFalse (stream.CanSeek);
+                                Assert.IsFalse (stream.CanTimeout);
+                                Assert.IsFalse (stream.CanWrite);
+                                Assert.Throws<NotSupportedException> (() => { var _ = stream.Length; });
+                                Assert.Throws<NotSupportedException> (() => { var _ = stream.Position; });
+                                Assert.Throws<NotSupportedException> (() => { stream.Position = 1; });
+                                Assert.Throws<NotSupportedException> (() => { stream.Seek (1, SeekOrigin.Begin); });
+                                Assert.AreEqual (stream.Read (bytes, 0, text.Length), text.Length);
+                                Assert.AreEqual (text, Encoding.ASCII.GetString (bytes, 0, 1));
```
jonpryor: This appears to be the assert which is failing, and upon review it doesn't make sense to me: `Encoding.ASCII.GetBytes(bytes, 0, 1)` will only return a string with a single character, while `text` contains 24 characters.

Should this instead be?

```csharp
Assert.AreEqual (text, Encoding.ASCII.GetString (bytes, 0, text.Length));
```

catlan: Sorry about that, yes looks like I copied that last line without running it. text.Length instead of 1 is correct.

brendanzagaeski:
Thanks again for the contribution!

*Release status update*

A new Release version of Xamarin.Android has now been published that includes the new functionality from this item.

New functionality included in Xamarin.Android 10.3.1.0.

New functionality included on Windows in Visual Studio 2019 version 16.6. To get the new version that includes the fix, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) or install the latest version from <https://visualstudio.microsoft.com/downloads/>.

New functionality included on macOS in Visual Studio 2019 for Mac version 8.6. To get the new version that includes the fix, check for the latest updates on the **Stable** [updater channel](https://docs.microsoft.com/visualstudio/mac/update).

PR #4207, opened 2020-01-30
[build] Convert JavadocImporter to short-form style

pjcollins:
Context: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3429460&view=ms.vss-test-web.build-test-results-tab

We've been seeing some failures in the `DotNetPackageXASdkProject` test
on Windows [0] since commit 03743f32. This failure was a result of us
packaging the wrong version of `libZipSharp.dll` on Windows. The
build of `javadoc2mdoc.csproj` would run after the build of
`Xamarin.Android.Build.Tasks.csproj`, and as a result the `net45`
version of `libZipSharp.dll` was being copied on top of (and replacing)
the `netstandard2.0` version that was previously copied by
`Xamarin.Android.Build.Tasks.csproj`.

Fix the problem by updating JavadocImporter to a short form style
project targeting `netstandard2.0`, so that the `netstandard2.0` version
of `libZipSharp.dll` will be used everywhere.

In the future we may want to consider updating our short-form style
projects that set `AppendTargetFrameworkToOutputPath` to false. This was
done in commit 03743f32 to decrease the scope of change required to
target `netstandard2.0`, however it can lead to unexpected issues.

[0^]

    C:\A\xamarin-v00000L-1\_work\2\s\packages\xamarin.android.sdk\0.0.1\tools\Xamarin.Android.Common.targets(784,2): error MSB4062: The "Xamarin.Android.Tasks.ResolveSdks" task could not be loaded from the assembly C:\A\xamarin-v00000L-1\_work\2\s\packages\xamarin.android.sdk\0.0.1\tools\Xamarin.Android.Build.Tasks.dll. Could not load file or assembly 'libZipSharp, Version=1.0.7328.28304, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.


pjcollins:
Full test run is here - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3432398&view=results

pjcollins:
The only failure in the full run appears to be something previously introduced:
>   the reference 34559812 and actual size 34948932 for apk-Release-Aot differs too much

PR #4214, opened 2020-02-03
[Xamarin.Android.Build.Tasks] CleanBOM could leave behind a temp file

jonathanpeppers:
Fixes: http://work.azdo.io/774576

In reviewing `Path.GetTemp*` usage, I found the
`MonoAndroidHelper.CleanBOM` method could leave behind temp files if
an exception is thrown. It did not use a `try-finally` block.

I took the time to refactor the method and add a general test for it.

The other places I reviewed seem to be OK. They have `File.Delete`
calls in a `try-finally` block. Some are at the MSBuild task level.


jonathanpeppers:
There might be a perf regression somehow:
```
Build_AndroidResource_Change
Exceeded expected time of 4250ms, actual 4320ms
```
Need to look into it.

jonathanpeppers:
I didn't see a perf regression in the build log, it seems like these performance tests passed on re-run.

Reviewing the code, it seems like the new code would actually be slightly faster. It skips a hash check on both files--which would always differ.

PR #4222, opened 2020-02-05
[Xamarin.Android.Build.Tasks] Re-enable Resources.Designer.cs generation

brendanzagaeski:
Now that [03743f32][0] has switched the
`Xamarin.Android.Build.Tasks.csproj` file to the short-form project
style, some item metadata needs to be added explicitly to
`Resources.resx` and `Resources.Designer.cs` to tell Visual Studio to
generate `Resources.Designer.cs` from `Resources.resx` as expected.  The
short-form project style does add the files automatically to the
appropriate `@(EmbeddedResource)` and `@(Compile)` item lists via the
default wildcard includes, but it doesn't add this metadata.

The new items are almost the same as in [0342fe56][1], but they use the
`Update` attribute instead of the `Include` attribute because now the
items are already included via the wildcard includes.

[0]: https://github.com/xamarin/xamarin-android/commit/03743f32aec3b02edc280886061b12e0dde0bd1c
[1]: https://github.com/xamarin/xamarin-android/commit/0342fe5698b86e21e36c924732ff135b9a87e4af

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
@@ -427,6 +427,21 @@
     <ProjectReference Include="..\..\external\Java.Interop\src\Java.Interop\Java.Interop.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Compile Update="Properties\Resources.Designer.cs">
+      <DesignTime>True</DesignTime>
+      <AutoGen>True</AutoGen>
+      <DependentUpon>Resources.resx</DependentUpon>
+    </Compile>
+  </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Update="Properties\Resources.resx">
+      <Generator>ResXFileCodeGenerator</Generator>
+      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
+    </EmbeddedResource>
+  </ItemGroup>
+
   <!-- Import Microsoft.NET.Sdk targets before our targets so we can override behavior -->
   <!-- See https://github.com/microsoft/msbuild/pull/4922 -->
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
```diff
@@ -427,6 +427,21 @@
...
+    </Compile>
+  </ItemGroup>
+
+  <ItemGroup>
+    <EmbeddedResource Update="Properties\Resources.resx">
+      <Generator>ResXFileCodeGenerator</Generator>
+      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
+    </EmbeddedResource>
+  </ItemGroup>
+
```
brendanzagaeski: The discussion topic I have in mind while this is a Draft is whether it seems like a good time to try adding a dependency on dotnet/arcade so that Xamarin.Android.Build.Tasks can start using the [`GenerateResxSource` mechanism](https://github.com/dotnet/arcade/blob/e67d9f098029ebecedf11012a749b532d68ad2a9/Documentation/ArcadeSdk.md#generateresxsource-bool) to generate C# from the `.resx` file at build time instead of using the `ResXFileCodeGenerator` IDE mechanism.

Doing that would also slightly simplify the `.xlf` generation story since dotnet/arcade takes care of bringing in the xliff-tasks dependency. As I understand it, dotnet/arcade offers features on an opt-in basis, so it would be fine to use it just for these localization scenarios for now, with the possibility to use other features in the future too if desired.

Before I started down that path, I thought it would be good to check if it seemed like a good time to start on that, or if it would make more sense to hold off for now and keep using the `ResXFileCodeGenerator` for a while longer.

pjcollins: Sorry about breaking this, I should have pinged you on the short-form style conversion PR. We did discuss the loss of this metadata briefly but I thought it was acceptable given the limited information I could find at the time.

I like the idea of migrating to dotnet/arcade tooling sooner than later if it makes the localization workflow easier. Perhaps that change would also serve as an additional safeguard against the kind of breakage this PR is fixing?

brendanzagaeski: No worries on the breakage. It's easy to work around locally, and it's great to have the project in the short-form style now.

I'll plan to take a little look at using the dotnet/arcade approach today or tomorrow to see if it's quick and easy.

brendanzagaeski: I got back to looking into dotnet/arcade, and I learned that it would change a bit more than I was picturing. It turns out that not all of the Arcade features are strictly opt-in. For example, it places all build outputs in an `artifacts` directory. Although the top-level `artifacts` path is customizable, some of the subdirectories (including the `bin` and `obj` directories) are hard-coded relative to that top-level path:

https://github.com/dotnet/arcade/blob/dd45d827ad6d8e1467619a75264e23e1073f69e1/src/Microsoft.DotNet.Arcade.Sdk/tools/RepoLayout.props#L49-L52

So using Arcade is at least a bit more involved than I thought, and it's probably best to keep using the old `ResXFileCodeGenerator` mechanism for now in the context of this specific PR.

Onboarding xamarin-android to use Arcade can then happen separately (if it looks like a good fit), and the `.resx` code generation mechanism can be revisited later, depending on how that work goes.

PR #4223, opened 2020-02-05
[Xamarin.Android.Build.Tasks] Dont always lowercase "*Layout" element values

dellis1972:
Fixes http://work.azdo.io/1042658

Commit a6387737 made the assumption that ALL `*Layout` view
extension values needed to be lowercased. This is not always
the case. In the Issue listed above the user defined custom
enumeration values which were CamelCase. These were not lowercased
by our build process. So when we lowercased the value in the
layout we got the following build error

	Error		‘fixedsize’ is incompatible with attribute ChildLayout (attr) enum [Auto=0, FixedHeight=2, FixedSize=3, FixedWidth=1, MaxHeightAlignBottom=16, MaxHeightCenter=32, MaxWidthAlignRight=4, MaxWidthCenter=8]

To work around this issue this PR logs all of the `attr` values it
comes across during the conversion of the casing. If the value
for the view extension exists in that list it will not be lowercased.
Otherwise the code will proceed as normal.

The order in which the xml files are processed also changed to make
this work. Before we just processed them in the order they were found.
Now we process them by the directory they are in and in reverse. This
is so the `values` folder which contain the `attribs.xml` based files
are processed first.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
@@ -202,7 +202,7 @@ namespace Monodroid {
 		{
 			if (attr.Name.Namespace != res_auto)
 				return false;
-			if (!attr.Value.StartsWith ("@")) // only lowercase if we reference another resource.
+			if (!attr.Value.StartsWith ("@", StringComparison.Ordinal)) // only lowercase if we reference another resource.
 				return false;
 			if (attr.Name.LocalName.EndsWith ("Layout", StringComparison.Ordinal)) {
 				attr.Value = attr.Value.ToLowerInvariant ();
--- a/src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
@@ -202,6 +202,8 @@ namespace Monodroid {
 		{
 			if (attr.Name.Namespace != res_auto)
 				return false;
+			if (!attr.Value.StartsWith ("@")) // only lowercase if we reference another resource.
+				return false;
 			if (attr.Name.LocalName.EndsWith ("Layout", StringComparison.Ordinal)) {
 				attr.Value = attr.Value.ToLowerInvariant ();
 				return true;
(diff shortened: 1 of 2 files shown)


dellis1972:
@jonathanpeppers not sure if this is such a good idea now. Perhaps we should just check for the `@` symbol at the start of the value. That should indicate its a resource and should be lowercased.

jonathanpeppers:
@dellis1972 yeah I was trying to follow what is supposed to be lowercased here.

Should we even be changing a custom enum like this:
```xml
<resources>
	<declare-styleable name=""UserLayoutAttributes"">
		<attr name=""UserLayout"">
			<enum name=""FixedWidth"" value=""0""/>
		</attr>
	</declare-styleable>
</resources>
```
Is aapt2 OK with `app:UserLayout="FixedWidth"`? If it is uppercase in both places, maybe we need to leave it alone.

dellis1972:
@jonathanpeppers I had a though. Perhaps this can be even simpler. We should only lowercase stuff if it looks like a `resource` value. i.e `@layout/foo` or `@drawable/foo` etc. And leave all other values alone. I think we only need to lowercase stuff because we lowercase all the filenames.

src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
```diff
@@ -202,6 +202,8 @@ private static bool TryFixResAuto (XAttribute attr, Dictionary<string, string> a
 		{
 			if (attr.Name.Namespace != res_auto)
 				return false;
+			if (!attr.Name.LocalName.StartsWith ("@")) // only lowercase if we reference another resource.
```
jonathanpeppers: Should this actually be looking at `attr.Value`? I think the value of the attribute will have the `@` symbol?

dellis1972: 🤦‍♂

brendanzagaeski:
*Draft release notes*

```

  * [Developer Community 861086](https://developercommunity.visualstudio.com/content/problem/861086/enum-attribute-no-more-compatible-with-def-in-vs20.html):
    Starting in Xamarin.Android 10.1, errors similar to `'fixedsize' is
    incompatible with attribute ChildLayout (attr) enum` could prevent building
    successfully in projects that used custom attribute names that ended with
    the word `Layout`.
```

dellis1972:
Test failure is Build_JLO_Change, "Exceeded expected time of 8700ms, actual 8730ms" 
I don't think its related to this PR. Probably just the bot being a bit slow.

src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
```diff
@@ -202,6 +202,8 @@ private static bool TryFixResAuto (XAttribute attr, Dictionary<string, string> a
 		{
 			if (attr.Name.Namespace != res_auto)
 				return false;
+			if (!attr.Value.StartsWith ("@")) // only lowercase if we reference another resource.
```
jonathanpeppers: The only think we should probably do is:
```suggestion
			if (!attr.Value.StartsWith ("@", StringComparison.Ordinal)) // only lowercase if we reference another resource.
```

PR #4228, opened 2020-02-07
[Xamarin.Android.Build.Tasks] Split up XA5101 errors and make them localizable

brendanzagaeski:

Give the XA5101 errors in `NdkUtil` their own new error codes so they
can be distinguished from the `NdkUtilOld` errors in telemetry.

Move the message text for XA5101 and the other new error codes into the
`.resx` file so that they are ready for localization.

### DIFF
--- a/Documentation/guides/messages/README.md
+++ b/Documentation/guides/messages/README.md
@@ -107,6 +107,9 @@ ms.date: 01/24/2020
 
 ## XA5xxx: GCC and toolchain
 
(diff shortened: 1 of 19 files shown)


brendanzagaeski:
Hmm. I see the build failed this time because of:

```
Artifact Build Results - macOS already exists for build 3452253.
```

So the tests still didn't run for this PR. I think there was some chat today about possible strategies to give the builds different IDs on retries. In the mean time, I'll just add an empty commit to this branch to try the build once more.

PR #4229, opened 2020-02-07
[Xamarin.Android.Build.Tasks] Make XA1001 and XA1002 localizable

brendanzagaeski:

Move the XA1001 and XA1002 message text into the `.resx` file so that
they are ready for localization.

Other changes:

Remove "we" from the XA1002 message, and change "casing" to
"capitalization" to reduce potential ambiguity for the l10n team.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
@@ -27,7 +27,7 @@ namespace Monodroid {
 				}
 				return Xamarin.Android.Tasks.MonoAndroidHelper.CopyIfChanged (tmpfile, filename);
 			} catch (Exception e) {
-				logMessage?.Invoke (TraceLevel.Warning, $"AndroidResgen: Warning while updating Resource XML '{filename}': {e.Message}");
+				logMessage?.Invoke (TraceLevel.Warning, string.Format (Xamarin.Android.Tasks.Properties.Resources.XA1001, filename, e.Message));
 				return false;
 			} finally {
 				if (File.Exists (tmpfile)) {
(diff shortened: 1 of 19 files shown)


src/Xamarin.Android.Build.Tasks/Utilities/AndroidResource.cs
```diff
@@ -27,7 +27,7 @@ public static bool UpdateXmlResource (string res, string filename, Dictionary<st
 				}
 				return Xamarin.Android.Tasks.MonoAndroidHelper.CopyIfChanged (tmpfile, filename);
 			} catch (Exception e) {
-				logMessage?.Invoke (TraceLevel.Warning, $"AndroidResgen: Warning while updating Resource XML '{filename}': {e.Message}");
+				logMessage?.Invoke (TraceLevel.Warning, string.Format (Xamarin.Android.Tasks.Properties.Resources.XA1001, filename, e.Message));
```
brendanzagaeski: One little interesting thing about these two messages is that the `logMessage.Invoke()` call doesn't itself mention the message code. So I double-checked that all the callers of `UpdateXmlResurce()` agreed on the message code. `UpdateXmlResurce()` currently has two callers, and both do agree on the XA1001 message code:

https://github.com/xamarin/xamarin-android/blob/1273970bcc77191177617225d21ac9772afd62cd/src/Xamarin.Android.Build.Tasks/Tasks/ConvertResourcesCases.cs#L117-L118

and

https://github.com/xamarin/xamarin-android/blob/1273970bcc77191177617225d21ac9772afd62cd/src/Xamarin.Android.Build.Tasks/Tasks/CopyAndConvertResources.cs#L106-L107

I similarly checked the callers of `TryFixCustomView()` for the XA1002 message code.

PR #4230, opened 2020-02-07
Add support for legacy WebRequestHandler as a HttpHandler (for NTLM)

EgorBo:
Currently Android has two HttpHandler options: **Managed** (which is based on modern fully managed `SocketHttpHandler` from corefx) and **Android**. Unfortunately it turned out in order to support NTLM/Kerberos authentication, `SocketHttpHandler` needs some native pieces (GSS) we don't have for Android (basically we need to port [krb5](https://github.com/krb5/krb5) lib to Android NDK) - other platforms including iOS are fine.

However, our old legacy `WebRequestHandler` used to fully support NTLM. So the idea is to bring it back and allow users to select it via VS

image

`System.Net.Http.dll` assembly already has two things we need to create a legacy handler:
1) internal `System.Net.Http.MonoWebRequestHandler` type
2) internal `HttpClientHandler(IMonoHttpClientHandler handler)` ctor 

However it also possible to simply use `WebRequestHandler` from `System.Net.Http.WebRequest.dll`. 

So if I am not mistaken, in order to update that combox in VS we need to add `System.Net.Http.WebRequestHandler` here in xamarin/XamarinVS: https://github.com/xamarin/XamarinVS/blob/7c01004f5d8a308771ab0d02887b79067e100b6c/src/Features/VisualStudio.Properties/VisualStudio.Android.Properties/Views/OptionsPageViewModel.cs#L52

Optionally: update this UI tip, update localizations.
image

So basically, the idea is to fix the following code without changes on the user side:
```csharp
var handler = new HttpClientHandler { Credentials = ntlmCredentials }; 
var httpClient = new HttpClient(handler)
```

### DIFF
--- a/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
+++ b/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
@@ -363,6 +363,33 @@ namespace Android.Runtime {
 				handlerType = Type.GetType (envvar, false);
 			else
 				return null;
+
+			if (handlerType == null && envvar.Contains("System.Net.Http.WebRequestHandler"))
+			{
+				// if WebRequestHandler is null it means System.Net.Http.WebRequest assembly is not referenced,
+				// however, we stil can create it since underlying parts of it are located in System.Net.Http.
+				// So basically it can be create via reflection:
+				//
+				//    return new HttpClientHandler(new MonoWebRequestHandler());
+				//
+				Type monoHandlerType = Type.GetType("System.Net.Http.MonoWebRequestHandler, System.Net.Http");
+				if (monoHandlerType == null)
+					return null;
+
+				ConstructorInfo[] internalMonoHandlerCtors = monoHandlerType.GetConstructors(
+					BindingFlags.NonPublic | BindingFlags.Instance);
+				if (internalMonoHandlerCtors.Length < 1)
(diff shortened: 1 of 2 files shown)


jonpryor:
This PR is "working too hard": the [documentation](https://docs.microsoft.com/en-us/xamarin/android/deploy-test/building-apps/build-process) for the `$(AndroidHttpClientHandlerType)` property states:

> The value is an assembly-qualified type name of an `HttpMessageHandler` subclass, suitable for use with `System.Type.GetType(string)`.

Thus, the expectation -- from the Xamarin.Android side of things, not the IDE side -- is that the `.csproj` could contain something like:

```xml
<PropertyGroup>
  <AndroidHttpClientHandlerType>System.Net.Http.MonoWebRequestHandler, System.Net.Http</AndroidHttpClientHandlerType>
</PropertyGroup>
```

and the resulting app would Just Work™.  Most/all of this PR, as such, should not even be necessary.

What would be "nice" would be an NTLM test in a new or existing app, which might be "too big" of an ask.

...and IDE support, so that there is a 'nicer' user-visible string for the `System.Net.Http.MonoWebRequestHandler, System.Net.Http` value.

EgorBo:
@jonpryor looks like I indeed overengineered it and you're right 🙂

Issue #4231, opened 2020-02-07
Analyze the Java.Interop size increase after enabling NRT


radekdoulik:
Context: https://github.com/xamarin/xamarin-android/commit/5945e9a77988a01ceb062dcc82bda6c6bf407e14
Context: https://jenkins.mono-project.com/view/Xamarin.Android/job/xamarin-android/plot/Tests%20size/

I will compare the `Java.Interop.dll` before and after to see where does the increase come from. It might be caused by additional nullable attributes we might be able to remove in the linker.

PR #4232, opened 2020-02-07
[size] Update apk size reference

radekdoulik:
Mono.Android increase comes from these commits

https://github.com/xamarin/xamarin-android/commit/476921a7e526ffd522245de2bcfef93f1a91fd0a
https://github.com/xamarin/xamarin-android/commit/1273970bcc77191177617225d21ac9772afd62cd

First is NDK bump, which adds about 40-50kbytes to our tests apks and
we cannot do much about it.

Second is adding code to the test, so the size increase is
expected. In future we plan to stop using Mono.Android test for size
regression testing as it changes frequently. We will be using a
HelloWorld sample, which will be updated, so that it is close to XA
app template.

Then there is XF test size increase coming from Java.Interop bump
https://github.com/xamarin/xamarin-android/commit/5945e9a77988a01ceb062dcc82bda6c6bf407e14

For this one I have opened an issue https://github.com/xamarin/xamarin-android/issues/4231 to analyze the size increase from enabling NRT.


jonpryor:
Looks like the values still need changing, e.g. [the most recent build has .apk size warnings](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3454129&view=logs&j=8556562a-ae5f-5bd1-7c4d-bf1af4b6f1e1&t=00ff70e0-f6b2-5f4a-423a-26cf80f77b2b)

```
  /Users/runner/runners/2.164.8/work/1/s/build-tools/scripts/TestApks.targets(339,5): warning : the reference 16823756 and actual size 16877004 for apk-Release differs too much (>50kb). [/Users/runner/runners/2.164.8/work/1/s/src/Mono.Android/Test/Mono.Android-Tests.csproj]

```

PR #4233, opened 2020-02-07
[Xamarin.Android.Build.Tasks] deprecate [assembly: DoNotPackage]

jonathanpeppers:
The `<ResolveAssemblies/>` MSBuild task currently looks for an
attribute in every assembly:

    [assembly: Java.Interop.DoNotPackage ("foo.jar")]

`foo.jar` would be used at compile time for java code, but not
included in the final Android app. This seems like the wrong way to go
about things, you can simply add to the `@(AndroidExternalJavaLibrary)`
item group for the same behavior.

I also found very little usage of this on Github:

    https://github.com/search?l=C%23&p=1&q=%5Bassembly%3A+DoNotPackage&type=Code

Most of the projects using this are unsupported libraries from the
Xamarin Component store.

If we plan to completely remove `<ResolveAssemblies/>` this is a piece
we will have to deprecate for a period of time first.

### DIFF
--- a/Documentation/guides/messages/xa0122.md
+++ b/Documentation/guides/messages/xa0122.md
@@ -1,0 +1,32 @@
+---
+title: Xamarin.Android warning XA0122
+description: XA0122 warning code
+ms.date: 02/07/2020
+---
+# Xamarin.Android warning XA0122
+
+## Example messages
+
+```
+warning XA0122: Assembly 'Library1' is using a deprecated attribute '[assembly: Java.Interop.DoNotPackageAttribute]'. Use a newer version of this NuGet package or notify the library author.
+```
+
+## Issue
+
+The behavior implemented in the `DoNotPackageAttribute` is deprecated:
+
+    [assembly: Java.Interop.DoNotPackage ("foo.jar")]
+
+This would prevent `foo.jar` from being packaged in the app.
+
+Alternatively, you can use the `@(AndroidExternalJavaLibrary)` item
+group for including `foo.jar`. The java library will only be used at
+compile time, and will not be packaged in the final Android app.
+
+## Solution
+
+Some libraries using this feature can be simply updated to a newer
+version on NuGet.
+
+Library authors will need to remove usage of this attribute. Its
+functionality will be removed in a future release.
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -137,6 +137,10 @@
     <value>Ignoring Reference Assembly `{0}`.</value>
     <comment>{0} - File name of the assembly.</comment>
   </data>
+  <data name="XA0122" xml:space="preserve">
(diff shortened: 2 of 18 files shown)


jonpryor:
> If we plan to completely remove `<ResolveAssemblies/>` this is a piece
> we will have to deprecate for a period of time first.

I don't see how these are directly related?  Why does removing `<ResolveAssemblies/>` require deprecating `DoNotPackageAttribute`?

jonathanpeppers:
> @jonpryor: Why does removing <ResolveAssemblies/> require deprecating DoNotPackageAttribute?

`<ResolveAssemblies/>` was removed, we would still have to include some MSBuild task that opened every assembly looking for attributes. We couldn't move this check to `<FilterAssemblies/>` because that one runs _before_ `<Csc/>`, and it wouldn't catch any `[assembly: DoNotPackage]` in the main app project.

If `DoNotPackage` has some value that makes it useful, we can keep it, though.

Documentation/guides/messages/xa0122.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Xamarin.Android warning XA0122
+description: XA0122 warning code
+ms.date: 02/07/2020
+---
+# Xamarin.Android warning XA0122
+
```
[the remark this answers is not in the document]
jonpryor: This isn't actionable if you're the library author.  *What should you do* if you're the library author to avoid this message?

The answer: use the `@(AndroidExternalLibrary)` Build action *instead of* `[DoNotPackageAttribute]`.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -137,6 +137,10 @@
     <value>Ignoring Reference Assembly `{0}`.</value>
     <comment>{0} - File name of the assembly.</comment>
   </data>
+  <data name="XA0122" xml:space="preserve">
+    <value>Assembly '{0}' is using a deprecated attribute '[assembly: {1}]'. Use a newer version of this NuGet package or notify the library author.</value>
```
jonpryor: Instead of a generic "XA0122" warning for "deprecated attribute(s)", this should instead be a warning specific to `Java.Interop.DoNotPackageAttribute` *so that* the warning message can specifically suggest the use of the `@(AndroidExternalLibrary)` Build action as a replacement.

brendanzagaeski: Peppers might have some good ideas on ways to cover both audiences, but one little tricky thing about this warning is that it can appear either for library authors or library consumers. For the library consumers, the information about `@(AndroidExternalLibrary)` would potentially be confusing.

Based on the GitHub search results, I'm cautiously optimistic there will be almost no one who will see this warning. I searched around a little to gather a little more info on the Xamarin authored NuGet packages that used the attribute too. From what I could tell, they are delisted now. (An example is Xamarin.Amazon.DeviceMessaging.)

jonathanpeppers: I think I will make `[Obsolete]` mention `@(AndroidExternalLibrary)` so NuGet authors get that message.

The other message seems appropriate, I think.

PR #4236, opened 2020-02-07
[Xamarin.Android.Build.Tasks] perf improvements for <ResolveAssemblies/>

jonathanpeppers:
When building the SmartHotel360 app, I found that
`<ResolveAssemblies/>` was loading the NuGet `ProjectAssetFile` /
`LockFile` even though it was not used by the task at all. Putting a
`Stopwatch` around the code, I also found it to be somewhat expensive.

I reworked the code to use a `Lazy<T>` for accessing the `LockFile`.
This way it is not actually loaded until it is needed.

I also found several `LogDebugMessage` calls that seemed to be too
much... The `<ResolveAssemblies/>` was emitting hundreds of lines of
logging in the SmartHotel360 app.

I removed two messages such as:

    LogDebugMessage ($"Adding {resolved_assembly} to topAssemblyReferences");
    ...
    LogDebugMessage ("{0}={1}", assemblyName, apiLevel);

The `[Input]` and `[Output]` values for the task and the remaining log
message should be sufficient for understanding what we need from
customer logs:

    LogMessage ("{0}Adding assembly reference for {1}, recursively...", new string (' ', indent), assemblyName);

The results for a build with no changes with the SmartHotel360 app:

    Before:
    320 ms  ResolveAssemblies                          1 calls
    After:
    238 ms  ResolveAssemblies                          1 calls

This saves ~82ms, and since this task runs on every `Build` and
`Install`, it could save ~164ms from the dev-loop.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAssemblies.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAssemblies.cs
@@ -72,15 +72,6 @@ namespace Xamarin.Android.Tasks
 
 			var assemblies = new Dictionary<string, ITaskItem> (Assemblies.Length);
 			var topAssemblyReferences = new List<string> (Assemblies.Length);
-			var logger = new NuGetLogger((s) => {
-				LogDebugMessage ("{0}", s);
-			});
-
-			LockFile lockFile = null;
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
The build failures look to be on master, too: apk sizes, tls tests, designer, libZipSharp/DotNetPackageXASdkProject test.

PR #4246, opened 2020-02-11
[apkdiff] Put apkdiff tool in separate output directory

radekdoulik:
We were clashing with `Xamarin.Android.Build.Tasks` project assemblies in
`lib/xamarin.android/xbuild/Xamarin/Android/`, so put it in `lib/xamarin.android/xbuild/Xamarin/Android/apkdiff/` instead.

There were 2 issues:

* the `Xamarin.Android.Build.Tasks` project requires `netstandard2.0` version of
`libZipSharp.dll`. Sometimes the `apkdiff` is built after
`Xamarin.Android.Build.Tasks` and thus we end up with `net45` version,
which causes errors like this one:

    C:\A\xamarin-v00000K-1\_work\1\s\packages\xamarin.android.sdk\0.0.1\tools\Xamarin.Android.Common.targets(785,2): error MSB4062: The "Xamarin.Android.Tasks.ResolveSdks" task could not be loaded from the assembly C:\A\xamarin-v00000K-1\_work\1\s\packages\xamarin.android.sdk\0.0.1\tools\Xamarin.Android.Build.Tasks.dll. Could not load file or assembly 'libZipSharp, Version=1.0.7328.28304, Culture=neutral, PublicKeyToken=null'. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [C:\A\xamarin-v00000K-1\_work\1\s\bin\TestRelease\temp\DotNetPackageXASdkProjectTrue\UnnamedProject.csproj]

* the `Xamarin.Android.Build.Tasks` project uses `Newtonsoft.Json` assembly as well. It is removed during the build process (in `ILRepacker` target), so we end up without the assembly. That leads to fail when creating the installation. The error:

    /Users/builder/jenkins/workspace/xamarin-android/xamarin-android/build-tools/create-pkg/create-pkg.targets(34,5): error MSB3030: **Could not copy the file** "/Users/builder/jenkins/workspace/xamarin-android/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android/**Newtonsoft.Json.dll**" because it was not found. [/Users/builder/jenkins/workspace/xamarin-android/xamarin-android/build-tools/create-pkg/create-pkg.csproj]

Also use the `$(LibZipSharpVersion)` property.

Do not reference `Mono.Posix.NETStandard`, context: https://github.com/xamarin/xamarin-android/pull/4267

### DIFF
--- a/build-tools/installers/create-installers.targets
+++ b/build-tools/installers/create-installers.targets
@@ -105,7 +105,6 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\apkdiff.exe" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\apkdiff.pdb" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\Mono.Options.dll" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\Mono.Posix.NETStandard.dll" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\libZipSharp.dll" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\libZipSharp.dll.config" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\Newtonsoft.Json.dll" />
--- a/build-tools/installers/create-installers.targets
+++ b/build-tools/installers/create-installers.targets
@@ -102,8 +102,14 @@
   </ItemGroup>
   <ItemGroup>
     <_MSBuildFiles Include="$(MSBuildSrcDir)\android-support-multidex.jar" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff.exe" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff.pdb" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\apkdiff.exe" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\apkdiff.pdb" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\Mono.Options.dll" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\Mono.Posix.NETStandard.dll" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\libZipSharp.dll" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\libZipSharp.dll.config" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\Newtonsoft.Json.dll" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\Newtonsoft.Json-LICENSE.md" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\aprofutil.exe" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\aprofutil.pdb" />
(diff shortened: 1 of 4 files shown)


jonpryor:
This does not appear to have fixed the issue: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3461003&view=logs&j=cac0e8d3-0ef5-5d2b-b57e-e8fde7204df3&t=b0f00238-050a-5f78-0891-501fba8e3b4e

We're still getting failures in `Xamarin.Android.Build.Tests.XASdkTests.DotNetPackageXASdkProject()`.

jpobst:
Try removing the explicit reference to the `net45` version.  The `<PackageReference>` should add the reference automagically:

https://github.com/xamarin/xamarin-android/blob/ee6424aefc8e9d86cba44722aecfe08335960d90/tools/apkdiff/apkdiff.csproj#L37

radekdoulik:
> Try removing the explicit reference to the net45 version. The <PackageReference> should add the reference automagically

I think that would still choose net45 over netstandard :-( Let see whether the project dependency would help.

I am going to remove the references from that file soon anyway, like I already did here https://github.com/radekdoulik/apkdiff/blob/3a479dcea2bc35bb0be8b8ff97a71a7fdb2482e7/apkdiff.csproj

radekdoulik:
OK, it still didn't remove the issue, so it is either caused by something else or the dependency didn't work as expected.

To make things worse, there is also conflict with `Newtonsoft.Json` assembly, where `Xamarin.Android.Build.Tasks` project is deleting this assembly in `ILRepacker` on Jenkins.

I think I will try to make apkdiff install in a separate directory to avoid conflicts with other projects.

dellis1972:
@radekdoulik looks like the csproj is not using the LibZipSharpVersion property. Can you change that in this PR?

build-tools/installers/create-installers.targets
```diff
@@ -102,8 +102,14 @@
   </ItemGroup>
   <ItemGroup>
     <_MSBuildFiles Include="$(MSBuildSrcDir)\android-support-multidex.jar" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff.exe" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff.pdb" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\apkdiff.exe" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\apkdiff.pdb" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\Mono.Options.dll" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\apkdiff\Mono.Posix.NETStandard.dll" />
```
pjcollins: I think we can remove this addition, see https://github.com/xamarin/xamarin-android/pull/4267 for context.

tools/apkdiff/apkdiff.csproj
```diff
@@ -53,9 +54,9 @@
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="Mono.Options" Version="6.6.0.161" />
     <PackageReference Include="Mono.Posix.NETStandard" Version="1.0.0" />
```
pjcollins: I think we can remove this `PackageReference`, and the corresponding `Reference` above. See https://github.com/xamarin/xamarin-android/pull/4267 for context.

dellis1972: @pjcollins LibZipSharp uses Mono.Posix.NETStandard. So it should get pulled in automatically...

radekdoulik:
> @radekdoulik looks like the csproj is not using the LibZipSharpVersion property. Can you change that in this PR?

I was, but it got reverted with https://github.com/xamarin/xamarin-android/pull/4246/commits/ee6424aefc8e9d86cba44722aecfe08335960d90, damn :-)

PR #4248, opened 2020-02-11
Fixing acceptable breakages file after set error=true on obsolete items

gugavaro:
We are marking as obsolete error all const fields that already have an enum and related interfaceconsts.

Reference: https://github.com/xamarin/java.interop/pull/568

Additionally this remove the `API Compatibility` job from our build pipeline YAML.  This step has been superseded by the api checker added in https://github.com/xamarin/xamarin-android/pull/3884.


tests/api-compatibility/acceptable-breakages-vReference.txt
```diff
@@ -1271,4 +1271,3782 @@ CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'Androi
...
 CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'Android.Widget.Toolbar.SetCollapseIcon(System.Int32)' changed from '[RegisterAttribute("setCollapseIcon", "(I)V", "GetSetCollapseIcon_IHandler")]' in the contract to '[RegisterAttribute("setCollapseIcon", "(I)V", "GetSetCollapseIcon_IHandler", ApiSince=29)]' in the implementation.
-CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'Android.Widget.ViewFlipper.FlipInterval.get()' changed from '[RegisterAttribute("getFlipInterval", "()I", "GetGetFlipIntervalHandler")]' in the contract to '[RegisterAttribute("getFlipInterval", "()I", "GetGetFlipIntervalHandler", ApiSince=29)]' in the implementation.
\ No newline at end of file
+CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'Android.Widget.ViewFlipper.FlipInterval.get()' changed from '[RegisterAttribute("getFlipInterval", "()I", "GetGetFlipIntervalHandler")]' in the contract to '[RegisterAttribute("getFlipInterval", "()I", "GetGetFlipIntervalHandler", ApiSince=29)]' in the implementation.
```
jonpryor: What changed here?  These appear the same to me.

jonpryor:
Please remove the `external/xamarin-android-api-compatibility` submodule reference as well, to further indicate that it isn't used anymore.

jpobst:
👍  Removed submodule.

jonpryor:
Squash-and-merge "title":

```
Bump to xamarin/Java.Interop/master@8f30933a (#4248)
```

Squash-and-merge "body":

```
Context: https://github.com/xamarin/java.interop/commit/8f30933a89b2ad3259740f21587f50dc8ea1a0e2

Changes: https://github.com/xamarin/java.interop/compare/3226a4b57ad84574a69a151a310b077cfe69ee19...8f30933a89b2ad3259740f21587f50dc8ea1a0e2

  * xamarin/Java.Interop@8f30933: [generator] Mark some Obsolete fields as errors (#568)
  * xamarin/Java.Interop@47201c4: [generator] Change protected members in final class to private (#569)
  * xamarin/Java.Interop@c5815fb: [generator] Be smarter about when members with same names need "new" (#567)
  * xamarin/Java.Interop@bfc0273: [generator] Ensure property setter parameter name is "value" (#566)

This Java.Interop bump includes two changes which "break" API:

  * Certain fields are changed from `[Obsolete]` to
    `[Obsolete(error:true)]`, turning *use* of the field within C#
    code into a CS0619 error.

  * The C# compiler has long warned about the presence of `protected`
    members within `sealed` types, as they cannot be used.  However,
    they were still emitted, resulting in CS0628 warnings when
    compiling `Mono.Android.dll`.

We consider changing `[Obsolete]` to `[Obsolete(error:true)]` to be
acceptable because this is being done to `const` fields, so this
won't break ABI, as there are no IL references to `const` fields, and
because (1) there are "replacement" members which can be used, and
(2) these fields have been obsolete for *years*.

The "problem" here is that `Microsoft.DotNet.ApiCompat.exe`, which
the `<CheckApiCompatibility/>` task uses from
`src/Mono.Android/Mono.Android.targets`, sees every such change as
an API break.  This in and of itself is plausibly sensible, but the
error message *itself* is bananas, e.g.:

	namespace Android.AccessibilityServices {
	  public abstract partial class AccessibilityService : Android.App.Service {
	    [Register ("GESTURE_SWIPE_DOWN", ApiSince = 16)]
	    [Obsolete ("This constant will be removed in the future version. Use Android.AccessibilityServices.AccessibilityGesture enum directly instead of this field.", error: true)]
	    public const Android.AccessibilityServices.AccessibilityGesture GestureSwipeDown = (Android.AccessibilityServices.AccessibilityGesture) 2;
	  }
	}

results in:

	CannotRemoveAttribute : Attribute 'System.ObsoleteAttribute' exists on 'Android.AccessibilityServices.AccessibilityGesture Android.AccessibilityServices.AccessibilityService.GestureSwipeDown' in the contract but not the implementation.

The message implies that `[Obsolete]` was removed, which isn't
strictly true!  `[Obsolete]` is still there!  It's just that the
constructor parameters have changed.

This results in a ginormous change to
`tests/api-compatibility/acceptable-breakages-vReference.txt`.

The "`protected` members within `sealed` types" fix for CS0628 also
resulted in errors from the Tests > API Compatibility job, which runs
via `external/xamarin-android-api-compatibility`.

Given that commit 07e74771 was intended to *replace* the
Tests > API Compatibility job, *remove* the 
`external/xamarin-android-api-compatibility` submodule reference.
We will no longer be using that submodule for API compat checks.

Co-authored-by: Jonathan Pobst <monkey@jpobst.com>
```

tests/api-compatibility/acceptable-breakages-vReference.txt
```diff
@@ -1271,4 +1271,3794 @@ CannotChangeAttribute : Attribute 'Android.Runtime.RegisterAttribute' on 'Androi
...
+CannotRemoveAttribute : Attribute 'System.ObsoleteAttribute' exists on 'Android.App.Admin.PrivateDnsSet Android.App.Admin.DevicePolicyManager.PrivateDnsSetNoError' in the contract but not the implementation.
+CannotRemoveAttribute : Attribute 'System.ObsoleteAttribute' exists on 'Android.App.Admin.ResetPasswordFlags Android.App.Admin.DevicePolicyManager.ResetPasswordDoNotAskCredentialsOnBoot' in the contract but not the implementation.
+CannotRemoveAttribute : Attribute 'System.ObsoleteAttribute' exists on 'Android.App.Admin.ResetPasswordFlags Android.App.Admin.DevicePolicyManager.ResetPasswordRequireEntry' in the contract but not the implementation.
+CannotRemoveAttribute : Attribute 'System.ObsoleteAttribute' exists on 'Android.App.Admin.UserManagementFlags Android.App.Admin.DevicePolicyManager.LeaveAllSystemAppsEnabled' in the contract but not the implementation.
+CannotRemoveAttribute : Attribute 'System.ObsoleteAttribute' exists on 'Android.App.Admin.UserManageme
```
jonpryor: This and the following messages don't make sense to me.  These are *not* the `[Obsolete]` attribute changes above; instead, it sounds like we're no longer emitting some of these methods?  I am confused. :-(

jonpryor: ...oh; these are the "`protected` members within `sealed` classes" category of changes.  `PreferenceScreen` is `sealed`, but `PreferenceScreen.OnCreateRootAdapter()` is `protected`, which makes no sense.

As such, they *do* make sense.

Issue #4241, opened 2020-02-10
Build is using deprecated CredScan v1 warning


akoeplinger:
Just saw this on a PR, thought I'd file an issue:

> ##[warning]This build is running CredScan v1, which is on path to deprecation in the coming months. Please update this build definition to run CredScan v2, by selecting 'CredScan V2' in the 'Tool Major Version' dropdown in the build task UI (toolMajorVersion: V2 in yaml).

/cc @pjcollins

PR #4250, opened 2020-02-11
[ci] Bump credscan and artifact task versions

pjcollins:
Fixes: https://github.com/xamarin/xamarin-android/issues/4241
Context: https://github.com/xamarin/yaml-templates/commit/fc7eb6822ea92ee000258b6fff71e80b8d1022f5, https://github.com/xamarin/yaml-templates/commit/af29b0c60164f883bc597544a178fa6cd53eae9e

Bumps our static analysis template to a new version that will allow us
to specify the version of the `credscan` tool to use. This template will
use `V2` by default, but can be overridden as needed:

```yaml
    - template: security\xa-static-analysis\v2.yml@yaml
      parameters:
        credScanSuppressionsFile: $(System.DefaultWorkingDirectory)\build-tools\automation\CredScanSuppressions.json
        toolMajorVersion: V1
```

Artifact publishing and downloading tasks have also been bumped to their
latest versions.

### DIFF
--- a/build-tools/automation/CredScanSuppressions.json
+++ b/build-tools/automation/CredScanSuppressions.json
@@ -4,6 +4,34 @@
         {
             "file": "\\src\\Xamarin.Android.Build.Tasks\\Tests\\Xamarin.ProjectTools\\Resources\\Base\\test.keystore",
             "_justification": "Dummy keystore file used for testing."
+        },
+        {
+            "file": "tests\\MSBuildDeviceIntegration\\Tests\\InstallTests.cs",
+            "_justification": "Password of the dummy keystore file used only for testing."
+        },
+        {
+            "file": "external\\android-api-docs\\docs\\Mono.Android\\en\\Android.Webkit\\HttpAuthHandler.xml",
+            "_justification": "Android API documentation, does not contain a password."
+        },
+        {
+            "file": "external\\android-api-docs\\docs\\Mono.Android\\en\\Android.Webkit\\WebViewDatabase.xml",
+            "_justification": "Android API documentation, does not contain a password."
+        },
+        {
+            "file": "external\\android-api-docs\\docs\\Mono.Android\\en\\Java.Security\\KeyStore+PasswordProtection.xml",
+            "_justification": "Android API documentation, does not contain a password."
+        },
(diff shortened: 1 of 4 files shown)


jonpryor:
How do we know if it works?

Related: consider the **Code Analysis** > **CredScan** > [**Create Security Analysis Report**](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3462743&view=logs&j=d3f8649d-db33-5c22-064d-51efdff82f35&t=37444108-2d0c-5c51-b8cb-04ce0fb93d42) output.  It contains the word `error` in various contexts that make me wonder if we're not actually detecting errors?  The step is "green", yet:

```
external\android-api-docs\docs\Mono.Android\en\Javax.Security.Auth.Callback\PasswordCallback.xml(233,): error : SecretinFile : {{Searcher}}CSCAN-GENERAL0060
```

...which is probably an *ignorable* error -- it's in documentation! -- but it still says "error", yet the overall step is "green".

Additionally, isn't the Security Analysis Report supposed to check for known security issues, such as CVEs in dependencies?  If so, why didn't it report issues in SQLite?  (See PR #4247.). (Or did it, and we didn't notice?)

pjcollins:
I'll see if we can get the step/job to error when issues are detected. It looks like the updated version of the tool is finding more potential credentials, and I think we'll need a corresponding update to `CredScanSuppressions.json` (or perhaps our sources in some cases).

`Credscan` is only responsible for finding potential secrets/tokens in sources however. For other security issues (such as the SQLite mentioned above), there is a [separate report created by component governance](https://devdiv.visualstudio.com/DevDiv/_componentGovernance/112013?_a=alerts&typeId=538045&alerts-view-option=active) that unfortunately also doesn't report directly back to the build pipeline. It also doesn't seem to contain any mention of the particular SQLite issue mentioned above... There are likely other static analysis tools that we could look to adopt in the future to help improve this story.

pjcollins:
Here's an example of what a Credscan failure will now look like, from before the issue suppression file was updated:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3465426&view=results

I think this should be good to go now.

PR #4253, opened 2020-02-12
[Xamarin.Android.Build.Tasks] Update multidex warnings & make localizable

brendanzagaeski:

[3757be20][0] moved the XA4305 messages for `<CompileToDalvik/>` into
the `.resx` file, but it overlooked the uses of XA4305 in
`<CreateMultiDexMainDexClassList/>` and `<R8/>`.

The original uses of XA4305 in `<CreateMultiDexMainDexClassList/>` and
`<R8/>` were a little confusing because the `@(MultiDexMainDexList)`
MSBuild item corresponds to the `CustomMainDexListFiles` property in the
tasks, not the `MultiDexMainDexListFile` property.

To reduce the potential for confusion between `@(MultiDexMainDexList)`
and `MultiDexMainDexListFile` and to separate them in telemetry, give
the warning message related to `@(MultiDexMainDexList)` its own new
XA4309 warning code, reword the other XA4305 messages, and update the
documentation accordingly.

Along the way, move the messages for the new XA4309 code and the other
remaining multidex message code XA4306 into the `.resx` file so that
they are localizable.

[0]: https://github.com/xamarin/xamarin-android/commit/3757be20ba9fd1f34166f812daf9963f5d6dd22b

### DIFF
--- a/Documentation/guides/messages/xa4305.md
+++ b/Documentation/guides/messages/xa4305.md
@@ -8,22 +8,17 @@ ms.date: 02/11/2020
 ## Example messages
 
 ```
-Multidex is enabled, but the path of the merged keep file was not specified.
+Multidex is enabled, but `$(_AndroidMainDexListFile)` is empty.
 ```
 
 ```
-Multidex is enabled, but the merged keep file 'obj\Release\90\multidex.keep' does not exist.
+Multidex is enabled, but the `$(_AndroidMainDexListFile)` 'obj\Release\90\multidex.keep' does not exist.
 ```
-
-## Issue
-
-The path of the merged *multidex.keep* file for the `CompileToDalvik` or `R8`
-MSBuild task was not specified or the file was not found on disk.
 
 ## Solution
 
-The build process should automatically configure and populate the merged
-*multidex.keep* file, so please consider submitting a [bug][bug] if you are
-getting this warning under normal circumstances.
+The build process should automatically configure and populate the file for the
+`$(_AndroidMainDexListFile)` property, so please consider submitting a
+[bug][bug] if you are getting this warning under normal circumstances.
 
 [bug]: https://github.com/xamarin/xamarin-android/wiki/Submitting-Bugs,-Feature-Requests,-and-Pull-Requests
--- a/Documentation/guides/messages/xa4305.md
+++ b/Documentation/guides/messages/xa4305.md
@@ -1,28 +1,29 @@
 ---
 title: Xamarin.Android warning XA4305
 description: XA4305 warning code
-ms.date: 10/30/2018
+ms.date: 02/11/2020
 ---
 # Xamarin.Android warning XA4305
 
+## Example messages
+
+```
+Multidex is enabled, but the path of the merged keep file was not specified.
+```
+
+```
+Multidex is enabled, but the merged keep file 'obj\Release\90\multidex.keep' does not exist.
+```
+
 ## Issue
 
-The `CreateMultiDexMainDexClassList`, `CompileToDalvik` or `R8`
-MSBuild task encountered a *multidex.keep* file that was not found on
-disk. You can customize multidex settings for your Xamarin.Android
-application by adding files with the `MultiDexMainDexList` build item,
-which are combined into a final *multidex.keep* file.
-
-To learn more about multidex and how it relates to Android
-development, see the [Android documentation][android].
+The path of the merged *multidex.keep* file for the `CompileToDalvik` or `R8`
+MSBuild task was not specified or the file was not found on disk.
 
 ## Solution
 
-Verify you are not declaring a `MultiDexMainDexList` build item that
-does not exist.
+The build process should automatically configure and populate the merged
+*multidex.keep* file, so please consider submitting a [bug][bug] if you are
+getting this warning under normal circumstances.
 
-Consider submitting a [bug][bug] if you are getting this warning under
-normal circumstances.
-
-[android]: https://developer.android.com/studio/build/multidex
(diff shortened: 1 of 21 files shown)


src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -230,17 +230,24 @@
 {0} - Comma-separated list of the ignored library file names.</comment>
   </data>
   <data name="XA4305" xml:space="preserve">
-    <value>MultiDex is enabled, but '{0}' was not specified.</value>
-    <comment>The following are literal names and should not be translated: MultiDex
-{0} - The MSBuild property name.</comment>
+    <value>Multidex is enabled, but the path of the merged keep file was not specified.</value>
```
brendanzagaeski: I debated a bit if "path of the merged keep file" make sense and will work OK for translation to other languages. I think it's not bad, but I'm open to suggestions. Maybe "path for the generated keep file" would be better because R8 doesn't "merge" keep files anymore.

The idea behind "keep file" is that the [multidex documentation](https://developer.android.com/studio/build/multidex) mentions a `multiDexKeepFile` property for Gradle and a "Keep Options" section of the Proguard documentation. On the other hand, the documentation does not include the exact string "keep file" (with a space).

The idea behind "merged" is to give a hint that this file isn't something users are expected to specify or create on their own (unlike `@(MultiDexMainDexList)`).

jonpryor: How does one specify "the path to the merged keep file"?  If this message is to be actionable, a fix should be provided.  Is this some MSBuild property that must be overridden?  Something else?

brendanzagaeski: The tricky thing about these XA4305 messages in terms of actionability is that they should not appear in user builds *ever*, unless there is a bug in Xamarin.Android or the user is overwriting the "private" (undocumented and underscore-prefixed) `$(_AndroidMainDexListFile)` MSBuild property.

So the main recommended action would be to submit a bug report, but I didn't find any existing Xamarin.Android messages that suggest submitting bug reports directly as part of the message. At the moment, this PR lists that action instead in the documentation page for the message:

```
The build process should automatically configure and populate the merged
*multidex.keep* file, so please consider submitting a [bug][bug] if you are
getting this warning under normal circumstances.
```

Some other similar "bug-indicating" warnings that don't currently include a suggested action in the warning message are:

XA0107:

https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs#L297-L299

https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx#L133-L135

https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/Documentation/guides/messages/xa0107.md#L23-L24

XA0108:

https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/src/Xamarin.Android.Build.Tasks/Tasks/Lint.cs#L384

https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/Documentation/guides/messages/xa0108.md#L15-L16

XA5302:

https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/src/Xamarin.Android.Build.Tasks/Tasks/WriteLockFile.cs#L28

https://github.com/xamarin/xamarin-android/blob/cf6d51e4531ace51188d7920e2be1ef47ad9197c/Documentation/guides/messages/xa5302.md#L22-L23

brendanzagaeski:
**Changes in latest commit:**

Replace "merged keep file" with `$(_AndroidMainDexListFile)`.

Given that the XA4305 messages should only appear if there's a bug in Xamarin.Android or if the user is doing something unsupported like overriding `$(_AndroidMainDexListFile)` or invoking the `<R8/>` or `<CompileToDalvik/>` task using their own custom targets, maybe a decent compromise for now is to put the name `$(_AndroidMainDexListFile)` directly in the messages.

The main advantage of using the MSBuild property name rather than the property name from within the tasks (`MultiDexMainDexListFile`) is that it better matches other messages, like the other multidex messages XA4306 and XA4309.

For supported scenarios, where these messages indicate a bug in Xamarin.Android, this keeps the messages at least as actionable as they were originally. And it allows removing the debatable terms "merged" and "keep file."

The larger question of whether it would be more appropriate to throw an exception or include wording about submitting a bug can then perhaps be postponed and revisited in the future if a similar case comes up maybe based on user feedback or telemetry.

Documentation/guides/messages/xa4305.md
```diff
@@ -1,28 +1,24 @@
...
-application by adding files with the `MultiDexMainDexList` build item,
-which are combined into a final *multidex.keep* file.
+```
+Multidex is enabled, but `$(_AndroidMainDexListFile)` is empty.
+```
 
-To learn more about multidex and how it relates to Android
-development, see the [Android documentation][android].
+```
+Multidex is enabled, but the `$(_AndroidMainDexListFile)` 'obj\Release\90\multidex.keep' does not exist.
```
jonpryor: I would suggest inserting "file" between `$(_AndroidMainDexListFile)` and `obj\...`, e.g.

```
Multidex is enabled, but the `$(_AndroidMainDexListFile)` file 'obj\Release\90\multidex.keep' does not exist.
```

PR #4260, opened 2020-02-12
[linker] remove unneeded string.Format calls

jonathanpeppers:
I was using the Mono profiler for our build, and I noticed:

    Allocation summary
        Bytes      Count  Average Type name
    180765632    2142803       84 System.String
     53218472 bytes from:
            MonoDroid.Tuner.FixAbstractMethodsStep:FixAbstractMethods (Mono.Cecil.AssemblyDefinition)
            MonoDroid.Tuner.FixAbstractMethodsStep:FixAbstractMethodsUnconditional (Mono.Cecil.AssemblyDefinition)
            MonoDroid.Tuner.FixAbstractMethodsStep:FixAbstractMethods (Mono.Cecil.TypeDefinition)
            MonoDroid.Tuner.FixAbstractMethodsStep:HaveSameSignature (Mono.Cecil.TypeReference,Mono.Cecil.MethodDefinition,Mono.Cecil.MethodDefinition)
            (wrapper managed-to-native) string:FastAllocateString (int)

~53MB of string from this one method is a lot!

I found that one of these `string.Format` calls are running for almost
every method on every interface in every assembly:

    (string.Format ("{0}.{1}", iface.FullName, iMethod.Name) != tMethod.Name) :
    (string.Format ("{0}.{1}.{2}", iMethod.DeclaringType.DeclaringType, iface.Name, iMethod.Name) != tMethod.Name))

Even the simplest non-matching cases would call `string.Format`:

    iMethod.Name == "Foo"
    tMethod.Name == "Bar"

In all of these cases...

* Class implements interface implicitly
* Class implements interface explicitly
* Class implements abstract class

We found that the `IsInOverrides` check should work or merely compare
`iMethod.Name` and `tMethod.Name`. We don't seem to need the
`string.Format` calls at all?

## Results ##

An initial build of the Xamarin.Forms integration project on
Windows/.NET framework:

    Before:
    796 ms  LinkAssembliesNoShrink                     1 calls
    After:
    767 ms  LinkAssembliesNoShrink                     1 calls

The same project on macOS/Mono (slightly slower machine, too):

    Before:
    1341 ms  LinkAssembliesNoShrink                     1 calls
    After:
    1025 ms  LinkAssembliesNoShrink                     1 calls

String allocations are way better, too:

    Before:
    Allocation summary
        Bytes      Count  Average Type name
    180765632    2142803       84 System.String
    After:
    Allocation summary
        Bytes      Count  Average Type name
    127736832    1652070       77 System.String

Shows 53,028,800 bytes saved.

    Before:
    53218472 bytes from:
        MonoDroid.Tuner.FixAbstractMethodsStep:FixAbstractMethods (Mono.Cecil.AssemblyDefinition)
        MonoDroid.Tuner.FixAbstractMethodsStep:FixAbstractMethodsUnconditional (Mono.Cecil.AssemblyDefinition)
        MonoDroid.Tuner.FixAbstractMethodsStep:FixAbstractMethods (Mono.Cecil.TypeDefinition)
        MonoDroid.Tuner.FixAbstractMethodsStep:HaveSameSignature (Mono.Cecil.TypeReference,Mono.Cecil.MethodDefinition,Mono.Cecil.MethodDefinition)
    After:
     1933624 bytes from:
        MonoDroid.Tuner.FixAbstractMethodsStep:FixAbstractMethodsUnconditional (Mono.Cecil.AssemblyDefinition)
        MonoDroid.Tuner.FixAbstractMethodsStep:FixAbstractMethods (Mono.Cecil.TypeDefinition)
        MonoDroid.Tuner.FixAbstractMethodsStep:HaveSameSignature (Mono.Cecil.TypeReference,Mono.Cecil.MethodDefinition,Mono.Cecil.MethodDefinition)
        MonoDroid.Tuner.FixAbstractMethodsStep:HaveSameMethodName (Mono.Cecil.TypeReference,Mono.Cecil.MethodDefinition,Mono.Cecil.MethodDefinition)

Shows 51,284,848 bytes saved.

I don't know about the discrepancy between these two sets of numbers.
But I think this is roughly ~50MB less string allocations.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/LinkerTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/LinkerTests.cs
@@ -70,6 +70,69 @@ namespace Xamarin.Android.Build.Tests
 			}
 		}
 
+		[Test]
+		public void FixAbstractMethodsStep_Explicit ()
+		{
+			var path = Path.Combine (Path.GetFullPath (XABuildPaths.TestOutputDirectory), "temp", TestName);
+			var step = new FixAbstractMethodsStep ();
+			var pipeline = new Pipeline ();
+
+			Directory.CreateDirectory (path);
+
+			using (var context = new LinkContext (pipeline)) {
+
+				context.Resolver.AddSearchDirectory (path);
+
+				var myAssemblyPath = Path.Combine (path, "MyAssembly.dll");
+
+				using (var android = CreateFauxMonoAndroidAssembly ()) {
+					android.Write (Path.Combine (path, "Mono.Android.dll"));
+					CreateExplicitInterface (myAssemblyPath, android);
+				}
+
+				using (var assm = context.Resolve (myAssemblyPath)) {
+					step.Process (context);
+
+					var impl = assm.MainModule.GetType ("MyNamespace.MyClass");
+					Assert.AreEqual (2, impl.Methods.Count, "MyClass should contain 2 methods");
+					var method = impl.Methods.FirstOrDefault (m => m.Name == "MyNamespace.IMyInterface.MyMethod");
+					Assert.IsNotNull (method, "MyNamespace.IMyInterface.MyMethod should exist");
+					method = impl.Methods.FirstOrDefault (m => m.Name == "MyMissingMethod");
+					Assert.IsNotNull (method, "MyMissingMethod should exist");
+				}
+			}
+
+			Directory.Delete (path, true);
+		}
+
+		static void CreateExplicitInterface (string assemblyPath, AssemblyDefinition android)
+		{
+			using (var assm = AssemblyDefinition.CreateAssembly (new AssemblyNameDefinition ("NestedIFaceTest", new Version ()), "NestedIFaceTest", ModuleKind.Dll)) {
+				var void_type = assm.MainModule.ImportReference (typeof (void));
+
+				// Create interface
+				var iface = new TypeDefinition ("MyNamespace", "IMyInterface", TypeAttributes.Interface);
+
+				var iface_method = new MethodDefinition ("MyMethod", MethodAttributes.Abstract, void_type);
+				iface.Methods.Add (iface_method);
+				iface.Methods.Add (new MethodDefinition ("MyMissingMethod", MethodAttributes.Abstract, void_type));
+
+				assm.MainModule.Types.Add (iface);
+
+				// Create implementing class
+				var jlo = assm.MainModule.Import (android.MainModule.GetType ("Java.Lang.Object"));
+				var impl = new TypeDefinition ("MyNamespace", "MyClass", TypeAttributes.Public, jlo);
+				impl.Interfaces.Add (new InterfaceImplementation (iface));
+
+				var explicit_method = new MethodDefinition ("MyNamespace.IMyInterface.MyMethod", MethodAttributes.Abstract, void_type);
+				explicit_method.Overrides.Add (new MethodReference (iface_method.Name, void_type, iface));
+				impl.Methods.Add (explicit_method);
+
+				assm.MainModule.Types.Add (impl);
+				assm.Write (assemblyPath);
+			}
+		}
+
 		static AssemblyDefinition CreateFauxMonoAndroidAssembly ()
 		{
 			var assm = AssemblyDefinition.CreateAssembly (new AssemblyNameDefinition ("Mono.Android", new Version ()), "DimTest", ModuleKind.Dll);
--- a/Documentation/release-notes/4260.md
+++ b/Documentation/release-notes/4260.md
@@ -1,0 +1,24 @@
+### Linker Behavior Change
+
+Xamarin.Android has to consolidate changes to the Android APIs and how
+they affect C# types. If you consider an example such as:
+
+* API 28 has an `IFoo` interface in `Mono.Android.dll` for `v9.0`.
+* C# code implements `IFoo` in class `Bar`.
+* API 29 adds another method to `IFoo` in `Mono.Android.dll` for
+  `v10.0`.
+* Apps built targeting API 30 / `v10.0` will fix up `Bar` to implement
+  the new method, but throw `Java.Lang.AbstractMethodError` if it was
+  called.
+
+If the linker didn't make this change, you would instead hit a crash
+when trying to create an instance of type `Bar` such as:
+
+    System.TypeLoadException: VTable setup of type 'Bar' failed.
+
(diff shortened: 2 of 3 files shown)


jonpryor:
The short circuit and reduced memory usage is nice and all but...are we *still* doing too much?

The point behind PR#861 and the above `IsInOverrides()` invocation was to see if `tMethod` was supposed to override the interface method `iMethod`.

If that *fails*, then we fallback to...a whole mess of string work.

Why do we need to do that whole mess of string work *at all*?  Why do we *need* `HaveSameMethodName()`?  Why do we need to compare parameter types & return types?

Why isn't `IsInOverrides()` *by itself* "enough"?  What test case am I not thinking about?

(With an added point that method names being `{iface.FullName}.{iMethod.Name}` or `{declaringType}.{iface.Name}.{iMethod.Name}` is really only true for C# code, and IIRC isn't true for VB.NET code or F# code, so if we can avoid string comparisons *entirely* that would be Really Nice...)

@radekdoulik?

jonpryor:
More on the "F# names don't match the C# convention", consider this F# code which implements the [`IServiceProvider.GetService(Type)` method](https://docs.microsoft.com/en-us/dotnet/api/system.iserviceprovider.getservice?view=netframework-4.8):

```fsharp
open System;

type MyServiceProvider() =
  interface IServiceProvider with
    member this.GetService(serviceType : Type): System.Object = null
```

When using "Microsoft (R) F# Compiler version 10.2.3 for F# 4.5", the `monodic` output shows:

```
  .class nested public auto ansi serializable MyServiceProvider
  	extends [mscorlib]System.Object
  	implements [mscorlib]System.IServiceProvider  {
    .custom instance void class [FSharp.Core]Microsoft.FSharp.Core.CompilationMappingAttribute::'.ctor'(valuetype [FSharp.Core]Microsoft.FSharp.Core.SourceConstructFlags) =  (01 00 03 00 00 00 00 00 ) // ........

    // method line 2
    .method private virtual hidebysig newslot 
           instance default object 'System-IServiceProvider-GetService' (class [mscorlib]System.Type serviceType)  cil managed 
    {
        // Method begins at RVA 0x205c
	.override class [mscorlib]System.IServiceProvider::GetService
	// Code size 2 (0x2)
	.maxstack 8
	IL_0000:  ldnull 
	IL_0001:  ret 
    } // end of method MyServiceProvider::System-IServiceProvider-GetService

  } // end of class MyServiceProvider
```

That is, the method name which implements `IServiceProvider.GetService` is named `System-IServiceProvider-GetService`, which uses *hyphens* (dashes?  minus signs?) within the name, which is something we are *not* checking for.

Nor should we!  The IL explicitly states which interface method is implemented!

```
	.override class [mscorlib]System.IServiceProvider::GetService
```

radekdoulik:
> Why isn't IsInOverrides() by itself "enough"?

IIRC, the `.override` is not always present in IL.

jonpryor:
> IIRC, the `.override` is not always present in IL.

...and this is the point where I start burbling like a looney.

In fact, *virtual methods overrides* do *not* use `.override`!

```csharp
class App {
    public override string ToString ()
    {
        return "Hello, App!";
    }
}
```

disassembles (for `csc /optimize+` output) to:

```cil
    // method line 2
    .method public virtual hidebysig 
           instance default string ToString ()  cil managed 
    {
        // Method begins at RVA 0x2061
        // Code size 6 (0x6)
        .maxstack 8
        IL_0000:  ldstr "Hello, App!"
        IL_0005:  ret 
    } // end of method App::ToString
```

which in no way specifies what method it's overriding.

*This* is why we need to fallback to signature-based matching: for *`virtual`* methods that are overridden in derived types.  Interface methods get the "easy" path of `.override`.

😭

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/LinkerTests.cs
```diff
@@ -70,6 +70,75 @@ static void CreateAbstractIfaceImplementation (string assemblyPath, AssemblyDefi
...
+					Assert.AreNotEqual (0, method.Body.Instructions.Count, "MyMissingMethod should *not* be an empty method");
+				}
+			}
+
+			Directory.Delete (path, true);
+		}
+
+		static void CreateNestedInterface (string assemblyPath, AssemblyDefinition android)
+		{
+			using (var assm = AssemblyDefinition.CreateAssembly (new AssemblyNameDefinition ("NestedIFaceTest", new Version ()), "NestedIFaceTest", ModuleKind.Dll)) {
```
jonpryor: Possibly "silly" question, but why are we using System.Reflection.Emit to create `MyAssembly.dll` instead of using some C# code?

(Yes, `FixAbstractMethodsStep_NestedTypes()` does that too, but now I wonder the same thing as well!)

jonpryor:
> *This* is why we need to fallback to signature-based matching...

*Still*, that doesn't quite explain why we need the previous `string.Format()`-based logic.  Sure, virtual methods don't have `.override`, but virtual methods also have the *same name*.

The only instances -- *that I know of* -- where a method name would be `$"{iface.FullName}.{iMethod.Name}"` or `$"{declaringType}.{iface.Name}.{iMethod.Name}"` is when the method is part of an explicit interface implementation, in which case *it most certainly **should** have `.override`!*

Which means the "original" question is still unanswered:

> Why do we need to do that whole mess of string work at all? Why do we need `HaveSameMethodName()`?

An actual test case would be useful here. 😕

jonathanpeppers:
@jonpryor I think I did what we needed here:

1. Just check the method names and that's it.
2. Document the release notes.
3. Make the test case using an explicit interface implementation.

I also experimented with using C# source code for these tests, such as:

https://github.com/xamarin/java.interop/blob/master/tests/generator-Tests/Integration-Tests/Compiler.cs

I encountered a lot of issues trying to make it work (beyond the hacks in java.interop). It might be worth revisiting later I think.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/LinkerTests.cs
jonathanpeppers: I tried using CodeDom... It seems like sacrificing too many chickens, it required me adding global alias for System.dll:

https://github.com/jonathanpeppers/xamarin-android/commit/b259e62b144b4ff9439b0fb4878d32b388da638e

I didn't get to the point of how to "add" a new method to an interface of an existing assembly. I guess it would overwrite `Mono.Android.dll` with a new one?

Using only Mono.Cecil it's easy to fabricate the types to make `FixAbstractMethods` do its work.

PR #4265, opened 2020-02-13
[Xamarin.Android.Build.Tasks] don't use designtime/Resource.designer.cs

jonathanpeppers:
With .NET 5 coming, we need to ensure that Xamarin.Android can work in
the new project system in VS Windows.

You can test this with long-form (our current) projects by:

1. Add a `Directory.Build.props` file in the root of the project with
   the following content:
```xml
    <Project>
      <PropertyGroup>
        <Configurations>Debug;Release</Configurations>
        <Platforms>AnyCPU</Platforms>
      </PropertyGroup>
    </Project>
```
2. Changes the .sln entry for the Xamarin project to use the following
   GUID: `{9A19103F-16F7-4668-BE54-9A1E7A4F7556}` (instead of
   `{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}`)

After doing this, things will be buggy, but testable. I created a new
Xamarin.Android app project and removed all NuGet packages.
`<PackageReference/>` in particular causes issues.

The first problem I found was that the new project system saw two
`Resource.designer.cs` files.

Initial design-time builds were OK, but as soon as I did a full build
that generated `Resources\Resource.designer.cs` there were always two
in the project.

Note that I was still using a long-form project here no `**\*.cs`
wildcards were involved, it just seemed the `<Compile Remove="..."/>`
wasn't working.

I think we can make this work going forward by:

* DTBs will no longer generate
  `obj\Debug\designtime\Resource.designer.cs`, but will generate the
  main `Resources\Resource.designer.cs`.

* Builds will *have* to overwrite anything the DTB generates. I added
  a test checking this.

* A side-benefit, VS for Mac can now use the managed resource parser!

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/DesignerTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/DesignerTests.cs
@@ -2,7 +2,6 @@ using System;
 using System.IO;
 using System.Linq;
 using System.Text;
-using System.Linq;
 using System.Xml.Linq;
 using NUnit.Framework;
 using Xamarin.ProjectTools;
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
I will need to manually test the builds from this in both IDEs.

jonathanpeppers:
I haven't seen if we actually need this in .NET 6, closing.

PR #4266, opened 2020-02-13
[Mono.Android] Fix some warnings from dead code.

jpobst:
Currently, building `Mono.Android.dll` for `android-29` on `master` generates 26 warning (x2 TFV)s.  This large number of existing warnings makes it hard to notice any potential issues that show up as new warnings.  Because of this we would like to reduce (or eliminate!) the number of warnings generated.

This PR reduces the number of warnings for `android-29` down to 2 (x2 TFV):

```
 obj/Release/monoandroid10/android-29/mcw/Android.Media.RemoteControlClient.cs(1119,17): warning CS0169: The field 'RemoteControlClient.weak_implementor_SetOnGetPlaybackPositionListener' is never used [/Users/builder/azdo/_work/1/s/xamarin-android/src/Mono.Android/Mono.Android.csproj]
 obj/Release/monoandroid10/android-29/mcw/Android.Views.View.cs(21419,17): warning CS0169: The field 'View.weak_implementor_SetOnApplyWindowInsetsListener' is never used [/Users/builder/azdo/_work/1/s/xamarin-android/src/Mono.Android/Mono.Android.csproj]
```

These warnings probably will require some `generator` investigation and fixes, so we'll prioritize them later.


src/Mono.Android/Java.Lang/Object.cs
```diff
@@ -16,10 +16,8 @@ public partial class Object : IDisposable, IJavaObject, IJavaObjectEx
 #endif  // JAVA_INTEROP
 	{
 		IntPtr key_handle;
-		IntPtr weak_handle;
```
jonpryor: These fields can't be removed, as they're used from `monodroid-glue.cc`.  Ditto in `Throwable.cs`.

src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -296,12 +294,8 @@ internal static void PropagateUncaughtException (IntPtr env, IntPtr javaThreadPt
 		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
 		extern static void _monodroid_gc_wait_for_bridge_processing ();
 
-		static volatile bool BridgeProcessing; // = false
```
jonpryor: Should not be removed: https://github.com/xamarin/xamarin-android/blob/0ddcc3fbd8458a4749753b08dd9a081c1e6036c4/src/monodroid/jni/monodroid-glue.cc#L982

PR #4267, opened 2020-02-13
[installers] Remove Mono.Posix.NETStandard and add missing symbols

pjcollins:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3468220&view=logs&j=fd490c07-0b22-5182-fac9-6d67fe1e939b&t=da4239c9-cabf-50a7-e7ce-52be5c6d9883

A recent PR updating the Xamarin.Android version in Visual Studio
reported a new failure in SymbolCheck. The two files which were reported
to be missing symbols are:

    $msbuild\xamarin\android\mono.posix.netstandard.dll
    xamarin.android.sdk.dll

Some thorough sleuthing by @jonpryor indicated that we likely don't need
to be redistributing `Mono.Posix.NETStandard.dll` in our installers at
all. Various unecessary package references have been removed, and the
set of files required by our installers has been updated to reflect this
discovery. This removal should also appease the Symbol Check process.

We aren't sure why Symbol Check is now deciding to complain about
`Xamarin.Android.Sdk.dll`, but a small fix to include the corresponding
.pdb in the .vsix should be harmless.

Finally, the `Xamarin.LibZipSharp` NuGet was updated to contain symbol
files, so we can include the correct .pdb file for this assembly and
stop whitelisting it.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
@@ -337,12 +337,12 @@
     </ItemGroup>
   </Target>
 
-    <!-- Copy extra package ref content for our installers. '$(Pkg*)' props are set during NuGet restore when GeneratePathProperty="true" -->
+  <!-- Copy package ref symbols for our installers. '$(Pkg*)' props are set during NuGet restore when GeneratePathProperty="true" -->
   <Target Name="_CopyExtraPackageContent"
-      Inputs="$(PkgMono_Posix_NETStandard)\runtimes\osx\lib\netstandard2.0\Mono.Posix.NETStandard.dll;$(PkgXamarin_Build_AsyncTask)\lib\netstandard2.0\Xamarin.Build.AsyncTask.pdb"
-      Outputs="$(OutputPath)\Mono.Posix.NETStandard.dll;$(OutputPath)\Xamarin.Build.AsyncTask.pdb" >
+      Inputs="$(PkgXamarin_LibZipSharp)\lib\$(TargetFrameworkNETStandard)\libZipSharp.pdb;$(PkgXamarin_Build_AsyncTask)\lib\$(TargetFrameworkNETStandard)\Xamarin.Build.AsyncTask.pdb"
+      Outputs="$(OutputPath)\libZipSharp.pdb;$(OutputPath)\Xamarin.Build.AsyncTask.pdb" >
     <Copy
-        SourceFiles="$(PkgMono_Posix_NETStandard)\runtimes\osx\lib\netstandard2.0\Mono.Posix.NETStandard.dll;$(PkgXamarin_Build_AsyncTask)\lib\netstandard2.0\Xamarin.Build.AsyncTask.pdb"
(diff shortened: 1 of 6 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
```diff
@@ -337,12 +337,12 @@
...
+  <!-- Copy package ref symbols for our installers. '$(Pkg*)' props are set during NuGet restore when GeneratePathProperty="true" -->
   <Target Name="_CopyExtraPackageContent"
-      Inputs="$(PkgMono_Posix_NETStandard)\runtimes\osx\lib\netstandard2.0\Mono.Posix.NETStandard.dll;$(PkgXamarin_Build_AsyncTask)\lib\netstandard2.0\Xamarin.Build.AsyncTask.pdb"
-      Outputs="$(OutputPath)\Mono.Posix.NETStandard.dll;$(OutputPath)\Xamarin.Build.AsyncTask.pdb" >
+      Inputs="$(PkgXamarin_LibZipSharp)\lib\$(TargetFrameworkNETStandard)\libZipSharp.pdb;$(PkgXamarin_Build_AsyncTask)\lib\$(TargetFrameworkNETStandard)\Xamarin.Build.AsyncTask.pdb"
+      Outputs="$(OutputPath)\libZipSharp.pdb;$(OutputPath)\Xamarin.Build.AsyncTask.pdb" >
     <Copy
-        SourceFiles="$(PkgMono_Posix_NETStandard)\runtimes\osx\lib\netstandard2.0\Mono.Posix.NETStandard.dll;$(PkgXamarin_Build_AsyncTask)\lib\netstandard2.0\Xamarin.Build.AsyncTask.pdb"
+        SourceFiles="$(PkgXamarin_LibZipSharp)\lib\$(TargetFrameworkNETStandard)\libZipSharp.pdb;$(PkgXamarin_Build_AsyncTask)\lib\$(TargetFrameworkNETStandard)\Xamarin.Build.AsyncTask.pdb"
```
jonathanpeppers: Are these symbols not being copied to `$(OutputPath)` already? If `libZipSharp.dll` makes it there, I would think `libZipSharp.pdb` would be there, too?

I'm wondering if we can delete this entire target.

pjcollins: Short form style projects do not copy NuGet assets by default, but we override this behavior with https://github.com/xamarin/xamarin-android/blob/master/Directory.Build.props#L12 

From what I've seen in local testing, this property doesn't extend to symbol files. There's a long discussion about the lack of symbol copy functionality in the new project system in https://github.com/dotnet/sdk/issues/1458. As far as I can tell such functionality still doesn't exist. I think we have a couple options for the NuGets we own that would allow us to remove this hack in the future:

1) Update the NuGets we own to generate symbols in a `.snupkg` file: https://docs.microsoft.com/en-us/nuget/create-packages/symbol-packages-snupkg.
2) Include symbols in the assembly? - https://github.com/dotnet/sdk/issues/1458#issuecomment-576961024

PR #4269, opened 2020-02-14
[Xamarin.Android.Build.Tests] Add SmokeTests category

pjcollins:
Introduces a new `SmokeTests` category to `Xamarin.Android.Build.Tests`
as part of a broader effort to help split up our test execution and
improve turnaround times for our Azure Pipelines builds. My selection
of tests to add to this category was somewhat arbitrary, but it includes
a handful of build, rebuild, and incremental build scenarios, as well as
multiple project configurations (including AOT and Bundle Assemblies),
and native/java build actions.

We've also been seeing a somewhat rare case where the VS instance on our
Windows machines becomes unlaunchable and the `MSBuild` task falls back
to a very old version, breaking the build. I've attempted to fix this by
updating our script that reverts the version of XA installed with VS to
first try to resume a partial visual studio installation on disk.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/Aapt2Tests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/Aapt2Tests.cs
@@ -30,6 +30,7 @@ namespace Xamarin.Android.Build.Tests
 		}
 
 		[Test]
+		[Category ("SmokeTests")]
 		public void Aapt2Link ()
 		{
 			var path = Path.Combine (Root, "temp", "Aapt2Link");
@@ -84,6 +85,7 @@ namespace Xamarin.Android.Build.Tests
 		}
 
 		[Test]
+		[Category ("SmokeTests")]
 		public void Aapt2Compile ()
 		{
 			var path = Path.Combine (Root, "temp", "Aapt2Compile");
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/Aapt2Tests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/Aapt2Tests.cs
@@ -30,7 +30,6 @@ namespace Xamarin.Android.Build.Tests
 		}
 
 		[Test]
-		[Category ("SmokeTests")]
 		public void Aapt2Link ()
 		{
 			var path = Path.Combine (Root, "temp", "Aapt2Link");
@@ -85,7 +84,6 @@ namespace Xamarin.Android.Build.Tests
 		}
 
 		[Test]
-		[Category ("SmokeTests")]
 		public void Aapt2Compile ()
 		{
 			var path = Path.Combine (Root, "temp", "Aapt2Compile");
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
@@ -460,6 +460,7 @@ int xml myxml 0x7f140000
 		}
 
 		[Test]
+		[Category ("SmokeTests")]
 		public void CompareAapt2AndManagedParserOutput ()
 		{
 			var path = Path.Combine ("temp", TestName);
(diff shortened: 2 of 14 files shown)


pjcollins:
I cancelled this job after it resolved the issue:

pjcollins:
These changes cut the execution of the "Windows Build + Build Tree Tests" job down to 72 minutes, roughly half of the average of the last 4 runs against master (149 minutes). I want to kick off another run here for comparison, and open this up for further conversation and review.

@jpobst suggested the possibility of limiting the scope of "Windows Build + Build Tree Tests" only for PRs, and keep a full test execution against master/release branches. Personally, I think we can cut the Windows build tree test scope in both PRs and against master/release. It seems like overkill to run every Xamarin.Android build test against the Windows build environment, since we run these same tests against the .vsix we ship later in the pipeline.

@jonpryor @dellis1972 @jonathanpeppers what are your thoughts here? Should we try to find a universal "happy medium" of Windows build tree tests to run against both PR and master/release builds? Or should we only try to cut scope for PR builds?

jpobst:
I am definitely willing to defer to those who actually know what we're testing here.

My goals are:
- Faster PR builds
- Reasonable confidence that the Windows build tree isn't broken.

As long as the test owners are happy with Peter's subset it's a 👍 from me.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
```diff
@@ -460,6 +460,7 @@ public void UpdateLayoutIdIsIncludedInDesigner ([Values(true, false)] bool useRt
 		}
 
 		[Test]
+		[Category ("SmokeTests")]
```
jonathanpeppers: This test doesn't actually build a full project, it only runs `<Aapt*/>` MSBuild tasks directly. It took 501ms for me locally.

Should we pick a different, _slower_ test, here?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/Aapt2Tests.cs
```diff
@@ -84,6 +85,7 @@ public void Aapt2Link ()
 		}
 
 		[Test]
+		[Category ("SmokeTests")]
```
jonathanpeppers: The two tests in this file were fast, too: 410ms and 86ms.

pjcollins: My thinking was that we may want _some_ task only test in the smoke test category to make sure that aspect of the framework remains functional. I think these can be removed in favor of the ManagedResourceParserTest above though.

jonathanpeppers: Maybe I don't understand. Which builds will run the `SmokeTest` category and which run everything?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
pjcollins: I added a couple of other more comprehensive tests. I think we can keep this around to make sure we have some coverage for our individual Task tests in this category?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/Aapt2Tests.cs
pjcollins: The Windows Build + Build Tree tests job will only run `SmokeTests`, while our macOS and Windows test jobs which run against our .pkg and .vsix files will run the full suite.

pjcollins:
The augmented `Windows Build and Smoke Test` job completed in 61 minutes in this latest iteration, I think this should be good to go if everyone is on board. 

Let's wait until next week though to get some input from @dellis1972 before merging.

dellis1972:
@pjcollins Ideally I would like to see at least one test in the smoke tests which has special characters or spaces in a path. These is an area which continually regress on in windows. 
We the plan is to only run smoke tests on PR builds and wait for the full tests once a PR is merged when we should include one path test in the PR set.

pjcollins:
@dellis1972 sounds good, I will update the AOT test to instead use one with special characters. I believe I saw that while looking through these tests. Also just to note, the full set of tests will still run on every PR on both macOS and Windows. The tests which are running against our .pkg and .vsix installers are not being changed, this PR only cuts the scope of the tests running against the output of the Windows Build job.

PR #4171, opened 2020-01-24
[WIP] <ResolveAssemblies/> remove NuGet & recursive lookup

jonathanpeppers:
One of the remaining pain points in the Xamarin.Android build is the
`<ResolveAssemblies/>` MSBuild task.

1. It isn't particularly fast and runs on *every* build. During the
   devloop it runs twice for `Build` and `Install`. It runs for DTBs,
   designer builds, too.

2. It breaks for certain NuGet packages as mentioned on a few Github
   issues:

https://github.com/xamarin/xamarin-android/issues/3898
https://github.com/xamarin/xamarin-android/issues/4074

After some research and looking at build logs, we may be able to
rework the behavior:

1. Remove NuGet-related code and recursive assembly searching.

2. Use `@(_ReferencePath)` and `@(_ReferenceDependencyPaths)`.

3. Include assemblies even if they have `%(ResolvedFrom)` ==
   `ImplicitlyExpandDesignTimeFacades`.

We can also remove:

* Any of the `<PackageReference Include="NuGet.*" />` packages we are
  using.

* `MetadataResolver` class, `NuGetLogger` class

The first issue I found was that we have to include these references
by default:

    Java.Interop;
    Mono.Security;
    System.Net.Http;
    System.Runtime;
    System.Runtime.Serialization;
    System.ServiceModel.Internals;

`System.dll` references `Mono.Security.dll`. `Mono.Android.dll`
references `System.Net.Http.dll` due to `AndroidClientHandler`.

The dependency tree for `System.ServiceModel.Internals.dll` is
somewhat more convoluted:

    Adding assembly reference for Mono.Android, recursively...
      Adding assembly reference for System.Drawing.Common, recursively...
        Adding assembly reference for System.Runtime.Serialization, recursively...
          Adding assembly reference for System.ServiceModel.Internals, recursively...

The next issue I hit was building the SmartHotel360 app in `Release`:

[...]

PR #4275, opened 2020-02-17
Make preparation work when PackageReferences project style is default

kant2002:
Not sure what's specifically do not work on my PC, but I do not have `packages` folder. I have by default opt in into PackageReferences as default in VS. I also has `NUGET_PACKAGES` environment variable set which is reflected in my changes to `Mono.Android.targets`.

My change to `Mono.Android.targets` is just to show what's I have to change,to make build work. I try to use `NuGetPackageFolders` and `NuGetPackageRoot` instead of hardconding environment varibale, but both MSBuild variables was not working for me.

I comes to that changes after I try to play with https://github.com/xamarin/xamarin-android/pull/4171 in hope that it will fix me https://github.com/xamarin/MobileBlazorBindings/issues/70 or similar issues.


src/Mono.Android/Mono.Android.targets
```diff
@@ -177,7 +177,7 @@
       Inputs="$(TargetPath);@(ApiCompatibilityFiles)"
       Outputs="$(IntermediateOutputPath)CheckApiCompatibility.stamp">
     <CheckApiCompatibility
-        ApiCompatPath="$(XamarinAndroidSourcePath)\packages\microsoft.dotnet.apicompat\5.0.0-beta.20078.1\tools\net472\"
+        ApiCompatPath="$(NuGet_Packages)\microsoft.dotnet.apicompat\5.0.0-beta.20078.1\tools\net472\"
```
pjcollins: I think this change will only work when the `NUGET_PACKAGES`  variable is explicitly set in your environment? I don't believe that this is set by NuGet by default, so this is likely a breaking change.

jonpryor: This part feels like it'll fail if `$(NuGet_Packages)` isn't set, which is probably the case on Linux & macOS builds.

build-tools/xaprepare/xaprepare/ThirdPartyNotices/libzip.cs
```diff
@@ -18,5 +18,15 @@ class libzip_nih_at_libzip_TPN : ThirdPartyNotice
 		public override string LicenseText => null;
 
 		public override bool Include (bool includeExternalDeps, bool includeBuildDeps) => includeExternalDeps;
+
+		private static string GetNugetLocation()
```
pjcollins: Could we move this helper method into https://github.com/xamarin/xamarin-android/blob/master/build-tools/xaprepare/xaprepare/Application/Context.cs or maybe https://github.com/xamarin/xamarin-android/blob/master/build-tools/xaprepare/xaprepare/ConfigAndData/Configurables.cs so that it can be called by all places looking for a NuGet package path, rather than duplicating this logic across three classes?

src/Mono.Android/Mono.Android.targets
kant2002: I know that this is not full fix. I just want to point to location which I have to change for myself. I try different options, but not that good with Xamarin codebase to made changes. I technically can create separate MSBuild variable, but not sure that this is desired change. What to discuss first,

jonpryor:
> I do not have packages folder.

That's because you're setting the `%NUGET_PACKAGES%` environment variable.  From the [`globalPackagesFolder`](https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file) documentation:

> | **Key** | **Value** |
> |--------|-----------|
> | globalPackagesFolder (projects using PackageReference only) | The location of the default global packages folder. The default is `%userprofile%\.nuget\packages` (Windows) or `~/.nuget/packages` (Mac/Linux). A relative path can be used in project-specific `nuget.config` files. This setting is overridden by the `NUGET_PACKAGES` environment variable, which takes precedence. |

We explicitly set `globalPackagesFolder` so that various build steps would have a "stable" location for tool invocation purposes; see commit 3a4acf5dcb28ddea179fa1c2d9ab8953275dc5bb.  For example, the `src/Mono.Android` build invokes `Microsoft.DotNet.ApiCompat.exe` (via the `<CheckApiCompatibility/>` target, and overriding `globalPackagesFolder` means it's not in a "stable" location.

Why are you setting `%NUGET_PACKAGES%`?

jonpryor:
> Why are you setting `%NUGET_PACKAGES%`?

...though to a large extent, this is a pointless question and shouldn't actually *matter*; it's a documented environment variable, with documented semantics, and the xamarin-android build *should not break* "just because" this value is set ("within reason", poorly defined).

A "spit-balled" solution is to update `Configuration.props` with:

```xml
<XAPackagesDir Condition=" '$(XAPackagesDir)' == '' And '$(NUGET_PACKAGES)' != '' ">$(NUGET_PACKAGES)</XAPackagesDir>
<XAPackagesDir Condition=" '$(XAPackagesDir)' == '' And '$(OS)' == 'Windows' ">%userprofile%\.nuget\packages</XAPackagesDir>
<XAPackagesDir Condition=" '$(XAPackagesDir)' == '' And '$(OS)' != 'Windows' ">%(HOME)/.nuget/packages</XAPackagesDir>
```

Then use `$(XAPackagesDir)` instead of hardcoding `$(XamarinAndroidSourcePath)\packages`.

Which works with 2 seconds thought, but...  If someone then edits `NuGet.config` to override `globalPackagesFolder`, things will break (differently), as `$(XAPackagesDir)` won't match the `globalPackagesFolder` within `NuGet.config`.

😞

A slightly different take might be to use:

```xml
<XAPackagesDir Condition=" '$(XAPackagesDir)' == '' And '$(NUGET_PACKAGES)' != '' ">$(NUGET_PACKAGES)</XAPackagesDir>
<XAPackagesDir Condition=" '$(XAPackagesDir)' == '' >$(XamarinAndroidSourcePath)\packages</XAPackagesDir>
```

and continue to rely on the `globalPackagesFolder` value, though this approach still doesn't deal well with `NuGet.config` changes without corresponding build system changes, e.g. updating the `CheckApiCompatibility.ApiCompatPath` property value, which just feels "wrong".

jonpryor:
Perhaps `$(NuGetPackageRoot)` is the way to go?

https://stackoverflow.com/a/57048513

jonpryor:
Indeed, `$(NuGetPackageRoot)` *does* exist, at least with mono 6.6.0.155 / MSBuild 16.4.0.56401, and contains values which mirror that of NuGet.config.

Looks like we should take a page out of PR #4282 so that `xaprepare` can read the `$(NuGetPackageRoot)` property, and use `$(NuGetPackageRoot)` ~everywhere.

kant2002:
> Why are you setting `%NUGET_PACKAGES%`?
I have small system drive - 128 GB so I distribute a lot of caches to second HDD. This gives me unique perspective on some issues as you can see, but really I cannot do much. 

> Perhaps `$(NuGetPackageRoot)` is the way to go?
I thought the same, but this variable simply does not work for me. 


> ```
> <XAPackagesDir Condition=" '$(XAPackagesDir)' == '' And '$(NUGET_PACKAGES)' != '' ">$(NUGET_PACKAGES)</XAPackagesDir>
> <XAPackagesDir Condition=" '$(XAPackagesDir)' == '' >$(XamarinAndroidSourcePath)\packages</XAPackagesDir>
> ```
I'm in favor of that change. I think about it myself, but do not have idea how I can properly change name property, and that this change in the Xamarin spirit. If so, I completely understand that this is least invasive change which keep status quo.
Seems to be these variables controlled by `$(ExcludeRestorePackageImports)` property which I do not fully understand, and as such can create bigger mess then with suggested changes.

I made changes according to suggestions. Both by @jonpryor and @pjcollins (hopefully I properly recognize location for placing variables)


Side comment: `msbuild Xamarin.Android-Tests.sln` does not work for me. It fails with 
```
 Java.Interop\JnienvTest.cs(380,38): error CS0117: 'TypeManager' does not contain a definition for 'GetJavaToManagedType' [xamarin-android\src\Mono.Android\Test\Mono.Android-Tests.csproj]

// and 
obj\Debug\generated\src\Foo.IUnsignedInterface.cs(179,41): error CS1003: Syntax error, ',' expected [xamarin-android\tests\CodeGen-Binding\Xamarin.Android.McwGen-Tests\Xamarin.Android.McwGen-Tests.csproj]
```

grendello:
@kant2002 build failures may result from previous versions build artifacts that weren't properly updated or cleaned up. You should clean your repo and start from there. I use a script which does roughly the following to get me a fresh start:

```shell
#!/bin/bash -e
git clean -xdff
git submodule foreach --recursive git clean -xdff
git submodule foreach --recursive git reset --hard HEAD
git submodule sync
git submodule update --recursive --init --force
```

When working with master it's often necessary to start from scratch.

kant2002:
@grendello I do cleanup, and issues with building `Android-Test.sln` still there.

First issue comes from the fact, that test project has
```
<Reference Include="Mono.Android" />
<Reference Include="Mono.Android.Export" />
<Reference Include="Mono.Data.Sqlite" />
```
Looks like it does not pickup just build references, and instead of that use shipped with MS

jonpryor:
@kant2002 wrote:

> I thought the same, but [`$(NuGetPackageRoot)`] simply does not work for me.

Can you elaborate on the failure?  Which MSBuild version are you using?  (I wonder if its because your MSBuild is too old?)

jonpryor:
@kant2002 wrote:

> Side comment: `msbuild Xamarin.Android-Tests.sln` does not work for me. It fails with...

I think we have a "typo" in our documentation.  Instead of using `msbuild`, you should use e.g. `bin\Debug\bin\xabuild.exe`, as we do in our [YAML build script](https://github.com/xamarin/xamarin-android/blob/87ee20c59a09ba4314f4226e1c77ee2331c3f09f/build-tools/automation/azure-pipelines.yaml#L273)

    bin\Debug\bin\xabuild.exe Xamarin.Android-Tests.sln /restore /p:Configuration=Debug /bl:bin\TestDebug\msbuild-build-tests.binlog

kant2002:
@jonpryor wrote
> Can you elaborate on the failure? Which MSBuild version are you using? (I wonder if its because your MSBuild is too old?)

```
>msbuild -version
Microsoft (R) Build Engine version 16.4.0+e901037fe for .NET Framework
16.4.0.56107
```

@jonpryor wrote
> I think we have a "typo" in our documentation. Instead of using `msbuild`, you should use e.g. `bin\Debug\bin\xabuild.exe`, as we do in our [YAML build script](https://github.com/xamarin/xamarin-android/blob/87ee20c59a09ba4314f4226e1c77ee2331c3f09f/build-tools/automation/azure-pipelines.yaml#L273)
> 
> ```
> bin\Debug\bin\xabuild.exe Xamarin.Android-Tests.sln /restore /p:Configuration=Debug /bl:bin\TestDebug\msbuild-build-tests.binlog
> ```

Have couple problems. This can be discussed in separate issue if you like and I can submit separate PR after that.

1. XaBuild do not want to cleanup symlinks in folder `bin\Debug\lib\xamarin.android\xbuild\Microsoft`
```
bin\Debug\bin\xabuild.exe Xamarin.Android-Tests.sln /restore /p:Configuration=Debug /bl:bin\TestDebug\msbuild-build-tests.binlog /p:HOME=f:\d\build
Removing old symlink: bin\Debug\lib\xamarin.android\xbuild\Microsoft

Unhandled Exception: System.IO.IOException: The directory is not empty.
...
   at System.IO.Directory.Delete(String path)
   at Xamarin.Android.Build.XABuild.Main() in d:\d\kant\GitHub\xamarin-android\tools\xabuild\XABuild.cs:line 67
```
2. Android NDK cannot be found when installed in the different location
```
src\Mono.Android\Test\Mono.Android-Tests.targets(17,5): error : Could not locate Android NDK
```
To say true, I add `/p:HOME=f:\d\build` to your command line. Once I place variable in `Configuration.Override.props` everything build fine. Honestly I expect property to work, since syntax of XABuild too close to MSBuild syntax.

jonpryor:
@kant2002 wrote:

> 1. XaBuild do not want to cleanup symlinks in folder bin\Debug\lib\xamarin.android\xbuild\Microsoft

The first run needs to be run as Administrator so that it is *able* to create symlinks.  Which is something else I need to add to the docs... :-/


> 2. Android NDK cannot be found when installed in the different location

Meaning the NDK is not present within `%USERPROFILE%\android-toolchain\ndk`?  I'm not sure I properly understand what you're saying.

The xamarin-android expectation is that it will *not* use a pre-existing NDK directory, and instead will rely on the one that it installs into `%USERPROFILE%\android-toolchain\ndk`, as different NDK versions will have different compiler versions which support different C++ versions, and this is the easiest way for us to ensure that our code builds in a simple and reliable manner without having a "manual software installation list" that's a mile long...

> Honestly I expect property to work, since syntax of XABuild too close to MSBuild syntax.

This further confuses me.  You're saying that this command doesn't work?

```
> bin\Debug\bin\xabuild /p:HOME=f:\d\build ...
```

That seems odd, as `xabuild` *uses* MSBuild to parse the command-line arguments.

Does it work if you instead override `AndroidNdkDirectory`?

```
> bin\Debug\bin\xabuild /p:AndroidNdkDirectory=F:\d\build\android-toolchain\ndk ...
```

kant2002:
@jonpryor 
> 1. The first run needs to be run as Administrator so that it is _able_ to create symlinks. Which is something else I need to add to the docs... :-/
Actually it _was able_ to create without Admin, but _not_ able to delete on second run. Is this expected if I run under non-Admin?
See #4306

> 2. Meaning the NDK is not present within `%USERPROFILE%\android-toolchain\ndk`? I'm not sure I properly understand what you're saying.
See #4305 

I try to offload questions which really unrelated to that PR

jonpryor:
...and due to a glorious GitHub snafu, my intended commit message was lost. :-(

Intended commit message:

```
Context: 3a4acf5dcb28ddea179fa1c2d9ab8953275dc5bb

The `%NUGET_PACKAGES%` environment variable overrides the
[`globalPackagesFolder` value within `NuGet.config`][0], but our build
system did not support this:

	> set NUGET_PACKAGES=F:\packages
	> msbuild /t:Prepare Xamarin.Android.sln
	> msbuild Xamarin.Android.sln
	# eventual build failure as it attempts to find files within
	# ...\xamarin-android\packages, which does not exist.

Further investigation showed that `$(NuGetPackageRoot)` did not appear
to contain the `%NUGET_PACKAGES%` value in @kant2002's environment.

Support overriding `%NUGET_PACKAGES%` by adding a new
`$(XAPackagesDir)` MSBuild property to `Configuration.props`, which
contains the path to the NuGet `packages` directory to use.  It can
be overridden in `Configuration.Override.props` -- but probably
shouldn't, as that may cause mismatches between the build & NuGet --
and will default to using `%NUGET_PACKAGES%` if set, otherwise will
use the previous `...\xamarin-android\packages` directory.

[0]: https://docs.microsoft.com/en-us/nuget/reference/nuget-config-file
```

PR #4285, opened 2020-02-19
[ci] Improve test execution conditions

pjcollins:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3482854&view=logs&j=8a30e15e-2b03-5f17-c7f5-98058d3d9cd8&t=47d4d312-ac98-567e-7ba2-f57378c80633

Replaces `succeededOrFailed()` conditions with `continueOnError: true`
to ensure that steps that should be treated as fatal stop execution of
a job. Previously, all test steps would run regardless of the success of
a prior step, even if that step was necessary (e.g. building a test
suite before running it). A job should continue if any step with the new
`continueOnError: true` parameter fails, and we should now abort earlier
when a step without this parameter errors out.

A new template has been added to ensure jobs which encounter issues
in tests are still flagged as failed (rather than `SucceededWithIssues`).

Finally, I've increased the shallow clone depth for our regression test
dependencies so that checkout will still succeed if more commits land on
the branch that is being tracked after the pipeline instance is created.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -282,6 +282,7 @@ stages:
           /t:RunJavaInteropTests
           /p:TestAssembly="bin\Test$(XA.Build.Configuration)\generator-Tests.dll;bin\Test$(XA.Build.Configuration)\Java.Interop.Tools.JavaCallableWrappers-Tests.dll;bin\Test$(XA.Build.Configuration)\logcat-parse-Tests.dll;bin\Test$(XA.Build.Configuration)\Xamarin.Android.Tools.ApiXmlAdjuster-Tests.dll;bin\Test$(XA.Build.Configuration)\Xamarin.Android.Tools.Bytecode-Tests.dll"
           /bl:$(System.DefaultWorkingDirectory)\bin\Test$(XA.Build.Configuration)\msbuild-run-ji-tests.binlog
+      continueOnError: True
 
     - task: PublishTestResults@2
       displayName: publish test results
@@ -289,7 +290,6 @@ stages:
         testResultsFormat: NUnit
         testResultsFiles: TestResult-*.xml
         testRunTitle: Java Interop Tests - Windows Build Tree
-      condition: succeededOrFailed()
 
     - template: yaml-templates\run-nunit-tests.yaml
       parameters:
@@ -451,6 +451,7 @@ stages:
           /restore
           /t:Build
           /bl:$(System.DefaultWorkingDirectory)/bin/Test$(ApkTestConfiguration)/build-check-boot-times.binlog
+      continueOnError: true
 
     - task: MSBuild@1
       displayName: Run check-boot-times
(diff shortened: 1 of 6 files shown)


pjcollins:
The downside with this approach appears to be that failures in steps which set `continueOnError: true` do not cause the job/check to fail, which could cause us to overlook some test failures.

I'll look at alternatives, but at present I think the two options to move forward are:
1) Go back to the old approach
2) Add a step at the end of every test job that checks to see if any issues arose.

If we can find a reasonably elegant way to do 2. I think that would be preferable.

pjcollins:
Full test run triggered here - https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3488220&view=results

jonpryor:
in-progress commit message:

```
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3482854&view=logs&j=8a30e15e-2b03-5f17-c7f5-98058d3d9cd8&t=47d4d312-ac98-567e-7ba2-f57378c80633

What should happen when a job step fails?  Should execution continue,
or should the entire job fail?

Unfortunately, often the answer is "yes"; some jobs -- such as the
uploading of build logs -- we *always* want to happen, regardless of
errors in previous steps.  Other jobs shouldn't run if a previous
step fails.

Which brings us to the Context url, in which the
**Checkout xamarin/xamarin-forms-samples@master** step failed.
Without this step, ~everything after is meaningless, but what AzDO
shows is:

  * [✔︎] ...
  * [✘] Checkout xamarin/xamarin-forms-samples@master ...
  * [➤] Provision Android Dependencies
  * [➤] install mono
  * [➤] DownloadPipelineArtifact
  * ...
  * [✔︎] run Incremental Build Tests
  * [➤] kill leftover processes
  * [✔︎] PublishTestResults
  * [✘] run AOT Tests
  * ...

Restructure job invocations, and replace
`condition: succeededOrFailed()` with `continueOnError: true` to
ensure that steps which should be treated as fatal -- those *without*
`continueOnError: true` -- stop execution of the entire job.

Previously, all test steps would run regardless of the success of
a prior step, even if that step was necessary (e.g. building a test
suite before running it).  A job should continue if any step with the
new `continueOnError: true` parameter fails, and we should now abort
earlier when a step without this parameter errors out.

Finally, I've increased the shallow clone depth for our regression
test dependencies so that checkout will still succeed if more commits
land on the branch that is being tracked after the pipeline instance
is created.
```

jonpryor:
@pjcollins: question about the semantics: if the [`build remap-assembly-ref.csproj` step](https://github.com/xamarin/xamarin-android/blob/a98e4f5e47a59c6299456f1da94806a67a07d4ff/build-tools/automation/azure-pipelines.yaml#L650) fails -- which lacks `continueOnError` or `condition` -- will `yaml-templates/upload-results.yaml` still be executed?  'cause if `build remap-assembly-ref.csproj` fails, presumably we'd need the log files in order to ascertain what went wrong...

pjcollins:
@jonpryor The step to upload supplemental build/test results will always run, regardless of the state of previous steps - https://github.com/xamarin/xamarin-android/blob/c8b0cf2a641343ec25a49a9d7dce13f06c17e9b1/build-tools/automation/yaml-templates/upload-results.yaml#L20

PR #4293, opened 2020-02-20
Bump to xamarin/java.interop/master@b81cfbb

jonathanpeppers:
Changes: https://github.com/xamarin/java.interop/compare/8f30933...b81cfbb

* [Java.Interop] Call GetSimpleReference before custom attrs

* [Xamarin.Android.Tools.Bytecode] Kotlin name shadowing support

* [Java.Interop.Tools.Cecil] cache TypeReference.Resolve() calls


jonathanpeppers:
There seems to be on random, unrelated test failure:
```
Xamarin.Android.Build.Tests.ResolveSdksTaskTests.ResolveSdkTiming

Task should not take more than 2 seconds to run.
Expected: less than or equal to 00:00:02
But was: 00:00:02.7268197
```
If this happens on other PRs, I'll take a look at this test.

Otherwise, this looks OK.

PR #4294, opened 2020-02-20
Enable nullable reference types in xaprepare

grendello:
This started as a "quick" experiment which turned out to generate around
390 warnings regarding code affected by `null` one way or another.
Despite nullable reference types sometimes forcing to produce more
verbose code or instantiate more objects than was necessary before
enabling the feature, I think the net result is worth the effort. A few
code paths were fixed where a potential for null dereference existed,
nothing really serious but still well worth it to have stabler (albeit
slightly bigger and possibly a tiny bit slower) codebase.

### DIFF
--- a/build-tools/automation/yaml-templates/setup-test-environment.yaml
+++ b/build-tools/automation/yaml-templates/setup-test-environment.yaml
@@ -11,6 +11,27 @@ steps:
   parameters:
     provisionExtraArgs: ${{ parameters.provisionExtraArgs }}
 
+- task: UseDotNet@2
+  displayName: install .NET Core $(DotNetCorePreviewVersion)
+  inputs:
+    version: $(DotNetCorePreviewVersion)
+
+- task: UseDotNet@2
+  displayName: install .NET Core $(DotNetCoreVersion)
+  inputs:
+    version: $(DotNetCoreVersion)
+
+- script: |
+    dotnet tool install --global boots
+  displayName: install boots
+  condition: and(succeeded(), eq(variables['agent.os'], 'Darwin'))
+
+- script: |
+    export PATH="$PATH:/Users/runner/.dotnet/tools"
+    boots https://download.mono-project.com/archive/6.8.0/macos-10-universal/MonoFramework-MDK-6.8.0.105.macos10.xamarin.universal.pkg
+  displayName: update Mono to version 6.x
+  condition: and(succeeded(), eq(variables['agent.os'], 'Darwin'))
+
 - task: MSBuild@1
   displayName: build xaprepare
   inputs:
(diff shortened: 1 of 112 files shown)


jonpryor:
Looks like many of the machines we run tests on don't support C#8... :-/

grendello:
yep, seems to be the case :(

build-tools/automation/yaml-templates/setup-test-environment.yaml
```diff
@@ -11,6 +11,27 @@ steps:
...
+- script: |
+    dotnet tool install --global boots
+  displayName: install boots
+  condition: and(succeeded(), eq(variables['agent.os'], 'Darwin'))
+
+- script: |
+    export PATH="$PATH:/Users/runner/.dotnet/tools"
+    boots https://download.mono-project.com/archive/6.8.0/macos-10-universal/MonoFramework-MDK-6.8.0.105.macos10.xamarin.universal.pkg
+  displayName: update Mono to version 6.x
+  condition: and(succeeded(), eq(variables['agent.os'], 'Darwin'))
```
jonathanpeppers, line 33: Do we need these `boots` commands here, too? https://github.com/xamarin/xamarin-android/blob/42822e0488185cdf4bca7c0bd05b21ad03dfbd7e/build-tools/automation/azure-pipelines.yaml#L102-L115

If there is an old mono, then `xaprepare` wouldn't build and `make prepare-update-mono` would fail?

If so, do we need to run `make prepare-update-mono` at all?

grendello: The YAML changes here are specifically for our mac test bots which are imaged with Mono 5.12 and this is the only way to provision them. `prepare-update-mono` still serves its purpose on our **build** bots where the base Mono version is new enough to build NRT but we still **might** need a newer mono. Even on test bots - YAML installs some version of Mono sufficient to build NRT code and `prepare-update-mono` may still update the Mono version. That means we don't need to change the Mono version installed by YAML.

jonathanpeppers: So we rely on `Hosted Mac Internal Mojave` to have a new enough Mono 👍 :

https://github.com/xamarin/xamarin-android/blob/42822e0488185cdf4bca7c0bd05b21ad03dfbd7e/build-tools/automation/azure-pipelines.yaml#L54-L55

But `Hosted Mac Internal` does not have it...

jonpryor: > If so, do we need to run `make prepare-update-mono` at all?

It's a matter of duplication and synchronization.

"Mere mortals" like developers on their own machines *cannot* execute YAML files.  (Alas!). As such, there needs to be *a* mono version which is controlled *outside* YAML, currently in `Configuration.props` via `$(MonoDarwinPackageUrl)`.

So *a* "source" of the Mono framework to install must live in `Configuration.props`.

Unfortunately, that single source is insufficient in the context of this PR, because we need to install "a" mono *before* we build xaprepare, which is what reads `Configuration.props` to install the specified mono.

Which means a *second* source of truth, the YAML file.

(I suppose we could update xaprepare to instead extract the Mono URL from the YAML file, and then we'd be back to one source of truth…)

So long as there are two potential sources of truth, we need *both* the `UseDotNet@2` *and* `make prepare-update-mono` invocations.

grendello: @jonpryor yes, in the long run I think the idea of having a single source of the URL is the correct one. Extracting it from the YAML file would work just fine, but I'd rather not do the work right now and go along with what's in this PR.

PR #4308, opened 2020-02-21
Remove KeyTool from signing

BretJohnson:
Invoking keytool takes about 200ms (my machine, best case, likely mostly due to java startup). Peppers and I were looking for ways to speed up Apply Changes builds and Android builds in general. For Apply Changes, removing the call to keytool (or caching its result if it's needed) was next on my optimization list.

It looks like the KeyTool invocation isn't needed at all.
Here's the commit which added it, 5 years ago: https://github.com/xamarin/monodroid/commit/fe94c0a88ca68e78a639d0032fb59dd91409e9ca

The comment there says "emit the SHA1 during the _Sign target to make it easier for people to publish to google play".
But note that:
- KeyTool is only run when the AndroidKeyStore isn't supplied, so it's only run when using the default debug keystore, not when the user supplies one
- I don't actually see anything output in VS (at least on Windows) in the build log when keytool is run there - the output isn't captured.
- KeyTool output isn't used as other otherwise by our build (except for an error blowing out the build)
- If the user suppolies an invalid keystore or key alias password, the sign task itself will fail (I did a quick test to verify).

So it seems like KeyTool doesn't serve any useful purpose here & we'd like to get the perf bump of removing it.
Unless Dean can think of reasons why we shouldn't, it seems safe to pull it.

### DIFF
--- a/Documentation/release-notes/4308.md
+++ b/Documentation/release-notes/4308.md
@@ -15,3 +15,8 @@ Running `keytool -list` does *nothing* to benefit the development loop,
 while the archive workflow within Visual Studio can show the same information.
 
 [show-sha1]: https://xamarin.github.io/bugzilla-archives/13/13154/bug.html
+
+If you still need to determine the Keystore Signature, please follow the steps
+in the article [Finding your Keystore's Signature][keystore-sig].
+
+[keystore-sig]: https://docs.microsoft.com/en-us/xamarin/android/deploy-test/signing/keystore-signature?tabs=windows
--- a/Documentation/release-notes/4308.md
+++ b/Documentation/release-notes/4308.md
@@ -1,0 +1,17 @@
+### Build and deployment performance
+
+The `SignAndroidPackage` target will no longer execute the command:
+
+```
+keytool -list -keystore $(AndroidSigningKeyStore) --storepass $(AndroidSigningStorePass) -v
+```
+
+This was originally done [as a convenience][show-sha1], but this
+adds ~200ms to the execution time of the `SignAndroidPackage` target,
+which is executed as part of the inner development loop and the
+Apply Changes feature within the UI Designer.
+
+Running `keytool -list` does *nothing* to benefit the development loop,
+while the archive workflow within Visual Studio can show the same information.
+
+[show-sha1]: https://xamarin.github.io/bugzilla-archives/13/13154/bug.html
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -3001,15 +3001,6 @@ because xbuild doesn't support framework reference assemblies.
 		<_JarSignerSuffix Condition=" '$(AndroidPackageFormat)' != 'aab' ">-Signed-Unaligned</_JarSignerSuffix>
 		<_JarSignerSuffix Condition=" '$(AndroidPackageFormat)' == 'aab' ">-Signed</_JarSignerSuffix>
 	</PropertyGroup>
-	<KeyTool
-		KeyStore="$(_ApkKeyStore)"
-		KeyAlias="$(_ApkKeyAlias)"
-		KeyPass="$(_ApkKeyPass)"
-		StorePass="$(_ApkStorePass)"
-		ToolPath="$(KeytoolToolPath)"
-		ToolExe="$(KeytoolToolExe)"
-		Command="-list"
-		Condition="'$(AndroidKeyStore)'==''" />
 	<AndroidSignPackage Condition=" '$(AndroidUseApkSigner)' != 'true' "
 		UnsignedApk="%(ApkAbiFilesIntermediate.FullPath)"
 		SignedApkDirectory="$(OutDir)"


dellis1972:
@BretJohnson @jonathanpeppers do the original bug was not a bug but a feature request. Using allot of google play services need the hash of the cert used to sign the app. So we added this as a way for the users to just look at the output and pick up the required hash. It was to make the users life a bit easier. 

So while it's not "required" for building, it would be nice to have some way for a user to get the required information easily from MSbuild. This might mean a new target...

jonathanpeppers:
@dellis1972 we noticed it wasn't printing anything anymore. Maybe it got broken in: https://github.com/xamarin/xamarin-android/commit/7c38f601f8667749686bd464d3df649a1273fa3e#diff-d256cdd1b672f1afdc22f87b4a8ad29a

Do we need to fix it and hide the feature behind a property? It could be `$(AndroidPrintSigningHash)` or something.

However, if I broke it in 2018 and no one noticed. 👀 Maybe we don't need it?

dellis1972:
@jonathanpeppers doh :) 

putting it behind a documented feature flag might be nice. Also it would be nice to figure out how this works with aab files. i.e how do you link those to google play. Do we even need this anymore. Its been a while since I used a google service.

dellis1972:
I still think we should have a way of getting this in the product without having to use a Third Party addin. 

If we set the Verbose property we can get the SHA256 output as well

```
keytool -list -keystore foo.keystore --storepass android -v

Keystore type: JKS
Keystore provider: SUN

Your keystore contains 1 entry

Alias name: android
Creation date: 27-Feb-2020
Entry type: PrivateKeyEntry
Certificate chain length: 1
Certificate[1]:
Owner: CN=Android Debug, O=Android, C=US
Issuer: CN=Android Debug, O=Android, C=US
Serial number: 58be88c7
Valid from: Thu Feb 27 12:47:38 GMT 2020 until: Sat Feb 19 12:47:38 GMT 2050
Certificate fingerprints:
	 MD5:  5E:3E:0E:4C:94:77:F6:FC:A8:0D:BA:EF:81:7A:15:99
	 SHA1: 2D:70:CB:10:02:C5:AA:79:2F:EC:17:EC:3E:D3:97:07:52:D2:96:30
	 SHA256: F8:78:49:13:FE:59:DF:EA:CD:88:FF:09:D9:E5:5B:8A:41:CC:A7:00:9E:B8:F7:F6:88:46:B5:BF:40:70:6B:8F
	 Signature algorithm name: SHA256withRSA
	 Version: 3

Extensions: 

#1: ObjectId: 2.5.29.14 Criticality=false
SubjectKeyIdentifier [
KeyIdentifier [
0000: 3D 06 B7 9D 76 70 5A 6B   A2 F1 49 19 09 2E 51 93  =...vpZk..I...Q.
0010: D6 EF 5F 5F                                        ..__
]
]



```

I am ok with adding a new target for a user to call which will get this info, rather than having it happen during a sign.

BretJohnson:
Could we add doc for this & maybe just handle via doc?

It feels like no matter what we do we'd want some doc to explain the steps.
And given that, maybe doc is sufficient to explain the steps, including the keytool command above.

Though that said, it does seem like adding a target would be easy & may make our doc with the steps more concise.

dellis1972:
The reason I suggest adding a target is that our tooling knows where keytool is. the user might not have that in their PATH. It just makes it easy for the user to run say

`msbuild /t:PrintKeyStoreInfo /p:AndroidSigningKeyStore=foo.store /p:AndroidSigningStorePass=android`

but then again 

`keytool -list -keystore foo.keystore --storepass android -v`

looks just as simple...

jonpryor:
Could/should we add a new MSBuild property which controls whether or not `keytool` is run?

BretJohnson:
My feeling, fwiw, is:
- Update the doc to show the steps. That is the most important thing.
- And just have the doc show steps to run keytool directly. It could also maybe point the the marketplace extension above as an option.
@JonDouglas What are your thoughts here?

JonDouglas:
We expose _some_ of this information via Visual Studio's archive flow. It is a bit hidden from a UX perspective, but it's there. i.e. Double Click on a keystore:

image

In my opinion, these changes make more sense in IDE tooling than SDK. I'd be in favor of removing it!

BretJohnson:
The release notes update look good to me, as do the rest of the changes here.
This is ready to merge, from my perspective. Thanks all.

BretJohnson:
@brendanzagaeski To clarify, this didn't actually do anything useful in practice before, as the user couldn't see the keytool output here. So there's no need to document any loss of functionality.

brendanzagaeski:
@BretJohnson, I was able to get the SHA-1 fingerprint for my `debug.keystore` in the diagnostic build output with the current version. It looks roughly like this:

```
Task "KeyTool"
  KeyTool : -list
            C:\Users\user1\AppData\Local\Xamarin\Mono for Android\debug.keystore
  C:\Program Files\Android\Jdk\microsoft_dist_openjdk_1.8.0.25\bin\keytool.exe -list -alias androiddebugkey -storepass android -keypass android -keystore "C:\Users\user1\AppData\Local\Xamarin\Mono for Android\debug.keystore"
C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets(3004,2): message ANDKT0000: androiddebugkey, Aug 5, 2019, PrivateKeyEntry,
C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets(3004,2): message ANDKT0000: Certificate fingerprint (SHA1): 00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00
C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets(3004,2): message ANDKT0000: Warning: The JKS keystore uses a proprietary format. It is recommended to migrate to PKCS12 which is an industry standard format using "keytool -importkeystore -srckeystore C:\Users\user1\AppData\Local\Xamarin\Mono for Android\debug.keystore -destkeystore C:\Users\user1\AppData\Local\Xamarin\Mono for Android\debug.keystore -deststoretype pkcs12".
Done executing task "KeyTool".
```

brendanzagaeski:
That said, if the answers to the other questions are that there are no valid uses of the SHA-1 fingerprint anymore or that we should discourage using the `debug.keystore` for anything that needs a fingerprint, then the current invocation of the `<KeyTool/>` task is indeed not relevant to customers.

In that case, it would be fine to merge this PR. We could potentially create a separate follow-up item to discuss whether it would be useful to provide a property to let users run a new, updated version of the `<KeyTool/>` task that provides the SHA-256 fingerprints.

BretJohnson:
I see we have this doc with steps to find the fingerprint (using keytool): https://docs.microsoft.com/en-us/xamarin/android/deploy-test/signing/keystore-signature?tabs=windows

That addresses the user need here, right? Does everything look correct in that doc?

PR #4312, opened 2020-02-24
[xaprepare] Further steps to fix brew

grendello:
Turns out we can sometimes have a situation when a homebrew package is
installed but it's not linked into a directory that's found in the
user's `PATH` environment variable.

Remedy the situation by unconditionally linking all the packages we
depend on.

### DIFF
--- a/build-tools/xaprepare/xaprepare/Application/HomebrewProgram.MacOS.cs
+++ b/build-tools/xaprepare/xaprepare/Application/HomebrewProgram.MacOS.cs
@@ -190,6 +190,11 @@ namespace Xamarin.Android.Prepare
 				return;
 			}
 
+			// It may happen that the package is installed but not linked to the prefix that's in the user's PATH.
+			// Detecting whether the package is linked would require requesting and parsing JSON for all the packages which
+			// would be more trouble than it's worth. Let's just link the package
+			await runner.Link (Name, echoOutput: false, echoError: false);
+
 			if (!Pin)
 				return;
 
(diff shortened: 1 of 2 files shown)


build-tools/xaprepare/xaprepare/ToolRunners/BrewRunner.MacOS.cs
```diff
@@ -21,6 +21,8 @@ public BrewRunner (Context context, Log log = null, string toolPath = null)
 			: base (context, log, toolPath)
 		{
 			ProcessTimeout = TimeSpan.FromMinutes (30);
+			EchoStandardError = true;
```
jonpryor: Not sure I see why these changes are needed?  I don' understand the interaction between these properties and the parameters to `Link()`.

build-tools/xaprepare/xaprepare/Application/HomebrewProgram.MacOS.cs
```diff
@@ -190,6 +190,11 @@ protected override async Task AfterDetect (bool installed)
 				return;
 			}
 
+			// It may happen that the package is installed but not linked to the prefix that's in the user's PATH.
+			// Detecting whether the package is linked would require requesting and parsing JSON for all the packages which
+			// would be more trouble than it's worth. Let's just link the package
+			await runner.Link (Name, echoOutput: false, echoError: false);
```
jonpryor: Why *wouldn't* we want to echo stdout & stderr?

build-tools/xaprepare/xaprepare/ToolRunners/BrewRunner.MacOS.cs
grendello: It's needed because without this we won't **capture** the output and each `xaprepare` run will be marred with the messages below for each dependency check:

```
Warning: Already linked: /usr/local/Cellar/mingw-w64/7.0.0_1
To relink:
  brew unlink mingw-w64 && brew link mingw-w64
```

build-tools/xaprepare/xaprepare/Application/HomebrewProgram.MacOS.cs
grendello: See my reply above. The `Warning` line is printed to `stderr` and would be shown in red, the remaining two lines are sent to `stdout`.

PR #4321, opened 2020-02-25
[docs] fix broken image link in profiling.md

jonathanpeppers:
I had Windows-style directory separators, which doesn't appear to work in Github.

It worked in VS Code on Windows, so I didn't notice it there.


jonathanpeppers:
Yay, it works now if I go to: https://github.com/xamarin/xamarin-android/blob/78c15ff1ed7c1aaa28b1aaec00217175326e3d39/Documentation/guides/profiling.md

PR #4327, opened 2020-02-26
Adding API-R Beta 1 Preview Binding

gugavaro:
Adding API-R Beta 1 Preview Binding

Android R Beta 1 was released in late February.


brendanzagaeski:
*Draft release notes*

````
### Preview bindings for Android 11 Developer Preview

This version includes preview bindings for the [first Developer Preview of
Android 11][android-11-preview-1] from Google.  See the [Android 11 Developer
Preview documentation][android-11-upstream] for additional information about the
behavior and API changes in this new Android version.  To try the bindings for
the new APIs in a Xamarin.Android project, set the `$(TargetFrameworkVersion)`
to `v10.0.99` in the `.csproj` file:

```xml
<TargetFrameworkVersion>v10.0.99</TargetFrameworkVersion>
```

[android-11-preview-1]: https://android-developers.googleblog.com/2020/02/Android-11-developer-preview.html
[android-11-upstream]: https://developer.android.com/preview
````

File: tests/api-compatibility/acceptable-breakages-v10.0.99.txt [hunk 1]
```diff
@@ -0,0 +1,5 @@
+Compat issues with assembly Mono.Android:
+CannotAddAbstractMembers : Member 'Android.Telephony.CellInfo.CellIdentity' is abstract in the implementation but is missing in the contract.
```
jonpryor: This introduces a semantic question:

The *intention* "going forward" -- API-R and beyond -- is that we will not have any more API breakages.  There will *just* be `v10.0`.

As such, we (presumably?) do *not* want to update `acceptable-breakages-*.txt`, because they aren't acceptable. :-)

That said, in order *actually* not have any API breakages, we will need `generator` fixes: https://github.com/xamarin/java.interop/issues/509

My concern is that if we merge this PR as-is, we'll forget to remove this `acceptable-breakages-v10.0.99.txt` file, and then miss any "actual" breakages when API-R goes stable.

Thus, instead of adding this file, could we instead make the `<CheckApiCompatibility/>` task invocation within `Mono.Android.targets` *conditional* on `%(AndroidApiInfo.Stable)`==True?  That way *no* API compatibility checking will be performed for *unstable* API levels, reducing the likelihood of us accidentally accepting API breakage in stable versions.

*Alternatively*, could we have a "different set" of `acceptable-breakages-*.txt` files to track unstable API levels?  `unstable-acceptable-breakages-*.txt`, perhaps?

src/Mono.Android/metadata
```diff
@@ -1528,4 +1528,27 @@
...
 
+  <!-- Api 30 adjustments -->
+  <!-- We need to remove abstract method close from CameraOfflineSession since it is already defined on base abstract class CameraCaptureSession -->
+  <remove-node api-since="30" path="/api/package[@name='android.hardware.camera2']/class[@name='CameraOfflineSession']/method[@name='close' and count(parameter)=0]" />
+
+  <!-- Namespace android.view.inline should be renamed to Android.Views.Inline -->
+  <attr api-since="30" path="/api/package[@name='android.view.inline']" name="managedName">Android.Views.Inline</attr>
+
+  <!-- Google has added getCellIdentity and getCellSignalStrength as abstract method on the base class -->
+  <attr api-since="30" path="/api/package[@name='android.telephony']/class[@name='CellInfoCdma']/method[@name='getCellIdentity' and count(parameter)=0]" name="return">android.telephony.CellIdentity</attr>
```
jonpryor: Why do we need to override the `return` attribute?  That feels dubious/problematic.

tests/api-compatibility/acceptable-breakages-v10.0.99.txt
jonpryor: @gugavaro notes that the version number `v10.0.99` in this file should prevent us from accidentally including the contents into a stable version, so I think I'm overthinking things here.

src/Mono.Android/metadata
jonpryor:
```zsh
% javap -s -classpath ~/android-toolchain/sdk/platforms/android-29/android.jar android.telephony.CellInfoCdma
Compiled from "CellInfoCdma.java"
public final class android.telephony.CellInfoCdma extends android.telephony.CellInfo implements android.os.Parcelable {
  public android.telephony.CellIdentityCdma getCellIdentity();
    descriptor: ()Landroid/telephony/CellIdentityCdma;
  ...
}
```

The return signature is `Landroid/telephony/CellIdentityCdma;`.  This change should *prevent* the member from being usable, as `JNIEnv.GetMethodID(..., "getCellIdentity", "()Landroid/telephony/CellIdentity;")` should return NULL.

PR #4333, opened 2020-02-26
[Xamarin.Android.Build.Tasks] clean obj if $(XamarinAndroidVersion) changes

jonathanpeppers:
I found a crash at runtime, if you:

1. Build / Run a project with 16.4

2. Build / Run the same project with 16.6

The stack trace is:

    java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "jm_typemap" referenced by "/data/app/com.companyname.app59-bgIEcmaPpoMxpR948Wo8ZA==/lib/x86/libmonodroid.so"...
        at java.lang.Runtime.loadLibrary0(Runtime.java:1071)
        at java.lang.Runtime.loadLibrary0(Runtime.java:1007)
        at java.lang.System.loadLibrary(System.java:1667)
        at mono.MonoPackageManager.LoadApplication(MonoPackageManager.java:80)
        at mono.MonoRuntimeProvider.attachInfo(MonoRuntimeProvider.java:48)
        at android.app.ActivityThread.installProvider(ActivityThread.java:6983)
        at android.app.ActivityThread.installContentProviders(ActivityThread.java:6528)
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6445)
        at android.app.ActivityThread.access$1300(ActivityThread.java:219)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1859)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)

What happened is in ce2bc689, the `jm_typemap` and `mj_typemap`
symbols were removed. But we don't actually have anything that cleans
the generated assembly files when `$(XamarinAndroidVersion)` changes.

We have a `DeleteBinObjTest` test, why didn't that catch this?

Apparently the test was triggering:

    Building target "_CleanIntermediateIfNuGetsChange" completely.
    Input file "HelloForms.Android\obj\project.assets.json" is newer than output file "obj\Debug\90\stamp\_CleanIntermediateIfNuGetsChange.stamp".

And so the test wasn't actually *testing* much of anything!

As part of that test, we have to run a NuGet restore. The zip file
contains a project that was restored on another machine. We *have* to
run `/t:Restore` for things to build.

As a solution for this, I added some code to update the timestamp on
`_CleanIntermediateIfNuGetsChange.stamp`. Although not ideal, this
allows us to test what would really happen to a project during Xamarin
updates.

To actually fix the bug, I added a new
`_CleanIntermediateIfXAVersionChanges` target very similar to the
`_CleanIntermediateIfNuGetsChange` target. I think this will go a long
way to systematically fix #deletebinobj problems during Xamarin
updates.

I also updated the test to include projects built with 16.4 and verify
the app launches.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ReadXamarinAndroidVersion.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ReadXamarinAndroidVersion.cs
@@ -1,0 +1,40 @@
+using System;
+using System.IO;
+using Microsoft.Build.Framework;
+
+namespace Xamarin.Android.Tasks
+{
+	/// <summary>
+	/// Reads "XamarinAndroidVersion=10.2" from $(_AndroidBuildPropertiesCache) / build.props
+	/// </summary>
+	public class ReadXamarinAndroidVersion : AndroidTask
+	{
+		public override string TaskPrefix => "RXAV";
+
+		[Required]
+		public string BuildPropertiesCache { get; set; }
+
+		[Output]
+		public string XamarinAndroidVersion { get; set; }
+
+		public override bool RunTask ()
+		{
+			if (File.Exists (BuildPropertiesCache)) {
+				using (var reader = File.OpenText (BuildPropertiesCache)) {
+					while (!reader.EndOfStream) {
+						string line = reader.ReadLine ();
+						int index = line.IndexOf ('=');
+						if (index != -1) {
+							string key = line.Substring (0, index);
+							if (string.Equals ("XamarinAndroidVersion", key, StringComparison.OrdinalIgnoreCase)) {
+								XamarinAndroidVersion = line.Substring (index + 1, line.Length - index - 1);
+								break;
+							}
+						}
+					}
+				}
+			}
+			return !Log.HasLoggedErrors;
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
> Is it worth just writing the version to a file on its own in the intermediate dir? Then we can just use Read/WriteLinesToFile and not need to create a new Task?

The problem was it has to work with people's `obj` directories they have now in 16.4 (or older). I think we have to use the existing file to know what version they have.

jonathanpeppers:
It does look like I messed up the file I uploaded to Azure, maybe there is a typo in the URL:
```
System.Net.WebException : The remote server returned an error: (404) Not Found.
```

jonpryor:
Why didn't `@(_PropertyCacheItems)` and the `_CreatePropertiesCache` and related targets handle this already?  `@(_PropertyCacheItems)` contains `XamarinAndroidVersion=$(XamarinAndroidVersion)`, which *should* change on XA version, so...

Why do we need most of what's in this PR?  Why didn't our existing infrastructure handle it already?

jonathanpeppers:
> Why do we need most of what's in this PR? Why didn't our existing infrastructure handle it already?

`build.props` will trigger all the targets to run, that part works. Unfortunately, it doesn't actually _delete_ any leftover files, and that seems to be causes the crash above.

We have parts throughout the build that use wildcards to find files. At some point if we got around to refactoring everything, I think it would be possible to make the build completely correct when updating Xamarin.Android. That might be a dream, though.

Since `_CleanIntermediateIfNuGetsChange` was added in 16.0, I think it has considerably helped things.

jonathanpeppers:
Let me revisit this after my vacation.

I like the idea of using `build.props` to nuke `obj` -- that seems like it might be an even better solution.

jonathanpeppers:
I'm closing this in favor of: https://github.com/xamarin/xamarin-android/pull/4360

PR #4342, opened 2020-02-28
[xaprepare] Do less when updating Mono

grendello:
When updating Mono from an older version (e.g. 6.6) `xaprepare` can
crash as soon as the Mono package is installed with exceptions similar
to the ones below:

    System.InvalidOperationException: Failed to obtain version information from the cc compiler
      at Xamarin.Android.Prepare.Unix.DetectCompilers () [0x00344] in build-tools/xaprepare/xaprepare/OperatingSystems/Unix.cs:102
      at Xamarin.Android.Prepare.OS.ConfigureEnvironment () [0x00096] in build-tools/xaprepare/xaprepare/OperatingSystems/OS.cs:371
      at Xamarin.Android.Prepare.OS.Init () [0x000e6] in build-tools/xaprepare/xaprepare/OperatingSystems/OS.cs:254
      at Xamarin.Android.Prepare.Context.Init (System.String scenarioName) [0x0026c] in build-tools/xaprepare/xaprepare/Application/Context.cs:761
      at Xamarin.Android.Prepare.App.Run (System.String[] args) [0x0078a] in build-tools/xaprepare/xaprepare/Main.cs:153

    System.InvalidOperationException: Unable to determine the last version change commit
      at Xamarin.Android.Prepare.BuildInfo.DetermineLastVersionChangeCommit (Xamarin.Android.Prepare.Context context) [0x000fa] in build-tools/xaprepare/xaprepare/Application/BuildInfo.cs:130
      at Xamarin.Android.Prepare.BuildInfo.GatherGitInfo (Xamarin.Android.Prepare.Context context) [0x0006d] in build-tools/xaprepare/xaprepare/Application/BuildInfo.cs:36
      at Xamarin.Android.Prepare.Context.Init (System.String scenarioName) [0x00430] in build-tools/xaprepare/xaprepare/Application/Context.cs:780
      at Xamarin.Android.Prepare.App.Run (System.String[] args) [0x0078a] in build-tools/xaprepare/xaprepare/Main.cs:153

This is most probably caused by the new Mono changing assemblies/shared libraries
underneath the older Mono which is currently executing `xaprepare`.

In order to avoid the crash, do as little as possible to update Mono and quit as
quickly as possible.

### DIFF
--- a/build-tools/xaprepare/xaprepare/Application/Context.cs
+++ b/build-tools/xaprepare/xaprepare/Application/Context.cs
@@ -765,16 +765,21 @@ namespace Xamarin.Android.Prepare
 
 			Tools.Init (this);
 
-			Banner ("Updating Git submodules");
+			if (SelectedScenario.NeedsGitSubmodules) {
+				Banner ("Updating Git submodules");
 
-			var git = new GitRunner (this);
-			if (!await git.SubmoduleUpdate ()) {
-				Log.ErrorLine ("Failed to update Git submodules");
-				return false;
+				var git = new GitRunner (this);
+				if (!await git.SubmoduleUpdate ()) {
+					Log.ErrorLine ("Failed to update Git submodules");
+					return false;
+				}
 			}
 
 			BuildInfo = new BuildInfo ();
-			await BuildInfo.GatherGitInfo (this);
+			if (SelectedScenario.NeedsGitBuildInfo) {
+				await BuildInfo.GatherGitInfo (this);
+			}
+
 			AbiNames.LogAllNames (this);
 
 			if (MakeConcurrency == 0)
--- a/build-tools/xaprepare/xaprepare/Application/Scenario.cs
+++ b/build-tools/xaprepare/xaprepare/Application/Scenario.cs
@@ -10,6 +10,9 @@ namespace Xamarin.Android.Prepare
 		public string Description { get; }
 		public string LogFilePath { get; protected set; }
 		public List<Step> Steps   { get; } = new List<Step> ();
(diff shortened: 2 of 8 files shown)


build-tools/xaprepare/xaprepare/Application/Context.cs
```diff
@@ -765,16 +765,20 @@ public async Task<bool> Init (string scenarioName = null)
 
 			Tools.Init (this);
 
-			Banner ("Updating Git submodules");
```
jonpryor: Why is it that `Context.Init()` -- which runs for *all* scenarios -- always updates git submodules?  Wouldn't it make sense to move this into the specific scenarios which requires it, instead of *all* scenarios, so that the `UpdateMono` scenario *doesn't* update git submodules?

This would avoid adding `KnownConditions.MonoUpdateKludge`.

grendello: It makes sense to run submodule update only in the main scenario, but it wouldn't remove the need for `KnownConditions.MonoUpdateKludge` - the `BuildInfo` lines fail as well.

grendello: Also, `ConfigureEnvironment` has to be avoided as well in this case, so the condition is still needed

pjcollins: Can we make all of those steps use a scenario based condition rather than adding this new condition? It makes sense to skip all this stuff if we aren't running `Required` or `Standard` scenarios I think?

PR #4347, opened 2020-03-02
Bump to xamarin/java.interop@cedf4d0

pjcollins:
* Bump to xamarin/java.interop@cedf4d0
* [Mono.Android] Do not check target framework of referenced console apps
These project references only exist to ensure a particular build order,
so we can skip target framework version compatibility checks.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
@@ -322,7 +322,6 @@
       <_ILRepackArgs>/out:&quot;$(MSBuildThisFileDirectory)$(IntermediateOutputPath)$(AssemblyName).dll&quot; /internalize</_ILRepackArgs>
       <_ILRepackArgs>$(_ILRepackArgs) &quot;$(MSBuildThisFileDirectory)$(IntermediateOutputPath)$(AssemblyName).dll&quot;</_ILRepackArgs>
       <_ILRepackArgs>$(_ILRepackArgs) @(_InputAssembliesThatExist->'&quot;%(Identity)&quot;', ' ')</_ILRepackArgs>
-      <_ILRepackArgs>$(_ILRepackArgs) /lib:&quot;$(XAInstallPrefix)xbuild-frameworks\MonoAndroid\v1.0\facades&quot;</_ILRepackArgs> <!-- Path to netstandard.dll -->
     </PropertyGroup>
     <Exec
       Command="&quot;$(ILRepack)&quot; $(_ILRepackArgs)"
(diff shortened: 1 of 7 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
```diff
@@ -322,6 +322,7 @@
       <_ILRepackArgs>/out:&quot;$(MSBuildThisFileDirectory)$(IntermediateOutputPath)$(AssemblyName).dll&quot; /internalize</_ILRepackArgs>
       <_ILRepackArgs>$(_ILRepackArgs) &quot;$(MSBuildThisFileDirectory)$(IntermediateOutputPath)$(AssemblyName).dll&quot;</_ILRepackArgs>
       <_ILRepackArgs>$(_ILRepackArgs) @(_InputAssembliesThatExist->'&quot;%(Identity)&quot;', ' ')</_ILRepackArgs>
+      <_ILRepackArgs>$(_ILRepackArgs) /lib:&quot;$(XAInstallPrefix)xbuild-frameworks\MonoAndroid\v1.0\facades&quot;</_ILRepackArgs> <!-- Path to netstandard.dll -->
```
jonathanpeppers: I'm wondering if this is right... `netstandard.dll` in this case is coming from our build tree, and meant for Xamarin.Android projects. Xamarin.Android.Build.Tasks is a desktop assembly, should it be using Mono's system install of `netstandard.dll` instead?

jonpryor:

```
Changes: https://github.com/xamarin/java.interop/compare/bd7c60a6a5ea8abaf6f241da17c42d2d93226bf2...cedf4d04c7bf1f327d615a0316cbeabe59be5454

  * xamarin/Java.Interop@cedf4d0: [build] Convert more projects to Short-Form projects (#547)

Update `src/Mono.Android` to not check target framework of referenced
console apps.  These project references only exist to ensure a
particular build order, so we can skip target framework version
compatibility checks.
```

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
pjcollins: We were previously repacking against the netstandard.dll that was copied into this directory by some Java.Interop project. Maybe we can do something similar to what macios is doing here... https://github.com/xamarin/xamarin-macios/blob/0557f16e6a936f9d3c94fe020536141296b42640/msbuild/ILMerge.targets#L13

pjcollins: I just rebuilt on macOS with the latest changes, and we should now using a path from the .NET Core installation:

> /lib:"/usr/local/share/dotnet/sdk/NuGetFallbackFolder/netstandard.library/2.0.3/build/netstandard2.0/ref/"

pjcollins: Looks like Windows wasn't happy with this approach

pjcollins:
Re-running, we hit a bot which doesn't have powershell installed...
/azp run

PR #4351, opened 2020-03-03
[build] Remove doc generation duplication

pjcollins:
Context: https://github.com/xamarin/monodroid/pull/1076

Removes the duplicate docs reference and doc generation
from the inverted build system.

PR #4355, opened 2020-03-04
[build] Add BCL doc redirects to OSS installers

brendanzagaeski:
Context: 1d053ec45b416c2ad1d8e2b702ab201b595f7975

Commit 1d053ec4 added the classlib docs and doc redirect files for
`Mono.Android.dll` and `OpenTK-1.0.dll` to the OSS installers, but it
unintentionally overlooked the doc redirect files for the BCL
assemblies.

Those redirect files don't require anything from monodroid, so it makes
sense to generate them as part of the xamarin-android build and include
them into the Windows OSS installers.

Other changes:

Remove the unneeded doc redirect files for the `Mono.*` assemblies.
.NET Framework does not have documentation for those assemblies, so the
redirect files aren't useful.

Include `Xamarin.Android.NUnitLite.xml` into the Windows OSS installers.
`Xamarin.Android.NUnitLite.csproj` already generated that file during
OSS builds, but the file was not yet included in the installers.

### DIFF
--- a/build-tools/installers/create-installers.targets
+++ b/build-tools/installers/create-installers.targets
@@ -32,6 +32,9 @@
     <_MonoDocAssembly Include="@(_MsxDocAssembly);OpenTK-1.0;Xamarin.Android.NUnitLite">
       <SourceDir>$(_AndroidApiDocsPath)docs\%(Identity)\en\</SourceDir>
     </_MonoDocAssembly>
+    <_BclDocsAssembly Include="mscorlib;System.Core;System.ComponentModel.Composition;System.ComponentModel.DataAnnotations;System.Data;System" />
+    <_BclDocsAssembly Include="System.EnterpriseServices;System.Json;System.Numerics;System.Runtime.Serialization;System.ServiceModel" />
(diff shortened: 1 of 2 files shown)


brendanzagaeski:
These changes are motivated by Peter's nice investigation in <https://github.com/xamarin/monodroid/pull/1076> and <https://github.com/xamarin/xamarin-android/pull/4351>. This PR https://github.com/xamarin/xamarin-android/pull/4355 will help make those other PRs even more self-contained.

brendanzagaeski:
To make sure I got all of the remaining doc files this time, I have now checked the list of `.xml` files in the OSS PR `.vsix` against the list from the commercial master build (b054f38f). As desired, the only differences are for the removed `Mono*.xml` files. (The OSS PR build only includes one target framework version, but that is as expected for those PR builds.)

I also checked a few of the generated BCL doc redirect files and made sure they had the same content in the two installers.

brendanzagaeski:
I just realized the background story about `Xamarin.Android.NUnitLite.xml` that I put in the commit message isn't quite right. I'll push a quick change to fix that.

brendanzagaeski:

Correct the info about `Xamarin.Android.NUnitLite.xml` in the commit message.

PR #4360, opened 2020-03-05
[Xamarin.Android.Build.Tasks] clean $(IntermediateOutputPath) if build.props changes

jonathanpeppers:
I found a crash at runtime, if you:

1. Build / Run a project with 16.4

2. Build / Run the same project with 16.6

The stack trace is:

    java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "jm_typemap" referenced by "/data/app/com.companyname.app59-bgIEcmaPpoMxpR948Wo8ZA==/lib/x86/libmonodroid.so"...
        at java.lang.Runtime.loadLibrary0(Runtime.java:1071)
        at java.lang.Runtime.loadLibrary0(Runtime.java:1007)
        at java.lang.System.loadLibrary(System.java:1667)
        at mono.MonoPackageManager.LoadApplication(MonoPackageManager.java:80)
        at mono.MonoRuntimeProvider.attachInfo(MonoRuntimeProvider.java:48)
        at android.app.ActivityThread.installProvider(ActivityThread.java:6983)
        at android.app.ActivityThread.installContentProviders(ActivityThread.java:6528)
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:6445)
        at android.app.ActivityThread.access$1300(ActivityThread.java:219)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1859)
        at android.os.Handler.dispatchMessage(Handler.java:107)
        at android.os.Looper.loop(Looper.java:214)
        at android.app.ActivityThread.main(ActivityThread.java:7356)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:492)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:930)

What happened is in ce2bc689, the `jm_typemap` and `mj_typemap`
symbols were removed. But we don't actually have anything that cleans
the generated assembly files when `$(XamarinAndroidVersion)` changes.

We have a `DeleteBinObjTest` test, why didn't that catch this?

Apparently the test was triggering:

    Building target "_CleanIntermediateIfNuGetsChange" completely.
    Input file "HelloForms.Android\obj\project.assets.json" is newer than output file "obj\Debug\90\stamp\_CleanIntermediateIfNuGetsChange.stamp".

And so the test wasn't actually *testing* much of anything!

As part of that test, we have to run a NuGet restore. The zip file
contains a project that was restored on another machine. We *have* to
run `/t:Restore` for things to build. The changes in this PR remove
`_CleanIntermediateIfNuGetsChange`, so we should be OK going forward.

I also updated the test to include projects built with 16.4 and verify
the app launches.

## Actual Changes ##

To fix the real problem here, I think we need to fundamentally change
what happens when values in `build.props` /
`$(_AndroidBuildPropertiesCache)` changes. Currently we are using it
to trigger several MSBuild targets, but I believe we should also use
it to trigger the `_CleanMonoAndroidIntermediateDir` target if the
file changes.

1. Repurpose `_CleanIntermediateIfNuGetsChange` to instead be a
   `_CleanIntermediateIfNeeded` target that does not run on
   design-time builds.

2. Include the `GetLastWriteTime` value of `$(_NuGetAssetsFile)` in
   `build.props`.

3. Remove `DesignTimeBuild=$(DesignTimeBuild)` from `build.props`. We
   use a different path for DTBs at `obj\Debug\designtime\build.props`
   anyway.

4. `_CleanMonoAndroidIntermediateDir` can't delete `build.props`
   anymore. I moved this so it will still happen during a full
   `Clean`.

5. Remove the `_CleanIntermediateIfPackageNamingPolicyChanges` and
   `_CleanupOldStaticResources` MSBuild targets. They are no longer
   needed at all.

6. Remove `Xamarin.Android.DefaultOutputPaths.targets`.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -936,6 +936,7 @@ stages:
       inputs:
         artifactName: Test Results - Designer - macOS
         targetPath: $(Build.ArtifactStagingDirectory)/designer-binlogs
+      condition: always()
 
   # Check - "Xamarin.Android (Test Designer Windows)"
   - job: designer_integration_win
@@ -999,6 +1000,7 @@ stages:
       inputs:
         artifactName: Test Results - Designer - Windows
         targetPath: $(Build.ArtifactStagingDirectory)/designer-binlogs
+      condition: always()
 
 - stage: integrated_regression_test
   displayName: Regression Tests
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -572,9 +572,8 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
     _ValidateLinkMode;
     _CheckNonIdealConfigurations;
     _SetupDesignTimeBuildForBuild;
-    _CleanIntermediateIfNuGetsChange;
-    _CleanIntermediateIfPackageNamingPolicyChanges;
     _CreatePropertiesCache;
+    _CleanIntermediateIfNeeded;
     _CheckProjectItems;
     _CheckForContent;
     _CheckTargetFramework;
@@ -592,9 +591,8 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
     _ValidateLinkMode;
     _CheckNonIdealConfigurations;
      _SetupDesignTimeBuildForBuild;
-    _CleanIntermediateIfNuGetsChange;
-    _CleanIntermediateIfPackageNamingPolicyChanges;
-     _CreatePropertiesCache;
+    _CreatePropertiesCache;
+    _CleanIntermediateIfNeeded;
     _AddAndroidDefines;
     _CreateNativeLibraryArchive;
     _AddAndroidEnvironmentToCompile;
@@ -668,6 +666,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<CleanDependsOn>
 		$(CleanDependsOn);
 		_CleanMonoAndroidIntermediateDir;
+		_CleanAndroidBuildPropertiesCache;
 	</CleanDependsOn>
 </PropertyGroup>
 
@@ -721,37 +720,15 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	</PropertyGroup>
 </Target>
 
-<Target Name="_BeforeCleanIntermediateIfNuGetsChange">
-  <PropertyGroup>
-    <_NuGetAssetsFile Condition="Exists('$(ProjectLockFile)')">$(ProjectLockFile)</_NuGetAssetsFile>
-    <_NuGetAssetsFile Condition="'$(_NuGetAssetsFile)' == '' and Exists('packages.config')">packages.config</_NuGetAssetsFile>
-  </PropertyGroup>
-</Target>
-
-<Target Name="_CleanIntermediateIfNuGetsChange" DependsOnTargets="_BeforeCleanIntermediateIfNuGetsChange"
-    Inputs="$(_NuGetAssetsFile)"
-    Outputs="$(_AndroidStampDirectory)_CleanIntermediateIfNuGetsChange.stamp">
-  <!--NOTE: only want to clean if the stamp exists, *after* an initial build-->
-  <CallTarget Targets="_CleanMonoAndroidIntermediateDir" Condition=" Exists ('$(_AndroidStampDirectory)_CleanIntermediateIfNuGetsChange.stamp') " />
+<Target Name="_CleanIntermediateIfNeeded"
+    Condition=" '$(DesignTimeBuild)' != 'True' "
+    DependsOnTargets="_CreatePropertiesCache"
+    Inputs="$(_AndroidBuildPropertiesCache)"
+    Outputs="$(_AndroidStampDirectory)_CleanIntermediateIfNeeded.stamp">
+  <!--NOTE: only want to clean if build.props existed at the beginning of the build -->
+  <CallTarget Targets="_CleanMonoAndroidIntermediateDir" Condition=" '$(_AndroidBuildPropertiesCacheExists)' == 'True' " />
   <MakeDir Directories="$(_AndroidStampDirectory)" Condition=" !Exists('$(_AndroidStampDirectory)') " />
-  <Touch Files="$(_AndroidStampDirectory)_CleanIntermediateIfNuGetsChange.stamp" AlwaysCreate="true" />
-</Target>
-
-<Target Name="_CleanIntermediateIfPackageNamingPolicyChanges">
-  <ReadLinesFromFile File="$(IntermediateOutputPath)packagenamingpolicy.props">
-    <Output TaskParameter="Lines" PropertyName="_PackageNamingPolicy" />
-  </ReadLinesFromFile>
-  <CallTarget
-      Targets="_CleanMonoAndroidIntermediateDir"
-      Condition=" '$(AndroidPackageNamingPolicy)' != '$(_PackageNamingPolicy)' And Exists ('$(IntermediateOutputPath)packagenamingpolicy.props') "
-  />
-  <WriteLinesToFile
-      Condition=" '$(AndroidPackageNamingPolicy)' != '$(_PackageNamingPolicy)' "
-      File="$(IntermediateOutputPath)packagenamingpolicy.props"
-      Lines="$(AndroidPackageNamingPolicy)"
-      Overwrite="True"
-  />
-  <MakeDir Directories="$(_AndroidStampDirectory)" Condition=" !Exists('$(_AndroidStampDirectory)') " />
+  <Touch Files="$(_AndroidStampDirectory)_CleanIntermediateIfNeeded.stamp" AlwaysCreate="true" />
 </Target>
 
 <Target Name="_AddAndroidDefines"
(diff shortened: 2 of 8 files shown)


brendanzagaeski:
*Draft release notes*

Even though this PR itself is still a Draft, I took a try at writing up a release note for, in part just to see if it would make sense or be interesting at this stage. It was sort of interesting. This first draft might be a little overcautious, but maybe users will find the extra info useful anyway.

Depending on how this PR proceeds, feel free to discard the note or edit it as desired and add it to the PR. (Additional info about preparing draft release notes can now be found in the [release notes README](https://github.com/xamarin/xamarin-android/blob/master/Documentation/release-notes/README.md).)

```
### More aggressive incremental clean of `obj` directory for certain scenarios

In previous versions, Xamarin.Android reserved a costlier, more complete
incremental clean of the intermediate output directory for special scenarios
where a NuGet package changed or where the `$(AndroidPackageNamingPolicy)`
MSBuild property changed.  Xamarin.Android 10.3 expands this behavior to
cover several other MSBuild properties:

  *`$(AndroidAotMode)`
  *`$(AndroidCreatePackagePerAbi)`
  *`$(AndroidDexTool)`
  *`$(AndroidEmbedProfilers)`
  *`$(AndroidEnableProfiledAot)`
  *`$(AndroidExplicitCrunch)`
  *`$(AndroidGenerateJniMarshalMethods)`
  *`$(AndroidIncludeDebugSymbols)`
  *`$(AndroidLinkMode)`
  *`$(AndroidLinkSkip)`
  *`$(AndroidLinkTool)`
  *`$(AndroidNdkDirectory)`
  *`$(AndroidPackageFormat)`
  *`$(AndroidSdkBuildToolsVersion)`
  *`$(AndroidSdkDirectory)`
  *`$(AndroidSequencePointsMode)`
  *`$(AndroidUseLatestPlatformSdk)`
  *`$(AndroidUseSharedRuntime)`
  *`$(AotAssemblies)`
  *`$(BundleAssemblies)`
  *`$(DesignTimeBuild)`
  *`$(EmbedAssembliesIntoApk)`
  *`$(JavaSdkDirectory)`
  *`$(MonoSymbolArchive)`
  *`$(OS)`
  *`$(TargetFrameworkVersion)`
  *`$(XamarinAndroidVersion)`

This should reduce the number of cases where project authors need to run a full
clean by hand after changing one of these properties, such as after updating to
a new Xamarin.Android version.

> [!IMPORTANT]
> Changing any of these properties will now result in a longer incremental
> build time for the next build.
>
> That said, many of these properties are only recommended for Release
> configuration builds anyway, so they shouldn't be a concern for Debug
> configuration builds.
```

jonathanpeppers:
Thanks @brendanzagaeski, I'll add this, but I think we can leave off the `!IMPORTANT` note.

I don't think the build will actually be much slower if you change one of those properties. When `build.props` changes it triggers almost every target to run again already. The leftover files are what seem to cause a lot of problems. The extra file cleaning only seems to take about 500ms more time on my machine.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1745,32 +1728,6 @@ because xbuild doesn't support framework reference assemblies.
   </PropertyGroup>
 </Target>
 
-<Target Name="_CleanupOldStaticResources"
-    Condition=" Exists ('$(_AndroidIntermediateJavaSourceDirectory)mono\android\app\NotifyTimeZoneChanges.java') ">
-  <ItemGroup>
-    <_OldStaticResources Include="$(_AndroidIntermediateJavaSourceDirectory)mono\android\incrementaldeployment\MultiDexLoader.java" />
```
dellis1972: How many released did we switch over to the new pre built runtime? I think it should be safe to remove this code if we have had a couple of released between then and now

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1177,8 +1154,14 @@ because xbuild doesn't support framework reference assemblies.
 </PropertyGroup>
 
 <Target Name="_CreatePropertiesCache" DependsOnTargets="_SetupDesignTimeBuildForBuild;_SetLatestTargetFrameworkVersion;_ResolveMonoAndroidSdks">
+	<PropertyGroup>
+		<_AndroidBuildPropertiesCacheExists Condition=" Exists('$(_AndroidBuildPropertiesCache)') ">True</_AndroidBuildPropertiesCacheExists>
+		<_NuGetAssetsFile      Condition=" Exists('$(ProjectLockFile)') ">$(ProjectLockFile)</_NuGetAssetsFile>
+		<_NuGetAssetsFile      Condition=" '$(_NuGetAssetsFile)' == '' and Exists('packages.config') ">packages.config</_NuGetAssetsFile>
+		<_NuGetAssetsTimestamp Condition=" '$(_NuGetAssetsFile)' != '' ">$([System.IO.File]::GetLastWriteTime('$(_NuGetAssetsFile)').Ticks)</_NuGetAssetsTimestamp>
```
dellis1972: interesting trick :)

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
jonathanpeppers: I think the new clean behavior will delete these files now 👍 

We still have a test checking for these files getting deleted: https://github.com/xamarin/xamarin-android/blob/2fd73856bde0b0800b2455485cf15e6b4b393bec/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L3852-L3899

src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.DefaultOutputPaths.targets
```diff
@@ -1,25 +0,0 @@
-<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
jonpryor: Given that @adalon & @kzu *added* this file in 13d216fd86b4419eee8029abecf2ef24115edf26, we might want them to chem in on it's *removal*...

kzu: Since the purpose of this PR is to improve the way the incremental clean works, I'm fine having the previous mechanism to improve its perf/reliability be removed too. I suspect changing the target framework / API level for a project isn't that common anyway (maybe less so even in the .NET5 way some day...)

jonathanpeppers:
Don't merge this yet, we think there might be potentially some break in the designer:

```
1) Failed : Xamarin.AndroidDesigner.Tests.AndroidDesignerSessionEmptyBuildTest.CanRenderExternalJavaCustomControl
  Expected string length 34 but was 37. Strings differ at index 0.
  Expected: "Android.Support.V7.Widget.CardView"
  But was:  "com.android.layoutlib.bridge.MockView"
  -----------^
  at Xamarin.AndroidDesigner.Tests.AndroidDesignerSessionEmptyBuildTest.CanRenderExternalJavaCustomControl () [0x00000] in <1a9c0c81aa594cbfbde7c060877c8743>:0 
  at System.Runtime.CompilerServices.AsyncMethodBuilderCore+MoveNextRunner.InvokeMoveNext (System.Object stateMachine) [0x00000] in <436d1bba8d9a4d9da4d81e343dca2215>:0 
```

https://xamarinhq.slack.com/archives/C03CEGRUW/p1583941778431000

jonathanpeppers:
Doing another rerun since this got merged: https://github.com/xamarin/designer/pull/2821

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1193,9 +1176,9 @@ because xbuild doesn't support framework reference assemblies.
 		<_PropertyCacheItems Include="AndroidCreatePackagePerAbi=$(AndroidCreatePackagePerAbi)" />
 		<_PropertyCacheItems Include="AndroidGenerateJniMarshalMethods=$(AndroidGenerateJniMarshalMethods)" />
 		<_PropertyCacheItems Include="OS=$(OS)" />
-		<_PropertyCacheItems Include="DesignTimeBuild=$(DesignTimeBuild)" />
```
jonathanpeppers: @garuma I think this change is what is breaking the designer tests. Many of the targets skip now, and they used to _not_ skip.

Do you know if there are some cases where `DesignTimeBuild` is `False` with the designer?

garuma: Yeah we set it for all the targets we run, that's the rundown:

- `UpdateAndroidResources` w/ DesignTimeBuild=true and AndroidUseManagedDesignTimeResourceGenerator=false
- `PrepareResources;_GenerateCompileInputs` w/ BuildProjectReferences=false, ContinueOnError=true, DesignTimeBuild=true, DeferredBuildSupported=true
- `GetExtraLibraryLocationsTargetName` w/ DesignTimeBuild=true

jonathanpeppers: Do you think a test could be leaving it off somehow?

Let me try reverting this line and see what happens.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -999,6 +1000,7 @@ stages:
       inputs:
         artifactName: Test Results - Designer - Windows
         targetPath: $(Build.ArtifactStagingDirectory)/designer-binlogs
+      condition: always()
```
jonathanpeppers: An earlier incarnation of this PR was breaking the designer tests. I forgot to add this to the commit message:
```
* [build] designer tests should always upload artifacts, even if tests fail
```

PR #4365, opened 2020-03-06
Updating ApiCompat nuget package version to take advantage of DIM features

gugavaro:
Commit [8ff203f](https://github.com/dotnet/arcade/commit/8ff203f47d83e89ac4b9310b4e902188e0344bf0) enables the option to DIM feature not be considered an Api breakage.

This behavior is conditional to argument --allow-default-interface-methods be provided.
The behavior cannot be default because not all languages implement DIM or even previous versions of C# like C#7, does not uderstand DIM, so for those consumers DIM is considered an API breakage.

Those changes update the version of the nuget package and adds the --allow-default-interface-methods flag on our workflow.


jonpryor:
See also: https://github.com/dotnet/arcade/pull/4987

gugavaro:
See also: https://github.com/dotnet/arcade/pull/4995

gugavaro:
See Also: https://github.com/dotnet/runtime/pull/33514

PR #4372, opened 2020-03-09
[Mono.Android] Enable constants on interfaces.

jpobst:
Context: https://github.com/xamarin/java.interop/issues/509

Beginning with C#8, interfaces may now contain constants.  This allows us to better match the `android.jar` we are binding. 

This PR:
- Enables the `generator` flag: `--lang-features=interface-constants`, which will generates constants on interfaces using C#8.
- This does not remove any previously created constants.  We will need to `[Obsolete]` them in a future PR.

API diff between `master` `API-30` and this PR: https://gist.github.com/jpobst/31c692a67a7f0389895b59cca92480ff

### DIFF
--- a/src/Mono.Android/Mono.Android.targets
+++ b/src/Mono.Android/Mono.Android.targets
@@ -108,7 +108,7 @@
       <_Api>$(IntermediateOutputPath)mcw\api.xml</_Api>
       <_Dirs>--enumdir=$(IntermediateOutputPath)mcw</_Dirs>
       <_FullIntermediateOutputPath>$([System.IO.Path]::GetFullPath('$(IntermediateOutputPath)'))</_FullIntermediateOutputPath>
-      <_LangFeatures Condition="$(AndroidApiLevel) &gt;= 30">--lang-features=default-interface-methods,nested-interface-types</_LangFeatures>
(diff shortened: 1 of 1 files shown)


jonpryor:
API diff is *yuge*.

brendanzagaeski:
*Draft release note*

This sounds like it might be interesting to mention under the *Preview bindings for Android 11 Developer Preview* section that will be in the release notes.

Here's what I'm thinking of adding for this PR:

```
### Preview bindings for Android 11 Developer Preview

<!-- ... the other notes about this new API level ... -->

A new feature of these bindings is that they take advantage of C# 8.0's support
for static fields on interfaces to provide constants at locations that match the
original Android APIs. For example,
`Android.Views.TextClassifiers.TextClassifierConsts.HintTextIsEditable` is now
available as `Android.Views.TextClassifiers.ITextClassifier.HintTextIsEditable`.
```

To suggest something different or to recommend excluding it from the release notes, feel free to reply in this PR or add a `Documentation/release-notes/4372.md` file with the new suggestion. Thanks in advance!

jpobst:
@brendanzagaeski Good call, I modified your comment to add some more details.

PR #4375, opened 2020-03-09
[Xamarin.Android.Build.Tasks] remove $(AndroidExplicitCrunch)

jonathanpeppers:
`$(AndroidExplicitCrunch)` is an experimental feature that was aimed
at improving incremental build times for projects containing lots of
images. Unfortunately, I have found that it also can cause projects to
run `aapt` on every build. It may (or may not) help build times
depending on the project.

I think now is a good time to remove `$(AndroidExplicitCrunch)`, since
is not supported with `aapt2`. We may be in an `aapt2`-only world soon.

Removing `$(AndroidExplicitCrunch)` will make it easier for us to
refactor things going forward. Projects using it will continue to
work, but just use the default `aapt` behavior.

I also removed some tests with:

    [Ignore ("Enable once Merge Resources is finished")]

This was a "merge resources" feature that was never finished. It
relied on `$(AndroidExplicitCrunch)`, so we can remove these tests.


jonathanpeppers:
The one test failure we've seen off and on:
```
error ANDKT0000: keytool error: java.io.EOFException
```

brendanzagaeski:
*Draft release notes*

```
### AndroidExplicitCrunch feature disabled

In previous Xamarin.Android versions, setting the `AndroidExplicitCrunch`
MSBuild property to `true` would configure the build to
[crunch][crunch-upstream] PNG drawable resources using a separate MSBuild task
before passing them to AAPT.

This feature was never compatible with projects configured to use AAPT2 and was
automatically disabled in those projects.  For projects configured to use the
older AAPT tool, this feature could cause builds to rerun AAPT on every build,
increasing build times.  Due to these limitations, this feature has now been
removed from Xamarin.Android.  Project authors who added this property to their
*.csproj* files in the past can now remove it.

[crunch-upstream]: https://developer.android.com/topic/performance/reduce-apk-size#crunch
```

Documentation/release-notes/4375.md
```diff
@@ -0,0 +1,16 @@
...
+In previous Xamarin.Android versions, setting the
+`AndroidExplicitCrunch` MSBuild property to `true` would configure the
+build to [crunch][crunch-upstream] PNG drawable resources using a
+separate MSBuild task before passing them to AAPT.
+
+This feature was never compatible with projects configured to use
+AAPT2 and was automatically disabled in those projects.  For projects
+configured to use the older AAPT tool, this feature could cause builds
+to rerun AAPT on every build, which may or may not improve build times
+depending on the project.  Due to these limitations, this feature has
```
brendanzagaeski:
```suggestion
to rerun AAPT on every build.  Due to these limitations, this feature has
```

I think "improve build times" reads as backwards in the current context. One option could be something like: "which could hurt overall build times." But after rereading this, I think an even simpler option could be to leave off the last part of the sentence altogether because "rerun ... on every build" already suggests that the behavior was undesirable.

jonathanpeppers:
We can ignore the one test failure `BuildAotApplicationAndBundleAndÜmläüts("armeabi-v7a",False,True)`:
```
[aot-compiler stderr] C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\ndk\arm-linux-androideabi-strip: unable to rename 'obj\Release\aot\armeabi-v7a\libaot-Java.Interop.dll.so.tmp'; reason: File exists
```
It looks like the first build was cancelled/timed out. On the re-run, it looks like this file was left behind.

PR #4385, opened 2020-03-11
[Xamarin.Android.Build.Tasks] only extract specific .jar files from .aar

jonathanpeppers:
Context: https://github.com/xamarin/AndroidX/issues/73
Context: https://developer.android.com/studio/projects/android-library.html#aar-contents
Context: https://groups.google.com/forum/#!topic/lint-dev/BM8q5EQ3POE

With the latest AndroidX bindings, we were encountering a build error
from r8 such as:

    You have Jar libraries, 
        lint.jar, 
        lint.jar, 
        lint.jar, 
    that have the identical name with inconsistent file contents. 
    Please make sure to remove any conflicting libraries in EmbeddedJar, InputJar and AndroidJavaLibrary. 

The following `.aar` contains a  `lint.jar`:

https://maven.google.com/androidx/fragment/fragment/1.2.2/fragment-1.2.2.aar

Apparently there is a feature for implementing your own custom `lint`
rules. These are placed in `lint.jar` and Android Studio runs them as
needed.

Reading the spec for `.aar` files, it looks like there are different
`.jar` files that may exist within:

* `/classes.jar`
* `/libs/*.jar`
* `/lint.jar`
* `/api.jar`

However, we currently just unzip *everything* inside `.aar` files. Any
`.jar` file extracted is passed to javac, r8, etc.

To fix this issue, I think we should only extract `classes.jar` or
`libs/*.jar`. This should prevent us from unzipping unnecessary files
and solve this particular issue.

I added/updated some tests around this scenario.

### DIFF
--- a/Documentation/release-notes/4385.md
+++ b/Documentation/release-notes/4385.md
@@ -1,0 +1,7 @@
+#### Bindings projects
+
+  * [GitHub PR 4385](https://github.com/xamarin/xamarin-android/pull/4385):
+    *You have Jar libraries, lint.jar, lint.jar, ... that have the identical
+    name with inconsistent file contents* build error could prevent apps from
+    consuming custom bindings libraries for certain Java libraries.  This was
+    discovered while binding new versions of the AndroidX libraries.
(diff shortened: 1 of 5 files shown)


brendanzagaeski:
*Draft release notes*

Would this change be of potential interest to users making their own bindings libraries or third-party bindings libraries? If so, feel free to add a note along the following lines to `Documentation/release-notes/4385.md`:

```
#### Bindings projects

  * [GitHub PR 4385](https://github.com/xamarin/xamarin-android/pull/4385):
    *You have Jar libraries, lint.jar, lint.jar, ... that have the identical
    name with inconsistent file contents* build error could prevent apps from
    consuming custom bindings libraries for certain Java libraries.  This did
    not yet affect any bindings libraries published by the Xamarin team.
```

PR #4392, opened 2020-03-13
[build] include designer binlog files from tests

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/4360

I fear that #4360 is causing a real problem from the designer unit
tests. However, I need the `.binlog` files to understand what is
happening.

We currently do not have these as artifacts, so this will add them.


jonathanpeppers:
The previous build looks like it worked: https://build.azdo.io/3549287

I just renamed and cleaned up a few things.

PR #4395, opened 2020-03-13
[Xamarin.Android.Build.Tasks] don't use temp files in <ConvertResourcesCases/>

jonathanpeppers:
I noticed that we were still using temporary files for
`AndroidResource.UpdateXmlResource`. I reworked this to use
`MemoryStreamPool.CreateStreamWriter` and generate the file in-memory.

We also had the code:

    if (srcmodifiedDate <= lastUpdate) {
        Log.LogDebugMessage ("  Skipping: {0}  {1} <= {2}", file, srcmodifiedDate, lastUpdate);
        continue;
    }
    Log.LogDebugMessage ("  Processing: {0}   {1} > {2}", file, srcmodifiedDate, lastUpdate);

I removed the "Skipping" log message, because we have found that
MSBuild log messages can hurt performance. In this case, it could
potentially log a lot of files, and it will help incremental builds to
only log files that changed.

Building the Toggl app on Windows:

https://github.com/jonathanpeppers/mobileapp/tree/peppers-fixes

Originally from:

https://github.com/toggl/mobileapp

    Before:
    613 ms  ConvertResourcesCases                      2 calls
    After:
    508 ms  ConvertResourcesCases                      2 calls

For this app, this will save ~100ms on an initial build, or builds
where Android resources change.

I also cleaned up a few things:

* Removed an unused `logMessage` parameter.

* Removed `ConvertResourcesCases.AdditionalResourceDirectories` so
  that both aapt & aapt2 will invoke `<ConvertResourcesCases/>` the
  same way.

    ResourceDirectories="$(MonoAndroidResDirIntermediate);@(_LibraryResourceDirectories)"

### DIFF
--- a/Documentation/release-notes/4395.md
+++ b/Documentation/release-notes/4395.md
@@ -1,0 +1,7 @@
+### Build and deployment performance
+
+  * [GitHub PR 4395](https://github.com/xamarin/xamarin-android/pull/4395):
+    Remove temporary file usage in the `<ConvertResourcesCases/>` MSBuild
+    task. This reduced the time for this MSBuild task from 613ms to 508ms
+    in a real-world project in an initial clean build or incremental builds
+    with Android resource changes.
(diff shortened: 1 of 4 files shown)


brendanzagaeski:
*Draft release notes*

This looks like a nice user-facing build time improvement. If you get a chance, you can add a draft release note for it to `Documentation/release-notes/4395.md` based on the [build and deployment performance template](https://github.com/xamarin/xamarin-android/blob/master/Documentation/release-notes/README.md#template-for-build-time-or-app-performance-improvement). Feel free to pick and choose parts of the template to suit this PR. Thanks in advance!

jonathanpeppers:
I can't catch a break today:
```
Microsoft.CSharp.Core.targets(59,5): error : Roslyn compiler server reports different hash version than build task. [/Users/builder/azdo/_work/2/s/xamarin-android/build-tools/xaprepare/xaprepare/xaprepare.csproj]

    9 Warning(s)
    1 Error(s)
```
Going to try re-running it.

PR #4396, opened 2020-03-13
[Mono.Android] Add logging around RegisterNativeMembers()

jonpryor:
Context: https://xamarinhq.slack.com/archives/CFB4MC864/p1584020384371400

Logging of which Java types we're registering native methods for was
removed in af02a65a.

Reintroduce a similar form of logging so that we can determine which
type is causing `Delegate.CreateDelegate()` to throw from
`AndroidTypeManager.RegisterNativeMembers()`.

The new log messages will only be printed when `timing` logging is
enabled:

	$ adb shell setprop debug.mono.log timing

For example:

	Registering native methods for managed type `Xamarin.Forms.Platform.Android.PlatformRenderer, Xamarin.Forms.Platform.Android, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null`
	Finished registering native methods for managed type `Xamarin.Forms.Platform.Android.PlatformRenderer, Xamarin.Forms.Platform.Android, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null` on Java type `crc643f46942d9dd1fff9/PlatformRenderer`; elapsed: 0s:2::587500

### DIFF
--- a/src/monodroid/jni/monodroid-glue.cc
+++ b/src/monodroid/jni/monodroid-glue.cc
@@ -1648,11 +1648,21 @@ MonodroidRuntime::Java_mono_android_Runtime_register (JNIEnv *env, jstring manag
 {
 	timing_period total_time;
 
+	char *type = nullptr;
+
 	if (XA_UNLIKELY (utils.should_log (LOG_TIMING)))
 		total_time.mark_start ();
 
 	int managedType_len = env->GetStringLength (managedType);
 	const jchar *managedType_ptr = env->GetStringChars (managedType, nullptr);
+
+	if (XA_UNLIKELY (utils.should_log (LOG_TIMING))) {
+		const char *mt_ptr = env->GetStringUTFChars (managedType, nullptr);
+		type = utils.strdup_new (mt_ptr);
+		env->ReleaseStringUTFChars (managedType, mt_ptr);
+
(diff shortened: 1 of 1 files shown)


jonpryor:
Should instead update: https://github.com/xamarin/xamarin-android/blob/8599d63d28922a580c75aeb015554ae9d5034bf3/src/monodroid/jni/monodroid-glue.cc#L1683-L1695

brendanzagaeski:
*Draft release notes*

Although the logging information provided by this PR would technically be visible to users, it sounds like it will mostly be of interest to and actionable for the Xamarin.Android team, so by default I'll skip over this PR for the release notes.

If there's anything you'd like to highlight for this PR in the release notes, feel free to add it to a `Documentation/release-notes/4396.md` file as part of this PR.

Thanks!

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1648,12 +1648,22 @@ MonodroidRuntime::Java_mono_android_Runtime_register (JNIEnv *env, jstring manag
 {
 	timing_period total_time;
 
+	char *type = NULL;
```
grendello: `nullptr` instead of `NULL`, please :)

PR #4397, opened 2020-03-13
Bump to xamarin/java.interop/master@bd7c60a6

jpobst:
Breakages:

## acceptable-breakages-v10.0.99.txt 

These are because we are no longer generating XXXXConsts classes for DIM enabled bindings as of https://github.com/xamarin/java.interop/pull/600

## acceptable-breakages-v10.0.txt 

These are because there was a bug where we were not generating `[Obsolete]` on fields that were turned into properties.  Now the attribute is generated in API-28 (the "contract"), but they are no longer marked as deprecated by Google in API-29 (the "implementation"), so they appear as removing the attribute.

api-28.xml.in
```
<field deprecated="deprecated" final="false" name="BSSID" jni-signature="Ljava/lang/String;" static="false" transient="false" type="java.lang.String" type-generic-aware="java.lang.String" visibility="public" volatile="false">
```

api-29.xml.in
```
<field deprecated="not deprecated" final="false" name="BSSID" jni-signature="Ljava/lang/String;" static="false" transient="false" type="java.lang.String" type-generic-aware="java.lang.String" visibility="public" volatile="false">
```

## acceptable-breakages-v8.0.txt 

Same as above, a propified field was missing `[Obsolete]`, and Google later un-deprecated the field.

api-25.xml.in
```
<field deprecated="deprecated" final="false" name="number" jni-signature="I" static="false" transient="false" type="int" type-generic-aware="int" visibility="public" volatile="false">
```

api-26.xml.in
```
<field deprecated="not deprecated" final="false" name="number" jni-signature="I" static="false" transient="false" type="int" type-generic-aware="int" visibility="public" volatile="false">
```


brendanzagaeski:
*Draft release notes*

This bump appears to bring in a nice user-facing build time improvement from <https://github.com/xamarin/java.interop/commit/0a3354bca451beb4142e1f285b7cdcc6aacbbee0>. 
If you or @jonathanpeppers get a chance, you can add a draft release note for it to `Documentation/release-notes/4397.md` based on the [build and deployment performance template](https://github.com/xamarin/xamarin-android/blob/master/Documentation/release-notes/README.md#template-for-build-time-or-app-performance-improvement). Feel free to pick and choose parts of the template to suit that commit.

(Side note: The different authors between https://github.com/xamarin/java.interop/commit/0a3354bca451beb4142e1f285b7cdcc6aacbbee0 and this PR makes me realize it might be useful to adjust the release note workflow to cover that scenario. At least for members of the Xamarin.Android product team, maybe each new user-facing PR in a submodule can include a PR comment with a draft release note. Then the bump in xamarin-android can combine all those comments into a `Documentation/release-notes/*.md` file.)

For the the generator changes in this bump, it looks like they might be mostly specific to the bindings process for `Mono.Android.dll`, so feel free to skip those for the release notes if they aren't of interest to authors of other bindings projects.  If you do want to mention any of them, they can go in the same `4397.md` file, placed under their own headings.

Thanks in advance!

jonathanpeppers:
@jpobst for mine add:

```
### Build and deployment performance

  * [GitHub PR 596](https://github.com/xamarin/java.interop/pull/596):
    Use `File.Exists` instead of `DirectoryGetFile`.  This reduced the time for the
    `LinkAssembliesNoShrink` task from about 711ms to 426ms for a small test
    Xamarin.Forms app on an initial clean build.
```

jpobst:
👍 Release notes added.

Issue #4394, opened 2020-03-13
net5 & Delegate.CreateDelegate() changes


jonpryor:
Context: https://github.com/dotnet/runtime/issues/33531
Context: https://github.com/mono/mono/issues/19207
Context: https://github.com/xamarin/xamarin-android/commits/test-net5-drop

### Steps to Reproduce

1. Bribe @pjcollins to get a xamarin-android build environment which uses a (very) preview .NET 5 runtime & mono package.
2. Run a Xamarin.Forms app, e.g. our integration test, or from a default template.

### Expected Behavior

It works!

### Actual Behavior

It fails:

```
UNHANDLED EXCEPTION:
System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.ArgumentException: Late bound operations cannot be performed on types or methods for which ContainsGenericParameters is true. (Parameter 'target')
   at System.Delegate.CreateDelegate(Type type, Type target, String method, Boolean ignoreCase, Boolean throwOnBindFailure)
   at System.Delegate.CreateDelegate(Type type, Type target, String method)
   at Android.Runtime.AndroidTypeManager.RegisterNativeMembers(JniType nativeClass, Type type, String methods)
   at Android.Runtime.JNIEnv.RegisterJniNatives(IntPtr typeName_ptr, Int32 typeName_len, IntPtr jniClass, IntPtr methods_ptr, Int32 methods_len)
   at Java.Interop.JniEnvironment.Object.AllocObject(JniObjectReference type)
   at Java.Interop.JniType.AllocObject()
   at Java.Interop.JniPeerMembers.JniInstanceMethods.StartCreateInstance(String constructorSignature, Type declaringType, JniArgumentValue* parameters)
   at Xamarin.Forms.Platform.Android.FormsViewGroup..ctor(Context p0)
   at Xamarin.Forms.Platform.Android.VisualElementRenderer`1[[Xamarin.Forms.Page, Xamarin.Forms.Core, Version=2.0.0.0, Culture=neutral, PublicKeyToken=null]]..ctor(Context context)
   at Xamarin.Forms.Platform.Android.PageRenderer..ctor(Context context)
   at System.Reflection.RuntimeConstructorInfo.InternalInvoke(Object obj, Object[] parameters, Boolean wrapExceptions)
   --- End of inner exception stack trace ---
[...]

PR #4398, opened 2020-03-13
[Mono.Android] Delegate.CreateDelegate & generic type definitions

jonpryor:
Fixes: https://github.com/xamarin/xamarin-android/issues/4394

When using a preview drop of the .NET 5 BCL with a Xamarin.Forms app,
app startup crashes within
`AndroidTypeManager.RegisterNativeMembers()` becuase we attempt to use
`Delegate.CreateDelegate()` on a generic type definition, a
`typeof(VisualElementRenderer<>)`.

.NET Core and .NET Framework doesn't like this; mono does, which is
why it hasn't been a problem before.

We attempt to register the generic type definition for
`VisualElementRenderer<T>` because that's what the Java Callable
Wrapper does!  Given C#:

	// C#
	partial class FormsViewGroup : Android.Views.ViewGroup {
	}
	partial class VisualElementRenderer<T> : FormsViewGroup {
	}
	partial class PageRenderer : VisualElementRenderer<Page> {
	}

the Java Callable Wrappers will "mirror" that hierarchy, with
`Runtime.register()` calls in the class constructors:

	// Java Callable Wrappers
	public class FormsViewGroup extends android.view.ViewGroup {}
	public class VisualElementRenderer_1 extends FormsViewGroup {
	  static {
	    mono.android.Runtime.register ("VisualElementRenderer`1, …");
	  }
	}
	public class PageRenderer extends VisualElementRenderer_1 {
	  static {
	    mono.android.Runtime.register ("VisualElementRenderer`1, …");
	  }
	}

When a `PageRenderer` instance is created from C#, an instance of the
Java Callable Wrapper `PageRenderer` will also be created, but before
the instance an be constructed the static initializer of
`PageRenderer` and all base classes must be executed *first*, and
because there's a `Runtime.register()` for
``VisualElementRenderer`1``,
`AndroidTypeManager.RegisterNativeMembers()` will thus attempt to
register the generic type definition.

How should this be fixed?

Assuming that updating .NET Core to behave like Mono isn't an option,
there are three other solutions:

 1. Don't emit `Runtime.register()` in the Java Callable Wrappers for
    C# types which are generic.

 2. Within `AndroidRuntime.RegisterNativeMembers()`, skip generic type
    definitions entirely.

 3. Improve the semantics of connector method lookups.

@jonpryor is reticent to do something as significant as (1) or (2)
without significant testing, as the side effects may be non-obvious.
For example, could this break [`GenericHolder<T>`][0] or some
unknown variation thereof?

Instead, let's go with (3).

The whole reason that we can lookup connector methods on a generic
type definition in the first place is because the connector methods
are in fact present in a *non*-generic base class, and Mono looks at
all base classes to resolve a method before deciding to throw.
Instead of relying on Mono, we can do this ourselves.

[0]: https://github.com/xamarin/xamarin-android/blob/8599d63d28922a580c75aeb015554ae9d5034bf3/src/Mono.Android/Test/Java.Interop/JnienvTest.cs#L468-L472

### DIFF
--- a/src/Mono.Android/Android.Runtime/AndroidRuntime.cs
+++ b/src/Mono.Android/Android.Runtime/AndroidRuntime.cs
@@ -364,6 +364,9 @@ namespace Android.Runtime {
 
 		public override void RegisterNativeMembers (JniType nativeClass, Type type, string methods)
 		{
+			if (type.IsGenericTypeDefinition) {
+				return;
+			}
 			if (FastRegisterNativeMembers (nativeClass, type, methods))
 				return;
 
@@ -394,8 +397,15 @@ namespace Android.Runtime {
 						throw new InvalidOperationException (String.Format ("Specified managed method '{0}' was not found. Signature: {1}", mname, toks [1]));
 					callback = CreateDynamicCallback (minfo);
 				} else {
+					Type callbackDeclaringType = type;
+					if (toks.Length == 4) {
+						callbackDeclaringType = Type.GetType (toks [3], throwOnError: true);
+					}
+					while (callbackDeclaringType.ContainsGenericParameters) {
+						callbackDeclaringType = callbackDeclaringType.BaseType;
+					}
 					GetCallbackHandler connector = (GetCallbackHandler) Delegate.CreateDelegate (typeof (GetCallbackHandler),
-						toks.Length == 4 ? Type.GetType (toks [3], true) : type, toks [2]);
+						callbackDeclaringType, toks [2]);
 					callback = connector ();
 				}
 				natives [i] = new JniNativeMethodRegistration (toks [0], toks [1], callback);
--- a/src/Mono.Android/Android.Runtime/AndroidRuntime.cs
+++ b/src/Mono.Android/Android.Runtime/AndroidRuntime.cs
@@ -364,9 +364,6 @@ namespace Android.Runtime {
 
 		public override void RegisterNativeMembers (JniType nativeClass, Type type, string methods)
 		{
-			if (type.IsGenericTypeDefinition) {
-				return;
-			}
 			if (FastRegisterNativeMembers (nativeClass, type, methods))
 				return;
 
--- a/tests/CodeGen-Binding/Xamarin.Android.JcwGen-Tests/ExceptionTests.cs
+++ b/tests/CodeGen-Binding/Xamarin.Android.JcwGen-Tests/ExceptionTests.cs
@@ -16,7 +16,7 @@ namespace Xamarin.Android.JcwGenTests {
 		public void ManagedJavaManaged_FinallyExecuted ()
 		{
 			using (var t = new Bxc7634 ()) {
-				using (var r = new Java.Lang.Runnable (() => {throw new InvalidOperationException ();})) {
+				using (var r = new MyGenericRunnable<int> ()) {
 					Assert.IsFalse (t.FinallyBlockRun);
 					bool ioeThrown = false;
 					try {
@@ -39,6 +39,14 @@ namespace Xamarin.Android.JcwGenTests {
 				}
 				Assert.IsNotNull (t.ThrowableCaught);
 				Assert.AreEqual ("Android.Runtime.JavaProxyThrowable", t.ThrowableCaught.GetType ().FullName);
+			}
+		}
+
+		class MyGenericRunnable<T> : Java.Lang.Object, Java.Lang.IRunnable {
+
+			public void Run ()
+			{
+				throw new InvalidOperationException ();
 			}
 		}
 	}


grendello:
Wouldn't https://github.com/xamarin/xamarin-android/issues/4394#issuecomment-599136742 be a better solution? `VisualElementRenderer` is declared as `abstract` - does it even make sense to register classes we can't instantiate? Avoiding this would make this particular instance of the problem disappear (not saying other instances of it cannot exist) without hurting the runtime performance:

```csharp
public abstract class VisualElementRenderer<TElement> : FormsViewGroup, IVisualElementRenderer, IDisposedState,
                IEffectControlProvider where TElement : VisualElement
```

jonpryor:
> Wouldn't https://github.com/xamarin/xamarin-android/issues/4394#issuecomment-599136742 be a better solution? 

In the abstract, no. In this particular case, maybe.

We could plausibly skip `Runtime.register()` invocations on `abstract` classes, but that doesn't "fix" or otherwise handle the whole "why are generic type definitions being registered" scenario, briefly described https://github.com/xamarin/xamarin-android/issues/4394#issuecomment-599152565

I think I'm going to need to add an actual example of this scenario, implement 2(a), and verify that it breaks in precisely the manner I *think* it will...

jonpryor:
Demonstration of the problem with 2(a)/(3): `tests/CodeGen-Binding/Xamarin.Android.JcwGen-Tests/ExceptionTests.cs` was updated to use this new type from the `ExceptionTests.ManagedJavaManaged_FinallyExecuted()` test:

```csharp
	class MyGenericRunnable<T> : Java.Lang.Object, Java.Lang.IRunnable {

		public void Run ()
		{
			throw new InvalidOperationException ();
		}
	}
```

When `RegisterNativeMembers()` does:

```csharp
	if (type.IsGenericTypeDefinition) {
			return;
	}
```

skipping registration of generic type definitions, `ExceptionTests.ManagedJavaManaged_FinallyExecuted()` fails:

```
Java.Lang.UnsatisfiedLinkError : No implementation found for void crc64eadbb4d7d7a577cd.ExceptionTests_MyGenericRunnable_1.n_run() (tried Java_crc64eadbb4d7d7a577cd_ExceptionTests_1MyGenericRunnable_11_n_1run and Java_crc64eadbb4d7d7a577cd_ExceptionTests_1MyGenericRunnable_11_n_1run__)
  at Java.Interop.JniEnvironment+InstanceMethods.CallVoidMethod (Java.Interop.JniObjectReference instance, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args) [0x00097] in <f5f8c37841194e01aed7d12794c9d34d>:0 
  at Java.Interop.JniPeerMembers+JniInstanceMethods.InvokeVirtualVoidMethod (System.String encodedMember, Java.Interop.IJavaPeerable self, Java.Interop.JniArgumentValue* parameters) [0x00030] in <f5f8c37841194e01aed7d12794c9d34d>:0 
  at Com.Xamarin.Android.Bxc7634.RunFinallyBlock (Java.Lang.IRunnable p0) [0x00038] in <ec6fa48ee70b46ba8517f7afafe0ed9e>:0 
  at Xamarin.Android.JcwGenTests.ExceptionTests.ManagedJavaManaged_FinallyExecuted () [0x0001e] in <2966f57b35b54c108e107d8d83b47199>:0 
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <00e39b3333b74d8d922b340489226267>:0 
  --- End of managed Java.Lang.UnsatisfiedLinkError stack trace ---
java.lang.UnsatisfiedLinkError: No implementation found for void crc64eadbb4d7d7a577cd.ExceptionTests_MyGenericRunnable_1.n_run() (tried Java_crc64eadbb4d7d7a577cd_ExceptionTests_1MyGenericRunnable_11_n_1run and Java_crc64eadbb4d7d7a577cd_ExceptionTests_1MyGenericRunnable_11_n_1run__)
        at crc64eadbb4d7d7a577cd.ExceptionTests_MyGenericRunnable_1.n_run(Native Method)
        at crc64eadbb4d7d7a577cd.ExceptionTests_MyGenericRunnable_1.run(ExceptionTests_MyGenericRunnable_1.java:30)
        at com.xamarin.android.Bxc7634.runFinallyBlock(Bxc7634.java:10)
        at crc64f295cabbf85394f5.TestSuiteInstrumentation.n_onStart(Native Method)
        at crc64f295cabbf85394f5.TestSuiteInstrumentation.onStart(TestSuiteInstrumentation.java:35)
        at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2196)
```

jonpryor:
I'll wait for the current build to finish so that CI can show the above error, then I'll push a new commit which removes the `type.IsGenericTypeDefinition` check from  `RegisterNativeMembers()` to show that it works.

jonpryor:
Now, could we change things so that if the generic type definition is *abstract* we either omit the `Runtime.register()` or skip the `RegisterNativeMembers()` invocation?  I *suspect* not, because even if you have an abstract class, that class may override virtual or interface members from a base class/interface while *not* requiring that all subclasses implement it, e.g. if `MyGenericRunnable<T>` were abstract while `IRunnable.Run()` were *not* abstract, then subclasses wouldn't override `run()` in the Java Callable Wrapper either, and things would fail.

grendello:
Could we change the generator to do more analysis on such troublesome classes? If we see that the abstract generic (or just abstract one) does not implement any Java interface methods, does not override any virtual Java methods and does not register any of its methods with Java - could we ignore the type then? Just thinking that we might want to do more work on build time to have better performance during execution time.

jonpryor:
...and the "real" error: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3553407&view=ms.vss-test-web.build-test-results-tab&runId=11836442&resultId=100018&paneView=debug

PR #4219, opened 2020-02-04
Bump to mono:2020-02

directhex:

PR #4405, opened 2020-03-16
[ci] Don't build XA.sln in parallel on Windows, as it sometimes fails.

jpobst:
The [2020-02 Mono bump](https://github.com/xamarin/xamarin-android/pull/4219) enabled `msbuild /m` for parallel builds in our CI for Windows.  However we have not audited our `Xamarin.Android.sln` to ensure build dependencies are correctly set.  Something(s) in there are not correct and this causes a race condition where builds sometimes fail because expected files do no exist yet.

Often this looks like this:
```
Error MSB3644: The reference assemblies for MonoAndroid,Version=v10.0 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks
```

This disables parallel builds until a future audit is completed.


jonathanpeppers:
I'm somewhat confused why `/m` was added for _Windows_ during a mono bump? 👀 

Maybe they meant to do it on macOS?

PR #4407, opened 2020-03-16
[Xamarin.Android.Build.Tasks] improvements to Set*Writeable

jonathanpeppers:
Context: https://github.com/jonathanpeppers/Benchmarks/tree/setwriteable

`MonoAndroidHelper` has a couple methods that are called a lot during
a build:

1. `public static void SetDirectoryWriteable (string directory)`
2. `public static void SetWriteable (string source)`

Both have some room for performance improvements.

## SetWriteable ##

I changed usage of `FileInfo` to just use the static `File` methods.
We can do this for files, but not directories. There is no
`Directory.GetAttributes` method in the BCL.

We had a `File.Exists` check, which all callers were already doing. I
changed this to use `try-catch` around `File.GetAttributes`.

Benchmarking the difference:

    BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18362
    Intel Core i9-9900K CPU 3.60GHz (Coffee Lake), 1 CPU, 16 logical and 8 physical cores
    [Host]     : .NET Framework 4.8 (4.8.4121.0), X86 LegacyJIT

    |     Method |      Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
    |----------- |----------:|---------:|---------:|------:|------:|------:|----------:|
    | Writeable2 |  36.20 us | 0.531 us | 0.443 us |     - |     - |     - |         - |
    | Writeable1 |  61.27 us | 0.826 us | 0.732 us |     - |     - |     - |         - |
    |  Readonly2 | 100.97 us | 1.274 us | 1.129 us |     - |     - |     - |         - |
    |  Readonly1 | 126.21 us | 1.441 us | 1.348 us |     - |     - |     - |         - |

## SetDirectoryWriteable ##

The main improvement is to *not* call `Directory.EnumerateDirectories`
and `Directory.EnumerateFiles` independently. Use
`DirectoryInfo.EnumerateFileSystemInfos` instead. This way we aren't
enumerating the file system twice.

It also doesn't need to call `SetWriteable` for files.

Benchmarking the difference, it is an order of magnitude better now:

    BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18362
    Intel Core i9-9900K CPU 3.60GHz (Coffee Lake), 1 CPU, 16 logical and 8 physical cores
    [Host]     : .NET Framework 4.8 (4.8.4121.0), X86 LegacyJIT

    |     Method |      Mean |    Error |   StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
    |----------- |----------:|---------:|---------:|------:|------:|------:|----------:|
    | Writeable2 | 158.51 us | 1.870 us | 1.749 us |     - |     - |     - |         - |
    |  Readonly2 | 159.54 us | 2.172 us | 2.032 us |     - |     - |     - |         - |
    | Writeable1 | 758.41 us | 5.558 us | 5.199 us |     - |     - |     - |   16384 B |
    |  Readonly1 | 763.97 us | 8.778 us | 8.211 us |     - |     - |     - |   16384 B |

## Overall results ##

On an initial build of the Xamarin.Forms integration project in this
repo:

    Before:
     461 ms  GenerateJavaStubs                          1 calls
    1061 ms  ResolveLibraryProjectImports               1 call
    After:
     448 ms  GenerateJavaStubs                          1 calls
     993 ms  ResolveLibraryProjectImports               1 calls

These two targets would call into `Set*Writeable` a lot because they
write lots of files.

This saves ~90ms on an initial build. Incremental builds will have a
smaller improvement.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/Files.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/Files.cs
@@ -80,6 +80,38 @@ namespace Xamarin.Android.Tools
 			return changed;
 		}
 
+		/// <summary>
+		/// NOTE: assumes the file exists, will throw FileNotFoundExcpetion.
+		/// </summary>
+		public static void SetWriteableUnchecked (string source)
+		{
+			var fromAttr = File.GetAttributes (source);
+			var toAttr = fromAttr & ~FileAttributes.ReadOnly;
+			if (fromAttr != toAttr) {
+				File.SetAttributes (source, toAttr);
+			}
+		}
+
+		public static void SetDirectoryWriteable (string directory)
+		{
+			var root = new DirectoryInfo (directory);
+			if (!root.Exists)
+				return;
+			SetWriteable (root);
+
+			foreach (FileSystemInfo child in root.EnumerateFileSystemInfos ("*", SearchOption.AllDirectories)) {
+				SetWriteable (child);
+			}
+		}
+
+		static void SetWriteable (FileSystemInfo info)
+		{
+			var fromAttr = info.Attributes;
+			var toAttr = fromAttr & ~FileAttributes.ReadOnly;
+			if (fromAttr != toAttr)
+				info.Attributes = toAttr;
+		}
+
 		public static bool CopyIfChanged (string source, string destination)
 		{
 			if (HasFileChanged (source, destination)) {
@@ -88,10 +120,12 @@ namespace Xamarin.Android.Tools
 					Directory.CreateDirectory (directory);
 
 				if (!Directory.Exists (source)) {
-					MonoAndroidHelper.SetWriteable (destination);
-					File.Delete (destination);
+					if (File.Exists (destination)) {
+						SetWriteableUnchecked (destination);
+						File.Delete (destination);
+					}
 					File.Copy (source, destination);
-					MonoAndroidHelper.SetWriteable (destination);
+					SetWriteableUnchecked (destination);
 					File.SetLastWriteTimeUtc (destination, DateTime.UtcNow);
 					return true;
 				}
@@ -114,8 +148,10 @@ namespace Xamarin.Android.Tools
 				if (!string.IsNullOrEmpty (directory))
 					Directory.CreateDirectory (directory);
 
-				MonoAndroidHelper.SetWriteable (destination);
-				File.Delete (destination);
+				if (File.Exists (destination)) {
+					SetWriteableUnchecked (destination);
+					File.Delete (destination);
+				}
 				File.WriteAllBytes (destination, bytes);
 				return true;
 			}
@@ -129,8 +165,10 @@ namespace Xamarin.Android.Tools
 				if (!string.IsNullOrEmpty (directory))
 					Directory.CreateDirectory (directory);
 
-				MonoAndroidHelper.SetWriteable (destination);
-				File.Delete (destination);
+				if (File.Exists (destination)) {
+					SetWriteableUnchecked (destination);
+					File.Delete (destination);
+				}
 				using (var fileStream = File.Create (destination)) {
 					stream.Position = 0; //HasStreamChanged read to the end
 					stream.CopyTo (fileStream);
@@ -179,7 +217,7 @@ namespace Xamarin.Android.Tools
 			return false;
 		}
 
-		public static bool HasZipChanged (Stream source, string destination, out string hash)
+		static bool HasZipChanged (Stream source, string destination, out string hash)
 		{
 			hash = null;
 
@@ -196,7 +234,7 @@ namespace Xamarin.Android.Tools
 			return src_hash != dst_hash;
 		}
 
-		public static bool HasZipChanged (string source, string destination, out string hash)
+		static bool HasZipChanged (string source, string destination, out string hash)
 		{
 			hash = null;
 			if (!File.Exists (source))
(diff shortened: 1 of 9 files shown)


jonpryor:
In the `SetWritable` > Benchmarking the difference section, I don't know what the benchmarks *are*.  What is `Writeable` vs. `Writeable2` vs. `Readonly` vs. `Readonly2` regarding *files* vs. *directories*?

src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
```diff
@@ -363,32 +363,38 @@ public static bool IsForceRetainedAssembly (string assembly)
 
 		public static void SetWriteable (string source)
 		{
-			if (!File.Exists (source))
```
jonpryor: Replacing a `File.Exists()` with a try/catch block just seems "icky".

And if callers should *already* be checking `File.Exists()`... is this in the "contract"?  "Documented" anywhere?  *Verified*?  And if all that is true, why have the try/catch here at all?

jonathanpeppers: I'm fine to let it throw 👍. I checked all the callers with `Find All References` in the IDE.

I should add a `<summary/>` mentioning this.

jonpryor: It looks like `<ConvertResourcesCases/>` doesn't:

  * https://github.com/xamarin/xamarin-android/blob/720aa3c8e4e466339a0f08bdfba94cfbee5d88fb/src/Xamarin.Android.Build.Tasks/Tasks/ConvertResourcesCases.cs#L102

`Files.CopyIfChanged()` confuses me:

  * https://github.com/xamarin/xamarin-android/blob/720aa3c8e4e466339a0f08bdfba94cfbee5d88fb/src/Xamarin.Android.Build.Tasks/Utilities/Files.cs#L83-L101

  * `HasFileChanged()` can return true if *neither* `source` nor `destination` exists, meaning `File.Exists()` on either could be false.

Ditto for `Files.CopyIfBytesChanged()`, `Files.CopyIfStreamChanged()`, etc.

src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
```diff
@@ -363,32 +363,38 @@ public static bool IsForceRetainedAssembly (string assembly)
...
 
-			var dirInfo = new DirectoryInfo (directory);
-			if ((dirInfo.Attributes | FileAttributes.ReadOnly) == FileAttributes.ReadOnly)
-				dirInfo.Attributes &= ~FileAttributes.ReadOnly;
-
-			foreach (var dir in Directory.EnumerateDirectories (directory, "*", SearchOption.AllDirectories)) {
-				dirInfo = new DirectoryInfo (dir);
-				if ((dirInfo.Attributes | FileAttributes.ReadOnly) == FileAttributes.ReadOnly)
-					dirInfo.Attributes &= ~FileAttributes.ReadOnly;
+			foreach (FileSystemInfo child in root.EnumerateFileSystemInfos ("*", SearchOption.AllDirectories)) {
```
jonpryor: If the concern is GC use and reducing the garbage created, wouldn't we want to *avoid* the `DirectoryInfo` creation and use of `root.EnumerateFileSystemInfos()` here?

jonathanpeppers: What is the alternative? `DirectoryInfo` is the only way I know how to set directory attributes besides p/invoke. There is not a `Directory.SetAttributes` method.

jonpryor: Indeed.  Ugh.

src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
jonathanpeppers: This brings up a philosophical question.

Let's say we have:
```csharp
void Foo (string file)
{
    if (File.Exists (file)) {
        Bar (file);
        // Do something
    }
}

void Bar (string file)
{
    if (File.Exists (file)) {
        // Do something
    }
}
```

If `Foo` and `Bar` are both called a lot, let's say 1000's of times. What is the clean way to avoid the duplicate `File.Exists` call?

Do I need a `bool checkExists` parameter for `Bar`?

src/Xamarin.Android.Build.Tasks/Utilities/Files.cs
```diff
@@ -129,8 +165,10 @@ public static bool CopyIfStreamChanged (Stream stream, string destination)
 				if (!string.IsNullOrEmpty (directory))
 					Directory.CreateDirectory (directory);
 
-				MonoAndroidHelper.SetWriteable (destination);
-				File.Delete (destination);
+				if (File.Exists (destination)) {
+					SetWriteableUnchecked (destination);
+					File.Delete (destination);
+				}
```
jonathanpeppers: After reworking this PR, so things seem clean _and_ correct. I'm not seeing any gains anymore.

I will close this for now, but leave the branch here for later.

PR #4414, opened 2020-03-17
[Xamarin.Android.Build.Tasks] move DX into its own MSBuild targets file

jonathanpeppers:
Context: https://android-developers.googleblog.com/2020/02/the-path-to-dx-deprecation.html

As part of consolidation, cleanup, and the fact DX is about to be
unsupported:

* Move calls to DX and the `<CompileToDalvik/>` MSBuild task to a new
  `Xamarin.Android.DX.targets` file.

* This includes usage of ProGuard.

* Remove support for using ProGuard in combination with d8. r8 should
  just be used in this case.

Additionally, I cleaned up some logic in `Xamarin.Android.D8.targets`.
Resetting properties causes some unnecessary logging in MSBuild.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -304,6 +304,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<!-- NOTE: $(AndroidLinkTool) would be blank if code shrinking is not used at all -->
 	<AndroidLinkTool       Condition=" '$(AndroidLinkTool)' == '' And '$(AndroidEnableProguard)' == 'True' ">proguard</AndroidLinkTool>
 	<AndroidLinkTool       Condition=" '$(AndroidLinkTool)' == 'proguard' And '$(AndroidEnableDesugar)' == 'True' ">r8</AndroidLinkTool>
+	<AndroidLinkTool       Condition=" '$(AndroidLinkTool)' == 'proguard' And '$(AndroidDexTool)' == 'd8' ">r8</AndroidLinkTool>
 	<AndroidEnableProguard Condition=" '$(AndroidLinkTool)' != '' ">True</AndroidEnableProguard>
 	<AndroidEnableDesugar  Condition=" '$(AndroidEnableDesugar)' == '' And ('$(AndroidDexTool)' == 'd8' Or '$(AndroidLinkTool)' == 'r8') ">True</AndroidEnableDesugar>
 	<AndroidEnableDesugar  Condition=" '$(AndroidEnableDesugar)' == '' ">False</AndroidEnableDesugar>
@@ -401,6 +402,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 -->
 <!-- As we split up/refactor this file, put new imports here -->
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.D8.targets" />
+<Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.DX.targets" />
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt.targets" />
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt2.targets" />
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.SkipCases.projitems" />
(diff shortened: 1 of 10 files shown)


brendanzagaeski:
*Draft release notes*

Here's a candidate note for the ProGuard change:

```
### For D8 DEX compiler, R8 now used automatically instead of ProGuard

Xamarin.Android now selects the R8 code shrinker automatically for projects that
have the **Dex compiler** (`$(AndroidDexTool)`) set to **d8** and **Code
shrinker** (`$(AndroidLinkTool)`) set to **ProGuard**.
```

brendanzagaeski:
From the CI results, it looks like the `ProguardBOMError("d8","proguard")` test scenario might no longer be needed :smile:

jonathanpeppers:
@brendanzagaeski I think this test is still good, as it proves that using that combination still works even though I removed code like this:

https://github.com/xamarin/xamarin-android/blob/cf25620b45c7b8bdbd7712bcfed807e1bde03934/src/Xamarin.Android.Build.Tasks/Xamarin.Android.D8.targets#L49-L73

If we completely remove ProGuard support at some point, I would like to remove a lot of stuff like that, though. 👍

jonathanpeppers:
Actually the BOM one isn't valid I guess, looks like r8 doesn't care about BOMs in proguard files.

jonathanpeppers:
Test failures look similar to master or random:

```
InstallFailedException: Unexpected install output: cmd: Failure calling service package: Broken pipe (32)
...
The command ""/Library/Frameworks/Xamarin.Android.framework/Libraries/xbuild/Xamarin/Android/Darwin/aprofutil"  -s -v -f -p 9197 -o "custom.aprof"" exited with code 5.
```

The perf test should get fixed by: https://github.com/xamarin/xamarin-android/pull/4426

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -401,6 +402,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 -->
 <!-- As we split up/refactor this file, put new imports here -->
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.D8.targets" />
+<Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.DX.targets" />
```
dellis1972: is there a way to conditionally import this I wonder. So only if `AndroidLinkTool == dx`.

jonathanpeppers: I'll try it, I think I will just have to name the target the same in both files.

dellis1972: what we can do is add an extension point and then use that in the new Dx.targets.
it should work. the only call we have to Aapt/Aapt2 now I believe will be the CallTarget calls. 
Although I just thought of a way to work around that now I think...

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2377,8 +2377,7 @@ because xbuild doesn't support framework reference assemblies.
 
 <PropertyGroup>
 	<_CompileDexDependsOn>
-		_CompileToDalvikWithDx;
-		_CompileToDalvikWithD8;
+		_CompileToDalvik;
```
dellis1972: brilliant!!! :D

brendanzagaeski:
*Draft release notes adjustment*

I noticed a new detail about how narrow the user-facing visibility of the ProGuard change is, so I will plan to skip including any release notes for it.

*Background info*

I realized that `$(AndroidEnableDesugar)` has been enabled by default for D8 all along:

https://github.com/xamarin/xamarin-android/blob/988c8112d0592a4e9f7e89e5b129c3a07b78a8cf/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L311

Combined with https://github.com/xamarin/xamarin-android/commit/fbe4b44c7b40dcaacdecaed834c631b15490cc52, that means the only new scenario where R8 will be used after this PR is:

  * `$(AndroidEnableDesugar)` explicitly set to a non-empty string other than `True`.
  * `$(AndroidLinkTool)` = `ProGuard`
  * `$(AndroidDexTool)` = `d8`

Since having `$(AndroidEnableDesugar)` explicitly set to `False` (or another string) should be quite rare and since this only affects projects that are already using Java code shrinking (via ProGuard), I think in the end it'll be simplest to omit this from the release notes.

brendanzagaeski:
*Draft release notes un-adjustment*

Whoops. Disregard my previous comment. I overlooked the fact that the order of properties in the `Xamarin.Android.Common.targets` is important here. The line that sets `$(AndroidEnableDesugar)` to `True` by default comes *after* the line from https://github.com/xamarin/xamarin-android/commit/fbe4b44c7b40dcaacdecaed834c631b15490cc52 that set `$(AndroidLinkTool)` to `r8`, so the draft release note that was committed as part of this PR is accurate. I will include it in the final release notes.

PR #4417, opened 2020-03-18
[build] Do not generate and use sources.projitems

radekdoulik:
This file was only used for dependencies and wasn't imported until
second build. The dependencies already contain the source files.

It was causing problems in parallel configure builds as the cmake
processes were writting to the file simultaneously.

### DIFF
--- a/src/monodroid/monodroid.targets
+++ b/src/monodroid/monodroid.targets
@@ -4,7 +4,6 @@
   <Import Project="monodroid.props" />
   <Import Project="monodroid.projitems" />
   <Import Project="..\..\build-tools\scripts\JavaInteropDllConfigs.targets" />
-  <Import Project="sources.projitems" Condition="Exists ('sources.projitems')" />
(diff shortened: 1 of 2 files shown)


src/monodroid/monodroid.targets
```diff
@@ -102,7 +101,7 @@
 
   <Target Name="_BuildHostRuntimes"
       DependsOnTargets="_CreateJavaInteropDllConfigs"
-      Inputs="@(_MonodroidSource);@(_HostRuntime->'$(IntermediatePath)%(OutputDirectory)-Debug\CMakeCache.txt');@(_HostRuntime->'$(IntermediatePath)%(OutputDirectory)-Release\CMakeCache.txt');jni\*.cc;jni\*.h;jni\**\*.c;"
+      Inputs="@(_HostRuntime->'$(IntermediatePath)%(OutputDirectory)-Debug\CMakeCache.txt');@(_HostRuntime->'$(IntermediatePath)%(OutputDirectory)-Release\CMakeCache.txt');jni\*.cc;jni\*.h;jni\**\*.c;"
```
grendello: we need `*.hh` as well

radekdoulik: indeed

PR #4376, opened 2020-03-10
Bump to xamarin/Java.Interop/master@56c92c70

jonpryor:
***DO NOT MERGE***

Context: https://github.com/xamarin/java.interop/pull/597

Just testing to see if xamarin/java.interop#597 works...

PR #4419, opened 2020-03-18
[Mono.Android] Build latest stable version against netcoreapp3.1

pjcollins:
Adds `netcoreapp3.1` to the set of target frameworks that Mono.Android
will build against in preparation for .NET 5. Additionally, the project build
will effectively be a no-op when targeting `netcoreapp3.1` if
`$(AndroidFrameworkVersion)` is not equal to 
`$(AndroidLatestStableFrameworkVersion)`

### DIFF
--- a/src/Mono.Android/Mono.Android.csproj
+++ b/src/Mono.Android/Mono.Android.csproj
@@ -1,10 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project Sdk="Microsoft.NET.Sdk">
+<Project>
 
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
   <Import Project="..\..\Configuration.props" />
 
   <PropertyGroup>
-    <TargetFrameworks>monoandroid10;netstandard2.1</TargetFrameworks>
+    <TargetFrameworks>monoandroid10;netcoreapp3.1</TargetFrameworks>
     <RootNamespace>Android</RootNamespace>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\product.snk</AssemblyOriginatorKeyFile>
@@ -30,7 +31,7 @@
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
 
-   <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.1' ">
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.1' ">
     <OutputPath>$(XAInstallPrefix)xbuild-frameworks\Xamarin.Android.App\$(AndroidFrameworkVersion)\</OutputPath>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
@@ -78,9 +79,8 @@
     </Reference>
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.1' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.1' ">
     <ProjectReference Include="..\..\external\Java.Interop\src\Java.Interop\Java.Interop.csproj" />
-    <PackageReference Include="System.Drawing.Common" Version="4.7.0" />
   </ItemGroup>
 
   <ItemGroup>
@@ -340,5 +340,11 @@
     <ProjectReference Include="..\..\external\Java.Interop\tools\jcw-gen\jcw-gen.csproj" ReferenceOutputAssembly="false" SkipGetTargetFrameworkProperties="true" AdditionalProperties="TargetFramework=net472" />
     <ProjectReference Include="..\..\src\java-runtime\java-runtime.csproj" ReferenceOutputAssembly="false" />
   </ItemGroup>
+
+  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
+
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp3.1' And '$(AndroidFrameworkVersion)' != '$(AndroidLatestStableFrameworkVersion)' ">
+    <BuildDependsOn></BuildDependsOn>
+  </PropertyGroup>
 
 </Project>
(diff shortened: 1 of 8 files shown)


pjcollins:
New build is here, not sure what's up with the duplicate element error above 
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3569697&view=results

src/Mono.Android/Mono.Android.csproj
```diff
@@ -30,6 +30,11 @@
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
   </PropertyGroup>
 
+   <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.1' ">
```
jonpryor: Indentation here is "wonky".

Where should netcoreapp3.1 output go?

Rando thought that I'm trying in PR #4376: place netcoreapp3.1 output into `$(topdir)\bin\Debug-netcoreapp3.1`.

Better idea -- unimplemented because I don't know the details -- is for files to be included in the .NET 5 support package to follow the appropriate directory structure of the .NET 5 support package.  The problem is that I don't know what the directory structure should be...

pjcollins: For the targeting pack you can look at the rough destination paths here:
https://github.com/xamarin/xamarin-android/blob/net5-preview/src/Xamarin.Android.App/Xamarin.Android.App.Ref.csproj#L33

I like the idea of better mirroring this output, though there's a couple of questions/concerns. To do this we'll need to only build the latest stable version of `Mono.Android.dll` when targeting `netstandard2.1`, and not every supporting version. We also need ultimately need Mono.Android.dll in both targeting and runtime packs, so we'll need to account for that in the future as well. 

I'm not sure how much of this should be considered for this PR. I took a quick look at attempting to only build the latest `Mono.Android.dll` question and didn't find a great quick solution to that, but we may want to dedicate time to that aspect now at least so that we don't bloat our build times unnecessarily.

src/Mono.Android/Mono.Android.csproj
```diff
@@ -73,6 +78,11 @@
     </Reference>
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.1' ">
+    <ProjectReference Include="..\..\external\Java.Interop\src\Java.Interop\Java.Interop.csproj" />
+    <PackageReference Include="System.Drawing.Common" Version="4.7.0" />
```
pjcollins: I believe that this reference was only needed to ensure this attribute compiled successfully:

https://github.com/xamarin/xamarin-android/blob/master/src/Mono.Android/Properties/AssemblyInfo.cs.in#L29

This raises another question though, should we still be including these typeforwards, and the [`System.Drawing`](https://github.com/xamarin/xamarin-android/commit/c7a41e59f5391301008474f1469405aa5c7a5b62) sources when targeting `netstandard2.1` (or `net5` in the future)?

src/Mono.Android/Mono.Android.csproj
marek-safar: Isn't netstandard2.1 too restrictive? You won't be able to use any newer APIs.

pjcollins: I've updated the second target to `netcoreapp3.1`.

src/Mono.Android/Mono.Android.csproj
pjcollins: I was able to remove the `System.Drawing.Common` reference when targeting `netcoreapp3.1`, however I'm still not sure if we should be including the sources/typeforwards when building against this TargetFramework.

src/Mono.Android/Mono.Android.csproj
pjcollins: I fixed the indentation, and changed the output path to use the `TargetFramework` for now, now that only the latest stable version will be built when targeting `netcoreapp3.1`.

build-tools/xaprepare/xaprepare/ConfigAndData/Dependencies/Linux.DebianCommon.cs
```diff
@@ -13,6 +13,7 @@ abstract class LinuxDebianCommon : Linux
 			new DebianLinuxProgram ("cli-common-dev"),
 			new DebianLinuxProgram ("curl"),
 			new DebianLinuxProgram ("devscripts"),
+			new DebianLinuxProgram ("dotnet-sdk-3.1"),
```
grendello: There's one problem with this dependency - it's not in standard Debian/Ubuntu/etc repositories, so it will fail on an unprepared Debian machine. I think we should at least handle failure to detect or install it with a descriptive error message pointing to the dotnet website with instructions on how to add the repository.

src/Mono.Android/Mono.Android.csproj
jonpryor: > To do this we'll need to only build the latest stable version of `Mono.Android.dll` when targeting `netstandard2.1`, and not every supported version.

The intention is that net5-android will redistribute *only* API-R/API-30 `Mono.Android.dll`, and not one for every previous `$(TargetFrameworkVersion)` value.

build-tools/xaprepare/xaprepare/ConfigAndData/Dependencies/Linux.DebianCommon.cs
pjcollins: I've removed this, it looks like this .NET Core 3.1 installation issue is happening on other PRs as well, we can look to address this issue in a separate PR.

src/Mono.Android/Mono.Android.csproj
pjcollins: The latest changes ensure that we only build the `$(AndroidLatestStableFrameworkVersion )` version of `Mono.Android.dll` when targeting `netcoreapp3.1`.

jonpryor:
Commit message:

```
We expect the .NET 5 BCL to closely resemble the .NET Core 3.1 BCL.
Continue preparing for .NET 5 and build `Mono.Android.dll` against
the `netcoreapp3.1` target framework, to ensure that the APIs that we
need actually exist.

Additionally, the `Mono.Android.csproj` build will effectively be a
no-op when targeting `netcoreapp3.1` if `$(AndroidFrameworkVersion)`
is not equal to `$(AndroidLatestStableFrameworkVersion)`, as we
currently plan on only shipping *one* `Mono.Android.dll` assembly as
part of .NET 5, so there is no need to build `Mono.Android.dll`
against other `$(TargetFrameworkVersion)` values.

When building for `netcoreapp3.1`, `Mono.Android.dll` uses the
`Java.Interop.dll` which is built for .NET Standard, instead of the
one built against the `MonoAndroid` framework.  We may want to
consider updating `Java.Interop.csproj` to multitarget against
`netcoreapp3.1` as well.
```

PR #4422, opened 2020-03-19
[Xamarin.Android.Build.Tasks] APT0001 & XA2000 l10n support

brendanzagaeski:

Move the message strings for APT0001 & XA0002 into the `.resx` file so
that they are localizable.

Other changes:

Adjust the wording of the message for APT0001 so that it covers not only
the scenario where an `aapt` option is used with `aapt2` but also the
scenario where one of the provided `aapt2` command line arguments
contains a typo.

### DIFF
--- a/Documentation/guides/messages/apt0001.md
+++ b/Documentation/guides/messages/apt0001.md
@@ -8,7 +8,7 @@ ms.date: 03/31/2020
 ## Example messages
 
 ```
-error APT0001: Unknown option `--ignore-assets`. Please check `$(AndroidAapt2LinkExtraArgs)` and `$(AndroidAapt2CompileExtraArgs)` to see if they include any `aapt` command line arguments that are no longer valid for `aapt2` and ensure that all other arguments are valid for `aapt2`.
+error APT0001: Unknown option `--ignore-assets`. Please check `$(AndroidAapt2CompileExtraArgs)` and `$(AndroidAapt2LinkExtraArgs)` to see if they include any `aapt` command line arguments that are no longer valid for `aapt2` and ensure that all other arguments are valid for `aapt2`.
 ```
 
 ## Issue
--- a/Documentation/guides/messages/apt0001.md
+++ b/Documentation/guides/messages/apt0001.md
@@ -8,7 +8,7 @@ ms.date: 08/01/2018
 ## Example messages
 
 ```
-error APT0001: unknown option -- --no-crunch . This is the result of using `aapt` command line arguments with `aapt2`. The arguments are not compatible.
+error APT0001: Unknown option `--no-crunch`. Please check `$(AndroidAapt2CompileExtraArgs)` and `$(AndroidAapt2CompileExtraArgs)` to see if they include any `aapt` command line arguments that are no longer valid for `aapt2` and ensure that all other arguments are valid for `aapt2`.
 ```
 
 ## Issue
--- a/Documentation/guides/messages/apt0001.md
+++ b/Documentation/guides/messages/apt0001.md
@@ -1,14 +1,14 @@
 ---
 title: Xamarin.Android error APT0001
 description: APT0001 error code
-ms.date: 08/01/2018
+ms.date: 03/31/2020
 ---
 # Xamarin.Android error APT0001
 
 ## Example messages
 
 ```
-error APT0001: Unknown option `--no-crunch`. Please check `$(AndroidAapt2CompileExtraArgs)` and `$(AndroidAapt2CompileExtraArgs)` to see if they include any `aapt` command line arguments that are no longer valid for `aapt2` and ensure that all other arguments are valid for `aapt2`.
+error APT0001: Unknown option `--ignore-assets`. Please check `$(AndroidAapt2LinkExtraArgs)` and `$(AndroidAapt2CompileExtraArgs)` to see if they include any `aapt` command line arguments that are no longer valid for `aapt2` and ensure that all other arguments are valid for `aapt2`.
 ```
 
 ## Issue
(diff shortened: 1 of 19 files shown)


Documentation/guides/messages/apt0001.md
```diff
@@ -8,7 +8,7 @@ ms.date: 08/01/2018
 ## Example messages
 
 ```
-error APT0001: unknown option -- --no-crunch . This is the result of using `aapt` command line arguments with `aapt2`. The arguments are not compatible.
+error APT0001: Unknown option `--no-crunch`. Please check `$(AndroidAapt2CompileExtraArgs)` and `$(AndroidAapt2CompileExtraArgs)` to see if they include any `aapt` command line arguments that are no longer valid for `aapt2` and ensure that all other arguments are valid for `aapt2`.
```
jonathanpeppers: The only issue with this error message, is `--no-crunch` seems like it is an option for `aapt2`:
```
> .\bin\Debug\lib\xamarin.android\xbuild\Xamarin\Android\aapt2.exe compile --help
aapt2 compile [options] -o arg files...

Options:
 -o arg                                            Output path
 --dir arg                                         Directory to scan for resources
 --zip arg                                         Zip file containing the res directory to scan for resources
 --output-text-symbols arg                         Generates a text file containing the resource symbols in the
                                                   specified file
 --pseudo-localize                                 Generate resources for pseudo-locales (en-XA and ar-XB)
 --no-crunch                                       Disables PNG processing
 --legacy                                          Treat errors that used to be valid in AAPT as warnings
 --visibility arg                                  Sets the visibility of the compiled resources to the specified
                                                   level. Accepted levels: public, private, default
 -v                                                Enables verbose logging
 --trace-folder arg                                Generate systrace json trace fragment to specified folder.
 -h                                                Displays this help menu
```
Maybe they added this in a recent version of aapt2?

Should we just use `--banana` as an example or something?

dellis1972: :+1 for `--banana`

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -62,7 +62,7 @@ void CheckAppDomainUsageUnconditional (AssemblyDefinition assembly, Action<strin
 
 			foreach (var mr in assembly.MainModule.GetMemberReferences ()) {
 				if (mr.ToString ().Contains ("System.AppDomain System.AppDomain::CreateDomain")) {
-					warn ($"warning XA2000: Use of AppDomain.CreateDomain() detected in assembly: {assembly}. .NET 5 will only support a single AppDomain, so this API will no longer be available in Xamarin.Android once .NET 5 is released.");
+					warn (string.Format ("warning XA2000: " + Xamarin.Android.Tasks.Properties.Resources.XA2000, assembly));
```
brendanzagaeski: It occurred to me after some consideration that "warning" might need to be localized too. I'll have to do some quick investigation on that.

dellis1972: if you localise that is might perhaps break the MSBuild Error/Warning parser, which is what this depends on.

brendanzagaeski:

Change the example invalid AAPT2 option in the doc to `--ignore-assets`. I confirmed that that option is valid for `aapt` and invalid for both `aapt2 link` and `aapt2 compile`.

I considered using a nonsense option name, but I thought that looked a little funny in the context of the rest of the message that goes out of its way to mention "`aapt` command line arguments that are no longer valid for `aapt2`".

Add a "The following are literal names..." comment for XA2000 to be extra thorough.

Rebase on master.

**Intentionally *not* changed in the latest force-push:**

The following line is intentionally the same as before:

```
warn (string.Format ("warning XA2000: " + Xamarin.Android.Tasks.Properties.Resources.XA2000, assembly));
```
I confirmed that the `warning XA2000` prefix should *not* be localized. I ran MSBuild in a different locale and checked the `warning` prefix produced by `LogCodedWarning()`. The prefix is *not* localized even when the rest of the message is localized.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
brendanzagaeski: I confirmed that this is all set. The `warning XA2000` prefix should *not* be localized. I ran MSBuild in a different locale and checked the `warning` prefix produced by `LogCodedWarning()`. The prefix is *not* localized even when the rest of the message is localized.

brendanzagaeski:
**Test results:**

`Xamarin.Android.Build.Tests.BuildTest.CheckLibraryImportsUpgrade(True)` failed due to a build error:

```
"C:\A\vs2019xam000000-1\_work\1\s\bin\TestRelease\temp\CheckLibraryImportsUpgrade(True)\Library1\Library1.csproj" (Build target) (1) ->
(_ConvertResourcesCases target) -> 
  Resources\values\Strings.xml : warning XA1001: AndroidResgen: Warning while updating resource XML 'obj\Release\res\values\strings.xml': Access to the path 'C:\A\vs2019xam000000-1\_work\1\s\bin\TestRelease\temp\CheckLibraryImportsUpgrade(True)\Library1\obj\Release\res\values\strings.xml' is denied.

"C:\A\vs2019xam000000-1\_work\1\s\bin\TestRelease\temp\CheckLibraryImportsUpgrade(True)\Library1\Library1.csproj" (Build target) (1) ->
(_CompileResources target) ->
  Resources\values\Strings.xml : error APT2126: file not found. 
```

This is not specific to this PR. See for example <https://github.com/xamarin/xamarin-android/pull/4423> and <https://github.com/xamarin/xamarin-android/pull/4483>.

PR #4381, opened 2020-03-10
[Xamarin.Android.Build.Tests] Move debug.keystore into Project Tree

dellis1972:
When running unit tests we occasionally see a random
`ANDKT0000` build error. We suspect this happens when
two tests try to create the same `debug.keystore` in
the same locaiton.

To work around this issue all projects should use override
the `_ApkDebugKeyStore` property to make sure the file
is created in the project directory. This should stop
parallel tests creating the same file at the same time.

PR #4426, opened 2020-03-19
[tests] fix deletion of debug.keystore

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/4381

In 18957918, we made the `debug.keystore` reside in each test app
instead of system-wide. This was to improve random parallel build
failures.

Unfortunately, this file is getting cleaned in various incremental
build tests.

One such is a performance test where this target runs:

    _CreateAndroidDebugSigningKey (1.102s)

This appears to be happening on any incremental `SignAndroidPackage`
call in our tests.

I'm updating the file cleaning logic in Xamarin.ProjectTools to skip
over `debug.keystore` files.

After fixing this, I found we can lower the times by ~500ms for the
performance tests that install APKs. There is still a buffer:

    => Xamarin.Android.Build.Tests.PerformanceTest.Install_CSharp_Change
    [TESTLOG] Test InstallCSharpChange Starting
    ...
    Found Time Elapsed 00:00:04.7500000
    [TESTLOG] Test InstallCSharpChange Complete


jonathanpeppers:
Yay only one test failure! `BuildBasicApplicationAndAotProfileIt` which should be some separate issue.

PR #4443, opened 2020-03-23
[build] Don't build "real" debug DSOs

grendello:
We always build both release and debug binaries but the purpose of the
Debug build is not to be debuggable/unoptimized/etc in the native sense
but rather to contain code specific to and required by the managed Debug
builds.

So far, the Debug native builds have been built without any
optimizations thus making our Debug apps slower than they really need to
be. This commit changes the compiler optimization to `-O2` for the Debug
build, putting it on par with the Release builds, as far as optimization
is concerned.

This should make our Debug apps a bit faster.

The changes aren't dramatic, but there's a slight improvement:

= Before

Date (UTC): **3/25/2020 8:23:52 AM**
App configuration: **Debug**

Xamarin.Android
  - Version: **10.3.99-129**
  - Branch: **master**
  - Commit: **5099550c47d1ec39179a8cb7b05ca37fa3e61b74**

Device
  - Model: **Pixel 3 XL**
  - SDK version: **29**

= After

Date (UTC): **3/25/2020 9:48:59 AM**
App configuration: **Debug**

Xamarin.Android
  - Version: **10.3.99-130**
  - Branch: **always-optimize-native-code**
  - Commit: **00ac073bb855e2e88ed1b383745b3aa89b7a39cd**

Device
  - Model: **Pixel 3 XL**
  - SDK version: **29**

= Comparison

== Displayed time

| Before  | After   | Change    |
| ------- | ------- | --------- |
| 995.500 | 988.200 | -0.73% 🠅  |

== Native to managed transition time

| Before | After  | Change    |
| ------ | ------ | --------- |
| 50.187 | 50.164 | -0.04% 🠅  |

== Total native init time

| Before | After  | Change    |
| ------ | ------ | --------- |
| 89.513 | 87.692 | -2.03% 🠅  |

### DIFF
--- a/build-tools/cmake/xa_macros.cmake
+++ b/build-tools/cmake/xa_macros.cmake
@@ -49,9 +49,9 @@ macro(xa_common_prepare)
     fstrict-return
     Wa,-noexecstack
     fPIC
-	g
-	fomit-frame-pointer
-	O2
+    g
+    fomit-frame-pointer
+    O2
     )
 
   if(CMAKE_BUILD_TYPE STREQUAL Release)
--- a/build-tools/cmake/xa_macros.cmake
+++ b/build-tools/cmake/xa_macros.cmake
@@ -49,12 +49,12 @@ macro(xa_common_prepare)
     fstrict-return
     Wa,-noexecstack
     fPIC
+	g
+	fomit-frame-pointer
+	O2
     )
 
-  if(CMAKE_BUILD_TYPE STREQUAL Debug)
-    set(XA_COMPILER_FLAGS ${XA_COMPILER_FLAGS} ggdb3 fno-omit-frame-pointer O0)
-  else()
-    set(XA_COMPILER_FLAGS ${XA_COMPILER_FLAGS} g fomit-frame-pointer O2)
+  if(CMAKE_BUILD_TYPE STREQUAL Release)
     add_definitions("-DRELEASE")
   endif()
 
--- a/src/monodroid/CMakeLists.txt
+++ b/src/monodroid/CMakeLists.txt
@@ -229,8 +229,8 @@ if(CMAKE_BUILD_TYPE STREQUAL Debug)
   # will force a -O0 on us and our "debug" build is not really for debugging of our native code but
   # rather for "debug" builds of user apps - it has extra code but it has to be as fast as possible.
   if(ENABLE_NDK)
-	# This is specific to clang, enable only for Android builds
-	set(XA_COMPILER_FLAGS_DEBUG -fno-limit-debug-info)
+    # This is specific to clang, enable only for Android builds
+    set(XA_COMPILER_FLAGS_DEBUG -fno-limit-debug-info)
   endif()
   set(CMAKE_C_FLAGS_DEBUG "${XA_COMPILER_FLAGS_DEBUG}")
   set(CMAKE_CXX_FLAGS_DEBUG "${XA_COMPILER_FLAGS_DEBUG}")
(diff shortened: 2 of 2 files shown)


build-tools/cmake/xa_macros.cmake
```diff
@@ -51,10 +51,8 @@ macro(xa_common_prepare)
     fPIC
     )
 
-  if(CMAKE_BUILD_TYPE STREQUAL Debug)
-    set(XA_COMPILER_FLAGS ${XA_COMPILER_FLAGS} ggdb3 fno-omit-frame-pointer O0)
```
jonpryor: Alternatively, could/should we have `bin/Debug/...` be built with `-ggdb3 -0)` and `bin/Release/...` built with `-O2`?

Commercial build outputs come from `bin/Release/...`, so this arrangement would mean that local (default `bin/Debug` builds would be more easily debuggable, while what our customers get would be optimized.)

grendello: I wouldn't bother. We don't use any debuggers to debug our native code anyway. Let's keep things simple. And if we do need to use a debugger, it's easy to modify the cmake config locally to get unoptimized builds.

brendanzagaeski:
*Draft release note*

It sounds like this change might give a nice improvement in startup times for Debug configuration apps. When you get a chance, you can add a draft release note for it to `Documentation/release-notes/4443.md` based on the [app startup performance template](https://github.com/xamarin/xamarin-android/blob/master/Documentation/release-notes/README.md#template-for-build-time-or-app-performance-improvement). Feel free to pick and choose parts of the template to suit this PR.

Also feel free to hold off on adding the note for as long as needed, if for example the data on the estimated time improvement won't be available until a little later in the PR.

Thanks in advance!

grendello:
I found out that Android NDK was overriding our `-O2` and updated our code to disable the override. However, I don't think the results deserve mention in release notes. We are slightly faster in the native part, as you can see, but it doesn't translate much to the overall Displayed time speedup, so for the end users this information would just be inconsequential noise :)

brendanzagaeski:
Sounds sensible. I'll make a note to myself that this PR doesn't need any release notes. Thanks!

jonpryor:
Current draft release notes

---

`libmono-android.debug.so` and `libmono-android.release.so` differ
not in how easy they are to "debug" *themselves*, but rather in which
*app* configuration they should be included in:

  * `libmono-android.debug.so` is for "Debug" configuration apps, and
    supports loading assemblies from *outside* the `.apk`, e.g. for
    commercial Fast Deployment support.  It ***MUST NOT*** be used in
    "real" apps deployed to an App Store.

  * `libmono-android.release.so` is for "Release" configuration apps,
    and *only* loads assemblies from within the `.apk`.  (Even if the
    `.__override__` directory used by `libmono-android.debug.so`
    exists, *it will not be used*.)  This is the library that
    ***MUST*** be used in "real" apps deployed to an App Store.

Which is to say, the meaning of "debug" used in these file names does
not match the "conventional" use of "debug".

Unfortunately, we didn't inform `cmake` of this.  As such,
`libmono-android.debug.so` was also built with debug symbols and all
compiler optimizations *disabled* (`-O0`), which is unfortunate as
we want "Debug" configuration apps to start as quickly as possible,
to improve the inner dev loop.

Inform `cmake` and the NDK build system that we really want `-O2` to
be used to compile `libmono-android.debug.so`, which is the same set
of optimizations used for `libmono-android.release.so`.

This should make our Debug apps a bit faster.

The changes aren't dramatic, but there's a slight improvement:

Device:

  - Model: Pixel 3 XL
  - Native architecture: arm64-v8a
  - SDK version: 29

| Event                             | Before (ms) | After (ms) |     Δ     |
|-----------------------------------|------------:|-----------:|:---------:|
|                    Displayed time |     995.500 |    988.200 | -0.73% ⬆ |
| Native to managed transition time |      50.187 |     50.164 | -0.04% ⬆ |
|            Total native init time |      89.513 |     87.692 | -2.03% ⬆ |

PR #4450, opened 2020-03-24
[tests] rework how debug.keystore is used in tests

jonathanpeppers:
Context: https://build.azdo.io/3576318

This partially reverts:

* 18957918
* b6eba4ea

Some tests that deploy to an emulator with fast deployment were
failing with:

    [MESSAGE] jarsigner error: java.lang.RuntimeException:
    keystore load: /Users/runner/.local/share/Xamarin/Mono for Android/debug.keystore (No such file or directory)

This is because `_CreateAndroidDebugSigningKey` created a
`debug.keystore` local to the project, and the system-wide one doesn't
exist. `androidtools` apparently will *only* use the system-wide key,
and fails if it does not exist.

18957918 was a little heavy-handed, it was aiming to just fix
`CodeBehindBuildTests.csproj`. I reverted the global change to set
`$(_ApkDebugKeyStore)` in all tests, to just set it in
`CodeBehindBuildTests.csproj`.

We also don't need the fix in b6eba4ea anymore.


dellis1972:
I don't see the `CodeBehindBuildTests.csproj` changes?

jonathanpeppers:
@dellis1972 I left this line in there: https://github.com/xamarin/xamarin-android/blob/5099550c47d1ec39179a8cb7b05ca37fa3e61b74/tests/CodeBehind/BuildTests/CodeBehindBuildTests.csproj#L21

This is already on master right now ☝️

PR #4453, opened 2020-03-24
[Xamarin.Android.Build.Tasks] remove usage of $(TFV) in _ResolveSdks

jonathanpeppers:
As we migrate to .NET 5, we need a way for short-form MSBuild projects
to run a different set of MSBuild tasks/targets within the codebase.

At the top level `Xamarin.Android.Sdk.targets` will be imported
instead of `Xamarin.Android.CSharp.targets`. We can define
`$(UsingAndroidNETSdk)` as `true`, which is similar to
`$(UsingMicrosoftNETSdk)` already present for short-form MSBuild
projects.

Based on `$(UsingAndroidNETSdk)`:

* Existing long-form projects can import
  `Xamarin.Android.Legacy.targets`
* New short-form project can import other new files as needed.

The first two MSBuild tasks to change are:

* `<ValidateJavaVersion/>`
* `<ResolveAndroidTooling/>`

Both of these tasks rely on `$(TargetFrameworkVersion)`; however, we
can make versions of these tasks that don't need to use
`$(TargetFrameworkVersion)` at all.

`<ValidateJavaVersion/>` can just require Java 1.8 going forward.

`<ResolveAndroidTooling/>` will have any logic around
`$(TargetFrameworkVersion)` or `$(AndroidUseLatestPlatformSdk)` and
just targets the latest supported Android API level.

To make the two versions of these tasks:

* The core code of these tasks will remain.
* A new `Tasks\Legacy` folder and `Xamarin.Android.Tasks.Legacy`
  namespace will be created.
* Any "legacy" code can be moved into these classes that will inherit
  from the core tasks.
* Any MSBuild targets using these tasks will be included in
  `Xamarin.Android.Legacy.targets` via `$(UsingAndroidNETSdk)` as well
  as any `<UsingTask/>` declarations.

## Binding Projects ##

Since Binding projects have their own calls to
`<ValidateJavaVersion/>` and `<ResolveAndroidTooling/>`, now is the
time to consolidate this. We had some duplicated code here.

I moved the common targets into `Xamarin.Android.Tooling.targets`, all
projects will now use the `_ResolveSdks` target from this file.

I had to introduce a private property:

    <_AndroidIsBindingProject>True</_AndroidIsBindingProject>

We can use this to detect if the current project is a binding project
or not. This may evolve as we fully understand what needs to happen
for .NET 5.

### DIFF
--- a/Documentation/guides/DotNet5.md
+++ b/Documentation/guides/DotNet5.md
@@ -1,0 +1,32 @@
+# .NET 5 and Xamarin.Android
+
+_NOTE: this document is very likely to change, as the requirements for
+.NET 5 are better understood._
+
+A .NET 5 project for Xamarin.Android will look something like:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net5.0-android</TargetFramework>
+  </PropertyGroup>
+</Project>
+```
+
+See the [Target Framework Names in .NET 5][net5spec] spec for details.
+
+[net5spec]: https://github.com/dotnet/designs/blob/5e921a9dc8ecce33b3195dcdb6f10ef56ef8b9d7/accepted/2020/net5/net5.md
+
+## Changes to MSBuild tasks
+
+In .NET 5 the behavior of the following MSBuild tasks will change, but
+"legacy" projects will stay the same:
+
+* `<ValidateJavaVersion/>` - used to require Java 1.6, 1.7, or 1.8
+  based on the version of the Android Build Tools or
+  `$(TargetFrameworkVersion)`. .NET 5 will require Java 1.8.
+
+* `<ResolveAndroidTooling/>` - used to support the
+  `$(AndroidUseLatestPlatformSdk)` setting or multiple
+  `$(TargetFrameworkVersion)`. .NET 5 will always target the latest
+  Android APIs for `Mono.Android.dll`.
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -167,10 +167,10 @@ In this message, the phrase "should not be reached" means that this error messag
     <comment>The abbreviation "JDK" should not be translated.</comment>
   </data>
   <data name="XA0031" xml:space="preserve">
-    <value>Java SDK {0} or above is required when using $(TargetFrameworkVersion) {1}.</value>
-    <comment>The following are literal names and should not be translated: $(TargetFrameworkVersion)
+    <value>Java SDK {0} or above is required when using {1}.</value>
+    <comment>
 {0} - The Java SDK version number
-{1} - The target framework version number</comment>
+{1} - The feature being used: `$(TargetFrameworkVersion) v10.0` or `&lt;Project Sdk=&quot;Xamarin.Android.Sdk&quot;&gt;`</comment>
   </data>
   <data name="XA0032" xml:space="preserve">
     <value>Java SDK {0} or above is required when using Android SDK Build-Tools {1}.</value>
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Legacy/ValidateJavaVersion.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Legacy/ValidateJavaVersion.cs
@@ -1,0 +1,70 @@
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+using System.Text.RegularExpressions;
+
+namespace Xamarin.Android.Tasks.Legacy
+{
+	/// <summary>
+	/// ValidateJavaVersion's job is to shell out to java and javac to detect their version
+	/// </summary>
+	public class ValidateJavaVersion : Xamarin.Android.Tasks.ValidateJavaVersion
+	{
+		public string AndroidSdkBuildToolsVersion { get; set; }
+
+		public string TargetFrameworkVersion { get; set; }
+
+		protected override bool ValidateJava (string javaExe, Regex versionRegex)
+		{
+			Version requiredJavaForFrameworkVersion = GetJavaVersionForFramework ();
+			Version requiredJavaForBuildTools = GetJavaVersionForBuildTools ();
+			Version required = requiredJavaForFrameworkVersion > requiredJavaForBuildTools ? requiredJavaForFrameworkVersion : requiredJavaForBuildTools;
+
+			MinimumRequiredJdkVersion = required.ToString ();
+
+			try {
+				var versionNumber = GetVersionFromTool (javaExe, versionRegex);
+				if (versionNumber != null) {
+					Log.LogMessage (MessageImportance.Normal, $"Found Java SDK version {versionNumber}.");
+					if (versionNumber < requiredJavaForFrameworkVersion) {
+						Log.LogCodedError ("XA0031", Properties.Resources.XA0031, requiredJavaForFrameworkVersion, $"$(TargetFrameworkVersion) {TargetFrameworkVersion}");
+					}
+					if (versionNumber < requiredJavaForBuildTools) {
+						Log.LogCodedError ("XA0032", Properties.Resources.XA0032, requiredJavaForBuildTools, AndroidSdkBuildToolsVersion);
+					}
+					if (versionNumber > Version.Parse (LatestSupportedJavaVersion)) {
+						Log.LogCodedError ("XA0030", Properties.Resources.XA0030, versionNumber, LatestSupportedJavaVersion);
+					}
+				}
+			} catch (Exception ex) {
+				Log.LogWarningFromException (ex);
+				Log.LogCodedWarning ("XA0034", Properties.Resources.XA0034, required);
+				return false;
+			}
+
+			return !Log.HasLoggedErrors;
+		}
+
+		Version GetJavaVersionForFramework ()
+		{
+			var apiLevel = MonoAndroidHelper.SupportedVersions.GetApiLevelFromFrameworkVersion (TargetFrameworkVersion);
+			if (apiLevel >= 24)
+				return new Version (1, 8);
+			else if (apiLevel == 23)
+				return new Version (1, 7);
+			else
+				return new Version (1, 6);
+		}
+
+		Version GetJavaVersionForBuildTools ()
+		{
+			Version buildTools;
+			if (!Version.TryParse (AndroidSdkBuildToolsVersion, out buildTools)) {
+				return Version.Parse (LatestSupportedJavaVersion);
+			}
+			if (buildTools >= new Version (24, 0, 1))
+				return new Version (1, 8);
+			return Version.Parse (MinimumSupportedJavaVersion);
+		}
+	}
+}
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ValidateJavaVersion.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ValidateJavaVersion.cs
@@ -1,9 +1,8 @@
-﻿using Microsoft.Build.Framework;
-using Microsoft.Build.Utilities;
 using System;
 using System.IO;
 using System.Text;
 using System.Text.RegularExpressions;
+using Microsoft.Build.Framework;
 using Xamarin.Android.Tools;
 
 namespace Xamarin.Android.Tasks
@@ -15,18 +14,16 @@ namespace Xamarin.Android.Tasks
 	{
 		public override string TaskPrefix => "VJV";
 
-		public string TargetFrameworkVersion { get; set; }
-
-		public string AndroidSdkBuildToolsVersion { get; set; }
-
 		public string JavaSdkPath { get; set; }
 
 		public string JavaToolExe { get; set; }
 
 		public string JavacToolExe { get; set; }
 
+		[Required]
 		public string LatestSupportedJavaVersion { get; set; }
 
+		[Required]
 		public string MinimumSupportedJavaVersion { get; set; }
 
 		[Output]
@@ -37,7 +34,7 @@ namespace Xamarin.Android.Tasks
 
 		public override bool RunTask ()
 		{
-			if (!ValidateJava (TargetFrameworkVersion, AndroidSdkBuildToolsVersion))
+			if (!ValidateJava ())
 				return false;
 
 			Log.LogDebugMessage ($"{nameof (ValidateJavaVersion)} Outputs:");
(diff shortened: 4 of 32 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
```diff
@@ -0,0 +1,51 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <UsingTask TaskName="Xamarin.Android.Tasks.Legacy.ValidateJavaVersion" AssemblyFile="Xamarin.Android.Build.Tasks.dll" />
+  <UsingTask TaskName="Xamarin.Android.Tasks.Legacy.ResolveAndroidTooling" AssemblyFile="Xamarin.Android.Build.Tasks.dll" />
```
jonathanpeppers: @dellis1972 @pjcollins if you want to review the general idea ^^

src/Xamarin.Android.Sdk/targets/Xamarin.Android.Sdk.targets
```diff
@@ -7,11 +7,55 @@
...
   </PropertyGroup>
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
 
   <Import Project="$(MSBuildThisFileDirectory)..\tools\Xamarin.Android.DefaultOutputPaths.targets" Condition="'$(EnableDefaultOutputPaths)' == 'true'" />
   <Import Project="$(MSBuildThisFileDirectory)..\tools\Xamarin.Android.Common.targets" />
 
+  <UsingTask TaskName="Xamarin.Android.Tasks.ValidateJavaVersion" AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
+  <UsingTask TaskName="Xamarin.Android.Tasks.ResolveAndroidTooling" AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
+
```
jonathanpeppers: Should we put the additional `<UsingTask/>` and target in a new `.targets` file? Maybe `Xamarin.Android.Sdk.Common.targets`?

dellis1972: I would put the UsingTask in the target which is using them.

dellis1972:
I think we should move all the `ResolveAndroidTooling` stuff into a new target. This is because its currently duplicated between common and binding targets. We should look at unifying them. Not in this Pr but in the future. So we need them in on common target which can be imported.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -167,8 +167,8 @@ In this message, the phrase "should not be reached" means that this error messag
     <comment>The abbreviation "JDK" should not be translated.</comment>
   </data>
   <data name="XA0031" xml:space="preserve">
-    <value>Java SDK {0} or above is required when using $(TargetFrameworkVersion) {1}.</value>
-    <comment>The following are literal names and should not be translated: $(TargetFrameworkVersion)
+    <value>Java SDK {0} or above is required when using {1}.</value>
```
pjcollins: Should we change this to this to say `Java SDK {0} or above is required.`? I think continuing to print a target framework version could be confusing?

src/Xamarin.Android.Sdk/targets/Xamarin.Android.Sdk.targets
pjcollins: I think this makes sense for now, `Xamarin.Android.Sdk.targets` can be treated as our "Common" file, and we can create additional targets as necessarily. 

Depending on how far we want to split things up at this time however, we could consider putting these in a new `Xamarin.Android.Sdk.ValidateTooling.targets` file or something?

dellis1972: yes something like `Xamarin.Android.ValidateToolng.targets` or `Xamarin.Android.ResolveTools.targets`. That is was I was envisioned, so its all kinda self contained.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
jonathanpeppers: I was trying to use the exact same message in both old and new. Especially since it looks like we have it translated already...

And example of the message for legacy tasks in this PR will be:

    Java SDK 1.8.0 or above is required when using $(TargetFrameworkVersion) v10.0.

I moved the word `$(TargetFrameworkVersion)` to the legacy MSBuild task.

The new message will be:

    Java SDK 1.8.0 or above is required when using Xamarin.Android.Sdk.

`Xamarin.Android.Sdk` is kind of a placeholder. We should figure out what wording is ok for the new one, can we just put something in the `{1}` that will work? @brendanzagaeski what do you think?

src/Xamarin.Android.Sdk/targets/Xamarin.Android.Sdk.targets
jonathanpeppers: I put the targets in `Xamarin.Android.Tooling.targets`, which binding projects and regular projects share now.

It seems like all project types _build_ now, but let's see what breaks on CI.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
jonathanpeppers: Maybe it could be:
```
Java SDK 1.8.0 or above is required when using .NET 5.
```

pjcollins: If we want to go that route maybe we should do `... when using .NET 5 or newer.` or something like that, or we'll have to consider bumping this message regularly. I'm fine with `Xamarin.Android.Sdk` for now though.

jonathanpeppers: `... when using .NET 5 or newer` would just mean it needs to be translated.

brendanzagaeski: I think it'd be fine to add another string resource for the new tasks. It could use a property name like `XA0031_NET5`. That'll give flexibility to adjust the wording as the vocabulary around .NET 5 solidifies. For example, "when using .NET 5" might not be exactly true since a long-form project in a .NET 5 release of Xamarin.Android would still have the old behavior, if I'm understanding correctly?

An expansion on the original idea might be to include extra context about the term "Xamarin.Android.Sdk", like:

```
Java SDK 1.8.0 or above is required for projects that use `<Project Sdk="Xamarin.Android.Sdk">`.
```

The most solid related vocabulary term might be "[Sdk attribute](https://docs.microsoft.com/dotnet/core/tools/csproj#sdk-attribute)," but I have a hunch the example XML snippet might be more recognizable.

From what I understand, the individual strings are handled pretty quickly by the translation team, and I think they might even have tooling to reuse parts of translations that closely match existing strings. So I think it will be an OK trade-off to request a new translation for this special case.

src/Xamarin.Android.Build.Tasks/Tasks/ValidateJavaVersion.cs
```diff
@@ -57,36 +54,29 @@ public override bool RunTask ()
...
 				var versionNumber = GetVersionFromTool (javaExe, versionRegex);
 				if (versionNumber != null) {
 					Log.LogMessage (MessageImportance.Normal, $"Found Java SDK version {versionNumber}.");
-					if (versionNumber < requiredJavaForFrameworkVersion) {
-						Log.LogCodedError ("XA0031", Properties.Resources.XA0031, requiredJavaForFrameworkVersion, targetFrameworkVersion);
-					}
-					if (versionNumber < requiredJavaForBuildTools) {
-						Log.LogCodedError ("XA0032", Properties.Resources.XA0032, requiredJavaForBuildTools, buildToolsVersion);
+					if (versionNumber < required) {
+						Log.LogCodedError ("XA0032", Properties.Resources.XA0031, required, "Xamarin.Android.Sdk");
```
brendanzagaeski: It looks like the message code `"XA0032"` got carried over from copy-paste and is meant to be `"XA0031"` to match the Properties.Resources.XA0031 member.

src/Xamarin.Android.Sdk/targets/Xamarin.Android.Sdk.targets
```diff
@@ -7,11 +7,55 @@
...
+      <Output TaskParameter="ZipAlignPath"                PropertyName="ZipAlignToolPath"            Condition="'$(ZipAlignToolPath)' == ''" />
+      <Output TaskParameter="AndroidSequencePointsMode"   PropertyName="_SequencePointsMode"         Condition="'$(_SequencePointsMode)' == ''" />
+      <Output TaskParameter="LintToolPath"                PropertyName="LintToolPath"                Condition="'$(LintToolPath)' == ''" />
+      <Output TaskParameter="ApkSignerJar"                PropertyName="ApkSignerJar"                Condition="'$(ApkSignerJar)' == ''" />
+      <Output TaskParameter="AndroidUseApkSigner"         PropertyName="AndroidUseApkSigner"         Condition="'$(AndroidUseApkSigner)' == ''" />
+      <Output TaskParameter="AndroidUseAapt2"             PropertyName="_AndroidUseAapt2" />
+      <Output TaskParameter="Aapt2Version"                PropertyName="_Aapt2Version" />
+      <Output TaskParameter="Aapt2ToolPath"               PropertyName="Aapt2ToolPath"               Condition="'$(Aapt2ToolPath)' == ''" />
+    </ResolveAndroidTooling>
+  </Target>
```
jonathanpeppers: TODO: Move this to `Xamarin.Android.Sdk.Tooling.targets`

Documentation/guides/DotNet5.md
```diff
@@ -0,0 +1,30 @@
...
+
+_NOTE: this document is very likely to change, as the requirements for
+.NET 5 are better understood._
+
+A .NET 5 project for Xamarin.Android will look something like:
+
+```xml
+<Project Sdk="Xamarin.Android.Sdk/0.0.1">
+  <PropertyGroup>
+    <TargetFramework>netcoreapp5.0</TargetFramework>
```
jonpryor: As per the [**Target Framework Names in .NET 5** spec](https://github.com/terrajobst/designs/blob/25f4f0c7f710fb8205bb28147bb91b0471cb4840/accepted/2020/net5/net5.md), this will *probably* be `net5.0-android`, not `netcoreapp5.0`.

jonathanpeppers: I updated to match the spec, it also should have `<Project Sdk="Microsoft.NET.Sdk" />`.

Our current test projects have to import `Sdk="Xamarin.Android.Sdk/0.0.1"` right now, but we shouldn't document that.

pjcollins: I think the project SDK will also be `Microsoft.NET.Sdk`, and our SDK will be imported from there ultimately. Not sure if we should make this distinction now though.

src/Xamarin.Android.Build.Tasks/Tasks/Legacy/ValidateJavaVersion.cs
```diff
@@ -0,0 +1,77 @@
...
+			Version requiredJavaForFrameworkVersion = GetJavaVersionForFramework();
+			Version requiredJavaForBuildTools = GetJavaVersionForBuildTools();
+			Version required = requiredJavaForFrameworkVersion > requiredJavaForBuildTools ? requiredJavaForFrameworkVersion : requiredJavaForBuildTools;
+
+			MinimumRequiredJdkVersion = required.ToString();
+
+			try
+			{
+				var versionNumber = GetVersionFromTool(javaExe, versionRegex);
+				if (versionNumber != null)
```
jonpryor: Why is this file following a foreign coding convention?

For that matter, should this file be a `git cp` of `src/Xamarin.Android.Build.Tasks/Tasks/ValidateJavaVersion.cs` then with changes applied?

jonathanpeppers: Something weird happened, it's like this file got auto-formatted but not using our `.editorconfig`?

jonathanpeppers: I don't think there is a `git cp`: https://stackoverflow.com/questions/47401843/git-copy-file-as-opposed-to-git-mv

But I just fixed this file, I don't see others messed up.

jonathanpeppers:
The test failures look OK, F# should be fixed on master, and this one also happened on master:
```
"C:\A\vs2019xam000000-1\_work\1\s\bin\TestRelease\temp\BasicApplicationRepetitiveBuild\UnnamedProject.csproj" (Build;SignAndroidPackage target) (1) ->
(_ConvertResourcesCases target) -> 
  Resources\layout\Main.axml : warning XA1001: AndroidResgen: Warning while updating resource XML 'obj\Debug\res\layout\main.xml': Access to the path 'C:\A\vs2019xam000000-1\_work\1\s\bin\TestRelease\temp\BasicApplicationRepetitiveBuild\obj\Debug\res\layout\main.xml' is denied. [C:\A\vs2019xam000000-1\_work\1\s\bin\TestRelease\temp\BasicApplicationRepetitiveBuild\UnnamedProject.csproj]
```
I have a PR improving the stack trace, which will help figure out what is going on ^^

https://github.com/xamarin/xamarin-android/pull/4483/files#diff-c2cf55441f336d57dddc429b9728d4bbR30

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets
```diff
@@ -613,6 +550,8 @@ Copyright (C) 2012 Xamarin Inc. All rights reserved.
   </Target>
 
   <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.PCLSupport.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Tooling.targets" />
```
dellis1972: Does this need to be conditionally imported?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -407,6 +403,8 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt.targets" />
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt2.targets" />
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.SkipCases.projitems" />
+<Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Tooling.targets" />
```
dellis1972: same here 

```
Condition=" '$(UsingAndroidNETSdk)' == 'True' "
```

jonathanpeppers: This file is needed everywhere, should I take my own advice?

https://github.com/xamarin/xamarin-android/pull/4459#discussion_r400411074

jonathanpeppers:
@dellis1972 I added a header to each new `.targets` file explaining better where it is imported and why.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Tooling.targets
```diff
@@ -0,0 +1,72 @@
...
+  <Target Name="_SetLatestTargetFrameworkVersion" DependsOnTargets="$(_SetLatestTargetFrameworkVersionDependsOnTargets)">
+    <PropertyGroup>
+      <TargetFrameworkMoniker Condition=" '$(TargetFrameworkProfile)' == '' ">$(TargetFrameworkIdentifier),Version=$(TargetFrameworkVersion)</TargetFrameworkMoniker>
+      <TargetFrameworkMoniker Condition=" '$(TargetFrameworkProfile)' != '' ">$(TargetFrameworkIdentifier),Version=$(TargetFrameworkVersion),Profile=$(TargetFrameworkProfile)</TargetFrameworkMoniker>
+      <NuGetTargetMoniker>$(TargetFrameworkMoniker)</NuGetTargetMoniker>
+      <TargetFrameworkMonikerAssemblyAttributesPath>$(IntermediateOutputPath)$(TargetFrameworkMoniker).AssemblyAttributes$(DefaultLanguageSourceExtension)</TargetFrameworkMonikerAssemblyAttributesPath>
+    </PropertyGroup>
+    <ItemGroup>
+      <FileWrites Include="$(TargetFrameworkMonikerAssemblyAttributesPath)" />
+      <Reference Include="$(_JavaInteropReferences)" />
```
pjcollins: I think this can use a `Condition=" '$(UsingAndroidNETSdk)' != 'True' "` but we can address this case in a future PR that is dealing with ResolveAssemblies changes.

PR #4456, opened 2020-03-24
[Mono.Android.Export] Build against netcoreapp3.1

pjcollins:
We expect the .NET 5 BCL to closely resemble the .NET Core 3.1 BCL.
Continue preparing for .NET 5 and build `Mono.Android.Export.dll`
against the `netcoreapp3.1` target framework, to ensure that the APIs
that we need actually exist.


src/Mono.Android.Export/Mono.Android.Export.csproj
```diff
@@ -1,42 +1,30 @@
...
+    <NoStdLib>true</NoStdLib>
+    <ImplicitlyExpandDesignTimeFacades>false</ImplicitlyExpandDesignTimeFacades>
   </PropertyGroup>
-  <Import Project="..\..\Configuration.props" />
-  <PropertyGroup>
+
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'monoandroid10' ">
     <TargetFrameworkIdentifier>MonoAndroid</TargetFrameworkIdentifier>
-    <TargetFrameworkVersion>$(AndroidFrameworkVersion)</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v1.0</TargetFrameworkVersion>
```
pjcollins: @jonpryor I am missing some historical context here, do we need to be compiling against a specific MonoAndroid TargetFrameworkVersion here, or do you think these changes will be compatible?

src/Mono.Android.Export/CallbackCode.cs
```diff
@@ -511,9 +511,14 @@ static class DynamicCallbackFactory
 	{
 		static DynamicCallbackFactory ()
 		{
-			var ass = AppDomain.CurrentDomain.DefineDynamicAssembly (
+#if MONOANDROID1_0
+			var assembly = AppDomain.CurrentDomain.DefineDynamicAssembly (
 				new AssemblyName ("__callback_factory__"), AssemblyBuilderAccess.Run);
-			Module = ass.DefineDynamicModule ("__callback_factory__");
+#else
+			var assembly = AssemblyBuilder.DefineDynamicAssembly (
```
jonpryor: This method exists in the "normal" MonoAndroid profile.  We can skip the `#if MONOANDROID1_0` part and just use this directly.

src/Mono.Android.Export/Mono.CodeGeneration/CodeModule.cs
```diff
@@ -42,8 +42,10 @@ public CodeModule (string name)
 			myAsmName.Name = name;
 #if NET_2_1
 			AssemblyBuilder myAsmBuilder = myDomain.DefineDynamicAssembly (myAsmName, AssemblyBuilderAccess.Run);
-#else
+#elif MONOANDROID1_0
 			AssemblyBuilder myAsmBuilder = myDomain.DefineDynamicAssembly (myAsmName, AssemblyBuilderAccess.RunAndSave);
+#else
+			AssemblyBuilder myAsmBuilder = AssemblyBuilder.DefineDynamicAssembly (myAsmName, AssemblyBuilderAccess.Run);
```
jonpryor: Ditto here.

src/Mono.Android.Export/Mono.Android.Export.csproj
jonpryor: I would have thought that this *wouldn't* be compatible -- because I would have thought that we needed to use the "same" `$(TargetFrameworkVersion)` as `Mono.Android.dll` was built with so that it could be found.  Except there's a `@(ProjectReference)` to `Mono.Android.csproj`, which presumably means we *don't* need to use the same `$(TargetFrameworkVersion)` value!

*Furthermore* this change might solve the issues I was having on #4376!

pjcollins:
Now that (most if not all) of the regression test issues have been sorted, I'm going to manually run a full test build against this: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3583404&view=results

pjcollins:
Clearly there are still some regression test issues to sort out, but none of the regular PR checks seem to be reporting any issues out of the current set of failures happening elsewhere.

PR #4459, opened 2020-03-25
[build] Create Xamarin.Android.App packs

pjcollins:
Adds new `Xamarin.Android.App.Ref` and `Xamarin.Android.App.Runtime`
projects which will be used to create targeting and runtime packs for
the `Xamarin.Android.App` framework.

The `Xamarin.Android.App.Ref` NuGet is intended to contain only the
assets needed at compile time. Its structure currently looks like this:

```
Xamarin.Android.App.Ref
└─── data
│    └─── FrameworkList.xml
│    └─── AndroidApiInfo.xml
└─── ref
│    └─── netcoreapp5.0
│    │    └─── Java.Interop.dll
│    │    └─── Mono.Android.dll
│    │    └─── Mono.Android.Export.dll
```

The `Xamarin.Android.App.Runtime` NuGets contain the assets that
will be required at runtime. The structure of one of these packs currently
looks like this:

```
Xamarin.Android.App.runtime.android.21-arm64
└─── data
│    └─── RuntimeList.xml
└─── runtimes
│    └─── android.21-arm64
│    │    └─── lib
│    │    │    └─── netcoreapp5.0
│    │    │    │    └─── Java.Interop.dll
│    │    │    │    └─── Mono.Android.dll
│    │    │    │    └─── Mono.Export.dll
│    │    └─── native
│    │    │    └─── libmono-android.release.so
│    │    │    └─── libsqlite3_xamarin.so
│    │    │    └─── libxamarin-debug-app-helper.so
```

We will likely extend one of these packs to include both 
`mono.android.dex` and `mono.android.jar` alongside
`Mono.Android.dll` in a future PR.

The [Microsoft.DotNet.Build.Tasks.SharedFramework.Sdk][0] library
is used by these new Xamarin.Android.App projects to generate the 
`FrameworkList.xml` and `RuntimeList.xml` files required in the packages.

Versioning for existing SDK packs has been updated to use the shared
product versioning logic in `XAVersionInfo.targets`. In the future, we'll
likely want to use unique versioning logic for these three package types
(SDK, targeting, runtime).

The changes which unify versioning allow us to also generate and include
a `Xamarin.Android.Sdk.BundledVersions.props` file which will contain
information about the targeting and runtime packs that should be
associated with our SDK. This file declares a `KnownFrameworkReference`
that will allow our new framework to be resolved by the .NET Core build
system.

[0]: https://github.com/dotnet/arcade/tree/master/src/Microsoft.DotNet.Build.Tasks.SharedFramework.Sdk


src/Xamarin.Android.App/Xamarin.Android.App.Runtime.proj
```diff
@@ -0,0 +1,42 @@
+<Project Sdk="Microsoft.Build.NoTargets/1.0.88">
+
```
jonathanpeppers: I'm feeling as the number of MSBuild files start to grow...

@dellis1972 do you think we should start putting a short comment/description at the top of each new MSBuild file to explain what the file is?

So we would want just a short sentence explaining the difference between `App.Runtime.proj` and `App.Ref.proj`? `Xamarin.Android.Lite.proj`, too?

If I made a `Xamarin.Android.BuildOrder.targets` the name is almost self-explanatory, but would probably be helpful to say "this is where all `$(FooDependsOn)` properties go".

build-tools/automation/azure-pipelines.yaml
```diff
@@ -178,27 +178,29 @@ stages:
...
     - task: DotNetCoreCLI@2
-      displayName: pack Xamarin.Android.Sdk.Lite
+      displayName: pack sdk and ref nupkgs
       inputs:
         command: pack
-        packagesToPack: $(System.DefaultWorkingDirectory)/xamarin-android/build-tools/create-packs/Packager.csproj
+        packagesToPack: "\
+          $(System.DefaultWorkingDirectory)/xamarin-android/src/Xamarin.Android.App/Xamarin.Android.App.Ref.proj;
+          $(System.DefaultWorkingDirectory)/xamarin-android/src/Xamarin.Android.Sdk/Xamarin.Android.Sdk.proj;
+          $(System.DefaultWorkingDirectory)/xamarin-android/src/Xamarin.Android.Sdk/Xamarin.Android.Sdk.Lite.proj"
```
jonathanpeppers: Can this use the regular multi-line syntax?
```suggestion
        packagesToPack: |
          $(System.DefaultWorkingDirectory)/xamarin-android/src/Xamarin.Android.App/Xamarin.Android.App.Ref.proj;
          $(System.DefaultWorkingDirectory)/xamarin-android/src/Xamarin.Android.Sdk/Xamarin.Android.Sdk.proj;
          $(System.DefaultWorkingDirectory)/xamarin-android/src/Xamarin.Android.Sdk/Xamarin.Android.Sdk.Lite.proj
```
Yaml is weird, maybe it doesn't work.

pjcollins: `|` will keep new lines, and `>` will replace new lines with spaces. In this case the dotnet CLI yaml task expects `;` as the delimiter. This seemed like the only way to pass multiple projects on separate lines.

src/Xamarin.Android.App/Xamarin.Android.App.Runtime.proj
pjcollins: Alternatively in this case, we could use a single project/targets file that could create multiple NuGet packages depending on which target is called? These projects don't have any sources and are only here to try to simplify `dotnet pack` invocations.

jonathanpeppers: I think it's good to have separate files; it gives an opportunity to have a comment that explains what each `.nupkg` file is.

dellis1972: comments are good 👍

src/Xamarin.Android.App/Xamarin.Android.App.Runtime.proj
```diff
@@ -0,0 +1,51 @@
+<!--
```
jonpryor: What are the *semantics* of `src/Xamarin.Android.App`?  "Current" semantics is that `src/*` produce `.dll` files, `tools/*` produce `.exe` files, and `build-tools` is stuff used for the build.

This file doesn't "feel like" it's producing a `.dll` output.  Should `src/Xamarin.Android.App` instead be `build-tools/Xamarin.Android.App`?

pjcollins: Fair point, I've moved this into build-tools. I was wondering if I should move the SDK projects, but since that directory also contains new props/targets I think it can remain in `src`?

jonpryor: It doesn't need to be moved at this time.  (It *should* have gone into build-tools in the first place, but was overlooked.)

PR #4479, opened 2020-03-28
Bump to mono/mono/2020-02@beccf5ba

jonpryor:
Changes: https://github.com/mono/mono/compare/a7e29665d246d359a778bf4d692b4cb5fd2f3548...beccf5ba4386148534467bff03fabe92da6d9705

Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/999375/
Context: https://github.com/mono/mono/issues/18715
Context: https://github.com/mono/mono/issues/18917
Context: https://github.com/mono/mono/issues/19005

  * mono/mono@beccf5ba438: [2020-02] Use GArray instead of GList when getting custom attributes from an image (#19102)
  * mono/mono@5bea89d861f: [llvm] Disable running the llvm verifier by default, it was enabled by mistake, and it takes a long time. (#19332)
  * mono/mono@4b0424cd688: [2020- 02][debugger] Implementing step through multithreaded code. (#19343)
  * mono/mono@2444b38b3c5: [2020-02] [dim] Class overriding interface method that was already override by another interface (#19121)
  * mono/mono@93a2f4ea395: Fix suspend_policy that will be sent to debugger-libs, because on debugger-libs we want to resume the VM only when it's suspended, but we were sending the wrong suspend_policy in the case of EVENT_KIND_VM_START and mono is started with suspend=y. (#19330)
  * mono/mono@0565f6e2c0e: Do not access unloading domains in debugger (case 1013579) (#19322)
  * mono/mono@b7196c45e86: [jit] Initialize correct class for tls fields (#19300) (#19313)
  * mono/mono@88c1989895f: [2020-02] [merp] Add breadcrumb for StackHash (#19242)
  * mono/mono@d2be7389b11: [2020-02] [corlib] Suppress finalization of underlying console FileStreams (#19164)

### DIFF
--- a/.external
+++ b/.external
@@ -1,2 +1,2 @@
 xamarin/monodroid:master@34f5275d1db891e25af17d743b425cc318e6e7df
-mono/mono:2020-02@a7e29665d246d359a778bf4d692b4cb5fd2f3548
+mono/mono:2020-02@beccf5ba4386148534467bff03fabe92da6d9705


brendanzagaeski:
*Draft release notes*

Here are the notes I'm planning to include for this pull request:

```
### App startup performance

  * [Mono GitHub PR 19102](https://github.com/mono/mono/pull/19102):
    Use `GArray` instead of `GList` when getting custom attributes from an
    image.  This reduces the number of memory allocations during app startup for
    a standard Xamarin.Forms app, which should improve startup times by several
    milliseconds.

### Issues fixed

#### Application Mono Framework behavior on device and emulator

  * [Mono GitHub 18917](https://github.com/mono/mono/issues/18917): 
    Explicit implementations of interface properties were ignored if the
    interfaces provided default implementations.
```

(These draft notes for a mono/mono bump intentionally only cover changes where Xamarin.Android was mentioned in the issue description or in the commit.)

To suggest something different, feel free to reply in this PR or add a `Documentation/release-notes/4468.md` file with the new suggestion. Thanks!

PR #4489, opened 2020-03-31
[tests] Let Xamarin.Android-Tests.sln load in VSMac without errors

radekdoulik:
The VSMac was complaining loading the XF test subproject:

    Error while trying to load the project 'tests/Xamarin.Forms-Performance-Integration/Xamarin.Forms.Performance.Integration.csproj': Project Xamarin.Forms.Performance.Integration is being loaded as a Shared Assets project but has a different file extension. Please check the project type GUID in the solution file is correct.

The GUID used means: Visual C#: {FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}
https://docs.microsoft.com/en-us/visualstudio/extensibility/internals/adding-items-to-the-add-new-item-dialog-boxes?view=vs-2019


radekdoulik:
> You might link to:

> https://stackoverflow.com/questions/10802198/visual-studio-project-type-guids

> I can't find an MSDN doc, but that GUID seems like it means "C# project" instead of "Shared project".

Good point, added link to our docs.

PR #4490, opened 2020-03-31
[ci] Try to get Windows MSBuild tests running on Azure Hosted pool.

jpobst:


jpobst:
Although this now seems to work (🎉), it appears these agents are considerable slower (2x) than our current ones, so perhaps this isn't viable after all. (😭)

PR #4491, opened 2020-03-31
[build] /restore no longer needed on Windows

jonathanpeppers:
On macOS,  you can generally build with just:

    make prepare all

On Windows:

    msbuild Xamarin.Android.sln /t:Prepare
    msbuild Xamarin.Android.sln

In 295aff28, we introduced some changes that now require a change to
the second step:

    msbuild Xamarin.Android.sln /restore

I think we can add a call at the end of `/t:Prepare`:

    <MSBuild Projects="$(_TopDir)\Xamarin.Android.sln" Targets="Restore" />

This way you don't have to remember the `/restore` switch.

Lastly, I updated the Windows `Build and Smoke Test` phase on Azure
DevOps, so it will build on Windows without `/restore`.

### DIFF
--- a/build-tools/scripts/PrepareWindows.targets
+++ b/build-tools/scripts/PrepareWindows.targets
@@ -23,5 +23,6 @@
     />
     <Exec Command="$(_XAPrepareExe) $(_XAPrepareStandardArgs) -a" WorkingDirectory="$(_TopDir)" />
     <MSBuild Projects="$(_TopDir)\Xamarin.Android.BootstrapTasks.sln" Targets="Restore;Build" />
+    <MSBuild Projects="$(_TopDir)\Xamarin.Android.sln" Targets="Restore" />
   </Target>
 </Project>
--- a/Documentation/building/windows/instructions.md
+++ b/Documentation/building/windows/instructions.md
@@ -36,7 +36,7 @@ MSBuild version 15 or later is required.
 
  7. Build the project:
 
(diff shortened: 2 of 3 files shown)


build-tools/scripts/PrepareWindows.targets
```diff
@@ -9,7 +9,11 @@
   </PropertyGroup>
   <Import Project="$(_TopDir)\Configuration.props" />
   <Target Name="_BuildXAPrepare">
-    <MSBuild Projects="$(MSBuildThisFileDirectory)..\xaprepare\xaprepare.sln" Targets="Restore;Build" />
+    <MSBuild
+        Projects="$(MSBuildThisFileDirectory)..\xaprepare\xaprepare.sln"
+        Targets="Restore;Build"
+        RunEachTargetSeparately="True"
```
radekdoulik: What is the intention of using `RunEachTargetSeparately`? Why do we want to ignore failures in `Restore` target?

jonathanpeppers: It doesn't do what I thought it does, I'll be removing it when I bring it out of draft.

jonathanpeppers:
The single test failure seems unrelated:
```
NuGet.targets(124,5): error : Could not find file '/Users/runner/runners/2.165.2/work/1/s/packages/xamarin.forms/2.3.4.231/kqtnwvbj.ph4'.
```

PR #4493, opened 2020-03-31
[.NET 5] Start adding some project capabilities

pjcollins:
Context: https://github.com/xamarin/xamarin-android/pull/4383

Visual Studio uses CPS to load SDK-style projects. In CPS, most things
are keyed-off so-called [capabilities][0] which replace other methods that
were used previously like project type GUIDs.

While capabilities can be defined in the IDE, they make most sense to be
directly expressed in the targets themselves where CPS will pick them up
automatically.

After discussing with @kzu, simply using `Android` could be confusing
with native C++ Android project that export the same capability. By
combining `Managed + Mobile + Android` we can make a clear distinction
that we are dealing with a Xamarin.Android project.

[0]: https://github.com/microsoft/VSProjectSystem/blob/master/doc/overview/about_project_capabilities.md


pjcollins:
Pulling this out of the `net5-preview` branch as we're working to break it up and piecewise merge items back in.

PR #4496, opened 2020-04-01
Arch Linux build preparation support (xaprepare)

hadi77ir:
Implements basic support for `pacman` and some of AUR helpers.


build-tools/xaprepare/xaprepare/Application/Program.ArchLinux.cs
```diff
@@ -1,29 +1,139 @@
...
+				return Path.GetFullPath(fileName);
+
+			var values = Environment.GetEnvironmentVariable("PATH");
+			foreach (var path in values.Split(Path.PathSeparator))
+			{
+				var fullPath = Path.Combine(path, fileName);
+				if (File.Exists(fullPath))
+					return fullPath;
+			}
+			return null;
```
grendello, line 57: You don't need this code, you should instead use the [OS.Which](https://github.com/xamarin/xamarin-android/blob/master/build-tools/xaprepare/xaprepare/OperatingSystems/OS.cs#L453-L464) method via `Context.OS` (or `Context.Instance.OS`, as applicable).

build-tools/xaprepare/xaprepare/Application/Program.ArchLinux.cs
```diff
@@ -1,29 +1,139 @@
...
-	class ArchLinuxProgram : LinuxProgram
-	{
-		public ArchLinuxProgram (string packageName, string executableName = null)
-			: base (packageName, executableName)
-		{}
+    class ArchLinuxProgram : LinuxProgram
+    {
+        private static string aurHelper = null;
+		// TODO: check if binary name is right.
+        private static string[] aurHelpersToCheck = new string[] { "yay","aurman","pacaur","pakku","aura","pikaur","trizen","pbget","auracle","repofish","pb", "bb", "rua" };
```
grendello, line 12: Can you explain, in a comment here, what are these programs for?

grendello:
First of all, thanks for the PR :) In addition to the comments I made above for the code, I have to small requests, if you don't mind:

1. Could you review [this list of packages](https://github.com/xamarin/xamarin-android/blob/master/build-tools/xaprepare/xaprepare/ConfigAndData/Dependencies/Linux.Arch.cs#L8-L39) and see if they're enough to build XA on your Arch system?
2. Please reformat the changes so that they follow the [Mono coding guidelines](https://www.mono-project.com/community/contributing/coding-guidelines/#style-guidelines), in line with the rest of the `xaprepare` code

Thanks again! :)

build-tools/xaprepare/xaprepare/Application/Program.ArchLinux.cs
hadi77ir: Arch Linux has two sources for installing packages, first is through "pacman" and repositories containing prebuilt packages, the other is "Arch User Repository" (AUR) which contains PKGBUILD files for building packages from scratch and then installing them. To install Oracle's JDK, you should use its AUR package, as it is not in the official repositories. To ease usage of AUR, many "AUR Helper"s have been developed and their CLI arguments are merely the same as pacman's. I decided to use them instead of pacman when available for package installation. 6 hours ago I thought it would be a great thing, but now I think I should remove it and leave it to user to install those custom AUR packages before running "xaprepare". :joy: And instead, complete the work on recognizing those custom packages.

grendello: Thanks for the explanation :) Yes, I think your approach to let the users take care of installation of those helpers so that we don't force anyone to use what they don't need/like/want. It's, indeed, much better to do more work on our side and adapt to the user's environment.

PR #4498, opened 2020-04-01
[.NET 5] initial implementation of "dotnet publish"

jonathanpeppers:
There are currently two "verbs" we are aiming to get working in
Xamarin.Android:

    dotnet build
    dotnet publish

Currently in .NET Core (aka .NET 5), `dotnet publish` is where all the
work to produce a self-contained "app" happens:

  * The linker via the `<IlLink/>` MSBuild task
  * .NET Core's version of AOT, named "ReadyToRun"

https://docs.microsoft.com/en-us/dotnet/core/whats-new/dotnet-core-3-0#readytorun-images

This means Xamarin.Android would run the following during `dotnet
build`:

  * Run `aapt` to generate `Resource.designer.cs` and potentially emit
    build errors for issues in `@(AndroidResource)` files.
  * Compile C# code

Almost everything else happens during `donet publish`:

  * Generate java stubs, `AndroidManifest.xml`, etc. This must happen
    after the linker.
  * Compile java code via `javac`
  * Convert java code to `.dex` via d8/r8
  * Create an `.apk` and sign it

This is highly subject to change, however.

To get everything working:

  * I moved any `$(FooDependsOn)` properties that determine MSBuild
    target ordering for "legacy" projects to

  * The equivalent for .NET 5 will be in
    `Xamarin.Android.Sdk.BuildOrder.targets`.

  * I modified the private `$(_CorePublishTargets)` property to run
    Xamarin.Android-specific targets after `<IlLink/>`. This will
    change after this PR lands:

    https://github.com/dotnet/sdk/pull/11073

  * I moved the existing `_ResolveAssemblies` MSBuild target to

  * The .NET 5 version of `_ResolveAssemblies` should be able to make
    use of item groups provided by MSBuild. This file lives in
    `Xamarin.Android.Sdk.AssemblyResolution.targets`. It will
    eventually change when we have Mono runtime packs for .NET 5.

Future changes:

  * We will need to use different item groups in the
    `_ResolveAssemblies` target once a Mono runtime pack for .NET 5 is
    available.

  * `dotnet publish` still uses the current Mono linker. We will need
    to refactor and integrate the new linker documented at:

https://github.com/mono/linker/tree/ec2cbdb894e11065f42b1223f664d9688b749323/src/linker#adding-custom-steps-to-the-linker

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetCLI.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetCLI.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Text;
@@ -10,17 +11,25 @@ namespace Xamarin.ProjectTools
 		public string BuildLogFile { get; set; }
 		public string ProcessLogFile { get; set; }
 		public string Verbosity { get; set; } = "diag";
-		public string AndroidSdkPath {get; set; } = AndroidSdkResolver.GetAndroidSdkPath ();
-		public string AndroidNdkPath {get; set; } = AndroidSdkResolver.GetAndroidNdkPath ();
+		public string AndroidSdkPath { get; set; } = AndroidSdkResolver.GetAndroidSdkPath ();
+		public string AndroidNdkPath { get; set; } = AndroidSdkResolver.GetAndroidNdkPath ();
 
-		string Executable = "dotnet";
+		const string Executable = "dotnet";
+		readonly XASdkProject project;
+		readonly string projectOrSolution;
+
+		public DotNetCLI (XASdkProject project, string projectOrSolution)
+		{
+			this.project = project;
+			this.projectOrSolution = projectOrSolution;
+		}
 
 		/// <summary>
 		/// Runs the `dotnet` tool with the specified arguments.
 		/// </summary>
 		/// <param name="args">command arguments</param>
 		/// <returns>Whether or not the command succeeded.</returns>
-		protected bool Execute (params string[] args)
+		protected bool Execute (params string [] args)
 		{
 			if (string.IsNullOrEmpty (ProcessLogFile))
 				ProcessLogFile = Path.Combine (XABuildPaths.TestOutputDirectory, $"dotnet{DateTime.Now.ToString ("yyyyMMddHHmmssff")}-process.log");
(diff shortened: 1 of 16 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetCLI.cs
```diff
@@ -69,22 +91,27 @@ public bool Build (string projectOrSolution, string configuration, string target
...
+				$"/flp1:LogFile=\"{BuildLogFile}\";Encoding=UTF-8;Verbosity={Verbosity}",
 				$"/bl:\"{Path.Combine (testDir, "msbuild.binlog")}\""
 			};
-
-			bool succeeded = Execute (execArgs);
-			if (succeeded && Directory.Exists (testDir)) {
-				FileSystemUtils.SetDirectoryWriteable (testDir);
-				Directory.Delete (testDir, true);
+			if (restore) {
+				arguments.Add ("/restore");
```
pjcollins, line 103: I don't think we need a restore param here, as all `dotnet` commands should run a `restore` first by default if needed?

jonathanpeppers: I think I added this in confusion when something wasn't working 👍

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Tooling.targets
```diff
@@ -41,16 +41,27 @@ projects.
     </ItemGroup>
   </Target>
 
-  <Target Name="_ResolveSdks" DependsOnTargets="_GetReferenceAssemblyPaths">
+  <PropertyGroup Condition=" '$(UsingAndroidNETSdk)' == 'True' ">
+    <_ResolveSdksDependsOnTargets>ResolveFrameworkReferences</_ResolveSdksDependsOnTargets>
+    <TargetFrameworkDirectory>$(MSBuildThisFileDirectory)..\lib\MonoAndroid\v10.0\</TargetFrameworkDirectory>
```
pjcollins, line 46: Just as a note, this path will be changing shortly in an upcoming PR when these assemblies are removed from the SDK.

src/Xamarin.Android.Sdk/targets/Xamarin.Android.Sdk.DefaultItems.props
```diff
@@ -29,4 +29,13 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
```
pjcollins, line 30: We should remove this explicit reference ItemGroup now that the `FrameworkReference` below is in place.

Let's move this in to `Xamarin.Android.Sdk.Lite.props` (e.g. https://github.com/xamarin/xamarin-android/blob/net5-preview/src/Xamarin.Android.Sdk/targets/Xamarin.Android.Sdk.Lite.props#L17)

PR #4499, opened 2020-04-01
[Xamarin.Android.Build.Tasks] Remove TargetFrameworkDirectory usage

pjcollins:
The `TargetFrameworkDirectory` property is no longer set by the .NET
Core build system.  Removing seemingly dead references to this property
will help simplify the places where we need to override this value in
.NET 5.

I could not find any references of `_XAInboxNETStandardFolders` in any
relevant [GitHub sources][0], so I've removed that ItemGroup declaration
entirely.

`Javac` (and its parents) also don't seem to make use of this property
as far as I can tell, so those references have been removed.

[0]: https://github.com/search?q=_XAInboxNETStandardFolders&type=Code


pjcollins:
Full test run triggered here - https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3602059&view=results

pjcollins:
Test results in the pr and full build so far look to be issues we've seen elsewhere:

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3602049&view=ms.vss-test-web.build-test-results-tab

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3602059&view=ms.vss-test-web.build-test-results-tab

Issue #4036, opened 2019-12-13
[16.4] Keystore was tampered with or password was incorrect


xdr-synapse:
After upgrade to VS 16.4, I was unable to sign again my application for Google Play with my keystore. Tought it was a bug so I waited for for an update to see if it was fixed, but no (16.4.1)

I tried so sign my APK with jarsigner from Android JDK and it worked, so I confirm that I don't lost the password or alias.

### Steps to Reproduce

1. Keystore password and alias the same (generated by VS old version). Password have the next symbols: @ ^  ] $
1. Distribute project with AdHoc/Google Play
2. Get error

### Expected Behavior
Sign the APK
### Actual Behavior
Failed to load signer "signer 1"
java.io.IOException: Keystore was tampered with, or password was incorrect.
### Version Information
Microsoft Visual Studio Enterprise 2019
Version 16.4.1
VisualStudio.16.Release/16.4.1+29609.76
Microsoft .NET Framework
Version 4.8.03752

Installed Version: Enterprise

Azure App Service Tools v3.0.0   16.4.457.38025
Azure App Service Tools v3.0.0

C# Tools   3.4.1-beta4-19607-02+52d275c4f82f329a9732b078c7f7fa0e45cd3e84
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.2.0 (d16-2@8b56e20)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

IntelliCode Extension   1.0
IntelliCode Visual Studio Extension Detailed Info

Microsoft JVM Debugger   1.0
Provides support for connecting the Visual Studio debugger to JDWP compatible Java Virtual Machines

Microsoft MI-Based Debugger   1.0
Provides support for connecting Visual Studio to MI compatible debuggers

Mono Debugging for Visual Studio   16.5.23 (1b51e8c)
Support for debugging Mono processes with Visual Studio.

NuGet Package Manager   5.4.0
NuGet Package Manager in Visual Studio. For more information about NuGet, visit https://docs.nuget.org/

ProjectServicesPackage Extension   1.0
ProjectServicesPackage Visual Studio Extension Detailed Info

Visual Basic Tools   3.4.1-beta4-19607-02+52d275c4f82f329a9732b078c7f7fa0e45cd3e84
[...]

PR #4504, opened 2020-04-02
[Xamarin.Android.Build.Tests] Add string password keystore tests

dellis1972:
Context #4036
    
Add unit tests to make sure we can use a strong password with the keytool and signing tools.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/KeyToolTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/KeyToolTests.cs
@@ -6,6 +6,7 @@ using System.IO;
 using Xamarin.Android.Tasks;
 using Xamarin.Android.Tools;
 using Xamarin.ProjectTools;
+using System.Linq;
 
 namespace Xamarin.Android.Build.Tests
 {
@@ -15,6 +16,7 @@ namespace Xamarin.Android.Build.Tests
 	{
 		List<BuildErrorEventArgs> errors;
 		List<BuildWarningEventArgs> warnings;
+		List<BuildMessageEventArgs> messages;
 		MockBuildEngine engine;
 		string temp;
 		string keyToolPath;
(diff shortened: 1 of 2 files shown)


jonpryor:
@dellis1972: I don't understand the interaction between this PR and https://github.com/xamarin/xamarin-android/issues/4036#issuecomment-607830864 .

Specifically, the comment on Issue #4036 implies that the Android+Java toolchain doesn't like "certain characters" such as `@`.  Yet this PR uses a password that contains `@`, *it works*, and I don't see any relevant changes in the PR which would cause it to work with this PR but not before this PR.

What am I missing?

dellis1972:
@jonpryor It seems its not very consistent. See https://github.com/xamarin/xamarin-android/issues/4036#issuecomment-613424371. With certain characters I get

```
keytool -genkeypair -alias teststringkey -storepass "r@^]$test12" -keypass "r@^]$test12" -keystore foo.keystore -dname "CN=Android Debug,O=Android,C=US" -keyalg RSA -validity 10000 -storetype pkcs12
keytool error: java.lang.Exception: Key password must be at least 6 characters
```

BUT those same characters work fine in the unit test. I have a feeling keytool is a bit weird.

Anyway I updated the PR to just say we are adding a strong password unit test.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/KeyToolTests.cs
```diff
@@ -163,5 +165,47 @@ public void CreateDebugKeyStore ()
...
+				KeyPass = pass,
+				Command = "-list",
+				ToolPath = keyToolPath,
+			};
+
+			Assert.IsTrue (keyToolTask.Execute (), "Task should have succeeded.");
+			Assert.AreEqual (0, errors.Count, "Task should have no errors.");
+			Assert.AreEqual (0, warnings.Count, "Task should have no warnings.");
+			Assert.AreEqual (0, task.ExitCode, "ExitCode should have been 0");
+			Assert.IsTrue (messages.Select (x => x.Message).Contains ("Certificate fingerprint (SHA1):"), "Certificate SHA1 should have been printed.");
```
jonpryor, line 208: This assert is failing: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3640507&view=ms.vss-test-web.build-test-results-tab&runId=12594914&resultId=100127&paneView=debug

```
Certificate SHA1 should have been printed.
 Expected: True
 But was: False
```

dellis1972: Weird cos its in the output `Certificate fingerprint (SHA1):` ... I'll take a look

Issue #4486, opened 2020-03-31
App builds but fails to launch after VS update


bolk:
As suggested by @brendanzagaeski in https://github.com/xamarin/xamarin-android/issues/4470#issuecomment-606318200 I open a new issue with diagnostic MSBuild output.

### Steps to Reproduce

1. Installed latest VS2019 stable Update on Mac (Xamarin.Android: 10.2.0.100)
2. Multidex enabled, DX true
3. Project build Successfully but Crashes on Launch

### Expected Behavior

App to build and Launch

### Actual Behavior

App Builds successfully
App Crashes on Launch:

> am start -n "it.arcasgr.apparcafondi/crc64117ee1069bfbf50b.SplashActivity"
> Starting: Intent { cmp=it.arcasgr.apparcafondi/crc64117ee1069bfbf50b.SplashActivity launchParam=MultiScreenLaunchParams { mDisplayId=0 mFlags=0 } }

[AndroidRuntime] Caused by: java.lang.ClassNotFoundException: Didn't find class "crc64117ee1069bfbf50b.MainApplication" on path: DexPathList[[],nativeLibraryDirectories=[/data/app/it.arcasgr.apparcafondi-2/lib/arm64, /data/app/it.arcasgr.apparcafondi-2/base.apk!/lib/arm64-v8a, /data/app/it.arcasgr.apparcafondi-2/split_config.arm64_v8a.apk!/lib/arm64-v8a, /data/app/it.arcasgr.apparcafondi-2/split_config.en.apk!/lib/arm64-v8a, /data/app/it.arcasgr.apparcafondi-2/split_config.it.apk!/lib/arm64-v8a, /data/app/it.arcasgr.apparcafondi-2/split_config.xxhdpi.apk!/lib/arm64-v8a, /syst03-31 09:25:05.904 V/EnterpriseBillingPolicyInternal( 3785): EBPInternal Constructor called: true
[Zygote] v2
[libpersona] KNOX_SDCARD checking this for 10579
[libpersona] KNOX_SDCARD not a persona
[Zygote] accessInfo : 0
[SELinux] SELinux selinux_android_compute_policy_index : Policy Index[2],  Con:u:r:zygote:s0 SPD:SEPF_SECMOBILE_7.0_0006 RAM:SEPF_SECMOBILE_7.0_0009, [-1 -1 -9 -1 0 1]
[...]

Issue #4478, opened 2020-03-28
error ADB0010: Mono.AndroidTools.InstallFailedException: Failure [INSTALL_FAILED_INVALID_APK:


vravindranath:
### Steps to Reproduce

1. Updated the VS for mac Version 8.5 (build 3183)
2. Try to install the debug build in phone.
3. Fails with the following error.

### Actual Behavior Unable to install the app into device. The install fails with the following errors

```
error ADB0010: Mono.AndroidTools.InstallFailedException: Failure [INSTALL_FAILED_INVALID_APK: com.test.teaseer.==/base.apk code is missing]

error ADB0010:   at Mono.AndroidTools.Internal.AdbOutputParsing.CheckInstallSuccess (System.String output, System.String packageName) [0x00152] in /Users/runner/runners/2.165.0/work/1/s/xamarin-android/external/monodroid/tools/msbuild/external/androidtools/Mono.AndroidTools/Internal/AdbOutputParsing.cs:341 
 error ADB0010:   at Mono.AndroidTools.AndroidDevice+<>c__DisplayClass95_0.<InstallPackage>b__0 (System.Threading.Tasks.Task`1[TResult] t) [0x00016] in /Users/runner/runners/2.165.0/work/1/s/xamarin-android/external/monodroid/tools/msbuild/external/androidtools/Mono.AndroidTools/AndroidDevice.cs:753 
error ADB0010:   at System.Threading.Tasks.ContinuationTaskFromResultTask`1[TAntecedentResult].InnerInvoke () [0x00024] in <f9d1b832704f410aa8ec771f4fe80552>:0 
error ADB0010:   at System.Threading.Tasks.Task.Execute () [0x00000] in /Users/builder/jenkins/workspace/build-package-osx-mono/2019-10/external/bockbuild/builds/mono-x64/external/corert/src/System.Private.CoreLib/src/System/Threading/Tasks/Task.cs:2319 
 error ADB0010: --- End of stack trace from previous location where exception was thrown ---
error ADB0010: 
[...]

PR #4505, opened 2020-04-02
[Xamarin.Android.Build.Tasks] trailing / causes missing classes.dex

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/4478
Fixes: https://github.com/xamarin/xamarin-android/issues/4486

We with the release of VS for Mac 8.5 going stable have been getting
some reports of:

    error ADB0010:  Deployment failed
    Mono.AndroidTools.InstallFailedException: Failure [INSTALL_FAILED_INVALID_APK: Package couldn't be installed in /data/app/com.contoso.androidapp1-RTZZFPyLkRI7Bk7VDgGkDg==: Package /data/app/com.contoso.androidapp1-RTZZFPyLkRI7Bk7VDgGkDg==/base.apk code is missing]
        at Mono.AndroidTools.Internal.AdbOutputParsing.CheckInstallSuccess (System.String output, System.String packageName) [0x00159] in <65710797f19f43cc9d8f9e05353b9615>:0
        at Mono.AndroidTools.AndroidDevice+<>c__DisplayClass95_0.<InstallPackage>b__0 (System.Threading.Tasks.Task`1[TResult] t) [0x0001c] in <65710797f19f43cc9d8f9e05353b9615>:0
        at System.Threading.Tasks.ContinuationTaskFromResultTask`1[TAntecedentResult].InnerInvoke () [0x00024] in <f9d1b832704f410aa8ec771f4fe80552>:0
        at System.Threading.Tasks.Task.Execute () [0x00000] in <f9d1b832704f410aa8ec771f4fe80552>:0
        at AndroidDeviceExtensions.PushAndInstallPackage (Mono.AndroidTools.AndroidDevice device, System.String apkFile, System.String packageName, System.Boolean reinstall, Mono.AndroidTools.Adb.AdbProgressReporter notifyProgress, System.Threading.CancellationToken token) [0x001d3] in <81f66a65e9434daf9d06602cf6ac0bf9>:0
        at AndroidDeviceExtensions.PushAndInstallPackage (Mono.AndroidTools.AndroidDevice device, System.String apkFile, System.String packageName, System.Boolean reinstall, Mono.AndroidTools.Adb.AdbProgressReporter notifyProgress, System.Threading.CancellationToken token) [0x00402] in <81f66a65e9434daf9d06602cf6ac0bf9>:0
        at Xamarin.AndroidTools.AndroidDeploySession.InstallPackage () [0x003be] in <81f66a65e9434daf9d06602cf6ac0bf9>:0
        at Xamarin.AndroidTools.AndroidDeploySession.RunAsync (System.Threading.CancellationToken token) [0x003f8] in <81f66a65e9434daf9d06602cf6ac0bf9>:0
        at Xamarin.AndroidTools.AndroidDeploySession.RunLoggedAsync (System.Threading.CancellationToken token) [0x00077] in <81f66a65e9434daf9d06602cf6ac0bf9>:0  [/Users/macuser/Projects/AndroidApp1/AndroidApp1/AndroidApp1.csproj]

The error means the `.apk` is missing `classes.dex`.

It seems to affect apps where:

  * `AndroidDexTool=dx`, this would need to be set explicitly or it
    would default to `d8`.
  * The issue appears to occur on macOS-only.

Brendan was able to reproduce the issue:

https://github.com/xamarin/xamarin-android/issues/4478#issuecomment-607632589

Using the zip file, there is an `obj` directory sitting next to the
`.sln` file:

  * `obj`
  * `AndroidApp1`
  * `AndroidApp1/AndroidApp1.csproj`
  * `AndroidApp1/obj`
  * `AndroidApp1.sln`

Note that the actual `$(BaseIntermediateOutputPath)` would be
`AndroidApp1/obj`. Brendan just moved an `obj` directory next to the
`.sln` file while trying to reproduce the issue.

However, it appears the presence of this phantom `obj` triggers the
issue! You can rename it to `obj2` to *solve* the issue. It is a
directory with sub-directories, but no files.

If you look at the build log, the main difference is:

    CompileToDalvik
        Parameters
    -       ClassesOutputDirectory = obj/Release/android/bin/classes/
    +       ClassesOutputDirectory = obj/Release/android/bin/classes

The presence of the trailing `/` breaks this code:

    cmd.AppendSwitchIfNotNull ("--output ", Path.GetDirectoryName (ClassesOutputDirectory));

`classes.dex` is output into `obj/Release/android/bin/classes` instead
of `obj/Release/android/bin` and doesn't make it into the final
`.apk`.

So tracking back to what creates the input value of
`ClassesOutputDirectory`, it is the MSBuild property
`$(_AndroidIntermediateJavaClassDirectory)`, it is merely:

    <_AndroidIntermediateJavaClassDirectory>$(IntermediateOutputPath)android\bin\classes\</_AndroidIntermediateJavaClassDirectory>

But then comparing broken vs working logs:

    Broken:
        _InitialBaseIntermediateOutputPath = obj/
        BaseIntermediateOutputPath = obj/
        IntermediateOutputPath = obj/Release/
        _AndroidIntermediateJavaClassDirectory = obj/Release/android/bin/classes/
    Working:
        _InitialBaseIntermediateOutputPath = obj\
        BaseIntermediateOutputPath = obj\
        IntermediateOutputPath = obj\Release\
        _AndroidIntermediateJavaClassDirectory = obj\Release\android\bin\classes\

It appears the existence of this random `obj` directory influences
MSBuild's path combining behavior?

The trailing `\` works, but `/` does not, due to `TrimEnd`:

    <CompileToDalvik
        ...
        ClassesOutputDirectory="$(_AndroidIntermediateJavaClassDirectory.TrimEnd('\'))"

The breakage was introduced in 185c5297, which added a trailing `\` to
the directory passed to `<CompileToDalvik/>`. We are unsure what is
causing the strange MSBuild behavior.

However, after reviewing `<CompileToDalvik/>`, I think we should just
remove the `Path.GetDirectoryName` call. We should just pass in the
directory name we want to use from MSBuild targets.

I could add a test for this scenario by setting a global property:

    _AndroidIntermediateJavaClassDirectory=obj/Debug/android/bin/classes/

Setting a global property could test the trailing `/` character
regardless of MSBuild behavior.

I also cleaned things up a bit by create a new
`$(_AndroidIntermediateDexOutputDirectory)` property.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
@@ -677,5 +677,23 @@ namespace App1
 				Assert.IsTrue (xab.LastBuildOutput.ContainsText (expected), $"Build should be using {expected}");
 			}
 		}
+
+		[Test]
+		public void MissingDexFile ()
+		{
+			//NOTE: The trailing / was breaking <CompileToDalvik /> in 16.5
+			var parameters = new [] { "_AndroidIntermediateJavaClassDirectory=obj/Debug/android/bin/classes/" };
+			var proj = new XamarinAndroidApplicationProject {
+				DexTool = "dx"
+			};
+			using (var b = CreateApkBuilder ()) {
+				Assert.IsTrue (b.Build (proj, parameters: parameters), "Build should have succeeded.");
+				var apk = Path.Combine (Root, b.ProjectDirectory,
+					proj.IntermediateOutputPath, "android", "bin", "UnnamedProject.UnnamedProject.apk");
+				using (var zip = ZipHelper.OpenZip (apk)) {
+					Assert.IsTrue (zip.ContainsEntry ("classes.dex"), "Apk should contain classes.dex");
+				}
+			}
+		}
 	}
 }
--- a/Documentation/release-notes/4505.md
+++ b/Documentation/release-notes/4505.md
@@ -1,0 +1,8 @@
+#### Application and library build and deployment
+
+  * [GitHub 4478](https://github.com/xamarin/xamarin-android/issues/4478),
+    [GitHub 4486](https://github.com/xamarin/xamarin-android/issues/4486):
+    On macOS, starting in Xamarin.Android 10.2, *\[INSTALL_FAILED_INVALID_APK:
+    ...  base.apk code is missing\]* and *Module 'base' has no dex files*
+    prevented deploying successfully for certain projects configured to use the
+    DX DEX compiler.
--- a/src/Xamarin.Android.Build.Tasks/Tasks/CompileToDalvik.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/CompileToDalvik.cs
@@ -114,7 +114,7 @@ namespace Xamarin.Android.Tasks
 					cmd.AppendSwitchIfNotNull ("--main-dex-list=", MultiDexMainDexListFile);
 				}
 			}
-			cmd.AppendSwitchIfNotNull ("--output ", Path.GetDirectoryName (ClassesOutputDirectory));
+			cmd.AppendSwitchIfNotNull ("--output ", ClassesOutputDirectory);
 
 			using (var sw = new StreamWriter (path: inputListFile, append: false,
 					encoding: MonoAndroidHelper.UTF8withoutBOM)) {
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -1024,6 +1024,7 @@ because xbuild doesn't support framework reference assemblies.
 	<MonoAndroidCodeBehindDir>$(IntermediateOutputPath)generated</MonoAndroidCodeBehindDir>
 	<MonoAndroidIntermediateResourceCache>$(IntermediateOutputPath)resourcecache</MonoAndroidIntermediateResourceCache>
 	<_AndroidIntermediateJavaSourceDirectory>$(IntermediateOutputPath)android\src\</_AndroidIntermediateJavaSourceDirectory>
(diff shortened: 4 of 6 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
```diff
@@ -677,5 +677,23 @@ public void CheckTheCorrectRuntimeAssemblyIsUsedFromNuget ()
 				Assert.IsTrue (xab.LastBuildOutput.ContainsText (expected), $"Build should be using {expected}");
 			}
 		}
+
+		[Test]
+		public void MissingDexFile ()
+		{
+			//NOTE: The trailing / breaks <CompileToDalvik />
```
brendanzagaeski, line 684: Not important really, but should this be in the past tense like:

```
//NOTE: The trailing / broke <CompileToDalvik /> in d16-5
```

The dependence on `Trim ('\')` got into d16-5 sneakily via the relatively harmless-looking <https://github.com/xamarin/xamarin-android/commit/185c5297a296163eb9a64cc15488e8e829e26c89>.

brendanzagaeski:
In case it might be a helpful extra reassurance, it looks like `GetDirectoryName` came in with: https://github.com/xamarin/monodroid/commit/30c22c7f0affc11fc9bdc96bca8716d452869893#diff-974b9024f51680c37819ca34a137b935R108

Based on that, it looks like it was added because originally the value passed to the `ClassesOutputDirectory` parameter on `<CompileToDalvik>` was set to match exactly with the value passed to `ClassesOutputDirectory` on the `<Javac>` task, and the `<CompileToDalvik>` task was once upon a time modifying that *directory* name to be a *file* name by appending `.dex`.

But then in 30c22c7f, the `<CompileToDalvik>` task started needing to adjust that directory name to be the name of the *parent* directory.

Another complicating factor back at that time was that the values passed to the `ClassesOutputDirectory` parameters were not set to include the conventional trailing slash used in MSBuild for properties containing directory names. That led to a potentially slightly funny technique of using `Path.GetDirectoryName()` instead of `Directory.GetParent()` to get the parent directory.

jonathanpeppers:
Somehow this check showed up twice:
image
image

This appears to be green, but the Github checks are messed up.

Issue #4457, opened 2020-03-25
Issue with R8 Compilation 


rkshnair:
In my app I am using Zebra [EMDK](https://www.nuget.org/packages/Symbol.XamarinEMDK/) library for integrating Barcode Scanner.
In order to Creating a common Application to run on Zebra and Non-Zebra devices I am adding this line of code in Android Manifest 
`<uses-library android:name="com.symbol.emdk" android:required="false" />` I have implemented this as per Zebra [documentation](https://techdocs.zebra.com/emdk-for-android/4-0/guide/programming_practices/)

The issue I am facing is, when I enable R8, D8 compilation it is not building the application and I am getting the below error

**R8: Error: Compilation can't be completed because some library classes are missing.**  

**Android Build configuration of the app is**

![image](https://user-images.githubusercontent.com/3647522/77518711-871d4b00-6e7e-11ea-8b7e-f97d9a9f419f.png)

**Error List**
![image](https://user-images.githubusercontent.com/3647522/77517528-93080d80-6e7c-11ea-875c-9df9105e08c2.png)

I am attaching the source code of the same for your reference in below link.

[Code](https://1drv.ms/u/s!AmsqbFKyP-CPggB48BOwtIikEETI?e=NDnyQt)

dellis1972:
I did a bit of investigation.
```
Task "DetermineJavaLibrariesToCompile" (TaskId:489)
  Task Parameter:
      MonoPlatformJarPaths=
          /Users/dean/Documents/Sandbox/Xamarin/xamarin-android/bin/Debug/lib/xamarin.android/xbuild-frameworks/MonoAndroid/v10.0/mono.android.jar
          /Users/dean/Documents/Sandbox/Xamarin/xamarin-android/bin/Debug/lib/xamarin.android/xbuild/Xamarin/Android/java_runtime_fastdev.jar (TaskId:489)
  Task Parameter:
      JavaLibraries=
          /Users/dean/.nuget/packages/xamarin.androidx.annotation/1.1.0/buildTransitive/monoandroid90/../../jar/androidx.annotation.annotation.jar
                  AndroidXSkipAndroidXMigration=true
          /Users/dean/.nuget/packages/xamarin.androidx.collection/1.1.0/buildTransitive/monoandroid90/../../jar/androidx.collection.collection.jar
                  AndroidXSkipAndroidXMigration=true
[...]

PR #4509, opened 2020-04-02
[Xamarin.Android.Build.Tasks] -ignorewarnings option to make r8 less strict

jonathanpeppers:
Context: https://www.guardsquare.com/products/proguard/manual/usage
Fixes: https://feedback.devdiv.io/959121
Fixes: https://feedback.devdiv.io/960245
Fixes: https://github.com/xamarin/xamarin-android/issues/4457

A common problem Xamarin.Android developers are running into is `r8`
failing with:

    Warning: Missing class: org.apache.http.client.methods.HttpEntityEnclosingRequestBase
    Error: Compilation can't be completed because some library classes are missing.

The simplest way to hit this is:

  * Use Google Play Services
  * Use multi-dex

In this case, we are only using `r8` for multi-dex support and *not*
code shrinking. Developers aren't setting proguard rules at all in
this scenario.

The correct fix is to do this in `AndroidManifest.xml`:

      <uses-library android:name="org.apache.http.legacy" android:required="false" />
    </application>

This will automatically include `--lib` to
`android-sdk\platforms\android-29\optional\org.apache.http.legacy.jar`
in the call to `r8`.

This is just completely different behavior than you got using `dx`.
Even though the fix is *correct*, it's a problem developers don't want
to run into. Most of the time everything will work fine without doing
this.

This error message can surface in other forms beyond usage of
`org.apache.http.legacy`. One example: A NuGet package has a
dependency on Google Play Services, but doesn't actually include the
`<dependency/>` in its `.nuspec` file. The fix is to add a
`<PackageReference/>` in the project yourself.

What is really weird about this, is `r8` only fails on *some*
warnings, not all.

`r8` is completely fine to emit warnings like these and continue on:

    Warning: The rule `-keep public class * extends java.lang.annotation.Annotation {
    *;
    }` uses extends but actually matches implements.
    Warning in obj\Debug\lp\44\jl\classes.jar:com/google/android/gms/internal/zzam.class:
    Type `org.apache.http.impl.cookie.DateUtils` was not found, it is required for default or static interface methods desugaring of `long com.google.android.gms.internal.zzam.zzf(java.lang.String)`
    Warning in obj\Debug\lp\44\jl\classes.jar:com/google/android/gms/internal/zzas.class:
    Type `android.net.http.AndroidHttpClient` was not found, it is required for default or static interface methods desugaring of `com.google.android.gms.internal.zzs com.google.android.gms.internal.zzas.zza(android.content.Context, com.google.android.gms.internal.zzan)`
    Warning: Resource 'META-INF/MANIFEST.MF' already exists.

I found a configuration option for proguard rules that appears to do
what we want:

    -ignorewarnings

    Specifies to print any warnings about unresolved references and other
    important problems, but to continue processing in any case. Ignoring
    warnings can be dangerous. For instance, if the unresolved classes or
    class members are indeed required for processing, the processed code
    will not function properly. Only use this option if you know what
    you're doing!

Even though there is a scary message, this seems to do exactly what we
need:

  1. r8 continues with its job if there are warnings.
  2. We still actually *emit* the warnings.

The name `-ignorewarnings` is somewhat confusing as it does not
suppress them at all. It just prevents an overall failure.

I added a new `$(AndroidR8IgnoreWarnings)` (`True` by default) MSBuild
property that will automatically include the `-ignorewarnings` option
in the ProGuard configuration file Xamarin.Android generates.

I updated a test for this scenario.

### DIFF
--- a/Documentation/guides/BuildProcess.md
+++ b/Documentation/guides/BuildProcess.md
@@ -1168,6 +1168,15 @@ resources.
   command-line options to pass to the **aapt** command when
   processing Android assets and resources.
 
+- **AndroidR8IgnoreWarnings** &ndash; Automatically specifies
+  the `-ignorewarnings` proguard rule for `r8`. This allows `r8`
+  to continue with dex compilation even if certain warnings are
+  encountered. Defaults to `True`, but can be set to `False` to
+  enforce more strict behavior. See the [ProGuard manual](https://www.guardsquare.com/products/proguard/manual/usage)
+  for details.
+
+  Added in Xamarin.Android 10.4.
+
 - **AndroidResgenFile** &ndash; Specifies the name of the Resource
   file to generate. The default template sets this to
   `Resource.designer.cs`.
--- a/Documentation/release-notes/4509.md
+++ b/Documentation/release-notes/4509.md
@@ -1,0 +1,42 @@
+### R8 uses `-ignorewarnings` to suppress common build errors
+
+Developers could hit a common problem in applications using Google
+Play Services:
+
+    Warning: Missing class: org.apache.http.client.methods.HttpEntityEnclosingRequestBase
+    Error: Compilation can't be completed because some library classes are missing.
+
+The correct fix is to adjust `AndroidManifest.xml` in the project:
+
+```xml
+<application ...>
+  <uses-library android:name="org.apache.http.legacy" android:required="false" />
+</application>
+```
+
+This issue could be encountered if:
+
+* `AndroidDexTool=d8` (this is default in Xamarin.Android 10.2 and higher)
+* `AndroidEnableMultiDex=True` or `AndroidLinkTool=r8`
+
+When using `AndroidDexTool=dx`, the same project would not encounter a
+build error--just warnings. However, `r8` is more strict and converts
+the `library classes are missing` warning into a failure.
+
+Xamarin.Android now applies a setting that will make `r8` more lenient
+to keep things closer to the existing `dx` behavior. By automatically
+passing `-ignorewarnings`, `r8` will not fail the build, but still
+emit warnings.
+
+To disable this behavior, a new `$(AndroidR8IgnoreWarnings)` MSBuild
+property can be set to `False` in your `.csproj` file:
+
+```xml
+<PropertyGroup>
+  <AndroidR8IgnoreWarnings>False</AndroidR8IgnoreWarnings>
+</PropertyGroup>
+```
+
+See the [ProGuard manual][proguard] for details about ProGuard rule syntax.
+
+[proguard]: https://www.guardsquare.com/products/proguard/manual/usage
--- a/src/Xamarin.Android.Build.Tasks/Tasks/R8.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/R8.cs
@@ -27,6 +27,7 @@ namespace Xamarin.Android.Tasks
 
 		// proguard-like configuration settings
 		public bool EnableShrinking { get; set; } = true;
+		public bool IgnoreWarnings { get; set; }
 		public string AcwMapFile { get; set; }
 		public string ProguardGeneratedReferenceConfiguration { get; set; }
 		public string ProguardGeneratedApplicationConfiguration { get; set; }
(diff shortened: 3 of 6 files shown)


brendanzagaeski:
I was thinking maybe it could be appropriate to limit the automatic `-ignorewarnings` to when `$(AndroidLinkTool)` is blank, that is, when R8 is being used only for multidex and not code shrinking? The reasoning I have in mind is that the Symbol.XamarinEMDK scenario from https://github.com/xamarin/xamarin-android/issues/4457 does also fail when using ProGuard code shrinking. It's the multidex scenario where the behavior changes in a trickier way when using D8.

jonathanpeppers:
@jonpryor or @dellis1972 do you have an opinion about putting `-ignorewarnings` always?

> @brendanzagaeski: maybe it could be appropriate to limit the automatic -ignorewarnings to when $(AndroidLinkTool) is blank, that is, when R8 is being used only for multidex and not code shrinking?

For multi-dex it makes sense, because developers don't have a way to pass in `proguard.cfg` files. If you are using `AndroidLinkTool=r8`, then you can configure `proguard.cfg` files as needed.

jonpryor:
> do you have an opinion about putting `-ignorewarnings` always?

My concern was largely about *existing* projects and existing customers encountering "new" build errors.

I think it's equally reasonable to set the property in *new* projects so that `-ignorewarnings` isn't used there.

Beyond that, I'm not entirely sure I understand the question.

brendanzagaeski:
> I think there have been cases where users moving from ProGuard to r8 have encountered new build errors with existing projects?
> 
> I think `-ignorewarnings` will allow their projects to build as before, and they will still get build warnings.

Yeah, I'm not sure what the balance is of projects that will have had ProGuard enabled and will get new errors from R8 vs. the number of projects that will have had multidex enabled with *no* code shrinking and will get new errors from R8 just because of multidex.

It's not a huge issue either way because as you noted the build *warnings* will still appear for R8 with `-ignorewarnings`, just not the *errors*. The little concern I had in mind was:

1. A project is currently using ProGuard and has made the build clean for ProGuard by fixing build *errors*.
2. The project moves to R8.
3. The project adds some new code that changes the Java dependencies.
4. The project *misses* the warnings from R8 because now there isn't an error.
5. The app aborts at run time instead.

jonathanpeppers:
> 5. The app aborts at run time instead.

I think this is basically what you got with dx + proguard ^^ You get runtime errors and have to read the build log to figure out what is wrong.

If I change this test to use dx+proguard instead of multi-dex:
https://github.com/xamarin/xamarin-android/blob/5397f99461ae79fa8593665ae6b3d64e2a3ff61a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L330-L333

I get warnings and it builds. But if I switch to r8, it will error unless we supply `-ignorewarnings`.

After `-ignorewarnings`, you get the same amount of warnings, but the build succeeds.

I think we should go with `-ignorewarnings` for all usage of r8, and I just need to add some release notes here?

brendanzagaeski:
> If I change this test to use dx+proguard instead of multi-dex:
> https://github.com/xamarin/xamarin-android/blob/5397f99461ae79fa8593665ae6b3d64e2a3ff61a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L330-L333
> 
> I get warnings and it builds. But if I switch to r8, it will error unless we supply `-ignorewarnings`.

It looks like there's an interesting little confounding factor. As currently set up, that test skips the `<Proguard/>` task when it builds the project:

```
Task "Proguard" skipped, due to false condition; ( '$(AndroidLinkTool)' == 'proguard' And '$(_ProguardProjectConfiguration)' != '' ) was evaluated as ( 'proguard' == 'proguard' And '' != '' ).
```

Adding `proj.IsRelease = true;` makes the test fail on the `<Proguard/>` task:

```
error MSB6006: "java.exe" exited with code 1.
Done executing task "Proguard" -- FAILED. (TaskId:202)
```

dellis1972:
The problem is there is no obvious way for us to tell the user how to fix the `Compilation can't be completed because some library classes are missing.`. Which is bad UX I think. I'm ok with turning this on by default, and letting the user decide if they want to turn it off.

PR #4519, opened 2020-04-03
[build] Fix Xamarin.Android.Sdk.Lite

pjcollins:
Recent updates to `Xamarin.Android.Sdk` have broken the `Lite` version.
While `Xamarin.Android.Sdk.Lite` is likely something that will be short
lived, we should try to keep it functioning. A new test has been added
to use this "Lite" SDK.

Removes `Xamarin.Android.TargetFrameworkInference.targets` from both
SDKs, and moves the pieces of this file that the Lite SDK depends on to
`Xamarin.Android.Sdk.Lite.targets`.

Conditionally imports `Xamarin.Android.Sdk.BundledVersions.props`, as
the Lite SDK won't generate or care about this file.

Further extends the way to opt out of default items by introducing a
`DisableImplicitAndroidFrameworkReference` property.  This was done in a
similar way to our Android file globbing conditions.  Users can opt out
by using the predefined .NET Core SDK properties, or by using Android
SDK specific properties.

Improves the SDK version info included in `Xamarin.Android.Build.Tests`
so that additional local commits will be accounted for.  The previously used
version files would not be re-generated during an incremental build of
`Xamarin.Android.Build.Tests`.

Finally, `$(MSBuildExtensionsPath)` can only be used in an `Import`
declaration on macOS, or the multiple search paths won't be respected.
Rather than overriding `$(LanguageTargets)` we can let the .NET SDK
import its own language targets, and then import our targets around
that.


jonpryor:
`BuildWithLiteSdk` [failed](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3615463&view=ms.vss-test-web.build-test-results-tab&runId=12364974&resultId=100186&paneView=attachments):

```
  /Users/runner/hostedtoolcache/dotnet/sdk/5.0.100-preview.1.20155.7/Sdks/Microsoft.NET.Sdk/Sdk/Sdk.targets(37,3): error MSB4019: The imported project "/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/xbuild/Xamarin/Android/Xamarin.Android.CSharp.targets" was not found. Confirm that the expression in the Import declaration "/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/xbuild/Xamarin/Android/Xamarin.Android.CSharp.targets" is correct, and that the file exists on disk. [/Users/runner/runners/2.165.2/work/1/s/bin/TestRelease/temp/BuildWithLiteSdk/UnnamedProject.csproj]
```

That path looks suspect; on my machine, it's `/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.CSharp.targets`, though I suspect that's not the desired/intended path either,.

PR #4522, opened 2020-04-03
[Xamarin.Android.Build.Tasks] Add XA1011 error for D8 with ProGuard

brendanzagaeski:
Context: 5563a79cbb6931818b06ddfb0c83d6957e3df480
Somewhat related: https://github.com/xamarin/xamarin-android/pull/4509#issuecomment-608640547

Rather than automatically changing `$(AndroidLinkTool)` from `proguard`
to `r8` when `$(AndroidDexTool)`=`d8`, provide an error that instructs
project authors to update the property themselves.

The motivation for this approach is that we have seen some cases where
changing from ProGuard to R8 can change behavior.  Forcing project
authors to change `$(AndroidLinkTool)` themselves will help let them
know that they should pay attention for any potential changes to Java
code shrinking behavior after the change.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -303,7 +303,6 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<!-- NOTE: $(AndroidLinkTool) would be blank if code shrinking is not used at all -->
 	<AndroidLinkTool       Condition=" '$(AndroidLinkTool)' == '' And '$(AndroidEnableProguard)' == 'True' ">proguard</AndroidLinkTool>
 	<AndroidLinkTool       Condition=" '$(AndroidLinkTool)' == 'proguard' And '$(AndroidEnableDesugar)' == 'True' ">r8</AndroidLinkTool>
-	<AndroidLinkTool       Condition=" '$(AndroidLinkTool)' == 'proguard' And '$(AndroidDexTool)' == 'd8' ">r8</AndroidLinkTool>
 	<AndroidEnableProguard Condition=" '$(AndroidLinkTool)' != '' ">True</AndroidEnableProguard>
 	<AndroidEnableDesugar  Condition=" '$(AndroidEnableDesugar)' == '' And ('$(AndroidDexTool)' == 'd8' Or '$(AndroidLinkTool)' == 'r8') ">True</AndroidEnableDesugar>
 	<AndroidEnableDesugar  Condition=" '$(AndroidEnableDesugar)' == '' ">False</AndroidEnableDesugar>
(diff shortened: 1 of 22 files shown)


brendanzagaeski:
I'm partly putting this up just to get thoughts on whether it looks like a good option. I'll be OK if we close this PR without merging if it seems better in the end to keep the automatic reassignment from `proguard` to `r8`.

It seemed the automatic conversions in d16-5 caused some trickiness that might have been a little less tricky if project authors had been forced to make the changes themselves to decide which new combination of settings they wanted to use.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -562,6 +561,10 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<CreateProperty Value="None" Condition="'$(AndroidLinkMode)' == 'SdkOnly' And '$(AndroidUseSharedRuntime)' == 'true'">
 		<Output TaskParameter="Value" PropertyName="AndroidLinkMode" />
 	</CreateProperty>
+	<Error Code="XA1011"
```
jonpryor, line 564: The problem with this approach, as I understand it, is that it isn't localizable.  This should probably be in a Task.

brendanzagaeski: Sounds good. I wasn't sure whether to use this PR to start a new pattern of mandating logging from within tasks for warnings and errors for localizability, but it does seem reasonable to use this PR to start that, so I'll adjust it!

brendanzagaeski: I found out that dotnet/sdk has a mechanism for this where it defines custom warning and error tasks that take resource names and format arguments instead of plain text:

  * [`NETSdkWarning.cs`](https://github.com/dotnet/sdk/blob/56a9d827c8408f98ab27caceae2c8e1936281401/src/Tasks/Common/NETSdkWarning.cs)
  * [`NETSdkError.cs`](https://github.com/dotnet/sdk/blob/56a9d827c8408f98ab27caceae2c8e1936281401/src/Tasks/Common/NETSdkError.cs)
  * [`MessageBase.cs`](https://github.com/dotnet/sdk/blob/56a9d827c8408f98ab27caceae2c8e1936281401/src/Tasks/Common/MessageBase.cs)

Example use: [`Microsoft.NET.RuntimeIdentifierInference.targets`](https://github.com/dotnet/sdk/blob/56a9d827c8408f98ab27caceae2c8e1936281401/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.RuntimeIdentifierInference.targets#L129-L134)

I'll tentatively plan to do a basic imitation of that approach for now. I'm guessing xamarin-android will be able to switch from that imitation to the upstream dotnet/sdk mechanism itself once the .NET 5 transition is complete.

jonpryor:
PR #4523 has been merged.  This PR should be rebased to use `<AndroidError/>` & `<AndroidWarning/>` now.

brendanzagaeski:

Rebase on master and switch to the new `<AndroidError/>` task to make the new error localizable.

Update the `MultiDexAndCodeShrinker`, `BuildProguardEnabledProject`, and `BuildApplicationWithSpacesInPath` tests to account for the new behavior, and remove the separate `XA1011` test because that scenario is now covered by both `BuildProguardEnabledProject` and `BuildApplicationWithSpacesInPath`.

brendanzagaeski:
There's one test failure for `AndroidUpdateResourcesTest.RepetitiveBuild`. It appears to be one of the unrelated issues that has been appearing intermittently for the CI tests:

```
```

brendanzagaeski:

Rebase on master to fix merge conflicts, and squash the commits along the way.

Correct a little mistake in `Localization.md` about what to pass in to the `ResourceName` parameter of the `<AndroidError/>` task.

I verified that the changed lines in the patch for this new commit are identical to the previous patch, apart from the new correction to `Localization.md`.

PR #4523, opened 2020-04-04
[Xamarin.Android.Build.Tasks] Add localizable message tasks

brendanzagaeski:
The built-in `<Error/>` and `<Warning/>` tasks from MSBuild do not
provide a way to look up string resources, so they don't allow logging
localizable warnings and errors from MSBuild targets.

Add custom `<AndroidError/>` and `<AndroidWarning/>` tasks to handle
this scenario.

This approach is inspired by the [`<NETSdkError/>`][0] and
[`<NETSdkWarning/>`][1] tasks in dotnet/sdk.

[0]: https://github.com/dotnet/sdk/blob/56a9d827c8408f98ab27caceae2c8e1936281401/src/Tasks/Common/NETSdkError.cs
[1]: https://github.com/dotnet/sdk/blob/56a9d827c8408f98ab27caceae2c8e1936281401/src/Tasks/Common/NETSdkWarning.cs

### DIFF
--- a/Documentation/workflow/Localization.md
+++ b/Documentation/workflow/Localization.md
@@ -22,6 +22,14 @@ so when adding a new message, follow these steps:
     Log.LogCodedError ("XA0000", Properties.Resources.XA0000);
     ```
 
+    Or, to log a message directly from an MSBuild target, pass the name of the
+    generated property to the `ResourceName` parameter of the `<AndroidError/>`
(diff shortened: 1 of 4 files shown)


brendanzagaeski:
The reason this topic of message localizability in MSBuild targets came up right now is so that I can use it in <https://github.com/xamarin/xamarin-android/pull/4522>.

PR #4526, opened 2020-04-06
[Xamarin.Android.Build.Tasks] Remove unused ResourceMerger class

brendanzagaeski:
[Xamarin.Android.Build.Tasks] Remove unused ResourceMerger class

Context: 480c006b80f910aec781da860e86e037b4867b9b
Context: ff84922e60808f12b60776a94323444c38dbf088

In the past, the `ResourceMerger` class was used by the
`<MergeResources/>` task and the `<CopyAndConvertResources/>` task.
Both of those tasks have now been removed, so the `ResourceMerger` class
can now also be removed.

Similarly, `$(_AndroidResourcesCacheFile)` is no longer used and can be
removed.

To make it easy to grep for the removed XA5215 and XA5216 messages codes
in the future, add them to `Documentation/guides/messages/README.md`
under a new `Removed messages` section.  That will help prevent
accidental reuse of the codes for new messages, which could result in a
confusing mixture of old and new web search results for the codes.

Other changes:

Move a few other messages that are no longer used to the `Removed
messages` section of the README.


Documentation/guides/messages/README.md
```diff
@@ -135,6 +135,8 @@ ms.date: 01/24/2020
 + [XA5207](xa5207.md): Could not find android.jar for API level `{compileSdk}`.
 + XA5211: Embedded Wear app package name differs from handheld app package name ({wearPackageName} != {handheldPackageName}).
 + XA5213: java.lang.OutOfMemoryError. Consider increasing the value of $(JavaMaximumHeapSize). Java ran out of memory while executing '{tool} {arguments}'
++ XA5215: Duplicate Resource found for {elementName}. Duplicates are in {filenames}
++ XA5216: Resource entry {elementName} is already defined in {filename}
```
brendanzagaeski, line 139: I added these lines so I could ask if it does seem useful to add them. The idea I had was that keeping old message codes in the README could make it easier to avoid reusing them in future, and avoiding reuse would mean the web search results for the codes won't get mixed up between the old and new uses.

Another way to search for old codes is to use `git log`, but having them in the README makes it easier.

jonpryor: This is a good idea, but we should probably add `[removed in Xamarin.Android 10.4]` to the end of this messages or something.

brendanzagaeski:

Add a new section to the messages README for removed messages. Move XA5215 and XA5216 to that section under the tentative release where they will be removed.

Remove the `$(_AndroidResourcesCacheFile)` MSBuild property that is no longer used.

Search for any remaining appearances `MergeResources` and `ResourceMerger`. I didn't find any on master. It looks like maybe the GitHub search index hadn't yet caught up with master for that line Dean noticed.

I have updated the pull request description to cover the additional changes.

Documentation/guides/messages/README.md
```diff
@@ -55,17 +55,14 @@ ms.date: 01/24/2020
 + [XA0107](xa0107.md): `{Assmebly}` is a Reference Assembly.
 + [XA0108](xa0108.md): Could not get version from `lint`.
 + [XA0109](xa0109.md): Unsupported or invalid `$(TargetFrameworkVersion)` value of 'v4.5'.
-+ [XA0110](xa0110.md): Disabling $(AndroidExplicitCrunch) as it is not supported by `aapt2`. If you wish to use $(AndroidExplicitCrunch) please set $(AndroidUseAapt2) to false.
```
jonpryor, line 58: As with various musings elsewhere, I don't think we want to *remove* these codes from the list.  We instead want to say that they're obsolete or something.

brendanzagaeski:
From what I could tell from my "Search for any remaining appearances," all the involved tests were already removed in <https://github.com/xamarin/xamarin-android/commit/ff84922e60808f12b60776a94323444c38dbf088>. For example, if I open 
 [`BuildTests.cs` from master](https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs), I can't find any mentions of `MergeResources`

PR #4529, opened 2020-04-06
[tests] CheckAapt2WarningsDoNotGenerateErrors non-parallelizable

jonathanpeppers:
Context: https://build.azdo.io/3615798
Context: https://build.azdo.io/3615816

This test commonly fails with:

    (Restore target) ->
    NuGet.targets(124,5): error : Could not find file '/Users/runner/runners/2.165.2/work/1/s/packages/xamarin.forms/2.3.4.231/ltlis65w.fez'.
        0 Warning(s)
        1 Error(s)

Looking at the other tests that use
`KnownPackages.XamarinForms_2_3_4_231`, they already have
`[NonParallelizable]`.

I think we should do the same for this test.


jonathanpeppers:
Some performance tests are failing since 7117414ca27d88a71b4a272705a0207f772423bd, I'll look into that in a separate PR.

PR #4534, opened 2020-04-07
[build] bump to .NET Core 3.1.201

jonathanpeppers:
Seeing if this helps anything at all.

### DIFF
--- a/build-tools/automation/build.groovy
+++ b/build-tools/automation/build.groovy
@@ -175,7 +175,7 @@ timestamps {
                 }
             }
             // Install .NET Core and temporarily append it to PATH
-            sh "curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin -version 3.1.100"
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
The problem here is something related to how `$PATH` is being set:
```
08:32:09  + curl -sSL https://dot.net/v1/dotnet-install.sh
08:32:09  + bash /dev/stdin -version 3.1.201
08:32:09  dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Sdk/3.1.201/dotnet-sdk-3.1.201-osx-x64.tar.gz
08:32:14  dotnet-install: Extracting zip from https://dotnetcli.azureedge.net/dotnet/Sdk/3.1.201/dotnet-sdk-3.1.201-osx-x64.tar.gz
08:32:16  dotnet-install: Adding to current process PATH: `/Users/builder/.dotnet`. Note: This change will be visible only when sourcing script.
08:32:16  dotnet-install: Installation finished successfully.
```

PR #4537, opened 2020-04-07
[build] remove remaining packages.config usage

jonathanpeppers:
In an effort to port the remaining projects in xamarin-android to
short-form MSBuild projects, a first step is make sure we are using
`<PackageReference/>` everywhere.

I went ahead and migrated a few "easy" projects to short-form:

* `api-merge.csproj`
* `jnienv-gen.csproj`
* `jit-times.csproj`

As a result of these changes `xaprepare` no longer needs to run:

    .nuget\NuGet.exe restore Xamarin.Android.sln

against any of the `.sln` files in this repository. We can rely on
`msbuild /restore` to build and restore everything now.

We still need `.nuget\NuGet.exe`, as it will be used during MSBuild
tests using `packages.config`.


jonathanpeppers:
It seems like all the Windows build had provisionator failures.

I'll re-run those later, once we figure out what is going on.

Makefile
```diff
@@ -102,7 +102,7 @@ MSBUILD_FLAGS += /p:AndroidApiLevel=$(API_LEVEL) /p:AndroidFrameworkVersion=$(wo
 endif
 
 all-tests::
-	MSBUILD="$(MSBUILD)" $(call MSBUILD_BINLOG,all-tests,tools/scripts/xabuild) $(MSBUILD_FLAGS) Xamarin.Android-Tests.sln
+	MSBUILD="$(MSBUILD)" $(call MSBUILD_BINLOG,all-tests,tools/scripts/xabuild) /restore $(MSBUILD_FLAGS) Xamarin.Android-Tests.sln
```
jonpryor, line 105: I'm surprised that 295aff28371fc9f3cbe1e1f937eb7fa6634ad670 didn't need this...

build-tools/create-vsix/create-vsix.csproj
```diff
@@ -45,63 +45,26 @@
...
-      <HintPath>..\..\packages\Microsoft.VisualStudio.Threading.15.0.240\lib\net45\Microsoft.VisualStudio.Threading.dll</HintPath>
-    </Reference>
-    <Reference Include="Microsoft.VisualStudio.Utilities, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.VisualStudio.Utilities.15.0.26201\lib\net45\Microsoft.VisualStudio.Utilities.dll</HintPath>
-      <Private>True</Private>
-    </Reference>
-    <Reference Include="Microsoft.VisualStudio.Validation, Version=15.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.VisualStudio.Validation.15.0.82\lib\net45\Microsoft.VisualStudio.Validation.dll</HintPath>
-    </Reference>
+    <PackageReference Include="Microsoft.VisualStudio.CoreUtility" Version="15.0.26201" />
```
jonpryor, line 54: Should these include `Private="True"` as well?  Is `%(PackageReference.Private)` even A Thing™?

pjcollins: I think `Private` controlled "copy this assembly to the output dir" with packages.config style projects. I think the closest thing to this for PackageReference would be the usage of `<CopyLocalLockFileAssemblies>` (see https://github.com/xamarin/xamarin-android/blob/master/Directory.Build.props#L12).

Something else is funky with these project changes though, as it seems that our builds are no longer creating a .vsix file, though the attempt to do so is still succeeding...

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3628840&view=logs&j=96fd57f5-f69e-53c7-3d47-f67e6cf9b93e&t=38f83f46-bc21-5edd-1614-e44f20babf10&l=46587

build-tools/create-vsix/create-vsix.csproj
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\Microsoft.VSSDK.BuildTools.15.0.26201\build\Microsoft.VSSDK.BuildTools.props" Condition="Exists('..\..\packages\Microsoft.VSSDK.BuildTools.15.0.26201\build\Microsoft.VSSDK.BuildTools.props')" />
+  <Import Project="..\..\packages\microsoft.vssdk.buildtools\15.0.26201\build\Microsoft.VSSDK.BuildTools.props" Condition="Exists('..\..\packages\microsoft.vssdk.buildtools\15.0.26201\build\Microsoft.VSSDK.BuildTools.props')" />
```
jonathanpeppers, line 3: @pjcollins this was the line that broke the vsix, I needed to update this path.

jonathanpeppers:
The one test failure:
```
Exceeded expected time of 9800ms, actual 9880ms
TearDown : System.IO.FileNotFoundException : Test attachment file path could not be found.
```
It seems unrelated, I will bump this number in a separate PR if we keep seeing it.

Issue #1560, opened 2018-04-16
Add error codes to Warnings


joj:
There is a `LogCodedWarning` method in Utilities (https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Utilities/MSBuildExtensions.cs#L119) that is never used, as far as I can tell. We need warnings to provide proper error codes, so VS can show them in the correct column in the error list and we can react automatically to known errors that may be fixed easily.

As part of this, we need to make sure that `LogCodedError` is used instead of `LogError` (or that `LogError` is always used with errorCode as a parameter), for the same reason.

> VS bug [#601721](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/601721)

bholmes:
Jon found 2 instances of `LogCodedWarning ` via grep.  We do need more.

jonpryor:
What we should *also* do is *document* these codes. We should add a new `Documentation/guides/messages` directory, which contains one `.md` file per warning/error code, which contains information about the message.

For example, `Documentation/guides/messages/xa4212.md` would contain information about error code XA4212, added in PR#797.

jonathanpeppers:
@brendanzagaeski we made progress towards this in d16-0, but there are probably more codes...

Since this task is so open-ended, should we re-milestone to d16-1?

brendanzagaeski:
Re-milestoning to d16-1 sounds good to me. I'll continue to prioritize the remaining items roughly based on how much project customization is required to hit each one, and I'll keep a tentative goal to complete all of them by the d16-1 milestone.

jpobst:
This seems largely done.  Additional issue(s) should be opened for any remaining occurrences found.

PR #4538, opened 2020-04-08
[Xamarin.Android.Build.Tasks] Layout code-behind l10n support

brendanzagaeski:
Context: https://github.com/xamarin/xamarin-android/issues/1560

Add codes to the existing warnings and errors in the layout code-behind
tasks, and move the message strings into the `.resx` file so that they
are localizable.

Remove `<GenerateCodeBehindForLayout/>` so that it doesn't need to be
localized.  The original uses of that task were removed in commit
ab3773cf.


dellis1972:
When did we stop using `GenerateCodeBehindForLayout` ?

brendanzagaeski:
From `git log -G GenerateCodeBehindForLayout`, what I saw was that `<GenerateCodeBehindForLayout/>` was removed from the *.targets* in <https://github.com/xamarin/xamarin-android/commit/ab3773cf439ab620934b04e1bd01f0d9cf61a6ae> and then the only other changes related to that task afterward were cleanup change across all the build tasks: 28e1f9308c8697bc47e9e3220ffb5bd32c961620, d4c8f077faefdfc6174355848a9d8c74ecaa8f56.

PR #4544, opened 2020-04-09
[.NET 5] Resolve app content from runtime packs

pjcollins:
Partially reworks the `BuildApk` task to accept a dynamic list of native
libraries needed for the app. System library names, and the logic which
would conditionally include them that was previously defined in the
`BuildApk` task has been moved into Xamarin.Android.Common.targets.

The `PreferNativeLibrariesWithDebugSymbols` parameter has been removed
from `BuildApk`.  We have not been shipping native libraries with debug
symbols, and as a result this check was not changing project packaging
behavior.

A RuntimeIdentifier <-> AndroidAbi mapping has been added to 
`MonoAndroidHelper.cs` to ensure that components coming from the 
`@(ResolvedFileToPublish)` ItemGroup would end up in the right location
in the .apk file.

The .NET 5 behavior for `_ResolveAssemblies` has been updated to also
use the `@(ResolvedFileToPublish)` items, to ensure that we include the
implementation version of these assemblies in our .apk / .aab files.

Linking has temporarily been disabled by default for .NET 5 apps while
a compatible solution is being worked on.

Shared runtime usage has also been disabled by default for .NET 5
apps.

AOT cross compilers and Android architecture specific native libraries
have been removed from Xamarin.Android.Sdk.

A couple of other runtime tweaks from
https://github.com/xamarin/xamarin-android/tree/test-net5-drop
have been applied to ensure a self published app can run.

### How to test

  1) Install the [latest .NET 5 preview][0].
  2) Download the .nupkg files from a CI build, and create a 
    `NuGet.config` file that has a package source pointing to them and
    the .NET 5 feed:

```xml
<configuration>
  <packageSources>
    <add key="xanetsdk" value="packages" />
    <add key="dotnet5" value="https://dnceng.pkgs.visualstudio.com/public/_packaging/dotnet5/nuget/v3/index.json" />
  </packageSources>
</configuration>
```

  3) Open an existing Android project (ideally something minimal) and
    tweak it as shown below. The version should match the version of the
    packages that you downloaded.

```xml
<Project Sdk="Xamarin.Android.Sdk/10.4.99.7">
  <PropertyGroup>
    <TargetFramework>netcoreapp5.0</TargetFramework>
    <RuntimeIdentifier>android.21-arm64</RuntimeIdentifier>
    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
  </PropertyGroup>
</Project>
```

  4) Publish (and optionally install) the project:

```
dotnet publish -t:Install -p:Configuration=Release *.csproj --self-contained
```

[0]:  https://github.com/dotnet/installer#installers-and-binaries


pjcollins:
@dellis1972 @jonathanpeppers @jonpryor @Redth This is still in a very early stage and has a handful of hacks that we'll need to address, but I wanted to get some extra eyes, testing, and feedback whenever possible.

src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
```diff
@@ -19,7 +19,7 @@ static void get_runtime_types ()
 			mono_unhandled_exception_method = typeof (System.Diagnostics.Debugger).GetMethod (
 				"Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
 			if (mono_unhandled_exception_method == null)
-				AndroidEnvironment.FailFast ("Cannot find System.Diagnostics.Debugger.Mono_UnhandledException");
+				//AndroidEnvironment.FailFast ("Cannot find System.Diagnostics.Debugger.Mono_UnhandledException");
```
jonathanpeppers, line 22: This is also called here:

https://github.com/xamarin/xamarin-android/blob/d48cf04f9749664bf48fc16bcb920d5d941cccab/src/Mono.Android/Android.Runtime/JNIEnv.cs#L242-L279

@marek-safar / @steveisok is there an equivalent for `System.Diagnostics.Debugger.Mono_UnhandledException`?

We think this is what tells the debugger to break in the IDE if an unhandled exception is thrown.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2386,12 +2408,10 @@ because xbuild doesn't support framework reference assemblies.
     CreatePackagePerAbi="$(AndroidCreatePackagePerAbi)"
     UseSharedRuntime="$(AndroidUseSharedRuntime)"
     Debug="$(AndroidIncludeDebugSymbols)"
-    PreferNativeLibrariesWithDebugSymbols="$(AndroidPreferNativeLibrariesWithDebugSymbols)"
```
dellis1972, line 2389: @pjcollins I believe while this is not used by users, it might have been used by @grendello to help with debugging our runtime in an embedded assemblies scenario. @grendello ever used this property?

src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
marek-safar: I think @lambdageek commented on that in XI doc. Do you know why is it needed?

/cc @thaystg

thaystg: I don't think we have anything similar for XI.
This is used for trigger the debugged process to send an unhandled exception message to the agent. This is needed since MonoDroid can't know at the Java->C# boundary if a given C# exception is unmanaged or not.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
pjcollins: The scenario for local debugging could still be supported by these changes, you'd just need to change these file names in your common targets: https://github.com/xamarin/xamarin-android/pull/4544/files/319ae454657c479f57c1d0e4e5c6cddfd2ad438d#diff-db39bf07b89b50eaab8ab2001546310cR2196

pjcollins:
Full test run triggered here: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3648768&view=results

src/Xamarin.Android.Sdk/targets/Xamarin.Android.Sdk.props
```diff
@@ -32,14 +32,16 @@
 
   <!--  User-facing configuration-specific defaults -->
   <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
-    <AndroidUseSharedRuntime Condition=" '$(AndroidUseSharedRuntime)' == '' ">true</AndroidUseSharedRuntime>
+    <!--FIXME: Disable the shared runtime by default -->
+    <AndroidUseSharedRuntime Condition=" '$(AndroidUseSharedRuntime)' == '' ">false</AndroidUseSharedRuntime>
```
jonathanpeppers, line 36: Also default `$(EmbedAssembliesIntoApk)` to `true` for .NET 5 projects, so we won't be using Fast Deployment yet at all.

pjcollins:
The regression test failures look fine, I fixed one instance that was unrelated here: https://github.com/xamarin/QualityAssurance/commit/0e8f369bfddf1f903c53ff1ae070a3adc7e226d6.

It does look like some instant run scenarios are potentially broken now though, I need to investigate further. This may require additional monodroid changes.

PR #4545, opened 2020-04-09
[tests] New apk size regression testing

radekdoulik:
Use apkdiff and description files to check for apk size regression.

The new `ApkDiffCheckRegression` tool task handles running `apkdiff`
tool and creates `TestResult` file for convenient CI reporting.


radekdoulik:
Note that this is draft PR with wrong `Xamarin.Forms_Performance_Integration-Signed-Debug.apkdesc` to test the creation of `TestResult`  xml file and picking it up by CI.

radekdoulik:
Looks like we don't run instrumentation tests with Debug configuration. Maybe we should start running them again?

PR #4547, opened 2020-04-13
[docs] Update release note for R8 -ignorewarnings

brendanzagaeski:
Restructure the R8 `-ignorewarnings` release note to provide details
about two examples of common warnings and errors that users have seen
when using R8 for multidex.

Remove the description about "just warnings" when using `dx` because in
my local testing, using DX for multidex did not produce any related
warnings, at least not for these two common known scenarios.

### DIFF
--- a/Documentation/release-notes/4509.md
+++ b/Documentation/release-notes/4509.md
@@ -1,42 +1,142 @@
-### R8 uses `-ignorewarnings` to suppress common build errors
+### R8 configured to suppress common build errors
 
-Developers could hit a common problem in applications using Google
-Play Services:
+Xamarin.Android now configures R8 to skip emitting an error for Java dependency
+warnings such as:
 
-    Warning: Missing class: org.apache.http.client.methods.HttpEntityEnclosingRequestBase
-    Error: Compilation can't be completed because some library classes are missing.
-
-The correct fix is to adjust `AndroidManifest.xml` in the project:
-
-```xml
-<application ...>
-  <uses-library android:name="org.apache.http.legacy" android:required="false" />
-</application>
+```
+R8 : warning : Missing class: org.apache.http.client.methods.HttpEntityEnclosingRequestBase
+R8 : warning : Missing class: java.lang.ClassValue
+R8 : error : Compilation can't be completed because some library classes are missing.
 ```
 
-This issue could be encountered if:
+This allows a more gradual transition to R8, particularly for projects that are
+using R8 for multidex only and not code shrinking.
 
-* `AndroidDexTool=d8` (this is default in Xamarin.Android 10.2 and higher)
-* `AndroidEnableMultiDex=True` or `AndroidLinkTool=r8`
-
-When using `AndroidDexTool=dx`, the same project would not encounter a
-build error--just warnings. However, `r8` is more strict and converts
-the `library classes are missing` warning into a failure.
-
-Xamarin.Android now applies a setting that will make `r8` more lenient
-to keep things closer to the existing `dx` behavior. By automatically
-passing `-ignorewarnings`, `r8` will not fail the build, but still
-emit warnings.
-
-To disable this behavior, a new `$(AndroidR8IgnoreWarnings)` MSBuild
-property can be set to `False` in your `.csproj` file:
+Although R8 warnings will no longer cause the build to fail, project authors are
+encouraged to address the warnings at their earliest convenience and then set R8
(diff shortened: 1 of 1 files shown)


Documentation/release-notes/4509.md
```diff
@@ -1,42 +1,131 @@
...
+
+ 3. Edit the _.csproj_ file in a text editor and add `--pg-conf
+    proguard-rules.pro` to the `AndroidR8ExtraArguments` MSBuild property, for
+    example by adding the following lines:
+
+    ```xml
+    <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+      <AndroidR8ExtraArguments>--pg-conf proguard-rules.pro</AndroidR8ExtraArguments>
+    </PropertyGroup>
+    ```
```
brendanzagaeski, line 106: I used `$(AndroidR8ExtraArguments)` rather than `@(ProguardConfiguration)` because `@(ProguardConfiguration)` items are currently only passed to R8 if code shrinking is enabled.

I thought about that a bit, and as far as I have been able to imagine so far, I think it would work nicely to change the `<R8/>` task slightly so that it supplies `@(ProguardConfiguration)` items to R8 in the multidex-only scenario too. Now that we know that R8 does many or all of the same dependency checks in multidex-only mode as in code shrinking mode, it seems advantageous to pass the `@(ProguardConfiguration)` items in that scenario too. I haven't yet been able to think of any problems with doing that.

I'll submit a PR for that as soon as I get a chance, if no one beats me to it.

Documentation/release-notes/4509.md
```diff
@@ -1,42 +1,131 @@
...
+[java.lang package][java-lang-android].
+
+One known scenario that produces this warning is when an app references the
+Xamarin.Firebase.Firestore NuGet package.  That package depends on the
+cross-platform Google Guava library.  Guava uses the `java.lang.ClassValue` type
+only on platforms where it is available, so it is safe to suppress the _Missing
+class: java.lang.ClassValue_ warning in that case.
+
+A future version of the Xamarin.Guava.Google NuGet package will provide the
+`-dontwarn java.lang.ClassValue` R8 rule to suppress the warning automatically.
```
brendanzagaeski, line 123: I started work on adding the appropriate configuration file to Xamarin.Guava.Google: <https://github.com/xamarin/XamarinComponents/pull/830>.

Along the way, I realized that the Xamarin.Guava.Google NuGet package had recently become a little unusual in its NuGet dependency story due to the fix of using `@(EmbeddedReferenceJar)` to get `guava.jar` from the NuGet package into dependent bindings projects and then into apps. That was going to make it awkward to include the R8 configuration file. So I set up a new fix to make the NuGet dependencies work as expected again: <https://github.com/xamarin/XamarinComponents/pull/831>.

(At the moment, <https://github.com/xamarin/XamarinComponents/pull/830> is not yet rebased on <https://github.com/xamarin/XamarinComponents/pull/831>.)

If we decide that is OK to change `<R8/>` to pass `@(ProguardConfiguration)` items to R8 in multidex-only mode, then <https://github.com/xamarin/XamarinComponents/pull/830> can be simplified even more.

brendanzagaeski:
As I'm looking at how long these notes are with everything I put in, I'm thinking there might be someplace better to move that content to keep the release notes a little shorter. Maybe a devblogs.microsoft.com post or a non-release-notes page on docs.microsoft.com.

Documentation/release-notes/4509.md
jonathanpeppers: They could hit this error with `AndroidLinkTool=r8`, too. Should this first mention the normal way to use proguard rules as the common case?

brendanzagaeski:

Update the wording to match what I settled on for the Xamarin.Android 10.3.0.74 Preview release notes. From Peppers's suggestion, the new wording now includes a note about how to set `@(ProguardConfiguration)` when `$(AndroidLinkTool)`=`r8`.

This can now be merged. After this is merged, other suggestions or edits are still welcome via comments on this PR or candidate changes to the file in follow-up PRs.

I kept the somewhat lengthy "How to solve" sections for now. I might revisit that before the Xamarin.Android 10.3 Release, but for now I think it's fine. I remembered that there isn't yet a mention of D8 or R8 under <https://docs.microsoft.com/xamarin/android/deploy-test/>, so there isn't an existing page where these examples would fit. I'll keep thinking about it a little to figure out if I'll move the examples somewhere else before the 10.3 Release.

Thanks!

PR #4549, opened 2020-04-13
Support for sending plots to AppInsights

mjbond-msft:
Provides the capability to send plot information to AppInsights using a command line tool ProcessPlotCSVFile.exe. The tool extracts the plot information from the provided CSV file, which contains series labels with associated values.  The tool simply parses out this information and sends it to AppInsights along with additional information such as the build number, commit id, date/time, etc.

Configuration for each plot is quite similar to the configuration found for processing of the same plot information by the Jenkins build.  To process plots within a CSV file, add a template reference to  `yaml-templates/plots-to-appinsights.yaml` to the YAML workflow.  You need to provide a plot group such as `Test size` and a plot title such as `Runtime test sizes` along with the name of the CSV file containing the plots.

Plot graphs are available on the [Xamarin Android Plots](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/dashboard/arm/subscriptions/64e11c84-c922-4ffd-bea9-67ab39354edd/resourcegroups/dashboards/providers/microsoft.portal/dashboards/6dd56150-d5c1-4e0b-b57d-147551b050c2) dashboard
Plot data is available in the AppInsights account: [XamarinAndroidMetrics](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/64e11c84-c922-4ffd-bea9-67ab39354edd/resourceGroups/XamarinMetrics/providers/microsoft.insights/components/XamarinAndroidMetrics/logs)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -685,6 +685,76 @@ stages:
...
+      inputs:
+        solution: build-tools/plots-to-appinsights/ProcessPlotCSVFile.csproj
+        configuration: $(ApkTestConfiguration)
+        msbuildArguments: >
+          /restore
+          /t:Build
+          /v:normal
+          /bl:$(System.DefaultWorkingDirectory)/bin/Tools$(ApkTestConfiguration)plots-to-appinsights.binlog
+      continueOnError: true
+      condition: and(succeeded(), or(eq(variables['Build.Reason'], 'IndividualCI'), eq(variables['Build.Reason'], 'Manual')))
```
jpobst, line 700: @mjbond-msft What are these additional `condition`'s for?

@jonpryor @pjcollins Should we run this on PR builds?  It feels like we shouldn't.  On one hand we wouldn't catch a regression before we merged, but we also don't want PR tests messing up our graphs.

build-tools/plots-to-appinsights/App.config
```diff
@@ -0,0 +1,14 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
jpobst, line 1: Can kill this file, it doesn't do anything.

build-tools/automation/yaml-templates/plots-to-appinsights.yaml
```diff
@@ -0,0 +1,63 @@
+parameters:
+  framework: net472     # TargetFramework setting for the build-tools/plots-to-appinsights/ProcessPlotCSVFile source project that becomes part of the build output directory
```
jpobst, line 2: Optionally, you can add `<AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>` to the `csproj` since we aren't multi-targeting this project and then it won't be nested in a `net472` output directory and this can be killed.

jpobst:
I think our team should revisit what data we want to generate and chart later on, but this looks like it solves our problem of ***where*** to put the data and ***how*** to get it there.

build-tools/plots-to-appinsights/App.config
jonpryor: Indentation is also wonky, but killing the file will fix the indentation. :-)

build-tools/plots-to-appinsights/AppInsights.cs
```diff
@@ -0,0 +1,39 @@
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Extensibility;
+using System;
+using System.Collections.Generic;
+
+namespace ProcessPlotCSVFile
+{
+	internal class AppInsights
+	{
+		private TelemetryClient TelemetryClient { get; set; } = null;
```
jonpryor, line 10: No need for this to be a *property* when it's `private`; leave it a field.

If the property weren't `private`, no need for it to have a setter either, or a non-private setter, and no need for `= null` either.

build-tools/plots-to-appinsights/AppInsights.cs
```diff
@@ -0,0 +1,39 @@
...
+		}
+
+		public void SendTelemetry (string eventName, Dictionary<string, string> properties, Dictionary<string, double> metrics)
+		{
+			if (TelemetryClient != null && !string.IsNullOrWhiteSpace (TelemetryClient.InstrumentationKey)) {
+				lock (TelemetryClient) {
+					TelemetryClient.TrackEvent (eventName, properties, metrics);
+				}
+			} else {
+				throw new Exception ("TelemetryClient not initialized or AppInsights instrumentation key not set");
```
jonpryor, line 33: Is it possible for `TelemetryClient` to be `null`?  It shouldn't be, and thus this `else` block can be removed.

build-tools/plots-to-appinsights/AppInsights.cs
```diff
@@ -0,0 +1,39 @@
+using Microsoft.ApplicationInsights;
+using Microsoft.ApplicationInsights.Extensibility;
+using System;
+using System.Collections.Generic;
+
+namespace ProcessPlotCSVFile
```
jonpryor, line 6: Namespace should have a `Xamarin.Android.Tools` prefix, e.g. `Xamarin.Android.Tools.Pots`.

build-tools/plots-to-appinsights/ProcessPlotCSVFile.sln
```diff
@@ -0,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
```
jonpryor, line 2: No need for this `.sln` file.

Instead, `build-tools/plots-to-appinsights/*.csproj` should be part of `Xamarin.Android.sln`.

build-tools/plots-to-appinsights/Program.cs
```diff
@@ -0,0 +1,342 @@
...
+				}
+
+				tw.WriteLine ("    Metrics:");
+				foreach (var metric in metrics.OrderBy (x => x.Key)) {
+					tw.WriteLine ($"      {metric.Key}: {metric.Value}");
+				}
+			}
+		}
+
+		static void SendTelemetry_Warnings (AppInsights appInsightsClient, string eventName, List<string> warnings, Settings settings, TextWriter tw)
```
jonpryor, line 256: Method names shouldn't have `_` except in unit tests.

build-tools/plots-to-appinsights/ProcessPlotCSVFile.sln
jpobst: I think this is more akin to `xaprepare` which has a `xaprepare.sln` so it can be built on each test agent rather than being built in the `Mac Build` step and then included in the installer.  This tool is only used on the agent running `APK Instrumentation` so it probably makes more sense to only build it there.  

https://github.com/xamarin/xamarin-android/blob/master/build-tools/automation/yaml-templates/setup-test-environment.yaml#L14-L19

build-tools/automation/azure-pipelines.yaml
mjbond-msft: the conditions are there to ensure that we do not collect or report plot information during PR builds. basically, applied the same policy as found for the current `xamarin-android` Jenkins build where plots are only reported for CI builds

mjbond-msft: i should note that i'm enabling plots for _manual_ builds. this mostly serves as a convenience to be able to launch a manual build against a feature branch to test that plots are getting processed correctly since they won't be processed as part of a PR

mjbond-msft: @jpobst regarding the comment "don't want PR tests messing up our graphs" the Build.Reason (PR, CI, Manual) is included in the telemetry being sent. so _if_ the decision is made to include plots as part of PRs, the telemetry query backing the graph would simply filter on Build.Reason == "CI" (and i'll make sure i have this in the backing query) and so PR plots wouldn't mess up the graphs

build-tools/plots-to-appinsights/App.config
mjbond-msft: So long App.config. Commit: 71c3a38587a0dc8e4cd76b74b3302acb980484bf

build-tools/automation/yaml-templates/plots-to-appinsights.yaml
mjbond-msft: Excellent suggestion and will avoid maintenance headaches down the road.  Addressed with commit: f4a541069b53112d50dbae1395449f840eb14dd0

build-tools/plots-to-appinsights/AppInsights.cs
mjbond-msft: Indeed the null/empty checks aren't needed as the constructor should keep everything honest.
Addressed with commit: 135324e3afd3da9c3c91a9de110ae83e56df3353

build-tools/plots-to-appinsights/AppInsights.cs
mjbond-msft: Indeed the null/empty checks aren't needed as the constructor should keep everything honest.
Addressed with commit: 135324e3afd3da9c3c91a9de110ae83e56df3353

build-tools/plots-to-appinsights/AppInsights.cs
mjbond-msft: Ah, good to know for any future tools.
Commit: 8cf10948bf271e7e37d026afd255c6003ad24cbd

build-tools/plots-to-appinsights/ProcessPlotCSVFile.sln
mjbond-msft: Indeed. I pattered this after tools that build & run as part of the test jobs such as `APK Instrumentation`

build-tools/plots-to-appinsights/Program.cs
mjbond-msft: Nuked the underscores. Commit: 56a0b717242a0d9ad9665993b7bfb9dff258eb2d

PR #4551, opened 2020-04-13
[docs] Put newest download locations first in README

brendanzagaeski:
When the xamarin-macios team recently added a downloads section to their
README, they noticed that users were by default picking the first link,
so they decided to switch the order to sort from newest to oldest.

Update the commercial download locations in the xamarin-android README
to use newest-to-oldest order like the xamarin-macios README.

Leave the OSS master download as the last entry because that link is
expected to be used less frequently than the links for the commercial
versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,27 +23,34 @@ Xamarin.Android provides open-source bindings of the Android SDK for use with
 
 # Downloads
 
-| Platform        | Status |
+## Current
+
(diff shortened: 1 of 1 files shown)


brendanzagaeski:

Group the downloads into "Current" and "Previous" sections so that the OSS master download can go at the bottom of the "Current" section.

PR #4553, opened 2020-04-14
[Xamarin.Android.Build.Tasks] Use @(ProguardConfiguration) for R8 multidex

brendanzagaeski:
Context: https://github.com/xamarin/xamarin-android/issues/4369#issuecomment-610553571
Context: https://developercommunity.visualstudio.com/content/problem/979562/r8-warning-missing-class-javalangclassvalue.html
Context: https://github.com/xamarin/AndroidX/issues/74

Currently, Xamarin.Android always uses `com.android.tools.r8.R8` to
generate DEX files if `$(AndroidDexTool)`=`d8` and
`$(AndroidEnableMultiDex)`=`True`, even if `$(AndroidLinkTool)`=``.

This causes some tricky behavior because, unlike DX and
`com.android.tools.r8.D8`, R8 always performs dependency checks,
including [warning about missing Java classes][0].  As a consequence,
switching from `$(AndroidDexTool)`=`dx` to `$(AndroidDexTool)`=`d8`
while `$(AndroidEnableMultiDex)`=`True` can introduce new build
warnings, somewhat unexpectedly.

Commit 6ee55837 improved the user experience around this scenario by
suppressing the additional error that R8 was emitting when it emitted
any `Missing class` warnings.

Add another improvement for this scenario by adjusting `<R8/>` to pass
`@(ProguardConfiguration)` items to R8, regardless of whether code
shrinking is enabled or not.  This allows the normal expected R8
configuration file mechanism to work in multidex-only mode.

Leave the lines that add `proguard_xamarin.cfg` and the lines that
generate a ProGuard configuration file based on `AcwMapFile` where they
are because `-keep` and `-keepclassmembers` rules are not relevant in
multidex-only mode.

TODO:

Android Studio avoids this tricky scenario altogether by using D8 rather
than R8 unless code shrinking is enabled.  Because D8 is designed to
have only a `--main-dex-list` option and no `--main-dex-rules` option,
Android Studio pre-processes any `--main-dex-rules` files into the
`--main-dex-list` format using a separate [`D8MainDexListTask`][1] step
that runs [`com.android.tools.r8.GenerateMainDexList`][2] before calling
D8.

Xamarin.Android could consider switching to that same approach.  On the
other hand, the warnings that R8 produces are accurate.  They indicate
missing dependencies that could cause apps to abort at run time.
Getting warnings for those issues at build time is an advantage of the
current approach.

[0]: https://r8.googlesource.com/r8/+/refs/tags/1.5.68/src/main/java/com/android/tools/r8/R8.java#291
[1]: https://android.googlesource.com/platform/tools/base/+/81f184dbf8d7caefd09ee7309c1427425da6415c/build-system/gradle-core/src/main/java/com/android/build/gradle/internal/TaskManager.java#2206
[2]: https://r8.googlesource.com/r8/+/refs/tags/1.5.68/src/main/java/com/android/tools/r8/GenerateMainDexListCommand.java

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/R8.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/R8.cs
@@ -104,23 +104,6 @@ namespace Xamarin.Android.Tasks
 						}
 					}
 				}
-				if (!string.IsNullOrEmpty (ProguardConfigurationFiles)) {
-					var configs = ProguardConfigurationFiles
-						.Replace ("{sdk.dir}", AndroidSdkDirectory + Path.DirectorySeparatorChar)
-						.Replace ("{intermediate.common.xamarin}", ProguardCommonXamarinConfiguration)
-						.Replace ("{intermediate.references}", ProguardGeneratedReferenceConfiguration)
-						.Replace ("{intermediate.application}", ProguardGeneratedApplicationConfiguration)
-						.Replace ("{project}", string.Empty) // current directory anyways.
-						.Split (';')
-						.Select (s => s.Trim ())
-						.Where (s => !string.IsNullOrWhiteSpace (s));
-					foreach (var file in configs) {
-						if (File.Exists (file))
-							cmd.AppendSwitchIfNotNull ("--pg-conf ", file);
-						else
-							Log.LogCodedWarning ("XA4304", file, 0, Properties.Resources.XA4304, file);
-					}
-				}
 			} else {
 				//NOTE: we may be calling r8 *only* for multi-dex, and all shrinking is disabled
 				cmd.AppendSwitch ("--no-tree-shaking");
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/R8.cs
```diff
@@ -139,6 +122,23 @@ protected override CommandLineBuilder GetCommandLineBuilder ()
 				tempFiles.Add (temp);
 				cmd.AppendSwitchIfNotNull ("--pg-conf ", temp);
 			}
+			if (!string.IsNullOrEmpty (ProguardConfigurationFiles)) {
```
jonathanpeppers, line 125: I think  you might need to adjust which files are passed in by default:
```
BuildHasNoWarnings(False,True,True,"apk")
...
Should have no more than 1 MSBuild warnings.
...
(_CompileToDalvik target) -> 
  obj/Debug/proguard/proguard_xamarin.cfg : warning XA4304: ProGuard configuration file 'obj/Debug/proguard/proguard_xamarin.cfg' was not found.
```
I would guess that we are putting `proguard_xamarin.cfg` in the item group always, and we need to only do that when `$(AndroidLinkTool)` isn't blank?

brendanzagaeski:
Another thing I just realized is that when Xamarin.Android moves to a minimum supported version of Android 5.0 Lollipop (API level 21), the support for `--main-dex-list` and `--main-dex-rules` can be removed entirely, which would allow using D8 for multidex with no preprocessing step.

brendanzagaeski:
*Changes in latest commit:*

Fix the `BuildHasNoWarnings(False,True,True,"apk")` test failure by avoiding trying to add the generated config files like `proguard_xamarin.cfg` to the R8 command line when `$(AndroidLinkTool)` is blank. Those files aren't generated and aren't applicable in the multidex-only scenario.

PR #4554, opened 2020-04-14
[Xamarin.Android.Build.Tasks] Ignore `global` Alias when emitting code.

### LINKED ISSUE #4542: "extern alias" added to auto generated Resource.designer.cs causes syntax errors
I've attached a sample project and the diagnostic build output when building in Visual Studio 16.4.6/Xamarin.Android.10.1 and VS 16.5.3/Xamarin.Android.10.2. The problematic behavior is also seen in VS 16.5.2.

When built with 10.1 the project builds as expected. No `extern alias` lines are added to the generated resource file.

When built with 10.2 the project fails to build because `extern alias` lines (in my sample there are 2) with invalid syntax from the project file are added to the generated resource file.

I understand this may be connected to #3808 and #4409 and their fixes, but it seems there is still an issue with how these lines are added or that they are added at all as they seem to be unnecessary. The aliased libraries are not referenced in the resource file and the project builds fine without them in 10.1.

I modeled my sample project after the real project that has `<Aliases>` tags added for references to the same two nuget packages used in my sample project. I didn't write this code, so there may be something else going on. It might be that it is not even correct and that is the root cause of the problem, but there is definitely a difference in the way it is handled between 10.1 and 10.2 that is less than ideal.

EDIT: One more thing that I should probably mention is that this behavior isn't seen when the packages are added as `<PackageReference>` tags in the project file. When they are added the "old way" with a package.config file and are added to the project file with a `<Reference>` tag the issue appears.

It might also be worth pointing out that I was hoping to update to 10.2 to avoid this issue: [#18106](https://github.com/mono/mono/issues/18106) so right it seems like I am between a rock and a hard place.

[xamarin.android.10.1.txt](https://github.com/xamarin/xamarin-android/files/4452005/xamarin.android.10.1.txt)
[xamarin.android.10.2.txt](https://github.com/xamarin/xamarin-android/files/4452007/xamarin.android.10.2.txt)
[App1.zip](https://github.com/xamarin/xamarin-android/files/4451989/App1.zip)

dellis1972:
Fixes #4542

Our previous fix d18c824c did not take into account that users might
need to add `global` to their list of aliases in the `ProjectReferece`
MetaData. This then causes the following error

	Error CS1681: You cannot redefine the global extern alias (CS1681)

So to fix this we need to ignore the `global` entry if one is present.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
@@ -29,7 +29,7 @@ namespace Xamarin.Android.Tasks
 
 		[Required]
 		public string ResourceDirectory { get; set; }
-		
+
 		public ITaskItem[] AdditionalResourceDirectories { get; set; }
 
 		[Required]
@@ -99,7 +99,7 @@ namespace Xamarin.Android.Tasks
 				var parser = new JavaResourceParser () { Log = Log };
 				resources = parser.Parse (JavaResgenInputFile, IsApplication, resource_fixup);
 			}
-			
+
 			var extension = Path.GetExtension (NetResgenOutputFile);
 			var language = string.Compare (extension, ".fs", StringComparison.OrdinalIgnoreCase) == 0 ? "F#" : CodeDomProvider.GetLanguageFromExtension (extension);
 			bool isVB = string.Equals (extension, ".vb", StringComparison.OrdinalIgnoreCase);
@@ -136,10 +136,17 @@ namespace Xamarin.Android.Tasks
 					ITaskItem item = new TaskItem (assemblyPath);
 					assembly.CopyMetadataTo (item);
 					assemblies.Add (item);
-					string alias = assembly.GetMetadata ("Aliases");
-					if (!string.IsNullOrEmpty (alias)) {
-						foreach (var a in alias.Split (new [] {','}, StringSplitOptions.RemoveEmptyEntries))
-							aliases.Add (a.Trim ());
+					string aliasMetaData = assembly.GetMetadata ("Aliases");
+					if (!string.IsNullOrEmpty (aliasMetaData)) {
+						foreach (var alias in aliasMetaData.Split (new [] {','}, StringSplitOptions.RemoveEmptyEntries)) {
+							string aliasName = alias.Trim ();
+							// don't emit an `extern alias global` as it is implicitly done.
+							if (string.Compare ("global", aliasName, StringComparison.Ordinal) == 0)
+								continue;
+							aliases.Add (aliasName);
+							// only add the first alias for each reference.
+							break;
+						}
 					}
 					Log.LogDebugMessage ("Scan assembly {0} for resource generator", fileName);
 				}
@@ -162,7 +169,7 @@ namespace Xamarin.Android.Tasks
 		// Remove private constructor in F#.
 		// Add static constructor. (but ignored in F#)
 		void AdjustConstructor (CodeTypeDeclaration type)
-		{			
+		{
 			var staticCtor = new CodeTypeConstructor () { Attributes = MemberAttributes.Static };
 			staticCtor.Statements.Add (
 				new CodeExpressionStatement (
@@ -256,7 +263,7 @@ namespace Xamarin.Android.Tasks
 		static string NormalizeAlternative (string value)
 		{
 			int s = value.IndexOfAny (new[] { Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar });
-			
+
 			if (s < 0)
 				return value;
 
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
@@ -18,25 +18,39 @@ namespace Xamarin.Android.Build.Tests
 	public class AndroidUpdateResourcesTest : BaseTest
 	{
 		[Test]
-		public void CheckProjectReferenceAlias ()
+		public void CheckMultipleLibraryProjectReferenceAlias ([Values (true, false)] bool withGlobal)
 		{
 			var path = Path.Combine (Root, "temp", TestName);
-			var library = new XamarinAndroidLibraryProject () {
+			var library1 = new XamarinAndroidLibraryProject () {
 				ProjectName = "Library1",
+			};
+			var library2 = new XamarinAndroidLibraryProject () {
+				ProjectName = "Library2",
+				RootNamespace = "Library1"
 			};
 			var proj = new XamarinAndroidApplicationProject () {
 				References = {
-					new BuildItem.ProjectReference (Path.Combine("..", library.ProjectName, Path.GetFileName (library.ProjectFilePath)), "Library1") {
-						Metadata = { { "Aliases", "Lib1, LibA1,LibB1" } },
+					new BuildItem.ProjectReference (Path.Combine("..", library1.ProjectName, Path.GetFileName (library1.ProjectFilePath)), "Library1") {
+						Metadata = { { "Aliases", withGlobal ? "global,Lib1A,Lib1B" : "Lib1A,Lib1B" } },
+					},
+					new BuildItem.ProjectReference (Path.Combine("..", library2.ProjectName, Path.GetFileName (library2.ProjectFilePath)), "Library2") {
+						Metadata = { { "Aliases", withGlobal ? "global,Lib2A,Lib2B" : "Lib2A,Lib2B" } },
 					},
 				},
 			};
-			using (var builder = CreateDllBuilder (Path.Combine (path, library.ProjectName), cleanupAfterSuccessfulBuild: false, cleanupOnDispose: false)) {
-				builder.ThrowOnBuildFailure = false;
-				Assert.IsTrue (builder.Build (library), "Library should have built.");
-				using (var b = CreateApkBuilder (Path.Combine (path, proj.ProjectName), cleanupAfterSuccessfulBuild: false, cleanupOnDispose: false)) {
-					b.ThrowOnBuildFailure = false;
-					Assert.IsTrue (b.Build (proj), "Project should have built.");
+			using (var builder1 = CreateDllBuilder (Path.Combine (path, library1.ProjectName), cleanupAfterSuccessfulBuild: false, cleanupOnDispose: false)) {
+				builder1.ThrowOnBuildFailure = false;
+				Assert.IsTrue (builder1.Build (library1), "Library should have built.");
+				using (var builder2 = CreateDllBuilder (Path.Combine (path, library2.ProjectName), cleanupAfterSuccessfulBuild: false, cleanupOnDispose: false)) {
+					builder2.ThrowOnBuildFailure = false;
+					Assert.IsTrue (builder2.Build (library2), "Library should have built.");
+					using (var b = CreateApkBuilder (Path.Combine (path, proj.ProjectName), cleanupAfterSuccessfulBuild: false, cleanupOnDispose: false)) {
+						b.ThrowOnBuildFailure = false;
+						Assert.IsTrue (b.Build (proj), "Project should have built.");
+						string[] text = File.ReadAllLines (Path.Combine (Root, path, proj.ProjectName, "Resources", "Resource.designer.cs"));
+						Assert.IsTrue (text.Count (x => x.Contains ("Library1.Resource.String.library_name")) == 2, "library_name resource should be present exactly once for each library");
+						Assert.IsTrue (text.Count (x => x == "extern alias Lib1A;" || x == "extern alias Lib1B;") <= 1, "No more than one extern alias should be present for each library.");
+					}
 				}
 			}
 		}
@@ -88,7 +102,7 @@ namespace Xamarin.Android.Build.Tests
 		[Category ("SmokeTests")]
 		public void DesignTimeBuild ([Values(false, true)] bool isRelease, [Values (false, true)] bool useManagedParser, [Values (false, true)] bool useAapt2)
 		{
-			var regEx = new Regex (@"(?<type>([a-zA-Z_0-9])+)\slibrary_name=(?<value>([0-9A-Za-z])+);", RegexOptions.Compiled | RegexOptions.Multiline ); 
+			var regEx = new Regex (@"(?<type>([a-zA-Z_0-9])+)\slibrary_name=(?<value>([0-9A-Za-z])+);", RegexOptions.Compiled | RegexOptions.Multiline );
 
 			var path = Path.Combine (Root, "temp", $"DesignTimeBuild_{isRelease}_{useManagedParser}_{useAapt2}");
 			var lib = new XamarinAndroidLibraryProject () {
@@ -174,7 +188,7 @@ namespace Xamarin.Android.Build.Tests
 				var image_data = new byte [stream.Length];
 				stream.Read (image_data, 0, (int)stream.Length);
 				image = new AndroidItem.AndroidResource ("Resources\\drawable\\Image.png") { BinaryContent = () => image_data };
-				proj.AndroidResources.Add (image); 
+				proj.AndroidResources.Add (image);
 			}
 			using (var b = CreateApkBuilder ("temp/MoveResource")) {
 				Assert.IsTrue (b.Build (proj), "First build should have succeeded.");
@@ -232,9 +246,9 @@ namespace Xamarin.Android.Build.Tests
 					var image_data = new byte [stream.Length];
 					stream.Read (image_data, 0, (int)stream.Length);
 					image1 = new AndroidItem.AndroidResource ("Resources\\drawable\\Image1.png") { BinaryContent = () => image_data };
-					proj.AndroidResources.Add (image1); 
+					proj.AndroidResources.Add (image1);
 					image2 = new AndroidItem.AndroidResource ("Resources\\drawable\\Image2.png") { BinaryContent = () => image_data };
-					proj.AndroidResources.Add (image2); 
+					proj.AndroidResources.Add (image2);
 				}
 				b.ThrowOnBuildFailure = false;
 				Assert.IsTrue (b.Build (proj), "First build was supposed to build without errors");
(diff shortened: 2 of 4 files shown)


brendanzagaeski:
I have one little question that wouldn't change this PR about `global`, so it can be probably be considered in a separate PR (if it sounds sensible).

In <https://github.com/xamarin/xamarin-android/commit/49d12fdce20f1d0abd926dae3fbbeb56f037d36a>, the `declaringTypeName` line was was updated so it always uses the *first* alias in the list: <https://github.com/xamarin/xamarin-android/blob/49d12fdce20f1d0abd926dae3fbbeb56f037d36a/src/Xamarin.Android.Build.Tasks/Utilities/ResourceDesignerImportGenerator.cs#L63>

To minimize potential conflicts in `Resource.designer.cs`, is it worth making the same adjustment to the line that emits the `extern alias` statements?

https://github.com/xamarin/xamarin-android/blob/d48cf04f9749664bf48fc16bcb920d5d941cccab/src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs#L206-L207

src/Xamarin.Android.Build.Tasks/Utilities/ResourceDesignerImportGenerator.cs
```diff
@@ -49,7 +49,7 @@ public void CreateImportMethods (IEnumerable<ITaskItem> libraries)
 						continue;
 					}
 					string alias = assemblyPath.GetMetadata ("Aliases");
-					bool hasAlias = !string.IsNullOrEmpty (alias);
+					bool hasAlias = !string.IsNullOrEmpty (alias) && !aliases.Contains ("global");
```
brendanzagaeski, line 52: I just noticed yesterday that Visual Studio includes `global` in the **Aliases** property input by default, so it's somewhat likely for projects to have `global` included even if they don't need it. With that in mind, I'm wondering if it would be best to handle the following scenario:

  * There are two assemblies that provide the same namespace, and both contain Android resources.

  * Each assembly includes `global` in the `%(Aliases)` along with a unique name, so for example:

    ```xml
    <Aliases>global,assembly1</Aliases>
    ```

    ```xml
    <Aliases>global,assembly2</Aliases>
    ```

  * Everywhere in hand-written C# files, the project is careful to use `extern alias assembly1` and `extern alias assembly2` and to prefix each type name or using statement with the appropriate alias.

  * With this `&& !aliases.Contains ("global")` change, I think the generated `Resource.designer.cs` file will not prefix the types from either assembly?

brendanzagaeski: Perhaps even more interesting is to add two more points to the scenario:

  * In addition to the conflicting namespaces, each assembly also contains its own namespace that does conflict with the other assembly.

  * Some of the hand-written C# files use those non-conflicting namespaces, and they don't use any `extern alias` statements because both namespaces are accessible directly under `global` without conflicts.

dellis1972: I think if there is a `global` in the Aliases Metadata provided by the user, we don't need to emit ANY `extern alias` values. since the normal `global` one will just work. 

If `global` is not present in that metadata, then global will NOT work, so we need the alias.

brendanzagaeski: Here's a test that covers the scenario I'm picturing. Removing `&& !aliasMetaData.Contains ("global")` from `ResourceDesignerImportGenerator.cs` and `GenerateResourceDesigner.cs` allows this test to pass:

```csharp
[Test]
public void CheckMultipleLibraryProjectReferenceAlias ([Values (true, false)] bool withGlobal)
{
	var path = Path.Combine (Root, "temp", TestName);
	var library1 = new XamarinAndroidLibraryProject () {
		ProjectName = "Library1",
	};
	var library2 = new XamarinAndroidLibraryProject () {
		ProjectName = "Library2",
		RootNamespace = "Library1"
	};
	var proj = new XamarinAndroidApplicationProject () {
		References = {
			new BuildItem.ProjectReference (Path.Combine("..", library1.ProjectName, Path.GetFileName (library1.ProjectFilePath)), "Library1") {
				Metadata = { { "Aliases", withGlobal ? "global,Lib1A,Lib1B" : "Lib1A,Lib1B" } },
			},
			new BuildItem.ProjectReference (Path.Combine("..", library2.ProjectName, Path.GetFileName (library2.ProjectFilePath)), "Library2") {
				Metadata = { { "Aliases", withGlobal ? "global,Lib2A,Lib2B" : "Lib2A,Lib2B" } },
			},
		},
	};
	using (var builder1 = CreateDllBuilder (Path.Combine (path, library1.ProjectName), cleanupAfterSuccessfulBuild: false, cleanupOnDispose: false)) {
		builder1.ThrowOnBuildFailure = false;
		Assert.IsTrue (builder1.Build (library1), "Library should have built.");
		using (var builder2 = CreateDllBuilder (Path.Combine (path, library2.ProjectName), cleanupAfterSuccessfulBuild: false, cleanupOnDispose: false)) {
			builder2.ThrowOnBuildFailure = false;
			Assert.IsTrue (builder2.Build (library2), "Library should have built.");
			using (var b = CreateApkBuilder (Path.Combine (path, proj.ProjectName), cleanupAfterSuccessfulBuild: false, cleanupOnDispose: false)) {
				b.ThrowOnBuildFailure = false;
				Assert.IsTrue (b.Build (proj), "Project should have built.");
				string[] text = File.ReadAllLines (Path.Combine (Root, path, proj.ProjectName, "Resources", "Resource.designer.cs"));
				Assert.IsTrue (text.Count (x => x.Contains ("Library1.Resource.String.library_name")) == 2, "library_name resource should be present exactly once for each library");
				Assert.IsTrue (text.Count (x => x == "extern alias Lib1A;" || x == "extern alias Lib1B;") == 1, "Exactly one extern alias should be present for each library.");
			}
		}
	}
}
```

dellis1972: True, however I was trying to make this work in a way that if `global` is present, then we don't need to emit ANY aliases in order for the designer file to compile. This is what the customer was expecting https://github.com/xamarin/xamarin-android/issues/4542#issuecomment-613553781. And it's how it used to work when we completely ignored Aliases. `global` means we don't need anything.

brendanzagaeski: Whoops. I think I might have obscured the key concern by including additional asserts. It's the step of building the app project that concerns me. That _fails_ right now because `Resource.designer.cs` contains exactly the kind of C# compiler conflict that aliases solve:

```
Build FAILED.

"C:\Source\xamarin-android\bin\TestDebug\temp\CheckMultipleLibraryProjectReferenceAliasTrue\UnnamedProject\UnnamedProject.csproj" (Build;SignAndroidPackage target) (1) ->
(CoreCompile target) -> 
  Resources\Resource.designer.cs(28,4,28,29): error CS0433: The type 'Resource' exists in both 'Library1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Library2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
  Resources\Resource.designer.cs(29,4,29,29): error CS0433: The type 'Resource' exists in both 'Library1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Library2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
  Resources\Resource.designer.cs(30,4,30,29): error CS0433: The type 'Resource' exists in both 'Library1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Library2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
  Resources\Resource.designer.cs(31,4,31,29): error CS0433: The type 'Resource' exists in both 'Library1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Library2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'

    0 Warning(s)
    4 Error(s)
```

Due to the following generated lines:

```cs
global::Library1.Resource.Drawable.Icon = global::UnnamedProject.Resource.Drawable.Icon;
global::Library1.Resource.String.library_name = global::UnnamedProject.Resource.String.library_name;
global::Library1.Resource.Drawable.Icon = global::UnnamedProject.Resource.Drawable.Icon;
global::Library1.Resource.String.library_name = global::UnnamedProject.Resource.String.library_name;
```

That is, even though both libraries have `global` in the list of aliases, `Resource.designer.cs` must _not_ use `global` for both libraries or there will be a conflict. It seems to me the simplest way to handle an arbitrary number of conflicting libraries (admittedly unusual, but might as well handle it if we can) is probably to use the first non-`global` alias for every libraries, which is the nice behavior this PR gives after removing `&& !aliasMetaData.Contains ("global")`.

dellis1972: So with the proposed change we will ALWAYS use the alias even if global is present. I guess the downside of that is users will see a change in their Resource.designer.cs when they upgrade. 

Shame we don't have any data on how often the RootNamespace is changed to match that of an identical library :(

brendanzagaeski: Oh yeah, that is an interesting trade-off to consider.

I suppose one little reassurance is that anyone who would get a change in `Resource.designer.cs` with this PR would already have a change in behavior in Xamarin.Android 10.2 (d16-5) and Xamarin.Android 10.3 (d16-6) due to the initial feature addition from <https://github.com/xamarin/xamarin-android/commit/d18c824c9e96dac59d75be3da2b46fecd18ade51>. So only users updating all at once from Xamarin.Android 10.1 (d16-4) or lower up to Xamarin.Android 10.4 (d16-7) or higher could benefit from changes in this PR. If there hasn't been much feedback about the change yet, then it's probably not too terribly common in user projects. (Though admittedly, there's a chance that projects that are slower to update Visual Studio versions might also be more likely to use `%(Aliases)` metadata.)

Maybe, like the reporter of #4542 was mentioning, an idea could be to provide an extra way to control how aliases are used in `Resource.designer.cs`. I wonder if any other .NET project type has a way to handle this. I think code-behind for XAML wouldn't have the issue since XAML files have to specify the CLR namespaces. I think `Resources.Designer.cs` files for `.resx` files wouldn't have to handle it either since there is only one assembly involved. Maybe ASP.NET or WCF has a scenario?

brendanzagaeski:
*Review update*

To help make it easy to skim to the latest status of the discussion on this PR at a glance, I'll add the results from running the [`CheckMultipleLibraryProjectReferenceAlias()`](https://github.com/xamarin/xamarin-android/pull/4554/#discussion_r410026051) test with the latest updated commit. The [build failure due to the conflicting types](https://github.com/xamarin/xamarin-android/pull/4554/#discussion_r410443666) remains:

```
Build FAILED.

"C:\Source\xamarin-android\bin\TestDebug\temp\CheckMultipleLibraryProjectReferenceAliasTrue\UnnamedProject\UnnamedProject.csproj" (Build;SignAndroidPackage target) (1) ->
(CoreCompile target) -> 
  Resources\Resource.designer.cs(28,4,28,29): error CS0433: The type 'Resource' exists in both 'Library1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Library2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
  Resources\Resource.designer.cs(29,4,29,29): error CS0433: The type 'Resource' exists in both 'Library1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Library2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
  Resources\Resource.designer.cs(30,4,30,29): error CS0433: The type 'Resource' exists in both 'Library1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Library2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'
  Resources\Resource.designer.cs(31,4,31,29): error CS0433: The type 'Resource' exists in both 'Library1, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null' and 'Library2, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'

    0 Warning(s)
    4 Error(s)
```

dellis1972:
@brendanzagaeski I reworked this and added your test. Can you re-review please.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
```diff
@@ -136,10 +136,17 @@ public override bool RunTask ()
...
+					string aliasMetaData = assembly.GetMetadata ("Aliases");
+					if (!string.IsNullOrEmpty (aliasMetaData)) {
+						foreach (var alias in aliasMetaData.Split (new [] {','}, StringSplitOptions.RemoveEmptyEntries)) {
+							string aliasName = alias.Trim ();
+							// dont emit an `extern alias global` as it is implicitly done.
+							if (string.Compare ("global", aliasName, StringComparison.OrdinalIgnoreCase) == 0)
+								continue;
+							// only add the first alias for each reference.
+							aliases.Add (aliasName);
+							break;
```
brendanzagaeski, line 148: No need to adjust this unless it sounds like a good idea. When I was skimming these lines quickly while rereading the PR, I realized my eye paused on this comment for a moment because `aliases.Add()` by itself could in theory add more than just 1 alias, so a tiny idea popped to mind that the comment might look nice moved one line down:

```
aliases.Add (aliasName);
// only add the first alias for each reference.
break;
```

src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
```diff
@@ -136,10 +136,17 @@ public override bool RunTask ()
...
 					assemblies.Add (item);
-					string alias = assembly.GetMetadata ("Aliases");
-					if (!string.IsNullOrEmpty (alias)) {
-						foreach (var a in alias.Split (new [] {','}, StringSplitOptions.RemoveEmptyEntries))
-							aliases.Add (a.Trim ());
+					string aliasMetaData = assembly.GetMetadata ("Aliases");
+					if (!string.IsNullOrEmpty (aliasMetaData)) {
+						foreach (var alias in aliasMetaData.Split (new [] {','}, StringSplitOptions.RemoveEmptyEntries)) {
+							string aliasName = alias.Trim ();
+							// dont emit an `extern alias global` as it is implicitly done.
```
brendanzagaeski, line 143: Optional tiny typo fix:

```diff
- // dont emit an `extern alias global` as it is implicitly done.
+ // don't emit an `extern alias global` as it is implicitly done.
```

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
```diff
@@ -37,6 +81,9 @@ public void CheckProjectReferenceAlias ()
 				using (var b = CreateApkBuilder (Path.Combine (path, proj.ProjectName), cleanupAfterSuccessfulBuild: false, cleanupOnDispose: false)) {
 					b.ThrowOnBuildFailure = false;
 					Assert.IsTrue (b.Build (proj), "Project should have built.");
+					string text = File.ReadAllText (Path.Combine (Root, path, proj.ProjectName, "Resources", "Resource.designer.cs"));
+					Assert.IsTrue (text.Contains ("extern alias Lib1"), "Resource.designer.cs should contain  extern alias Lib1.");
+					Assert.IsFalse (text.Contains ("extern alias Lib2"), "Resource.designer.cs should contain  extern alias Lib2.");
```
brendanzagaeski, line 86: I think the `Assert.IsFalse()` here might not be testing anything too interesting? I think the the test would need a `Build()` call for `library2.ProjectName` to create a project directory and files for `library2` so that the project reference from the application project would have files available to reference. Then the assert could be an `Assert.IsTrue()`.

Maybe it would be OK to clean this up by removing this test and adding the corresponding asserts into `CheckMultipleLibraryProjectReferenceAlias` instead?

Sketch of the change to `CheckMultipleLibraryProjectReferenceAlias()`:

```diff
Assert.IsTrue (b.Build (proj), "Project should have built.");
string[] text = File.ReadAllLines (Path.Combine (Root, path, proj.ProjectName, "Resources", "Resource.designer.cs"));
Assert.IsTrue (text.Count (x => x.Contains ("Library1.Resource.String.library_name")) == 2, "library_name resource should be present exactly once for each library");
-Assert.IsTrue (text.Count (x => x == "extern alias Lib1A;" || x == "extern alias Lib1B;") <= 1, "No more than one extern alias should be present for each library.");
+Assert.IsTrue (text.Contains ("extern alias Lib1A;"), "Resource.designer.cs should contain extern alias Lib1A.");
+Assert.IsFalse (text.Contains ("extern alias Lib1B;"), "No more than one extern alias should be present for each library.");
+Assert.IsTrue (text.Contains ("extern alias Lib2A;"), "Resource.designer.cs should contain extern alias Lib2A.");
+Assert.IsFalse (text.Contains ("extern alias Lib2B;"), "No more than one extern alias should be present for each library.");
```

dellis1972: done :)

src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
dellis1972: done :)

jonpryor:
In progress commit message:

```
Fixes: https://github.com/xamarin/xamarin-android/issues/4542

Our previous fix d18c824c did not take into account that users might
need to add `global` to their `%(ProjectReferece.Aliases)` metadata,
e.g.

	<ProjectReferece>
	  <Aliases>global,Lib1A,Lib1B</Aliases>
	</ProjectReferece>

This then causes the following error:

	Error CS1681: You cannot redefine the global extern alias (CS1681)

because our `Resource.designer.cs` generation emitted:

	extern alias global;

To fix this we need to ignore the `global` entry if one is present.
```

src/Xamarin.Android.Build.Tasks/Utilities/ResourceDesignerImportGenerator.cs
```diff
@@ -59,8 +59,14 @@ public void CreateImportMethods (IEnumerable<ITaskItem> libraries)
 						var declaringTypeName = $"{reader.GetString (declaringType.Namespace)}.{reader.GetString (declaringType.Name)}";
 						if (declaringTypeName == resourceDesignerName) {
 							if (hasAlias) {
-								string [] aliases = alias.Split (new [] {','}, StringSplitOptions.RemoveEmptyEntries);
-								declaringTypeName = $"{aliases[0].Trim ()}::{declaringTypeName}";
+								string [] aliases = aliasMetaData.Split (new [] {','}, StringSplitOptions.RemoveEmptyEntries);
+								foreach (var alias in aliases) {
+									string aliasName = alias.Trim ();
+									if (string.Compare ("global", aliasName, StringComparison.OrdinalIgnoreCase) == 0)
```
jonpryor, line 65: I'm not certain that `OrdinalIgnoreCase` is appropriate here, as C# is a case sensitive language.  I'm not sure it's "unreasonable" for someone to use `Global` as an alias.

…unless MSBuild metadata is *by design* case insensitive, in which case this is appropriate.  I don't think that's the case (?).

PR #4564, opened 2020-04-15
[.NET 5] exclude shared runtime APKs from Xamarin.Android.Sdk

jonathanpeppers:
Using `AndroidUseSharedRuntime=True` doesn't work in .NET 5 yet, and
it is still up in the air if we will support it.

For now, we can remove these from `Xamarin.Android.Sdk.nupkg` to save
on file size. This save ~164MB of file size on disk when the NuGet
package is extracted.

### DIFF
--- a/build-tools/installers/create-installers.targets
+++ b/build-tools/installers/create-installers.targets
@@ -376,12 +376,12 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\jar2xml.jar" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.AndroidTools.dll" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.AndroidTools.pdb" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-arm64-v8a.apk" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-armeabi-v7a.apk" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-debug.apk" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-debug.xml" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-x86.apk" />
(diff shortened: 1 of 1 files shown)


build-tools/installers/create-installers.targets
```diff
@@ -376,12 +376,12 @@
...
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-debug.xml" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-x86.apk" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-x86_64.apk" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-arm64-v8a.apk"   Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-armeabi-v7a.apk" Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-debug.apk"       Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-debug.xml"       Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-x86.apk"         Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Mono.Android.DebugRuntime-x86_64.apk"      Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
```
pjcollins, line 384: I've been thinking about a way to make this more elegant WRT Xamarin.Android.Sdk, the conditional inclusion here is definitely getting unwieldy - https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Sdk/Xamarin.Android.Sdk.proj#L37

I was considering using more ItemGroups with different names `_MSBuildFiles` vs `_MSBuildFilesLegacy` or something like that, but I think I like this condition better.

PR #4562, opened 2020-04-15
JetBrains OpenJDK 11

grendello:
This PR is a test to see what works and what breaks with JetBrains
OpenJDK 11 as far as Xamarin.Android is concerned.

PR #4567, opened 2020-04-16
JetBrains OpenJDK 11

grendello:
Moved from: https://github.com/xamarin/xamarin-android/pull/4562

This PR is a test to see what works and what breaks with JetBrains
OpenJDK 11 as far as Xamarin.Android is concerned.


jonpryor:
Various fixes required to allow the updated Java.Interop to build are also part of https://github.com/xamarin/xamarin-android/pull/4569.

PR #4569 should be merged before this.

Configuration.props
```diff
@@ -118,6 +118,10 @@
     <MingwCommandPrefix32>i686-w64-mingw32</MingwCommandPrefix32>
     <MingwCommandPrefix64>x86_64-w64-mingw32</MingwCommandPrefix64>
   </PropertyGroup>
+  <PropertyGroup>
+    <JavacSourceVersion>1.8</JavacSourceVersion>
+    <JavacTargetVersion>1.8</JavacTargetVersion>
```
jonathanpeppers, line 123: I think we still have support for JDK 1.6/1.7, if we merge this, do we need to require 1.8?

I think the relevant code is here:

https://github.com/xamarin/xamarin-android/blob/c39f1fe4ee43b72d41d3a99b520aa8fdabf9cfd5/src/Xamarin.Android.Build.Tasks/Tasks/Legacy/ValidateJavaVersion.cs#L48-L68

jonpryor:
Depends on: https://github.com/xamarin/monodroid/pull/1083

Configuration.props
jonpryor: Good question, I don't know the answers.

If we use `javac -target 1.8`, the resulting `.class` file requires Java 1.8.  However, is the code impacted by this PR impacted by that?  For example, `src/java-runtime` is only run on-device, not on the build machine, so requiring 1.8 for `src/java-runtime` shouldn't break any existing customers who have Java 1.6 installed.

`jar2xml.jar`, on the other hand, *may* be executed on customer machines.  (I wish that weren't the case, and I thought I saw a suggestion to obsolete & remove for .net5?). https://github.com/xamarin/jar2xml/commit/4125da6 and https://github.com/xamarin/monodroid/pull/1083 would *require* Java 1.8 in order to execute it.

What's the `javac -target` for `src/r8` and `src/manifestmerger`?  I don't see an explicit Language version there, which means we may *already* be requiring Java 1.8 be installed?  I don't know.

Finally, the `<ValidateJavaVersion/>` code you pointed to mentions `AndroidSdkBuildToolsVersion`; that's the Android SDK `build-tools` version, so this may need to be updated to check for build-tools 30 and require Java 11 there.

jonathanpeppers: For `r8` and `manifestmerger`, I bet it's doing this by default:
```
java {
    sourceCompatibility = JavaVersion.VERSION_1_8
    targetCompatibility = JavaVersion.VERSION_1_8
}
```
But we should probably put this in every `build.gradle` file explicitly.

jonpryor: If that "bet" is correct, then we're *already* in a scenario where customers *must* be using JDK 8 or they can't use `r8.jar`, which is required (I think?).  Meaning JDK 6/7 support is already gone.

jonpryor:
Looks like macOS is able to build, but [not Windows](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3655757&view=logs&j=cac0e8d3-0ef5-5d2b-b57e-e8fde7204df3&t=d0eaacf3-179f-5217-7519-cd03bc45e5e4):

```
  "C:\Users\dlab14\android-toolchain\sdk\build-tools\29.0.2\dx" --dex --no-strict --output="C:\A\vs2019xam00000H-1\_work\2\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\java_runtime.dex" "C:\A\vs2019xam00000H-1\_work\2\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\java_runtime.jar"
  
##[debug]Processed: ##vso[task.logdetail id=dffb25d4-cb0b-4e58-94d2-4b9a4f215bed;parentid=;name=EXEC;type=Build;starttime=2020-04-19T01:21:10.0332656Z;state=InProgress;]
##[error]EXEC(0,0): Error : No suitable Java found. In order to properly use the Android Developer
##[debug]Processed: ##vso[task.logissue type=Error;sourcepath=EXEC;linenumber=0;columnnumber=0;code=;]No suitable Java found. In order to properly use the Android Developer
##[debug]Processed: ##vso[task.logdetail id=dffb25d4-cb0b-4e58-94d2-4b9a4f215bed;parentid=;type=Build;result=Failed;finishtime=2020-04-19T01:21:10.0488457Z;progress=100;state=Completed;parentid=;name=;]
EXEC : error : No suitable Java found. In order to properly use the Android Developer [C:\A\vs2019xam00000H-1\_work\2\s\src\java-runtime\java-runtime.csproj]
  Tools, you need a suitable version of Java JDK installed on your system.
  We recommend that you install the JDK version of JavaSE, available here:
    http://www.oracle.com/technetwork/java/javase/downloads
  
  If you already have Java installed, you can define the JAVA_HOME environment
  variable in Control Panel / System / Avanced System Settings to point to the
  JDK folder.
  
  You can find the complete Android SDK requirements here:
    http://developer.android.com/sdk/requirements.html
  
##[error]src\java-runtime\java-runtime.targets(59,3): Error MSB3073: The command ""C:\Users\dlab14\android-toolchain\sdk\build-tools\29.0.2\dx" --dex --no-strict --output="C:\A\vs2019xam00000H-1\_work\2\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\java_runtime.dex" "C:\A\vs2019xam00000H-1\_work\2\s\bin\Release\lib\xamarin.android\xbuild\Xamarin\Android\java_runtime.jar"" exited with code -1.
```

jonpryor:
Discussed the [Windows failure](https://github.com/xamarin/xamarin-android/pull/4567#issuecomment-616008109), and the consensus was that we should `java-runtime.targets` to instead use `r8.jar`:

https://github.com/xamarin/xamarin-android/blob/ee3a0c582c82c91936020611e8081d644ec2fbaa/src/java-runtime/java-runtime.targets#L59-L62

jonpryor:
The [AllActivityAttributeProperties("manifestmerger.jar")](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3655757&view=ms.vss-test-web.build-test-results-tab&runId=12712476&resultId=100135&paneView=attachments) unit test fails:

```
error AMM0000: java.lang.UnsupportedClassVersionError: com/xamarin/manifestmerger/Main has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0 [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.lang.ClassLoader.defineClass1(Native Method) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.lang.ClassLoader.defineClass(ClassLoader.java:763) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.net.URLClassLoader.defineClass(URLClassLoader.java:468) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.net.URLClassLoader.access$100(URLClassLoader.java:74) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.net.URLClassLoader$1.run(URLClassLoader.java:369) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.net.URLClassLoader$1.run(URLClassLoader.java:363) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.security.AccessController.doPrivileged(Native Method) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.net.URLClassLoader.findClass(URLClassLoader.java:362) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.lang.ClassLoader.loadClass(ClassLoader.java:424) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at java.lang.ClassLoader.loadClass(ClassLoader.java:357) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: 	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:495) [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: Error: A JNI error has occurred, please check your installation and try again [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: Exception in thread "main"  [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
error AMM0000: java.lang.UnsupportedClassVersionError :  com/xamarin/manifestmerger/Main has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0 [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\AllActivityAttributePropertiesmanifestmerger.jar\UnnamedProject.csproj]
```

It appears that if our Gradle-based projects such as [`src/manifestmerger`](https://github.com/xamarin/xamarin-android/tree/master/src/manifestmerger) don't explicitly specify the Java target version, it uses the JDK version -- which make sense! -- and thus implicitly requires JDK 11 on the customer machines.

We'll need to update our Gradle projects accordingly.

src/java-runtime/java-runtime.targets
```diff
@@ -57,8 +57,12 @@
     Inputs="@(_RuntimeOutput->'%(Identity)')"
     Outputs="@(_RuntimeOutput->'%(OutputDex)')">
   <Exec
-      Command="&quot;$(AndroidSdkDirectory)\build-tools\$(XABuildToolsFolder)\dx&quot; --dex --no-strict --output=&quot;%(_RuntimeOutput.OutputDex)&quot; &quot;%(_RuntimeOutput.OutputJar)&quot;"
+      Command="&quot;$(JavaPath)&quot; -classpath &quot;$(OutputPath)\r8.jar&quot; com.android.tools.r8.D8 --release --no-desugaring --output &quot;%(_RuntimeOutput.IntermediateRuntimeOutputPath)&quot; &quot;%(_RuntimeOutput.OutputJar)&quot;"
```
jonathanpeppers, line 60: This looks like it should be correct to me 👍

Configuration.props
jonpryor: Given that using `manifestmerger.jar` *broke* when we didn't have `java.targetCompatibility` set, the bet *is* correct: JDK 8 is already a minimum requirement for all functionality.

jonathanpeppers:
☝️ comparing the current JDK we use on macOS:

     $JAVA_HOME/jre/lib/server/libjvm.dylib

Looking at the contents of the `.tar.gz` file, it is at:

    /jbrsdk/Contents/Home/lib/server/libjvm.dylib

Which seems like it's probably getting extracted to:

     $JAVA_HOME/lib/server/libjvm.dylib

And there isn't a `jre` directory?

jonpryor:
It looks like https://github.com/xamarin/monodroid/pull/1091 works, in that we're no longer getting XA5300 errors on every MSBuild Unit Test build.

Yay.

Instead, we're getting `javac` errors on every MSBuild Unit Test build:

```
Task "Javac" (TaskId:154)
  Task Parameter:JavaPlatformJarPath=/Users/runner/Library/Android/sdk/platforms/android-29/android.jar (TaskId:154)
  Task Parameter:ClassesOutputDirectory=obj/Release/android/bin/classes/ (TaskId:154)
  Task Parameter:ClassesZip=obj/Release/android/bin/classes.zip (TaskId:154)
  Task Parameter:StubSourceDirectory=obj/Release/android/src/ (TaskId:154)
  Task Parameter:ToolPath=/Users/runner/Library/Android/jdk/bin (TaskId:154)
  Task Parameter:
      Jars=
          /Library/Frameworks/Mono.framework/External/xbuild-frameworks/MonoAndroid/v10.0/mono.android.jar
          /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/java_runtime.jar (TaskId:154)
  /Users/runner/Library/Android/jdk/bin/javac -J-Dfile.encoding=UTF8 "@/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp2eba46d8.tmp"  (TaskId:154)
JAVAC : error : option --boot-class-path not allowed with target 11 [/Users/runner/runners/2.168.2/work/1/s/bin/TestRelease/temp/ProguardBOMError/UnnamedProject.csproj]
  The command exited with code 2. (TaskId:154)
```

Doh. :-(

jonpryor:
The `<Javac/>` error can be avoided by setting `$(JavacSourceVersion)` and `$(JavacTargetVersion)` to 1.8.

The problem is that if `javac -source` and `javac -target` aren't provided, they default to JDK 11, which doesn't allow providing `--boot-class-path`, which we need in order to use `android.jar` as the boot class path (and get `javac` to resolve e.g. `java.lang.String` from `android.jar`, so that you don't use JDK 11 methods in JDK 8).

Given that, what's the solution?  Update `Xamarin.Android.Common.targets` & co. to export those MSBuild properties?

jonpryor:
Fewer failures!  But still some failures:

[BuildHasNoWarnings(False,False,False,"aab")](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3716524&view=ms.vss-test-web.build-test-results-tab&runId=13136934&resultId=100121&paneView=attachments):

```
  /Users/runner/Library/Android/jdk/bin/java -Xmx1G -jar /Library/Frameworks/Xamarin.Android.framework/Libraries/xbuild/Xamarin/Android/bundletool.jar build-bundle --modules obj/Debug/android/bin/base.zip --output obj/Debug/android/bin/UnnamedProject.UnnamedProject.aab --config /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp2229842b.tmp  (TaskId:167)
BUILDAPPBUNDLE : warning : An illegal reflective access operation has occurred [/Users/runner/runners/2.168.2/work/1/s/bin/TestRelease/temp/BuildHasNoWarningsFalseFalseFalseaab/UnnamedProject.csproj]
BUILDAPPBUNDLE : warning : Illegal reflective access by com.google.protobuf.UnsafeUtil (file:/Library/Frameworks/Xamarin.Android.framework/Versions/10.4.100.28/lib/xamarin.android/xbuild/Xamarin/Android/bundletool.jar) to field java.nio.Buffer.address [/Users/runner/runners/2.168.2/work/1/s/bin/TestRelease/temp/BuildHasNoWarningsFalseFalseFalseaab/UnnamedProject.csproj]
BUILDAPPBUNDLE : warning : Please consider reporting this to the maintainers of com.google.protobuf.UnsafeUtil [/Users/runner/runners/2.168.2/work/1/s/bin/TestRelease/temp/BuildHasNoWarningsFalseFalseFalseaab/UnnamedProject.csproj]
BUILDAPPBUNDLE : warning : Use --illegal-access=warn to enable warnings of further illegal reflective access operations [/Users/runner/runners/2.168.2/work/1/s/bin/TestRelease/temp/BuildHasNoWarningsFalseFalseFalseaab/UnnamedProject.csproj]
BUILDAPPBUNDLE : warning : All illegal access operations will be denied in a future release [/Users/runner/runners/2.168.2/work/1/s/bin/TestRelease/temp/BuildHasNoWarningsFalseFalseFalseaab/UnnamedProject.csproj]
Done executing task "BuildAppBundle". (TaskId:167)
```

Is there a newer version of `bundletool` we should use?

[Xamarin.Android.Build.Tests.BuildTest.MultiDexAndCodeShrinker("proguard")](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3716524&view=ms.vss-test-web.build-test-results-tab&runId=13136934&resultId=100214&paneView=attachments):

```
  /Users/runner/Library/Android/jdk/bin/java -Djava.ext.dirs=/Users/runner/Library/Android/sdk/build-tools/29.0.2/lib com.android.multidex.MainDexListBuilder /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tnq6w4jr.sj0.jar '/Library/Frameworks/Mono.framework/External/xbuild-frameworks/MonoAndroid/v10.0/mono.android.jar':'/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/java_runtime.jar':'/Library/Frameworks/Xamarin.Android.framework/Libraries/xbuild/Xamarin/Android/android-support-multidex.jar':'obj/Release/android/bin/classes/../classes.zip'  (TaskId:159)
CREATEMULTIDEXMAINDEXCLASSLIST : error : Could not create the Java Virtual Machine. [/Users/runner/runners/2.168.2/work/1/s/bin/TestRelease/temp/MultiDexAndCodeShrinkerproguard/UnnamedProject.csproj]
CREATEMULTIDEXMAINDEXCLASSLIST : error : A fatal exception has occurred. Program will exit. [/Users/runner/runners/2.168.2/work/1/s/bin/TestRelease/temp/MultiDexAndCodeShrinkerproguard/UnnamedProject.csproj]
  -Djava.ext.dirs=/Users/runner/Library/Android/sdk/build-tools/29.0.2/lib is not supported.  Use -classpath instead. (TaskId:159)
  The command exited with code 1. (TaskId:159)
Done executing task "CreateMultiDexMainDexClassList" -- FAILED. (TaskId:159)
```

Many of the failures mirror this failure.

At a guess, Build-tools 29.0.2 doesn't support JDK 11, and we'll need to try Build-tools r30?

brendanzagaeski:
*Draft release notes*

No terrible hurry on this while this PR is still under active revision, but if I understand correctly, I think once this PR is eventually merged it will enable OpenJDK 11 compatibility for user builds? If that's correct, then it might be appropriate to add release note to `Documentation/release-notes/4567.md` as part of this PR.

Here's a very rough first draft of one possible idea:

```
### OpenJDK 11 build compatibility

The Xamarin.Android build tasks are now compatible with OpenJDK 11.

The primary motivation for this change is that the current preview Android SDK
Build-Tools version 30 rc4 that aligns with the Android 11 Developer Preview
requires JDK 9 or higher to run.

Project authors who are not yet targeting the Android 11 Developer Preview can
continue to use OpenJDK 8.
```

On the other hand, this PR itself might not need a release note, if for example the Visual Studio installer will be updated to provide OpenJDK 11 by default in the same release where OpenJDK 11 compatibility ships. In that case, the main topic in the release notes can be the new default OpenJDK version.

In either case, there will likely be some docs to update, such as: <https://docs.microsoft.com/xamarin/android/troubleshooting/questions/jdk9-errors>

jonpryor:
Now that things are [kinda/sorta building, -ish](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3783526&view=results), we hit The Question:

What does OpenJDK 11 support *look* like?

A key blocker in this current PR is that the Designer integration tests *fail*.  One of those reasons is that the Java code in the designer *will not build against JDK11*, as e.g. [`javax.annotation` was removed in JDK11](https://blog.codefx.org/java/java-11-migration-guide/).

Plus, the whole reason JDK11 is *required* in this PR is because `src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props.in` sets the default value of `$(AndroidSdkBuildToolsVersion)` to `30.0.0-rc4`.  Without that change, JDK8 would be usable.

It is also very unlikely that we'd *want* to have `$(AndroidSdkBuildToolsVersion)` default to 30, as (1) it's not stable, and (2) it'll likely only be required to use API-30, which also isn't yet stable.

---

Assuming the current PR is as green as it looks like it'll be, next step will be to "revert" the `$(AndroidSdkBuildToolsVersion)` change, and look into reverting the JDK install changes that we threw at the Designer.  That should allow *everything* to be "green enough" to merge.

---

Which brings us to "usage": to *use* JDK11, install JDK 11 and use Visual Studio to update the preferences so that JDK 11 will be used: https://docs.microsoft.com/en-us/xamarin/android/troubleshooting/questions/android-sdk-location?tabs=windows

This is a somewhat less interesting question.

The "real underlying question" is, if I want to use API-30/Android 11, what does that workflow look like?

  1. Set `$(TargetFrameworkVersion)`=v11.0
  2. Which will presumably likewise require setting `$(AndroidSdkBuildToolsVersion)` to 30.0.0, but this setting isn't in the IDE.  (Hm...)
  3. Which will fail to *package* unless JDK 11 is installed.
  4. We have no good story to actually install JDK11 that doesn't involve manual interaction by the customer.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -905,14 +910,23 @@ stages:
...
-        provisioning_script: $(System.DefaultWorkingDirectory)/bot-provisioning/dependencies.csx
+        provisioning_script: $(System.DefaultWorkingDirectory)/designer/bot-provisioning/dependencies.csx
         provisioning_extra_args: -remove Xamarin.Android -vv DEVDIV_PKGS_NUGET_TOKEN=$(DevDiv.NuGet.Token) SECTOOLS_PKGS_NUGET_TOKEN=$(SecTools.NuGet.Token)
 
-    - template: yaml-templates/run-installer.yaml
+    - template: yaml-templates/setup-test-environment.yaml
+      parameters:
+        xaSourcePath: $(System.DefaultWorkingDirectory)/xamarin-android
+
+    - script: echo "##vso[task.setvariable variable=JavaSdkDirectory]$HOME/android-toolchain/$($(XA.Jdk11.Folder)"
```
pjcollins, line 920: There are some extra characters in this line that need to be removed.

jonpryor:
Current build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3791298&view=results

jonpryor:
Squash-and-merge summary:

```
[ci] Preliminary JDK11 Support (#4567)
```

Body:

```
Context: https://issuetracker.google.com/issues/150189789
Context: https://developer.android.com/preview/features#signature-scheme-v4
Context: https://developer.android.com/preview/features#incremental

Preliminary JDK 11 support.

The Android SDK Build-tools r30.0.0-rc4 package currently requires
JDK 11 in order to use `apksigner.jar`:

	Task "AndroidApkSigner"
	  Task Parameter:ApkSignerJar=C:\Users\dlab14\android-toolchain\sdk\build-tools\30.0.0-rc4\lib\apksigner.jar
	  …
	  Task Parameter:ToolPath=C:\Program Files\Android\jdk\microsoft_dist_openjdk_1.8.0.25\bin
	  Task Parameter:ManifestFile=obj\Release\android\AndroidManifest.xml
	  C:\Program Files\Android\jdk\microsoft_dist_openjdk_1.8.0.25\bin\java.exe -jar C:\Users\dlab14\android-toolchain\sdk\build-tools\30.0.0-rc4\lib\apksigner.jar sign --ks "C:\Users\dlab14\AppData\Local\Xamarin\Mono for Android\debug.keystore" --ks-pass pass:android --ks-key-alias androiddebugkey --key-pass pass:android --min-sdk-version 21 --max-sdk-version 29  "C:\A\vs2019xam00000Y-1\_work\1\s\bin\TestRelease\temp\BuildAotApplication AndÜmläüts_x86_64_True_True\bin\Release\UnnamedProject.UnnamedProject-Signed.apk"
	  java.lang.UnsupportedClassVersionError: com/android/apksigner/ApkSignerTool has been compiled by a more recent version of the Java Runtime (class file version 53.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	  at java.lang.ClassLoader.defineClass1(Native Method)
	  …
	…\Xamarin.Android.Common.targets(2559,2): error MSB6006: "java.exe" exited with code 1.

This could be triggered by using the Android SDK Manager to install
Build-tools r30.0.0-rc4, then overriding the
`$(AndroidSdkBuildToolsVersion)` MSBuild property in `App.csproj`:

	<PropertyGroup>
	  <AndroidSdkBuildToolsVersion>30.0.0-rc4</AndroidSdkBuildToolsVersion>
	</PropertyGroup>

Many previous commits have added support for JDK 11, e.g. d99facb1,
dcee2c84, 4742d50d, 89f3dc15, and 895b7bc5 (among many others).

Additionally, c8ab4555 updated the build environment so that *both*
JetBrains OpenJDK 1.8 and 11 were installed, though 1.8 was still used.

Now it's time to "flip the build environment," and use JDK 11 to build
the product, installers, and run (most) unit tests:

  * Update `azure-pipelines.yaml` and `azure-pipelines-oss.yaml` so
    that JDK 11 is used to build everything.

  * Update `xaprepare` so that when `$(JavaSdkDirectory)` isn't
    overridden, JDK 11 is used by default.

  * Override the `JI_JAVA_HOME` environment variable when preparing
    `external/Java.Interop` so that it uses the desired JDK.

  * Update the default `$(LatestSupportedJavaVersion)` value to
    11.0.4, the version of the JDK we install.

  * Update the `<ValidateJavaVersions/>` task so that JDK 11 is
    required when `$(AndroidSdkBuildToolsVersion)` is >= r30.
    This will cause an XA0032 error if JDK 1.8 is used to build a
    project when `$(AndroidSdkBuildToolsVersion)` is >= r30.

  * `apksigner` from Build-tools r30 creates a new `.idsig` file
    next to the `.apk`.  This is to enable a new
    `adb install --incremental` feature.  Add the `.idsig` file
    to `@(FileWrites)`.

  * Update various unit test `.csproj` files to consistently
    `<Import/>` the `Configuration.props` file and use properties
    within it for default values.  This in particular allowed building
    e.g. `samples/HelloWorld` to implicitly use JDK 11.

  * Increase "expected" times in
    `tests/msbuild-times-reference/MSBuildDeviceIntegration.csv`
    as JDK 11 appears to be slower than JDK 1.8 (🙁).

  * *Disable* support for `@(JavaSourceJar)` when JDK 11 is used.
    JDK 11 changes the HTML generated by the `javadoc` command, and
    we need to update our HTML parser to handle it.  In the interest
    of expediency, disable this for now.

The Android Designer integration tests continue to use JDK 1.8,
as the Designer doesn't currently build under JDK 11.

TODO:

  * Fix `@(JavaSourceJar)` support
  * Allow Android Designer to work with JDK 11, run Designer
    integration tests using JDK 11.
  * Figure out how to not accidentally bitrot JDK 1.8 support.

Co-authored-by: Jonathan Pryor <jonpryor@vt.edu>
Co-authored-by: Jonathan Pobst <jonathan.pobst@microsoft.com>
Co-authored-by: Peter Collins <pecolli@microsoft.com>
```

Documentation/release-notes/4567.md
```diff
@@ -0,0 +1,18 @@
+### Preliminary JDK 11 Support
+
+Xamarin.Android can now use JDK 11 to build libraries and applications.
+
+To do so:
+
+ 1. Install JDK 11
+ 2. Set the `$(JavaSdkDirectory)` MSBuild property to refer to the root
+    of the JDK installation from (1).
```
brendanzagaeski, line 9: Can project authors configure the JDK via the Java Development Kit location setting in Visual Studio options (for example, on Windows: **Tools > Options > Xamarin > Android Settings > Java Development Kit Location**)? If so, I think it could be good to mention that approach first.

Then, if a different approach might work best on CI systems, such as Azure Pipelines, that could be included as a second option.

Documentation/release-notes/4567.md
```diff
@@ -0,0 +1,18 @@
...
+
+#### Known Issues
+
+When JDK 11 is used, the `@(JavaDocJar)` Build action is not supported,
```
brendanzagaeski, line 13: It could be handy to create a GitHub issue for this known issue so that it will already have a tracking location, but it's OK to skip that for now too.

Documentation/release-notes/4567.md
```diff
@@ -0,0 +1,18 @@
...
+and binding projects will not parse Javadoc to determine parameter names.
+This may a change in parameter names.
```
brendanzagaeski, line 15: It looks like there's a word or two missing here unintentionally, probably just due to rearrangements in editing. Maybe this can be something like:

```
This may result in different parameter names for existing bindings projects.
Authors of bindings projects affected by this change are recommended to continue
to use JDK 8 to build the bindings projects until this issue is resolved.
```

The additional candidate sentence also helps reinforce the point that this issue is primarily relevant to bindings projects authors and doesn't affect app project authors who are only consuming bindings provided by the Xamarin team.

Documentation/release-notes/4567.md
```diff
@@ -0,0 +1,18 @@
+### Preliminary JDK 11 Support
+
+Xamarin.Android can now use JDK 11 to build libraries and applications.
```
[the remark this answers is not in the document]
jonpryor: > Visual Studio will start installing the JetBrains Runtime OpenJDK 11 by default in the same release that contains this Xamarin.Android change.

Almost certainly *not*, not for d16-7 P3 anyway, and I *really* doubt the likelihood of d16-7 *period*.  For starters, it's only *needed* if Build-tools r30 is used, and --at present! -- it's *not* required to target API-R/Android 11.

(Even if it *were* required, I rather doubt the number of people who will be *using* Android 11 in the preview this is released in.)

More importantly, the Android Designer doesn't currently work with JDK 11.  We *can't* change the default, even if we wanted to.

Additionally, I don't know what the *rest* of the VS "ecosystem" looks like, and whether it can be updated to JDK 11.

Finally, let's take a gander at API-29, shall we?  That didn't become the *default* `$(TargetFrameworkVersion)` until *d16-6*, ~9 *months* after a Xamarin.Android release containing a stable Android 10.0 binding.

I think there is *no chance at all* of migrating VS to JDK 11 "soon."*

*Definition of "soon" is deliberately undefined.''

> If it's not yet known when Visual Studio will start installing OpenJDK 11 by default…

Will do, though I think it'll be "moot."  One of the (random) ideas is that we'll *delete* the `Documentation/release-notes` contents when a release note has gone stable.  My guess is that XA d16-7 will be stable *long before* VS adds JDK 11, at which point this release note *will likely not exist*.

> If it is already known that Visual Studio will *not* install OpenJDK 11 by default

99.99% certain of this.

> it could be good to add a sentence about when users might want to go to the trouble of finding and installing JDK 11 themselves.

I'm not sure *I* can write this, not quickly.  The only reason I know of to want/need JDK11 support is for Android SDK Build-tools r30+, but *why* would anybody want to use that, *especially* considering that it's still a "release candidate"?  *I don't know* why this would be desirable.

I like the blurb you provided.

Documentation/release-notes/4567.md
jonpryor: Looks like these locations have documentation for how to set that value:

  * https://docs.microsoft.com/en-us/xamarin/android/troubleshooting/questions/update-jdk?tabs=windows
  * https://docs.microsoft.com/en-us/xamarin/android/get-started/installation/windows#configuration

PR #4583, opened 2020-04-20
[Mono.Android] Hide "internal" members of JLO from IDEs.

### LINKED ISSUE #4582: Hide Java.Lang.Object members meant only for internal use
Hi, this is not a very important issue, but something I guess would make a dev's life a bit easier :)

I work with Xamarin.Android so I use mostly objects that are bound from Java libs so they inherit from the `Java.Lang.Object` class. The _problem_ here is that it has a few public members that are intended for internal Java.Interop use only. For instance `public IntPtr Handle` or `public JniObjectReference PeerReference`.

So I'm wondering whether it'd be possible to hide these members as they're making an unwanted noise in IntelliSense and could also possibly lead to some confusion or errors when mistakenly used. The best would of course be to make them internal or something like that or, if that's not possible, the `EditorBrowsable(EditorBrowsableState.Never)` attribute could be added to remove that noise.

jpobst:
Fixes #4582.

`Java.Lang.Object` contains various `public` and `protected` members that are required by our bindings, such as `J.L.Object.Handle` and `J.L.Object.PeerReference`.  We cannot make them `internal` because our bindings need them, but we can make them "invisible" to users using IDEs by applying `[EditorBrowsable (Never)]` and `[DebuggerBrowsable (Never)]`.  This will prevent them from showing up in Intellisense and debugger displays, as they should not be used by users.

Note they are still usable if needed, they simply do not show up in Intellisense.  You have to type them out manually.

Intellisense for a class deriving JLO before:
before

After:
after


src/Mono.Android/Java.Lang/Object.cs
```diff
@@ -91,21 +93,29 @@ internal void SetHandleOnDeserialized (StreamingContext context)
...
+		[DebuggerBrowsable (DebuggerBrowsableState.Never)]
+		[EditorBrowsable (EditorBrowsableState.Never)]
 		public virtual JniPeerMembers JniPeerMembers {
 			get { return _members; }
 		}
 #endif  // JAVA_INTEROP
 
+		[DebuggerBrowsable (DebuggerBrowsableState.Never)]
+		[EditorBrowsable (EditorBrowsableState.Never)]
 		public IntPtr Handle {
```
jonathanpeppers, line 119: Would this also hide usage of `Handle` like:
```csharp
private Java.Lang.Object foo;
//...
if (foo != null && foo.Handle != IntPtr.Zero) {
   //...
}
```
If so, maybe we shouldn't hide `Handle` as developers might often use it?

Xamarin.Forms uses `Handle` in these extension methods a lot: 

https://github.com/xamarin/Xamarin.Forms/blob/954cc8b0f174f179fd3395b4b1f257201f9701af/Xamarin.Forms.Platform.Android/Extensions/JavaObjectExtensions.cs#L7-L31

jpobst: Yes, it would hide `Handle` in that case (from Intellisense).  My gut says users wouldn't be using 'Handle' if they are writing applications or libraries.  (The authors of XF itself are a special case.)

But that's just an assumption and it's certainly up for discussion.

jonathanpeppers: You also might use `Handle` if you are working with JNI directly? Some folks might do that?

src/Mono.Android/Java.Lang/Object.cs
```diff
@@ -91,21 +93,29 @@ internal void SetHandleOnDeserialized (StreamingContext context)
 		}
 
 #if JAVA_INTEROP
+		[DebuggerBrowsable (DebuggerBrowsableState.Never)]
+		[EditorBrowsable (EditorBrowsableState.Never)]
 		public int JniIdentityHashCode {
```
jonpryor, line 98: I would think that when debugging you *would* want *something* to designate "object identity", e.g. to easily answer "are these two variables referring to the same instance"?

Is this (still?) a desirable/useful feature?  If so, what property would best be expose to the debugger?  `JniIdentityHashCode` is useful if there are *separate `Java.Lang.Object` instances wrapping the same Java instance, while `Handle` and `PeerReference` are good for seeing distinct wrappers.

jonpryor:
We should give `Java.Lang.Throwable` the same treatment.

src/Mono.Android/Java.Lang/Object.cs
```diff
@@ -91,21 +93,29 @@ internal void SetHandleOnDeserialized (StreamingContext context)
 		}
 
 #if JAVA_INTEROP
+		[DebuggerBrowsable (DebuggerBrowsableState.Never)]
```
jonpryor, line 96: This property should be shown to the debugger.

src/Mono.Android/Java.Lang/Object.cs
```diff
@@ -91,21 +93,29 @@ internal void SetHandleOnDeserialized (StreamingContext context)
...
 		}
 
+		[DebuggerBrowsable (DebuggerBrowsableState.Never)]
+		[EditorBrowsable (EditorBrowsableState.Never)]
 		public virtual JniPeerMembers JniPeerMembers {
 			get { return _members; }
 		}
 #endif  // JAVA_INTEROP
 
+		[DebuggerBrowsable (DebuggerBrowsableState.Never)]
```
jonpryor, line 117: This property should be shown to the debugger.

bramborman:
Just remembered about the `Android.Runtime.IJavaObject` iface which could get the same treatment too

src/Mono.Android/Java.Lang/Object.cs
bramborman: Agree, but just wondering whether that couldn't be implemented in VS debugger [the same way it is for .NET objects](https://stackoverflow.com/a/4251470)? This way we wouldn't have to check complex ids and hashes and it would feel more _native_ to devs imo.

jpobst:
Addressed feedback to let the debugger see a couple of instance identifiers, and added `Java.Lang.Throwable`.

The only member on `IJavaObject` is `Handle`. Presumably if you've casted your object to an `IJavaObject` it's probably ok to see `Handle` since there's no other reason to cast to `IJavaObject`.

src/Mono.Android/Java.Lang/Object.cs
jonpryor: I didn't know that was a feature that existed.

That said, I assume that **Make Object ID** only works for identical instances.  If you have two different C# instances which wrap the same Java instance -- presumably a rare but still possible scenario -- then Make Object ID wouldn't help.  `JniIdentityHashCode` *could*.

bramborman: Well that's not exactly what I mean... I mean if there could be added a new similar functionality, probably another context menu option specialized for this usecase, something like _Make Java object ID_  that wouldn't care about the .NET instance but about the native Java instance probably using the mentioned `JniIdentityHashCode` property. That would make looking for Java instance equality much easier compared to writing down some hash codes or so. I've heard something that in .NET 5 there should me a more native interoperability with Java (and I believe that Java.Interop is the thing to be used for that?) so if that's true, this change would benefit not only Xamarin.Android users.

jpobst: I suspect the primary reason .NET has `Make Object ID` is because .NET objects do not inherently have an ID and thus you cannot tell if 2 objects are the same.  We already have that information easily available with these fields.  In essence, they had to build a workaround for something we already have.

Adding `Make Java object ID` for both VS2019 and VSMac would be a decent amount of work for a different team (Android IDE team).  I suspect it will be placed in their backlog at much lower priority than the other tasks they have and probably never implemented.

As such, let's keep these properties visible to the debugger.  Overall this is still a good win by cleaning up the IntelliSense dropdown and some of the Debugger window.

bramborman: You're right, the effort wouldn't be worth the outcome especially since it's imo something devs don't do really often.

PR #4584, opened 2020-04-20
[build] Use API-21 for all native builds

grendello:
Up until this commit, we have been building the 64-bit targets (`x86-64`
and `armv8-a` against NDK API 21 while the 32-bit targets (`x86` and
`armv7-a`) against NDK API 16.  This commit changes it so that all
targets build against API 21.  This gives us the ability to speed up
Debug builds on 32-bit platforms via use of  POSIX APIs for directory
traversal that are available starting from API 21 onwards (`openat` and
`fstatat`).  The switch to API 21 was initially planned for .NET5 but
there's no harm in making the change now.

In addition, this commit updates versions of several components:

  * Bump emulator version to 30.0.5
  * Bump NDK version to 21b (21.1.6352462)
  * Bump x86 system image revision to 7

Also, `XABuildConfig.NDKMinimumApiAvailable` is now set to the minimum
supported NDK API level instead of the minimum API level *installed* by
the NDK.


brendanzagaeski:
*Questions* (based on previous similar change in Xamarin.Android 9.3)

Will the changes in this PR also automatically update the warning message for XA4216? It looks like an equivalent question is: Will the changes automatically update the `NDKMinimumApiAvailable` output from `Xamarin.Android.BuildTools.PrepTasks.NDKInfo`?

*Draft release note*

Here's a candidate release note for this change based on the release note for [a previous update of the minimum compatible API level](https://docs.microsoft.com/xamarin/android/release-notes/9/9.3#android-ndk-r19-compatibility). Feel free to suggest changes or even a whole new wording.

````
### Minimum compatible Android version now Android 5.0 Lollipop (API level 21)

The minimum compatible Android version for Xamarin.Android apps is now Android
5.0 Lollipop (API level 21).  Warning XA4216 has been updated to help highlight
this change for any projects that might still have an older version set for the
`minSdkVersion` in the `AndroidManifest.xml` file:

```
warning XA4216: AndroidManifest.xml //uses-sdk/@android:minSdkVersion '20' is less than API-21, this configuration is not supported.
```

#### Background information

The unmanaged native libraries that are part of Xamarin.Android are now built to
target Android 5.0 Lollipop (API level 21), so apps built using this release can
abort due to incompatible APIs if run on earlier Android versions.
````

**Comment by grendello** [comment]
> _Questions_ (based on previous similar change in Xamarin.Android 9.3)
> 
> Will the changes in this PR also automatically update the warning message for XA4216? It looks like an equivalent question is: Will the changes automatically update the `NDKMinimumApiAvailable` output from `Xamarin.Android.BuildTools.PrepTasks.NDKInfo`?
> 

The warning message will be updated automatically, yes - it uses a file generated by `xaprepare` which stores the minimum values. This PR is irrelevant for `NKInfo` which gathers information about the installed NDK - i.e. what's available on the system.

File: build-tools/xaprepare/xaprepare/ConfigAndData/Dependencies/AndroidToolchain.cs [hunk 1]
```diff
@@ -38,13 +38,16 @@ public AndroidToolchain ()
 				new AndroidPlatformComponent ("platform-25_r03",   apiLevel: "25", pkgRevision: "3"),
 				new AndroidPlatformComponent ("platform-26_r02",   apiLevel: "26", pkgRevision: "2"),
 				new AndroidPlatformComponent ("platform-27_r03",   apiLevel: "27", pkgRevision: "3"),
-				new AndroidPlatformComponent ("platform-28_r04",   apiLevel: "28", pkgRevision: "4"),
-				new AndroidPlatformComponent ("platform-29_r01",   apiLevel: "29", pkgRevision: "1"),
-				new AndroidPlatformComponent ("platform-R_r02",    apiLevel: "R",  pkgRevision: "2"),
+				new AndroidPlatformComponent ("platform-28_r06",   apiLevel: "28", pkgRevision: "6"),
+				new AndroidPlatformComponent ("platform-29_r04",   apiLevel: "29", pkgRevision: "4"),
+				new AndroidPlatformComponent ("platform-R_r03",    apiLevel: "R",  pkgRevision: "3"),
```
jonpryor, line 43: Let's *not* update API-R to Developer Preview 3, please.  This should be done separately, as there's no guarantee that it's API compatible with DP2.

PR #4586, opened 2020-04-20
[.NET 5] map $(RuntimeIdentifier) to $(AndroidSupportedAbis)

jonathanpeppers:
Context: https://docs.microsoft.com/dotnet/core/rid-catalog

.NET Core has a concept of `$(RuntimeIdentifier)` (or RID), which
means you would `publish` an `arm64-v8a` build for Android via:

    dotnet publish foo.csproj --runtime-identifier android-arm64 --self-contained

To make this work with our existing MSBuild targets we will need to
map RIDs to `$(AndroidSupportedAbis)`. I made a helper method that
looks for the second `-arm64` portion of the RID.

There is still work ongoing to support multiple RIDs in a single build
in .NET 5. We will need to rework things when this is available.

I updated the .NET 5 MSBuild tests to validate multiple RIDs.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
@@ -256,17 +256,6 @@ namespace Xamarin.Android.Tasks
 			"x86_64",
 		};
 
-		static readonly Dictionary<string, string> RidAbiMap = new Dictionary<string, string> {
-			{ "android.21-arm64", "arm64-v8a" },
-			{ "android.21-arm", "armeabi-v7a" },
-			{ "android.21-x86", "x86" },
-			{ "android.21-x64", "x86_64" },
-			{ "android-arm64", "arm64-v8a" },
-			{ "android-arm", "armeabi-v7a" },
-			{ "android-x86", "x86" },
-			{ "android-x64", "x86_64" },
-		};
-
 		public static string GetNativeLibraryAbi (string lib)
 		{
 			// The topmost directory the .so file is contained within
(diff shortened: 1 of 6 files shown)


src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
```diff
@@ -692,5 +679,38 @@ public static string FixUpAndroidResourcePath (string file, string resourceDirec
...
+				return "arm64-v8a";
+			}
+			if (runtimeIdentifier.EndsWith ("-arm", StringComparison.OrdinalIgnoreCase)) {
+				return "armeabi-v7a";
+			}
+			if (runtimeIdentifier.EndsWith ("-x86", StringComparison.OrdinalIgnoreCase)) {
+				return "x86";
+			}
+			if (runtimeIdentifier.EndsWith ("-x64", StringComparison.OrdinalIgnoreCase)) {
+				return "x64";
```
jonpryor, line 711: Shouldn't this be `x86_64`, as per the comment on line 694?

jonathanpeppers: Yeah, this is wrong 🤦‍♂️ . I updated the test so it will try every architecture & RID.

src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
```diff
@@ -692,5 +679,38 @@ public static string FixUpAndroidResourcePath (string file, string resourceDirec
...
+				return "armeabi-v7a";
+			}
+			if (runtimeIdentifier.EndsWith ("-x86", StringComparison.OrdinalIgnoreCase)) {
+				return "x86";
+			}
+			if (runtimeIdentifier.EndsWith ("-x64", StringComparison.OrdinalIgnoreCase)) {
+				return "x86_64";
+			}
+			return "";
+		}
```
Nirmal4G, line 714: Just a suggestion:

Use an `androidAbiGraph.json` to define the mappings just like `runtimeIdentifierGraph.json` in dotnet sdk.

If it's 1:1 mapping, then, no need for that complexity.

Issue #4550, opened 2020-04-13
Application crash from finalizer in AndroidHttpHandler when disposing


dotMorten:
I appears to me that the `AndroidClientHandler` does not appreciate being disposed while a request is in progress / getting cancelled. This can lead to application crashes.

### Steps to Reproduce

Run the following code and see the unobserved task exception is hit (even though it's being observed in the continuation method:

```cs
        private async void TestRequests()
        {
            TaskScheduler.UnobservedTaskException += (object sender, UnobservedTaskExceptionEventArgs args) =>
            {
                // App will crash here if we don't handle it
                args.SetObserved();
                Debugger.Break();
            };
            for (int i = 0; i < 100; i++)
            {
                //using (var handler = new System.Net.Http.HttpClientHandler()) // Doesn't crash
                using (var handler = new AndroidClientHandler()) 
                {
                    using (HttpClient client = new HttpClient(handler))
                    {
                        using (var request = new HttpRequestMessage(HttpMethod.Get, "http://sampleserver3.arcgisonline.com/ArcGIS/rest/services/Earthquakes/EarthquakesFromLastSevenDays/MapServer/0/query?text=&geometry=&geometryType=esriGeometryPoint&inSR=&spatialRel=esriSpatialRelIntersects&relationParam=&objectIds=&where=1%3D1&time=&returnCountOnly=false&returnIdsOnly=false&returnGeometry=true&maxAllowableOffset=&outSR=&outFields=&f=pjson&t=" + DateTime.Now.Ticks.ToString()))
                        {
                            CancellationTokenSource tcs = new CancellationTokenSource();
                            tcs.CancelAfter(100);
                            try
                            {
                                _ = client.SendAsync(request, tcs.Token).ContinueWith(t =>
                                {
                                });
                            }
                            catch { }
                        }
                    }
                }
                GC.Collect();
                GC.WaitForPendingFinalizers();
            }
        }
```

This'll throw
> A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (Cannot access a disposed object.
> Object name: 'AndroidClientHandler'.)
> StackTrace
```   at Xamarin.Android.Net.AndroidClientHandler.AssertSelf () [0x00009] in <4ccdb3137d974856b786e1aeebbfbab6>:0 
[...]

PR #4589, opened 2020-04-21
[AndroidClientHandler] Observe cancellation requests more

grendello:
Context: https://github.com/xamarin/xamarin-android/issues/4550

It is possible that, in certain situations, a cancellation request
arriving from code external to `AndroidClientHandler` will not be
satisfied while starting the task to acquire connection status, because
the cancellation request appears after handler already established
connection to the remote host but before the request status is obtained.

Failure to check that the cancellation was requested may result in the
status retrieval task starting only to run into the issue that the
`AndroidClientHandler` object itself is already disposed (since the
cancellation request has been applied) and throw the
`ObjectDisposedException` instead of cancelling the status request
quietly.  Note that this is likely happen only in situations when the
application code aggressively cancels and collects (via calls to the
`GC` class) `AndroidClientHandler` instances or in other situations
which induce potential for timing issues in the TPL or `async/await`
state machine, thus leading to possible race conditions there.

This commit passes cancellation token to the task which runs the status
request, thus preventing it from starting in the situation described
above.  The possibility of a race condition still remains, but is
seriously reduced.

### DIFF
--- a/Documentation/release-notes/4589.md
+++ b/Documentation/release-notes/4589.md
@@ -1,0 +1,7 @@
+#### Application behavior on device and emulator
+
+  * [GitHub PR 4582](https://github.com/xamarin/xamarin-android/pull/4589):
+    `ObjectDisposedException` could in theory be thrown during cancellation of
+    `AndroidClientHandler` requests, depending on the particular timing of
+    cancellation and object disposal.  The time window where this can happen 
+    is now narrower.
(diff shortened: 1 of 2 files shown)


dotMorten:
This didn't fix it for me. See details: https://github.com/xamarin/xamarin-android/issues/4550#issuecomment-617259064

dotMorten:
This didn't fix it for me. See details: https://github.com/xamarin/xamarin-android/issues/4550#issuecomment-617259064

dotMorten:
I'm not sure this actually makes it narrower. During GC I consistently found this change had no effect what so ever, and I was only able to reproduce this issue during GC.
Apart from that it seems like a safe change, I'm not sure this PR is really worth it, as there's no known test case that actually proves this PR has any effect.

jonpryor:
> whether this change could be considered an improvement in following Task idioms even if it doesn't correspond to a known test case

*I* will do so.

PR #4598, opened 2020-04-21
[.NET 5] fix warning for $(TFV) of 5.0

jonathanpeppers:
In our current .NET 5 builds, you always hit:

    warning XA1008: The TargetFrameworkVersion (Android API level 21) is lower than the targetSdkVersion (29).
    Please increase the `$(TargetFrameworkVersion)` or decrease the `android:targetSdkVersion` in the `AndroidManifest.xml` so that the API levels match.

It is giving a warning for Google Play, because
`$(TargetFrameworkVersion)` is equal to 5.0.

To fix this, I modified the `<CheckGoogleSdkRequirements/>` MSBuild
task so that `TargetFrameworkVersion` is optional. For .NET 5 builds,
we can pass in `$(_AndroidApiLevel)` instead.

This way we will still get some important warnings for .NET 5 such as
`targetSdkVersion` or `minSdkVersion` that need to be changed in the
`AndroidManifest.xml` file.

I also fixed up `LastBuildOutput`, we needed the value for .NET 5
MSBuild tests and I removed unnecessary usage of `yield return`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/CheckGoogleSdkRequirements.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/CheckGoogleSdkRequirements.cs
@@ -14,8 +14,15 @@ namespace Xamarin.Android.Tasks
 	{
 		public override string TaskPrefix => "CGS";
 
-		[Required]
+		/// <summary>
+		/// This will be blank for .NET 5 builds
+		/// </summary>
 		public string TargetFrameworkVersion { get; set; }
+
+		/// <summary>
+		/// This is used instead of TargetFrameworkVersion for .NET 5 builds
+		/// </summary>
+		public int ApiLevel { get; set; }
 
 		[Required]
 		public string ManifestFile { get; set; }
@@ -24,7 +31,9 @@ namespace Xamarin.Android.Tasks
 		{
 			ManifestDocument manifest = new ManifestDocument (ManifestFile);
 
-			var compileSdk = MonoAndroidHelper.SupportedVersions.GetApiLevelFromFrameworkVersion (TargetFrameworkVersion);
+			var compileSdk = string.IsNullOrEmpty (TargetFrameworkVersion) ?
+				ApiLevel :
+				MonoAndroidHelper.SupportedVersions.GetApiLevelFromFrameworkVersion (TargetFrameworkVersion);
 
 			if (!int.TryParse (manifest.GetMinimumSdk (), out int minSdk)) {
 				minSdk = 1;
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Tooling.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Tooling.targets
@@ -11,9 +11,10 @@ called for "legacy" projects in Xamarin.Android.Legacy.targets.
 
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <UsingTask TaskName="Xamarin.Android.Tasks.ValidateJavaVersion" AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
-  <UsingTask TaskName="Xamarin.Android.Tasks.ResolveAndroidTooling" AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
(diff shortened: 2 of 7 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/CheckGoogleSdkRequirements.cs
```diff
@@ -14,17 +14,26 @@ public class CheckGoogleSdkRequirements : AndroidTask
...
-		[Required]
+		/// <summary>
+		/// This will be blank for .NET 5 builds
+		/// </summary>
 		public string TargetFrameworkVersion { get; set; }
 
+		/// <summary>
+		/// This is used instead of TargetFrameworkVersion for .NET 5 builds
+		/// </summary>
+		public int ApiLevel { get; set; }
```
Nirmal4G, line 25:
```suggestion
		public int TargetApiLevel { get; set; }
```

This is more clear in intent than the previous name.

src/Xamarin.Android.Build.Tasks/Tasks/CheckGoogleSdkRequirements.cs
```diff
@@ -14,17 +14,26 @@ public class CheckGoogleSdkRequirements : AndroidTask
...
 		[Required]
 		public string ManifestFile { get; set; }
 
 		public override bool RunTask ()
 		{
 			ManifestDocument manifest = new ManifestDocument (ManifestFile);
 
-			var compileSdk = MonoAndroidHelper.SupportedVersions.GetApiLevelFromFrameworkVersion (TargetFrameworkVersion);
+			var compileSdk = string.IsNullOrEmpty (TargetFrameworkVersion) ?
+				ApiLevel :
```
Nirmal4G, line 35:
```suggestion
				TargetApiLevel :
```

src/Xamarin.Android.Sdk/targets/Xamarin.Android.Sdk.Tooling.targets
```diff
@@ -54,4 +55,12 @@ called for "legacy" projects in Xamarin.Android.Legacy.targets.
     </ResolveAndroidTooling>
   </Target>
 
+  <Target Name="_CheckGoogleSdkRequirements"
+      Condition="Exists('$(IntermediateOutputPath)android\AndroidManifest.xml') And '$(AndroidEnableGooglePlayStoreChecks)' == 'true' ">
+    <CheckGoogleSdkRequirements
+        ApiLevel="$(_AndroidApiLevel)"
```
Nirmal4G, line 61:
```suggestion
        TargetApiLevel="$(_AndroidApiLevel)"
```

src/Xamarin.Android.Build.Tasks/Tasks/CheckGoogleSdkRequirements.cs
jonpryor: Should this instead be `string TargetAndroidId` (plus associated chicanery and conversions), so that it can “reasonably” support preview API levels, such as this year’s current `R` (and presumably next year’s `S`, and...)?

`AndroidManifest.xml` has historically allowed `//uses-sdk/@android:targetSdkVersion` to contain values such as `R`, and we’ve historically allowed `$(TargetFrameworkVersion)` to be e.g. `v10.0.99` to  use the preview API level, and we need *some* form of equivalent in .NET 5, especially since we’ll be *losing* `$(TargetFrameworkVersion)`...

jonathanpeppers: @jonp for .NET 5 the incoming value here will be `$(_AndroidApiLevel)`, which is currently coming from:

https://github.com/xamarin/xamarin-android/blob/3c94f6f9ae7902919620c1ca26d7ad77ff080620/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAndroidTooling.cs#L188

I don't think we have a way to set this to an unstable API level yet in .NET 5, but the value would be set to `29` instead of `R`.

I think the name `ApiLevel` is OK, because we've used this for other tasks that take `$(_AndroidApiLevel)` as input:

https://github.com/xamarin/xamarin-android/blob/3c94f6f9ae7902919620c1ca26d7ad77ff080620/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L810

https://github.com/xamarin/xamarin-android/blob/3c94f6f9ae7902919620c1ca26d7ad77ff080620/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1860

https://github.com/xamarin/xamarin-android/blob/3c94f6f9ae7902919620c1ca26d7ad77ff080620/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L2366

A couple older tasks use the name `AndroidSdkPlatform`, though.

PR #4599, opened 2020-04-21
[build] see if we can add Windows Defender exclusions

jonathanpeppers:
Context: https://docs.microsoft.com/powershell/module/defender/add-mppreference

Our build and test phases on Windows currently hit `IOException` in
many random places.

This is a test to see if adding Windows Defender exclusions help
anything. It might also improve performance.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -244,6 +244,8 @@ stages:
 
     - template: yaml-templates\kill-processes.yaml
 
+    - template: yaml-templates\windows-defender.yaml
+
     - template: yaml-templates\clean.yaml
 
     - task: UseDotNet@2
--- a/build-tools/automation/yaml-templates/run-msbuild-win-tests.yaml
+++ b/build-tools/automation/yaml-templates/run-msbuild-win-tests.yaml
@@ -15,6 +15,8 @@ jobs:
 
     - template: kill-processes.yaml
 
+    - template: windows-defender.yaml
+
     - template: clean.yaml
 
     - template: setup-test-environment.yaml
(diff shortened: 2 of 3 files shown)


jonathanpeppers:
Initially it looks like it may not have helped:
```
```
Let me add some logging to actually print the paths and commands it is running.

Unfortunately, `Add-MpPreference` doesn't print anything if it succeeds.

jonathanpeppers:
Even if this doesn't fix the `IOException` it might be worth the build time improvement, comparing the builds:
```
Before:
    Windows              54m 21s
    MSBuild Windows - 1  51m 37s
After:
    Windows              57m 30s
    MSBuild Windows - 1  57m 48s
```
We'll see if the next one seems quicker, too.

jonathanpeppers:
I'm not sure if this change is worth it, as it doesn't fix the random `IOException` we see, or the equivalent for an AOT build:

```
  [aot-compiler stderr] C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\ndk\arm-linux-androideabi-strip: unable to rename 'obj\Release\aot\armeabi-v7a\libaot-System.Core.dll.so.tmp'; reason: File exists (TaskId:183)
  [aot-compiler stderr] AOT of image C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile\obj\Release\android\assets\shrunk\System.Core.dll failed. (TaskId:183)
C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets(2361,3): error XA3001: Could not AOT the assembly: System.Core.dll [C:\A\vs2019xam00000B-1\_work\2\s\bin\TestRelease\temp\BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile\UnnamedProject.csproj]
```

Issue #4127, opened 2020-01-15
Some of the new wrapped Java exceptions use BCL exceptions that differ from the related BCL types


brendanzagaeski:
### Context

<https://github.com/xamarin/xamarin-android/commit/60363ef45cd6994a322adac1b73d4271b0658c1e> added an option to wrap `Java` exception types in .NET exception types for certain Android APIs.

`JavaDictionary.Get()` is an example:

https://github.com/xamarin/xamarin-android/blob/8d7557aad23a3c0724e1996aade3072b1a09f298/src/Mono.Android/Android.Runtime/JavaDictionary.cs#L72-L86

### Expected behavior

`JavaDictionary` inherits from `Java.Lang.Object` and implements `IDictionary`:

https://github.com/xamarin/xamarin-android/blob/8d7557aad23a3c0724e1996aade3072b1a09f298/src/Mono.Android/Android.Runtime/JavaDictionary.cs#L12

The documentation for the [`IDictionary.Item[Object]`](https://docs.microsoft.com/dotnet/api/system.collections.idictionary.item?view=netstandard-2.1) indexer property mentions two exceptions:

  * `ArgumentNullException`
  * `NotSupportedException`

### Actual behavior

`JavaDictionary.Get()` can currently throw the following two different exceptions instead:

  * `InvalidCastException`
  * `NullReferenceException`

Depending on the goals of the Java exception wrapping, throwing different System exceptions from `JavaDictionary.Get()` than the ones documented for `IDictionary.Item[Object]` might be unintended. Maybe only the exception conditions documented for `IDictionary.Item[Object]` (like "`key` is `null`") need to be surfaced as System exceptions, and other exceptions, if any, can be left as Java exceptions?

### Related scenarios

If this `JavaDictionary.Get()` scenario does look good to change, then there are probably a few additional similar cases to change too.

Another slightly different case is `JavaCollection.Add()`. I think it might have been wrapped because `JavaCollection` implements `ICollection`, but I just noticed today that the non-generic [`ICollection`](https://docs.microsoft.com/dotnet/api/system.collections.icollection?view=netstandard-2.1) interface doesn't include an `Add()` method. Only the generic [`ICollection<T>`](https://docs.microsoft.com/dotnet/api/system.collections.generic.icollection-1?view=netstandard-2.1) includes an `Add()` method. So maybe `JavaCollection.Add()` doesn't need to wrap any exceptions?

brendanzagaeski:
[...]

Issue #3216, opened 2019-06-13
Android HttpClient implementation in .NET Standard Project throwing Java Exception


muhaym:
### Steps to Reproduce

1. Set the AndroidHttpClientHandlerType to Xamarin.Android.Net.AndroidClientHandler
![image](https://user-images.githubusercontent.com/10905479/59449402-9a49b780-8e24-11e9-8f4f-9e1cabbb924f.png)

2. Replicate poor internet or no internet case using emulator / device

Different exception thrown will be like
```json
[...]

PR #4601, opened 2020-04-22
[AndroidClientHandler] Wrap Java exceptions in .NET ones

grendello:
Context: 60363ef45cd6994a322adac1b73d4271b0658c1e

This commit is, in a way, a follow-up to 60363ef4 in that it causes less
Java exceptions to "leak" to applications which use code shared between
platforms that might not be able to check for Java exception types
directly (because they don't link against `Mono.Android.dll`).  However,
instead of making the behavior optional, as implemented in 60363ef4,
this commit unconditionally wraps `Java.IO.Exception` and some of its
derivatives in the standard .NET `WebException`.  This makes the handler
to behave in a manner that's closer to how other `HttpClient` custom
handlers behave.

### DIFF
--- a/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
+++ b/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
@@ -292,13 +292,13 @@ namespace Xamarin.Android.Net
 						throw new InvalidOperationException ("Request redirected but no new URI specified");
 					request.Method = redirectState.Method;
 				} catch (Java.Net.SocketTimeoutException ex) when (JNIEnv.ShouldWrapJavaException (ex)) {
-					throw new WebException (ex.Message, ex, WebExceptionStatus.Timeout, null);
+					throw new HttpRequestException (ex.Message, ex, WebExceptionStatus.Timeout, null);
 				} catch (Java.Net.UnknownServiceException ex) when (JNIEnv.ShouldWrapJavaException (ex)) {
-					throw new WebException (ex.Message, ex, WebExceptionStatus.ProtocolError, null);
+					throw new HttpRequestException (ex.Message, ex, WebExceptionStatus.ProtocolError, null);
 				} catch (Java.Lang.SecurityException ex) when (JNIEnv.ShouldWrapJavaException (ex)) {
-					throw new WebException (ex.Message, ex, WebExceptionStatus.SecureChannelFailure, null);
+					throw new HttpRequestException (ex.Message, ex, WebExceptionStatus.SecureChannelFailure, null);
 				} catch (Java.IO.IOException ex) when (JNIEnv.ShouldWrapJavaException (ex)) {
-					throw new WebException (ex.Message, ex, WebExceptionStatus.UnknownError, null);
+					throw new HttpRequestException (ex.Message, ex, WebExceptionStatus.UnknownError, null);
 				}
 			}
 		}
@@ -833,7 +833,7 @@ namespace Xamarin.Android.Net
 			try {
 				httpConnection.RequestMethod = request.Method.ToString ();
 			} catch (Java.Net.ProtocolException ex) when (JNIEnv.ShouldWrapJavaException (ex)) {
-				throw new WebException (ex.Message, ex, WebExceptionStatus.ProtocolError, null);
+				throw new HttpRequestException (ex.Message, ex, WebExceptionStatus.ProtocolError, null);
 			}
 
 			// SSL context must be set up as soon as possible, before adding any content or
--- a/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
+++ b/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
@@ -292,13 +292,13 @@ namespace Xamarin.Android.Net
 						throw new InvalidOperationException ("Request redirected but no new URI specified");
 					request.Method = redirectState.Method;
 				} catch (Java.Net.SocketTimeoutException ex) when (JNIEnv.ShouldWrapJavaException (ex)) {
-					throw new HttpRequestException (ex.Message, ex, WebExceptionStatus.Timeout, null);
+					throw new WebException (ex.Message, ex, WebExceptionStatus.Timeout, null);
 				} catch (Java.Net.UnknownServiceException ex) when (JNIEnv.ShouldWrapJavaException (ex)) {
-					throw new HttpRequestException (ex.Message, ex, WebExceptionStatus.ProtocolError, null);
+					throw new WebException (ex.Message, ex, WebExceptionStatus.ProtocolError, null);
 				} catch (Java.Lang.SecurityException ex) when (JNIEnv.ShouldWrapJavaException (ex)) {
-					throw new HttpRequestException (ex.Message, ex, WebExceptionStatus.SecureChannelFailure, null);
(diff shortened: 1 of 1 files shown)


brendanzagaeski:
- Since this changes behavior (e.g., if apps are currently catching the Java exception types, those catches would stop working after updating Xamarin.Android versions), maybe this new behavior can go under an MSBuild property too, like 60363ef45cd6994a322adac1b73d4271b0658c1e?

- Depending on the intended scope of this PR, maybe it would be handy to cover the exceptions mentioned in <https://github.com/xamarin/xamarin-android/issues/3216> (cross-reference: <https://github.com/xamarin/xamarin-android/issues/4127#issuecomment-579971538>) so that item could be closed?

  For example, `Java.Net.UnknownHostException` can be wrapped in `System.Net.Http.HttpRequestException` for:

  ```cs
  try {
  	var client = new HttpClient ();
  	await client.GetStringAsync ("https://does.not.exist.contoso.com");
  } catch (System.Net.Http.HttpRequestException ex) {
  	Console.WriteLine (ex);
  }
  ```

  (This catches in a .NET Core app but not yet with `AndroidClientHandler`.)

  And `Javax.Net.Ssl.SSLHandshakeException` can be wrapped in `System.Net.Http.HttpRequestException` too:


  ```cs
  try {
  	var client = new HttpClient ();
  	await client.GetStringAsync ("https://has.an.expired.tls.cert.contoso.com");
  } catch (System.Net.Http.HttpRequestException ex) {
  	Console.WriteLine (ex);
  }
  ```

  (This also catches in a .NET Core app but not yet with `AndroidClientHandler`.)


- Maybe `HttpRequestException` fits well for the other wrappings added in this PR too instead of `WebException`? (Reference: [`HttpClient.SendAsync()`](https://docs.microsoft.com/dotnet/api/system.net.http.httpclient.sendasync?view=netcore-3.1) ,  [`HttpClientHandler.SendAsync()`](https://docs.microsoft.com/dotnet/api/system.net.http.httpclienthandler.sendasync?view=netcore-3.1))

  For example, using `192.168.1.255` as a placeholder for an endpoint that times out, and assuming the app is [configured to allow cleartext HTTP traffic](https://developer.android.com/training/articles/security-config) to that address:

  ```cs
  try {
  	var client = new HttpClient(new AndroidClientHandler {
  		ConnectTimeout = TimeSpan.FromMilliseconds (10),
  		ReadTimeout = TimeSpan.FromMilliseconds(10)
  	});
  	await client.GetStringAsync ("http://192.168.1.255");
  } catch (System.Net.Http.HttpRequestException ex) {
  	Console.WriteLine (ex);
  }
  ```

  This catches in a .NET Core app (using the default `SocketsHttpHandler`) but not yet in a Xamarin.Android app with `AndroidClientHandler`. With the changes from this PR, the exception would be wrapped from `Java.Net.SocketTimeoutException` into a `WebException` instead of an `HttpRequestException`.

- Perhaps even the older `WebException` wrapping for `Java.Net.ConnectException` could be adjusted conditionally to `HttpRequestException` based on the new MSBuild property? For example, using `192.168.1.255:22` as a placeholder for an endpoint that refuses a connection:

  ```cs
  try {
  	var client = new HttpClient ();
  	});
  	await client.GetStringAsync ("http://192.168.1.255:22");
  } catch (System.Net.Http.HttpRequestException ex) {
  	Console.WriteLine (ex);
  }
  ```

  This catches in a .NET Core app but not yet in a Xamarin.Android app with `AndroidClientHandler` because it throws a `WebException` instead of an `HttpRequestException`.

- Maybe tests could also be added for some of the scenarios like the 4 examples in this comment? Maybe the existing test servers for other connection tests can produce these exceptions?

brendanzagaeski:
> Maybe the existing test servers for other connection tests can produce these exceptions?

Or maybe the BCL tests for Mono or .NET Core include some `HttpClient` exception scenarios that could help ensure matching behavior for `AndroidClientHandler`?

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
```diff
@@ -282,14 +282,24 @@ protected virtual IHostnameVerifier GetSSLHostnameVerifier (HttpsURLConnection c
...
+				try {
+					HttpURLConnection httpConnection = await SetupRequestInternal (request, java_connection).ConfigureAwait (continueOnCapturedContext: false);
+					HttpResponseMessage response = await ProcessRequest (request, java_url, httpConnection, cancellationToken, redirectState).ConfigureAwait (continueOnCapturedContext: false);
+					if (response != null)
+						return response;
+
+					if (redirectState.NewUrl == null)
+						throw new InvalidOperationException ("Request redirected but no new URI specified");
+					request.Method = redirectState.Method;
+				} catch (Java.Net.SocketTimeoutException ex) {
```
jonpryor, line 294: As @brendanzagaeski pointed out, these `catch` blocks should use `when (JNIEnv.ShouldWrapJavaException (ex))`, as was done in 60363ef45.

grendello:
@brendanzagaeski I'm in two minds regarding `WebException` vs `HttpRequestException` - the former is thrown by the default `HttpClient` handler implementation which uses the "old" .NET networking stack, the latter may be thrown by other handlers and is thrown by `HttpClient` itself but switching to it breaks backward compatibility of `AndroidClientHandler`. Mixing the two is, IMO, not advisable as it would introduce some confusion on which exception is thrown when - we should use one of them. I don't have an opinion on which is "better" - it's ultimately a strategic decision at this point. @jonpryor ?

Regarding tests... Yeah, it would be nice, but testing network behaviors is hard and in our case we can't mock the requests - we must issue actual requests to actual servers, so that the Java backend actually kicks in. Yes, we could make our code mockable by introducing a fake Java backend that would throw certain exceptions at certain points, but I honestly don't see much value in it. Regarding the test servers - we've been wanting to do it for quite a while now (to use https://httpbin.org/ on the test servers so we can perform all kinds of tests in controlled manner) but so far it hasn't happened. When we have a server we can rely on and fully control (in the `httpbin` sense), then sure.

jonpryor:
The underlying intent behind 60363ef was that ".NET methods" should throw *only* documented exception types.

[`HttpClient.GetAtringAsync(string)`](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.getstringasync?view=netcore-3.1) only documents two exception types:

  * `ArgumentNullException`
  * `HttpRequestException`

`WebException` is *not* on this list.  Consequently, we should be throwing `HttpRequestException`.

jonpryor:
…except:

1. `HttpRequestException` lacks additional *useful* information such as [`WebException.Status`](https://docs.microsoft.com/en-us/dotnet/api/system.net.webexception.status?view=netcore-3.1)
2. The `HttpClient` docs are quasi-meaningless, because
3. We *actually* care about `HttpClientHandler`, *not* `HttpClient`, and
4. [`HttpClientHandler.SendAsync()`](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.sendasync?view=netcore-3.1) is *more* useless, because it *only* documents `ArgumentNullException`.

Nor are the `HttpClient` docs complete; `ObjectDisposedException`, `OperationCanceledException`, and `TimeoutException` aren't documented either, and we see places where those can be thrown from `HttpClient`!

Thus, if everything is meaningless, go with the type that provides the most useful info: `WebException`.

PR #4606, opened 2020-04-22
[Xamarin.Android.Build.Tasks] refactor common logic for binding projects

jonathanpeppers:
For .NET 5, I am looking into getting binding projects working.

There may not need to be a concept of a "binding project" *at all*, if
we can get the item groups to work in a regular .NET 5 Android
application or library.

My first thought was to just try something like this in
`Microsoft.Android.Sdk.targets`:

    <Import Project="$(MSBuildThisFileDirectory)..\tools\Xamarin.Android.Bindings.targets" />
    <Import Project="$(MSBuildThisFileDirectory)..\tools\Xamarin.Android.Common.targets" />

However, there is a lot of duplicate MSBuild targets in
`Xamarin.Android.Bindings.targets`. There are targets with the same
name that do slightly different things.

To try to consolidate this, I've created two new files:

* `Xamarin.Android.DesignTime.targets` - for design-time build things
* `Xamarin.Android.EmbeddedResource.targets` - for creation/extraction
  of `__AndroidLibraryProjects__.zip` or `__AndroidNativeLibraries__.zip`.

Now Xamarin.Android class libraries and binding projects will share
the same targets in these files.

This will bring binding projects closer to working in .NET 5. If we
get `Xamarin.Android.Binding.targets` as short as it can be, it will
be much easier to get working.

Other changes:

* A duplicate `_GetReferenceAssemblyPaths` target could be removed. It
  is already present in `Xamarin.Android.Legacy.targets`.
* I left the `ResolveLibraryProjects` target intact, as it seems to
  have a very public-looking name. I changed it to depend on a private
  `_CreateBindingResourceArchive` target.

### DIFF
--- a/build-tools/installers/create-installers.targets
+++ b/build-tools/installers/create-installers.targets
@@ -221,7 +221,9 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.CSharp.targets" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.D8.targets" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Designer.targets" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.DesignTime.targets" />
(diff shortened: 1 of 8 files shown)


dellis1972:
This is looking great :)

PR #4607, opened 2020-04-23
Fix build break on newly reimaged bot

mjbond-msft:
In the process of creating a new Mac OSS PR workflow, I encountered an issue where the build would consistently break when executing the `prepare-update-mono` target.  The break occurred on the following command on _a newly reimaged agent (bot)_:
```bash
pgrep -lfi VBCSCompiler.exe
```
The command returns 0 if the managed compiler server `VBCSCompiler.exe` is running; otherwise, it returns 1 if the process is not running.  When executing the build pipeline on the newly re-imaged machine, the process is not running, the `pgrep` command returns 1 and the build fails with the following error: 

```
pgrep -lfi VBCSCompiler.exe
make: *** [prepare-update-mono] Error 1
```

Turns out that on established build machines in the pool, `VBCSCompiler.exe` is consistently found to be running at this stage of the build.  If you inspect a build log, you will see that the `pgrep` command logged that `VBCSCompiler.exe` is running.  Here's an example from a recent build log:

```
pgrep -lfi VBCSCompiler.exe
36187 /Library/Frameworks/Mono.framework/Versions/6.10.0/bin/mono-sgen64 /Library/Frameworks/Mono.framework/Versions/6.10.0/lib/mono/msbuild/Current/bin/Roslyn/VBCSCompiler.exe -pipename:sX2CFXy8VTADY0R+V+y_1o2brmCa+1ch6aJTNiSJU4w
```

In discussing the issue with @directhex, the intent of the command is quite the opposite. The intent is to ensure that `VBCSCompiler.exe` is not running at this point of time. Based on the purpose of the target to perform a mono update, we should ensure that processes associated with the `mono` toolset are shutdown and not running prior to the update.  Since `VBCSCompiler.exe` is part of the `mono` toolset it seems to make more logical sense to ensure that it is not running.  In other words, ensure that the next invocation of `MSBuild` starts the compiler server from the new location if `mono` is updated.

**Note:** The original change was introduced to `master` on 2020-03-13 via this commit: 8599d63d2. That change was verified and continued to run, as with most changes, on a well established build agent (bot).

**Build validation**
Log output from the most recent [PR build](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3670578&view=logs&j=96fd57f5-f69e-53c7-3d47-f67e6cf9b93e&t=967deba6-588a-5db9-6e70-909908c847bd) indicates that the `VBCSCompiler` process was successfully terminated

```
pgrep -lfi VBCSCompiler.exe 2>/dev/null || true
/Library/Frameworks/Mono.framework/Versions/6.12.0/bin/mono-sgen64 /Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/Roslyn/VBCSCompiler.exe -pipename:9+YB2btX0dwaCLheWLK88CakD6Q4tWAcacUOQ2hCxbI
VBCSCompiler process ID (if running): 72723
Terminating the VBCSCompiler '72723' server process prior to updating mono
Verified that the VBCSCompiler server process no longer exists
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,8 +210,7 @@ prepare-help: prepare-build
 
 .PHONY: prepare-update-mono
 prepare-update-mono: prepare-build
-	# Ensure VBCSCompiler.exe process isn't running during the mono update
-	# pgrep -lfi VBCSCompiler.exe
+	# Ensure the VBCSCompiler.exe process isn't running during the mono update
 	$(eval pid=$(shell sh -c "pgrep VBCSCompiler.exe 2>/dev/null" | awk '{ print $$1 }'))
 ifneq ($(pid),)
 	@echo "VBCSCompiler.exe process '$(pid)' is running. Killing process prior to mono update ..."
--- a/Makefile
+++ b/Makefile
@@ -215,7 +215,7 @@ shutdown-compiler-server:
 	$(eval pid=$(shell sh -c "pgrep -lfi VBCSCompiler.exe 2>/dev/null" | awk '{ print $$1 }'))
 	@echo "VBCSCompiler process ID (if running): $(pid)" ;\
 	if [[ -n "$(pid)" ]]; then \
-		echo "VBCSCompiler.exe process '$(pid)' is running. Destroying process prior to updating mono" ;\
+		echo "Terminating the VBCSCompiler.exe '$(pid)' process prior to updating mono" ;\
 		pgrep -lfi VBCSCompiler.exe ;\
 		kill -HUP $(pid) 2>/dev/null ;\
 	fi
--- a/Makefile
+++ b/Makefile
@@ -213,7 +213,7 @@ prepare-update-mono: prepare-build
 	# Ensure VBCSCompiler.exe process isn't running during the mono update
 	# pgrep -lfi VBCSCompiler.exe
 	$(eval pid=$(shell sh -c "pgrep VBCSCompiler.exe 2>/dev/null" | awk '{ print $$1 }'))
-ifeq ($(pid),)
+ifneq ($(pid),)
 	@echo "VBCSCompiler.exe process '$(pid)' is running. Killing process prior to mono update ..."
 	kill -HUP $(pid)
 endif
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -208,11 +208,22 @@ prepare: prepare-build
 prepare-help: prepare-build
 	mono --debug $(PREPARE_EXE) -h
 
+.PHONY: shutdown-compiler-server
+shutdown-compiler-server:
+	# Ensure the VBCSCompiler.exe process isn't running during the mono update
+	pgrep -lfi VBCSCompiler.exe 2>/dev/null || true
+	$(eval pid=$(shell sh -c "pgrep -lfi VBCSCompiler.exe 2>/dev/null" | awk '{ print $$1 }'))
```
jonpryor, line 215: This line needs to end with a `\` in order for `$$pid` to be usable on the following line.

(By default, each line in a Makefile gets its own shell.)

I also don't understand why this. is using `$(eval)` as well.

Makefile
```diff
@@ -208,11 +208,22 @@ prepare: prepare-build
 prepare-help: prepare-build
 	mono --debug $(PREPARE_EXE) -h
 
+.PHONY: shutdown-compiler-server
+shutdown-compiler-server:
```
jonpryor, line 212: A more "conventional" implementation would be:

```make
.PHONY: shutdown-compiler-server
shutdown-compiler-server:
	# Ensure the VBCSCompiler.exe process isn't running during the mono update
	pgrep -lfi VBCSCompiler.exe 2>/dev/null || true
	pid=`pgrep -lfi VBCSCompiler.exe 2>/dev/null | awk '{ print $$1 }'`;                    \
	echo "VBCSCompiler process ID (if running): $$pid" ;                                   \
	if [ -n "$$pid" ]; then                                                                 \
		echo "Terminating the VBCSCompiler '$$pid' server process prior to updating mono" ; \
		kill -HUP $$pid 2>/dev/null ;                                                       \
		echo "VBCSCompiler server process should no longer exist" ;                         \
		pgrep -lfi VBCSCompiler.exe 2>/dev/null || true ;                                   \
	fi
```

mjbond-msft: updated to align on bash. indeed much cleaner than the Frankenstein approach of combining make and bash. commit: 8a3d64455a394724a94b78a28b9a035897ac8bfb

Makefile
mjbond-msft: See comment below. Addressed with commit 8a3d64455a394724a94b78a28b9a035897ac8bfb

PR #4614, opened 2020-04-24
[Xamarin.Android.Build.Tasks] Fixup the AndroidManifest.xml path.

dellis1972:
Fixes 662837

We were not fixing up the path for the AndroidManifest.xml
file if we get errors in it. Normally it points to the
file in `$(IntermediateOutputPath)android\AndroidManifest.xml`.

The problem here is that file is NOT a file the user should
edit. Since it will be overwritten on the next build.
So lets fix up the path to point to the user editable one.

One issue with this is what happens if there is an error in
the manifest that is in the generated parts of the file...
In which case the line numbers and even the actual xml will
NOT be present in the user manifest. Not sure how we can deal
with that issue. For now this seems like a better solution
than what we currently have.


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs
```diff
@@ -119,6 +119,7 @@ public void Bug12935 ([Values (true, false)] bool useAapt2)
 				Assert.IsFalse (builder.Build (proj), "Build for TargetFrameworkVersion 15 should have failed");
 				StringAssertEx.Contains (useAapt2 ? "APT2259: " : "APT1134: ", builder.LastBuildOutput);
 				StringAssertEx.Contains (useAapt2 ? "APT2067" : "", builder.LastBuildOutput);
+				StringAssertEx.Contains ("Properties/AndroidManifest.xml(6)", builder.LastBuildOutput);
```
jonpryor, line 122: Looks like this new assert is failing: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3690462&view=ms.vss-test-web.build-test-results-tab&runId=12949274&resultId=100141&paneView=debug

```
String did not contain 'Properties/AndroidManifest.xml(6)'!
```

Also, might this assert be macOS-specific?

dellis1972: Its is Os specific, a Path separator :'( 

`Properties\AndroidManifest.xml(6)` vs `Properties/AndroidManifest.xml(6)`

🤦

jonpryor:
"Better" is we give them a meaningful line number. :-)

The problem is *what* a meaningful line number *is*, and how we get that.

For example, in your `[assembly: UsesPermission]` sample, the meaningful line number is the `[assembly: UsesPermission]` line.  Can we *get* that line? Not easily.  It gets worse if that line is in another referenced assembly or (*shudder*) a NuGet package.

Thus, perhaps the answer is https://knowyourmeme.com/memes/why-not-both-why-dont-we-have-both

Why don't we do both?

Which is to say, instead of *one* error message which (currently) lists the generated `AndroidManifest.xml` file, we emit *two* error messages, one listing the generated file -- with clear wording that it's *generated* -- and another error listing the "real" file, and ensure that the error mentions that the generated file *contains **generated** contents* and that errors may "come from" elsewhere.

dellis1972:
So we should emit both then? I'll take a look at doing that.

dellis1972:
ok, so what text should we add to the actual error...  will it need to be localised as well. 

> <blah error>. Note this file is auto generated. The error itself might be in <path to manifest>, code or a library your application is consuming.

We should add a bunch of xml comments to the generated manifest file as well I guess saying the same thing, so even if the user does open it, it will be obvious  that its a generated file.

dellis1972:
Hmm errors are 

```
```

src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs
```diff
@@ -154,6 +156,20 @@ XElement GetActivityWithName (XElement app, string name)
 			return app.Elements ("activity").FirstOrDefault (e => ToFullyQualifiedName ((string) e.Attribute (androidNs + "name")) == name);
 		}
 
+		void InjectAutoGeneratedComment (XElement root, string templateFilename = "")
+		{
+			var sb = new StringBuilder ();
+			sb.AppendLine ();
+			sb.AppendLine ("This code was generated by a tool.");
+			if (!string.IsNullOrEmpty (templateFilename)) {
+				sb.AppendLine ($"It was generated from {templateFilename}");
```
jonpryor, line 165: Very nice.  Now should *this* be localized as well?

¯\_(ツ)_/¯

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
```diff
@@ -166,12 +168,19 @@ protected bool LogAapt2EventsFromOutput (string singleLine, MessageImportance me
...
+				}
+
 				// Strip any "Error:" text from aapt's output
 				if (message.StartsWith ("error: ", StringComparison.InvariantCultureIgnoreCase))
 					message = message.Substring ("error: ".Length);
 
 				if (level.Contains ("error") || (line != 0 && !string.IsNullOrEmpty (file))) {
 					LogCodedError (GetErrorCode (message), message, file, line);
+					if (manifestError)
+						LogCodedError (GetErrorCode (message), message, AndroidManifestFile.ItemSpec, line);
```
jonpryor, line 183: On the plus side, I believe that the original message comes from `aapt`/etc., and thus we don't need to localize it.

On the downside, should we be emitting the *same* error message for *both* filenames?  Or should we distinguish these messages somehow?

```
obj\Debug\…\AndroidManifest.xml:2: error APT2001: already defined a <flag> …
Properties\AndroidManifest.xml:2: error APT2001: Template file for preceding error message
```

If we do this, we'll need to localize the Template file for preceding error message" text.

src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs
dellis1972: Well we don't localise the generated text in the `Resource.designer.cs` so I'm not sure.

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
dellis1972: 🤔

brendanzagaeski: To throw in my 2 cents, I like the sound of two slightly different messages. If I'm following the PR correctly, I believe there will be cases where there will be errors that only affect the generated version of the file, for example due to a typo in a resource name within an `ActivityAttribute.Label` property. Having a different message for the template file will hopefully make the error messages for that scenario a little less confusing.

brendanzagaeski:
*Draft release notes*

As this PR is finalized for merge, when you get a chance, you can add a release note for it to `Documentation/release-notes/4614.md` as part of the PR. Here's a rough first draft. Thanks!

```
  * [GitHub PR 4614](https://github.com/xamarin/xamarin-android/pull/4614):
    Syntax errors in `AndroidManifest.xml` always referred to the generated
    version of the file (like
    `obj\Debug\90\android\manifest\AndroidManifest.xml`), even when the syntax
    error was also present in the hand-written input
    `Properties\AndroidManifest.xml` file.  Now two errors are provided for any
    syntax errors in `AndroidManifest.xml`, one matching the generated file, and
    one matching the hand-written file.
```

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -130,6 +130,23 @@
...
+    <comment></comment>
+  </data>
+  <data name="XA_Manifest_AutoGenerated_Header" xml:space="preserve">
+    <value>
+    This code was generated by a tool.
+    It was generated from {0}
+    Changes to this file may cause incorrect behavior and will be lost if
+    the contents are regenerated.
+    </value>
+    <comment>{0} - The path to AndroidManiest.xml</comment>
```
brendanzagaeski, line 148: Candidate comment wording:

```
<comment>{0} - The path of the template file.</comment>
```

dellis1972:
@brendanzagaeski you gave me an idea. We don't need two errors we can do one and add the additional text to it. That should simplify things somewhat.

brendanzagaeski:
It's an interesting idea to try to get it down to one error, but I think with the current wording, it brings the discussion back to the complications raised in <https://github.com/xamarin/xamarin-android/pull/4614#pullrequestreview-403820314> and <https://github.com/xamarin/xamarin-android/pull/4614#issuecomment-622197914>.

For example, with the current latest version of this PR, if I intentionally add a syntax error to line 7 of my `Properties\AndroidManifest.xml`, I get the following error:

```
C:\Temp\AndroidApp1\AndroidApp1\Properties\AndroidManifest.xml(14): error APT2260: resource mipmap/ic_launche (aka com.contoso.androidapp1:mipmap/ic_launche) not found.. This error is likely caused by an issue with the AndroidManifest.xml file or an Android manifest generation attribute in a source code file.
```

So the line number provided in this error message is misleading. And this approach also goes back to losing the feature mentioned in <https://github.com/xamarin/xamarin-android/pull/4614#pullrequestreview-403820314>:

> Right now they could double-click in the IDE and see the line with the invalid value, and we lose that.

dellis1972:
Well we can never fix up the Line numbers without a lot of work. I don't think it's worth the effort. 
While the two error messages was a nice idea , the fact that the IDE's may or may not show them in order is a problem. 

I'm now back to the idea of just fixing up the file path and adding a comment to say that the error might not have anything to do with that file and might be caused by other things.

Documentation/release-notes/4614.md
```diff
@@ -0,0 +1,8 @@
+* [GitHub PR 4614](https://github.com/xamarin/xamarin-android/pull/4614):
+    Syntax errors in `AndroidManifest.xml` always referred to the generated
+    version of the file (like
+    `obj\Debug\90\android\manifest\AndroidManifest.xml`), even when the syntax
+    error was also present in the hand-written input
+    `Properties\AndroidManifest.xml` file.  Now two errors are provided for any
```
jonpryor, line 6: I believe that these release notes are now "out of date", as this PR only emits *one* error message, not two.

brendanzagaeski:
Sounds fair. OK by me on losing the little feature of double-clicking the error message to see the context within the generated file.

Would it be OK to remove the line number from the error message then? Given what I saw with introducing an error in my `Properties/AndroidManifest.xml` file, it seems like having the line number included might usually be confusing?

One other question is whether it might be appropriate for the message to mention that other `AndroidManifest.xml` files merged from NuGet packages could be involved? Off-hand, I'm not sure if manifest merging can result in this kind of error, but it seems possible?

brendanzagaeski:
Another almost silly little thing is maybe the `message` value passed in for the `{0}` format item could have `.` trimmed off the end if it's present so the final message won't have a double `..` 😄

brendanzagaeski:
I think the message wording is still the same since I last checked. I still get:

```
C:\Temp\AndroidApp1\AndroidApp1\Properties\AndroidManifest.xml(14): error APT2260: resource mipmap/ic_launche (aka com.contoso.androidapp1:mipmap/ic_launche) not found.. This error is likely caused by an issue with the AndroidManifest.xml file or an Android manifest generation attribute in a source code file.
```

To recap, three questions I had on this wording were:

1. Would it be OK to remove the line number from the error message since it is incorrect?

2. Could the `message` value passed in for the `{0}` format item could have `.` trimmed off the end if it's present.
   
   Or maybe easier, a `.` could be appended to the message if it doesn't yet have one, and then the `.` could be removed from the resource in the `.resx` file instead.

3. Can manifest merging result in this kind of error?

   I tested this. The answer is it is indeed possible, so I will suggest a tiny wording adjustment from "the" to "an" to cover this scenario too.

Here's a rough draft `git apply`-able diff with what I'm thinking:

```diff
diff --git a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
index 0d9ea217..ea75e5eb 100644
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -131,8 +131,8 @@
     <comment>{0} - The regular expression that the file name must match</comment>
   </data>
   <data name="AAPTManifestError" xml:space="preserve">
-    <value>{0}. This error is likely caused by an issue with the AndroidManifest.xml file or an Android manifest generation attribute in a source code file.</value>
-    <comment>{0} - The error message returned by the AAPT or AAPT2 tool</comment>
+    <value>{0} This error is likely caused by an issue with an AndroidManifest.xml file or an Android manifest generation attribute in a source code file.</value>
+    <comment>{0} - The error message returned by the AAPT or AAPT2 tool. Always ends with a period "." character.</comment>
   </data>
   <data name="XA_Manifest_AutoGenerated_Header" xml:space="preserve">
     <value>
diff --git a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt.cs b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt.cs
index 09697ef1..940f4dd3 100644
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt.cs
@@ -419,7 +419,7 @@ namespace Xamarin.Android.Tasks
 
 				if (level.Contains ("error") || (line != 0 && !string.IsNullOrEmpty (file))) {
 					if (manifestError)
-						LogCodedError (GetErrorCode (message), string.Format (Xamarin.Android.Tasks.Properties.Resources.AAPTManifestError, message), AndroidManifestFile.ItemSpec, line);
+						LogCodedError (GetErrorCode (message), string.Format (Xamarin.Android.Tasks.Properties.Resources.AAPTManifestError, message.EndsWith (".") ? message : $"{message}."), AndroidManifestFile.ItemSpec, 0);
 					else
 						LogCodedError (GetErrorCode (message), message, file, line);
 					return;
diff --git a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
index d7708aee..d8e44df1 100644
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
@@ -179,7 +179,7 @@ namespace Xamarin.Android.Tasks {
 
 				if (level.Contains ("error") || (line != 0 && !string.IsNullOrEmpty (file))) {
 					if (manifestError)
-						LogCodedError (GetErrorCode (message), string.Format (Xamarin.Android.Tasks.Properties.Resources.AAPTManifestError, message), AndroidManifestFile.ItemSpec, line);
+						LogCodedError (GetErrorCode (message), string.Format (Xamarin.Android.Tasks.Properties.Resources.AAPTManifestError, message.EndsWith (".") ? message : $"{message}."), AndroidManifestFile.ItemSpec, 0);
 					else
 						LogCodedError (GetErrorCode (message), message, file, line);
 					return true;
diff --git a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs
index f4579ad7..7508e73c 100644
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs
@@ -119,7 +119,7 @@ namespace Bug12935
 				Assert.IsFalse (builder.Build (proj), "Build for TargetFrameworkVersion 15 should have failed");
 				StringAssertEx.Contains (useAapt2 ? "APT2259: " : "APT1134: ", builder.LastBuildOutput);
 				StringAssertEx.Contains (useAapt2 ? "APT2067" : "", builder.LastBuildOutput);
-				StringAssertEx.Contains (Path.Combine ("Properties", "AndroidManifest.xml(12)"), builder.LastBuildOutput);
+				StringAssertEx.Contains (Path.Combine ("Properties", "AndroidManifest.xml : error"), builder.LastBuildOutput);
 				StringAssertEx.Contains ($"{(useAapt2 ? "2" : "1")} Error(s)", builder.LastBuildOutput);
 			}
 		}
```

dellis1972:
I think this is good to go. Test failure is 

```
error XA3001: Could not AOT the assembly: System.dll 
```

Not related to this PR.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="utf-8"?>
 <root>
-  <!-- 
-    Microsoft ResX Schema 
-    
+  <!--
+    Microsoft ResX Schema
+
```
jonathanpeppers, line 5: @brendanzagaeski and I usually edit the `.resx` file in VS Windows and I guess it's choosing different whitespace settings. Are we going to have an IDE whitespace fight going forward? 😜 

Should we avoid whitespace changes in `Resources.resx` and `Resource.designer.cs`?

Normal C# files should be fine if it is using the `.editorconfig` file to format.

dellis1972: I'm just editing this in VScode so it should be using the .editorconfig settings...

dellis1972:
@brendanzagaeski I used `TrimEnd('.')` rather than the EndsWith stuff. It seemed simpler :)

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
brendanzagaeski: Unfortunately, yes, the Visual Studio `.resx` resource editor recreates this comment with trailing spaces every time it saves the file. And unfortunately, as far as I can find, `.editorconfig` does not provide settings to trim trailing whitespace within comments in Visual Studio. The current `.editorconfig` settings do trim trailing whitespace outside of comments in both C# and XML files, but not within comments. Apparently Visual Studio Code works differently and does trim trailing whitespace within comments.

Maybe the simplest compromise for now would be to set `trim_trailing_whitespace = false` for `.resx` files in the `.editorconfig` file. That would leave the responsibility on contributors to trim the whitespace by hand as appropriate for any changed lines and not to change trailing whitespace for any other lines. I have submitted that change in <https://github.com/xamarin/xamarin-android/pull/4954>.

For this PR, I have pushed a commit to the PR branch that resets the `Resource.Designer.cs` and `Resources.resx` files to match what Visual Studio generates. Feel free to squash that into the other commit as this PR moves to completion.

brendanzagaeski:

Tidy up the generated resource files. I pushed a commit to set contents of the `Resources.Designer.cs` and `Resources.resx` files to match what Visual Studio generates. The changes are just in whitespace, comment formatting, and ordering of XML elements. Feel free to squash that into the other commit as this PR moves to completion.

jonpryor:
Work-in-progress commit message:

```markdown
Fixes: http://work.devdiv.io/662837

During the build process, the `$(AndroidManifest)` file -- usually
`Properties\AndroidManifest.xml` -- is "merged" with generated XML
fragments based on custom attributes such as
`Android.App.ActivityAttribute`.

If there is an error in the generated file -- often because of an
error within `$(AndroidManifest)`, e.g. an invalid reference to an
Android resource -- then `aapt`/`aapt2`/etc. will emit an error message
which mentions `$(IntermediateOutputPath)android\AndroidManifest.xml`:

	…\obj\Debug\MonoAndroid70\android\manifest\AndroidManifest.xml(7): error APT2258: resource mipmap/invalid (aka com.xamarin.android.helloworld:mipmap/invalid) not found

Unfortunately, there's no easy way to accurately determine if the cause
of the error message is due to `$(AndroidManifest)` or instead due to
the *generated* portions.  Is the error in the customer's code, or in
*Xamarin.Android* code?

Update the error message to mention the multiple causes of the error:

	…\obj\Debug\MonoAndroid70\android\manifest\AndroidManifest.xml(7): error APT2258:
	resource mipmap/invalid (aka com.xamarin.android.helloworld:mipmap/invalid) not found.
	This error is likely caused by an issue with the AndroidManifest.xml file or an Android
	manifest generation attribute in a source code file.

Additionally, add an XML comment to the top of generated
`AndroidManifest.xml` file, indicating that the file is generated:

	<!--
	  This code was generated by a tool.
	  It was generated from {0}
	  Changes to this file may cause incorrect behavior and will be lost if
	  the contents are regenerated.
	  -->
```

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -130,6 +130,19 @@
     <value>Invalid file name: It must contain only {0}.</value>
     <comment>{0} - The regular expression that the file name must match</comment>
   </data>
+  <data name="AAPTManifestError" xml:space="preserve">
+    <value>{0}. This error is likely caused by an issue with the AndroidManifest.xml file or an Android manifest generation attribute in a source code file.</value>
```
jonpryor, line 134: @dellis1972: while attempting to write the [above commit message](https://github.com/xamarin/xamarin-android/pull/4614#issuecomment-674600906), I turn my attention to this error message.  I have two philosophical concerns with it:

1. We emit only one error message, and it's a (potentially) very *long* error message at that.  While the error message is lined wrapped, e.g. [this VS error list image](https://blogimages.builttoroam.com/OT-Visual-Studio-2019-Makes-Build-Output_D884/image_3.png), a message of this length may look "weird" in context.

    Is there a better way?

    @brendanzagaeski: Did we ever figure out how to sanely do multi-line error messages that integrate nicely in VS?

2. The message mentions "the AndroidManifest.xml" file.  Should we mention *which* `AndroidManifest.xml` file?

    Additionally, particularly in our new/current `manifestmerger.jar` world order (2c6f5cd1, f7ea4a3db003da7be85d34eff4813c6052368e90), there can be *multiple* `AndroidManifest.xml` files.  *Presumably* these other files won't be the cause of errors, but can we be absolutely certain?

    It may thus not be possible to *know* "which" `AndroidManifest.xml` file is the cause of the error either.

dellis1972: We should only deal with "actionable" files I think. Pointing the user to one of the Intermediate files does not help them as if they edit the file it will just be overwritten.  

The error links to the main Manifest file so when the user double clicks on it I the IDE that is the one which will be opened.

brendanzagaeski: > @brendanzagaeski: Did we ever figure out how to sanely do multi-line error messages that integrate nicely in VS?

As far as I know, it's fine to use line breaks within the string for a single call to `LogCodedWarning()` or `LogCodedError()` if desired. So for example, it would be OK to do something like:

```
<value>{0}.

This error is likely caused by an issue with the AndroidManifest.xml file or an Android manifest generation attribute in a source code file.</value>
```

This appears in the command line output like:

```
C:\Temp\Solution1\AndroidXAppSingleView1\Properties\AndroidManifest.xml : error APT2260: resource mipmap/ic_launche (aka com.companyname.androidappsingleview1:mipmap/ic_launche) not found.
C:\Temp\Solution1\AndroidXAppSingleView1\Properties\AndroidManifest.xml : error APT2260: 
C:\Temp\Solution1\AndroidXAppSingleView1\Properties\AndroidManifest.xml : error APT2260: This error is likely caused by an issue with the AndroidManifest.xml file or an Android manifest generation attribute in a source code file.
```

And it appears in Visual Studio like:

image

Compared to no line breaks:

image

jonpryor: Please add line breaks.

dellis1972: done

dellis1972:
Test failures are BCL related. There is one JLO_Change which is slower `Exceeded expected time of 10000ms, actual 10395.979ms` but I suspect that is not this PR.

PR #4618, opened 2020-04-24
[WIP] Interpreted apps

grendello:
Doesn't work yet - XA runtime is not placed in the `interpreter-` dirs yet

### DIFF
--- a/build-tools/xaprepare/xaprepare/ConfigAndData/Runtimes.cs
+++ b/build-tools/xaprepare/xaprepare/ConfigAndData/Runtimes.cs
@@ -10,21 +10,49 @@ namespace Xamarin.Android.Prepare
 		public readonly List<Runtime> Items = new List <Runtime> {
 			new MonoJitRuntime (
 				abiName: AbiNames.TargetJit.AndroidArmV7a,
+				interpreter: false,
+				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidArmV7a)
+			),
+
+			new MonoJitRuntime (
+				abiName: AbiNames.TargetJit.AndroidArmV7a,
+				interpreter: true,
 				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidArmV7a)
 			),
 
 			new MonoJitRuntime (
 				abiName: AbiNames.TargetJit.AndroidArmV8a,
+				interpreter: false,
+				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidArmV8a)
+			),
+
+			new MonoJitRuntime (
+				abiName: AbiNames.TargetJit.AndroidArmV8a,
+				interpreter: true,
 				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidArmV8a)
 			),
 
 			new MonoJitRuntime (
 				abiName: AbiNames.TargetJit.AndroidX86,
+				interpreter: false,
+				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidX86)
+			),
+
+			new MonoJitRuntime (
+				abiName: AbiNames.TargetJit.AndroidX86,
+				interpreter: true,
 				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidX86)
 			),
 
 			new MonoJitRuntime (
 				abiName: AbiNames.TargetJit.AndroidX86_64,
+				interpreter: false,
+				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidX86_64)
+			),
+
+			new MonoJitRuntime (
+				abiName: AbiNames.TargetJit.AndroidX86_64,
+				interpreter: true,
 				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidX86_64)
 			),
 
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -196,6 +196,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<_XAMajorVersionNumber>1</_XAMajorVersionNumber>
 	<_XASupportsFastDev Condition=" Exists ('$(MSBuildThisFileDirectory)Xamarin.Android.Common.Debugging.targets') ">True</_XASupportsFastDev>
 	<_XASupportsFastDev Condition=" '$(_XASupportsFastDev)' == '' ">False</_XASupportsFastDev>
+	<_AndroidUseInterpreter Condition=" '$(_AndroidUseInterpreter)' == '' ">False</_AndroidUseInterpreter>
 	<AndroidApplication Condition="'$(AndroidApplication)' == ''">false</AndroidApplication>
 	<AndroidUseSharedRuntime Condition=" '$(AndroidUseSharedRuntime)' == '' Or '$(_XASupportsFastDev)' == 'False' ">False</AndroidUseSharedRuntime>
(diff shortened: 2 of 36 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2179,29 +2181,31 @@ because xbuild doesn't support framework reference assemblies.
...
   <ItemGroup>
     <_TargetLibDir Include="$(MSBuildThisFileDirectory)lib\%(_TargetArchitecture.Identity)" Condition=" Exists('$(MSBuildThisFileDirectory)lib\%(_TargetArchitecture.Identity)') "/>
+	<_TargetLibInterpreterDir Include="$(MSBuildThisFileDirectory)lib\$(_TargetInterpreterPrefix)%(_TargetArchitecture.Identity)" Condition=" Exists('$(MSBuildThisFileDirectory)lib\$(_TargetInterpreterPrefix)%(_TargetArchitecture.Identity)') "/>
   </ItemGroup>
   <SplitProperty Value="$(AndroidEmbedProfilers)" Condition=" '$(AndroidEmbedProfilers)' != '' ">
     <Output TaskParameter="Output" ItemName="_EmbedProfilers" />
   </SplitProperty>
   <ItemGroup Condition=" '$(UsingAndroidNETSdk)' != 'True' ">
     <AndroidNativeLibrary Include="%(_TargetLibDir.Identity)\libsqlite3_xamarin.so" Condition="$(_Assemblies.Contains('Mono.Data.Sqlite.dll'))" />
-    <AndroidNativeLibrary Include="%(_TargetLibDir.Identity)\libMonoPosixHelper.so" Condition="$(_Assemblies.Contains('Mono.Posix.dll'))" />
```
dellis1972, line 2194: @grendello I'm confused.. so we not need these files when not using the interpreter?

grendello: We do, but they come from a different directory

jonpryor:
`$(_AndroidUseInterpreter)` is a "private" property.  The question: what property & semantics should we have for this to be public?

Options:

  * Make it public: `$(AndroidUseInterpreter)` boolean property
  * Add to `$(AndroidAotMode)`?
  * New enum property: `$(AndroidRuntimeModes)`
  * Something else?

### `$(AndroidAotMode)`

`$(AndroidAotMode)` is an "enum" property with the values: `None`, `Normal`, `Hybrid`, `Full`.  Could we add an `Interpreter` value?

While this could be done, it was felt that this was exposing an "implementation detail" -- that while an "interpreter" value is used with `mono_jit_set_aot_mode()`, that need not be the case for all time.

The other problem is that we'll invariably want to be able to specify *both* interpreter and an AOT mode.  A separate `$(AndroidUseInterpreter)` property works, or we could turn `$(AndroidAotMode)` into a comma-separated "flags enum" property as described in `$(AndroidRuntimeModes)`.

##  `$(AndroidRuntimeModes)`

A possible scenario with more "robust" interpreter support is that Mono could support "tiered JIT" use: interpreter for starters, then JIT frequently used methods.

This suggests we could have a "`[Flags]`-enum like" comma-separated property value, being one or more of:

  * `Aot`
  * `Jit`
  * `Interpreter`

This allows combinations of:

  * `Aot`: `$(AndroidAotMode)`=Full [not currently supported]
  * `Jit`: aka `$(AndroidAotMode)`=None [or Normal?]; current default.
  * `Interpreter`: *only* interpreter, no JIT
  * `Interpreter,Jit`: Tiered JIT use, if/when that is ever supported.  
  * `Aot,Jit`: `$(AndroidAotMode)`=Normal or Hybrid [ambiguous; suggests this entire idea needs more thought]
  * `Aot,Interpreter`: `$(AndroidAotMode)`=Normal or Hybrid [also ambiguous]
  * `Aot,Interpreter,Jit`: All The Things! Not sure if this combination actually makes sense, but if it did, it would be tiered JIT for non-AOT'd methods.

On the one hand, this could plausibly "subsume" `$(AndroidAotMode)` allowing for a nice "semantic cleanup," but on the other hand such "subsume-ation" makes for potential confusion about priority and ordering of properties when *both* `$(AndroidRuntimeModes)` and `$(AndroidAotMode)` are set.  It also contains ambiguities around Hybrid vs. Full AOT.

grendello:
@jonp I'd use `$(AndroidUseInterpreter)` as it doesn't expose any technicalities to the users.

build-tools/xaprepare/xaprepare/ConfigAndData/Runtimes.cs
```diff
@@ -10,21 +10,49 @@ partial class Runtimes
...
 				abiName: AbiNames.TargetJit.AndroidX86,
+				interpreter: false,
 				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidX86)
 			),
 
+			new MonoJitRuntime (
+				abiName: AbiNames.TargetJit.AndroidX86,
+				interpreter: true,
+				enabledCheck: (Context ctx) => ctx.IsTargetJitAbiEnabled (AbiNames.TargetJit.AndroidX86)
+			),
```
lewurm, line 45: Alas we do not support the interpreter on `x86`.

Is this fairly common? Or does everyone use `x86_64` anyway as emulator?

src/monodroid/jni/android-system.hh
```diff
@@ -125,6 +137,14 @@ namespace xamarin::android::internal
 		long max_gref_count = 0;
 		MonoAotMode aotMode = MonoAotMode::MONO_AOT_MODE_NONE;
 		bool running_in_emulator = false;
+
+		// This is a hack because of the way Mono currently switches the full interpreter (no JIT) mode. In Mono
+		// **internal** headers there's an AOT mode macro, `MONO_EE_MODE_INTERP`, whose value is exactly the same as
+		// MonoAotMode::MONO_AOT_MODE_LAST.  However, we use `MonoAotMode::MONO_AOT_MODE_LAST` as a sentinel to indicate
+		// that we want to use the default Mono AOT/JIT mode and so we can't "overload" it to mean something else for
+		// the sake of using Mono's internal functionality.  Until Mono makes `MONO_EE_MODE_INTERP` part of the public
+		// `MonoAotMode` enum and its value is not in conflict with the sentinel, we will use this hack.
```
lewurm, line 146: might wanna add a link to this issue: https://github.com/mono/mono/issues/18893

build-tools/xaprepare/xaprepare/ConfigAndData/Runtimes.cs
grendello: We actually use `x86` emulators on CI...

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
dellis1972: So should we be including them from the old directory when not using the interpreter?

grendello: We are - look at the diff above this section - it conditionally includes the interpreter path in `$(_TargetLibInterpreterDir)` which is otherwise identical to `$(_TargetLibDir)` and has the same value if the interpreter isn't used.

build-tools/xaprepare/xaprepare/ConfigAndData/Runtimes.cs
lewurm: I meant with "everyone" the average Xamarin.Android developer 🙂

grendello: Developers will most definitely run emulators using either `x86` or `x86_64` images, but I don't know which one of those is more popular. Both are accelerated and `x86_64` can run `x86` apps, so I'd **hope** that this one's more popular, but that's just my hope not supported by any data :)

lewurm: ok, I created https://github.com/mono/mono/issues/19669

jonathanpeppers: I think most devs use x86 right now as Visual Studio suggests it by default.

JonDouglas: @lewurm x86 is like 90% with x86 in the < 10% range. ARM even lower here.

JonDouglas: ABI/CPU - 

88/10/3/2% for x86/x86_64/armeabi-v7a/arm64-v8a on Mac. 

90/7/1/1% for x86/x86_64/armeabi-v7a/arm64-v8a on Windows.

PR #4619, opened 2020-04-24
[.NET 5] $(AndroidBoundExceptionType)=System

jonpryor:
Commit 60363ef4 added support for the `$(AndroidBoundExceptionType)`
MSBuild property, which controls whether or not Java exception types
are wrapped in appropriate .NET exception types from ".NET methods"
such as `IList.Add()` via `Android.Runtime.JavaList`/etc.

By default `$(AndroidBoundExceptionType)`=Java, meaning "don't wrap",
and `System` can be used to enable exception wrapping.

Commit 60363ef4 mentioned:

> [When .NET 5 is released with Xamarin.Android support][1], the
> default value will be `System`.

Do it!  Within .NET 5, `$(AndroidBoundExceptionType)`=System,
by default.

### DIFF
--- a/Documentation/guides/DotNet5.md
+++ b/Documentation/guides/DotNet5.md
@@ -35,6 +35,13 @@ In .NET 5 the behavior of the following MSBuild tasks will change, but
 
 `$(AndroidUseIntermediateDesignerFile)` will be `True` by default.
 
+`$(AndroidBoundExceptionType)` will be `System` by default.  This will
+[alter the types of exceptions thrown from various methods][abet-sys] to
+better align with existing .NET 5 semantics, at the cost of compatibility with
+previous Xamarin.Android releases.
+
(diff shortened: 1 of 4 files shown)


brendanzagaeski:
As a caution https://github.com/xamarin/xamarin-android/issues/4127 has not been addressed. I was hesitant to recommend `AndroidBoundExceptionType`=`System` to users until that item was resolved because the new set of exception types would change again if that item is fixed, meaning users would need to update their `catch` blocks again for the updated exception types.

brendanzagaeski:
*Draft release notes adjustment*

I previously didn't recognize that `Microsoft.Android.Sdk.props` was only applicable to projects that are importing the new `.targets` for .NET 5. Since that is the case, this PR does not change behavior of existing projects, so it doesn't need its own release note after all. Information about this change has instead been added to the guide about .NET 5 changes.

PR #4620, opened 2020-04-24
[build] bump to dotnet 5.0.100-preview.3.20216.6

jonathanpeppers:
This doesn't give us any new functionality, but I've noticed the
MSBuild `Node-2` tests are hanging. These are the ones that run the
`XASdkTests`, which I think is suspicious.

It's worth updating anyway, since preview 1 is from March 16.


jonathanpeppers:
This still hangs, we probably don't need to bump this until Preview 4.

PR #4621, opened 2020-04-24
[tests] mark hanging test [NonParallelizable]

jonathanpeppers:
Hanging test prints:

    [TESTLOG] Test CheckClassIsReplacedWithMd5 Outcome=Passed

This started happening since 835c59bc, but I don't see how the commit
would cause it.

Working build (https://build.azdo.io/3669040) prints this after
`CheckClassIsReplacedWithMd5`:

    => Xamarin.Android.Build.Tests.ResolveSdksTaskTests.LatestTFV_OldTargetSdkVersion

After adding logging, etc. I couldn't figure out if this was a bug in
NUnit or our tests. Adding `[NonParallelizable]` for the entire
`ResolveSdksTaskTests` class seems to solve the hang, so I think we
should just go with this for now.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ResolveSdksTaskTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ResolveSdksTaskTests.cs
@@ -13,7 +13,7 @@ namespace Xamarin.Android.Build.Tests {
 
 	[TestFixture]
 	[Category ("Node-2")]
-	[Parallelizable (ParallelScope.Self)]
+	[NonParallelizable] // NOTE: This test was hanging without this
 	public class ResolveSdksTaskTests : BaseTest {
 #pragma warning disable 414
 


jonathanpeppers:
Didn't hang so retry once ^^

If it works a second time, I'll close this, I guess.

jonathanpeppers:
Ok, it hung this time, yay! yay?
```
=> Xamarin.Android.Build.Tests.ConvertResourcesCasesTests.CheckClassIsReplacedWithMd5
TestSetup Begin
[TESTLOG] Test CheckClassIsReplacedWithMd5 Starting
TestSetup End
Message:   AndroidConversionFlagFile modified: 1/1/0001 12:00:00 AM
Message:   Processing: /Users/runner/runners/2.166.4/work/1/s/bin/TestRelease/temp/CheckClassIsReplacedWithMd5/res/layout/main.xml   4/28/2020 5:35:57 PM > 1/1/0001 12:00:00 AM
Message: Fixed up Custom Views in /Users/runner/runners/2.166.4/work/1/s/bin/TestRelease/temp/CheckClassIsReplacedWithMd5/res/layout/main.xml
Message: compiled compiled.stamp
Message: compiled compiled.stamp
CleanupTest Begin
[TESTLOG] Test CheckClassIsReplacedWithMd5 Complete
[TESTLOG] Test CheckClassIsReplacedWithMd5 Outcome=Passed
```
I don't see `CleanupTest End`, but I forgot to put it in a try-finally block.

It is weird I don't see `ResolveSdksTaskTests` in the log output at all? This makes me wonder if there is some NUnit bug?

jonathanpeppers:
It seems like this change prevented the hang ~3 times in a row, so maybe we should go with this for now.

PR #4626, opened 2020-04-27
OSS PR Mac build stage

mjbond-msft:
Add a Mac build stage to the OSS pipeline to serve as a replacement for the Jenkins OSS PR build:
https://jenkins.mono-project.com/view/Xamarin.Android/job/xamarin-android-pr-pipeline-release/

Plan is to use the existing `Xamarin.Android-OSS` pipeline definition, which will include both the Mac and Linux build stages:
https://dev.azure.com/xamarin/public/_build?definitionId=48&_a=summary

Testing has been performed against the following `Xamarin.Android-OSS-PR` pipeline definition that will be deprecated once this PR merges to master:
https://dev.azure.com/xamarin/public/_build?definitionId=52&_a=summary

**Note:** This change includes support for Release builds only and not Debug builds.  Debug builds could be added as parallel stages

Also includes the following fixes:
 - Produce the `create-all-packs.binlog` in the expected location (927b3aa3d62e6e6b4967bcee87e63720c7041e1a)
 - Avoid `prepare-update-mono` build breaks when building `xaprepare.csproj` on newly reimaged agents (a378aebc67e61bbf43589ed60646879088616249)
 - Remove the duplicate step to produce plots for Locale (873dcf4fae336a23403007656d5732865fd36c5d)


pjcollins:
One issue we previously discussed with having multiple pipelines across projects is [described here](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#create-pipelines-in-multiple-azure-devops-organizations-and-projects):
> if you create pipelines for a single repository in multiple Azure DevOps organizations, only the first organization's pipelines can be automatically triggered by GitHub commits or pull requests. Manual or scheduled builds are still possible in secondary Azure DevOps organizations.

We should potentially try to investigate a way around this limitation as part of this PR. Perhaps using a GitHub action to trigger the OSS pipeline would work? https://github.com/marketplace/actions/azure-pipelines-action

mjbond-msft:
@pjcollins the limitation you cited continues to exist. Azure DevOps prescribes a one-to-one mapping from a Azure DevOps org to a Xamarin org.  However, in the case of Azure DevOps we're split across two orgs: DevDiv and Xamarin.  Consequently, we cannot use the xamarin github service connection.

I found that we can use a github token-based service connection works just fine; particularly, the following features all work:
 - PR & CI triggers
 - GitHub status updates based on the `Report build status` setting
 - Rebuild for PR & CI builds

build-tools/automation/azure-pipelines-oss.yaml
```diff
@@ -2,66 +2,186 @@
...
+
+stages:
+- stage: mac_stage
+  displayName: Mac
+  condition: eq(variables['EnableMacStage'], 'true')  # The variable is defined on the pipeline definition
+  jobs:
+  - job: mac_build
+    displayName: Mac Build
+    pool:
+      name: $(XA.Build.MacOSSPool)  # The variable is defined on the pipeline definition
```
pjcollins, line 37: Could we move the pool names back into YAML? I think it's easier to test future changes to the pool we're using when the pool is defined in this version-able file rather than having it defined in the static pipeline settings.

build-tools/automation/azure-pipelines-oss.yaml
```diff
@@ -2,66 +2,186 @@
...
 name: $(Build.SourceBranchName)-$(Build.SourceVersion)-$(Rev:r)
 
-trigger:
-  - master
-  - d16-*
+# Triggers are enabled on the pipeine definition in place of being defined here
+#  https://dev.azure.com/xamarin/public/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=48&view=Tab_Triggers
+# trigger:
+#  - master
+#  - d16-*
```
pjcollins, line 9: Do we need to override these YAML defined triggers? Is this what allows us to get automatic triggers in both Azure DevOps organizations?

build-tools/automation/azure-pipelines-oss.yaml
```diff
@@ -2,66 +2,186 @@
...
+    workspace:
+      clean: all
+    steps:
+    - checkout: self
+      submodules: recursive
+
+    - script: |
+        # Report used vs. available disk space on each volume
+        df -H
+      displayName: report disk info
```
pjcollins, line 148: I think we can remove this script, I'm working to address this issue in https://github.com/xamarin/xamarin-android/pull/4674

build-tools/automation/azure-pipelines-oss.yaml
mjbond-msft: Agree that it would be better to include the pool names within the YAML file itself.  Addressed with commit bc68d05623c2b643fd3b629e1b2797ab9f7290c0

build-tools/automation/azure-pipelines-oss.yaml
mjbond-msft: Agree that incorporating the trigger settings within the YAML file itself provides greater flexibility and documention.  Addressed with commit f2783df6fb499a19a4da9742ec2981dd15bdd6e6

build-tools/automation/azure-pipelines-oss.yaml
mjbond-msft: Removed. Commit: e6d8876378c36b290aadbbcfcd7ad9058307b0ef

build-tools/automation/azure-pipelines-oss.yaml
```diff
@@ -3,65 +3,188 @@
...
-    inputs:
-      artifactName: Linux Packages
-      targetPath: $(System.DefaultWorkingDirectory)/bin/Build$(OSSBuildConfiguration)/linux-artifacts
-
-  - template: yaml-templates/upload-results.yaml
-    parameters:
-      configuration: $(OSSBuildConfiguration)
-      artifactName: OSS Build Results - Linux
+  EnableMacStage: true
+  EnableLinuxStage: false
```
dellis1972, line 20: should this value be true?

pjcollins: We're working on fixing this build in https://github.com/xamarin/xamarin-android/pull/4674. I think it's best to keep this disabled until that PR is working since the linux build has been broken for a while.

dellis1972: ok no worries.

build-tools/automation/azure-pipelines-oss.yaml
```diff
@@ -3,65 +3,188 @@
...
+    - template: yaml-templates/use-dot-net.yaml
+      parameters:
+        version: $(DotNetCorePreviewVersion)
+
+    - template: yaml-templates/use-dot-net.yaml
+      parameters:
+        version: $(DotNetCoreVersion)
+
+    # Prepare and build everything
+    - script: |
```
jonpryor, line 66: Given d3e0fc831b98b5481f220fcd9c0967d096bf6ee8, I'm wary of using `script: |`, and changing this to use `script: >` should be straightforward.

build-tools/automation/azure-pipelines-oss.yaml
```diff
@@ -3,65 +3,188 @@
...
+        msbuildArguments: /t:CreateAllPacks /restore /bl:$(Build.SourcesDirectory)/bin/Build$(XA.Build.Configuration)/create-all-packs.binlog
+
+    - task: PublishPipelineArtifact@1
+      displayName: upload nupkgs
+      inputs:
+        artifactName: $(NuGetArtifactName)
+        targetPath: $(Build.SourcesDirectory)/bin/Build$(XA.Build.Configuration)/$(NuGetArtifactName)
+
+    - script: |
+        exitCode=0
```
jonpryor, line 123: I don't understand why you need this pattern.  `script:` blocks will fail if the last command executed fails, so all that this `script:` block *should* need is:

```
- script: >
  make run-performance-tests CONFIGURATION=$(XA.Build.Configuration) V=1
```

If it fails, then the entire script step *also* fails.  There shouldn't be a need for everything else.

build-tools/xaprepare/xaprepare/xaprepare.csproj
```diff
@@ -8,7 +8,7 @@
     <RootNamespace>Xamarin.Android.Prepare</RootNamespace>
     <AssemblyName>xaprepare</AssemblyName>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <LangVersion>latest</LangVersion>
+    <LangVersion>preview</LangVersion>
```
jonpryor, line 11: If this needs to change, it should probably be changed to `8.0`, not `preview`.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -188,7 +188,7 @@ stages:
       inputs:
         solution: $(System.DefaultWorkingDirectory)/xamarin-android/build-tools/create-packs/Microsoft.Android.Sdk.proj
         configuration: $(XA.Build.Configuration)
-        msbuildArguments: /t:CreateAllPacks /restore /bl:$(System.DefaultWorkingDirectory)/bin/Build$(XA.Build.Configuration)/create-all-packs.binlog
+        msbuildArguments: /t:CreateAllPacks /restore /bl:$(System.DefaultWorkingDirectory)/xamarin-android/bin/Build$(XA.Build.Configuration)/create-all-packs.binlog
```
jonpryor, line 191: Why the insertion of `xamarin-android/`?

build-tools/automation/azure-pipelines-oss.yaml
mjbond-msft: Changed to `script: >` per 4244cb01c8ff5b9ec43149f4da76ec982427e877

build-tools/automation/azure-pipelines-oss.yaml
mjbond-msft: Ah, yes. This was a literal translation of the Jenkins groovy script here: https://github.com/xamarin/xamarin-android/blob/3f438e46d7b166a3a3ef54c9ffafb5f426760468/build-tools/automation/build.groovy#L304-L306.  Was needed to ensure the Jenkins pipeline stage result was set correctly. In the Azure DevOps world, the same logic isn't needed anymore and we can just rely on the pass/fail of the command. Fixed with commit f6b997c39d8fa35aed2c5f44d0ca86c1ff7c82f8

build-tools/automation/azure-pipelines.yaml
pjcollins: This path was incorrect before. `$(DefaultWorkingDirectory)` changes based on the jobs usage of the multi-repo checkout feature, and in this case the extra `xamarin-android` is needed to make sure this log ends up in the right place.

mjbond-msft: indeed. and i just confirmed with this change that the `xa-build-status` zip artifact now includes `create-all-packs.binlog`

build-tools/xaprepare/xaprepare/xaprepare.csproj
mjbond-msft: 8.0 it is then. b5221c9a5f24e8d612b2379ac33bcff3c9949253

PR #4622, opened 2020-04-25
LOC CHECKIN | xamarin/xamarin-android d16-6 | 20200424

brendanzagaeski:
Add all of the latest `.xlf` file changes from the 20200424 localization drop except for the changes related to XA0031 because those are for https://github.com/xamarin/xamarin-android/commit/231bf2a4381af4996c795e27fa65b15d63e3fc0b, which is only present on master.

Cherry-pick all of the corresponding `.resx` file changes from master.

This leaves two `.resx` file changes on master that are not yet cherry-picked because they do not yet have any corresponding translations in the `.xlf` files: 3fee2438 and ef12d24e.

PR #4627, opened 2020-04-27
Try a build with no changes

jonpryor:
Context: https://github.com/xamarin/xamarin-android/pull/4622

PR #4622 is consistently failing in the **APK Instrumentation** step
because `javac` wasn't found (?!).  Is this specific to PR #4622?
Or is d16-6 just "broken in general"?

### DIFF
--- a/build-tools/automation/yaml-templates/setup-test-environment.yaml
+++ b/build-tools/automation/yaml-templates/setup-test-environment.yaml
@@ -20,15 +20,27 @@ steps:
 
 - script: |
     mono build-tools/xaprepare/xaprepare/bin/${{ parameters.configuration }}/xaprepare.exe --s=UpdateMono --auto-provision=yes --auto-provision-uses-sudo=yes --no-emoji --run-mode=CI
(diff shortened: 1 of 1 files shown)


jonpryor:
My grep is wrong; I should have searched for `Initializing scenario`:

```bash
% grep 'Initializing scenario' prepare-20200427T203*
prepare-20200427T203008.log:[00:00:00.2793986] Initializing scenario UpdateMono
prepare-20200427T203116.log:[00:00:00.4118305] Initializing scenario Required
prepare-20200427T203230.log:[00:00:00.2998047] Initializing scenario AndroidToolchain
```

This *is* https://github.com/MicrosoftDocs/azure-devops-docs/issues/3265 : `xaprepare -s=Required` fails, but the failure is "eaten", and execution continues.

Thus the followup: why's it fail on d16-6 and succeed on master?  The answer is c69bb31d987ea1aac24972452c4b233bec0029ac: we need to update the MinGW compiler version to appease Homebrew, which would allow `xaprepare -s=Required` to complete successfully.

PR #4629, opened 2020-04-28
[xaprepare] Install essential utilities in the `Required` scenario

grendello:
Context: a5748347da50ea002ea7288821c8813fb82be4fa

a5748347 made `xaprepare` ignore "essential" tools if they are missing
from the system, so that certain scenarios (currently `UpdateMono`) can
complete their task even without the essentials being
installed (provided the scenario doesn't actually need the tool, of
course).  However, what was missing is the part to actually install the
tools, since they're not going to magically appear on the host we're
currently running on.  This, however, requires a task that can both
ignore missing essential tools *and* install them (the idea behind
essential tools is that the scenarios may blindly assume the tools
exist, thus making the code simpler).  One such task is the `Required`
scenario which ensures that all necessary bits are there.  However, even
with a5748347 the scenario would fail as it would expect the essentials
to be there:

    Required program '7za' could not be found
    System.InvalidOperationException: Required program '7za' could not be found
      at Xamarin.Android.Prepare.OS.Which (System.String programPath, System.Boolean required)
      at Xamarin.Android.Prepare.EssentialTools.Init (Xamarin.Android.Prepare.Context context)
      at Xamarin.Android.Prepare.Context+<Init>d__201.MoveNext ()
      at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess (System.Threading.Tasks.Task task)
      at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification (System.Threading.Tasks.Task task)
      at System.Runtime.CompilerServices.TaskAwaiter.ValidateEnd (System.Threading.Tasks.Task task)
      at System.Runtime.CompilerServices.TaskAwaiter`1[TResult].GetResult ()
      at Xamarin.Android.Prepare.App+<Run>d__3.MoveNext ()

This commit fixes the problem by making the `Required` scenario capable
of installing the essentials that are missing (applies to `7zip` and
`git` as other programs like `pkgutil` and `brew` are real prerequisites
which must exist before `xaprepare` runs)

### DIFF
--- a/build-tools/xaprepare/xaprepare/Application/EssentialTools.cs
+++ b/build-tools/xaprepare/xaprepare/Application/EssentialTools.cs
@@ -25,7 +25,7 @@ namespace Xamarin.Android.Prepare
 
 		public void Init (Context context)
 		{
-			bool require = context.CheckCondition (KnownConditions.AllowProgramInstallation);
+			bool require = !context.CheckCondition (KnownConditions.EnsureEssential) && context.CheckCondition (KnownConditions.AllowProgramInstallation);
 
 			Log.StatusLine ();
 			Log.StatusLine ("Locating essential tool binaries", ConsoleColor.DarkGreen);
--- a/build-tools/xaprepare/xaprepare/Application/KnownConditions.cs
+++ b/build-tools/xaprepare/xaprepare/Application/KnownConditions.cs
@@ -24,5 +24,11 @@ namespace Xamarin.Android.Prepare
 		///   the top of the repository. Default: unset
 		/// </summary>
 		IncludeCommercial,
+
+		/// <summary>
+		///   If this condition is set, the current scenario will take care of missing essential (<see
+		///   cref="EssentialTools"/>) programs should they be missing.
+		/// </summary>
+		EnsureEssential,
 	}
 }
--- a/build-tools/xaprepare/xaprepare/Scenarios/Scenario_Required.cs
+++ b/build-tools/xaprepare/xaprepare/Scenarios/Scenario_Required.cs
@@ -18,6 +18,9 @@ namespace Xamarin.Android.Prepare
 			if (context == null)
 				throw new ArgumentNullException (nameof (context));
 
+			// Install essential tools, should they be missing
+			context.SetCondition (KnownConditions.EnsureEssential, true);
+
 			Steps.Add (new Step_GenerateFiles (atBuildStart: true, onlyRequired: true));
 			Steps.Add (new Step_PrepareExternalJavaInterop ());
 		}


jonpryor:
Do we need to rethink this/potentially *revert* a5748347da50ea002ea7288821c8813fb82be4fa?

What was the *original* problem that a5748347da50ea002ea7288821c8813fb82be4fa tried to address?

The original issue was that we'd see unit test failures on CI within the **Test** > **APK Instrumentation** job, [e.g.](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3679916&view=logs&j=8556562a-ae5f-5bd1-7c4d-bf1af4b6f1e1&t=5076e147-fc66-561e-6c69-3aa777afefc5):

```
  /Users/runner/runners/2.166.4/work/1/s/external/Java.Interop/src/Java.Interop/Java.Interop.targets(37,5): error MSB3073: The command """ -source 1.6 -target 1.6 -bootclasspath "" -d "obj/Release/ji-classes" java/com/xamarin/java_interop/internal/JavaProxyObject.java java/com/xamarin/java_interop/internal/JavaProxyThrowable.java java/com/xamarin/java_interop/GCUserPeerable.java java/com/xamarin/java_interop/ManagedPeer.java" exited with code 127. [/Users/runner/runners/2.166.4/work/1/s/external/Java.Interop/src/Java.Interop/Java.Interop.csproj]
```

The above error made no sense, until we discovered the existence of https://github.com/MicrosoftDocs/azure-devops-docs/issues/3265.

In short, if you have a YAML `script` block that executes multiple commands, *any* of the commands can fail while the entire `script` block is "successful".  I xamarin-android CI terms, `xaprepare -s=UpdateMono` and `xaprepare -s=Required` *failed*, which in turn meant that `make -C external/Java.Interop prepare` was never executed, which is why `$(JavaCPath)` wasn't set, which is why we saw the above failure.

(Why'd `xaprepare` fail?  Because MinGW was "too old", and we fixed this via PR #4438.)

Marking certain binaries as "not essential", as a5748347da50ea002ea7288821c8813fb82be4fa does, was just a *bandaid* around the original problem regarding a *lack of error propagation* in the YAML `script` block in CI.

Instead of continuing down the path of a5748347da50ea002ea7288821c8813fb82be4fa, let's *revert* it: *all* required binaries are in fact required, because if they're not required we'll just fail later *anyway* (when it turns out the "required but not required enough to install" utilities are actually required, and trying to use them results in a failure).

Instead, we should consider "directly" working around MicrosoftDocs/azure-devops-docs#3265 and place each command invocation into a separate YAML `script` block, on the (hopefully correct?) assumption that it *at least* will check the exit status of the *last* command executed.  (If that's not the case, then we need to avoid/remove all `script` blocks entirely, as they're unreliable.)

I took a stab at this here:

https://github.com/xamarin/xamarin-android/pull/4627/commits/af0d40dc92e536998754556dffd6a1ef1f86b336

Note: commit message in ^^ needs re-editing, as it was written before we remembered about the azure-devops-docs issue.

jonpryor:
Alternate approach: PR #4634

grendello:
@jonp the original problem a574834 fixed was that `7za` was missing even though it was treated as a pre-requisite - we had a catch-22 problem: `prepare-update-mono` insisted on `7za` being there but it wasn't and since the required programs check ran before the scenario started, xaprepare bailed out before it ran the scenario. So both a574834 and this PR (which fixes the same issue with the `Required` scenario, the difference being here that `Required` will actually install the missing programs) are still needed.

PR #4633, opened 2020-04-29
Remove the LaunchProfiles capability for the CPS integration

kzu:
Implements https://work.azdo.io/1112733 as a workaround for the conflicts between 
the built-in launchsettings.json-based .NET Core debugger and our Mono debugger.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
@@ -27,6 +27,10 @@
     <ProjectCapability Include="Mobile" />
     <ProjectCapability Include="Android" />
     <ProjectCapability Include="AndroidApplication" Condition="$(AndroidApplication)" />
+    
+    // See https://work.azdo.io/1112733
+    <!-- Conflicts with our targets generator in VS+CPS -->
+    <ProjectCapability Remove="LaunchProfiles" />
   </ItemGroup>
 
   <!-- NOTE: We have to replace the Run target after Microsoft.NET.Sdk.targets are imported -->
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
```diff
@@ -27,6 +27,10 @@
     <ProjectCapability Include="Mobile" />
     <ProjectCapability Include="Android" />
     <ProjectCapability Include="AndroidApplication" Condition="$(AndroidApplication)" />
+    
+    // See https://work.azdo.io/1112733
```
jonpryor, line 31: Not sure how MSBuild deals with "content" between items in an ItemGroup, but it's likely not a good idea, regardless.  This "comment" should be moved into the *actual* XML comment on the following line.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
```diff
@@ -27,6 +27,10 @@
...
+    <!-- Conflicts with our targets generator in VS+CPS -->
+    <ProjectCapability Remove="LaunchProfiles" />
```
jonathanpeppers, line 33:
```suggestion
    <!-- Conflicts with our targets generator in VS+CPS, see https://work.azdo.io/11127 -->
    <ProjectCapability Remove="LaunchProfiles" />
```
It's weird that raw text content was valid XML.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
kzu: doh!

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
kzu: it wasn't :(

PR #4635, opened 2020-04-29
[.NET 5] Update NuGet package version scheme

pjcollins:
Updates the version for all .nupkg files we are generating for .NET 5 to
follow the dotnet [SDK versioning scheme][0].  For more information, see
Documentation/guides/DotNet5.md.

Versioning for release candidates has not yet been automated.  This
commit only introduces the following two package version types:

    * 10.0.100-ci.branchname.commitdistance+sha.1b2c3d4e
    * 10.0.100-ci.pr.gh1234.commitdistance+sha.1b2c3d4f

Version file generation has been updated to generate new files for our
nupkgs.  Mono.Android assembly info generation has also been updated to
include different version information depending on $(TargetFramework).

[0]: https://github.com/dotnet/designs/blob/master/accepted/2018/sdk-version-scheme.md

### DIFF
--- a/Documentation/guides/DotNet5.md
+++ b/Documentation/guides/DotNet5.md
@@ -145,11 +145,11 @@ This is the package version scheme: `OS-Major.OS-Minor.InternalRelease[-prerelea
     * If we have a particular feature we want people to subscribe to (such as
       an Android preview release), we publish previews with a custom pre-release
       identifier:
-        * Example: `10.1.100-android_r.beta.1`
+        * Example: `10.1.100-android-r.beta.1`
         * This way people can sign up for only official previews, by
-          referencing `*-android_r.beta.*`
-        * It's still possible to sign up for all `android_r` builds, by
-          referencing `*-ci.android_r.*`
+          referencing `*-android-r.beta.*`
+        * It's still possible to sign up for all `android-r` builds, by
+          referencing `*-ci.android-r.*`
 * Build metadata: Required Hash
     * This is `sha.` + the short commit hash.
         * Use the short hash because the long hash is quite long and
@@ -162,8 +162,8 @@ This is the package version scheme: `OS-Major.OS-Minor.InternalRelease[-prerelea
       version string refers to a stable version or not.
         * Example: `10.0.100`: incomplete version
         * Example: `10.0.100+sha.1a2b3c`: stable
-        * Example: `10.0.100-ci.d17_0.1234+sha.1a2b3c`: CI build
-        * Example: `10.0.100-android_r.beta.1+sha.1a2b3c`: official
+        * Example: `10.0.100-ci.d17-0.1234+sha.1a2b3c`: CI build
+        * Example: `10.0.100-android-r.beta.1+sha.1a2b3c`: official
           preview
             * Technically it's possible to remove the prerelease part, but
               we’d still be able to figure out it’s not a stable version by
--- a/Documentation/guides/DotNet5.md
+++ b/Documentation/guides/DotNet5.md
@@ -106,7 +106,7 @@ The following instructions can be used for early preview testing.
     packages you want to use:
 
 ```xml
-<Project Sdk="Microsoft.Android.Sdk/10.4.99.24">
+<Project Sdk="Microsoft.Android.Sdk/10.0.100">
   <PropertyGroup>
     <TargetFramework>netcoreapp5.0</TargetFramework>
     <RuntimeIdentifier>android.21-arm64</RuntimeIdentifier>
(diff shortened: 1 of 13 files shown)


Documentation/guides/DotNet5.md
```diff
@@ -120,4 +120,58 @@ The following instructions can be used for early preview testing.
 dotnet publish -t:Install *.csproj
 ```
 
-[0]:  https://github.com/dotnet/installer#installers-and-binaries
+## Package Versioning Scheme
+
+Our NuGet packages are versioned using [Semver 2.0.0][2].
+
+This is the scheme: `OsMajor.OsMinor.InternalRelease[-prereleaseX]+sha.1b2c3d4`.
```
jonpryor, line 127: Suggest using `OS-Major` and `OS-Minor` instead of `OsMajor` and `OsMinor`; I find the latter ugly.

Documentation/guides/DotNet5.md
```diff
@@ -120,4 +120,58 @@ The following instructions can be used for early preview testing.
 dotnet publish -t:Install *.csproj
 ```
 
-[0]:  https://github.com/dotnet/installer#installers-and-binaries
+## Package Versioning Scheme
+
+Our NuGet packages are versioned using [Semver 2.0.0][2].
```
jonpryor, line 125: Another pet peeve of mine, but I would suggest *removing **all** mention* of SemVer.  It *can't* reasonably apply to us, as *we* have *no idea* when semantics breaks occur!

This is more noticeable in the "current world order" in which we bundle Mono, and Mono may import some blob of code from mono/referencesource, which then proceeds to break a bunch of previously expected *semantics*, e.g. https://docs.microsoft.com/en-us/xamarin/android/release-notes/10/10.0#mono-framework-version-update-to-64

Semantic compatibility is *hard*, and SemVer is *all about* ***semantic*** compatibility, and at a very *deep* level, we're so far removed from "semantically relevant" changes that we have *no* idea if some of our semantics have broken.

This might plausibly change in a .NET 5 world order, in that *we* are no longer redistributing the BCL, but I'm still very leery of the entire idea.

Documentation/guides/DotNet5.md
```diff
@@ -120,4 +120,58 @@ The following instructions can be used for early preview testing.
...
+        * This way people can sign up for only official previews, by
+          referencing `Android *-android_q.beta.*`
+        * It's still possible to sign up for all `android_q` builds, by
+          referencing `Android *-ci.android_q.*`
+* Build metadata: Required Hash
+    * This is `sha.` + the short commit hash.
+        * Use the short hash because the long hash is quite long and
+          cumbersome. This leaves the complete version open for duplication,
+          but this is extremely unlikely.
+    * Example: `Android 10.0.100+sha.1a2b3c`
```
jonpryor, line 160: Shouldn't the examples use `Microsoft.Android.Sdk` instead of `Android`?

build-tools/scripts/XAVersionInfo.targets
```diff
@@ -40,7 +40,7 @@
   <Target Name="GetXAVersionInfo"
       DependsOnTargets="_GetSubmodulesVersionInfo">
     <GitBlame
-        FileName="Configuration.props"
```
jonpryor, line 43: A problem with this change is that it'll *reset* our current commit distances!  Once this is merged, instead of `XA 10.4.99.51` (as of 42822e0488185cdf4bca7c0bd05b21ad03dfbd7e; see https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3687020&view=artifacts&type=publishedArtifacts), it'll become `XA 10.4.99.0`.

If we're going to change the file that contains `$(ProductVersion)`, then we need to *also* bump `$(ProductVersion)` to e.g. `10.4.100.0`, so that "new" master builds don't go "backwards".

Documentation/guides/DotNet5.md
pjcollins: This versioning information applies to all targeting, runtime, and SDK packs. I am not sure what the best naming choice here would be. This content is pretty much a direct replica of https://github.com/xamarin/xamarin-macios/commit/2d6fc9aaad4810b791dbafb4f927834c1451eb5e#diff-df8aabef466d05fbc410668620f6c412R1.

Maybe we should just omit the name prefix?

build-tools/scripts/XAVersionInfo.targets
```diff
@@ -68,6 +83,10 @@
     </GitBranch>
     <PropertyGroup>
       <XARepositoryName Condition=" '$(XARepositoryName)' == '' ">xamarin-android</XARepositoryName>
+      <!-- See Azure Pipelines predefined variables. -->
+      <_AndroidPackLabel Condition=" '$(SYSTEM_PULLREQUEST_PULLREQUESTNUMBER)' != '' ">ci.pr.gh$(SYSTEM_PULLREQUEST_PULLREQUESTNUMBER).$(PackVersionCommitCount)</_AndroidPackLabel>
+      <_AndroidPackLabel Condition=" '$(_AndroidPackLabel)' == '' ">ci.$(XAVersionBranch).$(PackVersionCommitCount)</_AndroidPackLabel>
```
rolfbjarne, line 88: It doesn't look like `$(XAVersionBranch)` is limited to alphanumeric characters.

pjcollins:
@jonathanpeppers thanks, I forgot about updating the push behavior. We'll now push all non PR builds by default, while still allowing PR builds to be pushed if a build is queued with a particular variable set.

build-tools/scripts/XAVersionInfo.targets
pjcollins: Thanks, I've added a Regex.Replace to handle this case.

PR #4636, opened 2020-04-29
[Microsoft.Android.Sdk] change dotnet publish/build behavior

jonathanpeppers:
Currently, we are producing `.apk` files with .NET 5 with:

    dotnet publish foo.csproj

The problem with this is that `dotnet build` needs to produce
something that is "runnable". `dotnet build` is the developer loop,
while `dotnet publish` is something you run for a release.

Ideally, to produce an `.apk` you would instead do:

    dotnet build foo.csproj

`dotnet publish` would be used for distribution to Google Play. Our
first implementation would likely support re-signing the app and
copying the output to a local `publish` directory.

For now, I've removed our usage of `$(_CorePublishTargets)` and moved
these targets to `$(BuildDependsOn)`. The benefit here, is we are no
longer using any private MSBuild properties for .NET 5.

I updated our tests in these areas. I dropped the tests running
`dotnet publish`, as they are no longer valid.

Down the road:

* Make `dotnet publish` do *something*. Once we know what it should do.
* Implement `dotnet run`.

### DIFF
--- a/Documentation/guides/DotNet5.md
+++ b/Documentation/guides/DotNet5.md
@@ -53,13 +53,14 @@ Default Android related file globbing behavior is defined in `Microsoft.Android.
 
 ## dotnet cli
 
-There are currently two "verbs" we are aiming to get working in
+There are currently a few "verbs" we are aiming to get working in
 Xamarin.Android:
 
     dotnet build
     dotnet publish
+    dotnet run
 
-Currently in .NET Core (aka .NET 5), `dotnet publish` is where all the
+Currently in .NET 5 console apps, `dotnet publish` is where all the
 work to produce a self-contained "app" happens:
 
 * The linker via the `<IlLink/>` MSBuild task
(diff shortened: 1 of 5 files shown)


Documentation/guides/DotNet5.md
```diff
@@ -53,11 +53,12 @@ Default Android related file globbing behavior is defined in `Microsoft.Android.
...
 
-There are currently two "verbs" we are aiming to get working in
+There are currently a few "verbs" we are aiming to get working in
 Xamarin.Android:
 
     dotnet build
     dotnet publish
+    dotnet run
 
 Currently in .NET Core (aka .NET 5), `dotnet publish` is where all the
```
pjcollins, line 63: These publish lines may need a bit of rewording as well.

Documentation/guides/DotNet5.md
```diff
@@ -110,14 +122,13 @@ The following instructions can be used for early preview testing.
...
 </Project>
 ```
 
-  4) Publish (and optionally install) the project:
+  4) Build (and optionally run) the project:
 
-```
-dotnet publish -t:Install *.csproj
-```
+    dotnet build *.csproj -t:Run
```
pjcollins, line 132: I think we need to also update  https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets#L33 for this to work?

Documentation/guides/DotNet5.md
jonathanpeppers: This should probably say "in .NET 5 console apps" 👍

dellis1972:
Crazy idea.... 

How about we skip all of the fastdev stuff (e.g GetPrimaryCpuAdi etc) when we do the publish work flow? If the intention there is to only produce a release package I'm sure there are a ton of targets we can just not run in that scenario.

PR #4595, opened 2020-04-21
[.NET 5] Rename new SDK, Ref, and Runtime packs

pjcollins:
These changes rename and reorganize the sources and projects used to
create our .NET 5 SDK, Targeting, and Runtime packs.

The `src/Xamarin.Android.Sdk/` directory has been mostly moved to
`src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/`, as it contains
our .NET 5 props and targets files and logic that is very closely tied
to `Xamarin.Android.Build.Tasks` content.

All `.proj` files which were used to create a `nupkg` file have been
moved into `build-tools/create-packs/`.

A new msbuild target has been added to the common `create-packs` targets
file to allow all packs to be created with one simple command:

    msbuild /t:CreateAllPacks build-tools/create-packs/Microsoft.Android.Sdk.proj

PR #4638, opened 2020-04-30
[packaging] Package interpreter-enabled versions of the Mono runtime

grendello:
Context: 42822e0488185cdf4bca7c0bd05b21ad03dfbd7e

42822e04 installed interpreter versions of the Mono runtime into our
build tree but did not add code to package them.  This commit adds the
necessary entries to include the interpreter runtime in .pkg and .vsix
installer packages.

### DIFF
--- a/build-tools/installers/create-installers.targets
+++ b/build-tools/installers/create-installers.targets
@@ -152,19 +152,25 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\jcw-gen.pdb" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\jit-times.exe" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\jit-times.pdb" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\jnimarshalmethod-gen.exe" Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\jnimarshalmethod-gen.pdb" Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\jnimarshalmethod-gen.exe" Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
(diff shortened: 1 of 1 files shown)


jonpryor:
Should this add an on-device test of some form which enables the interpreter?  Maybe not even on-device, but a packaging test which asserts that the interpreter version is actually embedded…

grendello:
Yes, we should do that, but it's not a small change - I think a material for a separate PR

build-tools/installers/create-installers.targets
```diff
@@ -165,6 +165,12 @@
...
+    <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\interpreter-%(Identity)\libmono-btls-shared.so')"         Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\interpreter-%(Identity)\libmono-profiler-aot.so')"        Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\interpreter-%(Identity)\libmono-profiler-log.so')"        Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\interpreter-%(Identity)\libmono-native.so')"              Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
+    <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\interpreter-%(Identity)\libMonoPosixHelper.so')"          Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' "/>
+    <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\interpreter-%(Identity)\libmonosgen-2.0.so')"             Condition=" '$(PackageId)' != 'Xamarin.Android.Sdk' " />
```
pjcollins, line 173: Oops, I missed this file in https://github.com/xamarin/xamarin-android/pull/4595.

This file should now conditionally include content based on a PackageId of `Microsoft.Android.Sdk` rather than `Xamarin.Android.Sdk`.

@grendello would you mind fixing up the rest of this file to include that substitution where necessary?

Issue #4596, opened 2020-04-21
Test failure in TestBundleIntegerArrayList2 when the shared runtime is enabled


pjcollins:
### Steps to Reproduce

1. Build commercial xamarin-android in the debug configuration
2. Build and run the runtime suite:
```
bin/Debug/bin/xabuild /restore src/Mono.Android/Test/Mono.Android-Tests.csproj /p:AndroidUseSharedRuntime=True /t:AcquireAndroidTarget,Install,CheckAndRecordApkSizes,RunTestApks
```
3. TestBundleIntegerArrayList2 fails:
```
04-21 14:52:22.677 18253 18271 E NUnitLite: Test 'Xamarin.Android.RuntimeTests.BundleTest.TestBundleIntegerArrayList2' failed: System.MemberAccessException : Cannot create an instance of Android.Runtime.JavaList`1[T] because Type.ContainsGenericParameters is true.
04-21 14:52:22.677 18253 18271 E NUnitLite:   at System.Reflection.RuntimeConstructorInfo.DoInvoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00054] in <a71bc0650544428ea4f1cb0a371f7a80>:0 
04-21 14:52:22.677 18253 18271 E NUnitLite:   at System.Reflection.RuntimeConstructorInfo.Invoke (System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x00000] in <a71bc0650544428ea4f1cb0a371f7a80>:0 
04-21 14:52:22.677 18253 18271 E NUnitLite:   at System.Reflection.ConstructorInfo.Invoke (System.Object[] parameters) [0x00000] in <a71bc0650544428ea4f1cb0a371f7a80>:0 
04-21 14:52:22.677 18253 18271 E NUnitLite:   at Java.Interop.TypeManager.CreateProxy (System.Type type, System.IntPtr handle, Android.Runtime.JniHandleOwnership transfer) [0x0001b] in <1d456e9a51254e87a1ca3e16f71731f5>:0 
[...]

PR #4642, opened 2020-04-30
[typemaps] Don't create `System.Type` instance for generic types

grendello:
Context: 7117414ca27d88a71b4a272705a0207f772423bd

7117414c made string lookup return into our type map support, for Debug
builds only, and with them the old problem of duplicate type names
returned as well - it is possible that a single Java type name will map
to a number of managed types and vice versa.  This is not a problem in
most cases, but whenever a Java type name maps to a generic managed
type (for instance "Android.Runtime.JavaList`1, Mono.Android"), the
code added in 7117414c runs into trouble:

    System.MemberAccessException : Cannot create an instance of Android.Runtime.JavaList`1[T] because Type.ContainsGenericParameters is true.

The reason this happens is due to the way the native code processes the
managed type name it obtained from the `Java -> managed` map.  Namely,
it uses Mono embedding API to return a `MonoType` instance corresponding
to the managed type name, then that `MonoType` instance is converted
into `MonoReflectionType*` which is returned to the managed code (it
corresponds to the `System.Type` instance there) and then an attempt to
instantiate this type is made in the usual way.  It works for
non-generic types, but whenever a generic type is processed in this way
it will result in us returning a `MonoReflectionType*` corresponding to
an **open** generic type which then results in the exception shown
above.

Given just the type name, the native runtime is unable to "close" the
generic type without information regarding the types of the generic
parameters needed to do it.  The solution, a workaround really, for now
is to look at the type name and return `null` to the managed code
whenever we see that the type name contains the "`" character, marking
it as a generic type.  The managed code will then proceed to look up the
type and instantiate it in the correct way.

In the future we need to find a better solution for this issue, but
doing so would require quite extensive changes to our current code.  The
solution implemented here is sufficient for now.

### DIFF
--- a/src/monodroid/jni/embedded-assemblies.cc
+++ b/src/monodroid/jni/embedded-assemblies.cc
@@ -210,6 +210,14 @@ EmbeddedAssemblies::typemap_java_to_managed (const char *java_type_name)
 	const char *managed_type_name = entry->to;
 	log_debug (LOG_DEFAULT, "typemap: Java type '%s' corresponds to managed type '%s'", java_type_name, managed_type_name);
 
+	const char *ch = managed_type_name;
+	while (ch != nullptr && *ch != '\0') {
+		if (*ch++ == '`') {
+			log_debug (LOG_DEFAULT, "typemap: managed type '%s' is a generic one, will create Type instance in managed code", managed_type_name);
+			return nullptr;
+		}
+	}
+
 	MonoType *type = mono_reflection_type_from_name (const_cast<char*>(managed_type_name), nullptr);
 	if (XA_UNLIKELY (type == nullptr)) {
 		log_warn (LOG_ASSEMBLY, "typemap: managed type '%s' (mapped from Java type '%s') could not be loaded", managed_type_name, java_type_name);


jonpryor:
I don't think that this fix is quite correct.

**The Problem** is *duplicates* or *aliasing*: when *multiple* C# types bind the *same* Java type.  For example, [`java.util.ArrayList`](https://developer.android.com/reference/java/util/ArrayList) contains (at least?) *three* bindings:

  * `Android.Runtime.JavaList`: https://github.com/xamarin/xamarin-android/blob/61f09bf2ef0c8f09550e2d77eb97f417432b315b/src/Mono.Android/Android.Runtime/JavaList.cs#L11-L13

  * `Android.Runtime.JavaList<T>`: https://github.com/xamarin/xamarin-android/blob/61f09bf2ef0c8f09550e2d77eb97f417432b315b/src/Mono.Android/Android.Runtime/JavaList.cs#L667-L668

This has *always* been potentially problematic.

In the "original" typemap world order, we'd create a *singular* mapping from Java names to C# names: https://github.com/xamarin/java.interop/blob/fc18c54b2ccf14f444fadac0a1e7e81f837cc470/src/Java.Interop.Tools.JavaCallableWrappers/Java.Interop.Tools.JavaCallableWrappers/TypeNameMapGenerator.cs#L149-L186

This iteration had a couple of attributes:

 1. "First" type (potentially) "wins"
 2. Unless the first type is an interface or abstract *and* the subsequent type is an interface or abstract, in which case the "base" type "wins"

Thus, consider `java.util.ArrayList`: what is the order of declarations?

```
% monodis --typedef bin/Debug/lib/xamarin.android/xbuild-frameworks/MonoAndroid/v10.0/Mono.Android.dll | grep 'JavaList\|ArrayList'
4471: Android.Runtime.JavaList (flist=40852, mlist=91734, flags=0x100001, extends=0x5a84)
4472: Android.Runtime.JavaList`1 (flist=40865, mlist=91783, flags=0x100001, extends=0x45dc)
4515: Java.Util.ArrayList (flist=41084, mlist=92660, flags=0x100001, extends=0x4660)
…
```

The first type is non-generic.  `Android.Runtime.JavaList` "wins".

What about `ArrayAdapter` ([#android ref](https://xamarinhq.slack.com/archives/C03CEGRUW/p1588268168086600))?

```
% monodis --typedef bin/Debug/lib/xamarin.android/xbuild-frameworks/MonoAndroid/v10.0/Mono.Android.dll | grep 'ArrayAdapter'       
1035: Android.Widget.ArrayAdapter (flist=9306, mlist=21614, flags=0x100001, extends=0x1038)
1180: Android.Widget.ArrayAdapter`1 (flist=11186, mlist=27455, flags=0x100001, extends=0x102c)
…
```

That's `Mono.Android.dll`.  What about the `typemap.jm` file?

```
% strings typemap.jm
…
android/widget/ArrayAdapter
Android.Widget.ArrayAdapter, Mono.Android
…
java/util/ArrayList
```

(As space inefficient as the `typemap` format was, it at least makes interrogation easy…)

Note that there is only *one* "match"/mapped type in the above output.

The problem with the new typemap system, [as mentioned in chat](https://xamarinhq.slack.com/archives/C03CEGRUW/p1588268402089900) -- assuming I'm understanding things correctly) -- is that *the typemap* ***also*** contains "duplicates".  It is "as if" the original `typemap.jm` file had:

```
android/widget/ArrayAdapter
Android.Widget.ArrayAdapter, Mono.Android
android/widget/ArrayAdapter
Android.Widget.ArrayAdapter`1, Mono.Android
…
java/util/ArrayList
java/util/ArrayList
java/util/ArrayList
```

The new typemap system works this way because it "shares" the same array data between the java-to-managed mappings and the managed-to-java mappings; see commit 7117414ca27d88a71b4a272705a0207f772423bd.  `DebugTypemapFileHeader::java_to_managed` and `DebugTypemapFileHeader::managed_to_java` are the *same size*, each `DebugTypemapFileHeader::entry_count` entries in size.

The problem with the above is that it's *undefined* which entry **bsearch**(3) will find when there are such duplicates.  It could be the first; it could be the last.

What we *think* is happening in Issue #4596 is that `bsearch()` is (1) working, but (2) returning the *generic* type instead of the non-generic type.  *This* is the (plausibly) The Problem.

Which brings us to **The Solution** (?):

As laid out in this PR, it updates `EmbeddedAssemblies::typemap_java_to_managed()` so that if the target type is a generic type, `nullptr` is returned.

*This is insufficient*.  The ramifications of this will be a variant of the issue fixed in 
https://github.com/xamarin/monodroid/commit/192b1f1b71c98650d04bf0a4e52ee337a054ff4c.

Relevant code: https://github.com/xamarin/xamarin-android/blob/61f09bf2ef0c8f09550e2d77eb97f417432b315b/src/Mono.Android/Java.Interop/TypeManager.cs#L252-L264

Imagine:

 1. We hit `TypeManager.CreateInstance()`, where `targetType` is `null`, and the runtime Java type is a `java.util.ArrayList`.
 2. We hit the loop on lines 252-264, in which we start checking the runtime class of `handle`, and then start walking the base classes.
 3. In particular, we effectively do:

        type = GetJavaToManagedType ("java/util/ArrayList");

 4. If (3) returns `null`, we walk to the base class, which is `java.util.AbstractList`, which *isn't* aliased (yay?)
 5. (4) means that the instance returned by `TypeManager.CreateInstance()` is an instance of `Java.Util.AbstractListInvoker`

If the "caller" of `TypeManager.CreateInstance()` in turn *casts* the returned instance to something that *isn't* an `AbstractListInvoker`, then you get an `InvalidCastException`.

The fix in this PR fixes the originally reported `System.MemberAccessException`, but *potentially* "replaces" that with a later `InvalidCastException`.  Critically, `TestBundleIntegerArrayList2()` *won't* get an InvalidCastException, as it doesn't perform a runtime cast, but it's *imaginable*, because that's more-or-less the scenario that xamarin/monodroid#192b1f1b was fixing:

	sensor_manager = (SensorManager)GetSystemService (Activity.SensorService);

jonpryor:
Given that the apparent problem is that the array that `bsearch()` is searching contains duplicate keys, could we *instead* fix this by "probing" predecessor"elements" in the array that was searched, to see if they're *also* matches, and then return the earliest entry in the array?  No idea what this looks like; perhaps something like this?

```cpp
entry = binary_search<const char, TypeMapEntry, compare_type_name, false> (managed_type_name, module->managed_to_java, module->entry_count);
if (entry == nullptr)
	continue;
while (entry > module->java_to_managed &&
		compare_type_name (&entry [-1]->from, managed_type_name)) {
	entry = &entry [-1];
}
return entry;
```

Additionally, this PR *only* updates the *Debug* version of `EmbeddedAssemblies::typemap_java_to_managed()`.  I see no reason why the *Release* version wouldn't also potentially suffer from the same problem, and it should likewise be updated.

jonpryor:
*Actually*, it's `EmbeddedAssemblies::binary_search()` which should be updated to "look backwards" for earlier-appearing key entries…

jonpryor:
Alternate take: PR #4645

grendello:
> Given that the apparent problem is that the array that `bsearch()` is searching contains duplicate keys, could we _instead_ fix this by "probing" predecessor"elements" in the array that was searched, to see if they're _also_ matches, and then return the earliest entry in the array? No idea what this looks like; perhaps something like this?
> 
> ```c++
> entry = binary_search<const char, TypeMapEntry, compare_type_name, false> (managed_type_name, module->managed_to_java, module->entry_count);
> if (entry == nullptr)
> 	continue;
> while (entry > module->java_to_managed &&
> 		compare_type_name (&entry [-1]->from, managed_type_name)) {
> 	entry = &entry [-1];
> }
> return entry;
> ```
> 
> Additionally, this PR _only_ updates the _Debug_ version of `EmbeddedAssemblies::typemap_java_to_managed()`. I see no reason why the _Release_ version wouldn't also potentially suffer from the same problem, and it should likewise be updated.

`Release` doesn't use strings for the java -> managed lookup. Java **name** matches exactly one `MVID+TypeTokenID` entry (and vice-versa - the `MVID+TypeTokenID` entry will point to an exact java type, whatever its name)

grendello:
> Given that the apparent problem is that the array that `bsearch()` is searching contains duplicate keys, could we _instead_ fix this by "probing" predecessor"elements" in the array that was searched, to see if they're _also_ matches, and then return the earliest entry in the array? No idea what this looks like; perhaps something like this?

That wouldn't be sufficient, as we might be coming from the "top" or the "bottom" of the array, and so the correct thing to do, if we were to implement this, would be to look at the surrounding entries thus making 3 comparisons instead of one. As I stated in the OP, this PR is not the final solution for the issue and we need something better. I need to rewrite the code to remove the requirement that both java->managed and managed->java tables have identical sizes. Once that's done, we'll go back to the `typemap.jm` behavior in that there will be just a single entry in the map for any given type.  Alternatively, I can keep the identity relationship between map sizes and introduce an "alias" entry (for lack of better word) which would point duplicates to the correct entry. Or we can make all the duplicates point to the same managed type name (using the same algorithm as the old typemap generator in `Java.Interop` but instead of removing the duplicates, just using name of the type that is considered the "best" match for all the duplicate entries). However, I'm not sure if we have time for larger changes like this right now - if you think we do, then I'll close this PR and start working on one of the above solutions.

jonpryor:
Superseded by PR #4656.

PR #4643, opened 2020-04-30
[Xamarin.Android.Build.Tasks] add $(_AndroidAllowDeltaInstall) support

jonathanpeppers:
This PR:
-  Adds an experimental feature for speeding up incremental deploys of APKs, just sending over changes, when _AndroidAllowDeltaInstall is set. See details here https://github.com/xamarin/monodroid/pull/1089

- Adds the code to enable delta install (run on the PC) and the Android installers associated with it (run on the device) to androidtools and Xamarin Android. This code used to be in the designer repo and is now shared, part of Xamarin Android. This code is used for both delta install (new experimental feature, used for all APK deploys when _AndroidAllowDeltaInstall is enabled) and for Apply Changes (existing non-experimental feature), updating resources only on the device when the user invokes the Apply Changes command.

- Bump to xamarin/monodroid/master@e795bce

  Changes: https://github.com/xamarin/monodroid/compare/5849ea1...e795bce


tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
```diff
@@ -302,13 +302,24 @@ public override void OnCreate ()
...
 			},
+			new object[] {
+				/* useSharedRuntime */   true,
+				/* embedAssemblies */    false,
+				/* fastDevType */        "Assemblies:Dexes",
+			},
+			new object[] {
+				/* useSharedRuntime */   true,
+				/* embedAssemblies */    false,
+				/* fastDevType */        "Assemblies:Dexes",
```
dellis1972, line 313: The `Assemblies;Dexes` mode will not change the apk when an AndroidResource is altered. So I'm not sure this is a good test. We should just use the `Assemblies` mode. That way any resource change will end up in the apk.

BretJohnson: Updated

BretJohnson:
@brendanzagaeski - I (finally) added release notes. Can you please approve now.

build-tools/installers/create-installers.targets
```diff
@@ -386,6 +386,12 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Build.Debugging.Tasks.pdb" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Common.Debugging.props" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Common.Debugging.targets" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\arm64-v8a\installer" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\armeabi-v7a\installer" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\x86\installer" />
```
dellis1972, line 391: @BretJohnson I think there is an existing `lib` folder where we put all the native android libraries. Perhaps we should put these tools in there too? I was panning to put the native fastdev 2 stuff in the `lib/xxxx` folders.

BretJohnson: @dellis1972 I moved the installers to the lib directory, with this PR: https://github.com/xamarin/androidtools/pull/231
Can you please review & approve. With that, the `installer` executable will be in the e.g. `lib/x86` subdirectory, next to the other device specific binaries.

BretJohnson:
@brendanzagaeski It's really option 2. Yes, it aims to make deployments faster, but it's conceptually separate from fast deploy and complementary to it.
Fast deploy is about taking assemblies out of the APK and installing them quickly (for debug deployments).
While DeltaInstall (which is a term Google uses internally in their code though it's not publicized to external devs) makes APK deployments faster - whatever is left in the APK (like resources) can get deployed faster, by only updating parts of the APK that changed.

Android Studio does use this technology. It's used to make "Apply Changes" faster, when only resources are updated, and also make all deploys faster by only installing changes.

Still, this is experimental so we don't want to say much about it, yet. I'm also fine removing it from the release notes, but it's probably a tiny bit better to have it there.

brendanzagaeski:
> DeltaInstall (which is a term Google uses internally in their code though it's not publicized to external devs) 

Mhm. That's what I was seeing when I was trying web searches. So for me, from a user messaging perspective, if possible, I'd love to avoid making the term "delta install" dev-facing, especially since there are already quite a few vocabulary terms around how to get app changes from the IDE to devices.

>  it's conceptually separate from fast deploy and complementary to it

It's conceptually separate in terms of implementation, but in terms of customer feature use, it seems like a trickier distinction to draw. A problem is that the existing feature uses the term "fast deployment" instead of something like "deploy assemblies outside the APK." The generality of the term "fast deployment" gobbles up the vocabulary space. But since that term has already been around for a long time in Xamarin.Android, that's water under the bridge now

There might be a way around this by adding another term that's even more general than "fast deployment," but that'd take some brainstorming.

> Both this & fast deploy can be switched on & off separately, which is what we want.

The good and maybe bad (because like the name "fast deployment," this mechanism also eats up conceptual space with its generality) news is that a colon-separated item in `$(AndroidFastDeploymentType)` would still allow that. So for example, the default Release configuration could fast deploy only APK deltas, whereas the Debug configuration could fast deploy more things:

```xml
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
    <AndroidFastDeploymentType>ApkDiff:Assemblies</AndroidFastDeploymentType>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
    <AndroidFastDeploymentType>ApkDiff</AndroidFastDeploymentType>
  </PropertyGroup>
```

Using `Apk` as part of the new term also keeps the door open for a future potential possibility of doing similar "delta installs" for managed assemblies that can be controlled separately from the APK "delta installs."

@dellis1972 can probably tell me if I'm getting too loony in suggesting `$(AndroidFastDeployDexes)`. In that case, I'd suggest including `Apk` in the property name and, for similarity to existing properties, replacing `Allow` with `Use`:

- `AndroidUseApkDeltaInstall`
- Or maybe `AndroidUseApkDeltaDeployment` since the [build process doc](https://docs.microsoft.com/xamarin/android/deploy-test/building-apps/build-process) is under the _Deployment and testing_ docs section.

build-tools/installers/create-installers.targets
dellis1972: @BretJohnson That PR was merged can you update this PR with the new paths :)

BretJohnson:
@brendanzagaeski Maybe we should just remove it from the release notes for now. Would you be OK with that? This is an experimental feature, primarily to enable internal performance testing. We'd rather not make customers aware of it until later, when we have more confidence in how we'll productize it. Ideally, it'll be on by default eventually, but we're not there yet.

jonathanpeppers:
@BretJohnson if you want to leave it undocumented, prefix the property with an underscore, so it will be `$(_AndroidAllowDeltaInstall)`. (or if we have a better name now, that works, too)

build-tools/installers/create-installers.targets
BretJohnson: Thanks. And the paths are now updated.

BretJohnson:
On names, I rather like `AndroidUseApkDeltaInstall`. @brendanzagaeski @dellis1972 @jonathanpeppers What do you think? Anyone not like that?
If no objections, I'll rename the setting to it.

BretJohnson:
On second though, after discussing more with @jonathanpeppers , I'll rename to `_AndroidAllowDeltaInstall` and we'll leave this undocumented for now.

Later, when/if this is a productized feature, on by default, we can add a different option potentially to disable it.

BretJohnson:
I renamed to `_AndroidAllowDeltaInstall` and removed this from the doc.

See rename changes in this PR and in https://github.com/xamarin/monodroid/pull/1089

BretJohnson:
@brendanzagaeski per above, I made this setting internal only for now, renaming it to `_AndroidAllowDeltaInstall` and removing doc / release notes doc
If you're good with all that, can you please approve. Thanks.

build-tools/installers/create-installers.targets
```diff
@@ -386,6 +386,12 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Build.Debugging.Tasks.pdb" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Common.Debugging.props" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Common.Debugging.targets" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\lib\arm64-v8a\installer" />
```
jonpryor, line 389: As this means we're redistributing `installer`, this means that we need to update the `ThirdPartyNotices.txt` that we generate and include in the installers so that the licensing information for `installer` is available to our users.

jonpryor:
A link to some form of "spec" or "documentation" would also be handy.  You never know when you'll need to check a spec, e.g. https://github.com/xamarin/xamarin-android/pull/4681#discussion_r427681702

@BretJohnson?

build-tools/installers/create-installers.targets
BretJohnson: Per @garuma:

> they should already have the right entry, the license and copyright is the same as the other tooling (Apache 2.0 copyright Android Open Source Project)
> but they may need a specific entry for the entries, their system is different
>

BretJohnson: @jonathanpeppers do we need anything else for this?

jonathanpeppers: @jonpryor is this existing license sufficient?

https://github.com/xamarin/xamarin-android/blob/bea063b939d7d8d5f166c50564665d62ab140417/build-tools/xaprepare/xaprepare/ThirdPartyNotices/Xamarin.Android.Build.Tasks.cs#L7-L16

PR #4644, opened 2020-04-30
[build] install .NET 5 preview 4 nightly build

jonathanpeppers:
Context: https://github.com/xamarin/net5-samples/blob/7607d1cde77f0bef676cb7e1a4290e3150501c56/azure-pipelines.yml
Context: https://docs.microsoft.com/azure/devops/pipelines/tasks/tool/dotnet-core-tool-installer

Currently, on CI we've been doing:

    - task: UseDotNet@2
      inputs:
        version: 5.0.100-preview.1.20155.7

This `UseDotNet` Azure DevOps task works for any public release of
.NET 5 found here:

https://dotnet.microsoft.com/download/dotnet/5.0

However, we need .NET 5 Preview 4, which is not on this public
downloads page yet...

Luckily, we can use these install scripts *instead*:

https://docs.microsoft.com/dotnet/core/tools/dotnet-install-script

The main drawback is the powershell script *only* works on Windows, so
we have to conditionally use bash on macOS.

This allows us to download any build found here:

https://github.com/dotnet/installer#installers-and-binaries

Now we can get .NET 5 Preview 4 on CI *right now*. This will be
important for us to start using changes that may have only landed in
dotnet/installer/master down the road.

On macOS, we also have to set `$PATH` and `$DOTNET_ROOT` for the
remainder of the build using:

    echo "##vso[task.setvariable variable=DOTNET_ROOT]$DOTNET_ROOT" &&
    echo "##vso[task.setvariable variable=PATH]$PATH"

When using just `export`, future steps on Azure DevOps did not have the
new environment variables set.

At some point we can switch back to the `UseDotNet` task when .NET
public builds are available.

### DIFF
--- a/build-tools/automation/yaml-templates/use-dot-net.yaml
+++ b/build-tools/automation/yaml-templates/use-dot-net.yaml
@@ -1,0 +1,27 @@
+# This template enables installing master/nighly builds of .NET 5.
+# We can eventually switch back to the UseDotNet task.
+
+parameters:
+  version: ''
+
+steps:
+
+  - powershell: |
+      $ErrorActionPreference = 'Stop'
+      $ProgressPreference = 'SilentlyContinue'
+      Invoke-WebRequest -Uri "https://dot.net/v1/dotnet-install.ps1" -OutFile dotnet-install.ps1
+      & .\dotnet-install.ps1 -Version ${{ parameters.version }} -InstallDir "$env:ProgramFiles\dotnet\" -Verbose
+      & dotnet --list-sdks
+    displayName: install .NET Core ${{ parameters.version }}
+    condition: and(succeeded(), eq(variables['agent.os'], 'Windows_NT'))
+
+  - bash: >
+      DOTNET_ROOT=~/.dotnet/ &&
+      PATH="$DOTNET_ROOT:$PATH" &&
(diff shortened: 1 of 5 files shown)


jonathanpeppers:
> Do we need this working on Linux too?

We only need this installed for any of the `XASdk*` tests, and I don't think the Linux build runs any MSBuild tests.

build-tools/automation/yaml-templates/use-dot-net.yaml
```diff
@@ -0,0 +1,27 @@
+# This template enables installing master/nighly builds of .NET 5.
+# We can eventually switch back to the UseDotNet task.
+
+parameters:
+  version: ''
```
pjcollins, line 5: Minor nit would be to use `$(DotNetCoreVersion)` as the default parameter value here, but this looks good to me overall.

jonathanpeppers:
The designer/macOS test is failing on master, and I think we can ignore the single MSBuild test failure. A test was showing this target didn't skip: 
```
Building target "_UpdateAndroidResgen" completely.
Output file "obj/Debug/R.cs.flag" does not exist.
```
I think it may be an issue with the test if that file is completely missing. Will look into it if it keeps happening.

The previous run worked: https://build.azdo.io/3698980

PR #4645, opened 2020-05-01
[monodroid] Return first entry in binary_search

jonpryor:

Context: https://github.com/xamarin/monodroid/commit/192b1f1b71c98650d04bf0a4e52ee337a054ff4c
Context: https://github.com/xamarin/java.interop/blob/fc18c54b2ccf14f444fadac0a1e7e81f837cc470/src/Java.Interop.Tools.JavaCallableWrappers/Java.Interop.Tools.JavaCallableWrappers/TypeNameMapGenerator.cs#L149-L186

In some environments, `BundleTest.TestBundleIntegerArrayList2()` fails
when  using the commercial shared runtime:

	Test 'Xamarin.Android.RuntimeTests.BundleTest.TestBundleIntegerArrayList2' failed: System.MemberAccessException : Cannot create an instance of Android.Runtime.JavaList`1[T] because Type.ContainsGenericParameters is true.

The problem appears to be rooted in ce2bc689.  In a pre-ce2bc689 world,
there were two separate sets of "typemap" files that the application
could use:

  * Java-to-managed mappings, in `typemap.jm`
  * Managed-to-Java mappings, in `typemap.mj`.

These files did *not* need to have the same number of entries!

	% unzip /Library/Frameworks/Xamarin.Android.framework/Versions/10.2.0.100/lib/xamarin.android/xbuild/Xamarin/Android/Mono.Android.DebugRuntime-debug.apk typemap.mj
	% strings typemap.jm | wc -l
	   10318
	% strings typemap.mj | wc -l
	   11956

The reason why they have a different number of entries is *aliasing*:
there can be *multiple* bindings for a given Java type.  The
managed-to-Java mappings can contain *all* of them; the
Java-to-managed mapping *must* pick Only One.

For example, [`java.util.ArrayList`][0] contains (at least?) *three* bindings:

  * [`Android.Runtime.JavaList`][1]
  * [`Android.Runtime.JavaList<T>`][2]

In a pre-ce2bc689 world, this was "fine": we'd binary search each file
*separately* to find type mappings.

This changed in ce2bc689, as the typemap information was merged into a
*single* array in order to de-duplicate information.  This reduced
`.apk` size.

An unanticipated result of this is that the Java-to-managed mappings
can now contain *duplicate* keys, "as if" the original `typemap.jm`
had the contents:

  java/util/ArrayList
  java/util/ArrayList
  java/util/ArrayList

Whereas pre-ce2bc689 there would have only been the first entry.

"Sometimes" this duplication is "fine": the typemap search
"Just happens" to return the first entry, or the 3rd entry, and apps
continue to work as intended.

"Sometimes" it isn't: the typemap binary search finds the 2nd entry,
which is a generic type.  This results in attempting to instantiate a
generic type *without appropriate type parameters*, which results in
the aforementioned `MemberAccessException`.

Update `EmbeddedAssemblies::binary_search()` so that instead of
returning the "first" entry that the binary search happens to land on,
probe *previous* entries in the type mappings to see if they *also*
contain the same key.  `binary_search()` can instead return the
*first entry* within the mapping, not just the first it happens to
land upon while binary searching.

This should allow for more *consistent* behaviors, ensuring that
Java-to-managed typemap lookup obtains a *non*-generic type, thus
avoiding the `MemberAccessException`.

Additionally, update `TestBundleIntegerArrayList2()` so that it asserts
the runtime *type* of `obj` returned, asserting that it is in fact a
`JavaList` and not e.g. `Java.Util.ArrayList` or something.
(The linker could otherwise "change" things, and this assertion will
ensure that `JavaList` won't be linked away…)

[0]: https://developer.android.com/reference/java/util/ArrayList
[1]: https://github.com/xamarin/xamarin-android/blob/61f09bf2ef0c8f09550e2d77eb97f417432b315b/src/Mono.Android/Android.Runtime/JavaList.cs#L11-L13
[2]: https://github.com/xamarin/xamarin-android/blob/61f09bf2ef0c8f09550e2d77eb97f417432b315b/src/Mono.Android/Android.Runtime/JavaList.cs#L667-L668

### DIFF
--- a/Documentation/release-notes/4645.md
+++ b/Documentation/release-notes/4645.md
@@ -1,0 +1,5 @@
+#### Application behavior on device and emulator
+
+  * [GitHub 4596](https://github.com/xamarin/xamarin-android/issues/4596):
+    Ensure that Java-to-managed type mappings lookup the correct non-generic
+    binding type, avoiding a *System.MemberAccessException*.
--- a/src/Mono.Android/Test/Android.OS/BundleTest.cs
+++ b/src/Mono.Android/Test/Android.OS/BundleTest.cs
@@ -34,6 +34,7 @@ namespace Xamarin.Android.RuntimeTests {
 			b.PutIntegerArrayList ("key", new List<Java.Lang.Integer> () { Java.Lang.Integer.ValueOf (1) });
 			var obj = b.Get ("key");
 			Assert.NotNull (obj, "Missing 'key' in bundle");
+			Assert.IsTrue (obj is global::Android.Runtime.JavaList, "`obj` should be a JavaList!");
 			try {
 				var list = b.GetIntegerArrayList ("key");
 				Assert.NotNull (list, "'key' doesn't refer to a list of integers");
--- a/src/Mono.Android/Test/Environment.txt
+++ b/src/Mono.Android/Test/Environment.txt
@@ -1,0 +1,2 @@
+debug.mono.debug=1
+debug.mono.log=timing,default
--- a/src/Mono.Android/Test/Mono.Android-Tests.csproj
+++ b/src/Mono.Android/Test/Mono.Android-Tests.csproj
@@ -79,6 +79,9 @@
     <AndroidBoundLayout Include="Resources\layout\Main.axml" />
   </ItemGroup>
   <ItemGroup>
+    <AndroidEnvironment Include="Environment.txt" />
+  </ItemGroup>
+  <ItemGroup>
     <AndroidAsset Include="Assets\asset1.txt" />
     <AndroidAsset Include="Assets\subfolder\asset2.txt" />
   </ItemGroup>
--- a/src/monodroid/jni/embedded-assemblies.cc
+++ b/src/monodroid/jni/embedded-assemblies.cc
@@ -149,8 +149,8 @@ EmbeddedAssemblies::binary_search (const Key *key, const Entry *base, size_t nme
 	}
 
 	constexpr size_t size = sizeof(Entry);
+	const Entry *ret = nullptr;
 	while (nmemb > 0) {
-		const Entry *ret;
 		if constexpr (use_extra_size) {
 			ret = reinterpret_cast<const Entry*>(reinterpret_cast<const uint8_t*>(base) + ((size + extra_size) * (nmemb / 2)));
 		} else {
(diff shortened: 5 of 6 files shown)


Documentation/release-notes/4645.md
```diff
@@ -0,0 +1,5 @@
+#### Application behavior on device and emulator
+
+  * [GitHub 4596](https://github.com/xamarin/xamarin-android/issues/4596):
+    Ensure that Java-to-managed type mappings lookup the correct non-generic
+    binding type, avoiding a *System.MemberAccessException*.
```
brendanzagaeski, line 5: Looks good. As a heads-up, I'll plan to adjust this to the following for the final aggregation of the release notes:

```

  * [GitHub 4596](https://github.com/xamarin/xamarin-android/issues/4596):
    Starting in Xamarin.Android 10.3 Preview, *System.MemberAccessException*
    could cause apps to abort when using certain APIs involving generic types.
```

I've got in the habit of using the "Starting in..." phrase to let users know the scope of the fix for regressions and to help me see which items can removed from the final GA release notes. The other adjustment is to move the user-visible message text to appear near the beginning, just after that "Starting in..." phrase, because that text is what a user would be most likely to look for to recognize the issue if they were hitting it.

Further suggestions on the note are welcome, either as comments on this PR or as changes in the committed note, but no need to change the committed note otherwise, unless you want to. Thanks!

grendello:
I don't like this solution as it sacrifices performance doing on the runtime what should rather be done on the build time (and especially that it affects the `Release` builds which do not have any problem with duplicates), but since we need a solution now, I'm fine with it until we can implement the correct fix (e.g. one of [these ideas](https://github.com/xamarin/xamarin-android/pull/4642#issuecomment-622570290))

jonpryor:
Alas, crashes in the APK tests:

```
Abort message: 'JNI DETECTED ERROR IN APPLICATION: can't make objects of type javax.net.ssl.X509TrustManager: 0x7026cb88
    in call to AllocObject
    from void crc64f295cabbf85394f5.TestSuiteInstrumentation.n_onStart()'
    eax 00000000  ebx 00000dbb  ecx 00000dd0  edx 00000006
    edi eeb8431e  esi c6d7ada0
    ebp f23d1ad0  esp c6d7ad48  eip f23d1ad9
```

I'm not immediately sure why this would be happening.

jonpryor:
This approach won't fix Issue #4660, as it "just" uses the first entry, which within #4660 will *always* be a generic type.

This PR has been superseded by PR #4656.

PR #4646, opened 2020-05-01
[.NET 5] support for binding projects

jonathanpeppers:
The current idea for binding projects in .NET 5 is they won't be a
distinct project type at all anymore. It would be completely valid to
put this in a .NET 5 Xamarin.Android app with `$(OutputType)` of
`Exe`:

    <InputJar Include="foo.jar" />

A project with an `$(OutputType)` of `Library` would likely do:

    <EmbeddedJar Include="foo.jar" />

To make this work, I've split up `Xamarin.Android.Bindings.targets`
into a few different files:

* `Xamarin.Android.Bindings.Core.targets` - the targets shared between
  .NET 5 and "legacy" binding projects. It imports the other new files.
* `Xamarin.Android.Bindings.ClassParse.targets` - is only imported
  when `$(AndroidClassParser)` is `class-parse`.
* `Xamarin.Android.Bindings.JarToXml.targets` - is only imported
  when `$(AndroidClassParser)` is `jar2xml`.
* `Xamarin.Android.Bindings.Documentation.targets` - is imported by
  "legacy" binding projects, contains targets related to documentation
  extraction/generation. We will eventually use this in .NET 5 when
  `mdoc.exe` is available or we have an alternative.

Once these were split up, to get bindings working in .NET 5 projects I
had to:

1. Import `Xamarin.Android.Bindings.Core.targets`.
2. Update `Microsoft.Android.Sdk.BuildOrder.targets` to include the
   binding-specific targets.
3. Make the binding targets only run if there are any `@(InputJar)` or
   `@(EmbeddedJar)`.

One hangup is that bindings require:

    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>

We probably don't want this set for all projects by default. I set
this if any of the binding targets run, so that once you add a
`@(InputJar)` it gets turned on.

I updated `XASdkTests` to verify binding `.jar` files works in a
library project and an app project.

Other changes:

* Added a new `AndroidDotnetToolTask` for swapping between
`generator.exe` and `dotnet generator.dll` calls.
* I ported `javadoc2mdoc.csproj` to use short-form projects, and
multi-target both `net472` and `netcoreapp3.1`.
* The `<ImportJavaDoc/>` now extends `AndroidDotnetToolTask`.
* `$(ApplicationResolveReferencesDependsOn)` is unused, so I removed
usage of it.
* I removed the `_ResolveMonoAndroidFramework` target, it was not
needed.

TODO:

* Item groups like `@(JavaSourceJar)` can't fully work yet in .NET 5.
  `Xamarin.Android.Documentation.targets` uses `mdoc.exe`, which won't
  run under `dotnet` yet.
* For now, `Xamarin.Android.Documentation.targets` is not imported in
  .NET 5 projects.


Documentation/guides/DotNet5.md
```diff
@@ -45,6 +45,13 @@ In .NET 5 the behavior of the following MSBuild tasks will change, but
 better align with existing .NET 5 semantics, at the cost of compatibility with
 previous Xamarin.Android releases.
 
+`$(AndroidClassParser)` will be `class-parse` by default.
+
+`$(AndroidCodegenTarget)` will be `XAJavaInterop1` by default.
+
+If Java binding is enabled with `@(InputJar)` or `@(EmbeddedJar)`,
+then `$(AllowUnsafeBlocks)` must default to `True`.
```
brendanzagaeski, line 53: It might be handy to list `@(LibraryProjectZip)` here too since binding `.aar` libraries is probably fairly common and I think this PR already supports it. (`<CreateLibraryResourceArchive/>` will copy the `.jar` files from within the `.aar` archives to `$(IntermediateOutputPath)library_project_jars`, and `AddLibraryJarsToBind` will add them to `@(InputJar)`.)

For the sake of completeness, `@(LibraryProjectProperties)` could also be added. I doubt any users will be trying to bind those old `android create lib-project` library projects anymore, but it might be handy to have it listed for the teams own future reference.

jonpryor:
Note: we *do not* want to support `jar2xml` in .NET 5, and thus the .NET 5 targets *should not* import or distribute `Xamarin.Android.Bindings.JarToXml.targets`.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.Core.targets
```diff
@@ -0,0 +1,169 @@
...
+    <AvailableItemName Include="TransformFile" />
+    <AvailableItemName Include="LibraryProjectProperties" />
+    <AvailableItemName Include="LibraryProjectZip" />
+    <AvailableItemName Include="JavaDocIndex" />
+    <AvailableItemName Include="JavaDocJar" />
+    <AvailableItemName Include="JavaSourceJar" />
+  </ItemGroup>
+
+  <PropertyGroup>
+    <ExportJarToXmlDependsOnTargets>
```
jonpryor, line 40: `jar2xml` related *anything* should be in `Xamarin.Android.Bindings.JarToXml.targets`.  `jar2xml` properties/items/targets/etc. *should not* exist in .NET 5.

build-tools/installers/create-installers.targets
```diff
@@ -214,6 +214,10 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Analysis.targets" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Application.targets" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.targets" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.ClassParse.targets" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.Core.targets" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.Documentation.targets" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.JarToXml.targets" />
```
jonpryor, line 220: This file shouldn't be installed under .NET 5.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.Core.targets
```diff
@@ -0,0 +1,169 @@
...
+  <Target Name="CleanNativeLibraryIntermediaries">
+    <Delete Files="$(IntermediateOutputPath)__AndroidNativeLibraries__.zip" />
+    <RemoveDirFixed Directories="$(IntermediateOutputPath)native_library_imports" />
+    <RemoveDirFixed Directories="$(IntermediateOutputPath)libs" />
+    <RemoveDirFixed Directories="$(IntermediateOutputPath)$(_NativeLibraryImportsDirectoryName)" />
+  </Target>
+
+  <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Bindings.Documentation.targets" />
+  <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Bindings.ClassParse.targets" Condition=" '$(AndroidClassParser)' == 'class-parse' " />
+  <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Bindings.JarToXml.targets"   Condition=" '$(AndroidClassParser)' == 'jar2xml' " />
```
jonpryor, line 167: This import should instead be in some ".Legacy.targets" file so it's only done on desktop VS, not .NET 5.

Documentation/guides/DotNet5.md
jonathanpeppers: It looks like `AddLibraryJarsToBind` runs first and so `@(InputJar)` is filled out. Otherwise, we wouldn't get any C# bindings for them.

I don't think we need to list these?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.Core.targets
jonathanpeppers: The target is named `ExportJarToXml`, even for `class-parse`...

dellis1972:
@jonathanpeppers I'm not sure we can just delete `Xamarin.Android.Bindings.targets`. Existing binding projects will be doing the following

`<Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.Bindings.targets" />`

So we should have a "legacy" target file which just imports the new stuff. This is until we remove the old stuff completely.

jonathanpeppers:
> I'm not sure we can just delete `Xamarin.Android.Bindings.targets.`

@dellis1972 sorry Github's UI is misleading. I removed 255 lines, it's down to ~160 lines now.

Things would completely break if I removed it, `.csproj` files have:

```xml
<Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.Bindings.targets" />
```

Documentation/guides/DotNet5.md
brendanzagaeski: The main thing I was thinking was that a user reading this document might think `@(LibraryProjectZip)` was not supported outside of bindings projects and that they'd need to keep creating bindings projects to use `@(LibraryProjectZip)`.

Maybe that info actually makes more sense in its own section anyway? So maybe this PR could add another short section with its own heading like "Consolidation of bindings projects and other project types" or "Changes to bindings projects" that says something like:

> Bindings projects are no longer a separate project type. All of the bindings project build actions like **LibraryProjectZip** and **TransformFile** can now be used directly in Xamarin.Android class library projects and Xamarin.Android application projects instead.

jonathanpeppers: Sorry, I realize now your commenting on the doc and not the code. 🤦‍♂️ That makes sense, thanks.

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
```diff
@@ -58,9 +57,20 @@ public class BindingsGenerator : AndroidRunToolTask
...
 		private List<Tuple<string, string>> transform_files = new List<Tuple<string,string>> ();
+		bool useDotNet = false;
+		string generatorAssemblyPath;
 
 		public override bool RunTask ()
 		{
+			var framework = RuntimeInformation.FrameworkDescription;
+			Log.LogDebugMessage ($"FrameworkDescription: {framework}");
+			if (!framework.StartsWith (".NET Framework", StringComparison.OrdinalIgnoreCase) &&
+					!framework.StartsWith ("Mono", StringComparison.OrdinalIgnoreCase)) {
```
jonathanpeppers, line 68: This is the only part I'm "iffy" about. Is there any better way to detect we're running on .NET 5/.NET Core from a netstandard library?

I got this idea from: https://github.com/dotnet/BenchmarkDotNet/blob/a08062fa473e3426c7d5f1a1cf2ad5cbbc46f7d1/src/BenchmarkDotNet/Environments/Runtimes/CoreRtRuntime.cs#L47-L51

Documentation/guides/DotNet5.md
brendanzagaeski: Ah, ha! No worries. The new additions to the doc from the follow-up commits look good!

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
jonpryor: Related to the "iffy-ness", should we "abstract-ify" this a bit, so that it's not *just* `generator.exe`/`generator.dll` which is special-cased?  From a quick `git grep -3 'override.*ToolName' src/Xamarin.Android.Build.Tasks`, these other utilities may also need to be similarly special-cased:

  * `javadoc-to-mdoc.exe` (`<ImportJavaDoc/>`)
  * `mdoc.exe` (`<MDoc/>`)
  * `mono-symbolicate.exe` (?) (`<MonoSymbolicate/>`)

(xamarin-android builds `javadoc-to-mdoc.exe`, and thus *can* add .NET Core target framework build support for that.  `mdoc.exe` and `mono-symbolicate.exe` come from Mono, and thus need some cross-team discussion to determine how to deal with them in the future.  *Presumably* they'll become .NET Core apps?)

Expanding our scope thus, how can/should these be supported?  For starters, we'd want `AndroidRunToolTask` to have the appropriate logic, not `BindingsGenerator`/`ImportJavaDoc`/etc.

Next, do we *need* care about ".NET Core" vs. everything else?  Or can we just *probe*, e.g. by having `GenerateCommandLineCommands()` do:

```csharp
string dotnetCommand = Path.ChangeExtension (ToolName, ".dll");
string dotnetCommandPath = Path.Combine (ToolPath, dotnetCommand);
if (File.Exists (dotnetCommandPath)) {
    // use `dotnet {dotnetCommandPath} …`
}
else {
    // use `{ToolPath}/{ToolName} …`
}
```

jpobst: MSBuild knows what it is running on if that helps: `$(MSBuildRuntimeType) = Core, Full, Mono`: https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-reserved-and-well-known-properties?view=vs-2019.

jonathanpeppers: Thanks, maybe I should switch to it, since MSBuild is defining these based on how it is built:
https://github.com/microsoft/msbuild/blob/6f41a4fa3c6957d430783b7d31fe16e6a13b6e8e/src/Build/Evaluation/Evaluator.cs#L1159-L1166

jonpryor:
> * `Xamarin.Android.Bindings.Documentation.targets` - is always
>     imported and contains everything related to documentation
>     extraction/generation.

Given that it uses `mdoc.exe`, don't we want to *skip* this file in .NET 5?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
```diff
@@ -91,7 +91,18 @@
     </None>
     <None Include="Xamarin.Android.Bindings.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <SubType>Designer</SubType>
+    </None>
+    <None Include="Xamarin.Android.Bindings.ClassParse.targets">
```
jonpryor, line 95: Should these added files instead go into: https://github.com/xamarin/xamarin-android/tree/master/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android

Doing so helps reduce the "file pollution" going in in `src/Xamarin.Android.Build.Tasks`, which makes *me* happier when doing `code -n src/Xamarin.Android.Build.Tasks/*.targets`…

jonathanpeppers:
I updated this part of the commit message, I just forgot to update it:

> * `Xamarin.Android.Bindings.Documentation.targets` - is imported by
  "legacy" binding projects, contains targets related to documentation
  extraction/generation. We will eventually use this in .NET 5 when
  `mdoc.exe` is available or we have an alternative.

jpobst:
Are you saying you can put an `<InputJar>` directly in an Android application project and consume it there?

Today you can only have 1 `<InputJar>` per binding project. Is that still going to be in place?  It seems weird to say you can add up to 1 `.jar` directly in your application.  But allowing more than 1 seems like it won't work, because how do we match `metadata/etc` files to each `jar`?

jonathanpeppers:
> Today you can only have 1 <InputJar> per binding project. Is that still going to be in place? 

I think multiple `.jar` files work today?

I think what happens is they are all merged into a single `api.xml` file: https://github.com/xamarin/xamarin-android/blob/7dab4ee432dcd1e51c52dfe579023fccaf09c471/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets#L264

If that doesn't work, then `@(InputJar)` and `@(EmbeddedJar)` shouldn't be item groups, they should be a property?

Right now this PR is only changing .NET 5 projects, but we could consider current Xamarin.Android apps to include bindings if that makes sense.

jpobst:
Ah, I think you are correct.  You can only bind a single `.aar` per project: https://docs.microsoft.com/en-us/xamarin/android/platform/binding-java-library/binding-an-aar.

Is that an issue with this?

jonathanpeppers:
> You can only bind a single .aar per project.

I think it's going to behave the same inside an app, too. I didn't have to actually change the logic we have much, it was mostly moving things around.

I'll work on porting something to .NET 5 next week, and we'll see what issues I hit.

jonathanpeppers:
The one test failure looks OK:
```
Warning while updating resource XML 'obj\Release\res\layout-v16\list_item2.xml': Access to the path 'C:\A\vs2019xam00000E-1\_work\2\s\bin\TestRelease\temp\CodeBehind\FailedBuildMany_ConflictingLinearLayout\Release\project\obj\Release\res\layout-v16\list_item2.xml' is denied.
```

tools/javadoc2mdoc/javadoc2mdoc.csproj
```diff
@@ -1,44 +1,30 @@
...
-  </PropertyGroup>
-  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-    <Optimize>true</Optimize>
-    <OutputPath>$(XAInstallPrefix)xbuild\Xamarin\Android\</OutputPath>
-    <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>4</WarningLevel>
-    <ExternalConsole>true</ExternalConsole>
-    <IntermediateOutputPath>obj\AnyCPU\Release</IntermediateOutputPath>
+  <PropertyGroup>
+    <OutputPath>$(XamarinAndroidSourcePath)bin\$(Configuration)-$(TargetFramework)\</OutputPath>
```
jonpryor, line 11: Shouldn't this be conditional on `$(TargetFramework)`, so that when building for `net472` the output path continues to be `$(XAInstallPrefix)xbuild\Xamarin\Android\`?

tools/javadoc2mdoc/javadoc2mdoc.csproj
```diff
@@ -1,44 +1,30 @@
...
-  </ItemGroup>
-  <ItemGroup>
-    <ProjectReference Include="..\..\src\Xamarin.Android.Tools.JavadocImporter\Xamarin.Android.Tools.JavadocImporter.csproj">
-      <Project>{E0890301-F75F-40E7-B008-54C28B3BA542}</Project>
-      <Name>Xamarin.Android.Tools.JavadocImporter</Name>
-    </ProjectReference>
+    <ProjectReference Include="..\..\src\Xamarin.Android.Tools.JavadocImporter\Xamarin.Android.Tools.JavadocImporter.csproj" />
   </ItemGroup>
-  <Import Project="$(MSBuildBinPath)\Microsoft.CSharp.targets" />
+  <Target Name="_CopyOutput" Condition=" '$(TargetFramework)' == 'net472' " AfterTargets="Build">
```
jonpryor, line 16: This `<Target/>` should be moved into a new `Directory.Build.targets` file so that the next time e.g. Visual Studio for Mac decides to "rewrite" all the `.csproj` files, this target won't be re-indented into oblivion.

Documentation/guides/DotNet5.md
```diff
@@ -45,6 +71,14 @@ In .NET 5 the behavior of the following MSBuild tasks will change, but
 better align with existing .NET 5 semantics, at the cost of compatibility with
 previous Xamarin.Android releases.
 
+`$(AndroidClassParser)` will be `class-parse` by default.
```
jonpryor, line 74: This should mention that not only is `class-parse` the default, but that `jar2xml` is not supported at all.

jpobst: Same for `$(AndroidCodegenTarget)`.

tools/javadoc2mdoc/javadoc2mdoc.csproj
jonathanpeppers: This was breaking because:
* `javadoc2mdoc` depends on `Xamarin.Android.Tools.JavadocImporter`
* `Xamarin.Android.Tools.JavadocImporter` uses `libZipSharp`
* It caused us to ship the `net45` version of `libZipSharp`, but `Xamarin.Android.Build.Tasks` was build against the `netstandard2.0` `libZipSharp`.

We might consider reworking how our build tree is created down the road. I feel like we could consider leaving `$(OutputPath)` default and having a way to copy outputs to the final tree.

PR #4652, opened 2020-05-04
[build] fix .NET Core install on Jenkins

jonathanpeppers:
I'm hoping to use the same tricks I used in:

https://github.com/xamarin/xamarin-android/pull/4644

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -180,10 +180,7 @@ run-apk-tests:
 	exit $$_r
 
 run-performance-tests:
-	_r=0 ; \
-	$(call MSBUILD_BINLOG,run-apk-tests,,Test) $(TEST_TARGETS) /t:RunApkTests /p:RunApkTestsTarget=RunPerformanceApkTests $(APK_TESTS_PROP) || _r=1 ; \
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
This does fix the .NET Core install, but more things are broken:
```
Xamarin.Android.Legacy.targets(223,5): error XA2002: Can not resolve reference: `Mono.Linq.Expressions`, referenced by `Java.Interop-Tests`. Please add a NuGet package or assembly reference for `Mono.Linq.Expressions`, or remove the reference to `Java.Interop-Tests`.
```
It seems like NuGet restore isn't happening for these performance tests?

Makefile
```diff
@@ -180,10 +180,7 @@ run-apk-tests:
 	exit $$_r
 
 run-performance-tests:
-	_r=0 ; \
-	$(call MSBUILD_BINLOG,run-apk-tests,,Test) $(TEST_TARGETS) /t:RunApkTests /p:RunApkTestsTarget=RunPerformanceApkTests $(APK_TESTS_PROP) || _r=1 ; \
```
jonpryor, line 184: Not sure why these changes are needed as part of a .NET Core installation change.

jonathanpeppers: I was mainly trying to get the build log shorter 👀 

It seems both the startup & MSBuild perf tests fail to build, which is odd.

PR #4654, opened 2020-05-04
[Xamarin.Android.Build.Tasks] simplify d8/r8 multi-dex support

jonathanpeppers:
Context: https://developer.android.com/studio/build/multidex#mdex-gradle

When `android:minSdkVersion` is 21 or higher, d8 can automatically
support multi-dex. It will merely generate multiple `classes.dex`
files as needed, and we won't have a need for an
`$(AndroidEnableMultiDex)` MSBuild property.

Changes when `$(AndroidDexTool)` is set to `d8`:

* We no longer use `android-support-multidex.jar`.
* We no longer use `android.support.multidex.MultiDexApplication`.
* `$(AndroidEnableMultiDex)` is not used for anything.

These changes are not made when `AndroidDexTool=dx`

This makes the `<R8/>` MSBuild task much simpler, as many of the
incoming values are no longer needed. `<R8/>` will now only be needed
when `AndroidLinkTool=r8` and developers are opting into code
shrinking.

### DIFF
--- a/Documentation/release-notes/4654.md
+++ b/Documentation/release-notes/4654.md
@@ -1,0 +1,31 @@
+### Multidex now handled automatically for projects using D8
+
+Because the minimum compatible Android version is now Android 5.0 Lollipop (API
+level 21), multidex configuration is now handled automatically for all
+Xamarin.Android projects that are configured to use the D8 DEX compiler.
+
+Any projects that have transitioned to D8 and are using the
+**MultiDexMainDexList** build action or the **Enable Multi-Dex** setting in the
+Visual Studio project property pages can now discard those settings.  These
+correspond to the `MultiDexMainDexList` MSBuild item type and the
+`AndroidEnableMultiDex` MSBuild property:
+
+```xml
+<ItemGroup>
+  <MultiDexMainDexList Include="multidex-config.txt" />
+</ItemGroup>
+<PropertyGroup>
+  <AndroidEnableMultiDex>true</AndroidEnableMultiDex>
+</PropertyGroup>
+```
+
+#### Background information
+
+Android 5.0 Lollipop (API level 21) and higher use the ART runtime, which
+pre-compiles DEX files during app installation a single _.oat_ file, so the
+Android build process automatically enables and configures multidex for apps
+that have a `minSdkVersion` of Android 5.0 (API level 21) or higher.
+
+See the [multidex documentation][multidex] for additional details.
+
+[multidex]: https://developer.android.com/studio/build/multidex#mdex-on-l
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -1348,56 +1348,6 @@ namespace UnnamedProject {
 		}
 
 		[Test]
-		public void MultiDexR8ConfigWithNoCodeShrinking ([Values (true, false)] bool useConfig)
-		{
-			var proj = new XamarinAndroidApplicationProject () {
-				IsRelease = true,
-				DexTool = "d8",
-			};
-			proj.SetProperty ("AndroidEnableMultiDex", "True");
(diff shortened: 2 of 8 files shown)


jonathanpeppers:
FYI, I don't know how much will break here yet.

brendanzagaeski:
*Draft release notes*

As this PR is finalized for merge, when you get a chance, you can add a release note for it to `Documentation/release-notes/4654.md` as part of the PR. Here's a candidate note based on the context about the new minimum API level from [Documentation/release-notes/4584.md](https://github.com/xamarin/xamarin-android/blob/567ef51014a010db5e5e86e93fdc3ef5e6367315/Documentation/release-notes/4584.md). This candidate note can be copied verbatim into `4654.md` if it looks OK. Thanks!

````
### Multidex now handled automatically for projects using D8

Because the minimum compatible Android version is now Android 5.0 Lollipop (API
level 21), multidex configuration is now handled automatically for all
Xamarin.Android projects that are configured to use the D8 DEX compiler.

Any projects that have transitioned to D8 and are using the
**MultiDexMainDexList** build action or the **Enable Multi-Dex** setting in the
Visual Studio project property pages can now discard those settings.  These
correspond to the `MultiDexMainDexList` MSBuild item type and the
`AndroidEnableMultiDex` MSBuild property:

```xml
<ItemGroup>
  <MultiDexMainDexList Include="multidex-config.txt" />
</ItemGroup>
<PropertyGroup>
  <AndroidEnableMultiDex>true</AndroidEnableMultiDex>
</PropertyGroup>
```

#### Background information

Android 5.0 Lollipop (API level 21) and higher use the ART runtime, which
pre-compiles DEX files during app installation a single _.oat_ file, so the
Android build process automatically enables and configures multidex for apps
that have a `minSdkVersion` of Android 5.0 (API level 21) or higher.

See the [multidex documentation][multidex] for additional details.

[multidex]: https://developer.android.com/studio/build/multidex#mdex-on-l
````

File: src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs [hunk 1]
```diff
@@ -1348,7 +1346,7 @@ public void MultiDexAndCodeShrinker ([Values ("proguard", "r8")] string linkTool
 		}
 
 		[Test]
-		public void MultiDexR8ConfigWithNoCodeShrinking ([Values (true, false)] bool useConfig)
+		public void MultiDexR8ConfigWithNoCodeShrinking ()
```
brendanzagaeski, line 1349: It occurs to me that this PR fully supersedes the changes from <https://github.com/xamarin/xamarin-android/commit/78dd02167e3b974f246ed1e1f5b386e698c15c0f>, so I think we can delete this test altogether now.

In fact, if it might be OK, maybe this PR could include a full revert of <https://github.com/xamarin/xamarin-android/commit/78dd02167e3b974f246ed1e1f5b386e698c15c0f>? That could be nice because it would remove remove the obsolete extra conditions for `@(ProguardConfigFiles)` in `Xamarin.Android.Common.targets`, and it would also remove the obsolete release note.

Documentation/release-notes/4654.md
```diff
@@ -0,0 +1,31 @@
+### Multidex now handled automatically for projects using D8
```
brendanzagaeski, line 1: Release note approved. Thanks!

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
jonathanpeppers: I added a commit to specifically revert https://github.com/xamarin/xamarin-android/commit/78dd02167e3b974f246ed1e1f5b386e698c15c0f, is it fine that it deletes the release notes?

brendanzagaeski: Yup. It's even kind of handy that it removes the release notes too. Thanks!

PR #4656, opened 2020-05-06
[typemaps] Handle Java to managed type maps properly

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/4660
Context: https://github.com/xamarin/monodroid/commit/192b1f1b71c98650d04bf0a4e52ee337a054ff4c
Context: https://github.com/xamarin/java.interop/blob/fc18c54b2ccf14f444fadac0a1e7e81f837cc470/src/Java.Interop.Tools.JavaCallableWrappers/Java.Interop.Tools.JavaCallableWrappers/TypeNameMapGenerator.cs#L149-L186
Context: https://github.com/xamarin/java.interop/blob/010161d1ef6625b762429334f02e7b15e1f7caae/src/Java.Interop.Tools.JavaCallableWrappers/Java.Interop.Tools.JavaCallableWrappers/TypeNameMapGenerator.cs#L155

~~ Debug Mode ~~

In some environments, `BundleTest.TestBundleIntegerArrayList2()` fails
when  using the commercial shared runtime:

	Test 'Xamarin.Android.RuntimeTests.BundleTest.TestBundleIntegerArrayList2' failed: System.MemberAccessException : Cannot create an instance of Android.Runtime.JavaList`1[T] because Type.ContainsGenericParameters is true.

The problem appears to be rooted in ce2bc689.  In a pre-ce2bc689 world,
there were two separate sets of "typemap" files that the application
could use:

  * Java-to-managed mappings, in `typemap.jm`
  * Managed-to-Java mappings, in `typemap.mj`.

These files did *not* need to have the same number of entries!

	% unzip /Library/Frameworks/Xamarin.Android.framework/Versions/10.2.0.100/lib/xamarin.android/xbuild/Xamarin/Android/Mono.Android.DebugRuntime-debug.apk typemap.mj
	% strings typemap.jm | wc -l
	   10318
	% strings typemap.mj | wc -l
	   11956

The reason why they have a different number of entries is *aliasing*:
there can be *multiple* bindings for a given Java type.  The
managed-to-Java mappings can contain *all* of them; the
Java-to-managed mapping *must* pick Only One.

For example, [`java.util.ArrayList`][0] contains (at least?) *three* bindings:

  * [`Android.Runtime.JavaList`][1]
  * [`Android.Runtime.JavaList<T>`][2]

In a pre-ce2bc689 world, this was "fine": we'd binary search each file
*separately* to find type mappings.

This changed in ce2bc689, as the typemap information was merged into a
*single* array in order to de-duplicate information.  This reduced
`.apk` size.

An unanticipated result of this is that the Java-to-managed mappings
can now contain *duplicate* keys, "as if" the original `typemap.jm`
had the contents:

  java/util/ArrayList
  java/util/ArrayList
  java/util/ArrayList

Whereas pre-ce2bc689 there would have only been the first entry.

"Sometimes" this duplication is "fine": the typemap search
"Just happens" to return the first entry, or the 3rd entry, and apps
continue to work as intended.

"Sometimes" it isn't: the typemap binary search finds the 2nd entry,
which is a generic type.  This results in attempting to instantiate a
generic type *without appropriate type parameters*, which results in
the aforementioned `MemberAccessException`.

Update typemap generation code in `Xamarin.Android.Build.Tasks` so that
all the duplicate Java type names will point to the same managed type
name.  Additionally, make sure we select the managed type in the same
fashion the old typemap generator in `Java.Interop` worked - prefer base
types to the derived ones if the type is an interface or an abstract
class.  The effect of this change is that no matter which entry
`binary_search` ends up selecting it will always return the same managed
type name for all the aliased Java types.

~~ Release mode ~~

Another issue introduced in ce2bc689 is that we no longer ignore
interface and generic types when generating the Java-to-managed maps.
This adversely affects the `Release` mode in which it is possible that
an attempt to instantiate an open generic type (or an interface) will
happen, leading to error similar to:

	FATAL EXCEPTION: main
	  Process: com.companyname.androidsingleviewapp1, PID: 24068
	  android.runtime.JavaProxyThrowable: System.MemberAccessException: Cannot create an instance of Com.Contoso.Javalibrary1.Class0`1[T] because Type.ContainsGenericParameters is true.
	    at Com.Contoso.Javalibrary1.Class1.Create ()
	    at AndroidSingleViewApp1.MainActivity.OnCreate (Android.OS.Bundle savedInstanceState)
	    at Android.App.Activity.n_OnCreate_Landroid_os_Bundle_ (System.IntPtr jnienv, System.IntPtr native__this, System.IntPtr native_savedInstanceState)
	    at (wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.1(intptr,intptr,intptr)
	         at crc647d7dcab8da8ee782.MainActivity.n_onCreate(Native Method)
	         at crc647d7dcab8da8ee782.MainActivity.onCreate(MainActivity.java:29)
	         at android.app.Activity.performCreate(Activity.java:7144)
	         at android.app.Activity.performCreate(Activity.java:7135)
	         at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1271)
	         at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:2931)
	         at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3086)
	         at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:78)
	         at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:108)
	         at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:68)
	         at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1816)
	         at android.os.Handler.dispatchMessage(Handler.java:106)
	         at android.os.Looper.loop(Looper.java:193)
	         at android.app.ActivityThread.main(ActivityThread.java:6718)
	         at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
	         at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:858)

Update typemap generation code to also ignore generic types and
interfaces when generating the Java to Managed map in `Release` mode. We
must not simply skip outputting entries for such types because we need
to maintain element count parity between java-to-managed and
managed-to-java lookup tables.  Therefore, the way we skip such types is
to output a type token ID of `0` instead of the actual one.  This will
cause the runtime code to fail to find a mapping, thus allowing the
managed code to deal with such a type properly.

~~ Test Updates ~~

Update `TestBundleIntegerArrayList2()` so that it asserts
the runtime *type* of `obj` returned, asserting that it is in fact a
`JavaList` and not e.g. `Java.Util.ArrayList` or something.
(The linker could otherwise "change" things, and this assertion will
ensure that `JavaList` won't be linked away…)

Add a unit test for issue #4660 (by @brendanzagaeski)

It's very cute. :-)

We sort the Java-to-managed type name table, so it can be binary
searched to find the Managed type.

Additionally, the "preferred" managed type is the *first* type in
`monodis --typedef` output; the first type that we encounter when
going through all types in an assembly.

Brendan's cute way of provoking the bug is to add a new *generic* type
which will come *before* the "normally preferred" (and correct!) type.
Thus, even if/when binary search returns the first entry, that'll
still be the *wrong entry*, because that'll be a generic type!

Very cute.

[0]: https://developer.android.com/reference/java/util/ArrayList
[1]: https://github.com/xamarin/xamarin-android/blob/61f09bf2ef0c8f09550e2d77eb97f417432b315b/src/Mono.Android/Android.Runtime/JavaList.cs#L11-L13
[2]: https://github.com/xamarin/xamarin-android/blob/61f09bf2ef0c8f09550e2d77eb97f417432b315b/src/Mono.Android/Android.Runtime/JavaList.cs#L667-L668

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/TypeMapGenerator.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/TypeMapGenerator.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Text;
 
+using Java.Interop.Tools.Cecil;
 using Mono.Cecil;
 
 namespace Xamarin.Android.Tasks
@@ -14,6 +15,7 @@ namespace Xamarin.Android.Tasks
 		const string TypeMapIndexMagicString = "XATI"; // Xamarin Android Typemap Index
 		const uint TypeMapFormatVersion = 2; // Keep in sync with the value in src/monodroid/jni/xamarin-app.hh
 		const string TypemapExtension = ".typemap";
+		const uint InvalidJavaToManagedMappingIndex = UInt32.MaxValue;
 
 		internal sealed class ModuleUUIDArrayComparer : IComparer<ModuleReleaseData>
 		{
@@ -52,6 +54,7 @@ namespace Xamarin.Android.Tasks
 			public uint Token;
 			public int AssemblyNameIndex = -1;
 			public int ModuleIndex = -1;
+			public bool SkipInJavaToManaged;
 		}
 
 		internal sealed class ModuleReleaseData
@@ -76,6 +79,8 @@ namespace Xamarin.Android.Tasks
 			public string ManagedLabel;
 			public int JavaIndex;
 			public int ManagedIndex;
+			public TypeDefinition TypeDefinition;
+			public bool SkipInJavaToManaged;
 		}
 
 		// Widths include the terminating nul character but not the padding!
@@ -87,6 +92,7 @@ namespace Xamarin.Android.Tasks
 			public List<TypeMapDebugEntry> JavaToManagedMap;
 			public List<TypeMapDebugEntry> ManagedToJavaMap;
 			public string OutputFilePath;
+			public string ModuleName;
 			public byte[] ModuleNameBytes;
 		}
 
@@ -125,7 +131,7 @@ namespace Xamarin.Android.Tasks
 			}
 		}
 
-		public bool Generate (bool debugBuild, bool skipJniAddNativeMethodRegistrationAttributeScan, List<TypeDefinition> javaTypes, string outputDirectory, bool generateNativeAssembly, out ApplicationConfigTaskState appConfState)
+		public bool Generate (bool debugBuild, bool skipJniAddNativeMethodRegistrationAttributeScan, List<TypeDefinition> javaTypes, TypeDefinitionCache cache, string outputDirectory, bool generateNativeAssembly, out ApplicationConfigTaskState appConfState)
 		{
 			if (String.IsNullOrEmpty (outputDirectory))
 				throw new ArgumentException ("must not be null or empty", nameof (outputDirectory));
@@ -140,25 +146,26 @@ namespace Xamarin.Android.Tasks
 			string typemapsOutputDirectory = Path.Combine (outputDirectory, "typemaps");
 
 			if (debugBuild) {
-				return GenerateDebug (skipJniAddNativeMethodRegistrationAttributeScan, javaTypes, typemapsOutputDirectory, generateNativeAssembly, appConfState);
+				return GenerateDebug (skipJniAddNativeMethodRegistrationAttributeScan, javaTypes, cache, typemapsOutputDirectory, generateNativeAssembly, appConfState);
 			}
 
 			return GenerateRelease (skipJniAddNativeMethodRegistrationAttributeScan, javaTypes, typemapsOutputDirectory, appConfState);
 		}
 
-		bool GenerateDebug (bool skipJniAddNativeMethodRegistrationAttributeScan, List<TypeDefinition> javaTypes, string outputDirectory, bool generateNativeAssembly, ApplicationConfigTaskState appConfState)
+		bool GenerateDebug (bool skipJniAddNativeMethodRegistrationAttributeScan, List<TypeDefinition> javaTypes, TypeDefinitionCache cache, string outputDirectory, bool generateNativeAssembly, ApplicationConfigTaskState appConfState)
 		{
 			if (generateNativeAssembly)
-				return GenerateDebugNativeAssembly (skipJniAddNativeMethodRegistrationAttributeScan, javaTypes, outputDirectory, appConfState);
-			return GenerateDebugFiles (skipJniAddNativeMethodRegistrationAttributeScan, javaTypes, outputDirectory, appConfState);
+				return GenerateDebugNativeAssembly (skipJniAddNativeMethodRegistrationAttributeScan, javaTypes, cache, outputDirectory, appConfState);
+			return GenerateDebugFiles (skipJniAddNativeMethodRegistrationAttributeScan, javaTypes, cache, outputDirectory, appConfState);
 		}
 
-		bool GenerateDebugFiles (bool skipJniAddNativeMethodRegistrationAttributeScan, List<TypeDefinition> javaTypes, string outputDirectory, ApplicationConfigTaskState appConfState)
+		bool GenerateDebugFiles (bool skipJniAddNativeMethodRegistrationAttributeScan, List<TypeDefinition> javaTypes, TypeDefinitionCache cache, string outputDirectory, ApplicationConfigTaskState appConfState)
 		{
 			var modules = new Dictionary<string, ModuleDebugData> (StringComparer.Ordinal);
 			int maxModuleFileNameWidth = 0;
 			int maxModuleNameWidth = 0;
 
+			var javaDuplicates = new Dictionary<string, List<TypeMapDebugEntry>> (StringComparer.Ordinal);
 			foreach (TypeDefinition td in javaTypes) {
 				UpdateApplicationConfig (td, appConfState);
 				string moduleName = td.Module.Assembly.Name.Name;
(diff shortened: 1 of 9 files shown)


grendello:
Commit message snatched from https://github.com/xamarin/xamarin-android/pull/4645

src/Xamarin.Android.Build.Tasks/Utilities/TypeMapGenerator.cs
```diff
@@ -329,12 +377,18 @@ bool GenerateRelease (bool skipJniAddNativeMethodRegistrationAttributeScan, List
...
+				// to output `0` as the token id for the type, thus effectively causing the runtime unable to match such
+				// a Java type name to a managed type. This fixes https://github.com/xamarin/xamarin-android/issues/4660
 				var entry = new TypeMapReleaseEntry {
 					JavaName = javaName,
 					JavaNameLength = outputEncoding.GetByteCount (javaName),
 					ManagedTypeName = td.FullName,
 					Token = td.MetadataToken.ToUInt32 (),
-					AssemblyNameIndex = knownAssemblies [assemblyName]
+					AssemblyNameIndex = knownAssemblies [assemblyName],
+					SkipInJavaToManaged = td.IsInterface | td.HasGenericParameters,
```
jonpryor, line 391: Shouldn't this be `||`, not `|`?

grendello: it can be either, since it's a `bool` it will work both ways, but I guess `||` would be more conventional...

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -208,6 +209,10 @@ EmbeddedAssemblies::typemap_java_to_managed (const char *java_type_name)
 	}
 
 	const char *managed_type_name = entry->to;
+	if (managed_type_name == nullptr) {
+		log_debug (LOG_DEFAULT, "typemap: Java type '%s' maps either to an open generic type or an interface type.");
```
jonpryor, line 213: This should probably use `LOG_ASSEMBLY` as `LOG_DEFAULT` requires an `adb shell setprop debug.mono.log default` to see the messages, which just *looks* weird.

src/Xamarin.Android.Build.Tasks/Utilities/TypeMapGenerator.cs
```diff
@@ -491,6 +554,10 @@ void OutputModule (ModuleDebugData moduleData)
 		//
 		void OutputModule (BinaryWriter bw, ModuleDebugData moduleData)
 		{
+			if ((uint)moduleData.JavaToManagedMap.Count == UInt32.MaxValue) {
```
jonpryor, line 557: This feels like dead/impossible code.  Sure, there's "defense in depth," but this feels like it's bordering on lunacy, as `List<T>` is backed by an array which can hold at most [`Array.MaxArrayLength`](https://github.com/dotnet/runtime/blob/0ec974709e311ef63e7e23ed7c74f1f33762aecc/src/libraries/System.Private.CoreLib/src/System/Array.cs#L27) elements.

This is admittedly less than `int.MaxValue`, and is an implementation detail.

Allowing `List<T>.Capacity` and `List<T>.Count` to *exceed* `int.MaxValue` would be a *yuge* semantic break, though.  Checking that `int.MaxValue` is less than `uint.MaxValue` is bizarre; it can't *not* be true.

(And if by some chance somebody *wanted* to allow `List<T>`/Arrays to hold more than int.MaxValue elements -- as mono could do at one point in time! -- you'd have to have `Count` *throw an exception* when Count exceeded `int.MaxValue`, otherwise currently semantically correct code that checks that `list.Count > 0` *will **fail***, which is very much not good.  Which means that *this* code would be *doubly* broken!)

If anything, use `int.MaxValue` as the limit, not `uint.MaxValue`.  Assuming we need this check at all.

src/Xamarin.Android.Build.Tasks/Utilities/TypeMapGenerator.cs
```diff
@@ -502,7 +569,7 @@ void OutputModule (BinaryWriter bw, ModuleDebugData moduleData)
 			foreach (TypeMapDebugEntry entry in moduleData.JavaToManagedMap) {
 				bw.Write (outputEncoding.GetBytes (entry.JavaName));
 				PadField (bw, entry.JavaName.Length, (int)moduleData.JavaNameWidth);
-				bw.Write (entry.ManagedIndex);
+				bw.Write (entry.SkipInJavaToManaged ? UInt32.MaxValue : (uint)entry.ManagedIndex);
```
jonpryor, line 572: This makes it look like we're using `uint.MaxValue` as a "placeholder" value for an "invalid" mapping.  If so, we should semantically name this constant:

```csharp
const uint InvalidJavaToManagedMappingIndex = uint.MaxValue;
```

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -736,7 +741,12 @@ EmbeddedAssemblies::typemap_load_file (BinaryTypeMapHeader &header, const char *
 		cur->from = reinterpret_cast<char*>(java_pos);
 
 		uint32_t idx = *(reinterpret_cast<uint32_t*>(java_pos + header.java_name_width));
-		cur->to = reinterpret_cast<char*>(managed_start + (managed_entry_size * idx));
+		if (idx < UINT32_MAX) {
```
jonpryor, line 744: "Ditto" here, instead of `UINT32_MAX`, we should use a semantically meaningful name, e.g. `INVALID_TARGET_INDEX`.

src/Xamarin.Android.Build.Tasks/Utilities/TypeMapGenerator.cs
grendello: It's a check for an "impossible" situation, for sure. But I don't want to rely on implementation details of `List<>` to check a condition that may affect us on the runtime. Also, if for some weird reason (a bug?) `.Count` is set to `-1`, then casting it to `uint` will yield `uint.MaxValue`. However, that is not the point of all this - `uint.MaxValue` is a special value used by the native runtime to determine that the entry should be skipped (is considered as "invalid") and this check makes this condition more explicit while not hurting anything.

PR #4658, opened 2020-05-06
[Mono.Android] Don't download or process Android platforms we no longer support.

jpobst:


jpobst:
Closing this.  Even though we do not support the previous versions we do rely on their history to produce API compatible bindings.

PR #4663, opened 2020-05-07
[.NET 5] NuGet workarounds for Xamarin.Forms projects

jonathanpeppers:
Context: https://github.com/xamarin/net5-samples/blob/d525d8a2d60700f52f86372c47e83d646d800aa4/Directory.Android.targets

Currently, .NET 5 will not restore *existing* Xamarin.Android NuGet
packages. It does not know to map `netcoreapp5.0` to
`MonoAndroid10.0`, `MonoAndroid9` etc.

We'll need to work around this until this lands:

* https://github.com/NuGet/NuGet.Client/pull/3339

To get a Xamarin.Forms project building & running *at all* in
`xamarin/net5-samples`, I had to:

1. Change `$(AssetTargetFallback)`
2. Write an MSBuild target to replace any instances of
   `netstandard2.0\Xamarin.Forms.Platform.dll` with
   `MonoAndroid10.0\Xamarin.Forms.Platform.dll`.
3. Add additional platform-specific assemblies in
   `$(PkgXamarin_Forms)\lib\MonoAndroid10.0\`

There are still some issues with this:

1. You get a lot of `NU1701` warnings due to `$(AssetTargetFallback)`.
2. Due to: https://github.com/NuGet/docs.microsoft.com-nuget/issues/1955
   You have to manually list *every* transitive dependency, which
   totals around ~36 packages for AndroidX.

These drawbacks make it completely impractical to run some subset of
our existing MSBuild tests on .NET 5. We would have to list the entire
tree of `<PackageReference/>` for every test.

Since we might be waiting a bit of time for this, I have been able to
come up with some workarounds to solve these problems for now:

1. Add a `Microsoft.Android.Sdk.NuGet.targets` with workarounds that
   we will completely remove down the road (I hope!).
2. Use `$(PackageTargetFallback)` instead of `$(AssetTargetFallback)`.
   It does not emit `NU1701` warnings, and transitive dependencies
   work! It is a deprecated MSBuild property, but should be fine as a
   workaround.
3. Write a `_FixupNuGetReferences` MSBuild target that runs after
   `ResolvePackageAssets`. A `<FixupNuGetReferences/>` MSBuild task
   will remove any `netstandard2.0` assemblies where platform-specific
   ones exist. It also needs to add platform-specific assemblies that
   are missing.

With this in place, a Xamarin.Forms `.csproj` with no hacks works!

```xml
 <Project Sdk="Microsoft.Android.Sdk">
   <PropertyGroup>
    <TargetFramework>netcoreapp5.0</TargetFramework>
    <OutputType>Exe</OutputType>
   </PropertyGroup>
   <ItemGroup>
    <PackageReference Include="Xamarin.Forms" Version="4.5.0.617" />
   </ItemGroup>
 </Project>
```

I created a new `XamarinFormsXASdkProject` class for use in our
existing .NET 5 MSBuild tests.

Other changes:

* `$(XamarinAndroidVersion)` needs to be filled out for the AndroidX
  MSBuild targets:
  https://github.com/xamarin/XamarinAndroidXMigration/blob/ea130ca0d0e9b3e20edccec02364f36da11ada6b/source/Xamarin.AndroidX.Migration/BuildTasks/Xamarin.AndroidX.Migration.targets#L107
* `$(RuntimeIdentifier)` should have a default value for application
  projects.
* Some tweaks to make `XASdkProject` more flexible.
* `Xamarin.ProjectTools.DotNetStandard` projects now support
  `<Import/>`.
* Added AndroidX-compatible `Tabbar.xml` and `Toolbar.xml`

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/FixupNuGetReferences.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/FixupNuGetReferences.cs
@@ -1,0 +1,78 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using Microsoft.Build.Framework;
+
+namespace Xamarin.Android.Tasks
+{
+	/// <summary>
+	/// This task contains *temporary* workarounds for NuGet in .NET 5.
+	/// </summary>
+	public class FixupNuGetReferences : AndroidTask
+	{
+		public override string TaskPrefix => "FNR";
+
+		[Required]
+		public string [] PackageTargetFallback { get; set; }
+
+		public ITaskItem [] CopyLocalItems { get; set; }
+
+		[Output]
+		public string [] AssembliesToAdd { get; set; }
+
+		[Output]
+		public ITaskItem [] AssembliesToRemove { get; set; }
+
+		public override bool RunTask ()
+		{
+			if (CopyLocalItems == null || CopyLocalItems.Length == 0)
+				return true;
+
+			var assembliesToAdd     = new Dictionary<string, string> ();
+			var assembliesToRemove  = new List<ITaskItem> ();
+			var fallbackDirectories = new HashSet<string> ();
+
+			foreach (var item in CopyLocalItems) {
+				var directory = Path.GetDirectoryName (item.ItemSpec);
+				var directoryName = Path.GetFileName (directory);
+				Log.LogDebugMessage ($"{directoryName} -> {item.ItemSpec}");
+				if (directoryName == "netstandard2.0") {
+					var parent = Directory.GetParent (directory);
+					foreach (var nugetDirectory in parent.EnumerateDirectories ()) {
+						var name = Path.GetFileName (nugetDirectory.Name);
+						foreach (var fallback in PackageTargetFallback) {
+							if (!string.Equals (name, fallback, StringComparison.OrdinalIgnoreCase))
+								continue;
+							var fallbackDirectory = Path.Combine (parent.FullName, fallback);
+							fallbackDirectories.Add (fallbackDirectory);
+
+							// Remove the netstandard assembly, if there is a platform-specific one
+							var path = Path.Combine (fallbackDirectory, Path.GetFileName (item.ItemSpec));
+							if (File.Exists (path)) {
+								Log.LogDebugMessage ($"Removing: {item.ItemSpec}");
+								assembliesToRemove.Add (item);
+							}
+						}
+					}
+				}
+			}
+
+			// Look for any platform-specific assemblies
+			foreach (var directory in fallbackDirectories) {
+				foreach (var assembly in Directory.GetFiles (directory, "*.dll")) {
+					var assemblyName = Path.GetFileName (assembly);
+					if (!assembliesToAdd.ContainsKey (assemblyName)) {
+						Log.LogDebugMessage ($"Adding: {assembly}");
+						assembliesToAdd.Add (assemblyName, assembly);
+					}
+				}
+			}
+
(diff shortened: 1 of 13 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
```diff
@@ -49,6 +49,7 @@
   <Import Project="Microsoft.Android.Sdk.DefaultItems.props" />
 
   <PropertyGroup>
+    <XamarinAndroidVersion>$(AndroidNETSdkVersion)</XamarinAndroidVersion>
```
pjcollins, line 52: It may make sense to move this into https://github.com/xamarin/xamarin-android/blob/47c260816e4a74304f0cad25d348eb2926a43de1/build-tools/create-packs/Microsoft.Android.Sdk.proj#L104 just to keep these version props together, though I'm not sure if that would have any other downside.

src/Xamarin.Android.Build.Tasks/Tasks/FixupNuGetReferences.cs
```diff
@@ -0,0 +1,78 @@
...
+							if (!string.Equals (name, fallback, StringComparison.OrdinalIgnoreCase))
+								continue;
+							var fallbackDirectory = Path.Combine (parent.FullName, fallback);
+							fallbackDirectories.Add (fallbackDirectory);
+
+							// Remove the netstandard assembly, if there is a platform-specific one
+							var path = Path.Combine (fallbackDirectory, Path.GetFileName (item.ItemSpec));
+							if (File.Exists (path)) {
+								Log.LogDebugMessage ($"Removing: {item.ItemSpec}");
+								assembliesToRemove.Add (item);
```
dellis1972, line 54: Should we set a flag so we break out of the `EnumerateDirectories` loop if we find a match? reading this code it looks like we will continue to enumerate directories

jonathanpeppers: The latest Forms NuGet package has multiple `MonoAndroid` directories:
```
> 7z l 'C:\Users\jopepper\Downloads\xamarin.forms.4.6.0.726.nupkg' | grep Xamarin.Forms.Platform.dll
2020-04-28 02:01:24 .....       138880        45480  lib\MonoAndroid10.0\Xamarin.Forms.Platform.dll
2020-04-28 02:01:30 .....       127104        43270  lib\MonoAndroid90\Xamarin.Forms.Platform.dll
2020-04-28 02:01:34 .....        13952         7171  lib\netstandard1.0\Xamarin.Forms.Platform.dll
2020-04-28 02:01:26 .....        13952         7164  lib\netstandard2.0\Xamarin.Forms.Platform.dll
2020-04-28 02:01:30 .....        19584         7924  lib\tizen40\Xamarin.Forms.Platform.dll
2020-04-28 02:01:36 .....        13952         7173  lib\uap10.0.14393\Xamarin.Forms.Platform.dll
2020-04-28 02:01:28 .....        13952         7164  lib\uap10.0.16299\Xamarin.Forms.Platform.dll
2020-04-28 02:01:26 .....        18560         7560  lib\Xamarin.iOS10\Xamarin.Forms.Platform.dll
2020-04-28 02:01:34 .....        13952         7166  lib\Xamarin.Mac\Xamarin.Forms.Platform.dll
```
I think we have to continue the loop in case the `MonoAndroid90` is the one that needs removed...

This is _hopefully_ going to be a hack we can remove later.

Issue #4675, opened 2020-05-12
"The project ... is missing Android SDKs required for building." blocks building without NDK for Enable Startup Tracing and AOT Compilation


brendanzagaeski:
### Steps to reproduce

1. Create a new **Android App (Xamarin) > Single View App**.
2. Set the active configuration to Release.
3. In the Visual Studio project property pages, turn on **Enable Startup Tracing** (`AndroidEnableProfiledAot`) or **AOT Compilation** (`AotAssemblies`).
4. When prompted for **Android SDK - License Acceptance**, click **Cancel**.
5. Attempt to build the project via **Build > Build Solution**.

### Expected behavior

The build completes successfully because [in Xamarin.Android 10.0 (d16-3) and higher, the Android NDK is not required](https://docs.microsoft.com/xamarin/android/release-notes/10/10.0#enable-startup-tracing-no-longer-requires-the-android-ndk) unless **Use LLVM Optimizing Compiler** (`EnableLLVM`) or **Bundle assemblies into native code** (`BundleAssemblies`) is enabled.

### Actual behavior

In Visual Studio 2019 version 16.4 and higher, due to the new support for installation of the NDK by the **Auto Install Android SDKs** feature, the IDE blocks the build.

The **Error List** shows:

```
The project AndroidApp1 is missing Android SDKs required for building. Double-click on this message and follow the prompts to install them.
```

The **Output** window shows:

```
Project 'AndroidApp1' requires the following components installed on your machine: 
NDK

In order to resolve this, please go to the Error List window, make sure "Build + IntelliSense" is selected and double-click on the error above.
Build has been canceled.
```

### Preliminary investigation

This behavior arises because the `GetAndroidDependencies` target returns that the Android NDK is needed.

```cmd
msbuild -t:GetAndroidDependencies -p:Configuration=Release AndroidApp1\AndroidApp1.csproj
```

```
Target "GetAndroidDependencies: (TargetId:9)" in file "C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets" from project "C:\Temp\AndroidApp1\AndroidApp1\AndroidApp1.csproj" (entry point):
Set Property: _ProjectAndroidManifest=C:\Temp\AndroidApp1\AndroidApp1\Properties\AndroidManifest.xml
Set Property: _NdkRequired=true
[...]

PR #4676, opened 2020-05-12
[Xamarin.Android.Build.Tasks] Update NDK dependency check

brendanzagaeski:
Context: f0b1e8a53054c5620b256b7f5613ef5b3610146c
Fixes: https://github.com/xamarin/xamarin-android/issues/4675

Thanks to commit f0b1e8a5, Xamarin.Android 10.0 and higher no longer
require the NDK when `$(AotAssemblies)`=`true` or
`$(AndroidEnableProfiledAot)`=`true`.  The NDK is now only required when
`$(EnableLLVM)`=`true` or when `$(BundleAssemblies)`=`true`.

Update `GetAndroidDependencies` to account for this new behavior.  This
allows users who have `$(AotAssemblies)`=`true` or
`$(AndroidEnableProfiledAot)`=`true` to build successfully from within
Visual Studio without being forced to install the NDK.  Previously, this
scenario was blocked by an error:

	The project AndroidApp1 is missing Android SDKs required for building. Double-click on this message and follow the prompts to install them.

This error provided the following additional information in the Output
window:

	Project 'AndroidApp1' requires the following components installed on your machine:
	NDK

	In order to resolve this, please go to the Error List window, make sure "Build + IntelliSense" is selected and double-click on the error above.
	Build has been canceled.

### DIFF
--- a/Documentation/release-notes/fix-4675.md
+++ b/Documentation/release-notes/fix-4675.md
@@ -1,0 +1,6 @@
+#### IDE compatibility
+
+- [GitHub 4675](https://github.com/xamarin/xamarin-android/issues/4675):
+  _The project ... is missing Android SDKs required for building._ prevented
+  building without the Android NDK installed for projects with **Enable Startup
+  Tracing** or **AOT Compilation** enabled.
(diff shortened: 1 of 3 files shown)


brendanzagaeski:
Interestingly, it seems the new behavior isn't yet working as expected on macOS since the new test in this PR isn't yet passing there.

I'll take a look at that.

brendanzagaeski:

Correct the new test to check only the relevant section of the build output so that it will pass on macOS too.

Correct a typo in the commit message.

PR #4678, opened 2020-05-12
[tests] One .NET MSBuild test infrastructure

jonathanpeppers:
As we transition to One .NET, we need a way to run our existing
MSBuild tests with `dotnet build` instead of `msbuild`.

To do this:

* Add a new `dotnet` category, on a subset of tests. Eventually, we
  may be able to run all tests, but not quite yet.
* Add a `--params dotnet=true` parameter to the `nunit3-console`
  invocation.
* `Xamarin.ProjectTools` does not have a reference to NUnit, so create
  a `Builder.UseDotNet` static property that is set in `BaseTest`.
* `Builder.UseDotNet` is used throughout `Xamarin.ProjectTools` to
  change the contents of generated projects.

For now, this enables 35 passing tests. We will fix issues and enable
more of them moving forward.

Workarounds:

* We have to set `$(XamarinAndroidSupportSkipVerifyVersions)` for
  existing support libraries to be used.
* We have to use a `Directory.Build.targets` and
  `$(VectorDrawableCheckBuildToolsVersionTaskBeforeTargets)` for
  existing support libraries.

Other fixes:

* Cleaned up `BuildBasicApplication`, so it is parameterized.
* Cleaned up `CheckWhichRuntimeIsIncluded`, `supportedAbi` does not
  need to be a `string[]`.
* `$(MonoSymbolArchive)` should default to `false`, since
  `mono-symbolicate` is not yet available for .NET 5. This fixed a few
  tests around `Release` builds.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidProject.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidProject.cs
@@ -13,22 +13,22 @@ namespace Xamarin.ProjectTools
 			: base (debugConfigurationName, releaseConfigurationName)
 		{
 			Language = XamarinAndroidProjectLanguage.CSharp;
-			TargetFrameworkVersion = Versions.Android10;
-			UseLatestPlatformSdk = true;
-			AddReferences ("System.Core", "System.Xml", "Mono.Android");
-			ProjectGuid = Guid.NewGuid ().ToString ();
 
-			SetProperty ("ProjectTypeGuids", () => "{" + Language.ProjectTypeGuid + "};{" + ProjectTypeGuid + "}");
-			SetProperty ("ProjectGuid", () => "{" + ProjectGuid + "}");
+			if (!Builder.UseDotNet) {
+				TargetFrameworkVersion = Versions.Android10;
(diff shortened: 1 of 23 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidProject.cs
```diff
@@ -13,21 +13,21 @@ protected XamarinAndroidProject (string debugConfigurationName = "Debug", string
...
+			}
 
 			SetProperty (KnownProperties.OutputType, "Library");
-			SetProperty ("MonoAndroidAssetsPrefix", "Assets");
-			SetProperty ("MonoAndroidResourcePrefix", "Resources");
-
-			SetProperty (KnownProperties.AndroidUseLatestPlatformSdk, () => UseLatestPlatformSdk ? "True" : "False");
-			SetProperty (KnownProperties.TargetFrameworkVersion, () => TargetFrameworkVersion);
-
 			SetProperty (ReleaseProperties, KnownProperties.AndroidUseSharedRuntime, "False");
```
pjcollins, line 31: Should this be moved up into the `!Builder.UseDotNet` condition?

dellis1972:
fantastic :D

Issue #4677, opened 2020-05-12
InstallAndroidDependencies target does not install platforms or build-tools components on Windows


brendanzagaeski:
### Example steps to reproduce

Adjust the `InstallAndroidDependenciesTest` to assert that the target Android SDK install directories are created:

```patch
diff --git a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs
index 73c32881..62066963 100644
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs
@@ -24,6 +24,8 @@ namespace Xamarin.Android.Build.Tests
 					string defaultTarget = b.Target;
 					b.Target = "InstallAndroidDependencies";
 					Assert.IsTrue (b.Build (proj, parameters: new string [] { "AcceptAndroidSDKLicenses=true" }), "InstallAndroidDependencies should have succeeded.");
+					Assert.IsTrue (Directory.Exists (Path.Combine (sdkPath, "platforms")), "At least one platform should have been installed");
+					Assert.IsTrue (Directory.Exists (Path.Combine (sdkPath, "build-tools")), "At least one Build Tools version should have been installed");
 					b.Target = defaultTarget;
 					Assert.IsTrue (b.Build (proj), "build should have succeeded.");
 				}
```

### Expected behavior

The test passes when running on Windows. The corresponding build log should include lines that start with:

```
Downloading archive from ...
```

and some lines like:

```
Installing Android SDK component: Android SDK Platform 29 r4 [Platform: API 29]
```

### Actual behavior

The test fails on Windows. The build does not include any `Downloading archive` or `Installing Android SDK component` lines. Instead it shows:

```
No components need to be installed
```

### Preliminary investigation

[...]

PR #4681, opened 2020-05-13
[Xamarin.Android.Build.Tasks] Use Path.Combine() in <CalculateProjectDependencies/>

brendanzagaeski:
Fixes: https://github.com/xamarin/xamarin-android/issues/4677

`<CalculateProjectDependencies/>` was always using `/` as the path
separator for the `build-tools` and `platforms` output items:

	Task "CalculateProjectDependencies"
	  Task Parameter:TargetFrameworkVersion=v10.0
	  Task Parameter:ManifestFile=C:\Source\xamarin-android\bin\TestDebug\temp\InstallAndroidDependenciesTest\Project\Properties\AndroidManifest.xml
	  Task Parameter:BuildToolsVersion=29.0.2
	  Task Parameter:PlatformToolsVersion=29.0.5
	  Task Parameter:ToolsVersion=26.1.1
	  Task Parameter:NdkVersion=16.1
	  Task Parameter:NdkRequired=False
	  Output Item(s):
	      AndroidDependency=
	          platforms/android-29
	          build-tools/29.0.2
	                  Version=29.0.2
	          platform-tools
	                  Version=29.0.5
	          tools
	                  Version=26.1.1
	Done executing task "CalculateProjectDependencies".

This meant the commercial `<InstallAndroidDependencies/>` task wouldn't
ever install the `build-tools` or `platforms` components on Windows
because none of the candidate `IAndroidComponent.FileSystemPath` values
that were available via the SDK installer matched the input
`ITaskItem.ItemSpec` values.

Fix this by changing `<CalculateProjectDependencies/>` to use
`Path.Combine()` instead of `/`.

Test changes:

`InstallAndroidDependenciesTest` did not yet enforce the expected
behavior for this bug because `<ResolveSdks/>` was automatically falling
back to the default Android SDK installation.

Update `InstallAndroidDependenciesTest` to enforce the expected
behavior.

To minimize download requirements for the test, add a mock
`tools/source.properties` file to the target `android-sdk` directory
during each test run.  This reduces the total downloads from about 1 GB
to under 100 MB, and, because the `tools` component isn't needed for the
current default project configuration, `b.Build (proj)` still completes
successfully.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/CalculateProjectDependencies.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/CalculateProjectDependencies.cs
@@ -53,8 +53,8 @@ namespace Xamarin.Android.Tasks
 				manifestApiLevel = manifest.TargetSdkVersion ?? manifest.MinSdkVersion ?? DefaultMinSDKVersion;
 			}
 			var sdkVersion = Math.Max (targetApiLevel.Value, manifestApiLevel);
-			dependencies.Add (CreateAndroidDependency ($"platforms/android-{sdkVersion}", $""));
-			dependencies.Add (CreateAndroidDependency ($"build-tools/{BuildToolsVersion}", BuildToolsVersion));
+			dependencies.Add (CreateAndroidDependency (Path.Combine ("platforms", $"android-{sdkVersion}"), $""));
+			dependencies.Add (CreateAndroidDependency (Path.Combine ("build-tools", BuildToolsVersion), BuildToolsVersion));
 			if (!string.IsNullOrEmpty (PlatformToolsVersion)) {
 				dependencies.Add (CreateAndroidDependency ("platform-tools", PlatformToolsVersion));
 			}
--- a/Documentation/release-notes/fix-4677.md
+++ b/Documentation/release-notes/fix-4677.md
@@ -1,0 +1,6 @@
+#### Application and library build and deployment
+
+- [GitHub 4677](https://github.com/xamarin/xamarin-android/issues/4677):
+  On Windows, the `InstallAndroidDependencies` MSBuild target would not yet
+  install the expected Android SDK Platform or Android SDK Build-Tools
+  components in a newly configured command line build environment.
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs
@@ -1,5 +1,7 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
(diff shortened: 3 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/CalculateProjectDependencies.cs
```diff
@@ -53,8 +53,8 @@ public override bool RunTask ()
 				manifestApiLevel = manifest.TargetSdkVersion ?? manifest.MinSdkVersion ?? DefaultMinSDKVersion;
 			}
 			var sdkVersion = Math.Max (targetApiLevel.Value, manifestApiLevel);
-			dependencies.Add (CreateAndroidDependency ($"platforms/android-{sdkVersion}", $""));
-			dependencies.Add (CreateAndroidDependency ($"build-tools/{BuildToolsVersion}", BuildToolsVersion));
+			dependencies.Add (CreateAndroidDependency (Path.Combine ("platforms", $"android-{sdkVersion}"), $""));
```
dellis1972, line 56: This must have changed behaviour in the sdk installer. I seem to remember the spec required the system to use `/` as a separator :/

jonpryor: Relevant spec: https://microsoft-my.sharepoint.com/:w:/p/mhutch/ETjdNgT_yfJMszzuRRW2hUYB133Goo831A5U2rAFwgI8bA?e=77b68c4ccf5841139a242a0d99cd0368

Forward slash `/` is explicitly mentioned in the spec:

> For items in the main [Android repository manifest](https://dl-ssl.google.com/android/repository/repository2-1.xml), the ID shall be the path formatted in \*nix format, e.g. “platforms/android-25” or “ndk-bundle”. For [add-ons](http://dl-ssl.google.com/android/repository/addon.xml), the ID shall be the add-on’s “name-id” property. We will also have our own manifest format where the fileSystemPath will be exposed directly. 

We *do not* want to use `Path.Combine()` here.

dellis1972: So in that case the issue must be in the task which calls the SDK Manager or the SDK Manager code itself I guess.

dellis1972:
Interestingly on windows doing the following does not install anything

`/t:InstallAndroidDependencies /p:AndroidSdkDirectory=c:\Sandbox\sdk /p:AcceptAndroidSDKLicenses=true`

where the `sdk` folder is empty. I would expect it to install everything the project needs. This is what we are telling our customers, and it used to work. I have a feeling that something in the `InstallAndroidDependencies` or the sdk installer itself is not behaving correctly now. We should investigate that to fix the issue in the sdk-installer.

dellis1972:
It works fine on a Mac.

dellis1972:
This was fixed in https://github.com/xamarin/android-sdk-installer/pull/452 so we can close this PR.

PR #4696, opened 2020-05-18
[tests] fix PerformanceTest

jonathanpeppers:
Context: https://build.azdo.io/3744248

Some of the MSBuild performance timings have started failing.

It seems like it is a change on the build machines, as I've not seen
anything in the logs pointing out a regression. I merely bumped a few
of the times here.

I also fixed the failure related to missing test attachments:

    TearDown : System.IO.FileNotFoundException : Test attachment file path could not be found.

If one of the files was missing in `DeviceTest.cs`, then it would fail
getting any subsequent attachments due to the thrown exception.


jonathanpeppers:
The perf tests passed...

PR #4697, opened 2020-05-18
[Xamarin.Android.Build.Tasks] update API level warning messages

jonathanpeppers:
Context: https://support.google.com/googleplay/android-developer/answer/113469#targetsdk

As of August 1, 2019, Google Play requires API level 28. The `XA0113` warning was checking for API 26, so it needed to be updated.

The `XA0117` warning also needs to be updated as API 21 is Xamarin.Android's new supported minimum API level.

### DIFF
--- a/Documentation/guides/messages/xa0113.md
+++ b/Documentation/guides/messages/xa0113.md
@@ -7,13 +7,13 @@ ms.date: 02/28/2019
 
 ## Issue
 
-As of August 1st 2018, new applications and application updates uploaded
-to the Google Play store need to target v8.0 (API 26) or above.
+As of August 1st 2019, new applications and application updates uploaded
+to the Google Play store need to target v9.0 (API 28) or above.
 
 ## Solution
 
 If you are seeing this warning, you should update the
-`$(TargetFrameworkVersion)` of your projects to be v8.0 or above.
+`$(TargetFrameworkVersion)` of your projects to be v9.0 or above.
 
 If you are not targeting the Google Play store and wish to hide these
 warnings, you can make use of the `/warnasmessage:XA0113` command line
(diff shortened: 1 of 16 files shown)


brendanzagaeski:
*Draft release notes*

Here's the note I'll plan to include for this pull request:

````
### Updated XA0113 warning for Google Play submission requirements

The XA0113 warning has been updated to reflect the [current minimum target
version of Android 9 (API level 28)][targetsdk] for submissions to the Google
Play store.  The following warning will now appear for projects that have an
earlier version set under **Compile using Android version: (Target Framework)**
in the Visual Studio project properties pages:

```
warning XA0113: Google Play requires that new applications and updates must use a TargetFrameworkVersion of v9.0 (API level 28) or above. You are currently targeting v8.0 (API level 26).
```

[targetsdk]: https://support.google.com/googleplay/android-developer/answer/113469#targetsdk
````

To suggest any changes, feel free to reply on this PR. Thanks!

For the XA0117 adjustment, I realized that won't be user-visible in d16-7 because the target framework directories have already been removed from the installer, so the XA0001 error will be shown instead:

```
error XA0001: Unsupported or invalid $(TargetFrameworkVersion) value of 'v4.4'. Please update your Project Options.
```

It might be worth considering backporting the XA0117 adjustment to a d16-6 servicing update to match the approach taken in Xamarin.Android 9.1 to warn about the [removals that were coming up in Xamarin.Android 9.2](https://docs.microsoft.com/xamarin/android/release-notes/9/9.2#removal-of-target-framework-versions-below-android-44-api-level-19)?

brendanzagaeski:
> It might be worth considering backporting the XA0117 adjustment to a d16-6 servicing update to match the approach taken in Xamarin.Android 9.1 to warn about the removals that were coming up in Xamarin.Android 9.2?

Ah actually, that's probably not a good idea since a servicing update ideally should not _add_ any warnings to builds. I hadn't quite thought that through.

PR #4710, opened 2020-05-21
[Mono.Android] API-R enumification

jpobst:
Enumification for the API-R developer preview 4.

- Convert constant fields to enums as appropriate.
- Update methods that take an `int` constant value to use new enums.

There is some `acceptable-breakage` because I renamed some of the enums that were added in API-R with auto-generated names that had already been committed in https://github.com/xamarin/xamarin-android/pull/4468.  (Note that our enums are not "API-aware" so all enumifications are added for all API levels.  That is, enums added in API-R get added to `android-21`, `android-22`, etc..)


src/Mono.Android/map.csv
```diff
@@ -1,29 +1,29 @@
...
-?,30,android/accessibilityservice/AccessibilityService.GESTURE_4_FINGER_DOUBLE_TAP_AND_HOLD,42,,,,
-?,30,android/accessibilityservice/AccessibilityService.GESTURE_4_FINGER_SINGLE_TAP,37,,,,
-?,30,android/accessibilityservice/AccessibilityService.GESTURE_4_FINGER_SWIPE_DOWN,34,,,,
-?,30,android/accessibilityservice/AccessibilityService.GESTURE_4_FINGER_SWIPE_LEFT,35,,,,
-?,30,android/accessibilityservice/AccessibilityService.GESTURE_4_FINGER_SWIPE_RIGHT,36,,,,
-?,30,android/accessibilityservice/AccessibilityService.GESTURE_4_FINGER_SWIPE_UP,33,,,,
-?,30,android/accessibilityservice/AccessibilityService.GESTURE_4_FINGER_TRIPLE_TAP,39,,,,
-?,30,android/accessibilityservice/AccessibilityService.GESTURE_DOUBLE_TAP,17,,,,
-?,30,android/accessibilityservice/AccessibilityService.GESTURE_DOUBLE_TAP_AND_HOLD,18,,,,
+E,30,android/accessibilityservice/AccessibilityService.GESTURE_2_FINGER_DOUBLE_TAP,20,Android.AccessibilityServices.AccessibilityGesture,TwoFingerDoubleTap,remove,
```
jonpryor, line 1: Silly Q, but does the `map2.csv`-parsing code in generator trim whitespace, such that we could make this "nicer" by adding spaces before/after the commas to make things line up "nicely"?  Or is that just "crazy"?

```
E,30,android/accessibilityservice/AccessibilityService.GESTURE_2_FINGER_DOUBLE_TAP_AND_HOLD,40,Android.AccessibilityServices.AccessibilityGesture,TwoFingerDoubleTapAndHold,remove,
E,30,android/accessibilityservice/AccessibilityService.GESTURE_2_FINGER_SINGLE_TAP         ,19,Android.AccessibilityServices.AccessibilityGesture,TwoFingerSingleTap       ,remove,
E,30,android/accessibilityservice/AccessibilityService.GESTURE_2_FINGER_SWIPE_DOWN         ,26,Android.AccessibilityServices.AccessibilityGesture,TwoFingerSwipeDown       ,remove,
```

jpobst: It does trim, but I don't think it's going to make things nicer to read.  You would have to pad to at least the size of the longest line which is 217 characters.

jonpryor: You wouldn't necessarily need to pad the entire file, just each group of enums.  If you don't think it'll be nicer to read, though, there's no point.

PR #4711, opened 2020-05-21
[tests] improve PerformanceTest accuracy

jonathanpeppers:
Context: https://msbuildlog.com
Context: https://github.com/microsoft/msbuild/blob/ba9a1d64a7abf15a8505827c00413156a3eb7f62/src/Build/Logging/BinaryLogger/BuildEventArgsReader.cs

I've noticed that we have been including some MSBuild overhead in our
timings in `PerformanceTest`. If you use the MSBuild binlog viewer,
you can see the "project build time" can be *seconds* shorter than the
overall build time. This extra time is spent for MSBuild startup,
shutdown, and writing log files, etc. We can exclude this time for
these tests.

We can get this number by parsing the `.binlog` file directly, using a
`BuildEventArgsReader` MSBuild API. I just had to update our
`<PackageReference/>` to `Microsoft.Build.*` to a newer version.

Keeping a `Stack` of "project started" event, you can sum them as
"project finished" events occur. These events occur in a tree-like
fashion for project dependencies, so a `Stack` is needed to calculate
the values properly.

I also turned off the other log files, to improve the accuracy
further.

Builds in `PerformanceTest` will be running:

    msbuild /noconsolelogger /bl

This is should be optimal for what we need here.

I was able to drop each timing by a decent amount. I think this will
be more accurate for catching regressions and reduce random failures.

### DIFF
--- a/build-tools/scripts/MSBuildReferences.projitems
+++ b/build-tools/scripts/MSBuildReferences.projitems
@@ -1,15 +1,16 @@
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <MSBuildReferenceVersion>15.1.0.0</MSBuildReferenceVersion>
+    <MSBuildPackageReferenceVersion>16.5</MSBuildPackageReferenceVersion>
   </PropertyGroup>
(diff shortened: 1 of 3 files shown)


dellis1972:
Designer tests seem to be failing. They shouldn't be related though.

PR #4722, opened 2020-05-26
[Xamarin.Android.Build.Tasks] Catch Exception when setting Console.InputEncoding

dellis1972:
Context https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1130414

For some unknown reason the setting of `Console.InputEncoding`
on a customers machine is throwing an IOException. This then
causes Visual Studio to completely crash.

```
Description: The process was terminated due to an unhandled exception.
Exception Info: System.IO.IOException
   at System.IO.__Error.WinIOError(Int32, System.String)
   at System.IO.__Error.WinIOError()
   at System.Console.set_InputEncoding(System.Text.Encoding)
   at Xamarin.Android.Tasks.Aapt2Daemon.Aapt2DaemonStart()
   at System.Threading.ThreadHelper.ThreadStart_Context(System.Object)
   at System.Threading.ExecutionContext.RunInternal(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object, Boolean)
   at System.Threading.ExecutionContext.Run(System.Threading.ExecutionContext, System.Threading.ContextCallback, System.Object)
   at System.Threading.ThreadHelper.ThreadStart()
```

At this time we are still unsure as to what is causing the problem.
But we should at least catch the exception to that Visual Studio does
not crash.

We need to figure out a way of reporting this issue somehow in the
exception handler so we can get more data.

Note: Not sure what is going on with the line ending changes :/

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
@@ -57,6 +57,7 @@ namespace Xamarin.Android.Tasks
 		long jobsRunning = 0;
 		long jobId = 0;
 		int maxInstances = 0;
+		Queue<string> daemonStartupWarnings = new Queue<string> ();
 
 		public CancellationToken Token => tcs.Token;
 
@@ -76,6 +77,8 @@ namespace Xamarin.Android.Tasks
 		public int MaxInstances => maxInstances;
 
 		public int CurrentInstances => daemons.Count;
+
+		public Queue<string> StartupWarnings => daemonStartupWarnings;
 
 		public Aapt2Daemon (string aapt2, int maxNumberOfInstances, int initalNumberOfDaemons)
 		{
@@ -162,21 +165,34 @@ namespace Xamarin.Android.Tasks
 				// and we are using netstandard 2.0
 				//StandardInputEncoding = Encoding.UTF8,
 			};
-			// We need to FORCE the StandardInput to be UTF8 so we can use 
+			// We need to FORCE the StandardInput to be UTF8 so we can use
 			// accented characters. Also DONT INCLUDE A BOM!!
 			// otherwise aapt2 will try to interpret the BOM as an argument.
-			Process aapt2;
+			Process aapt2 = null;
 			lock (lockObject) {
-				Encoding current = Console.InputEncoding;
 				try {
-					Console.InputEncoding = new UTF8Encoding (false);
-					aapt2 = new Process ();
-					aapt2.StartInfo = info;
-					aapt2.Start ();
-				} finally {
-					Console.InputEncoding = current;
+					Encoding current = Console.InputEncoding;
+					try {
+						try {
+							Console.InputEncoding = new UTF8Encoding (false);
+						} catch (IOException ioEx) {
+							// For some reason we can not set the InputEncoding. If this happens we don't
+							// want to crash Visual Studio with an Unhandled Exception.
+							// The downside of ignoring this is paths with accented characters will cause problems.
+							daemonStartupWarnings.Enqueue (ioEx.ToString ());
+						}
+						aapt2 = new Process ();
+						aapt2.StartInfo = info;
+						aapt2.Start ();
+					} finally {
+						Console.InputEncoding = current;
+					}
+				} catch (Exception ex) {
+					daemonStartupWarnings.Enqueue (ex.ToString ());
 				}
 			}
+			if (aapt2 == null)
+				return;
 			try {
 				foreach (var job in pendingJobs.GetConsumingEnumerable (tcs.Token)) {
 					Interlocked.Add (ref jobsRunning, 1);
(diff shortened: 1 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
```diff
@@ -162,21 +165,32 @@ private void Aapt2DaemonStart ()
...
 			lock (lockObject) {
 				Encoding current = Console.InputEncoding;
 				try {
-					Console.InputEncoding = new UTF8Encoding (false);
+					try {
+						Console.InputEncoding = new UTF8Encoding (false);
+					} catch (IOException ioEx) {
+						// For some reason we can not set the InputEncoding. If this happens we don't
+						// want to crash Visual Studio with an Unhandled Exception.
+						// The downside of ignoring this is paths with accented characters will cause problems.
```
brendanzagaeski, line 180: If we end up with this PR as intermediate fix until a complete solution is found, then when you get a chance, if you could whip up a little known issue that describes the symptom users would see when using an accented character, that would be perfect.

Fingers crossed that the complete solution is just over the horizon though.

brendanzagaeski: The known issue can go alongside the draft release note for this PR in a `Documentation/release-notes/4722.md` file as part of this PR, so something like:

```
### IDE compatibility

- [Developer Community 1038779](https://developercommunity.visualstudio.com/content/problem/1038779/visual-studio-crash-down.html):
  ...

### Known issues

- [GitHub nnnn]():
  ... prevents building successfully if project paths contain accented
  characters or other non-ASCII UTF8 characters for projects configured to use
  AAPT2.
```

dellis1972:
More info on the error

```
Xamarin.VisualStudio.UnhandledExceptionsManager|Error|0|An unhandled error occurred. Details: The handle is invalid.


System.IO.IOException: The handle is invalid.

   at System.IO.__Error.WinIOError()
   at System.Console.set_InputEncoding(Encoding value)
   at Xamarin.Android.Tasks.Aapt2Daemon.Aapt2DaemonStart()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state, Boolean preserveSyncCtx)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```

The error is `handle is invalid` 🤷 I have no idea why the `Console` handle would be invalid.

src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
dellis1972: @brendanzagaeski I can't get the actual error they might see as we have a bug stopping me from building in a directory with accented characters

src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
```diff
@@ -162,21 +165,34 @@ private void Aapt2DaemonStart ()
...
+						aapt2.StartInfo = info;
+						aapt2.Start ();
+					} finally {
+						Console.InputEncoding = current;
+					}
+				} catch (Exception ex) {
+					daemonStartupWarnings.Enqueue (ex.ToString ());
 				}
 			}
+			if (aapt2 == null)
```
jonpryor, line 194: How can this happen?  Line 184 will re-assign `aapt2` to a non-null value, unless the `Process` constructor throws.

If `aapt2` *could* be null, that sounds like a possible "semantic" violation, as other parts of the code will assume/require that `aapt2` be running, but if it *isn't* running, what'll happen?

src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
```diff
@@ -162,21 +165,34 @@ private void Aapt2DaemonStart ()
...
+							daemonStartupWarnings.Enqueue (ioEx.ToString ());
+						}
+						aapt2 = new Process ();
+						aapt2.StartInfo = info;
+						aapt2.Start ();
+					} finally {
+						Console.InputEncoding = current;
+					}
+				} catch (Exception ex) {
+					daemonStartupWarnings.Enqueue (ex.ToString ());
```
jonathanpeppers, line 191: If we can't start the daemon at all here, should this fail instead?

It seems like the warning would be OK for `IOException` but not in the case where it couldn't start the daemon at all?

jonpryor:
In retrospect, I *really* don't like the idea of setting `Console.InputEncoding` in the first place; it's a global solution for a local problem, the local problem being that [`ProcessStartInfo.StandardInputEncoding`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.processstartinfo.standardinputencoding?view=netcore-3.1) not being part of .NET Standard 2.0.

We should either use .NET Standard 2.1 (PR #4735, which apparently is failing), or use `System.Reflection` to lookup and set `ProcessStartInfo.StandardInputEncoding`, which is a local solution for a local problem.

Issue #4664, opened 2020-05-02
[Bug] ResourceManager.GetString always returning values for the default culture


marcellangmaier:
### Description
When having a Resource file (Resource1.resx) with default settings (ResXFileCodeGenerator, Embedded Resource) in my project with one translation (Resource1.en.resx) the ResourceManager is always returning the value from Resource1.resx even if queried with CultureInfo("en") Parameter.

Testing the same thing in a Windows Forms application works as expected.
### Steps to Reproduce

1.  Create default Shell Application
2.  Add Toolbar Item to ItemsPage.xaml with Clicked Handler
3.  Add resource file "Resource1.resx" to Common Project, Create Key "Key1" with value "Value1 neutral"
4. Add resource file "Resource1.en.resx" to Common Project. Create Key "Key1" with value "Value1 english"
5. Add the following code in Clicked Handler:

            var en = System.Globalization.CultureInfo.GetCultureInfo("en");

            var s = Resource1.ResourceManager.GetString("Key1");
            await DisplayAlert("Value (no culture)", s, "OK");

            s = Resource1.ResourceManager.GetString("Key1", en);
            await DisplayAlert("Value (en)", s, "OK");

### Expected Behavior
Culture of my device: de
First alert box should show the value: "Value1 neutral"
Second alert box should show the value: "Value1 english"

### Actual Behavior

Both boxes are showing the neutral value

### Basic Information

- Version with issue:
- Last known good version:
- IDE: Version 16.6.0 Preview 5.0
- Platform Target Frameworks: Android: 9.0
- Affected Devices: Samsung Galaxy S9, Adnroid Emulator

Sample Project
[resxtest.zip](https://github.com/xamarin/Xamarin.Forms/files/4567444/resxtest.zip)

thierryiseli:
I have the same issue. With iOS it works without problems.

thierryiseli:
I do not know exactly what the mistake was, but now it basically works. The only thing I did was to remove the assembly specification in XAML and update Xamarin.Forms to the latest version.

`xmlns:resources="clr-namespace:MyApp.Localization;assembly=MyApp"`
to
`xmlns:resources="clr-namespace:MyApp.Localization"`

Furthermore I have moved all resx-files into an own namespace.

DevEddy:
Same problem here for Android...

brendanzagaeski:
### Workaround

Open the _.csproj_ file for any affected app project, and replace the end of the file with the following lines:

```xml
[...]

PR #4724, opened 2020-05-26
[Xamarin.Android.Build.Tasks] fix .resx files for IDE builds

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/4664
Fixes: https://feedback.devdiv.io/1030901

It turns out that building *inside* Visual Studio is causing `.resx`
files to stop working. Satellite assemblies are missing from the APK
and/or were not deployed with Fast Deployment.

In 65a917a9, I removed a direct call to `<ResolveAssemblyReference/>`
for performance reasons. This appears to have caused this problem.

This is due to:

https://github.com/microsoft/msbuild/blob/dc485bce3427e9d2b020ce61c2400e7b5a76062c/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2101
https://github.com/microsoft/msbuild/blob/dc485bce3427e9d2b020ce61c2400e7b5a76062c/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2059

`<ResolveAssemblyReference/>` by default only finds satellite
assemblies when `$(BuildingProject)` is set. This does *not* happen
when `$(BuildingInsideVisualStudio)` is `true` during design-time
builds.

So inside the IDE:

* `Build` runs, `$(BuildingProject)` is `true` and everything works.
* `Install` runs, `$(BuildingProject)` is `false`. This is the part
  where everything falls apart.

I could reproduce this in a test by running `Build` then
`SignAndroidPackage` in two steps. Our MSBuild tests are already set
`$(BuildingInsideVisualStudio)` to `true` by default.

The fix for this is to make `SignAndroidPackage` depend on
`BuildOnlySettings`:

https://github.com/microsoft/msbuild/blob/dc485bce3427e9d2b020ce61c2400e7b5a76062c/src/Tasks/Microsoft.Common.CurrentVersion.targets#L1081-L1086

This makes `$(BuildingProject)` always `true` if `SignAndroidPackage`
or `Install` runs in the IDE.

### DIFF
--- a/Documentation/release-notes/4724.md
+++ b/Documentation/release-notes/4724.md
@@ -1,0 +1,8 @@
+#### Application behavior on device and emulator
+
+  - [Developer Community 1030901](https://developercommunity.visualstudio.com/content/problem/1030901/localization-in-android-doesnt-work-4.html),
+    [GitHub 4664](https://github.com/xamarin/xamarin-android/issues/4664):
+    Starting in Xamarin.Android 10.3, localized resources from _.resx_ files were
+    no longer deployed when building and deploying from within Visual Studio to
+    an attached device or emulator.  (In contrast, clean builds started via the
+    **Build > Archive** command or on the command line worked as expected.)
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
@@ -725,5 +725,42 @@ namespace App1
 				}
 			}
 		}
+
+		[Test]
+		public void MissingSatelliteAssembly ()
+		{
+			var path = Path.Combine ("temp", TestName);
+			var lib = new XamarinAndroidLibraryProject {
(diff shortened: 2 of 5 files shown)


jonathanpeppers:
I manually tested this, and it seems to solve the problem.

An alternate workaround for https://github.com/xamarin/xamarin-android/issues/4664 might be to update `$(SignAndroidPackageDependsOn)` in a `.csproj` file after `Xamarin.Android.CSharp.targets` is imported:

```xml
<Import Project="$(MSBuildExtensionsPath)\Xamarin\Android\Xamarin.Android.CSharp.targets" />
<PropertyGroup>
  <SignAndroidPackageDependsOn Condition=" '$(BuildingInsideVisualStudio)' == 'True' ">
    BuildOnlySettings;
    $(SignAndroidPackageDependsOn);
  </SignAndroidPackageDependsOn>
</PropertyGroup>
```

dellis1972:
no one. change anything, its green :P

jonpryor:
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1124466
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1130016

PR #4736, opened 2020-05-27
[tests] fix failing NuGet restore on TimeZone tests

jonathanpeppers:
TimeZone tests fail to build with `/restore`:

    NuGet.targets(124,5): The local source '/Users/runner/runners/2.168.2/work/1/s/bin/BuildRelease/nupkgs' doesn't exist.

In 10d814f2, I added support for running our MSBuild test suite under
.NET 5. This inadvertently added the local source to the
`NuGet.config` for the TimeZone tests.

To fix this, I made the change so the `NuGet.config` will only contain
the extra sources if:

* `XASdkProject` is used.
* `Builder.UseDotNet` is set via `nunit3-console --params dotnet=true`.


jonathanpeppers:
I triggered: https://build.azdo.io/3760996

I had to set `$(XA.RunAllTests)` to get the timezone tests to run.

PR #4741, opened 2020-05-29
[build] fixes for .nupkg's on Linux

jonathanpeppers:
`Mono.Terminal` has a timestamp:

    $ ls -l ./Debug/lib/xamarin.android/xbuild/Xamarin/Android/Mono.Terminal.dll
    -rwxr--r-- 1 grendel grendel 20480 Jan  1  1980 ./Debug/lib/xamarin.android/xbuild/Xamarin/Android/Mono.Terminal.dll

That causes the failure:

    "build-tools/create-packs/Microsoft.Android.Sdk.proj" (CreateAllPacks target) (1) ->
        /usr/share/dotnet/sdk/3.1.300/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(198,5): error : The DateTimeOffset specified cannot be converted into a Zip file timestamp. (Parameter 'value')

`Mono.Terminal` is only used by `logcat-parse`, so we should exclude
this file for .NET 5 packages. `logcat-parse` is already excluded.

There is also some incorrect casing that causes:

    (CreateAllPacks target) ->
        build-tools/create-packs/Microsoft.Android.Sdk.proj(57,5): error MSB3030: Could not copy the file "bin/Debug/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Tools.JavaDocImporter.pdb" because it was not found.
        build-tools/create-packs/Microsoft.Android.Sdk.proj(57,5): error MSB3030: Could not copy the file "bin/Debug/lib/xamarin.android/xbuild/Xamarin/Android/Xamarin.Android.Tools.JavaDocImporter.dll" because it was not found.

These should be `JavadocImporter` instead of `JavaDocImporter`.

I also fixed two issues in `Microsoft.Android.Sdk.proj` that was
preventing native libraries from being added to the `.nupkg` files on
Linux.

I added CI steps for Linux to create the `.nupkg` files and publish
them as build artifacts.

### DIFF
--- a/build-tools/automation/azure-pipelines-oss.yaml
+++ b/build-tools/automation/azure-pipelines-oss.yaml
@@ -166,6 +166,13 @@ stages:
     - script: make package-deb V=1 CONFIGURATION=$(XA.Build.Configuration)
       displayName: make package-deb
 
+    - task: MSBuild@1
+      displayName: pack all nupkgs
+      inputs:
+        solution: $(System.DefaultWorkingDirectory)/build-tools/create-packs/Microsoft.Android.Sdk.proj
+        configuration: $(XA.Build.Configuration)
+        msbuildArguments: /t:CreateAllPacks /restore /bl:$(System.DefaultWorkingDirectory)/bin/Build$(XA.Build.Configuration)/create-all-packs.binlog
+
     - script: >
         mkdir -p $(System.DefaultWorkingDirectory)/bin/Build$(XA.Build.Configuration)/linux-artifacts &&
         cp $(System.DefaultWorkingDirectory)/*xamarin.android*.tar.bz2 $(System.DefaultWorkingDirectory)/bin/Build$(XA.Build.Configuration)/linux-artifacts &&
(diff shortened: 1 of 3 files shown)


build-tools/automation/azure-pipelines-oss.yaml
```diff
@@ -180,6 +187,12 @@ stages:
         artifactName: Linux Packages
         targetPath: $(System.DefaultWorkingDirectory)/bin/Build$(XA.Build.Configuration)/linux-artifacts
 
+    - task: PublishPipelineArtifact@1
+      displayName: publish nupkgs
+      inputs:
+        artifactName: nupkgs
```
pjcollins, line 193: I think we'll need a unique name here to distinguish between mac and linux produced nupkg artifacts, as the OSS macOS build is using the same name:
https://github.com/xamarin/xamarin-android/blob/c94e1ab5c87a188948a77bbd1eca70e6b672fa3c/build-tools/automation/azure-pipelines-oss.yaml#L111

jonathanpeppers: I put the word `Linux` in front of it now.

PR #4742, opened 2020-05-29
[CI] Delete VS installer before provisioning VS

pjcollins:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3765530&view=logs&jobId=1cfdb47b-00d6-51ee-9189-006d3307b76f

Our Windows regression test jobs have been failing for a number of weeks
due to a timeout during provisioning.  After getting access to one of
these machines, I noticed that 'VSIXInstaller.exe' processes were being
kept open even after invocations of the tool completed succesfully.
I was able to avoid this issue by deleting the Visual Studio `Installer`
directory that was left behind, and running a stable VS Installer to
replace the old installer with a stable version.  It seems that an early
16.7 preview version of VSIXInstaller.exe was problematic, and the
latest public preview and stable installers do not have this issue.

### DIFF
--- a/build-tools/automation/yaml-templates/run-integrated-regression-tests.yaml
+++ b/build-tools/automation/yaml-templates/run-integrated-regression-tests.yaml
@@ -40,6 +40,11 @@ steps:
   clean: true
   fetchDepth: 10
 
+  # Clear out VS installer content before provisioning VS to avoid using unstable tools from pre-release installers.
+- powershell: Remove-Item "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer" -Recurse -Force -ErrorAction Ignore
+  displayName: Delete VS Installer
+  condition: and(succeeded(), eq(variables['agent.os'], 'Windows_NT'))
+
 - task: xamops.azdevex.provisionator-task.provisionator@2
   displayName: Provision Android Dependencies
   inputs:


pjcollins:
Full pipeline build with all test stages has been triggered here - https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3767597&view=results

build-tools/automation/yaml-templates/run-integrated-regression-tests.yaml
```diff
@@ -40,6 +40,11 @@ steps:
   clean: true
   fetchDepth: 10
 
+  # Clear out VS installer content before provisioning VS to avoid using unstable tools from pre-release installers.
+- powershell: Remove-Item "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer" -Recurse -Force -ErrorAction Ignore
+  displayName: Delete VS Installer
+  condition: and(succeeded(), eq(variables['agent.os'], 'Windows_NT'))
```
jonathanpeppers, line 46: What puts `C:\Program Files (x86)\Microsoft Visual Studio\Installer` back in place? provisionator?

Will this break things like the Azure DevOps `MSBuild` task? It uses `vswhere.exe` in this directory.

pjcollins: The provisionator call which invokes the .csx linked below will regenerate this folder as part of VS provisioning:
https://github.com/xamarin/QualityAssurance/blob/master/Scripts/Provisionator/xamarin-android.csx#L20

pjcollins:
Looking at https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=3767948&view=results it looks like this did address the provisioning hang on the Windows machines. There are still a handful of failures on these jobs but I think they should all be resolved by https://github.com/xamarin/QualityAssurance/commit/f9061c501cccc56dc24506db19b2dc63f0a9cc53.

PR #4743, opened 2020-05-29
[tests] ignore incremental Hybrid AOT test

jonathanpeppers:
This test is failing with:

    1) Failed : Xamarin.Android.Build.Tests.IncrementalBuildTest.BuildIncrementalAot("arm64-v8a","Hybrid",False,True)
    `_BuildApkEmbed` should be skipped on second build!
    Expected: True
    But was:  False
    at Xamarin.Android.Build.Tests.IncrementalBuildTest.BuildIncrementalAot

This is triggered due to the `<CilStrip/>` MSBuild task:

https://github.com/xamarin/xamarin-android/blob/0022554468a0d0ded26c7fd119a555cddd5525ff/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L2253-L2259

It modifies the assemblies in-place. This causes the generated native
assembly code to change, and so you see things like:

    Input file "obj/Release/android/typemaps.armeabi-v7a.o" is newer than output file "obj/Release/app_shared_libraries/armeabi-v7a/libxamarin-app.so".

So then the newer `libxamarin-app.so` file causes the `_BuildApkEmbed`
MSBuild target to run.

It is still unknown with One .NET, what will happen with `Hybrid` AOT
and `<CilStrip/>`. For now, I think we should just ignore this test.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
@@ -929,6 +929,11 @@ namespace Lib2
 					Assert.IsFalse (b.Output.IsTargetSkipped (target), $"`{target}` should *not* be skipped on first build!");
 				}
 
+				if (androidAotMode == "Hybrid") {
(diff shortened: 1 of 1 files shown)


dellis1972:
This is ok in the short term. We should look at moving the AOT stuff into new targets file and pull them out of the BuildApk target.

PR #4750, opened 2020-06-02
[Xamarin.Android.Build.Tasks] fix IsSdkAssembly() checks in LinkAssembliesNoShrink

jonathanpeppers:
I discovered the `Profile.IsSdkAssembly()` and
`Profile.IsProductAssembly()` checks only work against assembly names:

https://github.com/xamarin/xamarin-android/blob/a134aa72df62d7d1bfb4da52b5c642dbfe4d0490/src/Xamarin.Android.Build.Tasks/Linker/Mobile.Tuner/MobileProfile.cs#L12-L197

So we were actually checking for obsolete `AppDomain` usage in the BCL,
`Mono.Android.dll`, `Java.Interop.dll`, etc.

I fixed these checks by using `Path.GetFileNameWithoutExtension()`.
Now these assemblies shouldn't even be loaded with Mono.Cecil at all.

## Results ##

    Before:
    2596 ms  LinkAssembliesNoShrink                     1 calls
    After:
    1849 ms  LinkAssembliesNoShrink                     1 calls

Saved ~750ms.

This was running an initial build of the Xamarin.Forms integration
project on macOS. There will likely be not much of an improvement for
incremental builds, as the MSBuild target runs partially.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
@@ -57,7 +57,8 @@ namespace Xamarin.Android.Tasks
 					var destination = DestinationFiles [i];
 					AssemblyDefinition assemblyDefinition = null;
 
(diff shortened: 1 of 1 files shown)


dellis1972:
I'm going to merge this. Both test failures seem to be Nuget related.

PR #4757, opened 2020-06-03
Stop using `__Internal` in our p/invoke calls

grendello:
Context: https://github.com/xamarin/xamarin-android/pull/4914
Context: https://github.com/xamarin/xamarin-android/commit/d583b7c215fe9be2f732b4a52423be0aea7f0809

The .NET5 version of the Mono runtime "hijacked" `__Internal` for its
own purposes and the change causes us trouble when trying to resolve
p/invokes from our managed code to our runtime.  Instead we now use
`xa-internal-api` (added in d583b7c2) to specify the p/invoke library
name.

### DIFF
--- a/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
+++ b/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
@@ -280,7 +280,7 @@ namespace Android.Runtime {
 			}
 		}
 
-		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
+		[DllImport (AndroidRuntime.InternalDllName, CallingConvention = CallingConvention.Cdecl)]
 		static extern IntPtr _monodroid_timezone_get_default_id ();
 
 		// This is invoked by
@@ -301,7 +301,7 @@ namespace Android.Runtime {
 		// These are invoked by
 		// System.dll!System.AndroidPlatform.getifaddrs
 		// DO NOT REMOVE
-		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
+		[DllImport (AndroidRuntime.InternalDllName, CallingConvention = CallingConvention.Cdecl)]
 		static extern int _monodroid_getifaddrs (out IntPtr ifap);
 
 		static int GetInterfaceAddresses (out IntPtr ifap)
@@ -312,7 +312,7 @@ namespace Android.Runtime {
 		// These are invoked by
 		// System.dll!System.AndroidPlatform.freeifaddrs
 		// DO NOT REMOVE
-		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
+		[DllImport (AndroidRuntime.InternalDllName, CallingConvention = CallingConvention.Cdecl)]
                 static extern void _monodroid_freeifaddrs (IntPtr ifap);
 
 		static void FreeInterfaceAddresses (IntPtr ifap)
@@ -320,7 +320,7 @@ namespace Android.Runtime {
 			_monodroid_freeifaddrs (ifap);
 		}
 
-		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
+		[DllImport (AndroidRuntime.InternalDllName, CallingConvention = CallingConvention.Cdecl)]
 		static extern void _monodroid_detect_cpu_and_architecture (ref ushort built_for_cpu, ref ushort running_on_cpu, ref byte is64bit);
 
 		static void DetectCPUAndArchitecture (out ushort builtForCPU, out ushort runningOnCPU, out bool is64bit)
--- a/src/monodroid/jni/monodroid-glue.cc
+++ b/src/monodroid/jni/monodroid-glue.cc
@@ -1151,27 +1151,92 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
 {
 	unsigned int dl_flags = monodroidRuntime.convert_dl_flags (flags);
 	bool libmonodroid_fallback = false;
-
+	bool name_is_full_path = false;
+	bool name_needs_free = false;
 	/* name is nullptr when we're P/Invoking __Internal, so remap to libxa-internal-api */
-	if (name == nullptr) {
+	if (name == nullptr || strstr (name, "xa-internal-api") != nullptr) {
+#if defined (WINDOWS)
+		char *tmp_name = nullptr;
+
+		auto probe_dll_at = [&](const char *the_path) -> bool {
+			if (the_path == nullptr) {
+				return false;
+			}
+
+			const char *last_sep = strrchr (the_path, MONODROID_PATH_SEPARATOR_CHAR);
+			if (last_sep != nullptr) {
+				char *dir = utils.strdup_new (the_path, last_sep - the_path);
+				tmp_name = utils.string_concat (dir, MONODROID_PATH_SEPARATOR, API_DSO_NAME);
+				delete[] dir;
+				if (!utils.file_exists (tmp_name)) {
+					delete[] tmp_name;
+					tmp_name = nullptr;
+					return false;
+				}
+
+				return true;
+			}
+
+			return false;
+		};
+
+		//
+		// First try to see if it exist at the path pointed to by `name`. With p/invokes, currently (Sep 2020), we can't
+		// really trust the path since it consists of *some* directory path + p/invoke library name and it does not
+		// point to the location where the native library is. However, we still need to check the location first, should
+		// it point to the right place in the future.
+		//
+		// Context: https://github.com/mono/mono/issues/20295#issuecomment-679271621
+		//
+		bool found = probe_dll_at (name);
+		if (!found) {
+			// Next lets try the location of the XA runtime DLL, libxa-internal-api.dll should be next to it.
+			const char *path = get_my_location (false);
+			found = probe_dll_at (path);
+			if (path != nullptr) {
+				free (reinterpret_cast<void*>(const_cast<char*>(path)));
+			}
+
+			if (!found) {
+				log_warn (LOG_DEFAULT, "Failed to locate %s, using file name without the path", API_DSO_NAME);
+				name = API_DSO_NAME;
+			} else {
+				name = tmp_name;
+				name_is_full_path = true;
+				name_needs_free = true;
+			}
+		}
+#else // ndef WINDOWS
 		name = API_DSO_NAME;
+#endif // WINDOWS
 		libmonodroid_fallback = true;
 	}
 
-	void *h = androidSystem.load_dso_from_any_directories (name, dl_flags);
+	void *h = nullptr;
+	if (!name_is_full_path)
+		h = androidSystem.load_dso_from_any_directories (name, dl_flags);
 
 	if (h != nullptr) {
-		return monodroid_dlopen_log_and_return (h, err, name, false, libmonodroid_fallback);
+		return monodroid_dlopen_log_and_return (h, err, name, name_needs_free, libmonodroid_fallback);
 	}
 
 	if (libmonodroid_fallback) {
-		char *full_name = utils.path_combine (AndroidSystem::SYSTEM_LIB_PATH, API_DSO_NAME);
+		const char *full_name;
+		if (name_is_full_path) {
+			full_name = name;
+		} else {
+			if (name_needs_free) {
+				delete[] name;
+			}
+			full_name = utils.path_combine (AndroidSystem::SYSTEM_LIB_PATH, API_DSO_NAME);
+			name_needs_free = true;
+		}
 		h = androidSystem.load_dso (full_name, dl_flags, false);
-		return monodroid_dlopen_log_and_return (h, err, full_name, true, true);
+		return monodroid_dlopen_log_and_return (h, err, full_name, name_needs_free, true);
 	}
 
 	if (!utils.ends_with (name, ".dll.so") && !utils.ends_with (name, ".exe.so")) {
-		return monodroid_dlopen_log_and_return (h, err, name, false);
+		return monodroid_dlopen_log_and_return (h, err, name, name_needs_free);
 	}
 
 	char *basename_part = const_cast<char*> (strrchr (name, '/'));
@@ -1185,6 +1250,10 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
 
 	if (h != nullptr && XA_UNLIKELY (utils.should_log (LOG_ASSEMBLY)))
 		log_info_nocheck (LOG_ASSEMBLY, "Loaded AOT image '%s'", static_cast<const char*>(basename));
+
+	if (name_needs_free) {
+		delete[] name;
+	}
 
 	return h;
 }
@@ -1489,7 +1558,7 @@ MonodroidRuntime::typemap_managed_to_java (MonoReflectionType *type, const uint8
 
 #if defined (WINDOWS)
 const char*
-MonodroidRuntime::get_my_location ()
+MonodroidRuntime::get_my_location (bool remove_file_name)
 {
 	HMODULE hm = NULL;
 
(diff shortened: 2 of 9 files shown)


garuma:
Our code will not directly address those p/invoke so the change should be fine for us. Unless there is a specific concern I'm not seeing?

grendello:
@garuma that's what I thought, but seeing so many Designer tests fail I thought the changes here broke them somehow, thanks!

jonpryor:
@garuma: any idea then why custom control use would be disabled? https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3782292&view=ms.vss-test-web.build-test-results-tab&runId=13618984&resultId=100691&paneView=debug

garuma:
Looks like it's not resolving `monodroid` when run on desktop:

```
[2020-06-04 19:19:38.4] Renderer >> 4 [monodroid] Calling into managed runtime init
[2020-06-04 19:19:38.6] Renderer (error) >> 
[2020-06-04 19:19:38.6] Renderer (error) >> Unhandled Exception:
[2020-06-04 19:19:38.6] Renderer (error) >> System.DllNotFoundException: monodroid assembly:<unknown assembly> type:<unknown type> member:(null)
[2020-06-04 19:19:38.6] Renderer (error) >> 
[2020-06-04 19:19:38.6] Renderer (error) >>   at (wrapper managed-to-native) Android.Runtime.AndroidObjectReferenceManager._monodroid_gref_get()
[2020-06-04 19:19:38.6] Renderer (error) >> 
[2020-06-04 19:19:38.6] Renderer (error) >>   at Android.Runtime.AndroidObjectReferenceManager.get_GlobalReferenceCount () [0x00000] in /Users/builder/azdo/_work/2/s/xamarin-android/src/Mono.Android/Android.Runtime/AndroidRuntime.cs:107 
[2020-06-04 19:19:38.6] Renderer (error) >> 
[2020-06-04 19:19:38.6] Renderer (error) >>   at Java.Interop.JniRuntime+JniObjectReferenceManager.CreateGlobalReference (Java.Interop.JniObjectReference reference) [0x00011] in <b3fb8b2fd70944739118ed66394c3cef>:0 
[2020-06-04 19:19:38.6] Renderer (error) >> 
[2020-06-04 19:19:38.6] Renderer (error) >>   at Android.Runtime.AndroidObjectReferenceManager.CreateGlobalReference (Java.Interop.JniObjectReference value) [0x00000] in /Users/builder/azdo/_work/2/s/xamarin-android/src/Mono.Android/Android.Runtime/AndroidRuntime.cs:169 
[2020-06-04 19:19:38.6] Renderer (error) >> 
[2020-06-04 19:19:38.6] Renderer (error) >>   at Java.Interop.JniObjectReference.NewGlobalRef () [0x0000a] in <b3fb8b2fd70944739118ed66394c3cef>:0 
[2020-06-04 19:19:38.6] Renderer (error) >> 
[2020-06-04 19:19:38.6] Renderer (error) >>   at Java.Interop.JniRuntime..ctor (Java.Interop.JniRuntime+CreationOptions options) [0x001c4] in <b3fb8b2fd70944739118ed66394c3cef>:0 
[2020-06-04 19:19:38.6] Renderer (error) >> 
[2020-06-04 19:19:38.6] Renderer (error) >>   at Android.Runtime.AndroidRuntime..ctor (System.IntPtr jnienv, System.IntPtr vm, System.Boolean allocNewObjectSupported, System.IntPtr classLoader, System.IntPtr classLoader_loadClass, System.Boolean jniAddNativeMethodRegistrationAttributePresent) [0x00000] in /Users/builder/azdo/_work/2/s/xamarin-android/src/Mono.Android/Android.Runtime/AndroidRuntime.cs:25 
[2020-06-04 19:19:38.6] Renderer (error) >> 
[2020-06-04 19:19:38.6] Renderer (error) >>   at Android.Runtime.JNIEnv.Initialize (Android.Runtime.JnienvInitializeArgs* args) [0x000f8] in /Users/builder/azdo/_work/2/s/xamarin-android/src/Mono.Android/Android.Runtime/JNIEnv.cs:177 
```

grendello:
> Looks like it's not resolving `monodroid` when run on desktop:
> 

That's because because the XA runtime is named `libmono-android.{debug,release}.{dll,dylib,so}` on desktop... ugh

jonpryor:
Can't merge until designer-side changes are in place.

TODO: link to designer-side changes?

src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
```diff
@@ -283,7 +283,7 @@ static string GetDefaultTimeZone ()
 			}
 		}
 
-		[DllImport ("__Internal", CallingConvention = CallingConvention.Cdecl)]
+		[DllImport ("monodroid", CallingConvention = CallingConvention.Cdecl)]
```
dellis1972, line 286: since we are hitting all these locations, can we change these to use a const? so we can update it if need be in one place.

grendello: Good idea. I need to update this PR to match the recent changes on `master`

jonpryor:
Something else that I've discovered as part of xamarin/java.interop#691 and xamarin/xamarin-android#5031 is that we *also* need to migrate away from `__Internal` in our `config.xml` files as well!

https://github.com/xamarin/xamarin-android/blob/faf1d16692ead1115d34fdf7b84c256c94a321da/src/monodroid/config.xml#L2

src/monodroid/config.xml
```diff
@@ -1,5 +1,9 @@
 <configuration>
-	<dllmap dll="java-interop" target="__Internal" />
+	<dllmap dll="java-interop">
+		<dllentry os="linux"   target="xa-internal-api" />
```
dellis1972, line 3: so we know the dllmap file support is going away for .net 5/6. is it worth updating this to fix that issue now or do it in a separate PR?

grendello: I think you're right, especially that the change appears to break stuff for some reason (the library isn't found). We need to migrate to programmatic dllmap setup anyway, that is probably best left for a separate PR

jonpryor:
Windows designer integration tests are [failing](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4045461&view=logs&j=bfe1c8f0-0562-5169-4501-8cf84831035d&t=79783496-89f6-5d59-544a-94fd06d3acc1):

```
[2020-09-07 13:18:07.1] Renderer >> 5 [monodroid] MonodroidRuntime::monodroid_dlopen ("C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\ReferenceAssemblies\Microsoft\Framework\MonoAndroid\v9.0\xa-internal-api", 0, 000000001dc4c1a0, 0000000000000000)
[2020-09-07 13:18:07.1] Renderer >> 7 [monodroid] Internal API library is required
[2020-09-07 13:18:07.2] ERROR: : ReadMessage failed
System.IO.IOException: Unable to read data from the transport connection: An existing connection was forcibly closed by the remote host. ---> System.Net.Sockets.SocketException: An existing connection was forcibly closed by the remote host
   at System.Net.Sockets.NetworkStream.Read(Byte[] buffer, Int32 offset, Int32 size)
   --- End of inner exception stack trace ---
   at System.Net.Sockets.NetworkStream.Read(Byte[] buffer, Int32 offset, Int32 size)
   at Xamarin.AndroidDesigner.Java.JavaCommunicationContext.ReadMessageLoop() in D:\a\1\s\designer\Xamarin.Designer.Android\Xamarin.AndroidDesigner\Java\JavaCommunicationContext.cs:line 237
[2020-09-07 13:18:07.3] ERROR: : Layout inflation failed
…
```

The `Internal API library is required` message is from `MonodroidRuntime::init_internal_api_dso()`: https://github.com/xamarin/xamarin-android/blob/030442e6aeaf51f42309a2433ef05f06a8b7fd8d/src/monodroid/jni/monodroid-glue.cc#L1086

Thus, the problem is that `libxa-internal-api.dll` isn't being loaded on Windows. :-(

dellis1972 (approved):

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1151,27 +1151,91 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
 {
 	unsigned int dl_flags = monodroidRuntime.convert_dl_flags (flags);
 	bool libmonodroid_fallback = false;
-
+	bool name_is_full_path = false;
+	bool name_needs_free = false;
 	/* name is nullptr when we're P/Invoking __Internal, so remap to libxa-internal-api */
-	if (name == nullptr) {
+	if (name == nullptr || strstr (name, "xa-internal-api") != nullptr) {
```
jonpryor, line 1157: Should this instead be `strstr (name, MONODROID_PATH_SEPARATOR "xa-internal-api") != nullptr`, ensuring that we get e.g. `/xa-internal-api`?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1151,27 +1151,91 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
...
+#if defined (WINDOWS)
+		char *tmp_name = nullptr;
+
+		auto probe_dll_at = [&](const char *the_path) -> bool {
+			const char *last_sep = strrchr (the_path, MONODROID_PATH_SEPARATOR_CHAR);
+			if (last_sep != nullptr) {
+				tmp_name = utils.strdup_new (the_path, last_sep - the_path);
+				tmp_name = utils.string_concat (tmp_name, MONODROID_PATH_SEPARATOR, API_DSO_NAME);
```
jonpryor, line 1165: This is a memory leak.  The preceding line `utils.strdup_new()` does a `new char[…]`, and `utils.string_concat()` *also* returns a `new char[…]` *without* a `delete[] s1`.  Consequently, this sequence of statements "loses" the first `new char[…]` held by `tmp_name`, causing it to leak memory.

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1151,27 +1151,91 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
...
+			if (!found) {
+				log_warn (LOG_DEFAULT, "Failed to locate %s, using file name without the path", API_DSO_NAME);
+				name = API_DSO_NAME;
+			} else {
+				name = tmp_name;
+				name_is_full_path = true;
+				name_needs_free = true;
+			}
+		}
+#else
```
jonpryor, line 1208: Should be `#else // ndef WINDOWS`

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1151,27 +1151,91 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
...
+		// First try to see if it exist at the path pointed to by `name`. With p/invokes, currently (Sep 2020), we can't
+		// really trust the path since it consists of *some* directory path + p/invoke library name and it does not
+		// point to the location where the native library is. However, we still need to check the location first, should
+		// it point to the right place in the future.
+		bool found = probe_dll_at (name);
+		if (!found) {
+			// Next lets try the location of the XA runtime DLL, libxa-internal-api.dll should be next to it.
+			HMODULE hm;
+			const void *func = reinterpret_cast<const void*>(&Java_mono_android_Runtime_initInternal);
+			if (GetModuleHandleEx (GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, static_cast<LPCSTR>(func), &hm) == 0) {
```
jonpryor, line 1187: This looks like `MonodroidRuntime::get_my_location()`? Perhaps that should be used instead?

src/monodroid/jni/monodroid-glue.cc
grendello: I'd rather not do that. We can't be sure we'll always be getting a path here. Right now we do get **a** path (which points to a location where there's no libraries we ship) but it may change and I'd rather be flexible.

src/monodroid/jni/monodroid-glue.cc
grendello: heh, forgot that we had that one :)

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1151,27 +1151,88 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
...
+				return true;
+			}
+
+			return false;
+		};
+
+		// First try to see if it exist at the path pointed to by `name`. With p/invokes, currently (Sep 2020), we can't
+		// really trust the path since it consists of *some* directory path + p/invoke library name and it does not
+		// point to the location where the native library is. However, we still need to check the location first, should
+		// it point to the right place in the future.
```
jonpryor, line 1186: TODO (fix after current PR build completes): to link to: https://github.com/mono/mono/issues/20295#issuecomment-679271621

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1151,27 +1151,88 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
...
+		bool found = probe_dll_at (name);
+		if (!found) {
+			// Next lets try the location of the XA runtime DLL, libxa-internal-api.dll should be next to it.
+			const char *path = get_my_location ();
```
jonpryor, line 1190: Silly question: why does `get_my_location()` return a `const char*` when we're supposed to `free()` it? Isn't convention to return a *non*-`const` type for values which should be freed?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1151,27 +1151,88 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
...
+			found = probe_dll_at (path);
+			if (path != nullptr) {
+				free (reinterpret_cast<void*>(const_cast<char*>(path)));
```
jonpryor, line 1193: …cause this is just bizarre.

src/monodroid/jni/monodroid-glue.cc
grendello: `const char*` doesn't mean it cannot be `free`'d, it means it {should,must} not be modified. Pointers can point to r/o (by convention) data after all..

src/monodroid/jni/monodroid-glue.cc
grendello: It is, but we have a no-warnings policy, so... :)

jonpryor:
Summary:

```
[Mono.Android] Don't use `__Internal` in our p/invoke calls (#4757)
```

Content:

```markdown
Context: d583b7c215fe9be2f732b4a52423be0aea7f0809
Context: d583b7c215fe9be2f732b4a52423be0aea7f0809

Context: https://github.com/mono/mono/issues/20295#issuecomment-679271621

Context? https://github.com/mono/mono/pull/17369
Context? https://github.com/mono/mono/commit/d5b6cf33b3d2a89189be4e6bddcf0207e9887ecd

As a runtime extension, Mono long supported `[DllImport("__Internal")]`
as a synonym for [`dlopen(NULL, 0)`][0].  Unfortunately, this
[did not work on Android][1] -- Bionic would SIGSEGV when the
filename parameter was `NULL` -- and as "fallout" of this change the
`loader_func` that Xamarin.Android registers with mono via
`mono_dl_fallback_register()` would be invoked *instead of*
`dlopen(NULL)` when loading `__Internal`…but *only* on Android.

Use of this feature remains widely used in Xamarin.Android, e.g.
[`java-interop` is mapped to `__Internal`][2], avoiding the need to
build and ship a separate `libjava-interop.so` library.

Unfortunately, as part of .NET 5, MonoVM will be overhauling the
semantics of `[DllImport("__Internal")]`, preventing `loader_func`
from being invoked on Android to resolve `__Internal`.

Consequently, we need to begin moving away from use of `__Internal`.

Begin this process, and remove all P/Invokes to `__Internal` within
`Mono.Android.dll` with P/Invokes to `xa-internal-api` (d583b7c2).

[0]: https://linux.die.net/man/3/dlopen
[1]: https://github.com/mono/mono/commit/9b73cd9c33507ba27829728ea82d10d98335e1d6
[2]: https://github.com/xamarin/xamarin-android/blob/faf1d16692ead1115d34fdf7b84c256c94a321da/src/monodroid/config.xml#L2
```

PR #4763, opened 2020-06-04
[dotnet] Consolidate default item properties

pjcollins:
Context: https://github.com/xamarin/xamarin-macios/pull/8752

Reduces the amount of SDK / Target Framework Profile specific properties
which control default item inclusion behavior.  For now, we will only
have one property, $(EnableDefaultAndroidItems), which controls default
inclusion for all Android specific file types.

Default Android item inclusion behavior has been moved from a .props
to a .targets file for better consistency with xamarin-macios SDKs, and to
allow usage of the default value of `$(EnableDefaultItems)` which is set in
Microsoft.NET.Sdk.DefaultItems.targets.

$(DisableImplicitAndroidFrameworkReference) has also been removed as it
was only used by our "Lite" SDK, and we can explicitly remove the
Microsoft.Android @(FrameworkReference) from that SDK instead.

Finally, The `Xamarin.Android.DefaultOutputPaths.targets` import has been
removed as we should use the default OutputPath suffix behavior from
Microsoft.NET.Sdk on both macOS and Windows if possible.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultItems.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultItems.targets
@@ -1,5 +1,9 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <PropertyGroup>
+    <EnableDefaultAndroidItems Condition=" '$(EnableDefaultAndroidItems)' == '' ">$(EnableDefaultItems)</EnableDefaultAndroidItems>
+  </PropertyGroup>
+
   <!-- Default Resource file inclusion -->
   <!-- https://developer.android.com/guide/topics/resources/providing-resources -->
   <ItemGroup Condition=" '$(EnableDefaultAndroidItems)' == 'true' ">
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
@@ -8,7 +8,6 @@
     <MinimumSupportedJavaVersion Condition=" '$(MinimumSupportedJavaVersion)' == '' ">1.8.0</MinimumSupportedJavaVersion>
     <EnableDefaultOutputPaths Condition=" '$(EnableDefaultOutputPaths)' == '' And '$(OS)' != 'Windows_NT' ">false</EnableDefaultOutputPaths>
     <EnableDefaultOutputPaths Condition=" '$(EnableDefaultOutputPaths)' == '' ">true</EnableDefaultOutputPaths>
-    <EnableDefaultItems Condition=" '$(EnableDefaultItems)' == '' ">true</EnableDefaultItems>
     <EnableDefaultAndroidItems Condition=" '$(EnableDefaultAndroidItems)' == '' ">$(EnableDefaultItems)</EnableDefaultAndroidItems>
     <AndroidBoundExceptionType Condition=" '$(AndroidBoundExceptionType)' == '' ">System</AndroidBoundExceptionType>
     <!-- Enable nuget package conflict resolution -->
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
@@ -6,9 +6,6 @@
     <UsingAndroidNETSdk>true</UsingAndroidNETSdk>
     <LatestSupportedJavaVersion  Condition=" '$(LatestSupportedJavaVersion)' == '' ">1.8.0</LatestSupportedJavaVersion>
     <MinimumSupportedJavaVersion Condition=" '$(MinimumSupportedJavaVersion)' == '' ">1.8.0</MinimumSupportedJavaVersion>
-    <EnableDefaultOutputPaths Condition=" '$(EnableDefaultOutputPaths)' == '' And '$(OS)' != 'Windows_NT' ">false</EnableDefaultOutputPaths>
-    <EnableDefaultOutputPaths Condition=" '$(EnableDefaultOutputPaths)' == '' ">true</EnableDefaultOutputPaths>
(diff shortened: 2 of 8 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
```diff
@@ -8,9 +8,8 @@
     <MinimumSupportedJavaVersion Condition=" '$(MinimumSupportedJavaVersion)' == '' ">1.8.0</MinimumSupportedJavaVersion>
     <EnableDefaultOutputPaths Condition=" '$(EnableDefaultOutputPaths)' == '' And '$(OS)' != 'Windows_NT' ">false</EnableDefaultOutputPaths>
     <EnableDefaultOutputPaths Condition=" '$(EnableDefaultOutputPaths)' == '' ">true</EnableDefaultOutputPaths>
-    <DisableImplicitAndroidFrameworkReference Condition=" '$(DisableImplicitAndroidFrameworkReference)' == '' ">false</DisableImplicitAndroidFrameworkReference>
-    <EnableDefaultAndroidResourceItems Condition=" '$(EnableDefaultAndroidResourceItems)' == '' ">true</EnableDefaultAndroidResourceItems>
-    <EnableDefaultAndroidAssetItems Condition=" '$(EnableDefaultAndroidAssetItems)' == '' ">true</EnableDefaultAndroidAssetItems>
+    <EnableDefaultItems Condition=" '$(EnableDefaultItems)' == '' ">true</EnableDefaultItems>
```
rolfbjarne, line 11: Isn't this set automatically by `Microsoft.NET.Sdk`?

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Xamarin.Android.Sdk.Lite.props
```diff
@@ -12,6 +12,11 @@
     <UsingAndroidNETSdk />
   </PropertyGroup>
 
+  <ItemGroup>
+    <!-- Remove Microsoft.Android FrameworkReference in favor of default references below. -->
+    <FrameworkReference Remove="Microsoft.Android" />
```
jonathanpeppers, line 17: Does this need to check some property, like `$(DisableImplicitFrameworkReferences)`?

rolfbjarne: We check `DisableImplicitFrameworkReferences`: https://github.com/xamarin/xamarin-macios/blob/bf88adadafa71f899ca8310555d43407dada63d5/dotnet/targets/Xamarin.Shared.Sdk.DefaultItems.targets#L26

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
pjcollins: It should be, and we should always be imported after, so it should be fine to remove this.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Xamarin.Android.Sdk.Lite.props
pjcollins: This is just for the Lite SDK which will never support a Xamarin flavored @(FrameworkReference), as such a concept is incompatible with the legacy SDK. I don't think we can/should make this conditional.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
rolfbjarne: Actually it's set in `Microsoft.NET.Sdk.DefaultItems.targets`: https://github.com/dotnet/sdk/blob/32496b832d0e7702a0dca105160d97bf0c84c7ae/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets#L16, which means I don't think it'll be set in `Microsoft.Android.Sdk.props`.

I moved the default inclusion logic to our .targets file because of this: https://github.com/rolfbjarne/xamarin-macios/commit/84e62e117f4bdfa6768b3e50952c7017e6e1673c

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultItems.targets
```diff
@@ -1,8 +1,12 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
+  <PropertyGroup>
+    <EnableDefaultAndroidItems Condition=" '$(EnableDefaultAndroidItems)' == '' ">$(EnableDefaultItems)</EnableDefaultAndroidItems>
```
jonpryor, line 4: Question: is there a naming convention here?  It would be nice if this were consistent with iOS -- whatever that may be -- and I thus wonder if it should be the current `$(EnableDefault[Nount]Items)` or it should instead be `$(Enable[Noun]DefaultItems)`, e.g. `$(EnableAndroidDefaultItems)`.

rolfbjarne: We're using `$(EnableDefault[Platform]Items)`: https://github.com/xamarin/xamarin-macios/blob/b8dac12704c162a61d0c273746b5e29b129dfff9/dotnet/targets/Xamarin.Shared.Sdk.DefaultItems.targets#L8-L11

That also seems to be the pattern in .NET: `$(EnableDefault*Items)`: https://github.com/dotnet/sdk/blob/32496b832d0e7702a0dca105160d97bf0c84c7ae/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.targets#L17-L19

pjcollins: The xamarin-macios SDKs are following the same pattern: https://github.com/xamarin/xamarin-macios/blob/ab6b41848362b4faeb5f55eaa59d51933d3236d0/dotnet/targets/Xamarin.Shared.Sdk.DefaultItems.targets#L8-L11

PR #4767, opened 2020-06-04
[One .NET] support for multiple $(RuntimeIdentifiers)

jonathanpeppers:
To mirror what we have for `$(AndroidSupportedAbis)` in the existing
Xamarin.Android world, we need a way for .NET 6 apps to support
multiple architectures.

A `HelloAndroid.csproj` targeting all architectures might look like:

      <PropertyGroup>
        <TargetFramework>net5.0-android</TargetFramework>
        <RuntimeIdentifiers>android.21-x86;android.21-x64;android.21-arm;android.21-arm64</RuntimeIdentifiers>
        <OutputType>Exe</OutputType>
      </PropertyGroup>
    </Project>

To achieve this, we can do a nested `<MSBuild/>` call:

    <ItemGroup>
      <_RIDs Include="$(RuntimeIdentifiers)" />
    </ItemGroup>
    <MSBuild
        Projects="$(MSBuildProjectFile)"
        Targets="_ComputeFilesToPublishForRuntimeIdentifiers"
        Properties="RuntimeIdentifier=%(_RIDs.Identity)">
      <Output TaskParameter="TargetOutputs" ItemName="ResolvedFileToPublish" />
    </MSBuild>

For a new target:

    <Target Name="_ComputeFilesToPublishForRuntimeIdentifiers"
        DependsOnTargets="ResolveReferences;ComputeFilesToPublish"
        Returns="@(ResolvedFileToPublish)">
    </Target>

This is equivalent to what would happen in the legacy
`_ResolveAssemblies` target.

The linker then outputs files to:

    obj\Release\net5.0\$(RuntimeIdentifier)\linked\

One problem with this, is you receive multiple BCL assemblies for each
architecture:

    obj\Release\net5.0\android.21-arm\linked\System.Linq.dll
    obj\Release\net5.0\android.21-arm64\linked\System.Linq.dll
    obj\Release\net5.0\android.21-x86\linked\System.Linq.dll
    obj\Release\net5.0\android.21-x64\linked\System.Linq.dll

I created a `<ProcessAssemblies/>` MSBuild task that checks the mvid
of each assembly to detect duplicates. Non-duplicates will be placed
in a sub-directory within the `.apk` file, such as:

* `assemblies\HelloAndroid.dll`
* `assemblies\System.Linq.dll`
* `assemblies\arm64-v8a\System.Private.CoreLib.dll`
* `assemblies\armeabi-v7a\System.Private.CoreLib.dll`
* `assemblies\x86\System.Private.CoreLib.dll`
* `assemblies\x86_64\System.Private.CoreLib.dll`

I had to make a few build & runtime changes for these ABI
sub-directories to work.

Other notes:

* We must pass `@(IntermediateAssembly)` from the outer build to the
  nested `<MSBuild/>` calls. Otherwise, the inner calls look for
  `obj\Release\net5.0\android.21-arm\HelloAndroid.dll` instead of
  `obj\Release\net5.0\HelloAndroid.dll` that actually exists.
* `_ComputeFilesToPublishForRuntimeIdentifiers` was running some extra
  targets causing `<ResolveSdks/>` to run 5 times for 4 ABIs. I had to
  make changes in `BuildOrder.targets` to fix this.
* `<ResolveAssemblies/>` now sets `%(IntermediateLinkerOutput)` on
  each assembly item. This allows .NET 6 builds to support multiple
  linker directories.

### DIFF
--- a/Documentation/guides/messages/xa0035.md
+++ b/Documentation/guides/messages/xa0035.md
@@ -1,0 +1,37 @@
+---
+title: Xamarin.Android error XA0035
+description: XA0035 error code
+ms.date: 06/17/2020
+---
+# Xamarin.Android error XA0035
+
+## Issue
+
+Xamarin.Android was not able to determine the application's target
+[Android ABIs][abis] as specified by the `.csproj` file.
+
+[abis]: https://developer.android.com/ndk/guides/abis
+
+## Solution
+
+Open the project file [in Visual Studio][edit-project-files] or another text
+editor and make sure all of the values in the `RuntimeIdentifiers` MSBuild
+property are valid:
+
+```xml
+<PropertyGroup>
+  <RuntimeIdentifiers>android.21-arm;android.21-arm64;android.21-x86;android.21-x64</RuntimeIdentifiers>
+</PropertyGroup>
+```
+
+See the Microsoft documentation on [runtime identifiers][rids] for more
+information.
+
+[edit-project-files]: https://docs.microsoft.com/visualstudio/msbuild/visual-studio-integration-msbuild#edit-project-files-in-visual-studio
+[rids]: https://docs.microsoft.com/dotnet/core/rid-catalog
+
+## Example messages
+
+```
+error XA0035: Unable to determine the Android ABI from the value 'XYZ'. Edit the project file in a text editor and set the 'RuntimeIdentifiers' MSBuild property to contain only valid identifiers for the Android platform.
+```
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
@@ -45,18 +45,8 @@ projects, these properties are set in Xamarin.Android.Legacy.targets.
     </BuildDependsOn>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <CompileDependsOn>
-      _SetupDesignTimeBuildForCompile;
-      _AddAndroidDefines;
-      _IncludeLayoutBindingSources;
-      AddLibraryJarsToBind;
-      $(CompileDependsOn);
-    </CompileDependsOn>
-    <CoreCompileDependsOn>
-      UpdateGeneratedFiles;
-      $(CoreCompileDependsOn)
-    </CoreCompileDependsOn>
+  <!-- Targets that run unless we are running _ComputeFilesToPublishForRuntimeIdentifiers -->
+  <PropertyGroup Condition=" '$(_ComputeFilesToPublishForRuntimeIdentifiers)' != 'true' ">
     <CoreResolveReferencesDependsOn>
       _SeparateAppExtensionReferences;
       _PrepareWearApplication;
@@ -71,6 +61,20 @@ projects, these properties are set in Xamarin.Android.Legacy.targets.
       AddEmbeddedJarsAsResources;
       AddEmbeddedReferenceJarsAsResources;
     </ResolveReferencesDependsOn>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <CompileDependsOn>
+      _SetupDesignTimeBuildForCompile;
+      _AddAndroidDefines;
+      _IncludeLayoutBindingSources;
+      AddLibraryJarsToBind;
+      $(CompileDependsOn);
+    </CompileDependsOn>
+    <CoreCompileDependsOn>
+      UpdateGeneratedFiles;
+      $(CoreCompileDependsOn)
+    </CoreCompileDependsOn>
     <DeferredBuildDependsOn>
       $(DeferredBuildDependsOn);
       UpdateAndroidResources;
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Linker.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Linker.targets
@@ -10,6 +10,7 @@ This file contains the .NET 5-specific targets to customize ILLink
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <Target Name="_PrepareLinking"
+      Condition=" '$(PublishTrimmed)' == 'true' "
       AfterTargets="ComputeResolvedFilesToPublishList"
       DependsOnTargets="GetReferenceAssemblyPaths">
     <ItemGroup>
@@ -22,9 +23,7 @@ This file contains the .NET 5-specific targets to customize ILLink
     </ItemGroup>
     <PropertyGroup>
       <!-- make the output verbose to see what the linker is doing. FIXME: make dependent upon verbosity level -->
-      <_ExtraTrimmerArgs>$(_ExtraTrimmerArgs) --verbose --custom-data XATargetFrameworkDirectories="$(_XATargetFrameworkDirectories)"</_ExtraTrimmerArgs>
-
-      <IntermediateLinkDir>$(MonoAndroidIntermediateAssemblyDir)</IntermediateLinkDir>
+      <_ExtraTrimmerArgs>$(_ExtraTrimmerArgs) --verbose --deterministic --custom-data XATargetFrameworkDirectories="$(_XATargetFrameworkDirectories)"</_ExtraTrimmerArgs>
 
       <!-- we don't want to ignore stuff we can't find -->
       <_ExtraTrimmerArgs>$(_ExtraTrimmerArgs) --skip-unresolved false</_ExtraTrimmerArgs>
@@ -43,4 +42,9 @@ This file contains the .NET 5-specific targets to customize ILLink
       </_TrimmerCustomSteps>
     </ItemGroup>
   </Target>
+
+  <Target Name="_LinkAssemblies"
+      DependsOnTargets="_ResolveAssemblies;_CreatePackageWorkspace;$(_BeforeLinkAssemblies);_GenerateJniMarshalMethods;_LinkAssembliesNoShrink"
+  />
+
 </Project>
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
@@ -9,7 +9,9 @@
     <AndroidApplication Condition=" '$(AndroidApplication)' == '' And '$(OutputType)' == 'Exe' ">true</AndroidApplication>
     <AndroidApplication Condition=" '$(AndroidApplication)' == '' ">false</AndroidApplication>
     <SelfContained Condition=" '$(SelfContained)' == '' And '$(AndroidApplication)' == 'true' ">true</SelfContained>
-    <RuntimeIdentifier Condition=" '$(RuntimeIdentifier)' == '' And '$(AndroidApplication)' == 'true' ">android.21-x86</RuntimeIdentifier>
+    <!-- Prefer $(RuntimeIdentifiers) plural -->
+    <RuntimeIdentifiers Condition=" '$(RuntimeIdentifier)' == '' And '$(RuntimeIdentifiers)' == '' And '$(AndroidApplication)' == 'true' ">android.21-arm64;android.21-x86</RuntimeIdentifiers>
+    <RuntimeIdentifier  Condition=" '$(RuntimeIdentifiers)' != '' And '$(RuntimeIdentifier)' != '' " />
     <AndroidManifest Condition=" '$(AndroidManifest)' == '' And '$(AndroidApplication)' == 'true' ">Properties\AndroidManifest.xml</AndroidManifest>
     <!-- We don't ever need a `static void Main` method, so switch to Library here-->
     <OutputType Condition=" '$(OutputType)' == 'Exe' ">Library</OutputType>
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -193,6 +193,15 @@ In this message, the phrase "should not be reached" means that this error messag
     <value>Failed to get the Java SDK version. Please ensure you have Java {0} or above installed.</value>
     <comment>{0} - The Java version number</comment>
   </data>
+  <data name="XA0035" xml:space="preserve">
+    <value>Unable to determine the Android ABI from the value '{0}'. Edit the project file in a text editor and set the 'RuntimeIdentifiers' MSBuild property to contain only valid identifiers for the Android platform.</value>
+    <comment>The following are literal names and should not be translated: Android, ABI, RuntimeIdentifiers, MSBuild
+{0} - The $(RuntimeIdentifier) supplied by the user.</comment>
+  </data>
+  <data name="XA0036" xml:space="preserve">
+    <value>The 'AndroidSupportedAbis' MSBuild property is no longer supported. Edit the project file in a text editor, remove any uses of 'AndroidSupportedAbis', and use the 'RuntimeIdentifiers' MSBuild property instead.</value>
+    <comment>The following are literal names and should not be translated: AndroidSupportedAbis, MSBuild, RuntimeIdentifiers</comment>
+  </data>
   <data name="XA0100" xml:space="preserve">
     <value>EmbeddedNativeLibrary '{0}' is invalid in Android Application projects. Please use AndroidNativeLibrary instead.</value>
     <comment>The following are literal names and should not be translated: EmbeddedNativeLibrary, AndroidNativeLibrary
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -327,23 +327,24 @@ namespace Xamarin.Android.Tasks
 				sourcePath = CompressAssembly (assembly);
 
 				// Add assembly
-				AddFileToArchiveIfNewer (apk, sourcePath, GetTargetDirectory (assembly.ItemSpec) + "/"  + Path.GetFileName (assembly.ItemSpec), compressionMethod: UncompressedMethod);
+				var assemblyPath = GetAssemblyPath (assembly, frameworkAssembly: false);
+				AddFileToArchiveIfNewer (apk, sourcePath, assemblyPath + Path.GetFileName (assembly.ItemSpec), compressionMethod: UncompressedMethod);
 
 				// Try to add config if exists
 				var config = Path.ChangeExtension (assembly.ItemSpec, "dll.config");
-				AddAssemblyConfigEntry (apk, config);
+				AddAssemblyConfigEntry (apk, assemblyPath, config);
 
 				// Try to add symbols if Debug
 				if (debug) {
 					var symbols = Path.ChangeExtension (assembly.ItemSpec, "dll.mdb");
 
 					if (File.Exists (symbols))
-						AddFileToArchiveIfNewer (apk, symbols, AssembliesPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
+						AddFileToArchiveIfNewer (apk, symbols, assemblyPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
 
 					symbols = Path.ChangeExtension (assembly.ItemSpec, "pdb");
 
 					if (File.Exists (symbols))
-						AddFileToArchiveIfNewer (apk, symbols, AssembliesPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
+						AddFileToArchiveIfNewer (apk, symbols, assemblyPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
 				}
 				count++;
 				if (count == ZipArchiveEx.ZipFlushLimit) {
@@ -368,20 +369,21 @@ namespace Xamarin.Android.Tasks
 				}
 
 				sourcePath = CompressAssembly (assembly);
-				AddFileToArchiveIfNewer (apk, sourcePath, AssembliesPath + Path.GetFileName (assembly.ItemSpec), compressionMethod: UncompressedMethod);
+				var assemblyPath = GetAssemblyPath (assembly, frameworkAssembly: true);
+				AddFileToArchiveIfNewer (apk, sourcePath, assemblyPath + Path.GetFileName (assembly.ItemSpec), compressionMethod: UncompressedMethod);
 				var config = Path.ChangeExtension (assembly.ItemSpec, "dll.config");
-				AddAssemblyConfigEntry (apk, config);
+				AddAssemblyConfigEntry (apk, assemblyPath, config);
 				// Try to add symbols if Debug
 				if (debug) {
 					var symbols = Path.ChangeExtension (assembly.ItemSpec, "dll.mdb");
 
 					if (File.Exists (symbols))
-						AddFileToArchiveIfNewer (apk, symbols, AssembliesPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
+						AddFileToArchiveIfNewer (apk, symbols, assemblyPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
 
 					symbols = Path.ChangeExtension (assembly.ItemSpec, "pdb");
 
 					if (File.Exists (symbols))
-						AddFileToArchiveIfNewer (apk, symbols, AssembliesPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
+						AddFileToArchiveIfNewer (apk, symbols, assemblyPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
 				}
 				count++;
 				if (count == ZipArchiveEx.ZipFlushLimit) {
@@ -409,7 +411,8 @@ namespace Xamarin.Android.Tasks
 					return assembly.ItemSpec;
 				}
 
-				if (compressedAssembliesInfo.TryGetValue (Path.GetFileName (assembly.ItemSpec), out CompressedAssemblyInfo info) && info != null) {
+				var key = CompressedAssemblyInfo.GetDictionaryKey (assembly);
+				if (compressedAssembliesInfo.TryGetValue (key, out CompressedAssemblyInfo info) && info != null) {
 					EnsureCompressedAssemblyData (assembly.ItemSpec, info.DescriptorIndex);
 					AssemblyCompression.CompressionResult result = AssemblyCompression.Compress (compressedAssembly);
 					if (result != AssemblyCompression.CompressionResult.Success) {
@@ -429,6 +432,8 @@ namespace Xamarin.Android.Tasks
 						return assembly.ItemSpec;
 					}
 					return compressedAssembly.DestinationPath;
+				} else {
+					Log.LogDebugMessage ($"Assembly missing from {nameof (CompressedAssemblyInfo)}: {key}");
 				}
 
 				return assembly.ItemSpec;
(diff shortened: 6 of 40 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
```diff
@@ -267,4 +269,118 @@ projects. .NET 5 projects will not import this file.
...
+  </Target>
+
+  <Target Name="_LinkAssemblies"
+      DependsOnTargets="_ResolveAssemblies;_CreatePackageWorkspace;$(_BeforeLinkAssemblies);_GenerateJniMarshalMethods;_LinkAssembliesNoShrink;_LinkAssembliesShrink"
+  />
+
+  <Target Name="_LinkAssembliesShrink"
+      Condition="'$(AndroidLinkMode)' != 'None'"
+      Inputs="@(ResolvedUserAssemblies);$(_AndroidBuildPropertiesCache)"
+      Outputs="$(_AndroidLinkFlag)">
```
jonathanpeppers, line 335: I moved the linker targets here, that shouldn't be called anymore in .NET 6.

dellis1972:
@jonathanpeppers its green!

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
```diff
@@ -45,18 +45,8 @@ projects, these properties are set in Xamarin.Android.Legacy.targets.
     </BuildDependsOn>
   </PropertyGroup>
 
-  <PropertyGroup>
```
jonpryor, line 48: Why did this property group need to be moved to be after `$(CoreResolveReferencesDependsOn)` and `$(ResolveReferencesDependsOn)`?

jonathanpeppers: I needed to add a `Condition` so that `ResolveSdks` didn't run in the nested `<MSBuild/>` call:

> _ComputeFilesToPublishForRuntimeIdentifiers was running some extra
targets causing <ResolveSdks/> to run 5 times for 4 ABIs. I had to
make changes in BuildOrder.targets to fix this.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
```diff
@@ -9,7 +9,9 @@
     <AndroidApplication Condition=" '$(AndroidApplication)' == '' And '$(OutputType)' == 'Exe' ">true</AndroidApplication>
     <AndroidApplication Condition=" '$(AndroidApplication)' == '' ">false</AndroidApplication>
     <SelfContained Condition=" '$(SelfContained)' == '' And '$(AndroidApplication)' == 'true' ">true</SelfContained>
-    <RuntimeIdentifier Condition=" '$(RuntimeIdentifier)' == '' And '$(AndroidApplication)' == 'true' ">android.21-x86</RuntimeIdentifier>
+    <!-- Prefer $(RuntimeIdentifiers) plural -->
+    <RuntimeIdentifiers Condition=" '$(RuntimeIdentifier)' == '' And '$(RuntimeIdentifiers)' == '' And '$(AndroidApplication)' == 'true' ">android.21-arm;android.21-arm64</RuntimeIdentifiers>
```
jonpryor, line 13: The default `$(RuntimeIdentifiers)` are arm32 & arm64?  Why?  Especially considering that the *previous* default was x86, and the "normal" Xamarin.Android default -- via xamarin/monodroid, for commercial users -- is "all the ABIs!" `arm64-v8a;armeabi-v7a;x86;x86_64`.

Given that (1) we don't know what the "commercial integration" looks like, and (2) it would be Really Really Handy to be able to run the `.apk`s on x86(_64?) emulators, this default seems "weird" to me.

src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
```diff
@@ -0,0 +1,101 @@
...
+{
+	/// <summary>
+	/// Processes .dll files coming from @(ResolvedFileToPublish). Removes duplicate .NET assemblies by MVID.
+	/// 
+	/// Also sets some metadata:
+	/// * %(FrameworkAssembly)=True to determine if framework or user assembly
+	/// * %(HasMonoAndroidReference)=True for incremental build performance
+	/// * %(AbiDirectory) if an assembly has an architecture-specific version
+	/// </summary>
+	public class ProcessAssemblies : FilterAssemblies
```
jonpryor, line 20: Not sure I see why this is a `FilterAssemblies` subclass, other than to have access to the `HasReference()` method, which could be split out "elsewhere".

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
jonathanpeppers: This is the same default we have here: https://github.com/xamarin/xamarin-android/blob/5a5580cd3cce0559da5e9d097b03d356a6630c05/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L234

I can put them all, if we think that makes more sense.

jonathanpeppers: This now defaults to `android.21-arm64;android.21-x86`.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
jonpryor: I still don't comprehend this answer.  The `Condition`al `<PropertyGroup/>` covers `$(CoreResolveReferencesDependsOn)` and `$(ResolveReferencesDependsOn)`, which doesn't *immediately* appear to cover/include `$(CompileDependsOn)` or `$(CoreCompileDependsOn)`.

Rephrased, I fail to see how *this* wouldn't work, which would result in a smaller diff (I think?):

```xml
  <PropertyGroup>
    <CompileDependsOn>
      _SetupDesignTimeBuildForCompile;
      _AddAndroidDefines;
      _IncludeLayoutBindingSources;
      AddLibraryJarsToBind;
      $(CompileDependsOn);
    </CompileDependsOn>
    <CoreCompileDependsOn>
      UpdateGeneratedFiles;
      $(CoreCompileDependsOn)
    </CoreCompileDependsOn>
  </PropertyGroup>

  <!-- Targets that run unless we are running _ComputeFilesToPublishForRuntimeIdentifiers -->
  <PropertyGroup Condition=" '$(_ComputeFilesToPublishForRuntimeIdentifiers)' != 'true' ">
    <CoreResolveReferencesDependsOn>
      _SeparateAppExtensionReferences;
      _PrepareWearApplication;
      $(ResolveReferencesDependsOn);
      _AddAndroidCustomMetaData;
    </CoreResolveReferencesDependsOn>
    <ResolveReferencesDependsOn>
      $(CoreResolveReferencesDependsOn);
      UpdateAndroidInterfaceProxies;
      UpdateAndroidResources;
      AddBindingsToCompile;
      AddEmbeddedJarsAsResources;
      AddEmbeddedReferenceJarsAsResources;
    </ResolveReferencesDependsOn>
  </PropertyGroup>

  <PropertyGroup>
    <DeferredBuildDependsOn>
      $(DeferredBuildDependsOn);
      UpdateAndroidResources;
    </DeferredBuildDependsOn>
    <PrepareForRunDependsOn>
      $(PrepareForRunDependsOn);
      _CollectMonoAndroidOutputs;
    </PrepareForRunDependsOn>
    <CleanDependsOn>
      $(CleanDependsOn);
      _CleanMonoAndroidIntermediateDir;
      _CleanAndroidBuildPropertiesCache;
      CleanBindingsOutput;
      CleanLibraryProjectIntermediaries;
      CleanNativeLibraryIntermediaries;
    </CleanDependsOn>
    <ExportJarToXmlDependsOnTargets>
      _ResolveMonoAndroidSdks;
      _ExtractLibraryProjectImports;
      _GetLibraryImports;
      _ExportJarToXml;
    </ExportJarToXmlDependsOnTargets>
  </PropertyGroup>
```

…though I suppose the "problem" with ^^ is that it looks weird; why separate two non-`Condition`al property groups like that?  (The answer being "smaller patches!", but that might not actually be a good rationale…)

---

My question was *not* "why did `$(CoreResolveReferencesDependsOn)` and `$(ResolveReferencesDependsOn)` need to become conditional?".

My question was "why is this diff so noisy?"

src/Xamarin.Android.Build.Tasks/Tasks/RuntimeIdentifierToAbi.cs
```diff
@@ -14,17 +14,21 @@ public class RuntimeIdentifierToAbi : AndroidTask
...
 
+		public string [] RuntimeIdentifiers { get; set; }
+
 		[Output]
 		public string SupportedAbis { get; set; }
 
 		public override bool RunTask ()
 		{
-			if (!string.IsNullOrEmpty (RuntimeIdentifier)) {
+			if (RuntimeIdentifiers != null && RuntimeIdentifiers.Length > 0) {
```
jonpryor, line 24: Implicit to this is that `$(RuntimeIdentifiers)` takes *precedence* over `$(AndroidSupportedAbis)`.  Is this desirable/intentional?  What will be the fallout to this decision?

For example, if I have an existing `App.csproj` which overrides `$(AndroidSupportedAbis)`, and then I migrate to .NET 5/6 and *don't* touch `$(RuntimeIdentifiers)`… What Happens?

Given that this PR always provides a default `$(RuntimeIdentifiers)` value, what I *think* will happen is that we'll effectively *ignore* `$(AndroidSupportedAbis)`, replacing the "original overridden value" with the defaults that *we* provide.

Is this useful or desirable?  (It might be!  I don't know!  I *am* wondering if there's been any consideration here, particularly as the commit message doesn't mention this semantic.)

Meanwhile, is the *opposite* possible?  If `$(AndroidSupportedAbis)` is overridden, and `$(RuntimeIdentifiers)` *isn't* -- though making this determination may itself be difficult! -- would it be at all possible to populate `$(RuntimeIdentifiers)` based on the value of `$(AndroidSupportedAbis)`?

src/Xamarin.Android.Build.Tasks/Tasks/RuntimeIdentifierToAbi.cs
```diff
@@ -14,17 +14,21 @@ public class RuntimeIdentifierToAbi : AndroidTask
...
+				SupportedAbis = string.Join (";", RuntimeIdentifiers.Select (MonoAndroidHelper.RuntimeIdentifierToAbi));
+			} else if (!string.IsNullOrEmpty (RuntimeIdentifier)) {
 				SupportedAbis = MonoAndroidHelper.RuntimeIdentifierToAbi (RuntimeIdentifier);
 			}
 			if (string.IsNullOrEmpty (SupportedAbis)) {
```
jonpryor, line 29: Different question: why isn't this a dead code path?  How could we possibly get here, given that *both* `$(AndroidSupportedAbis)` and `$(RuntimeIdentifiers)` provide default values if they're initially empty?

Though I suppose if someone explicitly overrides on the command-line to be empty -- is that possible? -- that might trigger this codepath?

    msbuild /p:AndroidSupportedAbis= /p:RuntimeIdentifiers=

though that feels like an "unkind" move, and perhaps *deserves* a build failure.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
```diff
@@ -267,4 +269,118 @@ projects. .NET 5 projects will not import this file.
     />
   </Target>
 
+  <Target Name="_PrepareAssemblies"
+    DependsOnTargets="$(_PrepareAssembliesDependsOnTargets)">
```
jonpryor, line 273: TODO: Indentation

src/Xamarin.Android.Build.Tasks/Tasks/RuntimeIdentifierToAbi.cs
jonathanpeppers: It would get here if you provided `RuntimeIdentifiers=somethinginvalid`, so maybe this just needs to be an error instead of silently using some default.

src/Xamarin.Android.Build.Tasks/Tasks/RuntimeIdentifierToAbi.cs
jonathanpeppers: Eventually, I think that `RuntimeIdentifier` and/or `RuntimeIdentifiers` will be implicitly set via:

    <TargetFrameworkVersion>net5.0-android</TargetFrameworkVersion>

And so, it seems like `$(AndroidSupportedAbis)` would always be ignored going forward? Is it worth trying to support both?

Either way, I should document something about `$(RuntimeIdentifiers)` being preferred in `DotNet5.md`.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
jonathanpeppers: >  I suppose the "problem" with ^^ is that it looks weird; why separate two non-Conditional property groups like that? (The answer being "smaller patches!", but that might not actually be a good rationale…)

I think it should be ok to move code around that is being changed so the final code looks better? This would be like creating a new C# method when it gets too long. Diff size isn't a huge concern to me if it was code that is already changing.

If this was auto-formatting unrelated files, we shouldn't do that, of course.

src/Xamarin.Android.Build.Tasks/Tasks/RuntimeIdentifierToAbi.cs
jonathanpeppers: I added `XA0035`, which seems a lot better.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Linker.targets
```diff
@@ -10,6 +10,7 @@ This file contains the .NET 5-specific targets to customize ILLink
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <Target Name="_PrepareLinking"
+      Condition=" '$(PublishTrimmed)' == 'true' "
```
jonpryor, line 13: Am I to infer `$(PublishTrimmed)` is set when linking is enabled within `dotnet publish`?

What happens when linking *isn't* enabled during `dotnet publish`?  We still need to run a *subset* of linker steps, even in "Debug config builds", which presumably would be "`dotnet publish` when `$(PublishTrimmed)`=False" (?).  See also f96fcf93e157472072576bcc0a8698302899e8cf.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -327,23 +327,24 @@ void AddAssemblies (ZipArchiveEx apk, bool debug, bool compress, IDictionary<str
...
-				AddAssemblyConfigEntry (apk, config);
+				AddAssemblyConfigEntry (apk, assemblyPath, config);
 
 				// Try to add symbols if Debug
 				if (debug) {
 					var symbols = Path.ChangeExtension (assembly.ItemSpec, "dll.mdb");
 
 					if (File.Exists (symbols))
-						AddFileToArchiveIfNewer (apk, symbols, AssembliesPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
+						AddFileToArchiveIfNewer (apk, symbols, assemblyPath + Path.GetFileName (symbols), compressionMethod: UncompressedMethod);
```
jonpryor, line 342: Instead of `assemblyPath +`, should this instead be `Path.Combine()`?

src/Xamarin.Android.Build.Tasks/Tasks/RuntimeIdentifierToAbi.cs
jonpryor: If `$(AndroidSupportedAbis)` will be *ignored* going forward, then we should consider adding a warning if the property is set at all, as having something set which doesn't do anything will be a source of confusion.

This may complicate migrating from "long-form" to "short-form" projects, i.e. it'll be a property that needs to be *removed* as part of the migration, but this may be for the best.

Alternatively, we would need to ensure that if `$(AndroidSupportedAbis)` is set, the values within it *contribute to* `$(RuntimeIdentifiers)`, so that when it's set, the specified ABIs will still appear in the resulting `.apk`/`.aab`.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Linker.targets
jonathanpeppers: `$(PublishTrimmed)` is the flag that decides if the `ILLink` target should run or not:

https://github.com/dotnet/sdk/blob/d30e9ba1fe915d99cd3455bda9b41cde42775fbe/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L32-L35

Our targets will enable this if `$(AndroidLinkMode)` is not `None`.

We could remove this `Condition` and it would not break anything. I just changed this for consistency: we don't really need to change `ILLink` settings if it isn't going to run.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
jonathanpeppers: This is the path within the zip file, so we always want to use `/`.

Documentation/guides/messages/xa0035.md
```diff
@@ -0,0 +1,28 @@
...
+```xml
+<PropertyGroup>
+  <RuntimeIdentifiers>android.21-arm;android.21-arm64;android.21-x86;android.21-x64</RuntimeIdentifiers>
+</PropertyGroup>
+```
+
+See the Microsoft documentation on [runtime identifiers][rids] for more
+information.
+
+[rids]: https://docs.microsoft.com/dotnet/core/rid-catalog
```
brendanzagaeski, line 28: Probably it was already in mind when adding this info, but this link might need to be updated if the eventual page that talks about Android RIDs ends up being elsewhere. I think what's written here is good for now. I'm mostly just thinking out loud about the future.

src/monodroid/jni/basic-android-system.cc
```diff
@@ -61,3 +62,15 @@ BasicAndroidSystem::determine_primary_override_dir (jstring_wrapper &home)
...
 }
+
+const char*
+BasicAndroidSystem::get_built_for_abi_name ()
+{
+	if (built_for_abi_name == nullptr) {
+		unsigned short built_for_cpu = 0, running_on_cpu = 0;
+		unsigned char is64bit = 0;
+		_monodroid_detect_cpu_and_architecture (&built_for_cpu, &running_on_cpu, &is64bit);
+		built_for_abi_name = android_abi_names [built_for_cpu];
```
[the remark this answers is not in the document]
grendello: As long as we stick to the `CPU_*` macros (and we do, by using `_monodroid_detect_cpu_and_architecture`), [we're fine](https://github.com/xamarin/xamarin-android/blob/3dcac85d716f30dc2ccea38ae11503be77881a84/src/monodroid/jni/basic-android-system.hh#L18-L27). If you want to be absolutely sure we're within the boundaries of the array, we can compare `built_for_cpu` with `ANDROID_ABI_NAMES_SIZE`

Documentation/guides/messages/xa0035.md
```diff
@@ -0,0 +1,28 @@
...
+## Issue
+
+Xamarin.Android was not able to determine the application's target
+[Android ABIs][abis] as specified by the `.csproj` file.
+
+[abis]: https://developer.android.com/ndk/guides/abis
+
+## Solution
+
+Make sure you are declaring valid values for `$(RuntimeIdentifiers)`:
```
brendanzagaeski, line 17: I think this wording is OK for now since, as far as I know, the full IDE experience is not yet know. Depending on how the IDE integration turns out, this can be updated to give a few more details about how to edit the `.csproj` file, like:

````
Open the project file [in Visual Studio][edit-project-files] or another text
editor and make sure all of the values in the `RuntimeIdentifiers` MSBuild
property are valid:

```xml
<PropertyGroup>
  <RuntimeIdentifiers>android.21-arm;android.21-arm64;android.21-x86;android.21-x64</RuntimeIdentifiers>
</PropertyGroup>
```

[edit-project-files]: https://docs.microsoft.com/visualstudio/msbuild/visual-studio-integration-msbuild#edit-project-files-in-visual-studio
````

File: src/Xamarin.Android.Build.Tasks/Properties/Resources.resx [hunk 13]
```diff
@@ -193,6 +193,11 @@ In this message, the phrase "should not be reached" means that this error messag
     <value>Failed to get the Java SDK version. Please ensure you have Java {0} or above installed.</value>
     <comment>{0} - The Java version number</comment>
   </data>
+  <data name="XA0035" xml:space="preserve">
+    <value>Unable to determine the Android ABI from the value '{0}'. Set $(RuntimeIdentifiers) to valid identifiers for the Android platform.</value>
```
brendanzagaeski, line 197: It's not too important in this case since I bet the wording will be updated again as the .NET 6 story evolves, but for an extra margin of safety, I'll suggest making the initial wording:

```
<value>Unable to determine the Android ABI from the value '{0}'. Edit the project file in a text editor and set the 'RuntimeIdentifiers' MSBuild property to contain only valid identifiers for the Android platform.</value>
```

Motivation: Give users across a range of familiarity with MSBuild and across all the different translations of the message the best chance at finding the `RuntimeIdentifiers` MSBuild property (a lesson I learned recently in <https://github.com/xamarin/xamarin-android/commit/91022a97214bc6611e61354f723e3b47a0b119e0>).

src/Xamarin.Android.Build.Tasks/Tasks/RuntimeIdentifierToAbi.cs
jonathanpeppers: Since `$(RuntimeIdentifiers)` is needed for `Restore` to even work... I think we really should require it over `$(AndroidSupportedAbis)`.

I added `XA0036` if `$(AndroidSupportedAbis)` is present, @brendanzagaeski you might review what I added for this, thanks.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -193,6 +193,15 @@ In this message, the phrase "should not be reached" means that this error messag
     <value>Failed to get the Java SDK version. Please ensure you have Java {0} or above installed.</value>
     <comment>{0} - The Java version number</comment>
   </data>
+  <data name="XA0035" xml:space="preserve">
+    <value>Unable to determine the Android ABI from the value '{0}'. Edit the project file in a text editor and set the 'RuntimeIdentifiers' MSBuild property to contain only valid identifiers for the Android platform.</value>
+    <comment>The following are literal names and should not be translated: Android, ABI, RuntimeIdentifiers, MSBuild
+{0} - The $(RuntimeIdentifier) supplied by the user.</comment>
+  </data>
+  <data name="XA0036" xml:space="preserve">
+    <value>The 'AndroidSupportedAbis' MSBuild property is no longer supported. Edit the project file in a text editor and use the 'RuntimeIdentifiers' MSBuild property instead.</value>
```
brendanzagaeski, line 202: This might be silly, but for bonus instructiveness points, maybe:

```
The 'AndroidSupportedAbis' MSBuild property is no longer supported. Edit the project file in a text editor, remove any uses of 'AndroidSupportedAbis', and use the 'RuntimeIdentifiers' MSBuild property instead.
```

jonpryor:

```
For .NET 5, obsolete the `$(AndroidSupportedAbis)` MSBuild property
and replace with the `$(RuntimeIdentifiers)` MSBuild property.

The `$(AndroidSupportedAbis)` MSBuild property has long been used to
control which Android ABIs are included in a `.apk` or `.aab` file,
unless `$(AndroidCreatePackagePerAbi)`=True.

The .NET Core analogue is the [`$(RuntimeIdentifiers)` property][0],
which is a `;`-separated list of "RID" values from the
[.NET Core RID catalog][1].

The mapping between `$(AndroidSupportedAbis)` and
`$(RuntimeIdentifiers)` values is as follows:

  * armeabi-v7a: android.21-arm
  *   arm64-v8a: android.21-arm64
  *         x86: android.21-x86
  *      x86_64: android.21-x64

Within a .NET 5+/`dotnet build` context, the `$(AndroidSupportedAbis)`
property is *no longer supported*; `$(AndroidSupportedAbis)` will not
be supported in .NET 5 because `$(RuntimeIdentifiers)` may be
implicitly set based on the `$(TargetFrameworks)` property; it may not
be possible to override `$(RuntimeIdentifiers)` "early enough" to take
`$(AndroidSupportedAbis)` into consideration.

If the `$(AndroidSupportedAbis)` property present, an XA0036 warning
will be generated:

	warning XA0036: The 'AndroidSupportedAbis' MSBuild property is no longer supported.
	Edit the project file in a text editor, remove any uses of 'AndroidSupportedAbis',
	and use the 'RuntimeIdentifiers' MSBuild property instead.\

A `HelloAndroid.csproj` targeting all architectures might look like:

	  <PropertyGroup>
	    <TargetFramework>net5.0-android</TargetFramework>
	    <RuntimeIdentifiers>android.21-x86;android.21-x64;android.21-arm;android.21-arm64</RuntimeIdentifiers>
	    <OutputType>Exe</OutputType>
	  </PropertyGroup>
	</Project>

To achieve this, we can do a nested `<MSBuild/>` call:

	<ItemGroup>
	  <_RIDs Include="$(RuntimeIdentifiers)" />
	</ItemGroup>
	<MSBuild
	    Projects="$(MSBuildProjectFile)"
	    Targets="_ComputeFilesToPublishForRuntimeIdentifiers"
	    Properties="RuntimeIdentifier=%(_RIDs.Identity)">
	  <Output TaskParameter="TargetOutputs" ItemName="ResolvedFileToPublish" />
	</MSBuild>

For a new target:

	<Target Name="_ComputeFilesToPublishForRuntimeIdentifiers"
	    DependsOnTargets="ResolveReferences;ComputeFilesToPublish"
	    Returns="@(ResolvedFileToPublish)">
	</Target>

This is equivalent to what would happen in the legacy
`_ResolveAssemblies` target.

The linker then outputs files to:

	obj\Release\net5.0\$(RuntimeIdentifier)\linked\

One problem with this is that you receive multiple BCL assemblies for
each architecture:

	obj\Release\net5.0\android.21-arm\linked\System.Linq.dll
	obj\Release\net5.0\android.21-arm64\linked\System.Linq.dll
	obj\Release\net5.0\android.21-x86\linked\System.Linq.dll
	obj\Release\net5.0\android.21-x64\linked\System.Linq.dll

I created a `<ProcessAssemblies/>` MSBuild task that checks the MVID
of each assembly to detect duplicates.  Non-duplicates will be placed
in a sub-directory within the `.apk` file, such as:

  * `assemblies\HelloAndroid.dll`
  * `assemblies\System.Linq.dll`
  * `assemblies\arm64-v8a\System.Private.CoreLib.dll`
  * `assemblies\armeabi-v7a\System.Private.CoreLib.dll`
  * `assemblies\x86\System.Private.CoreLib.dll`
  * `assemblies\x86_64\System.Private.CoreLib.dll`

I had to make a few build & runtime changes for these ABI
sub-directories to work.

Other notes:

  * We must pass `@(IntermediateAssembly)` from the outer build to the
    nested `<MSBuild/>` calls.  Otherwise, the inner calls look for
    `obj\Release\net5.0\android.21-arm\HelloAndroid.dll` instead of
    `obj\Release\net5.0\HelloAndroid.dll` that actually exists.

  * `_ComputeFilesToPublishForRuntimeIdentifiers` was running some
    extra targets causing `<ResolveSdks/>` to run 5 times for 4 ABIs.
    I had to make changes in `BuildOrder.targets` to fix this.

  * `<ResolveAssemblies/>` now sets `%(IntermediateLinkerOutput)` on
    each assembly item.  This allows .NET 6 builds to support multiple
    linker directories.

[0]: https://docs.microsoft.com/en-us/dotnet/core/tools/csproj#runtimeidentifiers
[1]: https://docs.microsoft.com/en-us/dotnet/core/rid-catalog
```

PR #4768, opened 2020-06-05
Revert api21

grendello:


jonpryor:
Closing, as we reverted the changes on master, so I'll just CP those over.

PR #4774, opened 2020-06-05
[tests] set /p:JavaSdkDirectory for MSBuild tests

jonpryor:
This makes the behavior match what we are doing for the
`$(AndroidSdkDirectory)` and `$(AndroidNdkDirectory)` properties.

Additionally, if `$(JavaSdkDirectory)` isn't set, try to use the
`JI_JAVA_HOME` and `JAVA_HOME` environment variables, to allow paths
to be reasonably overridden for test purposes.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs
@@ -302,6 +302,10 @@ namespace Xamarin.ProjectTools
 				if (Directory.Exists (ndkPath)) {
 					sw.WriteLine (" /p:AndroidNdkDirectory=\"{0}\" ", ndkPath);
 				}
+				string jdkPath = AndroidSdkResolver.GetJavaSdkPath ();
+				if (Directory.Exists (jdkPath)) {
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs
```diff
@@ -302,6 +302,10 @@ protected bool BuildInternal (string projectOrSolution, string target, string []
 				if (Directory.Exists (ndkPath)) {
 					sw.WriteLine (" /p:AndroidNdkDirectory=\"{0}\" ", ndkPath);
 				}
+				string jdkPath = AndroidSdkResolver.GetJavaSdkPath ();
+				if (Directory.Exists (jdkPath)) {
+					sw.WriteLine (" /p:JavaSdkDirectory=\"{0}\" ", jdkPath);
+				}
```
jonathanpeppers, line 308: There is a change like this in `DotNetCLI.cs`, too:

https://github.com/xamarin/xamarin-android/pull/4567/files#diff-0bf2fe5f45104a6362159a989e222e3bR132

jonpryor: merged 4a320361c3cd2f720960ee9955038938d46d088e.

Thanks!

PR #4779, opened 2020-06-06
[Xamarin.Android.Build.Tasks] Improve XA1011 error wording

brendanzagaeski:
Context: https://developercommunity.visualstudio.com/content/problem/1043110/xamarin-%E9%A1%B9%E7%9B%AE%E5%9C%A8vs165%E7%89%88%E6%9C%AC%E7%BC%96%E8%AF%91%E6%AD%A3%E5%B8%B8%E5%8D%87%E7%BA%A7%E5%88%B0vs166%E6%97%A0%E6%B3%95%E7%BC%96%E8%AF%91%E8%BF%90%E8%A1%8C.html

Seeing the localized version of the message for XA1011 in context from a
real user's build made @brendanzagaeski realize that the original
wording he wrote for that message was problematic.

One issue was that messages for Xamarin.Android build errors are not
currently accompanied by their numbered error codes when they appear in
the Error List in Visual Studio.  This means that users who see a
localized translation for a message will likely search online using the
localized text instead of the numbered code and will have a more
challenging time finding corresponding release notes and documentation
pages.

Another problem was that the error message used "`$(AndroidLinkTool)`"
as a kind of abbreviation for "`AndroidLinkTool` MSBuild property".  The
idea was that `$(AndroidLinkTool)` matched the MSBuild syntax for
retrieving a property value and would therefore indicate that
`AndroidLinkTool` was an MSBuild property, but Xamarin.Android users
don't typically need to interact too much with MSBuild syntax, so they
wouldn't necessarily recognize the meaning of the abbreviation.  On top
of that, in user projects, `AndroidLinkTool` would be more likely to
appear in an XML tag like `<AndroidLinkTool>` than in a property
retrieval like `$(AndroidLinkTool)`, so the use of `$(AndroidLinkTool)`
in the error message could potentially cause some confusion.

To help address these problems, adjust the wording of the XA1011 error.
In addition to replacing "`$(AndroidLinkTool)`" with "`AndroidLinkTool`
MSBuild property", make the message more actionable by adding
information about how to change the corresponding code shrinker setting
in the Visual Studio project property pages.

[A number of the messages from the dotnet/sdk repo][0] use similar
wording to refer to MSBuild properties.  One difference is that those
messages say "property" instead of "MSBuild property."  @brendanzagaeski
prefers "MSBuild property" because he believes it will help prevent
ambiguity, especially in the localized translations.

TODO: Audit all of the other errors and warnings that currently use the
`$()` syntax, and make similar changes as appropriate.

[0]: https://github.com/dotnet/sdk/blob/1ce8d7029b267c335bf606c591a97bfb9824ece1/src/Tasks/Common/Resources/Strings.resx

### DIFF
--- a/Documentation/guides/messages/README.md
+++ b/Documentation/guides/messages/README.md
@@ -81,7 +81,7 @@ ms.date: 01/24/2020
 + [XA1008](xa1008.md): The TargetFrameworkVersion (Android API level {compileSdk}) is lower than the targetSdkVersion ({targetSdk}).
 + [XA1009](xa1009.md): The {assembly} is Obsolete. Please upgrade to {assembly} {version}
 + [XA1010](xa1010.md): Invalid \`$(AndroidManifestPlaceholders)\` value for Android manifest placeholders. Please use \`key1=value1;key2=value2\` format. The specified value was: `{placeholders}`
-+ [XA1011](xa1011.md): Using ProGuard with the D8 DEX compiler is no longer supported. Please update \`$(AndroidLinkTool)\` to \`r8\`.
++ [XA1011](xa1011.md): Using ProGuard with the D8 DEX compiler is no longer supported. Please set the code shrinker to \`r8\` in the Visual Studio project property pages or edit the project file in a text editor and set the \`AndroidLinkTool\` MSBuild property to \`r8\`.
 + XA1012: Included layout root element override ID '{id}' is not valid.
 + XA1013: Failed to parse ID of node '{name}' in the layout file '{file}'.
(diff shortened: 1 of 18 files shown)


brendanzagaeski:
There might be another wave of users who will see this XA1011 message in d16-7 if they change their DEX compiler after seeing the new warning about DX from <https://github.com/xamarin/xamarin-android/commit/2662fd15ab9cd840b93281456480aaf24db2a2d8>. So if this wording change looks good, it could be a good candidate to cherry-pick to d16-7.

PR #4783, opened 2020-06-08
[Xamarin.Android.Build.Tasks] Add XA1027 & XA1028 warnings for old ProGuard properties

brendanzagaeski:
Context: https://developercommunity.visualstudio.com/content/problem/1043110/index.html
Context: b3baf32

If a project with `$(AndroidDexTool)`=`d8` had
`$(EnableProguard)`=`true` or `$(AndroidEnableProguard)`=`true` set in
addition to `$(AndroidLinkTool)`=`proguard`, the build would only emit
the XA1011 error about `$(AndroidLinkTool)`:

	error XA1011: Using ProGuard with the D8 DEX compiler is no longer supported. Please set the code shrinker to `r8` in the Visual Studio project property pages or edit the project file in a text editor and set the `AndroidLinkTool` MSBuild property to `r8`.

If upon seeing this error message, the project author decided to try
clearing `$(AndroidLinkTool)` to `` as a first sanity check before
setting it to `r8`, they would get a confusing result.  The XA1011 error
would still appear, with no new hints about why the error was still
appearing!

Improve that scenario by adding new warnings that describe how
`$(EnableProguard)` and `$(AndroidEnableProguard)` automatically set
`$(AndroidLinkTool)` if it is blank.

Other changes:

Use single quotes in the new messages and change the quote character in
the XA1011 message from backquote to single quote to align more closely
with the format of strings from the `.resx` files in dotnet/sdk and
microsoft/msbuild.


brendanzagaeski:
I still need to add a couple things to this PR:

- Documentation pages for the new warnings
- A draft release note

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -269,11 +269,13 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
         <_AndroidApplicationSharedLibraryPath>$(IntermediateOutputPath)app_shared_libraries\</_AndroidApplicationSharedLibraryPath>
 	<_ResolvedUserAssembliesHashFile>$(IntermediateOutputPath)resolvedassemblies.hash</_ResolvedUserAssembliesHashFile>
 
+	<AndroidDexTool   Condition=" '$(AndroidDexTool)' != 'dx' ">d8</AndroidDexTool>
```
brendanzagaeski, line 272: I moved this `$(AndroidDexTool)` reassignment a few lines up so that the new properties are evaluated _after_ this reassignment but _before_ `$(AndroidEnableProguard)` gets reassigned. That way they see `d8` as expected when it is turned on for an empty or invalid `$(AndroidDexTool)` from the app project, but they see the original `$(EnableProguard)` and `$(AndroidEnableProguard)` values from the app project.

Documentation/guides/messages/README.md
```diff
@@ -81,7 +81,7 @@ ms.date: 01/24/2020
 + [XA1008](xa1008.md): The TargetFrameworkVersion (Android API level {compileSdk}) is lower than the targetSdkVersion ({targetSdk}).
 + [XA1009](xa1009.md): The {assembly} is Obsolete. Please upgrade to {assembly} {version}
 + [XA1010](xa1010.md): Invalid \`$(AndroidManifestPlaceholders)\` value for Android manifest placeholders. Please use \`key1=value1;key2=value2\` format. The specified value was: `{placeholders}`
-+ [XA1011](xa1011.md): Using ProGuard with the D8 DEX compiler is no longer supported. Please set the code shrinker to \`r8\` in the Visual Studio project property pages or edit the project file in a text editor and set the \`AndroidLinkTool\` MSBuild property to \`r8\`.
++ [XA1011](xa1011.md): Using ProGuard with the D8 DEX compiler is no longer supported. Please set the code shrinker to 'r8' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidLinkTool' MSBuild property to 'r8'.
```
brendanzagaeski, line 84: I don't have a strong opinion about this, so I'd be fine switching these back and using `` ` `` in the new messages for now. My thinking was to take one step closer to aligning with dotnet/sdk message formatting as .NET 6 approaches.

Documentation/guides/messages/xa1024.md
```diff
@@ -0,0 +1,43 @@
...
+description: XA1024 warning code
+ms.date: 06/19/2020
+---
+# Xamarin.Android warning XA1024
+
+## Example messages
+
+```
+warning XA1024: The 'EnableProguard' MSBuild property is set to 'true' and the 'AndroidLinkTool' MSBuild property is empty, so 'AndroidLinkTool' will default to 'proguard'.
+error XA1011: Using ProGuard with the D8 DEX compiler is no longer supported. Please set the code shrinker to 'r8' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidLinkTool' MSBuild property to 'r8'.
```
brendanzagaeski, line 12: All in all, the new messages might be slightly excessive because they always appear alongside the XA1011 message, which does say to set `$(AndroidLinkTool)` to `r8`, and that will indeed resolve the error in call cases. So these new warnings are catching a fairly slim scenario where a user wants to try to disable code shrinking entirely instead of moving to `r8`. On top of that, since these new messages are warnings, if the user happens to have warnings temporarily hidden in the Error List, they wouldn't even see them alongside the XA1011 error.

I think there's a chance the new warnings could still help a few users here and there as DX moves closer to deprecation, but I'd also be OK with skipping this PR if it does indeed seem excessive. These messages will all be irrelevant anyway once DX compatibility is removed.

brendanzagaeski:
The one test failure in the initial test run was unrelated. Rerunning the tests because a couple of the phases timed out.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -4306,6 +4306,62 @@ public void XA1018 ()
 			}
 		}
 
+		[Test]
+		public void XA1024XA1025 ([Values (null, "proguard", "r8")] string linkTool, [Values (null, "true", "false")] string enableProguard, [Values (null, "true", "false")] string androidEnableProguard)
```
jonathanpeppers, line 4310: Since this is adding 9 tests (3 * 3 * 3)... can we use `[TestCaseSource]` and list a few less combinations?

Maybe we could get by with 3-5 combinations or something?

brendanzagaeski:
**Changes in the latest force-push and commit:**

- Rebase.
- Increment the message code numbers so they don't conflict with now already existing XA1024.
- Use `[TestCaseSource]` to reduce the number of scenarios for the new test from twenty-seven down to five.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -265,11 +265,13 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
...
+	<_AndroidXA1025 Condition=" '$(EnableProguard)' == 'true' And '$(AndroidEnableProguard)' == '' And '$(AndroidDexTool)' == 'd8' And $(AndroidLinkTool) == '' ">true</_AndroidXA1025>
+	<_AndroidXA1026 Condition=" '$(AndroidEnableProguard)' == 'true' And '$(AndroidDexTool)' == 'd8' And $(AndroidLinkTool) == '' ">true</_AndroidXA1026>
 	<!-- $(EnableProguard) is an obsolete property that should be removed at some stage. -->
 	<AndroidEnableProguard Condition=" '$(AndroidEnableProguard)' == '' ">$(EnableProguard)</AndroidEnableProguard>
 	<AndroidEnableProguard Condition=" '$(AndroidLinkTool)' == 'proguard' ">True</AndroidEnableProguard>
 	<AndroidEnableMultiDex Condition=" '$(AndroidEnableMultiDex)' == '' ">False</AndroidEnableMultiDex>
-	<AndroidDexTool   Condition=" '$(AndroidDexTool)' != 'dx' ">d8</AndroidDexTool>
 	<AndroidDexTool   Condition=" '$(AndroidLinkTool)' == 'r8' ">d8</AndroidDexTool>
```
jonathanpeppers, line 275: I know this isn't this PR, but maybe we should go ahead and change this line:
```suggestion
	<AndroidDexTool   Condition=" '$(AndroidLinkTool)' == 'r8' and '$(AndroidDexTool)' == 'dx' ">d8</AndroidDexTool>
```
I think this will prevent MSBuild from logging that the property changed from `d8` to `d8`.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -4245,6 +4245,91 @@ public void XA1018 ()
...
+			var proj = new XamarinAndroidApplicationProject {
+				// Make sure the test covers the scenario where `$(AndroidDexTool)` is not explicitly configured
+				DexTool = null,
+				LinkTool = linkTool
+			};
+			proj.SetProperty ("EnableProguard", enableProguard);
+			proj.SetProperty ("AndroidEnableProguard", androidEnableProguard);
+			using (var builder = CreateApkBuilder ()) {
+				builder.Target = "_CheckNonIdealConfigurations";
+				if (linkTool == "r8") {
```
dellis1972, line 4290: In other tests rather than having these kinds of `if/else` blocks, we include the expected results in the input parameters and in the `TestCaseSource` arguments.  e.g

```
new object [] {
 	/* linkTool */                   null,
 	/* enableProguard */             null,
 	/* androidEnableProguard */      "true",
        /* expectedBuildResult */ false,
        /* expectNumWarnings */ 0,
       /* expectedWarningCode */ "XA1025",                       
}
```

This would allow you to collapse all that also identical code into maybe one code block.

Also you should be able make the `androidEnableProguard` a nullable bool. 'bool? androidEnableProguard' so you can do an actual boolean evaluation rather than doing a string comparison.

dellis1972: A good example is here https://github.com/xamarin/xamarin-android/blob/c477f1175b9b07fe4b6d4765aecb011dea1f0db8/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/AndroidRegExTests.cs#L170

brendanzagaeski: Very nice. Applied that change!

brendanzagaeski:
**Changes in the latest force-push and commit:**

- Rebase.
- Increment the message code numbers so they don't conflict with the now in-progress XA1026.
- Switch to including the expected results in the `TestCaseSource` arguments to simplify the test logic.

**Questions:**

There's a chance it might be appropriate to close this PR rather than merging it. If Xamarin.Android will remove DX and ProGuard support in the Visual Studio release that comes after February 2021, then this PR probably isn't needed anymore. On the other hand, if Xamarin.Android legacy projects will retain backwards compatibility with DX and ProGuard for a little while after February 2021 for users who manually install and select older Android SDK Build-Tools versions, then this PR might still offer some value.

For some sense of scale, on Visual Studio 2019 version 16.7, over the past week or so, the ratio of number of builds with `d8` enabled to builds with `dx` enabled was about 25:1. Adding in the builds from Visual Studio 2019 version 16.4, the ratio shifts to about 10:1.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
brendanzagaeski: Oh shoot. I forgot to take a look at adding this in the latest changes today. I'll plan to take a look the next time I'm working.

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -3886,6 +3886,67 @@ public void XA1018 ()
...
+				/* enableProguard */             null,
+				/* androidEnableProguard */      "true",
+				/* expectedBuildResult */        false,
+				/* expectedWarning */            "warning XA1028:",
+			}
+		};
+
+		[Test]
+		[TestCaseSource (nameof (XA1027XA1028Source))]
+		public void XA1027XA1028 (string linkTool, string enableProguard, string androidEnableProguard, bool expectedBuildResult, string expectedWarning)
```
jonathanpeppers, line 3929: I think this new test fails in some cases:

image

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
brendanzagaeski: I got a chance to experiment with this today, and it looks like MSBuild has some built-in smarts about this.

Here's an excerpt of the diff of the diagnostic build output, comparing `$(AndroidDexTool)`=`dx` + `$(AndroidLinkTool)`=`r8` versus `$(AndroidDexTool)`=`d8` + `$(AndroidLinkTool)`=`r8`:

```diff
@@ -250,7 +250,6 @@ Property reassignment: $(ResolveNuGetPackageAssetsDependsOn)="
                _SetLatestTargetFrameworkVersion;
                ResolveProjectReferences
        " (previous value: "ResolveProjectReferences") at C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets (148,2)
-Property reassignment: $(AndroidDexTool)="d8" (previous value: "dx") at C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets (274,2)
 Property reassignment: $(AndroidEnableProguard)="True" (previous value: "") at C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets (279,2)
 Property reassignment: $(AssemblySearchPaths)="
                {HintPathFromItem};
```

That is, MSBuild logs a message when it changes the property value from `dx` to `d8`, but when the property value is already `d8` and MSBuild sets it set to `d8` again, it remains silent. I also see that this comment has now been marked as resolved. So I'll leave this `<AndroidDexTool/>` `Condition` as-is for now, but I'll be happy to adjust it if it's still a desired change for tidiness or similar.

brendanzagaeski:
**Changes in the commit:**

Correct the new test so it runs in the Release configuration and doesn't get warnings about using R8 and ProGuard in the Debug configuration.

jonathanpeppers (commented):
@brendanzagaeski whoops, there are conflicts now, sorry!

brendanzagaeski:
**Changes in the latest force-push and commit:**

- Rebase.
- Tidy formatting.

brendanzagaeski:

Rebase and squash.

PR #4786, opened 2020-06-08
[Xamarin.Android.Build.Tasks] fix incremental <BuildApk/> on Windows

jonathanpeppers:
Fixes: https://developercommunity.visualstudio.com/content/problem/1045818/after-update-vs-2019-xamarin-app-debug-throws-exce.html
Fixes: https://feedback.devdiv.io/1045818

On Windows, if your project contains an `AndroidAsset` such as:

    Assets\foo\bar.txt

And you do these steps:

1. Build & run the project.
2. Disconnect/reconnect the device or close the emulator.
3. Build & run the project again.

At this phase, the new `.apk` file will be missing
`Assets\foo\bar.txt`! Any code depending on it, would fail.

Reviewing the build log for the `<BuildApk/>` MSBuild task:

    Registering item assets/foo/bar.txt
    Deregistering item assets/foo\bar.txt
    Removing assets/foo/bar.txt as it is not longer required.

In 1c4727cb, support was added to make `<BuildApk/>` incrementally
update the `.apk` file. However, it did not take into account the
problem encountered in f35622082. On Windows, `aapt2` will produce
`assets/foo\bar.txt` with weird mixture of path separators within the
`.apk` file.

The simplest fix here, is to check for `\\` within the existing
`packaged_resources` file. Then fix up the path as needed.

I added logging and a test around this situation.


brendanzagaeski:
_Draft release notes_

During the finalization of this PR for merging, when you get a chance, you can add a release note for it to `Documentation/release-notes/4786.md` as part of the PR. Here's the template to fill out. Thanks!

```

- [Developer Community 1045818](https://developercommunity.visualstudio.com/content/problem/1045818/after-update-vs-2019-xamarin-app-debug-throws-exce.html):
  Starting in Xamarin.Android 10.3, errors similar to
  _Java.Lang.RuntimeException: 'Font asset not found Fonts/fontello.ttf'_ ...
```

jonathanpeppers:
I thought to test out: https://mvnrepository.com/artifact/com.android.tools.build/aapt2/4.0.0-6051327

And it seems like Google has fixed the problem. I think we should  just bump `aapt2` instead?

jonathanpeppers:
Nevermind, this version still does not work:

https://mvnrepository.com/artifact/com.android.tools.build/aapt2/4.0.0-6051327

The regex was just not matching:

```diff
diff --git a/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAndroidTooling.cs b/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAndroidTooling.cs
index 063c8bac..776de449 100644
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAndroidTooling.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAndroidTooling.cs
@@ -213,7 +213,8 @@ namespace Xamarin.Android.Tasks
                }

                //  Android Asset Packaging Tool (aapt) 2:19
-               static readonly Regex Aapt2VersionRegex = new Regex (@"(?<version>[\d\:]+)(\d+)?");
+               //  Android Asset Packaging Tool (aapt) 2.19-6051327
+               static readonly Regex Aapt2VersionRegex = new Regex (@"(?<version>[\d\:|\.]+)(\d+)?");
```

This was just making it revert to `aapt`, which works fine.

Documentation/release-notes/4786.md
```diff
@@ -0,0 +1,13 @@
+#### Application behavior on device and emulator
+
+- [Developer Community 1045818][0]: Starting in Xamarin.Android 10.3,
+  errors similar to _Java.Lang.RuntimeException: 'Font asset not found
+  Fonts/fontello.ttf'_ could occur during an incremental build in
+  Visual Studio on Windows. There is a [known issue][1] with the
```
brendanzagaeski, line 6: I think the issue occurs when the app is running on device rather than during the build itself? So maybe a rewording of the first sentence could be:

```
[Developer Community 1045818][0]: Starting in Xamarin.Android 10.3 on Windows,
errors similar to _Java.Lang.RuntimeException: 'Font asset not found
Fonts/fontello.ttf'_ could cause apps to abort unexpectedly on device after they
were updated and redeployed from Visual Studio.
```

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -146,22 +146,28 @@ void ExecuteWithAbi (string [] supportedAbis, string apkInputPath, string apkOut
 					var lastWriteInput = File.GetLastWriteTimeUtc (apkInputPath);
 					using (var packaged = new ZipArchiveEx (apkInputPath, FileMode.Open)) {
 						foreach (var entry in packaged.Archive) {
-							Log.LogDebugMessage ($"Deregistering item {entry.FullName}");
-							existingEntries.Remove (entry.FullName);
+							// NOTE: aapt2 is creating zip entries on Windows such as `assets\subfolder/asset2.txt`
+							var entryName = entry.FullName;
+							if (entryName.Contains ("\\")) {
```
brendanzagaeski, line 151: Very possibly a silly question, but on the small chance it might be useful make this code match f356220827c55a488ccf5732e6fd7e4e3333f55a even more closely, this could pass a `char` argument:

```cs
if (entryName.Contains ('\\')) {
```

So it would match:
https://github.com/xamarin/xamarin-android/blob/61fd95e52f2466888c7f6c6a2d0158599fef4723/src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs#L120

jonathanpeppers: Actually, I think the `"\\"` one should be used. The `ZipArchiveEx` is using:
```csharp
public static class Enumerable
{
    public static bool Contains<TSource>(this IEnumerable<TSource> source, TSource value);
````
I've seen this one is a bit slower than `string.Contains` ^^

brendanzagaeski: Ah, interesting. Sounds good!

Issue #4098, opened 2020-01-08
mono.android.runtime.JavaObject cannot be cast to byte[] when using List<byte[]>


BobbyShoe:
### Steps to Reproduce

1.   Declare a List<byte[]> in C#.   Pass this to a Java library and the mono.android.runtime.JavaObject cannot be cast to byte[] error is received when the object is accessed in Java.

Here is the sample call from the Java library.  The call to csdBuffer.get throws the exception.

public static void setCsdBuffers(MediaFormat format, List<byte[]> csdBuffers) {
    for (int i = 0; i < csdBuffers.size(); i++) {
      format.setByteBuffer("csd-" + i, ByteBuffer.wrap(csdBuffers.get(i)));
    }
  }

Seems like a similar issue to the java.lang.ClassCastException information mentioned on this page
https://docs.microsoft.com/en-us/xamarin/android/troubleshooting/troubleshooting#javalangclasscastexception-monoandroidruntimejavaobject-cannot-be-cast-to

However the work around suggested there states to use the similar Java collection for the inner collection in the generic.  However in this case a byte[] is the inner collection, so is there some comparable Java type that could be used?

### Expected Behavior
A List<byte[]> can be passed between C# and Java code.

### Actual Behavior
ClassCastException is thrown when the Java code attempts to access the byte array.

### Version Information
Xamarin 16.4.000.307

Running on Android 7.1

### Log File
 Caused by: java.lang.ClassCastException: mono.android.runtime.JavaObject cannot be cast to byte[]

jonpryor:
The bug is that `JavaConvert.LocalJniHandleConverters` doesn't special-case array types: https://github.com/xamarin/xamarin-android/blob/38a10a17eaf16b052e3c5ffbee502565acc688f5/src/Mono.Android/Java.Interop/JavaConvert.cs#L310-L356

Consider this Java method:

```java
public class Gxa4098 {
	public static int genericListMarshaling (List<byte[]> bytes) {
		int count = 0;
		for (byte[] value : bytes ) {
			count += value.length;
		}
		return count;
	}
}
```

Which results in the binding:

```
[Register ("genericListMarshaling", "(Ljava/util/List;)I", "")]
public static unsafe int GenericListMarshaling (global::System.Collections.Generic.IList<byte[]> p0)
{
	const string __id = "genericListMarshaling.(Ljava/util/List;)I";
[...]

PR #4788, opened 2020-06-08
[Mono.Android] Properly marshal generic lists of primitive arrays.

jpobst:


src/Mono.Android/Java.Interop/JavaConvert.cs
```diff
@@ -314,44 +314,74 @@ public static T FromJavaObject<T>(IJavaObject? value, out bool set)
...
 					return JNIEnv.ToLocalJniHandle (v);
 			} },
+			{ typeof (bool[]), value => {
+				return JNIEnv.NewArray ((bool[]) value);
+			} },
 			{ typeof (byte),   value => {
 				using (var v = new Java.Lang.Byte ((sbyte) (byte) value))
 					return JNIEnv.ToLocalJniHandle (v);
 			} },
+			{ typeof (byte[]), value => {
```
jonpryor, line 324: My concern with this approach is that I don't think it'll be "recursive" enough.  This handles `List<byte[]>`, but what happens when we hit a `List<byte[][]>`?  Or a `List<byte[][][]>`? Or a …?

jonpryor: …which is why in [my original comment](https://github.com/xamarin/xamarin-android/issues/4098#issuecomment-572818288) I said:

> but we might want to instead consider using following the "recursive type" approach used around `JNIEnv.NewArray(Array, Type)`.

jonpryor:
Would this patch to `JavaConvert.cs` pass the unit tests?

```diff
diff --git a/src/Mono.Android/Java.Interop/JavaConvert.cs b/src/Mono.Android/Java.Interop/JavaConvert.cs
index 910839f5..2872a08d 100644
--- a/src/Mono.Android/Java.Interop/JavaConvert.cs
+++ b/src/Mono.Android/Java.Interop/JavaConvert.cs
@@ -352,6 +352,9 @@ namespace Java.Interop {
 				using (var v = new Java.Lang.String (value.ToString ()!))
 					return JNIEnv.ToLocalJniHandle (v);
 			} },
+			{ typeof (Array), value => {
+				return JNIEnv.NewArray ((Array) value);
+			} },
 			{ typeof (Android.Runtime.JavaObject), value => {
 				return value == null ? IntPtr.Zero : JNIEnv.ToLocalJniHandle (new Android.Runtime.JavaObject (value));
 			} },
@@ -364,6 +367,8 @@ namespace Java.Interop {
 				return converter;
 			if (value != null && LocalJniHandleConverters.TryGetValue (value.GetType (), out converter))
 				return converter;
+			if (sourceType.IsArray)
+				return LocalJniHandleConverters [typeof (Array)];
 			return LocalJniHandleConverters [typeof (Android.Runtime.JavaObject)];
 		}

```

BenBtg:
Thanks for pushing through this fix. I can confirm it resolves the PointR binding issue.

PR #4792, opened 2020-06-09
[compression] Store assembly information with unique key

grendello:

d236af54 introduced assembly compression and one part of the process is
to gather information about the size of all the assemblies to be
packaged with the application, in order to generate native assembly code
which pre-allocates memory for them.  To do that, the native runtime must
be sure that the the assembly being currently decompressed will fit in
the pre-allocated memory area and thus it stores the original file size
both in the compressed assembly header and in the generated native
assembler source (compiled into `libxamarin-app.so`).  If the sizes
mismatch, we will see an error in the logcat similar to:

    F/monodroid-assembly(11348): Compressed assembly 'mscorlib.dll' is larger than when the application was built (expected at most 146432, got 2043392). Assemblies don't grow just like that!

In addition to the assembly size, the compressed assembly header
contains also an index into an array of assembly descriptors which is
placed in the generated native assembler source code.  The
index **must** point to the correct descriptor or bad things happen.
This is the reason for the particular error shown above, coming from a
test app (here)[0].

During build time, the information about assembly sizes is carried
between the involved tasks in an object registered with MSBuild using
its `BuildEngine4.RegisterTaskObject` API.  The API uses a key to access
the registered object which, in d236af54, was defined to be a constant
string, shared across all the builds.  However, it turns out
that *somehow* if two or more projects store the registered object with
the same key the end result is that the cache contains entries from more
than one object (how it happened is unclear, it should be "impossible")
which results in the descriptor indexes in compressed assembly header to
point to incorrect locations.

In order to fix the issue, this commit makes two changes:

 * Make sure the assembly info object is unregistered by the `BuildApk`.
   This ensures that "merging" of two dictionaries (however/wherever it
   happened) will not take place.
 * Make the key to the registered object unique by addition of the
   project GUID (from the `$(ProjectGuid)` MSBuild property).
   This ensures correct operation during builds which take advantage of
   MSBuild's parallel build feature.

[0]: https://github.com/xamarin/QualityAssurance/tree/master/Samples/android/Bug15162
    
### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -88,6 +88,9 @@ namespace Xamarin.Android.Tasks
 		//[Required]
 		public bool EnableCompression { get; set; }
 
+		[Required]
+		public string ProjectFullPath { get; set; }
+
 		[Output]
 		public ITaskItem[] OutputFiles { get; set; }
 
@@ -114,7 +117,7 @@ namespace Xamarin.Android.Tasks
 
 		List<string> existingEntries = new List<string> ();
 
-		void ExecuteWithAbi (string [] supportedAbis, string apkInputPath, string apkOutputPath)
+		void ExecuteWithAbi (string [] supportedAbis, string apkInputPath, string apkOutputPath, bool debug, bool compress, IDictionary<string, CompressedAssemblyInfo> compressedAssembliesInfo)
 		{
 			if (InterpreterEnabled) {
 				foreach (string abi in supportedAbis) {
@@ -180,7 +183,7 @@ namespace Xamarin.Android.Tasks
 				}
 
 				if (EmbedAssemblies && !BundleAssemblies)
-					AddAssemblies (apk);
+					AddAssemblies (apk, debug, compress, compressedAssembliesInfo);
 
 				AddRuntimeLibraries (apk, supportedAbis);
 				apk.Flush();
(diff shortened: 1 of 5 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2112,6 +2112,7 @@ because xbuild doesn't support framework reference assemblies.
       SupportedAbis="@(_BuildTargetAbis)"
       Debug="$(AndroidIncludeDebugSymbols)"
       EnableCompression="$(AndroidEnableAssemblyCompression)"
+      ProjectGuid="$(ProjectGuid)"
```
jonathanpeppers, line 2115: It looks like when I build a .NET 5 project, `$(ProjectGuid)` is blank?

```
dotnet build foo.csproj -bl
```
image

Maybe this has to use [`$(MSBuildProjectFile)`](https://docs.microsoft.com/visualstudio/msbuild/msbuild-reserved-and-well-known-properties)?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2302,7 +2303,8 @@ because xbuild doesn't support framework reference assemblies.
     LibraryProjectJars="@(ExtractedJarImports)"
     TlsProvider="$(AndroidTlsProvider)"
     UncompressedFileExtensions="$(AndroidStoreUncompressedFileExtensions)"
-	InterpreterEnabled="$(_AndroidUseInterpreter)">
+	InterpreterEnabled="$(_AndroidUseInterpreter)"
+    ProjectFullPath="$(MSBuildProjectFullPath)">
```
jonathanpeppers, line 2307: Fast Deployment has a different call for `<BuildApk/>` so this change will be needed here, too: 

https://github.com/xamarin/monodroid/blob/e795bcef375bc77da0bbe995fc9e57560ed5ddc0/tools/msbuild/Xamarin.Android.Common.Debugging.targets#L271-L297

You might also add the missing `InterpreterEnabled` value, too.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -300,7 +303,11 @@ private void AddAssemblies (ZipArchiveEx apk)
 			IDictionary<string, CompressedAssemblyInfo> compressedAssembliesInfo = null;
 
 			if (compress) {
-				compressedAssembliesInfo = BuildEngine4.GetRegisteredTaskObject (CompressedAssemblyInfo.CompressedAssembliesInfoKey, RegisteredTaskObjectLifetime.Build) as IDictionary<string, CompressedAssemblyInfo>;
+				string key = $"{CompressedAssemblyInfo.CompressedAssembliesInfoKey}:{ProjectFullPath}";
```
jonpryor, line 306: Given that this string is repeated in two places, could we move it to a method e.g. `MonoAndroidHelper.GetCompressedAssemblyInfoKey(ProjectFullPath)`?

jonathanpeppers: Or a `static` method on `CompressedAssemblyInfo` such as `CompressedAssemblyInfo.GetKey (ProjectFullPath)`. Then `CompressedAssembliesInfoKey` could also be `private`.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateCompressedAssembliesNativeSourceFiles.cs
```diff
@@ -61,7 +64,9 @@ void GenerateCompressedAssemblySources ()
 				kvp.Value.DescriptorIndex = index++;
 			}
 
-			BuildEngine4.RegisterTaskObject (CompressedAssemblyInfo.CompressedAssembliesInfoKey, assemblies, RegisteredTaskObjectLifetime.Build, false);
+			string key =CompressedAssemblyInfo.GetKey (ProjectFullPath);
```
jonathanpeppers, line 67: You might just fix when updating this to target monodroid/master:
```suggestion
			string key = CompressedAssemblyInfo.GetKey (ProjectFullPath);
```

PR #4793, opened 2020-06-09
VS Android App template sample to measure apk size

radekdoulik:
Add clean project created in VS using Android App template. The main idea here is to have clean project without any changes, so that it can be easily updated with newer versions when needed. This can be done by replacing the `samples/VSAndroidApp` directory with newer project generated in VS from the Android App template.

The project itself is located in `samples` directory and is used through the proxy project `VSAndroidAppProxy.csproj`, which makes  it possible to build it within XA repo.

It is also run during apk instrumentation step on Azure and the apk size is compared using `apkdiff` to the reference `com.companyname.vsandroidapp-Signed-Release.apkdesc` file. The test result with the usual error is generated when the regression test fails.


pjcollins:
@radekdoulik I'm starting to put together some [dashboards for app size and build time performance][0], I wonder if we may be overlapping a bit here? My work will help highlight trends but not provide a failing test result on PRs, so perhaps the partial duplication is fine if we want this kind of check on our PRs builds.

[0]: https://msit.powerbi.com/groups/4d661d85-12d3-4609-a5fb-c61cae72824b/dashboards/fe3c4102-63af-4e21-b84c-e8986f2ce017

radekdoulik:
@pjcollins this PR just extends the regression testing we already have from last fall, when I added apk size regression testing to our PR builds. So far we were using mostly XF test and this adds non-XF app so that we cover broader area.

Some time ago I improved the regression testing with usage of `apkdiff` tool (https://www.nuget.org/packages/apkdiff/). Maybe it can be useful for you as well to collect data for the dashboard?

samples/VSAndroidApp/Resources/AboutResources.txt
```diff
@@ -0,0 +1,44 @@
+Images, layout descriptions, binary blobs and string dictionaries can be included
```
pjcollins, line 1: I think we can exclude this file and the `AboutAssets.txt` file to reduce the additions a bit.

radekdoulik: I would like to keep it just clean template app, without any changes, so I think it is ok to include these.

pjcollins (approved):

PR #4800, opened 2020-06-10
Bump to Mono.Posix.NETStandard 5.20.1-preview

jonpryor:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1138785/

Includes `Mono.Posix.NETStandard.pdb` in the installer artifacts.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
@@ -381,13 +381,13 @@
     <_ExtraPackageTarget Include="$(OutputPath)\Xamarin.Build.AsyncTask.pdb" />
     <_ExtraPackageSource Include="$(PkgMono_Posix_NETStandard)\ref\$(TargetFrameworkNETStandard)\Mono.Posix.NETStandard.pdb" />
     <_ExtraPackageTarget Include="$(OutputPath)\Mono.Posix.NETStandard.pdb" />
-
-    <_ExtraPackageInput Include="$(PkgMono_Posix_NETStandard)\ref\$(TargetFrameworkNETStandard)\Mono.Posix.NETStandard.dll" />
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
```diff
@@ -374,13 +374,25 @@
...
+
   <!-- Copy package ref symbols for our installers. '$(Pkg*)' props are set during NuGet restore when GeneratePathProperty="true" -->
   <Target Name="_CopyExtraPackageContent"
-      Inputs="$(PkgXamarin_LibZipSharp)\lib\$(TargetFrameworkNETStandard)\libZipSharp.pdb;$(PkgXamarin_Build_AsyncTask)\lib\$(TargetFrameworkNETStandard)\Xamarin.Build.AsyncTask.pdb;$(PkgMono_Posix_NETStandard)\ref\$(TargetFrameworkNETStandard)\Mono.Posix.NETStandard.dll"
-      Outputs="$(OutputPath)\libZipSharp.pdb;$(OutputPath)\Xamarin.Build.AsyncTask.pdb" >
+      Inputs="@(_ExtraPackageSource);@(_ExtraPackageInput)"
+      Outputs="@(_ExtraPackageTarget)" >
     <Copy
-        SourceFiles="$(PkgXamarin_LibZipSharp)\lib\$(TargetFrameworkNETStandard)\libZipSharp.pdb;$(PkgXamarin_Build_AsyncTask)\lib\$(TargetFrameworkNETStandard)\Xamarin.Build.AsyncTask.pdb;$(PkgMono_Posix_NETStandard)\ref\$(TargetFrameworkNETStandard)\Mono.Posix.NETStandard.dll"
+        SourceFiles="@(_ExtraPackageSource)"
```
pjcollins, line 393: I think we also need to copy `@(_ExtraPackageInput)` here. It may also make sense to move the

> '$(Pkg*)' props are set ...

comment above the new item group where those properties are used.

jonpryor: Sorry, misread the original file; `Mono.Posix.NETStandard.dll` should be in `@(_ExtraPackageSource)`. :-/

Issue #4804, opened 2020-06-11
Wrong AndroidManifest.xml packaged to the APK while using binding library


olegtyshcneko:
### Steps to Reproduce

1. Create blank android project
2. Create binding library, add jar with androidmanifest.xml in it (in our case it is gigya android sdk aar file)
3. Build project
4. Try to deploy to the device the apk

### Expected Behavior
Deployment of the APK to the device is the successful

### Actual Behavior
Error happens 
Failed to parse APK info: failed to parse AndroidManifest.xml, error: %!s(<nil>)
deploy failed, error: failed to get apk infos, output: W/ResourceType( 5266): Bad XML block: header size 28024 or total size 1702240364 is larger than data size 570
ERROR: AndroidManifest.xml is corrupt, error: exit status 1

### Additional info
When APK is opened - it contains wrong AndroidManifest.xml file, it copies it from gigya binding aar in our case. 
When project is built with /v:d option we can see following inside
Adding AndroidManifest.xml as the archive file is out of date. 
This error doesn't happen on the older version of xamarin android build tools.
Presumably it happened after commit 1c4727c(Dean Ellis) Allow BuildApk incrementally, because before that there was a check in BuildTask.cs on line 190: 
`if (apk.Archive.Any (e=> e.FullName == path))`
and now there is no such check. 

### Version Information

=== Visual Studio Community 2019 for Mac (Preview) ===

Version 8.7 Preview (8.7 build 1077)
Installation UUID: d5f9f469-b751-4def-b628-74d658efb1bd
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 6.18.0.23 (d16-6 / 088c73638)

	Package version: 612000074

=== Mono Framework MDK ===

Runtime:
	Mono 6.12.0.74 (2020-02/e9d3af508e4) (64-bit)
	Package version: 612000074

=== Roslyn (Language Service) ===

3.7.0-2.20264.2+0f7f7354f2624cc089dca14b3b1c32ee46ce1349

=== NuGet ===

Version: 5.6.0.6591

=== .NET Core SDK ===

SDK: /usr/local/share/dotnet/sdk/3.1.300/Sdks
SDK Versions:
	3.1.300
	3.1.202
	3.1.200
	3.1.102
	3.1.101
	3.1.100
	3.1.100-preview3-014645
	3.0.100
	3.0.100-rc1-014190
	3.0.100-preview9-014004
	3.0.100-preview5-011568
[...]

PR #4812, opened 2020-06-12
[Xamarin.Android.Build.Tasks] Wrong AndroidManifest.xml packaged to the APK while using binding library

dellis1972:
Fixes https://github.com/xamarin/xamarin-android/issues/4804

We were overwriting the AndroidManifest.xml file in the apk with
ones from Support Libraries. This manifests itself with the following error

```
Failed to parse APK info: failed to parse AndroidManifest.xml, error: %!s()
deploy failed, error: failed to get apk infos, output: W/ResourceType( 5266): Bad XML block: 
```

This commit fixes this issue by ignoring files from .jar files
which already exist in the apk. This stops things from being
overwritten. A test has also been added.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -28,7 +28,7 @@ namespace Xamarin.Android.Tasks
 
 		[Required]
 		public string ApkInputPath { get; set; }
-		
+
 		[Required]
 		public string ApkOutputPath { get; set; }
 
@@ -101,7 +101,7 @@ namespace Xamarin.Android.Tasks
 		}
 
 		SequencePointsMode sequencePointsMode = SequencePointsMode.None;
-		
+
 		public ITaskItem[] LibraryProjectJars { get; set; }
 		string [] uncompressedFileExtensions;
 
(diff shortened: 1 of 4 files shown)


dellis1972:
@jonathanpeppers waiting on a good repo from the custom. The steps they provided did not work

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -241,6 +241,10 @@ void ExecuteWithAbi (string [] supportedAbis, string apkInputPath, string apkOut
 								Log.LogDebugMessage ($"Skipping {path} as the archive file is up to date.");
 								continue;
 							}
+							if (apk.Archive.Any (e => e.FullName == path)) {
+								Log.LogMessage ("Warning: failed to add jar entry {0} from {1}: the same file already exists in the apk", name, Path.GetFileName (jarFile));
```
jonathanpeppers, line 245: Should this be a `LogDebugMessage` and not have the "warning" text in the message?

A developer probably can't (or won't) fix it if it is coming from a java library.

dellis1972: good point

Issue #4805, opened 2020-06-11
ResX Localization Broken


divil5000:
You seem to have broken ResX localization in a recent version of Xamarin.Android. I found issue #4664 and downloaded the latest version of VS (I'm on 16.6.2) but the issue persists. This has broken our app and is now holding up deployment to tens of thousands of users.

I reproduced the issue with a blank app straight from one of your templates, which I will attach.

It has a ResX file with a single string resource, which has been localized into German. Pressing the floating button in the demo app will pop up a toast that should contain the default English text then the German text, but instead the default English text appears twice because for some reason it is not finding the German resources.

NOTE: I stripped the bin and obj folders out of the zip file because it was 54mb with them.
[app1.zip](https://github.com/xamarin/xamarin-android/files/4765623/app1.zip)

### Steps to Reproduce

1. Run the sample app
2. Press the floating button
3. Observe in the toast that the English text is displayed twice

### Expected Behavior

Should display Some English Text, German!

### Actual Behavior

Displays Some English Text, Some English Text

### Version Information

Microsoft Visual Studio Professional 2019
Version 16.6.2
VisualStudio.16.Release/16.6.2+30204.135
Microsoft .NET Framework
Version 4.8.03752

Installed Version: Professional

ADL Tools Service Provider   1.0
This package contains services used by Data Lake tools

ASA Service Provider   1.0

ASP.NET and Web Tools 2019   16.6.948.25768
ASP.NET and Web Tools 2019

ASP.NET Core Razor Language Services   16.1.0.2020603+b3ac44798c16fff5b95dbcfe62dea84aa9a1bd72
Provides languages services for ASP.NET Core Razor.

ASP.NET Web Frameworks and Tools 2019   16.6.948.25768
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   16.6.948.25768
Azure App Service Tools v3.0.0

Azure Data Lake Node   1.0
This package contains the Data Lake integration nodes for Server Explorer.

Azure Data Lake Tools for Visual Studio   2.4.6000.0
Microsoft Azure Data Lake Tools for Visual Studio

Azure Functions and Web Jobs Tools   16.6.948.25768
Azure Functions and Web Jobs Tools

Azure Stream Analytics Tools for Visual Studio   2.4.6000.0
Microsoft Azure Stream Analytics Tools for Visual Studio

C# Tools   3.6.0-4.20251.5+910223b64f108fcf039012e0849befb46ace6e66
[...]

PR #4813, opened 2020-06-12
[Xamarin.Android.Build.Tasks] fix IDE builds missing satellite assemblies in apps

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/4805
Context: https://github.com/xamarin/xamarin-android/issues/4664

There are still reports of `.resx` files not working inside IDEs...
65a917a9 was the original cause of this problem.

Unfortunately, the original problem in #4664 still occurs if the
`.resx` files are in the Xamarin.Android application project. The
original fix in d298fdaa, only solved the problem for missing
satellite assemblies in referenced projects.

Two missing targets need to run for `SignAndroidPackage` in IDE builds:

    PrepareResources;
    CreateSatelliteAssemblies;

These targets fill out the
`@(IntermediateSatelliteAssembliesWithTargetPath)` item group, that
contains all satellite assemblies for the current project.

In IDE builds, this item group was empty and `<BuildApk/>` was not
adding satellite assemblies to the `.apk` file at all.

I was able to reproduce this problem pretty easily with a test. I
reworked the `MissingSatelliteAssembly` test to have a version for
libraries and apps.

I also cleaned the test up a bit, in general, using less hardcoded
strings.

### DIFF
--- a/Documentation/release-notes/4805.md
+++ b/Documentation/release-notes/4805.md
@@ -1,0 +1,9 @@
+#### Application behavior on device and emulator
+
+  - [GitHub 4805](https://github.com/xamarin/xamarin-android/issues/4805):
+    Starting in Xamarin.Android 10.3, localized resources from _.resx_
+    files in Xamarin.Android application projects were no longer
+    deployed when building and deploying from within Visual Studio to
+    an attached device or emulator.  (In contrast, clean builds
+    started via the **Build > Archive** command or on the command line
+    worked as expected.)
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
@brendanzagaeski the main issue with the old approach is we were calling `<ResolveAssemblyReference/>` directly. If they changed parameters on that task, it could break. It's basically relying on a private API. I'm not actually sure if it would have worked in .NET 5 or not.

The case of having 0 `.resx` files, seems like it would spend hardly any time in `PrepareResources` / `CreateSatelliteAssemblies`, too. When I was an app developer we always used native localization instead of `.resx`, but I don't know about the usage statistics there.

brendanzagaeski:
Right. And the time savings for the old approach, even in my original test scenario, which was pretty much the best case for it, was only on the order of 10 milliseconds. I only would have been concerned if the old approach was noticeably faster in the majority of cases, but that wasn't the case.

PR #4816, opened 2020-06-12
[Xamarin.Android.Build.Tasks] @(LinkDescription) should be in Inputs for the linker

jonathanpeppers:
Fixes: https://developercommunity.visualstudio.com/content/problem/1061307/changes-not-applied-on-build-after-linker-configur.html

Changes to a `@(LinkDescription)` file did not trigger the
`_LinkAssembliesShrink` MSBuild target to run again.

The only way to see changes from this file was to `Rebuild` every time.

:eyes:

Adding a test for this scenario is also nice, as it didn't seem like
we had an MSBuild test verifying that `@(LinkDescription)` works in
general.

### DIFF
--- a/Documentation/release-notes/incremental-linkdescription.md
+++ b/Documentation/release-notes/incremental-linkdescription.md
@@ -1,0 +1,7 @@
+#### Application and library build and deployment
+
+  * [Developer Community 1061307][0]: Fixes an issue where modifying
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
This will conflict with https://github.com/xamarin/xamarin-android/pull/4767, but I can fix the conflict after one of these is merged.

dellis1972:
Test failures look un related.

PR #4820, opened 2020-06-15
[build] enable One .NET MSBuild tests

jonathanpeppers:
This will enable all MSBuild tests in
`Xamarin.Android.Build.Tests.dll` except for the new categories:
`DotNetIgnore`, `AOT`, `FSharp`, `LibraryProjectZip`, `MkBundle`,
`MonoSymbolicate`, `PackagesConfig`, and `StaticProject`.

Reasoning:

* `AOT` - need Mono AOT support in .NET 5.
* `FSharp` - [Xamarin.Android.FSharp.ResourceProvider][0] support
  needed on .NET 5.
* `LibraryProjectZip` - will need future support in Xamarin.Android.
* `MkBundle` - may not ever come to .NET 5.
* `MonoSymbolicate` - need symbolication support in .NET 5.
* `PackagesConfig` - n/a for .NET 5.
* `StaticProject` a few tests that were migrated to
  `Xamarin.Android.Build.Tests` in d447aa66 are ignored for now.

This also reworks our build pipeline to run the MSBuild tests across 12
CI machines:

* Windows - Node 1 - Legacy
* Windows - Node 2 - Legacy
* Windows - Node 3 - Legacy
* Windows - Node 1 - One .NET
* Windows - Node 2 - One .NET
* Windows - Node 3 - One .NET
* macOS - Node 1 - Legacy
* macOS - Node 2 - Legacy
* macOS - Node 3 - Legacy
* macOS - Node 1 - One .NET
* macOS - Node 2 - One .NET
* macOS - Node 3 - One .NET

Since this is a lot of machines:

1. I moved all the `One .NET` tests to their own phase.
2. The phase only runs when `RunAllTests=true`. This means the `One .NET` tests will run on master, release branches, and Mono bumps. You can manually queue a build to enable `RunAllTests` on a PR.

To be completed in another PR, there are still more test assemblies
that need to be run under a `dotnet` context:

* `Xamarin.Android.Build.Tests.Commercial.dll`
* `MSBuildDeviceIntegration.dll`

[0]: https://github.com/xamarin/Xamarin.Android.FSharp.ResourceProvider

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -63,6 +63,7 @@ variables:
   # - This is a non-fork branch with name containing "mono-" (for Mono bumps)
   IsMonoBranch: $[and(eq(variables['XA.Commercial.Build'], 'true'), ne(variables['System.PullRequest.IsFork'], 'True'), or(contains(variables['Build.SourceBranchName'], 'mono-'), contains(variables['System.PullRequest.SourceBranch'], 'mono-')))]
   RunAllTests: $[or(eq(variables['XA.RunAllTests'], true), eq(variables['IsMonoBranch'], true))]
+  DotNetNUnitCategories: '&& cat != DotNetIgnore && cat != AOT && cat != FSharp && cat != LibraryProjectZip && cat != MkBundle && cat != MonoSymbolicate && cat != PackagesConfig && cat != StaticProject'
 
 # Stage and Job "display names" are shortened because they are combined to form the name of the corresponding GitHub check.
 stages:
@@ -796,31 +797,45 @@ stages:
 
     - template: yaml-templates/fail-on-issue.yaml
 
-  # Xamarin.Android (Test MSBuild - macOS)
+  # Xamarin.Android (Test MSBuild - macOS - Legacy)
   - template: yaml-templates\run-msbuild-mac-tests.yaml
     parameters:
       node_id: 1
+      job_name: mac_msbuild_tests_1
+      job_suffix: Legacy
 
   - template: yaml-templates\run-msbuild-mac-tests.yaml
     parameters:
       node_id: 2
+      job_name: mac_msbuild_tests_2
+      job_suffix: Legacy
+      run_extra_tests: true
 
   - template: yaml-templates\run-msbuild-mac-tests.yaml
     parameters:
       node_id: 3
+      job_name: mac_msbuild_tests_3
+      job_suffix: Legacy
 
-  # Xamarin.Android (Test MSBuild - Windows)
+  # Xamarin.Android (Test MSBuild - Windows - Legacy)
   - template: yaml-templates\run-msbuild-win-tests.yaml
     parameters:
       node_id: 1
+      job_name: win_msbuild_tests_1
+      job_suffix: Legacy
 
   - template: yaml-templates\run-msbuild-win-tests.yaml
     parameters:
       node_id: 2
+      job_name: win_msbuild_tests_2
+      job_suffix: Legacy
+      run_extra_tests: true
 
   - template: yaml-templates\run-msbuild-win-tests.yaml
     parameters:
       node_id: 3
+      job_name: win_msbuild_tests_3
+      job_suffix: Legacy
 
   # Check - "Xamarin.Android (Test MSBuild With Emulator - macOS)"
   - job: mac_msbuilddevice_tests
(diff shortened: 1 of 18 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -796,31 +797,95 @@ stages:
...
+      job_suffix: Legacy
+      run_extra_tests: true
+
+  - template: yaml-templates\run-msbuild-win-tests.yaml
+    parameters:
+      node_id: 3
+      job_name: win_msbuild_tests_3
+      job_suffix: Legacy
+
+  # Xamarin.Android (Test MSBuild - Windows - One .NET)
```
pjcollins, line 865: I'm starting to worry about fanning out to too many machines on each PR build, especially in the case of Windows where we only have 9 instances available. I think we'll need to address execution scope for `Xamarin.Android.Build.Tests` on PR builds in the somewhat near future. I could go either way here, but we may want to consider only running these new One .NET msbuild test jobs against main/release branches for now as part of this PR?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/MakeBundleTests.cs
```diff
@@ -17,6 +17,7 @@
 namespace Xamarin.Android.Build.Tests
 {
 	[Category ("Node-2")]
+	[Category ("StaticProject")] // TODO: enable for .NET 5
```
pjcollins, line 20: I believe this test is only validating `BundleAssemblies` behavior, and could use the `MkBundle` category instead?

jonathanpeppers: Yeah, I should probably put both categories and can drop the `//TODO:` comment.

`StaticProject` was a category of tests that have static `.csproj` files in our repo. I haven't figured out how these should run under .NET 5 yet.

build-tools/automation/azure-pipelines.yaml
jonathanpeppers: I'm experimenting with adding `condition: eq(variables['RunAllTests'], true)` to all the One .NET tests.

I'll need to queue a build manually and set `RunAllTests` at queue time to verify that works.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/MakeBundleTests.cs
pjcollins: Sounds good, and hopefully this particular instance won't need to be updated if MkBundle remains a legacy feature.

build-tools/automation/azure-pipelines.yaml
jonathanpeppers: I think the `condition` is correct now. Here is the manual build: https://build.azdo.io/3948573

jpobst:
My preference would be to move these to a new `Test - One .NET` stage like `Regression Tests` is.  I think this will help keep the graph more readable, particularly if we're skipping them by default on PR runs:

image

I would put the stage between `Test` and `Regression Tests`.

jonathanpeppers:
I moved it around so it will look like this now:

image

I'll need to manually queue one with `RunAllTests=true` to make sure that works.

jonathanpeppers:
The PR build here looks OK so far, started a `RunAllTests=true` build here: https://build.azdo.io/3952696

jonpryor:
```markdown
*Invert* the MSBuild unit tests & One .NET Relationship: previously,
we would selectively *enable* our MSBuild tests to run under .NET 5+.
Now, run *all* tests under .NET 5+, *except* for the categories:

  * `AOT` and `MonoSymbolicate`: requires Mono AOT support in .NET 5
    and packaging of required tooling; see also, but not limited to,
    https://github.com/dotnet/runtime/issues/35852

  * `FSharp`: [Xamarin.Android.FSharp.ResourceProvider][0] support
    is needed to run on .NET 5.

  * `LibraryProjectZip`, `StaticProject`: Require future support
    within Xamarin.Android, but don't need to be done *before*
    inverting the unit test regime.

  * `MkBundle`: `mkbundle.exe` might not make it to .NET 5 -- see
    also dotnet/runtime#35852 -- and it may be moot with d236af54.

  * `PackagesConfig`: .NET 5+ won't support `packages.config` files.

Also rework our build pipeline to run the MSBuild tests across 12
CI machines:

  * Windows - Node 1 - Legacy
  * Windows - Node 2 - Legacy
  * Windows - Node 3 - Legacy
  * Windows - Node 1 - One .NET
  * Windows - Node 2 - One .NET
  * Windows - Node 3 - One .NET
  * macOS - Node 1 - Legacy
  * macOS - Node 2 - Legacy
  * macOS - Node 3 - Legacy
  * macOS - Node 1 - One .NET
  * macOS - Node 2 - One .NET
  * macOS - Node 3 - One .NET

Since this is a lot of machines:

 1. I moved all the `One .NET` tests to their own phase.
 2. The phase only runs when `RunAllTests=true`.  This means the
    `One .NET` tests will run on master, release branches, and
    Mono bumps.

    You can manually queue a build to enable `RunAllTests` on a PR.

To be completed in another PR, there are still more test assemblies
that need to be run under a `dotnet` context:

  * `Xamarin.Android.Build.Tests.Commercial.dll`
  * `MSBuildDeviceIntegration.dll`

[0]: https://github.com/xamarin/Xamarin.Android.FSharp.ResourceProvider
```

PR #4825, opened 2020-06-15
[Xamarin.Android.Build.Tasks] use MSBuild item groups for ProGuard rules

jonathanpeppers:
Context: https://github.com/xamarin/monodroid/commit/e82633e9eaf36b2ac3530a681c86d6926bf7f8c8

When using `dotnet build`, usage of R8 or ProGuard could encounter
build warnings on macOS such as:

    warning XA4304: ProGuard configuration file '/Users/runner/Library/Android/sdk//tools\proguard\proguard-android.txt' was not found.

Reviewing how this worked, the `$(ProguardConfigFiles)` MSBuild
property:

    <ProguardConfigFiles Condition="'$(ProguardConfigFiles)' == '' And '$(AndroidLinkTool)' != ''">
        {sdk.dir}tools\proguard\proguard-android.txt;
        {intermediate.common.xamarin};
        {intermediate.references};
        {intermediate.application};
        @(ProguardConfiguration);
    </ProguardConfigFiles>

Was using string replacements such as:

    var configs = ProguardConfigurationFiles
        .Replace ("{sdk.dir}", AndroidSdkDirectory + Path.DirectorySeparatorChar)
        .Replace ("{intermediate.common.xamarin}", ProguardCommonXamarinConfiguration)
        .Replace ("{intermediate.references}", ProguardGeneratedReferenceConfiguration)
        .Replace ("{intermediate.application}", ProguardGeneratedApplicationConfiguration)
        .Replace ("{project}", string.Empty) // current directory anyways.
        .Split (';')
        .Select (s => s.Trim ())
        .Where (s => !string.IsNullOrWhiteSpace (s));

There is some behavior difference, where this path worked running
under Mono:

    /Users/runner/Library/Android/sdk//tools\proguard\proguard-android.txt

But why are we even replacing strings like this?

Reading the ProGuard documentation, `{sdk.dir}` and `{project}` do not
seem to be a placeholder we need to support. [One mention][0] of
`${sdk.dir}` (with a dollar sign) which would produce:

    $/Users/runner/Library/Android/sdk//tools\proguard\proguard-android.txt

It leads me to believe Xamarin.Android invented its own string
templating for this instead of embracing MSBuild. It was a long time
ago (~2014), so xbuild was also likely a factor then.

To clean things up and fix .NET 5 at the same time:

1. Use `@(_ProguardConfiguration)` instead of `$(ProguardConfigFiles)`.
2. List the files as you would in MSBuild item groups normally.
3. Use `string[]` as the value type in the `<Proguard/>` and `<R8/>`
   MSBuild tasks.
4. Continue to support the `$(ProguardConfigFiles)` property for
   backwards compatibility. It can remain an escape hatch if someone
   wants to replace *every* file.

Other changes:

* Updated the `BuildHasNoWarnings` test, so it will run under `dotnet
  build`. Don't set `$(MonoSymbolArchive)` to `true` for `dotnet`.
* The `AndroidSdkDirectory` property is no longer needed in the
  `<Proguard/>` and `<R8/>` MSBuild tasks.
* Removed System.Linq usage.

[0]: https://www.guardsquare.com/en/products/proguard/manual/examples


jonpryor:
@jonathanpeppers: Please add a release-notes blurb.

dellis1972:
Looks good. Do we need to document how users upgrade to the new system? Or will it just happen automagically?

jonathanpeppers:
@dellis1972 I couldn't find any examples of using `$(ProguardConfigFiles)`, it is undocumented:

https://github.com/search?l=&p=1&q=%22%3CProguardConfigFiles%3E%22+extension%3A.csproj+extension%3A.targets+extension%3A.props&ref=advsearch&type=Code

It will still _work_ to use this property, just not the string replacement part.

I think everyone should just use `@(ProguardConfiguration)` as the docs say:

https://docs.microsoft.com/xamarin/android/deploy-test/release-prep/proguard#customizing-proguard

PR #4828, opened 2020-06-16
[Xamarin.ProjectTools] use Directory.Build.props for $(Configuration)

jonathanpeppers:
When running `dotnet build` and setting `IsRelease=true` in our
MSBuild tests, the current MSBuild tests generate `.csproj` files such
as:

    <Project Sdk="Microsoft.Android.Sdk">
      <PropertyGroup>
        <Configuration>Release</Configuration>

The problem with this is how the ordering of imports work in SDK-style
projects:

1. Any `Directory.Build.props` are imported.
2. The SDK targets are imported, such as:

https://github.com/dotnet/sdk/blob/b7c7f33aa44b23dc1f984a9a1162ee2d57439897/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.props#L39-L45

3. The `.csproj` is imported.

The result, is even though `Configuration=Release` in this project,
some defaults are set such as `Optimize=False` and `DebugSymbols=True`.

This works for legacy projects because
`Xamarin.Android.CSharp.targets` is imported at the bottom of
`.csproj` files. `Directory.Build.props` is imported after the user's
`.csproj` file.

To fix this:

* If `Builder.UseDotNet` is `true`, set `$(Configuration)` in a
  `Directory.Build.props`.
* "legacy" projects continue the old way.

To make everything work, I had to make changes to
`Xamarin.ProjectTools`. `IsRelease` is now just a plain C# property
instead of calling `SetProperty("Configuration", value)`.

Once `Release` builds were working, a couple tests were really broken
under `dotnet build`:

* `_RemoveRegisterAttribute` was running on every `Release` build. I
  had to make a new target to touch `$(_AndroidLinkFlag)` when
  `ILLink` runs. Due to the way the nested `<MSBuild/>` can run
  `ILLink` multiple times, I think we have to do it this way.
* Several tests needed the workaround for:
  https://github.com/mono/linker/issues/1139
  I moved this to an extension method.
* I needed to update the version of Xamarin.Forms in
  `XamarinFormsAndroidApplicationProject`.
* I needed one exception for `CheckTimestamps(False)`, as `ILLink`
  will output some assemblies with older timestamps. This is probably
  not a problem at all for `Release` builds.

Other changes:

* We don't need a `Condition` when setting `$(Configuration)`, as it
  is the very first property evaluated.
* Fixed how `Directory.Build.props` and `Directory.Build.targets` are
  handled when saving `.csproj` files.
* Added one test to the `dotnet` category that verifies a `Release`
  build works along with a `<ProjectReference/>`.
* Fixed an NRE that could occur when using `<Import/>`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Linker.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Linker.targets
@@ -47,4 +47,13 @@ This file contains the .NET 5-specific targets to customize ILLink
       DependsOnTargets="_ResolveAssemblies;_CreatePackageWorkspace;$(_BeforeLinkAssemblies);_GenerateJniMarshalMethods;_LinkAssembliesNoShrink"
   />
 
+  <Target Name="_TouchAndroidLinkFlag"
+      AfterTargets="ILLink"
+      Condition=" '$(PublishTrimmed)' == 'true' and Exists('$(_LinkSemaphore)') "
+      Inputs="$(_LinkSemaphore)"
+      Outputs="$(_AndroidLinkFlag)">
+    <!-- This file is an input for _RemoveRegisterAttribute -->
+    <Touch Files="$(_AndroidLinkFlag)" AlwaysCreate="true" />
+  </Target>
+
 </Project>
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/KnownTargets.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/KnownTargets.cs
@@ -6,6 +6,7 @@ namespace Xamarin.ProjectTools
 	{
 		public const string LinkAssembliesNoShrink = "_LinkAssembliesNoShrink";
 
-		public static string LinkAssembliesShrink => Builder.UseDotNet ? "ILLink" : "_LinkAssembliesShrink";
+		// _RunILLink is found at: https://github.com/dotnet/sdk/blob/1ed51bc8f9cb06760c5b2f26798ee0278bd75f54/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L69-L72
+		public static string LinkAssembliesShrink => Builder.UseDotNet ? "_RunILLink" : "_LinkAssembliesShrink";
 	}
 }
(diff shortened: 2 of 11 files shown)


jonathanpeppers:
This has brought up some real .NET 5 failures now that they are truly being built in release mode:

image

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Linker.targets
```diff
@@ -47,4 +47,13 @@ This file contains the .NET 5-specific targets to customize ILLink
...
   />
 
+  <Target Name="_TouchAndroidLinkFlag"
+      AfterTargets="ILLink"
+      Condition=" '$(PublishTrimmed)' == 'true' and Exists('$(_LinkSemaphore)') "
+      Inputs="$(_LinkSemaphore)"
+      Outputs="$(_AndroidLinkFlag)">
+    <!-- This file is an input for _RemoveRegisterAttribute -->
+    <Touch Files="$(_AndroidLinkFlag)" AlwaysCreate="true" />
+  </Target>
```
jonathanpeppers, line 57: I couldn't find a way other than to use `AfterTargets="ILLink"`: 

https://github.com/dotnet/sdk/blob/e448c2bd3482d0dbd9bab06de589e85db5f47e56/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L34-L37

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/KnownTargets.cs
```diff
@@ -6,6 +6,7 @@ public static class KnownTargets
 	{
 		public const string LinkAssembliesNoShrink = "_LinkAssembliesNoShrink";
 
-		public static string LinkAssembliesShrink => Builder.UseDotNet ? "ILLink" : "_LinkAssembliesShrink";
+		// _RunILLink is found at: https://github.com/dotnet/sdk/blob/1ed51bc8f9cb06760c5b2f26798ee0278bd75f54/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L69-L72
+		public static string LinkAssembliesShrink => Builder.UseDotNet ? "_RunILLink" : "_LinkAssembliesShrink";
```
jonathanpeppers, line 10: `_RunILLink` can actually be skipped, while `ILLink` never skips it just depends on `_RunILLink`.

PR #4829, opened 2020-06-17
[Xamarin.Android.Build.Tasks] Remove Fast Dev of AndroidResources

dellis1972:
Our current enhanced fast deployment system makes use of grey listed API's which are going to be removed in a future version of android. This PR and its companion in monodroid removes support for fast deploying the `packaged_resosurces` file to the override directory. It also removes the resource patching support from the debug runtime.


tests/MSBuildDeviceIntegration/Tests/MonoAndroidExportTest.cs
```diff
@@ -135,8 +135,9 @@ protected override void OnCreate (Bundle bundle)
 </manifest>";
 				Assert.True (b.Install (proj), "Project should have installed.");
 				ClearAdbLogcat ();
+				b.BuildLogFile = "run.log";
 				if (CommercialBuildAvailable)
-					Assert.True (b.RunTarget (proj, "_Run"), "Project should have run.");
+					Assert.True (b.RunTarget (proj, "_Run", doNotCleanupOnUpdate: true), "Project should have run.");
```
jonathanpeppers, line 140: Should we redo code like this so they use `StartAndroidActivity` instead?

https://github.com/xamarin/xamarin-android/blob/77464dc92ee53a5c0ef62b1df78d4e2bb4af5992/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Application.targets#L15

dellis1972: Well the StartAndroidActivity will work in some cases, but it does not call the internal "_Run" target which sets up the debugger so we can't use it everywhere.

jonpryor:
This PR should probably be a draft, and cannot be merged until this PR is merged:

https://github.com/xamarin/monodroid/pull/1099

dellis1972:
This is now dependent on https://github.com/xamarin/xamarin-android/pull/4838 being merged.

PR #4830, opened 2020-06-17
Mark API-30 as stable.

jpobst:
Set API-R as API-30, v11.0, and stable.

Notes:
- Requires `monodroid` bump https://github.com/xamarin/monodroid/pull/1102.
- Removes testing of `jnimarshalmethod-gen.exe` because running it on `monoandroid11` hits a crash in Mono trying to enumerate interfaces with default members: https://github.com/dotnet/runtime/issues/34389
- Updated `tests/api-compatibility/acceptable-breakages-vReference.txt` because the default compare will now be against `API-30` instead of `API-29`, and we purposely removed `xxxConst` classes we marked as `[Obsolete]` 5+ years ago.


jonpryor:
Squash-and-merge-

Summary:

```
[Mono.Android] API-30 stabilization (#4830)
```

Body:

```
Context: https://developer.android.com/android11

Changes: https://github.com/xamarin/monodroid/compare/6d05c4929db1ba12ba3a04174938256e435dbbd4...fb5542fa193d5daef1c907f172e0775f9ea4b0d2

  * xamarin/monodroid@fb5542fa1 [Make.config] Update API-R to API-30 (#1102)

API-30 has been enumified (8f9c6382).  Time to declare it stable!

Set API-R as API-30, v11.0, and stable.

Update unit test files so `$(TargetFrameworkVersion)`= v11.0 (API-30).

Rename `src/Mono.Android/Profiles/api-R.params.txt` to
`api-30.params.txt`, to match the stable API-level ID.

Updated `tests/api-compatibility/acceptable-breakages-vReference.txt`
because the default API compare will now be against `API-30`, not
`API-29`, and we deliberately removed the `XxxConst` classes which
were `[Obsolete]`d 5+ years ago in API-30, resulting in an intentional
API break.

TODO:

  * `jnimarshalmethod-gen.exe` tests were disabled as they ran into a
    [runtime crash][0].  Once the runtime crash is fixed, we need to
    update `src/Mono.Android/Test/Mono.Android-Tests.csproj` so that
    `$(AndroidGenerateJniMarshalMethods)` is True when
    `'$(HostOS)' == 'Darwin'`.

[0]: https://github.com/dotnet/runtime/issues/34389
```

Issue #4791, opened 2020-06-09
[Bug] "typemap: module matching MVID [xxxxxxx] not found." A lot of android logs


kuramochia:
Release build using Xamarin Android SDK v10.3.1.4, I will see a large number of logs below.
```
monodroid-assembly typemap: module matching MVID [49B5F91A-2F8E-439E-AD20-F837EB41AFF4] not found.
```

The following command was executed.

```
> adb shell setprop mono.debug.log all
```

And when you view the log, it became the following log.

```
monodroid-assembly typemap: module matching MVID [49B5F91A-2F8E-439E-AD20-F837EB41AFF4] not found.
monodroid typemap: failed to map managed type to Java type: System.Byte, mscorlib, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cecToken=7cec85d7bea7798e (Module: ID: Module: 49b5f91a-2f8e-439e-ad20-f837eb41aff4; Type token: 33554569)
```

I think its same issue with #4727. I want you to fix this, too.

_Originally posted by @kuramochia in https://github.com/xamarin/xamarin-android/issues/4727#issuecomment-641055033_

jonpryor:
The fix for Issue #4727 in 755a521d0925e648159acaf04998e43c1e462f04 was to turn the messages from *warning* messages, which were *always* written, to *info* messages, which would only be shown if "assembly" messages were enabled.

By running `adb shell set prop debug.mono.log all`, you have indicated that you want *all* messages shown.  As such, *all* messages are shown, and is generally considered to be a Bad Idea™, as LREF logging is also enabled ("all" means *all*), which will quickly drown out logcat.

This is By Design™.

If you don't want all messages shown, then don't enable them all, which can be done by clearing the `debug.mono.log` system property:

    adb shell setprop debug.mono.log "''"

kuramochia:
@jonpryor below message can change to info level?
```
monodroid-assembly typemap: module matching MVID [49B5F91A-2F8E-439E-AD20-F837EB41AFF4] not found.
```
this message shown in large number of logs.

`adb shell setprop debug.mono.log` just checked why this log is logged.

roninstar:
Just to make sure I am not missing something.  I never had any monodroid warning error in my Release build before this most recent Visual Studio upgrade to 16.6.2. Now I get these errors every few seconds. Why did this just start?  How can I prevent this from logging in a production device if its a benign warning?
[...]

PR #4831, opened 2020-06-17
[runtime] Downgrade some typemap message severity

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/4791

Make a couple of messages related to MVID use `log_info` instead of
`log_warn` so that they are not shown by default (even though they are
useful for diagnostic purposes)

### DIFF
--- a/src/monodroid/jni/embedded-assemblies.cc
+++ b/src/monodroid/jni/embedded-assemblies.cc
@@ -295,7 +295,7 @@ EmbeddedAssemblies::typemap_java_to_managed (const char *java_type_name)
 
 	MonoType *type = mono_reflection_type_from_name (const_cast<char*>(managed_type_name), nullptr);
 	if (XA_UNLIKELY (type == nullptr)) {
-		log_warn (LOG_ASSEMBLY, "typemap: managed type '%s' (mapped from Java type '%s') could not be loaded", managed_type_name, java_type_name);
+		log_info (LOG_ASSEMBLY, "typemap: managed type '%s' (mapped from Java type '%s') could not be loaded", managed_type_name, java_type_name);
 		return nullptr;
 	}
 
@@ -505,7 +505,7 @@ EmbeddedAssemblies::typemap_managed_to_java ([[maybe_unused]] MonoType *type, Mo
 
 	const TypeMapModule *match = binary_search<uint8_t, TypeMapModule, compare_mvid> (mvid, map, map_entry_count);
 	if (match == nullptr) {
-		log_warn (LOG_ASSEMBLY, "typemap: module matching MVID [%s] not found.", MonoGuidString (mvid).get ());
+		log_info (LOG_ASSEMBLY, "typemap: module matching MVID [%s] not found.", MonoGuidString (mvid).get ());
 		return nullptr;
 	}
 


brendanzagaeski:
To ask a potentially silly question for the sake of sanity checking, are all of the other remaining uses of `log_warn()` in `embedded-assemblies.cc` for typemaps actionable by users? Or to put it another way, if one of the remaining warnings appeared in a log, would that most likely mean something was incorrect (either with the application or with Xamarin.Android itself)?

brendanzagaeski:
Also, for the other message from #4791:

```
monodroid typemap: failed to map managed type to Java type: System.Byte, mscorlib, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cecToken=7cec85d7bea7798e (Module: ID: Module: 49b5f91a-2f8e-439e-ad20-f837eb41aff4; Type token: 33554569)
```

Does that indicate a problem with the app or Xamarin.Android, or would the following line from `JNIEnv.cs` also need to be adjusted to use `LogLevel.Info` and `LogCategories.Assembly`?

https://github.com/xamarin/xamarin-android/blob/17db4dc1e8ba6438fc02fdbb0e2fbe6c16fb4b58/src/Mono.Android/Android.Runtime/JNIEnv.cs#L710

Thanks!

grendello:
> To ask a potentially silly question for the sake of sanity checking, are all of the other remaining uses of `log_warn()` in `embedded-assemblies.cc` for typemaps actionable by users? Or to put it another way, if one of the remaining warnings appeared in a log, would that most likely mean something was incorrect (either with the application or with Xamarin.Android itself)?

None of those messages are actionable by users, but they are valuable information for us should a crash happen on some devices they don't have access to. They can indicate a problem with both Xamarin.Android and the user app.

grendello:
> Also, for the other message from #4791:
> 
> ```
> monodroid typemap: failed to map managed type to Java type: System.Byte, mscorlib, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cecToken=7cec85d7bea7798e (Module: ID: Module: 49b5f91a-2f8e-439e-ad20-f837eb41aff4; Type token: 33554569)
> ```
> 
> Does that indicate a problem with the app or Xamarin.Android, or would the following line from `JNIEnv.cs` also need to be adjusted to use `LogLevel.Info` and `LogCategories.Assembly`?
> 
> https://github.com/xamarin/xamarin-android/blob/17db4dc1e8ba6438fc02fdbb0e2fbe6c16fb4b58/src/Mono.Android/Android.Runtime/JNIEnv.cs#L710
> 
> Thanks!

In this particular case Xamarin.Android managed runtime should never try to look up type mapping for a standard type (or, rather, more general - not derived from `Java.Lang.Object`) but it's something we don't have a fix for now. The warning message, if it applied to a different type than `System.Byte`, something derived from JLO, **would** indicate a problem with the maps generated for the app.

brendanzagaeski:
Thanks for the additional background info! Given the additional information that (a) it's known that the `module matching MVID ... not found` message was more common than originally expected due to a particular way the managed code currently tries to look up type maps for types that do not inherit from `Java.Lang.Object` and (b) the remaining messages are currently expected to be quite rare, the current scope of this PR to adjust the verbosity level of the `module matching MVID ... not found` message in particular will be useful to include as is.

For the other message from #4791:

```
monodroid typemap: failed to map managed type to Java type: System.Byte, mscorlib, Version=2.0.5.0, Culture=neutral, PublicKeyToken=7cecToken=7cec85d7bea7798e (Module: ID: Module: 49b5f91a-2f8e-439e-ad20-f837eb41aff4; Type token: 33554569)
```

I'll raise a question with the original reporter of the issue to see if they can provide a code sample that reproduces that behavior so the team can take a look at the best way to resolve it. For example, one idea is that it might indicate a case where adjusting the app code to use a .NET BCL type instead of a Java type code could improve performance.

brendanzagaeski:
_Draft release notes_

Here's what I'll plan to include for this PR in the release notes:

```

- [GitHub 4791](https://github.com/xamarin/xamarin-android/issues/4791):
  Starting in Xamarin.Android 10.3, many _typemap: module matching MVID \[...\]
  not found._ messages were logged to the application output for certain running
  apps, obscuring other logging statements of interest.  These messages were
  diagnostic and did not indicate problems in the apps.
```

To suggest any changes, feel free to reply on this PR. Thanks!

PR #4835, opened 2020-06-18
Updating the version of cmdline-tools to "latest"

joj:
Google changed the version indicator of cmdline-tools to "latest", and now it installs in a directory named `cmdline-tools/latest`. I'm updating this file to reflect that.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props.in
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props.in
@@ -9,7 +9,7 @@
 		<YieldDuringToolExecution Condition="'$(YieldDuringToolExecution)' == ''">true</YieldDuringToolExecution>
 		<LatestSupportedJavaVersion Condition="'$(LatestSupportedJavaVersion)' == ''">11.0.4</LatestSupportedJavaVersion>
 		<MinimumSupportedJavaVersion Condition="'$(MinimumSupportedJavaVersion)' == ''">1.6.0</MinimumSupportedJavaVersion>
-		<AndroidCommandLineToolsVersion Condition=" '$(AndroidCommandLineToolsVersion)' == '' ">1.0</AndroidCommandLineToolsVersion>
+		<AndroidCommandLineToolsVersion Condition=" '$(AndroidCommandLineToolsVersion)' == '' ">latest</AndroidCommandLineToolsVersion>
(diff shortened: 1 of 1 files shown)


jonpryor:
> Google changed the version indicator of cmdline-tools to "latest",

Not exactly correct.  The Google repository has always had multiple versions listed for cmdline-tools: https://dl.google.com/android/repository/repository2-1.xml

Right now it has three:

```powershell
cat ~/Downloads/repository2-1-3.xml | Select-String -Pattern "cmdline-tools"
	<remotePackage path="cmdline-tools;2.0">
	<remotePackage path="cmdline-tools;latest">
	<remotePackage path="cmdline-tools;1.0">
```

You can consider `latest` to be a "symbolic link" to the latest version.  (Kinda, sorta.) Except that when installed, it's *not* a symbolic link, it's an actual directory, named `latest`:

```bash
$ tools/bin/sdkmanager  "cmdline-tools;latest"
$ cat cmdline-tools/latest/tools/source.properties 
Pkg.Revision=2.0
Pkg.Path=cmdline-tools;2.0
Pkg.Desc=Android SDK Command-line Tools
```

The problem with `latest` is that *we have no idea what it is*.

There are two primary ideas being having (user-overridable) versions for packages, such as `$(AndroidSdkBuildToolsVersion)` and `$(AndroidCommandLineToolsVersion)`:

 1. So that *we can ensure things work*, and
 2. Repeatable builds.

If your dependencies vary in uncontrollable ways, you don't have repeatable builds. The end.

Meanwhile, consider `$(AndroidSdkBuildToolsVersion)`.  If that "followed" `$(AndroidCommandLineToolsVersion)` and had a "latest" value, then (presumably?) people would *automatically* start using r30.0.0, which would break ~everyone.  (Build-tools r30 requires JDK 11, JDK 11 is *barely* supported, and *probably* breaks the designer, and…)

In short, to my mind, a "floating" dependency version such as `latest` is a Bad Idea™.

If we want people to use a newer version, then bump it to `2.0`.  Please don't bump to `latest`.  That's an *abdication* of our own responsibilities.

jonpryor:
`tools/bin/sdkmanager  "cmdline-tools;2.0" also Works For Me™ to explicitly install the 2.0 version and not "latest".

joj:
@jonpryor is right (which is not surprising). We have changed our tools to accommodate this versioning scheme from Google, which is unlike any other scheme they did in the past (in general, when the folder name includes the version, that means that versions can be installed side-by-side; versions can be installed side-by-side in this case, but that was unexpected because it doesn't _mean_ much). I'm closing this PR. It's not needed and not correct. As for going to 2.0, I'll let the xamarin android team decide that based on what they think is best. That's not a hard requirement on my side of things.

PR #4838, opened 2020-06-19
[tests] update and consolidate NUnit packages

jonathanpeppers:
Update to:

* NUnit 3.12.0
* NUnit.ConsoleRunner 3.11.1
* NUnit3TestAdapter 3.16.1

These are all different numbers, but these are the latest stable
versions on nuget.org.

I also went ahead and consolidated all NUnit `@(PackageReference)` to
a `NUnitReferences.projitems` file. This way we can list the versions
for all NUnit packages in one MSBuild file.

NUnit.ConsoleRunner was still listed in a few other files we will have
to manually update.

Bump to xamarin/monodroid/master@82bea9c6

Changes: https://github.com/xamarin/monodroid/compare/6e43a20e...82bea9c6

### DIFF
--- a/.external
+++ b/.external
@@ -1,2 +1,2 @@
-xamarin/monodroid:master@6e43a20ebaa2cd59532e130ef64585c5829e3373
+xamarin/monodroid:master@82bea9c6b963cb855a6e01775084182ab45cf9ac
 mono/mono:2020-02@83105ba22461455f4343d6bb14976eba8b0b3f39
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,7 +8,7 @@
             "name": "Launch",
             "type": "mono",
             "request": "launch",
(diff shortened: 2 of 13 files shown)


jonathanpeppers:
It doesn't look like this helps the timeout we see sometimes for `Test MSBuild - 1`, but it is probably worth updating these anyway, nothing appears to be broken.

jonathanpeppers:
This will require a change on the monodroid side: 

https://github.com/xamarin/xamarin-android/blob/d44e48ab93d0562cc8f4c4ede440af18ec141f1f/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Xamarin.Android.Build.Tests.csproj#L19-L23

jonathanpeppers:
It seems like updating NUnit makes the test run _consistently_ hang? I'll revisit this at some point.

Issue #4817, opened 2020-06-13
new() constraint and XElement doesn't work in Release mode


StoatsOnAPlane:
I have found that when I try to use an XElement and a generic method with the new() constraint, the app terminates immediately in Release Mode

Not sure if this is an issue with Xamarin or what, but it has suddenly started giving me issues.  Took me a while to track it down, but I can recreate the issues in these steps.  Only crashes in release mode.  Linking set to None does not fix the issue.

I've attached a demo project to reproduce
[BreakingStuff.zip](https://github.com/xamarin/xamarin-android/files/4775328/BreakingStuff.zip)

### Steps to Reproduce

1. Create a new Android App
2. Add references to System.Xml and System.Xml.Linq
3. Create a generic method that takes in an XElement and has "where T : new()" as a constraint
4. Call the method

Here's the code in the project I'm testing with
```
protected override void OnCreate(Bundle savedInstanceState)
        {
            base.OnCreate(savedInstanceState);
            Xamarin.Essentials.Platform.Init(this, savedInstanceState);

            XElement element = new XElement("Test");

            Log.WriteLine(LogPriority.Info, "TestApp", "About to call method");
            bool testBool = GetT<bool>(element, "TestBool");
            Log.WriteLine(LogPriority.Info, "TestApp", "method called");

            if (testBool == true)
            {
                return;
            }

            // Set our view from the "main" layout resource
            SetContentView(Resource.Layout.activity_main);
        }

public T GetT<T>(XElement element, string attributeName) where T : new()
        {
            XAttribute xAttribute = element.Attribute(attributeName);

            if(xAttribute == null)
            {
                return new T();
            }

            return (T)Convert.ChangeType(xAttribute.Value, typeof(T), CultureInfo.InvariantCulture);
        }

```
### Expected Behavior
The app runs in both Debug and Release mode

### Actual Behavior
The app runs in Debug but in Release the app terminates unexpectedly, posts this in Logcat

Time	Device Name	Type	PID	Tag	Message
06-13 14:53:33.886	Google Pixel 4	Error	14108	libc	Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 14108 (anyname.testapp), pid 14108 (anyname.testapp)
06-13 14:53:33.886	Google Pixel 4	Info	14108	TestApp	About to call method
[...]

PR #4844, opened 2020-06-22
Bump mono to mono/mono@109b4b6f4

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/4817

Changes: https://github.com/mono/mono/compare/87ef5557017a8d822ae31587846a54fcd66daf1b...2020-02

  * mono/mono@109b4b6f445 [jit] Fix the Activator.CreateInstance ()
    intrinsic for primitive types. (#19989) (#19994)
  * mono/mono@15782aa7404 If there is a crash and the debugger is
    attached, we call mini_get_dbg_callbacks ()->send_crash(). That call
    ends up in mono_debugger_agent_send_crash which does all kinds of
    questionable stuff like trying to take the loader lock. Eventually
    it ends up calling socket_transport_send which tries to do a
    transition to GC Safe mode. I inserted a MONO_ENTER_GC_UNSAFE in
    mono_debugger_agent_send_crash function and this fixes the assert. (#19992)
  * mono/mono@6ea4cbcee04 [2020-02] [merp] Add API methods for getting
    hashcode/reason of last crash (#19978)
  * mono/mono@d45ff3af75b [merp] Add tests for crashing via POSIX signal (#19977)
  * mono/mono@bc1dd071a71 [merp] Add a test for Mono.Runtime.SendExceptionToTelemetry() (#19966)


jonpryor:
Related: https://github.com/xamarin/xamarin-macios/pull/8928

Issue #4852, opened 2020-05-08
[Bug] Crash with Android 5.x and x64 devices


stevecluk:
### Description
I have an app that's been in the Play Store for a few years now. It was recently updated with the latest nugets and Forms, and the Pre-Launch report started showing crashes in the Huawei p8 Lite.
The crash only seems to happen on x64 devices (and emulators) that are running Android 5.0 or 5.1.  Android 5.x 32-bit devices, and all 6+ devices are fine.

The main crash I was seeing was a result of trying to check if a folder exists from the MainActivity's OnCreate(), showing a Dll issue: 
```
android.runtime.JavaProxyThrowable: System.DllNotFoundException: libmono-native.so assembly:<unknown assembly> type:<unknown type> member:(null)
  at (wrapper managed-to-native) Interop+Sys.Stat(byte&,Interop/Sys/FileStatus&)
  at Interop+Sys.Stat (System.ReadOnlySpan`1[T] path, Interop+Sys+FileStatus& output) [0x00028] in <ac2cc6a1f7c64bf5b08e0682cda293fb>:0 
  at System.IO.FileSystem.FileExists (System.ReadOnlySpan`1[T] fullPath, System.Int32 fileType, Interop+ErrorInfo& errorInfo) [0x00007] in <ac2cc6a1f7c64bf5b08e0682cda293fb>:0 
  at System.IO.FileSystem.DirectoryExists (System.ReadOnlySpan`1[T] fullPath, Interop+ErrorInfo& errorInfo) [0x00000] in <ac2cc6a1f7c64bf5b08e0682cda293fb>:0 
  at System.IO.FileSystem.DirectoryExists (System.ReadOnlySpan`1[T] fullPath) [0x00000] in <ac2cc6a1f7c64bf5b08e0682cda293fb>:0 
  at System.IO.Directory.Exists (System.String path) [0x0001e] in <ac2cc6a1f7c64bf5b08e0682cda293fb>:0 
```
To reproduce the problem, I'd created an empty Xamarin Forms app using the latest Visual Studio, and tried to run it (debug or release) on the Android x64 emulator, which also crashed on startup but with a different library error this time.
As with the Interop issue, this crash only happens on 64-bit builds of 5.0/5.1 emulator (and devices), where as 32-bit builds work fine.  Android 6+ devices run either x86/x64 fine.
```
05-08 11:55:05.178: E/AndroidRuntime(7531): Process: com.companyname.crashtest, PID: 7531
[...]

Issue #4772, opened 2020-06-03
System.DllNotFoundException: libmono-native.so assembly:<unknown assembly> type:<unknown type> member:(null)


todorspetkov:
### Description
Observing the following Exception in Firebase Crashlytics only on Android 5.0.x devices, 90% on Huawei devices. Release build (no shared mono runtime). Anyone seen somthething similar?

```
Fatal Exception: android.runtime.JavaProxyThrowable: System.DllNotFoundException: libmono-native.so assembly:<unknown assembly> type:<unknown type> member:(null)
  at (wrapper managed-to-native) Interop+Sys.GetNonCryptographicallySecureRandomBytes(byte*,int)
  at Interop.GetRandomBytes (System.Byte* buffer, System.Int32 length) [0x00000] in <1b7a5cd4b5194667aea235f87b2a861e>:0 
  at System.Guid.NewGuid () [0x00000] in <1b7a5cd4b5194667aea235f87b2a861e>:0 
  Comune.InitializeComponent () [0x00182] in <1170faa3308346b08cbc4f21dc5916bc>:0 
  at xxx.Views.SelezioneComune..ctor (System.Boolean isCambia) [0x0000c] in <1170faa3308346b08cbc4f21dc5916bc>:0 
  at xxx.App.OnStart () [0x000aa] in <1170faa3308346b08cbc4f21dc5916bc>:0 
  at Xamarin.Forms.Application.SendStart () [0x00000] in <9608dd67eb804f0494e04a66ff90c168>:0 
  at Xamarin.Forms.Platform.Android.FormsAppCompatActivity.OnStateChanged () [0x0001b] in <63e7c9910ab446ac9b78b1ac3ac3e5bb>:0 
  at Xamarin.Forms.Platform.Android.FormsAppCompatActivity.OnStart () [0x00046] in <63e7c9910ab446ac9b78b1ac3ac3e5bb>:0 
  at xxx.Droid.MainActivity.OnStart () [0x00000] in <db697bb0cebf40899ddd1dbba31ef430>:0 
  at Android.App.Activity.n_OnStart (System.IntPtr jnienv, System.IntPtr native__this) [0x00008] in <26ea6ddb724e4be2b66b6bde37c83c15>:0 
  at (wrapper dynamic-method) Android.Runtime.DynamicMethodNameCounter.14(intptr,intptr)
       at crc6473e84085518c5ff7.MainActivity.n_onStart(MainActivity.java)
[...]

PR #4861, opened 2020-06-23
[runtime] Preload `libmono-native.so` before initializing XA

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/4772
Fixes: https://github.com/xamarin/xamarin-android/issues/4852

Fixes apply to the the **Release** builds on Android 5.x devices.

Android 5.x devices have a bug in their 64-bit dynamic linker which
prevents it from correctly resolving shared library dependencies when a
library required by the one currently being loaded is not found in the
same directory.  The problem is that the Android dynamic linker on those
devices fails to properly interpret the internal library name (so-called
`SONAME`, contained in the `DT_SONAME` field in the ELF executable
header) when checking whether the library is already loaded in the
process memory.  Failing that check, the loader proceeds to look for the
library **file** on the disk/storage medium in the predefined paths.
Since `libmonosgen-2.0.so` and other libraries are not found in those
paths, the search fails and the linker returns an error.  Later versions
of Android fix the bug in that the `DT_SONAME` field is used to properly
look up already loaded shared libraries and thus the dependency
resolution succeeds.

The solution here is to use the Java API, `System.loadLibrary`, to
preload `libmono-native.so` because the Java API is able to somehow
figure out that the dependent libraries are found in the same directory
and load them to satisfy the requirements.  We could probably implement
the same algorithms in our native code, but it feels it would be a
wasted effort since Android versions above 5.x work correctly.

### DIFF
--- a/src/java-runtime/java/mono/android/MonoPackageManager.java
+++ b/src/java-runtime/java/mono/android/MonoPackageManager.java
@@ -51,7 +51,7 @@ public class MonoPackageManager {
 				String externalLegacyDir = new java.io.File (
 							external0,
 							"../legacy/Android/data/" + context.getPackageName () + "/files/.__override__").getAbsolutePath ();
-				boolean embeddedDSOsEnabled = (runtimePackage.flags & FLAG_EXTRACT_NATIVE_LIBS) == 0;
+				boolean embeddedDSOsEnabled = android.os.Build.VERSION.SDK_INT >= 23 && (runtimePackage.flags & FLAG_EXTRACT_NATIVE_LIBS) == 0;
 				String runtimeDir = getNativeLibraryPath (runtimePackage);
 				String[] appDirs = new String[] {filesDir, cacheDir, dataDir};
 				String[] externalStorageDirs = new String[] {externalDir, externalLegacyDir};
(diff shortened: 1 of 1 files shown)


jonpryor:
Commit message for review:

```
Fixes: https://github.com/xamarin/xamarin-android/issues/4772
Fixes: https://github.com/xamarin/xamarin-android/issues/4852

Context: https://issuetracker.google.com/issues/36950617

Fixes apply to the the **Release** builds on Android 5.x devices.

A Xamarin.Android application contains multiple native libraries:

	$ adb shell pm list packages -f | grep -i hello
	package:/data/app/com.xamarin.android.helloworld-_-zYezFQQsuaiegto4HlCw==/base.apk=com.xamarin.android.helloworld

	$ adb shell ls /data/app/com.xamarin.android.helloworld-_-zYezFQQsuaiegto4HlCw==/lib/arm64
	…
	libmono-native.so
	libmonosgen-2.0.so
	…

Native libraries may have dependencies, e.g. `libmono-native.so`
depends on `libmonosgen-2.0.so` in order to load.

Once Upon A Time, Android's native linker was "not helpful" when
loading native libraries through **dlopen**(3): it wouldn't use the
library's directory to resolve dependencies, and thus
`dlopen("libmono-native.so")` would fail.

[This limitation was (mostly) fixed in Android 4.3 (API-18)][0], but
a variation of this limitation is also present on some Android 5.x
devices (API-21, API-22) when running *64-bit apps*.  For example, if
a C# API call requires loading `libmono-native.so`, then on these
devices the load will fail, as `libmonosgen-2.0.so` can't be found --
which is rather unexpected, as `libmonosgen-2.0.so` is *already*
loaded (!) -- which results in an app crash:


	UNHANDLED EXCEPTION:
	System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation. --->
		System.TypeInitializationException: The type initializer for 'PanCardView.Droid.CardsViewRenderer' threw an exception. --->
		System.TypeInitializationException: The type initializer for 'System.Random' threw an exception. --->
		System.DllNotFoundException: libmono-native.so assembly:<unknown assembly> type:<unknown type> member:(null)
	  at (wrapper managed-to-native) Interop+Sys.GetNonCryptographicallySecureRandomBytes(byte*,int)
	  at Interop.GetRandomBytes (System.Byte* buffer, System.Int32 length) <0x7f86c3bbf0 + 0x0001b> in <39e41a681ea94f00bcc87fb130853dc2>:0 
	  at System.Random.GenerateGlobalSeed () <0x7f86cb33f0 + 0x00017> in <39e41a681ea94f00bcc87fb130853dc2>:0 
	  at System.Random..cctor () <0x7f86cb3690 + 0x0000b> in <39e41a681ea94f00bcc87fb130853dc2>:0 
	  …

The problem is that the Android dynamic linker on these devices fails
to properly interpret the internal library name (so-called `SONAME`,
contained in the `DT_SONAME` field in the ELF executable header) when
checking whether the library is already loaded in the process memory.
Failing that check, the loader proceeds to look for the library
**file** on the disk/storage medium in the predefined paths.  Since
`libmonosgen-2.0.so` and other libraries are not found in those paths,
the search fails and the linker returns an error.

Later versions of Android fix the bug in that the `DT_SONAME` field
is used to properly look up already loaded shared libraries and thus
the dependency resolution succeeds.

The solution here is to use the `System.loadLibrary()` Java API to
preload `libmono-native.so`, as the Java API is able to figure out
that the dependent libraries are found in the same directory and load
them to satisfy the requirements.

[0]: https://issuetracker.google.com/issues/36950617
```

PR #4863, opened 2020-06-24
Bump to xamarin/monodroid/master@6e43a20e

jonathanpeppers:
Changes: https://github.com/xamarin/monodroid/compare/fb5542fa...6e43a20e

* [tools/msbuild] handle multiple abi's in ro.product.cpu.abilist64
* [build] restore API 19 & 20
* Bump to xamarin/android-sdk-installer/master@23fc8d8
* Bump to xamarin/androidtools/master@3f51412d
* Bump to xamarin/xamarin-android/master@8576118a

### DIFF
--- a/.external
+++ b/.external
@@ -1,2 +1,2 @@
-xamarin/monodroid:master@fb5542fa193d5daef1c907f172e0775f9ea4b0d2
+xamarin/monodroid:master@6e43a20ebaa2cd59532e130ef64585c5829e3373
 mono/mono:2020-02@83105ba22461455f4343d6bb14976eba8b0b3f39


jonathanpeppers:
It seems like this breaks some tests:

```
Xamarin.Android.Bcl_Tests, MonoTests.Mono.Unix.Native.MemfdTest.TestMemfd / Release

System.ArgumentOutOfRangeException : Current platform doesn't support this value.
Parameter name: value
Actual value was MFD_CLOEXEC, MFD_ALLOW_SEALING.
```

```
Mono.Android_TestsMultiDex, System.NetTests.NetworkInterfacesTest.DotNetInterfacesShouldEqualJavaInterfaces / Release

8.3 (lo not found in Java interfaces)
Expected: True
But was: False
```

Then an instance of `Mono.Android-Tests` crashed.

jonathanpeppers:
I think we should bump the API 30 emulator in a separate PR that fixes the runtime issues.

Issue #4853, opened 2020-06-23
"error XA0030: Building with JDK version `11.0.7` is not supported" when attempting to use a JDK 11 version that is a bit too new


brendanzagaeski:
### Steps to reproduce

1. Download the current version of the _jbrsdk_ JetBrains Runtime that's available on <https://confluence.jetbrains.com/display/JBR/JetBrains+Runtime>.

   Today, that's <https://bintray.com/jetbrains/intellij-jbr/jbrsdk11-windows-x64/765.64>, OpenJDK version 11.0.7.

2. Extract the files to a directory.

3. Set Xamarin.Android builds to use that JDK 11 directory. For example, open **Tools > Options** in Visual Studio, select the **Xamarin > Android Settings** node, and set **Java Development Kit Location** to that directory.

4. Attempt to build a Xamarin.Android app project.

### Expected behavior

Maybe the build should succeed? That is, maybe the explicit version check for the JDK version number can now use just the major.minor part of the version number instead of the full major.minor.build? It seems like the third place of JDK 11 version numbers (the `Build` number in `System.Version` terminology) changes more frequently than it did for the old 1.8.0 version number scheme of JDK 8.

### Actual behavior

The build fails because the third place of the JDK version number is higher than the current `$(LatestSupportedJavaVersion)`:

```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android\Xamarin.Android.Legacy.targets(163,5): error XA0030: Building with JDK version `11.0.7` is not supported. Please install JDK version `11.0.4`. See https://aka.ms/xamarin/jdk9-errors
```

### Version information

Microsoft Visual Studio Enterprise 2019 Int Preview
Version 16.7.0 Preview 4.0 [30222.8.master]

Xamarin.Android SDK   10.4.0.0 (d16-7/de70286)

jonathanpeppers:
@jonpryor should this be 11.0.99?

https://github.com/xamarin/xamarin-android/blob/88215f93fdcf14fa1a2aebbb499c09c6bdb8fa8c/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props.in#L10

Or what do we actually do here? It seems like they will bump the minor version all the time.

brendanzagaeski:
_Release status update_

A new Preview version of Xamarin.Android has now been published that includes the fix for this item. The fix is not yet included in a Release version. I will update this again when a Release version is available that includes the fix.

[...]

PR #4876, opened 2020-06-26
[Xamarin.Android.Build.Tasks] Set LatestSupportedJavaVersion=v11.0

jonpryor:
Fixes: https://github.com/xamarin/xamarin-android/issues/4853

`$(LatestSupportedJavaVersion)` is currently v11.0.4, which means if
e.g. JDK 11.0.7 is used -- which one would expect to be compatible! --
the build will fail with an XA0030 error:

	error XA0030: Building with JDK Version `11.0.7` is not supported. Please install JDK version `11.0.4`.

We *could* set `$(LatestSupportedJavaVersion)`=v11.0.99, but that
value is also used as part of the XA0030 message text, which could
result in an error message mentioning a version which doesn't exist:

	error XA0030: Building with JDK Version `14.0.0` is not supported. Please install JDK version `11.0.99`.

Instead of using a three-tuple in the `$(LatestSupportedJavaVersion)`
value -- which is "forwarded" to the XA0030 error -- instead use a
two-tuple value of `11.0`, which will result in a "saner" XA0030 error
message of `11.0` and not `11.0.4` or (worse!) `11.0.99`.

To bridge the gap between three-tuples (from `java -version`) and
the two-tuple in `$(LatestSupportedJavaVersion)`, add a new
`ValidateJavaVersion.CheckJavaGreaterThanLatestSupported()` method
which will *append* `.99` to `$(LatestSupportedJavaVersion)` if it
only contains two values, allowing us to compare `11.0.7 >= 11.0.99`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.props
@@ -5,7 +5,7 @@
   <PropertyGroup>
     <UsingAndroidNETSdk>true</UsingAndroidNETSdk>
     <AndroidCommandLineToolsVersion Condition=" '$(AndroidCommandLineToolsVersion)' == '' ">1.0</AndroidCommandLineToolsVersion>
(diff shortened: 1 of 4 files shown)


jonpryor:
Superseded by PR #4875.

Issue #3745, opened 2019-10-07
Assembly ending with "resources" causing mislocated assembly in apk


erikpowa:
short:
"user" assembly -> `foo-resources.dll`
assembly location in apk -> `\assemblies\assets\foo-resources.dll`

### Steps to Reproduce

1. Create empty app
2. Create a library
3. Add property to library `<AssemblyName>foo-resources</AssemblyName>`
4. Reference the library
5. BuildApk

### Expected Behavior
`\assemblies\foo-resources.dll`

### Actual Behavior
`\assemblies\assets\foo-resources.dll`

### Version Information

master branch

erikpowa:
Can't repro anymore. Tested at https://github.com/xamarin/xamarin-android/commit/da46500671b82fa7f15209ea2a091581cbf820f4

erikpowa:
My test was faulty 🤦 which used `foo-Resources`, but still can repro with lower-case `foo-resources`.

brendanzagaeski:
_Notes for the Xamarin.Android team_

This issue will by default result in a _FileNotFoundException: Could not load file or assembly_ exception if an app attempts to load a non-satellite assembly that has a name ending with _.resources.dll_.

I encountered this behavior myself today in a test failure while working on and unrelated PR, so I started writing an automated build test for it before I found this existing issue. In case it might be handy for some point in the future, here's the test:

```cs
[Test]
public void AssembliesThatAreNamedLikeSatelliteAssembliesButAreNot ()
{
	var path = Path.Combine ("temp", TestName);
	var app = new XamarinAndroidApplicationProject {
		ProjectName = "App",
		Sources = {
			new BuildItem.Source ("Class1.cs") {
				TextContent = () => "public class Class1 : Library1.resources.Class1 { }"
			},
		},
		IsRelease = true
	};
	var lib = new DotNetStandard {
		ProjectName = "Library1.resources",
		Sdk = "Microsoft.NET.Sdk",
		TargetFramework = "netstandard2.0",
		Sources = {
			new BuildItem.Source ("Class1.cs") {
				TextContent = () => "namespace Library1.resources { public class Class1 { } }"
			}
		}
	};
	app.References.Add (new BuildItem.ProjectReference ($"..\\{lib.ProjectName}\\{lib.ProjectName}.csproj", lib.ProjectName, lib.ProjectGuid));

[...]

PR #4887, opened 2020-07-01
[Xamarin.Android.Build.Tasks] Deploy changes to transitive reference assembly dependencies

brendanzagaeski:
Fixes: https://developercommunity.visualstudio.com/content/problem/1086457/index.html
Context: 8f2ae24813ec3cdab56f688c4c3b611a33759333
Context: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/938468

If a Xamarin.Android app referenced a .NET Standard library that had
`$(ProduceReferenceAssembly)`=`true`, and that library in turn
referenced a second .NET Standard library, then changes to the second
library would not be deployed during incremental deployments in Visual
Studio unless the app project also directly referenced the second
library.

For additional context, note that as discussed in internal work item
938468, if a .NET Framework console application references a .NET
Standard library that has `$(ProduceReferenceAssembly)`=`true`, then
changes to the library are similarly missing in the running console app
after incremental builds in Visual Studio.

Although the problem in Xamarin.Android apps was similar to the .NET
Framework behavior, an important difference was that Xamarin.Android
apps did reflect changes correctly for libraries referenced directly by
the app.  Only transitive references caused trouble in Xamarin.Android.

The reason direct references worked was that Visual Studio always ran
Xamarin.Android's `Install` target on each incremental deployment, and
the `<ResolveAssemblies/>` task in that target correctly located the
up-to-date assemblies for direct references.

For example, for the included build test, the `<ResolveAssemblies/>`
task resolved the `Library1` reference to:

	C:\Source\xamarin-android\bin\TestDebug\temp\TransitiveDependencyProduceReferenceAssembly\Library1\bin\Debug\netstandard2.0\Library1.dll

In contrast, the `<ResolveAssemblies/>` task resolved the `Library2`
reference to:

	C:\Source\xamarin-android\bin\TestDebug\temp\TransitiveDependencyProduceReferenceAssembly\App\bin\Debug\Library2.dll

This was problematic because that copy of the library was left over from
the previous deployment.  In a sense, this was a circular reference: The
input for the current deployment was the output from the previous
deployment.

To fix this behavior, include more of the items from
`@(ReferenceCopyLocalPaths)` in the `@(FilteredAssemblies)` input to
`<ResolveAssemblies/>` by removing the
`'%(ReferenceCopyLocalPaths.RelativeDir)' == ''` condition.

For the included build test, this means the following additional item is
now included in `@(FilteredAssemblies)`:

	C:\Source\xamarin-android\bin\TestDebug\temp\TransitiveDependencyProduceReferenceAssembly\Library1\bin\Debug\netstandard2.0\Library2.dll

This path is then treated as a top-level reference in
`<ResolveAssemblies/>`, so the task no longer needs to try to resolve it
from elsewhere.

In theory, another way to fix the problem could be to change how
`<ResolveAssemblies/>` resolves transitive dependencies, but
`<ResolveAssemblies/>` is only used in the legacy build targets, so
adding more sophistication now would have limited value.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
@@ -517,6 +517,68 @@ namespace Lib2
 		}
 
 		[Test]
+		public void TransitiveDependencyProduceReferenceAssembly ()
+		{
+			var path = Path.Combine (Root, "temp", TestName);
+			var app = new XamarinAndroidApplicationProject {
+				ProjectName = "App",
+				Sources = {
+					new BuildItem.Source ("Class1.cs") {
+						TextContent = () => "public class Class1 : Library1.Class1 { }"
+					},
+				}
+			};
+			var lib1 = new DotNetStandard {
+				ProjectName = "Library1",
+				Sdk = "Microsoft.NET.Sdk",
+				TargetFramework = "netstandard2.0",
+				Sources = {
+					new BuildItem.Source ("Class1.cs") {
+						TextContent = () => "namespace Library1 { public class Class1 { } }"
+					},
+					new BuildItem.Source ("Class2.cs") {
+						TextContent = () => "namespace Library1 { public class Class2 : Library2.Class1 { } }"
+					}
+				}
+			};
+			lib1.SetProperty ("ProduceReferenceAssembly", "True");
+			var lib2 = new DotNetStandard {
+				ProjectName = "Library2",
+				Sdk = "Microsoft.NET.Sdk",
+				TargetFramework = "netstandard2.0",
+				Sources = {
+					new BuildItem.Source ("Class1.cs") {
+						TextContent = () => "namespace Library2 { public class Class1 { } }"
+					},
+				}
+			};
+			lib2.SetProperty ("ProduceReferenceAssembly", "True");
+			lib1.OtherBuildItems.Add (new BuildItem.ProjectReference ($"..\\{lib2.ProjectName}\\{lib2.ProjectName}.csproj", lib2.ProjectName, lib2.ProjectGuid));
+			app.References.Add (new BuildItem.ProjectReference ($"..\\{lib1.ProjectName}\\{lib1.ProjectName}.csproj", lib1.ProjectName, lib1.ProjectGuid));
+
+			using (var lib2Builder = CreateDllBuilder (Path.Combine (path, lib2.ProjectName), cleanupAfterSuccessfulBuild: false))
+			using (var lib1Builder = CreateDllBuilder (Path.Combine (path, lib1.ProjectName), cleanupAfterSuccessfulBuild: false))
+			using (var appBuilder = CreateApkBuilder (Path.Combine (path, app.ProjectName))) {
+				Assert.IsTrue (lib2Builder.Build (lib2), "first Library2 build should have succeeded.");
+				Assert.IsTrue (lib1Builder.Build (lib1), "first Library1 build should have succeeded.");
+				Assert.IsTrue (appBuilder.Build (app), "first app build should have succeeded.");
+
+				lib2.Sources.Add (new BuildItem.Source ("Class2.cs") {
+					TextContent = () => "namespace Library2 { public class Class2 { } }"
+				});
+
+				Assert.IsTrue (lib2Builder.Build (lib2, doNotCleanupOnUpdate: true), "second Library2 build should have succeeded.");
+				Assert.IsTrue (lib1Builder.Build (lib1, doNotCleanupOnUpdate: true, saveProject: false), "second Library1 build should have succeeded.");
+				appBuilder.Target = "SignAndroidPackage";
+				Assert.IsTrue (appBuilder.Build (app, doNotCleanupOnUpdate: true, saveProject: false), "app SignAndroidPackage build should have succeeded.");
+
+				var lib2Output = Path.Combine (path, lib2.ProjectName, "bin", "Debug", "netstandard2.0", $"{lib2.ProjectName}.dll");
+				var lib2InAppOutput = Path.Combine (path, app.ProjectName, app.IntermediateOutputPath, "android", "assets", $"{lib2.ProjectName}.dll");
+				FileAssert.AreEqual (lib2Output, lib2InAppOutput, "new Library2 should have been copied to app output directory");
+			}
+		}
+
+		[Test]
 		[Category ("dotnet")]
 		public void LinkAssembliesNoShrink ()
 		{
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
@@ -259,13 +259,17 @@ projects. .NET 5 projects will not import this file.
           Condition="'%(ReferenceCopyLocalPaths.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And '%(ReferenceCopyLocalPaths.Extension)' == '.dll' And '%(ReferenceCopyLocalPaths.RelativeDir)' == '' And Exists('%(ReferenceCopyLocalPaths.Identity)') "
       />
       <FilteredAssemblies
+          Include="@(_ReferenceDependencyPaths)"
+          Condition="'%(_ReferenceDependencyPaths.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And Exists('%(_ReferenceDependencyPaths.Identity)') "
+      />
+      <FilteredAssemblies
           Include="@(_ReferencePath)"
           Condition="'%(ReferencePath.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And Exists('%(_ReferencePath.Identity)') "
       />
     </ItemGroup>
     <!-- Find all the assemblies this app requires -->
     <ResolveAssemblies
-        Assemblies="@(FilteredAssemblies)"
+        Assemblies="@(FilteredAssemblies->Distinct())"
         I18nAssemblies="$(MandroidI18n)"
         LinkMode="$(AndroidLinkMode)"
         ProjectFile="$(MSBuildProjectFullPath)"
--- a/Documentation/release-notes/transitive-refasm.md
+++ b/Documentation/release-notes/transitive-refasm.md
@@ -1,0 +1,8 @@
+#### Application and library build and deployment
+
+- [Developer Community 1086457](https://developercommunity.visualstudio.com/content/problem/1086457/index.html):
+  Changes to libraries referenced by the .NET Standard library in a default
+  Xamarin.Forms project were not reflected in the running app without a clean
+  rebuild.  More generally, this issue affected any library referenced
+  indirectly via a .NET Standard library that had the
+  `ProduceReferenceAssembly` MSBuild property set to `true`.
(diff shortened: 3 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
```diff
@@ -223,7 +223,7 @@ projects. .NET 5 projects will not import this file.
       />
       <FilteredAssemblies
           Include="@(ReferenceCopyLocalPaths)"
-          Condition="'%(ReferenceCopyLocalPaths.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And '%(ReferenceCopyLocalPaths.Extension)' == '.dll' And '%(ReferenceCopyLocalPaths.RelativeDir)' == '' And Exists('%(ReferenceCopyLocalPaths.Identity)') "
+          Condition="'%(ReferenceCopyLocalPaths.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And '%(ReferenceCopyLocalPaths.Extension)' == '.dll' And Exists('%(ReferenceCopyLocalPaths.Identity)') "
```
brendanzagaeski, line 226: In the original commit that added the `%(ReferenceCopyLocalPaths.RelativeDir)' == ''` condition (https://github.com/xamarin/xamarin-android/commit/8f2ae24813ec3cdab56f688c4c3b611a33759333), it sounded like the condition wasn't a critical part of the functionality. Additionally, my understanding is that the mechanism for resolving the proper runtime assembly when given a reference assembly (such as for NuGet packages) has also been improved within `<ResolveAssemblies/>` since that time. So I believe that this condition can be removed without undesired consequences.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
```diff
@@ -514,6 +514,71 @@ public void ProduceReferenceAssembly ()
...
 
+		[Test]
+		public void TransitiveDependencyProduceReferenceAssembly ()
+		{
+			var path = Path.Combine (Root, "temp", TestName);
+			var app = new XamarinAndroidApplicationProject {
+				ProjectName = "App",
+				//NOTE: so _BuildApkEmbed runs in commercial tests
+				EmbedAssembliesIntoApk = true,
+				AndroidUseSharedRuntime = false,
```
jonathanpeppers, line 525: Does this test still work if you remove these settings? Just so it will have fast deployment enabled like most apps by default.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
jonathanpeppers: Yeah, I don't see why that condition would be there...

jonathanpeppers:
I changed:

```
Other changes:

  * The `BuildWithResolveAssembliesFailure` test using `packages.config`
    gets a slightly different error message due to the head project
    knowing about `Microsoft.Azure.EventHubs`. I think the new behavior
    is OK.
  * `@(FilteredAssemblies)` can now have duplicates, so we need to use
    `->Distinct()`. Note that MSBuild preserves ordering:
 
 https://github.com/microsoft/msbuild/blob/9c2d922959e80ba72fefb7addec21ede227639c3/src/Build/Evaluation/Expander.cs#L2484-L2513
```

jonpryor:
@brendanzagaeski: Looks like this (somehow) broke satellite assembly use? https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3894877&view=ms.vss-test-web.build-test-results-tab&runId=14457216&resultId=100003&paneView=debug

```
System.InvalidOperationException : Satellite assembly resource lookup failed for resource 'LibString' from culture 'fr-FR'.
Expected='Lib French'
Actual='Lib English'
```

brendanzagaeski:
Shoot, yup. Looking into that satellite assembly behavior now.

brendanzagaeski:

To fix the broken `SatelliteAssemblyValues` test, undo the original fix, and instead add the `@(ReferenceDependencyPaths)` items to `@(FilteredAssemblies)`:

```diff
+      <FilteredAssemblies
+          Include="@(ReferenceDependencyPaths)"
+          Condition="'%(ReferenceDependencyPaths.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And Exists('%(ReferenceDependencyPaths.Identity)') "
+      />
```

Update the commit message accordingly:

```diff
-To fix this behavior, include more of the items from
-`@(ReferenceCopyLocalPaths)` in the `@(FilteredAssemblies)` input to
-`<ResolveAssemblies/>` by removing the
-`'%(ReferenceCopyLocalPaths.RelativeDir)' == ''` condition.
+To fix this behavior, add the `@(ReferenceDependencyPaths)` items that
+are found by the `<ResolveAssemblyReference/>` task to the
+`@(FilteredAssemblies)` input.
```

_Explanation:_

Having satellite assemblies included in the input to `<ResolveAssemblies/>` causes trouble, so use a different item list (`@(ReferenceDependencyPaths)`) that still includes the desired transitive dependencies but doesn't include the satellite assemblies. The `@(ReferenceDependencyPaths)` item list is also nice because its name lines up closely with the scenario this PR aims to fix.

_Longer explanation:_

After the original fix, two satellite assemblies were being passed as additional inputs to `<ResolveAssemblies/>` in the failing test:

```
C:\Source\xamarin-android\bin\TestRelease\de-DE\LibraryResources.resources.dll
C:\Source\xamarin-android\bin\TestRelease\fr-FR\LibraryResources.resources.dll
```

`<ResolveAssemblies/>` did not expect this. It discarded the `de-DE` assembly entirely, but kept the `fr-FR` assembly and treated it just like any other non-satellite assembly.

As a consequence, `bin\TestRelease\fr-FR\LibraryResources.resources.dll` was copied to:

```
obj\Release\110\android\assets\LibraryResources.resources.dll
```

And then the linking step created:

```
obj\Release\110\android\assets\shrunk\LibraryResources.resources.dll
```

After that, the `<BuildApk/>` task placed the `shrunk` assembly into the APK, oddly preserving the `shrunk` subdirectory, so that the final path in the APK was:

```
assemblies/shrunk/LibraryResources.resources.dll
```

(This is part of the problem happens with non-satellite assemblies too, as a user reported a little while back: <https://github.com/xamarin/xamarin-android/issues/3745>.)

Finally, when the APK included the incorrect `assemblies/shrunk/LibraryResources.resources.dll` entry in addition to the correct `assemblies/fr-FR/LibraryResources.resources.dll` entry, that somehow interfered with the resource lookup process. I'll plan to investigate this a little part of the problem a little more tomorrow because the results I've seen so far are interesting. In one case, I got the test to fail in an APK where the only difference compared to the working APK was an additional `NOTICES` entry.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
```diff
@@ -258,14 +258,18 @@ projects. .NET 5 projects will not import this file.
           Include="@(ReferenceCopyLocalPaths)"
           Condition="'%(ReferenceCopyLocalPaths.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And '%(ReferenceCopyLocalPaths.Extension)' == '.dll' And '%(ReferenceCopyLocalPaths.RelativeDir)' == '' And Exists('%(ReferenceCopyLocalPaths.Identity)') "
       />
+      <FilteredAssemblies
+          Include="@(ReferenceDependencyPaths)"
+          Condition="'%(ReferenceDependencyPaths.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And Exists('%(ReferenceDependencyPaths.Identity)') "
+      />
       <FilteredAssemblies
           Include="@(_ReferencePath)"
           Condition="'%(ReferencePath.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And Exists('%(_ReferencePath.Identity)') "
```
jonathanpeppers, line 267: This is somewhat unrelated to this PR, but I see a small mistake here, should we just fix this since I noticed it?
```suggestion
          Condition="'%(_ReferencePath.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And Exists('%(_ReferencePath.Identity)') "
```

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
```diff
@@ -516,6 +516,68 @@ public void ProduceReferenceAssembly ()
...
+				Sdk = "Microsoft.NET.Sdk",
+				TargetFramework = "netstandard2.0",
+				Sources = {
+					new BuildItem.Source ("Class1.cs") {
+						TextContent = () => "namespace Library2 { public class Class1 { } }"
+					},
+				}
+			};
+			lib2.SetProperty ("ProduceReferenceAssembly", "True");
+			lib1.OtherBuildItems.Add (new BuildItem.ProjectReference ($"..\\{lib2.ProjectName}\\{lib2.ProjectName}.csproj", lib2.ProjectName, lib2.ProjectGuid));
```
jonpryor, line 556: Why does this use `lib1.OtherBuildItems` and not `lib1.References`?

brendanzagaeski: Oh yeah, that's a little tricky thing I learned when setting up this test. What happens is that `XmlUtils.ToXml()` _only_ adds `OtherBuildItems` in projects that have `$(EnableDefaultItems)`=`true`:

https://github.com/xamarin/xamarin-android/blob/99ec9f231bc52b4e9c451b955f2493517afc1fc2/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Utilities/XmlUtils.cs#L32-L51

And `DotNetStandard` has `$(EnableDefaultItems)`=`true`:

https://github.com/xamarin/xamarin-android/blob/99ec9f231bc52b4e9c451b955f2493517afc1fc2/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetStandard.cs#L42

brendanzagaeski:
_brendanzagaeski commented on Jul 16 <https://github.com/xamarin/xamarin-android/pull/4887#issuecomment-659881458>_

> Finally, when the APK included the incorrect `assemblies/shrunk/LibraryResources.resources.dll` entry in addition to the correct `assemblies/fr-FR/LibraryResources.resources.dll` entry, that somehow interfered with the resource lookup process. I'll plan to investigate this a little part of the problem a little more tomorrow because the results I've seen so far are interesting. In one case, I got the test to fail in an APK where the only difference compared to the working APK was an additional `NOTICES` entry.

* * *

I got a chance to dig into this more and found that the difference for the failing APKs was related to `$(AndroidEnableAssemblyCompression)` and whether the satellite assemblies were compressed or not. I put the details in a new issue <https://github.com/xamarin/xamarin-android/issues/4967>. Long story short, that behavior isn't a concern for this PR anymore because this PR no longer adds satellite assemblies to any extra item lists.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
```diff
@@ -258,14 +258,18 @@ projects. .NET 5 projects will not import this file.
           Include="@(ReferenceCopyLocalPaths)"
           Condition="'%(ReferenceCopyLocalPaths.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And '%(ReferenceCopyLocalPaths.Extension)' == '.dll' And '%(ReferenceCopyLocalPaths.RelativeDir)' == '' And Exists('%(ReferenceCopyLocalPaths.Identity)') "
       />
+      <FilteredAssemblies
+          Include="@(ReferenceDependencyPaths)"
+          Condition="'%(ReferenceDependencyPaths.ResolvedFrom)' != 'ImplicitlyExpandDesignTimeFacades' And Exists('%(ReferenceDependencyPaths.Identity)') "
+      />
```
jonathanpeppers, line 264: I think this needs to use `@(_ReferenceDependencyPaths)`:

https://github.com/xamarin/xamarin-android/blob/eb6ea0b55c89ce7e0b107c72900be9d019fb4b5b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L438-L442

This way we will include the `%(AndroidSkip...)` metadata, it matches the use of `@(_ReferencePath)` below.

dellis1972: @brendanzagaeski I think with this change ^ we can probably merge this. Its looking good.

brendanzagaeski: Ooo, nice catch. I've added that change now.

brendanzagaeski:

Change `ReferenceDependencyPaths` to `_ReferenceDependencyPaths` so that the items include the `%(AndroidSkip...)` metadata.

jonathanpeppers:
Something weird happened when I merged this... GitHub failed, then displayed a retry button. When I clicked retry, it lost what I edited in the commit message?

Issue #2441, opened 2018-11-19
Remove use of EmbeddedResources for res, assets etc. 


dellis1972:
In order to better support reference assemblies and faster incremental builds we need to stop using `EmbeddedResource` to handle moving res, assets etc between projects.

The problem with `EmbeddedResource` is that we cannot easily detect what has been updated in a library project. It might only be one file or just some C# code. In order to find that out we need to extract the entire zip and process it. 

The new method will place the required resources/assets etc next to the final assembly. This format will be Nuget Friendly in that the normal Nuget packaging with "just work".  We will also autogenerate a `.targets` file which can be used to make sure the required files and folders are included in the build. This `.targets` will need to be in a `build` directory in order for it to be picked up by nuget. Other files need to be in a `content` folder or some other folder which nuget will automatically pick up. 

```
MyProject.dll
content\assets\*
content\res\*
content\lib\*
build\MyProject.targets
```

This layout will be updated as the idea is expanded upon.

dellis1972:
@Redth it would be helpful to know the current folder layout used in the Support/Google Nuget packages (for Aar files) to make sure we are on the right track.

Redth:
@dellis1972 nothing is released with these yet, so blank slate here.

I think we still want to use `AndroidAarLibrary` for some packages which require the downloading of them with Xamarin.Build.Download, but if we're bundling things inside the nupkg otherwise, I guess it doesn't hurt to have the .aar extracted already into folders which you are suggesting.

I think if we do this it should follow the exact same layout as an aar file itself, just under some directory.  I'm thinking it might make sense to nest these under an `aar` directory just to be explicit at what this layout is?

Keep in mind we should support multitargeting in this scenario so we will want to have something like:

```
lib\MonoAndroid90\Some.dll
build\MonoAndroid90\Some.targets
content\MonoAndroid90\aar\assets\*
content\MonoAndroid90\aar\res\*
content\MonoAndroid90\aar\AndroidManifest.xml
content\MonoAndroid90\aar\annotations.zip
content\MonoAndroid90\aar\classes.jar
content\MonoAndroid90\aar\proguard.txt
content\MonoAndroid90\aar\public.txt
content\MonoAndroid90\aar\R.txt
```

Redth:
Actually, another thought, perhaps `AndroidAarLibrary` could be updated to support a directory path of unzipped aar contents as well, and in doing so you could use both this nupkg layout as well as .aar files themselves...

[...]

PR #4896, opened 2020-07-07
[One .NET] exclude MSBuild targets for .dll.config files

jonathanpeppers:
Running under .NET 5+, the `BuildBasicApplicationCheckConfigFiles`
test fails with:

    UnnamedProject.dll.config was must be copied to Intermediate directory
    Expected: True
    But was: False

After review, it doesn't seem like Xamarin.Android needs to support
`.dll.config` files in .NET 5+ at all.

1. `<dllmap>` is not supported. It was a feature of Mono.
2. `appSettings` only work via a compat NuGet package:

```xml
<PackageReference Include="System.Configuration.ConfigurationManager" Version="4.7.0" />
```

This NuGet package does not work in the current Xamarin.Android
(throws `PlatformNotSupportedException`), and there is no other
mechanism to support `appSettings`.

Any MSBuild targets related to `.config` files have been moved to

I added `XA1024` to warn going forward and updated tests around the
`_CopyConfigFiles` MSBuild target to run under `dotnet` context.

### DIFF
--- a/Documentation/guides/messages/xa1024.md
+++ b/Documentation/guides/messages/xa1024.md
@@ -1,0 +1,36 @@
+---
+title: Xamarin.Android warning XA1024
+description: XA1024 warning code
+ms.date: 07/08/2020
+---
+# Xamarin.Android warning XA1024
+
+## Example messages
+
+```
+warning XA1024: Ignoring configuration file 'Foo.dll.config'. .NET configuration files are not supported in Xamarin.Android projects that target .NET 5 or higher.
+```
+
+## Issue
+
+No support for [configuration files][config] such as `Foo.dll.config`
+or `Foo.exe.config` is available in Xamarin.Android projects targeting
+.NET 5 or higher. [`<dllmap>`][dllmap] configuration elements are not
+supported in .NET 5 at all, and other element types for compatibility
+packages like [System.Configuration.ConfigurationManager][nuget] have
+never been supported in Xamarin.Android projects.
+
+[config]: https://docs.microsoft.com/dotnet/framework/configure-apps/
+[nuget]: https://www.nuget.org/packages/System.Configuration.ConfigurationManager/
+
+## Solution
+
+Use a supported alternative for [`<dllmap>`][dllmap], such as the
+[NativeLibrary][native] API.
+
+Use an alternative for application settings such as
+[Preferences][preferences] in Xamarin.Essentials.
+
+[dllmap]: https://github.com/dotnet/coreclr/blob/master/Documentation/design-docs/dllmap.md
+[native]: https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.nativelibrary
+[preferences]: https://docs.microsoft.com/xamarin/essentials/preferences
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -1608,11 +1608,11 @@ namespace App1
 		}
 
 		[Test]
+		[Category ("dotnet")]
 		public void BuildBasicApplicationCheckConfigFiles ()
 		{
 			var proj = new XamarinAndroidApplicationProject ();
-			using (var b = CreateApkBuilder ("temp/BuildBasicApplicationCheckConfigFiles", false)) {
-				b.Verbosity = Microsoft.Build.Framework.LoggerVerbosity.Diagnostic;
+			using (var b = CreateApkBuilder ()) {
 				var config = new BuildItem.NoActionResource ("UnnamedProject.dll.config") {
 					TextContent = () => {
 						return "<?xml version='1.0' ?><configuration/>";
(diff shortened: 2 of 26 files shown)


jonpryor:
> To make things work as they should:

Therein lies the question: how *should* `.dll.config` files work in a .NET 6 world?  *At present* -- and presumably for the foreseeable future! -- .NET 5+ mono no longer supports using `.dll.config` files for `<dllmap/>` purposes.

Thus, should they be included *at all*?

Would it instead make more sense to ignore them entirely, and perhaps emit a warning if they're found, though this may be "too much" in that there's nothing a dev could do about them if e.g. a NuGet package includes them...

jonathanpeppers:
@jonpryor `.config` files are used for other things besides `dllmap`.

You can put string values, for example: 
```xml
<?xml version="1.0" encoding="utf-8" ?>  
<configuration>  
  <appSettings>  
    <add key="Foo" value="Bar"/>
  </appSettings>  
</configuration>  
```

Accessed by:
```csharp
var foo = System.Configuration.ConfigurationManager.AppSettings["Foo"];
```
https://docs.microsoft.com/dotnet/api/system.configuration.configurationmanager.appsettings?view=net-5.0

This worked in .NET Framework, and it appears to be back in .NET 5.

It doesn't seem like it works in current Xamarin.Android, though.

akoeplinger:
> This worked in .NET Framework, and it appears to be back in .NET 5.

Note that the functionality is not provided by the inbox .NET libraries but rather through an out of box NuGet package that you need to explicitly reference.

This means it's more considered a compatibility thing rather than something you should use for new development.

jonathanpeppers:
So yeah it looks like I can use this in a .NET 5 console app:

https://www.nuget.org/packages/System.Configuration.ConfigurationManager/4.7.0

So perhaps we don't need this at all in Xamarin? Unless anyone else knows a usage of `.config` files we need to support, I will change this to move the targets to `Xamarin.Android.Legacy.targets` instead.

jonpryor:
@jonathanpeppers: the [`System.Configuration.ConfigurationManager`](https://docs.microsoft.com/en-us/dotnet/api/system.configuration.configurationmanager?view=dotnet-plat-ext-3.1) usage [you mentioned](https://github.com/xamarin/xamarin-android/pull/4896#issuecomment-654951843):

 1. only loads content from `.exe.config` files, *not* `.dll.config` files, and
 2. requires `System.Configuration.dll`, which is *not* part of the "mobile"/MonoAndroid profile, nor is it part of .NET Standard 2.0.

Which means the `ConfigurationManager.AppSettings["Foo"]` example is not relevant.

Thus, I believe that .NET 5+ doesn't need to retain `.dll.config` file support.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -1610,9 +1610,12 @@ public void BuildBasicApplicationCheckMdbAndPortablePdb ()
 		[Test]
 		public void BuildBasicApplicationCheckConfigFiles ()
 		{
+			if (Builder.UseDotNet) {
+				Assert.Ignore ("'.dll.config' files are not supported in .NET 5+.");
```
jonpryor, line 1614: Should this test be ignored under `dotnet`?  Or should it instead assert that the resulting `.apk` file doesn't contain `.dll.config` files?

Additionally, there is the [migration question](https://github.com/xamarin/xamarin-android/pull/4896#issuecomment-654944896): if an existing project contains a `.dll.config` file, should *anything else* be done in addition to "ignoring" the presence of the file?  Should we emit a diagnostic message, or a warning, or *anything*?

If we *do* decide to "do something", that "something" could be asserted in this test.

jonathanpeppers: I added a `XA1024` warning for any `.config` file found in `@(ResolvedFileToPublish)`.

Documentation/guides/messages/xa1024.md
```diff
@@ -0,0 +1,34 @@
...
+warning XA1024: Use of 'Foo.dll.config' is not supported in Xamarin.Android for .NET 5 and higher.
+```
+
+## Issue
+
+No mechanism for supporting `Foo.dll.config` or `Foo.exe.config` files
+exists in Xamarin.Android for .NET 5 and higher. [`<dllmap>`][dllmap]
+is not supported in .NET 5 at all. Compatibility packages like
+[System.Configuration.ConfigurationManager][nuget] have never been
+supported in Xamarin.Android either.
```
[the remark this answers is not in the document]
brendanzagaeski: One other question crossed my mind about `<assemblyBinding>` elements, but it looks like that probably doesn't need to be mentioned since I think those aren't used by Xamarin.Android legacy projects either, and it seems those elements also aren't supported in .NET 5 at all.

jonathanpeppers:
Context: https://xamarinhq.slack.com/archives/C03CEGRUW/p1594415651239600

> From @jonpryor: when (if?) should “normal” XA emit XA1024, to help assist future migrations to .NET 6?

If we emitted this in current stable Xamarin.Android:

> warning XA1024: Ignoring configuration file 'Foo.dll.config'. .NET configuration files are not supported in Xamarin.Android projects that target .NET 5 or higher.

The problem is I don't know what you do to make the warning go away? The [NativeLibrary][native] API isn't an option.

[native]: https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.nativelibrary

jonpryor:
> The problem is I don't know what you do to make the warning go away?

Exactly.

Perhaps the solution isn't to emit XA1024 at "reference" time -- when `App.csproj` has a `@(PackageReference)` to something that has `Lib.dll.config` -- but *instead* to have it at "NuGet packaging" time, when *creating* the NuGet package that contains `Lib.dll.config`.

The problem there is that we don't currently *have* a well-defined "NuGet packaging time" construct, though we kinda want one as part of/to mirror `dotnet pack`, perhaps alongside https://github.com/xamarin/xamarin-android/issues/2441 ?

> The [NativeLibrary](https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.nativelibrary) API isn't an option.

That certainly doesn't help matters. :-(

jonpryor:
```markdown
Running under .NET 5+, the
`BuildTest.BuildBasicApplicationCheckConfigFiles()` test fails with:

	UnnamedProject.dll.config was must be copied to Intermediate directory
	Expected: True
	But was: False

After review, it doesn't seem like Xamarin.Android needs to support
`.dll.config` files in .NET 5+ at all.

 1. `<dllmap/>` is not supported.  It was a feature of Mono.
    [`NativeLibrary`][0] is the replacement, but only exists in
    .NET Core 3.0+.

 2. `appSettings` only work via a compat NuGet package:

        <PackageReference Include="System.Configuration.ConfigurationManager" Version="4.7.0" />

    However, this NuGet package does not work in the current
    Xamarin.Android (it throws `PlatformNotSupportedException`), and
    there is no other mechanism to support `appSettings`.

All MSBuild targets related to `.config` files have been moved to

If a .NET 5+ build encounters a `.dll.config` file, an XA1024 warning
will be generated:

	warning XA1024: Ignoring configuration file 'Foo.dll.config'.
	.NET configuration files are not supported in Xamarin.Android projects that target .NET 5 or higher.

This is unfortunately an un-actionable warning message, so we may want
to reconsider this approach.

TODO/Possible answer: if we ever have a mechanism to support NuGet
package creation a'la `dotnet pack`, we could emit the XA1024 warning
at package creation if the `.dll.config` is included in the NuGet.

I added updated tests around the`_CopyConfigFiles` MSBuild target to
run under `dotnet` context.

[0]: https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.nativelibrary
```

PR #4900, opened 2020-07-07
[Xamarin.Android.Build.Tasks] deprecate %(ProjectReference.IsAppExtension)

jonathanpeppers:
Context: https://developer.android.com/training/wearables/apps/standalone-apps

Going forward in .NET 5+, we will be dropping support for embedding
Android Wear applications inside a "main" application.

From a `Foo.Android.csproj` you can:

    <ProjectReference Include="..\Foo.Wear\Foo.Wear.csproj">
      <Name>Wearable</Name>
      <IsAppExtension>True</IsAppExtension>
    </ProjectReference>

This would embed `com.foo.wear.apk` *inside* `com.foo.android.apk` in
`Resources/raw`.

We are removing this functionality, because it is not widely used. You
can distribute Android Wear as standalone applications since Android
Wear 2.0, so this should be the option going forward. I added a new
`XA4312` warning to be emitted for projects using this feature.

I moved the relevant MSBuild targets to
`Xamarin.Android.Wear.targets`, which is only imported by "legacy"
Xamarin.Android projects. .NET 5+ projects will not even import this
file.

I also added a `Wear` NUnit category to be used for excluding tests
down the road.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -655,6 +655,11 @@ In this message, the term "bundled" is a short way of saying "included into the 
     <comment>The following are literal names and should not be translated: Wear, APK, SignAndroidPackage
 "Wear" is a short version of the full product name "Wear OS" and so should not be translated.</comment>
   </data>
+  <data name="XA4312" xml:space="preserve">
+    <value>Referencing the Android Wear application project '{0}' from an Android application project is deprecated and will no longer be supported in a future version of Xamarin.Android. Remove the Android Wear application project reference from the Android application project and distribute the Wear application as a standalone application instead.</value>
+    <comment>The following are literal names and should not be translated: Android Wear, Android, Wear.
+{0} - The referenced Android Wear project.</comment>
+  </data>
   <data name="XA5101" xml:space="preserve">
     <value>Missing Android NDK toolchains directory '{0}'. Please install the Android NDK.</value>
     <comment>{0} - The path of the missing directory</comment>
--- a/Documentation/guides/DotNet5.md
+++ b/Documentation/guides/DotNet5.md
@@ -102,6 +102,15 @@ If Java binding is enabled with `@(InputJar)`, `@(EmbeddedJar)`,
 `@(LibraryProjectZip)`, etc. then `$(AllowUnsafeBlocks)` will default
 to `True`.
 
+Referencing an Android Wear project from an Android application will
+not be supported:
+
+```xml
+<ProjectReference Include="..\Foo.Wear\Foo.Wear.csproj">
+  <IsAppExtension>True</IsAppExtension>
+</ProjectReference>
+```
+
 [rids]: https://docs.microsoft.com/dotnet/core/rid-catalog
 [abet-sys]: https://github.com/xamarin/xamarin-android/issues/4127
 
--- a/Documentation/guides/messages/README.md
+++ b/Documentation/guides/messages/README.md
@@ -152,6 +152,7 @@ ms.date: 01/24/2020
 + [XA4309](xa4309.md): 'MultiDexMainDexList' file '{file}' does not exist.
 + [XA4310](xa4310.md): \`$(AndroidSigningKeyStore)\` file \`{keystore}\` could not be found.
 + XA4311: The application won't contain the paired Wear package because the Wear application package APK is not created yet. If building on the command line, be sure to build the "SignAndroidPackage" target.
++ [XA4312](xa4312.md): Referencing an Android Wear application project from an Android application project is deprecated.
 
 ## XA5xxx: GCC and toolchain
 
--- a/Documentation/guides/messages/xa4312.md
+++ b/Documentation/guides/messages/xa4312.md
@@ -1,0 +1,29 @@
+---
+title: Xamarin.Android error/warning XA4312
+description: XA4312 error/warning code
+ms.date: 07/07/2020
+---
+# Xamarin.Android error/warning XA4312
+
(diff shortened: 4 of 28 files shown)


jonpryor:
> Going forward in .NET 5+, we will be dropping support for embedding Android Wear applications inside a "main" application.

Rationale is from the [Android docs](https://developer.android.com/training/wearables/apps/standalone-apps):

> A watch APK ***should*** be distributed using that feature; do not embed a watch APK in a phone APK.

Emphasis mine.

This strongly supports that our existing "embedding Android Wear applications" support can be removed.

One unfortunate aspect are all the caveats in that document; it's a bit of a mess:

> Generally, the minimum and target API level for a standalone app, and for Wear 2.0, is level 25…
> 
> **Note**: Due to a latency issue affecting an app's availability on Wear 1.x watches, if you build a standalone Wear 2.0 APK and will continue to have a Wear 1.0 APK, do both of the following:
>
> * Provide a standalone version of the Wear APK, and
> * Continue embedding a version of the Wear APK in your phone APK

So while you shouldn't embed Wear apps, you may need to?

> **Caution**:  If you publish an update to your existing, production phone APK that has removed an embedded Wear APK, production users who update the phone APK before installing your standalone Wear APK will lose their existing Wear app and its data. If you publish an update to your existing, production phone APK, continue to embed your watch APK into that phone APK.

Clear as mud.

I'm not at all sure what we should tell our customers here, but we'll have a year+ to figure out the details before .NET 6 is stable…

---

In the meantime, it feels like there are *two separate* warning/error messages we should start emitting -- warnings for Xamarin.Android 11.x, errors for .NET 5+:

  * A warning/error when the Wear app's minSdkVersion is < 25, directing the developer to update the app to be a Wear 2.0 app with minSdkVersion >= 25, and
  * A warning/error when `%(ProjectReference.IsAppExtension)` is True

jonathanpeppers:
@jonpryor for this one:

> A warning/error when the Wear app's minSdkVersion is < 25, directing the developer to update the app to be a Wear 2.0 app with minSdkVersion >= 25

How do we detect a project is a Wear app? Should MSBuild targets in the Xamarin.Android.Wear NuGet package be the thing that emits this warning? It seems like this NuGet package should require API >= 25.

Additionally, should we unlist Xamarin.Android.Wear 1.x packages from NuGet?

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -655,6 +655,10 @@ In this message, the term "bundled" is a short way of saying "included into the
     <comment>The following are literal names and should not be translated: Wear, APK, SignAndroidPackage
 "Wear" is a short version of the full product name "Wear OS" and so should not be translated.</comment>
   </data>
+  <data name="XA4312" xml:space="preserve">
+    <value>Embedding an Android Wear application inside an Android application is deprecated. Distribute the Wear application as a standalone application instead.</value>
```
brendanzagaeski, line 659: To brainstorm on wording a little, maybe:

```
Referencing an Android Wear application project from an Android application project is deprecated and will no longer be supported in a future version of Xamarin.Android. Remove the Android Wear application project reference from the Android application project and distribute the Wear application as a standalone application instead.
```

Motivation: I did a quick search to see if "embed" or "embedding" appeared on any of the Android Wear docs on docs.microsoft.com. It looks like those terms don't currently appear there, so maybe it's simplest to describe the scenario in generic project referencing terms.

If the project reference terminology feels too imprecise, another option could be to use a phrase like "packaging ... as a resource" to align with the terminology on <https://docs.microsoft.com/xamarin/android/wear/deploy-test/packaging>.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -655,6 +655,10 @@ In this message, the term "bundled" is a short way of saying "included into the
...
+    <comment>The following are literal names and should not be translated: Android, Wear.</comment>
```
brendanzagaeski, line 660: For the sake of being extra explicit to minimize any questions the localization might have, I'll suggest including "Android Wear" as a literal name too:

```
<comment>The following are literal names and should not be translated: Android Wear, Android, Wear.</comment>
```

jonpryor:
Commit message?

```markdown
Context: https://developer.android.com/training/wearables/apps/standalone-apps

Going forward in .NET 5+, we will be dropping support for embedding
Android Wear applications inside a "main" application.

From a `Foo.Android.csproj` you can:

	<ProjectReference Include="..\Foo.Wear\Foo.Wear.csproj">
	  <IsAppExtension>True</IsAppExtension>
	</ProjectReference>

This would embed `com.foo.wear.apk` *inside* `com.foo.android.apk` in
`Resources/raw`.

We are removing this functionality in .NET 5+ because Google
[states that][0]:

> A watch APK should be distributed using [multiple APKs];
> do not embed a watch APK in a phone APK.

This advice is the recommendation for Android Wear 2.0+ (API-25+),
released on 2016-Oct-19, or ~3.5+ years ago.  Given that standalone
APK files have been the recommendation for such a long period of time,
we see no reason to continue supporting the Wear 1.0 apps when
building apps with .NET 5+.

Xamarin.Android 11.x will continue to support the "embedding" Wear 1.0
apps into their referencing `App.apk` packages, but doing so will now
elicit an XA4312 warning message, so that developers can begin
migrating their projects for eventual support under .NET 6.

	warning XA4312: Referencing an Android Wear application project from an Android application project is deprecated.

I moved the relevant MSBuild targets to `Xamarin.Android.Wear.targets`,
which is only imported by "legacy" Xamarin.Android projects.
.NET 5+ projects will not even import this file.

`WearTests.cs` can now run under `dotnet` context.
```

PR #4901, opened 2020-07-07
[tests] don't delete NuGet.config files during MSBuild tests

jonathanpeppers:
Running under .NET 5+, the `AndroidResourceChange` test would always
fail with:

    The target _ResolveLibraryProjectImports should have been skipped.

It turns out our test infrastructure was deleting `NuGet.config` on
incremental builds. This triggers the `Restore` target to do its work
again on every build changing `$(ProjectAssetsFile)`. This is an input
of `_ResolveLibraryProjectImports`:

    <Target Name="_ResolveLibraryProjectImports"
        Inputs="$(ProjectAssetsFile)...

We shouldn't ever delete `NuGet.config` files during MSBuild tests.
This allowed the test to pass, and we can now run it under `dotnet`.

The same was happening for `build.log`, `msbuild.binlog`, and
`process.log`, so I thought it would be good to skip deletion of
`.log` or `.binlog` files.

I also cleaned up the test a little to use `AssertTargetIsSkipped`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
@@ -1077,6 +1077,7 @@ namespace Lib2
 		}
 
 		[Test]
+		[Category ("dotnet")]
 		public void AndroidResourceChange ()
 		{
 			var proj = new XamarinAndroidApplicationProject ();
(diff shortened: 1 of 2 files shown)


dellis1972:
I like this. There is also a problem where `.log` files get deleted during Incremental builds as well, unless we set the `doNotCleanOnUpdate` flag. Should we skip `.log` files too?

jonathanpeppers:
> I like this. There is also a problem where .log files get deleted during Incremental builds as well, unless we set the doNotCleanOnUpdate flag. Should we skip .log files too?

Yeah let me audit every file that is deleted and create an exclude list.

PR #4904, opened 2020-07-09
[mac] Add support for Apple silicon

grendello:
**NOTE: THIS PR WILL NOT BUILD WITH EXPECTED OUTPUT UNTIL MONO ARCHIVE WITH UNIVERSAL BINARIES IS AVAILABLE**
**NOTE: TO BUILD THE UNIVERSAL BINARIES WE NEED XCODE 12 UNIVERSAL APPS BETA2 ON THE BOTS**

Context: https://github.com/mono/mono/pull/20014
Context: https://www.apple.com/newsroom/2020/06/apple-announces-mac-transition-to-apple-silicon/

Add support for building universal binaries for macOS host which
target the `x86-64` and the upcoming `ARM64` macOS architectures.

The build requires that `Xcode 12 Universal Apps beta2` (or newer)
is installed and is made the default Xcode with:

    $ sudo xcode-select -s /Applications/Xcode-beta.app

If the above step isn't done the build will still work but it will
produce just the `x86-64` binaries.


build-tools/cmake/xa_macros.cmake
```diff
@@ -77,3 +77,43 @@ macro(xa_common_prepare)
...
+        string(REPLACE "." ";" COMPILER_VERSION_LIST ${COMPILER_VERSION})
+        list(GET COMPILER_VERSION_LIST 0 COMPILER_VERSION_MAJOR)
+        message(STATUS "major compiler version: ${COMPILER_VERSION_MAJOR}")
+     else()
+        message(WARNING "Unable to detect macOS C/C++ compiler version")
+     endif()
+
+     set(COMPILER_SUPPORTS_ARM64 False)
+     if(COMPILER_VERSION_MAJOR)
+        if(${COMPILER_VERSION_MAJOR} GREATER_EQUAL 12)
```
jonpryor, line 103: A possible problem here is that we're assuming that if the compiler version is >= 12, then `-arch arm64` is supported.

The problem is that there are *two* Xcode 12 downloads:

  * **Xcode 12 beta 2**: 
    > This distribution of Xcode 12 beta does not include support for building macOS Universal apps. 
  * **Xcode 12 for macOS Universal Apps beta 2**:
    > This version of Xcode 12 beta is intended exclusively for building macOS Universal apps that run on Apple Silicon Macs

I haven't downloaded the former to verify that it reports a compiler version of 12, but I imagine it *would* report a compiler version of 12 while *not* supporting `clang -arch arm64`.

Is there a better/"more robust" way of determining that `clang -arch arm64` is supported?

jonpryor: Then again…?

My `/Applications/Xcode.app` is Xcode 11 (stable), with `clang --version` reporting:

```
Apple clang version 11.0.3 (clang-1103.0.32.62)
```

Does Xcode 11 support `-arch arm64`?  Kinda!

```bash
$ /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -o h hw.c -arch arm64 \
-isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include/sys/cdefs.h:807:2: error: Unsupported architecture
#error Unsupported architecture
 ^
```

Note that this isn't from the compiler; it's from a header file.  If we use `clang -isysroot` using an Xcode 12 path, it "works":

```bash
$ /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang -o h hw.c -arch arm64 \
    -isysroot /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.0.sdk
ld: warning: ignoring file /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/11.0.3/lib/darwin/libclang_rt.osx.a, missing required architecture arm64 in file /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/11.0.3/lib/darwin/libclang_rt.osx.a (3 slices)
$ echo $?
0
```

It works in that the exit code is 0.  There is a warning about `libclang_rt.osx.a`, but the file *is* created, and contains arm64 output:

```bash
$ file h
h: Mach-O 64-bit executable arm64
```

Which is somewhat bizarre.

jonpryor: Which is all to suggest that maybe a "`clang` version check" is the wrong course of action here, and perhaps instead check for a *file* within the SDK which suggests that arm64 is supported?

Swift documentation appears to be per-ABI, resulting in files such as:

```
/Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.0.sdk/System/iOSSupport/usr/lib/swift/CoreNFC.swiftmodule/arm64.swiftdoc
```

Xcode 11 has a `CoreNFC.swiftmodule/x86_64.swiftdoc`, but no `arm64.swiftdoc` file.

Alternatively, the `SDKSettings.json` file mentions it:

```
/* /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.0.sdk//SDKSettings.json */
…
"DeploymentTargetSettingName":"MACOSX_DEPLOYMENT_TARGET","Archs":["x86_64","x86_64h","arm64"]
…
```

It's all one line, but the Xcode 11 `SDKSettings.json` file doesn't contain *any* mention of arm64`, meaning `grep` may be sufficient:

```
$ grep arm64 /Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX11.0.sdk//SDKSettings.json 2>&1 > /dev/null && echo yes
yes
```

build-tools/cmake/xa_macros.cmake
```diff
@@ -77,3 +77,52 @@ macro(xa_common_prepare)
...
+   if(APPLE)
+     set(SDK_SUPPORTS_ARM64 False)
+     set(SDK_SUPPORTS_X86_64 False)
+     execute_process(
+       COMMAND xcode-select -p
+       RESULT_VARIABLE XCODE_SELECT_RESULT
+       OUTPUT_VARIABLE XCODE_DEVELOPER_PATH
+     )
+     if(NOT ${XCODE_SELECT_RESULT} EQUAL "0")
+       message(WARNING "xcode-select failed with result ${XCODE_SELECT_RESULT}")
```
jonpryor, line 91: Shouldn't this *also* include `{XCODE_DEVELOPER_PATH}`, which contains stdout & stderr?  Or will stdout & stderr be printed *anyway*?

grendello: stdout won't be printed, stderr will. If `xcode-select` failed then `XCODE_DEVELOPER_PATH` will be empty anyway:

```
$ xcode-select -pp 2> /dev/null
$ xcode-select -pp > /dev/nullxcode-select: error: invalid argument '-pp'
Usage: xcode-select [options]
```

build-tools/cmake/xa_macros.cmake
```diff
@@ -77,3 +77,51 @@ macro(xa_common_prepare)
...
+  if(APPLE)
+    set(SDK_SUPPORTS_ARM64 False)
+    set(SDK_SUPPORTS_X86_64 False)
+    execute_process(
+      COMMAND xcode-select -p
+      RESULT_VARIABLE XCODE_SELECT_RESULT
+      OUTPUT_VARIABLE XCODE_DEVELOPER_PATH
+      )
+    if(NOT ${XCODE_SELECT_RESULT} EQUAL "0")
+      message(WARNING "xcode-select failed with result ${XCODE_SELECT_RESULT}")
```
jonpryor, line 91: As per the [cmake `execute_process()` docs](https://cmake.org/cmake/help/latest/command/execute_process.html), `RESULT_VARIABLE` is an *integer*, the *exit code* of the process, while `OUTPUT_VARIABLE` is the unified stdout+stderr of the process run.

Question: when `execute_process()` is run, does it *also* print stdout/stderr to *`cmake`s stdout* -- i.e. our AzDO log file -- so that we can see the result of the executed command?

If `xcode-select -p` fails, we'll want to know *why* it failed, which means we'll need access to stdout/stderr.  If `execute_process()` always "forwards" this output, we're fine.  If it *doesn't* "forward" this output, then we need to ensure that this `message()` includes it:

```cmake
message(WARNING "xcode-select failed with result ${XCODE_SELECT_RESULT} output ```${XCODE_DEVELOPER_PATH}```")
```

build-tools/cmake/xa_macros.cmake
```diff
@@ -77,3 +77,51 @@ macro(xa_common_prepare)
...
+      set(SDKSETTINGS_PATH "${XCODE_DEVELOPER_PATH}/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/SDKSettings.plist")
+
+      # CAUTION: do NOT ever remove the '-o -' parameter, without '-o' plutil will overwrite the .plist file
+      execute_process(
+		COMMAND plutil -extract SupportedTargets.macosx.Archs json -o - "${SDKSETTINGS_PATH}"
+        RESULT_VARIABLE PLUTIL_RESULT
+        OUTPUT_VARIABLE SDK_ARCHITECTURES
+		)
+      if(NOT ${PLUTIL_RESULT} EQUAL 0)
+        message(WARNING "plutil failed to read ${SDKSETTINGS_PATH}, returned with result ${PLUTIL_RESULT}")
```
jonpryor, line 103: Ditto, if `{SDK_ARCHITECTURES}` contains `plutil` stdout/stderr, it should also be provided here if possible.

build-tools/cmake/xa_macros.cmake
```diff
@@ -77,3 +77,51 @@ macro(xa_common_prepare)
...
+    else()
+      string(STRIP "${XCODE_DEVELOPER_PATH}" XCODE_DEVELOPER_PATH)
+      set(SDKSETTINGS_PATH "${XCODE_DEVELOPER_PATH}/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/SDKSettings.plist")
+
+      # CAUTION: do NOT ever remove the '-o -' parameter, without '-o' plutil will overwrite the .plist file
+      execute_process(
+		COMMAND plutil -extract SupportedTargets.macosx.Archs json -o - "${SDKSETTINGS_PATH}"
+        RESULT_VARIABLE PLUTIL_RESULT
+        OUTPUT_VARIABLE SDK_ARCHITECTURES
+		)
```
jonpryor, line 101: Indentation here is inconsistent.

build-tools/cmake/xa_macros.cmake
grendello:
```$ ninja -j1 -v
ninja: warning: bad deps log signature or version; starting over
[0/1] /Users/grendel/android-toolchain/sdk/cmake/3.10.2.4988404/bin/cmake -H/Users/grendel/vc/xamarin/xamarin-android/src/monodroid -B/Users/grendel/vc/xamarin/xamarin-android/src/monodroid/obj/Debug/host-Darwin-Debug
-- BIN_DIR exists
xcode-select: error: invalid argument '-pp'
Usage: xcode-select [options]

Print or change the path to the active developer directory. This directory
controls which tools are used for the Xcode command line tools (for example, 
xcodebuild) as well as the BSD development commands (such as cc and make).

Options:
  -h, --help                  print this help message and exit
  -p, --print-path            print the path of the active developer directory
  -s <path>, --switch <path>  set the path for the active developer directory
  --install                   open a dialog for installation of the command line developer tools
  -v, --version               print the xcode-select version
  -r, --reset                 reset to the default command line tools path
CMake Warning at /Users/grendel/vc/xamarin/xamarin-android/build-tools/cmake/xa_macros.cmake:91 (message):
  xcode-select failed with result 1
Call Stack (most recent call first):
  CMakeLists.txt:95 (xa_macos_prepare_arm64)
```

jonpryor:
CMP0066 & CMP0067 docs:

  * https://cmake.org/cmake/help/v3.10/policy/CMP0066.html
  * https://cmake.org/cmake/help/v3.10/policy/CMP0067.html

build-tools/cmake/xa_macros.cmake
grendello: Behavior is the same as with `xcode-select`

build-tools/cmake/xa_macros.cmake
grendello: Updated

src/sqlite-xamarin/CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
 cmake_minimum_required(VERSION 3.6.0)
 
+#
+# MUST be included before project()!
+#
+include("../../build-tools/cmake/xa_common.cmake")
```
jonpryor, line 6: I suspect that this is basically a no-op, as there is no `libsqlite3_xamarin.dylib` in my build tree.  I think this is only built for Android.

jonpryor:
Commit message:

```markdown
Context: https://github.com/mono/mono/pull/20014
Context: https://www.apple.com/newsroom/2020/06/apple-announces-mac-transition-to-apple-silicon/

***NOTE: THIS WILL NOT FULLY SUPPORTABLE BUILD UNTIL MONO ARCHIVE WITH
   UNIVERSAL BINARIES IS AVAILABLE***

***NOTE: TO BUILD THE UNIVERSAL BINARIES WE NEED XCODE 12 UNIVERSAL APPS
   BETA2 ON THE BOTS***

Add support for building the following macOS native libraries as
[universal binaries][0] on macOS hosts, which target the `x86-64` and
the forthcoming `arm64` macOS architectures:

  * `libmono-android.debug.dylib`
  * `libmono-android.release.dylib`

Support for creating universal binaries requires:

  * Building on macOS Catalina 10.15.4 or later
  * Building with [Xcode 12 for macOS Universal Apps][1]

Once **Xcode 12 for macOS Universal Apps** has been installed, it must
be made the default Xcode toolchain via:

	$ sudo xcode-select -s /Applications/Xcode-beta.app

If `xcode-select` isn't using Xcode 12 for macOS Universal Apps, then
the built macOS native libraries will only support x86-64.

Additionally, [CMAKE_POLICY_DEFAULT_CMP0066][2] and
[CMAKE_POLICY_DEFAULT_CMP0067]][3] are enabled when building under
CMake 3.7 or later so that our `c_compiler_has_flag()` and related
checks continue to work as intended on more recent CMake versions.

[0]: https://developer.apple.com/documentation/xcode/building_a_universal_macos_binary
[1]: https://developer.apple.com/documentation/xcode-release-notes/xcode-12-for-macos-universal-apps-beta-release-notes/
[2]: https://cmake.org/cmake/help/v3.10/policy/CMP0066.html
[3]: https://cmake.org/cmake/help/v3.10/policy/CMP0067.html
```

PR #4905, opened 2020-07-09
[One .NET] fixes for IncrementalClean and FileWrites

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/blob/792d61d116cc0293bd457d1c418d1972829ec226/Documentation/guides/MSBuildBestPractices.md#filewrites-and-incrementalclean

Running under `dotnet`, a test is failing with:

    GenerateJavaStubsAndAssembly(True)
    android\environment.armeabi-v7a.o is not in FileWrites!
    Expected: some item equal to "C:\A_work\1\s\bin\TestRelease\temp\GenerateJavaStubsAndAssemblyTrue\obj\Release\android\environment.armeabi-v7a.o"

When I authored `Microsoft.Android.Sdk.BuildOrder.targets`, I didn't
bring over the workaround to make `IncrementalClean` and `FileWrites`
work appropriately:

    <CoreBuildDependsOn>
      $([MSBuild]::Unescape($(CoreBuildDependsOn.Replace('IncrementalClean;', '$(_BeforeIncrementalClean);IncrementalClean;'))))
    </CoreBuildDependsOn>

It appears this is still needed in the new `dotnet` world order. I
brought over the missing pieces from `Xamarin.Android.Legacy.targets`.

We also need to specify the `armeabi-v7a` ABI in this test, because
the default ABIs are different in .NET 5+.

I added some helper methods to check `Builder.UseDotNet` to decide if
`$(RuntimeIdentifier)` should be used instead of `$(AndroidSupportedAbis)`.
I updated a couple tests that should use these helpers.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
@@ -27,6 +27,12 @@ projects, these properties are set in Xamarin.Android.Legacy.targets.
       _CopyPackage;
       _Sign;
     </BuildDependsOn>
+    <_BeforeIncrementalClean>
+      _PrepareAssemblies;
+      _CompileDex;
+      $(_AfterCompileDex);
+      _AddFilesToFileWrites;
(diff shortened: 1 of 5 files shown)


jonathanpeppers:
Eventually we can use https://github.com/microsoft/msbuild/pull/5499 here instead.

PR #4910, opened 2020-07-10
[ci] Add Anti-Malware Scan to code analysis job

pjcollins:
Context: https://www.1eswiki.com/wiki/AntiMalware_Scan_Build_Task
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1139581

Adds the Anti-Malware Scan build task to our code analysis job.  This
tool uses Windows Defender to scan the specified folder for malware,
and will ensure that we don't check any offending binaries into source
control.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -91,6 +91,7 @@ stages:
       displayName: Run AntiMalware (Defender) Scan
       inputs:
         FileDirPath: $(System.DefaultWorkingDirectory)
+        EnableServices: true
       condition: succeededOrFailed()
 
     - task: securedevelopmentteam.vss-secure-development-tools.build-task-report.SdtReport@1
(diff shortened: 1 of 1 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -87,6 +87,13 @@ stages:
       parameters:
         exclusionFile: $(System.DefaultWorkingDirectory)\build-tools\automation\PoliCheckExclusions.xml
 
+    - task: securedevelopmentteam.vss-secure-development-tools.build-task-antimalware.AntiMalware@3
```
jonpryor, line 90: What is `@3`?  I've seen `@yaml` before to pull in files from xamarin/yaml-templates, but `@3` is new to me and this file, so I'm not at all sure where this comes from or where to find the source.

pjcollins: `@3` is the major version of the task, which allows for changes to the task to be released without breaking existing usage. Every task that is using a full name (not a shorthand name like `script` or `powershell`) will have a trailing `@$version`. This is seen elsewhere in the pipeline:
https://github.com/xamarin/xamarin-android/blob/8f5d32969c70f3347af019fac57df28139cd3a23/build-tools/automation/azure-pipelines.yaml#L214-L236

Issue #4893, opened 2020-07-02
Android app bundle (.aab) crashes on startup when using Android 9.0 emulator


codympearson:
### Steps to Reproduce

1. Set your Android project as the startup project
2. Select the `Release` configuration
3. In your Android project settings, under Android Build > General, set the package format to `aab`
4. Under Advanced, ensure x86 is enabled as a supported ABI
4. Build and deploy your app to a simulator running Android 9.0 (Pie) - x86
5. The app freezes on the splash screen (before OnCreate)
6. Logs show the app is throwing a mono exception, crashing, and restarting

As a note, a crash does not occur on Android 7.1, 8.1, and 10.0.

### Expected Behavior
The app does not crash on launch when using `aab` on Android 9.0.

### Actual Behavior
The app crashes on Android 9.0 after a mono exception and appears to freeze on the launch screen.

### Version Information
=== Visual Studio Community 2019 for Mac ===

Version 8.6.5 (build 23)
Installation UUID: 569a2032-b97f-4417-8a14-45ffe1b65f41
	GTK+ 2.24.23 (Raleigh theme)
	Xamarin.Mac 6.18.0.23 (d16-6 / 088c73638)

	Package version: 610000104

=== Mono Framework MDK ===

Runtime:
	Mono 6.10.0.104 (2019-12/5d03a6fe116) (64-bit)
	Package version: 610000104

=== Roslyn (Language Service) ===

3.6.0-3.20210.9+4eafdcb1bcbd8d3573f2ba6065e56d9b9ce4f8a3

=== NuGet ===

Version: 5.6.0.6591

=== .NET Core SDK ===

SDK: /usr/local/share/dotnet/sdk/3.1.301/Sdks
SDK Versions:
	3.1.301
	3.1.300
	3.1.200
	3.1.101
	3.1.100
	3.0.101
	3.0.100
	2.1.701
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/6.10.0/lib/mono/msbuild/Current/bin/Sdks

=== .NET Core Runtime ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	3.1.5
	3.1.4
	3.1.2
	3.1.1
	3.1.0
	3.0.1
	3.0.0
	2.1.19
	2.1.18
	2.1.16
	2.1.15
	2.1.14
	2.1.13
	2.1.12

=== Xamarin.Profiler ===

Version: 1.6.12.29
[...]

PR #4914, opened 2020-07-13
Work around an Android 9 dlopen bug on x86

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/4893

This commit fixes an issue in Android 9 (API 28) 32-bit (ONLY) dynamic
loader issue where if the same library is loaded more than once, instead
of returning a handle to the already loaded copy, `dlopen` will
reinitialize the library thus resetting its state (BSS section is
cleared, global instances of C++ classes are re-instantiated etc).

When the library is reinitalized, however, our runtime is not - since
the Java `Runtime.initialize`, which ends up in our native code,
is *not* called because the library load is not part of the process
startup, when `Runtime.initialize` is invoked.  This in turn leads to
situation when certain class members initialized in `Runtime.initialize`
and assumed to be correct throughout the application life are, in fact,
invalid and lead to crashes.  In the instance of issue #4893, the
problem was that the `MonodroidRuntime::Class_getName` field was not
initialized to point to the `Java.Lang.Class.getName` Java method and
that, in turn, led to Xamarin.Android not being able to look up Java
types during application execution time.

`libmonodroid.so` is reloaded whenever Mono tries to resolve an
`__Internal` p/invoke (via a call to Xamarin.Android's
`monodroid_dlopen` fallback handler) - these p/invokes are assumed to
live in `libmonodroid` so we need to return a handle to it, so that Mono
can subsequently use `dlsym` to look up the required symbol.  Given the
situation described in the previous paragraph, we need a way to not
reload `libmonodroid` or do something else to avoid it being
reinitialized.

Since there's no portable way to check whether a DSO is already
loaded we need to resort to "something else", implemented in this
commit.  We move all the functions used in `__Internal` p/invokes to a
separate DSO which can be reloaded & reinitialized over and over again
without shredding our global state.  This is done by way of introducing
a C++ interface class, `MonoAndroidExternalAPI` which is passed to the
new DSO on load time, thus giving it a way to invoke all the necessary
internal Xamarin.Android functions the `__Internal` p/invokes need in
order to work as expected.


build-tools/installers/create-installers.targets
```diff
@@ -159,6 +159,7 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\LayoutBinding.cs" />
     <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\%(Identity)\libmono-android.debug.so')"       Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
     <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\%(Identity)\libmono-android.release.so')"     Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
+    <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\%(Identity)\libmonodroid-mono-api.so')"       Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
```
jonathanpeppers, line 162: I think you need this new file in .NET 5+ packages:
```suggestion
    <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\%(Identity)\libmonodroid-mono-api.so')" />
```

jonathanpeppers: Actually list it here, and leave this line as-is:

https://github.com/xamarin/xamarin-android/blob/b987d1b6bdbef4f1e858ed2c6d2f69504f614243/build-tools/create-packs/Microsoft.Android.Runtime.proj#L39-L42

grendello: @jonathanpeppers I mirrored the action for `libmono-android*.so` - the two don't work without each other, so is the change you suggested also required for `libmono-android*.so`?

jonpryor, line 162: Naming "quibble": we've been "trying" (kinda/sorta) to keep "monodroid" out of filenames, and nominally been successful:

```bash
% find /Library/Frameworks/Xamarin.Android.framework/Versions/Current -iname \*monodroid\*
# No matches
```

(Within `.apk` naming is another matter…)

`src/monodroid` creates `libmono-android*` files, not `libmonodroid*` files.

I think we should continue this, and instead use `libandroid-mono-api.so` or something like that.

src/monodroid/jni/cpu-arch-detect.cc
```diff
@@ -225,7 +225,7 @@ get_running_on_cpu (unsigned short *running_on_cpu)
 	*running_on_cpu = CPU_KIND_UNKNOWN;
 }
 
-MONO_API void
```
jonpryor, line 228: Why remove `MONO_API`?  Is this going to become a non-"public" function?

src/monodroid/jni/external-api.cc
```diff
@@ -1,29 +1,69 @@
...
+constexpr int TRUE = 1;
+constexpr int FALSE = 0;
 
 using namespace xamarin::android;
-using namespace xamarin::android::internal;
+
+static MonodroidExternalAPI *external_api = nullptr;
+
+MONO_API bool
+__init_external_api (MonodroidExternalAPI *api)
```
jonpryor, line 14: This is a name reserved by the compiler (even if something frequently used by others).  Any particular reason for a double-underscore `__` prefix and not `_monodroid_`?

src/monodroid/jni/cpu-arch-detect.cc
grendello: Yes, it was moved to `external-api.cc` and it is not expected to be called externally on `libmonodroid.so`

src/monodroid/jni/monodroid-glue-external.hh
```diff
@@ -0,0 +1,67 @@
...
+#include <jni.h>
+
+#include "monodroid.h"
+#include "logger.hh"
+#include "timing.hh"
+#include "xamarin_getifaddrs.h"
+
+namespace xamarin::android
+{
+	class MonodroidExternalAPI
```
jonpryor, line 16: What I find _fascinating_ about this is that it *feels like* a reversion of 6be4bdbcfd9f2fbe6267aceb934394bcdd0f13b4, which is "Stop using dlopen/dlsym to load Mono symbols."  It's *not* -- there are far fewer `dlsym()` calls with this approach -- but it *feels* like it, and I just find that...funny.

jonpryor: Unfortunately, it means that instead of a direct call to certain functions, it's a *virtual* call to certain functions.  *Some* overhead, but not much.

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1079,30 +1080,49 @@ MonodroidRuntime::monodroid_dlopen_log_and_return (void *handle, char **err, con
 		delete[] full_name;
 	}
 
+	if (!need_api_init)
+		return handle;
+
+	auto api = new MonodroidExternalAPI_Impl ();
```
jonpryor, line 1086: I believe this is a possible memory leak: there's a `new` here, but no corresponding `delete external_api` anywhere in this PR.

If I understand the scenario correctly, *every time* we `dlopen("libxa-mono-api.so")`, Android will *clear* the memory used for `static`/global variables of that native lib (the BSS section), and without some form of `_fini()` in `libxa-mono-api.so` -- assuming Bionic even supports `_init`/`_fini`/`__attribute((destructor))`/etc.  (See also the [**dlopen**(3) man page](https://linux.die.net/man/3/dlopen)?).

Without a way to free the instance that `libxa-mono-api.so!external_api` refers to, every *subsequent* `dlopen("libxa-mono-api.so")` will thus leak the `MonodroidExternalAPI_Impl` allocated from the *previous* `dlopenI("libxa-mono-api.so")` invocation.

A possible fix is to *avoid* `new`/`delete`, and instead use a global:

```cc
static MonodroidExternalAPI_Impl _monodroid_ext_api;

void*
MonodroidRuntime::monodroid_dlopen_log_and_return (…)
{
    …
    if (!api_init (&_monodroid_ext_api)) {
        …
    }
    …
}
```

This way we only ever create *one* instance of `MonodroidExternalAPI_Impl`, and it's never leaked.

src/monodroid/jni/monodroid-glue-external.hh
grendello: There are no virtual calls involved, not really. Note that this class is pure virtual, it's an interface class, the one we derive from it, `MonodroidExternalAPI_Impl` is marked as `final` (like C# `sealed` really( and all of its methods are also `final`. The compiler will know it and will generate plain calls.

src/monodroid/jni/monodroid-glue.cc
grendello: This isn't a safe way. Yes, the bug is that the memory is cleared on this particular Android emulator image (and the scenario was that an `armeabi` app was running on an `x86` emulator). But there's no guarantee that this will always be the case (in fact, the app I tested before our `HelloWorld` sample, did reload the DSO, just rechecked it, and return a different value of its handle) and, even though we don't have any state in the API class **now**, I'd rather leak a tiny amount of memory than risk sharing state with an unknown number of the `DSO` instances.

grendello: I will, however, investigate what bionic supports wrt destructors

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1079,30 +1080,49 @@ MonodroidRuntime::monodroid_dlopen_log_and_return (void *handle, char **err, con
...
 void*
 MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[maybe_unused]] void *user_data)
 {
+	constexpr char api_dso[] = "libxa-mono-api.so";
+
 	int dl_flags = monodroidRuntime.convert_dl_flags (flags);
 	bool libmonodroid_fallback = false;
 
-	/* name is nullptr when we're P/Invoking __Internal, so remap to libmonodroid */
+	/* name is nullptr when we're P/Invoking __Internal, so remap to libmonodroid-mono-api */
```
jonpryor, line 1110: Comment is now out of date; needs to mention `libxa-mono-api.so`.

src/monodroid/jni/monodroid-glue.cc
jonpryor: How often is this codepath hit?  How much memory could we possibly leak?

Are we hitting this codepath *every time* we *invoke* a P/Invoke to a `__Internal` method, in which case we could be hitting this code path *thousands*-*millions* of times, which could be a significant memory leak?

Or do we hit this code path *every time* we *resolve* a P/Invoke to a `__Internal` method, in which case we could *at most* leak dozens of instances, *maybe*?  (Though each instance is probably on a separate code page, so that could still add up…)

What's our worst-case scenario here?

src/monodroid/jni/monodroid-glue-external.hh
```diff
@@ -0,0 +1,67 @@
...
+	{
+	public:
+		static constexpr char INIT_FUNCTION_NAME[] = "_monodroid_init_external_api";
+
+	public:
+		virtual mono_bool monodroid_get_network_interface_up_state (const char *ifname, mono_bool *is_up) = 0;
```
jonpryor, line 22: Should this type have a public virtual destructor?

src/monodroid/jni/monodroid-glue-external.hh
jonpryor, line 16: Naming quibble (?): the scenario being addressed is around P/Invokes to `__Internal` functions, in which mono eventually calls `MonodroidRuntime::monodroid_dlopen()`, which eventually calls `dlopen("libxa-mono-api.so")`.

This entire codepath is around P/Invokes to `__Internal`.  This is in no way "external".  (Unless we're using "external" as a "mirror" to "internal", at which point my brain potentially starts hurting.)

Then we have PR #4757, in which -- to address .NET mono "hijacking" `__Internal` for its own use, preventing us from using it -- changes our P/Invokes to use `monodroid` instead of `__Internal`.

Wouldn't this just "reintroduce" the same problem, as a `[DllImport("monodroid")]` would still result in a subsequent `monodroid_dlopen("monodroid")`, which would clear BSS, and break everything?

Furthermore, the whole reason we can't merge #4757 yet is *because* by using `[DllImport("monodroid")]`, we *break* desktop use, as there is no `monodroid.dll`/`libmonodroid.dylib` file for the Designer to load.

---

I thus suggest the following, in broad strokes:

 1. "Merge" this PR & PR #4757.
 2. As part of the merging, *instead of* `[DllImport("monodroid")]`, we do `[DllImport("xa-internal-api")]`, which in turn means we need to *build* `libxa-internal-api.dylib`/etc for desktop platforms.  (Or maybe `xa-private`, to "mirror" `System.Private.CoreLib`/etc.?)
 3. `libxa-mono-api.so` *is* `libxa-internal-api.so`.  
    (On further thought, I'm not really sure why `mono` is in the library name, when *most* of the member function are for XA use, not mono…)
 4. *This* type, `MonodroidExternalAPI`, should instead be `MonoAndroidInternalCalls` (kinda/sorta "mirroring" `Mono.Android.dll`).

src/monodroid/jni/external-api.cc
```diff
@@ -1,29 +1,69 @@
...
+_monodroid_init_external_api (MonodroidExternalAPI *api)
+{
+	if (api == nullptr)
+		return false;
+
+	external_api = api;
+	return true;
+}
+
+/* !DO NOT REMOVE! Used by Mono BCL (System.Net.NetworkInformation.NetworkInterface) */
```
jonpryor, line 23: Not relevant to this PR, but since I'm curious: how *does* the .NET 5 "hijacking" of `__Internal` mentioned in PR #4757 impact *this* function?  How does `__Internal` *work* in .NET 5, in a way that *we* are even in the picture?  (*Are* we in the picture?)  What happens with `System.Net.NetworkInformation.NetworkInterface`?

src/monodroid/jni/monodroid-glue.cc
jonpryor: > But there's no guarantee that this will always be the case…

I don't understand this response.

In the solution of this PR, we have two libraries in the `.apk` of interest:

  * `libmonodroid.so`
  * `libxa-mono-api.so`

We're setting things up such that we *only ever* load `libmonodroid.so` *once*, by using `System.loadLibrary("monodroid")` in the *Java* bootstrap code.  All C/C++ `dlopen()` invocations, meanwhile, *never* load `libmonodroid.so`, but instead *always* load `libxa-mono-api.so`.

In this setup, the suggested `_monodroid_ext_api` would be loaded in `libmonodroid.so`.  Thus, the object would only ever be created once, and would never move around in memory, and would never have it's BSS cleared.

The loading of `libxa-mono-api.so`, meanwhile, would *itself* contain only *one* pointer, which would be "reset" whenever we `dlopen("libxa-mono-api.so")` via our invocation of `libxa-mono-api.so!_monodroid_init_external_api`.

> in fact, the app I tested before our HelloWorld sample, did reload the DSO, just rechecked it, and return a different value of its handle

*Which* DSO was reloaded?  I'm *assuming* you mean `libxa-mono-api.so`, right?  So what would it matter if the address returned by `dlopen("libxa-mono-api.so")` differs?  We don't *need* to allocate a different `MonodroidExternalAPI_Impl` instance *per `libxa-mono-api.so` loaded*; all loaded `libxa-mono-api.so` libraries can all use the *same* instance, no?

> I'd rather leak a tiny amount of memory than risk sharing state with an unknown number of the DSO instances.

Given the nature of things, I'm not sure that there's much risk to this sharing of state.

grendello: > > But there's no guarantee that this will always be the case…
> 
> I don't understand this response.
> 
> In the solution of this PR, we have two libraries in the `.apk` of interest:
> 
>     * `libmonodroid.so`
> 
>     * `libxa-mono-api.so`
> 
> 
> We're setting things up such that we _only ever_ load `libmonodroid.so` _once_, by using `System.loadLibrary("monodroid")` in the _Java_ bootstrap code. All C/C++ `dlopen()` invocations, meanwhile, _never_ load `libmonodroid.so`, but instead _always_ load `libxa-mono-api.so`.
> In this setup, the suggested `_monodroid_ext_api` would be loaded in `libmonodroid.so`. Thus, the object would only ever be created once, and would never move around in memory, and would never have it's BSS cleared.
> The loading of `libxa-mono-api.so`, meanwhile, would _itself_ contain only _one_ pointer, which would be "reset" whenever we `dlopen("libxa-mono-api.so")` via our invocation of `libxa-mono-api.so!_monodroid_init_external_api`.

1. Mono asks us to resolve `__Internal`, we dlopen the DSO, initialize it with the API, return the handle
2. The DSO somehow affects "state" of the API (e.g. one of the methods there needs to store some state and it's modified as a result of some call)
3. Mono asks us to resolve `__Internal` again (for instance for a different code path which hasn't been hit before), so we repeat 1. and 2.
4. *Both* instances created above are used by Mono at different times, stepping on each other's toes by modifying the state of the shared API class

> 
> > in fact, the app I tested before our HelloWorld sample, did reload the DSO, just rechecked it, and return a different value of its handle
> 
> _Which_ DSO was reloaded? I'm _assuming_ you mean `libxa-mono-api.so`, right? So what would it matter if the address returned by `dlopen("libxa-mono-api.so")` differs? We don't _need_ to allocate a different `MonodroidExternalAPI_Impl` instance _per `libxa-mono-api.so` loaded_; all loaded `libxa-mono-api.so` libraries can all use the _same_ instance, no?

Yes, `libxa-mono-api.so`, since this is the "DSO" featured by this PR. If the address differs, then we have TWO or more copies of the assembly, all using a SINGLE instance of the API which *may* have some state. the fact it doesn't have anything now doesn't matter - in the future we may introduce some state and get bitten by the shared nature of the class  if there are multiple instances of DSOs using it.

> 
> > I'd rather leak a tiny amount of memory than risk sharing state with an unknown number of the DSO instances.
> 
> Given the nature of things, I'm not sure that there's much risk to this sharing of state.

I disagree, as replied above.

grendello: > How often is this codepath hit? How much memory could we possibly leak?
> 
> Are we hitting this codepath _every time_ we _invoke_ a P/Invoke to a `__Internal` method, in which case we could be hitting this code path _thousands_-_millions_ of times, which could be a significant memory leak?
> 
> Or do we hit this code path _every time_ we _resolve_ a P/Invoke to a `__Internal` method, in which case we could _at most_ leak dozens of instances, _maybe_? (Though each instance is probably on a separate code page, so that could still add up…)
> 
> What's our worst-case scenario here?

From what I see, the `__Internal` DSO lookup is done once per assembly, so we'd leak N times where N equals the number of assemblies.

src/monodroid/jni/monodroid-glue-external.hh
grendello: In theory, yes, in practice, no. It's a pure interface class, no code here ever runs. And the only descendant is our Impl class, from which no further classes can be derived.

src/monodroid/jni/monodroid-glue-external.hh
grendello: > Naming quibble (?): the scenario being addressed is around P/Invokes to `__Internal` functions, in which mono eventually calls `MonodroidRuntime::monodroid_dlopen()`, which eventually calls `dlopen("libxa-mono-api.so")`.
> 
> This entire codepath is around P/Invokes to `__Internal`. This is in no way "external". (Unless we're using "external" as a "mirror" to "internal", at which point my brain potentially starts hurting.)

The `__Internal` name has been a misnomer forever and it's irrelevant here. The class is named `External` because it exposes external API functions from XA/Monodroid - the fact they are used for "internal" p/invokes is not relevant in this light IMO.

> 
> Then we have PR #4757, in which -- to address .NET mono "hijacking" `__Internal` for its own use, preventing us from using it -- changes our P/Invokes to use `monodroid` instead of `__Internal`.
> 
> Wouldn't this just "reintroduce" the same problem, as a `[DllImport("monodroid")]` would still result in a subsequent `monodroid_dlopen("monodroid")`, which would clear BSS, and break everything?

It would, and 4757 needs to be updated, obviously. This issue herewasn't known when #4757 was opened...

> 
> Furthermore, the whole reason we can't merge #4757 yet is _because_ by using `[DllImport("monodroid")]`, we _break_ desktop use, as there is no `monodroid.dll`/`libmonodroid.dylib` file for the Designer to load.

While this PR builds `libxa-mono-api.{dll,dylib}`... So we should be able to update #4757 to use `libxa-mono-api` instead of `monodroid`

> 
> I thus suggest the following, in broad strokes:
> 
>     1. "Merge" this PR & PR #4759 
We can do that, but this PR fixes a real user issue while #4757 is for the world 12 months from now really. I'm fine with it if you think it's ok to merge them, though

> 
>     2. As part of the merging, _instead of_ `[DllImport("monodroid")]`, we do `[DllImport("xa-internal-api")]`, which in turn means we need to _build_ `libxa-internal-api.dylib`/etc for desktop platforms.  (Or maybe `xa-private`, to "mirror" `System.Private.CoreLib`/etc.?)
Already builds for desktop

> 
>     3. `libxa-mono-api.so` _is_ `libxa-internal-api.so`.
>        (On further thought, I'm not really sure why `mono` is in the library name, when _most_ of the member function are for XA use, not mono…)
It's not internal API as far as XA is concerned - it's an export of API calls, so it's by all means external - to be used by external code. However, I don't really care enough about the name, I'll just update it to internal in a while.

> 
>     4. _This_ type, `MonodroidExternalAPI`, should instead be `MonoAndroidInternalCalls` (kinda/sorta "mirroring" `Mono.Android.dll`).
:+1:

src/monodroid/jni/external-api.cc
grendello: I have no idea what is the answer here. I haven't investigated the .NET5 `__Internal` stuff yet, not even sure if it's already decide how it will work in the end. However, I *think* `System.Net.NetworkInformation.NetworkInterface` will not use our internal code but some dotnetcore equivalent.

src/monodroid/jni/monodroid-glue-external.hh
jonpryor: > "Merge" this PR & PR #4759 

In retrospect, let's not do this.  Instead, let's keep it in mind: the implicit intent *here* is that we shouldn't necessarily be P/Invoking to `monodroid` or `__Internal`, but instead we should *eventually* be P/Invoking to… (1) *something* that we control, and (2) a "real" filename, not a "virtual" one, which is the problem with `monodroid` & PR #4759: there is no "real" `monodroid` library *unless* you're on-device, which isn't the case for the Designer.

In this case, `libxa-internal-api.so` should be our preferred "stable" P/Invoke target.

src/monodroid/jni/monodroid-glue-external.hh
jonpryor: Depending on how your "can we have a `_fini` function/`__attribute__((destructor))`-based cleanup code in `libxa-internal-api.so`?" investigation goes, we *may* want a virtual destructor, so that when `libxa-internal-api.so!internal_calls` is destructed/cleaned up, it'll do the correct thing.

For example, we could (maybe, perhaps, at least in spirit?) make `internal_calls` a `std::unique_ptr<MonoAndroidInternalCalls>`, and when the library is unloaded -- assuming it's unloaded or otherwise cleaned up! -- then the `MonoAndroidInternalCalls` destructor would be executed.

Needs more investigation, to be sure.

src/monodroid/jni/monodroid-glue.cc
jonpryor: Next question is *how much is leaked*?  Are we leaking a byte?  A *page*?  Multiple pages?  Unfortunately, this is likely based on **malloc**(3) behavior, and I don't know offhand what it's likely to do. :-/

We "only" likely to have app in the low-hundreds of assemblies (scary, but pull in Forms & AndroidX & NuGet and…), so if it's only once per assembly, this could either be an ignorable amount (1 byte per allocation) or a shockingly large amount (1 or more pages per allocation).

Maybe a custom allocator would be appropriate?

src/monodroid/jni/monodroid-glue-external.hh
grendello: From a very quick scan of the Mono source, I don't think Mono calls `dlclose` on the handle it gets from us during runtime, it closes the handle(s) only as part of the app shutdown cleanup. So I'm not sure we'd gain anything from `_fini` or `std::unique_ptr`.

src/monodroid/jni/monodroid-glue.cc
grendello: I don't think it's once per *every* assembly - it's more likely per assembly which has any `__Internal` p/invokes which right now is two assemblies (?) in our collection. Size of the class is 8 bytes. Bionic uses `jemalloc` which will round up the allocation size to at least `sizeof(double)` and the allocated objects are packed closely together avoid wasting space (http://jemalloc.net/jemalloc.3.html - see the `IMPLEMENTATION NOTES` section) so I really think we shouldn't go overboard with custom allocator and that we don't have too much to worry about in this regard

src/monodroid/CMakeLists.txt
```diff
@@ -306,6 +313,15 @@ if(UNIX)
     )
 endif()
 
+add_library(xa-internal-api SHARED ${XA_INTERNAL_API_SOURCES})
+target_link_libraries(xa-internal-api ${EXTRA_LINKER_FLAGS})
+set_target_properties(
+  xa-internal-api
+  PROPERTIES
+  COMPILE_FLAGS -fvisibility=default
+  LINK_FLAGS -fvisibility=default
```
jonpryor, line 322: Why is this `-fvisibility=default` instead of `-fvisibility=hidden`?

src/monodroid/jni/internal-pinvoke-api.cc
```diff
@@ -0,0 +1,333 @@
...
+MONO_API void
+_monodroid_counters_dump (const char *format, ...)
+{
+	va_list args;
+	va_start (args, format);
+	internal_calls->dump_counters (format, args);
+	va_end (args);
+}
+
+/* !DO NOT REMOVE! Used by libgdiplus.so */
```
jonpryor, line 200: *How* is this "used by `libgdiplus.so`"? Is this *actually* resolvable outside the "normal" `__Internal` mechanism?

(Do we even care?  The "`System.Drawing` for Xamarin.Android" was never really shipped or used, was it?)

src/monodroid/jni/internal-pinvoke-api.cc
```diff
@@ -0,0 +1,333 @@
...
+MONO_API int
+_monodroid_get_display_dpi (float *x_dpi, float *y_dpi)
+{
+	return internal_calls->monodroid_get_display_dpi (x_dpi, y_dpi);
+}
+
+MONO_API int
```
jonpryor, line 207: I'm not sure this is viable, which suggests we need a better "cross-reference mechanism".  For example, `set_assemblies_prefix()` is from Embeddinator-4000: 

  * https://github.com/mono/Embeddinator-4000/blob/d844acae5324fd6ac8eaf6c5c9c254b2b6a8a59e/support/mono_embeddinator.c#L56-L63
  * https://github.com/mono/Embeddinator-4000/blob/d844acae5324fd6ac8eaf6c5c9c254b2b6a8a59e/support/java/Runtime.java#L95
  * https://github.com/mono/Embeddinator-4000/blob/d844acae5324fd6ac8eaf6c5c9c254b2b6a8a59e/support/java/android/AndroidImpl.java#L26

It *looks like* Embeddinator-4000 implicitly requires that `monodroid_embedded_assemblies_set_assemblies_prefix()` be located in `monodroid`, because that's the only library it actually loads.

---

We *really* should use GitHub URLs, if at all possible, including commit hash. :-/

---

That said, a cursory investigation shows that Embeddinator-4000 has bit-rotten to the extent that it likely *can't* work with current xamarin-android -- `monodroid` can't be loaded until *after* `libmonosgen-2.0.so` is loaded, plus other dependencies! -- so perhaps it's not worth worrying about at all.

Which really proves that the only way for "Embeddinator-4000" to work reliably is for the core bits to be in xamarin-android, so that they won't get inadvertently broken.

src/monodroid/CMakeLists.txt
grendello: this DSO is supposed to have only external symbols, so it's here instead/in addition to `MONO_API`

src/monodroid/jni/internal-pinvoke-api.cc
grendello: I have no idea if it was ever shipped. I did the porting and handed it over. I've just checked sources, it uses `dlopen` to load `libmonodroid.so` explicitly, so it won't work after this PR.

src/monodroid/jni/internal-pinvoke-api.cc
```diff
@@ -0,0 +1,333 @@
...
+}
+
+MONO_API char**
+monodroid_strsplit (const char *str, const char *delimiter, size_t max_tokens)
+{
+	return internal_calls->monodroid_strsplit (str, delimiter, max_tokens);
+}
+
+MONO_API char*
+monodroid_strdup_printf (const char *format, ...)
```
jonpryor, line 238: Silly Q, but why is this (and others?) exported in the first place?  It's not used from `Mono.Android.dll`, and I can't find any uses of this symbol outside of `src/monodroid` anywhere on GitHub either: https://github.com/search?p=1&q=monodroid_strdup_printf&type=Code

src/monodroid/jni/internal-pinvoke-api.cc
```diff
@@ -0,0 +1,333 @@
...
+}
+
+MONO_API void
+monodroid_strfreev (char **str_array)
+{
+	internal_calls->monodroid_strfreev (str_array);
+}
+
+MONO_API char**
+monodroid_strsplit (const char *str, const char *delimiter, size_t max_tokens)
```
jonpryor, line 232: Ditto this symbol.

src/monodroid/jni/internal-pinvoke-api.cc
jonpryor: If it won't work after this PR, let's just remove the symbol entirely.

src/monodroid/jni/internal-pinvoke-api.cc
jonpryor: Which I suppose is a long way of saying "let's just remove this symbol as well".

src/monodroid/jni/internal-pinvoke-api.cc
grendello: I have no clue - I just moved things around, haven't checked if and where from they are still used.

src/monodroid/jni/internal-pinvoke-api.cc
grendello: Isn't this also used by the Desginer?

jonpryor:
Related: https://github.com/xamarin/designer/pull/3100

build-tools/installers/create-installers.targets
```diff
@@ -159,6 +159,7 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\LayoutBinding.cs" />
     <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\%(Identity)\libmono-android.debug.so')"       Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
     <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\%(Identity)\libmono-android.release.so')"     Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
+    <_MSBuildFiles Include="@(AndroidSupportedTargetJitAbi->'$(MSBuildSrcDir)\lib\%(Identity)\libxa-internal-api.so')"          Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
```
jonpryor, line 162: [Elsewhere](https://xamarinhq.slack.com/archives/C03CEGRUW/p1595283571077500), @garuma asked:

> I tried installing the .vsix from your PR and I see some libxa-internal-api.so but nothing similar for desktop


We need to update the installer to likewise install `libxa-internal-api.dylib` (macOS) and `libxa-internal-api.dll` (Windows).  (Related question: shouldn't it be `xa-internal-api.dll` on Windows, *without* the `lib` prefix?)

These files *are* built, as per the log file, they're just not in the installers.

grendello: Fixed the installers. Regarding the name, I considered it but all the other files have the `lib` prefix so I decided to keep it - it's just a convention and the Designer loads them directly with the `lib` prefix already, so I don't think there's a point in changing name of this new DLL.

src/monodroid/jni/internal-pinvoke-api.cc
```diff
@@ -0,0 +1,326 @@
...
+	internal_calls->monodroid_freeifaddrs (ifa);
+}
+
+MONO_API void
+_monodroid_detect_cpu_and_architecture (unsigned short *built_for_cpu, unsigned short *running_on_cpu, unsigned char *is64bit)
+{
+	internal_calls->monodroid_detect_cpu_and_architecture (built_for_cpu, running_on_cpu, is64bit);
+}
+
+/* !DO NOT REMOVE! Used by Mono BCL (System.Net.NetworkInformation.NetworkInterface) */
```
jonpryor, line 42: As per previous discussions, I would like to start having "full" references to what's using the code, so that if/when the referencing code is updated, we can more easily determine that a function is removable, e.g.

https://github.com/mono/mono/blob/e59c1cd70f4a7171a0ff5e1f9f4937985d6a4d8d/mcs/class/System/System.Net.NetworkInformation/LinuxNetworkInterface.cs#L250-L261

src/monodroid/jni/internal-pinvoke-api.cc
```diff
@@ -0,0 +1,326 @@
...
+}
+
+MONO_API int
+monodroid_embedded_assemblies_set_assemblies_prefix (const char *prefix)
+{
+	return internal_calls->monodroid_embedded_assemblies_set_assemblies_prefix (prefix);
+}
+
+MONO_API managed_timing_sequence*
+monodroid_timing_start (const char *message)
```
jonpryor, line 207: I find it somewhat amusing that some symbols start with `_monodroid`, and some start with `monodroid`, and I *think* the idea was that `_monodroid` was for "things internal to `Mono.Android.dll`."

By that logic, I think this should be `_monodroid`?  Or we just ignore this "weird" difference?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1418,6 +1451,45 @@ MonodroidRuntime::typemap_managed_to_java (MonoReflectionType *type, const uint8
 	return embeddedAssemblies.typemap_managed_to_java (type, mvid);
 }
 
+#if defined (WINDOWS)
+const char*
+MonodroidRuntime::get_my_location ()
+{
+	HMODULE hm = NULL;
+
+	if (GetModuleHandleEx (GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
```
jonpryor, line 1460: As per elsewhere, we should *avoid* use of the "macro WinAPI functions", and use the "real" functions directly.  Furthermore, we should avoid use of the `A`Functions, and should prefer the `W` functions.

As such:

```cpp
DWORD handle_flags = GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS | GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT;
if (GetModuleHandleExW (handle_flags, (LPCWSTR) &get_xamarin_android_msbuild_path, &hm) == 0) {
    …
}
```

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1418,6 +1451,45 @@ MonodroidRuntime::typemap_managed_to_java (MonoReflectionType *type, const uint8
...
+	                      GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
+	                      (LPCSTR) &get_xamarin_android_msbuild_path, &hm) == 0) {
+		int ret = GetLastError ();
+		log_warn (LOG_DEFAULT, "GetModuleHandle failed, error = %d\n", ret);
```
jonpryor, line 1464: Given that we're immediately returning, there is no "context" to this message. *Anything* could cause it.  Additionally, this message looks very similar to the `GetModuleFileName failed` message, meaning *I* will get confused when trying to deduce what went wrong. :-)

The messages should thus be updated to mention what's happening, e.g.

```cpp
log_warn (LOG_DEFAULT, "Unable to get HANDLE to `libmono-android.debug.dll`; GetModuleHandle returned %d\n", ret);
```

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1418,6 +1451,45 @@ MonodroidRuntime::typemap_managed_to_java (MonoReflectionType *type, const uint8
...
+		return nullptr;
+	}
+
+	WCHAR path[MAX_PATH * 2];
+	if (GetModuleFileNameW (hm, path, sizeof(path)) == 0) {
+		int ret = GetLastError ();
+		log_warn (LOG_DEFAULT, "GetModuleFileName failed, error = %d\n", ret);
```
jonpryor, line 1471: Given that we're immediately returning, there is no "context" to this message. *Anything* could cause it.  Additionally, this message looks very similar to the `GetModuleFileName failed` message, meaning *I* will get confused when trying to deduce what went wrong. :-)

The messages should thus be updated to mention what's happening, e.g.

```cpp
log_warn (LOG_DEFAULT, "Unable to get filename to `libmono-android.debug.dll`; GetModuleFileName returned %d\n", ret);
```

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1418,6 +1451,45 @@ MonodroidRuntime::typemap_managed_to_java (MonoReflectionType *type, const uint8
...
+		return nullptr;
+	}
+
+	PathRemoveFileSpecW (path);
```
jonpryor, line 1475: This *can* return an error: https://docs.microsoft.com/en-us/windows/win32/api/shlwapi/nf-shlwapi-pathremovefilespecw

No idea what that would *mean*, and the docs don't say how or why it can fail, but it does have a `BOOL` return type, which *presumably* we should be checking…

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1418,6 +1451,45 @@ MonodroidRuntime::typemap_managed_to_java (MonoReflectionType *type, const uint8
...
+
+	return utils.utf16_to_utf8 (path);
+}
+#elif defined (APPLE_OS_X)
+const char*
+MonodroidRuntime::get_my_location ()
+{
+	Dl_info info;
+	if (dladdr (reinterpret_cast<const void*>(&MonodroidRuntime::get_my_location), &info) == 0) {
+		log_warn (LOG_DEFAULT, "dladdr failed: %s", dlerror ());
```
jonpryor, line 1485: Same-but-different: warning message should provide more context:

```cpp
log_warn (LOG_DEFAULT, "Could not lookup library containing `MonodroidRuntime::get_my_location()`; dladdr failed: %s", dlerror ());
```

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1498,6 +1570,30 @@ MonodroidRuntime::Java_mono_android_Runtime_initInternal (JNIEnv *env, jclass kl
...
+#endif
+	}
+
+	if (api_dso_handle == nullptr) {
+		log_info (LOG_DEFAULT, "Attempting to load %s with \"bare\" dlopen", API_DSO_NAME);
+		api_dso_handle = dlopen (API_DSO_NAME, RTLD_NOW);
+	}
+
+	if (api_dso_handle == nullptr)
+#endif
```
jonpryor, line 1593: This *looks* weird, because of the indentation. I would suggest moving this `#endif` *above* the `if`, which means the Android build source would be:

```cpp
void *api_dso_handle = nullptr;
if (api_dsl_handle == nullptr)
    api_dso_handle = androidSystem.load_dso_from_any_directories (API_DSO_NAME, RTLD_NOW);
```

Which *looks* better.  Arguably there's an "extra null check" in here, but the compiler should be able to optimize it away anyway…

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1068,8 +1073,30 @@ MonodroidRuntime::convert_dl_flags (int flags)
...
-MonodroidRuntime::monodroid_dlopen_log_and_return (void *handle, char **err, const char *full_name, bool free_memory)
+force_inline void
+MonodroidRuntime::init_internal_api_dso (void *handle)
+{
+	if (handle == nullptr) {
+		log_fatal (LOG_DEFAULT, "Internal API library is required");
+		exit (FATAL_EXIT_MONO_MISSING_SYMBOLS);
+	}
+
+	auto api = new MonoAndroidInternalCalls_Impl ();
```
jonpryor, line 1084: The "memory leak" here should be explicitly called out in the commit message, and how we're fine with it.

src/monodroid/jni/internal-pinvoke-api.cc
grendello: I don't like the leading underscore, it makes for weird looking invocations like, in this case, `JNIEnv._monodroid_timing_start` - it looks just... weird

src/monodroid/jni/monodroid-glue.cc
grendello: The return value doesn't indicate an error. It tells us whether "file spec" (i.e. file name) was removed from the path or not. We don't really care so there's no point in checking that.

src/monodroid/jni/internal-pinvoke-api.cc
jonpryor: A benefit to moving them all into one file is that they're a bit more obvious.  That's nice.

src/monodroid/jni/monodroid-glue.cc
jonpryor: If I'm reading this right, the [idea on how to slightly limit the memory leak](https://discordapp.com/channels/732297728826277939/732297837953679412/736269643232182322) is to only incur the memory leak on Android 9.0 targets which have the dlopen "bug" in the first place, as per above `if (handle == api_dso_handle)` check.

While this does limit things -- and in a manner which doesn't explicitly check the API Level, which is nice if/when a vendor fixes the dlopen issue on their own fork -- it doesn't help *on Android 9 targets*.  Thus, Android 9 targets will still have a leak, and the constraints of this leak should be described.

grendello: The latest push should have plugged the leak entirely. I wanted to avoid this kind of "heavy" work, but with it limited to few devices and with the mutex held anyway, I think it's OK

jonpryor:
Commit message for review:

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/4893

The setup:

 1. App with `AndroidManifest.xml` which sets
    `/application/@android:extractNativeLibs` to `false`, e.g.:

        [assembly:Application (ExtractNativeLibs=false)]

 2. Build (1) as a *Release* app

 3. Run (2) on an Android 9.0 (API-28) *x86* emulator (32-bit!)

The result: app crash during process startup!

	E/mono(14126): Unhandled Exception:
	E/mono(14126): System.ArgumentNullException: Value cannot be null.
	E/mono(14126): Parameter name: key
	E/mono(14126):   at System.Collections.Generic.Dictionary`2[TKey,TValue].FindEntry (TKey key)
	E/mono(14126):   at System.Collections.Generic.Dictionary`2[TKey,TValue].TryGetValue (TKey key, TValue& value)
	E/mono(14126):   at Java.Interop.TypeManager.RegisterType (System.String java_class, System.Type t)
	E/mono(14126):   at Android.Runtime.JNIEnv.RegisterJniNatives (System.IntPtr typeName_ptr, System.Int32 typeName_len, System.IntPtr jniClass, System.IntPtr methods_ptr, System.Int32 methods_len)

The cause of the crash is an apparent Android bionic bug, a'la:

	void* h1 = dlopen ("…path/to/libmonodroid.so", RTLD_LAZY);
	// time passses
	void* h2 = dlopen ("…path/to/libmonodroid.so", RTLD_LAZY);

"Normal"/expected semantics are that `h1` is the same as `h2`, with
the second `dlopen()` invocation effectively being a no-op, as the
library has already been loaded.

On 32-bit Android 9 -- but *only* when `extractNativeLibs`=false! --
what *instead* happens is that that native library is *re-loaded* such
that `h1` != `h2` (potentially?!), *and* the global state of `h1` is
cleared; the library is reinitialized.

Where this comes in is with P/Invokes to `__Internal`, which we
"redirect" to `libmonodroid.so`.  However, since `dlopen(NULL)`
has a history on Android of generating a SIGSEGV, we avoid the crash
by explicitly loading `libmonodroid.so` every time we need to satisfy
a request for `__Internal` (in `MonodroidRuntime::monodroid_dlopen()`
when `name` is `nullptr`).  As we re-load `libmonodroid.so` and
`dlopen()` potentially *clears* global data, what happens is that
certain class members initialized in `Runtime.initInternal()` become
`nullptr`.  In this particular case, `MonodroidRuntime::Class_getName`
became null, which meant that we couldn't lookup Java class names,
which caused the subsequent crash.

The fix is to introduce a new `libxa-internal-api.so` which contains
only three things of consequence:

 1. Explicit "init" and "shutdown" exports:

        MONO_API bool _monodroid_init_internal_api (MonoAndroidInternalCalls *api);
        MONO_API bool _monodroid_shutdown_internal_api (void);

 2. A pointer to a new `MonoAndroidInternalCalls` "interface",
    implemented in `libmonodroid.so`, and

 3. Other "C" exports that `__Internal` needs to invoke, which are
    implemented by delegating through `MonoAndroidInternalCalls`.

`libxa-internal-api.so` thus becomes the *only* library that we ever
`dlopen()` more than once, and as we control when it's loaded, we can
call `_monodroid_init_internal_api()` and
`_monodroid_shutdown_internal_api()` as appropriate, ensuring that
when we run on Android 9+x86+`extractNativeLibs`=false, the "damage"
from subsequent `dlopen()` invocations is minimized.
```

PR #4920, opened 2020-07-15
[One .NET] drop support for DX

jonathanpeppers:
Context: https://android-developers.googleblog.com/2020/02/the-path-to-dx-deprecation.html

On February 1, 2021 DX will be removed from the Android SDK. It is
very possible this will happen before .NET 6 ships. There are also
currently a few MSBuild tests that fail when using DX.

* Drop support for DX in .NET 5+. Emit `XA1023` as an *error*.
* Don't even include `Xamarin.Android.DX.targets` in .NET 5+ packages.
* Add several tests to the `dotnet` category.
* Any tests using `AndroidDexTool=dx` will be ignored.

I left the `PackagingTest.CheckClassesDexIsIncluded` test so that it
verifies you get an `XA1023` build error when using DX with .NET 5+.

Other changes:

* There were some related warning messages that I moved to the
  `_CheckNonIdealConfigurations` MSBuild target. These seem like they
  should all be in the same MSBuild target.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -395,8 +395,12 @@ In this message, the term "binding" means a piece of generated code that makes i
 {0} - The file name of the library</comment>
   </data>
   <data name="XA1023" xml:space="preserve">
-    <value>Using the DX DEX Compiler is deprecated. Please update `$(AndroidDexTool)` to `d8`.</value>
-    <comment>The following are literal names and should not be translated: D8, DEX, `$(AndroidDexTool)`, `d8`</comment>
+    <value>Using the DX DEX Compiler is deprecated. Please set the DEX compiler to 'd8' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidDexTool' MSBuild property to 'd8'.</value>
+    <comment>The following are literal names and should not be translated: DX, DEX, d8, AndroidDexTool.</comment>
+  </data>
+  <data name="XA1023_dotnet" xml:space="preserve">
+    <value>Using the DX DEX Compiler is not supported in Xamarin.Android projects that target .NET 5 or higher. Please set the DEX compiler to 'd8' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidDexTool' MSBuild property to 'd8'.</value>
+    <comment>The following are literal names and should not be translated: DX, DEX, Xamarin.Android, .NET 5, d8, AndroidDexTool.</comment>
   </data>
   <data name="XA1024" xml:space="preserve">
     <value>Ignoring configuration file '{0}'. .NET configuration files are not supported in Xamarin.Android projects that target .NET 5 or higher.</value>
--- a/Documentation/guides/OneDotNet.md
+++ b/Documentation/guides/OneDotNet.md
@@ -104,6 +104,9 @@ previous Xamarin.Android releases.
 `$(AndroidClassParser)` will be `class-parse` by default. `jar2xml`
 will not be supported.
 
+`$(AndroidDexTool)` will be `d8` by default. `dx` will not be
+supported.
+
 `$(AndroidCodegenTarget)` will be `XAJavaInterop1` by default.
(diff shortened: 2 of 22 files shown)


src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -398,6 +398,10 @@ In this message, the term "binding" means a piece of generated code that makes i
     <value>Using the DX DEX Compiler is deprecated. Please update `$(AndroidDexTool)` to `d8`.</value>
     <comment>The following are literal names and should not be translated: D8, DEX, `$(AndroidDexTool)`, `d8`</comment>
   </data>
+  <data name="XA1023_DOTNET" xml:space="preserve">
```
brendanzagaeski, line 401: It's not important at all, but this could use other capitalization options too, like `XA1023_dotnet`, for example. I have a hunch the all-caps might have been following `XA0119_AAB`, which is perfectly good, but the other resource names have some additional variety in capitalization like `XA0000_TargetFrameworkVersion_for_API`, so there's flexibility to use other capitalizations too if desired.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -398,6 +398,10 @@ In this message, the term "binding" means a piece of generated code that makes i
...
+    <value>Using the DX DEX Compiler is not supported in Xamarin.Android projects that target .NET 5 or higher. Please update `$(AndroidDexTool)` to `d8`.</value>
+    <comment>The following are literal names and should not be translated: D8, DEX, Xamarin.Android, .NET 5, `$(AndroidDexTool)`, `d8`</comment>
```
brendanzagaeski, line 403: It's arguably a little weird to do it since it will temporarily make this message less similar to the existing XA1023 message, but I'll suggest using the "new" wording approach for how to change MSBuild properties:

```
<value>Using the DX DEX Compiler is not supported in Xamarin.Android projects that target .NET 5 or higher. Please set the DEX compiler to 'd8' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidDexTool' MSBuild property to 'd8'.</value>
<comment>The following are literal names and should not be translated: DX, DEX, Xamarin.Android, .NET 5, d8, AndroidDexTool</comment>
```

My thought is that the original XA1023 message can be updated to follow this new wording style too. If it's simplest, that change could be included in this PR, but I can also add it in a separate PR if that'll help keep this PR more on-topic.

Admittedly, I'm not sure if we know yet whether the IDE will display the DEX compiler option in the same way in the project property pages for Android projects in the .NET 5+ project system, but we can update that wording later if needed.

jonathanpeppers: 👍 I updated both `XA1023` messages.

jonathanpeppers:
The test failures look unrelated, one is network:
```
"/Users/runner/work/1/s/bin/TestRelease/temp/CheckNoVersionVectorsTrue/UnnamedProject.csproj" (Restore target) (1) ->
(Restore target) -> 
  /Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/NuGet.targets(124,5): error : Failed to retrieve information about 'Xamarin.Forms' from remote source 'https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/1a5f89f6-d8da-4080-b15f-242650c914a8/nuget/v3/flat2/xamarin.forms/index.json'. [/Users/runner/work/1/s/bin/TestRelease/temp/CheckNoVersionVectorsTrue/UnnamedProject.csproj]
/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/NuGet.targets(124,5): error :   No such host is known [/Users/runner/work/1/s/bin/TestRelease/temp/CheckNoVersionVectorsTrue/UnnamedProject.csproj]

    0 Warning(s)
    1 Error(s)
```
`Windows Build and Smoke Test` is fixed in 6675be8489cbf101e78ecc2694e64f06ad28a600.

PR #4923, opened 2020-07-15
[ci] Trim new line from multi-line MSBuild arguments

pjcollins:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3893162&view=logs&j=cac0e8d3-0ef5-5d2b-b57e-e8fde7204df3&t=ecb52562-b867-5d86-aae7-56a38766a1ce&l=170

The Java.Interop test step has recently started failing on our Windows
builds due to a strange error which seems to indicate that there is an
extra new line character in an invalid location.  I believe that this
has to do with the `>` character we're using to allow for blocks of
strings in various places.  Updating this to `>-` should hopefully trim
any trailing new lines that aren't intended according to this resource
for YAML multi-line string syntax https://yaml-multiline.info/.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -524,10 +524,7 @@ stages:
       inputs:
         solution: build-tools/check-boot-times/check-boot-times.csproj
         configuration: $(ApkTestConfiguration)
-        msbuildArguments: >
-          /restore
-          /t:Build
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
I restarted two of the stages that said:
```
##[error]We stopped hearing from agent Azure Pipelines 6. Verify the agent machine is running and has a healthy network connection. Anything that terminates an agent process, starves it for CPU, or blocks its network access can cause this error. For more information, see: https://go.microsoft.com/fwlink/?linkid=846610
```

PR #4925, opened 2020-07-15
[One .NET] fix design-time builds

jonathanpeppers:
Design-time builds were not working as expected under a `dotnet`
context, mostly due to differences in build ordering.

`dotnet` design-time builds would fail with:

    Xamarin.Android.Common.targets(577,2): Could not locate MonoAndroid SDK.

Reviewing the `_ResolveMonoAndroidSdks` MSBuild target, it depends on
values that the `_ResolveSdks` provides.

After updating `$(_ResolveMonoAndroidSdksDependsOn)` to include
`_ResolveSdks`, I hit:

    Xamarin.Android.Common.targets(716,2): error MSB4044: The "GetAndroidDefineConstants" task was not given a value for the required parameter "AndroidApiLevel".

`$(_AndroidApiLevel)` was blank, which is provided by the
`_ResolveAndroidTooling` target.

After updating `$(_ResolveMonoAndroidSdksDependsOn)` again, I hit:

    CSC error CS2001: Source file 'C:\src\xamarin-android\bin\TestDebug\temp\BuildAppWithManagedResourceParser\App1\obj\Release\Resource.designer.cs' could not be found.

It turns out using `$(AndroidUseIntermediateDesignerFile)` breaks
design-time builds. We set this by default for .NET 5+. The fix here
was to make the `_RemoveLegacyDesigner` target not run when
`$(ManagedDesignTimeBuild)` is true.

After these changes, I could add the `dotnet` category to a few tests
around design-time builds.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
@@ -99,7 +99,7 @@ namespace Xamarin.Android.Build.Tests
 		}
 
 		[Test]
-		[Category ("SmokeTests")]
+		[Category ("SmokeTests"), Category ("dotnet")]
 		public void DesignTimeBuild ([Values(false, true)] bool isRelease, [Values (false, true)] bool useManagedParser, [Values (false, true)] bool useAapt2)
 		{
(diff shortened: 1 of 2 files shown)


dellis1972:
@jonathanpeppers do we know what causes the error when ManagedDesignTimeBuild is on? We should fix that, it should work when `AndroidUseIntermediateDesignerFile` is true.

jonathanpeppers:
> do we know what causes the error when ManagedDesignTimeBuild is on?

It causes:

> CSC error CS2001: Source file 'C:\src\xamarin-android\bin\TestDebug\temp\BuildAppWithManagedResourceParser\App1\obj\Release\Resource.designer.cs' could not be found.

When the actual file is: `obj\Debug\designtime\Resource.designer.cs`. It seems like we just need to skip the `_RemoveLegacyDesigner` MSBuild target.

PR #4926, opened 2020-07-16
[build] bump to .NET 5.0.100-rc.1.20365.11

jonathanpeppers:
Context: https://github.com/dotnet/installer

This is still a "master" build, they are just using the `rc1` name after `preview7`.

This may break some tests using Xamarin.Forms, we'll see what happens.


jonathanpeppers:
It looks like everything worked, I was afraid this test might fail, but it doesn't use a Forms project:

https://github.com/xamarin/xamarin-android/blob/9c880983914e2dbc3f384dea16b5351696732682/tests/MSBuildDeviceIntegration/Tests/XASdkDeployTests.cs#L51-L59

I'll make other versions of this test down the road.

jonathanpeppers:
Closing in favor of #5007

Issue #4548, opened 2020-04-13
throwing a caught exception in View.Draw causes sigsev while debugging


roubachof:
NOTE: the SigSev only happens while debugging.
If in RELEASE on launch in DEBUG mode wo debugging, everything is neat.

I am here trying to create blur views for `Xamarin.Forms`, for the android implementation I chose https://github.com/mmin18/RealtimeBlurView.

I translated in C# all the classes, you can find them here:

https://github.com/roubachof/Sharpnado.Presentation.Forms/tree/secret/Sharpnado.Presentation.Forms.Droid/Renderers/MaterialFrame/RealtimeBlurView

In his java implementation, mmin18 has the following code:

https://github.com/mmin18/RealtimeBlurView/blob/master/library/src/com/github/mmin18/widget/RealtimeBlurView.java#L259-L267

For stoping the drawing process he throws a custom "STOP_EXCEPTION" in the `Draw` method of the blur view:

https://github.com/mmin18/RealtimeBlurView/blob/master/library/src/com/github/mmin18/widget/RealtimeBlurView.java#L321-L330

Remark: you have to interrupt the drawing of all the next children, so return in the blur view's Draw method is not enough....

It seems to be fine in plain android since I couldn't find an issue talking about SIGSEV and STOP_EXCEPTION in the original repo.

But in Xamarin.Android I get a sig sev exception:
```
Loaded assembly: System.Runtime.Serialization.dll [External]
**Sharpnado.Presentation.Forms.Droid.Renderers.MaterialFrame.RealtimeBlurView.RealtimeBlurView+StopException:** 'Exception of type 'Sharpnado.Presentation.Forms.Droid.Renderers.MaterialFrame.RealtimeBlurView.RealtimeBlurView+StopException' was thrown.'

04-13 14:29:09.259 D/Mono    ( 3665): DllImport attempting to load: '/system/lib/liblog.so'.
04-13 14:29:09.259 D/Mono    ( 3665): DllImport loaded library '/system/lib/liblog.so'.
04-13 14:29:09.260 D/Mono    ( 3665): DllImport searching in: '/system/lib/liblog.so' ('/system/lib/liblog.so').
04-13 14:29:09.260 D/Mono    ( 3665): Searching for '__android_log_print'.
04-13 14:29:09.260 D/Mono    ( 3665): Probing '__android_log_print'.
04-13 14:29:09.260 D/Mono    ( 3665): Found as '__android_log_print'.
[...]

PR #10, opened 2016-04-28
Xamarin.Android.Build.Tasks:_Foo - Set Inputs and Outputs to fix ..

radical:
.. incremental builds.

PR #4927, opened 2020-07-16
[debugger] Removing usage of mono_unhandled_exception

thaystg:
We don't need to call mono_unhandled_exception the runtime will handle with it.

Relates to https://github.com/mono/mono/pull/20314.
We don't need to wait for mono/mono PR to be merged, mono PR only removes the unhandled_exception function.

### DIFF
--- a/src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
+++ b/src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
@@ -7,7 +7,6 @@ using System.Threading;
 namespace Android.Runtime {
 	public static class JNINativeWrapper {
 
-		static MethodInfo? mono_unhandled_exception_method;
 		static MethodInfo? exception_handler_method;
 		static MethodInfo? wait_for_bridge_processing_method;
 
@@ -15,12 +14,6 @@ namespace Android.Runtime {
 		{
 			if (exception_handler_method != null)
 				return;
-#if MONOANDROID1_0
-			mono_unhandled_exception_method = typeof (System.Diagnostics.Debugger).GetMethod (
-				"Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
-			if (mono_unhandled_exception_method == null)
-				AndroidEnvironment.FailFast ("Cannot find System.Diagnostics.Debugger.Mono_UnhandledException");
-#endif
 			exception_handler_method = typeof (AndroidEnvironment).GetMethod (
 				"UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
 			if (exception_handler_method == null)
@@ -70,15 +63,8 @@ namespace Android.Runtime {
 			ig.Emit (OpCodes.Leave, label);
 
 			bool  filter = Debugger.IsAttached || !JNIEnv.PropagateExceptions;
-			if (filter && mono_unhandled_exception_method != null) {
-				ig.BeginExceptFilterBlock ();
 
-				ig.Emit (OpCodes.Call, mono_unhandled_exception_method);
-				ig.Emit (OpCodes.Ldc_I4_1);
-				ig.BeginCatchBlock (null!);
-			} else {
-				ig.BeginCatchBlock (typeof (Exception));
-			}
+			ig.BeginCatchBlock (typeof (Exception));
 
 			ig.Emit (OpCodes.Dup);
 			ig.Emit (OpCodes.Call, exception_handler_method!);
--- a/src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
+++ b/src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
@@ -51,26 +51,34 @@ namespace Android.Runtime {
 
 			ig.Emit (OpCodes.Call, wait_for_bridge_processing_method!);
 
-			var label = ig.BeginExceptionBlock ();
+			bool  filter = Debugger.IsAttached || !JNIEnv.PropagateExceptions;
+			if (!filter) {
+				var label = ig.BeginExceptionBlock ();
 
-			for (int i = 0; i < param_types.Length; i++)
-				ig.Emit (OpCodes.Ldarg, i);
-			ig.Emit (OpCodes.Call, dlg.Method);
+				for (int i = 0; i < param_types.Length; i++)
+					ig.Emit (OpCodes.Ldarg, i);
+				ig.Emit (OpCodes.Call, dlg.Method);
 
-			if (retval != null)
-				ig.Emit (OpCodes.Stloc, retval);
+				if (retval != null)
+					ig.Emit (OpCodes.Stloc, retval);
 
-			ig.Emit (OpCodes.Leave, label);
+				ig.Emit (OpCodes.Leave, label);
 
-			ig.BeginCatchBlock (typeof (Exception));
+				ig.BeginCatchBlock (typeof (Exception));
 
-			ig.Emit (OpCodes.Dup);
-			ig.Emit (OpCodes.Call, exception_handler_method!);
+				ig.Emit (OpCodes.Dup);
+				ig.Emit (OpCodes.Call, exception_handler_method!);
 
-			if (filter)
-				ig.Emit (OpCodes.Throw);
+				ig.EndExceptionBlock ();
+			}
+			else { //let the debugger handle the exception
+				for (int i = 0; i < param_types.Length; i++)
+					ig.Emit (OpCodes.Ldarg, i);
+				ig.Emit (OpCodes.Call, dlg.Method);
 
-			ig.EndExceptionBlock ();
+				if (retval != null)
+					ig.Emit (OpCodes.Stloc, retval);
+			}
 
 			if (retval != null)
 				ig.Emit (OpCodes.Ldloc, retval);
--- a/src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
+++ b/src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
@@ -62,8 +62,6 @@ namespace Android.Runtime {
 
 			ig.Emit (OpCodes.Leave, label);
 
-			bool  filter = Debugger.IsAttached || !JNIEnv.PropagateExceptions;
-
 			ig.BeginCatchBlock (typeof (Exception));
 
 			ig.Emit (OpCodes.Dup);
--- a/src/Mono.Android/Android.Runtime/JNIEnv.cs
+++ b/src/Mono.Android/Android.Runtime/JNIEnv.cs
@@ -237,18 +237,10 @@ namespace Android.Runtime {
 			GC.SuppressFinalize (obj);
 		}
 
-		static Action<Exception> mono_unhandled_exception = null!;
 		static Action<AppDomain, UnhandledExceptionEventArgs> AppDomain_DoUnhandledException = null!;
 
 		static void Initialize ()
 		{
-			if (mono_unhandled_exception == null) {
-				var mono_UnhandledException = typeof (System.Diagnostics.Debugger)
-					.GetMethod ("Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
-				if (mono_UnhandledException != null)
-					mono_unhandled_exception = (Action<Exception>) Delegate.CreateDelegate (typeof(Action<Exception>), mono_UnhandledException);
-			}
-
 			if (AppDomain_DoUnhandledException == null) {
 				var ad_due = typeof (AppDomain)
 					.GetMethod ("DoUnhandledException",
@@ -275,10 +267,6 @@ namespace Android.Runtime {
 			}
 
 			var javaException = JavaObject.GetObject<Java.Lang.Throwable> (env, javaExceptionPtr, JniHandleOwnership.DoNotTransfer)!;
-
-			// Disabled until Linker error surfaced in https://github.com/xamarin/xamarin-android/pull/4302#issuecomment-596400025 is resolved
-			//System.Diagnostics.Debugger.Mono_UnhandledException (javaException);
-			mono_unhandled_exception?.Invoke (javaException);
 
 			try {
 				var jltp = javaException as JavaProxyThrowable;
--- a/src/Mono.Android/Android.Runtime/UncaughtExceptionHandler.cs
+++ b/src/Mono.Android/Android.Runtime/UncaughtExceptionHandler.cs
@@ -7,8 +7,6 @@ namespace Android.Runtime {
 
 	sealed class UncaughtExceptionHandler : Java.Lang.Object, Java.Lang.Thread.IUncaughtExceptionHandler {
 
-		Action<Exception>? mono_unhandled_exception;
-
 		Action<AppDomain, UnhandledExceptionEventArgs>?      AppDomain_DoUnhandledException;
 
 		Java.Lang.Thread.IUncaughtExceptionHandler defaultHandler;
@@ -30,7 +28,6 @@ namespace Android.Runtime {
 				Android.Runtime.AndroidEnvironment.FailFast ($"Unable to initialize UncaughtExceptionHandler. Nested exception caught: {e}");
 			}
 
-			mono_unhandled_exception! (ex);
 			if (AppDomain_DoUnhandledException != null) {
 				try {
 					var jltp = ex as JavaProxyThrowable;
@@ -48,11 +45,6 @@ namespace Android.Runtime {
 
 		void Initialize ()
 		{
-			if (mono_unhandled_exception == null) {
-				var mono_UnhandledException = typeof (System.Diagnostics.Debugger)
-					.GetMethod ("Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
-				mono_unhandled_exception = (Action<Exception>) Delegate.CreateDelegate (typeof(Action<Exception>), mono_UnhandledException);
-			}
 
 			if (AppDomain_DoUnhandledException == null) {
 				var ad_due = typeof (AppDomain)


jonpryor:
@thaystg, @jonathanpeppers: "semantics" question: if you have a managed1 > managed2 call stack in which managed2 throws an exception which is caught by managed1:

```csharp
public void managed1 ()
{
    try {
        managed2 ();
    }
    catch (Exception e) {
        Console.WriteLine (e);
    }
}

public void managed1 ()
{
    throw new Exception ("Wee!");
}
```

what happens when ^^ is run within the debugger?  Will we get a "first chance exception" on the `throw new Exception("Wee!")` line?  Is a "last-chance notification" ever raised?  (Presumably not?)

How does this align with what .NET & Visual Studio do (no mono in the picture)?

jonathanpeppers:
Visual Studio on Windows _can_ break on this line:

```csharp
throw new Exception ("Wee!");
```

But only if you have the `Debug > Windows > Exception Settings` boxes checked appropriately:

image

VS for Mac probably has something similar?

jonpryor:
@jonathanpeppers: which indicates the *Visual Studio with .NET* has a first-chance exception, which is entirely expected. :-)

I'm not at all sure how to reasonably test the "last chance notification": https://docs.microsoft.com/en-us/windows/win32/debug/debugger-exception-handling

> The second attempt to notify the debugger occurs only if the system is unable to find a frame-based exception handler that handles the exception. This is known as *last-chance notification*.

i.e. if `managed1()` *didn't* have a `try`/`catch` block, we'd theoretically be hitting the last-chance notification scenario, but I don't know how to tell (if it's possible to tell?) the difference between the first-chance and last-chance exception.

Regardless, the question *here* isn't entirely about "VS with .NET", it's about "VS with *Mono* with @thaystg's changes".  What happens *there*?

thaystg:
I've tested with my PR using VSMac and mono, and VSMac can break on this line: 
`throw new Exception ("Wee!");`
Only if you have the Debug > Breakpoints > System.Exception checked
image

src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
```diff
@@ -51,26 +51,34 @@ public static Delegate CreateDelegate (Delegate dlg)
 
 			ig.Emit (OpCodes.Call, wait_for_bridge_processing_method!);
 
-			var label = ig.BeginExceptionBlock ();
+			bool  filter = Debugger.IsAttached || !JNIEnv.PropagateExceptions;
+			if (!filter) {
```
jonpryor, line 55: @thaystg: This presents a question: is there any way to "mimic" the "normal" .NET Unhandled Exception behaviors while still catching the exception?

Related: https://github.com/xamarin/xamarin-android/issues/4548

The problem is that when the `catch` block doesn't exist, in a managed > Java > managed callstack, if the rightmost frame throws (and no `catch` block is present around it) and execution *doesn't* return to Java, and we unwind the rightmost Java > managed frames, then JVM state may be corrupted.

In an ideal world, *every* "marshal method" would have a `try{…} catch (Exception) {…}` handler, so that when a managed method throws an exception it can be properly caught & propagated back to Java. The problem with this ideal is it means every exception is *by definition* "handled", thus "mooting" the entire concept of "*first chance* unhandled exceptions".  (There'd still *eventually* be an unhandled exception, but that would be on the *leftmost* managed frame, after the rightmost frame had been unwound.)

Is this possible?

thaystg: Are you telling me that removing the try catch when the debugger is attached is a problem because when we have a case like:
managed -> Java -> managed (which throws the unhandled exception), debugger will stop in the exception, but when we continue the execution we may get an error because JVM state may be corrupted? Which kind of error? shouldn't the process be finished in this case?
If I test #4548 I will reproduce the case that you are saying?

jonpryor: > Are you telling me that removing the try catch when the debugger is attached is a problem because when we have a case like:
> managed -> Java -> managed (which throws the unhandled exception), debugger will stop in the exception, but when we continue the execution we may get an error because JVM state may be corrupted?

Yup. Because the debugger is running, there are no `try`/`catch` blocks at Java-to-managed boundaries, and because there are `try`/`catch` blocks, *if* there is an exception thrown from managed code, the JVM state *is* corrupted, and Android aborts the process.

*Desktop* repro:

```bash
$ git clone https://github.com/xamarin/java.interop.git
$ cd java.interop
$ make prepare
$ cat > d <<EOF
diff --git a/tests/Java.Interop-Tests/Java.Interop/TestType.cs b/tests/Java.Interop-Tests/Java.Interop/TestType.cs
index 4640d5f1..2cd108b1 100644
--- a/tests/Java.Interop-Tests/Java.Interop/TestType.cs
+++ b/tests/Java.Interop-Tests/Java.Interop/TestType.cs
@@ -172,7 +172,8 @@ namespace Java.InteropTests
 		static Delegate GetMethodThrowsHandler ()
 		{
 			Action<IntPtr, IntPtr> h = MethodThrowsHandler;
-			return JniEnvironment.Runtime.MarshalMemberBuilder.CreateMarshalToManagedDelegate (h);
+//			return JniEnvironment.Runtime.MarshalMemberBuilder.CreateMarshalToManagedDelegate (h);
+			return h;
 		}
 
 		static void MethodThrowsHandler (IntPtr jnienv, IntPtr n_self)

EOF
$ git apply < d
$ make
$ make run-tests TESTS=bin/TestDebug/Java.Interop-Tests.dll
```

Desired behavior: nothing bad happens. ;-)

Actual results: horrible flaming death:

```bash
$ make run-tests TESTS=bin/TestDebug/Java.Interop-Tests.dll
r=0; \
	 	msbuild /p:Configuration=Debug  build-tools/scripts/RunNUnitTests.targets /p:TestAssembly=bin/TestDebug/Java.Interop-Tests.dll || r=1; \
	exit $r;
…
  	Native Crash Reporting
  Got a segv while executing native code. This usually indicates
  used by your application.
  
  	Native stacktrace:
  	0x101da3779 - /Library/Frameworks/Mono.framework/Versions/Current/Commands/mono : mono_dump_native_crash_info
  	0x101d3b5be - /Library/Frameworks/Mono.framework/Versions/Current/Commands/mono : mono_handle_native_crash
  	0x101d9d8f8 - /Library/Frameworks/Mono.framework/Versions/Current/Commands/mono : altstack_handle_and_restore
  	0x10c849e4e - /Users/jon/android-toolchain/jdk-11/lib/server/libjvm.dylib : _ZN5Chunk9next_chopEv
  	0x10cb5ccec - /Users/jon/android-toolchain/jdk-11/lib/server/libjvm.dylib : checked_jni_GetBooleanField
  	0x10c299140 - /Users/jon/Dropbox/Developer/Java.Interop/bin/TestDebug/libjava-interop.dylib : java_interop_jnienv_get_boolean_field
  	0x1650e3841 - Unknown
  
  	Telemetry Dumper:
EXEC : warning : SIGSEGV handler expected:libjvm.dylib+0x5f61c3  found:0x0000000000000000 [/Users/jon/Dropbox/Developer/Java.Interop/build-tools/scripts/RunNUnitTests.targets]
  Signal Handlers:
  SIGSEGV: SIG_DFL, sa_mask[0]=00000000000000000000000000000000, sa_flags=none
…
```

Again, this crash is "expected", because by *not* catching all exceptions (which is what `JniEnvironment.Runtime.MarshalMemberBuilder.CreateMarshalToManagedDelegate()` does), we're propagating a Mono exception "through" a Java-to-managed boundary, and Java Does Not Like That. (Deservedly so!)

The Good And Proper solution is to catch *all* exceptions at *all* Java-to-managed boundaries, but in doing so we "kill" the normal/expected IDE unhandled exception behavior… because now all exceptions are handled.

Is there a way to have our cake and eat it, too?

marek-safar: @thaystg what is the best path forward here?

thaystg: I'm not sure how can we do this, but I would like to try to generate a try catch with some "mark" so we would know that the debugger should stop because it is not a really handled exception, it's only an exception that we catch in our forced catch.
I will come back to soon to this issue to make more tests.

jonpryor:
Example Project: 
[Scratch.JMJMException.zip](https://github.com/xamarin/xamarin-android/files/6779572/Scratch.JMJMException.zip)

Repro Steps:
 1. Download & Build [Scratch.JMJMException.zip](https://github.com/xamarin/xamarin-android/files/6779572/Scratch.JMJMException.zip)
 2. In a command-line, run `adb logcat > log.txt`; peruse contents for later discussion
 3. Within the IDE, *Run* the app with the debugger *attached*.  Click the "Click Me!" button which fills the screen.

    What should happen is that the IDE will "break" at the `throw new Exception("…")` line in `MainActivity.cs`, with the following message written to `log.txt`:

    ```
    V mono-stdout: # jon: Should be in a Java > Managed [MainActivity.OnCreate] > Java [Run.tryCatchFinally] > Managed [Run] frame. Throwing an exception...
    ```

    When you *Continue* execution, the app will *exit*; it will no longer be running.  Checking `log.txt` from (2) will result in viewing a lines similar to:

    ```
    F libc    : Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x40 in tid 26527 (h.jmjmexception), pid 26527 (h.jmjmexception)
    I crash_dump64: obtaining output fd from tombstoned, type: kDebuggerdTombstone
    I crash_dump64: performing dump of process 26527 (target tid = 26527)
    F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
    F DEBUG   : Build fingerprint: 'google/redfin/redfin:11/RQ1A.210105.003/7005429:user/release-keys'
    F DEBUG   : Revision: 'MP1.0'
    F DEBUG   : ABI: 'arm64'
    F DEBUG   : Timestamp: 2021-07-07 14:56:02-0400
    F DEBUG   : pid: 26527, tid: 26527, name: h.jmjmexception  >>> scratch.jmjmexception <<<
    F DEBUG   : uid: 10291
    F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x40
    F DEBUG   : Cause: null pointer dereference
    F DEBUG   :     x0  0000007fc13ec9d0  x1  0000000000000001  x2  0000000000000001  x3  000000747edb4b4c
    F DEBUG   :     x4  000000747edb4b4c  x5  b4000074af42d43e  x6  0000000000000020  x7  0000000000000020
    F DEBUG   :     x8  0000000000000008  x9  0000000000000000  x10 b4000074af42d427  x11 0000000000000010
    F DEBUG   :     x12 3d2120746e696f70  x13 7274706c6c756e20  x14 0000000000000000  x15 0000000000000005
    F DEBUG   :     x16 000000747f3b3678  x17 000000771396524c  x18 0000000000000008  x19 0000007fc13ec8f0
    F DEBUG   :     x20 0000007fc13ec9d0  x21 0000000000000001  x22 00000077149d5000  x23 0000000000000000
    F DEBUG   :     x24 b40000758f3db010  x25 b4000075ef3d40c8  x26 0000007414437210  x27 0000007fc13ec9d0
    F DEBUG   :     x28 0000007fc13ece38  x29 0000007fc13ec8b0
    F DEBUG   :     lr  000000747eec66f8  sp  0000007fc13ec890  pc  000000747eec5c30  pst 0000000000001000
    F DEBUG   : backtrace:
    F DEBUG   :       #00 pc 00000000001a8c30  /apex/com.android.art/lib64/libart.so (art::ArtMethod::PrettyMethod(bool)+76) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #01 pc 00000000001a96f4  /apex/com.android.art/lib64/libart.so (art::ArtMethod::GetOatQuickMethodHeader(unsigned long)+592) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #02 pc 0000000000576718  /apex/com.android.art/lib64/libart.so (void art::StackVisitor::WalkStack<(art::StackVisitor::CountTransitions)0>(bool)+516) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #03 pc 00000000005a35d0  /apex/com.android.art/lib64/libart.so (art::Thread::GetCurrentMethod(unsigned int*, bool, bool) const+156) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #04 pc 0000000000382004  /apex/com.android.art/lib64/libart.so (art::JavaVMExt::JniAbort(char const*, char const*)+620) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #05 pc 0000000000382804  /apex/com.android.art/lib64/libart.so (art::JavaVMExt::JniAbortV(char const*, char const*, std::__va_list)+108) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #06 pc 000000000037463c  /apex/com.android.art/lib64/libart.so (art::(anonymous namespace)::ScopedCheck::AbortF(char const*, ...)+148) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #07 pc 0000000000373120  /apex/com.android.art/lib64/libart.so (art::(anonymous namespace)::ScopedCheck::CheckPossibleHeapValue(art::ScopedObjectAccess&, char, art::(anonymous namespace)::JniValueType)+3116) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #08 pc 0000000000371d84  /apex/com.android.art/lib64/libart.so (art::(anonymous namespace)::ScopedCheck::Check(art::ScopedObjectAccess&, bool, char const*, art::(anonymous namespace)::JniValueType*)+612) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #09 pc 00000000003674e4  /apex/com.android.art/lib64/libart.so (art::(anonymous namespace)::CheckJNI::NewString(_JNIEnv*, unsigned short const*, int)+680) (BuildId: aeb876e57f112c8539df5053f2da77d4)
    F DEBUG   :       #10 pc 0000000000008b50  /data/app/~~xWY_FPMiBGkhLQyx_hhKpQ==/scratch.jmjmexception-K9NCUEsEY32ORi4xuwFBQw==/lib/arm64/libxa-internal-api.so (java_interop_jnienv_new_string+44) (BuildId: 0657f6418c67601c3f4fe59c63d9b239293d6c68)
    F DEBUG   :       #11 pc 000000000000bda8  <anonymous:7463644000>
    ```

 4. Run the app *without* the debugger being attached.

    What happens is that the app *doesn't* crash, which differs from (3).  Furthermore, *more* `# jon` messages are also present in `log.txt`, which weren't present in (3):

    ```
    V mono-stdout: # jon: MyCatchHandler.OnCatch: t=Android.Runtime.JavaProxyThrowable: Exception of type 'Android.Runtime.JavaProxyThrowable' was thrown.
    …
    V mono-stdout: # jon: from Java finally block
    ```

jonathanpeppers:
Here is the same project ported to .NET 6: [Scratch.JMJMException.zip](https://github.com/xamarin/xamarin-android/files/6779689/Scratch.JMJMException.zip)

Crashes (and doesn't break) with: [log.txt](https://github.com/xamarin/xamarin-android/files/6779693/log.txt)

marek-safar:
@thaystg what are the next steps here?

thaystg:
As discussed with @lewing  I will work on it again as soon as possible. Only finishing 2 implementations: one about debugging after hotreload on android and another about show custom displays on wasm debugger, then I will work on it again.

jonpryor:
@lambdageek  suggests that Android use:

```c
/*
* mono_install_ftnptr_eh_callback:
*
*   Install a callback that should be called when there is a managed exception
*   in a native-to-managed wrapper. This is mainly used by iOS to convert a
*   managed exception to a native exception, to properly unwind the native
*   stack; this native exception will then be converted back to a managed
*   exception in their managed-to-native wrapper.
*/
void
mono_install_ftnptr_eh_callback (MonoFtnPtrEHCallback callback)
```

jonpryor:
@lambdageek: while looking at what `mono_install_ftnptr_eh_callback()` does, I see:

https://github.com/dotnet/runtime/blob/21dc0040b5d6cd52501d9c502fb3ddbc23db97d7/src/mono/mono/mini/mini-exceptions.c#L2364-L2372

In particular, *this* worries me:

```c
g_error ("Did not expect ftnptr_eh_callback to return.");
```

The (not quite documented?) requirement, then, is that the `callback` provided to `mono_install_ftnptr_eh_callback()` ***must not return***.

This does *not* work for JNI stack unwind.  JNI is limited to the C ABI, which doesn't have exceptions as a language or ABI feature, and thus stack unwind is *cooperative*, by necessity.

See also:

  * https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#java_exceptions

Thus, the way to get Java to unwind its stack is to *return execution to Java*.

I suppose `setjmp()`/`longjmp()` could be used to "return execution to Java", except that I don't *want* to be using `setjmp()`/`longjmp()`, *particularly* not from managed code!

Thus, I don't consider `mono_install_ftnptr_eh_callback()` to be a viable alternative on Android. :-(

jonpryor:
@thaystg: Re-upping and re-framing this: the problems with the current approach (described in part in this PR and in #4927) is:

1. If a managed exception is thrown from managed code and the debugger is attached, it is *possible* that JVM state will be corrupted if stack unwind crosses a JNI boundary.  This means there is a potential behavioral difference between app runs when a debugger is attached vs. when no debugger is attached.
2. The way it works -- bad as it is -- involves the use of [*exception filters*](https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/exceptions/exception-handling), either via Mono extensions to `System.Reflection.Emit` and in some "optimized" code paths:
    https://github.com/xamarin/xamarin-android/blob/8bc7a3e84f95e70fe12790ac31ecd97957771cb2/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs#L21-L26
  a. IIRC, AOT (or AOT+LLVM?) doesn't like the use of exception filters, and thus their use should be avoided?

Additionally, we would like to update `generator` output so that more exception handling behavior is "baked into" the code, removing the need for `System.Reflection.Emit`, but as part of that we'd like to know what "ABI" we should be using.  We could emit exception filters, but then vargaz won't be happy with us :wink:, especially given that Profiled AOT is enabled by default in .NET Android 6 (7?).

jonpryor:
@thaystg [previously mentioned](https://github.com/xamarin/xamarin-android/pull/4927#discussion_r552223778):

>  I would like to try to generate a try catch with some "mark" so we would know that the debugger should stop because it is not a really handled exception, it's only an exception that we catch in our forced catch.

I just read that -- not sure why I didn't see it before? Or don't remember having seen it before? ¯\\_(ツ)_/¯  -- and thought of an "interesting" way to "mark" an exception: with another exception!

```csharp
namespace System.Runtime.ExceptionServices {
    public partial class StackUnwinding : Exception
    {
        public StackUnwinding(Exception innerException);
    }
}
```

where `StackUnwinding` allows code to explicitly participate in native stack unwind:

```csharp
void JniBoundary(IntPtr jnienv, IntPtr self)
{
    var t = new JniTransition(jnienv);
    try {
        // code which may throw an exception
    }
    catch (StackUnwinding e) {
        t.SetPendingException(e.InnerException);
    }
    finally {
        t.Dispose();
    }
}
```

Given a `Managed1 > JniCall1 > Java > JniBoundary1 > Managed2 > JniCall2 > Java > JniBoundary2 > Managed3` callstack, and the rightmost `Managed3` throws e.g. `InvalidOperationException`. If a debugger is attached, then a "first chance exception" will be raised.

***Unhandled Exception***

Next, "first pass" is performed: the  entire *managed* callstack is scanned to see if anything catches `InvalidOperationException`.  If nothing catches `InvalidOperationException` and a debugger is attached, then an "unhandled exception" is raised in the debugger, *still* at the initial `throw` site.

Finally, the "second pass" is performed to unwind the callstack, *but* each managed frame is unwound, right-to-left, in a *two-check" process:

 1. Execute a `StackUnwinding` block, if present
 2. Otherwise, unwind the stack normally, executing relevant `catch` blocks, which shouldn't exist in this example, and any relevant `finally` blocks.

This would allow us to be involved *only* during stack unwind *back* to a Java caller, without the use of exception filters, in a manner which is consistent whether or not a debugger is attached.

If there are multiple `catch` blocks which involve `StackUnwinding` and any other `ExceptionType`, that should be a "style-cop" warning and *only* the `StackUnwinding` catch block will be evaluated.

***Handled Exception***

Next, "first pass" is performed: the  entire *managed* callstack is scanned to see if anything catches `InvalidOperationException`.  If `Managed1` catches `InvalidOperationException` and a debugger is attached, the debugger does *not* alert about an unhandled exception; the only Debugger event is the previously described "first chance exception".

For each managed frame between `Managed1` and `Managed2`, the "second pass" is performed to unwind the callstack, in the same "two-check" process.  The `catch(StackUnwinding)` block in `JnITransition2` is hit, which will marshal the exception to Java, then return control to Java.  Java will see a pending exception, execute any relevant catch/finally blocks, and should eventually return control to `JniCall2`, which will see the pending exception from JNI and re-throw it.  This will cause `Managed2` to be unwound, returning control to `JniTransition1`, which will catch the exception and marshal to Java, which …


***Problems***
A problem with this sketch is that it doesn't support cross-VM catches: if a `Java.Lang.RuntimeException` is thrown, there is no way to ask the Java callstacks if that type is caught as part of the "first pass".  Thus, if a debugger is attached, you could have a scenario where the debugger alerts you to an "unhandled exception" (no relevant `catch` blocks found!), and when execution is continued, execution continues when a Java method catches & handles the exception.

This also requires "special-casing" a particular exception type, a'la [`ThreadAbortException`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadabortexception?view=net-6.0#remarks).  I am not sure how much appetite there is for a new "special" exception.

Alternatively, it could be a `struct`!  Exception C#2 and later requires that `catch` blocks only catch `Exception` instances, triggering a [CS0155](https://learn.microsoft.com/en-us/dotnet/csharp/misc/cs0155) if you try to catch anything else.  This *could* be changed -- IL doesn't require it -- but now we're looking at a C# language change, which is presumably a much larger "ask".

jonpryor:
Closing this PR as it's kinda moot; in our forthcoming CoreCLR & NativeAOT world orders, `mono_unhandled_exception()` won't exist.  We need to figure out what the replacement is, and use said replacement, in order to have proper unban deed exception semantics when running atop CoreCLR & NativeAOT.

What's currently here will not help with that future world order.  Additionally, what's here now doesn't directly apply (things have moved), and there's nobody around to take it over. (Assuming we still want the approach in here?)

Issue #4907, opened 2020-07-09
Bump macOS native toolchain used.


jonpryor:
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1139574

When building on macOS, Xamarin.Android uses Xcode 10.2.1 to build native code, e.g. https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3874736&view=logs&j=96fd57f5-f69e-53c7-3d47-f67e6cf9b93e&t=65256bb7-a34c-5353-bc4d-c02ee25dc133

```
InstalledDir: /Applications/Xcode_10.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

We should look into using a more recent Xcode version, or see if we can auto provision *and use* a more recent Xcode Command Line Tools version.

moljac:
Who has the access to the bots? To assign it?

PR #4930, opened 2020-07-16
[ci] Install and build with Xcode 11

pjcollins:
Context: https://github.com/xamarin/xamarin-android/issues/4907

A new script has been added to manage the Xcode version that we build
against on macOS.  Our three macOS build pools all appear to be using
Mojave agents, and as a result Xcode 11.3.1 will be used.  This script
will support a future migration to Catalina and will use the latest
Xcode currently available (11.6) when building on that macOS version.

### DIFF
--- a/build-tools/automation/azure-pipelines-oss.yaml
+++ b/build-tools/automation/azure-pipelines-oss.yaml
@@ -19,7 +19,7 @@ pr:
 variables:
   XA.Jdk8.Folder: jdk-1.8
   XA.Jdk11.Folder: jdk-11
-  XA.Build.MacOSSPool: macOS-10.15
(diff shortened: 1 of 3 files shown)


jonpryor:
@pjcollins: something else to keep in mind is PR #4904: we'll be wanting/needing Xcode *12* (beta!) at some point, so that we can produce universal binaries.

For the time being it's kinda moot, as to be *really* useful *mono* needs to also be a Universal binary (timeframe unknown), but it *is* something to keep in mind.

pjcollins:
>  we'll be wanting/needing Xcode 12 (beta!) at some point, so that we can produce universal binaries.

👍 this should be trivial to support in the future after this PR is up and running.

pjcollins:
Taking this out of "draft mode" queued some new builds, the latest OSS build should be here - https://dev.azure.com/xamarin/public/_build/results?buildId=25443&view=results.

PR #4931, opened 2020-07-16
[build] PackDotNet target to simplify .NET 6 development

jonathanpeppers:
When working on .NET 6, I use the following powershell function for my
"developer loop":

    function pack-xa-nugets([string] $configuration = 'Debug')
    {
        & msbuild $xa\build-tools\xa-prep-tasks\xa-prep-tasks.csproj
        & msbuild $xa\Xamarin.Android.sln -p:DisableApiCompatibilityCheck=true
        & msbuild /t:CreateAllPacks build-tools/create-packs/Microsoft.Android.Sdk.proj
        Remove-Item -r $xa\packages\microsoft.android.* -ErrorAction SilentlyContinue
    }

The steps here are:

1. Build `xa-prep-tasks.csproj` so the `_CreateVersion` MSBuild target
   can run and update git versioning information:

https://github.com/xamarin/xamarin-android/blob/9c880983914e2dbc3f384dea16b5351696732682/build-tools/xa-prep-tasks/xa-prep-tasks.targets#L4

2. Build `Xamarin.Android.sln`.

3. Run `CreateAllPacks`, to create all the `.nupkg` files.

4. Delete any previously extracted files in `packages/microsoft.android.*`

Step 4 is particularly important for sanity. Otherwise, I have
attempted to build a project, but NuGet did not extract the new
`.nupkg` files because their versions were the same!

Because no one else has this script, I ported it to MSBuild to it can
live this in repo and work on all platforms.

On Windows:

    msbuild Xamarin.Android.sln /t:PackDotNet

On macOS:

    make pack-dotnet

Down the road we could improve this workflow further by copying files
directly. The `PackDotNet` target currently takes about 40 seconds on
my Windows desktop machine.

Other naming suggestions welcome.

### DIFF
--- a/build-tools/scripts/DotNet.targets
+++ b/build-tools/scripts/DotNet.targets
@@ -1,0 +1,16 @@
+<Project>
+  <Target Name="PackDotNet">
+    <PropertyGroup>
+      <_TopDir>$(MSBuildThisFileDirectory)..\..\</_TopDir>
+    </PropertyGroup>
+    <MSBuild Projects="$(_TopDir)build-tools\xa-prep-tasks\xa-prep-tasks.csproj" />
+    <MSBuild Projects="$(_TopDir)Xamarin.Android.sln" Properties="DisableApiCompatibilityCheck=true" />
+    <MSBuild Projects="$(_TopDir)build-tools\create-packs\Microsoft.Android.Sdk.proj" Targets="CreateAllPacks" />
+    <!-- Clean up old, previously restored packages -->
+    <ItemGroup>
+      <_OldPackages Include="$(_TopDir)packages\microsoft.android.*\**\*.nupkg" />
+      <_DirectoriesToRemove Include="%(_OldPackages.RootDir)%(_OldPackages.Directory)" />
+    </ItemGroup>
+    <RemoveDir Directories="@(_DirectoriesToRemove)" />
+  </Target>
+</Project>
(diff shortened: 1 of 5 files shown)


build-tools/scripts/DotNet.targets
```diff
@@ -0,0 +1,16 @@
...
+    <MSBuild Projects="$(_TopDir)build-tools\xa-prep-tasks\xa-prep-tasks.csproj" />
+    <MSBuild Projects="$(_TopDir)Xamarin.Android.sln" Properties="DisableApiCompatibilityCheck=true" />
+    <MSBuild Projects="$(_TopDir)build-tools\create-packs\Microsoft.Android.Sdk.proj" Targets="CreateAllPacks" />
+    <!-- Clean up old, previously restored packages -->
+    <ItemGroup>
+      <_OldPackages Include="$(_TopDir)packages\microsoft.android.*\**\*.nupkg" />
+      <_DirectoriesToRemove Include="%(_OldPackages.RootDir)%(_OldPackages.Directory)" />
+    </ItemGroup>
+    <RemoveDir Directories="@(_DirectoriesToRemove)" />
+  </Target>
```
pjcollins, line 15: This looks great. I had a similar helper target in a branch while working on the Mono.Android tests, and it may be a good candidate for this as well? It generates `nuget.config` and `global.json` which point to the newly built packages. If that seems useful we could add this target and output these files to `.\bin\BuildDebug\nupkgs\`. Otherwise I will add it in a future PR. The target is here:
https://github.com/pjcollins/xamarin-android/commit/bb1dd6df7c8e5525c22b2f2f348992b19f0bf96e#diff-6fc2685fd76642c79c8ad32367445f42R6-R48

jonathanpeppers: `build-tools\scripts\DotNet.targets` might be a good place to put `GenerateAndroidSdkProjectConfigs`.

But maybe we save it for later for when it is needed?

pjcollins: I'm happy to add it later 👍

PR #4933, opened 2020-07-17
[One .NET] fix build ordering & BuildItem.Deleted

jonathanpeppers:
The `AndroidUpdateResourcesTest.CheckFilesAreRemoved` test was failing
with:

    Resources\values\Theme.xml error XA2001: Source file 'Resources\values\Theme.xml' could not be found.

When looking through logs, I noticed lots of differences in build
ordering from legacy projects.

1. `$(BuildDependsOn)` ordering - In legacy projects
   `ImplicitlyExpandDesignTimeFacades` is listed last, while it was in
   the middle for .NET 5+ projects.

This was caused by the ordering of `<Import/>`.
`Microsoft.Android.Sdk.BuildOrder.targets` needs to be imported before
`Xamarin.Android.Common.targets`.

2. `$(BuildDependsOn)` also listed these targets in legacy projects:

    _CheckTargetFramework;
    _RemoveLegacyDesigner;

`_CheckTargetFramework` is not needed, but it needs to be moved to

`_RemoveLegacyDesigner` should be listed in `$(BuildDependsOn)`.

3. `_UpdateAndroidResources` had different values for `DependsOnTargets`.

In c87ef952, I removed `$(CoreResolveReferencesDependsOn)` to just
simplify things. It turns out `_UpdateAndroidResources` uses this
property. I put things back the way they were.

After fixing the build ordering, the test still failed.

Then I noticed the *actual* cause to the failing test... There were
additional `@(AndroidResource)` items in the `.csproj` file.
`BuildItem.Deleted` wasn't being checked when short-form `.csproj`
files are generated. We should omit build items with this flag set.

After doing this, we can run a couple additional tests under `dotnet`
context.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
@@ -77,6 +77,7 @@ projects, these properties are set in Xamarin.Android.Legacy.targets.
       _CreateNativeLibraryArchive;
       _AddAndroidEnvironmentToCompile;
       _CheckForContent;
+      _RemoveLegacyDesigner;
       _ValidateAndroidPackageProperties;
       $(BuildDependsOn);
     </BuildDependsOn>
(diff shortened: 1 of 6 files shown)


dellis1972:
Not sure if the performance error is within the margin of error

```
Exceeded expected time of 7250ms, actual 7361.866ms
```

I would say yes.... I think :) 

I'm pretty sure the Designer error is unrelated `System.Threading.Tasks.TaskCanceledException : A task was canceled.` for `CanReconnectSeveralTimeAndHaveOneContext`.

jonathanpeppers:
> Exceeded expected time of 7250ms, actual 7361.866ms

We might need to bump some of the times again, but it didn't happen on the second run.

PR #4935, opened 2020-07-18
[xaprepare] CGManifest.json generation

jonpryor:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1139601
Fixes: https://dev.azure.com/devdiv/DevDiv/_componentGovernance/112013/alert/3185600?typeId=1016455

Context: https://docs.opensource.microsoft.com/tools/cg/cgmanifest.html
Context: https://github.com/xamarin/xamarin-macios/commit/32c4999822da706e0d374c253b76a293ba80cd8a

For a variety of "sanity and security" reasons, we need to better
integrate with "Component Governance", which "detects open source [used]
and alerts you to whether it has security vulnerabilities…".

Much of Component Governance is automagic, scanning the source tree
before and after the build occurs -- meaning it can scan NuGet
dependencies & such which are "in tree" after `make prepare` -- but it
can't have "insight" into the *meaning* of certain dependencies.

Consider the [7-Zip.CommandLine][0] NuGet package: we use this *only*
to build the xamarin-android repo itself; it is not redistributed.
However, it gets flagged by Component Governance due to LGPL-2 use.

There is a way to tell Component Governance that a file isn't
redistributed: provide a [`CGManifest.json` file][1] and set the
`DevelopmentDependency` to `true`.

Update `xaprepare` so that it (manually) "knows" which packages are
Development Dependencies, by listing their names in
`Step_GenerateCGManifest.DevelopmentDependencies`, and then scan the
build tree for NuGet package references and Git submodule info, so
that an appropriate `CGManifest.json` file can be generated.

[0]: https://www.nuget.org/packages/7-Zip.CommandLine/
[1]: https://docs.opensource.microsoft.com/tools/cg/cgmanifest.html

### DIFF
--- a/build-tools/Xamarin.Android.Tools.BootstrapTasks/result-packaging.targets
+++ b/build-tools/Xamarin.Android.Tools.BootstrapTasks/result-packaging.targets
@@ -8,6 +8,7 @@
     <_BuildStatusFiles Include="$(BootstrapOutputDirectory)XABuildConfig.cs" Condition="Exists('$(BootstrapOutputDirectory)XABuildConfig.cs')" />
     <_BuildStatusFiles Include="$(BootstrapOutputDirectory)*.binlog" />
     <_BuildStatusFiles Include="$(BootstrapOutputDirectory)prepare*.log" />
+    <_BuildStatusFiles Include="$(BootstrapOutputDirectory)*.json" />
     <_BuildStatusFiles Include="$(BootstrapOutputDirectory)*.mk" />
     <_BuildStatusFiles Include="$(BootstrapOutputDirectory)*.projitems" />
     <_BuildStatusFiles Include="$(BootstrapOutputDirectory)*.cmake" />
--- a/build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
+++ b/build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
@@ -1,0 +1,317 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+
+namespace Xamarin.Android.Prepare
+{
+	partial class Step_GenerateCGManifest : Step
+	{
+		static readonly HashSet<string> DevelopmentDependencies = new HashSet<string> (StringComparer.OrdinalIgnoreCase) {
+			"7-Zip.CommandLine",
+			"icsharpcode/NRefactory",
+			"Kajabity.Tools.Java",
+			"Mono.Linq.Expressions",
+			"mono/debugger-libs",
+			"Newtonsoft.Json",
+			"System.CommandLine",
+			"Xamarin.Forms",
+			"xunit",
+			"xunit.abstractions",
+			"xunit.analyzers",
+			"xunit.assert",
+			"xunit.core",
+			"xunit.extensibility.core",
+			"xunit.extensibility.execution",
+			"xunit.runner.utility",
+
+			// ???
+			"Microsoft.Build",
+			"Microsoft.Build.Framework",
+			"Microsoft.Build.Tasks.Core",
+			"Microsoft.Build.Utilities.Core",
+			"Microsoft.VisualStudio.CoreUtility",
+			"Microsoft.VisualStudio.Imaging",
+			"Microsoft.VisualStudio.OLE.Interop",
+			"Microsoft.VisualStudio.Shell.15.0",
+			"Microsoft.VisualStudio.Shell.Framework",
+			"Microsoft.VisualStudio.Shell.Interop",
+			"Microsoft.VisualStudio.Shell.Interop.8.0",
+			"Microsoft.VisualStudio.Shell.Interop.9.0",
+			"Microsoft.VisualStudio.TextManager.Interop",
+			"Microsoft.VisualStudio.TextManager.Interop.8.0",
+			"Microsoft.VisualStudio.Threading",
+			"Microsoft.VisualStudio.Utilities",
+			"Microsoft.VisualStudio.Validation",
+			"Microsoft.VSSDK.BuildTools",
+		};
+
+		public Step_GenerateCGManifest ()
+			: base ("Generate CGManifest.json")
+		{}
+
+		protected override async Task<bool> Execute (Context context)
+		{
+			var nugets              = MSBuildPackageReferenceInfo.GetPackageReferences ();
+			var git                 = new GitRunner (context);
+			var gitSubmoduleInfo    = await git.ConfigList (new[]{"--blob", "HEAD:.gitmodules"});
+			var gitSubmoduleStatus  = await git.SubmoduleStatus ();
+			var gitSubmodules       = GitSubmoduleInfo.GetGitSubmodules (gitSubmoduleInfo, gitSubmoduleStatus);
+
+			var cgManifestEntries   = ((IEnumerable<CGManifestEntry>) nugets).Concat (gitSubmodules)
+				.OrderBy (e => e.Name);
+
+			var jsonPath    = Path.Combine (Configurables.Paths.BuildBinDir, "CGManifest.json");
+			using var json  = File.CreateText (jsonPath);
+
+			json.WriteLine ("{");
+			json.WriteLine ("    \"Registrations\": [");
+
+			var properties = new Dictionary<string, string> ();
+
+			bool first = true;
+
+			foreach (var entry in cgManifestEntries) {
+				if (first) {
+					first   = false;
+				} else {
+					json.WriteLine (",");
+				}
+
+				WriteComponent (json, entry, properties);
+			}
+
+			json.WriteLine ();
+			json.WriteLine ("    ]");
+			json.WriteLine ("}");
+
+			return true;
+		}
+
+		void WriteComponent (TextWriter json, CGManifestEntry entry, Dictionary<string, string> properties)
+		{
+			string dev = DevelopmentDependencies.Contains (entry.Name)
+				? "true"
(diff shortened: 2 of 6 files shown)


jonpryor:
To add to the commit message, snippet of generated `CGManifest.json` file:

```json
{
    "Registrations": [
        {
            "Component": {
                "Type": "nuget",
                "Nuget": {
                    "Name": "7-Zip.CommandLine",
                    "Version": "18.1.0"
                }
            },
            "DevelopmentDependency":true
        },
        {
            "Component": {
                "Type": "git",
                "Git": {
                    "RepositoryUrl": "https://github.com/dlfcn-win32/dlfcn-win32",
                    "CommitHash": "ef7e412d3fa2b511e660342e237536c235e51ee4"
                }
            },
            "DevelopmentDependency":false
        },
…
```

build-tools/Xamarin.Android.Tools.BootstrapTasks/result-packaging.targets
```diff
@@ -8,6 +8,7 @@
     <_BuildStatusFiles Include="$(BootstrapOutputDirectory)XABuildConfig.cs" Condition="Exists('$(BootstrapOutputDirectory)XABuildConfig.cs')" />
     <_BuildStatusFiles Include="$(BootstrapOutputDirectory)*.binlog" />
     <_BuildStatusFiles Include="$(BootstrapOutputDirectory)prepare*.log" />
+    <_BuildStatusFiles Include="$(BootstrapOutputDirectory)*.json" />
```
jonpryor, line 11: This line appears wrong, in that the generated `xa-build-status-11.1.99.16_8105993-Darwin-Release.zip` doesn't contain any `CGManifest.json` files. :-(

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
```diff
@@ -0,0 +1,302 @@
...
+				if (i >= 0)
+					return RepositoryUrl.Substring (i + github.Length);
+				return RepositoryUrl;
+			}
+		}
+
+		public  string      RepositoryUrl   {get; private set;}
+		public  string      CommitHash      {get; private set;}
+
+		private GitSubmoduleInfo ()
```
grendello, line 116: `private` is the default visibility for all class members, no need to us the keyword

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
```diff
@@ -0,0 +1,302 @@
...
+			}
+
+			if (url != null && path != null)
+				yield return CreateSubmoduleInfo (url, path);
+
+			GitSubmoduleInfo CreateSubmoduleInfo (string url, string path)
+			{
+				string? hash = null;
+				foreach (var e in submoduleStatus) {
+					int pi  = e.IndexOf (path, StringComparison.Ordinal);
```
grendello, line 188: I think `StringComparison.OrdinalIgnoreCase` would be safer here - URLs are not case-sensitive and the status paths come from the filesystem which may have a different case (especially on case-aware but case-insensitive filesystems)

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	sealed class MSBuildPackageReferenceInfo : CGManifestEntry {
+
+		string name;
+
+		public  override    string      Name    => name;
+		public  string      Version {get; private set;}
+
+		private MSBuildPackageReferenceInfo (string name, string version)
```
grendello, line 218: Ditto regarding `private`

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
```diff
@@ -0,0 +1,302 @@
...
+				if (contents.IndexOf ("PackageReference", StringComparison.Ordinal) < 0 ||
+						contents.IndexOf ("PropertyGroup", StringComparison.Ordinal) < 0)
+					continue;
+				var proj                = XDocument.Parse (contents);
+				var packageReferences   = proj.Elements (MSBuildXmlns + "Project")
+					.Elements (MSBuildXmlns + "ItemGroup")
+					.Elements (MSBuildXmlns + "PackageReference");
+				foreach (var packageReference in packageReferences) {
+					var name    = (string) packageReference.Attribute ("Include");
+					var version = (string) packageReference.Attribute ("Version") ??
```
grendello, line 263: Shouldn't the casts here and in the previous line be to `string?`? Line 265 below makes a comparison against `null` and `string` cannot be `null` with NRT.

build-tools/Xamarin.Android.Tools.BootstrapTasks/result-packaging.targets
jonpryor: Lol, I never added `Step_GenerateCGManifest` to any scenario! Thus, it never ran!

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
```diff
@@ -0,0 +1,302 @@
...
+			json.WriteLine ("    \"Registrations\": [");
+
+
+			bool first = true;
+
+			foreach (var entry in cgManifestEntries) {
+				if (!first) {
+					json.WriteLine (",");
+				}
+				first = false;
```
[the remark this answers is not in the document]
jonpryor: Nope, won't work: on first iteration, `first==true`, so line 76 *won't* execute, so if it were in that block `first=false` would likewise never execute.

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
jonpryor: What this assumes is *consistency* between `git config --blob HEAD:.gitmodules --list` and `git submodule status`:

```bash
$ git config --blob HEAD:.gitmodules --list
…
submodule.external/Java.Interop.path=external/Java.Interop
submodule.external/Java.Interop.url=https://github.com/xamarin/java.interop.git
submodule.external/Java.Interop.branch=master
…

$ git submodule status
 60efc990cf7d33a55ff7864f215569d14ce51e34 external/Java.Interop (remotes/origin/HEAD-4-g60efc99)
…
```

Here, `path` would be `external/Java.Interop`, which is then "keyed" to the appropriate `git submodule status` line.

I *hope*/*assume* that these should be identical between the two commands, in which case `.Ordinal` should be sufficient.

The `url` isn't relevant here, and is "passed through as-is".

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
jonpryor: Indeed!

jonpryor:
I *believe* that `CGManifest.json` is used by the **Component Detection** job, e.g. on master: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=3900360&view=logs&j=96fd57f5-f69e-53c7-3d47-f67e6cf9b93e&t=ca5ba72a-d988-558d-8a6b-324437a88f8a

"Oddly" (?), this job doesn't appear to run on PR builds.  Not sure what the semantics should be here?  Makes it hard to see if this PR *does* anything…

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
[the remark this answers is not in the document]
grendello: Right, I took a shortcut - move it to the block, invert the condition, set `first` to `false` in there and call `json.WriteLine` in `else`

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
grendello: `.gitmodules` is just a text file, it can potentially have files using any case, especially if created on a case-insensitive host. `git submodule status` reads the file, but it also reads the filesystem (reading the `.git` dir for information about the submodules) and who knows which case it will use if there's a discrepancy. It's just a precaution that doesn't cost much but may save a minor headache.

build-tools/Xamarin.Android.Tools.BootstrapTasks/result-packaging.targets
jonpryor: Now it's in there.

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
jonpryor: Done; it uses `.OrdinalIgnoreCase` now.

build-tools/xaprepare/xaprepare/Steps/Step_GenerateCGManifest.cs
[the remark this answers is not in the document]
jonpryor: I think I did as you asked, more or less, in the most recent force-push?

pjcollins:
> "Oddly" (?), this job doesn't appear to run on PR builds. Not sure what the semantics should be here? Makes it hard to see if this PR does anything…

I think we can add the following line to the bottom of our macOS build job YAML https://github.com/xamarin/xamarin-android/blob/c06558b12e6f615dfbb1dc7b9efd25fbb9872df4/build-tools/automation/azure-pipelines.yaml#L241 temporarily to test this: 
```yaml
- task: ComponentGovernanceComponentDetection@0
```

PR #4937, opened 2020-07-20
Bump LibZipSharp to 1.0.16

grendello:


grendello:
Still broken on Windows :(

PR #4945, opened 2020-07-21
[xaprepare] Redirect some git messages from stderr to stdout

grendello:
`git` outputs its progress messages to the standard error stream which
`xaprepare` intercepts and prints on screen in red and in the following
format:

    stderr | Cloning into 'monodroid'...
    stderr | remote: Enumerating objects: 58, done
    stderr | Receiving objects:   0% (1/79000)
    stderr | Receiving objects:   1% (790/79000)
    stderr | Submodule 'external/Java.Interop' (https://github.com/xamarin/java.interop.git) registered for path 'external/Java.Interop'
    stderr | Submodule 'external/android-api-docs' (https://github.com/xamarin/android-api-docs) registered for path 'external/android-api-docs'

This may be confused for an actual error even though it is most
certainly a harmless progress output.  This commit adds code to
intercept the know progress messages and redirect them to standard
output using normal text formatting.

Note: this will work only when English is used!

### DIFF
--- a/build-tools/xaprepare/xaprepare/ToolRunners/GitRunner.cs
+++ b/build-tools/xaprepare/xaprepare/ToolRunners/GitRunner.cs
@@ -7,6 +7,26 @@ namespace Xamarin.Android.Prepare
 {
 	partial class GitRunner : ToolRunner
 	{
+		// Redirects git progress output from standard error to standard output so that we show less red on the screen
+		sealed class GitProgressStderrWrapper : ProcessStandardStreamWrapper
+		{
+			public readonly List<string> Messages = new List<string> ();
+
+			protected override string PreprocessMessage (string message, ref bool writeLine, out bool ignoreLine)
+			{
+				ignoreLine = true;
+
+				if (message.Length == 0)
+					return message;
+
+				Messages.Add (message);
+				Log.Instance.MessageLine (message);
+				ignoreLine = true;
+
+				return message;
+			}
+		}
+
 		// These are passed to `git` itself *before* the command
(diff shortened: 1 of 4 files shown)


jonpryor:
I had a different idea for a solution: "time travel".

Instead of printing out each line as it is encountered, requiring that we "guess" whether this is a "real" error vs. an informational message, we instead:

 1. Log *all* messages, both stdout & stderr, as "normal" output (i.e. *not* red).
 2. *Capture* stderr messages into a `List<string>`.
 3. When the process exits, check the status code.  If it's non-zero, *re-print* the captured stderr messages in red.

This also nicely deals with the localization conundrum: we don't need to care about English at all.

build-tools/xaprepare/xaprepare/ToolRunners/GitRunner.cs
```diff
@@ -187,10 +207,20 @@ public async Task<bool> IsRepoUrlHttps (string workingDirectory)
...
 			runner.WorkingDirectory = workingDirectory;
+			if (useCustomStderrWrapper) {
+				runner.StandardErrorEchoWrapper = new GitProgressStderrWrapper {
+					LoggingLevel = runner.EchoStandardErrorLevel,
+					CustomSeverityName = ProcessRunner.StderrSeverityName,
+				};
+			}
+
+			// We need English for line matching in the stderr handler
+			runner.Environment ["LANG"] = "C";
```
jonpryor, line 222: I don't think that this line is required anymore, as we're not comparing the messages to anything, just capturing them.

PR #4946, opened 2020-07-21
[Xamarin.Android.Build.Tasks] remove $(UseShortFileNames)

jonathanpeppers:
Setting `$(UseShortFileNames)` = `false` would tell Xamarin.Android's
MSBuild targets to use the "long" folder names such as:

* `obj\Debug\lp` -> `__library_projects__`
* `obj\Debug\lp\*\jl` -> `library_project_imports`
* `obj\Debug\lp\*\nl` -> `native_library_imports`

This was useful when `$(UseShortFileNames)` was a new feature, giving
developers a way to "opt out" if they hit a bug. However,
`$(UseShortFileNames)` has defaulted to `true` since ~July 2017 to
help the [MAX_PATH][0] limit on Windows.

I'm finding basically zero usage of `UseShortFileNames=false` on
Github:

https://github.com/search?q=UseShortFileNames+false+extension%3Acsproj&type=Code

This makes it an easy choice.

I have removed all usage of `$(UseShortFileNames)`. The remaining
behavior will act as if `UseShortFileNames=true`.

[0]: https://docs.microsoft.com/windows/win32/fileio/naming-a-file#maximum-path-length-limitation


brendanzagaeski:
To throw out a quick idea since I noticed this draft PR by chance, maybe for the sake of extra thoroughness, `$(UseShortFileNames)`=`false` could produce a build warning too? Admittedly, providing a warning in this case might be a little excessive, but including a warning would fit with a loose pattern that has been used in some previous cases of providing warnings for deprecated and obsoleted features.

jonathanpeppers:
@brendanzagaeski I don't know of a reason why you'd set `UseShortFileNames=false`, and it is [undocumented](https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/BuildProcess.md).

If it was a _useful_, known MSBuild property a warning makes sense. But I don't think we need a warning?

brendanzagaeski:
Works for me on leaving out the warning. I too would be surprised if there were any negative consequences or inconveniences of removing this feature for existing projects, so it's quite possible the warning wouldn't be useful to any users. As long as allowing a little wiggle room for less messaging around lower impact feature removals works for everyone else, it's OK by me too.

> undocumented

Out of curiosity, I did a quick search of old release notes to see if it ever appeared there, and I did find:

https://github.com/xamarin/release-notes-archive/blob/9bac84d3db0a16bcb258c602f71eccfa814ba0a0/release-notes/android/xamarin.android_8/xamarin.android_8.0/index.md#shorter-internal-path-names-experimental

But even with this mention in the release notes, this feature removal likely still has enough wiggle room to allow no warning. For example, even though the Xamarin.Android 8.0 (d15-4) release notes do mention the feature, the subsequent Xamarin.Android 8.1 (d15-5) release notes do not mention the feature, despite the fact that d15-5 was when the feature was promoted out of being "experimental" and when `$(UseShortFileNames)` was set to `True` by default.

PR #4959, opened 2020-07-23
[Microsoft.Android.Sdk] fix %(AndroidSkip*) item metadata

jonathanpeppers:
The `CheckMetadataSkipItemsAreProcessedCorrectly` test was failing
under a `dotnet` context due to item metadata missing, such as:

    <AndroidSkipAddToPackage>True</AndroidSkipAddToPackage>
    <AndroidSkipJavaStubGeneration>True</AndroidSkipJavaStubGeneration>
    <AndroidSkipResourceExtraction>True</AndroidSkipResourceExtraction>

These metadata support AndroidX.Migration, and they are completely
missing from the item groups generated in
`Microsoft.Android.Sdk.AssemblyResolution.targets`.

After the `<ProcessAssemblies/>` MSBuild task is called, we should
call `<AppendCustomMetadataToItemGroup/>` to populate the missing
metadata.

The `CheckMetadataSkipItemsAreProcessedCorrectly` test now passes.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
@@ -60,10 +60,20 @@ _ResolveAssemblies MSBuild target.
         IntermediateAssemblyDirectory="$(MonoAndroidIntermediateAssemblyDir)"
         UseSharedRuntime="$(AndroidUseSharedRuntime)"
         LinkMode="$(AndroidLinkMode)">
-      <Output TaskParameter="OutputAssemblies" ItemName="ResolvedAssemblies" />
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
> Should these metadata items also be on _ShrunkAssemblies Items as well?

Yeah, I should add this.

> I'm curious why _AddAndroidCustomMetaData is not working as intended in this case? It appears in the BuildOrder.targets, but does not seem to work?

`_AddAndroidCustomMetaData` still works "before the build", like for `_ResolveLibraryProjectImports`.

After the build in "legacy" Xamarin.Android, `@(_ReferencePath)` is passed through to `<ResolveAssemblies/>`:

https://github.com/xamarin/xamarin-android/blob/946f269e656fb42fba98b2e3490737bef77e682b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets#L259-L273

But in the new world, we use the `@(ResolvedFileToPublish)` item group coming from the dotnet/sdk. So we are basically starting over with any metadata we need.

But the bright side is, we don't call `<ResolveAssemblies/>` anymore!

PR #4960, opened 2020-07-24
[Xamarin.Android.Build.Tasks] set AndroidApiLevel sooner in <ResolveAndroidTooling/>

jonathanpeppers:
Context: https://build.azdo.io/3906855

Under a `dotnet` context, the
`GetDependencyWhenBuildToolsAreMissingTest` test was failing with:

    platforms/android-26 should be a dependency.
      at Xamarin.Android.Build.Tests.StringAssertEx.Contains (System.String text, System.Collections.Generic.IEnumerable`1[T] collection, System.String message) [0x00042] in <8269b6f7f46a4de4813f522c4edb1727>:0
      at Xamarin.Android.Build.Tests.BuildTest.GetDependencyWhenBuildToolsAreMissingTest () [0x000f2] in <8269b6f7f46a4de4813f522c4edb1727>:0

In d7227b80, I partially fixed the `GetAndroidDependencies` MSBuild
target, but there appears to be another issue.

Reviewing the build log, the `GetAndroidDependencies` target was
returning:

    Done executing task "CalculateProjectDependencies". (TaskId:52)
    Target output items:
        platforms/android-21

This was due to the `$(_AndroidApiLevel)` MSBuild property being
blank, when the `<CalculateProjectDependencies/>` MSBuild task is
called. It used the `$(TargetFrameworkVersion)` of `v5.0` to decide
that API 21 is needed...

The `<ResolveAndroidTooling/>` MSBuild task was failing (due to
missing `build-tools`) and not setting `$(_AndroidApiLevel)` at all.

To solve this issue, I moved things around so `AndroidApiLevel` is set
*first*. `Validate` can remain an empty method so it is overridden by
the legacy `<ResolveAndroidTooling/>` task.

Then I hit another problem on Windows:

    _ValidateAndroidPackageProperties target) ->
        error XA5207: Could not find android.jar for API level 30. This means the Android SDK platform for API level 30 is not installed. Either install it in the Android SDK Manager (Tools > Android > Android SDK Manager...), or change the Xamarin.Android project to target an API version that is installed. (C:\Program Files (x86)\Android\android-sdk\platforms\android-30\android.jar missing.)

We need this change, I ignored the test on Windows for now:

https://github.com/xamarin/android-sdk-installer/pull/450

This also moves any tests related to the `GetAndroidDependencies`
MSBuild target to `AndroidDependenciesTests`. We can now run them all
in a `dotnet` context.


jonathanpeppers:
The one failed test is odd:
```
Xamarin.ProjectTools.FailedBuildException : Build failure: Library1.csprojBuild log recorded at /Users/runner/work/1/s/bin/TestRelease/temp/BuildApplicationWithMonoEnvironmentNormal/Library1/build.log
```
But there are no errors:
```
Build succeeded.
    0 Warning(s)
    0 Error(s)
```
I think `dotnet build` must have returned a non-zero exit code? Not seen that before, but seems unrelated. We can keep an eye out for it.

PR #4961, opened 2020-07-24
[tests] combine 3 test assemblies into Xamarin.Android.Build.Tests

jonathanpeppers:
Moving the following assemblies into `Xamarin.Android.Build.Tasks.dll`:

* `CodeBehindUnitTests.dll`
* `EmbeddedDSOTests.dll`
* `Xamarin.Android.MakeBundle-UnitTests.dll`

This will simplify the `.yaml` definitions for running these tests and
makes it possible to parameterize them for a future `One .NET` test
phase.

The only code changes are:

* Move any classes in the top-level namespace to nested classes.
* Add a `BaseTest.GetPathToZipAlign()` method instead of generating
  `Config.cs` at build time.

These tests are still passing.


jonathanpeppers:
I spot checked a test from each assembly, and they are running & passing on Node-2:

image

image

image

Issue #1597, opened 2018-04-24
AIDL file with method that uses a Map parameter causes build failure


jgold6:
When building the attached test project, the build fails with error:
`
/path/to/obj/Debug/aidl/IMxFrameworkService.cs(17,17): Error CS7036: There is no argument given that corresponds to the required formal parameter 'loader' of 'Parcel.ReadHashMap(ClassLoader)' (CS7036) (MobileAndroid)
`

The `IMxFrameworkService.cs` is the file generated from the AIDL file. There is this line of code in the `OnTransact` method:

`arg1 = data.ReadHashMap ();`

Note that there is no `ClassLoader` parameter. In the `Android.OS.Parcel` class, there is only one overload for the `ReadHashMap` method and it takes a `ClassLoader` (an abstract class) as an argument. There is no overload that takes no arguments. 

### Steps to Reproduce

1. Open the attached test project
2. Build the project

[Microsoft-118042018043876.zip](https://github.com/xamarin/xamarin-android/files/1944147/Microsoft-118042018043876.zip)

### Expected Behavior
Project will build

### Actual Behavior
build fails with error noted above.

### Version Information

=== Visual Studio Enterprise 2017 for Mac ===

Version 7.4.3 (build 10)
Installation UUID: f86726f2-bd5d-4610-867e-44e82f306ca2
Runtime:
	Mono 5.8.1.0 (2017-10/6bf3922f3fd) (64-bit)
	GTK+ 2.24.23 (Raleigh theme)

	Package version: 508010000

=== NuGet ===

Version: 4.3.1.4445

=== .NET Core ===

Runtime: /usr/local/share/dotnet/dotnet
Runtime Versions:
	2.0.5
	2.0.0
	1.1.1
	1.0.4
	1.0.0
SDK: /usr/local/share/dotnet/sdk/2.1.4/Sdks
SDK Versions:
	2.1.4
	2.0.0
	1.0.1
	1.0.0-preview2-003121
MSBuild SDKs: /Library/Frameworks/Mono.framework/Versions/5.8.1/lib/mono/msbuild/15.0/bin/Sdks

=== Xamarin.Profiler ===

Version: 1.6.1
Location: /Applications/Xamarin Profiler.app/Contents/MacOS/Xamarin Profiler

=== Xamarin.Android ===

Version: 8.2.0.16 (Visual Studio Enterprise)
Android SDK: /Users/jongoldberger/Library/Developer/Xamarin/android-sdk-macosx
[...]

PR #4963, opened 2020-07-25
[Xamarin.Android.Tools.Aidl] fix classloader retrieval for List and Map (#1597)

ssimek:
Fixes incorrect code generation from AIDL for cases which require the class loader (ReadArrayList/ReadHashMap)

The current code could not possibly work, as the required parameters were simply not provided to the called methods causing build failures. The one case which provided null instead of class loader would likely cause runtime failures under some circumstances.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/xamarin/xamarin-android?pullRequest=4963) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/xamarin/xamarin-android?pullRequest=4963) before we can accept your contribution.<br/><br/>:x: ssimek [sign now](https://cla.dotnetfoundation.org/xamarin/xamarin-android?pullRequest=4963)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/xamarin/xamarin-android?pullRequest=4963) it.</sub>

jonpryor:
I believe this would fix: https://github.com/xamarin/xamarin-android/issues/1597

jonpryor:
@ssimek wrote:

> The one case which provided null instead of class loader would likely cause runtime failures under some circumstances.

Can you please elaborate?  The Android documentation certainly states that `null` is a valid value for the parameter, e.g. both [`Parcel.readArrayList(ClassLoader)`](https://developer.android.com/reference/android/os/Parcel#readArrayList(java.lang.ClassLoader)) and [`Parcel.readHashMap(ClassLoader)`](https://developer.android.com/reference/android/os/Parcel#readHashMap(java.lang.ClassLoader)) state:

> `loader`: `ClassLoader`: This value may be `null`.

jonpryor:
@ssimek: please also see/try: PR #5003.

ssimek:
Hi @jonpryor,

my fix is based the fix on [original AIDL tool code](https://android.googlesource.com/platform/system/tools/aidl/+/refs/heads/master/aidl_to_java.cpp) and I honestly did not investigate much into why the value is required/should be provided. However, at a glance, I'd say that the call with a `null` classloader would somehow fail if there were another `Parcelable` inside the List/Map, since

> The given class loader will be used to load any enclosed Parcelables.

It is not clear what happens when no class loader is provided. I assume it is safer to do the same as the original AIDL tool instead of hoping that providing `null`will work correctly.

jonpryor:
Superseded by PR#5003.

The `ClassLoader` parameter support isn't part of PR #5003, and is being tracked at: https://github.com/xamarin/xamarin-android/issues/5012

PR #4965, opened 2020-07-27
[Xamarin.Android.Build.Tasks] Add ResponseFile support for Generator Task

### LINKED ISSUE #4956: Add response file support to BindingsGenerator task
With the explosion of referenced assemblies due to AndroidX, the BindingsGenerator task is bumping up against the 32K limit for command lines:
```
C:\Program Files (x86)\Microsoft Visual Studio\2019\Preview\MSBuild\Xamarin\Android
\Xamarin.Android.Bindings.targets(325,5): warning MSB6002: The command-line for the
"BindingsGenerator" task is too long. Command-lines longer than 32000 characters are likely to
fail. Try reducing the length of the command-line by breaking down the call to "BindingsGenerator" 
into multiple calls with fewer parameters per call. [D:\X\GPS-FB-ax_u_202002\generated
\com.google.firebase.firebase-inappmessaging\com.google.firebase.firebase-inapp
messaging.csproj]
```
We added support for using a response file for `generator.exe` here: https://github.com/xamarin/java.interop/pull/665, however we still need to update the MSBuild task to build and pass the response file.

Details for usage are in the linked PR.

dellis1972:
Fixes #4956

We added support for using a response file for generator.exein
xamarin/java.interop#665

This commit updates the BindingGenerator Task to use a response
file rather than passing all the argumets on the command line.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
@@ -31,7 +31,7 @@ namespace Xamarin.Android.Tasks
 		public string ApiXmlInput { get; set; }
 
 		public string AssemblyName { get; set; }
-		
+
 		public string CodegenTarget { get; set; }
 
 		public bool NoStdlib { get; set; }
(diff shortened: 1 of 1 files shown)


jpobst:
This may already do this, but it would be great if the command line arguments that are in the response file could still be printed to the MSBuild output, as I use that for debugging a lot.

dellis1972:
It will. This is still a Work in progress and its not working lol :)

dellis1972:
Designer Failure looks unrelated :D

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
```diff
@@ -110,46 +110,60 @@ public override bool RunTask ()
...
-			if (UseShortFileNames)
-				cmd.AppendSwitch ("--use-short-file-names");
-
-			if (EnableInterfaceMembersPreview && SupportsCSharp8)
-				cmd.AppendSwitch ("--lang-features=interface-constants,default-interface-methods");
+				if (EnableInterfaceMembersPreview && SupportsCSharp8)
+					sw.WriteLine ("--lang-features=interface-constants,default-interface-methods");
+			}
 
+			Log.LogDebugMessage ("[Generator] response file {0}: {1}", responseFile, File.ReadAllText (responseFile));
```
jonathanpeppers, line 164: Instead of reading the file back into memory, could we make a helper method that logs and writes for each call?

Something like:
```csharp
LogDebugMessage ("[Generator] response file: {0}", responseFile);

WriteLine (sw, "--global");
WriteLine (sw, "--public");

// The other switches...

void WriteLine (StreamWriter sw, string line)
{
    sw.WriteLine (line);
    Log.LogDebugMessage (line);
}
```

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
```diff
@@ -110,46 +110,60 @@ public override bool RunTask ()
...
-			}
-			
-			if (ReferencedManagedLibraries != null)
-				foreach (var lib in ReferencedManagedLibraries)
-					cmd.AppendSwitchIfNotNull ("--ref=", Path.GetFullPath (lib.ItemSpec));
-			if (AnnotationsZipFiles != null)
-				foreach (var zip in AnnotationsZipFiles)
-					cmd.AppendSwitchIfNotNull ("--annotations=", Path.GetFullPath (zip.ItemSpec));
+				foreach (var tf in transform_files)
+					sw.WriteLine ($"--{tf.Item2}={tf.Item1}");
```
jonpryor, line 146: If `transform_files` contains filenames, then we should probably quote this *entire* string:

```csharp
sw.WriteLine ($"\"--{tf.Item2}={tf.Item1}\"");
```

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
dellis1972: Yup I thought about doing that lol.

PR #4980, opened 2020-08-05
[tests, build] multi-target Xamarin.Android.Build.Tests

jonathanpeppers:
There are various tests in `Xamarin.Android.Build.Tests` that execute
MSBuild tasks directly. This means the changes in #4820 are not quite
complete. We need to actually *run* the tests under .NET Core to see
if these tests would pass under a One .NET context.

The solution (you would think), would be to make the test assembly
`netstandard2.0` so you could run either:

    > nunit3-console Xamarin.Android.Build.Tests.dll
    > dotnet test Xamarin.Android.Build.Tests.dll

`nunit3-console` would execute the tests under .NET framework/Mono and
`dotnet test` would run under .NET Core or .NET 5+.

Unfortunately, `nunit3-console` can't run tests from a
`netstandard2.0` library!

So we multi-target:

    <TargetFrameworks>net472;net5.0</TargetFrameworks>

Unfortunately, the `net5.0` causes MSBuild to fail with:

    > msbuild Xamarin.Android.sln
    ...
    (GetReferenceAssemblyPaths target) ->
      C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(1177,5): error MSB3644: The reference assemblies for .NETFramework,Version=v5.0 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks

I think there is some expectation to use `dotnet build` for .NET 5
instead. We don't want to move our *entire* build over to `dotnet
build` yet (we probably can't!), so instead we can do:

    <TargetFrameworks>net472;netcoreapp3.1</TargetFrameworks>

This allows `msbuild Xamarin.Android.sln` to continue to work. This
might also be something that gets fixed in a future MSBuild version.

I updated various `.yaml` definitions so we can run NUnit tests via
`nunit3-console` or `dotnet test` as needed.

Other things that broke here:

* We need `Microsoft.NET.Test.Sdk` for `dotnet test` to work.
* `Builder.UseDotNet` can now be set based on a `#if` instead of an
  NUnit parameter.
* `Xamarin.Android.Tools.Aidl` needs to be `netstandard2.0`. This
  required including a `System.CodeDom` package.
* `Xamarin.Android.Build.Tasks` can't use `FSharp.Core` in
  `netcoreapp3.1`. I just removed it, since it is not used.
* Updated various paths to `Xamarin.Android.Build.Tests.dll`.

### DIFF
--- a/build-tools/installers/create-installers.targets
+++ b/build-tools/installers/create-installers.targets
@@ -121,7 +121,6 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\cil-strip.pdb"   Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\class-parse.exe" Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\class-parse.pdb" Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\FSharp.Core.dll" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\generator.exe"   Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\generator.pdb"   Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\HtmlAgilityPack.dll" />
@@ -401,6 +400,7 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\lib\x86\installer" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\lib\x86_64\installer" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\protobuf-net.dll" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\System.CodeDom.dll" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\System.Collections.Immutable.dll" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.AndroidTools.dll" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.AndroidTools.pdb" />
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
@@ -270,9 +270,6 @@
       BeforeTargets="CoreCompile"
       Inputs="@(_SharedRuntimeAssemblies)"
       Outputs="$(_GeneratedProfileClass)">
-    <ItemGroup>
-      <_SharedRuntimeAssemblies Remove="$(_SharedRuntimeBuildPath)v1.0\FSharp.Core.dll" />
-    </ItemGroup>
     <GenerateProfile Files="@(_SharedRuntimeAssemblies)" OutputFile="$(_GeneratedProfileClass)" />
     <WriteLinesToFile
         File="$(IntermediateOutputPath)$(CleanFile)"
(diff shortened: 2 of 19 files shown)


jonathanpeppers:
Here is a build with a full run of the `One .NET` tests: https://build.azdo.io/3979182

build-tools/installers/create-installers.targets
```diff
@@ -121,7 +121,6 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\cil-strip.pdb"   Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\class-parse.exe" Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\class-parse.pdb" Condition=" '$(PackageId)' != 'Microsoft.Android.Sdk' " />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\FSharp.Core.dll" />
```
pjcollins, line 124: If we're removing this we may have a couple other areas that could be worth cleaning up:
https://github.com/xamarin/xamarin-android/blob/master/build-tools/create-vsix/symbols/ExternalWhiteList.csv#L9
and 
https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets#L273-L275

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
```diff
@@ -270,9 +270,6 @@
       BeforeTargets="CoreCompile"
       Inputs="@(_SharedRuntimeAssemblies)"
       Outputs="$(_GeneratedProfileClass)">
-    <ItemGroup>
-      <_SharedRuntimeAssemblies Remove="$(_SharedRuntimeBuildPath)v1.0\FSharp.Core.dll" />
-    </ItemGroup>
```
jonathanpeppers, line 275: I checked `src\Xamarin.Android.Build.Tasks\obj\Debug\Profile.g.cs` and `FSharp.Core.dll` isn't being written to it at all, anyway:

```csharp
∩╗┐using System.Collections.Generic;

namespace Xamarin.Android.Tasks {
	public partial class Profile {
		// KEEP THIS SORTED ALPHABETICALLY, CASE-INSENSITIVE
		public static readonly string [] SharedRuntimeAssemblies = new []{
			"I18N.CJK.dll",
			"I18N.dll",
			"I18N.MidEast.dll",
			"I18N.Other.dll",
			"I18N.Rare.dll",
			"I18N.West.dll",
			"Java.Interop.dll",
			"Microsoft.CSharp.dll",
			"Mono.Android.dll",
			"Mono.Android.Export.dll",
			"Mono.Btls.Interface.dll",
			"Mono.CompilerServices.SymbolWriter.dll",
			"Mono.CSharp.dll",
			"Mono.Data.Sqlite.dll",
			"Mono.Data.Tds.dll",
			"Mono.Posix.dll",
			"Mono.Security.dll",
			"mscorlib.dll",
			"OpenTK.dll",
			"OpenTK-1.0.dll",
			"System.ComponentModel.Composition.dll",
			"System.ComponentModel.DataAnnotations.dll",
			"System.Core.dll",
			"System.Data.DataSetExtensions.dll",
			"System.Data.dll",
			"System.Data.Services.Client.dll",
			"System.dll",
			"System.EnterpriseServices.dll",
			"System.IdentityModel.dll",
			"System.IO.Compression.dll",
			"System.IO.Compression.FileSystem.dll",
			"System.Json.dll",
			"System.Net.dll",
			"System.Net.Http.dll",
			"System.Net.Http.WinHttpHandler.dll",
			"System.Numerics.dll",
			"System.Numerics.Vectors.dll",
			"System.Reflection.Context.dll",
			"System.Runtime.CompilerServices.Unsafe.dll",
			"System.Runtime.Serialization.dll",
			"System.Security.dll",
			"System.ServiceModel.dll",
			"System.ServiceModel.Internals.dll",
			"System.ServiceModel.Web.dll",
			"System.Transactions.dll",
			"System.Web.Services.dll",
			"System.Windows.dll",
			"System.Xml.dll",
			"System.Xml.Linq.dll",
			"System.Xml.Serialization.dll",
		};
	}
}
```

Issue #4565, opened 2020-04-16
Release builds fail if imported AAR library contains a lint.jar file


davidschreiber:
When building a Xamarin-Android app using an AAR library that contains a `lint.jar` file, release builds fail due to Xamarin-Android trying to process the `lint.jar` file using R8.

### Steps to Reproduce

1. Use a AAR library binding, where the original AAR file contains a `lint.jar` file. The possibility of having a `lint.jar` is described here: https://developer.android.com/studio/projects/android-library#aar-contents
2. Perform a release build, so that the R8 task is executed.
3. Observe error messages like this:

```
1>  AndroidSample -> ...\bin\Release\AndroidSample.dll
1>R8 : warning : Missing class: org.conscrypt.ConscryptHostnameVerifier
1>R8 : warning : Missing class: com.android.tools.lint.detector.api.Detector
1>R8 : warning : Missing class: com.android.tools.lint.detector.api.Detector$UastScanner
1>R8 : warning : Missing class: com.android.tools.lint.client.api.UElementHandler
1>R8 : warning : Missing class: com.android.tools.lint.detector.api.Detector$XmlScanner
1>R8 : warning : Missing class: com.android.tools.lint.client.api.IssueRegistry
1>R8 : error : Compilation can't be completed because some library classes are missing.
```

### Expected Behavior

The release build should succeed without errors. The `lint.jar` file should be excluded from R8, as well as from any other build steps, as it is not meant to be included inside the app (it is only used by the IDE to provide linter warnings).

### Actual Behavior

R8 fails to minify the project's classes, because it picks up the `lint.jar` file and tries to minimize it too.

I assume that this is the case, because all JAR files inside the AAR are used as library class files, and the `lint.jar` file is not properly filtered, for example here:

https://github.com/xamarin/xamarin-android/blob/7249199988f42dd2f8da1035cb2e7a334f0ad28a/src/Xamarin.Android.Build.Tasks/Tasks/ResolveLibraryProjectImports.cs#L354-L358

Instead, the `lint.jar` file is only important for the IDE, and should not be included on the compile classpath or runtime classpath of the app.

### Version Information

Microsoft Visual Studio Community 2019
Version 16.4.5
VisualStudio.16.Release/16.4.5+29806.167
Microsoft .NET Framework
Version 4.8.03752

Installed Version: Community

Visual C++ 2019   00435-60000-00000-AA767
Microsoft Visual C++ 2019

Microsoft Visual Studio VC Package   1.0
[...]

PR #4982, opened 2020-08-05
[Xamarin.Android.Build.Tasks] fix for lint.jar and @(LibraryProjectZip)

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/4565
Context: https://github.com/xamarin/xamarin-android/issues/4565#issuecomment-668341395

In f230b87c, I fixed `lint.jar` from being included in apps from the
`@(AndroidAarLibrary)` item group. Unfortunately, this did not fix the
scenario for `@(LibraryProjectZip)` in binding projects!

I moved the `skipCallback` used in the
`<ResolveLibraryProjectImports/>` to a common place in the `Files`
class so it can also be used by the `<CreateLibraryResourceArchive/>`
MSBuild task.  We need to use this for `.aar` files *only*, because
`.zip` files could be *any* format.

I brought over Brendan's test that reproduces the bug as well.

### DIFF
--- a/Documentation/release-notes/4982.md
+++ b/Documentation/release-notes/4982.md
@@ -1,0 +1,5 @@
+#### Binding projects
+
+* [GitHub Issue 4565](https://github.com/xamarin/xamarin-android/issues/4565):
+  *Missing class: com.android.tools.lint...* build error could prevent apps from
+  consuming custom binding libraries for `.AAR` files.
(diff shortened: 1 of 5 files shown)


jonathanpeppers:
Whoops, I'll add release notes for this tomorrow.

PR #4993, opened 2020-08-10
Work around Android 9+ bug with extractNativeLibs flag

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/4983
Context: https://developer.android.com/reference/android/content/pm/ApplicationInfo.html#FLAG_EXTRACT_NATIVE_LIBS

Xamarin.Android has supported the [extractNativeLibs][0]
attribute (introduced by Android API 23) on the `<application>` element
in AndroidManifest.xml since 95ca1025f0b7e678055ea922a4091b8b9710b18f,
with a single significant modification in
feb9ea2a53034b73d5bd62ed3caea4e8f658d1e5 after we discovered that
Android build system can set the flag when constructing the APK after we
are done packaging.

feb9ea2a5 introduced a runtime check to see whether the
`FLAG_EXTRACT_NATIVE_LIBS` is *not* set, which meant that the native
libraries are to stay in the APK file and we need to set up our DSO
search paths to point to the inside of APK files instead of the
traditional filesystem location.

However, it appears that Android 10 (API 29, on both devices and in emulators)
and Android 9 (API 28, on *just* the devices) broke the
`FLAG_EXTRACT_NATIVE_LIBS` semantics in that it is possible for the flag
to be *set* (which means libraries are *extracted*) with the libraries
not extracted from APKs, thus breaking the logic implemented in
feb9ea2a5.

It is possible that other Android versions are affected as well, which
means we can no longer trust the flag value and need to implement
another way of checking whether the libraries are on the filesystem or
in the APK.  The simplest approach is to check for existence of a known
library in the filesystem location, regardless of the API level, and
assume the flag is *not* set if the shared library is missing.  This is
what this commit implements.  The check is performed once on application
startup, thus minimizing the performance impact.

[0]: https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs

Co-Authored by: Jonathan Peppers (@jonathanpeppers)

### DIFF
--- a/src/monodroid/jni/basic-android-system.cc
+++ b/src/monodroid/jni/basic-android-system.cc
@@ -1,6 +1,7 @@
 #include <cerrno>
 
 #include "basic-android-system.hh"
+#include "cpp-util.hh"
 #include "globals.hh"
 
 using namespace xamarin::android;
@@ -12,13 +13,29 @@ size_t BasicAndroidSystem::app_lib_directories_size = 0;
 const char* BasicAndroidSystem::built_for_abi_name = nullptr;
 
 void
-BasicAndroidSystem::setup_app_library_directories (jstring_array_wrapper& runtimeApks, jstring_array_wrapper& appDirs, int androidApiLevel)
+BasicAndroidSystem::detect_embedded_dso_mode (jstring_array_wrapper& appDirs) noexcept
 {
-	if (androidApiLevel < 23 || !is_embedded_dso_mode_enabled ()) {
+	// appDirs[2] points to the native library directory
+	simple_pointer_guard<char[]> libmonodroid_path = utils.path_combine (appDirs[2].get_cstr (), "libmonodroid.so");
+	log_debug (LOG_ASSEMBLY, "Checking if libmonodroid was unpacked to %s", libmonodroid_path.get ());
+	if (!utils.file_exists (libmonodroid_path)) {
+		log_debug (LOG_ASSEMBLY, "%s not found, assuming application/android:extractNativeLibs == false", libmonodroid_path.get ());
+		set_embedded_dso_mode_enabled (true);
+	} else {
+		log_debug (LOG_ASSEMBLY, "Native libs extracted to %s, assuming application/android:extractNativeLibs == true", appDirs[2].get_cstr ());
+		set_embedded_dso_mode_enabled (false);
+	}
+}
+
+void
+BasicAndroidSystem::setup_app_library_directories (jstring_array_wrapper& runtimeApks, jstring_array_wrapper& appDirs)
+{
+	if (!is_embedded_dso_mode_enabled ()) {
 		log_info (LOG_DEFAULT, "Setting up for DSO lookup in app data directories");
 		BasicAndroidSystem::app_lib_directories_size = 1;
 		BasicAndroidSystem::app_lib_directories = new const char*[app_lib_directories_size]();
 		BasicAndroidSystem::app_lib_directories [0] = utils.strdup_new (appDirs[2].get_cstr ());
+		log_debug (LOG_ASSEMBLY, "Added filesystem DSO lookup location: %s", appDirs[2].get_cstr ());
 	} else {
 		log_info (LOG_DEFAULT, "Setting up for DSO lookup directly in the APK");
 		BasicAndroidSystem::app_lib_directories_size = runtimeApks.get_length ();
@@ -48,6 +65,7 @@ BasicAndroidSystem::add_apk_libdir (const char *apk, size_t index, [[maybe_unuse
 	assert (user_data != nullptr);
 	assert (index < app_lib_directories_size);
 	app_lib_directories [index] = utils.string_concat (apk, "!/lib/", static_cast<const char*>(user_data));
+	log_debug (LOG_ASSEMBLY, "Added APK DSO lookup location: %s", app_lib_directories[index]);
 }
 
 void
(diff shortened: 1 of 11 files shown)


src/monodroid/jni/basic-android-system.cc
```diff
@@ -12,13 +13,29 @@ size_t BasicAndroidSystem::app_lib_directories_size = 0;
...
+	simple_pointer_guard<char[]> libmonodroid_path = utils.path_combine (appDirs[2].get_cstr (), "libmonodroid.so");
+	log_debug (LOG_ASSEMBLY, "Checking if libmonodroid was unpacked to %s", libmonodroid_path.get ());
+	if (!utils.file_exists (libmonodroid_path)) {
+		log_debug (LOG_ASSEMBLY, "%s not found, assuming application/android:extractNativeLibs == false", libmonodroid_path.get ());
+		set_embedded_dso_mode_enabled (true);
+	} else {
+		log_debug (LOG_ASSEMBLY, "Native libs extracted to %s, assuming application/android:extractNativeLibs == true", appDirs[2].get_cstr ());
+		set_embedded_dso_mode_enabled (false);
+	}
+}
```
jonathanpeppers, line 28: If a user has `AndroidFastDeploymentType=Assemblies::Dexes`, to enable fast deployment of native libraries & dexes.

Would this `file_exists` check also fail in that case?

grendello: Yes, because the path refers to the system-created library path, not the override directories. With fastdev we will check the override locations anyway, regardless of this setting.

src/monodroid/jni/basic-android-system.cc
```diff
@@ -12,13 +13,29 @@ size_t BasicAndroidSystem::app_lib_directories_size = 0;
 const char* BasicAndroidSystem::built_for_abi_name = nullptr;
 
 void
-BasicAndroidSystem::setup_app_library_directories (jstring_array_wrapper& runtimeApks, jstring_array_wrapper& appDirs, int androidApiLevel)
+BasicAndroidSystem::detect_embedded_dso_mode (jstring_array_wrapper& appDirs) noexcept
 {
-	if (androidApiLevel < 23 || !is_embedded_dso_mode_enabled ()) {
+	// appDirs[2] points to the native library directory
+	simple_pointer_guard<char[]> libmonodroid_path = utils.path_combine (appDirs[2].get_cstr (), "libmonodroid.so");
```
jonpryor, line 19: Should this verify that `appDirs[2]` is a valid index before using it?  Or will that  cause an immediate exception/SIGSEGV if `appDirs[2]` is invalid?

jonpryor: Won't cause an exception; instead `appDirs[2]` will return an "invalid string", and `.get_cstr()` will return `nullptr`.

jonpryor: Result is that if `appDris[2]` isn't valid then `libmonodroid_path` will be `"libmonodroid.so"` (no directory info), which will thus depend upon "current-working-directory" behavior within **stat**(2).

grendello: It's tied tightly with our Java code - if we don't send enough indices, let it segfault, we'll notice it and fix.

jonpryor:
Copy-edited squash-and-merge message:

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/4983

Context: https://developer.android.com/reference/android/content/pm/ApplicationInfo.html#FLAG_EXTRACT_NATIVE_LIBS

Android API-23 add support for
[`//application/@android:extractNativeLibs`][0], which allows an app
to state that native libs within the `.apk` should *not* be extracted
as part of app installation, to save on overall installation size.

Support for `extractNativeLibs` was added in 95ca1025 and feb9ea2a.

Unfortunately, we have determined that on some Android 9 devices and
Android 10 devices & emulators, the [`FLAG_EXTRACT_NATIVE_LIBS`][1]
check used in commit feb9ea2a doesn't work as intended or expected:

	boolean embeddedDSOsEnabled = (runtimePackage.flags & ApplicationInfo.FLAG_EXTRACT_NATIVE_LIBS) == 0;

What we observe on the offending devices is that `embeddedDSOsEnabled`
is True when `//application/@extractNativeLibs` is true, which is
the *opposite* of what we expect.  Consequently, when we *should* be
looking for native libs "on-disk", we don't!

The result is that we can't find `libmonodroid.so`:

	DllImport error loading library '__Internal': 'Could not load library: Library '/system/lib/libmonodroid.so' not found.'.

Note that *path* that is being checked: `/system/lib/libmonodroid.so`,
because we *think* we should be checking within the `.apk`, but the
`.apk` doesn't contain a loadable `libmonodroid.so` file, so we
"fallback" to various other paths, including `/system/lib`, none
of which have `libmonodroid.so`.

The result is that, eventually, the app crashes during startup:

	Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x6f63736d in tid 7607 (…), pid 7607 (…)

It is possible that other Android versions are affected as well, which
means we can no longer trust the flag value and need to implement
another way of checking whether the libraries are on the filesystem or
in the `.apk`.

Work around this Android issue by checking for the presence of a known
library in a filesystem location, regardless of the API level.
Specifically, check for the presence of `libmonodroid.so`.
If `libmonodroid.so` is not found, assume that all libraries should be
loaded from the `.apk`, as if `//application/@extractNativeLibs`=false.

The check is performed once on application startup, thus minimizing the
performance impact.

[0]: https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs
[1]: https://developer.android.com/reference/android/content/pm/ApplicationInfo#FLAG_EXTRACT_NATIVE_LIBS

Co-Authored by: Jonathan Peppers (@jonathanpeppers)
```

PR #4999, opened 2020-08-11
[runtime] Introduce simple local/stack string classes

grendello:
This commit introduces a couple of top-level classes to work with
strings allocated either on stack or from the heap, which are always
released when leaving the scope.  The classes allocate string buffer on
stack if the size doesn't exceed a certain maximum value (provided as a
template parameter) or, if the space needed is more than size of the
on-stack buffer, allocate memory from heap as necessary.  If we choose
the stack allocation size correctly, this has the advantage of being
very quick since no calls are necessary for (de)allocation of memory -
it's either part of the function prolog/epilog where appropriate stack
for local variables space is reserved, or part of a scope setup which
operation also isn't expensive.

Two introduced classes are `static_local_string` and
`dynamic_local_string`.  The former is useful in situations when we know
the string size beforehand and we know it will never exceed that size.
The class doesn't support resizing of the buffer.  The latter adds the
ability to extend the available space, possibly transfering data from
the stack to a dynamically allocated buffer.

Both classes derive from a single base class which can be configured to
use any kind of memory storage implementation, if it's needed in the
future.

= Performance changes (app Displayed Time)

Device:
  - Model: **Pixel 3 XL**
  - SDK version: **29**

== Release
Release performance borders on the statistical error, but it is still
slightly better (on slower devices it will be more pronounced):

| Before  | After   | Δ        | Notes                          |
| ------- | ------- | -------- | ------------------------------ |
| 859.000 | 862.800 | +0.44% ✗ | preload enabled; 32-bit build  |
| 853.400 | 851.800 | -0.19% ✓ | preload disabled; 32-bit build |
| 867.600 | 852.200 | -1.78% ✓ | preload enabled; 64-bit build  |
| 855.000 | 858.000 | +0.35% ✗ | preload disabled; 64-bit build |

== Debug
Debug performance has bigger gains because the runtime does much more
parsing and string processing than the Release one:

| Before   | After    | Δ        | Notes                          |
| -------- | -------- | -------- | ------------------------------ |
| 1124.700 | 1113.400 | -1.00% ✓ | preload enabled; 32-bit build  |
| 1120.500 | 1114.200 | -0.56% ✓ | preload disabled; 32-bit build |
| 1122.700 | 1123.000 | +0.03% ✗ | preload enabled; 64-bit build  |
| 1113.700 | 1112.100 | -0.14% ✓ | preload disabled; 64-bit build |

### DIFF
--- a/src/monodroid/jni/android-system.cc
+++ b/src/monodroid/jni/android-system.cc
@@ -6,10 +6,6 @@
 #include <ctype.h>
 #include <fcntl.h>
 
-#ifdef ANDROID
-#include <sys/system_properties.h>
-#endif
-
 #if defined (WINDOWS)
 #include <windef.h>
 #include <winbase.h>
@@ -51,11 +47,6 @@ BundledProperty *AndroidSystem::bundled_properties = nullptr;
 #if defined (WINDOWS)
 std::mutex AndroidSystem::readdir_mutex;
 char *AndroidSystem::libmonoandroid_directory_path = nullptr;
-#endif
-
-#if !defined (ANDROID)
-static constexpr uint32_t PROP_NAME_MAX = 32;
-static constexpr uint32_t PROP_VALUE_MAX = 92;
 #endif
 
 #if defined (DEBUG) || !defined (ANDROID)
@@ -207,8 +198,8 @@ AndroidSystem::_monodroid__system_property_get (const char *name, char *sp_value
 		return -1;
 
 	char *buf = nullptr;
-	if (sp_value_len < PROP_VALUE_MAX + 1) {
-		size_t alloc_size = ADD_WITH_OVERFLOW_CHECK (size_t, PROP_VALUE_MAX, 1);
+	if (sp_value_len < PROPERTY_VALUE_BUFFER_LEN) {
+		size_t alloc_size = ADD_WITH_OVERFLOW_CHECK (size_t, PROPERTY_VALUE_BUFFER_LEN, 1);
 		log_warn (LOG_DEFAULT, "Buffer to store system property may be too small, will copy only %u bytes", sp_value_len);
 		buf = new char [alloc_size];
 	}
@@ -225,12 +216,31 @@ AndroidSystem::_monodroid__system_property_get (const char *name, char *sp_value
 #endif
 
 int
+AndroidSystem::monodroid_get_system_property (const char *name, dynamic_local_string<PROPERTY_VALUE_BUFFER_LEN>& value)
+{
+	int len = _monodroid__system_property_get (name, value.get (), value.size ());
+	if (len > 0) {
+		// Clumsy, but if we want direct writes to be fast, this is the price we pay
+		value.set_length_after_direct_write (static_cast<size_t>(len));
+		return len;
+	}
+
+	size_t plen;
+	const char *v = lookup_system_property (name, plen);
+	if (v == nullptr)
+		return len;
+
+	value.assign (v, plen);
+	return ADD_WITH_OVERFLOW_CHECK (int, plen, 0);
+}
+
+int
 AndroidSystem::monodroid_get_system_property (const char *name, char **value)
 {
 	if (value)
 		*value = nullptr;
 
-	char  sp_value [PROP_VALUE_MAX+1] = { 0, };
+	char  sp_value [PROPERTY_VALUE_BUFFER_LEN] = { 0, };
 	char *pvalue = sp_value;
 	int len = _monodroid__system_property_get (name, sp_value, sizeof (sp_value));
 
@@ -334,18 +344,20 @@ AndroidSystem::create_update_dir (char *override_dir)
 	log_warn (LOG_DEFAULT, "Creating public update directory: `%s`", override_dir);
 }
 
-char*
-AndroidSystem::get_full_dso_path (const char *base_dir, const char *dso_path, bool &needs_free)
+bool
+AndroidSystem::get_full_dso_path (const char *base_dir, const char *dso_path, dynamic_local_string<SENSIBLE_PATH_MAX>& path)
 {
-	needs_free = false;
 	if (dso_path == nullptr)
-		return nullptr;
+		return false;
 
 	if (base_dir == nullptr || utils.is_path_rooted (dso_path))
 		return const_cast<char*>(dso_path); // Absolute path or no base path, can't do much with it
 
-	needs_free = true;
-	return utils.path_combine (base_dir, dso_path);
+	path.assign (base_dir)
+		.append (MONODROID_PATH_SEPARATOR)
+		.append (dso_path);
+
+	return true;
 }
 
 void*
@@ -375,13 +387,12 @@ AndroidSystem::load_dso_from_specified_dirs (const char **directories, size_t nu
 	if (dso_name == nullptr)
 		return nullptr;
 
-	bool needs_free = false;
-	char *full_path = nullptr;
+	dynamic_local_string<SENSIBLE_PATH_MAX> full_path;
 	for (size_t i = 0; i < num_entries; i++) {
-		full_path = get_full_dso_path (directories [i], dso_name, needs_free);
-		void *handle = load_dso (full_path, dl_flags, false);
-		if (needs_free)
-			delete[] full_path;
+		if (!get_full_dso_path (directories [i], dso_name, full_path)) {
+			continue;
+		}
+		void *handle = load_dso (full_path.get (), dl_flags, false);
 		if (handle != nullptr)
 			return handle;
 	}
@@ -414,43 +425,35 @@ AndroidSystem::load_dso_from_any_directories (const char *name, unsigned int dl_
 	return handle;
 }
 
-char*
-AndroidSystem::get_existing_dso_path_on_disk (const char *base_dir, const char *dso_name, bool &needs_free)
+bool
+AndroidSystem::get_existing_dso_path_on_disk (const char *base_dir, const char *dso_name, dynamic_local_string<SENSIBLE_PATH_MAX>& path)
 {
-	needs_free = false;
-	char *dso_path = get_full_dso_path (base_dir, dso_name, needs_free);
-	if (utils.file_exists (dso_path))
-		return dso_path;
+	if (get_full_dso_path (base_dir, dso_name, path) && utils.file_exists (path.get ()))
+		return true;
 
-	needs_free = false;
-	delete[] dso_path;
-	return nullptr;
+	return false;
 }
 
-char*
-AndroidSystem::get_full_dso_path_on_disk (const char *dso_name, bool &needs_free)
+bool
+AndroidSystem::get_full_dso_path_on_disk (const char *dso_name, dynamic_local_string<SENSIBLE_PATH_MAX>& path)
 {
-	needs_free = false;
 	if (is_embedded_dso_mode_enabled ())
-		return nullptr;
+		return false;
 
-	char *dso_path = nullptr;
 #ifndef RELEASE
 	for (size_t i = 0; i < AndroidSystem::MAX_OVERRIDES; i++) {
 		if (AndroidSystem::override_dirs [i] == nullptr)
 			continue;
-		dso_path = get_existing_dso_path_on_disk (AndroidSystem::override_dirs [i], dso_name, needs_free);
-		if (dso_path != nullptr)
-			return dso_path;
+		if (get_existing_dso_path_on_disk (AndroidSystem::override_dirs [i], dso_name, path))
+			return true;
 	}
 #endif
 	for (size_t i = 0; i < app_lib_directories_size; i++) {
-		dso_path = get_existing_dso_path_on_disk (app_lib_directories [i], dso_name, needs_free);
-		if (dso_path != nullptr)
-			return dso_path;
+		if (get_existing_dso_path_on_disk (app_lib_directories [i], dso_name, path))
+			return true;
 	}
 
-	return nullptr;
+	return false;
 }
 
 int
@@ -491,10 +494,10 @@ AndroidSystem::get_max_gref_count_from_system (void)
 		max = 51200;
 	}
 
-	char *override;
-	if (androidSystem.monodroid_get_system_property (Debug::DEBUG_MONO_MAX_GREFC, &override) > 0) {
+	dynamic_local_string<PROPERTY_VALUE_BUFFER_LEN> override;
+	if (androidSystem.monodroid_get_system_property (Debug::DEBUG_MONO_MAX_GREFC, override) > 0) {
 		char *e;
-		max       = strtol (override, &e, 10);
+		max       = strtol (override.get (), &e, 10);
 		switch (*e) {
 			case 'k':
 				e++;
@@ -508,10 +511,9 @@ AndroidSystem::get_max_gref_count_from_system (void)
 		if (max < 0)
 			max = INT_MAX;
 		if (*e) {
-			log_warn (LOG_GC, "Unsupported '%s' value '%s'.", Debug::DEBUG_MONO_MAX_GREFC, override);
+			log_warn (LOG_GC, "Unsupported '%s' value '%s'.", Debug::DEBUG_MONO_MAX_GREFC, override.get ());
 		}
 		log_warn (LOG_GC, "Overriding max JNI Global Reference count to %i", max);
-		delete[] override;
 	}
 	return max;
 }
--- a/src/monodroid/jni/basic-utilities.hh
+++ b/src/monodroid/jni/basic-utilities.hh
@@ -7,6 +7,7 @@
 #include <cstdlib>
 #include <cstdio>
 #include <cstring>
+#include <type_traits>
 
 #include <unistd.h>
 
@@ -15,73 +16,13 @@
 #include <dirent.h>
 
 #include "java-interop-util.h"
-
-#if __cplusplus >= 201703L
-#define UNUSED_ARG [[maybe_unused]]
-#else
-#if defined (__GNUC__)
-#define UNUSED_ARG __attribute__((__unused__))
-#else
-#define UNUSED_ARG
-#endif
-#endif
-
-#if WINDOWS
-#define MONODROID_PATH_SEPARATOR      "\\"
-#define MONODROID_PATH_SEPARATOR_CHAR '\\'
-#else
-#define MONODROID_PATH_SEPARATOR      "/"
-#define MONODROID_PATH_SEPARATOR_CHAR '/'
-#endif
-
-#if WINDOWS
-typedef struct _stat monodroid_stat_t;
-#define monodroid_dir_t _WDIR
-typedef struct _wdirent monodroid_dirent_t;
-#else
-typedef struct stat monodroid_stat_t;
-#define monodroid_dir_t DIR
-typedef struct dirent monodroid_dirent_t;
-#endif
-
-#define DEFAULT_DIRECTORY_MODE S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH
-
-#if defined (_MSC_VER)
-#define inline __inline
-#define force_inline __forceinline
-#elif defined (__GNUC__)
-#ifndef XA_LIKELY
-#define XA_LIKELY(expr) (__builtin_expect ((expr) != 0, 1))
-#endif
-
-#ifndef XA_UNLIKELY
-#define XA_UNLIKELY(expr) (__builtin_expect ((expr) != 0, 0))
-#endif
-
-#define force_inline inline __attribute__((always_inline))
-#endif
-
-#ifndef force_inline
-#define force_inline inline
-#endif
-
-#ifndef inline
-#define inline inline
-#endif
-
-#ifndef XA_LIKELY
-#define XA_LIKELY(expr) (expr)
-#endif
-
-#ifndef XA_UNLIKELY
-#define XA_UNLIKELY(expr) (expr)
-#endif
+#include "helpers.hh"
+#include "cpp-util.hh"
+#include "platform-compat.hh"
+#include "strings.hh"
 
 namespace xamarin::android
 {
-#define ADD_WITH_OVERFLOW_CHECK(__ret_type__, __a__, __b__) utils.add_with_overflow_check<__ret_type__>(__FILE__, __LINE__, (__a__), (__b__))
-#define MULTIPLY_WITH_OVERFLOW_CHECK(__ret_type__, __a__, __b__) utils.multiply_with_overflow_check<__ret_type__>(__FILE__, __LINE__, (__a__), (__b__))
-
 	class BasicUtilities
 	{
 	public:
@@ -94,7 +35,7 @@ namespace xamarin::android
 		char           **monodroid_strsplit (const char *str, const char *delimiter, size_t max_tokens);
 		char            *monodroid_strdup_printf (const char *format, ...);
 		char            *monodroid_strdup_vprintf (const char *format, va_list vargs);
-		char*            path_combine(const char *path1, const char *path2);
+		char*            path_combine (const char *path1, const char *path2);
 		void             create_public_directory (const char *dir);
 		int              create_directory (const char *pathname, mode_t mode);
 		void             set_world_accessable (const char *path);
@@ -102,6 +43,59 @@ namespace xamarin::android
 		bool             file_exists (const char *file);
 		bool             directory_exists (const char *directory);
 		bool             file_copy (const char *to, const char *from);
+
+
+		// Make sure that `buf` has enough space! This is by design, the methods are supposed to be fast.
+		template<size_t MaxStackSpace, typename TBuffer>
+		void path_combine (TBuffer& buf, const char* path1, const char* path2) noexcept
+		{
+			path_combine (buf, path1, path1 == nullptr ? 0 : strlen (path1), path2, path2 == nullptr ? 0 : strlen (path2));
+		}
+
+		// internal::static_local_string<MaxStackSpace>
+		template<size_t MaxStackSpace, typename TBuffer>
+		void path_combine (TBuffer& buf, const char* path1, size_t path1_len, const char* path2, size_t path2_len) noexcept
+		{
+			assert (path1 != nullptr || path2 != nullptr);
+
+			if (path1 == nullptr) {
+				buf.append (path2);
+				return;
+			}
+
+			if (path2 == nullptr) {
+				buf.append (path1);
+				return;
+			}
+
+			buf.append (path1, path1_len);
+			buf.append (MONODROID_PATH_SEPARATOR, MONODROID_PATH_SEPARATOR_LENGTH);
+			buf.append (path2, path2_len);
+		}
+
+		template<size_t MaxStackSpace>
+		void path_combine (internal::static_local_string<MaxStackSpace>& buf, const char* path1, const char* path2) noexcept
+		{
+			path_combine <MaxStackSpace, decltype(buf)> (buf, path1, path2);
+		}
+
+		template<size_t MaxStackSpace>
+		void path_combine (internal::static_local_string<MaxStackSpace>& buf, const char* path1, size_t path1_len, const char* path2, size_t path2_len) noexcept
+		{
+			path_combine <MaxStackSpace, decltype(buf)> (buf, path1, path1_len, path2, path2_len);
+		}
+
+		template<size_t MaxStackSpace>
+		void path_combine (internal::dynamic_local_string<MaxStackSpace>& buf, const char* path1, const char* path2) noexcept
+		{
+			path_combine <MaxStackSpace, decltype(buf)> (buf, path1, path2);
+		}
+
+		template<size_t MaxStackSpace>
+		void path_combine (internal::dynamic_local_string<MaxStackSpace>& buf, const char* path1, size_t path1_len, const char* path2, size_t path2_len) noexcept
+		{
+			path_combine <MaxStackSpace, decltype(buf)> (buf, path1, path1_len, path2, path2_len);
+		}
 
 		bool ends_with_slow (const char *str, const char *end)
 		{
@@ -137,7 +131,7 @@ namespace xamarin::android
 				return nullptr;
 			}
 
-			size_t alloc_size = add_with_overflow_check<size_t>(__FILE__, __LINE__, len, 1);
+			size_t alloc_size = ADD_WITH_OVERFLOW_CHECK (size_t, len, 1);
 			auto ret = new char[alloc_size];
 			memcpy (ret, s, len);
 			ret[len] = '\0';
@@ -154,13 +148,11 @@ namespace xamarin::android
 			return strdup_new (s, strlen (s));
 		}
 
-		// Without <type_traits> it's a little bit open for abuse (bad stuff will happen if
-		// a type different than `char*` is used to specialize the function and we can't
-		// assert this condition on compile time), but we can take that risk since it's
-		// internal, controlled API
-		template<typename StringType = const char*, typename ...Strings>
-		char* string_concat (const char *s1, StringType s2, Strings... strings)
+		template<typename CharType = char, typename ...Strings>
+		char* string_concat (const char *s1, const CharType* s2, Strings... strings)
 		{
+			assert_char_type<CharType> ();
+
 			size_t len = calculate_length (s1, s2, strings...);
 
 			char *ret = new char [len + 1];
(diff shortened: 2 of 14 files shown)


src/monodroid/jni/android-system.cc
```diff
@@ -222,13 +213,32 @@ AndroidSystem::_monodroid__system_property_get (const char *name, char *sp_value
...
+		return len;
+	}
+
+	size_t plen;
+	const char *v = lookup_system_property (name, plen);
+	if (v == nullptr)
+		return len;
+
+	value.assign (v, plen);
+	return static_cast<int>(plen);
```
jonpryor, line 232: I think this should be `ADD_WITH_OVERFLOW_CHECK (int, plen, 0)`, assuming that `ADD_WITH_OVERFLOW_CHECK()`/`__builtin_add_overflow()` will reasonably do what I want: ensure that we don't try to return a value larger than `MAX_INT`.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+#include <type_traits>
+#include <unistd.h>
+
+#include "platform-compat.hh"
+#include "logger.hh"
+
+namespace xamarin::android::internal
+{
+	constexpr size_t SENSIBLE_TYPE_NAME_LENGTH = 128;
+	constexpr size_t SENSIBLE_PATH_MAX = 512;
```
jonpryor, line 17: I suspect this could be 256.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+	// 64-bit unsigned or 64-bit signed with sign
+	constexpr size_t MAX_INTEGER_DIGIT_COUNT_BASE10 = 21;
+
+#if DEBUG
+	constexpr bool BoundsCheck = true;
+#else
+	constexpr bool BoundsCheck = true;
```
jonpryor, line 24: Why is this value conditional on `#if DEBUG` if both values are the same?  Should this instead be `false`?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -181,7 +182,7 @@ EmbeddedAssemblies::zip_read_cd_info (int fd, uint32_t& cd_offset, uint32_t& cd_
 		return false;
 	}
 
-	auto buf = new uint8_t[alloc_size];
+	simple_pointer_guard<uint8_t[]> buf (new uint8_t[alloc_size]);
```
jonpryor, line 185: a little late, but should `alloc_size` -- declared line 178 -- be a `const size_t` or `constexpr`?  It doesn't look like it's ever changed after declaration…

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -351,22 +351,20 @@ EmbeddedAssemblies::zip_read_field (uint8_t* buf, size_t buf_len, size_t index,
...
 {
 	if (!zip_ensure_valid_params (buf, buf_len, index, count)) {
 		return false;
 	}
 
-	size_t alloc_size = ADD_WITH_OVERFLOW_CHECK (size_t, count, 1);
-	characters = new char[alloc_size];
-	memcpy (characters, buf + index, count);
-	characters [count] = '\0';
+	memcpy (characters.get (), buf + index, count);
```
jonpryor, line 360: This feels unsafe without some form of check that `count` < `SENSIBLE_PATH_MAX`.  Quick codepath check, and I see that in `zip_read_entry_info()`, `zip_read_field()` is called with `count` coming from `file_name_length`, a `uint16_t`.  It is thus *plausible*, without further checking, that `count` could be `MAX_SHORT`, or 64k, which can be longer than `SENSIBLE_PATH_MAX`.

I think instead there should be a `dynamic_local_string::assign(char* source, size_t offset, size_t count)` method, which can appropriately reallocate the underlying storage if necessary.

jonpryor: I see that there's already a `dynamic_local_string::append()` member function.  Why not use that instead?

```cpp
characters.clear ();
characters.assign (buf + index, count);
```

src/monodroid/jni/platform-compat.hh
```diff
@@ -0,0 +1,66 @@
+#ifndef __PLATFORM_COMPAT_HH
+#define __PLATFORM_COMPAT_HH
+
+#if __cplusplus >= 201703L
+#define UNUSED_ARG [[maybe_unused]]
+#else
```
jonpryor, line 6: Please, as per convention elsewhere in `src/monodroid/jni`, include comments on `#else` and `#endif` so that mere mortals (read: me) can more easily determine what a given block is *for*.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+
+			if (_length != s_length) {
+				return false;
+			}
+
+			if (_length == 0) {
+				return true;
+			}
+
+			return memcmp (_start, s, length ()) == 0;
```
jonpryor, line 72: Why use `_length` in some places and `length()` in others, in the same method?

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+		}
+
+		force_inline bool starts_with (const char *s) const noexcept
```
jonpryor, line 75: I feel like there should be an overload for string constants here, too:

```cpp
template<size_t Size>
force_inline bool starts_with (const char (&s)[Size]) const noexcept
{
    return starts_with (s, Size);
}
```

This avoids the `strlen(s)` call.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+		{
+			if (s == nullptr)
+				return false;
+
+			return starts_with (s, strlen (s));
+		}
+
+		force_inline bool starts_with (const char *s, size_t s_length) const noexcept
+		{
+			if (!can_access (s_length)) {
```
jonpryor, line 85: This overload should *also* verify that `s != nullptr`, lest the `memcmp()` assert.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+			static_assert (MaxStackSize > 0, "MaxStackSize must be more than 0");
+			init_store (size < MaxStackSize ? MaxStackSize : size);
+		}
+
+		virtual ~local_storage ()
+		{
+			free_store ();
+		}
+
+		T* get () const noexcept
```
jonpryor, line 219: I think this should be overloaded based on "`const`-ness":

```cpp
const T* get () const noexcept
{
    return …
}

T *get () noexcept
{
    return …
}
```

The presence of the `const_cast<T*>()` is itself a smell, and wouldn't be needed if `get()` were properly overloaded.

Additionally, as-is there's a potential runtime assert here (unless C++ has changed in this area in the past 20 years), e.g. if you have a global `const local_storage<10, false> v` instance, which is placed by the linker into read-only memory, then `*v.get() = 'a'` will crash.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+			if (token.initialized ())
+				assign (token.start (), token.length ());
+		}
+
+		force_inline size_t index () const noexcept
+		{
+			return idx;
+		}
+
+		force_inline size_t length () const noexcept
```
jonpryor, line 370: Why have two member functions with the same semantics, `index()` and `length()`?

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+
+		force_inline void clear () noexcept
+		{
+			set_index (0);
+			buffer.get ()[0] = NUL;
+		}
+
+		force_inline void terminate () noexcept
+		{
+			buffer.get ()[idx] = NUL;
```
jonpryor, line 393: How do we know that `idx` is a valid index here?

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+			}
+
+			std::memcpy (buffer.get () + idx, s, length);
+			idx += length;
+			buffer.get ()[idx] = NUL;
+
+			return *this;
+		}
+
+		string_base& append (const char *s) noexcept
```
jonpryor, line 413: Ditto earlier comment about overloading `append()` to support string literals w/o requiring `strlen()`

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+
+		string_base& assign (const char *s) noexcept
+		{
+			if (s == nullptr)
+				return *this;
+
+			return assign (s, strlen (s));
+		}
+
+		force_inline bool next_token (size_t start_index, const TChar separator, string_segment& token) const noexcept
```
jonpryor, line 477: Does "tokenization" *really* belong in the same type?

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+				LOG_DEFAULT,
+				"Index %u is out of range (0 - %u)",
+				access_index, index ()
+			);
+			exit (1);
+		}
+
+		force_inline void ensure_enough_space (size_t length) noexcept
+		{
+			if (index () + length > buffer.size ()) {
```
jonpryor, line 676: `ensure_enough_space()` is not consistent with `resize_if_needed()`: the latter has an additional `+1` in the calculation, for the terminating null.

jonpryor: All addition should use `ADD_WITH_OVERFLOW_CHECK()`.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+				);
+				exit (1);
+			}
+		}
+
+		force_inline void resize_if_needed (size_t needed_space) noexcept
+		{
+			if constexpr (TStorage::has_resize) {
+				size_t current_size = buffer.size ();
+				if (index () + needed_space + 1 > current_size) {
```
jonpryor, line 690: All addition should use `ADD_WITH_OVERFLOW_CHECK()`.

src/monodroid/jni/strings.hh
jonpryor: Reviewing this PR, `ensure_enough_space(42)`/`resize_if_needed(42)` are not doing me any favors, in that it isn't explicit if `42` is the *total* size required, or the *additional* size required.  Perhaps `ensure_have_extra(42)`/`resize_for_extra(42)` would be clearer?

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+					buffer.resize (current_size + (current_size / 2) + needed_space + 1);
```
jonpryor, line 691: Use our appropriate macros to prevent overflow & underflow.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+			return *this;
+		}
+
+		force_inline string_base& assign (const TChar* s, size_t length) noexcept
+		{
+			idx = 0;
+			return append (s, length);
+		}
+
+		string_base& assign (const char *s) noexcept
```
jonpryor, line 469: This should be overloaded for string literals, to avoid the `strlen()`.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+
+		force_inline bool starts_with (const TChar *s, size_t s_length) noexcept
+		{
+			if (s == nullptr || s_length == 0 || s_length > buffer.size ())
+				return false;
+
+			return memcmp (buffer.get (), s, s_length) == 0;
+		}
+
+		force_inline bool starts_with (const char* s) const noexcept
```
jonpryor, line 537: Please overload for string literals.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+
+		force_inline void set_length_after_direct_write (size_t new_length) noexcept
+		{
+			set_index (new_length);
+			terminate ();
+		}
+
+		force_inline void set_at (size_t index, const TChar ch) noexcept
+		{
+			if (index >= buffer.size ())
```
jonpryor, line 553: Why do some "out of range" checks result in *process abort* -- e.g. `ensure_enough_space()` -- while others result in just ignoring the operation?  Should all "out of range" access just abort?

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+				return;
+
+			TChar *p = buffer + index;
```
jonpryor, line 556: I haven't traced through the semantics, but can't `this->idx` differ from `buffer.size()`?  Wouldn't it be possible for `this->idx` to be *before* `buffer+index`, meaning we're attempting to write *after* the current `NUL`?  Is that desirable?

Sure, this class has a comment saying "it does no bounds checking; child classes should do that", but *this* method is (1) not `virtual`, and (2) `public`, meaning it is directly usable regardless of class.

Rephrased: does allowing this make any sense?

```cpp
static_local_string<10> s;
for (int i = 0; i < 10; ++i) s.set_at (i, '0' + i);
s.set_index(0);
s.set_at (5, 'a');
```

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,731 @@
...
+				return;
+			}
+
+			std::memcpy (new_allocated_store, base::get_local_store ().data (), MaxStackSize);
+		}
+	};
+
+	// This class mostly performs NO BOUNDS CHECKING - child classes MUST implement that!
+	template<size_t MaxStackSize, typename TStorage, typename TChar = char>
+	class string_base
```
jonpryor, line 342: High-level, could we just have this be a closer mirror of `std::basic_string`?

src/monodroid/jni/strings.hh
grendello: The idea was to be flexible and crash out only when absolutely necessary but I'm not sure I like the idea anymore, so I'll probably make it crash every time.

src/monodroid/jni/strings.hh
grendello: It makes sense, hence the presence of `length()` and `index()` methods you mentioned earlier. However, the question is - do we *need* the functionality in your code excerpt? My initial idea here was to implement only what we actually use and add functionality when need arises

src/monodroid/jni/strings.hh
grendello: Without the `const` overload the `*v.get() = 'a'` code wouldn't build:

```cpp
const local_storage<20, false> _test (10);
*_test.get () = 'a';
```

would give

```
android-system.cc:222:16: error: read-only variable is not assignable
        *_test.get () = 'a';
        ~~~~~~~~~~~~~ ^
```

I will, however, add the `const` overload as it will come useful elsewhere (and will remove the need for the `const_cast`)

src/monodroid/jni/strings.hh
grendello: The only public function that sets the index/length (after the recent changes) is `set_length` and I'm putting a guard in it. `idx` is otherwise set properly everywhere else

src/monodroid/jni/strings.hh
grendello: I think yes. It's an operation on the string, thus it belongs in the string type.

src/monodroid/jni/strings.hh
grendello: What do you mean? We have `assign`, `append`, `clear`, `starts_with` just like in `std::string`  with similar semantics, what else would you like to change to make it closer to `std::string`?

src/monodroid/jni/strings.hh
jonpryor: I'm not at all sure why we'd *want* the fragment `const local_storage<…> _test(10); *_test.get() = 'a'` to build.  I don't understand why this would be desirable.

src/monodroid/jni/android-system.cc
```diff
@@ -207,7 +198,7 @@ AndroidSystem::_monodroid__system_property_get (const char *name, char *sp_value
 		return -1;
 
 	char *buf = nullptr;
-	if (sp_value_len < PROP_VALUE_MAX + 1) {
+	if (sp_value_len < PROPERTY_VALUE_BUFFER_LEN) {
 		size_t alloc_size = ADD_WITH_OVERFLOW_CHECK (size_t, PROP_VALUE_MAX, 1);
```
jonpryor, line 202: This should be updated to likewise use `PROPERTY_VALUE_BUFFER_LEN`.

src/monodroid/jni/basic-utilities.hh
```diff
@@ -186,69 +178,34 @@ namespace xamarin::android
...
-		//   https://github.com/android-ndk/ndk/issues/295
-		//   https://bugs.llvm.org/show_bug.cgi?id=16404
-		//
-		// Using templated parameter types for `a` and `b` would make clang generate that tries to
-		// use 128-bit integers and thus output code that calls `__muloti4` and so linking would
-		// fail
-		//
-		template<typename Ret>
-		inline Ret multiply_with_overflow_check (const char *file, uint32_t line, size_t a, size_t b) const
+		template<typename StringType = char, typename ...Strings>
```
jonpryor, line 187: Shouldn't this be `CharType`, so the parameter is `const CharType* s1`?

src/monodroid/jni/basic-utilities.hh
```diff
@@ -154,13 +148,11 @@ namespace xamarin::android
 			return strdup_new (s, strlen (s));
 		}
 
-		// Without <type_traits> it's a little bit open for abuse (bad stuff will happen if
-		// a type different than `char*` is used to specialize the function and we can't
-		// assert this condition on compile time), but we can take that risk since it's
-		// internal, controlled API
-		template<typename StringType = const char*, typename ...Strings>
-		char* string_concat (const char *s1, StringType s2, Strings... strings)
+		template<typename StringType = char, typename ...Strings>
```
jonpryor, line 151: I do wonder if we're "prematurely designing" here, introducing template `StringType` parameter types (`CharType` parameter types? see other comment), when the *only* type this will *currently* be is `char`.  Why do we need the complexity of generic type parameters *now*?

src/monodroid/jni/basic-utilities.hh
```diff
@@ -186,69 +178,34 @@ namespace xamarin::android
...
+			assert_char_type<StringType> ();
 
-			if (XA_UNLIKELY (__builtin_mul_overflow (a, b, &ret))) {
-				log_fatal (LOG_DEFAULT, "Integer overflow on multiplication at %s:%u", file, line);
-				exit (FATAL_EXIT_OUT_OF_MEMORY);
-				return static_cast<Ret>(0);
-			}
-
-			return ret;
+			return strlen (s1) + calculate_length (strings...);
```
jonpryor, line 192: Should this also have a compiler assert to prevent overflow here?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1180,11 +1155,19 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err, [[m
 	else
 		basename_part = (char*)name;
 
-	simple_pointer_guard<char[]> basename = utils.string_concat ("libaot-", basename_part);
-	h = androidSystem.load_dso_from_any_directories (basename, dl_flags);
+	constexpr char LIBAOT_PREFIX[] = "libaot-";
+	constexpr size_t LIBAOT_PREFIX_SIZE = sizeof(LIBAOT_PREFIX);
+
+	size_t base_len = strlen (basename_part);
+	size_t len = base_len + LIBAOT_PREFIX_SIZE; // includes the trailing \0
```
jonpryor, line 1162: This this also use the overflow-checking macro?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1219,145 +1202,176 @@ MonodroidRuntime::set_environment_variable_for_directory (const char *name, jstr
...
-	delete[] value;
+	mono_jit_set_trace_options (value.get ());
 }
 
 inline void
 MonodroidRuntime::set_profile_options ()
 {
-	constexpr const char mlpd_ext[] = "mlpd";
-	constexpr const char aot_ext[] = "aotprofile";
+	log_warn (LOG_DEFAULT, __PRETTY_FUNCTION__);
```
jonpryor, line 1287: I assume we shouldn't preserve this statement, in which case I would suggest using `log_warn(LOG_DEFAULT, "#grendel"`-prefixed messages, so that it's easier to "see" the log messages which should be *removed* before merging.

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1219,145 +1202,176 @@ MonodroidRuntime::set_environment_variable_for_directory (const char *name, jstr
...
+	// path without allocation
+	dynamic_local_string<SENSIBLE_PATH_MAX + PROPERTY_VALUE_BUFFER_LEN> value;
+	{
+		dynamic_local_string<PROPERTY_VALUE_BUFFER_LEN> prop_value;
+		if (androidSystem.monodroid_get_system_property (Debug::DEBUG_MONO_PROFILE_PROPERTY, prop_value) == 0)
+			return;
 
-	constexpr const char output_arg[] = "output=";
-	constexpr const size_t output_arg_len = sizeof(output_arg) - 1;
+		value.assign (prop_value.get (), prop_value.length ());
```
jonpryor, line 1296: Seeing this statement makes me wonder if there should instead be an `operator=()` or a "move" operator for `value = std::move(prop_value)`;

src/monodroid/jni/basic-utilities.hh
grendello: I want to use `char8_t` for UTF-8 strings once NDK updates to a version of clang that supports it. At the same time, we will STILL need to be able to use `char` for the string pointers we pass to Mono (for simplicity, as `char8_t` cannot be just simply cast to `char`)

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,772 @@
...
+					return static_cast<ssize_t>(start_index);
+				}
+				start_index++;
+			}
+
+			return -1;
+		}
+
+		template<typename T>
+		force_inline bool to_integer (T &val, size_t start_index = 0, int base = 10) const noexcept
```
jonpryor, line 129: Why do we want to inline a member function as large as this?

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,772 @@
...
+	protected:
+		size_t       _last_index = 0;
+		bool         _fresh = true;
+		const char  *_start = nullptr;
+		size_t       _length = 0;
+
+		template<size_t MaxStackSize, typename TStorage, typename TChar> friend class string_base;
+	};
+
+	template<size_t MaxStackSize, bool HasResize, typename T = char>
```
jonpryor, line 205: Ditto earlier comment about "premature generalization": why have `T` when we only support `char`s?

src/monodroid/jni/basic-utilities.hh
grendello: you mean `ADD_WITH_OVERFLOW_CHECK`? If yes, then I'd rather avoid it as this code could generate a number of calls to the built-in thus killing the performance of this function. The likelihood that it will ever overlfow is very, very, very low considering we don't merge user-provided strings with this function. It's used to merge path names with literal strings, mostly. Let's not overengineer... :)

src/monodroid/jni/monodroid-glue.cc
grendello: The likelihood of a file path + some static string length exceeding `18446744073709551615` is quite low... :)

src/monodroid/jni/monodroid-glue.cc
grendello: Not sure if move semantics makes sense since the local buffers are by definition ephemeral so there's no need to add an overhead of implementing proper move semantics (the standard requires that the object moved from is left in some valid state but that the data is gone - that means extra assignments and checks while the object is going away with the scope end anyway). The `=` operator can't take the length parameter and it would become merely a wrapper for `assign(const char*)` - syntactic sugar, not much more. I can add it if you want, but it wouldn't be used here anyway since the `strlen` it would incur is simply not necessary given that `prop_value.length()` does the job without scanning the string for the `NUL` terminator.

src/monodroid/jni/strings.hh
grendello: It's not big at all. There's a lot of fluff in the code, syntactic niceties. And when this code is compiled, it is reduced to basically a single `strto*ll` call + some error checking, the same as if you wrote it by hand. The entire function takes 124 bytes of `aarch64` assembly:

```
 5c4:   350007e9        cbnz    w9, 6c0 <xamarin::android::internal::MonodroidRuntime::parse_runtime_args(xamarin::android::internal::dynamic_local_string<93ul, char>&, xamarin::android::internal::MonodroidRuntime::RuntimeOptions*)+0x6c0>
 5c8:   f1002518        subs    x24, x8, #0x9
 5cc:   540007a3        b.cc    6c0 <xamarin::android::internal::MonodroidRuntime::parse_runtime_args(xamarin::android::internal::dynamic_local_string<93ul, char>&, xamarin::android::internal::MonodroidRuntime::RuntimeOptions*)+0x6c0>  // b.lo, b.ul, b.last
 5d0:   f85383a9        ldur    x9, [x29, #-200]
 5d4:   b4000769        cbz     x9, 6c0 <xamarin::android::internal::MonodroidRuntime::parse_runtime_args(xamarin::android::internal::dynamic_local_string<93ul, char>&, xamarin::android::internal::MonodroidRuntime::RuntimeOptions*)+0x6c0>
                        char s[slen + 1];
 5d8:   91001d08        add     x8, x8, #0x7
 5dc:   910003ea        mov     x10, sp
 5e0:   927ced08        and     x8, x8, #0xfffffffffffffff0
 5e4:   cb080157        sub     x23, x10, x8
 5e8:   910002ff        mov     sp, x23
                        memcpy (s, start () + start_index, slen);
 5ec:   91002521        add     x1, x9, #0x9
 5f0:   aa1703e0        mov     x0, x23
 5f4:   aa1803e2        mov     x2, x24
 5f8:   94000000        bl      0 <memcpy>
                        s[slen] = '\0';
 5fc:   38386aff        strb    wzr, [x23, x24]
                        errno = 0;
 600:   94000000        bl      0 <__errno>
 604:   aa0003f8        mov     x24, x0
 608:   b900001f        str     wzr, [x0]
                                ret = strtoll (s, &endp, base);
 60c:   d102e3a1        sub     x1, x29, #0xb8
 610:   52800142        mov     w2, #0xa                        // #10
 614:   aa1703e0        mov     x0, x23
 618:   94000000        bl      0 <strtoll>
```

src/monodroid/jni/strings.hh
grendello: The explanation is the same :)

src/monodroid/jni/helpers.hh
```diff
@@ -0,0 +1,57 @@
...
+namespace xamarin::android
+{
+#define ADD_WITH_OVERFLOW_CHECK(__ret_type__, __a__, __b__) xamarin::android::Helpers::add_with_overflow_check<__ret_type__>(__FILE__, __LINE__, (__a__), (__b__))
+#define MULTIPLY_WITH_OVERFLOW_CHECK(__ret_type__, __a__, __b__) xamarin::android::Helpers::multiply_with_overflow_check<__ret_type__>(__FILE__, __LINE__, (__a__), (__b__))
+
+	class Helpers
+	{
+	public:
+		template<typename Ret, typename P1, typename P2>
+		static force_inline Ret add_with_overflow_check (const char *file, uint32_t line, P1 a, P2 b) noexcept
```
jonpryor, line 18: Why a static member function instead of a global function?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -860,57 +886,6 @@ MonodroidRuntime::LocalRefsAreIndirect (JNIEnv *env, jclass runtimeClass, int ve
 	return 1;
 }
 
-int
```
jonpryor, line 863: Any reason to remove this member function in *this* PR and not a separate one?

src/monodroid/jni/helpers.hh
grendello: I like to keep as much as possible in the classes, that's the only reason.

src/monodroid/jni/monodroid-glue.cc
grendello: I don't think it warrants a full PR to remove a single function... Such a PR would bring nothing but a waste of time IMO

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -298,29 +300,42 @@ MonodroidRuntime::open_from_update_dir (MonoAssemblyName *aname, [[maybe_unused]
...
 
 	for (uint32_t oi = 0; oi < AndroidSystem::MAX_OVERRIDES; ++oi) {
 		override_dir = androidSystem.get_override_dir (oi);
 		if (override_dir == nullptr || !utils.directory_exists (override_dir))
 			continue;
 
-		const char *format = utils.ends_with (name, ".dll") ? format_none : format_dll;
-		simple_pointer_guard<char, false> fullpath (utils.monodroid_strdup_printf (format, override_dir, pname.get ()));
+		size_t override_dir_len = strlen (override_dir);
+		static_local_string<SENSIBLE_PATH_MAX> fullpath (override_dir_len + file_name_len);
```
jonpryor, line 330: Do we need to include the count for the terminating null here?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -506,104 +524,106 @@ MonodroidRuntime::parse_gdb_options ()
...
-						if ((endp == arg) || (*endp != '\0')) {
-							log_error (LOG_DEFAULT, "Invalid --debug argument.");
-							continue;
-						}
-					} else if (*endp != '\0') {
-						log_error (LOG_DEFAULT, "Invalid --debug argument.");
-						continue;
+			options->debug = true;
+
+			if (token.has_at ('=', 5)) {
```
jonpryor, line 553: To reduce "manual counting", shouldn't the `5` become `ARG_DEBUG_LENGTH+1`?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -506,104 +524,106 @@ MonodroidRuntime::parse_gdb_options ()
...
+				static_local_string<SMALL_STRING_PARSE_BUFFER_LEN> hostport (token.length () - 6);
```
jonpryor, line 554: similarly, where is `6` coming from? It's from `ARG_DEBUG_LENGTH+2`, but that isn't really obvious, meaning *more* manual counting.

src/monodroid/jni/monodroid-glue.cc
grendello: This is to account for the path separator

src/monodroid/jni/monodroid-glue.cc
grendello: if we were to use a constant, it would need to be `ARG_DEBUG_LENGTH` since the 2nd argument to `has_at` is 0-based

src/monodroid/jni/monodroid-glue.cc
grendello: It's from `ARG_DEBUG_LENGTH + 1` and includes the `=`

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1288,145 +1271,175 @@ MonodroidRuntime::set_environment_variable_for_directory (const char *name, jstr
...
-
-	if (androidSystem.monodroid_get_system_property (Debug::DEBUG_MONO_ENV_PROPERTY, &value) == 0)
+	dynamic_local_string<PROPERTY_VALUE_BUFFER_LEN> value;
+	if (androidSystem.monodroid_get_system_property (Debug::DEBUG_MONO_ENV_PROPERTY, value) == 0)
 		return;
 
-	char **args = utils.monodroid_strsplit (value, "|", 0);
-	delete[] value;
+	auto log_envvar = [](const char *name, const char *v) {
+		log_info (LOG_DEFAULT, "Env variable '%s' set to '%s'.", name, v);
```
jonpryor, line 1311: I don't think we previously logged this info; any reason to begin doing so?

grendello: It should use `log_debug` and the reason is to double-check that parsing is correct, if need be.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,772 @@
...
+		}
+
+		force_inline bool empty () const noexcept
+		{
+			return length () == 0;
+		}
+
+		force_inline bool equal (const char *s) const noexcept
+		{
+			if (s == nullptr)
```
jonpryor, line 53: Is it possible for `_start` to ever be `nullptr`? If so, how should such an instance compare to `segment.equal(nullptr)`?

grendello: `_start` is `nullptr` initially and if `next_token` returns `false` on the **first** invocation. If `next_token` returns `false` then `token` should be effectively ignored - so comparison against `nullptr` passed to `equal` should always be `false` for a **valid** `string_segment` (i.e. when `next_token` returned `true`).  The `equal(nullptr, 0)` case is a bug.

src/monodroid/jni/strings.hh
```diff
@@ -0,0 +1,775 @@
...
+				std::memcpy (new_allocated_store, old_allocated_store, old_size);
+				delete[] old_allocated_store;
+				return;
+			}
+
+			std::memcpy (new_allocated_store, base::get_local_store ().data (), MaxStackSize);
+		}
+	};
+
+	template<size_t MaxStackSize, typename TStorage, typename TChar = char>
```
jonpryor, line 355: Is there a naming convention for type parameters?  Some places you have a `Type` suffix, e.g. `CharType`, and other places you have a C#-style `T` prefix, e.g. `TChar`.  Should these be consistent?

grendello: Undecided :)

Issue #4990, opened 2020-08-08
Mono.AndroidTools.InstallFailedException: Failure [INSTALL_FAILED_INVALID_APK: Failed to extract native libraries, res=-2]


alexshikov:
After update to latest Visual Studio for Mac the Android app is not deployed to the device anymore.
The previous version worked well.

### Steps to Reproduce

1. Try to run the app on the Emulator

### Expected Behavior

The app runs

### Actual Behavior

Deployment failed with error:
```
    /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.Debugging.targets(618,2): error ADB0010:  Deployment failed
    /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.Debugging.targets(618,2): error ADB0010: Mono.AndroidTools.InstallFailedException: Failure [INSTALL_FAILED_INVALID_APK: Failed to extract native libraries, res=-2]
    /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.Debugging.targets(618,2): error ADB0010:   at Mono.AndroidTools.Internal.AdbOutputParsing.CheckInstallSuccess (System.String output, System.String packageName) [0x00151] in /Users/builder/azdo/_work/22/s/xamarin-android/external/monodroid/tools/msbuild/external/androidtools/Mono.AndroidTools/Internal/AdbOutputParsing.cs:344 
    /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.Debugging.targets(618,2): error ADB0010:   at Mono.AndroidTools.AndroidDevice+<>c__DisplayClass97_0.<InstallPackage>b__0 (System.Threading.Tasks.Task`1[TResult] t) [0x00016] in /Users/builder/azdo/_work/22/s/xamarin-android/external/monodroid/tools/msbuild/external/androidtools/Mono.AndroidTools/AndroidDevice.cs:784 
    /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.Debugging.targets(618,2): error ADB0010:   at System.Threading.Tasks.ContinuationTaskFromResultTask`1[TAntecedentResult].InnerInvoke () [0x00024] in <b62523977edc47f990e8bfd12c39abb8>:0 
[...]

PR #5001, opened 2020-08-12
[Xamarin.Android.Build.Tasks] <BuildApk/> should check compression

jonathanpeppers:
Fixes: https://feedback.devdiv.io/1144021
Fixes: https://github.com/xamarin/xamarin-android/issues/4990

If you deploy an app with `android:extractNativeLibs="true"` (most of
the time the default when omitted).

Then change it to `android:extractNativeLibs="false"` and deploy again
you get a deploy error:

    ADB0010: Mono.AndroidTools.InstallFailedException: Failure [INSTALL_FAILED_INVALID_APK: Failed to extract native libraries, res=-2]

Looking at the resulting `.apk` file, the native libraries are
compressed when they shouldn't be. A `Rebuild` solves the issue.

In 1c4727cb, the changes to the `<BuildApk/>` for better incremental
build performance did not take into account what happens if the
compression mode changes for zip entries.

There are two cases:

1. We are copying entries produced by `aapt` from `packaged_resources`
   to the `.apk` file. This needs to check if the `CompressedSize`
   matches or not.
2. We are adding entries from disk. This needs to check the if
   `CompressionMode` matches in the existing entry.

The only hangup is usage of `CompressionMode.Default`, as it generates
entries with `CompressionMode.Deflate`. We can check against
`CompressionMode.Store` in this case.

I added tests around these scenarios.

### DIFF
--- a/Documentation/release-notes/5001.md
+++ b/Documentation/release-notes/5001.md
@@ -1,0 +1,8 @@
+#### Application and library build and deployment
+
+* [Developer Community 1144021][0] and
+  [GitHub Issue 4990](https://github.com/xamarin/xamarin-android/issues/4990):
+  *ADB0010: [INSTALL_FAILED_INVALID_APK: Failed to extract native libraries, res=-2]*
+  build error could prevent apps from being deployed after `AndroidManifest.xml` changes.
+
+[0]: https://developercommunity.visualstudio.com/content/problem/1144021/vs-1670-compile-and-archive-issues.html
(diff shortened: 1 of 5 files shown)


jonathanpeppers:
There is one test failure on macOS:
```
SkipExistingFile returned unexpected value for /Users/runner/work/1/s/bin/TestRelease/temp/SkipExistingFile_Compression.zip /Users/runner/work/1/s/bin/TestRelease/temp/SkipExistingFile_Compression/A.txt A.txt
8/12/2020 6:28:43 PM (disk 08/12/2020 18:28:43:000) <= 8/12/2020 6:28:43 PM (zip 08/12/2020 18:28:43:000) = False
Expected: True
But was:  False
```

jonathanpeppers:
The `Aapt2DaemonInstances` tests should hopefully be fixed in: https://github.com/xamarin/xamarin-android/pull/5006

PR #5006, opened 2020-08-13
[tests] fix Aapt2DaemonInstances on 4 core machines

jonathanpeppers:
Context: https://build.azdo.io/3974634

Several of the `Aapt2DaemonInstances` tests can fail with:

    Expected 6 but was 3
    Expected: 6
    But was:  3

It seems like this is happening randomly, and we think the reason is
that some macOS CI machines only have 4 processors/cores.

To solve this:

* I moved the `Environment.ProcessorCount` call to a `virtual`
  property.
* The test overrides the value to 8, so the test will pass on any
  machine.

I verified that if you change the value to 4, the test fails in the
same way we are seeing on CI.


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/Aapt2Tests.cs
```diff
@@ -111,6 +111,13 @@ public void Aapt2DaemonInstances (int maxInstances, int expectedMax, int expecte
 			Directory.Delete (Path.Combine (Root, path), recursive: true);
 		}
 
+		class Aapt2CompileWithProcessorCount : Aapt2Compile
+		{
+			protected override int ProcessorCount { get; }
+
+			public Aapt2CompileWithProcessorCount (int processorCount) => ProcessorCount = processorCount;
```
dellis1972, line 118: Hmm interesting way to solve this issue :)

PR #5007, opened 2020-08-13
[Microsoft.Android.Sdk] use net5.0-android $(TargetFramework)

jonathanpeppers:
After bumping to .NET 5.0.100-rc.2.20458.11, we can now use:

    <TargetFramework>net5.0-android</TargetFramework>

On the first try, this failed with:

Our SDK needs to specify:

    <TargetPlatformSupported Condition=" '$(TargetPlatformIdentifier)' == 'Android' ">true</TargetPlatformSupported>

After doing this, everything seemed to work as before.

I updated tests so they will use `net5.0-android` going forward.

## @(SdkSupportedTargetPlatform) ##

A second part of this change is the need to fill out
`$(TargetPlatformVersion)`, `@(AndroidSdkSupportedTargetPlatform)`,
and `@(SdkSupportedTargetPlatform)` similar to:

https://github.com/dotnet/sdk/blob/c43785736edc8d4bc927df3b557c1609d028cd93/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.WindowsSdkSupportedTargetPlatforms.props

Using this data, the dotnet/sdk will run the
`GenerateTargetPlatformDefineConstants` MSBuild target:

https://github.com/dotnet/sdk/blob/a0e17b2b5d2ec75c16e1083b1f926a258126c73a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets#L188-L198

This fills out `$(DefineConstants)` with `ANDROID;ANDROID21;...;ANDROID30`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,8 +19,7 @@ logcat-*.txt
 apk-sizes-*.txt
 *.rawproto
 *.binlog
-src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props
-src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BundledVersions.props
+*.ProjectImports.zip
 *~
 external/monodroid/
 external/mono/
--- a/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
+++ b/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
@@ -1,0 +1,83 @@
+using System.IO;
+using System.Linq;
+using System.Xml;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+
+namespace Xamarin.Android.Tools.BootstrapTasks
+{
+	/// <summary>
+	/// Generates Microsoft.Android.Sdk.SupportedPlatforms.props
+	/// Similar to: https://github.com/dotnet/sdk/blob/18ee4eac8b3abe6d554d2e0c39d8952da0f23ce5/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.WindowsSupportedTargetPlatforms.props
+	/// </summary>
+	public class GenerateSupportedPlatforms : Task
+	{
+		/// <summary>
+		/// A list of AndroidApiInfo.xml files produced by Mono.Android.targets
+		/// </summary>
+		[Required]
+		public string [] AndroidApiInfo { get; set; }
+
+		/// <summary>
+		/// The output file to generate
+		/// </summary>
+		[Required]
+		public string OutputFile { get; set; }
+
+		/// <summary>
+		/// $(AndroidMinimumDotNetApiLevel) from Configuration.props
+		/// </summary>
+		[Required]
+		public int MinimumApiLevel { get; set; }
+
+		public override bool Execute ()
+		{
+			if (AndroidApiInfo.Length == 0) {
+				Log.LogError ("This task requires at least one AndroidApiInfo.xml file!");
+				return false;
+			}
+
+			var versions = new AndroidVersions (
+				AndroidApiInfo.Select (d => Path.GetDirectoryName (d)));
+			var settings = new XmlWriterSettings {
+				OmitXmlDeclaration = true,
+				Indent = true,
+			};
+			using (var writer = XmlWriter.Create (OutputFile, settings)) {
+
+				writer.WriteComment ($@"
+***********************************************************************************************
+{Path.GetFileName (OutputFile)}
+
+Specifies the supported Android platform versions for this SDK.
+
+***********************************************************************************************
+");
+				writer.WriteStartElement ("Project");
+
+				writer.WriteStartElement ("PropertyGroup");
+				writer.WriteStartElement ("TargetPlatformVersion");
+				writer.WriteAttributeString ("Condition", " '$(TargetPlatformVersion)' == '' ");
+				writer.WriteString (versions.MaxStableVersion.ApiLevel.ToString ());
+				writer.WriteEndElement (); // </TargetPlatformVersion>
+				writer.WriteEndElement (); // </PropertyGroup>
+
+				writer.WriteStartElement ("ItemGroup");
+				foreach (AndroidVersion version in versions.InstalledBindingVersions
+						.Where (v => v.ApiLevel >= MinimumApiLevel)
+						.OrderBy (v => v.ApiLevel)) {
+					writer.WriteStartElement ("AndroidSdkSupportedTargetPlatform");
+					writer.WriteAttributeString ("Include", version.ApiLevel.ToString ());
+					writer.WriteEndElement (); // </AndroidSdkSupportedTargetPlatform>
+				}
+				writer.WriteStartElement ("SdkSupportedTargetPlatform");
+				writer.WriteAttributeString ("Condition", " '$(TargetPlatformIdentifier)' == 'Android' ");
+				writer.WriteAttributeString ("Include", "@(AndroidSdkSupportedTargetPlatform)");
+
+				writer.WriteEndDocument (); // </Project>
+			}
+
+			return !Log.HasLoggedErrors;
+		}
+	}
+}
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
@@ -19,9 +19,13 @@
   </PropertyGroup>
 
   <Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
+  <PropertyGroup>
+    <TargetPlatformSupported Condition=" '$(TargetPlatformIdentifier)' == 'Android' ">true</TargetPlatformSupported>
+  </PropertyGroup>
 
   <!-- Default item includes (globs and implicit references) -->
   <Import Project="Microsoft.Android.Sdk.DefaultItems.targets" />
+  <Import Project="Microsoft.Android.Sdk.SupportedPlatforms.props" />
   <!-- Build ordering, should be imported before Xamarin.Android.Common.targets -->
   <Import Project="$(MSBuildThisFileDirectory)Microsoft.Android.Sdk.BuildOrder.targets" />
 
--- a/Configuration.props
+++ b/Configuration.props
@@ -20,6 +20,8 @@
     <!-- Should correspond to the first value from `$(API_LEVELS)` in `build-tools/api-xml-adjuster/Makefile` -->
     <AndroidFirstFrameworkVersion Condition="'$(AndroidFirstFrameworkVersion)' == ''">v4.4</AndroidFirstFrameworkVersion>
     <AndroidFirstApiLevel Condition="'$(AndroidFirstApiLevel)' == ''">19</AndroidFirstApiLevel>
+    <!-- The min API level supported by Microsoft.Android.Sdk, should refactor/remove when this value is the same as $(AndroidFirstApiLevel) -->
+    <AndroidMinimumDotNetApiLevel Condition="'$(AndroidMinimumDotNetApiLevel)' == ''">21</AndroidMinimumDotNetApiLevel>
     <AndroidFirstPlatformId Condition="'$(AndroidFirstPlatformId)' == ''">$(AndroidFirstApiLevel)</AndroidFirstPlatformId>
     <_IsRunningNuGetRestore Condition="$(RestoreTaskAssemblyFile.EndsWith('NuGet.exe', StringComparison.InvariantCultureIgnoreCase))">True</_IsRunningNuGetRestore>
     <!-- *Latest* *stable* API level binding that we support; used when building src/Xamarin.Android.Build.Tasks -->
--- a/Documentation/guides/OneDotNet.md
+++ b/Documentation/guides/OneDotNet.md
@@ -206,7 +206,7 @@ The following instructions can be used for early preview testing.
 ```xml
 <Project Sdk="Microsoft.Android.Sdk/10.0.100">
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net5.0-android</TargetFramework>
     <RuntimeIdentifier>android.21-arm64</RuntimeIdentifier>
     <OutputType>Exe</OutputType>
   </PropertyGroup>
--- a/build-tools/automation/azure-pipelines-oss.yaml
+++ b/build-tools/automation/azure-pipelines-oss.yaml
@@ -27,7 +27,7 @@ variables:
   PREPARE_FLAGS: PREPARE_CI=1 PREPARE_CI_PR=1
   DotNetCoreVersion: 3.1.201
   # Version number from: https://github.com/dotnet/installer#installers-and-binaries
-  DotNetCorePreviewVersion: 5.0.100-preview.7.20307.3
+  DotNetCorePreviewVersion: 5.0.100-rc.2.20459.1
 
 stages:
 - stage: mac_stage
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -52,7 +52,7 @@ variables:
   NUnitConsoleVersion: 3.11.1
   DotNetCoreVersion: 3.1.201
   # Version number from: https://github.com/dotnet/installer#installers-and-binaries
-  DotNetCorePreviewVersion: 5.0.100-preview.7.20307.3
+  DotNetCorePreviewVersion: 5.0.100-rc.2.20459.1
   HostedMacMojave: Hosted Mac Internal Mojave
   HostedMac: Hosted Mac Internal
   HostedWinVS2019: Hosted Windows 2019 with VS2019
--- a/src/Xamarin.Android.Build.Tasks/.gitignore
+++ b/src/Xamarin.Android.Build.Tasks/.gitignore
@@ -1,2 +1,4 @@
+/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BundledVersions.props
+/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.SupportedPlatforms.props
 /Xamarin.Android.Common.props
 /Xamarin.Android.BuildInfo.txt
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XASdkProject.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XASdkProject.cs
@@ -73,7 +73,7 @@ $@"{{
 		public XASdkProject (string outputType = "Exe")
 		{
 			Sdk = $"Microsoft.Android.Sdk/{SdkVersion}";
-			TargetFramework = "net5.0";
+			TargetFramework = "net5.0-android";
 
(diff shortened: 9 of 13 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
```diff
@@ -43,6 +43,9 @@
     <!-- Conflicts with our targets generator in VS+CPS: See https://work.azdo.io/1112733 -->
     <ProjectCapability Remove="LaunchProfiles" />
   </ItemGroup>
+  <PropertyGroup>
+    <TargetPlatformSupported Condition=" '$(TargetPlatformIdentifier)' == 'Android' ">true</TargetPlatformSupported>
+  </PropertyGroup>
```
jonathanpeppers, line 48: This is setting `$(TargetPlatformSupported)` shortly after importing:
```xml
<Import Sdk="Microsoft.NET.Sdk" Project="Sdk.targets" />
```

jonathanpeppers:
/cc @dsplaisted @sfoslund

I think there is something else needed here: listing supported target platform versions?

dsplaisted:
> /cc @dsplaisted @sfoslund
> 
> I think there is something else needed here: listing supported target platform versions?

Yes, include `SupportedTargetPlatform` items like we do for Windows here: https://github.com/dotnet/sdk/blob/master/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.WindowsSupportedTargetPlatforms.props

dsplaisted:
Also, the OS version for iOS and Android should generally be specified in the `TargetFramework`.  So `net5.0-android21.0` instead of `net5.0-android`.  (`net5.0-android21` should also eventually be OK if that's preferrable, but I'm not sure whether the NuGet fix is in to understand this as version 21 instead of 2.1)

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
```diff
@@ -0,0 +1,70 @@
...
+				foreach (AndroidVersion version in versions.InstalledBindingVersions) {
+					writer.WriteStartElement ("AndroidSupportedTargetPlatform");
+					writer.WriteAttributeString ("Include", version.ApiLevel.ToString ());
+					writer.WriteEndElement (); // </AndroidSupportedTargetPlatform>
+				}
+				writer.WriteStartElement ("SupportedTargetPlatform");
+				writer.WriteAttributeString ("Condition", " '$(TargetPlatformIdentifier)' == 'Android' ");
+				writer.WriteAttributeString ("Include", "@(AndroidSupportedTargetPlatform)");
+
+				writer.WriteEndDocument (); // </Project>
```
jonathanpeppers, line 64: Once we generate these item groups, we get:

https://github.com/dotnet/sdk/blob/a0e17b2b5d2ec75c16e1083b1f926a258126c73a/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets#L188-L198

This fills out `$(DefineConstants)` with `ANDROID;ANDROID21;...;ANDROID30`.

@dellis1972 @jonpryor do we need to get rid of the old `__ANDROID__` defines? The .NET 5 design spec doesn't list these:

https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md#preprocessor-symbols

So should we stop running this task?

https://github.com/xamarin/xamarin-android/blob/e61ff790ca524ac173be3965fbbad510d8303b43/src/Xamarin.Android.Build.Tasks/Tasks/GetAndroidDefineConstants.cs#L34-L38

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
```diff
@@ -0,0 +1,70 @@
...
+				writer.WriteStartElement ("TargetPlatformVersion");
+				writer.WriteAttributeString ("Condition", " '$(TargetPlatformVersion)' == '' ");
+				writer.WriteString (versions.MaxStableVersion.ApiLevel.ToString ());
+				writer.WriteEndElement (); // </TargetPlatformVersion>
+				writer.WriteEndElement (); // </PropertyGroup>
+
+				writer.WriteStartElement ("ItemGroup");
+				foreach (AndroidVersion version in versions.InstalledBindingVersions) {
+					writer.WriteStartElement ("AndroidSupportedTargetPlatform");
+					writer.WriteAttributeString ("Include", version.ApiLevel.ToString ());
```
jonathanpeppers, line 57: This is supposed to be the API level and not the Android OS version?

Would list 29, 30 and not 10.0, 11.0?

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
dsplaisted: It would make it easier to port code if you left it in, but then there would be a bunch of duplicate constants.

Redth:
Is there no way to make 21.0 the default if not specified at all? I suppose it's not a huge deal as future versions you'll mostly want to specify it anyway.

dsplaisted:
> Is there no way to make 21.0 the default if not specified at all? I suppose it's not a huge deal as future versions you'll mostly want to specify it anyway.

Per [the design](https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md#os-versions), the plan is to support a TargetFramework without an OS version, but not to use that in the templates as the canonical version.  So Android should add logic (similar to Windows) that sets the TargetPlatformVersion if it's not set.  But the templates, samples, etc. should be explicit about the version.

(Note that the "canonical" version is different for Windows, as Windows Forms and WPF developers often will not care about specifying a specific version of Windows, so for those templates the `TargetFramework` will be `net5.0-windows` instead of `net5.0-windows7.0` or `net5.0-windows10.0.xyz`.)

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
dellis1972: So we have a few options. 

1. a hard break , users need to switch to the new defines for net5. (not desirable I think).
2. define the old `__` prefixed ones in net5..
3. define the new non prefixed ones in our product now as well as the old, but issue warnings the old are deprecated (if possible) to allow users to migrate. We can then phase our the old style.

Personally I would prefer option 3.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
Redth: Is this perhaps an opportunity to discuss where we use API level vs OS version?  Though we need to keep `__ANDROID_29__` and such for existing code...

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
jonathanpeppers: I don't see a way to deprecate usage of `#if __ANDROID__`, etc.?

It would have to be something in Roslyn, because  it would require looking through every C# source file.

I guess we leave both old & new `$(DefineConstants)` for now.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
dsplaisted: I think we were planning to still use the API level, though I don't see that specifically mentioned in the specs (https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md#os-versions, https://github.com/dotnet/designs/blob/master/accepted/2020/minimum-os-version/minimum-os-version.md).

@mhutch, can you comment?

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
dsplaisted: It occurs to me that Roslyn is now shipping a standard set of Analyzers in the .NET SDK, so if we wanted to we could probably add an analyzer there that detected usage of the old Android compilation constants format.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
jonathanpeppers: The current generated file is:
```xml
<Project>
  <PropertyGroup>
    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">30</TargetPlatformVersion>
  </PropertyGroup>
  <ItemGroup>
    <AndroidSupportedTargetPlatform Include="26" />
    <AndroidSupportedTargetPlatform Include="23" />
    <AndroidSupportedTargetPlatform Include="19" />
    <AndroidSupportedTargetPlatform Include="27" />
    <AndroidSupportedTargetPlatform Include="22" />
    <AndroidSupportedTargetPlatform Include="21" />
    <AndroidSupportedTargetPlatform Include="30" />
    <AndroidSupportedTargetPlatform Include="29" />
    <AndroidSupportedTargetPlatform Include="25" />
    <AndroidSupportedTargetPlatform Include="20" />
    <AndroidSupportedTargetPlatform Include="28" />
    <AndroidSupportedTargetPlatform Include="24" />
    <SupportedTargetPlatform Condition=" '$(TargetPlatformIdentifier)' == 'Android' " Include="@(AndroidSupportedTargetPlatform)" />
  </ItemGroup>
</Project>
```

I need to drop 19 & 20, and looks like they need to be sorted, too. 🤦‍♀️

jonathanpeppers: Latest file is:
```xml
<Project>
  <PropertyGroup>
    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">30</TargetPlatformVersion>
  </PropertyGroup>
  <ItemGroup>
    <AndroidSupportedTargetPlatform Include="21" />
    <AndroidSupportedTargetPlatform Include="22" />
    <AndroidSupportedTargetPlatform Include="23" />
    <AndroidSupportedTargetPlatform Include="24" />
    <AndroidSupportedTargetPlatform Include="25" />
    <AndroidSupportedTargetPlatform Include="26" />
    <AndroidSupportedTargetPlatform Include="27" />
    <AndroidSupportedTargetPlatform Include="28" />
    <AndroidSupportedTargetPlatform Include="29" />
    <AndroidSupportedTargetPlatform Include="30" />
    <SupportedTargetPlatform Condition=" '$(TargetPlatformIdentifier)' == 'Android' " Include="@(AndroidSupportedTargetPlatform)" />
  </ItemGroup>
</Project>
```

akoeplinger: @dsplaisted I filed an issue on the dotnet/designs repo a while back about whether we want to use the API level or OS version for android TFMs: https://github.com/dotnet/designs/issues/136.

It doesn't seem like there was a conclusion but I agree this is something that needs to be decided sooner than later.

.gitignore
```diff
@@ -19,8 +19,7 @@ logcat-*.txt
 apk-sizes-*.txt
 *.rawproto
 *.binlog
-src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.props
```
dellis1972, line 22: Was this removal intentional? This file is generated?

jonathanpeppers: There is a `src/Xamarin.Android.Build.Tasks/.gitignore`, where we actually had some things duplicated.

I moved the rules that seemed like they should be moved to `src/Xamarin.Android.Build.Tasks/.gitignore`.

jonathanpeppers:
I started a manual test run that will runs all our "One .NET tests": https://build.azdo.io/3992680

If nothing appears to be broken, I think we can merge.

Let me know if anyone has anything outstanding, or if we need to change any of these:

* We are using Android API levels and not OS versions.
* We are leaving the existing `__ANDROID__` defines in place for now.
* The generated MSBuild file looks OK: https://github.com/xamarin/xamarin-android/pull/5007#discussion_r470878979

jonathanpeppers:
Ok, many of the tests seem to fail when a `net5.0-android` application references a `net5.0-android` library:
```
"/Users/runner/work/1/s/bin/TestRelease/temp/BuildAppWithSystemNamespace/UnnamedProject/UnnamedProject.csproj" (Build;SignAndroidPackage target) (1:7) ->
                   (_GetProjectReferenceTargetFrameworkProperties target) -> 
                     /Users/runner/.dotnet/sdk/5.0.100-rc.1.20414.5/Microsoft.Common.CurrentVersion.targets(1656,5): error : Project '../Library1.System/Library1.System.csproj' targets 'net5.0-android'. It cannot be referenced by a project that targets '.NETCoreApp,Version=v5.0'. [/Users/runner/work/1/s/bin/TestRelease/temp/BuildAppWithSystemNamespace/UnnamedProject/UnnamedProject.csproj]

    0 Warning(s)
    1 Error(s)
```
I think @rolfbjarne was seeing the same thing on iOS.

rolfbjarne:
> I think @rolfbjarne was seeing the same thing on iOS.

Yes, that's the same error I was seeing.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
jonpryor: @jonathanpeppers [asked](https://github.com/xamarin/xamarin-android/pull/5007#discussion_r470265770)

> do we need to get rid of the old __ANDROID__ defines?

***No***, let's not. :-)

dsplaisted:
> Ok, many of the tests seem to fail when a `net5.0-android` application references a `net5.0-android` library:

This should be fixed in preview 8.  I'm not sure of the current state of RC1, but if it is not fixed in the latest builds, we will be working on fixing it as a high priority.

jonathanpeppers:
Thanks @dsplaisted, if they added a Preview 8 column to the downloads here:

https://github.com/dotnet/installer#installers-and-binaries

I think we could go ahead and start using this. (Or we can wait until master/rc1 has it)

jonathanpeppers:
I started a new test run using `.NET 5.0.100-rc.1.20426.3` here: https://build.azdo.io/4012976

@rolfbjarne FYI, I think the project references from a `net5.0-android` app to a `net5.0-android` library are working now. (one worked for me locally)

jonathanpeppers:
Only remaining test failure here is a bit weird:
```
Build FAILED.
    0 Warning(s)
    0 Error(s)
```
Normally I do `msbuild /v:diag msbuild.binlog | grep '-- FAILED'` and I find a message like "task returned false but logged no errors", but I don't see that.

Posted in MSBuild teams channel for now.

dellis1972 (approved):

jonathanpeppers:
The Windows build failure is caused by text in the commit message:
```
Microsoft.NET.TargetFrameworkInference.targets(113,5): error NETSDK1139: The target platform identifier android was not recognized.
```
This should be fixed in https://github.com/xamarin/java.interop/pull/705.

jonpryor:
Commit message?

```markdown
After bumping to .NET 5.0.100-rc.2.20459.1, we can now use:

	<TargetFramework>net5.0-android</TargetFramework>

On the first try, this failed with:

Our SDK needs to specify:

	<TargetPlatformSupported Condition=" '$(TargetPlatformIdentifier)' == 'Android' ">true</TargetPlatformSupported>

After doing this, everything seemed to work as before.

I updated tests so they will use `net5.0-android` going forward.


~~ @(SdkSupportedTargetPlatform) ~~

When building *current* Xamarin.Android projects, a *set* of
`__ANDROID_*__` symbols are `#define`d by the compiler, one
per API level <= the API level of the `$(TargetFrameworkVersion)`;
see e.g. 669d5153.  This allows code to be "guarded" by a compile-
time version check that *doesn't* require updates when the next
API level is supported:

	#if __ANDROID_30__
	  // Use types or methods that only exist in Xamarin.Android v11.0+
	#endif  // __ANDROID_30__

.NET 5+ will have a similar feature: a set of C# `#define`s will be
enabled corresponding to all supported previous versions *as well as*
the current version.  For example, when building a `net5.0` project,
*both* `NETCOREAPP3_1` *and* `NET5_0` will be defined.

To "unify" the experience between current Xamarin.Android and .NET 5,
we want to *also* define a set of `ANDROID*` symbols, *also* one per
supported API level.

The unified experience uses the new
[`GenerateTargetPlatformDefineConstants`][0] and
[`GeneratePlatformCompatibleDefineConstants`][1] MSBuild targets,
which use a new `@(SdkSupportedTargetPlatform)` item group to control
which symbols are defined.
See [`Microsoft.NET.WindowsSdkSupportedTargetPlatforms.props`][2]
for how `net5.0-windows` is supported.

For `net5.0-android` support, the new `<GenerateSupportedPlatforms/>`
build-tools task will *generate* a
`Microsoft.Android.Sdk.SupportedPlatforms.props` file based on the
contents of the `bin/…/Mono.Android/AndroidApiInfo.xml` files:

	<Project>
	  <PropertyGroup>
	    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">30</TargetPlatformVersion>
	  </PropertyGroup>
	  <ItemGroup>
	    <AndroidSupportedTargetPlatform Include="21" />
	    <!-- … -->
	    <AndroidSupportedTargetPlatform Include="30" />
	    <SdkSupportedTargetPlatform
	        Condition=" '$(TargetPlatformIdentifier)' == 'Android' "
	        Include="@(AndroidSupportedTargetPlatform)"
	    />
	  </ItemGroup>
	</Project>

`@(SdkSupportedTargetPlatform)` will cause the
`GeneratePlatformCompatibleDefineConstants` target to define the
symbols `ANDROID21;ANDROID22;…;ANDROID30`.

[0]: https://github.com/dotnet/sdk/blob/6924789288071724fa71d8c564bd6e418d5490b1/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets#L189-L200
[1]: https://github.com/dotnet/sdk/blob/6924789288071724fa71d8c564bd6e418d5490b1/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.BeforeCommon.targets#L214-L221
[2]: https://github.com/dotnet/sdk/blob/c43785736edc8d4bc927df3b557c1609d028cd93/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.WindowsSdkSupportedTargetPlatforms.props
```

Issue #4998, opened 2020-08-11
Cannot create AAB Release bundle with app with native compiled libraries


luminixinc:
Hi there,

We have hit an issue trying to bundle a *Release* build of our Xamarin.Forms/Xamarin.Android app as AAB.

The bundling step fails with the following build output:

```
C:\Program Files\Android\Jdk\microsoft_dist_openjdk_1.8.0.25\bin\java.exe -Xmx1G -jar "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\Android\bundletool.jar" build-bundle --modules obj\Release\90\android\bin\base.zip --output obj\Release\90\android\bin\com.companyname.nativehelloapp.aab --config C:\Users\BuildUser\AppData\Local\Temp\tmpDBCF.tmp 
Files under lib/ must have .so extension, found 'lib/x86/gdbserver'.
com.android.tools.build.bundletool.model.exceptions.BundleFileTypesException$InvalidFileExtensionInDirectoryException: Files under lib/ must have .so extension, found 'lib/x86/gdbserver'.
```

Here is a [reduced testcase](https://github.com/luminixinc/XamarinAndroidNativeDebugTest) that showcases the build issue in detail, and speculates about a potential fix.

jonathanpeppers:
@luminixinc it doesn't seem like `bundletool` allows a file in `lib` without an `.so` extension:

https://github.com/google/bundletool/blob/ae0fc0162fd80d92ef8f4ef4527c066f0106942f/src/main/java/com/android/tools/build/bundletool/validation/BundleFilesValidator.java#L97-L103

So, I tried something like this: https://github.com/xamarin/xamarin-android/compare/master...jonathanpeppers:bundletool-libs-without-so

But it just yields a different error:

```
[BT File 'root/lib/arm64-v8a/gdbserver' uses reserved file or directory name 'lib'.
```

It seems like `bundletool` does not have a way for this to work. Can you put the file in a different place, or name it `gdbserver.so`?

If you see another way to solve this problem, let me know, thanks.

luminixinc:
[...]

PR #5008, opened 2020-08-13
[Microsoft.Cpp.Android] fix for app bundles and C++ projects

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/4998
Context: https://github.com/luminixinc/XamarinAndroidNativeDebugTest

The above C++ project fails to build when using
`$(AndroidPackageFormat)` = `aab` with:

    Files under lib/ must have .so extension, found 'lib/x86/gdbserver'.

Reviewing `Microsoft.Cpp.Android.targets`, it adds `gdbserver` to the
`@(AndroidNativeLibrary)` item group.

`bundletool` fails on any non-`.so` files added to the `lib` directory:

https://github.com/google/bundletool/blob/ae0fc0162fd80d92ef8f4ef4527c066f0106942f/src/main/java/com/android/tools/build/bundletool/validation/BundleFilesValidator.java#L97-L103

This should probably only be done for debug builds. We should only
include this file when `$(AndroidIncludeDebugSymbols)` is `true`
and `$(AndroidPackageFormat)` is not `aab`.

### DIFF
--- a/Documentation/release-notes/cpp-aab.md
+++ b/Documentation/release-notes/cpp-aab.md
@@ -1,0 +1,6 @@
+#### C++ projects
+
+* [GitHub Issue 4998](https://github.com/xamarin/xamarin-android/issues/4998):
+  *Files under lib/ must have .so extension, found 'lib/x86/gdbserver'.* build error
+  prevented building app bundles for projects with references to Android C++ library
+  projects.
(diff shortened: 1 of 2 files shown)


Documentation/release-notes/cpp-aab.md
```diff
@@ -0,0 +1,5 @@
+#### C++ projects
+
+* [GitHub Issue 4998](https://github.com/xamarin/xamarin-android/issues/4998):
+  *Files under lib/ must have .so extension, found 'lib/x86/gdbserver'.* build error
+  could prevent C++ projects from building app bundles.
```
brendanzagaeski, line 5: Looks great on formatting. For wording, if I'm skimming the issue correctly, I think the error happens when attempting to build an app bundle for a Xamarin.Android app project that references a C++ Android library project, so maybe something like:

```
* [GitHub 4998](https://github.com/xamarin/xamarin-android/issues/4998):
  *Files under lib/ must have .so extension, found 'lib/x86/gdbserver'.* build error
  prevented building app bundles for projects with references to Android C++ library
  projects.
```

PR #5019, opened 2020-08-16
Fix nullable annotation on JavaCast

AArnott:
Assuming `JavaCast<TResult>` never returns null except when its input is null, this annotation conveys the idea so C# realizes that given a non-null input, a non-null output will come out, thereby avoiding compiler warnings in calling code caused by the addition of the less precise nullable annotations added earlier.

### DIFF
--- a/src/Mono.Android/Android.Runtime/Extensions.cs
+++ b/src/Mono.Android/Android.Runtime/Extensions.cs
@@ -6,7 +6,7 @@ namespace Android.Runtime {
 
 	public static class Extensions {
 
-		[return: NotNullIfNotNull("instance")]
+		[return: NotNullIfNotNull ("instance")]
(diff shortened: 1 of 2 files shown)


brendanzagaeski:
_Release status update_

A new Release version of Xamarin.Android has now been published that includes the fix from this item. Thanks again for the contribution!

Fix included in Xamarin.Android SDK version 11.1.0.17.

Fix included on Windows in Visual Studio 2019 version 16.8. To get the new version that includes the fix, [check for the latest updates](https://docs.microsoft.com/visualstudio/install/update-visual-studio?view=vs-2019) or install the most recent release from <https://visualstudio.microsoft.com/downloads/>.

Fix included on macOS in Visual Studio 2019 for Mac version 8.8. To get the new version that includes the fix, check for the latest updates on the **Stable** [updater channel](https://docs.microsoft.com/visualstudio/mac/update).

PR #5020, opened 2020-08-17
[runtime] package and use unstripped DSOs

grendello:
To save on distribution size, Xamarin.Android hasn't been shipping the
unstripped Mono native libraries for quite a while now.  However, that
causes stack traces, when the native side crashes, quite tedious to read
since they usually provide liitle to no information about the exact
location of the crash since the binaries have no non-public symbol
information.

This commit introduces several changes in this area:

 * We now package the .d.so versions (full debug info) of the DSOs
 * The .d.so versions are packaged into Debug APKs if
   the **unsupported** `$(_AndroidDebugNativeLibraries)` is set to
   `True` (it defauls to `False`)
 * The .so libraries which so far have been stripped of all the
   symbols, are now stripped of just the full debug information leaving
   the local/hidden symbol table in the binary.  These DSOs are the ones
   which are packaged into Release APKs.  APK size increase is around %10.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -1858,6 +1858,7 @@ because xbuild doesn't support framework reference assemblies.
   <PropertyGroup>
     <_Assemblies>@(_ResolvedFrameworkAssemblies)</_Assemblies>
 	<_TargetInterpreterPrefix Condition=" '$(_AndroidUseInterpreter)' != 'False' ">interpreter-</_TargetInterpreterPrefix>
+    <_AndroidDebugNativeLibraries Condition=" '$(_AndroidDebugNativeLibraries)' == '' ">False</_AndroidDebugNativeLibraries>
   </PropertyGroup>
   <ItemGroup>
     <_TargetArchitecture Include="$(_Android32bitArchitectures);$(_Android64bitArchitectures)" />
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1879,18 +1879,22 @@ because xbuild doesn't support framework reference assemblies.
...
     <FrameworkNativeLibrary Include="%(_TargetLibInterpreterDir.Identity)\libmono-profiler-aot.so" Condition=" @(_EmbedProfilers->Count()) > 0 and (@(_EmbedProfilers->AnyHaveMetadataValue('Identity', 'aot')) or @(_EmbedProfilers->AnyHaveMetadataValue('Identity', 'all'))) " />
     <FrameworkNativeLibrary Include="%(_TargetLibInterpreterDir.Identity)\libmono-profiler-log.so" Condition=" '$(AndroidIncludeDebugSymbols)' == 'True' or (@(_EmbedProfilers->Count()) > 0 and (@(_EmbedProfilers->AnyHaveMetadataValue('Identity', 'log')) or @(_EmbedProfilers->AnyHaveMetadataValue('Identity', 'all')))) " />
-    <FrameworkNativeLibrary Include="%(_TargetLibInterpreterDir.Identity)\libmonosgen-2.0.so" Condition=" '$(AndroidUseSharedRuntime)' != 'True' " />
+    <FrameworkNativeLibrary Include="%(_TargetLibInterpreterDir.Identity)\libmonosgen-2.0.so" Condition=" '$(AndroidUseSharedRuntime)' != 'True' and '$(AndroidIncludeDebugSymbols)' != 'True' " />
+    <FrameworkNativeLibrary Include="%(_TargetLibInterpreterDir.Identity)\libmonosgen-2.0.d.so" Condition=" '$(AndroidUseSharedRuntime)' != 'True' and '$(AndroidIncludeDebugSymbols)' == 'True' " ArchiveFileName="libmonosgen-2.0.so" />
```
jonathanpeppers, line 1897: It seems like this will make Debug `.apk` files several MB larger, hurting build & deploy times.

Can we make a new MSBuild property to enable this instead? I don't think we want every user to have this as the default in their app.

grendello: If we do that then we can be sure people won't enable that because it would make their deployment slower, so that would defy the purpose of this PR...

jonathanpeppers: If we give it a private name like `$(_AndroidDebugNativeLibraries)` or something, it won't be enabled unless we ask them to do it.

It would be more helpful to have this option than what we have now, for sure.

jonpryor:
```markdown
To save on `.pkg` and `.vsix` installer size, Xamarin.Android hasn't
been shipping the unstripped Mono native libraries for quite a while
now.  However, this means that when a native crash occurs, the stack
traces are tedious/impossible to read, as they usually provide little-
to-no information about the exact location of the crash, as the
binaries have no non-public symbol information.

This commit introduces several changes in this area:

  * We now package the `.d.so` versions of the shared libraries, which
    contain full debug information.

  * Add support for a new *unsupported* boolean MSBuild property,
    `$(_AndroidDebugNativeLibraries)`, which defaults to False.
    When True, the `.d.so` versions will be used in the `.apk`,
    e.g. `libmonosgen-2.0.d.so` will be added to the `.apk` as
    `libmonosgen-2.0.so`.

  * The "normal" `.so` libraries, which previously had been `strip`ed
    of *all* symbols, now use `strip --strip-debug`, which leaves the
    local & hidden symbol tables within the `.so`.

The use of `--strip-debug` causes the `.so` files to increase in size
by ~12%, e.g. `x86_64/libmonosgen-2.0.so` increases in size from
4.18MB to 4.67MB.

The inclusion of the `.d.so` files into the installers likewise
increases the `.vsix` install size ~12%, from ~334MB to ~377MB.
```

Issue #4986, opened 2020-08-06
After Upgrade from Visual Studio 2019 16.6 to 16.7 APK Size Increases significantly from 24 MB to 64 MB


inforithmics:
### Steps to Reproduce

1.  Open Attached Xamarin.Android.zip
2.  Change To Release Mode
3.  Execute Deploy (The generated Xamarin.Android.apk is 64 MB)

[Xamarin.Android.zip](https://github.com/xamarin/xamarin-android/files/5037458/Xamarin.Android.zip)

### Expected Behavior
Size is 24MB

### Actual Behavior
Size is 64MB
### Version Information
Visual Studio 2019 16.7
### Work around
Changing following line in the AndroidManifest.xml
from
`<uses-sdk android:minSdkVersion="23" android:targetSdkVersion="29" />`
to
`<uses-sdk android:minSdkVersion="22" android:targetSdkVersion="29" />`
reduces the apk size to 24MB

The reason why the apk with minsdkversion 23 is so much bigger is that that the libs aren't anymore compressed
![image](https://user-images.githubusercontent.com/2290863/89570863-0fe1c580-d827-11ea-8403-876a5aeedbb8.png)

Here the apk with MinSdkVersion="22"

![image](https://user-images.githubusercontent.com/2290863/89571066-5c2d0580-d827-11ea-925e-178d68df8edf.png)

With Visual Studio Version 2019 16.6 The size was 24MB even with minSDKVersion:="23"

Currently only x86 is selected as compiled target the app doesn't work is only used as example how the apk size is increased

jonathanpeppers:
@inforithmics the difference in the bigger `.apk` file is this value in the `AndroidManifest.xml`:
```xml
<application android:extractNativeLibs="false"
```

I think this value is emitted by default for `minSdkVersion="23"` and higher. It is something the Android toolchain is doing automatically. Google claims this really cuts down on _install_ size, as the native libraries are loading directly from the `.apk` file. Otherwise on install, they get extracted to a directory and take up more space.

I could change the manifest to this to get the smaller `.apk` file:

```xml
<?xml version="1.0" encoding="utf-8"?>
<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0" package="com.companyname.app52">
    <uses-sdk android:minSdkVersion="23" android:targetSdkVersion="29" />
[...]

PR #5021, opened 2020-08-17
[Xamarin.Android.Build.Task] emit android:extractNativeLibs="true" by default

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/4986
Context: https://github.com/xamarin/xamarin-android/commit/86737ca159d851f2743c01ffdf66428df07082b5
Context: https://android.googlesource.com/platform/tools/base/+/fd2ac00ab76dfdececce5a25cb7ad23b2b6f5d29%5E%21

When we bumped `manifest-merger` from 26.5.0 to 27.0.0, it started
emitting this in `AndroidManifest.xml` when `minSdkVersion` is 23 or
higher:

    <application android:extractNativeLibs="false">

When *omitted* this value defaults to `true`, meaning that native
libraries are compressed in `.apk` files and extracted to disk on
installation.

When `extractNativeLibs` is `false`, native libraries are
*uncompressed* in `.apk` files and loaded directly from the `.apk`.
This saves on install size at the cost of larger `.apk` size. This
behavior is only supported on API 23 devices and higher.

The Android docs say:

> The default value is "true". However, when building your app using
> Android Gradle plugin 3.6.0 or higher, this property is set to
> "false" by default.

Ignoring the related bugs caused by this behavior:

* https://github.com/xamarin/xamarin-android/commit/bf657e815b200e57098725c0ad963d70e881ea69
* https://github.com/xamarin/xamarin-android/commit/8828fef1331490352da15f1f43d7f4e3d7ccf220

The new `manifest-merger` will implicitly make `.apk` files larger. In
the case of a customer using AOT, an app grew from 24MB to 60MB!

To solve this:

* If not specified in apps' `Properties\AndroidManifest.xml`, we will
  emit `android:extractNativeLibs="true"` by default.

This way Xamarin.Android applications will have consistent behavior.
Developers can set `extractNativeLibs="false"` if they prefer that
behavior.

This behavior also works fine on an API 19 emulator, since
`extractNativeLibs="true"` is already the default and seems to be
ignored.

### DIFF
--- a/Documentation/release-notes/extractNativeLibs.md
+++ b/Documentation/release-notes/extractNativeLibs.md
@@ -1,0 +1,35 @@
+#### Application and library build and deployment
+
+* [GitHub Issue 4986](https://github.com/xamarin/xamarin-android/issues/4986):
+  Updates to Android tooling (`manifest-merger`), caused
+  `//application/@android:extractNativeLibs` to be set to `false` by
+  default. This can cause an undesirable `.apk` file size increase
+  that is more noticeable for Xamarin.Android applications using AOT.
+  Xamarin.Android now sets `extractNativeLibs` to `true` by default.
+
+According to the [Android documentation][extractNativeLibs],
+`extractNativeLibs` affects `.apk` size and install size:
+
+> Whether or not the package installer extracts native libraries from
+> the APK to the filesystem. If set to false, then your native
+> libraries must be page aligned and stored uncompressed in the APK.
+> No code changes are required as the linker loads the libraries
+> directly from the APK at runtime. The default value is "true".
+
+This is a tradeoff that each developer should decide upon on a
+per-application basis. Is a smaller install size at the cost of a
+larger download size preferred?
+
+Since Xamarin.Android now emits `android:extractNativeLibs="true"` by
+default, you can get the opposite behavior with an
+`AndroidManifest.xml` such as:
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0" package="com.companyname.hello">
+    <uses-sdk android:minSdkVersion="23" android:targetSdkVersion="30" />
+    <application android:label="Hello" android:extractNativeLibs="false" />
+</manifest>
+```
+
+[extractNativeLibs]: https://developer.android.com/guide/topics/manifest/application-element
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
@@ -191,6 +191,9 @@ namespace Xamarin.Android.Build.Tests
 			using (var b = CreateApkBuilder (Path.Combine ("temp", TestName))) {
 				Assert.IsTrue (b.Build (proj), "first build should have succeeded");
 
+				var manifest = Path.Combine (Root, b.ProjectDirectory, proj.IntermediateOutputPath, "android", "AndroidManifest.xml");
+				AssertExtractNativeLibs (manifest, extractNativeLibs: false);
+
 				var apk = Path.Combine (Root, b.ProjectDirectory,
 						proj.IntermediateOutputPath, "android", "bin", "UnnamedProject.UnnamedProject.apk");
 				AssertEmbeddedDSOs (apk);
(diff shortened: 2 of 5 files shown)


jonpryor:
Commit message for review:

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/4986

The [`//application/@android:extractNativeLibs`][0] attribute within
`AndroidManifest.xml` has focused yet large-scale impact on `.apk`s:

When `extractNativeLibs` is `true`, then (1) native libraries within
the `.apk` are *compressed*, and (2) the native libraries within the
`.apk` are extracted during installation time.  This allows the `.apk`
to be smaller, while resulting in a *larger* installation footprint,
as two copies of the native libs -- one compressed, one uncompressed --
are present on the device.

When `extractNativeLibs` is `false`, then (1) native libraries within
the `.apk` are stored *uncompressed*, and (2) the native libraries
within the `.apk` are *not* extracted during installation time, but
*only* on API-23 and later devices.  This makes for a larger `.apk`,
but a *smaller* installation footprint.

When *omitted* this value defaults to `true`.

Commit 86737ca1 bumped `manifest-merger` from 26.5.0 to 27.0.0, which
[altered `manifest-merger.jar` behavior][1] so that it started setting
`//application/@android:extractNativeLibs`=false when `minSdkVersion`
is 23 or higher, "as if" `Properties\AndroidManifest.xml` were:

	<application android:extractNativeLibs="false">

In the case of Issue #4986, which is an application which uses
Xamarin.Forms and sets `$(AotAssemblies)`=True, resulting in native
libraries being produced and included for *every assembly*, this
"minor" change to `extractNativeLibs` caused a 24MB `.apk` to balloon
in size to 64MB, which was unexpected.

This unexpected size increase was not caught by unit tests, as our
tests which check for `.apk` size regressions don't use Xamarin.Forms,
and it's AndroidX/Android Support -- used by Xamarin.Forms -- which
causes `manifest-merger.jar` to be used.  No `manifest-merger.jar`,
no unexpected size increase.

This `manifest-merger.jar` change is *also* the underlying cause of
bf657e81 and 8828fef1.

To solve this, if not explicitly specified within apps'
`Properties\AndroidManifest.xml`, we will now insert
`android:extractNativeLibs="true"` by default.

This way Xamarin.Android applications will have consistent behavior.
Developers can set `extractNativeLibs="false"` if they prefer that
behavior.

This behavior also works fine on an API-19 emulator, since
`extractNativeLibs="true"` is already the default and seems to be
ignored.

[0]: https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs
[1]: https://android.googlesource.com/platform/tools/base/+/fd2ac00ab76dfdececce5a25cb7ad23b2b6f5d29%5E%21
```

brendanzagaeski:
_Release notes_

In the final collated release notes, I'll plan to rearrange this a tiny bit, basically so it has two entries in the release notes, one as a default configuration change, and one as an issue fix.

I'll also add a paragraph about the backstory.

Feel free to comment back if anything looks bad about this rearrangement though.

````markdown
#### Deprecations, removals, and default configuration changes

- [`android:extractNativeLibs` set to `"true"` by default][#androidextractnativelibs-explicitly-set-to-true-by-default]

### `android:extractNativeLibs` set to `"true"` by default

Starting in Xamarin.Android 11.0, projects that had references to the
Xamarin.AndroidX NuGet packages or that were otherwise configured to
use the [improved manifest merging option][xa-102-manifest-merging]
would get a value of `"false"` by default for the
[`android:extractNativeLibs`][extractNativeLibs] Android manifest
attribute due to changes in the latest version of the manifest merger
tool from Google.  Because this value had previously been `"true"` when
left blank, this could cause unexpected behavior changes for
Xamarin.Android applications.  Xamarin.Android now now sets this
attribute it to `"true"` by default even when using the improved
manifest merging.

According to the [Android documentation][extractNativeLibs],
`extractNativeLibs` affects `.apk` size and install size:

> Whether or not the package installer extracts native libraries from
> the APK to the filesystem. If set to `false`, then your native
> libraries must be page aligned and stored uncompressed in the APK.
> No code changes are required as the linker loads the libraries
> directly from the APK at runtime.

This is a tradeoff that each developer should decide upon on a
per-application basis. Is a smaller install size at the cost of a
larger download size preferred?

Since Xamarin.Android now emits `android:extractNativeLibs="true"` by
default, you can get the opposite behavior with an
`AndroidManifest.xml` such as:

```xml
<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0" package="com.companyname.hello">
    <uses-sdk android:minSdkVersion="23" android:targetSdkVersion="30" />
    <application android:label="Hello" android:extractNativeLibs="false" />
</manifest>
```

[xa-102-manfiest-merging]: https://docs.microsoft.com/xamarin/android/release-notes/10/10.2#improved-android-manifest-merging
[extractNativeLibs]: https://developer.android.com/guide/topics/manifest/application-element

### Issues fixed

* [GitHub Issue 4986](https://github.com/xamarin/xamarin-android/issues/4986):
  Updates to Android tooling (`manifest-merger`), caused
  `//application/@android:extractNativeLibs` to be set to `false` by
  default. This could cause an undesirable `.apk` file size increase
  that was more noticeable for Xamarin.Android applications using AOT.
  Xamarin.Android now sets `extractNativeLibs` to `true` by default.
````

PR #5023, opened 2020-08-17
[setup-windows] Import Configuration.props earlier

brendanzagaeski:
`setup-windows.exe` was being output to:

	xamarin-android\tools\setup-windows\bin\Debug\bin\setup-windows.exe

This caused a little trouble because `setup-windows.exe` currently only
creates the correct symlinks if it is run from:

	xamarin-android\bin\Debug\bin\setup-windows.exe

The trouble was that `Configuration.props` was being imported after the
use of `$(XamarinAndroidSourcePath)`, so `$(XamarinAndroidSourcePath)`
was empty when it was used.

To fix that, move the `Configuration.props` import before the use of
`$(XamarinAndroidSourcePath)`.

### DIFF
--- a/tools/setup-windows/setup-windows.csproj
+++ b/tools/setup-windows/setup-windows.csproj
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\Configuration.props" />
+
   <PropertyGroup>
     <TargetFrameworks>net472</TargetFrameworks>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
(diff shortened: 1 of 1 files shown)


tools/setup-windows/setup-windows.csproj
```diff
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\Configuration.props" />
```
brendanzagaeski, line 3: I only looked at the pull request for 455202000ee9beefc25e07eabe52ca910bd495dd to see if there was a discussion about an important reason to have `Configuration.props` imported later, and I didn't see anything there. But there might have been earlier discussion in another PR about a reason it should be imported at the end of the file. So feel free to close this PR and take a different approach if needed. Thanks!

PR #5025, opened 2020-08-18
[tests] run CodeBehindTests under dotnet

jonathanpeppers:
`CodeBehindTests` use static files for the projects used in its tests:

* tests/CodeBehind/BuildTests/CodeBehindBuildTests.csproj
* tests/CodeBehind/CommonSampleLibrary/CommonSampleLibrary.csproj

The best way to run these under a `dotnet` context is to make a new
`CodeBehindBuildTestsDotNet.csproj` and `CommonSampleLibraryDotNet.csproj`.

We could change this in the future if there is a way to multi-target
"legacy" Xamarin.Android and Microsoft.Android.Sdk.

Other changes:

* We need support for generating `NuGet.config` and `global.json`
  files in these tests. This can be reused in future tests.
* Removed `.sln` files for these tests. We can build the `.csproj`
  files directly.
* We don't need to set `$(TargetFrameworkVersion)` under a `dotnet`
  context.
* The `Regex` that looked for errors shouldn't use the `^` to match
  the beginning of the string. C# compiler errors include full paths
  under a `dotnet` context, which appears to be new behavior?


jonathanpeppers:
Manual run here: https://build.azdo.io/3989337

Seems like these tests are working:

image

dellis1972 (approved):

grendello (approved):

Issue #5024, opened 2020-08-18
error XA0119: Using the shared runtime and Android App Bundles  at the same time is not currently supported.


cschwarz:
### Steps to Reproduce

1. Create an empty Xamarin Forms Android project
2. msbuild.exe XamarinAabIssue.Android.csproj "/p:Configuration=Release;AndroidPackageFormat=aab" --> compiles successful
3. Add an empty Xamarin Android Library project and reference it in the Android project
4. msbuild.exe XamarinAabIssue.Android.csproj "/p:Configuration=Release;AndroidPackageFormat=aab" -->

error XA0119: Using the shared runtime and Android App Bundles at the same time is not currently supported. Use the shared runtime for Debug configurations and Android App Bundles for Release configurations.

### Expected Behavior

This should work like in versions before VS 16.7

### Actual Behavior

error XA0119: Using the shared runtime and Android App Bundles at the same time is not currently supported. Use the shared runtime for Debug configurations and Android App Bundles for Release configurations.

### Version Information

Microsoft Visual Studio Community 2019 (2)
Version 16.7.1
VisualStudio.16.Release/16.7.1+30406.217
Microsoft .NET Framework
Version 4.8.04084

Installed Version: Community

Visual C++ 2019   00435-60000-00000-AA178
Microsoft Visual C++ 2019

ASP.NET and Web Tools 2019   16.7.532.28833
ASP.NET and Web Tools 2019

ASP.NET Core Razor Language Services   16.1.0.2035807+72d099b977d3a85e65fa3b0614ca8cfc803fef02
Provides languages services for ASP.NET Core Razor.

ASP.NET Web Frameworks and Tools 2019   16.7.532.28833
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   16.7.532.28833
Azure App Service Tools v3.0.0

Azure Functions and Web Jobs Tools   16.7.532.28833
Azure Functions and Web Jobs Tools

C# Tools   3.7.0-6.20375.2+34202cc2f3e869fd70a26d8237f4552cf9e192cf
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.2.6 (master@34d6af2)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

IntelliCode Extension   1.0
[...]

PR #4672, opened 2020-05-11
Bump to proguard 6.2.2

jonathanpeppers:
Testing: https://github.com/xamarin/proguard/pull/1
Changes: https://github.com/Guardsquare/proguard/compare/9d4660e5...ebe9000c

PR #5026, opened 2020-08-18
[Xamarin.Android.Build.Tasks] _CheckNonIdealConfigurations only runs for apps

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5024

Building a class library with `$(AndroidPackageFormat)` set to `aab`
fails with:

    error XA0119: Using the shared runtime and Android App Bundles at the same time is not currently supported.
        Use the shared runtime for Debug configurations and Android App Bundles for Release configurations.

In 5d8bc5c1, we upgraded this message from a warning to an error.
Unfortunately, this error is being emitted for class libraries as
well as apps!

You could easily hit the issue running MSBuild command-line:

    msbuild Foo.Android.csproj -p:Configuration=Release -p:AndroidPackageFormat=aab

If `Foo.Android.csproj` referenced a Xamarin.Android class library, it
would have the global `$(AndroidPackageFormat)` property set. Class
libraries also appear to have `$(AndroidUseSharedRuntime)` set by
default when omitted.

Looking through the `_CheckNonIdealConfigurations` MSBuild target, all
of the warnings & errors it emits only apply to Xamarin.Android
*application* projects:

* `'$(AndroidUseSharedRuntime)' == 'True' And '$(AotAssemblies)' == 'True'`
* `'$(AndroidUseSharedRuntime)' == 'True' And '$(AndroidLinkMode)' != 'None'`
* `'$(AndroidUseSharedRuntime)' == 'True' And '$(AndroidLinkTool)' != ''`
* `'$(AndroidUseSharedRuntime)' == 'True' And '$(AndroidPackageFormat)' == 'aab'`
* `'$(AndroidLinkTool)' == 'proguard' And '$(AndroidDexTool)' == 'd8'`
* `'$(AndroidDexTool)' == 'dx' and '$(UsingAndroidNETSdk)' != 'true'`
* `'$(AndroidDexTool)' == 'dx' and '$(UsingAndroidNETSdk)' == 'true'`

I think we can simply omit `_CheckNonIdealConfigurations` from
`$(BuildDependsOn)` in class library projects.

`$(AndroidUseSharedRuntime)` is being removed in #4672, so we probably
don't need to change its default value for class libraries.

I could reproduce this issue pretty easily in a unit test.

### DIFF
--- a/Documentation/release-notes/class-libraries-aab.md
+++ b/Documentation/release-notes/class-libraries-aab.md
@@ -1,0 +1,7 @@
+#### Application and library build and deployment
+
+* [GitHub Issue 5024](https://github.com/xamarin/xamarin-android/issues/5024):
+  *error XA0119: Using the shared runtime and Android App Bundles at
+  the same time is not currently supported* build error could
+  mistakenly appear for Xamarin.Android class libraries when
+  `$(AndroidPackageFormat)` is set to `aab`.
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
@@ -69,7 +69,6 @@ projects, these properties are set in Xamarin.Android.Legacy.targets.
   <PropertyGroup Condition=" '$(AndroidApplication)' != 'True' ">
     <BuildDependsOn>
       _ValidateLinkMode;
-      _CheckNonIdealConfigurations;
       _SetupDesignTimeBuildForBuild;
       _CreatePropertiesCache;
       _CleanIntermediateIfNeeded;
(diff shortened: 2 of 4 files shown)


dellis1972 (approved):

brendanzagaeski:
Didn't quite catch this to approve before merge, but just to comment quickly after the fact, thanks a bunch for the quick fix here, which looks good to me too, and very nice formatting on the release note as well. Thanks!

In the final collated release notes, I'll plan to make one tiny change to use a sort of more "boring" way to refer to the MSBuild property that'll also match how the property looks when setting it on the command line. Feel free to comment back if anything looks bad about this adjustment though.

````markdown

* [GitHub 5024](https://github.com/xamarin/xamarin-android/issues/5024):
  *error XA0119: Using the shared runtime and Android App Bundles at
  the same time is not currently supported* build error could
  mistakenly appear for Xamarin.Android class libraries when
  the `AndroidPackageFormat` MSBuild property was set to `aab`.
````

PR #692, opened 2017-07-12
[Xamarin.Android.Build.Tasks] Implement a new process for defining versionCode.

dellis1972:
Context https://bugzilla.xamarin.com/show_bug.cgi?id=51620
Context https://bugzilla.xamarin.com/show_bug.cgi?id=51618
Context https://bugzilla.xamarin.com/show_bug.cgi?id=51145

Our current version system for multiple apk's for each Abi
is a bit broken [1]. If a user for example has a versionCode
set which is 123 the final version code for an x86_64 build
ends up as 327803.
This is completely transparent to the user and also does not
follow the guidance in the documentation at [1] and [2].

So we need a new system :) but as usual we have to support the
old system. So we are introducing a new system which is more
flexible. This will only apply when the `$(AndroidCreatePackagePerAbi)`
is set to `True`.

The new system has two new properties

	<AndroidVersionCodePattern/>
	<AndroidVersionCodeProperties/>

The first allows the developer to define the Pattern to be used
for the versonCode. The pattern will be made up of a format string
which will contain keys. These keys will be replaced with values
form one of the known keys or a custom user defined one.

We define a few known key values

- abi : The current target abi converted to an int where
	- 'armeabi' = 1,
	- 'armeabi-v7a' = 2,
	- 'x86' = 3,
	- 'arm64-v8a' = 4,
	- 'x86_64' = 5,

- minSDK : The minSDK value from the manifest or 11 if not present.

- versionCode : The versionCode from the manifest.

With these keys the user can define a pattern of

	{abi}{minSDK}{versionCode}

or if they way to include zero padding they can use

	{abi}{minSDK}{versionCode:D4}

similar to the left padding formats used in string.Format ().

Users can also use the `$(AndroidVersionCodeProperties)` property
to define new custom keys. This string will be in the form of a
semi-colon delimited key=value pairs. For example

	foo=12;bar=$(SomeBuildProperty)

when can then be used in the pattern.

	{abi}{foo}{bar}{versionCode}

Lets work through an example. The user defines a version code of '123'
in the manifest and enables `$(AndroidCreatePackagePerAbi)`. They define
a `$(AndroidVersionCodePattern)` of `{abi}{versionCode:D5}`.
This will result in the following version code being produced for the
'x86' build.

	300123

The first 3 is the `{abi}` value. The rest is the left zero padded
versionCode.
A slightly more complex pattern would be `{abi}{minSDK:D2}{versionCode:D4}`
which would produce

	3140123

[...]

PR #689, opened 2017-07-11
[monodroid] Update to new profiler init symbol name.

alexrp:
**Do not merge yet.** This is intended for the next monthly integration branch, not master.

PR #5031, opened 2020-08-19
Bump to xamarin/java.interop/master@dcaf794d

jonpryor:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1139578

Context: https://github.com/xamarin/java.interop/pull/691
Context: https://liquid.microsoft.com/Web/Object/Read/ms.security/Requirements/Microsoft.Security.SystemsADM.10039#guide

Changes: https://github.com/xamarin/java.interop/compare/007b35b489173010de038c95cb2a0d3ec514f30d...dcaf794d26206834a0b08bda85ba85b1c15b7bd1

  * xamarin/Java.Interop@dcaf794d: [generator] Fix error and warning codes values (#697)
  * xamarin/Java.Interop@08ff4db1: [java-interop, Java.Interop] Securely load native libs (#691)
  * xamarin/Java.Interop@09c68911: [generator] Clean up generated whitespace (#692)
  * xamarin/Java.Interop@d664e90e: [generator] Make warning and error messages localizable. (#689)

For improved security, the guidance is that when loading `.dll` files
on Windows, the default [`LoadLibrary()`][0] behavior is to be
*avoided*, as the [Dynamic-Link Library Search Order][1] includes the
"current working directory" and directories listed in `%PATH%`, both
of which are under control of an "attacker" and not the application.

Instead, [`LoadLibraryEx()`][2] should be used, providing a set of
`LOAD_LIBRARY_SEARCH_*` values which constrain the set of directories
that will be searched for the library (if a full path is not used)
and/or the library's dependencies.

Historically, Xamarin.Android hasn't directly called any
`LoadLibrary()` function.  Instead, xamarin-android called
[**dlopen**(3)][3], then used dlfcn-win32/dlfcn-win32@ef7e412d to
implement `dlopen()` in terms of `LoadLibraryEx()`.

Unfortunately, dlfcn-win32/dlfcn-win32@ef7e412d calls `LoadLibraryEx()`
with `LOAD_WITH_ALTERED_SEARCH_PATH`, which while a security
improvement, does not go far enough for our internal requirements.

Migrate away from dlfcn-win32 and instead use the new
`java-interop-dlfcn.h` family of functions added in
xamarin/java.interop@08ff4db1.  These new functions appropriately use
`LoadLibraryEx()` with a constrained set of `LOAD_LIBRARY_SEARCH_*`
flags.

*Non-obvious semantic note*: On Mono for all platforms, P/Invoking
`__Internal` is equivalent to trying to load the path `NULL`, e.g.
on macOS `dlopen(NULL, …)`.  However, only Android allows a custom
"`loader_func`" to intercept the attempted load of `NULL` and "do
something" with it, which xamarin-android *does* intercept, "remapping"
`NULL` to `libxa-internal-api.so`.  On all other platforms,
`loader_func` cannot intercept this invocation.

…which made for an "interesting" interaction: on macOS, if we used:

	api_dso_handle = java_interop_lib_load (dso_path.get (), JAVA_INTEROP_LIB_LOAD_LOCALLY, nullptr);

then the macOS Designer integration tests would fail!

	Renderer >> 4 [monodroid] Calling into managed runtime init
	Renderer (error) >>
	Renderer (error) >> Unhandled Exception:
	Renderer (error) >> System.EntryPointNotFoundException: java_interop_jnienv_get_java_vm assembly:<unknown assembly> type:<unknown type> member:(null)
	Renderer (error) >> at (wrapper managed-to-native) Java.Interop.NativeMethods.java_interop_jnienv_get_java_vm(intptr,intptr&)
	Renderer (error) >> at Java.Interop.JniEnvironment+References.GetJavaVM (System.IntPtr jnienv, System.IntPtr& vm) [0x00000] in <0f003a4904fd44d0a8cc6a63962ab40b>:0
	Renderer (error) >> at Java.Interop.JniEnvironmentInfo.set_EnvironmentPointer (System.IntPtr value) [0x00037] in <0f003a4904fd44d0a8cc6a63962ab40b>:0
	Renderer (error) >> at Java.Interop.JniEnvironmentInfo..ctor (System.IntPtr environmentPointer, Java.Interop.JniRuntime runtime) [0x00006] in <0f003a4904fd44d0a8cc6a63962ab40b>:0
	Renderer (error) >> at Java.Interop.JniRuntime..ctor (Java.Interop.JniRuntime+CreationOptions options) [0x0017b] in <0f003a4904fd44d0a8cc6a63962ab40b>:0
	Renderer (error) >> at Android.Runtime.AndroidRuntime..ctor (System.IntPtr jnienv, System.IntPtr vm, System.Boolean allocNewObjectSupported, System.IntPtr classLoader, System.IntPtr classLoader_loadClass, System.Boolean jniAddNativeMethodRegistrationAttributePresent) [0x00000] in /Users/builder/azdo/_work/4/s/xamarin-android/src/Mono.Android/Android.Runtime/AndroidRuntime.cs:25

In order for the macOS Designer integration tests to succeed,
`libxa-internal-api.dylib` *must* be loaded as `RTLD_GLOBAL`, which is
done via `JAVA_INTEROP_LIB_LOAD_GLOBALLY`.

[0]: https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibrarya
[1]: https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order
[2]: https://docs.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryexa
[3]: https://linux.die.net/man/3/dlopen


jonpryor:
Related: https://github.com/xamarin/java.interop/pull/691

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -10,7 +10,10 @@
 #include <errno.h>
 #include <limits.h>
 
+#if defined (APPLE_OS_X)
```
grendello, line 13: What about Linux? :)

grendello (changes requested):

src/monodroid/jni/monodroid-glue.cc
jonpryor: The only reason it's needed *at all* is for `MonodroidRuntime::get_my_location()`, which on macOS uses `dladdr()`.  Linux doesn't use `dladdr()`, and thus it isn't needed.

radekdoulik (commented):
There is something wrong with warning codes, like in `Xamarin.Android.Build.Tests.BuildTest.GeneratorValidateEventName(False,True,"","")` test. Where we expect `BG8405` warning, but we get `BG2138`:

```
  BINDINGSGENERATOR : warning BG2138: Event name for 'Com.Xamarin.Testing.Test.IOn123Listener.SetOn123Listener' is invalid. A valid 'eventName' or 'argsType' can be assigned by adding a rule to the Metadata.xml transforms file. [/Users/runner/work/1/s/bin/TestRelease/temp/GeneratorValidateEventNameFalseTrue/UnnamedProject.csproj]
```

The `Localized` message seems to have the right number:
```
public static LocalizedMessage WarningInvalidEventName => new LocalizedMessage (8504, Java.Interop.Localization.Resources.Generator_BG8504);
```

@jpobst could you please take a look?

radekdoulik:
Oh, I see where is the problem. We print it as hexadecimal value, which I guess is not what we want. I will try to fix it.

radekdoulik:
We have some hexadecimal coded warnings, so I ended up with making the codes as hexdecimal values as well, so that we can print everything as before the localization changes.

This PR https://github.com/xamarin/java.interop/pull/697 should fix it and we can update this PR once it is merged.

radekdoulik (approved):

jonpryor:
Forgot to also mention: https://github.com/mono/mono/issues/20295

PR #5040, opened 2020-08-21
[tests] enable One .NET tests for F#

jonathanpeppers:
Context: https://github.com/xamarin/Xamarin.Android.FSharp.ResourceProvider/pull/6

Xamarin.Android.FSharp.ResourceProvider 1.0.1 is now built for
`netstandard2.0` and published to an internal feed.

We should be able to enable F# tests under a `dotnet` context now.

The `_RemoveLegacyDesigner` MSBuild target was running, when it
shouldn't be for F# projects and adding a `Resource.designer.cs`
`@(Compile)` item. I needed to add a `$(Language)` check.

One workaround is for:

https://github.com/dotnet/sdk/issues/12954

We need to set `$(ProduceReferenceAssembly)` to `false` for .NET 5 F#
projects at the moment.


dellis1972 (approved):

jonathanpeppers:
This looks good to merge, I opened an issue for the F# workarounds we have: https://github.com/xamarin/xamarin-android/issues/5076

PR #5046, opened 2020-08-24
[build] add attempt # for designer tests

jonathanpeppers:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4000236&view=logs&j=db00894d-3ef4-5d97-073c-254fbd613a41&t=2d91500b-42fa-5f4d-9c78-a0a207f1288c&l=2042

A re-run of the designer tests will fail with:

    ##[error]Artifact Test Results - Designer - macOS already exists for build 4000236.

All of our other calls to `PublishPipelineArtifact` go through the
steps in `upload-results.yaml`. I moved the logic that sets
`$(UploadAttemptSuffix)` so that it can be used by the designer tests
as well.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -955,12 +955,11 @@ stages:
         flattenFolders: true
       condition: always()
 
-    - task: PublishPipelineArtifact@1
-      displayName: upload designer binlogs
-      inputs:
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
Attempt 1 and 2 both failed of the designer/mac tests, but the artifact name looks fixed:

image

Issue #5049, opened 2020-08-26
Some of the C#8 features don't compile - Visual Studio v16.7.2


opcodewriter:
### Steps to Reproduce

1. In Visual Studio **v16.7.2** (it's the latest released) on Windows 10, create a Xamarin Android app

2. In the MainActivity.cs, in the MainActivity class, add the following C#8 code which uses a **switch expression**:

```
        int FromRainbow(Rainbow colorBand) => colorBand switch
        {
            Rainbow.Red => 0,
            Rainbow.Orange => 1,
             _ => throw new ArgumentException(message: "invalid enum value", paramName: nameof(colorBand))
        };
```
3.  Compiler shows error:
`
CS8370	Feature 'recursive patterns' is not available in C# 7.3. Please use language version 8.0 or greater.`
 
Interestingly, if I'm using a **local function** which is also a C# 8 feature, it compiles fine:
```
public class MainActivity : AppCompatActivity
{
        protected override void OnCreate(Bundle savedInstanceState)
        {
              ...

             LocalFunction();

              void LocalFunction() => Console.WriteLine("A");
        }
}
```

Why would one C# 8 feature work but not the other.

I am aware I could try edit the csproj and force by using `LangVersion`, but shouldn't this not be necessary anymore?

By default, the project targets Android 9. I also tried with Android 10, but it didn't work, gives the same compilation error.

**I could swear the switch expression worked fine in some previous version of Visual Studio 2019!** I have some code which compiled fine some months ago this year

### Expected Behavior

Compile

### Actual Behavior

Doesn't compile

### Log File

```
Microsoft Visual Studio Enterprise 2019
Version 16.7.2
VisualStudio.16.Release/16.7.2+30413.136
Microsoft .NET Framework
Version 4.8.03752
```
<Details>
```
Installed Version: Enterprise

Visual C++ 2019   00433-90101-78013-AA222
Microsoft Visual C++ 2019

ASP.NET and Web Tools 2019   16.7.532.28833
ASP.NET and Web Tools 2019

ASP.NET Core Razor Language Services   16.1.0.2035807+72d099b977d3a85e65fa3b0614ca8cfc803fef02
Provides languages services for ASP.NET Core Razor.

ASP.NET Web Frameworks and Tools 2019   16.7.532.28833
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   16.7.532.28833
Azure App Service Tools v3.0.0

[...]

PR #5054, opened 2020-08-27
[Xamarin.Android.Build.Tasks] fix C# 8.0 support by default

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5049
Context: https://github.com/dotnet/roslyn/commit/12f1b8f6c17c0bcd271525b5f4db0f621944c49d#diff-125e2d8c52dd9033f9b248f79f78c3f8
Context: https://github.com/dotnet/roslyn/pull/44911#discussion_r477418327

The `BuildTest.CSharp8Features` test was failing for me locally where
I have Visual Studio 2019 16.7.2 installed:

    Foo.cs(1,23): error CS8370: Feature 'using declarations' is not available in C# 7.3. Please use language version 8.0 or greater.

After review, it appears that a change in Roslyn's MSBuild targets
removed `$(MaxSupportedLangVersion)` and made the property private:
`$(_MaxSupportedLangVersion)`.

After some discussion around C# 9.0:

* It is unknown if C# 9.0 will be supported by mono/mono/2020-02. Some
  features like covariant return types and function pointers require
  runtime changes.
* Using a build of .NET 5 rc1, `$(LangVersion)` is already being set
  to `9.0` by default.

We should just change the current behavior of:

    <MaxSupportedLangVersion Condition=" '$(MaxSupportedLangVersion)' == '' ">8.0</MaxSupportedLangVersion>

And change it to:

    <LangVersion Condition=" '$(LangVersion)' == '' ">8.0</LangVersion>

This matches what Roslyn's default behavior is:

https://github.com/dotnet/roslyn/blob/1aeda2e94fb9371b96d4bfd94b074860064ec3d2/src/Compilers/Core/MSBuildTask/Microsoft.CSharp.Core.targets#L6-L21

.NET 5+ projects should be unaffected by this change, as they do not
import `Xamarin.Android.CSharp.targets`.

If at some point, mono/mono/2020-02 fully supports C# we can bump this
value. Users can opt into `9.0` on their own if desired.

`BuildTest.CSharp8Features` now passes for me locally.

### DIFF
--- a/Documentation/release-notes/csharp-8.md
+++ b/Documentation/release-notes/csharp-8.md
@@ -1,0 +1,5 @@
+#### Application and library build and deployment
+
+* [GitHub Issue 5049](https://github.com/xamarin/xamarin-android/issues/5049):
+  *Feature 'XYZ' is not available in C# 7.3. Please use language version 8.0 or greater.*
+  build error could prevent projects from using C# 8.0 language features.
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets
@@ -44,7 +44,7 @@ Copyright (C) 2012 Xamarin Inc. All rights reserved.
     <TargetFrameworkIdentifier>MonoAndroid</TargetFrameworkIdentifier>
     <MonoAndroidVersion>v5.0</MonoAndroidVersion>
     <TargetFrameworkVersion Condition="'$(TargetFrameworkVersion)' == '' And '$(AndroidUseLatestPlatformSdk)' == 'False'">v5.0</TargetFrameworkVersion>
-    <MaxSupportedLangVersion Condition=" '$(MaxSupportedLangVersion)' == '' ">8.0</MaxSupportedLangVersion>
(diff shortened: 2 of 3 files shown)


Documentation/release-notes/csharp-8.md
```diff
@@ -0,0 +1,5 @@
+#### Application and library build and deployment
+
+* [GitHub Issue 5049](https://github.com/xamarin/xamarin-android/issues/5049):
+  *Feature 'XYZ' is not available in C# 7.3. Please use language version 8.0 or greater.*
+  build error could prevent projects from using C# 8.0 language features.
```
brendanzagaeski, line 5: Release note approved. Thanks!

In the final collated release notes, I'll plan to make one tiny addition to include the version where the behavior started. Feel free to comment back if anything looks bad about this adjustment though.

```markdown

* [GitHub Issue 5049](https://github.com/xamarin/xamarin-android/issues/5049):
  Starting in Visual Studio 2019 version 16.7, build errors similar to *Feature
  'XYZ' is not available in C# 7.3. Please use language version 8.0 or greater.*
  could prevent projects from using C# 8.0 language features.
```

PR #5062, opened 2020-08-31
[XF test] Re-enable jnimarshalmethod-gen

radekdoulik:
Commit 2181e786147d43c048c1164db210db9cb0238988 disabled the use of `jnimarshalmethod-gen` in our XF test. Let see whether the problem still exists.


radekdoulik:
Looks like the problem still happens.

Relevant mono issue: https://github.com/dotnet/runtime/issues/34389

PR #5071, opened 2020-09-01
[Xamarin.Android.Build.Tasks] fix for `aapt compile` and .DS_Store

jonathanpeppers:
Context: https://developercommunity.visualstudio.com/content/problem/1141659/android-build-process-never-stops.html
Context: https://feedback.devdiv.io/1141659

If you add the following file to a project:

    <AndroidResource Include="Resources\drawable\.DS_Store" />

It can just be an empty file.

The `<Aapt2Compile/>` MSBuild task will block indefinitely with:

    Executing compile -o C:\src\xamarin-android\samples\HelloWorld\obj\Debug\110\flat\ C:\src\xamarin-android\samples\HelloWorld\obj\Debug\110\res\drawable\.ds_store

In VS for Mac, the "stop" button will not stop the build either.
Clicking this button has no effect. At the command-line, Ctrl+C works
properly.

The above `aapt2 compile` command does not error and does not output
any files.

At first I found this code:

    const char * const gDefaultIgnoreAssets =
        "!.svn:!.git:!.ds_store:!*.scc:.*:<dir>_*:!CVS:!thumbs.db:!picasa.ini:!*~";

https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/master/tools/aapt/AaptAssets.cpp#61

So I tested the following paths in a project:

* Resources\raw\.git
* Resources\raw\.svn
* Resources\raw\.DS_Store
* Resources\raw\picasa.ini
* Resources\raw\Thumbs.db
* Resources\raw\foo - with FileAttributes.Hidden

Only the files that started with an actual `.` character caused the
problem. All the other examples worked fine. This let me to realize I
was reading the wrong source code.

The skipping behavior is here:

    // Skip hidden input files
    if (file::IsHidden(path)) {
      continue;
    }

https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/master/tools/aapt2/cmd/Compile.cpp#652

Which simply checks for the `.` character:

    bool IsHidden(const android::StringPiece& path) {
      return util::StartsWith(GetFilename(path), ".");
    }

https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/master/tools/aapt2/util/Files.cpp#174

The following loop will block forever in this case:

    // wait for the file we expect to be created. There can be a delay between
    // the daemon saying "Done" and the file finally being written to disk.
    if (!string.IsNullOrEmpty (job.OutputFile) && !errored) {
        while (!File.Exists (job.OutputFile)) {
            Thread.Sleep (10);
        }
    }

Two things to fix here:

1. Cancellation: the outermost `AsyncTask.CancellationToken` as well
   as the `CancellationTokenSource` need to have
   `ThrowIfCancellationRequested ()` called within the `while` loop.

2. Let's make the `<AndroidComputeResPaths/>` MSBuild task check for
   these files and skip them silently (but log a message).

I would not say these changes fix DevCom 1141659, as the root cause is
that VS for Mac adds `.DS_Store` files to a project when using the
`Select files to add from folder` dialog.

### DIFF
--- a/Documentation/release-notes/5071.md
+++ b/Documentation/release-notes/5071.md
@@ -1,0 +1,8 @@
+#### Application and library build and deployment
+
+  * [Developer Community 1141659][0]: Starting in Xamarin.Android
+    11.0, the *Updating Resources...* build step could get stuck
+    indefinitely if `.DS_Store` or any file starting with `.` was
+    included in the project with the **AndroidResource** Build Action.
+
+[0]: https://developercommunity.visualstudio.com/content/problem/1141659/android-build-process-never-stops.html
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
@@ -85,7 +85,7 @@ namespace Xamarin.Android.Tasks {
 		protected long RunAapt (string [] args, string outputFile)
 		{
 			LogDebugMessage ($"Executing {string.Join (" ", args)}");
-			long jobid = daemon.QueueCommand (args, outputFile);
+			long jobid = daemon.QueueCommand (args, outputFile, CancellationToken);
 			jobs.Add (jobid);
 			return jobid;
 		}
--- a/src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
@@ -117,6 +117,10 @@ namespace Xamarin.Android.Tasks
 				}
 
 				string baseFileName = LowercaseFilenames ? rel.ToLowerInvariant () : rel;
+				if (Path.GetFileName (baseFileName).StartsWith (".", StringComparison.Ordinal)) {
+					Log.LogDebugMessage ($"Skipping ignored file: {baseFileName}");
+					continue;
+				}
 				if (Path.GetExtension (baseFileName) == ".axml")
 					baseFileName = Path.ChangeExtension (baseFileName, ".xml");
 				if (baseFileName != rel)
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
@@ -1376,5 +1376,25 @@ namespace UnnamedProject
 				Assert.Fail ($"aapt log message was not found: {aaptCommand}");
 			}
 		}
+
+		[Test]
+		public void InvalidFilenames ()
+		{
+			BuildItem CreateItem (string include) =>
+				new AndroidItem.AndroidResource (include) {
+					TextContent = () => "",
+				};
+
+			var proj = new XamarinAndroidApplicationProject ();
+			proj.AndroidResources.Add (CreateItem ("Resources\\raw\\.foo"));
+			proj.AndroidResources.Add (CreateItem ("Resources\\raw\\.git"));
+			proj.AndroidResources.Add (CreateItem ("Resources\\raw\\.svn"));
+			proj.AndroidResources.Add (CreateItem ("Resources\\raw\\.DS_Store"));
+			using (var b = CreateApkBuilder ()) {
+				Assert.IsTrue (b.Build (proj), "first build should have succeeded.");
+				Assert.IsTrue (b.Build (proj), "second build should have succeeded.");
+				b.Output.AssertTargetIsSkipped ("_CompileResources");
+			}
+		}
 	}
 }
--- a/src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/Aapt2Daemon.cs
@@ -34,13 +34,15 @@ namespace Xamarin.Android.Tasks
 			public long JobId { get; private set; }
 			public string OutputFile { get; private set; }
 			public bool Succeeded { get; set; }
+			public CancellationToken Token { get; private set; }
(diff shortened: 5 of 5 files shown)


src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -130,6 +130,14 @@
     <value>Invalid file name: It must contain only {0}.</value>
     <comment>{0} - The regular expression that the file name must match</comment>
   </data>
+  <data name="APT0004" xml:space="preserve">
+    <value>Invalid directory name. Directory names must not match the pattern '{0}'.</value>
+    <comment>{0} - The regular expression that the directory name must *not* match.</comment>
+  </data>
+  <data name="APT0005" xml:space="preserve">
+    <value>Invalid file name. File names must not match the pattern '{0}'.</value>
```
dellis1972, line 138: Should we mention `AndroidResouce` here? as its specific that that ItemGroup cannot contain those items.

jonathanpeppers: This ignores the files silently now, so we don't need a new error code.

dellis1972 (approved):

Documentation/release-notes/5071.md
```diff
@@ -0,0 +1,7 @@
+#### Application and library build and deployment
+
+  * [Developer Community 1141659][0]: Fixes an issue where including a
+    `.DS_Store` or any file starting with `.` as an `AndroidResource`
+    could cause application builds to hang indefinitely.
```
brendanzagaeski, line 5: To provide an example of the user-visible symptoms, include the regression information, and mention the IDE terminology "Build Action," maybe:

```markdown
   * [Developer Community 1141659][0]: Starting in Xamarin.Android 11.0,
     the *Updating Resources...* and *Cleaning solution* build steps
     could get stuck indefinitely if `.DS_Store` or any file starting
     with `.` was included in the project with the **AndroidResource**
     Build Action.
```

jonathanpeppers: Looks good, I don't think this would actually cause `Clean` to hang because it doesn't run `aapt2` at all, so I left that out.

brendanzagaeski: Sweet, and good point about cleaning solution. Not sure what was happening for the user on the Developer Community issue in that case, but I wouldn't be surprised if it was just an odd side-effect of the earlier problem when building that'll all be taken care of now. Anyway, sounds good to me on leaving out the `*Cleaning solution*` part from the release note :+1:

brendanzagaeski (approved):
Release note approved. Thanks!

jonpryor:
```markdown
Context: https://developercommunity.visualstudio.com/content/problem/1141659/android-build-process-never-stops.html
Context: https://feedback.devdiv.io/1141659

If you add an empty file named `.DS_Store` as an `@(AndroidResource)`
to your project, e.g.:

	<AndroidResource Include="Resources\drawable\.DS_Store" />

then the `<Aapt2Compile/>` MSBuild task will block indefinitely with:

	Executing compile -o C:\src\xamarin-android\samples\HelloWorld\obj\Debug\110\flat\ C:\src\xamarin-android\samples\HelloWorld\obj\Debug\110\res\drawable\.ds_store

(This can be accidentally done within Visual Studio for Mac if one
uses **Add** > **Files from Folder…** > [* select folder *], then
*not* "clearing" the checkbox for any `DS_Store` files within the
**Select files to add from Folder** dialog.)

Furthermore, within Visual Studio for Mac, the **Stop** button will
not stop the build either!  Clicking this button has no effect.

(Command-line builds *will* stop when Ctrl+C is used.)

It's not files named `.DS_Store` which causes the hang; that was just
the first one encountered.  Any filename which starts with `.` will
trigger a hang.

The hang is *not* (solely) within `aapt2`, either.
`aapt2 compile` exits.

The cause of the hang is due to the interaction between `aapt2` and
our `Aapt2Daemon`; see commit 0d6fd83b.

An `aapt2` "issue" is described in 0d6fd83b:

> [T]he `aapt2 link` command sometimes reports it is "Done"
> before it has even written the archive for the file.  We can get into
> a position where we think we are done but the file is not on disk.

Because of this "delay" between when `aapt2` reports that it's done
and when it's *actually* done -- output file exists -- the
`Aapt2Daemon.Aapt2DaemonStart()` method blocks until the expected
output file actually exists.

The problem with `.DS_Store` is that `aapt2` exits,
[having *skipped* "hidden" input files][0]

	// Skip hidden input files
	if (file::IsHidden(path)) {
	  continue;
	}

…which [simply checks for the `.` character][1]:

	bool IsHidden(const android::StringPiece& path) {
	  return util::StartsWith(GetFilename(path), ".");
	}

…but in the process of "skipping" `.DS_Store`, `aapt2 compile` doesn't
*do* anything.  As such, the output file never exists, and the polling
that `Aapt2Daemon.Aapt2DaemonStart()` to grab the created file never
ends:

	// wait for the file we expect to be created. There can be a delay between
	// the daemon saying "Done" and the file finally being written to disk.
	if (!string.IsNullOrEmpty (job.OutputFile) && !errored) {
	    while (!File.Exists (job.OutputFile)) {
	        Thread.Sleep (10);
	    }
	}

Two things to fix here:

 1. Cancellation: the outermost `AsyncTask.CancellationToken` as well
    as the `CancellationTokenSource` need to have
    `ThrowIfCancellationRequested ()` called within the `while` loop.

 2. Let's make the `<AndroidComputeResPaths/>` MSBuild task check for
    these files and skip them silently (but log a message).

I would not say these changes fix DevCom 1141659, as the root cause is
that VS for Mac adds `.DS_Store` files to a project when using the
`Select files to add from folder` dialog.

[0]: https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/android10-dev/tools/aapt2/cmd/Compile.cpp#651
[1]: https://android.googlesource.com/platform/frameworks/base.git/+/refs/heads/android10-dev/tools/aapt2/util/Files.cpp#174
```

PR #5064, opened 2020-08-31
Add DevDiv required Roslyn analyzers to all projects.

jpobst:

PR #4973, opened 2020-07-31
Migration to sdk short style projects 03

moljac:
src/ folder

PR #4950, opened 2020-07-22
Migration to sdk short style projects 02 (20200722)

moljac:
Further migratons to SDK/short style projects

```
./build-tools/remap-assembly-ref/remap-assembly-ref.csproj
./build-tools/unix-distribution-setup/unix-distribution-setup.csproj
./build-tools/create-pkg/create-pkg.csproj
./build-tools/check-boot-times/check-boot-times.csproj
./build-tools/conjure-xamarin-android-cecil/conjure-xamarin-android-cecil.csproj
./build-tools/proprietary/proprietary.csproj
./build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks.csproj
./build-tools/plots-to-appinsights/ProcessPlotCSVFile.csproj
./build-tools/timing/timing.csproj
./build-tools/create-vsix/create-vsix.csproj
./build-tools/api-merge/api-merge.csproj
./build-tools/api-xml-adjuster/api-xml-adjuster.csproj
./build-tools/xa-prep-tasks/xa-prep-tasks.csproj
./build-tools/jnienv-gen/jnienv-gen.csproj
./build-tools/xaprepare/xaprepare/xaprepare.csproj
```

PR #5073, opened 2020-09-01
[csproj] Convert non-managed projects to SDK-style "Microsoft.Build.NoTargets" projects.

jpobst:
Context: https://github.com/xamarin/xamarin-android/pull/5064

Xamarin.Android has many "projects" that are `.csproj`, but do not actually compile a managed assembly.  Instead, the `Build` and `Clean` targets are overridden to perform other tasks, like compile Java.  When using a `Directory.Build.props` file to add a NuGet to all projects, these projects failed to build due to a NuGet error.

Nowadays there is a better solution for these types of projects that produces more correct projects with fewer side-effects: https://github.com/microsoft/MSBuildSdks/tree/master/src/NoTargets.

This commit converts these projects to SDK-style projects based on `Microsoft.Build.NoTargets`.

Additionally, we add the `Microsoft.Build.NoTargets` version to a `global.json` so the version is consistent among all the projects.

This PR supersedes some of the changes in:
- https://github.com/xamarin/xamarin-android/pull/4950
- https://github.com/xamarin/xamarin-android/pull/4973

### DIFF
--- a/src/aapt2/aapt2.csproj
+++ b/src/aapt2/aapt2.csproj
@@ -1,14 +1,13 @@
-<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project Sdk="Microsoft.Build.NoTargets">
   <PropertyGroup>
-    <Configuration Condition="'$(Configuration)'==''">Debug</Configuration>
-    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-    <ProjectGuid>0c31de30-f9df-4312-bffe-dcad558ccf08</ProjectGuid>
-    <OutputType>Exe</OutputType>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+  </PropertyGroup>
+  
+  <PropertyGroup>
     <OutputPath Condition="'$(OutputPath)'==''">bin\$(Configuration)</OutputPath>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)'=='Debug'" />
-  <PropertyGroup Condition="'$(Configuration)'=='Release'" />
+  
   <Import Project="..\..\Configuration.props" />
-  <Import Project="$(MSBuildBinPath)\Microsoft.Common.targets" />
   <Import Project="aapt2.targets" />
 </Project>
--- a/build-tools/proprietary/proprietary.csproj
+++ b/build-tools/proprietary/proprietary.csproj
@@ -1,24 +1,11 @@
-<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project Sdk="Microsoft.Build.NoTargets" >
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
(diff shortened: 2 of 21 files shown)


src/aapt2/aapt2.csproj
```diff
@@ -1,14 +1,13 @@
-<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project Sdk="Microsoft.Build.NoTargets/2.0.1">
```
jonathanpeppers, line 1: Can you put a `global.json` in the top directly like done here:

https://github.com/xamarin/net6-samples/blob/main/global.json

```json
{
    "msbuild-sdks": {
            "Microsoft.Build.NoTargets": "2.0.1"
    }
}
```

I think that would allow us to put `<Project Sdk="Microsoft.Build.NoTargets">` in every file and omit the version number? Hopefully it works for sub-directories.

jpobst: Looks like it, much cleaner!

jonpryor:
Summary:

```
[build] Use Microsoft.Build.NoTargets for non-managed (#5073)
```

Commit:

```markdown
Context: https://github.com/xamarin/xamarin-android/pull/5064

Xamarin.Android has many "projects" that are `.csproj` files, but do
not actually compile a managed assembly.  Instead, their `Build` and
`Clean` targets are overridden to perform other tasks, like compile
Java code.

When using a `Directory.Build.props` file to add a NuGet to all
projects, as is done in PR #5064, these projects fail to build due to
a NuGet error:

	…/Microsoft.NuGet.targets(186,5): error : Object reference not set to an instance of an object [/Users/builder/azdo/_work/3/s/xamarin-android/src/r8/r8.csproj]
	…/Microsoft.NuGet.targets(186,5): error :   at Microsoft.NuGet.Build.Tasks.ResolveNuGetPackageAssets.IsFileValid (System.String file, System.String expectedLanguage, System.String unExpectedLanguage)
	…/Microsoft.NuGet.targets(186,5): error :   at Microsoft.NuGet.Build.Tasks.ResolveNuGetPackageAssets.TryGetFile (System.String packageName, System.String packageVersion, System.String packageRelativePath, System.String file, System.String& path)
	…/Microsoft.NuGet.targets(186,5): error :   at Microsoft.NuGet.Build.Tasks.ResolveNuGetPackageAssets.GetAnalyzers (Newtonsoft.Json.Linq.JObject lockFile)
	…/Microsoft.NuGet.targets(186,5): error :   at Microsoft.NuGet.Build.Tasks.ResolveNuGetPackageAssets.ExecuteCore ()
	…/Microsoft.NuGet.targets(186,5): error :   at Microsoft.NuGet.Build.Tasks.ResolveNuGetPackageAssets.Execute ()

Nowadays there is a better solution for these types of projects which
produces more correct projects with fewer side-effects:
The [`Microsoft.Build.NoTargets` SDK][0].

Convert these projects to SDK-style projects using the
`Microsoft.Build.NoTargets` SDK.

Normally, using this SDK would require specifying the package version
in each `<Package/>` element:

	<Project Sdk="Microsoft.Build.NoTargets/2.0.1">

Add a `global.json` file so that the `/Project/@Sdk` attribute can
*lack* the version number -- `global.json` provides the version --
allowing `Microsoft.Build.NoTarets` projects to uniformly use the
same package version:

	<Project Sdk="Microsoft.Build.NoTargets">

This supersedes some of the changes in:

  - https://github.com/xamarin/xamarin-android/pull/4950
  - https://github.com/xamarin/xamarin-android/pull/4973

[0]: https://github.com/microsoft/MSBuildSdks/tree/59125812d15b1e8a45666bb42e381d62c382c022/src/NoTargets
```

PR #5081, opened 2020-09-03
[build] dotnet test should set NUnit.NumberOfTestWorkers

jonathanpeppers:
Context: https://stackoverflow.com/a/57999644

Our nunit3-console commands pass:

    --workers=4

This helped to reduce test parallelism, which could slow the CI
machines to a crawl.

We aren't passing the same switch to instances where we run `dotnet
test`. We can do this by adding `-- NUnit.NumberOfTestWorkers=4` at
the end of the `dotnet test` command.

You can see this applied from the `dotnet test` log:

    > dotnet test .\bin\TestDebug\netcoreapp3.1\Xamarin.Android.Build.Tests.dll --logger trx --filter "Name~BuildBasicApplication" --results-directory . --diag log.txt -- NUnit.NumberOfTestWorkers=4

From `log.txt`:

    TpTrace Information: 0 : 13252, 1, 2020/09/03, 11:10:20.363, 3521331719316, vstest.console.dll, TestRunRequest.ExecuteAsync: Starting run with settings:TestRunCriteria:
    KeepAlive=False,FrequencyOfRunStatsChangeEvent=10,RunStatsChangeEventTimeout=00:00:01.5000000,TestCaseFilter=Name~BuildBasicApplication,TestExecutorLauncher=
    Settingsxml=<RunSettings>
      ...
      <NUnit>
        <NumberOfTestWorkers>4</NumberOfTestWorkers>
      </NUnit>
      ...
    </RunSettings>

We also should start doing this on macOS, as we are commonly hitting:

> We stopped hearing from agent Azure Pipelines 12. Verify the agent
> machine is running and has a healthy network connection. Anything
> that terminates an agent process, starves it for CPU, or blocks its
> network access can cause this error. For more information, see:
> https://go.microsoft.com/fwlink/?linkid=846610

We might certainly be "starving it for CPU"!

I think we should try limiting the NUnit workers to 4 on both Windows
& macOS, to see if things improve.

I setup a new `$(NUnit.NumberOfTestWorkers)` variable to be used throughout
our build, so the value can be changed in one place.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -64,6 +64,7 @@ variables:
   IsMonoBranch: $[and(eq(variables['XA.Commercial.Build'], 'true'), ne(variables['System.PullRequest.IsFork'], 'True'), or(contains(variables['Build.SourceBranchName'], 'mono-'), contains(variables['System.PullRequest.SourceBranch'], 'mono-')))]
   RunAllTests: $[or(eq(variables['XA.RunAllTests'], true), eq(variables['IsMonoBranch'], true))]
   DotNetNUnitCategories: '& TestCategory != DotNetIgnore & TestCategory != AOT & TestCategory != LibraryProjectZip & TestCategory != MkBundle & TestCategory != MonoSymbolicate & TestCategory != PackagesConfig & TestCategory != StaticProject'
+  NUnit.NumberOfTestWorkers: 4
 
 # Stage and Job "display names" are shortened because they are combined to form the name of the corresponding GitHub check.
 stages:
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
I'm not actually sure this helps anything.

I'm still seeing a lot of these on macOS:
```
We stopped hearing from agent Azure Pipelines 12. Verify the agent machine is running and has a healthy network connection. Anything that terminates an agent process, starves it for CPU, or blocks its network access can cause this error. For more information, see: https://go.microsoft.com/fwlink/?linkid=846610
```

I think we should try setting `--workers` and `NUnit.NumberOfTestWorkers` on Windows & macOS.

jonathanpeppers:
I'm going to try putting the number back to 4, this PR:

image

vs master:

image

The only thing weird is that the BCL tests hit this:
```
##[error]We stopped hearing from agent Azure Pipelines 76. Verify the agent machine is running and has a healthy network connection. Anything that terminates an agent process, starves it for CPU, or blocks its network access can cause this error. For more information, see: https://go.microsoft.com/fwlink/?linkid=846610
```
It make me think there is also something going on with the pool. I don't think the emulator should be able to completely peg the CPU of the CI machine?

jonathanpeppers:
I think the latest test run looks good, there is one random failure in `One .NET Mac 3`:
```
error ANDKT0000: keytool error: java.lang.Exception: Key pair not generated, alias <androiddebugkey> already exists
```
I think we've seen this occasionally in the past.

pjcollins (approved):

PR #5086, opened 2020-09-04
[l10n] Tidy generated resource files

brendanzagaeski:
Add the automatic changes that Visual Studio makes when saving the
resource files so there won't be any need to worry about them getting
included unintentionally as part of an unrelated future change to the
`.resx` file.

Other changes:

Remove some space and newline characters from the `AAPTManifestError`
resource.  I think what happened was that the text editor auto-indented
the content of that resource in the original commit 70790837.  Since the
`.resx` XML parser preserves whitespace, those spaces will carry over
into the final MSBuild error message that the app developer sees, which
I think was not intended in this case.

* * *

**Questions:**

This is a very tiny thing, but while I'm adjusting the spacing in `AAPTManifestError`, I remember a question from https://github.com/xamarin/xamarin-android/pull/4614 was whether it would be best to have a blank line between the AAPT error and the second line or not. I could adjust that now too if desired.

With a blank line (current):

image

Without a blank line:

image


jonathanpeppers (approved):
I don't have much of an opinion on the newlines, both look fine to me.

jonpryor:
I personally like the newline, though I *am* [the one who requested the extra newlines](https://github.com/xamarin/xamarin-android/pull/4614#discussion_r479517465)…

PR #5094, opened 2020-09-07
[C++] Add support for native code static checkers and sanitizers

grendello:
We now support the [`clang-tidy`][0] static analysis of the C++ code
and the [`ASAN`][1] and [`UBSAN`][2] clang sanitizers on the runtime to find issues
that cannot be detected by static analysis. All of the tools are shipped
with the Android NDK which currently ships the clang 9 toolchain.

All of the tools require separate builds of the native code.
`clang-tidy` builds the entire codebase to generate an AST and then
analyze it, reporting possible errors but it does not output any
binaries.

The two runtime sanitizers, however, instrument the generated code in
numerous ways thus making it much (up to 3x) slower to execute.
Therefore, we now build our native code 4 additional times - for each
sanitizer in both `Debug` and `Release` modes. The output binaries have
the `-checked+SANITIZER_NAME` infix added to their name.

The instrumented binaries will not be currently distributed to the end
users since they are intended for internal use.

Also added are two [`wrap.sh`][3] scripts, one for `ASAN` and the other
for `UBSAN` and can be found in the `build-tools/wrap.sh/` directory.

The instrumented binaries can then be placed in the APK instead of the
standard binaries by adding the following to the project file:

```xml
<PropertyGroup>
  <AndroidIncludeWrapSh Condition=" '$(UseASAN)' != '' Or '$(UseUBSAN)' != '' ">true</AndroidIncludeWrapSh>
  <_AndroidCheckedBuild Condition=" '$(UseASAN)' != '' ">asan</_AndroidCheckedBuild>
  <_AndroidCheckedBuild Condition=" '$(UseUBSAN)' != '' ">ubsan</_AndroidCheckedBuild>
  <_ASANScript>../path/to/build-scripts/wrap.sh/asan.sh</_ASANScript>
	<_UBSANScript>../path/to/build-scripts/wrap.sh/ubsan.sh</_UBSANScript>
</PropertyGroup>

<ItemGroup>
  <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
    <Link>lib\arm64-v8a\wrap.sh</Link>
  </AndroidNativeLibrary>
  <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
    <Link>lib\armeabi-v7a\wrap.sh</Link>
  </AndroidNativeLibrary>
  <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
    <Link>lib\x86\wrap.sh</Link>
  </AndroidNativeLibrary>
  <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
    <Link>lib\x86_64\wrap.sh</Link>
  </AndroidNativeLibrary>

  <AndroidNativeLibrary Include="$(_UBSANScript)" Condition=" '$(UseUBSAN)' != '' ">
    <Link>lib\arm64-v8a\wrap.sh</Link>
  </AndroidNativeLibrary>
  <AndroidNativeLibrary Include="$(_UBSANScript)" Condition=" '$(UseUBSAN)' != '' ">
    <Link>lib\armeabi-v7a\wrap.sh</Link>
  </AndroidNativeLibrary>
  <AndroidNativeLibrary Include="$(_UBSANScript)" Condition=" '$(UseUBSAN)' != '' ">
    <Link>lib\x86\wrap.sh</Link>
  </AndroidNativeLibrary>
  <AndroidNativeLibrary Include="$(_UBSANScript)" Condition=" '$(UseUBSAN)' != '' ">
    <Link>lib\x86_64\wrap.sh</Link>
  </AndroidNativeLibrary>
</ItemGroup>
```

The `Link` metadata value **must** be valid path to the native library
directory inside the APK since the wrapper scripts are
achitecture-specific and are expected by the Android loader to be found in
architecture-specific library directories.

The checked build is enabled by setting the `$(_AndroidCheckedBuild)`
MSBuild property to either `asan` or `ubsan` value.

[0]: https://releases.llvm.org/9.0.0/tools/clang/tools/extra/docs/clang-tidy/
[1]: https://releases.llvm.org/9.0.0/tools/clang/docs/AddressSanitizer.html
[2]: https://releases.llvm.org/9.0.0/tools/clang/docs/UndefinedBehaviorSanitizer.html
[3]: https://developer.android.com/ndk/guides/wrap-script

### DIFF
--- a/Documentation/guides/building-apps/build-properties.md
+++ b/Documentation/guides/building-apps/build-properties.md
@@ -555,6 +555,50 @@ The most common values for this property are:
 
 Added in Xamarin.Android 6.1.
 
+## AndroidIncludeWrapSh
+
+A boolean value which indicates whether the Android wrapper script
+([`wrap.sh`](https://developer.android.com/ndk/guides/wrap-script))
+should be packaged into the APK. The property defaults to `false`
+since the wrapper script may significantly influence the way the
+application starts up and works and the script should be included only
+when necessary e.g. for debugging or otherwise changing the
+application startup/runtime behavior.
+
+The script is added to the project using the
+[`@(AndroidNativeLibrary)`](~/android/deploy-test/building-apps/build-items.md#androidnativelibrary)
+build action, because it is placed in the same directory where
+architecture-specific native libraries, and must be named `wrap.sh`.
+
+The easiest way to specify path to the `wrap.sh` script is to put it
+in a directory named after the target architecture. This approach will
+work if you have just one `wrap.sh` per architecture:
+
+```xml
+<AndroidNativeLibrary Include="path/to/arm64-v8a/wrap.sh" />
+```
+
+However, if your project needs more than one `wrap.sh` per
+architecture, for different purposes, this approach won't work.
+Instead, in such cases the name can be specified using the `Link`
+metadata of the `AndroidNativeLibrary`:
+
+```xml
+<AndroidNativeLibrary Include="/path/to/my/arm64-wrap.sh">
+  <Link>lib\arm64-v8a\wrap.sh</Link>
+</AndroidNativeLibrary>
+```
+
+If the `Link` metadata is used, the path specified in its value must
+be a valid native architecture-specific library path, relative to the
+APK root directory. The format of the path is `lib\ARCH\wrap.sh` where
+`ARCH` can be one of:
+
++ `arm64-v8a`
++ `armeabi-v7a`
++ `x86_64`
++ `x86`
+
 ## AndroidKeyStore
 
 A boolean value which indicates whether
--- a/src/Mono.Android/Test/Mono.Android-Tests.csproj
+++ b/src/Mono.Android/Test/Mono.Android-Tests.csproj
@@ -24,6 +24,11 @@
     <TargetFrameworkVersion>v11.0</TargetFrameworkVersion>
     <AndroidDexTool Condition=" '$(AndroidDexTool)' == '' ">d8</AndroidDexTool>
     <_SkipJniAddNativeMethodRegistrationAttributeScan>True</_SkipJniAddNativeMethodRegistrationAttributeScan>
+    <AndroidIncludeWrapSh Condition=" '$(UseASAN)' != '' Or '$(UseUBSAN)' != '' ">true</AndroidIncludeWrapSh>
+    <_AndroidCheckedBuild Condition=" '$(UseASAN)' != '' ">asan</_AndroidCheckedBuild>
+    <_AndroidCheckedBuild Condition=" '$(UseUBSAN)' != '' ">ubsan</_AndroidCheckedBuild>
+    <_ASANScript>..\..\..\build-tools\wrap.sh\asan.sh</_ASANScript>
+    <_UBSANScript>..\..\..\build-tools\wrap.sh\ubsan.sh</_UBSANScript>
   </PropertyGroup>
   <Import Project="Mono.Android-Test.Shared.projitems" Label="Shared" Condition="Exists('Mono.Android-Test.Shared.projitems')" />
   <Import Project="..\..\..\Configuration.props" />
@@ -123,5 +128,32 @@
   </ItemGroup>
   <ItemGroup>
     <Folder Include="Localization\" />
+  </ItemGroup>
+  <ItemGroup>
+    <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
+      <Link>lib\arm64-v8a\wrap.sh</Link>
+    </AndroidNativeLibrary>
+    <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
+      <Link>lib\armeabi-v7a\wrap.sh</Link>
+    </AndroidNativeLibrary>
+    <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
+      <Link>lib\x86\wrap.sh</Link>
+    </AndroidNativeLibrary>
+    <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
+      <Link>lib\x86_64\wrap.sh</Link>
+    </AndroidNativeLibrary>
+
+    <AndroidNativeLibrary Include="$(_UBSANScript)" Condition=" '$(UseUBSAN)' != '' ">
+      <Link>lib\arm64-v8a\wrap.sh</Link>
+    </AndroidNativeLibrary>
+    <AndroidNativeLibrary Include="$(_UBSANScript)" Condition=" '$(UseUBSAN)' != '' ">
+      <Link>lib\armeabi-v7a\wrap.sh</Link>
+    </AndroidNativeLibrary>
+    <AndroidNativeLibrary Include="$(_UBSANScript)" Condition=" '$(UseUBSAN)' != '' ">
+      <Link>lib\x86\wrap.sh</Link>
+    </AndroidNativeLibrary>
+    <AndroidNativeLibrary Include="$(_UBSANScript)" Condition=" '$(UseUBSAN)' != '' ">
+      <Link>lib\x86_64\wrap.sh</Link>
+    </AndroidNativeLibrary>
   </ItemGroup>
 </Project>
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -88,7 +88,7 @@ namespace Xamarin.Android.Tasks
 		public bool IncludeWrapSh { get; set; }
 
 		public string CheckedBuild { get; set; }
-		
+
 		[Required]
 		public string ProjectFullPath { get; set; }
 
@@ -579,7 +579,6 @@ namespace Xamarin.Android.Tasks
 				sanitizerName = "asan";
 			} else if (String.Compare ("ubsan", mode, StringComparison.Ordinal) == 0) {
 				sanitizerName = "ubsan_standalone";
-				throw new NotImplementedException ();
 			} else {
 				Log.LogWarning ($"Unknown checked build mode '{CheckedBuild}'");
 				return;
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -87,6 +87,8 @@ namespace Xamarin.Android.Tasks
 
 		public bool IncludeWrapSh { get; set; }
 
+		public string CheckedBuild { get; set; }
+		
 		[Required]
 		public string ProjectFullPath { get; set; }
 
@@ -567,6 +569,56 @@ namespace Xamarin.Android.Tasks
 				);
 
 			AddNativeLibraries (files, supportedAbis, libs);
+
+			if (String.IsNullOrWhiteSpace (CheckedBuild))
+				return;
+
+			string mode = CheckedBuild.Trim ().ToLowerInvariant ();
+			string sanitizerName;
+			if (String.Compare ("asan", mode, StringComparison.Ordinal) == 0) {
+				sanitizerName = "asan";
+			} else if (String.Compare ("ubsan", mode, StringComparison.Ordinal) == 0) {
+				sanitizerName = "ubsan_standalone";
+				throw new NotImplementedException ();
+			} else {
+				Log.LogWarning ($"Unknown checked build mode '{CheckedBuild}'");
+				return;
+			}
+
+			if (!IncludeWrapSh) {
+				Log.LogError ("Checked builds require the wrapper script to be packaged. Please set the `$(AndroidIncludeWrapSh) MSBuild property to `true` in your project");
+				return;
+			}
+
+			if (!libs.Any (info => String.Compare (Path.GetFileName (info.Path), "wrap.sh", StringComparison.Ordinal) == 0)) {
+				// TODO: consider including various wrapper scripts with XA and adding them here
+				Log.LogError ($"Checked builds require the wrapper script to be packaged. Please add `wrap.sh` appropriate for the {CheckedBuild} checker to your project");
+				return;
+			}
+
+			NdkUtil.Init (AndroidNdkDirectory);
+			string clangDir = NdkUtil.GetClangDeviceLibraryPath (AndroidNdkDirectory);
+			if (String.IsNullOrEmpty (clangDir)) {
+				Log.LogError ($"Unable to find the clang compiler directory. Is NDK installed?");
+				return;
+			}
+
+			foreach (string abi in supportedAbis) {
+				string clangAbi = MonoAndroidHelper.MapAndroidAbiToClang (abi);
+				if (String.IsNullOrEmpty (clangAbi)) {
+					Log.LogError ($"Unable to map Android ABI {abi} to clang ABI");
+					return;
+				}
+
+				string sanitizerLib = $"libclang_rt.{sanitizerName}-{clangAbi}-android.so";
+				string sanitizerLibPath = Path.Combine (clangDir, sanitizerLib);
+				if (!File.Exists (sanitizerLibPath)) {
+					Log.LogError ($"Unable to find sanitizer runtime for the {CheckedBuild} checker at {sanitizerLibPath}");
+					return;
+				}
+
+				AddNativeLibrary (files, sanitizerLibPath, abi, sanitizerLib);
+			}
 		}
 
 		string GetNativeLibraryAbi (ITaskItem lib)
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -617,39 +617,38 @@ namespace Xamarin.Android.Tasks
 			} else if (String.Compare ("ubsan", mode, StringComparison.Ordinal) == 0) {
 				sanitizerName = "ubsan_standalone";
 			} else {
-				Log.LogWarning ($"Unknown checked build mode '{CheckedBuild}'");
+				LogSanitizerWarning ($"Unknown checked build mode '{CheckedBuild}'");
 				return;
 			}
 
 			if (!IncludeWrapSh) {
-				Log.LogError ("Checked builds require the wrapper script to be packaged. Please set the `$(AndroidIncludeWrapSh) MSBuild property to `true` in your project");
+				LogSanitizerError ("Checked builds require the wrapper script to be packaged. Please set the `$(AndroidIncludeWrapSh)` MSBuild property to `true` in your project.");
 				return;
 			}
 
 			if (!libs.Any (info => IsWrapperScript (info.Path, info.Link))) {
-				// TODO: consider including various wrapper scripts with XA and adding them here
-				Log.LogError ($"Checked builds require the wrapper script to be packaged. Please add `wrap.sh` appropriate for the {CheckedBuild} checker to your project");
+				LogSanitizerError ($"Checked builds require the wrapper script to be packaged. Please add `wrap.sh` appropriate for the {CheckedBuild} checker to your project.");
 				return;
 			}
 
 			NdkUtil.Init (AndroidNdkDirectory);
 			string clangDir = NdkUtil.GetClangDeviceLibraryPath (AndroidNdkDirectory);
 			if (String.IsNullOrEmpty (clangDir)) {
-				Log.LogError ($"Unable to find the clang compiler directory. Is NDK installed?");
+				LogSanitizerError ($"Unable to find the clang compiler directory. Is NDK installed?");
 				return;
 			}
 
 			foreach (string abi in supportedAbis) {
 				string clangAbi = MonoAndroidHelper.MapAndroidAbiToClang (abi);
 				if (String.IsNullOrEmpty (clangAbi)) {
-					Log.LogError ($"Unable to map Android ABI {abi} to clang ABI");
+					LogSanitizerError ($"Unable to map Android ABI {abi} to clang ABI");
 					return;
 				}
 
 				string sanitizerLib = $"libclang_rt.{sanitizerName}-{clangAbi}-android.so";
 				string sanitizerLibPath = Path.Combine (clangDir, sanitizerLib);
 				if (!File.Exists (sanitizerLibPath)) {
-					Log.LogError ($"Unable to find sanitizer runtime for the {CheckedBuild} checker at {sanitizerLibPath}");
+					LogSanitizerError ($"Unable to find sanitizer runtime for the {CheckedBuild} checker at {sanitizerLibPath}");
 					return;
 				}
 
@@ -703,6 +702,18 @@ namespace Xamarin.Android.Tasks
 				return;
 			}
 			files.Add (item);
+		}
+
+		// This method is used only for internal warnings which will never be shown to the end user, therefore there's
+		// no need to use coded warnings.
+		void LogSanitizerWarning (string message)
+		{
+			Log.LogWarning (message);
+		}
+
+		void LogSanitizerError (string message)
+		{
+			Log.LogError (message);
 		}
 	}
 }
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -503,9 +503,10 @@ namespace Xamarin.Android.Tasks
 			return assembliesPath;
 		}
 
-		class LibInfo
+		sealed class LibInfo
 		{
 			public string Path;
+			public string Link;
 			public string Abi;
 			public string ArchiveFileName;
 		}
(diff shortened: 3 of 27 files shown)


src/monodroid/jni/monodroid-glue.cc
```diff
@@ -652,6 +652,7 @@ MonodroidRuntime::mono_runtime_init ([[maybe_unused]] dynamic_local_string<PROPE
 		log_error (LOG_DEFAULT, "Failed to parse runtime args: '%s'", runtime_args.get ());
 	} else if (options.debug && cur_time > options.timeout_time) {
 		log_warn (LOG_DEBUGGER, "Not starting the debugger as the timeout value has been reached; current-time: %lli  timeout: %lli", cur_time, options.timeout_time);
+		delete[] options.host;
```
jonpryor, line 655: This should instead be done on line ~737, before the `#else`, so that `options.host` is deleted in all circumstances.

src/Mono.Android/Test/Mono.Android-Tests.csproj
```diff
@@ -125,4 +130,31 @@
   <ItemGroup>
     <Folder Include="Localization\" />
   </ItemGroup>
+  <ItemGroup>
+    <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
+        <Link>lib\arm64-v8a\wrap.sh</Link>
```
jonpryor, line 135: This should be a 2 space indent.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -544,25 +549,119 @@ void AddRuntimeLibraries (ZipArchiveEx apk, string [] supportedAbis)
 			}
 		}
 
+		bool IsWrapperScript (string path, string link)
+		{
+			if (Path.DirectorySeparatorChar == '/') {
```
jonpryor, line 554: How/why would `path` contain `\` when running on Linux/macOS? That feels like something that shouldn't happen?

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -544,25 +549,119 @@ void AddRuntimeLibraries (ZipArchiveEx apk, string [] supportedAbis)
...
+				return;
+
+			string mode = CheckedBuild;
+			string sanitizerName;
+			if (String.Compare ("asan", mode, StringComparison.Ordinal) == 0) {
+				sanitizerName = "asan";
+			} else if (String.Compare ("ubsan", mode, StringComparison.Ordinal) == 0) {
+				sanitizerName = "ubsan_standalone";
+			} else {
+				Log.LogWarning ($"Unknown checked build mode '{CheckedBuild}'");
```
jonpryor, line 627: `LogWarning()` should be `LogCodedWarning()`, along with corresponding docs in `Documentation/guides/messages`.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -544,25 +549,119 @@ void AddRuntimeLibraries (ZipArchiveEx apk, string [] supportedAbis)
...
+				return;
+			}
+
+			if (!IncludeWrapSh) {
+				Log.LogError ("Checked builds require the wrapper script to be packaged. Please set the `$(AndroidIncludeWrapSh) MSBuild property to `true` in your project");
```
jonpryor, line 632: `LogError()` should be `LogCodedError()`, along with corresponding docs in `Documentation/guides/messages`.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -544,25 +549,119 @@ void AddRuntimeLibraries (ZipArchiveEx apk, string [] supportedAbis)
...
+				return;
+			}
+
+			if (!libs.Any (info => IsWrapperScript (info.Path, info.Link))) {
+				// TODO: consider including various wrapper scripts with XA and adding them here
+				Log.LogError ($"Checked builds require the wrapper script to be packaged. Please add `wrap.sh` appropriate for the {CheckedBuild} checker to your project");
```
jonpryor, line 638: `LogError()` should be `LogCodedError()`, along with corresponding docs in `Documentation/guides/messages`.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -544,25 +549,119 @@ void AddRuntimeLibraries (ZipArchiveEx apk, string [] supportedAbis)
...
+				return;
+			}
+
+			NdkUtil.Init (AndroidNdkDirectory);
+			string clangDir = NdkUtil.GetClangDeviceLibraryPath (AndroidNdkDirectory);
+			if (String.IsNullOrEmpty (clangDir)) {
+				Log.LogError ($"Unable to find the clang compiler directory. Is NDK installed?");
```
jonpryor, line 645: `LogError()` should be `LogCodedError()`, along with corresponding docs in `Documentation/guides/messages`.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -544,25 +549,119 @@ void AddRuntimeLibraries (ZipArchiveEx apk, string [] supportedAbis)
...
+				return;
+			}
+
+			foreach (string abi in supportedAbis) {
+				string clangAbi = MonoAndroidHelper.MapAndroidAbiToClang (abi);
+				if (String.IsNullOrEmpty (clangAbi)) {
+					Log.LogError ($"Unable to map Android ABI {abi} to clang ABI");
```
jonpryor, line 652: `LogError()` should be `LogCodedError()`, along with corresponding docs in `Documentation/guides/messages`.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -544,25 +549,119 @@ void AddRuntimeLibraries (ZipArchiveEx apk, string [] supportedAbis)
...
+					return;
+				}
+
+				string sanitizerLib = $"libclang_rt.{sanitizerName}-{clangAbi}-android.so";
+				string sanitizerLibPath = Path.Combine (clangDir, sanitizerLib);
+				if (!File.Exists (sanitizerLibPath)) {
+					Log.LogError ($"Unable to find sanitizer runtime for the {CheckedBuild} checker at {sanitizerLibPath}");
```
jonpryor, line 659: `LogError()` should be `LogCodedError()`, along with corresponding docs in `Documentation/guides/messages`.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1452,7 +1454,8 @@ because xbuild doesn't support framework reference assemblies.
       FrameworkDirectories="$(_XATargetFrameworkDirectories);$(_XATargetFrameworkDirectories)Facades"
       AcwMapFile="$(_AcwMapFile)"
       SupportedAbis="@(_BuildTargetAbis)"
-      SkipJniAddNativeMethodRegistrationAttributeScan="$(_SkipJniAddNativeMethodRegistrationAttributeScan)">
+      SkipJniAddNativeMethodRegistrationAttributeScan="$(_SkipJniAddNativeMethodRegistrationAttributeScan)"
+      CheckedBuild="$(AndroidCheckedBuild)">
```
jonpryor, line 1458: `$(AndroidCheckedBuild)` should be documented within `Documentation/guides/building-apps/build-properties.md`.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -2151,7 +2161,9 @@ because xbuild doesn't support framework reference assemblies.
       LibraryProjectJars="@(ExtractedJarImports)"
       TlsProvider="$(AndroidTlsProvider)"
       UncompressedFileExtensions="$(AndroidStoreUncompressedFileExtensions)"
-      ProjectFullPath="$(MSBuildProjectFullPath)">
+      ProjectFullPath="$(MSBuildProjectFullPath)"
+      IncludeWrapSh="$(AndroidIncludeWrapSh)"
```
jonpryor, line 2165: Should `$(AndroidIncludeWrapSh)` be documented in `Documentation/guides/building-apps/build-properties.md`?

src/monodroid/CMakeLists.txt
```diff
@@ -116,6 +118,33 @@ set(TEST_COMPILER_ARGS_CPP ${XA_COMPILER_FLAGS}
   Wuninitialized
   )
 
+if(ENABLE_NDK)
+  if (ENABLE_CLANG_ASAN OR ENABLE_CLANG_UBSAN)
+	set(TEST_COMPILER_ARGS_CPP
```
jonpryor, line 123: Please update `.editorconfig` so that `CMakeLists.txt` always expands tabs, then re-save this file so that tabs are spaces.

src/monodroid/CMakeLists.txt
```diff
@@ -116,6 +118,33 @@ set(TEST_COMPILER_ARGS_CPP ${XA_COMPILER_FLAGS}
...
+	  ${TEST_COMPILER_ARGS_CPP}
+	  fno-omit-frame-pointer
+	  fno-optimize-sibling-calls
+	  )
```
jonpryor, line 127: I *think* this should align with the `s` in `set(`?

src/monodroid/CMakeLists.txt
```diff
@@ -357,22 +400,27 @@ if(NOT WIN32 AND NOT MINGW AND CMAKE_BUILD_TYPE STREQUAL Debug)
...
+    ${XAMARIN_DEBUG_APP_HELPER}
     PUBLIC -DDEBUG_APP_HELPER
     )
   if(APPLE)
     set_target_properties(
-      xamarin-debug-app-helper
+      ${XAMARIN_DEBUG_APP_HELPER}
       PROPERTIES
       OSX_ARCHITECTURES "${XA_OSX_ARCHITECTURES}"
-    )
```
jonpryor, line 370: This whitespace shouldn't change.

src/monodroid/CMakeLists.txt
```diff
@@ -357,22 +400,27 @@ if(NOT WIN32 AND NOT MINGW AND CMAKE_BUILD_TYPE STREQUAL Debug)
...
+	  )
   endif()
+  target_compile_options(
+	${XAMARIN_DEBUG_APP_HELPER}
+	PRIVATE ${CHECKED_COMPILER_FLAGS}
+	)
```
jonpryor, line 419: This `)` should vertically align with the `t` in `target_compile_options(`.

src/monodroid/CMakeLists.txt
```diff
@@ -403,8 +452,13 @@ elseif(NOT ANDROID AND NOT MINGW AND NOT WIN32)
   set(LINK_LIBS "-pthread ${LINK_LIBS} -ldl")
 endif()
 
+target_compile_options(
+  ${MONO_ANDROID_LIB}
+  PRIVATE ${CHECKED_COMPILER_FLAGS}
+  )
```
jonpryor, line 458: This `)` should align with the `t` in `target_compile_options(`.

src/monodroid/jni/monodroid-glue.cc
jonpryor, line 655: This should happen before the `#else` on line 737.

Additionally, the declaration of `options` on line 646 should become:

```cpp
RuntimeOptions options{};
```

src/monodroid/monodroid.targets
```diff
@@ -87,16 +127,36 @@
...
-      Outputs="@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android.debug.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android.release.so');@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Debug\libxamarin-app.so');@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Release\libxamarin-app.so')">
+      Outputs="@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android.debug.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android-asan.debug.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android-ubsan.debug.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android.release.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android-asan.release.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android-ubsan.release.so');@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Debug\libxamarin-app.so');@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Release\libxamarin-app.so')">
```
jonpryor, line 130: These `Outputs` don't match what's in `create-installers.targets`, which includes:

```
libmono-android-checked+asan.debug.so
```

The new libs need a `checked+` inserted.

src/monodroid/monodroid.targets
```diff
@@ -87,16 +127,36 @@
...
     <Exec
         Command="$(NinjaPath) -v"
         WorkingDirectory="$(IntermediateOutputPath)%(AndroidSupportedTargetJitAbi.Identity)-Debug"
     />
 
+    <Exec
+        Command="$(NinjaPath) -v"
+        WorkingDirectory="$(IntermediateOutputPath)%(AndroidSupportedTargetJitAbi.Identity)-asan-Debug"
```
jonpryor, line 138: I don't understand how this works; `$(NinjaPath)` (above) isn't given any command-line arguments, so how does the `WorkingDirectory` value control which native lib options are used?

tests/Xamarin.Forms-Performance-Integration/Droid/Xamarin.Forms.Performance.Integration.Droid.csproj
```diff
@@ -79,6 +84,33 @@
     <AndroidResource Include="Resources\layout\Toolbar.axml" />
     <AndroidResource Include="Resources\values\styles.xml" />
   </ItemGroup>
+  <ItemGroup>
+    <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
+        <Link>lib\arm64-v8a\wrap.sh</Link>
```
jonpryor, line 89: This should be indented only 2 spaces from `<AndroidNativeLibrary/>`, not 4.  4 is for attributes, not elements.

jonpryor (changes requested):

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: The answer is on the line of your previous comment :)

```xml
<Link>lib\arm64-v8a\wrap.sh</Link>
```

This will be passed to the task as-is and since our convention is to use Windows path separator character in MSBuild files, this is entirely expected.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: Even if this warning won't show up on the user's machines (since we're not going to ship the checked builds)?

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: Ditto

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
grendello: It was there, but we decided we're not shipping the checked builds - does it make sense to document the property?

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: This message will be logged only for checked builds, that is not on the end user machine.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: This message will be logged only for checked builds, that is not on the end user machine.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: This message will be logged only for checked builds, that is not on the end user machine.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: This message will be logged only for checked builds, that is not on the end user machine.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: This message will be logged only for checked builds, that is not on the end user machine.

src/monodroid/CMakeLists.txt
grendello: No, that would be a wall of text - the entries are inside the list, they should be indented in relation to `set`

src/monodroid/monodroid.targets
grendello: The `{rules,build}.ninja` files are generated in the `WorkingDirectory` directory and since it's the cwd when `ninja` runs, it reads the correct one for the given build.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
jonathanpeppers: Can we make it?
```xml
<_AndroidCheckedBuild Condition=" '$(_AndroidCheckedBuild)' == '' ">false</_AndroidCheckedBuild>
```

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -88,6 +88,10 @@ public class BuildApk : AndroidTask
 		//[Required]
 		public bool EnableCompression { get; set; }
 
+		public bool IncludeWrapSh { get; set; }
+
+		public string CheckedBuild { get; set; }
```
jonathanpeppers, line 93: Should this be `bool`, or can it have more than two values?

jonathanpeppers: Nevermind, I guess this can be `asan` or `ubsan`.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -198,6 +198,8 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidGenerateLayoutBindings Condition=" '$(AndroidGenerateLayoutBindings)' == '' ">False</AndroidGenerateLayoutBindings>
 	<AndroidFragmentType Condition=" '$(AndroidFragmentType)' == '' ">Android.App.Fragment</AndroidFragmentType>
 	<AndroidEnableAssemblyCompression Condition=" '$(AndroidEnableAssemblyCompression)' == '' ">True</AndroidEnableAssemblyCompression>
+	<AndroidIncludeWrapSh Condition=" '$(AndroidIncludeWrapSh)' == '' ">False</AndroidIncludeWrapSh>
+	<AndroidCheckedBuild Condition=" '$(AndroidCheckedBuild)' == '' "></AndroidCheckedBuild>
```
jonathanpeppers, line 202: I think you can remove this line if it defaults to empty.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: Yes, possibly other sanitizer names in the future

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
grendello: @jonpryor thought it was good to put it in when I asked a few weeks back. @jonpryor, thoughts?

src/Mono.Android/Test/Mono.Android-Tests.csproj
```diff
@@ -125,4 +130,31 @@
   <ItemGroup>
     <Folder Include="Localization\" />
   </ItemGroup>
+  <ItemGroup>
+    <AndroidNativeLibrary Include="$(_ASANScript)" Condition=" '$(UseASAN)' != '' ">
```
dellis1972, line 134: This is something users can use correct?
If so would it be possible for us to include a target which will auto add all this stuff based on a MSbuild boolean flag? like the `AndroidIncludeWrapSh` ? or `AndroidNativeCodeCheck`.

Just wondering if its worth the effort to make this easier for users to use. So they just enable one setting and we do the rest.

grendello: The trouble is that `wrap.sh` can do anything - it can be used not only for sanitizers, but also for debugging, for setting Mono runtime environment variables, logging etc etc. We will not be shipping the checkers/sanitizers and so we won't be shipping the ASAN and UBSAN `wrap.sh` scripts - there's nothing to add automatically. 
The idea behind this mechanism here is that we want to be able to have the `AndroidNativeLibrary` items in the project but we don't want to package them into APK for normal builds. To include them users need to set `AndroidIncludeWrapSh`, probably from command line, for one-off builds when they need `wrap.sh` help for something.
Once I finish the gdb support we will include the gdb `wrap.sh` automatically the way you suggest, but until then there's nothing to include.

Documentation/guides/building-apps/build-properties.md
```diff
@@ -549,6 +549,39 @@ The most common values for this property are:
...
+
+The script is added to the project using the
+[`@(AndroidNativeLibrary)`](./build-items.md#androidnativelibrary)
+build action, because it is placed in the same directory where
+architecture-specific native libraries, and must be named `wrap.sh`.
+The name can be specified using the `Link` metadata of the
+`AndroidNativeLibrary`:
+
+```xml
+<AndroidNativeLibrary Include="/path/to/my/arm64-wrap.sh" Condition=" '$(UseASAN)' != '' ">
```
jonpryor, line 570: The `Condition` here will likely add more questions than answers; I think it should be removed.

jonpryor, line 570: "Semantic" question: for native libs, the ABI can be "deduced" via "path sniffing".  Is the same possible with `wrap.sh` as well, e.g. given:

```xml
<AndroidNativeLibrary Include="path/to/lib/arm64-v8a/wrap.sh" />
```

will the ABI be inferred as `arm64-v8a`, thus allowing us to *skip* the `%(AndroidNativeLibrary.Link)` metadata?

If not, can this be done?

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
jonpryor: mono's `msbuild` "normalizes" paths.  `dotnet msbuild`, on the other hand, does not. :-(

Consider `x.targets`:

```xml
<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <ItemGroup>
    <Foo Include="item1">
      <Link>path\to\whatever</Link>
    </Foo>
  </ItemGroup>
  <Target Name="Foo">
    <Message Text="%(Foo.Link)" />
  </Target>
</Project>
```

With Mono's `msbuild`, it's normalized:

```shell
$ msbuild /nologo /v:m x.targets
  path/to/whatever
```

It's *not* normalized with `dotnet msbuild`:

```
% dotnet msbuild /nologo /v:m x.targets
  path\to\whatever
```

Documentation/guides/building-apps/build-properties.md
grendello: Yes, it will properly "sniff" the architecture, but since you **must** name the script `wrap.sh` you won't be able to store two different scripts in the same architecture's directory - this is when `Link` is needed (as it is the case in this PR)

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: I'm not sure I want to rely on this behavior... There's no guarantee msbuild on *nix won't revert to the dotnet msbuild way, plus, what happens when we switch to net5/6? I'd rather be on the safe side and keep this code

Documentation/guides/building-apps/build-properties.md
```diff
@@ -549,6 +549,50 @@ The most common values for this property are:
...
+A boolean value which indicates whether the Android wrapper script
+([`wrap.sh`](https://developer.android.com/ndk/guides/wrap-script))
+should be packaged into the APK. The property defaults to `false`
+since the wrapper script may significantly influence the way the
+application starts up and works and the script should be included 
+only when necessary e.g. for debugging or otherwise changing the
+application startup/runtime behavior.
+
+The script is added to the project using the
+[`@(AndroidNativeLibrary)`](./build-items.md#androidnativelibrary)
```
jonpryor, line 563: I fear this needs to be a "fuller" url, e.g. `(~/android/deploy-test/building-apps/build-items.md#androidnativelibrary)`, lest it cause complications when we next sync with MicrosoftDocs.

jonpryor: This does *not* need to be a full URL; relative URLs are fine. https://teams.microsoft.com/l/message/19:446e3cd63c894194826e96b9d1b3032c@thread.skype/1602579498335?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=42121604-a67e-4f9a-9f3f-57c3c6d7ccd3&parentMessageId=1602541101300&teamName=Xamarin%20Platform%20Engineering&channelName=Docs&createdTime=1602579498335

grendello: I changed it to use the same path as the other references in the document

PR #5095, opened 2020-09-07
[Xamarin.Android.Build.Tasks] Dont copy assets for Library Projects.

dellis1972:
We currently copy all the assets for all Library Projects
into one main directory. This can slow down build times
if the assets are large. So what we should do is use those
assets in place.

This will also allow us to consume .nuget assets in place as
well sometime in the future.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
@@ -14,7 +14,7 @@ using System.Collections.Generic;
 using Xamarin.Android.Tools;
 
 namespace Xamarin.Android.Tasks {
-	
+
 	//aapt2 link -o resources.apk.bk --manifest Foo.xml --java . --custom-package com.infinitespace_studios.blankforms -R foo2 -v --auto-add-overlay
 	public class Aapt2Link : Aapt2 {
 		static Regex exraArgSplitRegEx = new Regex (@"[\""].+?[\""]|[\''].+?[\'']|[^ ]+", RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.Multiline);
@@ -44,6 +44,8 @@ namespace Xamarin.Android.Tasks {
 		public string AndroidComponentResgenFlagFile { get; set; }
 
 		public string AssetsDirectory { get; set; }
+
+		public ITaskItem [] AdditionalAndroidAssetPaths { get; set; }
 
 		public string ExtraPackages { get; set; }
 
@@ -175,7 +177,7 @@ namespace Xamarin.Android.Tasks {
 				cmd.Add ("--custom-package");
 				cmd.Add (PackageName.ToLowerInvariant ());
 			}
-			
+
 			if (AdditionalResourceArchives != null) {
 				for (int i = AdditionalResourceArchives.Length - 1; i >= 0; i--) {
 					var flata = Path.Combine (WorkingDirectory, AdditionalResourceArchives [i].ItemSpec);
@@ -226,7 +228,7 @@ namespace Xamarin.Android.Tasks {
 					}
 				}
 			}
-			
+
 			cmd.Add ("--auto-add-overlay");
 
 			if (!string.IsNullOrWhiteSpace (UncompressedFileExtensions))
@@ -263,6 +265,8 @@ namespace Xamarin.Android.Tasks {
 				}
 			}
 
+			// When adding Assets the first item found takes precedence.
+			// So we need to add the applicaiton Assets first.
 			if (!string.IsNullOrWhiteSpace (AssetsDirectory)) {
 				var assetDir = AssetsDirectory.TrimEnd ('\\');
 				if (!Path.IsPathRooted (assetDir))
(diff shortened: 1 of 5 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -98,11 +98,11 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 <!--
 *******************************************
   Extensibility hook that allows VS to
-  provide extra behavior without modifying 
+  provide extra behavior without modifying
```
dellis1972, line 101: ugh. vscode applied a ton of white space changes :(

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1643,10 +1643,16 @@ because xbuild doesn't support framework reference assemblies.
       LibraryProjectImportsDirectoryName="$(_LibraryProjectImportsDirectoryName)">
     <Output TaskParameter="ExtraPackages" PropertyName="AaptExtraPackages" />
   </GetExtraPackages>
-  
+
```
dellis1972, line 1646: This is the only real change.

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
```diff
@@ -263,6 +265,16 @@ public async override System.Threading.Tasks.Task RunTaskAsync ()
...
 
+			if (AdditionalAndroidAssetPaths != null) {
+				for (int i = AdditionalAndroidAssetPaths.Length - 1; i >= 0; i--) {
+					var assetDir = AdditionalAndroidAssetPaths [i].ItemSpec.TrimEnd ('\\');
+					if (!string.IsNullOrWhiteSpace (assetDir) && Directory.Exists (assetDir)) {
+						cmd.Add ("-A");
+						cmd.Add (GetFullPath (assetDir));
+					}
+				}
+			}
```
jonathanpeppers, line 276: Would `<ResolveLibraryProjectImports/>` also need to skip extracting `assets` directories?

dellis1972: In this case, we are still extracting the directories to the `lp` directory, but we skip the copying over to the global assets directory and use them in place. This is a first step to allow us to support using assets/res/etc from the project reference/nuget location.

dellis1972: We already had a concept of `AdditionalResourcePaths` but did't have the same for assets. I figured this would be an easy win, especially if you have large assets (like movie files).

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
```diff
@@ -263,6 +265,16 @@ public async override System.Threading.Tasks.Task RunTaskAsync ()
 				}
 			}
 
+			if (AdditionalAndroidAssetPaths != null) {
+				for (int i = AdditionalAndroidAssetPaths.Length - 1; i >= 0; i--) {
```
jonpryor, line 269: Why go through `@(LibraryAssetDirectories)` *backwards*? Do later `-A PATH` options "take precedence" over earlier ones?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
jonpryor: If this is the only "real" change, could we revert the file and apply *just* this change?

Additionally, could the `Condition="…"` value be the *first* attribute instead of the last?

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
dellis1972: We go through backwards like we do for other resources. I assume its like the `-R` option where ` The last conflicting resource given takes precedence.`. I will have to test it.

dellis1972: First Item takes precedence it seems. So I switches the loop order and also made sure we add the main app assets first.

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Link.cs
```diff
@@ -272,6 +276,17 @@ public async override System.Threading.Tasks.Task RunTaskAsync ()
...
 				}
 			}
+
+			if (AdditionalAndroidAssetPaths != null) {
+				for (int i = 0; i < AdditionalAndroidAssetPaths.Length; i++) {
+					var assetDir = AdditionalAndroidAssetPaths [i].ItemSpec.TrimEnd ('\\');
+					if (!string.IsNullOrWhiteSpace (assetDir) && Directory.Exists (assetDir)) {
+						cmd.Add ("-A");
+						cmd.Add (GetFullPath (assetDir));
+					}
```
jonathanpeppers, line 286: As per the bug we were seeing with `WorkingDirectory` and `<Aapt2Link/>`, I think this needs to be:
```suggestion
					var assetDir = GetFullPath (AdditionalAndroidAssetPaths [i].ItemSpec.TrimEnd ('\\'));
					if (Directory.Exists (assetDir)) {
						cmd.Add ("-A");
						cmd.Add (assetDir);
					} else {
						LogDebugMessage ("asset directory did not exist: " + assetDir);
					}
```

dellis1972: done

jonathanpeppers (approved):

PR #5098, opened 2020-09-09
Add DevDiv required Roslyn analyzers to all projects and fix errors.

jpobst:
As part of building secure software, Microsoft DevDiv has a set of [Roslyn anaylzers][0] dealing with security that should be run on every managed assembly.

Adds these analyzers and fix any errors they introduce.

## Running Analyzers

In order to run the Roslyn analyzers, the NuGet package [`Microsoft.CodeAnalysis.FxCopAnalyzers`][1] must be added to each
project.  Rather than do this manually now, and for each new project in the future, we instead add this to the `Directory.Build.props` file, which automatically adds it to all projects.

By default, adding the NuGet package runs all included analyzers at each analyzer's default severity level.  At this time, we are only
concerned with the prescribed security set, so we use `.editorconfig` to set those analyzers as `error`, and all other analyzers as `none`.

Projects that wish to opt out of running the analyzers can set `<DisableRoslynAnalyzers>True</DisableRoslynAnalyzers>`.

Several security errors were surfaced in `NRefactory`. Because the code is upstream, and we only use this for running tests, we have chosen to not run the analyzers on `NRefactory`.  This code is not shipped to users.

## Fixing Errors

A few errors were suppressed due to being a false positive or using MD5 to implement industry standard RFC2617.  Justifications are included in the suppression attribute.

The only other errors surfaced by these analyzers is [CA3075: Insecure DTD Processing][2].  These were fixed by using `new XmlReaderSettings { XmlResolver = null }`, which will not attempt to resolve and download any DTD files.

Link to the build where it ran: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4117694.

[0]: https://github.com/dotnet/roslyn-analyzers
[1]: https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/
[2]: https://docs.microsoft.com/en-us/visualstudio/code-quality/ca3075?view=vs-2019


grendello (approved):

src-ThirdParty/NUnitLite/Constraints/XmlSerializableConstraint.cs
```diff
@@ -58,7 +59,8 @@ public override bool Matches(object actual)
 
                 stream.Seek(0, SeekOrigin.Begin);
 
-                object value = serializer.Deserialize(stream);
+                var reader = XmlReader.Create (stream, new XmlReaderSettings { XmlResolver = null });
+                object value = serializer.Deserialize(reader);
```
jonathanpeppers, line 63: I think `XmlReader` needs to be in a `using` block.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/EmbeddedDSOTests.cs
```diff
@@ -161,7 +161,8 @@ public void AndroidManifestHasFlag ()
 			string manifest = Path.Combine (testProjectPath, "obj", XABuildPaths.Configuration, "android", "manifest", "AndroidManifest.xml");
 			Assert.That (new FileInfo (manifest), Does.Exist, $"File {manifest} should exist");
 
-			var doc = new XPathDocument (manifest);
+			var reader = XmlReader.Create (manifest, new XmlReaderSettings { XmlResolver = null });
+			var doc = new XPathDocument (reader);
```
jonathanpeppers, line 165: Same here, `using` needed.

tools/xabuild/XABuild.cs
```diff
@@ -113,8 +113,9 @@ static StreamWriter OpenSysLinksFile (XABuildPaths paths)
 
 		static void CreateConfig (XABuildPaths paths)
 		{
-			var xml = new XmlDocument ();
-			xml.Load (paths.MSBuildConfig);
+			var xml = new XmlDocument { XmlResolver = null };
+			var reader = XmlReader.Create (paths.MSBuildConfig, new XmlReaderSettings { XmlResolver = null });
```
jonathanpeppers, line 117: `using`

src-ThirdParty/NUnitLite/Constraints/XmlSerializableConstraint.cs
jpobst: Good call, updated these 3 areas.

jonathanpeppers (approved):

dellis1972 (approved):

jonpryor:
```markdown
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1139592

As part of building secure software, Microsoft DevDiv has a set of
[Roslyn anaylzers][0] dealing with security that should be run on every
managed assembly.

Add these analyzers and fix any errors they introduce.


~~ Running Analyzers ~~

In order to run the Roslyn analyzers, the NuGet package
[`Microsoft.CodeAnalysis.FxCopAnalyzers`][1] must be added to each
project.  Rather than do this manually now, and for each new project
in the future, we instead add this to the `Directory.Build.props` file,
which automatically adds it to all projects.

By default, adding the NuGet package runs all included analyzers at
each analyzer's default severity level.  At this time, we are only
concerned with the prescribed security set, so we use
`.editorconfig` to set those analyzers as `error`, and all other
analyzers as `none`.

By default, *no* analyzers are run, as we found that enabling them
added ~90 minutes to our build times.  To enable the analyzers, set
the `$(EnableRoslynAnalyzers)` MSBuild property to True:

	msbuild /p:EnableRoslynAnalyzers=True Xamarin.Android.sln

We will create a separate scheduled build job to run the analyzers
on a regular basis.


~~ Fixing Errors ~~

A few errors were suppressed due to being a false positive, or for
using MD5 to implement industry standard RFC2617.  Justifications are
included in the [`SuppressMessage`][2] attribute.

The only other errors surfaced by these analyzers is
[CA3075: Insecure DTD Processing][3].  These were fixed by using
`new XmlReaderSettings { XmlResolver = null }`, which will not attempt
to resolve and download any DTD files.

[0]: https://github.com/dotnet/roslyn-analyzers
[1]: https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/
[2]: https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.suppressmessageattribute
[3]: https://docs.microsoft.com/visualstudio/code-quality/ca3075
```

PR #5099, opened 2020-09-09
[spec] for @(AndroidLibrary) and avoiding EmbeddedResource

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/issues/2441

See `OneDotNetEmbeddedResources.md` for the new spec for
`@(AndroidLibrary)` and removal of `EmbeddedResource` usage in
Xamarin.Android class libraries.

Future proposals can be commited to `Documentation/proposals`. When
implemented, the proposal can be reworked and moved to
`Documentation/guides` if it is still useful

Easiest to read [here](https://github.com/jonathanpeppers/xamarin-android/blob/spec-embeddedresource/Documentation/proposals/OneDotNetEmbeddedResources.md).

### DIFF
--- a/Documentation/proposals/OneDotNetEmbeddedResources.md
+++ b/Documentation/proposals/OneDotNetEmbeddedResources.md
@@ -214,7 +214,7 @@ potentially any other library that depends on AndroidX.
 
 ### Next Phase
 
-Things that could happen by .NET 6 RTM:
+Things that could happen sometime after .NET 6 Preview 1:
 
 * Support for referencing Android Studio projects in order to build
   them and consume the resulting `.aar` file.
--- a/Documentation/proposals/OneDotNetEmbeddedResources.md
+++ b/Documentation/proposals/OneDotNetEmbeddedResources.md
@@ -1,0 +1,247 @@
+# [proposal] Embedded Resources and `.nupkg` Files
+
+Traditionally, a Xamarin.Android class library can contain many types
+of Android-specific files:
+
+* `@(AndroidAsset)` files in `Assets/`
+* `@(AndroidEnvironment)` text files
+* `@(AndroidJavaLibrary)` `.jar` files
+* `@(AndroidResource)` files in `Resources/`
+* `@(EmbeddedNativeLibrary)` files in `lib/[arch]/*.so`
+
+These are packaged in different ways as `EmbeddedResource` files in
+the output assembly:
+
+* `@(AndroidEnvironment)`: stored as
+  `__AndroidEnvironment__%(Filename)%(Extension)` embedded resources
+* `@(AndroidAsset)`, `@(AndroidResource)`: merged into a single
+  `__AndroidLibraryProjects__.zip` embedded resource.
+  `@(AndroidAsset)` is copied into the `assets` path, and
+  `@(AndroidResource)` is copied into the `res` path.
+* `@(EmbeddedNativeLibrary)`: merged into a single
+  `__AndroidNativeLibraries__.zip` embedded resource.
+* `@(AndroidJavaLibrary)`: stored separately as embedded resources.
+
+The problem with this approach, is we have to inspect every assembly
+at build time and extract these files to a directory. Because we have
+a custom format that Android does not understand, we can't leave the
+files as-is. Android command-line tooling like `aapt2` works with
+files and directories on disk.
+
+Java binding projects have a slight variation:
+
+* `@(LibraryProjectZip)` can be `.aar` or `.zip` files and is
+  repackaged into `__AndroidLibraryProjects__.zip`. You can only have
+  one `@(LibraryProjectZip)` per project.
+* No support for `@(AndroidResource)` or `@(AndroidAsset)`.
+* `@(EmbeddedJar)` and `@(EmbeddedReferenceJar)` are packaged directly
+  as `EmbeddedResource` files.
+
+## New Approach in .NET 6
+
+We want to consider an approach that is NuGet-centric. When the above
+implementation was written, NuGet was just getting started. Developers
+still had the pattern of copying .NET assemblies from `bin` and
+committing it to their source control of choice. The single file
+approach worked well for this scenario.
+
+If we look at the general structure of a [NuGet package][nuget]:
+
+    lib/
+        net6.0-android29/Foo.dll
+        net6.0-android30/Foo.dll
+    # Optional reference assemblies
+    ref/
+        net6.0-android29/Foo.dll
+        net6.0-android30/Foo.dll
+    # Optional native libraries
+    runtimes/
+        android-arm/native/libFoo.so
+        android-arm64/native/libFoo.so
+        android-x86/native/libFoo.so
+        android-x64/native/libFoo.so
+
+There is not a great place where all Android file types would fit
+following this pattern.
+
+In Android Studio, Android libraries are packaged as [`.aar`][aar]
+files. A Xamarin.Android library, `Foo.csproj`, could also generate an
+[`.aar`][aar] file in its `$(OutputPath)`:
+
+    Foo.aar
+        classes.jar
+        res/
+        assets/
+        libs/*.jar
+        jni/[arch]/*.so
+
+`@(AndroidEnvironment)` files are specific to Xamarin.Android, so we
+can make one addition to the file format:
+
+    Foo.aar
+        .netenv/MyEnvironment.txt
+
+The name `.netenv` would be unlikely to collide with anything Google
+creates in the `.aar` file format in the future. It should also be
+completely ignored by Android tooling.
+
+So the output of `Foo.csproj` would look like:
+
+    bin/Debug/$(TargetFramework)/
+        Foo.dll
+        Foo.aar
+    bin/Release/$(TargetFramework)/
+        Foo.dll
+        Foo.aar
+
+If you ran the `Pack` target, you would get a `.nupkg` file with:
+
+    lib/
+        net6.0-android29/Foo.dll
+        net6.0-android29/Foo.aar
+        net6.0-android30/Foo.dll
+        net6.0-android30/Foo.aar
+
+When consuming the `.nupkg` files, Xamarin.Android will still have to
+extract [`.aar`][aar] files on disk for command-line tooling such as
+`javac`, `d8/r8` and `manifestmerger`. If users want to copy around
+loose build output and consume it, there will only be 1 additional
+[`.aar`][aar] file they will need to copy.
+
+### Java/Kotlin Dependencies
+
+A `Foo.csproj` might include `bar.aar` and `baz.jar` files that are
+Java/Kotlin dependencies.
+
+`.aar` files should sit alongside the .NET assembly:
+
+    lib
+        net6.0-android29/Foo.dll
+        net6.0-android29/Foo.aar
+        net6.0-android29/bar.aar
+        net6.0-android30/Foo.dll
+        net6.0-android30/Foo.aar
+        net6.0-android30/bar.aar
+
+The `baz.jar` file would be included in the above `Foo.aar` file at:
+
+    Foo.aar
+        classes.jar
+        libs/baz.jar
+
+### Native Libraries
+
+Since both `.aar` and `.nupkg` files support native libraries,
+Xamarin.Android should support consuming native libraries from both
+locations. A Xamarin.Android class library, `Foo.csproj` will place
+native libraries in the `Foo.aar` file by default.
+
+Collisions encountered on the same native library should be ignored.
+Applications with large numbers of NuGet packages could have a 
+duplicate native library and everything could still work at runtime.
+
+### MSBuild Item Groups
+
+Right now we have a confusing collection of MSBuild item groups:
+
+* `@(InputJar)` - `.jar` file to create a Java binding for, but not
+  embed.
+* `@(EmbeddedJar)` - `.jar` file to create a Java binding and embed.
+* `@(ReferenceJar)` - `.jar` file to reference for the binding
+  process.
+* `@(EmbeddedReferenceJar)` - `.jar` file to reference and embed, but
+  not bind.
+* `@(LibraryProjectZip)` - `.aar` or `.zip` file to bind and embed.
+  Only one can be used in a project!
+* `@(AndroidAarLibrary)` - `.aar` file to include in a application
+  project.
+* `@(AndroidJavaLibrary)` - `.jar` file to include in a application
+  project.
+* `@(EmbeddedNativeLibrary)` - `.so` file to embed in a class library
+  project or include in an application project.
+* `@(AndroidNativeLibrary)` - `.so` file to include in an application
+  project.
+
+Let's simplify this, we could support all of the above with a new
+`@(AndroidLibrary)` item group:
+
+```xml
+<!-- Include and bind -->
+<AndroidLibrary Include="foo.aar" Bind="true" />
+<!-- Just include, do not bind -->
+<AndroidLibrary Include="bar.aar" />
+<AndroidLibrary Include="baz.jar" />
+<!--
+  Bind but do not include in NuGet package.
+  %(Pack) is built into NuGet MSBuild targets.
+-->
+<AndroidLibrary Include="bar.aar" Bind="true" Pack="false" />
+<!-- Native libraries need ABI directory -->
+<AndroidLibrary Include="armeabi-v7a\libfoo.so" />
+<AndroidLibrary Include="x86\libfoo.so" />
+```
+
+The new `@(AndroidLibrary)` item group will simply translate to the
+old ones for backwards compatibility. The extension of the file can be
+used to determine what kind of library each item is. `%(Bind)` will be
+`false` by default, and `%(Pack)` will be `true` by default. `%(Pack)`
+will not do anything in application projects.
+
+The deprecated item groups will no longer embed, but pack into
+`.nupkg` files instead:
+
+* `@(EmbeddedJar)`
+* `@(EmbeddedReferenceJar)`
+
+`@(LibraryProjectZip)` will not be supported at all.
+
+## Implementation
+
+Since there is a lot of work here, I've split up the what needs to be
+done into different phases.
+
+### "Must Have" for .NET 6 Preview 1
+
+Things we need for the first public release:
+
+* Item group support for `@(AndroidLibrary)` and `%(Bind)=true`.
+* Support for packing `.aar` files in `$(OutputPath)` for
+  Xamarin.Android class libraries.
+* Support for consuming `.nupkg` files that contain `.aar` files.
+
+These should enable us to compile AndroidX for `net6.0-android`. Then
+potentially any other library that depends on AndroidX.
+
+### Next Phase
+
+Things that could happen by .NET 6 RTM:
+
+* Support for referencing Android Studio projects in order to build
+  them and consume the resulting `.aar` file.
+
+* Support for including Maven dependencies. This might be inclusion of
+  a `build.gradle` file that would enable us to run `gradle` and
+  download dependencies from Maven repositories.
+
+* Consider not extracting everything in `.aar` files. We can probably
+  skip the `assets` and `res` directories and let `aapt2` consume
+  `.aar` files directly.
+
+* `<GenerateJavaStubs/>` performance
+  * `.aar` files contain `AndroidManifest.xml` files. We could add
+    support for developers to add their own manifest to Xamarin.Android
+    class libraries.
+  * We could also generate the `AndroidManifest.xml` so attributes
+    like `[assembly:UsesPermission]` are done ahead of time.
+  * We could generate Java stubs for every `Java.Lang.Object`
+  * If all of the above was completed, `<GenerateJavaStubs/>` could
+    skip over the assembly completely.
+
+* Build performance could be further improved by included various
+  other build-related cache files in the `.aar` file.
+
+It is possible some of these might get pushed out to a future .NET 7
+release.
+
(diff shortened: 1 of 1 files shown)


Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,268 @@
+# [spec] Embedded Resources and `.nupkg` Files
+
+Traditionally, a Xamarin.Android class library can contain many types
+of Android-specific files:
+
+* `@(AndroidAsset)` files in `Assets\`
+* `@(AndroidEnvironment)` text files
+* `@(AndroidJavaLibrary)` `.jar` files
+* `@(AndroidNativeLibrary)` files in `lib\[arch]\*.so`
```
dotMorten, line 9: Missing `EmbeddedNativeLibrary` used for class libraries

jonathanpeppers: I'll list `@(EmbeddedNativeLibrary)`.

We are aspiring to change all these to:
```xml
<!-- Include and bind -->
<AndroidLibrary Include="foo.aar" Bind="true" />
<!-- Just include, do not bind -->
<AndroidLibrary Include="bar.aar" />
<AndroidLibrary Include="bar.jar" />
<!-- Native libraries need ABI directory -->
<AndroidLibrary Include="armeabi-v7a\libfoo.so" />
<AndroidLibrary Include="x86\libfoo.so" />
```

dotMorten: I like that change - much simpler to have a single build action for these.

Redth: Should this support an attribute to explicitly specify ABI for .so's?

jonathanpeppers: > an attribute to explicitly specify ABI

I thought about this, but maybe it should just be the directory name?

How many times would people need to do something like this:
```
<AndroidLibrary Include="foo\libFoo.so" Abi="armeabi-v7a" />
<AndroidLibrary Include="bar\libFoo.so" Abi="x86" />
```

Native libraries will likely all have the same file name, too, or `[DllImport]` gets complicated.

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+* `@(AndroidEnvironment)` text files
+* `@(AndroidJavaLibrary)` `.jar` files
+* `@(AndroidResource)` files in `Resources\`
+* `@(EmbeddedNativeLibrary)` files in `lib\[arch]\*.so`
+
+These are packaged in different ways as `EmbeddedResource` files in
+the output assembly:
+
+* `__AndroidEnvironment__[filename]`
+  * `@(AndroidEnvironment)` files as-is
```
jonpryor, line 16: I think it would be clearer if these were "reversed":

```markdown
* `@(AndroidEnvironment)`: stored as `__AndroidEnvironment__%(Filename)%(Extension)` embedded resources
* `@(AndroidAsset)`, `@(AndroidResource)`: merged into a single `__AndroidLibraryProjects__.zip` embedded resource.  `@(AndroidAsset)` is copied into the `assets` path, and `@(AndroidResource)` is copied into the `res` path.
* `@(EmbeddedNativeLibrary)`: merged into a single `__AndroidNativeLibraries__.zip` embedded resource.
* `@(AndroidJavaLibrary)`: stored separately as embedded resources.
…
```

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+
+We want to consider an approach that is NuGet-centric. When the above
+implementation was written, NuGet was just getting started. Developers
+still had the pattern of copying .NET assemblies from `bin` and
+committing it to their source control of choice. The single file
+approach worked well for this scenario.
+
+If we look at the general structure of a [NuGet package][nuget]:
+
+    lib
```
jonpryor, line 50: Please append `\`.

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+        net6.0-android29\Foo.dll
+        net6.0-android30\Foo.dll
+    # Optional reference assemblies
+    ref
```
jonpryor, line 54: Please append `\`.

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+        net6.0-android29\Foo.dll
+        net6.0-android30\Foo.dll
+    # Optional native libraries
+    runtimes
```
jonpryor, line 58: Please append `\`.

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+        res/
+        assets/
+        libs/*.jar
+        jni/[arch]/*.so
+
+`@(AndroidEnvironment)` files are specific to Xamarin.Android, so we
+can make one addition to the file format:
+
+    Foo.aar
+        env/MyEnvironment.txt
```
jonpryor, line 82: If this is a Xamarin.Android-specific extension, should we use a name that is less-likely to "potentially conflict" in the future?  Perhaps `.xa-env` or `_xa-env`?  (Which, *yes*, the former results in a hidden dir when extracting, but either one is highly unlikely to be used by Google in `.aar` files, ever.)

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+    bin/Debug/[targetframework]/
+        Foo.dll
+        Foo.aar
+    bin/Release/[targetframework]/
+        Foo.dll
+        Foo.aar
+
+If you ran the `Pack` target, you would get a `.nupkg` file with:
+
+    lib
```
jonpryor, line 95: Please append `\`.

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+
+`@(AndroidEnvironment)` files are specific to Xamarin.Android, so we
+can make one addition to the file format:
+
+    Foo.aar
+        env/MyEnvironment.txt
+
+So the output of `Foo.csproj` would look like:
+
+    bin/Debug/[targetframework]/
```
jonpryor, line 86: Kinda wondering what the convention is, if any, for `/` vs. `\` in paths.  I *thought* you were going for "`/` for `.zip` contents, `\` for on-disk contents," but here we have on-disk contents, and `/` is being used.

Should this instead use `\`?

Furthermore, instead of `[targetframework]` should this be `$(TargetFramework)`?

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+        Foo.dll
+        Foo.aar
+    bin/Release/[targetframework]/
+        Foo.dll
+        Foo.aar
+
+If you ran the `Pack` target, you would get a `.nupkg` file with:
+
+    lib
+        net6.0-android29\Foo.dll
```
jonpryor, line 96: …and here we have the same philosophical question, but from the `.nupkg` (`.zip`) perspective: these paths are *inside* the `.nupkg`.  Should they instead use `/`?

Documentation/guides/OneDotNetEmbeddedResources.md
jonathanpeppers: I think `.netenv/` might be nice, so it is hidden and says ".net".

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+
+The `baz.jar` file would be included in the above `Foo.aar` file at:
+
+    Foo.aar
+        classes.jar
+        libs/baz.jar
+
+### Native Libraries
+
+Since both `.aar` and `.nupkg` files support native libraries,
```
jonpryor, line 130: What happens in the result of "collision", e.g. a referenced `.aar` and `.nupkg` *both* provide `libConflictingName.so`? Which "wins"?  (Should this even be specified? We could leave it explicitly unspecified, perhaps?)

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+locations. A Xamarin.Android class library, `Foo.csproj` will place
+native libraries in the `Foo.aar` file by default.
+
+### Assembly-Level Attributes
+
+To mark each `.aar` file that is needed for an assembly, a new custom
+attribute is needed:
+
+```csharp
+[assembly: Android.Runtime.Dependency ("Foo.aar")]
```
jonpryor, line 141: Let's please use `Java.Interop` and not `Android.Runtime`, and it can live in `Java.Interop.dll`.

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+[assembly: Android.Runtime.Dependency ("bar.aar")]
+```
+
+We can emit a build error if any files are missing.
+
+We also need to emit an attribute to mark the assembly as using the
+new system:
+
+```csharp
+[assembly: Android.Runtime.Features (Android.Runtime.Feature.NativeDependencies)]
```
jonpryor, line 151: Let's please use `Java.Interop` and not `Android.Runtime`, and it can live in `Java.Interop.dll`.

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+during the `<ResolveLibraryProjectImports/>` MSBuild task. We would
+need to emit this for every Xamarin.Android class library, as new
+libraries might not contain any `[assembly: Dependency]` attributes.
+
+Usage of an enum would allow us to add new things later:
+
+```csharp
+public enum Feature
+{
+    NativeDependencies   = 0x1,
```
jonpryor, line 164: This feels error-prone.  If I add a new "type" of resource to my NuGet package, I must remember to add a new `Feature` flag, otherwise my new resource won't be included.  This just feels painful, and I'm not sure I understand the actual build time benefit, as we'd have to crack open an assembly to read the attribute *anyway*…

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,280 @@
...
+* Support for generating the above assembly-level attributes.
+
+These should enable us to compile AndroidX for `net6.0-android`. Then
+potentially any other library that depends on AndroidX.
+
+### Next Phase
+
+Things that could happen by .NET 6 RTM:
+
+* Support for referencing Android Studio projects in order to build
```
jonpryor, line 255: Punt punt punt punt!

Documentation/guides/OneDotNetEmbeddedResources.md
jonathanpeppers: Developers won't actually write these attributes, the build adds them automatically.

We could go the route of using a single attribute and a flags-based enum, or create lots of attributes?

In the future, the build will need to know if an assembly has no `EmbeddedResource` at all and skip looking at it in `<ResolveLibraryProjectImports/>`. Future changes could possibly skip an assembly for `<GenerateJavaStubs/>` if we pregenerate `AndroidManifest.xml` or java stubs.

Documentation/guides/OneDotNetEmbeddedResources.md
jonathanpeppers: I think we should emit a build error on duplicates, I don't think we do this at all now. I added a comment on this.

Whichever native library was added to the `.apk` _last_ appears like it would win.

Documentation/guides/OneDotNetEmbeddedResources.md
jonathanpeppers: I used `/` everywhere now to simplify things. The only place to use `\` would be in an MSBuild item group.

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,288 @@
...
+<!-- Native libraries need ABI directory -->
+<AndroidLibrary Include="armeabi-v7a\libfoo.so" />
+<AndroidLibrary Include="x86\libfoo.so" />
+```
+
+The new `@(AndroidLibrary)` item group will simply translate to the
+old ones for backwards compatibility. The extension of the file can be
+used to determine what kind of library each item is. `%(Bind)` will be
+`false` by default, and `%(Pack)` will be `true` by default. `%(Pack)`
+will not do anything in application projects.
```
jonathanpeppers, line 231: @Redth mentioned we probably need a way to exclude a `.jar` or `.aar` file in an application project.

There are cases, with like Amazon Kindle Fire SDKs or Symbol/Zebra EMDK, you might want C# bindings for the library, but the actual device has the library installed system wide. In this case, you should not include the code in the `classes.dex` in your app.

In that case would `Pack="false"` make sense, if you for some reason want to exclude a `.jar`, `.aar`, `.so` from your `.apk` file?

dellis1972:
@jonathanpeppers we can probably just extract the bits that need to be on disk rather than all of it. aapt2 compile can use res folders from with in zip file. I suspect that would be the largest folder anyway. So we can probably just extract assets manifest and native libs.

Documentation/guides/OneDotNetEmbeddedResources.md
jonpryor: I *don't* think this should be a build error.  "One can imagine" versioning scenarios in which NuGet A requires Nv1, while NuGet B requires Nv2, and -- through chicanery? -- you're able to get a final `.apk` which includes native libs from Nv2, and -- through testing -- know that it doesn't break NuGet A.

If this were an error, you couldn't even *build* such a thing in order to test it.

Do we want to *actively* prevent such a thing?  ¯\_(ツ)_/¯

Documentation/guides/OneDotNetEmbeddedResources.md
jonpryor: > the build adds them automatically

How would the build go about adding them?  Something like the `[assembly:TargetFrameworkAttribute]` done at build time?

jonpryor: > In the future, the build will need to know if an assembly has no `EmbeddedResource` at all and skip looking at it in `<ResolveLibraryProjectImports/>`.

Yes, but…. Could we (continue to) differentiate between "initial" builds and "incremental" builds?  In an incremental build scenario, there should be no need to re-process the assembly in the first place, so the custom attributes shouldn't help there.

That leaves the "initial" build scenario, *or* a "I just updated a referenced library project, and now I'm rebuilding the app project" scenario, and *either way* if we have to open the assembly *anyway* (just to check for custom attributes), I'm not sure how much "better" that would be vs. attempting to extract `@(EmbeddedResource)`s when an assembly doesn't have any.

(Some timing data might be interesting there…?)

"*Best*" performance improvements will be achieved by *not touching the assembly at all*.  I don't currently see how this enum helps prevent "touching the assembly".

jonathanpeppers: Yeah we could generate a `.cs` file like they do:
```
> cat '.\HelloAndroid\obj\Debug\net5.0-android\.NETCoreApp,Version=v5.0.AssemblyAttributes.cs'
// <autogenerated />
using System;
using System.Reflection;
[assembly: global::System.Runtime.Versioning.TargetFrameworkAttribute(".NETCoreApp,Version=v5.0", FrameworkDisplayName = "")]
```

Documentation/guides/OneDotNetEmbeddedResources.md
jonpryor: How are those files *currently* excluded?

We're going to need to have a migration path from current projects to .NET6+ projects, and the easier this migration can be done, the better.  Presumably if we can reuse the current exclusion mechanism, it'll simplify migration.

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,288 @@
...
+
+When consuming the `.nupkg` files, Xamarin.Android will still have to
+extract [`.aar`][aar] files on disk for command-line tooling such as
+`javac`, `d8/r8` and `manifestmerger`. If users want to copy around
+loose build output and consume it, there will only be 1 additional
+[`.aar`][aar] file they will need to copy.
+
+### Java/Kotlin Dependencies
+
+A `Foo.csproj` might include `bar.aar` and `baz.jar` files that are
```
jonpryor, line 113: If you're going to mention a `baz.jar` file, then the following example output should also mention it. :-)

Documentation/guides/OneDotNetEmbeddedResources.md
jonathanpeppers: Right now if you wanted to distribute a NuGet package of a binding like this, you would have to:
1. Use `@(InputJar)` for the binding.
2. Add a `.targets` file the NuGet package that includes the `.jar` file as `@(AndroidExternalJavaLibrary)` for application projects.

Maybe we just leave it as-is, I don't see a good way to do better at the moment. Checking for `%(Pack)`= `false` would only help when adding `.aar`/`.jar` files directly to an application project. It doesn't help NuGet packages, as you still need an MSBuild `.targets` file.

Documentation/guides/OneDotNetEmbeddedResources.md
jonathanpeppers: > attempting to extract @(EmbeddedResource)s when an assembly doesn't have any.

Maybe I need to see if this can work without any assembly-level attributes at all.

jonathanpeppers: I guess the question is, how do you know which `.aar` files go with an assembly reference? We would just look for `*.aar` in the same directory, maybe that is fine?

Documentation/guides/OneDotNetEmbeddedResources.md
jonathanpeppers: I think we should leave it as it would be today: you could have multiple `.dll` files with the same `@(EmbeddedNativeLibrary)` and we don't emit errors for this.

Whichever one happens to make it to the `.apk` is what you get.

jonathanpeppers:
> we can probably just extract the bits that need to be on disk rather than all of it. aapt2 compile can use res folders from with in zip file. I suspect that would be the largest folder anyway. So we can probably just extract assets manifest and native libs.

@dellis1972 I thought we could start with extracting everything (like today), and add performance gains like this later. I put this idea in the `Next Phase` section.

Documentation/guides/OneDotNetEmbeddedResources.md
jonpryor: > how do you know which `.aar` files go with an assembly reference?

Do we need to know that? I'm not sure why that would be immediately needed.  (One idea that came to mind was to restrict the `.aar`/`.jar` files provided to `javac` when building the Java Callable Wrappers, but it would likely be easier to just pass *all* the `.aar`/`.jar` files to all `javac` invocations *anyway*…)

Documentation/guides/OneDotNetEmbeddedResources.md
jonpryor: I don't think we can leave it "as-is" -- I think we want to *drop* the `@(InputJar)` name? -- but I think it would be useful to be *similar* to the current state of affairs, and "repurpose" `@(AndroidExternalJavaLibrary)` as a "synonym" for `@(Inputjar)` within projects.

Or perhaps go with `@(AndroidExternalLibrary)` to "mirror" `@(AndroidLibrary)`?.

Otherwise we have to go with `%(AndroidLibrary.Pack)`=False, which is also entirely "sensible" and "workable", but I'm less certain about *tooling support*.  We might need @joj to chime in on what would be easiest for the IDE to support.

Documentation/guides/OneDotNetEmbeddedResources.md
jonathanpeppers: I don't know if it is a security concern, but anyone could drop a `.jar` or `.aar` file next to a `.dll` file and it would get used throughout the build.

jonathanpeppers: This would probably also work in an application project:
```xml
<None Include="foo.aar" CopyToOutputDirectory="PreserveNewest"/>
```

Documentation/proposals/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,247 @@
...
+    support for developers to add their own manifest to Xamarin.Android
+    class libraries.
+  * We could also generate the `AndroidManifest.xml` so attributes
+    like `[assembly:UsesPermission]` are done ahead of time.
+  * We could generate Java stubs for every `Java.Lang.Object`
+  * If all of the above was completed, `<GenerateJavaStubs/>` could
+    skip over the assembly completely.
+
+* Build performance could be further improved by included various
+  other build-related cache files in the `.aar` file.
```
jonathanpeppers, line 241: @dellis1972 I tried to put several ideas on how we could use the `.aar` file to improve build performance in the future.

Documentation/guides/OneDotNetEmbeddedResources.md
jonpryor: > which is maybe fine?

I don't immediately see why that *wouldn't* be fine…

PR #5104, opened 2020-09-10
[build] consolidate global.json for Microsoft.Build.NoTargets

jonathanpeppers:
Let's remove `build-tools/create-packs/global.json` and have a single
`global.json` file in the root. This allows us to remove the version
number for `Sdk="Microsoft.Build.NoTargets"` in several files.

We weren't able to do this in 079197ac because of the extra
`global.json` file that took precendence over the one in the root.


jpobst (approved):
Ah, so that's why it didn't work when I tried removing the version number from these with my root `global.json`!

PR #5111, opened 2020-09-14
[Xamarin.Android.Build.Tasks] Fix ManagedResourceParser to handle duplicate `attr` elements.

dellis1972:
Context https://work.azdo.io/1184573

`aapt2` is captable of processing the following input without
error.

```
<declare-styleable name="CustomFonts">
  <attr name="android:scrollX" />
  <attr name="customFont" />
  <attr name="customFont" />
</declare-styleable>
```

Our `ManagedResourceParser` will generate the following exception

	 System.ArgumentException: An entry with the same key already exists.
		at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
		at System.Collections.Generic.TreeSet`1.AddIfNotPresent(T item)
		at System.Collections.Generic.SortedDictionary`2.Add(TKey key, TValue value)
		at Xamarin.Android.Tasks.ManagedResourceParser.Parse(String resourceDirectory, IEnumerable`1 additionalResourceDirectories, Boolean isApp, Dictionary`2 resourceMap)
		at Xamarin.Android.Tasks.GenerateResourceDesigner.RunTask()

This PR reworks the `ManagedResourceParser` to handle these
duplicate `attr` elements. It also introduces a new Task property
which will let us speficy the exact location of the `R.txt` file.
This means we no longer need to rely on relative pathing.


src/Xamarin.Android.Build.Tasks/Utilities/ManagedResourceParser.cs
```diff
@@ -198,35 +204,42 @@ public CodeTypeDeclaration Parse (string resourceDirectory, IEnumerable<string>
...
+				arrayValues.Sort (new CompareTuple ());
+				Log.LogDebugMessage ($"{field.Name} has {arrayValues.Count} items.");
 				int index = 0;
 				foreach (var arrayValue in arrayValues) {
 					value = arrayValue.Key;
 					CodeMemberField f = arrayValue.Value;
 					CodePrimitiveExpression code = expression.Initializers [index] as CodePrimitiveExpression;
+					Log.LogDebugMessage ($"Setting {field.Name} {index} = {value}");
 					code.Value = value;
+					Log.LogDebugMessage ($"Creating styleable {field.Name}_{f.Name}");
```
jonathanpeppers, line 242: Are these going to be too many `LogDebugMessage` calls? The task already logs quite a few lines.

src/Xamarin.Android.Build.Tasks/Utilities/ManagedResourceParser.cs
```diff
@@ -235,6 +248,7 @@ public CodeTypeDeclaration Parse (string resourceDirectory, IEnumerable<string>
 				}
 				field.Comments.Add (new CodeCommentStatement ($"aapt resource value: {{ {sb} }}"));
 			}
+			Log.LogDebugMessage ($"Done styleable array values");
```
jonathanpeppers, line 251: This seems like it might be extra, too.

src/Xamarin.Android.Build.Tasks/Utilities/ManagedResourceParser.cs
```diff
@@ -646,7 +662,7 @@ void ProcessStyleable (XmlReader reader)
 				CodeArrayCreateExpression c = field.InitExpression as CodeArrayCreateExpression;
 				if (c == null)
 					return;
-				
+				Log.LogDebugMessage ($"Adding array Mapping {field.Name} {fields.Count}");
```
jonathanpeppers, line 665: Maybe you meant to take these `LogDebugMessage` out, and they were just left for debugging?

jonathanpeppers (approved):

jonpryor:
There are some unit test failures, e.g. https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4066261&view=ms.vss-test-web.build-test-results-tab&runId=15761876&resultId=100085&paneView=attachments

```
  MainActivity.cs(22,36,22,40): error CS0117: 'Resource.Layout' does not contain a definition for 'Main' [/Users/runner/work/1/s/bin/TestRelease/temp/InstantRunTargetsSkipped_True/UnnamedProject.csproj]
```

dellis1972:
ok so the problem with the failing tests is because we are not correcting the casing of the items the ManagedParser is generating. Its because we are now using the `R.txt` file if it exists. Before the path was wrong so it was never picking  it up. 
I need to look into how to fix this.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
```diff
@@ -70,11 +72,14 @@ public override bool RunTask ()
 
 			// Create our capitalization maps so we can support mixed case resources
 			foreach (var item in Resources) {
-				if (!item.ItemSpec.StartsWith (ResourceDirectory))
+				var path = Path.GetFullPath (item.ItemSpec);
+				if (!path.StartsWith (ResourceDirectory))
```
jonpryor, line 76: This `.StartsWith()` should be `path.StartsWith (ResourceDirectory, StringComparison.OrdinalIgnoreCase)`…

src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
```diff
@@ -70,11 +72,14 @@ public override bool RunTask ()
...
 					continue;
 
-				var name = item.ItemSpec.Substring (ResourceDirectory.Length);
+				var name = path.Substring (ResourceDirectory.Length).TrimStart ('/', '\\');
```
jonpryor, line 79: …otherwise the `.StartsWith()` will be a "culture aware" comparison, and `path.Substring(ResourceDirectory.Length)` won't remove what you expect.

For example:

```csharp
// a & b are 'é'
var a = "\u00e9";
var b = "\u0065\u0301";                              // combining chars!
a.StartsWith(b);                                     // true
a.StartsWith(b, StringComparison.OrdinalIgnoreCase); // false
```

Now it's possible I'm overthinking this, because on Windows (at least) filesystem names are normalized, so there shouldn't be a normalization mismatch between `ResourceDirectory` and `Resources`.  However, other platforms (Linunx, macOS+APFS) perform *no* normalization, which frankly just makes this entire scenario *scarier*.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
```diff
@@ -94,7 +99,7 @@ public override bool RunTask ()
 			CodeTypeDeclaration resources;
 			if (UseManagedResourceGenerator) {
 				var parser = new ManagedResourceParser () { Log = Log, JavaPlatformDirectory = javaPlatformDirectory, ResourceFlagFile = ResourceFlagFile };
-				resources = parser.Parse (ResourceDirectory, AdditionalResourceDirectories?.Select (x => x.ItemSpec), IsApplication, resource_fixup);
+				resources = parser.Parse (ResourceDirectory, RTxtFile ?? string.Empty, AdditionalResourceDirectories?.Select (x => x.ItemSpec), IsApplication, resource_fixup);
```
jonpryor, line 102: Given the "multi-threading" concerns elsewhere (e.g. 51f56aacd9946231672f8c173e265bfe68cb52da), shouldn't this use `Path.GetFullPath(x.ItemSpec)`?

jonpryor: Ditto `RTxtFile`, which likewise won't be an absolute path.

jonpryor (changes requested):

src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
jonathanpeppers: I think the `GetFullPath()` calls would only be needed if this were an `AsyncTask` and this code was on a background thread.

`GenerateResourceDesigner` looks to be a regular `Task`.

Documentation/release-notes/5111.md
```diff
@@ -0,0 +1,9 @@
+#### Application and library build and deployment
+
+  * [Developer Community 1184573][0]: Xamarin Android falsely
+    saying that Resource does not exist in the current context (CS0103).
+    Duplicate `attr` elements in `declare-styleable`  styles were
+    causing the DesignTime Build system to crash. As a result the
+    `Resource.designer.cs` was not being updated.
+
+[0]: https://developercommunity.visualstudio.com/content/problem/1167082/xamarin-android-falsely-saying-that-resource-does.html
```
brendanzagaeski, line 9: To facilitate collation into the final release notes, this can inline the devcom link and use the phrase "design-time build":

- [Developer Community 1184573](https://developercommunity.visualstudio.com/content/problem/1167082/xamarin-android-falsely-saying-that-resource-does.html):
  Xamarin Android falsely saying that Resource does not exist in the
  current context (CS0103).  Duplicate `attr` elements in
  `declare-styleable`  styles were causing the design-time build system
  to crash. As a result, the _Resource.designer.cs_ file was not being
  updated.

brendanzagaeski (commented):
Apologies for the repetition if this was covered somewhere outside of this PR, but is it already understood why the reporting user reported this as a regression? For example, perhaps the user coincidentally updated some NuGet packages that introduced a case of a duplicate `attr` element in a `declare-styleable`?

The main reason to ask about this is that if the behavior is somehow a regression in Xamarin.Android itself, then it might be important to try to backport it to d16-8 before d16-8 is promoted from Preview to Release.

brendanzagaeski:
>  is it already understood why the reporting user reported this as a regression?

Just as a quick sanity check, I confirmed that I could hit the particular problem addressed by this PR when using Visual Studio 2019 version 16.4. That is, the design-time build version of _Resources.designer.cs_ was not updated when a `declare-styleable` contained a duplicated `attr` element. So it seems the precise behavior addressed by this PR is not itself a regression.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
dellis1972: I agree this is not an AsyncTask so the node will not be released to another task and then reused. This task will keep hold of the node until the task is complete. So I don't think we need the full paths here.

src/Xamarin.Android.Build.Tasks/Utilities/ManagedResourceParser.cs
```diff
@@ -198,13 +200,13 @@ public CodeTypeDeclaration Parse (string resourceDirectory, IEnumerable<string>
 			}
 
 			var sb = new StringBuilder ();
-			SortedDictionary<int, CodeMemberField> arrayValues = new SortedDictionary<int, CodeMemberField> ();
+
+			List<(int Key, CodeMemberField Value)> arrayValues = new List<(int Key, CodeMemberField Value)> ();
```
jonathanpeppers, line 204: 👍  I haven't seen this syntax yet, seems better than this:
```csharp
List<Tuple<int, CodeMemberField>>
```

src/Xamarin.Android.Build.Tasks/Utilities/ManagedResourceParser.cs
```diff
@@ -214,13 +216,13 @@ public CodeTypeDeclaration Parse (string resourceDirectory, IEnumerable<string>
...
 						value = Convert.ToInt32 (element?.Attribute ("id")?.Value ?? "0x0", fromBase: 16);
-						
 					} else {
 						CodePrimitiveExpression initExpression = fields [i].InitExpression as CodePrimitiveExpression;
 						value = Convert.ToInt32 (initExpression.Value);
 					}
-					arrayValues.Add (value, fields [i]);
+					arrayValues.Add ((Key: value, Value: fields [i]));
 				}
+				arrayValues.Sort (new CompareTuple ());
```
jonathanpeppers, line 225: This is really minor, but it seems like the `CompareTuple` could be stored in a static field and reused. Seems like this loop could get called a lot?

jonathanpeppers (approved):

Documentation/release-notes/5111.md
```diff
@@ -0,0 +1,6 @@
+#### Application and library build and deployment
+
+  * [Developer Community 1184573](https://developercommunity.visualstudio.com/content/problem/1167082/xamarin-android-falsely-saying-that-resource-does.html):
+    Xamarin Android falsely saying that Resource does not exist in the current context (CS0103).
+    Duplicate `attr` elements in `declare-styleable`  styles were causing the DesignTime Build
+    system to crash. As a result the `Resource.designer.cs` was not being updated.
```
brendanzagaeski, line 6: To facilitate collation into the final release notes, this can use the phrase "design-time build"  (the phrase "[design-time build](https://www.bing.com/search?q=%2B"design-time+build"+site%3Adocs.microsoft.com)" is much more common than "[designtime build](https://www.bing.com/search?q=%2B"designtime+build"+site%3Adocs.microsoft.com) on docs.microsoft.com)" and use italics for the file name to match the, admittedly somewhat interesting, convention of italics for file names from the [docs guidelines](https://github.com/MicrosoftDocs/xamarin-docs/blob/c7d82b0684b1c26cdcbc16766b636894a64a80fd/contributing-guidelines/template.md#text-styling).

All together:

```markdown
- [Developer Community 1184573](https://developercommunity.visualstudio.com/content/problem/1167082/xamarin-android-falsely-saying-that-resource-does.html):
  Xamarin Android falsely saying that Resource does not exist in the
  current context (CS0103).  Duplicate `attr` elements in
  `declare-styleable`  styles were causing the design-time build system
  to crash. As a result, the _Resource.designer.cs_ file was not being
  updated.
```

brendanzagaeski (approved):
Release note approved. Thanks! (Files reviewed: `*.md`)

PR #5114, opened 2020-09-14
[Xamarin.Android.Build.Tests][CI] Automatically adjust PR test scope

pjcollins:
Reduces the default test scope executing on every commit to every pull
request.  We currently run a large swath of MSBuild tests and consume a
lot of machine hours in our PR builds, even for code changes which could
not affect the outcome of these tests.

A new `Smoke Test` stage has been introduced to replace our previous
`Test` stage.  The former `Test` stage has been broken up into
`Legacy Tests`, `One .NET Tests`, and `Regression Tests` stages.

A new scenario has been added to `xaprepare` that is responsible for
determining the set of files that have changed in a particular PR.
These files are compared against a list of known source areas to
determine which test jobs to run.  The `mac_build` job will run this
scenario and save an output variable named `TestConditions.TestAreas`.
This variable will contain a separated list of test jobs to run such as:
`MSBuild,MSBuildDevice,BCL,TimeZone,Designer`.

While we now have a decent way to automatically adjust test scope
for PR builds based on the source that has changed, there will still be
instances where test runs will need to be manually triggered when a
PR author or reviewer wants additional test coverage.  I am hoping
that [feature work][1] to improve parity between multi-stage YAML
pipelines and UI based release definitions can improve this story for
us in the future.

Previous test stage/job structure:

```
Mac Build 
|
└─── Test (Automatic)
│   │   APK Instrumentation
│   │   BCL With Emulator
│   │   MSBuild - Mac-1 - Legacy
│   │   MSBuild - Mac-2 - Legacy
│   │   MSBuild - Mac-3 - Legacy
│   │   MSBuild - Win-1 - Legacy
│   │   MSBuild - Win-2 - Legacy
│   │   MSBuild - Win-3 - Legacy
│   │   MSBuild With Emulator
│   │   Designer - macOS
│   │   Designer - Windows
|
└─── One .NET Tests (Optional)
|   │   MSBuild - Mac-1 - One .NET
|   │   MSBuild - Mac-2 - One .NET
|   │   MSBuild - Mac-3 - One .NET
|   │   MSBuild - Win-1 - One .NET
|   │   MSBuild - Win-2 - One .NET
|   │   MSBuild - Win-3 - One .NET
|
└─── Regression Tests (Optional)
    │   TimeZone - Mac-1
    │   TimeZone - Mac-1
    │   TimeZone - Mac-1
```

New test stage/job structure:

```
Mac Build 
|
└─── Smoke Test (Automatic)
│   │   APK Instrumentation
|   │   MSBuild Smoke - Legacy & One .NET - macOS
|   │   MSBuild Smoke - Legacy & One .NET - Windows
|   │   MSBuild Smoke Tests With Emulator
|
└─── Legacy Tests (Optional)
|   │   MSBuild - Legacy - macOS-1
|   │   MSBuild - Legacy - macOS-2
|   │   MSBuild - Legacy - macOS-2
|   │   MSBuild - Legacy - Windows-1
|   │   MSBuild - Legacy - Windows-2
|   │   MSBuild - Legacy - Windows-3
|
└─── One .NET Tests (Optional)
|   │   MSBuild - One .NET - macOS-1
|   │   MSBuild - One .NET - macOS-2
|   │   MSBuild - One .NET - macOS-3
|   │   MSBuild - One .NET - Windows-1
|   │   MSBuild - One .NET - Windows-2
|   │   MSBuild - One .NET - Windows-3
|
└─── MSBuild Emulator Tests (Optional)
|   │   macOS
|
└─── Designer Tests (Optional)
|   │   macOS
|   │   Windows
|
└─── TimezoneInfo Emulator Tests (Optional)
|   │   macOS-1
|   │   macOS-2
|   │   macOS-3
|
└─── BCL Emulator Tests (Optional)
    |   macOS
```

The new `MSBuild Smoke Test` jobs only run against the `net472` version
of the test assembly, One .NET coverage in these smoke test jobs is
limited to the `XASdkTests` test cases that are tagged with the smoke
category.

The scope of `SmokeTests` in `Xamarin.Android.Build.Tests` has been
reduced as part of these changes.  The `SmokeTests` category has also
been added to a few tests in the `MSBuildDeviceIntegration` suite.
Jobs in the optional MSBuild stages are set to exclude the `SmokeTest`
category if those tests had previously ran in the `Smoke Test` stage.

[1]: https://dev.azure.com/mseng/AzureDevOpsRoadmap/_workitems/edit/1663893


pjcollins:
Full test run for commit 8f00c16 - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4067340&view=results

pjcollins:
PR build for commit 54b479e: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4077462&view=results
Full test run for commit 54b479e: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4077977&view=results

jonathanpeppers (commented):
Could we split up the `Test` stage some?

This screen is useful to me, as sometimes I uncheck everything besides the One .NET tests:

image

It might be good to have stages like:
* Designer
* Time Zone
* MSBuild Legacy
* MSBuild One .NET
* BCL

jpobst:
I like Peppers suggestion to break it up some, but maybe not that much, as I would like this screen to remain usable:

image

Maybe?
* MSBuild Legacy
* MSBuild One .NET
* Regression Tests   <--  (Everything else)

jpobst:
Also, fantastic work!  Looking forward to sub-2 hour PR builds!!!

We just need to commit as a team to keeping `master` green so we can immediately identify new failures, since there will probably be a few that slip in due to running fewer tests on PRs.

grendello:
I'd love to have triggers based on the commit path(s) which can cause automatic running of possibly affected tests

pjcollins:
Thanks for the quick feedback! To also sum up some other external discussion, I'll try to take a look at @grendello's suggestion of being smarter about when certain test jobs run, rather than requiring the author manually queue a new instance for everything.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -88,7 +88,6 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
 
 		[Test]
 		[TestCaseSource (nameof (AotChecks))]
-		[Category ("SmokeTests")]
 		public void BuildAotApplicationAndÜmläüts (string supportedAbis, bool enableLLVM, bool expectedResult)
```
dellis1972, line 91: I do worry about not running at least some of this (and the Debugger test) during a PR build. Things like mono bumps can break these and unless we are careful we might miss something. 

One idea might be to keep the Category, but re work the TestCaseSource to be a method which returns the list of tests to run (we have done this before https://github.com/xamarin/xamarin-android/blob/6ef8ba47f2912cb7fe310924a8b85515a41d43a1/tests/MSBuildDeviceIntegration/Tests/DeploymentTest.cs#L206). The then might be able to query if we are in "SmokeTest" mode and only run a limited set of these (even just 1) to make sure we don't break anything.

pjcollins: My intent here was to reduce the default checks that run, and not to disable the ability to run all of our tests on any given PR. This does add a manual step that may make the authors work more burdensome in some cases, and as a result I am looking at having variability in the test stages that will run by default based on the code that is changing.

pjcollins: Mono bumps should now result in every test job running automatically on each commit to the PR with the .external file change.

pjcollins:
Sample of an automatic PR build which changes a file in `Xamarin.Android.Build.Tasks`: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4086007&view=results

jpobst:
Very clever way of doing this!  I've been trying to come up with something for months and came up empty handed.

As a future enhancement, we should be able to use `$(System.PullRequest.PullRequestId)` and the GitHub API to look up the labels on a PR and support the `full-mono-integration-build` label to manually mark a PR as needing a full build.

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
```diff
@@ -0,0 +1,65 @@
...
+				} else if (file.Contains ("external/Java.Interop")) {
+					testAreas.Add ("MSBuild");
+				} else if (file.Contains ("Xamarin.Android.Build.Tasks")) {
+					testAreas.Add ("MSBuild");
+				}
+			}
+
+			//foreach (var testArea in testAreas.Distinct ()) {
+			//	Log.MessageLine ($"##vso[task.setvariable variable={testArea};isOutput=true]True");
+			//}
```
jonathanpeppers, line 59: You might leave this in to always log this.

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
```diff
@@ -0,0 +1,65 @@
...
+
+			var git = new GitRunner (context);
+			var filesChanged = await git.RunCommandForOutputAsync (BuildPaths.XamarinAndroidSourceRoot, "diff", "--name-only", commitRevision, commitMessagePieces [3]);
+			if (filesChanged == null || filesChanged.Count < 1) {
+				Log.ErrorLine ($"Unable to determine if any files were changed in this PR.");
+				return false;
+			}
+
+			var testAreas = new List<string> ();
+			foreach (string file in filesChanged) {
```
jonathanpeppers, line 43: Do you think it's worth logging each file in `filesChanged`?

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
```diff
@@ -0,0 +1,65 @@
...
+				// MSBuild: Runs all legacy and One .NET Xamarin.Android.Build.Task tests, as well as designer tests.
+				// Mono: Runs all BCL and timezone unit tests, as well as designer tests.
+				if (file == ".external") {
+					testAreas.Add ("Mono");
+					testAreas.Add ("MSBuild");
+				} else if (file.Contains ("external/Java.Interop")) {
+					testAreas.Add ("MSBuild");
+				} else if (file.Contains ("Xamarin.Android.Build.Tasks")) {
+					testAreas.Add ("MSBuild");
+				}
```
jonathanpeppers, line 54: One missing here is if you modified `src/monodroid`: I think @grendello would want the designer tests to run. But maybe doesn't need the BCL tests in that case?

grendello: Yep, BCL tests wouldn't be necessary. Designer tests, however, need to be ran for sure (and possibly the XF integration test)

grendello: It would be a good idea to check for the test paths themselves - if a test is modified by the commit, it should be ran.

grendello: Then, `Mono.Android` changes should trigger `Mono.Android/Test` + its 2 derivatives (MultiDex, AppBundle)

grendello: Then any Android NDK or SDK updates should trigger all the tests, IMO

pjcollins: All of the APK tests (including XF integration) will continue to run against every PR. That job executes in under one hour and I felt that we didn't need to cut scope there at this time.

I don't want to overcomplicate this by introducing source checks and job conditions for every test suite that we have. If an author is modifying a test suite that didn't run as part of the PR build it should hopefully be clear to the author that the relevant stage needs to be manually queued. I think the same applies for NDK/SDK updates.

I'd like this PR to cover the most common source areas that authors may be uncertain about relevant test coverage for, not every source area.

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
pjcollins: I had temporarily commented this out to get "default" pr build results, this has been restored.

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
pjcollins: I added some logging here.

pjcollins:
Example of a 'default' PR build (no relevant src path changes detected): https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4086758&view=results,

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
pjcollins: I reworked this to use a single variable that will contain multiple test area values. Valid values are associated with test job types instead of test suites.

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
```diff
@@ -0,0 +1,98 @@
...
+					testAreas.Add ("Designer");
+				}
+
+				if (file.Contains ("src/Xamarin.Android.Build.Tasks")) {
+					testAreas.Add ("MSBuild");
+					testAreas.Add ("MSBuildDevice");
+					testAreas.Add ("Designer");
+				}
+
+				if (file.Contains ("src/monodroid")) {
```
jonpryor, line 71: This should likely also add `MSBuildDevice`, as it impacts app startup & runtime behavior.

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
```diff
@@ -0,0 +1,98 @@
...
+			}
+
+			var git = new GitRunner (context);
+			var filesChanged = await git.RunCommandForOutputAsync (BuildPaths.XamarinAndroidSourceRoot, "diff", "--name-only", commitRevision, commitMessagePieces [3]);
+			if (filesChanged == null || filesChanged.Count < 1) {
+				Log.ErrorLine ($"Unable to determine if any files were changed in this PR.");
+				return false;
+			}
+
+			var testAreas = new List<string> ();
```
jonpryor, line 42: Should this be a `HashSet<string>`, so that duplicates aren't duplicated in output?

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
```diff
@@ -0,0 +1,98 @@
...
+			foreach (string file in filesChanged) {
+				Log.InfoLine ($"Detected change in file: '{file}'.");
+				// Compare files changed against common areas requiring additional test scope.
+				// MSBuild: Runs all legacy and One .NET Xamarin.Android.Build.Task tests
+				// MSBuildDevice: Runs all MSBuildDeviceIntegration tests
+				// BCL: Runs BCL tests on emulator
+				// TimeZone: Runs timezone unit tests on emulator
+				// Designer: Runs designer integration tests
+				if (file == ".external") {
```
jonpryor, line 51: Aside: what runs the on-device "apk" tests? Is that a separate area?

Might need a semi-exhaustive listing of `src/*`?

   * `src/Microsoft.Android.Sdk.ILLink`: OneDotNet tests + apk tests
   * `src/Mono.Android`: apk tests, Designer
   * `src/Mono.Android.Export`: apk tests
   * `src/OpenTK-1.0`: apk tests
   * `src/System.EnterpriseServices`: ignore
   * `src/Xamarin.Android.Build.Tasks`: MSBuild, MSBuildDevice, Designer
   * `src/Xamarin.Android.NUnitLite`: apk tests
   * `src/Xamarin.Android.NamingCustomAttributes`: ignore
   * `src/Xamarin.Android.Tools.Aidl`: MSBuild
   * `src/Xamarin.Android.Tools.JavadocImporter`: MSBuild
   * `src/aapt2`: MSBuild
   * `src/apksigner`: MSBuild? MSBuildDevice?
   * `src/bundletool`: MSBuild? MSBuildDevice?
   * `src/java-runtime`: apk tests
   * `src/manifestmerger`: MSBuild? MSBuildDevice?
   * `src/monodroid`: apk tests
   * `src/proguard`: MSBuild? MSBuildDevice? apk tests?
   * `src/r8`: MSBuild? MSBuildDevice? apk tests?
   * `src/sqlite-xamarin`: apk tests

jonpryor:
> Reduces the default test scope executing on every commit to every pull request.

which presumably means PR builds will be faster, as fewer tests will be run.

Any idea how long a "quick" change will now be, e.g. a Documentation change (no tests at all!)? Can we get a "lower-bound" for turnaround time?  Will the "upper time" be any *worse*?

jpobst:
Another crazy idea: add `or $(System.JobAttempt) > 1` to the `condition`s, and then we could click the `Rerun stage` button to kick off tests from the UI.  😉 

image

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
pjcollins: The the on-device "apk" tests will continue to run on every PR build as they are part of the smoke test stage. That job executes in under one hour and I felt that we didn't need to try to cut scope there at this time. As a result, we won't need to associate any source changes with [that particular test job](https://github.com/xamarin/xamarin-android/blob/05620bb1dedf499aeb80ee2e1ea4d52dd1eef64e/build-tools/automation/azure-pipelines.yaml#L399-L658).

I think we could run extra MSBuild test jobs based on other `src` changes as suggested, @jonathanpeppers and/or @dellis1972 what are your thoughts on the source paths above?

pjcollins:
> Any idea how long a "quick" change will now be, e.g. a Documentation change (no tests at all!)? Can we get a "lower-bound" for turnaround time? Will the "upper time" be any worse?

From what I've seen so far these changes allow for ~2 hour PR builds. Our build job takes roughly an hour on average, and this PR will still run four test jobs by default for every PR build (the longest of which can also take up to an hour). In theory a documentation change wouldn't even require a build, perhaps we can look at improving this further for certain source paths in the future.

Upper bound timing should be mostly unchanged, though it's possible we see some minor improvements there. Our longest jobs are typically Nodes 1 and 3 of our Legacy or One .NET MSBuild tests, and these jobs will now skip tests marked with the `SmokeTest` category (as they will have ran during the smoke test stage). 


> Another crazy idea: add or $(System.JobAttempt) > 1 to the conditions, and then we could click the Rerun stage button to kick off tests from the UI. 😉

I really like the idea of using `$(System.StageAttempt)` or `$(System.JobAttempt)` as a way of running extra test jobs without having to queue a new instance, but I think I'd want to try to experiment with that in a separate PR as the scope has grown quite a bit already.

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
jonathanpeppers: The ones above look right to me. `proguard` doesn't need the MSBuild One .NET tests.

Maybe could also add:
* `src-ThirdParty/android-platform-tools-base` - MSBuild, MSBuildDevice
* `src-ThirdParty/bazel` - MSBuildDevice
* `tests/MSBuildDeviceIntegration` - MSBuildDevice
* `tests/msbuild-times-reference` - MSBuildDevice

dellis1972: `src/java-runtime` will also need MSBuild tests I think and MSBuildDevice. This folder contains the release/debug runtimes. I don't think the apk tests will test debug/fastdev at all, so we might not get good coverage just running those tests.

dellis1972: same for monodroid changes.

jpobst:
One last little nitpick. It is possible to skip the stage instead of the jobs?  Looking at this screen it really looks like all tests were run and are green, when really they were all skipped.

image

pjcollins:
@jpobst I could not get this to work at a stage level, but the documentation I could find seems to imply that it should work. However, if we do manage to make this work we'd have to do one of the following:

A) Break up the "Regression Test" stage into additional stages which represent each test job type (which I think you were opposed to earlier) or
B) Reduce the level of granularity we have in selecting test jobs (which I think others are opposed to)

Alternatively, we could probably set some conditions on the stages themselves to make sure the stage is skipped if all the jobs would be skipped. Aside from the added complexity of this approach, another downside would be that the 'Rerun stage' button wouldn't light up and we couldn't use `$(System.*Attempt)` variables to queue additional tests.

If none of those options seem appealing, we'll have to ensure everyone has awareness of the UI being a bit misleading and/or rely more so on the checks reported back to the PR.

build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs
pjcollins: I think I've addressed all of these source areas.

pjcollins:
Example of a default smoke test PR build (no changes detected) for commit bf9a345: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4094018&view=results

tests/MSBuildDeviceIntegration/Tests/XASdkDeployTests.cs
```diff
@@ -13,7 +13,7 @@
 namespace Xamarin.Android.Build.Tests
 {
 	[NonParallelizable]
-	[Category ("UsesDevices"), Category ("DotNetIgnore")] // These don't need to run under `--params dotnet=true`
+	[Category ("UsesDevices"), Category ("SmokeTests"), Category ("DotNetIgnore")] // These don't need to run under `--params dotnet=true`
```
jonathanpeppers, line 16: If you make any other changes here you might update the comment:
```suggestion
	[Category ("UsesDevices"), Category ("SmokeTests"), Category ("DotNetIgnore")] // These don't need to run under `dotnet test`
```
But it's fine to leave it if this is finished.

pjcollins:
I think this should be ready to go if everyone could take another review pass.

jonathanpeppers (approved):

grendello (approved):

dellis1972 (approved):

jonpryor:
Squash-and-merge message:

```markdown
Reduce the default test scope executing on every commit for every
pull request.  We currently run a large swath of MSBuild tests and
consume a lot of machine hours in our PR builds, even for code
changes which could not affect the outcome of these tests.

Reduce the default test scope by adding a new
`xaprepare -s:DetermineApplicableTests` scenario, which looks at the
set of files that have changed in the PR.  These files are compared
against a list of known source path prefixes to determine which test
jobs to run.  The `mac_build` job will run this scenario and save an
output variable named `TestConditions.TestAreas`.  This variable will
contain a separated list of test jobs to run such as:
`MSBuild,MSBuildDevice,BCL,TimeZone,Designer`.  This list of test jobs
controls which tests are actually executed.

A new `Smoke Test` stage is introduced to replace our previous `Test`
stage.  The former `Test` stage has been broken up into `Legacy Tests`,
`One .NET Tests`, and `Regression Tests` stages, which may or may not
be executed depending on the `TestConditions.TestAreas` variable:

	Mac Build 
	|
	└─── Smoke Test [Always executed]
	│   │   APK Instrumentation
	|   │   MSBuild Smoke - Legacy & One .NET - macOS
	|   │   MSBuild Smoke - Legacy & One .NET - Windows
	|   │   MSBuild Smoke Tests With Emulator
	|
	└─── Legacy Tests [Optional; depends on TestAreas]
	|   │   MSBuild - Legacy - macOS-1
	|   │   …
	|   │   MSBuild - Legacy - Windows-1
	|   │   …
	|
	└─── One .NET Tests [Optional; depends on TestAreas]
	|   │   MSBuild - One .NET - macOS-1
	|   │   …
	|   │   MSBuild - One .NET - Windows-1
	|   │   …
	|
	└─── MSBuild Emulator Tests [Optional]
	|   │   macOS
	|
	└─── Designer Tests [Optional; depends on TestAreas]
	|   │   macOS
	|   │   Windows
	|
	└─── TimezoneInfo Emulator Tests [Optional; depends on TestAreas]
	|   │   macOS-1
	|   │   …
	|
	└─── BCL Emulator Tests [Optional; depends on TestAreas]
	    |   macOS

While we now have a decent way to automatically adjust test scope
for PR builds based on the source that has changed, there will still
be instances where test runs will need to be manually triggered when
a PR author or reviewer wants additional test coverage.  I am hoping
that [feature work][1] to improve parity between multi-stage YAML
pipelines and UI based release definitions can improve this story for
us in the future.

The new `MSBuild Smoke Test` jobs only run against the `net472`
version of the test assembly, One .NET coverage in these smoke test
jobs is limited to the `XASdkTests` test cases that are tagged with
the smoke category.

The scope of `SmokeTests` in `Xamarin.Android.Build.Tests` has been
reduced as part of these changes.  The `SmokeTests` category has also
been added to a few tests in the `MSBuildDeviceIntegration` suite.
Jobs in the optional MSBuild stages are set to exclude the `SmokeTest`
category if those tests had previously ran in the `Smoke Test` stage.

[1]: https://dev.azure.com/mseng/AzureDevOpsRoadmap/_workitems/edit/1663893
```

pjcollins:
The stages changed a bit late in the PR cycle, the wording will need updating:

```markdown
A new `Smoke Test` stage is introduced to replace our previous `Test`
stage.  The former `Test` stage has been broken up into `Legacy Tests`,
`One .NET Tests`, and `Regression Tests` stages, which may or may not
be executed depending on the `TestConditions.TestAreas` variable:
```
to be replaced with something like:
```markdown
A new `Smoke Tests` stage is introduced to replace our previous `Test`
stage.  The former `Test` stage has been broken up into `Legacy Tests`,
`One .NET Tests`, `MSBuild Emulator Tests`, `Designer Tests`,
`TimezoneInfo Emulator Tests`, and `BCL Emulator Tests` stages, which
may or may not be executed depending on the
`TestConditions.TestAreas` variable:
```

PR #5115, opened 2020-09-15
[linker] set %(IsTrimmable) for @(KnownFrameworkReference)

jonathanpeppers:
Context: https://github.com/dotnet/installer/blob/97f447f228824c0217d2c8b845b003118673ff22/src/redist/targets/GenerateBundledVersions.targets#L208-L218
Context: https://docs.microsoft.com/dotnet/core/deploying/trimming-options

Right now we are modifying the `@(ResolvedFileToPublish)` item group
and setting `%(IsTrimmable)` for `Mono.Android.dll`,
`Mono.Android.Export.dll`, and `Java.Interop.dll`. Instead we can set
this metadata on `@(KnownFrameworkReference)`, which is the way that
`Microsoft.NETCore.App` is setup. This way these assembly references
will be "trimmable" by default.

We should also rename `Microsoft.Android.Sdk.Linker.targets` to
`Microsoft.Android.ILLink.targets` to better match the naming in .NET 5.

### DIFF
--- a/build-tools/create-packs/Microsoft.Android.Sdk.proj
+++ b/build-tools/create-packs/Microsoft.Android.Sdk.proj
@@ -113,17 +113,19 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
     <XamarinAndroidVersion>$(AndroidPackVersionLong)</XamarinAndroidVersion>
   </PropertyGroup>
   <ItemGroup>
-    <KnownFrameworkReference Include="Microsoft.Android"
-                              TargetFramework="$(_AndroidNETAppTargetFramework)"
-                              RuntimeFrameworkName="Microsoft.Android"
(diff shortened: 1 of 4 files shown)


dellis1972 (approved):

pjcollins (approved):

jonathanpeppers:
I think the `LibraryProjectTargetsDoNotBreak` test might have a failure from this, let me look at it.

jonpryor:
I'm not sure it's safe to set `%(IsTrimmable)` on `Mono.Android.Export.dll`, as I suspect it would be *entirely removed*.  I think this assembly needs to be excluded.

jonathanpeppers:
I need to start an issue or something to track some of the TODO items with the linker. Mono.Android.Export needs to be on the list.

This change didn't actually change the behavior, Mono.Android.Export is handled the same.

PR #5126, opened 2020-09-17
[net5/illink] Add ApplyPreserveAttribute substep

radekdoulik:
The ApplyPreserveAttribute susbtep  was not (yet?) upstreamed, so
make it build and add it to the pipeline and run it before other
substeps to not break ac5269d54e518d9db6dac31e49de26c8b288662c fix.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs
@@ -8,6 +8,10 @@ using Mono.Tuner;
 
 using Mono.Cecil;
 
+#if NET5_LINKER
+using Microsoft.Android.Sdk.ILLink;
+#endif
+
 namespace MonoDroid.Tuner {
 
 	public class ApplyPreserveAttribute : ApplyPreserveAttributeBase {
--- a/src/Microsoft.Android.Sdk.ILLink/ApplyPreserveAttributeBase.cs
+++ b/src/Microsoft.Android.Sdk.ILLink/ApplyPreserveAttributeBase.cs
@@ -1,0 +1,165 @@
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+using Mono.Linker;
+using Mono.Linker.Steps;
+
+using Mono.Cecil;
+
+namespace Microsoft.Android.Sdk.ILLink
+{
+
+	public abstract class ApplyPreserveAttributeBase : BaseSubStep {
+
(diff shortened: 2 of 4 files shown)


radekdoulik:
Let add a test for it as well.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs
```diff
@@ -8,6 +8,10 @@
 
 using Mono.Cecil;
 
+#if NET5_LINKER
```
jonathanpeppers, line 11: Should we change this `#if` to `NETCOREAPP3_1` and we could remove this `$(DefineConstants)`?

https://github.com/xamarin/xamarin-android/blob/eadd3bda8e2a81c6ca997f05d3faac7040b29a7b/src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj#L5

If we change this `.csproj` to `TargetFramework=net5.0` later on, then `NETCOREAPP3_1` will still be defined.

jonathanpeppers (commented):
We should check is if they can actually support `[Preserve]` in .NET 6 for us, so we wouldn't have to bring this over.

rolfbjarne:
For reference, this is xamarin-macios' implementation: https://github.com/xamarin/xamarin-macios/pull/9529

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs
radekdoulik: I think this was just changed recently (https://github.com/dotnet/sdk/pull/13615). So `NETCOREAPP3_1` will not work. I think we should use `#if NET` instead. I will try to use it.

radekdoulik: It is not defined yet in the version we are using, so I will keep `NET5_LINKER` around until we have the `NET` define.

radekdoulik:
> We should check is if they can actually support [Preserve] in .NET 6 for us, so we wouldn't have to bring this over.

I would like to get this in so that we have it now. It looks like the `[Preserve]` might not land in upstream, at least not soon?

jonathanpeppers (approved):
I think we should go ahead and bring this in, and we can remove it if it makes it upstream.

radekdoulik:
I will wait with adding test until the Full linker mode works, noted in https://github.com/xamarin/xamarin-android/issues/5132 to not forget it.

PR #5131, opened 2020-09-17
[One .NET] use .aar instead of EmbeddedResource

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/2441
Implements: Documentation/proposals/OneDotNetEmbeddedResources.md

Creation of `EmbeddedResource` files:

* `__AndroidEnvironment__*`
* `__AndroidLibraryProjects__.zip`
* `__AndroidNativeLibraries__.zip`
* `*.jar`

...has moved to `Xamarin.Android.Legacy.targets`.

A new `_CreateAar` MSBuild target and `<CreateAar/>` MSBuild task now
creates `.aar` files in `$(OutputPath)` for class libraries.

The `_ResolveAars` MSBuild target will locate `.aar` files next to
.NET assemblies and use them if found.

The MSBuild targets/tasks for consuming `EmbeddedResource` files in
assemblies remains in place for backwards compatibility. The only
change was to support the new `.netenv` directory inside `.aar` files.

The `_CategorizeAndroidLibraries` MSBuild target handles setting up
the new, simplified `@(AndroidLibrary)` item group.

I added several tests for these scenarios.

Other changes:

* A new `<ExtractJarsFromAar/>` MSBuild task is needed for
  `@(LibraryProjectZip)` items for Java bindings. This task extracts
  `.jar` files to `$(IntermediateOutputPath)library_project_jars` and
  `annotations.zip` items to `$(IntermediateOutputPath)library_project_annotations`.
* The `@(LibraryProjectZip)` item group now works in .NET 5+. We no
  longer need the `LibraryProjectZip` test category, as these tests
  pass. Several of these tests needed to set `ProjectName` as they had
  two projects named `UnnamedProject.csproj`.
* `<ResolveLibraryProjectImports/>` needs to store the full file *and*
  the extension in `obj\$(Configuration)\lp\map.cache` now as you
  would commonly have both a `Foo.dll` and `Foo.aar`. I had to adjust
  `AssemblyIdentityMap` to account for this.
* `.aar` files can no longer assume `%(AndroidSkipResourceProcessing)`
  is `false` by default. `.aar` files coming from Xamarin.Android
  class libraries could have custom views that need to be fixed up at
  build time in the main Xamarin.Android application project.
* When the nested, inner build for the
  `_ComputeFilesToPublishForRuntimeIdentifiers` MSBuild target runs, I
  further trimmed down which targets run. `IncrementalClean` was
  running multiple times and deleting files!
* Fixed where `Files.ExtractAll` was extracting a `.DS_Store` file in
  the root of `.aar` files.

### DIFF
--- a/Documentation/guides/OneDotNetEmbeddedResources.md
+++ b/Documentation/guides/OneDotNetEmbeddedResources.md
@@ -1,4 +1,4 @@
-# [proposal] Embedded Resources and `.nupkg` Files
+# One .NET: Embedded Resources and `.nupkg` Files
 
 Traditionally, a Xamarin.Android class library can contain many types
 of Android-specific files:
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
@@ -1,0 +1,121 @@
+<!--
+***********************************************************************************************
+Microsoft.Android.Sdk.AndroidLibraries.targets
+
+This file contains the .NET 5-specific targets related to
+@(AndroidLibrary) items and .aar files generation for class library
+projects.
+
+***********************************************************************************************
+-->
+<Project>
+
+  <UsingTask TaskName="Xamarin.Android.Tasks.CreateAar"          AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
+  <UsingTask TaskName="Xamarin.Android.Tasks.ExtractJarsFromAar" AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
+
+  <ItemGroup>
+    <AvailableItemName Include="AndroidLibrary" />
+  </ItemGroup>
+  <ItemDefinitionGroup>
+    <AndroidLibrary>
+      <Bind>false</Bind>
+      <Pack>true</Pack>
+    </AndroidLibrary>
+  </ItemDefinitionGroup>
+  <PropertyGroup>
+    <_AarCacheFile>$(IntermediateOutputPath)$(TargetName).aar.cache</_AarCacheFile>
+    <_AarOutputPath>$(OutputPath)$(TargetName).aar</_AarOutputPath>
+  </PropertyGroup>
+
+  <Target Name="_CategorizeAndroidLibraries">
+    <ItemGroup Condition=" '$(AndroidApplication)' == 'true' ">
+      <AndroidAarLibrary    Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.aar' " />
+      <AndroidJavaLibrary   Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.jar' " />
+      <AndroidNativeLibrary Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.so' " />
+    </ItemGroup>
+    <ItemGroup Condition=" '$(AndroidApplication)' != 'true' ">
+      <AndroidAarLibrary     Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.aar' and '%(AndroidLibrary.Bind)' != 'true' " />
+      <LibraryProjectZip     Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.aar' and '%(AndroidLibrary.Bind)' == 'true' " />
+      <AndroidJavaLibrary    Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.jar' and '%(AndroidLibrary.Bind)' != 'true' " />
+      <EmbeddedJar           Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.jar' and '%(AndroidLibrary.Bind)' == 'true' " />
+      <EmbeddedNativeLibrary Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.so' " />
+      <None Include="@(LibraryProjectZip)" CopyToOutputDirectory="PreserveNewest" Link="%(Filename)%(Extension)" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="_ResolveAars" Condition=" '$(AndroidApplication)' == 'true' ">
+    <ItemGroup>
+      <_AarSearchDirectory   Include="@(_ReferencePath->'%(RootDir)%(Directory)')" />
+      <_AarSearchDirectory   Include="@(_ReferenceDependencyPaths->'%(RootDir)%(Directory)')" />
+      <_AarDistinctDirectory Include="@(_AarSearchDirectory->Distinct())" />
+      <_AarFromLibraries     Include="%(_AarDistinctDirectory.Identity)*.aar" />
+    </ItemGroup>
+    <ItemGroup Condition=" '@(_AarFromLibraries->Count())' != '0' ">
+      <AndroidAarLibrary Include="@(_AarFromLibraries)" Exclude="@(AndroidAarLibrary)" />
+      <!--
+        NOTE: %(AndroidSkipResourceProcessing) is required for located .aar's; there could be custom views.
+        Update in a second step, in case there is was an existing @(AndroidAarLibrary) with the same path.
+      -->
+      <AndroidAarLibrary Update="@(_AarFromLibraries)" AndroidSkipResourceProcessing="false" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="_CreateAarCache"
+      Condition=" '$(AndroidApplication)' != 'true' ">
+    <ItemGroup>
+      <_CreateAarInputs Include="@(AndroidAsset)" />
+      <_CreateAarInputs Include="@(_AndroidResourceDest)" />
+      <_CreateAarInputs Include="@(AndroidEnvironment)" />
+      <_CreateAarInputs Include="@(EmbeddedJar)" />
+      <_CreateAarInputs Include="@(EmbeddedNativeLibrary)" />
+    </ItemGroup>
+    <Hash
+        ItemsToHash="@(_CreateAarInputs)"
+        IgnoreCase="false">
+      <Output TaskParameter="HashResult" PropertyName="_AarDependencyHash" />
+    </Hash>
+    <WriteLinesToFile
+        Condition=" '$(_AarDependencyHash)' != '' "
+        Lines="$(_AarDependencyHash)"
+        File="$(_AarCacheFile)"
+        Overwrite="True"
+        WriteOnlyWhenDifferent="True"
+    />
+    <ItemGroup Condition=" '$(_AarDependencyHash)' != '' ">
+      <_CreateAarInputs Include="$(_AarCacheFile)" />
+      <FileWrites       Include="$(_AarCacheFile)" />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="_CreateAar"
+      Condition=" '$(AndroidApplication)' != 'true' "
+      DependsOnTargets="_CreateAarCache"
+      Inputs="@(_CreateAarInputs)"
+      Outputs="$(_AarOutputPath)">
+    <CreateAar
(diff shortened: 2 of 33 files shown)


jonpryor:
How much of this new behavior is specific to .NET 6, and how much *should* be specific to .NET 6?  For example, can Xamarin.Android 11.2 builds produce `.aar` file outputs -- avoiding the `@(EmbeddedResource)` extraction/etc. -- so that customers can get reduced build times "now" and not in a year?

Would it be possible to support `@(LibraryProjectZip)` in both library & app projects *now*, reducing/removing the need for separate & distinct binding projects, and easing the migration to .NET 6?

jonathanpeppers:
> How much of this new behavior is specific to .NET 6, and how much should be specific to .NET 6?

These changes only work for .NET 6 because it relies on the fact that binding projects and regular projects are "merged".

It is would be possible to make `Xamarin.Android.CSharp.targets` support bindings, `.aar` files, etc. We should consider it if the timeline gets pushed back or something.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildAssetsTest.cs
```diff
@@ -150,16 +150,17 @@ public void FullPath ()
...
-					var entryName = "library_project_imports/assets/foo.txt";
-					var entry = zip.ReadEntry (entryName);
-					Assert.IsNotNull (entry, $"{libraryProjectImports} should contain {entryName}");
-					using (var memory = new MemoryStream ()) {
-						entry.Extract (memory);
-						memory.Position = 0;
-						Assert.AreEqual ("bar", Encoding.Default.GetString (memory.ToArray ()));
+				if (Builder.UseDotNet) {
+					var aarPath = Path.Combine (Root, b.ProjectDirectory, proj.OutputPath, $"{proj.ProjectName}.aar");
+					FileAssert.Exists (aarPath);
```
dellis1972, line 155: just a thought. do we need a test to make sure that we DONT include any EmbeddedResource like `__AndroidLibraryProjects__.zip` for dotnet projects?

jonathanpeppers: I added `FileAssert.DoesNotExist (libraryProjectImports);`, and in another test, did this:

```csharp
// Check EmbeddedResource files do not exist
using (var assembly = AssemblyDefinition.ReadAssembly (assemblyPath)) {
	Assert.AreEqual (0, assembly.MainModule.Resources.Count);
}
```

Documentation/guides/OneDotNetEmbeddedResources.md
```diff
@@ -0,0 +1,220 @@
+# [proposal] Embedded Resources and `.nupkg` Files
```
jonpryor, line 1: `[proposal]` isn't needed when it's actually implemented. :-)

Documentation/proposals/OneDotNetEmbeddedResources.md
```diff
@@ -1,216 +1,7 @@
 # [proposal] Embedded Resources and `.nupkg` Files
```
jonpryor, line 1: I think this file can be *removed*; that appears to be the practice in e.g. dotnet/csharplang, e.g. https://github.com/dotnet/csharplang/commit/c3df20406f43fcd460cfedd1cd61b6cc47d27250#diff-c81858101b134a23eb7a0df15e102f4d (which contains a `git mv` from `/proposals` to `/proposals/csharp-9.0`).

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
```diff
@@ -0,0 +1,115 @@
...
+  </Target>
+
+  <Target Name="_ResolveAars" Condition=" '$(AndroidApplication)' == 'true' ">
+    <ItemGroup>
+      <_AarSearchDirectory   Include="@(_ReferencePath->'%(RootDir)%(Directory)')" />
+      <_AarSearchDirectory   Include="@(_ReferenceDependencyPaths->'%(RootDir)%(Directory)')" />
+      <_AarDistinctDirectory Include="@(_AarSearchDirectory->Distinct())" />
+      <!-- NOTE: %(AndroidSkipResourceProcessing) is required for located .aar's; there could be custom views -->
+      <AndroidAarLibrary     Update="@(AndroidAarLibrary)"                    AndroidSkipResourceProcessing="true" />
+      <AndroidAarLibrary     Include="%(_AarDistinctDirectory.Identity)*.aar" AndroidSkipResourceProcessing="false" />
```
dellis1972, line 53: Is it possible we will end up with two of the same item here? One the user manually included via `AndroidAarLibrary ` and one we pick up from this `*.aar` ? What will happen to the metadata in that case? Do we need to take that into account?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
```diff
@@ -188,39 +188,35 @@ public void IncrementalCleanDuringClean ()
...
-					class1Source,
+					new BuildItem.Source ("Class1.cs") {
+						TextContent = () => "public class Class1 { }",
+					}
 				},
 			};
-			using (var b = CreateDllBuilder (Path.Combine (testPath, "Lib"))) {
-				Assert.IsTrue (b.Build (lib), "Build should have succeeded.");
-				Assert.IsTrue (b.LastBuildOutput.ContainsText ("LogicalName=__AndroidLibraryProjects__.zip") ||
+			using (var b = CreateDllBuilder ()) {
```
dellis1972, line 201: Do we have a test which includes a third level library reference? i.e A->B->C where B and C both have `.aar` files. What are we expecting in that case? the `.aar` files should all end up in the the `A` intermediate path. 
Or are we expecting the users to do A->B->C, A->C to get this to work ?

jonathanpeppers: I updated a test to have an App A that reference Lib B, Lib B -> Lib C.

It appears to already work without changing anythinig:

image

But the test now asserts that the app has an `@(AndroidAsset)` from Lib C without actually referencing it, which is a good thing to test.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
jonathanpeppers: I'm still thinking about this one, let me test it.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
dellis1972: nice :)

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
jonathanpeppers: I tested putting the same path as this `_ResolveAars` target will find:
```
<AndroidAarLibrary Include="..\LibraryB\bin\Debug\net5.0-android\LibraryB.aar"  />
```
The build succeeds even with an `.aar` listed twice, but yes it would set `AndroidSkipResourceProcessing=false`:

image

The usage of the `.aar` seems OK? It seems to reuse the same `2` directory here:

image

dellis1972: So does it then Convert the Resource cases for the duplicate aar? I'm thinking about targets that should not be run because we  did set `AndroidSkipResourceProcessing=True` but it was overridden..

jonathanpeppers: I updated the `DotNetBuildLibrary` test to check this now.

I think the new logic is better, as it will handle existing `.aar` files:

```xml
<_AarFromLibraries     Include="%(_AarDistinctDirectory.Identity)*.aar" />
<AndroidAarLibrary     Include="@(_AarFromLibraries)" Exclude="@(AndroidAarLibrary)" />
<!--
  NOTE: %(AndroidSkipResourceProcessing) is required for located .aar's; there could be custom views.
  Update in a second step, in case there is was an existing @(AndroidAarLibrary) with the same path.
-->
<AndroidAarLibrary Update="@(_AarFromLibraries)" AndroidSkipResourceProcessing="false" />
```

Seems to be working:

image

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tasks/CreateAar.cs
```diff
@@ -0,0 +1,106 @@
...
+						}
+						var directory = Path.GetDirectoryName (resource.ItemSpec);
+						var archivePath = "res/" + Path.GetFileName (directory) + "/" + Path.GetFileName (resource.ItemSpec);
+						aar.AddStream (File.OpenRead (resource.ItemSpec), archivePath);
+						existingEntries.Remove (archivePath);
+					}
+				}
+				if (AndroidEnvironment != null) {
+					foreach (var env in AndroidEnvironment) {
+						var archivePath = ".netenv/" + Path.GetFileName (env.ItemSpec);
```
jonpryor, line 68: While this is presumably not "standard practice", nothing prevents someone from using `@(AndroidEnvironment)` on a file in a directory:

```xml
<AndroidEnvironment Include="Env.txt" />
<AndroidEnvironment Include="Dir1/Env.txt" />
<AndroidEnvironment Include="Dir2/Env.txt" />
```

We thus probably need to "hash" the filenames, much as we do assembly names & namespaces to compute JCW package names, so that we don't have inadvertent collisions.  Or throw the directory name in there, somehow.

src/Xamarin.Android.Build.Tasks/Tasks/CreateAar.cs
```diff
@@ -0,0 +1,106 @@
...
+						aar.AddStream (File.OpenRead (env.ItemSpec), archivePath);
+						existingEntries.Remove (archivePath);
+					}
+				}
+				if (JarFiles != null) {
+					foreach (var jar in JarFiles) {
```
jonpryor, line 74: I fear a similar scenario here, too: would it be "unreasonable" for `@(EmbeddedJarLibrary)` to use different directories for the "same" filename?

src/Xamarin.Android.Build.Tasks/Tasks/ExtractJarsFromAar.cs
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		static void Extract (ZipEntry entry, MemoryStream stream, string destination)
+		{
+			stream.SetLength (0); //Reuse the stream
+			entry.Extract (stream);
+			try {
+				MonoAndroidHelper.CopyIfStreamChanged (stream, destination);
+			} catch (PathTooLongException) {
+				throw new PathTooLongException ($"Could not extract \"{entry.FullName}\" to \"{destination}\". Path is too long.");
```
jonpryor, line 76: Why are we wrapping a `PathToLongException` with a `PathToLongException`? Assuming we need to, shouldn't we "preserve" the nested exception?

Furthermore, can't this possibly wind up in build error output?  Does this message need to be localized?

jonathanpeppers: This was copied from another task, but yeah it seems like I should just let this throw.

src/Xamarin.Android.Build.Tasks/Tasks/CreateAar.cs
jonathanpeppers: I changed this to:
```csharp
var archivePath = $".netenv/{Files.HashString (env.ItemSpec)}.env";
```

jonathanpeppers: I also had to replace `\` with `/`, so we should get the same hash values on all platforms.

PR #5135, opened 2020-09-18
[One .NET ] consolidate linker MSBuild properties

jonathanpeppers:
Context: https://docs.microsoft.com/dotnet/core/deploying/trimming-options

This sets up sensible defaults for the linker while supporting the
legacy `$(AndroidLinkMode)` MSBuild property.

For `Release` builds by default:

* `PublishTrimmed=true`
* `TrimMode=link`

I documented these settings in `OneDotNet.md`.

This may cause some tests to fail because we are using `TrimMode=Link`, we'll see what happens.

/cc @rolfbjarne

### DIFF
--- a/Documentation/guides/OneDotNet.md
+++ b/Documentation/guides/OneDotNet.md
@@ -135,6 +135,38 @@ Default Android related file globbing behavior is defined in `Microsoft.Android.
 This behavior can be disabled for Android items by setting `$(EnableDefaultAndroidItems)` to `false`, or
 all default item inclusion behavior can be disabled by setting `$(EnableDefaultItems)` to `false`.
 
+## Linker (ILLink)
+
+.NET 5 and higher has new [settings for the linker][linker]:
+
+* `<PublishTrimmed>true</PublishTrimmed>`
+* `<TrimMode>copyused</TrimMode>` - Enable assembly-level trimming.
+* `<TrimMode>link</TrimMode>` - Enable member-level trimming.
+
+In Android application projects by default, `Debug` builds will not
+use the linker and `Release` builds will set `PublishTrimmed=true` and
+`TrimMode=link`. `TrimMode=copyused` is the default the dotnet/sdk,
+but it doesn't not seem to be appropriate for mobile applications.
+Developers can still opt into `TrimMode=copyused` if needed.
+
+If the legacy `AndroidLinkMode` setting is used, both `SdkOnly` and
+`Full` will default to equivalent linker settings:
+
+* `<PublishTrimmed>true</PublishTrimmed>`
+* `<TrimMode>link</TrimMode>`
+
+With `AndroidLinkMode=SdkOnly` only BCL and SDK assemblies marked with
+`%(Trimmable)` will be linked at the member level.
+`AndroidLinkMode=Full` will set `%(TrimMode)=link` on all .NET
+assemblies similar to the example in the [trimming
+documentation][linker-full].
+
+It is recommended to migrate to the new linker settings, as
+`AndroidLinkMode` will eventually be deprecated.
+
+[linker]: https://docs.microsoft.com/dotnet/core/deploying/trimming-options
+[linker-full]: https://docs.microsoft.com/dotnet/core/deploying/trimming-options#trimmed-assemblies
+
 ## dotnet cli
 
 There are currently a few "verbs" we are aiming to get working in
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.props
@@ -30,7 +30,6 @@
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
     <AndroidUseSharedRuntime Condition=" '$(AndroidUseSharedRuntime)' == '' ">false</AndroidUseSharedRuntime>
     <EmbedAssembliesIntoApk Condition=" '$(EmbedAssembliesIntoApk)' == '' ">true</EmbedAssembliesIntoApk>
-    <AndroidLinkMode Condition=" '$(AndroidLinkMode)' == '' ">SdkOnly</AndroidLinkMode>
     <AndroidManagedSymbols Condition=" '$(AndroidManagedSymbols)' == '' ">true</AndroidManagedSymbols>
   </PropertyGroup>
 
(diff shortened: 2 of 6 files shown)


Documentation/guides/OneDotNet.md
```diff
@@ -135,6 +135,36 @@ Default Android related file globbing behavior is defined in `Microsoft.Android.
...
+
+If the legacy `AndroidLinkMode` setting is used, both `SdkOnly` and
+`Full` will default to equivalent linker settings:
+
+* `<PublishTrimmed>true</PublishTrimmed>`
+* `<TrimMode>link</TrimMode>`
+
+With these settings only BCL and SDK assemblies marked with
+`%(Trimmable)` will be linked at the member level. There is not an
+exact equivalent of `AndroidLinkMode=Full`, but a custom MSBuild
```
rolfbjarne, line 158: I came up with this: https://github.com/xamarin/xamarin-macios/blob/1463967d003ad594c140489388fb1f906a0a9565/dotnet/targets/Xamarin.Shared.Sdk.targets#L208-L214 (which can probably be improved to only set TrimMode for assemblies that hasn't set it already) , because it's quite common for our customers to want to use Full to link away as much as possible (i.e. link every single assembly). Apple has app size limits which can be quite annoying... and a common first approach is to link everything and hope both that the app is small enough for Apple and continues to work (and it works often enough)

dellis1972 (approved):

Documentation/guides/OneDotNet.md
jonathanpeppers: I was thinking we would deprecate `$(AndroidLinkMode)` eventually, but I'll change our `.targets` to do this for now.

There isn't even a UI to setup most of these things in .NET 5 yet, you only get a checkbox that does `PublishTrimmed=true`:

image

rolfbjarne: That UI will have to change for Xamarin...

jonathanpeppers:
A few tests failed with:
```
ILLink : error IL1020: No files to link were specified. Use one of '-a|-r|-x' options
```
Will look into it.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
```diff
@@ -13,6 +13,21 @@ This file contains the .NET 5-specific targets to customize ILLink
...
+          TrimMode="link"
+      />
+      <!-- Linking $(TargetFileName) causes: error IL1020: No files to link were specified. Use one of '-a|-r|-x' options -->
+      <ResolvedFileToPublish
+          Update="@(ResolvedFileToPublish)"
+          Condition=" '%(ResolvedFileToPublish.Filename)%(ResolvedFileToPublish.Extension)' == '$(TargetFileName)' "
+          IsTrimmable="false"
+          TrimMode=""
+      />
+    </ItemGroup>
```
jonathanpeppers, line 30: @rolfbjarne @radekdoulik I had to exclude the main assembly for `AndroidLinkMode=full` or I was hitting:
```
error IL1020: No files to link were specified. Use one of '-a|-r|-x' options
```
I think it is saying there is no "root assembly", so the error message isn't quite right?

rolfbjarne: Yes, the error message isn't quite right.

This is what I did: https://github.com/xamarin/xamarin-macios/blob/e72ef18bebad34e4c26500bfed5d50410437fb80/dotnet/targets/Xamarin.Shared.Sdk.targets#L216-L217

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.props
```diff
@@ -43,7 +42,10 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(AndroidApplication)' == 'true' ">
     <SelfContained Condition=" '$(SelfContained)' == '' ">true</SelfContained>
-    <PublishTrimmed Condition=" '$(AndroidLinkMode)' == 'SdkOnly' or '$(AndroidLinkMode)' == 'Full' ">true</PublishTrimmed>
+    <PublishTrimmed Condition=" '$(PublishTrimmed)' == '' and ('$(AndroidLinkMode)' == 'SdkOnly' or '$(AndroidLinkMode)' == 'Full') ">true</PublishTrimmed>
+    <PublishTrimmed Condition=" '$(PublishTrimmed)' == '' and '$(Configuration)' == 'Release' and '$(AndroidLinkMode)' != 'None' ">true</PublishTrimmed>
+    <TrimMode Condition=" '$(TrimMode)' == '' and ('$(AndroidLinkMode)' == 'SdkOnly' or '$(AndroidLinkMode)' == 'Full') ">link</TrimMode>
```
radekdoulik, line 47: Shouldn't the `TrimMode` be set after the next line, where we set `AndroidLinkMode`?

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
radekdoulik: Yes, `TrimmerRootAssembly ` seems to be the way to specify the [root assembly](https://docs.microsoft.com/en-us/dotnet/core/deploying/trimming-options#root-assemblies).

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.props
jonathanpeppers: Right now this only sets `$(TrimMode)` if the user set `$(AndroidLinkMode)` in their `.csproj` file. It's using the dotnet/sdk default of `CopyUsed`, which maybe is not what we want for mobile?

@rolfbjarne are you using `TrimMode=Link` by default for `Release` builds?

rolfbjarne: @jonathanpeppers the default will most likely be `TrimMode=Link` for device builds (both debug and release), and `TrimMode=Copy` for simulator builds

jonathanpeppers: I updated the code and documentation, so we are defaulting to `TrimMode=link` for `Release` builds now.

I think this setting makes more sense for mobile apps.

radekdoulik (approved):

rolfbjarne (approved):

Documentation/guides/OneDotNet.md
```diff
@@ -135,6 +135,38 @@ Default Android related file globbing behavior is defined in `Microsoft.Android.
...
+* `<TrimMode>link</TrimMode>`
+
+With `AndroidLinkMode=SdkOnly` only BCL and SDK assemblies marked with
+`%(Trimmable)` will be linked at the member level.
+`AndroidLinkMode=Full` will set `%(TrimMode)=link` on all .NET
+assemblies similar to the example in the [trimming
+documentation][linker-full].
+
+It is recommended to migrate to the new linker settings, as
+`AndroidLinkMode` will eventually be deprecated.
```
jonpryor, line 165: If this will eventually be deprecated, we need to start emitting a deprecation warning around its use.

Which just raises a different question: is there/can there be a mechanism that is usable both in the Current World Order *and* in .NET 6?  Or *is* `$(AndroidLinkMode)` that mechanism, and you're suggesting that support for `$(AndroidLinkMode)` need not be removed until "later" (.NET 7? 8?).

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
```diff
@@ -111,7 +115,7 @@ public void CheckIncludedAssemblies ()
 						proj.IntermediateOutputPath, "android", "bin", "UnnamedProject.UnnamedProject.apk");
 				using (var zip = ZipHelper.OpenZip (apk)) {
 					var existingFiles = zip.Where (a => a.FullName.StartsWith ("assemblies/", StringComparison.InvariantCultureIgnoreCase));
-					var missingFiles = expectedFiles.Where (x => !zip.ContainsEntry ("assmelbies/" + Path.GetFileName (x)));
+					var missingFiles = expectedFiles.Where (x => !zip.ContainsEntry ("assemblies/" + Path.GetFileName (x)));
```
jonpryor, line 118: Always "glorious" to find a "meaningless" test… :-(

PR #5137, opened 2020-09-21
[tests] Fix process output collection

radekdoulik:
The `BaseTest.RunProcess` is redirecting the standard and error output,
so that it can be captured. It didn't work, because it was not using the
output related events and so if there was enough output, we got dead
lock and timeout as a result.

Context: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.standarderror

Example behaviour before the fix - notice the timeout:

    RunProcess: C:\Users\rodo\android-toolchain\sdk\platform-tools\adb.exe  logcat -d > "C:\Users\rodo\git\xa-master\bin\TestDebug\temp\otNetInstallAndRunTrueFalse\logcat-failed.log"
    RunProcess timed out: C:\Users\rodo\android-toolchain\sdk\platform-tools\adb.exe  logcat -d > "C:\Users\rodo\git\xa-master\bin\TestDbug\temp\DotNetInstallAndRunTrueFalse\logcat-failed.log"

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
@@ -238,7 +238,7 @@ namespace Xamarin.Android.Build.Tests
 				});
 				proc.ErrorDataReceived += new DataReceivedEventHandler ((sender, e) => {
 					if (!string.IsNullOrEmpty (e.Data))
-						errorOutput.Append (Environment.NewLine + e.Data);
+						errorOutput.AppendLine (e.Data);
 				});
 
 				proc.Start ();
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
@@ -234,7 +234,7 @@ namespace Xamarin.Android.Build.Tests
 				proc.StartInfo = info;
 				proc.OutputDataReceived += new DataReceivedEventHandler ((sender, e) => {
 					if (!string.IsNullOrEmpty (e.Data))
-						standardOutput.Append (Environment.NewLine + e.Data);
+						standardOutput.AppendLine (e.Data);
 				});
 				proc.ErrorDataReceived += new DataReceivedEventHandler ((sender, e) => {
 					if (!string.IsNullOrEmpty (e.Data))
@@ -560,4 +560,3 @@ namespace Xamarin.Android.Build.Tests
 		}
 	}
 }
-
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
```diff
@@ -229,14 +229,31 @@ protected static string RunProcess (string exe, string args)
...
+
 				if (!proc.WaitForExit ((int)TimeSpan.FromSeconds (30).TotalMilliseconds)) {
 					proc.Kill ();
 					TestContext.Out.WriteLine ($"{nameof (RunProcess)} timed out: {exe} {args}");
 					return null; //Don't try to read stdout/stderr
 				}
-				var result = proc.StandardOutput.ReadToEnd ().Trim () + proc.StandardError.ReadToEnd ().Trim ();
-				return result;
+
+				proc.WaitForExit ();
```
jonathanpeppers, line 254: Do we need the second `WaitForExit()`? If it got here, the first one returned `true`?

dellis1972: I agree awe shouldn't need this.

radekdoulik: Actually we do, from the documentation for WaitForExit(int):
_When standard output has been redirected to asynchronous event handlers, it is possible that output processing will not have completed when this method returns. To ensure that asynchronous event handling has been completed, call the WaitForExit() overload that takes no parameter after receiving a true from this overload._ (https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.waitforexit?view=netframework-4.7.2#System_Diagnostics_Process_WaitForExit_System_Int32_)

radekdoulik:
Also note that this should speedup the CI builds, as we will not be timeouting as before, saving upto 30 seconds per affected test :-)

jonathanpeppers (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
```diff
@@ -229,14 +229,31 @@ protected static string RunProcess (string exe, string args)
 				CreateNoWindow = true,
 				WindowStyle = ProcessWindowStyle.Hidden,
 			};
-			using (var proc = Process.Start (info)) {
+			using (var proc = new Process ()) {
+				StringBuilder standardOutput = new StringBuilder (), errorOutput = new StringBuilder ();
+				proc.StartInfo = info;
+				proc.OutputDataReceived += new DataReceivedEventHandler ((sender, e) => {
+					if (!string.IsNullOrEmpty (e.Data))
+						standardOutput.Append (Environment.NewLine + e.Data);
```
jonpryor, line 237: Why place `Environment.NewLine` *first*?  This will introduce an "empty line" at the *beginning* of output, which might look "weird".

This also results in extra string temporaries.  This would likely be "nicer" to the GC:

```csharp
standardOutput.Append (e.Data).Append (Environment.NewLine);
```

jonathanpeppers: There is also an `AppendLine` method:

https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.appendline?view=netframework-4.7.2

```suggestion
						standardOutput.AppendLine (e.Data);
```

PR #5138, opened 2020-09-21
[tests] Fix logcat output saving

radekdoulik:
The `RunAdbCommand` is not running a shell, so we cannot use output
redirection like:

    RunAdbCommand ($"logcat -d > \"{deviceLog}\"");

So instead just write the output to the file ourselves.

Before the output of the command was:

    Invalid filter expression 'C:\...\bin\TestDebug\temp\DotNetInstallAndRunTrueFalse\logcat-failed.log'
    Usage: logcat [options] [filterspecs]
    options include:
      -s              Set default filter to silent. Equivalent to filterspec '*:S'
      -f <file>, --file=<file>               Log to file. Default is stdout
      -r <kbytes>, --rotate-kbytes=<kbytes>
                      Rotate log every kbytes. Requires -f option
    ...

So we didn't get the logcat output we wanted.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
@@ -29,7 +29,8 @@ namespace Xamarin.Android.Build.Tests
 				string deviceLog = Path.Combine (outputDir, "logcat-failed.log");
 				string remote = "/data/local/tmp/screenshot.png";
 				RunAdbCommand ($"shell screencap {remote}");
-				RunAdbCommand ($"logcat -d > \"{deviceLog}\"");
+				var output = RunAdbCommand ($"logcat -d");
+				File.WriteAllText (deviceLog, output);
 				RunAdbCommand ($"pull {remote} \"{local}\"");
 				RunAdbCommand ($"shell rm {remote}");
 				if (File.Exists (local)) {


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
```diff
@@ -29,7 +29,8 @@ protected override void CleanupTest ()
 				string deviceLog = Path.Combine (outputDir, "logcat-failed.log");
 				string remote = "/data/local/tmp/screenshot.png";
 				RunAdbCommand ($"shell screencap {remote}");
-				RunAdbCommand ($"logcat -d > \"{deviceLog}\"");
+				var output = RunAdbCommand ($"logcat -d");
+				File.WriteAllText (deviceLog, output);
```
jonathanpeppers, line 33: Since this could be a 4MB string:

https://github.com/xamarin/xamarin-android/blob/c98cb40f35a208c04466bb3fccb3cd25d463781c/build-tools/scripts/TestApks.targets#L94-L100

Do you think it's worth making a `GetLogcatOutput` method that would write each line to a file at a time? Instead of making a giant string?

radekdoulik: For tests code I don't care much.

If we really want this, we should do it by extending the `BaseTest.RunCommand` to be able to write the output to the file instead of returning the string with the outputs and let it pass the file path through the `RunAdbCommand `.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
```diff
@@ -29,7 +29,8 @@ protected override void CleanupTest ()
 				string deviceLog = Path.Combine (outputDir, "logcat-failed.log");
 				string remote = "/data/local/tmp/screenshot.png";
 				RunAdbCommand ($"shell screencap {remote}");
-				RunAdbCommand ($"logcat -d > \"{deviceLog}\"");
+				var output = RunAdbCommand ($"logcat -d");
```
jonpryor, line 32: We should use `$"logcat -d -f \"{deviceLog}\""`:

```bash
$ adb logcat --help
Usage: logcat [options] [filterspecs]
options include:
  -s              Set default filter to silent. Equivalent to filterspec '*:S'
  -f <file>, --file=<file>               Log to file. Default is stdout
…
```

jonpryor:
Should `DeviceTest.cs` should "take a page" out of `<StartAndroidEmulator/>` and co: instead of `adb logcat -d`, which grabs current logcat then exits, should we instead start `adb logcat` "in the background", letting it write to a file/etc., and then when the test is complete quit the `adb` process and backup the captured logcat?  This way if we get more than 4MB of output, we can capture *all* of it, instead of just the last 4MB.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
radekdoulik: That would not work as `-f` option writes the file to the device and not to the host. Have been there :-)

radekdoulik:
> Should DeviceTest.cs should "take a page" out of <StartAndroidEmulator/> and co: instead of adb logcat -d, which grabs current logcat then exits, should we instead start adb logcat "in the background", letting it write to a file/etc., and then when the test is complete quit the adb process and backup the captured logcat? This way if we get more than 4MB of output, we can capture all of it, instead of just the last 4MB.

Maybe in future? Today I was just frustrated by not getting the adb output at all and finally noticed where the issue was (I ran into that multiple times before - always wondering why there's no adb logcat output on CI, when on-device tests failed). I agree it can be improved a lot, this PR is to quickly fix the problem.

BTW, the logcat I was working with today was ~90 lines long.

jonpryor:
I'm not sure what to make of the on-device tests timing out: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4082213&view=logs&j=4348ab7c-72f0-52b0-44fd-6f17a16a558f&t=8d5650f4-ac5b-518d-79b1-c2948ac0f57f

Re-running the tests; hopefully they'll finish this time?

PR #5142, opened 2020-09-22
[One .NET] cleanup build output

jonathanpeppers:
I noticed build output for `HelloAndroid.csproj` emits a collection of
junk files & directories that we do not need:

    > ls -r -n .\HelloAndroid\bin\Debug\net5.0-android\
    android.21-arm64
    android.21-x86
    ref
    com.microsoft.net5.helloandroid-Signed.apk
    com.microsoft.net5.helloandroid.apk
    HelloAndroid.deps.json
    HelloAndroid.dll
    HelloAndroid.pdb
    android.21-arm64\publish
    android.21-arm64\ref
    android.21-x86\publish
    android.21-x86\ref
    ref\HelloAndroid.dll

First issue to fix, is we shouldn't set `$(ProduceReferenceAssembly)`
for application projects. Applications shouldn't reference other
applications, so there isn't a need for `$(ProduceReferenceAssembly)`
to be `true` by default.

Xamarin.Android class libraries can have continue to set
`ProduceReferenceAssembly=true` by default as provided by the
dotnet/sdk:

https://github.com/dotnet/sdk/blob/dec3c441c2882dc7880e20e42835ce420b8d13f8/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.TargetFrameworkInference.targets#L236-L238

This gets us down to:

    > ls -r -n .\HelloAndroid\bin\Debug\net5.0-android\
    android.21-arm64
    android.21-x86
    com.microsoft.net5.helloandroid-Signed.apk
    com.microsoft.net5.helloandroid.apk
    HelloAndroid.deps.json
    HelloAndroid.dll
    HelloAndroid.pdb
    android.21-arm64\publish
    android.21-x86\publish

Next, I found that the `PrepareForPublish` MSBuild target creates
directories with no obvious way to prevent it:

https://github.com/dotnet/sdk/blob/44f381b62d466565639d51847c9127afbe7062a9/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Publish.targets#L102-L107

The solution here is to set `$(PublishDir)` to `$(OutputPath)` if we
are running a regular build. `dotnet publish` can continue to use a
custom `$(PublishDir)` if desired.

This gets us to:

    > ls -r -n .\HelloAndroid\bin\Debug\net5.0-android\
    android.21-arm64
    android.21-x86
    com.microsoft.net5.helloandroid-Signed.apk
    com.microsoft.net5.helloandroid.apk
    HelloAndroid.deps.json
    HelloAndroid.dll
    HelloAndroid.pdb

This project has `RuntimeIdentifiers=android.21-arm64;android.21-x86`
and so an inner build per RID is creating directories.

We can pass through `$(_OuterOutputPath)` and set `$(OutputPath)` and
`$(OutDir)` on the inner build running the
`_ComputeFilesToPublishForRuntimeIdentifiers` target.

Almost there! What is `HelloAndroid.deps.json`?

    > ls -r -n .\HelloAndroid\bin\Debug\net5.0-android\
    com.microsoft.net5.helloandroid-Signed.apk
    com.microsoft.net5.helloandroid.apk
    HelloAndroid.deps.json
    HelloAndroid.dll
    HelloAndroid.pdb

`HelloAndroid.deps.json` would only be needed if you wanted to
`dotnet  run HelloAndroid.dll`. This wouldn't make any sense to do this,
because you wouldn't be able to run an Android application on the host
machine.

Finally, if we set `GenerateDependencyFile=false` by default we can
get rid of `HelloAndroid.deps.json`!

    > ls -r -n .\HelloAndroid\bin\Debug\net5.0-android\
    com.microsoft.net5.helloandroid-Signed.apk
    com.microsoft.net5.helloandroid.apk
    HelloAndroid.dll
    HelloAndroid.pdb

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.props
@@ -13,6 +13,7 @@
     <AndroidEnableSGenConcurrent Condition=" '$(AndroidEnableSGenConcurrent)' == '' ">true</AndroidEnableSGenConcurrent>
     <AndroidHttpClientHandlerType Condition=" '$(AndroidHttpClientHandlerType)' == '' ">Xamarin.Android.Net.AndroidClientHandler</AndroidHttpClientHandlerType>
     <AndroidUseIntermediateDesignerFile Condition=" '$(AndroidUseIntermediateDesignerFile)' == '' ">true</AndroidUseIntermediateDesignerFile>
+    <GenerateDependencyFile Condition=" '$(GenerateDependencyFile)' == '' ">false</GenerateDependencyFile>
     <!-- jar2xml is not supported -->
     <AndroidClassParser>class-parse</AndroidClassParser>
     <!-- mono-symbolicate is not supported -->
@@ -42,6 +43,7 @@
     <OutputType Condition=" '$(OutputType)' == 'Exe' ">Library</OutputType>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(AndroidApplication)' == 'true' ">
+    <ProduceReferenceAssembly Condition=" '$(ProduceReferenceAssembly)' == '' ">false</ProduceReferenceAssembly>
     <SelfContained Condition=" '$(SelfContained)' == '' ">true</SelfContained>
     <PublishTrimmed Condition=" '$(AndroidLinkMode)' == 'SdkOnly' or '$(AndroidLinkMode)' == 'Full' ">true</PublishTrimmed>
     <!-- Prefer $(RuntimeIdentifiers) plural -->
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.props
```diff
@@ -42,6 +43,7 @@
     <OutputType Condition=" '$(OutputType)' == 'Exe' ">Library</OutputType>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(AndroidApplication)' == 'true' ">
+    <ProduceReferenceAssembly Condition=" '$(ProduceReferenceAssembly)' == '' ">false</ProduceReferenceAssembly>
```
jonathanpeppers, line 46: @rolfbjarne I don't know if you are seeing lots of extra files for iOS projects, but you might consider `GenerateDependencyFile=false` by default and `ProduceReferenceAssembly=false` by default for iOS application projects.

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.props
rolfbjarne: Good idea, I had `GenerateRuntimeConfigurationFiles=false` set somewhere too at some point, which iirc gets rid of *.runtimeconfig.json

Issue #5143, opened 2020-09-22
Use preserve lists with illink/net5


radekdoulik:
We should pack the linker preserve lists into our DotNet packages, next to the Sdk assemblies, and let illink use them.

PR #5145, opened 2020-09-23
[illink] Pack and use linker preservation files

radekdoulik:
Implements: https://github.com/xamarin/xamarin-android/issues/5143
Fixes: https://github.com/xamarin/xamarin-android/issues/5166

Use the `@TrimmerRootDescriptor` items to let illink use the preservation files.

The xml preservation files are "prefiltered" to include only those, which are relevant (have the same filename without extension, `Mono.Android` for example) to the referenced assemblies.

Example build log:

    ...
                   Added Item(s): 
                       _PreserveLists=
                           C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.135\targets\..\PreserveLists\Java.Interop.xml
                           C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.135\targets\..\PreserveLists\Mono.Android.xml
                           C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.135\targets\..\PreserveLists\mscorlib.xml
                           C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.135\targets\..\PreserveLists\OpenTK.xml
                           C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.135\targets\..\PreserveLists\System.Data.Services.Client.xml
                           C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.135\targets\..\PreserveLists\System.ServiceModel.xml
                   Added Item(s): TrimmerRootDescriptor=C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.135\targets\..\PreserveLists\Java.Interop.xml
                   Added Item(s): TrimmerRootDescriptor=C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.135\targets\..\PreserveLists\Mono.Android.xml
                   Added Item(s): TrimmerRootDescriptor=C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.135\targets\..\PreserveLists\mscorlib.xml
    Done building target "_PrepareLinking" in project "UnnamedProject.csproj".: (TargetId:251)


radekdoulik:
Locally I see this illink crash:
```
14:53:19.042   1:8>ILLink : error IL1009: Assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' reference 'System.Security.Permissions, Version=0.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' could not be resolved [C:\Users\rodo\git\xa-master\bin\TestDebug\temp\DotNetBuildRodoTrue\UnnamedProject.csproj]
                     Mono.Linker.LinkerFatalErrorException: ILLink: error IL1009: Assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089' reference 'System.Security.Permissions, Version=0.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' could not be resolved (TaskId:194)
                      ---> Mono.Cecil.AssemblyResolutionException: Failed to resolve assembly: 'System.Security.Permissions, Version=0.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' (TaskId:194)
                      ---> System.IO.FileNotFoundException: Unable to find 'System.Security.Permissions.dll' or 'System.Security.Permissions.exe' file (TaskId:194)
                        at Mono.Linker.DirectoryAssemblyResolver.Resolve(AssemblyNameReference name, ReaderParameters parameters) (TaskId:194)
                        at Mono.Linker.AssemblyResolver.Resolve(AssemblyNameReference name, ReaderParameters parameters) (TaskId:194)
                        at Mono.Linker.LinkContext.Resolve(IMetadataScope scope) (TaskId:194)
                        at Mono.Linker.LinkContext.ResolveReferences(AssemblyDefinition assembly) (TaskId:194)
                        --- End of inner exception stack trace --- (TaskId:194)
                        at Mono.Linker.LinkContext.ResolveReferences(AssemblyDefinition assembly) (TaskId:194)
                        at Mono.Linker.Steps.LoadReferencesStep.ProcessReferences(AssemblyDefinition assembly) (TaskId:194)
                        at Mono.Linker.Steps.ResolveFromXmlStep.GetAssembly(LinkContext context, AssemblyNameReference assemblyName) (TaskId:194)
                        at Mono.Linker.Steps.ProcessLinkerXmlStepBase.ProcessAssemblies(XPathNodeIterator iterator) (TaskId:194)
                        at Mono.Linker.Steps.ProcessLinkerXmlStepBase.ProcessXml(Boolean stripResource, Boolean ignoreResource) (TaskId:194)
                        at Mono.Linker.Steps.ResolveFromXmlStep.Process() (TaskId:194)
                        at Mono.Linker.Steps.BaseStep.Process(LinkContext context) (TaskId:194)
                        at Mono.Linker.Pipeline.ProcessStep(LinkContext context, IStep step) (TaskId:194)
                        at Mono.Linker.Pipeline.Process(LinkContext context) (TaskId:194)
                        at Mono.Linker.Driver.Run(ILogger customLogger) (TaskId:194)
```
@jonathanpeppers do you know, why we might be missing `System.Security.Permissions` reference?

jonathanpeppers:
@radekdoulik I found if you reference AndroidX, Support libraries, or Xamarin.Forms, you need these:

https://github.com/xamarin/xamarin-android/blob/e113bc1329a02a1e16d0c4e4cfab544f429929ab/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/KnownPackages.cs#L7-L19

Here is a list that would be easy to copy-paste:

https://github.com/xamarin/net6-samples/blob/6f349cc915f66e6d15fb00b7333e48c00f95e6fb/HelloForms.Droid/HelloForms.Droid.csproj#L17-L22

radekdoulik:
@jonathanpeppers thanks!

So we still need these. I forgot about them already.

It looks like we might need them for all projects now, with proposed linker changes, as I was using modified `DotNetBuild` test from `XASdkTests.cs`.

I wonder whether `System.Security.Permissions` should be included in `microsoft.netcore.app.runtime.android-*` packages, as the `mscorlib` in them forwards to `System.Security.Permissions` assembly.

jonathanpeppers:
I'm not seeing what the actual resolution was for:

https://github.com/mono/linker/issues/1139

It seems like if `mscorlib.dll` has a reference to it, it has to be missing?

radekdoulik:
> It seems like if mscorlib.dll has a reference to it, it has to be missing?

It is forwardning some types to it.

I talked with @marek-safar  about that and mscorlib is a legacy assembly in net5. The `System.Security.Permissions` is not part of `microsoft.netcore.app.runtime.android-*` nugets by design.

So I think we should try to add the compat package references to projects which reference mscorlib? And don't reference mscorlib in the new projects.

marek-safar:
> I found if you reference AndroidX, Support libraries, or Xamarin.Forms, you need these:

Only if they don't have .NET Core version and are run in .NET Framework compatibility mode.

jonathanpeppers:
> Only if they don't have .NET Core version and are run in .NET Framework compatibility mode.

@marek-safar just so you know, this is every existing Xamarin NuGet package. So we need something to keep these packages working at least in the beginning for migration purposes.

If we add something to automatically add `@(PackageReference)` for System.Security.Permissions, etc. should that code live in dotnet/sdk or the Xamarin SDKs?

marek-safar:
> If we add something to automatically add @(PackageReference) for System.Security.Permissions, etc. should that code live in dotnet/sdk or the Xamarin SDKs?

I see this as a measure required for incompletely migrated projects only. We already discussed that with core sdk teams and it won't be included there (for good reasons). I think we'd need this only for very specific old projects in migration tool.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
```diff
@@ -33,6 +33,12 @@ This file contains the .NET 5-specific targets to customize ILLink
         <Type>Microsoft.Android.Sdk.ILLink.SetupStep</Type>
       </_TrimmerCustomSteps>
     </ItemGroup>
+    <ItemGroup>
+      <_PreserveLists Include="$(MSBuildThisFileDirectory)..\PreserveLists\*.xml" />
+      <TrimmerRootDescriptor
+          Condition=" '@(ResolvedFileToPublish)' != '' and '%(Filename)' != '' "
```
jonathanpeppers, line 39: I wouldn't do `'@(ResolvedFileToPublish)' != ''` as this can expand to a large string.

Maybe this works instead?
```suggestion
          Condition=" '@(ResolvedFileToPublish->Count())' != '0' and '%(Filename)' != '' "
```

radekdoulik:
I checked the `DotNetInstallAndRun` test locally, and it fails in the same way as in this issue: https://github.com/xamarin/xamarin-android/issues/5156 So it is expected to fail.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
radekdoulik: Unfortunately that doesn't work. I think it breaks the way the msbuild iterates over the two item groups and so we don't get the intersection of them. When run with above change, the resulting item group contains all the preserve files:

```
                   Added Item(s): TrimmerRootDescriptor=C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.145\targets\..\PreserveLists\Java.Interop.xml
                   Added Item(s): TrimmerRootDescriptor=C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.145\targets\..\PreserveLists\Mono.Android.xml
                   Added Item(s): TrimmerRootDescriptor=C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.145\targets\..\PreserveLists\mscorlib.xml
                   Added Item(s): TrimmerRootDescriptor=C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.145\targets\..\PreserveLists\OpenTK.xml
                   Added Item(s): TrimmerRootDescriptor=C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.145\targets\..\PreserveLists\System.Data.Services.Client.xml
                   Added Item(s): TrimmerRootDescriptor=C:\Users\rodo\git\xa-master\packages\microsoft.android.sdk\11.0.100-ci.master.145\targets\..\PreserveLists\System.ServiceModel.xml
17:31:17.746   1:8>Done building target "_PrepareLinking" in project "UnnamedProject.csproj".: (TargetId:251)
```

jonathanpeppers: What about?
```suggestion
          Condition=" '%(ResolvedFileToPublish.Filename)' != '' "
```

radekdoulik: Wait, I had the parentheses wrong :-) It works nicely, thanks for the improvement.

marek-safar:
I just realized, why are you doing this? There is a better mechanism in .NET Core to do this and what is worse this is wrong for BCL assemblies and will issues warnings for customers.

radekdoulik:
> I just realized, why are you doing this? There is a better mechanism in .NET Core to do this and what is worse this is wrong for BCL assemblies and will issues warnings for customers.

We are trying to make the illink work with XA, so we want to get to a state where the linked apps run fine. The preserve lists are the ones we are using with the original linker. I am going to review these and update some of them in the following PR.

So far I see this warning:
```
17:39:13.625   1:8>ILLink : warning IL2001: System.AppDomainSetup: Type System.AppDomainSetup has no fields to preserve [C:\Users\rodo\git\xa-master\bin\TestDebug\temp\DotNetBuildRodoTrue\UnnamedProject.csproj]
```

What is this better mechanism you mention? I might try to use it when I am looking at it next week.

marek-safar:
> The preserve lists are the ones we are using with the original linker.

That's not really accurate, the ones you are have are used with Mono BCL. Using for example `mscorlib.xml` with .NET Core makes no sense as there is no mscorlib and the names which match to other assemblies could cause conflict with existing preservation rules.

> What is this better mechanism you mention?
You need the preserve files only for assemblies you don't own which in .NET Core case should be none. 

For the ones you own like `Mono.Android.xml` and you still need a hint for linker you would be better to migrate to .NET 5 attributes in https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis?view=net-5.0

radekdoulik:
Hopefully the whole `mscorlib.xml` might go away and the part of `Mono.Android.xml` which tries to preserve mscorlib methods, which don't exist anymore.

radekdoulik:
> Why this file?

Because it was in the original preserve lists.

My current plan is to get linker to a state where the linked app doesn't crash on runtime. Part of that is to make sure we have missing pieces from old XA linker in place. Once we are there I will review things like preserve lists, usage of [Preserve] attrbibute in our code, assembly sizes, ... I keep track of these in [ILlink support](https://github.com/xamarin/xamarin-android/projects/15) project

I hope this file can be removed as the code was probably replaced meanwhile. Have to dig in the old repos to see what was it fixing and add a test for the issue if there's not existing one.

radekdoulik:
> Why this one? You are not going to inject OpenTK inside your NuGet if there is already OpenTK NuGet package, right?

Not sure about that. I thought we are still shipping own OpenTK, but it doesn't seem to be the case for net5, at least not currently. So it should be safe to remove it.

radekdoulik:
@jonpryor @jonathanpeppers do we have a plan regarding OpenTK and NET5?

marek-safar:
> about System.ServiceModel.dll

The file does not ship with .NET Core. It's external NuGet built from https://github.com/dotnet/wcf and you have no control on which version will be used in developers project. I think if that NuGet is not linker safe the fix should be in https://github.com/dotnet/wcf as with any other NuGet packages.

rolfbjarne:
> @jonpryor @jonathanpeppers do we have a plan regarding OpenTK and NET5?

This is a complex topic, and the answer depends on how much breaking changes we can do. It's probably something we should have a discussion about at some point.

Ref: https://github.com/mono/opentk/issues/19

radekdoulik:
> It's probably something we should have a discussion about at some point.

Right, as we should keep in sync with iOS/MacOS in this regard

jonpryor:
`DotNetInstallAndRun(True,True)`

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4112445&view=ms.vss-test-web.build-test-results-tab&runId=16098318&resultId=100111&paneView=attachments

```
10-01 22:37:04.410  5549  5549 F monodroid: The type `Java.Lang.Throwable` is missing required instance fields! handle=0x7cf0f7175580 handle_type=0x7cf0f71755c0 refs_added=0x7cf0f71755e0 weak_handle=0x0
```

jonpryor:
Squash-and-merge message for review:

```
Fixes: https://github.com/xamarin/xamarin-android/issues/5143
Fixes: https://github.com/xamarin/xamarin-android/issues/5166

Context: https://github.com/xamarin/xamarin-android/issues/5156

As of commit 40644d91, "One .NET" builds enable the linker, which
causes the `XASdkDeployTests.DotNetInstallAndRun(isRelease:True,…)`
unit tests to *crash on startup*:

	F monodroid: The type `Java.Lang.Throwable` is missing required instance fields! handle=0x7cf0f7175580 handle_type=0x7cf0f71755c0 refs_added=0x7cf0f71755e0 weak_handle=0x0

The cause of the crash is that the .NET 5 linker is not preserving
the `Java.Lang.Throwable.weak_handle` field (see also 130905e1?), and
the `Throwable.weak_handle` field wasn't being preserved because the
.NET 5 linker wasn't using our linker preservation list XML files.

Fix the .NET 5 linker:

 1. Add new linker preservation XML files for `Java.Inteorp.dll` &
    `Mono.Android.dll` to
    `src/Microsoft.Android.Sdk.ILLink/PreserveLists/*.xml`.

    Note that these are new files, "filtered" to contain only types
    present in `Java.Interop.dll` & `Mono.Android.dll`, unlike the
    ones in `src/Xamarin.Android.Build.Tasks/Linker/PreserveLists`.

 2. Add the linker preservation files (1) to the
    `Microsoft.Android.Sdk` NuGet package contents.

 3. Update `Microsoft.Android.Sdk.ILLink.targets` so that the linker
    preservation files are added to the `@(TrimmerRootDescriptor)`
    item group.  This item group is used by the `<ILLink/>` task
    which performs linking in a .NET 5 context.

This allows the `DotNetInstallAndRun()` test to start without printing
the `missing required instance fields` message.

***Note***: this will *not* allow the `DotNetInstallAndRun()` test to
*pass*; it merely uncovers xamarin/xamarin-android#5156!

	Unhandled Exception:
	System.ArgumentException: Cannot bind to the target method because its signature is not cmpatible with that of the delegate type.
	   at System.Delegate.CreateDelegate(Type type, Type target, String method, Boolean ignorCase, Boolean throwOnBindFailure)
	   at System.Delegate.CreateDelegate(Type type, Type target, String method)
	   at Android.Runtime.AndroidTypeManager.RegisterNativeMembers(JniType nativeClass, Type ype, String methods)
	   at Android.Runtime.JNIEnv.RegisterJniNatives(IntPtr typeName_ptr, Int32 typeName_len, ntPtr jniClass, IntPtr methods_ptr, Int32 methods_len)

This `ArgumentException` will be addressed in a later commit.
```

PR #5146, opened 2020-09-23
[build] Add VSMac workaround for %(Link) metadata

brendanzagaeski:
Context: https://developercommunity.visualstudio.com/content/problem/1073405/index.html
Context: https://github.com/mono/monodevelop/blob/fb12ccb7f4245eca20803661deaa37ead0ec35af/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs#L336-L337

In Visual Studio for Mac, `Xamarin.Android.Build.Tasks.csproj` was not
loading due to:

	Error while trying to load the project '/Users/macuser/Projects/xamarin-android/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj': Invalid value for Link property

The problem was that the `%(Link)` metadata for some of the files in the
project started with `..\`, which Visual Studio for Mac currently
rejects.  For comparison, Visual Studio on Windows accepts `%(Link)`
metadata starting with `..\` but hides the files from the Solution
Explorer.

To work around the error that happens in Visual Studio for Mac, add a
`MSBuild\..\` directory before the first `..\` in the affected `%(Link)`
metadata paths.  At the moment, this has a side effect of introducing a
confusing recursive `Xamarin.Android.Build.Tasks` item directory under
the `MSBuild` directory in the Solution pad, so also explicitly mark the
items as `%(Visible)`=`false` to prevent that.  On Windows, the
`%(Visible)`=`false` metadata means that the items stay hidden like they
were before.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
@@ -18,6 +18,7 @@
     <BuildDependsOn>
       _CopyNDKTools;
       _GenerateXACommonProps;
+      _AssignRelativeParentLinkMetadata;
       $(BuildDependsOn);
       _CopyExtractedMultiDexJar;
       _BuildMonoScripts;
@@ -31,25 +32,29 @@
   <ItemGroup>
     <None Include="MSBuild\Novell\Novell.MonoDroid.Common.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\..\Novell\Novell.MonoDroid.Common.targets</Link>
+      <!-- Work around Visual Studio for Mac `Invalid value for Link property`
+           project load error by using custom `%(_RelativeParentLink)` metadata
+           that gets copied into `%(Link)` at build time by
+           `_AssignRelativeParentLinkMetadata`. -->
+      <_RelativeParentLink>..\..\Novell\Novell.MonoDroid.Common.targets</_RelativeParentLink>
     </None>
     <None Include="MSBuild\Novell\Novell.MonoDroid.CSharp.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\..\Novell\Novell.MonoDroid.CSharp.targets</Link>
+      <_RelativeParentLink>..\..\Novell\Novell.MonoDroid.CSharp.targets</_RelativeParentLink>
     </None>
     <None Include="MSBuild\Novell\MonoDroid.FSharp.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\..\Novell\MonoDroid.FSharp.targets</Link>
+      <_RelativeParentLink>..\..\Novell\MonoDroid.FSharp.targets</_RelativeParentLink>
     </None>
     <None
         Include="MSBuild\Xamarin\Xamarin.Android.Sdk.props">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\Xamarin.Android.Sdk.props</Link>
+      <_RelativeParentLink>..\Xamarin.Android.Sdk.props</_RelativeParentLink>
     </None>
     <None
         Include="MSBuild\Xamarin\Xamarin.Android.Sdk.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\Xamarin.Android.Sdk.targets</Link>
+      <_RelativeParentLink>..\Xamarin.Android.Sdk.targets</_RelativeParentLink>
     </None>
     <None
         Include="MSBuild\Xamarin\Android\Xamarin.Android.Bindings.Before.targets">
@@ -297,6 +302,13 @@
         Overwrite="false"/>
     <ItemGroup>
       <FileWrites Include="$(_GeneratedProfileClass)" />
+    </ItemGroup>
+  </Target>
+  <Target Name="_AssignRelativeParentLinkMetadata">
+    <ItemGroup>
+      <None Condition=" '%(None._RelativeParentLink)' != '' ">
+        <Link>%(None._RelativeParentLink)</Link>
+      </None>
     </ItemGroup>
   </Target>
 


brendanzagaeski:
Shucks. The `%(Link)` metadata is also being used to set the path in the output directory. I'll spend a moment brainstorming to see if any creative ideas come to mind about ways to adjust or ignore the metadata just in the Visual Studio for Mac Solution pad.

jonathanpeppers (commented):
Well... here is why:

https://github.com/mono/monodevelop/blob/58cfa46817143efde5fbee67ec9e1c2db1873db7/main/src/core/MonoDevelop.Core/MonoDevelop.Projects/ProjectFile.cs#L336-L337

Checking for `..` seems pretty arbitrary, lol!

Would it work to add a `Condition` checking for `$(MONO_GAC_PREFIX)`?

https://github.com/xamarin/xamarin-android/blob/5429b43b24e5af83c948e1b9773f1189063fbdba/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs#L442-L455

brendanzagaeski:

Thanks to jonathanpeppers's observation about where this exception is coming from in Visual Studio for Mac, I realized there was a slightly hacky way to work around it by adding a subdirectory before the relative path in the `%(Link)` metadata.  I adjusted the commit, commit message, and PR description to match that work around.

I checked that this looks OK in the Solution view in both Visual Studio for Mac and Visual Studio and that the changes in these files are propagated to the correct build output locations.

That said, no worries if this looks too ugly. I'll leave a comment on the Visual Studio for Mac bug about this behavior to mention how xamarin-android is using `..\` paths in the `%(Link)` metadata to set output locations so the Visual Studio for Mac team will know we're also interested in having behavior that more closely matches Visual Studio on Windows.

brendanzagaeski:
I had another idea that assigning directly to `%(TargetPath)` instead of `%(Link)` might be an interesting alternative, but that won't work easily because the [`<AssignTargetPath/>`](https://github.com/dotnet/msbuild/blob/250161603c003af239d6e276c4fc6357586ba301/src/Tasks/AssignTargetPath.cs) task overwrites any existing `%(TargetPath)` metadata, and that task is run on all the `@(None)` items by the [`AssignTargetPaths`](https://github.com/dotnet/msbuild/blob/250161603c003af239d6e276c4fc6357586ba301/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2877) target.

dellis1972:
I'm kinda ok with the hack of 'MSBuild\..\'. We should also raise an issue with VSForMac as its a difference in behaviour between windows and Mac.

brendanzagaeski:
> We should also raise an issue with VSForMac as its a difference in behaviour between windows and Mac.

:+1: I've added a note on the internal item: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1140731

brendanzagaeski:
Another option came to mind to use a custom metadata name like `%(_OutputLink)` and add a small extra target to `$(BuildDependsOn)` to copy that metadata into `%(Link)` at build time. That would have a small bonus that the files would then be visible in the Solution Explorer in Visual Studio on Windows.

```diff
diff --git a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
index 53021311..8ecee605 100644
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
@@ -13,6 +13,7 @@
     <BuildDependsOn>
       _CopyNDKTools;
       _GenerateXACommonProps;
+      _AssignOutputLinkMetadata;
       $(BuildDependsOn);
       _CopyExtractedMultiDexJar;
       _BuildMonoScripts;
@@ -26,25 +27,25 @@
   <ItemGroup>
     <None Include="MSBuild\Novell\Novell.MonoDroid.Common.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\..\Novell\Novell.MonoDroid.Common.targets</Link>
+      <_OutputLink>..\..\Novell\Novell.MonoDroid.Common.targets</_OutputLink>
     </None>
     <None Include="MSBuild\Novell\Novell.MonoDroid.CSharp.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\..\Novell\Novell.MonoDroid.CSharp.targets</Link>
+      <_OutputLink>..\..\Novell\Novell.MonoDroid.CSharp.targets</_OutputLink>
     </None>
     <None Include="MSBuild\Novell\MonoDroid.FSharp.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\..\Novell\MonoDroid.FSharp.targets</Link>
+      <_OutputLink>..\..\Novell\MonoDroid.FSharp.targets</_OutputLink>
     </None>
     <None
         Include="MSBuild\Xamarin\Xamarin.Android.Sdk.props">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\Xamarin.Android.Sdk.props</Link>
+      <_OutputLink>..\Xamarin.Android.Sdk.props</_OutputLink>
     </None>
     <None
         Include="MSBuild\Xamarin\Xamarin.Android.Sdk.targets">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>..\Xamarin.Android.Sdk.targets</Link>
+      <_OutputLink>..\Xamarin.Android.Sdk.targets</_OutputLink>
     </None>
     <None
         Include="MSBuild\Xamarin\Android\Xamarin.Android.Bindings.Before.targets">
@@ -236,6 +237,13 @@
         Replacements="@PACKAGE_VERSION@=$(ProductVersion);@PACKAGE_VERSION_BUILD@=$(XAVersionCommitCount);@NDK_ARMEABI_V7_API@=$(AndroidNdkApiLevel_ArmV7a);@NDK_ARM64_V8A_API@=$(AndroidNdkApiLevel_ArmV8a);@NDK_X86_API@=$(AndroidNdkApiLevel_X86);@NDK_X86_64_API@=$(AndroidNdkApiLevel_X86_64);@BUNDLETOOL_VERSION@=$(XABundleToolVersion)">
     </ReplaceFileContents>
   </Target>
+  <Target Name="_AssignOutputLinkMetadata">
+    <ItemGroup>
+      <None Condition=" '%(None._OutputLink)' != '' ">
+        <Link>%(None._OutputLink)</Link>
+      </None>
+    </ItemGroup>
+  </Target>
   <Target Name="_GenerateProfileClass"
       BeforeTargets="CoreCompile"
       Inputs="@(_SharedRuntimeAssemblies)"
```

jonathanpeppers:
This example does look a little better: https://github.com/xamarin/xamarin-android/pull/5146#issuecomment-698427126

Maybe use `%(_RelativeLink)` as the name, and I think it makes sense? You should put a small comment that it works around VS for Mac.

brendanzagaeski:

Switch to the approach of using a custom metadata name and an additional target.

For this candidate commit, I used `%(_RelativeParentLink)` as the metadata name. The thing I realized about `%(_RelativeLink)` was that technically all the `%(Link)` metadata are relative, so including `Parent` in the name can hopefully provide a little reminder at a glance about what's different about these particular relative paths. Happy to change that though!

jonathanpeppers (approved):

PR #5154, opened 2020-09-24
[build] plots-to-appinsights: Fix Status Enum to contain actual flag values

mjbond-msft:
I happened to notice that the plots-to-appinsights tool is calling HasFlag on the Status enum to check if Help is set.  HasFlag really 
only works correctly if the values within the enum occupy their own bit position.  So make this a true Flags enum.


pjcollins (approved):

jonpryor:
@mjbond-msft: is there a link to the AppInsights documentation which describes the Status enum?

PR #5159, opened 2020-09-25
[Xamarin.Android.Build.Tests] Import tests from legacy release definition

pjcollins:
Context: https://github.com/xamarin/QualityAssurance/pull/3648


src/Mono.Android/Test/Mono.Android-Tests.csproj
```diff
@@ -18,6 +18,7 @@
     <AndroidManifest>Properties\AndroidManifest.xml</AndroidManifest>
     <AndroidSupportedAbis>armeabi-v7a;x86</AndroidSupportedAbis>
     <AndroidUseSharedRuntime>False</AndroidUseSharedRuntime>
+    <AndroidEnableSGenConcurrent>true</AndroidEnableSGenConcurrent>
```
jonathanpeppers, line 21: Should we consider making this the default for all projects?

We already do it for .NET 6:

https://github.com/xamarin/xamarin-android/blob/885b57bdcf32e559961b183e1537844c5aa8143e/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.props#L13

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
```diff
@@ -88,5 +90,539 @@ public void SubscribeToAppDomainUnhandledException ()
 			}, Path.Combine (Root, builder.ProjectDirectory, "startup-logcat.log"), 45), $"Output did not contain {expectedLogcatOutput}!");
 		}
 
+
+		void SymbolicateAndAssert (string symbolArchivePath, string logcatFilePath, IEnumerable<string> expectedStackTraceContents)
+		{
+			// 09-22 14:21:07.064 12786 12786 I MonoDroid:   at UnnamedProject.MainActivity.OnCreate (Android.OS.Bundle bundle) [0x00051] in <b3164619c4824e379aecfb7335bd4cce>:0
+			var obfuscatedStackRegex = new Regex ("in <.*>:0");
```
jonathanpeppers, line 97: You might make this a member variable to reuse it and use `RegexOptions.Compiled`:
```csharp
Regex obfuscatedStackRegex = new Regex ("in <.*>:0", RegexOptions.Compiled);
```

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
```diff
@@ -88,5 +90,539 @@ public void SubscribeToAppDomainUnhandledException ()
...
+			var newLinkerTestInstance = new CustomLinkerDescriptionTests ();
+			Android.Util.Log.Info (TAG, newLinkerTestInstance.TryAccessNonXmlPreservedMethodOfLinkerModeFullClass ());
+
+			Android.Util.Log.Info (TAG, Library2.Bug21578.MulticastOption_ShouldNotBeStripped ());
+			Android.Util.Log.Info (TAG, Library2.Bug21578.MulticastOption_ShouldNotBeStripped2 ());
+			Android.Util.Log.Info (TAG, Library2.Bug35195.AttemptCreateTable ());
+			Android.Util.Log.Info (TAG, Library2.Bug36250.SerializeSearchRequestWithDictionary ());
+
+			Android.Util.Log.Info (TAG, ""All regression tests completed."");
+");
```
jonathanpeppers, line 595: Since the above is a massive string, should we make this an `EmbeddedResource` in its own `.cs` file, so you can read the file more easily? It wouldn't have intellisense, but at least you don't have to escape `""`.

You could load it similar to:
https://github.com/xamarin/xamarin-android/blob/e4405eb27e740c93cd90f993f13734309c013193/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinFormsAndroidApplicationProject.cs#L20-L21

jonathanpeppers: Same idea with the other large strings here. Maybe you would end up with 4 separate files?

The `.xml` file is short enough, it's probably fine.

src/Mono.Android/Test/Mono.Android-Tests.csproj
pjcollins: Good question, I remember us wanting to do this in the past but I think decided to only do so in the templates. Maybe enough time has passed since that change for us to want this everywhere, or maybe we shouldn't enable this for all existing projects. @jonpryor @steveisok do either of you have an opinion on this?

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
pjcollins: I've moved the four larger test files to @(EmbeddedResource) files.

jonathanpeppers (approved):
This looks good now, if the tests end up green. 👍

dellis1972 (approved):

PR #5162, opened 2020-09-28
[One .NET] deprecate support for DebugType=Full or DebugType=PdbOnly

jonathanpeppers:
In .NET 5+ `DebugType=full` or `DebugType=pdbonly` does not appear to
be supported. These silently fall back to `portable`.

Testing with .NET 5 rc2:

* `dotnet new console`
* Add `<DebugType>full</DebugType>` or `<DebugType>pdbonly</DebugType>`
* `dotnet build` results in portable `.pdb` files on both Windows and
  Mac.

This is good news, as we've been wanting to deprecate the other symbol
formats *anyway*.

1. Emit a XA0125 warning for `full` or `pdbonly` symbol files for
   "legacy" Xamarin.Android. Move any related MSBuild targets to

2. In Microsoft.Android.Sdk, we only use the symbol files returned by
   the `ComputeFilesToPublish` MSBuild target from the dotnet/sdk. The
   old symbol file formats appear to be ignored.

3. We no longer need to ship `pdb2mdb` in Microsoft.Android.Sdk.

I updated our MSBuild tests to default to using `DebugType=portable`.
I also updated a couple tests to rely on the new default value.


dellis1972 (commented):
This looks ok. However what will happen if a user in .net 5 references a nuget which is only shipping `.mdb` files or the old style `.pdb`. In that instance I guess no stack traces or anything will be available. 
If we are happy with that, then this PR looks good.

jonathanpeppers:
> However what will happen if a user in .net 5 references a nuget which is only shipping .mdb files or the old style .pdb. 

I tested this by adding `DebugType=full` or `pdbonly` in this project:

https://github.com/xamarin/net6-samples/blob/main/HelloForms/HelloForms.csproj

When you build `HelloForms.Droid`, the `@(ResolvedFileToPublish)` item group seems to completely ignore the old symbol files:

image

But it seems to work if you use `DebugType=portable`.

We would have to _search_ for the old symbol files and try to make them work, but I don't think it is worth it.

Documentation/guides/messages/xa0125.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Xamarin.Android warning XA0125
+description: XA0125 warning code
+ms.date: 09/28/2020
+---
+# Xamarin.Android warning XA0125
+
+## Issue
+
+Using symbol files with the `.mdb` extension or
```
jonpryor, line 10: This description doesn't quite go far enough; it can be easily interpreted as only referring to the "current" project.  It should explicitly call out that non-portable PDB symbols of *external dependencies* such as NuGet packages will *also* no longer be supported.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -317,6 +317,11 @@ When it appears in the middle of a sentence, "lint" is not capitalized.
   <data name="XA0124" xml:space="preserve">
     <value>Interpreter is not supported by the x86 ABI</value>
   </data>
+  <data name="XA0125" xml:space="preserve">
+    <value>'{0}' is using a deprecated symbol format. Use 'DebugType=portable' to use the newer, cross-platform symbol format.</value>
+    <comment>The following are literal names and should not be translated: DebugType=portable.
+{0} - The file name of a deprecated symbol file</comment>
+  </data>
```
brendanzagaeski, line 324: To align with the more verbose way to refer to MSBuild properties started in https://github.com/xamarin/xamarin-android/commit/91022a97214bc6611e61354f723e3b47a0b119e0, I'll suggest:

```xml
  <data name="XA0125" xml:space="preserve">
    <value>'{0}' is using a deprecated symbol format. Set the debugging information to Portable in the Visual Studio project property pages or edit the project file in a text editor and set the 'DebugType' MSBuild property to 'portable' to use the newer, cross-platform symbol format.</value>
    <comment>The following are literal names and should not be translated: 'DebugType', 'portable'
The capitalized word "Portable" that appears earlier in the message is plain text and should be translated, but the lowercase word "portable" later in the message is a literal value and should not be translated.
{0} - The file name of a deprecated symbol file</comment>
  </data>
```

One other, possibly silly adjustment could be to change "symbol format" to "debug information level" to align with the terminology for this particular MSBuild property on https://docs.microsoft.com/en-us/visualstudio/msbuild/common-msbuild-project-properties?view=vs-2019:

```xml
  <data name="XA0125" xml:space="preserve">
    <value>'{0}' is using a deprecated debug information level. Set the debugging information to Portable in the Visual Studio project property pages or edit the project file in a text editor and set the 'DebugType' MSBuild property to 'portable' to use the newer, cross-platform debug information level.</value>
    <comment>The following are literal names and should not be translated: 'DebugType', 'portable'
The capitalized word "Portable" that appears earlier in the message is plain text and should be translated, but the lowercase word "portable" later in the message is a literal value and should not be translated.
{0} - The file name of a deprecated symbol file</comment>
  </data>
```

_Reference info:_

In Visual Studio, the property is **Build > Advanced \[button\] > Debugging information** in the project property pages.

In Visual Studio for Mac, the property is **Build > Compiler > Debug information** in the project property pages.

"Debugging information" is hopefully a close enough compromise to cover both.

Documentation/guides/messages/xa0125.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Xamarin.Android warning XA0125
+description: XA0125 warning code
+ms.date: 09/28/2020
+---
+# Xamarin.Android warning XA0125
+
```
brendanzagaeski, line 7: To make this page potentially more directly discoverable via web search for anyone hitting the message, an _Example messages_ section could optionally be added:

````markdown
## Example messages

```
warning XA0125: 'AndroidApp1.pdb' is using a deprecated debug information level. Set the debugging information to Portable in the Visual Studio project property pages or edit the project file in a text editor and set the 'DebugType' MSBuild property to 'portable' to use the newer, cross-platform debug information level.
```
````

(The example text here matches the possible wording ideas mentioned in another comment in this PR review for the `.resx` file.)

File: Documentation/guides/messages/xa0125.md (hunk 1 above)
Comment (jonathanpeppers) [reply Documentation/guides/messages/xa0125.md:16]: I added a couple sentences:
```
...
This also includes NuGet packages or any library that distributes `.mdb` or non-portable `.pdb` files.
...
Notify the library author if you are encountering the `XA0125` warning from a NuGet package.
```

File: Documentation/guides/messages/xa0125.md [hunk 4]
```diff
@@ -0,0 +1,30 @@
...
+## Solution
+
+`full` or `pdbonly` symbol files are a Windows-specific format and
+`<DebugType>portable</DebugType>` should be used going forward. .NET 6
+and higher will not support the older formats. Notify the library
+author if you are encountering the `XA0125` warning from a NuGet package.
+
+In Visual Studio, go to `Properties > Build > Advanced...` and change
+`Debugging Information` to `Portable`.
```
brendanzagaeski, line 30: To align with the [docs guidelines](https://github.com/MicrosoftDocs/xamarin-docs/blob/c7d82b0684b1c26cdcbc16766b636894a64a80fd/contributing-guidelines/template.md#text-styling), this can use `**strong**` for UI elements:

```markdown
In Visual Studio, go to **Properties > Build > Advanced** in the project
property pages and change **Debugging information** to **Portable**.
```

And Visual Studio for Mac can be added:

```markdown
In Visual Studio for Mac, go to **Build > Compiler > Debug information**
in the project property pages and change **Debug information** to
**Portable**.
```

Documentation/guides/messages/xa0125.md
brendanzagaeski: Ooo. This raises a potentially interesting question about the wording of the message.

I think Jonathan Dick had some comments about providing more guidance to NuGet package authors for the `error XA0121: Assembly '{0}' is using '[assembly: {1}]', which is no longer supported. Use a newer version of this NuGet package or notify the library author.` messages, which were in some ways similar deprecations in the context of NuGet packages. As I'm writing that down though, I'm remembering that part of the challenge for those messages was that the new approach for the NuGet package authors wasn't just changing an MSBuild property but instead changing how external libraries were packaged in the NuGet package and adding a `.targets` file to the NuGet package. So in fact, in this new case, the current message is probably almost all set.

One little question is whether it might be worth adding something similar to the last sentence from those previous messages?

```markdown
If this file comes from a NuGet package, update to a newer version of
the NuGet package or notify the library author.
```

brendanzagaeski: Ooo. This raises a potentially interesting question about the wording of the message.

I think Jonathan Dick had some comments about providing more guidance to NuGet package authors for the `error XA0121: Assembly '{0}' is using '[assembly: {1}]', which is no longer supported. Use a newer version of this NuGet package or notify the library author.` messages, which were in some ways similar deprecations in the context of NuGet packages. As I'm writing that down though, I'm remembering that part of the challenge for those messages was that the new approach for the NuGet package authors wasn't just changing an MSBuild property but instead changing how external libraries were packaged in the NuGet package and adding a `.targets` file to the NuGet package. So in fact, in this new case, the current message is probably almost all set.

One little question is whether it might be worth adding something similar to the last sentence from those previous messages?

```markdown
If this file comes from a NuGet package, update to a newer version of
the NuGet package or notify the library author.
```

Documentation/release-notes/debugtype.md
```diff
@@ -0,0 +1,21 @@
+#### Deprecation of DebugType full and pdbonly
+
+* warning XA0125: 'AndroidApp1.pdb' is using a deprecated debug
+  information level. Set the debugging information to Portable in the
+  Visual Studio project property pages or edit the project file in a
+  text editor and set the 'DebugType' MSBuild property to 'portable'
+  to use the newer, cross-platform debug information level.
```
brendanzagaeski, line 7: As a tiny formatting idea to match some similar previous release notes, this could use a code block for the example warning:

````markdown
```
warning XA0125: 'AndroidApp1.pdb' is using a deprecated debug information level. Set the debugging information to Portable in the Visual Studio project property pages or edit the project file in a text editor and set the 'DebugType' MSBuild property to 'portable' to use the newer, cross-platform debug information level.
```
````

File: Documentation/guides/messages/xa0125.md (hunk 1 above)
Comment (jonathanpeppers) [reply Documentation/guides/messages/xa0125.md:16]: 👍  I updated some of the wording.

**Review by brendanzagaeski — APPROVED** [review]
Warning message wording, localizability, and documentation and release notes all approved. (Files approved: `*.resx`, `Resources.Designer.cs`, `*.xlf`, `*messages/*.md`, `*release-notes/*.md`)

Since I was already looking, I also did a quick sanity check of the `.targets` file changes and approve those too.

The one automated test that failed (`Failed to preload libmono-native.so (may not exist), ignoring`) is curious and doesn't seem like it could be caused by the `.targets` file changes. It looks like that test has been failing for other recent PR builds too.

**Comment by jonathanpeppers** [comment]
This always prints in the log for .NET 5 builds right now: `Failed to preload libmono-native.so (may not exist), ignoring`

Radek should hopefully be fixing this in #5145:
```
10-01 21:56:32.531  5589  5589 F monodroid: The type `Java.Lang.Throwable` is missing required instance fields! handle=0x73a822888580 handle_type=0x73a8228885c0 refs_added=0x73a8228885e0 weak_handle=0x0
```

**Comment by jonpryor** [comment]
@jonathanpeppers [wrote](https://github.com/xamarin/xamarin-android/pull/5162#issuecomment-702801681):

> This always prints in the log for .NET 5 builds right now: Failed to preload libmono-native.so (may not exist), ignoring

This message is unrelated to #5145. I don't know what actually causes this, or what the fix should be, but it's "ignorable".  @grendello?

>     10-01 21:56:32.531  5589  5589 F monodroid: The type `Java.Lang.Throwable` is missing required instance fields! handle=0x73a822888580 handle_type=0x73a8228885c0 refs_added=0x73a8228885e0 weak_handle=0x0
> Radek should hopefully be fixing this in #5145:

This is correct, and has been merged. Note however that #5145 does not address #5156; thus, instead of `XASdkDeployTests.DotNetInstallAndRun()` crashing because of `missing required instance fields`, it'll *intead* crash because of an unhandled `ArgumentException`.  (Baby steps!)

File: Documentation/guides/messages/xa0125.md (hunk 1 above)
Comment (jonpryor) [reply Documentation/guides/messages/xa0125.md:16]: Next I have a "rendering" concern: look at the formatted Markdown preview: https://github.com/xamarin/xamarin-android/blob/5934f47a2e869c109a7703103b39949f252d1e2e/Documentation/guides/messages/xa0125.md

The error message requires horizontal scrolling to view.  *Lots* of horizontal scrolling.

That might just be GitHub.  What about docs.microsoft.com?

https://docs.microsoft.com/en-us/xamarin/android/errors-and-warnings/xa0115

The above docs URL contains error message which is 274 characters "wide":

```
Invalid value 'armeabi' in $(AndroidSupportedAbis). This ABI is no longer supported. Please update your project properties to remove the old value. If the properties page does not show an 'armeabi' checkbox, un-check and re-check one of the other ABIs and save the changes.
```

This PR adds a message which is *424*:

```
'AndroidApp1.pdb' is using a deprecated debug information level. Set the debugging information to Portable in the Visual Studio project property pages or edit the project file in a text editor and set the 'DebugType' MSBuild property to 'portable' to use the newer, cross-platform debug information level. If this file comes from a NuGet package, update to a newer version of the NuGet package or notify the library author.
```

Perhaps it's just how *I* use web browsers, but these messages are "too long".

Can we get some newlines in there? A'la https://github.com/xamarin/xamarin-android/pull/4614#discussion_r471700258 ?

```
warning XA0125: 'AndroidApp1.pdb' is using a deprecated debug information level.
warning XA0125: Set the debugging information to Portable in the Visual Studio project property pages or edit the project file in a text editor and set the 'DebugType' MSBuild property to 'portable' to use the newer, cross-platform debug information level.
warning XA0125: If this file comes from a NuGet package, update to a newer version of the NuGet package or notify the library author.
```

(MSBuild/the IDE would "add" the extra `warning XA0125` bits…)

Comment (brendanzagaeski) [reply Documentation/guides/messages/xa0125.md:16]: Adding hard line breaks to the message in the _.resx_ file sounds fine to me too for this case. Just to throw out another option, if the message stayed as-is in the _.resx_ file, I think it would also be fine to hard wrap the text within the code block on the documentation page:

````markdown
```
warning XA0125: 'AndroidApp1.pdb' is using a deprecated debug
information level. Set the debugging information to Portable in the
Visual Studio project property pages or edit the project file in a text
editor and set the 'DebugType' MSBuild property to 'portable' to use the
newer, cross-platform debug information level. If this file comes from a
NuGet package, update to a newer version of the NuGet package or notify
the library author.
```
````

(For a small point of comparison, it seems the [NuGet errors and warnings reference](https://docs.microsoft.com/nuget/reference/errors-and-warnings/nu5127) doesn't do anything too fancy for long messages, so I think xamarin-android has some creative latitude to do something a little different.)

Comment (jonathanpeppers) [reply Documentation/guides/messages/xa0125.md:22]: I made the XA0125 message 3 lines, and wrapped the two docs files. 👍

File: Documentation/guides/messages/xa0125.md [hunk 6]
```diff
@@ -0,0 +1,41 @@
...
+and higher will not support the older formats. Update to a newer
+version or notify the library author if you are encountering the
+`XA0125` warning from a NuGet package.
+
+In Visual Studio, go to **Properties > Build > Advanced** in the project
+property pages and change **Debugging information** to **Portable**.
+
+In Visual Studio for Mac, go to **Build > Compiler > Debug information**
+in the project property pages and change **Debug information** to
+**Portable**.
```
brendanzagaeski, line 41: Since JonP raised some finer-tooth-comb docs formatting items in PR 5165, I'll paste a few tweaks here for the MSBuild property and file name formatting to follow the docs conventions even more precisely:

````markdown
Using symbol files with the _.mdb_ extension or built with the
`DebugType` MSBuild property set to `full` is deprecated. Using
non-portable symbol files with the _.pdb_ extension or `DebugType` set
to `pdbonly` is deprecated. This also includes NuGet packages or any
library that distributes _.mdb_ or non-portable _.pdb_ files.

## Solution

`full` or `pdbonly` symbol files are a Windows-specific format, and
`DebugType` should be set to `portable` going forward. .NET 6 and higher
will not support the older formats. Update to a newer version or notify
the library author if you are encountering the XA0125 warning from a
NuGet package.

In Visual Studio, go to **Properties > Build > Advanced** in the project
property pages and change **Debugging information** to **Portable**.

In Visual Studio for Mac, go to **Build > Compiler > Debug information**
in the project property pages and change **Debug information** to
**Portable**.
````

Since I was taking another read over it, here are some possible additional wording adjustments that came to mind in case any of them look good:

````markdown
Support for _.mdb_ or _.pdb_ symbols files that were built with the
`DebugType` MSBuild property set to `full` or `pdbonly` is now
deprecated. This applies to _.mdb_ and _.pdb_ files both in application
projects and in referenced libraries, such as NuGet packages.

## Solution

Set `DebugType` to `portable` in the application project as well all
library references.  `portable` is the recommended setting for all
projects from now on.  The older `full` and `pdbonly` settings are for
older Windows-specific file formats.  .NET 6 and higher will not support
those older formats.

In Visual Studio, go to **Properties > Build > Advanced** in the project
property pages and change **Debugging information** to **Portable**.

In Visual Studio for Mac, go to **Build > Compiler > Debug information**
in the project property pages and change **Debug information** to
**Portable**.

If the problematic symbol file comes from a NuGet package, update to a
newer version of the package or notify the library author.
````

dellis1972 (approved):

grendello:
> @jonathanpeppers [wrote](https://github.com/xamarin/xamarin-android/pull/5162#issuecomment-702801681):
> 
> > This always prints in the log for .NET 5 builds right now: Failed to preload libmono-native.so (may not exist), ignoring
> 
> This message is unrelated to #5145. I don't know what actually causes this, or what the fix should be, but it's "ignorable". @grendello?
> 
This is because we have no way of checking for presence of `libmono-native.so` (which ships with XA/current but doesn't ship with XA/net5), check the [code here](https://github.com/xamarin/xamarin-android/blob/7eb6d69f61136f3bf7218e4e754e7335df9a093e/src/java-runtime/java/mono/android/MonoPackageManager.java#L90-L97)

PR #5165, opened 2020-09-29
[One .NET] deprecate aapt(1) and AndroidUseAapt2=false

jonathanpeppers:
To follow suit with `AndroidDexTool=dx` being deprecated in
Xamarin.Android and removed in .NET 6, we should give aapt(1) the same
treatment.

* `AndroidUseAapt2=false` emits a XA1026 warning for "legacy"
  Xamarin.Android.
* `AndroidUseAapt2=false` emits a XA1026 error in
  Microsoft.Android.Sdk.
* `Xamarin.Android.Aapt.targets` is not imported or even shipped in
  Microsoft.Android.Sdk.
* Doing this allowed me to `Assert.Ignore()` lots of tests under a
  `dotnet` context. This should improve test times.


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -467,6 +464,14 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
...
       Condition="  '$(AndroidDexTool)' == 'dx' and '$(UsingAndroidNETSdk)' == 'true' "
   />
+  <AndroidWarning Code="XA1026"
+      ResourceName="XA1026"
+      Condition=" '$(AndroidUseAapt2)' == 'false' and '$(UsingAndroidNETSdk)' != 'true' "
+  />
+  <AndroidError Code="XA1026"
+      ResourceName="XA1026_dotnet"
+      Condition=" '$(AndroidUseAapt2)' == 'false' and '$(UsingAndroidNETSdk)' == 'true' "
+  />
```
dellis1972, line 474: Should there be a condition on a `CallTarget` in here , for the `Aapt` target.

Documentation/release-notes/aapt.md
```diff
@@ -0,0 +1,19 @@
...
+`aapt2` going forward. Xamarin.Android has now deprecated
+`<AndroidUseAapt2>false</AndroidUseAapt2>` as well.
+
+Update the `$(AndroidUseAapt2)` MSBuild property to `true` to select
+'aapt2'. This property corresponds to the **Use incremental Android
+packaging system (aapt2)** setting in the Visual Studio project
+properties pages. Alternatively, remove `<AndroidUseAapt2>` from the
+_.csproj_ file to let the build select the default value `true`.
+
+[aapt]: https://developer.android.com/studio/command-line/aapt2#aapt2_changes
```
dellis1972, line 19: We should probably warn users that changing to aapt2 might well break their apps. Since aapt2 is stricter than aapt.
We could also advise users to use ` --legacy` in the `AndroidAapt2CompileExtraArgs` so that those newer errors can be treated as warnings until they can fix them.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -434,6 +434,14 @@ In this message, the term "binding" means a piece of generated code that makes i
...
 {0} - The file name such as 'Foo.dll.config'</comment>
   </data>
+  <data name="XA1026" xml:space="preserve">
+    <value>Using 'aapt' is deprecated in favor of 'aapt2'. Please enable 'Use incremental Android packaging system (aapt2)' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidUseAapt2' MSBuild property to 'true'.</value>
+    <comment>The following are literal names and should not be translated: aapt, aapt2, Android, AndroidUseAapt2.</comment>
+  </data>
+  <data name="XA1026_dotnet" xml:space="preserve">
+    <value>Using 'aapt' is not supported in Xamarin.Android projects that target .NET 5 or higher. Please enable 'Use incremental Android packaging system (aapt2)' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidUseAapt2' MSBuild property to 'true'.</value>
+    <comment>The following are literal names and should not be translated: aapt, aapt2, Android, AndroidUseAapt2.</comment>
+  </data>
```
brendanzagaeski, line 444: Since `'aapt'` and `'aapt2'` aren't argument values for the MSBuild property, while `'true'` is, maybe:

```xml
  <data name="XA1026" xml:space="preserve">
    <value>Using AAPT is deprecated in favor of AAPT2. Please enable 'Use incremental Android packaging system (aapt2)' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidUseAapt2' MSBuild property to 'true'.</value>
    <comment>The following are literal names and should not be translated: AAPT, AAPT2, Android, AndroidUseAapt2, true.</comment>
  </data>
  <data name="XA1026_dotnet" xml:space="preserve">
    <value>Using AAPT is not supported in Xamarin.Android projects that target .NET 5 or higher. Please enable 'Use incremental Android packaging system (aapt2)' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidUseAapt2' MSBuild property to 'true'.</value>
    <comment>The following are literal names and should not be translated: AAPT, Android, AndroidUseAapt2, true.</comment>
  </data>
```

Documentation/guides/messages/xa1026.md
```diff
@@ -0,0 +1,32 @@
...
+
+```
+error XA1026: Using 'aapt' is not supported in Xamarin.Android projects that target .NET 5 or higher. Please enable 'Use incremental Android packaging system (aapt2)' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidUseAapt2' MSBuild property to 'true'.
+```
+
+## Issue
+
+[Google has deprecated][aapt] the `aapt` command-line tool in favor of
+`aapt2` going forward. Xamarin.Android has now deprecated
+`<AndroidUseAapt2>false</AndroidUseAapt2>` as well.
```
brendanzagaeski, line 22: Maybe to match the AAPT naming in the linked section of the developer.android.com doc:

```markdown
[Google has deprecated][aapt] the AAPT command-line tool in favor of
AAPT2 going forward. Xamarin.Android has accordingly now deprecated
`<AndroidUseAapt2>false</AndroidUseAapt2>` as well.
```

Documentation/guides/messages/xa1026.md
```diff
@@ -0,0 +1,32 @@
...
+
+[aapt]: https://developer.android.com/studio/command-line/aapt2#aapt2_changes
+
+## Solution
+
+Update the `$(AndroidUseAapt2)` MSBuild property to `true` to select
+'aapt2'. This property corresponds to the **Use incremental Android
+packaging system (aapt2)** setting in the Visual Studio project
+properties pages. Alternatively, remove `<AndroidUseAapt2>` from the
+_.csproj_ file to let the build select the default value `true`.
```
brendanzagaeski, line 32: To match the other comment about AAPT2 naming and use the slightly more "boring" (undecorated) syntax for MSBuild properties, maybe:

```markdown
Update the `AndroidUseAapt2` MSBuild property to `true` to select
AAPT2. This property corresponds to the **Use incremental Android
packaging system (aapt2)** setting in the Visual Studio project
properties pages. Alternatively, remove `<AndroidUseAapt2>` from the
_.csproj_ file to let the build select the default value `true`.
```

Documentation/release-notes/aapt.md
brendanzagaeski: In case it might be handy as a starting place or a verbatim copy-paste, the Xamarin.Android 10.0 release notes used the following text:

```markdown
> [!IMPORTANT]
> AAPT2 will in some cases enforce stricter rules on resource files than the
> previous AAPT, so some adjustments might be needed if you see new error
> messages that come from AAPT2 itself rather than from the Xamarin.Android
> build tasks.
```

Documentation/release-notes/aapt.md
```diff
@@ -0,0 +1,19 @@
+#### Deprecation of aapt in favor of aapt2
+
+* warning XA1026: Using 'aapt' is deprecated in favor of 'aapt2'.
+  Please enable 'Use incremental Android packaging system (aapt2)' in
+  the Visual Studio project property pages or edit the project file in
+  a text editor and set the 'AndroidUseAapt2' MSBuild property to
+  'true'.
```
brendanzagaeski, line 7: As a tiny formatting idea to match some similar previous release notes, this could use a code block for the example warning:

````markdown
#### Deprecation of AAPT in favor of AAPT2

```
warning XA1026: Using AAPT is deprecated in favor of AAPT2. Please enable 'Use incremental Android packaging system (aapt2)' in the Visual Studio project property pages or edit the project file in a text editor and set the 'AndroidUseAapt2' MSBuild property to 'true'.
```
````

(Tentatively incorporating the other naming changes from other comments in this PR review.)

File: src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets (hunk 1 above)
Comment (jonathanpeppers) [reply src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets:474]: My plan was to not import or ship `Xamarin.Android.Aapt.targets` at all:

https://github.com/xamarin/xamarin-android/blob/02e86b5a6bbd6f938c16cf30bc709543565bb34e/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L364-L365

However this is wrong ☝️ , it needs to be checking `$(UsingAndroidNETSdk)` instead.

This should also work:

https://github.com/xamarin/xamarin-android/blob/91b66985704f1592b8c55819f7ff48d5e7353900/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1160-L1161

Comment (jonathanpeppers) [reply src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets:474]: I changed this to:
```xml
<Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt.targets"  Condition=" '$(AndroidUseAapt2)' != 'true' and '$(UsingAndroidNETSdk)' != 'true' " />
<Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt2.targets" Condition=" '$(AndroidUseAapt2)' == 'true' or '$(UsingAndroidNETSdk)' == 'true' " />
```
A couple quick tests, things seem to work? Will need to check CI results.

dellis1972 (approved):

Documentation/guides/messages/xa1026.md
```diff
@@ -0,0 +1,45 @@
...
+
+> [!IMPORTANT]
+> AAPT2 will in some cases enforce stricter rules on resource files than the
+> previous AAPT, so some adjustments might be needed if you see new error
+> messages that come from AAPT2 itself rather than from the Xamarin.Android
+> build tasks.
+
+If needed, the `--legacy` switch can run AAPT2 in an AAPT
+compatibility mode. Add the following to your `.csproj` file:
```
brendanzagaeski, line 41: To align with the, admittedly somewhat interesting, convention of italics for filenames from the [docs guidelines](https://github.com/MicrosoftDocs/xamarin-docs/blob/c7d82b0684b1c26cdcbc16766b636894a64a80fd/contributing-guidelines/template.md#text-styling):

```markdown
If needed, the `--legacy` switch can run AAPT2 in an AAPT
compatibility mode. Add the following to your _.csproj_ file:
```

Documentation/guides/messages/xa1026.md
```diff
@@ -0,0 +1,45 @@
...
+
+```xml
+<AndroidAapt2CompileExtraArgs>--legacy</AndroidAapt2CompileExtraArgs>
+```
```
brendanzagaeski, line 45: To match the informal convention that's followed on-and-off in the other message docs, this could include an outer `<PropertyGroup/>` for context:

```xml
<PropertyGroup>
  <AndroidAapt2CompileExtraArgs>--legacy</AndroidAapt2CompileExtraArgs>
</PropertyGroup>
```

Documentation/guides/building-apps/build-properties.md
```diff
@@ -965,7 +965,8 @@ in their `.csproj`. Alternatively provide the property on the command line:
 /p:AndroidUseAapt2=True
 ```
 
-Added in Xamarin.Android 8.3.
+Added in Xamarin.Android 8.3. `AndroidUseAapt=false` is deprecated in
```
brendanzagaeski, line 968: Tiny typo:

```diff
-AndroidUseAapt=false
+AndroidUseAapt2=false
```

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -365,8 +362,8 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 <!-- As we split up/refactor this file, put new imports here -->
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.D8.targets" Condition=" '$(AndroidDexTool)' == 'd8' " />
 <Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.DX.targets" Condition=" '$(AndroidDexTool)' == 'dx' and '$(UsingAndroidNETSdk)' != 'true' " />
-<Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt.targets" />
-<Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt2.targets" />
+<Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt.targets"  Condition=" '$(AndroidUseAapt2)' != 'true' and '$(UsingAndroidNETSdk)' != 'true' " />
+<Import Project="$(MSBuildThisFileDirectory)Xamarin.Android.Aapt2.targets" Condition=" '$(AndroidUseAapt2)' == 'true' or '$(UsingAndroidNETSdk)' == 'true' " />
```
brendanzagaeski, line 366: In a legacy project, if the `_ResolveAndroidTooling` later changes `$(AndroidUseAapt2)`, outputting a different value into `$(_AndroidUseAapt2)`, is it hypothetically possible users could get a build error because the matching AAPT or AAPT2 targets wouldn't be imported? (Maybe there's an existing automated test that covers this and ensures it isn't a concern.)

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -467,6 +464,14 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
       ResourceName="XA1023_dotnet"
       Condition="  '$(AndroidDexTool)' == 'dx' and '$(UsingAndroidNETSdk)' == 'true' "
   />
+  <AndroidWarning Code="XA1026"
+      ResourceName="XA1026"
+      Condition=" '$(AndroidUseAapt2)' == 'false' and '$(UsingAndroidNETSdk)' != 'true' "
```
brendanzagaeski, line 469: It's a pretty small thing, but for consistent handling of any funny invalid property values (like maybe `aapt2`), is it worth making these warning conditions exactly match the `Import` conditions: `'$(AndroidUseAapt2)' != 'true'` instead of `'$(AndroidUseAapt2)' == 'true'`?

brendanzagaeski (commented):
I'll add a couple last comments for some possible small finishing touches for the documentation and release notes, and some little questions about the `.targets` files. The message wording and localizability look all set. (Files that are all set: `*.resx`, `Resources.Designer.cs`, `*.xlf`.)

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
jonathanpeppers: Considering we ship aapt2 now, @dellis1972 do you think I should turn these into build _errors_ now?

https://github.com/xamarin/xamarin-android/blob/9bc45b58ce0641890a0a3fbd6498eac515e9555a/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAndroidTooling.cs#L141-L154

I don't think we need to emit a warning and fall back to aapt(1) anymore, we should just give a nice error message instead? This would happen if `aapt2` didn't exist on disk, or didn't return a valid version number.

jonathanpeppers: I went ahead and made these changes in a second commit. It seems like it makes us able to clean up a lot of things?

I also updated some more error messages that @brendanzagaeski might want to take a look at. XA0111 and XA0112 should be pretty rare if we redistribute aapt2, but maybe they need more instructions?

Documentation/guides/building-apps/build-properties.md
```diff
@@ -965,7 +965,8 @@ in their `.csproj`. Alternatively provide the property on the command line:
 /p:AndroidUseAapt2=True
 ```
 
-Added in Xamarin.Android 8.3.
+Added in Xamarin.Android 8.3. `AndroidUseAapt2=false` is deprecated in
```
jonpryor, line 968: "Convention" usage question: should this be "`$(AndroidUseAapt2)=false` is deprecated…", adding the `$()`-fu?

*Is* there a convention here, in public docs?

Documentation/guides/messages/xa0111.md
```diff
@@ -1,13 +1,13 @@
...
+title: Xamarin.Android error XA0111
+description: XA0111 error code
 ms.date: 05/10/2019
 ---
 # Xamarin.Android warning XA0111
 
 ## Issue
 
-Xamarin.Android could not get the `aapt2` version via the command:
+Xamarin.Android could not get the AAPT2 version via the command:
```
jonpryor, line 10: Not sure we want this change?

Documentation/guides/messages/xa0111.md
```diff
@@ -17,14 +17,17 @@ aapt2 -version
 
 ## Solution
 
-Try re-installing the android build-tools.
+Check for a custom `<Aapt2ToolPath>` value in your _.csproj_ file and
```
jonpryor, line 20: Another docs "convention" question (@brendanzagaeski?): should this use MSBuild property syntax `$(Aapt2ToolPath)` or XML syntax `<Aapt2ToolPath>`?

Documentation/guides/messages/xa0111.md
```diff
@@ -17,14 +17,17 @@ aapt2 -version
...
+remove it, as Xamarin.Android redistributes a valid copy of AAPT2.
+Alternatively, try re-installing the Xamarin workload in Visual
+Studio.
 
-`aapt` is part of the Android SDK and is used internally by
+AAPT is part of the Android SDK and is used internally by
```
jonpryor, line 25: It occurs to me that one "problem" with ``s/`aapt`/AAPT/g`` is that maybe, *maybe*, Acrolinx will have problems with it, and "ding" us for it. See e.g. https://microsoft-ce-csi.acrolinx.cloud/api/v1/checking/scorecards/9d312bb2-4a2c-4646-9671-00ab2510881b (via https://github.com/MicrosoftDocs/xamarin-docs-pr/pull/1833), wherein Acrolinx "complains" that the following words are misspelled:

  * binutils
  * dex
  * multi-dex
  * dex-compiler
  * aapt
  * abi
  * Iso (via `Hebew (Iso alternative)`)

In local testing, all-caps *appears* to be skipped by Acrolinx, which is good.  Using backpacks `` `aapt` `` also appears to appease Acrolinx, but bold `**aapt**` does *not* appease it.

Documentation/guides/messages/xa0111.md
jonathanpeppers: This is based on earlier review: https://github.com/xamarin/xamarin-android/pull/5165#discussion_r498497108

Documentation/guides/building-apps/build-properties.md
brendanzagaeski: When I first looked at it, I figured it was good enough not to worry about it too much. But now that the question is raised, I think the choice that would be most consistent with other general MSBuild docs would be something like:

```
Setting `AndroidUseAapt2` to `false` is deprecated...
```

References:

- In https://docs.microsoft.com/visualstudio/msbuild/common-msbuild-project-properties and https://docs.microsoft.com/dotnet/core/project-sdk/msbuild-props, the properties are always plain and not decorated with `$()`.

  Example quotation:

  > For example, if you want access to preview code analyzers, set `AnalysisLevel` to `preview`.

- In https://docs.microsoft.com/visualstudio/msbuild/msbuild-properties and https://docs.microsoft.com/visualstudio/msbuild/comparing-properties-and-items, the usage of `$()` is called out specifically, and the `$()` syntax is only used when referring to the properties in the context of retrieving values in a project file.

  For example, the [_Store XML in properties_](https://docs.microsoft.com/visualstudio/msbuild/msbuild-properties?view=vs-2019#store-xml-in-properties) section contains both syntaxes that reflect slightly different usages:

  > Properties can contain arbitrary XML, which can help in passing values to tasks or displaying logging information. The following example shows the `ConfigTemplate` property, which has a value that contains XML and other property references. MSBuild replaces the property references by using their respective property values. Property values are assigned in the order in which they appear. Therefore, in this example, `$(MySupportedVersion)`, `$(MyRequiredVersion)`, and `$(MySafeMode)` should have already been defined.

Documentation/guides/messages/xa0111.md
brendanzagaeski: The safest choice in terms of MSBuild property conventions would be a plain undecorated `Aapt2ToolPath`, but the way this sentence is phrased is a little special because it is talking about the XML content of the _.csproj_ file, so using the XML tag name instead of the MSBuild property name is not necessarily incorrect. One option could be to be explicit about the fact that it's an XML tag name:

```
Check for any `<Aapt2ToolPath>` elements in the _.csproj_ file and
remove them. Xamarin.Android redistributes its own recommended version
of AAPT2, so setting a custom value for the `Aapt2ToolPath` MSBuild
property is generally not recommended. Alternatively, try re-installing
the Xamarin workload in Visual Studio.
```

(This candidate wording also includes a little more detail for the second sentence about why it generally makes sense to remove this property.)

Example inspiration <https://docs.microsoft.com/dotnet/core/tools/csproj#sdk-attribute>:

> The root `<Project>` element of the _.csproj_ file has a new attribute called `Sdk`.

Documentation/guides/messages/xa0111.md
brendanzagaeski: I'll vote for "AAPT" when not referring specifically to the command line command. The motivation is the distinction between AAPT2 and `aapt2` in the introductory section of <https://developer.android.com/studio/command-line/aapt2>.

brendanzagaeski (commented):
The approach of making XA0111 and XA0112 into errors looks pretty nice to me. I like the simplifications it allows.

brendanzagaeski (commented):
I'll leave one tiny idea for a word to add in the documentation, but otherwise, this looks all set to me!

brendanzagaeski (approved):

jonpryor:
I am *conceptually* confused by this PR. If the intent is to *deprecate* support for `$(AndroidUseAapt2)`=False, then shouldn't we emit a *warning* or something when that's the case, informing the user that support for `$(AndroidUseAapt2`=False is disappearing in a future release (when? .NET 6?).

I don't see any code which explicitly checks for `$(AndroidUseAapt2`=False.

Instead, I see code in `ResolveAndroidTooling.RunTask()` which *requires* that `AndroidUseAapt2`=True, and *only then* checks that `aapt2` exists.  When`AndroidUseAapt2`=False, on the other hand, I see *nothing*: no warning, no potential error no message.

What am I missing?

jonpryor:
> What am I missing?

This: https://github.com/xamarin/xamarin-android/pull/5165/files#diff-db39bf07b89b50eaab8ab2001546310cR466

```xml
  <AndroidWarning Code="XA1026"
       ResourceName="XA1026"
       Condition=" '$(AndroidUseAapt2)' != 'true' and '$(UsingAndroidNETSdk)' != 'true' "
   />
```

PR #5168, opened 2020-09-30
Bump to r8 2.1.67

jonathanpeppers:
Context: https://mvnrepository.com/artifact/com.android.tools/r8/2.1.67
Context: https://r8.googlesource.com/r8/+/cc1556056e7317b5b4350c6d92800895e9115550

Google appears to have made several releases to r8 in the past few months; we should update.


brendanzagaeski (approved):
Release note approved, with one idea for an optional change. Thanks!

PR #5173, opened 2020-09-30
[ci] Improve test dependency installation times

pjcollins:
The `AndroidToolchain` scenario previously used for installing Android
test dependencies has been replaced by two new scenarios responsible
for less work overall.  The `AndroidTestDependencies` scenario is
responsible for installing both JDKs, as well as the core Android SDK
components required for all test jobs.  The `EmulatorTestDependencies`
scenario is responsible for installing the Android SDK `emulator` tool,
as well as the `x86_64` system image that is used for emulator tests.

The majority of test jobs now skip the download of roughly 2 GB of
components.  The unused `x86` emulator system image has been removed,
and the `x86_64` system image, emulator, docs, and m2repository
components should not be needed for tests that don't require a device.
Jobs requiring emulator components skip roughly 1.2 GB of unnecessary
components (`x86` image, docs, m2repository).

A minor bug in JDK 1.8 version detection has also been fixed, which
prevents JDK 1.8 from being re-installed every time.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1080,6 +1080,9 @@ stages:
     steps:
     - template: yaml-templates/setup-test-environment.yaml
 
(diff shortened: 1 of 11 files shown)


pjcollins:
Full test run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4114060&view=results

grendello (approved):

PR #5180, opened 2020-10-02
[tests] fix PackagingTest.CheckIncludedAssemblies

jonathanpeppers:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4116152&view=ms.vss-test-web.build-test-results-tab&runId=16129032&resultId=100065

This test has been failing with:

    The following Expected files are missing.
    Expected: True
    But was:  False

I recently fixed a typo in 40644d91:

    -var missingFiles = expectedFiles.Where (x => !zip.ContainsEntry ("assmelbies/" + Path.GetFileName (x)));
    +var missingFiles = expectedFiles.Where (x => !zip.ContainsEntry ("assemblies/" + Path.GetFileName (x)));

At the time the test was passing, but after review, it still seems
wrong?

    Assert.IsTrue (missingFiles.Any (),

We shouldn't be missing any files?

After changing it to:

    Assert.IsFalse (missingFiles.Any (),

I actually see the lists of assembly names in the test are wrong...

When running under a "legacy" context, the linker is removing almost
all of the assemblies listed. I added usage of `Mono.Data.Sqlite`, as
it seemed important for the test. I removed the other missing assembly
names.

When running under a `dotnet` context, there are additional assemblies
that need to be listed. This may have changed in 91b66985 or other
past commits where we bumped to newer .NET 5 versions.

I double-checked the `.apk` file and I think the test is correct now?


dellis1972 (approved):

jonathanpeppers:
This test looks to be fixed now:

image

PR #2129, opened 2018-08-31
[Xamarin.Android.Build.Tasks] Add ConvertCustomView Task

dellis1972:
Fixed #2100

Currently `ConvertResourcesCases` is called twice on ALL
resources referenced by the project. This is terribly inefficient.
However it is required. The first pass is done before a `Compile`,
this fixes up the casing for things like drawables etc ready to
be processed by aapt. The second is done after `GenerateJavaStubs`
which happens AFTER the `Compile` step. This is to replace any
custom view references with the correct `{md5}.View` style references.
This is because we replace the normal readable namespace with an md5 hash.

The problem was that `ConvertResourcesCases` is doing a TON of
work it doesn't really need to do the second time around. For example
checking if it needs to lower case names of items. The second pass
really only needs to worry about custom views. In addition to that
it was also re-scanning ALL the files again.

This commit introduces a new Task `ConvertCustomView`. The sole
purpose of this task is to fix up the layout files which contain
custom views. It does nothing else. To make this even quicker we
modify `ConvertResourcesCases` to emit a mapping file (class-map.txt). This file
contains items like

	MonoDroid.Example.MyLayout;/fullpath/to/file.xml
	android.support.v7.widget.ActionBarOverlayLayout;/fullpath/to/some/other/file.xml

This allows us to know were the files are which contain ANY layout.

With this information in conjuction with the `acw_map.txt` file will
allow us to do a targeted update. So we go through the `acw_map.txt`
values and fix up those files where we have entires in the `class-map.txt`.

This reduces the amount of time spent processing files quite a bit.
For a Blank Xamarin Forms app from a clean build.

	2639 ms  ConvertResourcesCases                      1 calls
	   3 ms  ConvertCustomView                          1 calls

Normally the `ConvertResourcesCases` would be called twice and would
take a total of 5-6 seconds.

- [x] Fix up Paths in Error Message
- [X] Write a decent Commit Message
- [x] General Clean up.

PR #5184, opened 2020-10-05
[Xamarin.Android.Build.Tasks] remove acw-map.txt usage in <ConvertResourcesCases/>

jonathanpeppers:
The `<ConvertResourcesCases/>` loads the `acw-map.txt` file and passes
the dictionary throughout several methods in `AndroidResource`.
However, this dictionary is completely unused throughout?

Since 1886e6f1, the logic using `acw-map.txt` was moved to the
`<ConvertCustomView/>` MSBuild task. We no longer need to use this
file in the `<ConvertResourcesCases/>` task.

At the time `<ConvertResourcesCases/>` runs (before `Build`),
`acw-map.txt` does not even exist. It is created by the
`<GenerateJavaStubs/>` MSBuild task, that runs after `Build`.

I was working on a performance improvement, where I was attempting to
use `acw-map.txt` -- unfortunately an empty dictionary.

To eliminate confusion and clean things up, I removed all usage of
`acw-map.txt` in the `<ConvertResourcesCases/>` MSBuild task and
`AndroidResource` class. I was not able to see a performance
improvement with this change, as it is likely too small to notice.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Aapt.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Aapt.targets
@@ -37,7 +37,6 @@ Copyright (C) 2019  Microsoft Corporation. All rights reserved.
       Condition="'$(_AndroidUseAapt2)' != 'True'"
       ResourceDirectories="$(MonoAndroidResDirIntermediate);@(LibraryResourceDirectories)"
       ResourceNameCaseMap="$(_AndroidResourceNameCaseMap)"
-      AcwMapFile="$(_AcwMapFile)"
       CustomViewMapFile="$(_CustomViewMapFile)"
       AndroidConversionFlagFile="$(_AndroidResgenFlagFile)"
   />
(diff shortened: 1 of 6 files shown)


jonathanpeppers:
After testing this some more, this change might not be valid.

It seems that `acw-map.txt` doesn't exist on the first build at all? I will rework this so `<ConvertResourcesCases/>` doesn't attempt to load `acw-map.txt`, although it will be for cleanup purposes and probably not show much performance improvement.

dellis1972:
My guess is this used used to fix up Custom Views as part of the ConvertCases process. But that job has now been taken over by [ConvertCustomView](https://github.com/xamarin/xamarin-android/blob/0d6fd83b3299406225015a4e381f252428926413/src/Xamarin.Android.Build.Tasks/Tasks/ConvertCustomView.cs#L14) which was added in https://github.com/xamarin/xamarin-android/pull/2129. So this code **should** longer be needed in ConvertResourcesCases.

Might be worth updating the commit message with some of these links so we can understand why this change is ok?

jonathanpeppers:
I think this additional sentence will get us the right links:

> Since 1886e6f1, the logic using `acw-map.txt` was moved to the
`<ConvertCustomView/>` MSBuild task. We no longer need to use this
file in the `<ConvertResourcesCases/>` task.

We can add this when merging (or I'll add it if I make another change).

dellis1972 (approved):

PR #5187, opened 2020-10-06
Utilize GitHub token in compliance with GitHub service account restrictions

mjbond-msft:
Related to VS #[1225839](https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1225839)

As part of the GitHub Single Sign-On (SSO) effort, GitHub service accounts (i.e., any GitHub account not linked with an individual Microsoft account) will be denied GitHub org (Microsoft, Xamarin) access rights. The Xamarin.Android pipelines currently make use of a GitHub token tied to the Xamarin org (e.g., a Microsoft org).  That token is passed to Provisionator.  This change involves replacing that token with a GitHub token tied to a GitHub service account that does not have access to any Microsoft GitHub org (Microsoft or Xamarin).

### DIFF
--- a/build-tools/automation/azure-pipelines-oss.yaml
+++ b/build-tools/automation/azure-pipelines-oss.yaml
@@ -29,6 +29,8 @@ variables:
   DotNetCoreVersion: 3.1.201
   # Version number from: https://github.com/dotnet/installer#installers-and-binaries
   DotNetCorePreviewVersion: 5.0.100-rc.2.20459.1
+  GitHub.Token: $(github--pat--vs-mobiletools-engineering-service2)
+
 
 stages:
 - stage: mac_stage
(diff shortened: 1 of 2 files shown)


build-tools/automation/azure-pipelines-oss.yaml
```diff
@@ -29,6 +29,8 @@ variables:
   DotNetCoreVersion: 3.1.201
   # Version number from: https://github.com/dotnet/installer#installers-and-binaries
   DotNetCorePreviewVersion: 5.0.100-rc.2.20459.1
+  GitHub.Token: $(github--pat--vs-mobiletools-engineering-service2)
```
mjbond-msft, line 32: @pjcollins Alternatively, this could be a variable added to the build definition (same goes for the change to azure-pipelines.yaml). Let me know what you prefer.

pjcollins: I think this works fine as is, I generally prefer these types of variables to be defined in the build definition UI only if they have different values across different builds.

pjcollins (approved):

PR #5188, opened 2020-10-06
[Xamarin.Android.Build.Tasks] remove $(_AndroidResourceNameCaseMap) for performance

jonathanpeppers:
Reviewing a customer's large solution, `$(_AndroidResourceNameCaseMap)`
contains a ~20,000 character string that is logged and parsed multiple
times throughout the build.

Improvements:

* The `<AndroidComputeResPaths/>` MSBuild task can create a
  `Dictionary<string,string>` directly and store it via
  `RegisterTaskObject`. No need to create a giant string at all.
* `<AndroidComputeResPaths/>` shouldn't log all the output values, as
  it adds hundreds of log messages per `LogDebugTaskItems()` call in
  this customer's project. The surrounding log has enough context, so
  these messages are not needed.
* `MonoAndroidHelper.LoadResourceCaseMap` should only be called when
  the value is actually needed. This is normally for error messages.
* `MonoAndroidHelper.LoadResourceCaseMap` can call
  `GetRegisteredTaskObject` and will no longer need to parse or split
  any strings.
* The `$(_AndroidResourceNameCaseMap)` MSBuild property can be
  completely removed. We no longer need to pass it into various
  MSBuild tasks.

## Results ##

Testing this change in the customer's large solution:

    Before:
       17 ms  ConvertCustomView                          1 calls
      103 ms  AndroidComputeResPaths                    10 calls
     7692 ms  Aapt2Link                                 10 calls
    14288 ms  Aapt2Compile                              10 calls
    After:
       16 ms  ConvertCustomView                          1 calls
       99 ms  AndroidComputeResPaths                    10 calls
     7643 ms  Aapt2Link                                 10 calls
    13994 ms  Aapt2Compile                              10 calls

This should save ~348ms in an initial build of this large solution or
incremental builds with `AndroidResource` changes.

The `.binlog` file produced (which is a compressed format) is also
334,415 bytes smaller.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ConvertResourcesCases.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ConvertResourcesCases.cs
@@ -23,15 +23,13 @@ namespace Xamarin.Android.Tasks
 
 		public string AndroidConversionFlagFile { get; set; }
 
-		public string ResourceNameCaseMap { get; set; }
-
-		Dictionary<string,string> resource_name_case_map;
+		Dictionary<string,string> _resource_name_case_map;
 		Dictionary<string, HashSet<string>> customViewMap;
+
+		Dictionary<string, string> resource_name_case_map => _resource_name_case_map ??= MonoAndroidHelper.LoadResourceCaseMap (BuildEngine4);
 
 		public override bool RunTask ()
 		{
-			resource_name_case_map = MonoAndroidHelper.LoadResourceCaseMap (ResourceNameCaseMap);
-
 			if (CustomViewMapFile != null)
 				customViewMap = Xamarin.Android.Tasks.MonoAndroidHelper.LoadCustomViewMapFile (BuildEngine4, CustomViewMapFile);
 
--- a/src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
@@ -615,15 +615,14 @@ namespace Xamarin.Android.Tasks
 			return Path.Combine (platformPath, "android.jar");
 		}
 
-		public static Dictionary<string, string> LoadResourceCaseMap (string resourceCaseMap)
-		{
-			var result = new Dictionary<string, string> ();
-			if (resourceCaseMap != null) {
-				foreach (var arr in resourceCaseMap.Split (';').Select (l => l.Split ('|')).Where (a => a.Length == 2))
-					result [arr [1]] = arr [0]; // lowercase -> original
-			}
-			return result;
-		}
+		static readonly string ResourceCaseMapKey = $"{nameof (MonoAndroidHelper)}_ResourceCaseMap";
+
+		public static void SaveResourceCaseMap (IBuildEngine4 engine, Dictionary<string, string> map) =>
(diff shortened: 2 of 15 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/ConvertResourcesCases.cs
```diff
@@ -94,9 +92,11 @@ void FixupResources (ITaskItem item)
 					resourcedirectories, (level, message) => {
 						switch (level) {
 						case TraceLevel.Error:
+							resource_name_case_map ??= MonoAndroidHelper.LoadResourceCaseMap (ResourceNameCaseMap);
```
jonpryor, line 95: I wonder if it would be "easier to maintain" if we instead turned the `resource_name_case_map` field into a *property*:

```csharp
Dictionary<string, string>? _resource_name_case_map;
Dictionary<string, string> resource_name_case_map =>
    _resource_name_case_map ??= MonoAndroidHelper.LoadResourceCaseMap (ResourceNameCaseMap);
```

This would reduce (remove?) the possibility of `NullReferenceException`s because `resource_name_case_map` was used before it was initialized.

jonathanpeppers: I made this change, and also used `RegisterTaskObject` so we no longer create a giant string at all.

Instead of the `$(_AndroidResourceNameCaseMap)` property, we can just use the same `Dictionary<string,string>` throughout the build.

This gave us another 50ms or so.

src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
```diff
@@ -615,15 +615,14 @@ public static string TryGetAndroidJarPath (TaskLoggingHelper log, string platfor
...
-			return result;
-		}
+		static readonly string ResourceCaseMapKey = $"{nameof (MonoAndroidHelper)}_ResourceCaseMap";
+
+		public static void SaveResourceCaseMap (IBuildEngine4 engine, Dictionary<string, string> map) =>
+			engine.RegisterTaskObject (ResourceCaseMapKey, map, RegisteredTaskObjectLifetime.Build, allowEarlyCollection: false);
+
+		public static Dictionary<string, string> LoadResourceCaseMap (IBuildEngine4 engine) =>
+			engine.GetRegisteredTaskObject (ResourceCaseMapKey, RegisteredTaskObjectLifetime.Build)
+				as Dictionary<string, string> ?? new Dictionary<string, string> (0);
```
dellis1972, line 625: does the empty value need to include a StringComparer like we do when we created it at https://github.com/xamarin/xamarin-android/pull/5188/files#diff-a154f765074e2f487da8dc8a1fd28ac3R76

jonathanpeppers: If the `??` evaluates, then something really went wrong. I thought an empty dictionary would be better than a `null`.

It could `LogError` instead, but it means we have something broken not-actionable for users. Should it be an error instead?

dellis1972: erring out with a non actionable error is probably not the best. Let's leave this as it is.

dellis1972 (approved):

brendanzagaeski (approved):
Release note looks good. Thanks!

(Files approved: `Documentation/release-notes/5188.md`)

PR #5191, opened 2020-10-07
[ci] Move .NET preview installation to xaprepare

pjcollins:
Context: https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workloads.md

In One .NET, our legacy installer payload content will be distributed
through a new "Optional Workload" concept.  Early support for workloads
has now landed in VS and the .NET SDK, and we'd like to begin to convert
our monolithic MSBuild SDK NuGet into various workload packs.  In an
effort to support dev and test efforts to this end, it will be helpful
to introduce a non-system installation of .NET that we can modify freely
without admin access or fear of leaving machines in a broken state.


pjcollins:
Full run for commit f6dc8fb: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=4128812&view=results

jpobst:
Shouldn't we be moving towards a future where `xaprepare` is built and run by .NET Core instead of .NET Framework/Mono?  How is this going to affect that?

pjcollins:
@jpobst The goal of these changes is to provide a non-system based installation of the .NET SDK that we can use for One .NET work, and modify freely for local dev and test efforts without worrying about breaking system installations. It also allows us to consolidate our .NET SDK version dependency information into a single location.  

Building xamarin-android still requires a system installation of .NET even with these changes (unless you modify your $PATH to prefer this installation). In the future if we do move `xaprepare` to target net5 or something newer, we'll likely want to consider introducing some system installation bootstrap for .NET.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
```diff
@@ -60,6 +60,17 @@ public static string GetJavaSdkPath ()
 			return JavaSdkPath;
 		}
 
+		// Cache the result, so we don't run MSBuild on every call
+		static string DotNetPath;
+		public static string GetDotNetPath ()
+		{
+			if (string.IsNullOrEmpty (DotNetPath))
+				DotNetPath = Environment.GetEnvironmentVariable ("DOTNET_ROOT");
```
jonathanpeppers, line 68: I was thinking about this, maybe we should take these two lines out?

I don't know of a situation where someone would want their `$DOTNET_ROOT` to take precedence of what is set in `Configuration.props`.

They can use `Configuration.Override.props` to change the location of `$(DotNetPath)`.

pjcollins: Maybe we should just change the ordering? Though, I would also be happy to remove this to make test execution more explicit. Maybe it's better to fail if the .NET installation specified in `Configuration.props` can't be found.

jonathanpeppers: Reordering like we do for JavaSdkDirectory would be fine:

https://github.com/xamarin/xamarin-android/blob/eb4b49aab01df7a20f78e1c90d03b983e6f17a14/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs#L48-L61

jpobst:
It feels like we should keep the .NET Core 3.1 install the way it is today, and then install any additional needed .NETs with the new stuff.  I think there's no reason we couldn't be running `xaprepare` with .NET Core today, other than no one has made the change.

build-tools/xaprepare/xaprepare/ConfigAndData/Configurables.MacOS.cs
```diff
@@ -21,5 +21,10 @@ partial class Paths
 			public static readonly string UnstrippedLibMonoSgenName = $"{LibMonoSgenBaseName}.d{Defaults.NativeLibraryExtension}";
 			public static readonly string StrippedLibMonoSgenName = $"{LibMonoSgenBaseName}{Defaults.NativeLibraryExtension}";
 		}
+
+		partial class Urls
+		{
+			public static readonly Uri DotNetInstallScript = new Uri ("https://dot.net/v1/dotnet-install.sh");
```
grendello, line 27: Since URLs for `macOS` and `Linux` are the same, this partial class should be added to `Configurables.Unix.cs` instead

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetSDKs.cs
```diff
@@ -0,0 +1,113 @@
...
+{
+	class Step_InstallDotNetSDKs : StepWithDownloadProgress
+	{
+		public Step_InstallDotNetSDKs ()
+			: base ("Install required .NET Core SDKs locally")
+		{ }
+
+		protected override async Task<bool> Execute (Context context)
+		{
+			if (context == null)
```
grendello, line 17: This check shouldn't be necessary anymore since `xaprepare` is compiled with nullable refs and thus `context` can be assumed not to be `null` here.

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetSDKs.cs
```diff
@@ -0,0 +1,113 @@
...
+					Utilities.DeleteDirectory (packToRemove);
+			}
+
+			// Delete any unnecessary SDKs if they exist.
+			var sdkPath = Path.Combine (dotnetPath, "sdk");
+			if (Directory.Exists (sdkPath)) {
+				var installedSdks = Directory.EnumerateDirectories (sdkPath).Select (d => new DirectoryInfo (d));
+				foreach (var sdkToRemove in installedSdks.Where (s => s.Name != dotnetVersion && s.Name != dotnetPreviewVersion))
+					Utilities.DeleteDirectory (sdkToRemove.FullName);
+			}
```
grendello, line 38: I feel the deletion code should be moved till after the `InstallDotNetAsync` calls, so that we delete old dotnet only if the new was installed successfully, thus assuring that the system is left in the same state as before `xaprepare` was invoked

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetSDKs.cs
```diff
@@ -0,0 +1,113 @@
...
+			Log.StatusLine ($"Installing dotnet SDK '{version}'...");
+
+
+			if (Context.IsWindows) {
+				return Utilities.RunCommand ("powershell.exe", new string [] {
+					"-NoProfile", "-ExecutionPolicy", "unrestricted", "-file", dotnetScriptPath,
+					"-Version", version, "-InstallDir", dotnetPath, "-Verbose"
+				});
+			} else {
+				return Utilities.RunCommand ("bash", new string [] {
```
grendello, line 97: `bash` is no longer the default shell on `macOS` (it now uses `zsh`). Running `sh` might be safer.

grendello (changes requested):

pjcollins:
> It feels like we should keep the .NET Core 3.1 install the way it is today, and then install any additional needed .NETs with the new stuff. 

I see some downside with this approach. We'd be increasing the number of places where we are tracking .NET SDK dependency information, and we'd have to keep the 3.1 version in sync across multiple YAML files (including monodroid which is already out of sync, but we don't do a great job of syncing there in general). I'd also prefer to keep the .NET SDK versions we depend on for CI/testing in the same installation directory for my own sanity. Maybe there are benefits that outweigh these costs that I am overlooking?

I know @grendello was opposed to making this a .NET Core app when this was being written, I assume some of that reasoning still stands today? This tool seems like one of the lower priority things we'd want/need to upgrade to .NET Core when it comes to our build system,

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetSDKs.cs
pjcollins: `sh` didn't work on Linux unfortunately, as this option `set -o pipefail` is used in the script, which the `sh` on our Linux VMs didn't seem to support. We could separate this out to use `sh` on macOS and `bash` on Linux, or we can keep it as is assume that this won't cause issues anytime soon.

jpobst:
I think there are some benefits:
- Mono and .NET Framework are deprecated, we should be working towards using them less.  It'll be a very long process, but we shouldn't be going backwards.
- The same command runs on all platforms, so we don't need the extra complexity of `xaprepare.exe` versus `mono xaprepare.exe`.
- We could avoid the hacks in place so that `xaprepare` can update the system Mono while it is running on Mono.  (https://github.com/xamarin/xamarin-android/blob/master/build-tools/automation/azure-pipelines.yaml#L164-L167)

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetSDKs.cs
grendello: Let's keep it as-is then. If it breaks then we'll fix it

grendello:
> I think there are some benefits:
> 
>     * Mono and .NET Framework are deprecated, we should be working towards using them less.  It'll be a very long process, but we shouldn't be going backwards.
> 
>     * The same command runs on all platforms, so we don't need the extra complexity of `xaprepare.exe` versus `mono xaprepare.exe`.
> 
>     * We could avoid the hacks in place so that `xaprepare` can update the system Mono while it is running on Mono.  (https://github.com/xamarin/xamarin-android/blob/master/build-tools/automation/azure-pipelines.yaml#L164-L167)

One big disadvantage of this is that we would require *both* Mono and DotNet to be installed on the system with no, currently, clear advantage to using dotnet. `xaprepare.exe` is invoked via `make` on `*nix` so there's no benefit here either (on Windows it's invoked via MSBuild with `/t:Prepare`)

grendello:
> > It feels like we should keep the .NET Core 3.1 install the way it is today, and then install any additional needed .NETs with the new stuff.
> 
> I see some downside with this approach. We'd be increasing the number of places where we are tracking .NET SDK dependency information, and we'd have to keep the 3.1 version in sync across multiple YAML files (including monodroid which is already out of sync, but we don't do a great job of syncing there in general). I'd also prefer to keep the .NET SDK versions we depend on for CI/testing in the same installation directory for my own sanity. Maybe there are benefits that outweigh these costs that I am overlooking?
> 
> I know @grendello was opposed to making this a .NET Core app when this was being written, I assume some of that reasoning still stands today? This tool seems like one of the lower priority things we'd want/need to upgrade to .NET Core when it comes to our build system,

Indeed. I'd much prefer to convert `xaprepare` to .net core only after we absolutely require dotnet to be installed.

jpobst:
There will never be a "big bang" change where we convert 100% to dotnet in a single commit.  The only way to do it is slowly when the opportunities arise.  Putting it off is simply fighting against the inevitable future.

jpobst:
> One big disadvantage of this is that we would require both Mono and DotNet to be installed on the system with no, currently, clear advantage to using dotnet. xaprepare.exe is invoked via make on *nix so there's no benefit here either (on Windows it's invoked via MSBuild with /t:Prepare)

We do require both Mono and DotNet to be installed on the system.  This is simply about which is used to bootstrap setting them up.

grendello:
> There will never be a "big bang" change where we convert 100% to dotnet in a single commit. The only way to do it is slowly when the opportunities arise. Putting it off is simply fighting against the inevitable future.

I just don't see any compelling advantage from this change with regards to `xaprepare` as of yet

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetSDKs.cs
pjcollins: @grendello this new sandbox is going to be primarily for testing, and modified regularly to drop in various Android workload components. I'd like to be regularly cleaning it up even if not installing a new version so that we try to reduce the chances of getting into a weird state.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
pjcollins: I removed this, we'll try to not set _any_ variables to point at this preview SDK.

grendello (approved):

jonathanpeppers (approved):

jpobst (approved):

jonpryor:
```markdown
Context: https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workloads.md

For .NET 6, our .NET 6 installer payload content will be distributed
through a new "Optional Workload" concept.

Early support for workloads has now landed in VS and the .NET SDK,
and we'd like to begin to convert our monolithic MSBuild SDK NuGet into
various workload packs.  In an effort to support dev and test efforts
to this end, it will be helpful to introduce a non-system installation
of .NET that we can modify freely without admin access or fear of
leaving machines in a broken state.

Update the `xaprepare -s:Standard` scenario to install multiple
`dotnet` versinos into `$(DotNetPreviewPath)`, which defaults to
`$(AndroidToolchainDirectory)\dotnet`.

The currently installed `dotnet` versions are:

  * 3.1.3: Stable version, for some tests.
  * 5.0.100-rc.2.20480.7: Preview .NET 5 RC which contains support
    for the Optional Workload concept.
```

PR #5195, opened 2020-10-09
[One .NET] Initial workload support

pjcollins:
Context: https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workloads.md

A new Microsoft.NET.Workload.Android NuGet package has been added which
contains the WorkloadManifest.json and WorkloadManifest.targets files
required for the [workload resolver][1].

The Microsoft.Android.Sdk NuGet package has been convered from an
[MSBuild project SDK][2] to an SDK workload pack.  The most noticeable
difference in this change is that the `Sdk.props` import hook has been
replaced by an `AutoImport.props` file.  The `AutoImport.props` file
should not set any MSBuild properties, and should only contain ItemGroup
glob behavior that can be modified by the main project file.  The
`AutoImport.props` file included in Microsoft.Android.Sdk will be
imported by all project using Microsoft.NET.Sdk once workloads are
enabled by default.  As a result, it is crucial that all of the content
in this file is conditionally included based on some Android specific
property value.

All other .props files from this repo which are included in the
Microsoft.Android.Sdk pack have been renamed to .targets files.  This
was done to follow the convention that .props files are loaded before
the main project file, and .targets files are loaded after.  The only
file that is loaded before the project file is `AutoImport.props`.

All One .NET tests have been updated to use the Microsoft.NET.Sdk
project SDK, rather than Microsoft.Android.Sdk.  In order to support
this, a new `/t:ExtractWorkloadPacks` target has been added. This target
can be used with any .proj file in build-tools/create-packs/, and it is
responsible for extracting Microsoft.NET.Workload.Android,
Microsoft.Android.Sdk, and Microsoft.Android.Ref into our sandbox .NET
preview installation location.  This target will also clean up any stale
Microsoft.Android.* packs that were previously installed.

The Xamarin.Android.Sdk.Lite package has been removed, as it continues
to be a bit of a headache to maintain moving forward.  Old versions of
this package will continue to be available on the internal NuGet feed as
needed, however teams should soon be able to switch over to our workload
for their own dev/test efforts.

Finally, I've bumped the .NET Preview version to 5.0.100-rtm.20509.5.

[1]: https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workload-resolvers.md
[2]: https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk


build-tools/create-packs/Directory.Build.targets
```diff
@@ -62,7 +62,30 @@
       DependsOnTargets="CreateAllRuntimePacks" >
     <Exec Command="$(DotNetPreviewTool) pack -p:Configuration=$(Configuration) &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Ref.proj&quot;" />
     <Exec Command="$(DotNetPreviewTool) pack -p:Configuration=$(Configuration) &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Sdk.proj&quot;" />
-    <Exec Command="$(DotNetPreviewTool) pack -p:Configuration=$(Configuration) &quot;$(MSBuildThisFileDirectory)Xamarin.Android.Sdk.Lite.proj&quot;" />
+    <Exec Command="$(DotNetPreviewTool) pack -p:Configuration=$(Configuration) &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Workload.proj&quot;" />
+  </Target>
+
+  <Target Name="ExtractWorkloadPacks"
```
jonathanpeppers, line 68: Right now I've been doing this for my .NET 6 dev-loop:

```
> msbuild Xamarin.Android.sln -t:PackDotNet
```

Should we make a new target that does both `PackDotNet` and `ExtractWorkloadPacks`?

build-tools/create-packs/Directory.Build.targets
```diff
@@ -62,7 +62,30 @@
...
+      <!-- <_WLPacks Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nupkgs\Microsoft.Android.Runtime.*.nupkg" /> -->
+    </ItemGroup>
+    <Unzip
+        SourceFiles="@(_WLManifest)"
+        DestinationFolder="$(DotNetPreviewPath)sdk-manifests\$(DotNetPreviewVersionBand)\Microsoft.Android.Workload"
+    />
+    <Unzip
+        SourceFiles="@(_WLPacks)"
+        DestinationFolder="$(DotNetPreviewPath)packs\$([System.String]::Copy('%(_WLPacks.Filename)').Replace('.$(AndroidPackVersionLong)', ''))\$(AndroidPackVersionLong)"
+    />
```
jonathanpeppers, line 84: I worry about when I switch branches, what happens to the old versions?

Should we consider nuking the whole `packs` directory?

build-tools/create-packs/Directory.Build.targets
pjcollins: I added the new Extract target to the `PackDotNet` target here - https://github.com/xamarin/xamarin-android/pull/5195/files#diff-9209e878d1934d143c08fe9e00beb658R9, would you rather create a new target?

build-tools/create-packs/Directory.Build.targets
pjcollins: `xaprepare` will clean up packs in our preview install path every time it runs - https://github.com/xamarin/xamarin-android/blob/master/build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetPreview.cs#L26, but I think it would make sense to run a similar clean up as part of this target. We shouldn't delete everything in `packs`, but everything `Android` related can be removed before extraction.

build-tools/create-packs/Directory.Build.targets
jonathanpeppers: I missed that, that is probably fine.

pjcollins: I was considering moving the Windows build + smoke test over to use this target to simplify the YAML a bit, do you think that would make sense?
`msbuild /t:Prepare`
`msbuild /t:PackDotNet`

jonathanpeppers: Yes, that would also help to know the target works, as we rely on it for CI. 👍

build-tools/create-packs/Microsoft.Android.Workload.proj
```diff
@@ -0,0 +1,86 @@
...
+  <Target Name="_GenerateXAWorkloadContent"
+      DependsOnTargets="_GetDefaultPackageVersion" >
+    <PropertyGroup>
+      <WorkloadManifestTargetsPath>$(OutputPath)\workload-manifest\WorkloadManifest.targets</WorkloadManifestTargetsPath>
+      <WorkloadManifestJsonPath>$(OutputPath)\workload-manifest\WorkloadManifest.json</WorkloadManifestJsonPath>
+      <WorkloadManifestTargetsContent>
+<![CDATA[
+<Project>
+    <Import Project="Sdk.targets"
+        Sdk="Microsoft.Android.SDK"
```
rolfbjarne, line 39: Shouldn't this be:

```suggestion
        Sdk="Microsoft.Android.Sdk"
```

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultItems.props
```diff
@@ -1,12 +1,20 @@
...
+Default item includes (globs and implicit references).
+
+***********************************************************************************************
+-->
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
   <!-- Default Resource file inclusion -->
   <!-- https://developer.android.com/guide/topics/resources/providing-resources -->
-  <ItemGroup Condition=" '$(EnableDefaultAndroidItems)' == 'true' ">
+  <ItemGroup Condition=" '$(TargetPlatformIdentifier)' == 'Android' and '$(EnableDefaultAndroidItems)' == 'true' ">
```
pjcollins, line 17: @dsplaisted Is this `$(TargetPlatformIdentifier)` check restrictive enough for these default items being imported by `AutoImport.props`?

dsplaisted: I think in this case you don't need the `TargetPlatformIdentifier` check, because `EnableDefaultAndroidItems` will only be true if when targeting Android anyway.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultItems.props
```diff
@@ -34,11 +42,11 @@
...
-  <ItemGroup>
+  <ItemGroup Condition=" '$(TargetPlatformIdentifier)' == 'Android' ">
     <ProjectCapability Include="Mobile" />
     <ProjectCapability Include="Android" />
     <ProjectCapability Include="AndroidApplication" Condition="$(AndroidApplication)" />
-    
+
     <!-- Conflicts with our targets generator in VS+CPS: See https://work.azdo.io/1112733 -->
     <ProjectCapability Remove="LaunchProfiles" />
   </ItemGroup>
```
dsplaisted, line 52: Is there a reason the ProjectCapability items need to be defined in the .props?  Could they just be defined in a .targets file?

dsplaisted (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
```diff
@@ -3,16 +3,19 @@
...
+    <ResolveAssemblyConflicts>true</ResolveAssemblyConflicts>
+    <_AndroidResourceDesigner>Resource.designer.cs</_AndroidResourceDesigner>
   </PropertyGroup>
 
-  <!-- Default property settings, imported before Microsoft.NET.Sdk -->
+  <Import Project="Microsoft.Android.Sdk.BundledVersions.props" Condition=" Exists('Microsoft.Android.Sdk.BundledVersions.props') " />
+  <Import Project="Microsoft.Android.Sdk.SupportedPlatforms.props" />
   <Import Project="Microsoft.Android.Sdk.DefaultProperties.props" />
   <Import Project="$(MSBuildThisFileDirectory)..\tools\Xamarin.Android.Common.Debugging.props"
       Condition="Exists('$(MSBuildThisFileDirectory)..\tools\Xamarin.Android.Common.Debugging.props')"/>
```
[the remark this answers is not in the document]
pjcollins: @dsplaisted Am I correct to assume that `AutoImport.props` is the only pre-project file import hook that workloads have access to? It probably makes sense for us to rename these to `.targets` files now since we no longer have an `Sdk.props` import hook, and we shouldn't be setting properties in `AutoImport.props`.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultItems.props
pjcollins: I'll move these into a targets file.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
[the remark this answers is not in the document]
jonathanpeppers: Is there a way a workload has a chance to set properties after the `.csproj`? We were setting `$(ProduceReferenceAssembly)` to `false` by default for "application" projects, as the reference assembly is somewhat useless. It needs to be set before `Microsoft.NET.Sdk/Sdk.targets` but depends on `$(OutputType)` set to `Exe` in the `.csproj`.

Likewise, I think we should set `$(GenerateDependencyFile)` to `false`? We don't need to generate a `.deps.json` file?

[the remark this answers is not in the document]
dsplaisted: > Is there a way a workload has a chance to set properties after the `.csproj`? We were setting `$(ProduceReferenceAssembly)` to `false` by default for "application" projects, as the reference assembly is somewhat useless. It needs to be set before `Microsoft.NET.Sdk/Sdk.targets` but depends on `$(OutputType)` set to `Exe` in the `.csproj`.
> 
> Likewise, I think we should set `$(GenerateDependencyFile)` to `false`? We don't need to generate a `.deps.json` file?

I discussed this a bit with @pjcollins, we may need to add another targets hook for this scenario.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XASdkProject.cs
```diff
@@ -54,25 +48,12 @@ public static void SaveNuGetConfig (string directory)
...
-    }}
-}}");
-		}
-
 		public string PackageName { get; set; }
 		public string JavaPackageName { get; set; }
 
 		public XASdkProject (string outputType = "Exe")
 		{
-			Sdk = $"Microsoft.Android.Sdk/{SdkVersion}";
```
dellis1972, line 75: This might be a silly question. With the "old" system this `{SdkVersion}` was used to allow Nuget/MSbuild to find the SDK files in our local Nuget cache. How will this work with the new workload system, can it still look in the local cache? Or will it try to download things from nuget.org. What is going to happen on the bots when it's trying to install a new workload because our tests are requesting it. (I should probably read the spec for the workload stuff).

Do we need to be able to support both types of sdk? I'm thinking from a local development point of view.

dsplaisted: Workloads won't automatically be downloaded from NuGet.org.  They need to be installed.  The version that is used is determined by the workload manifest, which the workload installation process should install or update.

For testing, you would want to "install" the workload you want to test.  For now this basically means copying some files into the dotnet folder.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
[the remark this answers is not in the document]
rolfbjarne: I'm running into a similar scenario for Xamarin.iOS: we have to set the `DebuggerSupport` property before Microsoft.NET.Sdk is imported, but we need to know other properties in the user's csproj to know which value to set `DebuggerSupport` to.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XASdkProject.cs
pjcollins: @dellis1972 I think we can migrate entirely away from the old MSBuild style SDK package. For local development, we are now installing a .NET 5 preview build into `$(AndroidToolchainDirectory )/dotnet`. This PR adds a new `ExtractWorkloadPacks` msbuild target to extract the relevant workload .nupkg files into this new .NET 5 installation path for tests running both locally and in CI to use.

For local dev/test efforts the following steps will prepare, build, pack .nupkgs, and install our workload components into the .NET 5 preview sandbox:

`msbuild /Prepare Xamarin.Android.sln`
`msbuild /t:PackDotNet Xamarin.Android.sln`
or 
`make jenkins`
`msbuild /t:CreateAllPacks,ExtractWorkloadPacks build-tools/create-packs/Microsoft.Android.Sdk.proj`

This works for everything aside from runtime packs at the moment, those will still need to be pulled from _some_ NuGet source. Our tests use this local source by default: https://github.com/xamarin/xamarin-android/blob/a7bda88d05b58efd26c17be3956c7140d05868e1/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/XamarinProject.cs#L94

Eventually the work described in https://github.com/dotnet/sdk/issues/14044 will allow us to install runtime packs into our .NET 5 sandbox and we shouldn't have to use NuGet feeds for anything.

jonpryor: @pjcollins wrote:

> This PR adds a new `ExtractWorkloadPacks` msbuild target to extract the relevant workload .nupkg files into this new .NET 5 installation path for tests running both locally and in CI to use.

Does this mean that `$(AndroidToolchainDirectory)\dotnet` will be updated to include a copy of in-tree build artifacts as part of `msbuild /t:PackDotNet`?

If so, is there anything we should to to prevent "non-clean" environments in which `$(AndroidToolchainDirectory)\dotnet` contains "possibly corrupt" or "out-of-sync" file contents due to repeated invocations of `msbuild /t:PackDotNet` over time?

Then there's presumably the unit test scenario: won't `msbuild /t:PackDotNet` and an update of `$(AndroidToolchainDirectory)\dotnet` be required to run `dotnet build`-based unit tests?

I'm slightly concerned about the prospect of "corrupting" local `dotnet` installs which will necessitate nuking the `$(AndroidToolchainDirectory)\dotnet` directory, but I don't really have an alternative idea either. 😞

build-tools/create-packs/Directory.Build.targets
```diff
@@ -62,7 +62,41 @@
...
+    <ItemGroup>
+      <_UnixExecutables Include="$(DotNetPreviewPath)packs\Microsoft.Android.Sdk\*\tools\$(HostOS)\**\*.*" />
+    </ItemGroup>
+    <Exec
+        Condition="$([MSBuild]::IsOSPlatform('osx'))"
```
jonpryor, line 89: Shouldn't this also check for linux?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XASdkProject.cs
dsplaisted: For the dotnet/sdk repo, we keep our testing environment separate.  We have a "stage 0" copy of the .NET SDK which is installed under `<REPO ROOT>\.dotnet`.  When we build the repo it creates a "stage 2" version of the .NET SDK in the output folder.  That's what the tests run against.

We may be able to add some hooks which would allow you to set environment variables to point to local workloads without having to copy them into a dotnet install.

pjcollins: > Does this mean that $(AndroidToolchainDirectory)\dotnet will be updated to include a copy of in-tree build artifacts as part of msbuild /t:PackDotNet?

Yes it does. `/t:PackDotNet` is a convenience target for the Windows build + smoke job, and for local dev. It effectively wraps `Build`, `CreateAllPacks`, and `ExtractWorkloadPacks`.  The `/t:CreateAllPacks` target can still be used to create all of our .NET .nupkg files without also installing them into `$(AndroidToolchainDirectory)\dotnet\packs\`.

We have a couple of steps which will clean out `$(AndroidToolchainDirectory)\dotnet\packs` to ensure that there are no stale workloads left behind whenever we try to "install" a new one. 
1) `/t:ExtractWorkloadPacks` will delete all `Microsoft.Android.*` content in `$(AndroidToolchainDirectory)\dotnet\packs`, and 
2) `xaprepare` deletes `Android` packs when ensuring that the correct version of .NET Preview is installed:
https://github.com/xamarin/xamarin-android/blob/master/build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetPreview.cs#L26

> Then there's presumably the unit test scenario: won't msbuild /t:PackDotNet and an update of $(AndroidToolchainDirectory)\dotnet be required to run dotnet build-based unit tests?

For unit tests in CI, we download the .nupkgs artifact produced by the build job and then call the `/t:ExtractWorkloadPacks` to "install" them for our tests.

build-tools/create-packs/Directory.Build.targets
pjcollins: Thanks, I updated this to check $(HostOS) for either Darwin or Linux.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultItems.props
```diff
@@ -1,8 +1,16 @@
...
-    <EnableDefaultAndroidItems Condition=" '$(EnableDefaultAndroidItems)' == '' ">$(EnableDefaultItems)</EnableDefaultAndroidItems>
-  </PropertyGroup>
+<!--
+***********************************************************************************************
+Microsoft.Android.Sdk.DefaultItems.props
+WARNING:  This file is imported by AutoImport.props, and will be imported by all
+          projects using Microsoft.NET.Sdk.  All Item includes in this file *MUST* be
+          hidden behind an Android property based condition.
+
+Default item includes (globs and implicit references).
```
rolfbjarne, line 9: Why do we want to define globs and implicit references here (where they'll be imported by all projects using Microsoft.NET.Sdks) as opposed to in a .targets file later (which would only be imported when building for Android), when they will be behind an Android-specific property anyway?

jonathanpeppers: This is the design here:

https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workload-resolvers.md#workload-props-files

I think it because the wildcards need to be defined before the body of the `.csproj` file. @dsplaisted might know more details.

@pjcollins we could maybe move the contents of file to `AutoImport.props`, maybe would be clearer that it needs to check for `$(TargetPlatformIdentifier)` everywhere?

dsplaisted: Yes, the default items need to be included before the body of the project file in order for VS to be able to override them.  IE if there is a `.cs` file and in VS you change the build action to `Content` in the VS UI, it will add something like the following to your project file:

```xml
<ItemGroup>
    <Compile Remove="ActuallyContent.cs" />
    <Content Include="ActuallyContent.cs" />
</ItemGroup>
```

For the `Remove` to work, the item needs to have been included beforehand, so the default items need to go in .props files before the body of the project.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultItems.props
```diff
@@ -33,14 +41,4 @@
     />
   </ItemGroup>
```
jonathanpeppers, line 42: The `@(FrameworkReference)` in this file breaks builds for `netstandard2.0` libraries:

    Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(324,5): error NETSDK1073: The FrameworkReference 'Microsoft.Android' was not recognized

I'll help look into this.

dsplaisted: The condition on the ItemGroup also needs to check that the TargetFrameworkIdentifier is Android.

jonathanpeppers: I moved things around, so the `@(FrameworkReference)` isn't in `AutoImport.props` at all anymore. I think this will be fine?

dsplaisted: That should also work.

jonathanpeppers:
Something odd happened:
```
error MSB3932: Failed to unzip file "/Users/runner/work/1/s/bin/BuildRelease/nupkgs/Microsoft.Android.Workload.11.0.100-ci.pr.gh5195.206.nupkg" because the file does not exist or is inaccessible.
```

But the build artifact has a git "distance" of 205: `Microsoft.Android.Workload.11.0.100-ci.pr.gh5195.205.nupkg`.

I'll fix the merge conflict, maybe it will work on the next try.

jonathanpeppers (approved):

dellis1972 (approved):

jonpryor:
For review:

```markdown
Context: https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workloads.md

Add a new Microsoft.NET.Workload.Android NuGet package which contains
the `WorkloadManifest.json` and `WorkloadManifest.targets` files
required for the [workload resolver][1].

The Microsoft.Android.Sdk NuGet package has been converted from an
[MSBuild project SDK][2] to an SDK workload pack.  The most noticeable
difference in this change is that the `Sdk.props` import hook has been
replaced by an `AutoImport.props` file.  The `AutoImport.props` file
should not set any MSBuild properties, and should only contain
ItemGroup glob behavior that can be modified by the main project file.
The `AutoImport.props` file included in Microsoft.Android.Sdk will be
imported by all project using Microsoft.NET.Sdk once workloads are
enabled by default.  As a result, it is crucial that all of the content
in this file be conditionally included based on some Android specific
property value; in this case, when `$(EnableDefaultAndroidItems)`=True.

All other `.props` files from this repo which are included in the
Microsoft.Android.Sdk pack have been renamed to `.targets` files.
This was done to follow the convention that `.props` files are loaded
before the main project file, and `.targets` files are loaded after.

The only file that is loaded before the project file is
`AutoImport.props`.

All One .NET tests have been updated to use the Microsoft.NET.Sdk
project SDK, rather than Microsoft.Android.Sdk.  In order to support
this, a new `/t:ExtractWorkloadPacks` target has been added.
This target can be used `build-tools/create-packs/*.proj`, and is
responsible for extracting Microsoft.NET.Workload.Android,
Microsoft.Android.Sdk, and Microsoft.Android.Ref into our sandbox .NET
preview installation location (cdc35b60).  This target will also clean
up any stale Microsoft.Android.* packs that were previously installed.

The Xamarin.Android.Sdk.Lite package has been removed, as it continues
to be a bit of a headache to maintain moving forward.  Old versions of
this package will continue to be available on the internal NuGet feed as
needed, however teams should soon be able to switch over to our workload
for their own dev/test efforts.

Finally, I've bumped the .NET Preview version to 5.0.100-rtm.20509.5.

[1]: https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workload-resolvers.md
[2]: https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk
```

dsplaisted:
🎉 🎉 🎉

PR #5209, opened 2020-10-14
[CI] Skip required program installation for most test jobs

pjcollins:
The setup-test-environment.yaml template has been updated to reduce the
amount of dependency installation that happens by default for all test
jobs.  The majority of these jobs do not need all dependencies defined
by xaprepare, and we can save time provisioning by skipping these steps.


pjcollins:
Full test run for 9983e46: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4158188&view=results

grendello (approved):

Issue #5214, opened 2020-10-16
Compare current and NET5/6 release apks


radekdoulik:
Use `apkdiff` to analyze differences between current and NET5/6 apks for simple XA and XA/Forms apps.

This will show where we are with apk sizes and hopefully also information how illink output compares to old XA linker.

jonathanpeppers:
This is finished, closing.

PR #5215, opened 2020-10-16
[tests] Add test for apk size comparison

radekdoulik:
Context: https://github.com/xamarin/xamarin-android/issues/5214

Build simple XA and XA/Forms apps in tests with current and NET5/6 XA.
Do not remove the temporary projects, so that it is possible to compare
the apks after tests run.

The apps are built with single arm64 ABI, in Release configuration.

The test can be run like this.

current XA:

    msbuild Xamarin.Android.sln /t:RunNunitTests /p:TEST="Xamarin.Android.Build.Tests.BuildTest.BuildReleaseArm64"

NET5/6:

    dotnet test -v diag --filter BuildTest.BuildReleaseArm64 .\bin\TestDebug\netcoreapp3.1\Xamarin.Android.Build.Tests.dll

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -56,6 +56,21 @@ namespace Xamarin.Android.Build.Tests
 			}
 		}
 
+		[Test]
+		public void BuildReleaseArm64 ([Values (false, true)] bool forms)
+		{
+			var proj = forms ?
+				new XamarinFormsAndroidApplicationProject () :
+				new XamarinAndroidApplicationProject ();
+			proj.IsRelease = true;
+			proj.SetAndroidSupportedAbis ("arm64-v8a");
+
+			// use BuildHelper.CreateApkBuilder so that the test directory is not removed in tearup
+			using (var b = BuildHelper.CreateApkBuilder (Path.Combine ("temp", TestName))) {
+				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
+			}
+		}
+
 		static readonly object [] BuildHasNoWarningsSource = new object [] {
 			new object [] {
 				/* isRelease */     false,


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
```diff
@@ -67,6 +67,25 @@ public void DotNetBuildBinding ()
...
+			if (!forms)
+				proj.OtherBuildItems.Add (new AndroidItem.InputJar ("javaclasses.jar") {
+					BinaryContent = () => Convert.FromBase64String (InlineData.JavaClassesJarBase64)
+				});
+
+			proj.SetProperty (KnownProperties.RuntimeIdentifier, "android.21-arm64");
+
+			var dotnet = CreateDotNetBuilder (proj, false);
+			Assert.IsTrue (dotnet.Build (), "`dotnet build` should succeed");
+		}
```
jonathanpeppers, line 87: I think the `BuildTest.BuildReleaseArm64` test by itself will be sufficient to compare.

We multi-target this project:

https://github.com/xamarin/xamarin-android/blob/b3d2910af1b523fb86d8edeb0ba7770c9446ece9/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Xamarin.Android.Build.Tests.csproj#L4

If you run the `netcoreapp3.1` version, it generates SDK-style projects and builds with `dotnet build`.

We might be missing a way to run this from the command-line easily. I normally use the IDE.

radekdoulik: I tried to use the `netcoreapp3.1` version on cmd line and it doesn't work with nunit3-console for me. I am getting:
```
NUnit.Engine.NUnitEngineException : The NUnit 3 driver encountered an error while executing reflected code.
  ----> System.InvalidCastException : Unable to cast transparent proxy to type 'System.Web.UI.ICallbackEventHandler'.
--NUnitEngineException
The NUnit 3 driver encountered an error while executing reflected code.
```

So I would like to keep the `DotNetBuildReleaseArm64` test around.

BTW, do you remember why do we need the following part?
```c#
				proj.OtherBuildItems.Add (new AndroidItem.InputJar ("javaclasses.jar") {
					BinaryContent = () => Convert.FromBase64String (InlineData.JavaClassesJarBase64)
				});
```

jonathanpeppers: >  it doesn't work with nunit3-console for me

You'll have to use `dotnet test`:

    > dotnet test .\bin\TestDebug\netcoreapp3.1\Xamarin.Android.Build.Tests.dll

> why do we need the following part?

The `@(InputJar)` was in the other test to check that the binding worked:

https://github.com/xamarin/xamarin-android/blob/c81435813751eb9b77f11620513fd4d048b8e209/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs#L145-L151

You can take this part out, I don't think you need to test a binding.

radekdoulik: Nice! The `dotnet test` works perfectly. I have updated the changes and the description.

jonathanpeppers (approved):

PR #5219, opened 2020-10-19
Remove Preserve attribute from IntentService ctor

radekdoulik:
It shouldn't be needed as the service based on `IntentService` should
have the constructor without parameters and thus it would preserve its
base constructor.

In case someone declares Android service without public default
parameterless constructor, the build fails with:

    error XA4213: The type 'MyService' must provide a public default constructor

Context: The constructor is called when the service is created by
Android and it uses the parameterless constructor.

### DIFF
--- a/src/Mono.Android/metadata
+++ b/src/Mono.Android/metadata
@@ -852,11 +852,6 @@
   <remove-node path="/api/package[@name='junit.framework']/class[@name='TestSuite']/method[parameter[@type='java.lang.Class']]" api-since="16" />
      -->
 
-  <!-- Preserve Mono.Android.App.IntentService::.ctor if the
-       type is used anywhere. -->
-
-  <attr path="/api/package[@name='mono.android.app']/class[@name='IntentService']/constructor[count(parameter) = 0]" name="customAttributes">[Preserve(Conditional = true)]</attr>
-
   <!--    Asyncification    -->
 
   <!-- Interfaces -->


jonpryor:
@radekdoulik: I believe I have recreated the problem.  Imagine we have the following type:

```csharp
	[Service]
	public class MyService : IntentService {
		public MyService() : base ("foo") {
		}

		protected override void OnHandleIntent(Intent? intent)
		{
		}
	}
```

This passes the XA4213 check, and compiles, and elicits no "meaningful" warnings.  (The above *does* cause a `warning CS0618: 'IntentService' is obsolete:`, which I'm not considering meaningful in this context.)

The problem is that it cannot possibly work at runtime:

 1. `AndroidManifest.xml` does declare the service:

    ```xml
        <service android:name="crc64382e7c390968e9cf.MyService" />
    ```

2. The `MyService` C# type does contain a default constructor, thus there is no XA4213 error

3. *But* the `MyService` *Java Callable Wrapper* does *not* contain a default constructor!

    ```java
    package crc64382e7c390968e9cf;
    public /* partial */ class MyService extends mono.android.app.IntentService implements mono.android.IGCUserPeer
    {
    /** … */
    	public MyService (java.lang.String p0)
    	{
    		super (p0);
    		if (getClass () == MyService.class)
    			mono.android.TypeManager.Activate ("HelloWorld.MyService, HelloWorld", "System.String, mscorlib", this, new java.lang.Object[] { p0 });
    	}
    }
    ```

4. (3) means that the service *cannot be started* from [`Context.startService()`](https://developer.android.com/reference/android/content/Context#startService(android.content.Intent)).

There *is* a way to make this work, but it is "non-obvious": provide `ExportAttribute.SuperArgumentsString` on the default constructor:

```csharp
// C#
partial class MyService : IntentService {
    [Export(SuperArgumentsString="\"foo\"")]
    public MyService()
        : base ("foo")
    {
    }
}
```

The use of `[Export]` allows the Java Callable Wrapper to contain a ("manually provided") default constructor, and the use of `ExportAttribute.SuperArgumentsString` allows the emitted constructor to call the super `IntentService(String)` constructor as opposed to the super `IntentService()` default constructor.

jonpryor:
Aside: the beginnings of a commit message, lest I forget…

```markdown
Context: https://github.com/xamarin/xamarin-android/issues/5197
Context: https://github.com/xamarin/monodroid/commit/1dee1e50e0
Context: https://github.com/xamarin/monodroid/commit/f976c0ffec
Context: https://github.com/xamarin/Java.Interop/commit/a76edb8ca5b3059cf325352bdee396bd5a8cb354

(From the "we have no useful historical archives" and "wait, what?!"
departments…)

Commit xamarin/monodroid@f976c0ff updated `generator` to support the
`customAttributes` attribute within Metadata XML files, while commit
xamarin/monodroid@1dee1e50 updated `src/Mono.Anrdoid/metadata` to add
custom attributes to the `IntentService` constructor:

	<!-- Preserve Mono.Android.App.IntentService::.ctor if the type is used anywhere. -->
	<attr path="/api/package[@name='mono.android.app']/class[@name='IntentService']/constructor[count(parameter) = 0]" name="customAttributes">[Preserve(Conditional = true)]</attr>

While the relevant commit messages don't explicitly state why this
change was added, we believe we have re-learned why the default
`IntentService` constructor was preserved: so that if a subclass
*did not call* the default constructor, the resulting Java Callable
Wrapper would still contain a default constructor:

	// C#
	[Service]
	public class MyService : IntentService {
	    public MyService()
	        : base ("foo")
	    {
	    }

	    protected override void OnHandleIntent(Intent? intent)
	    {
	    }
	}

In the above `MyService` class, the `IntentService(string)` constructor
is called, *not* the `IntentService` default constructor.

Consequently, the resulting Java Callable Wrapper will only contain
a `MyService(String)` constructor:

	// Java
	public /* partial */ class MyService extends mono.android.app.IntentService
	{
	    public MyService (java.lang.String p0)
	    {
	        super (p0);
	        if (getClass () == MyService.class)
	            mono.android.TypeManager.Activate ("HelloWorld.MyService, HelloWorld", "System.String, mscorlib", this, new java.lang.Object[] { p0 });
	    }
	}

However, types mentioned within `AndroidManifest.xml` such as `Service`
subclasses *must* provide a default constructor *from Java*, which is
not the case here.
```

jonpryor:
@radekdoulik: given the [scenario described above](https://github.com/xamarin/xamarin-android/pull/5219#issuecomment-713219902), what is the best fix?

Only two solutions come to mind:

 1. We try to "keep" `[Preserve]`, "replacing" it semantically with [`DynamicDependencyAttribute`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.dynamicdependencyattribute?view=net-5.0), or
 2. We update the linker so that `IntentService` default constructor is preserved if the `IntentService` type is preserved, thus "mimicking in spirit" the `[Preserve]` attribute without actually using it.

radekdoulik:
I think good solution here would be using `DynamicDependencyAttribute` to preserve `IntentService()` when `IntentService (string)` is preserved. That should solve the issue with

    public MyService() : base ("name")
    {
    }

radekdoulik:
> But the MyService Java Callable Wrapper does not contain a default constructor!

Indeed, because JCW's are generated after the linking.

I wonder if there are other cases similar to this one with `IntentService`. If so then we might rather do it in linker and preserve default constructors of all classes derived from `Service` (and `Activity`, or even `Context`?)

jonpryor:
> I wonder if there are other cases similar to this one

I am not aware of any other cases like this, and I assume that if there were, someone would have filed an issue already.

Which should not be interpreted as a reason to *not* do this in the linker. That might be better anyway, or not; I'm not sure.

jonpryor:
Obsoleted by #9126.

Issue #5218, opened 2020-10-19
compiling Xamarin.Forms Android App fails from network share


MagicAndre1981:
### Steps to Reproduce

1. [share your local drive](https://support.microsoft.com/en-us/help/4092694/windows-10-file-sharing-over-a-network)
2. open the .sln with VS2019 from network share and not from local drive
3. try to compile the android app

### Expected Behavior

app compiles fine

### Actual Behavior

app fails to compile:

```
System.IO.DirectoryNotFoundException: Could not find a part of the path 'D:\foo.app\obj\Debug\foo.App.Android\AnyCPU\lp\1\jl\__res_name_case_map.txt'
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.File.InternalDelete(String path, Boolean checkHost)
   at System.IO.File.Delete(String path)
   at Xamarin.Android.Tools.Files.CopyIfStreamChanged(Stream stream, String destination)
   at Xamarin.Android.Tools.Files.ExtractAll(ZipArchive zip, String destination, Action`2 progressCallback, Func`2 modifyCallback, Func`2 deleteCallback, Func`2 skipCallback)
   at Xamarin.Android.Tasks.ResolveLibraryProjectImports.Extract(IDictionary`2 jars, ICollection`1 resolvedResourceDirectories, ICollection`1 resolvedAssetDirectories, ICollection`1 resolvedEnvironments)
   at Xamarin.Android.Tasks.ResolveLibraryProjectImports.RunTask()
   at Xamarin.Android.Tasks.AndroidTask.Execute()	foo.App.Android	C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets	1107	
```

this is a regression from VS2017, here the app compiles fine.

### Version Information

```
Microsoft Visual Studio Enterprise 2019
Version 16.7.6
VisualStudio.16.Release/16.7.6+30611.23
Microsoft .NET Framework
Version 4.8.03752

Xamarin   16.7.000.456 (d16-7-6@f8d55b6)
Visual Studio extension to enable development for Xamarin.iOS and Xamarin.Android.

Xamarin Designer   16.7.0.495 (remotes/origin/d16-7@79c0c522c)
Visual Studio extension to enable Xamarin Designer tools in Visual Studio.

Xamarin Templates   16.7.85 (1bcbbdf)
[...]

PR #5221, opened 2020-10-19
[Xamarin.Android.Build.Tasks] fix DirectoryNotFoundException for builds on network shares

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5218

Builds could fail on a Windows network share with:

    System.IO.DirectoryNotFoundException: Could not find a part of the path 'XYZ'
        at System.IO.File.InternalDelete(String path, Boolean checkHost)
        at System.IO.File.Delete(String path)
        at Xamarin.Android.Tools.Files.CopyIfStreamChanged(Stream stream, String destination)
        at Xamarin.Android.Tools.Files.ExtractAll(ZipArchive zip, String destination, Action`2 progressCallback, Func`2 modifyCallback, Func`2 deleteCallback, Func`2 skipCallback)
        at Xamarin.Android.Tasks.ResolveLibraryProjectImports.Extract(IDictionary`2 jars, ICollection`1 resolvedResourceDirectories, ICollection`1 resolvedAssetDirectories, ICollection`1 resolvedEnvironments)
        at Xamarin.Android.Tasks.ResolveLibraryProjectImports.RunTask()

Normally a call to `File.Delete()` completes silently if the file does
not exist:

> If the file to be deleted does not exist, no exception is thrown.

https://docs.microsoft.com/dotnet/api/system.io.file.delete#remarks

However, this does not appear to be the case for a file on a network
share. `DirectoryNotFoundException` is thrown?!?

`Files.CopyIf*Changed` is a hot path, that is called many times
throughout a build. A simple `File.Exists()` check solves the problem,
but we don't want to add additional I/O calls. I made
`MonoAndroidHelper.SetWriteable()` able to optionally skip the
`File.Exists()` check.

When reviewing this code, I fixed some other issues: :eyes:

* `MonoAndroidHelper.SetDirectoryWriteable()` did not have the correct
  enum flags logic to check for `FileAttributes.ReadOnly`.
* `MonoAndroidHelper.SetWriteable()` can avoid using the
  `System.IO.FileInfo` API, as it creates an object on the heap.
  `File.GetAttributes()` and `File.SetAttributes()` are more
  appropriate for this case.

### DIFF
--- a/Documentation/release-notes/5221.md
+++ b/Documentation/release-notes/5221.md
@@ -1,0 +1,5 @@
+#### Application and library build and deployment
+
+* [GitHub Issue 5218](https://github.com/xamarin/xamarin-android/issues/5218):
+  Fixed a *System.IO.DirectoryNotFoundException* that could occur when
+  building a project on a Windows network share.
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/MonoAndroidHelperTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/MonoAndroidHelperTests.cs
@@ -204,5 +204,34 @@ namespace Xamarin.Android.Build.Tests
 			Assert.AreSame (expected, actual);
 			Assert.AreEqual (0, actual.Length);
 		}
+
+		[Test]
+		public void SetWriteable ()
+		{
+			File.WriteAllText (temp, contents: "foo");
+			File.SetAttributes (temp, FileAttributes.ReadOnly);
+
+			MonoAndroidHelper.SetWriteable (temp);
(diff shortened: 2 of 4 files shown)


MagicAndre1981:
I've replaced the current xamarin data from 16.7 with files from the `Xamarin.Android.Sdk-OSS-11.1.99.179_remotes-pull-5221-merge_949267d.vsix`, now the build failed because of missing `system.codedom.dll`, so I copied the 32Bit `system.codedom.dll` from `Program Files (x86)\dotnet\shared\Microsoft.WindowsDesktop.App\3.1.9` to `MSBuild\Xamarin\Android` folder and build works from network drive

dellis1972 (approved):

PR #5227, opened 2020-10-20
[.vscode] Update tasks to use Configuration.

dellis1972:
When building in vscode it was hardcoded to build in
`Debug`. This is a problem is you have to build in
`Release` since you can't use the ide.

This commit adds the ability to choose the configuration
when you launch the task. It will default to `Debug`.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -5,5 +5,6 @@
       "ms-vscode.mono-debug",
       "wghats.vscode-nxunit-test-adapter",
       "visualstudioexptteam.vscodeintellicode",
+      "streetsidesoftware.code-spell-checker",
     ]
 }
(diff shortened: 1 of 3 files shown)


jonathanpeppers (commented):
Does the `configuration` variable make it to this other file, too?

https://github.com/xamarin/xamarin-android/blob/caca5c892ecc829049f89a2d34cf4a02e0b85b3c/.vscode/settings.json#L4-L7

dellis1972:
> Does the `configuration` variable make it to this other file, too?
> 
> https://github.com/xamarin/xamarin-android/blob/caca5c892ecc829049f89a2d34cf4a02e0b85b3c/.vscode/settings.json#L4-L7

I don't think so since it requires user interaction. You get a drop down to select from, so as far as I know I will only apply to tasks.json and launch.json

dellis1972:
That said I should update `launch.json` too in that case

jonathanpeppers (approved):
Tested in VS Code, seems to work:

image

PR #5229, opened 2020-10-20
[tests/MSBuildDeviceIntegration] Add test for dll's which have the same size.

dellis1972:
Context https://github.com/xamarin/monodroid/issues/1123

There seems to be a bug in the new fast deployment system
where if the dll was modified BUT had exactly the same size
as the existing dll on the device it would not be uploade.

This commit adds a test to make sure this does work as
expected.

### DIFF
--- a/tests/MSBuildDeviceIntegration/Tests/InstallTests.cs
+++ b/tests/MSBuildDeviceIntegration/Tests/InstallTests.cs
@@ -478,23 +478,25 @@ namespace Xamarin.Android.Build.Tests
 			AssertCommercialBuild ();
 			AssertHasDevices ();
 
+			var class1src = new BuildItem.Source ("Class1.cs") {
+				TextContent = () => "namespace Library1 { public class Class1 { public static int foo = 0; } }"
+			};
 			var lib1 = new XamarinAndroidLibraryProject () {
 				ProjectName = "Library1",
 				Sources = {
-					new BuildItem.Source ("Class1.cs") {
-						TextContent = () => "namespace Library1 { public class Class1 { } }"
-					},
+					class1src,
 				}
 			};
 
+			var class2src = new BuildItem.Source ("Class2.cs") {
+				TextContent = () => "namespace Library2 { public class Class2 { public static int foo = 0; } }"
+			};
 			var lib2 = new DotNetStandard {
 				ProjectName = "Library2",
 				Sdk = "Microsoft.NET.Sdk",
 				TargetFramework = "netstandard2.0",
 				Sources = {
-					new BuildItem.Source ("Class2.cs") {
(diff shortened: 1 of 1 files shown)


radekdoulik (commented):
For peace of mind I would add one more `Assert` with check that the 2 assembly files are indeed the same size?

tests/MSBuildDeviceIntegration/Tests/InstallTests.cs
```diff
@@ -504,15 +506,22 @@ public void IncrementalFastDeployment ()
...
+				Assert.IsTrue (logLines.Any (l => l.Contains ("NotifySync CopyFile") && l.Contains ("UnnamedProject.dll")), "UnnamedProject.dll should have been uploaded");
+				Assert.IsTrue (logLines.Any (l => l.Contains ("NotifySync CopyFile") && l.Contains ("Library1.dll")), "Library1.dll should have been uploaded");
+				Assert.IsTrue (logLines.Any (l => l.Contains ("NotifySync CopyFile") && l.Contains ("Library2.dll")), "Library2.dll should have been uploaded");
 				var firstInstallTime = builder.LastBuildTime;
+				builder.BuildLogFile = "install2.log";
 				Assert.IsTrue (builder.Install (app, doNotCleanupOnUpdate: true, saveProject: false), "Second install should have succeeded.");
 				var secondInstallTime = builder.LastBuildTime;
 
 				var filesToTouch = new [] {
-					Path.Combine (rootPath, lib1.ProjectName, "Class1.cs"),
```
jonathanpeppers, line 515: Maybe I'm missing something, but does `Class1.cs` not ever get `Touch()` called on it?

dellis1972: We change the actual TextContent here https://github.com/xamarin/xamarin-android/pull/5229/files#diff-34b4814429c211ae0bf929ddaa36bbc4eb8c13de248a744fa75e7a8f333fb9aeR533

jonathanpeppers: I think the `TimeStamp` has to change for it to save the `.cs` file?

dellis1972: I'll test that out

dellis1972: The timestamp is set by this line https://github.com/xamarin/xamarin-android/pull/5229/files#diff-34b4814429c211ae0bf929ddaa36bbc4eb8c13de248a744fa75e7a8f333fb9aeR539

jonathanpeppers (approved):

PR #5231, opened 2020-10-20
[Xamarin.Android.Build.Tasks] do not require JDK 11 for build-tools 30

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5234

I was building a project and hit:

```
error XA0032: Java SDK 11.0 or above is required when using Android SDK Build-Tools 30.0.2.
```

This is actually not true anymore since we got c50df1c5 and other
changes in to redistribute `apksigner.jar`. Let's drop this check, and
just require JDK 1.8.

I'm not actually sure how others haven't run into this?

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Legacy/ValidateJavaVersion.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Legacy/ValidateJavaVersion.cs
@@ -76,8 +76,6 @@ namespace Xamarin.Android.Tasks.Legacy
 			if (!Version.TryParse (buildToolsVersionString, out buildTools)) {
 				return Version.Parse (LatestSupportedJavaVersion);
 			}
-			if (buildTools >= new Version (30, 0, 0))
-				return new Version (11, 0);
 			if (buildTools >= new Version (24, 0, 1))
 				return new Version (1, 8);
 			return Version.Parse (MinimumSupportedJavaVersion);
--- a/Documentation/release-notes/5234.md
+++ b/Documentation/release-notes/5234.md
@@ -1,0 +1,6 @@
+#### Application and library build and deployment
+
+* [GitHub Issue 5234](https://github.com/xamarin/xamarin-android/issues/5234):
+  Fixed *error XA0032: Java SDK 11.0 or above is required when using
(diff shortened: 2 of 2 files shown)


dellis1972 (approved):

jonpryor:
Everything looks good, though do wonder about the Designer integration test failure.  "Random", or did we break something, or is it an "upstream" bug?

I've restarted those tests, so we can possibly eliminate "random"…

jonathanpeppers:
When I click on the first instance of the designer test failing, it seems like it happened on one of Radek's PRs that isn't merged. I think they probably have a test failing on xamarin/uitools/master.

image

jonathanpeppers:
Since #5324 came in, let me add a release note for it.

src/Xamarin.Android.Build.Tasks/Tasks/Legacy/ValidateJavaVersion.cs
```diff
@@ -76,8 +76,6 @@ Version GetJavaVersionForBuildTools ()
 			if (!Version.TryParse (buildToolsVersionString, out buildTools)) {
 				return Version.Parse (LatestSupportedJavaVersion);
 			}
-			if (buildTools >= new Version (30, 0, 0))
```
jonpryor, line 79: It occurs to me that instead of removing this check entirely, we should instead be "more accurate".  We *know* Build-tools 30.0.0 and 30.0.1 require JDK 11, it's 30.0.2 (and later?) that work with JDK 8.  Thus, should we check for what we know?

```csharp
if (buildTools >= new Version (30, 0, 0) && buildTools < new Version (30, 0 2)) {
    return new Version (11, 0);
}
```

jonpryor (changes requested):

src/Xamarin.Android.Build.Tasks/Tasks/Legacy/ValidateJavaVersion.cs
jonathanpeppers: Since we redistribute `apksigner.jar` now, both 30.0.0 and 30.0.1 will work.

PR #5245, opened 2020-10-28
[One .NET] Create Host OS specific SDK packs

pjcollins:
The Microsoft.Android.Sdk pack has been broken up into three Host OS
packs.  Depending on the operating system used during the build, one or
more of the following packs will now be created:

 * Microsoft.Android.Sdk.linux-x64
 * Microsoft.Android.Sdk.osx-x64
 * Microsoft.Android.Sdk.windows-x64

This is still mostly missing Linux support, which can be addressed in
a future PR that adds a linux build job to our commercial pipeline.

I've moved around a few ItemGroups that were used to determine
what ends up in our ref and runtime packs.  These were previously
declared in a shared targets file as the assembly content did not differ
between the ref and runtime packs, but now that we are producing
reference assemblies for certain projects we can move the item
declarations into the project files that consume them.

Some duplicate .props/.targets imports have also been cleaned up
throughout build-tools/create-packs/*.

### DIFF
--- a/build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
+++ b/build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
@@ -34,7 +34,7 @@
       <_FilesToCopy Include="$(DotNetPreviewPath)\sdk\$(DotNetPreviewVersionFull)\EnableWorkloadResolver.sentinel" />
       <_FilesToCopy Include="$(DotNetPreviewPath)\sdk-manifests\$(DotNetPreviewVersionBand)\Microsoft.NET.Workload.Android\**\*" />
       <_FilesToCopy Include="$(DotNetPreviewPath)\packs\Microsoft.Android.Ref\**\*" />
-      <_FilesToCopy Include="$(DotNetPreviewPath)\packs\Microsoft.Android.Sdk\**\*" />
+      <_FilesToCopy Include="$(DotNetPreviewPath)\packs\Microsoft.Android.Sdk.osx-x64\**\*" />
     </ItemGroup>
     <MakeDir Directories="$(PkgOutputPath);$(PayloadDir);$(PkgResourcesPath);$(LicenseDestination)"/>
     <ReplaceFileContents
--- a/build-tools/create-packs/Directory.Build.targets
+++ b/build-tools/create-packs/Directory.Build.targets
@@ -6,24 +6,14 @@
   </ItemGroup>
 
   <Import Project="..\..\Configuration.props" />
-  <Import Project="..\..\build-tools\scripts\XAVersionInfo.targets" />
+  <Import Project="..\..\build-tools\installers\create-installers.targets" />
   <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Build.Tasks.SharedFramework.Sdk" />
   <UsingTask TaskName="CreateFrameworkListFile" AssemblyFile="$(DotNetBuildTasksSharedFrameworkTaskFile)"/>
 
   <PropertyGroup>
+    <_MonoAndroidNETOutputDir>$(XAInstallPrefix)xbuild-frameworks\Microsoft.Android\netcoreapp3.1\</_MonoAndroidNETOutputDir>
     <_WorkloadResolverFlagFile>$(DotNetPreviewPath)sdk\$(DotNetPreviewVersionFull)\EnableWorkloadResolver.sentinel</_WorkloadResolverFlagFile>
   </PropertyGroup>
-
-  <ItemGroup>
-    <!-- NOTE: we don't have a reference assembly for Java.Interop.dll yet -->
-    <_AndroidAppRefAssemblies Include="$(XamarinAndroidSourcePath)bin\$(Configuration)\lib\xamarin.android\xbuild-frameworks\Microsoft.Android\netcoreapp3.1\Java.Interop.dll" />
-    <_AndroidAppRefAssemblies Include="$(XamarinAndroidSourcePath)bin\$(Configuration)\lib\xamarin.android\xbuild-frameworks\Microsoft.Android\netcoreapp3.1\ref\Mono.Android.dll" />
-    <_AndroidAppRefAssemblies Include="$(XamarinAndroidSourcePath)bin\$(Configuration)\lib\xamarin.android\xbuild-frameworks\Microsoft.Android\netcoreapp3.1\ref\Mono.Android.Export.dll" />
-    <_AndroidAppPackAssemblies Include="$(XamarinAndroidSourcePath)bin\$(Configuration)\lib\xamarin.android\xbuild-frameworks\Microsoft.Android\netcoreapp3.1\Java.Interop.dll" />
-    <_AndroidAppPackAssemblies Include="$(XamarinAndroidSourcePath)bin\$(Configuration)\lib\xamarin.android\xbuild-frameworks\Microsoft.Android\netcoreapp3.1\Mono.Android.dll" />
-    <_AndroidAppPackAssemblies Include="$(XamarinAndroidSourcePath)bin\$(Configuration)\lib\xamarin.android\xbuild-frameworks\Microsoft.Android\netcoreapp3.1\Mono.Android.Export.dll" />
-    <FrameworkListFileClass Include="@(_AndroidAppPackAssemblies->'%(Filename)%(Extension)')" Profile="Android" />
-  </ItemGroup>
 
   <!-- LICENSE setup -->
   <PropertyGroup>
@@ -42,7 +32,8 @@
     </ItemGroup>
   </Target>
 
- <!-- https://github.com/dotnet/runtime/blob/0647ec314948904319da5eb15e9931f7c85ed1e2/src/installer/pkg/projects/Directory.Build.targets#L281 -->
+  <!-- https://github.com/dotnet/runtime/blob/0647ec314948904319da5eb15e9931f7c85ed1e2/src/installer/pkg/projects/Directory.Build.targets#L281 -->
+  <!-- TODO: Generate PlatformManifest.txt files? -->
   <Target Name="_GenerateFrameworkListFile" >
     <!-- Hardcode framework attributes -->
     <ItemGroup>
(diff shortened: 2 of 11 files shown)


build-tools/create-packs/Microsoft.Android.Sdk.proj
```diff
@@ -38,37 +38,41 @@ sdk pack imported by Microsoft.NET.Workload.Android.
...
-        SourceFiles="@(AndroidSdkBuildToolsUnix)"
-        DestinationFiles="@(AndroidSdkBuildToolsUnix->'$(ToolsSourceDir)%(RelativePath)')"
-        SkipUnchangedFiles="True"
-        Condition="$([MSBuild]::IsOSPlatform('osx')) or $([MSBuild]::IsOSPlatform('linux'))"
     />
+
     <ItemGroup>
+      <!-- Microsoft.Android.Sdk.ILLink output -->
+      <_PackageFiles Include="$(XAInstallPrefix)xbuild\Xamarin\Android\Microsoft.Android.Sdk.ILLink.dll" PackagePath="tools" />
+      <_PackageFiles Include="$(XAInstallPrefix)xbuild\Xamarin\Android\Microsoft.Android.Sdk.ILLink.pdb" PackagePath="tools" />
```
pjcollins, line 75: @radekdoulik I've updated the output directory for `Microsoft.Android.Sdk.ILLink.csproj` so that we don't include extra files that are automatically copied to the output dir that aren't required.

Are there any other files that I am missing that we need to include in the SDK pack? I did some diffing locally and these are the only other assemblies that seemed like they _could_ be related, but the .NET 6 tests we have that use the linker all seem to be working with only `Microsoft.Android.Sdk.ILLink.dll`:

 * Mono.Cecil.dll
 * Mono.Cecil.Mdb.dll
 * Mono.Cecil.Pdb.dll
 * Mono.Cecil.Rocks.dll
 * System.Text.Json.dll

radekdoulik: If the tests are working, then it should be fine. The Cecil and Json probably come with the ILLink and the `Java.Interop.Export` and `Xamarin.Android.Build.Tasks` are already included?

pjcollins: Ok sounds good, Java.Interop.Export and Xamarin.Android.Build.Tasks are added via their explicit inclusions in `create-installers.targets`: https://github.com/xamarin/xamarin-android/blob/master/build-tools/installers/create-installers.targets#L142
https://github.com/xamarin/xamarin-android/blob/master/build-tools/installers/create-installers.targets#L241

build-tools/xa-prep-tasks/Xamarin.Android.BuildTools.PrepTasks/ReplaceFileContents.cs
```diff
@@ -32,7 +32,8 @@ public override bool Execute ()
 				}
 			}
 
-			File.Delete (DestinationFile.ItemSpec);
+			if (File.Exists (DestinationFile.ItemSpec))
+				File.Delete (DestinationFile.ItemSpec);
```
jonathanpeppers, line 36: I think the only reason you would need this is if the directory did not exist. Was this throwing `DirectoryNotFoundException`?

`File.Delete()` shouldn't throw if the file doesn't exist:

https://docs.microsoft.com/dotnet/api/system.io.file.delete#remarks

build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
```diff
@@ -34,7 +34,7 @@
       <_FilesToCopy Include="$(DotNetPreviewPath)\sdk\$(DotNetPreviewVersionFull)\EnableWorkloadResolver.sentinel" />
       <_FilesToCopy Include="$(DotNetPreviewPath)\sdk-manifests\$(DotNetPreviewVersionBand)\Microsoft.NET.Workload.Android\**\*" />
       <_FilesToCopy Include="$(DotNetPreviewPath)\packs\Microsoft.Android.Ref\**\*" />
-      <_FilesToCopy Include="$(DotNetPreviewPath)\packs\Microsoft.Android.Sdk\**\*" />
+      <_FilesToCopy Include="$(DotNetPreviewPath)\packs\Microsoft.Android.Sdk.osx-x64\**\*" />
```
jonathanpeppers, line 37: Does this need to check `$(HostOS)` and use either: `osx-x64`, `windows-x64`, or `linux-x64`?

build-tools/create-packs/Directory.Build.targets
```diff
@@ -87,15 +78,19 @@
...
   </Target>
 
   <Target Name="ExtractWorkloadPacks"
       DependsOnTargets="DeleteExtractedWorkloadPacks" >
     <ItemGroup>
       <_WLManifest Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nupkgs\Microsoft.NET.Workload.Android.*.nupkg" />
-      <_WLPacks Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nupkgs\Microsoft.Android.Sdk.*.nupkg" />
+      <_WLPacks Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nupkgs\Microsoft.Android.Sdk.linux-x64.*.nupkg"   Condition=" '$(HostOS)' == 'Linux' " />
+      <_WLPacks Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nupkgs\Microsoft.Android.Sdk.osx-x64.*.nupkg"     Condition=" '$(HostOS)' == 'Darwin' " />
+      <_WLPacks Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nupkgs\Microsoft.Android.Sdk.windows-x64.*.nupkg" Condition=" '$(HostOS)' == 'Windows' " />
```
jonathanpeppers, line 93: Should we be able to create all three of these packs macOS? Should the `mac_build` stage be able to create them all and upload & push them to a feed?

build-tools/xa-prep-tasks/Xamarin.Android.BuildTools.PrepTasks/ReplaceFileContents.cs
pjcollins: Yeah I was using this with a path where the destination directory didn't exist, I figured this was easier than checking if the directory existed.

build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
pjcollins: I think this is only used for macOS .pkg creation, though if we wanted to use this to create a .pkg installer for linux in the future we'd need to update this file to check `$(HostOS)` as suggested.

build-tools/create-packs/Directory.Build.targets
pjcollins: I need to spend some more time looking at what needs to go into a Linux specific SDK and do some testing there, I could maybe just exclude the Linux pack creation from this list for now.

jonathanpeppers: I see `Microsoft.Android.Sdk.windows-x64.*.nupkg` and `Microsoft.Android.Sdk.osx-x64.*.nupkg` as artifacts, that is probably OK for now.

build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
jonathanpeppers: I misread that is this the `create-dotnet-pkg` project.

radekdoulik (approved):
ILLink related changes LGTM

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
```diff
@@ -0,0 +1,26 @@
...
+      ]
+    }
+  },
+  "packs": {
+    "Microsoft.Android.Sdk": {
+      "kind": "sdk",
+      "version": "@SDK_PACK_VERSION@",
+      "alias-to": {
+        "osx-x64": "Microsoft.Android.Sdk.osx-x64",
+        "windows-x64": "Microsoft.Android.Sdk.windows-x64"
```
pjcollins, line 18: We'll need to change this back to win-x86 once https://github.com/dotnet/sdk/pull/14393 lands.

build-tools/create-packs/Microsoft.NET.Workload.Android.proj
```diff
@@ -9,15 +9,14 @@ workload manifest pack containing information about the various Microsoft.Androi
 <Project Sdk="Microsoft.Build.NoTargets">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
```
jonathanpeppers, line 12: Just an FYI, it seems like it doesn't matter what `$(TargetFramework)` is set to for a `Microsoft.Build.NoTargets` project?

https://github.com/microsoft/MSBuildSdks/tree/master/src/NoTargets

`$(TargetFramework)` just has to be valid and not blank... Maybe there is a way we could this value from a `Directory.Build.targets` at some point? Then we wouldn't need it in these files.

build-tools/create-packs/Microsoft.Android.Ref.proj
```diff
@@ -36,10 +34,18 @@ by projects that use the Microsoft.Android framework in .NET 5.
     </PropertyGroup>
 
     <ItemGroup>
-      <_PackageFiles Include="@(_AndroidAppRefAssemblies)" PackagePath="$(_AndroidRefPackAssemblyPath)" TargetPath="$(_AndroidRefPackAssemblyPath)" />
-      <_PackageFiles Include="$(XAInstallPrefix)xbuild-frameworks\Microsoft.Android\netcoreapp3.1\mono.android.jar" PackagePath="$(_AndroidRefPackAssemblyPath)" />
-      <_PackageFiles Include="$(XAInstallPrefix)xbuild-frameworks\Microsoft.Android\netcoreapp3.1\mono.android.dex" PackagePath="$(_AndroidRefPackAssemblyPath)" />
-      <_PackageFiles Include="$(XAInstallPrefix)xbuild-frameworks\Microsoft.Android\netcoreapp3.1\AndroidApiInfo.xml" PackagePath="$(_AndroidRefPackAssemblyPath)" />
+      <!-- NOTE: we don't have a reference assembly for Java.Interop.dll yet -->
+      <_AndroidRefPackAssemblies Include="$(_MonoAndroidNETOutputDir)Java.Interop.dll" />
```
jonathanpeppers, line 38: You might check if we have this file now since we bumped Java.Interop:

https://github.com/xamarin/java.interop/commit/6cde08775fbec5506c02c69857df0f0f75995c94

pjcollins: I think we may need to copy it from `external/Java.Interop/bin/$(Configuration)-netcoreapp3.1/ref/Java.Interop.dll`, it doesn't look like it's automatically copied to the Mono.Android output dir. Maybe there is some msbuild property that controls ref assembly copying for project references that I can find.

jonpryor:
> This is still mostly missing Linux support, which can be addressed in
> a future PR that adds a linux build job to our commercial pipeline.

Another benefit to the frequently suggested but never implemented suggestion of moving the proprietary bits *out* of the "normal" packages is that we won't *need* a Linux commercial build, as it can use the same proprietary bits as macOS/Windows!  (hint hint?)

build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
jonpryor, line 37: This reminds me of a previous conversation around how "safe" it is to use `**\*` globs, as it could possibly pull in files that we don't expect ("build cruft", files unexpectedly copied from the system mono, etc.).

Should this instead be a more "constrained" glob expression and/or specific files?  See e.g. https://github.com/xamarin/xamarin-android/pull/2733#discussion_r258326881

pjcollins: This wildcard is safe, as all the content in this path was extracted from `Microsoft.Android.Sdk.osx-x64.$(Version).nupkg`. All of the content being added to the SDK .nupkg file is explicitly defined through https://github.com/xamarin/xamarin-android/blob/master/build-tools/installers/create-installers.targets and https://github.com/xamarin/xamarin-android/blob/master/build-tools/create-packs/Microsoft.Android.Sdk.proj#L71.

This PR did fix one instance of us accidentally including extra build output in our packs by changing the output directory for Microsoft.Android.Sdk.ILLink.

build-tools/create-packs/Microsoft.Android.Ref.proj
pjcollins: I updated this file to use the external JI path, I couldn't find a built in way to copy this over to the Mono.Android netcore output directory.

build-tools/create-packs/Microsoft.NET.Workload.Android.proj
pjcollins: I think this is only used when packing, but we ignore it mostly since we set explicit paths for everything in the packs... We can try to centralize it in the future and reduce the warnings coming from our pack invocations at the same time.

jonpryor:
Commit message?

```markdown
The `Microsoft.Android.Sdk` pack has been broken up into three Host OS
packs.  Depending on the operating system used during the build, one
or more of the following packs will now be created:

  * `Microsoft.Android.Sdk.linux-x64`
  * `Microsoft.Android.Sdk.osx-x64`
  * `Microsoft.Android.Sdk.win-x64`

This is still mostly missing Linux support, which can be addressed in
a future PR that adds a linux build step to our commercial pipeline.

TODO:

Once we bump to a `dotnet` install which contains
[dotnet/sdk#14393][0], we'll need to replace mentions of
`windows-x64` with `win-x64`.

[0]: https://github.com/dotnet/sdk/pull/14393
```

jonathanpeppers (approved):

dellis1972 (approved):

PR #5250, opened 2020-10-29
[One .NET] Add Microsoft.Android-Tests unit tests

pjcollins:
A new version of Mono.Android-Tests has been added to run against the
.NET 6 SDK:

 * `tests/Mono.Android-Tests/Runtime-Microsoft.Android.Sdk/Mono.Android.NET-Tests.csproj`

To support this, new .NET .csproj files have been added alongside all of
the existing legacy Xamarin.Android unit test library projects.  These
projects target `net5.0-android` and build against our .NET 6 SDK:

 * `src/Mono.Android/Test/Java.Interop-Tests/Java.Interop-Tests.NET.csproj`
 * `src/Mono.Android/Test/Mono.Android-Test.Library/Mono.Android-Test.Library.NET.csproj`
 * `src/Xamarin.Android.NUnitLite/Xamarin.Android.NUnitLite.NET.csproj`
 * `tests/TestRunner.Core/TestRunner.Core.NET.csproj`
 * `tests/TestRunner.NUnit/TestRunner.NUnit.NET.csproj`

The `_CreateAar` target was being skipped for the new Java.Interop-Tests.NET
project, as it had no inputs.  As a result, the `Mono.Android-Test-classes.jar`
file required by these tests was not being included in the application project.
To fix this, I've added `@(AndroidJavaLibrary)` to `@(_CreateAarInputs)` so that
these items will also be considered as inputs for this target/task.

All tests which have been ignored are tied back to one of the following issues:

 * Mono.Data.Sqlite is not supported.
 * Secondary AppDomains are not supported.
 * Missing crypto/networking support in dotnet/runtime.
 * https://github.com/dotnet/runtime/issues/35836
 * https://github.com/dotnet/runtime/issues/44358
 * https://github.com/xamarin/xamarin-android/issues/5265
 * https://github.com/xamarin/xamarin-android/issues/5275


src/Mono.Android/Test/Java.Interop-Tests/Java.Interop-Tests.NET.csproj
```diff
@@ -0,0 +1,53 @@
+﻿<Project>
```
jonathanpeppers, line 1: I used the suffix `DotNet` on two projects already:

* https://github.com/xamarin/xamarin-android/blob/master/tests/CodeBehind/BuildTests/CodeBehindBuildTestsDotNet.csproj
* https://github.com/xamarin/xamarin-android/blob/master/tests/CodeBehind/CommonSampleLibrary/CommonSampleLibraryDotNet.csproj

But maybe `.NET` is better, do you want to also rename these two files?

src/Mono.Android/Test/System.Net/SslTest.cs
```diff
@@ -11,6 +11,7 @@
 namespace System.NetTests {
 
 	[TestFixture, Category ("InetAccess")]
+	[Category ("DotNetIgnore")] // Missing crypto / networking support.
```
jonathanpeppers, line 14: Maybe we should use `InetAccess` instead? Would all `InetAccess` tests fail without crypto?

src/Mono.Android/Test/System.Net/WebSocketTests.cs
```diff
@@ -10,6 +10,7 @@ namespace System.NetTests
 	public class WebSocketTests
 	{
 		[Test, Category ("InetAccess")]
+		[Category ("DotNetIgnore")] // Missing crypto / networking support.
```
jonathanpeppers, line 13: Yeah, I wonder if we could just ignore `InetAccess` for now.

tests/TestRunner.Core/TestRunner.Core.NET.csproj
```diff
@@ -0,0 +1,11 @@
+<Project Sdk="Microsoft.NET.Sdk" >
+
+  <PropertyGroup>
+    <TargetFramework>net5.0-android</TargetFramework>
+    <OutputType>Library</OutputType>
+    <RootNamespace>Xamarin.Android.UnitTests</RootNamespace>
+    <AssemblyName>TestRunner.Core.NET</AssemblyName>
```
jonathanpeppers, line 7: I'm curious can remove `$(AssemblyName)` here and it uses the name of the `.csproj` file?

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetPreview.cs
```diff
@@ -20,15 +20,6 @@ protected override async Task<bool> Execute (Context context)
...
 
-			// Delete any custom Microsoft.Android packs that may have been installed by test runs. Other ref/runtime packs will be ignored.
-			var packsPath = Path.Combine (dotnetPath, "packs");
-			if (Directory.Exists (packsPath)) {
-				foreach (var packToRemove in Directory.EnumerateDirectories (packsPath).Where (p => new DirectoryInfo (p).Name.Contains ("Android"))) {
-					Log.StatusLine ($"Removing Android pack: {packToRemove}");
-					Utilities.DeleteDirectory (packToRemove);
-				}
-			}
-
```
jonathanpeppers, line 31: Does it say somewhere why this was removed?

grendello: Indeed, why is this code going away?

src/Mono.Android/Test/Java.Interop-Tests/Java.Interop-Tests.NET.csproj
dellis1972: I would go with `.NET` now that I think about it. Since it will match the case of `Sdk="Microsoft.NET.Sdk"`.

src/Mono.Android/Test/Java.Interop/JnienvTest.cs
```diff
@@ -453,7 +454,9 @@ public ContainsExportedMethods ()
 			Constructed = true;
 		}
 
+#if !NET // https://github.com/xamarin/xamarin-android/issues/5265
 		[Export]
```
radekdoulik, line 458: I think it should be enough to disable the `CreateTypeWithExportedMethods`  test on dotnet?

tests/Runtime-Microsoft.Android.Sdk/Microsoft.Android-Tests.projitems
```diff
@@ -0,0 +1,33 @@
...
+    </TestApk>
+  </ItemGroup>
+
+  <ItemGroup>
+    <TestApkInstrumentation Include="xamarin.android.runtimetests.NUnitInstrumentation">
+      <Package>$(_PackageName)</Package>
+      <ResultsPath>$(OutputPath)TestResult-$(_PackageName).xml</ResultsPath>
+    </TestApkInstrumentation>
+
+    <TestApkPermission Include="READ_EXTERNAL_STORAGE">
```
dellis1972, line 25: Do these permissions still work on API 30?

src/Mono.Android/Test/Java.Interop/JnienvTest.cs
radekdoulik: nm. The build error is happening even when `Mono.Android.Export` is referenced. I will look at how to fix #5265

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetPreview.cs
pjcollins: Sorry I didn't explain this in the description -- it has to do with test environment setup ordering for the APK test job. The `setup-test-environment.yaml` template installs our packs into the dotnet 5 test location, however this subsequent `Required` scenario run will delete them:
https://github.com/xamarin/xamarin-android/blob/master/build-tools/automation/azure-pipelines.yaml#L399

As an alternative to removing this snippet, we could either:
1) Extract the packs again after running the Required scenario required for APK tests
2) Build and run xaprepare -s:Required before test environment setup

Maybe 2 is better here?

tests/Runtime-Microsoft.Android.Sdk/Microsoft.Android-Tests.projitems
pjcollins: I copied these over to follow the convention that the other test projects set, they don't seem to be causing any issues though

> Executing: /Users/runner/Library/Android/sdk/platform-tools/adb   shell pm grant Microsoft.Android_Tests android.permission.READ_EXTERNAL_STORAGE

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetPreview.cs
jonathanpeppers: I think as long as each CI stage deletes and create a fresh `~/android-toolchain/dotnet` directory, it will be fine.

I think we've seen a couple cases where leaving old `dotnet`'s around causes problems, such as:

https://github.com/dotnet/runtime/issues/42039

tests/Runtime-Microsoft.Android.Sdk/Microsoft.Android-Tests.csproj
```diff
@@ -0,0 +1,148 @@
+<Project Sdk="Microsoft.NET.Sdk" >
```
jonpryor, line 1: Should this file *also* be added to `Xamarin.Android-Tests.sln`?

tests/Runtime-Microsoft.Android.Sdk/Microsoft.Android-Tests.csproj
```diff
@@ -0,0 +1,148 @@
...
+    <ProjectReference Include="..\TestRunner.Core\TestRunner.Core.NET.csproj" />
+    <ProjectReference Include="..\TestRunner.NUnit\TestRunner.NUnit.NET.csproj" />
+    <ProjectReference Include="$(_MonoAndroidTest)Mono.Android-Test.Library\Mono.Android-Test.Library.NET.csproj" />
+    <ProjectReference Include="$(_MonoAndroidTest)Java.Interop-Tests\Java.Interop-Tests.NET.csproj" />
+  </ItemGroup>
+
+  <Import Project="Microsoft.Android-Tests.projitems" />
+  <Import Project="..\..\build-tools\scripts\TestApks.targets" />
+
+  <Target Name="GenerateNuGetConfig" >
```
jonpryor, line 96: *If* this file is added to `Xamarin.Android-Tests.sln`, could you please test the following?

1. Within Visual Studio for Mac,
2. Load `Xamarin.Android-Tests.sln`.
3. Add a new project to the `.sln`.

What happens?

*Normally*, what would happen. is that *every `.csproj` in the `.sln`* is "re-formatted", screwing up whitespace, and requiring a `git checkout -f HEAD …` to clean things up.

I don't know if this is still the case for .NET 5+ projects/SDK short-form projects.  This is the case for "normal" projects.

It would likely be safer to move these `<Target/>`s into `Directory.Build.targets`.

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetPreview.cs
pjcollins: WRT 42039 the SDK clean up logic will remain unchanged: https://github.com/xamarin/xamarin-android/blob/8b367fdba4507be34c074611c2fcd618b5ab4815/build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetPreview.cs#L26

I'll rework this a bit to keep the Android pack clean up logic untouched as well.

tests/Runtime-Microsoft.Android.Sdk/Microsoft.Android-Tests.csproj
pjcollins: I don't think we should add this to `Xamarin.Android-Tests.sln` at the moment, as it will break our macOS build job. These new project files will only compile against a specific version of .NET 5 that also has our workload installed into it. Additionally, I think there is still IDE work underway that will be needed to be able to load and use a project with a target framework of `net5.0-android`. I will move the targets over, but I don't think we should include any `net5.0-android` projects in our legacy solutions, or if we do want to do so eventually we should at least wait until there is a .NET 6 public preview available that supports it.

build-tools/xaprepare/xaprepare/Steps/Step_InstallDotNetPreview.cs
pjcollins: The required scenario is only used by this APK tests job, so I removed the .NET Preview installation step from this scenario. The .NET preview will still be installed as part of the test environment setup yaml template.

tests/Mono.Android-Tests/Directory.Build.targets
```diff
@@ -33,4 +33,52 @@
...
+  error NETSDK1148: Found multiple publish output files with the same relative path: external\Java.Interop\bin\Release\java-interop.jar, \external\Java.Interop\bin\Release-netcoreapp3.1\java-interop.jar.
+    We do not need to include any of this content in the output directory, so remove it here:
+  -->
+  <Target Name="RemoveJavaInteropJarContent"
+      Condition=" '$(TargetFramework)' == 'net6.0-android' "
+      BeforeTargets="ComputeResolvedFilesToPublishList" >
+    <ItemGroup>
+      <ResolvedFileToPublish Remove="$(JavaInteropSourceDirectory)\bin\**\java-interop.jar" />
+    </ItemGroup>
+  </Target>
```
pjcollins, line 46: This Target addition was the only change required after upgrading to .NET 6, to work around a new check that was added: https://github.com/dotnet/sdk/blob/955c0fc7b06e2fa34bacd076ed39f61e4fb61716/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ConflictResolution.targets#L97

jonathanpeppers: I think this is fine for now, can you file an issue and assign to me?

We might need to do something to clear out `@(Content)` and give a build warning. I think customers might assume these files would make it their `.apk`, and they do not.

pjcollins: I think we already have some sort of check here, I filed this for follow up outside of this PR: https://github.com/xamarin/xamarin-android/issues/5299

PR #5252, opened 2020-10-30
[One .NET] <CreateAar/> should generate resource case map

jonathanpeppers:
The following situation was broken in .NET 6 after the new `.aar` file
implementation in fcd7cf8:

* A class library includes `Resources\layout\Foo.axml` with an
  uppercase name.
* An application references the library and uses the
  `Resource.designer.cs` definition of `Resource.Layout.Foo`.
* `Resource.Layout.Foo` will have the incorrect integer value and
  crash at runtime!

Reviewing the `Resource.designer.cs` file of the library:

    // aapt resource value: 0x7F030000
    public static int Foo = 2130903040;

Compared to the app:

    // aapt resource value: 0x7F030000
    public const int Foo = 2130903040;
    // aapt resource value: 0x7F030001
    public const int foo = 2130903041;

The reason this works in "legacy" Xamarin.Android is the
`__res_name_case_map.txt` file that is generated in
`__AndroidLibraryProjects__.zip`:

https://github.com/xamarin/xamarin-android/blob/57c5a5fde5efd23f5958cfd8119b7f9c31d9e39d/src/Xamarin.Android.Build.Tasks/Tasks/CreateManagedLibraryResourceArchive.cs#L100-L105

If we generate the same `__res_name_case_map.txt` file and place it in
the `.aar` file we generate, this problem is solved.

We decided to place the file in a `.net` directory, as well as moving
the `.netenv` directory for `@(AndroidEnvironment)` files to `.net/env`:

    .net/__res_name_case_map.txt
    .net/env/MyAndroidEnvironment.txt

Going forward, we could potentially put other files in the `.net` folder.

I updated a test for this scenario; it checks we get the proper casing
for a `MyLayout.axml` in `Resource.designer.cs`.

### DIFF
--- a/Documentation/guides/OneDotNetEmbeddedResources.md
+++ b/Documentation/guides/OneDotNetEmbeddedResources.md
@@ -75,15 +75,27 @@ files. A Xamarin.Android library, `Foo.csproj`, could also generate an
         libs/*.jar
         jni/[arch]/*.so
 
-`@(AndroidEnvironment)` files are specific to Xamarin.Android, so we
-can make one addition to the file format:
+Additionally, there is a need for Xamarin.Android-specific files
+within the [`.aar`][aar]. These are placed in a `.net` directory:
 
     Foo.aar
-        .netenv/MyEnvironment.txt
+        .net/__res_name_case_map.txt
+        .net/env/MyEnvironment.txt
 
-The name `.netenv` would be unlikely to collide with anything Google
-creates in the `.aar` file format in the future. It should also be
-completely ignored by Android tooling.
+The `__res_name_case_map.txt` file includes the original casing of
+`@(AndroidResource)` files. This is used during `Resource.designer.cs`
+generation to ensure that the casing matches the original file for a
+layout such as `Resource.Layout.Foo`. It is the same file format
+included in `__AndroidLibraryProjects__.zip` in "legacy" Xamarin.Android.
+
+`@(AndroidEnvironment)` files will be placed in `.net/env` so they can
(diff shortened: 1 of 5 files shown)


dellis1972 (commented):
question. Do we still want to use the root for this file? or should be put this in a custom folder like we do for other files? It would mean having to change the code to read the file, but it might be cleaner going forward once we remove that old code.

jonathanpeppers:
The only other Xamarin.Android specific files are `@(AndroidEnvironment)`, which go in the `.aar` at:

    .netenv/MyEnvironment.txt

Then we pick up every `.netenv` file as `@(AndroidEnvironment)` here:

https://github.com/xamarin/xamarin-android/blob/15b37758e9d65c82aebe1a327747b285f38ce791/src/Xamarin.Android.Build.Tasks/Tasks/ResolveLibraryProjectImports.cs#L386-L392

Should we make the file `.net/__res_name_case_map.txt`? I don't think we can put it in the `res` folder as other Android tooling might break.

dellis1972:
I'm ok with a `.net` folder. It could be a catch all for any other performance/cache files we want to include as well maybe.

dellis1972 (approved):

jonpryor:
So we're potentially adding *both* `.netenv` and `.net` folders?  Could we instead "rename" `.netenv` to `.net/env`?

I'm assuming nothing right now is using `.netenv` yet.

jonathanpeppers:
I updated `@(AndroidEnvironment)` to go in `.net/env`. Nothing should be using it yet--we still have time to change it.

Issue #4789, opened 2020-06-09
@(JavaSourceJar) & JDK11


jonpryor:
`@(JavaSourceJar)` can be is a way to extract parameter name information from Javadoc files, and use that parameter name information within a `.jar` binding.

See e.g. https://github.com/xamarin/xamarin-android/blob/a111f6704d640481a6a85234ec4aa86ddbc18c3a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BindingBuildTest.cs#L434-L451

With the introduction of JDK 11 support https://github.com/xamarin/xamarin-android/pull/4567, `@(JavaSourceJar)` support is unfortunately broken, in part because JDK 11's `javadoc` utility emits different HTML, which isn't currently supported.

### Steps to Reproduce

1. Revert the `JavaSourceJar`-related changes from PR #4567
2. Run the `BindingBuildTest.JavaSourceJar()` unit test
3. Watch it fail.

### Expected Behavior

The unit test should work!

### Actual Behavior

It doesn't.

jonpryor:
In-progress (and incomplete) xamarin-android patch to fix `javadoc-to-mdoc.exe`:

```diff
diff --git a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Bindings.Documentation.targets b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Bindings.Documentation.targets
index ad1e824a..c31086f5 100644
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Bindings.Documentation.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Bindings.Documentation.targets
@@ -16,12 +16,7 @@ This file is only used by binding projects. .NET 5 can eventually use it, once `
   <UsingTask TaskName="Xamarin.Android.Tasks.MDoc"          AssemblyFile="Xamarin.Android.Build.Tasks.dll" />
   <UsingTask TaskName="Xamarin.Android.Tasks.Unzip"         AssemblyFile="Xamarin.Android.Build.Tasks.dll" />
 
-  <PropertyGroup>
-    <_JavadocSupported Condition=" $(_JdkVersion.StartsWith ('1.8')) ">True</_JavadocSupported>
-  </PropertyGroup>
-
[...]

Issue #5200, opened 2020-10-13
API docs are out of date


gaurangnirmal:
### Steps to Reproduce
1. AppOpsManager.CheckOp and AppOpsManager.CheckOpNoThrow are deprecated.
No documentation found for updated content.

### Expected Behavior
- Updated methods like, 
unsafeCheckOp(java.lang.String, int, java.lang.String).
unsafeCheckOpNoThrow(java.lang.String, int, java.lang.String).

and documentation regarding it, should be available.
Please refer below link for more info,
https://developer.android.com/reference/android/app/AppOpsManager#public-methods

### Actual Behavior
- Deprecated methods like, 
AppOpsManager.CheckOp(java.lang.String, int, java.lang.String).
AppOpsManager.CheckOpNoThrow(java.lang.String, int, java.lang.String).

Found deprecated in latest apis but no updat in documentations or anything regarding it.
Please Refer links for more info.
https://docs.microsoft.com/en-us/dotnet/api/android.app.appopsmanager.checkop?view=xamarin-android-sdk-9
https://docs.microsoft.com/en-us/dotnet/api/android.app.appopsmanager.checkopnothrow?view=xamarin-android-sdk-9

### Version Information
Visual Studio v16.7.4
XF v4.8.0.1364
I am targeting my project to API-29.

gaurangnirmal:
These methods are not accessible from code, too. Giving compilation errors.

jpobst:
Ah, I misunderstood and thought only the documentation was missing.

The methods are missing due to a binding issue.  I have submitted a PR that fixes the methods here: https://github.com/xamarin/xamarin-android/pull/5208.

PR #5253, opened 2020-10-30
[Mono.Android] Generate Mono.Android.xml

jonpryor:
Fixes: https://github.com/xamarin/xamarin-android/issues/4789

Context: https://github.com/xamarin/java.interop/commit/7574f166008bb45c0df97315aae7907ac25f8602
Context: https://github.com/xamarin/xamarin-android/issues/5200

What do we want?  Updated documentation!  Better Bindings!

How do we get that?  Uh…

Historically, [Xamarin.Android API docs][0] were produced by parsing
Android's HTML documentation from `docs-24_r01.zip` and converting it
into [**mdoc**(5) documentation][1] via `tools/javadoc2mdoc`.
The problem is that Google hasn't released an updated `docs*.zip`
package since API-24 (~6 years ago), and thus our documentation is
woefully out of date.

We could have scraped developer.android.com/reference within that
time frame, but web scraping is annoying, and we'd still have to deal
with the pain of parsing HTML.

There is an alternative, though: with API-30, there is a new `sources`
package which contains the Java source code used for the API level,
which in turn contains Javadoc source code comments.  Previous API
levels similarly contained an `android-stubs-src.jar` file which
likewise contained Java source code containing Javadoc comments.

The new approach is to:

 1. Update `build-tools/xaprepare` to install the `sources` package.

 2. Use [`java-source-tool.jar`][2] to parse the Java source code,
    creating an `android-javadoc.xml` file.

 3. [Update `generator`][3] to consume `android-javadoc.xml` and
    convert the Javadocs into [C# XML documentation comments][4].

 4. Update `src/Mono.Android` so that `generator` will emit
    C# XML doc comments, and then produce a `Mono.Android.xml` file.

The result is a `Mono.Android.xml` file which contains imported
Android Javadoc documentation, e.g.

	<doc>
	  <assembly><name>Mono.Android</name></assembly>
	  <members>
	    <member name="T:Java.Lang.Object">
	      <summary>Class <c>Object</c> is the root of the class hierarchy.</summary>
	      <remarks>
	        <para>Class <c>Object</c> is the root of the class hierarchy.
	…

"Later", `Mono.Android.xml` can then be used alongside
[`mdoc update --import=Mono.Android.xml`][5] to update our published
API documentation.

Generation of `Mono.Android.xml` is disabled by default on CI PR
builds, as generating `Mono.Android.xml` increases `src/Mono.Android`
build times by an unacceptable amount.

With the inclusion of `java-source-tools.jar`, we can now fix the
TODO mentioned in commit 380e95e3, and re-add support for
`@(JavaSourceJar)` when running under JDK 11.

Update `Xamarin.Android.Bindings.Core.targets` so that before
invoking `generator`, we first run `java-source-utils.jar` on the
`@(JavaSourceJar)` files, producing Javadoc XML files.
The `<BindingsGenerator/>` task in turn is updated to accept a new
`BindingsGenerator.JavadocXml` property, which is converted into a
`generator --with-javadoc-xml=FILE` option.

The `@(JavaSourceJar)` item group is "extended" to support the
following item metadata:

  * `%(CopyrightFile)`: A path to a file that contains copyright
    information for the Javadoc contents, which will be appended to
	all imported documentation.

  * `%(UrlPrefix)`: A URL prefix to support linking to online
    documentation within imported documentation.

  * `%(UrlStyle)`: The "style" of URLs to generate when linking to
    online documentation.  Only one style is currently supported:
    `developer.android.com/reference@2020-Nov`.

For .NET 6 ("One .NET") integration purposes, provide a default item
group of `@(JavaSourceJar)` which includes `**\*-source*.jar` and
`**\*-src.jar`.  This will allow `dotnet build` of an "Android Java
Library Binding" project (`dotnet new android-bindinglib`) to
automatically process `.java` source code for documentation
translation purposes.

Add a new `$(AndroidJavadocVerbosity)` MSBuild property, which
controls "how much" of the Javadoc comments are converted into C#
XML documentation.  Supported values are:

  * `full`: Convert as much Javadoc as possible.
  * `intellisense`: Only emit XML documentation for IDE-useful
    constructs: summary, parameters, returns, exceptions.

The difference between the two is *build time* impact: `full` takes
longer, and thus may not always be desirable.

Finally, add a new `$(_UseLegacyJavadocImport)` MSBuild property
which *disables* use of `java-source-utils.jar` for Javadoc importing,
and instead use the previous, legacy, doesn't work on JDK 11,
approach of using `javadoc` and HTML parsing.  This shouldn't be
needed, but just in case it *is* needed…

[0]: https://github.com/xamarin/android-api-docs
[1]: http://docs.go-mono.com/?link=man%3amdoc(5)
[2]: https://github.com/xamarin/java.interop/commit/69e1b80afd81ac502494e4bc2a2de75f5171c73f
[3]: https://github.com/xamarin/java.interop/pull/687
[4]: https://docs.microsoft.com/dotnet/csharp/codedoc
[5]: http://docs.go-mono.com/?link=man%3amdoc-update(1)


jonpryor:
Current local build times…

```
% time msbuild /v:diag /restore > b.txt
msbuild /v:diag /restore > b.txt  1250.71s user 241.59s system 185% cpu 13:25.65 total

% time msbuild /v:diag /restore /p:IncludeAndroidJavadoc=False > b2.txt
msbuild /v:diag /restore /p:IncludeAndroidJavadoc=False > b2.txt  632.16s user 25.03s system 289% cpu 3:47.05 total
```

It takes ~3.5x longer for me to build `Mono.Android.csproj` (~4min vs. ~13.5min) with it generating generate the C# XML documentation comments vs. the current system, and this *doesn't* include how long it takes to run `java-source-utils.jar` to generate the input `android-javadoc.xml` file, which is *not* fast.

I should look into disabling C# doc comments for (1) PR builds, and (2) "down-level" `$(TargetFrameworkVersion)` values.  This should help reduce the build time hit for PR builds.

jonpryor:
From the "Argh! Build Times!" department, the `make jenkins` step on this PR took 1h22min, while on e.g. PR#5311 the `make jenkins` step took 44min.  *Only* an increase of 38 minutes!

…Ouch.

jonpryor:
Back on the "OMG The Build Times!" front:

> I should look into disabling C# doc comments for (1) PR builds, and (2) "down-level" $(TargetFrameworkVersion) values. This should help reduce the build time hit for PR builds.

I disabled "down-level" `$(TargetFrameworkVersion)` builds.  (At least I think I did that correctly?)

`make jenkins` build times:

  * 5886bc1 (master): 34min 45sec
  * This PR: 1h 19min 47sec

*Ouch*.

Why?  Looking at the `msbuild-*-leeroy-all.binlog`, and processing with:

```
msbuild /v:diag '/clp:PerformanceSummary;ShowTimestamp' msbuild-*-leeroy-all.binlog > msbuild-leeroy-all.txt
```

then `msbuild-leeroy-all.txt` will contain a `Target Performance Summary:` section, which makes for an "interesting" comparison, for master (5886bc1)-vs-#5253:

```diff
-      80745 ms  _GenerateApiDescription                    2 calls
+     109191 ms  _GenerateApiDescription                    2 calls
```

35% slower for `_GeneraetApiDescription`.  I can't think of any actual reason for this; it could be hardware-related, e.g. "real" hardware vs. VM.

```diff
-     126573 ms  _BuildAndroidRuntimes                      1 calls
+     172634 ms  _BuildAndroidRuntimes                      1 calls
```

~36% slower for `_BuildAndroidRuntimes`, which I also wouldn't expect to be slower.  More fuel for the VM argument?  (Which doesn't entirely make sense, as other PRs have similar build times as master?)

This PR *adds* a new target, which isn't present in master:

```diff
+     179719 ms  _BuildAndroidJavadocXml                    2 calls
```

~3 minutes to process the API-30 sources and produce an `android-javadoc.xml` file.

```
-      44955 ms  _GenerateBinding                           2 calls
+     548532 ms  _GenerateBinding                           2 calls
-     114905 ms  CoreCompile                               66 calls
+     849433 ms  CoreCompile                               67 calls
```

*Here* we start seeing the cause of the increased build time:

`_GenerateBinding` takes 12 *times* longer -- *1200%* increase -- taking *9 minutes* to execute.  This was expected; `generator` *is* changed, after all, though *9 minutes* on CI is [rather longer than expected](https://github.com/xamarin/xamarin-android/pull/5253#issuecomment-727092906), vs ~45 seconds on master.  

*Unexpected* is `CoreCompile`, which takes 7.4 *times* longer to execute -- a *7400%* increase -- taking *14 minutes* to execute, vs. ~2 minutes on master.

Apparently `csc /doc` is *not* fast, especially when processing *218MB* of source code and producing a *72MB* `Mono.Android.xml` file.

Additionally, those Target times only amount to an increase of 21 minutes, which doesn't explain the *45 minute* increase in `make jenkins` time.  More analysis is required.

jonpryor:
Fortunately, commit **[ci] Don't convert Javadoc to Xmldoc for PR builds** did what I hoped: `make jenkins` for the PR build is back to ~35 minutes, so it does properly *skip* the import…

As an alternative approach, @jpobst [suggested](https://github.com/xamarin/java.interop/pull/687#issuecomment-743266123) that instead of trying to convert *all* the docs, we just do `<summary/>`, `<param/>`, and `<returns/>`.  I'll look into that approach, see how it impacts build times.

jonpryor:
I have a local prototype which just emits `<summary/>` and related information, skipping `<remarks/>`.

*Locally*:

  * `generator` takes ~10 seconds less time to run, 224s vs 233s.
  *  The resulting output directory is smaller, 109MB (1755905 lines) vs. 116MB (2073437 lines).

Then there's the build; no doc comments

```
git clean -xdf .
time msbuild /restore  /v:diag /p:IncludeAndroidJavadoc=False '/clp:PerformanceSummary;ShowTimestamp' > b.txt
# msbuild /restore /v:diag /p:IncludeAndroidJavadoc=False  > b.txt  1211.56s user 48.44s system 223% cpu 9:24.12 total
…
    49784 ms  _GenerateBinding                           2 calls
    71968 ms  _GenerateApiDescription                    2 calls
    123601 ms  _CheckApiCompatibility                     2 calls
    260118 ms  CoreCompile                               20 calls
    556707 ms  DispatchToInnerBuilds                      1 calls
```

Summary Remarks output:

```
git clean -xdf .
time msbuild /restore  /v:diag /p:IncludeAndroidJavadoc=True '/clp:PerformanceSummary;ShowTimestamp' > b.txt
# msbuild /restore /v:diag /p:IncludeAndroidJavadoc=True  > b.txt  1208.99s user 92.55s system 190% cpu 11:22.73 total
…
    73456 ms  _GenerateApiDescription                    2 calls
    98804 ms  _BuildAndroidJavadocXml                    2 calls
    99111 ms  CoreCompile                               20 calls
    354481 ms  _GenerateBinding                           2 calls
    675286 ms  DispatchToInnerBuilds                      1 calls
```

(Compare to [CI times](https://github.com/xamarin/xamarin-android/pull/5253#issuecomment-743232941)!)

Full Remarks Output:

```
git clean -xdf .
time msbuild /restore  /v:diag /p:IncludeAndroidJavadoc=True '/clp:PerformanceSummary;ShowTimestamp' > b.txt
# msbuild /restore /v:diag /p:IncludeAndroidJavadoc=True  > b.txt  2116.18s user 353.22s system 125% cpu 32:49.99 total
…
    73619 ms  _GenerateApiDescription                    2 calls
    101309 ms  _BuildAndroidJavadocXml                    2 calls
    864551 ms  _GenerateBinding                           2 calls
    869092 ms  CoreCompile                               21 calls
    1959969 ms  DispatchToInnerBuilds                      1 calls
```

*Locally*, we see that the "summary" version takes less than *half* the `_GenerateBinding` time, and less than an *8th* of the `CoreCompile` time.  It's still slower than *not doing anything at all*, which is to be expected, but instead of [`CoreCompile` taking ~7.4x longer](https://github.com/xamarin/xamarin-android/pull/5253#issuecomment-743232941), it instead takes 1.4x longer…*locally*.

jonpryor:
[*On CI*](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4310456&view=logs&s=1afc3bfe-122c-538b-e9ad-2a86c2efcfef&j=96fd57f5-f69e-53c7-3d47-f67e6cf9b93e), `make jenkins` took 43min 44s, vs. the build for PR #5434, in which `make jenkins` took 34min 31s.  Total **Mac** > **Build** time is 1h 1min 40s vs. 52min 34s, adding ~9 minutes to the build.

This feels far more reasonable.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
```diff
@@ -33,9 +33,10 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <!-- Default XPath transforms for bindings -->
     <TransformFile Include="Transforms\**\*.xml" />
     <!-- Default Java or native libraries -->
-    <AndroidLibrary       Include="**\*.jar" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <AndroidLibrary       Include="**\*.jar" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder);**\*-source?.jar;**\*-src.jar" />
```
jonathanpeppers, line 36: What does the `?` in `**\*-source?.jar` do?

jonpryor: It's a [filesystem glob](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-items?view=vs-2019#use-wildcards-to-specify-items); `?` matches zero or one characters.

In this case, it's to match `*-sources.jar` (plural) and `*-source.jar` (singular).  (And if someone has a `*-sourcez.jar`, that'll be grabbed too, arguably inadvertently, but I think that's highly unlikely.)

jonpryor: …or I misread! Looks like MSBuild documents `?` as "a single character", not zero-or-one.  Guess I need to be more explicit then. :-/

jonpryor: The globs have been moved into a new `$(_DefaultJavaSourceJarPattern)` property, which is used in both Include & Exclude, as required.

src/Mono.Android/Mono.Android.csproj
```diff
@@ -24,6 +24,20 @@
...
+    <IncludeAndroidJavadoc Condition=" '$(IncludeAndroidJavadoc)' == '' And '$(CONVERT_JAVADOC_TO_XMLDOC)' == 'true' And '$(AndroidFrameworkVersion)' == '$(AndroidLatestStableFrameworkVersion)'">True</IncludeAndroidJavadoc>
+    -->
+    <!-- TODO: replace with above; for CI timing purposes only! -->
+    <IncludeAndroidJavadoc Condition=" '$(IncludeAndroidJavadoc)' == '' And '$(AndroidFrameworkVersion)' == '$(AndroidLatestStableFrameworkVersion)'">True</IncludeAndroidJavadoc>
+    <AndroidJavadocVerbosity Condition=" '$(AndroidJavadocVerbosity)' == '' ">intellisense</AndroidJavadocVerbosity>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(IncludeAndroidJavadoc)' == 'True' ">
+    <DocumentationFile>$(OutputPath)Mono.Android.xml</DocumentationFile>
+    <NoWarn>CS1572;CS1573;CS1574;CS1587;CS1591;</NoWarn>
```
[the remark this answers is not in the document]
jonpryor: The `$(NoWarn)` has been added.

jonathanpeppers (approved):

jonpryor:
Squash-and-merge commit message (plus PR message updated):

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/4789

Context: https://github.com/xamarin/java.interop/pull/687
Context: https://github.com/xamarin/xamarin-android/issues/5200

What do we want?  Updated documentation!  Better Bindings!

How do we get that?  Uh…

Historically, [Xamarin.Android API docs][0] were produced by parsing
Android's HTML documentation from `docs-24_r01.zip` and converting it
into [**mdoc**(5) documentation][1] via [`tools/javadoc2mdoc`].
The problem is that Google hasn't released an updated `docs*.zip`
package since API-24 (~6 years ago), and thus our documentation is
woefully out of date.

We could have scraped developer.android.com/reference within that
time frame, but web scraping is annoying, and we'd still have to deal
with the pain of parsing HTML.

There is an alternative, though: with API-30, there is a new `sources`
package which contains the Java source code used for the API level,
which in turn contains Javadoc source code comments.  Previous API
levels similarly contained an `android-stubs-src.jar` file which
likewise contained Java source code containing Javadoc comments.

The new approach is to:

 1. Update `build-tools/xaprepare` to install the `sources` package.

 2. Use [`java-source-tool.jar`][2] to parse the Java source code,
    creating an `android-javadoc.xml` file.

 3. [Update `generator`][3] to consume `android-javadoc.xml` and
    convert the Javadocs into [C# XML documentation comments][4].

 4. Update `src/Mono.Android` so that `generator` will emit
    C# XML doc comments, and then produce a `Mono.Android.xml` file.

The result is a `Mono.Android.xml` file which contains imported
Android Javadoc documentation, e.g.

	<doc>
	  <assembly><name>Mono.Android</name></assembly>
	  <members>
	    <member name="T:Java.Lang.Object">
	      <summary>Class <c>Object</c> is the root of the class hierarchy.</summary>
	      <remarks>
	        <para>Class <c>Object</c> is the root of the class hierarchy.
	…

"Later", `Mono.Android.xml` can then be used alongside
[`mdoc update --import=Mono.Android.xml`][5] to update our published
API documentation.

Generation of `Mono.Android.xml` is disabled by default on CI PR
builds, as generating `Mono.Android.xml` increases `src/Mono.Android`
build times by an unacceptable amount.

With the inclusion of `java-source-tools.jar`, we can now fix the
TODO mentioned in commit 380e95e3, and re-add support for
`@(JavaSourceJar)` when running under JDK 11.

Update `Xamarin.Android.Bindings.Core.targets` so that before
invoking `generator`, we first run `java-source-utils.jar` on the
`@(JavaSourceJar)` files, producing Javadoc XML files.
The `<BindingsGenerator/>` task in turn is updated to accept a new
`BindingsGenerator.JavadocXml` property, which is converted into a
`generator --with-javadoc-xml=FILE` option.

The `@(JavaSourceJar)` item group is "extended" to support the
following item metadata:

  * `%(CopyrightFile)`: A path to a file that contains copyright
    information for the Javadoc contents, which will be appended to
	all imported documentation.

  * `%(UrlPrefix)`: A URL prefix to support linking to online
    documentation within imported documentation.

  * `%(UrlStyle)`: The "style" of URLs to generate when linking to
    online documentation.  Only one style is currently supported:
    `developer.android.com/reference@2020-Nov`.

For .NET 6 ("One .NET") integration purposes, provide a default item
group of `@(JavaSourceJar)` which includes `**\*-source*.jar` and
`**\*-src.jar`.  This will allow `dotnet build` of an "Android Java
Library Binding" project (`dotnet new android-bindinglib`) to
automatically process `.java` source code for documentation
translation purposes.

Add a new `$(AndroidJavadocVerbosity)` MSBuild property, which
controls "how much" of the Javadoc comments are converted into C#
XML documentation.  Supported values are:

  * `full`: Convert as much Javadoc as possible.
  * `intellisense`: Only emit XML documentation for IDE-useful
    constructs: summary, parameters, returns, exceptions.

The difference between the two is *build time* impact: `full` takes
longer, and thus may not always be desirable.

Finally, add a new `$(_UseLegacyJavadocImport)` MSBuild property
which *disables* use of `java-source-utils.jar` for Javadoc importing,
and instead use the previous, legacy, doesn't work on JDK 11,
approach of using `javadoc` and HTML parsing.  This shouldn't be
needed, but just in case it *is* needed…

[0]: https://github.com/xamarin/android-api-docs
[1]: http://docs.go-mono.com/?link=man%3amdoc(5)
[2]: https://github.com/xamarin/java.interop/commit/69e1b80afd81ac502494e4bc2a2de75f5171c73f
[3]: https://github.com/xamarin/java.interop/pull/687
[4]: https://docs.microsoft.com/dotnet/csharp/codedoc
[5]: http://docs.go-mono.com/?link=man%3amdoc-update(1)
```

src/Xamarin.Android.Build.Tasks/Tasks/JavaSourceUtils.cs
```diff
@@ -0,0 +1,181 @@
...
+			if (!string.IsNullOrEmpty (JavaOptions)) {
+				cmd.AppendSwitch (JavaOptions);
+			}
+
+			if (!string.IsNullOrEmpty (JavaMaximumHeapSize)) {
+				cmd.AppendSwitchIfNotNull("-Xmx", JavaMaximumHeapSize);
+			}
+
+			// Arguments sent to java.exe
+			cmd.AppendSwitchIfNotNull ("-jar ", Path.Combine (MonoAndroidToolsDirectory, "java-source-utils.jar"));
```
dellis1972, line 93: Do we want to hard code this `java-source-utils.jar` ? Or have it as a property like we do for `ToolName` in other tasks?

jonpryor: We should follow the other tasks.

dellis1972 (approved):

jonathanpeppers (approved):

AmrAlSayed0:
Great work here! 👍 Is this change gonna be released next year with .NET 6 or with the next XA release?

PR #5266, opened 2020-11-05
[Xamarin.Android.Build.Tasks] Fix up ZipArchiveEx to handle updates.

dellis1972:
While investigating another bug, I found that the code to look up
an existing entry in the zip was not working at all.

The problem was the filename we were looking up included a `/` at
the start of the path. The zip entires in the zip do not include a
`/` at the start. So the code was never finding a match. As a result
we were always trying to add a file rather than skipping those which
were up to date.

In addition I found that we do not always need to create entires in
the zip for directories. So when we create the `classes.zip` in the
`Javac` task we dont need to create directory entires. So we should
have a property to allow us to skip directories.

Also we should have the ability to turn the auto flush on/off.

There seems to be a small performance boost with these changes.

Old
```
749 ms  Javac                     
746 ms  Javac                   
```

New
```
728 ms  Javac 
724 ms  Javac 
```

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
@@ -8,10 +8,11 @@ namespace Xamarin.Android.Tasks
 	public class ZipArchiveEx : IDisposable
 	{
 
-		public static int ZipFlushLimit = 50;
+		public static int ZipFlushLimit = 500 * 1024 * 1024;
 
 		ZipArchive zip;
 		string archive;
+		long filesWrittenTotalSize = 0;
 
 		public ZipArchive Archive {
 			get { return zip; }
@@ -31,16 +32,15 @@ namespace Xamarin.Android.Tasks
 			zip = ZipArchive.Open(archive, filemode);
 		}
 
-		public void Flush (bool forceFlush = false)
+		public void Flush ()
 		{
-			if (!AutoFlush && !forceFlush)
-				return;
 			if (zip != null) {
 				zip.Close ();
 				zip.Dispose ();
 				zip = null;
 			}
 			zip = ZipArchive.Open (archive, FileMode.Open);
+			filesWrittenTotalSize = 0;
 		}
 
 		string ArchiveNameForFile (string filename, string directoryPathInZip)
@@ -58,14 +58,17 @@ namespace Xamarin.Android.Tasks
 			return pathName.Replace ("\\", "/").TrimStart ('/');
 		}
 
-		string TrimEntryName (string entryName, bool appendSeparator = false)
+		void AddFileAndFlush (string filename, long fileLength, string archiveFileName, CompressionMethod compressionMethod)
 		{
-			return entryName.Replace ("\\", "/").TrimStart ('/') + (appendSeparator ? "/" : "");
+			filesWrittenTotalSize += fileLength;
+			zip.AddFile (filename, archiveFileName, compressionMethod: compressionMethod);
+			if (filesWrittenTotalSize > ZipArchiveEx.ZipFlushLimit && AutoFlush) {
+				Flush ();
+			}
 		}
 
 		void AddFiles (string folder, string folderInArchive, CompressionMethod method)
 		{
-			int count = 0;
 			foreach (string fileName in Directory.GetFiles (folder, "*.*", SearchOption.TopDirectoryOnly)) {
 				var fi = new FileInfo (fileName);
 				if ((fi.Attributes & FileAttributes.Hidden) != 0)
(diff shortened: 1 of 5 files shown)


src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
```diff
@@ -52,18 +58,23 @@ string ArchiveNameForFile (string filename, string directoryPathInZip)
 			return pathName.Replace ("\\", "/");
 		}
 
+		string TrimEntryName (string entryName, bool appendSeparator = false)
+		{
+			return entryName.Replace ("\\", "/").TrimStart ('/') + (appendSeparator ? "/" : "");
+		}
```
jonathanpeppers, line 64: Could this be done in `ArchiveNameForFile`? Just so it's all in the same place.

dellis1972: not really, I don't think. its also used here https://github.com/xamarin/xamarin-android/pull/5266/files#diff-d2b91e191cec60e8b907d7d6ddb09d24e4ad619c57d7146537af332a13bfa4faR116. That logic if it was passed though `ArchiveNameForFile` would change the resulting directory.

I mean I can try it and see what breaks.

jonpryor:
@dellis1972 : please review some of the unit test failures: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4286406&view=ms.vss-test-web.build-test-results-tab

e.g. https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4286406&view=ms.vss-test-web.build-test-results-tab&runId=17549180&resultId=100251&paneView=debug

```
Expected string length 31 but was 32. Strings differ at index 18.
Expected: "temp/A.txt\ntemp/B/\ntemp/B/B.txt"
But was:  "temp/A.txt\ntemp/B//\ntemp/B/B.txt"
------------------------------^
```

src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
```diff
@@ -17,6 +17,10 @@ public class ZipArchiveEx : IDisposable
 			get { return zip; }
 		}
 
+		public bool AutoFlush { get; set; } = true;
```
jonpryor, line 20: Related suggestion from our 1:1: instead of `AutoFlush` causing `ZipArchiveEx` to "flush" contents after a set number of `Add()`s -- 50, iirc -- perhaps it should instead "flush" after a threshold *size* of added files.  The primary point to flushing was to reduce memory consumption, after all…

src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
```diff
@@ -17,6 +17,10 @@ public class ZipArchiveEx : IDisposable
...
+
+		public bool CreateDirectoriesInZip { get; set; } = true;
```
jonpryor, line 22: Naming: I'm not sure we need the `InZip` suffix; what else could it refer to?

Also, as discussed earlier, I don't think we need this *at all*: in what circumstances in our toolchain do we need an explicit directory entry in the first place?  We can probably just skip this property and the `zip.CreateDirectory (fullDirPath)` invocation entirely.

src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
```diff
@@ -102,9 +113,11 @@ public void AddDirectory (string folder, string folderInArchive, CompressionMeth
 				var internalDir = dir.Replace (folder, string.Empty);
 				string fullDirPath = folderInArchive + internalDir;
 				try {
-					zip.CreateDirectory (fullDirPath);
+					string dirPathInZip = TrimEntryName (fullDirPath, appendSeparator: true);
+					if (CreateDirectoriesInZip && !zip.ContainsEntry (dirPathInZip))
+						zip.CreateDirectory (dirPathInZip);
 				} catch (ZipException) {
-					
+
```
jonpryor, line 120: Shouldn't we *log* the exception?  Silent errors are a *menace*. 

(And *yes*, this should have been noted in 86888322a46261193d4006a97aedd34acee4f921.  My bad.)

src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
```diff
@@ -27,8 +31,10 @@ public ZipArchiveEx(string archive, FileMode filemode)
 			zip = ZipArchive.Open(archive, filemode);
 		}
 
-		public void Flush ()
+		public void Flush (bool forceFlush = false)
```
jonpryor, line 34: I don't understand this change.   If someone is calling `.Flush()`, why *shouldn't* we just flush?

dellis1972:
@jonpryor I will add a size limit , but I need to keep the file count limit as well because of this error

```
Xamarin.Tools.Zip.ZipException: Failure to create temporary file: Too many open files
```

which is :'(

src/Xamarin.Android.Build.Tasks/Utilities/ZipArchiveEx.cs
```diff
@@ -49,12 +58,20 @@ string ArchiveNameForFile (string filename, string directoryPathInZip)
...
 			}
-			return pathName.Replace ("\\", "/");
+			return pathName.Replace ("\\", "/").TrimStart ('/');
+		}
+
+		void AddFileAndFlush (string filename, long fileLength, string archiveFileName, CompressionMethod compressionMethod)
+		{
+			filesWrittenTotalSize += fileLength;
+			zip.AddFile (filename, archiveFileName, compressionMethod: compressionMethod);
+			if ((filesWrittenTotalSize > ZipArchiveEx.ZipFlushSizeLimit || filesWrittenTotalCount > ZipArchiveEx.ZipFlushFilesLimit) && AutoFlush) {
```
jonathanpeppers, line 68: This is really minor, but the other places are doing `>=` and this does `>`.

Other places have:
```csharp
if (count >= ZipArchiveEx.ZipFlushFilesLimit) {
```

jonathanpeppers (approved):

jonathanpeppers:
You might need to rerun some tests later, I see provisionator failures.

PR #5269, opened 2020-11-05
Bump to .NET 6.0.100-alpha.1.20562.2

jonathanpeppers:
Changes: https://github.com/dotnet/installer/compare/53e0c8c7...efac3f25

This version of .NET 6 includes support for
`$(AfterMicrosoftNETSdkTargets)`, and so we now have a slightly
different import ordering:

1. `AutoImport.props`
2. The user's `.csproj` file
3. Some early parts of `Microsoft.NET.Sdk` / `Sdk.targets`
4. `Microsoft.Android.Sdk.targets`
5. The rest of `Microsoft.NET.Sdk` / `Sdk.targets`
6. `Microsoft.Android.Sdk.After.targets`

Anything that needs to happen *after* the `Microsoft.NET.Sdk` targets
can go in `After.targets`.

This allows us to set up some defaults that we lost in 91498d55 when
implementing Android as a .NET "workload":

* `$(GenerateDependencyFile)` defaults to `false`
* `$(ProduceReferenceAssembly)` defaults to `false` for application
  projects.

Using `$(SelfContained)` also triggers new behavior in .NET 6 that we
need to turn off:

* `$(GenerateRuntimeConfigurationFiles)` to `false` or we get a
  `*.runtimeconfig.dev.json` file generated for Android applications.
* `$(CopyLocalLockFileAssemblies)` to `false` - otherwise every `.dll`
  file is copied to `$(OutputPath)`, and we don't need this behavior.
  This appears to be new behavior in .NET 6 when `$(SelfContained)` is
  `true`.
* `$(ComputeNETCoreBuildOutputFiles)` to `false` or we get a `*.exe`
  .NET app host in the output directory.

I also adjusted the naming for `windows-x64` to `win-x64`, as
described in 011a7713.

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -5,8 +5,7 @@
     <!-- ensure only the sources defined below are used -->
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" protocolVersion="3" />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
-    <add key="dotnet internal feed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" protocolVersion="3" />
-    <add key="illink" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet5/nuget/v3/index.json" />
+    <add key="dotnet6" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json" />
 
      <!-- This is needed (currently) for the Xamarin.Android.Deploy.Installer dependency, getting the installer -->
      <!-- Android binary, to support delta APK install -->
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
@@ -13,12 +13,15 @@
     <AndroidEnableSGenConcurrent Condition=" '$(AndroidEnableSGenConcurrent)' == '' ">true</AndroidEnableSGenConcurrent>
     <AndroidHttpClientHandlerType Condition=" '$(AndroidHttpClientHandlerType)' == '' ">Xamarin.Android.Net.AndroidClientHandler</AndroidHttpClientHandlerType>
     <AndroidUseIntermediateDesignerFile Condition=" '$(AndroidUseIntermediateDesignerFile)' == '' ">true</AndroidUseIntermediateDesignerFile>
+    <GenerateDependencyFile Condition=" '$(GenerateDependencyFile)' == '' ">false</GenerateDependencyFile>
+    <GenerateRuntimeConfigurationFiles Condition=" '$(GenerateRuntimeConfigurationFiles)' == '' ">false</GenerateRuntimeConfigurationFiles>
+    <CopyLocalLockFileAssemblies Condition=" '$(CopyLocalLockFileAssemblies)' == '' ">false</CopyLocalLockFileAssemblies>
+    <ComputeNETCoreBuildOutputFiles Condition=" '$(ComputeNETCoreBuildOutputFiles)' == '' ">false</ComputeNETCoreBuildOutputFiles>
     <!-- jar2xml is not supported -->
     <AndroidClassParser>class-parse</AndroidClassParser>
     <!-- mono-symbolicate is not supported -->
     <MonoSymbolArchive>false</MonoSymbolArchive>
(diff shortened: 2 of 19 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -65,6 +65,8 @@ variables:
   DotNetNUnitCategories: '& TestCategory != DotNetIgnore & TestCategory != AOT & TestCategory != MkBundle & TestCategory != MonoSymbolicate & TestCategory != PackagesConfig & TestCategory != StaticProject'
   NUnit.NumberOfTestWorkers: 4
   GitHub.Token: $(github--pat--vs-mobiletools-engineering-service2)
+  # Temporary workaround for: https://github.com/dotnet/msbuild/issues/5863
+  MSBuildDisableFeaturesFromVersion: 16.10
```
jonathanpeppers, line 69: We probably shouldn't merge this until we get a build with a fix for: https://github.com/dotnet/msbuild/issues/5863

It would require everyone to set this env var when doing development.

jonathanpeppers: /cc @rolfbjarne this is that env var if you need it.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -13,12 +13,15 @@
     <AndroidEnableSGenConcurrent Condition=" '$(AndroidEnableSGenConcurrent)' == '' ">true</AndroidEnableSGenConcurrent>
     <AndroidHttpClientHandlerType Condition=" '$(AndroidHttpClientHandlerType)' == '' ">Xamarin.Android.Net.AndroidClientHandler</AndroidHttpClientHandlerType>
     <AndroidUseIntermediateDesignerFile Condition=" '$(AndroidUseIntermediateDesignerFile)' == '' ">true</AndroidUseIntermediateDesignerFile>
+    <GenerateDependencyFile Condition=" '$(GenerateDependencyFile)' == '' ">false</GenerateDependencyFile>
+    <GenerateRuntimeConfigurationFiles Condition=" '$(GenerateRuntimeConfigurationFiles)' == '' ">false</GenerateRuntimeConfigurationFiles>
+    <CopyLocalLockFileAssemblies Condition=" '$(CopyLocalLockFileAssemblies)' == '' ">false</CopyLocalLockFileAssemblies>
+    <ComputeNETCoreBuildOutputFiles Condition=" '$(ComputeNETCoreBuildOutputFiles)' == '' ">false</ComputeNETCoreBuildOutputFiles>
```
jonathanpeppers, line 19: @rolfbjarne you might also want to set all these, I was getting lots of files in the build output:
* `.json` files
* every `.dll` file
* an `.exe` file that looked like a .NET app host

grendello (approved):

pjcollins (approved):

radekdoulik (approved):

NuGet.config
```diff
@@ -6,7 +6,7 @@
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" protocolVersion="3" />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
     <add key="dotnet internal feed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" protocolVersion="3" />
```
akoeplinger, line 8: you can probably remove this feed too

jonathanpeppers: I think it was originally added for this tool:

https://github.com/xamarin/xamarin-android/blob/a22a0368dc791a8fa8d505a8b52c9e1f3906339a/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs#L146

I'll check if it works to remove it, though.

jonathanpeppers: It does seem to be in this other feed:
```
OK https://pkgs.dev.azure.com/dnceng/9ee6d478-d288-47f7-aacc-f6e6d082ae6d/_packaging/1a5f89f6-d8da-4080-b15f-242650c914a8/nuget/v3/flat2/microsoft.dotnet.apicompat/5.0.0-beta.20181.7/microsoft.dotnet.apicompat.5.0.0-beta.20181.7.nupkg 261ms
```
I'll remove it.

jonathanpeppers:
Hit this twice on NuGet restore, trying one last time:
```
[ERROR] FATAL UNHANDLED EXCEPTION: System.InvalidOperationException: Failed to compare two elements in the array. ---> System.Threading.ThreadAbortException: Thread was being aborted.
```
I think it is a random failure, but maybe I should try putting the `NuGet.config` back the way it was?

PR #5270, opened 2020-11-06
[l10n] Use Arcade to generate resource properties

brendanzagaeski:

Now that `Xamarin.Android.Build.Tasks.csproj` is a short-form project,
add a dependency on the Microsoft.DotNet.Arcade.Sdk NuGet package and
switch to using the [`GenerateResxSource` mechanism][0] instead of using
`%(EmbeddedResource.Generator)`=`ResXFileCodeGenerator`.

For this initial use of Microsoft.DotNet.Arcade.Sdk, import only
`..\tools\BuildTasks.props` and `..\tools\GenerateResxSource.targets`.
Avoid importing the full `Sdk.props` and `Sdk.targets` for now because
those files have other consequences that aren't desired at this time.
For example, `Sdk.props` indirectly imports `ProjectDefaults.props`,
which sets `$(DebugType)`=`embedded` by default.

An advantage of `GenerateResxSource` compared to `ResXFileCodeGenerator`
is that contributors no longer need to worry about updating
`Resources.Designer.cs` when updating `Resources.resx`.  This is
especially helpful for contributors who use an editor other than Visual
Studio on Windows because it means they no longer need to hand-edit the
`Resource.Designer.cs` file to match the format that
`ResXFileCodeGenerator` generates on Windows.

The generated
`obj/$(Configuration)/Xamarin.Android.Tasks.Properties.Resources.cs`
file has essentially the same setup as the previous
`Resources.Designer.cs` file.  The properties are still `internal` and
still include `<summary/>` comments to surface the non-localized
messages in IntelliSense, and the file is still updated by a design-time
build any time the `.resx` file changes.

One difference is that the new generated `.cs` file is not shown in the
Solution Explorer because it's located in `obj/`.  If any contributors
were previously using `Resources.Designer.cs` as a starting place to do
"find all references" on the property names, they would now need to open
the generated file from the `obj/` directory instead.

[0]: https://github.com/dotnet/arcade/blob/c6f24c0d1264f6f80ceb39d2b0477d054c3f8ffc/Documentation/ArcadeSdk.md#generateresxsource-bool


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
```diff
@@ -1,6 +1,7 @@
 <Project>
 
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+  <Import Project="..\tools\BuildTasks.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
```
brendanzagaeski, line 4: The use of relative paths in an `Sdk` import probably looks a little weird, but it seems to work beautifully smoothly. The NuGet restore process downloads the SDK package as desired, and the properties are imported into this project as expected.

Avoiding the full `Sdk.props` and `Sdk.targets` is helpful because those files have other consequences. For example, `Sdk.props` indirectly imports `ProjectDefaults.props`, which sets `$(DebugType)`=`embedded` by default.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
```diff
@@ -17,6 +18,7 @@
     <AndroidGeneratedClassDirectory Condition=" '$(AndroidGeneratedClassDirectory)' == '' ">..\..\src\Mono.Android\obj\$(Configuration)\monoandroid10\android-$(AndroidLatestStablePlatformId)\mcw</AndroidGeneratedClassDirectory>
     <LangVersion>latest</LangVersion>
     <NoWarn>8632</NoWarn>
+    <GenerateResxSource>true</GenerateResxSource>
```
brendanzagaeski, line 21: It's not directly related to this source line, but one nice feature of `GenerateResxSource` is that it can optionally create methods like `string FormatXA0001 (object p0)` for resources that contain format strings. So after this initial setup PR for `GenerateResxSource` is merged, Xamarin.Android could consider enabling that by setting `$(GenerateResxSourceEmitFormatMethods)`=`true` and updating uses of resources containing format items to use the new generated methods.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
```diff
@@ -445,24 +447,10 @@
...
-    <EmbeddedResource Update="Properties\Resources.resx">
-      <Generator>ResXFileCodeGenerator</Generator>
-      <LastGenOutput>Resources.Designer.cs</LastGenOutput>
-    </EmbeddedResource>
-  </ItemGroup>
-
   <!-- Import Microsoft.NET.Sdk targets before our targets so we can override behavior -->
   <!-- See https://github.com/microsoft/msbuild/pull/4922 -->
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
+  <Import Project="..\tools\GenerateResxSource.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
```
brendanzagaeski, line 453: In theory, this project could also now import `..\tools\Localization.targets` instead of using an explicit `<PackageReference/>` for XliffTasks, but `..\tools\Localization.targets` makes some additional assumptions about other properties that need to be set. One way to split the difference would be to import just `..\tools\DefaultVersions.Generated.props` to get `$(XliffTasksVersion)` from the Microsoft.DotNet.Arcade.Sdk NuGet package, but even then, since the change isn't specifically required for this PR, it's probably cleanest to consider it in a separate PR.

jonathanpeppers: Is there documentation on what files you can import like this? It seems to me that the names of these files might change between versions of Microsoft.DotNet.Arcade.Sdk, unless they expect people to import them directly.

They have a lot of documentation here, it just doesn't seem to be organized too well:

https://github.com/dotnet/arcade/tree/master/Documentation

jonathanpeppers: Ok, I played around with:
```xml
<Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
...
<Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
```
It seems like this Sdk does a lot more things than we need:
* computes many types of version numbers
* signs with a Microsoft .snk
* generates `[assembly:InternalsVisibleTo]`
* sets up `Test` targets

Using this diff worked, but I don't think we should do this:

```diff
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
@@ -1,7 +1,7 @@
 <Project>

   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
-  <Import Project="..\tools\BuildTasks.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
+  <Import Project="Sdk.props" Sdk="Microsoft.DotNet.Arcade.Sdk" />
   <Import Project="..\..\Configuration.props" />
   <Import Project="..\..\build-tools\scripts\LocalizationLanguages.projitems" />
   <Import Project="..\..\build-tools\scripts\MSBuildReferences.projitems" />
@@ -19,6 +19,7 @@
     <LangVersion>latest</LangVersion>
     <NoWarn>8632</NoWarn>
     <GenerateResxSource>true</GenerateResxSource>
+    <SignAssembly>false</SignAssembly>
   </PropertyGroup>

   <ItemGroup>
@@ -450,7 +451,7 @@
   <!-- Import Microsoft.NET.Sdk targets before our targets so we can override behavior -->
   <!-- See https://github.com/microsoft/msbuild/pull/4922 -->
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
-  <Import Project="..\tools\GenerateResxSource.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
+  <Import Project="Sdk.targets" Sdk="Microsoft.DotNet.Arcade.Sdk" />
   <Import Project="Xamarin.Android.Build.Tasks.targets" />
```

These attributes got added:
```
[assembly: AssemblyInformationalVersion("1.0.0-dev")]
[assembly: AssemblyCopyright("© Microsoft Corporation. All rights reserved.")]
[assembly: AssemblyFileVersion("42.42.42.42424")]
[assembly: AssemblyVersion("42.42.42.42")]
```
Some obsession with 42 here?

Is there another team using this for localization-only? I looked at dotnet/sdk, and they use it to do all of the above.

If we want to use this, I think what you did is the best option, since we don't want all this other stuff in here.

dellis1972: 42 is the answer is it not?  :P

I don't like the idea of doing all that extra stuff. Having looked at the code, there appears no way to turn bits off via properties if you import the `Sdk.targets`.

brendanzagaeski: Just to throw out one other idea, it looks like another option could instead be to copy just the one file for the desired task ([_GenerateResxSource.cs_](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/src/GenerateResxSource.cs)) into xamarin-android-tools (so it can be used by all the Xamarin.Android repos), probably also along with the _.targets_ file. Based on a hunch, I tried a quick experiment of copying that single _GenerateResxSource.cs_ file into an empty C# project, and it did build, so it looks like it doesn't currently depend on anything else in Arcade.

At least one disadvantage would be that any potential fixes from upstream would need to be copied by hand into xamarin-android-tools instead of just updating the NuGet package. (There might also be a few details to work out about how to consume the task from xamarin-android-tools via the build process of each repository.) On the other hand, it would provide some more transparency in cases where an API change or file rename in new Arcade versions might break xamarin-android's direct import of _GenerateResxSource.targets_. If there's precedent for preferring the single-file approach, maybe it'd be an acceptable option here too?

brendanzagaeski: I just realized the single-file approach would be messy in the case of LibZipSharp because LibZipSharp would also need its own separate copy of the task and the targets, unless it added a submodule dependency on xamarin-android-tools, and adding that dependency doesn't seem right for LibZipSharp because that library is not specific to Xamarin.Android. So I'm thinking that copying the single file around is probably uglier overall than importing the specific files from the NuGet package.

If we wanted to either use the current approach of this PR temporarily or abandon it for now but try to come back to using `GenerateResxSource` again in the future, one possible option might be to see if the the Arcade maintainers would be OK with splitting that task into its own NuGet package eventually, or see if they'd be OK documenting API stability for `..\tools\GenerateResxSource.targets`.

jonpryor:
Should we move forward with this? @jonathanpeppers appears to think we can: https://discordapp.com/channels/732297728826277939/732297837953679412/775780326014058560

> Importing a couple specific files seemed ok

Perhaps this needs more investigation that it doesn't introduce unexpected assembly changes?

brendanzagaeski:
With the current approach of the PR, it's possible to audit the entire contents of [_tools\GenerateResxSource.targets_](https://github.com/dotnet/arcade/blob/5537c88d462e239a42375e19cc0ec2ab1ebeec64/src/Microsoft.DotNet.Arcade.Sdk/tools/GenerateResxSource.targets) and [_tools\BuildTasks.props_](https://github.com/dotnet/arcade/blob/5537c88d462e239a42375e19cc0ec2ab1ebeec64/src/Microsoft.DotNet.Arcade.Sdk/tools/BuildTasks.props) imports, along with the [_GenerateResxSource.cs_](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/src/GenerateResxSource.cs) task. There is nothing in those that would change the output assemblies. In particular, the _.targets_ only touches item lists and properties that apply to _*.resx_ files. In the current version of Arcade, those files don't import any other dependencies, so there can't be anything hiding elsewhere that would change the assemblies unexpectedly.

One gotcha would be if at some point in the future Xamarin.Android.Build.Tasks added another _.resx_ file that for some reason should not have a generated _.cs_ file. With the current setup, all the _.resx_ files would automatically be included in the _.cs_ file generation.

My tentative vote would be to move ahead with the current approach of this PR. At least in theory, it wouldn't be too difficult to back out these changes if it's decided later on that it is actually more maintenance trouble than it's worth. Mostly, backing it out would mean regenerating the _Resources.Desginer.cs_ file from within Visual Studio and adding it back to the repo.

jonathanpeppers (approved):
I'm good with this if everyone else is.

The best part is the lines removed:

image

Issue #5271, opened 2020-11-06
[illink] Set TrimMode of support and compatibility assemblies in SdkOnly linker mode


radekdoulik:
Currently the support and compatability assemblies are not linked. In future they will inform illink themselves about linkability.

Meanwhile set the TrimMode for them to link.

This can be removed once NET6 versions of packages containing them will be released.

jonathanpeppers:
Assemblies can do this now:
```
[assembly: AssemblyMetadata ("IsTrimmable", "True")]
```
Such as: https://github.com/xamarin/AndroidX/pull/247/files#diff-b885226e4f4904e7fd5e9962c2b93df8a656026c6312f31c4d25f745804ab5dcR10

I think we can close this now.

PR #5272, opened 2020-11-06
[illink] Set TrimMode of support assemblies

radekdoulik:
Context: https://github.com/xamarin/xamarin-android/issues/5271

Temporarily set trim mode of support assemblies to `link`.

apk size comparison of XF BuildReleaseArm64True test after the change, current monodroid vs net5:
```
Summary:
  +          29 Other entries 0.00% (of 750,919)
  -         116 Dalvik executables -0.00% (of 3,455,436)
  +   1,604,215 Assemblies 33.66% (of 4,765,637)
  -   6,488,608 Shared libraries -61.38% (of 10,570,992)
  -     394,325 Package size difference -3.73% (of 10,572,124)
```
apk size comparison of XF BuildReleaseArm64True test before and after the change (net5):
```
Summary:
  -         184 Other entries -0.02% (of 751,132)
  -      35,860 Dalvik executables -1.03% (of 3,491,180)
  -   1,179,296 Assemblies -15.62% (of 7,549,148)
  -     144,888 Shared libraries -3.43% (of 4,227,272)
  -   1,210,846 Package size difference -10.63% (of 11,388,645)
```
So it reduces the apk size by cca 1.2M.

For completeness, the difference between monodroid and net5 before the change:
```
Summary:
  +         213 Other entries 0.03% (of 750,919)
  +      35,744 Dalvik executables 1.03% (of 3,455,436)
  +   2,783,443 Assemblies 58.41% (of 4,765,705)
  -   6,360,784 Shared libraries -60.08% (of 10,588,056)
  +     815,915 Package size difference 7.72% (of 10,572,730)
```
Note that the apk is now smaller in net5 compared to monodroid. OTOH the assembly size is still larger by cca 16%. That could be later partially improved by linking the compatibility assemblies coming from nugets.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
@@ -22,7 +22,7 @@ This file contains the .NET 5-specific targets to customize ILLink
       />
       <!-- TODO: remove setting the trim mode here, once the support packages are updated to NET6 and compatability packages not needed -->
       <ResolvedFileToPublish
-          Condition=" '$(AndroidLinkMode)' == 'SdkOnly' and ( $([System.String]::Copy(%(Filename)).StartsWith ('System.')) or $([System.String]::Copy(%(Filename)).StartsWith ('Xamarin.Android')) ) ">
+          Condition=" '$(AndroidLinkMode)' == 'SdkOnly' and ( $([System.String]::Copy(%(Filename)).StartsWith ('Xamarin.AndroidX.')) or $([System.String]::Copy(%(Filename)).StartsWith ('Xamarin.Android.Support.')) or $([System.String]::Copy(%(Filename)).StartsWith ('Xamarin.Google.')) or $([System.String]::Copy(%(Filename)).StartsWith ('Xamarin.GooglePlayServices.')) ) ">
         <TrimMode>link</TrimMode>
       </ResolvedFileToPublish>
       <!-- Mark our entry assembly as a root assembly. -->
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
@@ -20,6 +20,11 @@ This file contains the .NET 5-specific targets to customize ILLink
           Condition=" '$(AndroidLinkMode)' == 'Full' and '%(ResolvedFileToPublish.Extension)' == '.dll' and '%(ResolvedFileToPublish.AssetType)' != 'native' "
           TrimMode="link"
       />
+      <!-- TODO: remove setting the trim mode here, once the support packages are updated to NET6 and compatability packages not needed -->
+      <ResolvedFileToPublish
+          Condition=" '$(AndroidLinkMode)' == 'SdkOnly' and ( $([System.String]::Copy(%(Filename)).StartsWith ('System.')) or $([System.String]::Copy(%(Filename)).StartsWith ('Xamarin.Android')) ) ">
+        <TrimMode>link</TrimMode>
+      </ResolvedFileToPublish>
       <!-- Mark our entry assembly as a root assembly. -->
       <TrimmerRootAssembly Include="$(AssemblyName)" />
     </ItemGroup>


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
```diff
@@ -20,6 +20,11 @@ This file contains the .NET 5-specific targets to customize ILLink
           Condition=" '$(AndroidLinkMode)' == 'Full' and '%(ResolvedFileToPublish.Extension)' == '.dll' and '%(ResolvedFileToPublish.AssetType)' != 'native' "
           TrimMode="link"
       />
+      <!-- TODO: remove setting the trim mode here, once the support packages are updated to NET6 and compatability packages not needed -->
+      <ResolvedFileToPublish
+          Condition=" '$(AndroidLinkMode)' == 'SdkOnly' and ( $([System.String]::Copy(%(Filename)).StartsWith ('System.')) or $([System.String]::Copy(%(Filename)).StartsWith ('Xamarin.Android')) ) ">
```
jonathanpeppers, line 25: Should we make this a little more specific and check if it starts with:
* `Xamarin.Android.Support.`
* `Xamarin.AndroidX.`
* `Xamarin.Google.` - https://www.nuget.org/packages/Xamarin.Google.Android.Material/
* `Xamarin.GooglePlayServices.`

jonathanpeppers, line 25: Did you need the `$([System.String]::Copy())` here? I think you only need it when batching also had to call something like `.Replace()`, like we did here:

https://github.com/xamarin/xamarin-android/blob/b864f93df36dc1d8dcf11b5219d37c4c7681baa3/build-tools/create-packs/Directory.Build.targets#L107

dellis1972: I'm not sure using `System` is a good idea here. Users can name assemblies what every they want, so `System.Foo.UI` is a perfectly valid assembly name. 

 is there existing Metadata we can leverage to detect a) system assemblies (like FrameworkFileName? ) and support packages. 

for example, can we get away with using the `_AndroidAssemblySkipCases`  item group from https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Xamarin.Android.SkipCases.projitems#L24 to drive this.

jonathanpeppers: I think @radekdoulik is trying to get assemblies like these that are compat packages on NuGet:

https://github.com/xamarin/net6-samples/blob/c4a4f4a38f062f385bd4ed98602e8d9f9764b33e/HelloForms/Directory.Build.props#L70-L86

I don't think you can use `%(FrameworkFileName)`, because these are normal NuGet packages.

radekdoulik: I tried various ways to call `StartsWith` on item's `%Filename` and it didn't work for me. In the end I went with the `[System.String]` call.

Do you know a simpler expression which would work here?

radekdoulik: It is temporary solution, so I didn't want to complicate it much. OTOH, who knows how long it will stay that way, so I will refine it a bit.

Maybe I will just use a list of known assemblies?

jonathanpeppers: Maybe you always need it when batching... I see the `%()` now.

jonathanpeppers: If you look at `Xamarin.GooglePlayServices` or `Xamarin.AndroidX`, there are a lot! It might be a couple hundred.

radekdoulik: OK, made it mode strict about support libraries names.

For compat assemblies I will create a list of names. Probably from typeforwarders collected from `mscorlib`.

I will put that list to the next PR.

jonathanpeppers (approved):
Since, this is hopefully temporary for .NET 6, I think it will be fine for now. 👍

Issue #5263, opened 2020-11-04
[illink] System.Collections.Concurrent.dll linked away causes test to fail


jonathanpeppers:
This test fails for `Release` builds:

https://github.com/xamarin/xamarin-android/blob/7bf2216f36719b2ae8ec92c7768fccf7f1cae57f/tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs#L31-L61

The app crashes on startup with:
```
11-04 14:48:51.847 17530 17530 E mono    : Unhandled Exception:
11-04 14:48:51.847 17530 17530 E mono    : System.TypeLoadException: Could not set up parent class, due to: Could not load type of field 'Java.Interop.JniRuntime:TrackedInstances' (31) due to: Could not load file or assembly 'System.Collections.Concurrent, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. assembly:Java.Interop.dll type:JniRuntime member:(null)
11-04 14:48:51.848 17530 17530 E mono-rt : [ERROR] FATAL UNHANDLED EXCEPTION: System.TypeLoadException: Could not set up parent class, due to: Could not load type of field 'Java.Interop.JniRuntime:TrackedInstances' (31) due to: Could not load file or assembly 'System.Collections.Concurrent, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. assembly:Java.Interop.dll type:JniRuntime member:(null)
```

I'm thinking that something in `System.Collections.Concurrent` is being used in Java.Interop through using `ViewTreeObserver.GlobalLayout`, but it was completely linked away.

It might also be a bug in `ILLink`, so we might file upstream if that is the case: https://github.com/mono/linker

Here is the test project zipped up, it might help diagnose:

[Bug29730-True.zip](https://1drv.ms/u/s!Apmn9psh-S8hnWki4fzkfD5Hxksp?e=hvhxh2)

jonathanpeppers:
This test crashes in the exact same way, too:

https://github.com/xamarin/xamarin-android/blob/7bf2216f36719b2ae8ec92c7768fccf7f1cae57f/tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs#L274-L275

So maybe this is a more general problem?

radekdoulik:
Observation from attached test archive:

[...]

PR #5276, opened 2020-11-06
[One .NET] better logic for deduplication of architecture-specific assemblies

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5263

In .NET 6, we currently run the `ILLink` MSBuild target for each
`$(RuntimeIdentifier)`, as the Mono runtime packs have some .NET
assemblies that differ per architecture.

To make this happen, the `<ProcessAssemblies/>` MSBuild task works
with input such as:

* `android.21-arm\linked\Foo.dll`
* `android.21-x86\linked\Foo.dll`
* `android.21-arm\linked\System.Private.CoreLib.dll`
* `android.21-x86\linked\System.Private.CoreLib.dll`

And "merges" this to what we need in the output of the `.apk`:

* `assemblies\Foo.dll`
* `assemblies\armeabi-v7a\System.Private.CoreLib.dll`
* `assemblies\x86\System.Private.CoreLib.dll`

`Foo.dll` is identical for all architectures, so we only need to use a
single copy of it.

A bug in our implementation was for an assembly when building for
three RIDs:

* `android.21-arm\linked\System.Collections.Concurrent.dll`
* `android.21-arm64\linked\System.Collections.Concurrent.dll`
* `android.21-x86\linked\System.Collections.Concurrent.dll`

Two of these assemblies have the same MVID... This didn't quite work.

So I reworked the logic to instead do:

* Group by assembly file name
* Count the unique MVIDs
* If we have only 1 MVID, we just need the first assembly.
* Otherwise, place the assemblies in architecture-specific directories.

The resulting code seems like it is a bit simpler, and likely a little
faster. I was able to call `MonoAndroidHelper.HasMonoAndroidReference()`
once per file name, which should help performance.

I updated `XASdkTests` to check the broken scenario.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
@@ -38,7 +38,7 @@ namespace Xamarin.Android.Tasks
 
 		public override bool RunTask ()
 		{
-			var output = new Dictionary<Guid, ITaskItem> ();
+			var output = new List<ITaskItem> ();
 			var symbols = new Dictionary<string, ITaskItem> ();
 
 			if (ResolvedSymbols != null) {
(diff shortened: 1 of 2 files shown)


dellis1972 (approved):

radekdoulik (approved):

src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
```diff
@@ -47,53 +47,60 @@ public override bool RunTask ()
...
+					using (var pe = new PEReader (File.OpenRead (assembly.ItemSpec))) {
+						var reader = pe.GetMetadataReader ();
+						var module = reader.GetModuleDefinition ();
+						var mvid = reader.GetGuid (module.Mvid);
+						mvids.Add (mvid);
+
+						// Calculate %(FrameworkAssembly) and %(HasMonoAndroidReference) for the first
+						if (frameworkAssembly == null) {
+							string packageId = assembly.GetMetadata ("NuGetPackageId");
+							frameworkAssembly = packageId.StartsWith ("Microsoft.NETCore.App.Runtime.") ||
```
jonpryor, line 65: Can `packageId` be `null`?  Should we use `?.StartsWith(…)`?

jonathanpeppers: `GetMetadata` should return `string.Empty` if not found:

* https://github.com/dotnet/msbuild/blob/7452552ce911efdca6aea5a189c409f083db7bc7/src/Utilities/TaskItem.cs#L116-L140
* https://github.com/dotnet/msbuild/blob/7452552ce911efdca6aea5a189c409f083db7bc7/src/Utilities/TaskItem.cs#L407-L428

But we could add a `null` check, as it is still possible.

jonpryor:
```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/5263

In .NET 6, we currently run the `ILLink` MSBuild target for each
`$(RuntimeIdentifier)`, as the Mono runtime packs have some .NET
assemblies that differ per architecture.

To make this happen, the `<ProcessAssemblies/>` task (51fb93e1)
works with inputs such as:

  * `android.21-arm\linked\Foo.dll`
  * `android.21-x86\linked\Foo.dll`
  * `android.21-arm\linked\System.Private.CoreLib.dll`
  * `android.21-x86\linked\System.Private.CoreLib.dll`

And "merges" this to what we need in the output of the `.apk`:

  * `assemblies/Foo.dll`
  * `assemblies/armeabi-v7a/System.Private.CoreLib.dll`
  * `assemblies/x86/System.Private.CoreLib.dll`

`Foo.dll` is identical for all architectures, so we only need to use
a single copy of it.

A bug in our implementation was for an assembly that differed between
32-bit and 64-bit platforms, but was identical within those sets:

  * `android.21-arm\linked\System.Collections.Concurrent.dll`
  * `android.21-x86\linked\System.Collections.Concurrent.dll`
  * `android.21-arm64\linked\System.Collections.Concurrent.dll`
  
The `-arm` and `-x86` copies of these assemblies have the same MVID.

The `<ProcessAssemblies/>` task would *skip* processing any assembly
with a previously "seen" MVID.  Consequently, the `.apk` contained:

  * `assemblies/armeabi-v7a/System.Collections.Concurrent.dll`
  * `assemblies/arm64-v8a/System.Collections.Concurrent.dll`

Notably absent was a copy in `assemblies/x86`.

The result was a runtime crash when running on the x86 emulator:

	Unhandled Exception:
	System.TypeLoadException: Could not set up parent class, due to: 
	  Could not load type of field 'Java.Interop.JniRuntime:TrackedInstances' (31) due to:
	    Could not load file or assembly 'System.Collections.Concurrent, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. assembly:Java.Interop.dll type:JniRuntime member:(null)

I reworked the logic to instead do:

  * Group by assembly file name
  * Count the unique MVIDs
  * If we have only 1 MVID, we just need the first assembly.
  * Otherwise, place the assemblies in architecture-specific directories.

The resulting code seems like it is a bit simpler, and likely a
little faster.  I was able to call
`MonoAndroidHelper.HasMonoAndroidReference()` once per file name,
which should help performance.

I updated `XASdkTests.DotNetBuild()` to check this scenario.
```

PR #5278, opened 2020-11-09
[linker] Add AddKeepAlivesStep

brendanzagaeski:
Context: https://github.com/xamarin/java.interop/issues/719

Modify "old" bindings, so that they hold the object reference
long enough to avoid above issue.

It is implemented as linker step, which looks for offended methods
and injects call to `System.GC.KeepAlive` to prevent premature
collection.

The injection can be disabled with `AndroidAddKeepAlives` property
set to false.

The property's default value depends on `AndroidIncludeDebugSymbols`
property and is thus true for release builds and false for debug.

The impact on release build of XA forms template (flyout variety) is
cca 137ms on my machine.

Disabled:

    12079 ms  LinkAssemblies                             1 calls

Enabled:

    12216 ms  LinkAssemblies                             1 calls

Example updated method body:
```
try {
	JniArgumentValue* ptr = stackalloc JniArgumentValue[1];
	*ptr = new JniArgumentValue((windowToken == null) ? IntPtr.Zero : ((Java.Lang.Object)windowToken).Handle);
	ptr[1] = new JniArgumentValue((int)flags);
	bool result = JniPeerMembers.InstanceMethods.InvokeAbstractBooleanMethod("...", this, ptr);
} finally {
	GC.KeepAlive(windowToken);
	return result;
}
```

Added new `AndroidAddKeepAlives ` test to `LinkerTests.cs` to check
whether registered method was updated.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
@@ -33,15 +33,18 @@ namespace MonoDroid.Tuner
 
 		internal bool AddKeepAlives (AssemblyDefinition assembly)
 		{
-			if (assembly.MainModule.HasTypeReference ("Java.Lang.Object"))
+			if (assembly.Name.Name != "Mono.Android" && !assembly.MainModule.HasTypeReference ("Java.Lang.Object"))
 				return false;
+
 			bool changed = false;
 			List<TypeDefinition> types = assembly.MainModule.Types.ToList ();
 			foreach (TypeDefinition type in assembly.MainModule.Types)
 				AddNestedTypes (types, type);
+
 			foreach (TypeDefinition type in types)
 				if (MightNeedFix (type))
 					changed |= AddKeepAlives (type);
+
 			return changed;
 		}
 
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
@@ -65,22 +65,46 @@ namespace MonoDroid.Tuner
 			return !type.IsAbstract && type.IsSubclassOf ("Java.Lang.Object", cache);
 		}
 
+		MethodDefinition methodKeepAlive = null;
+
 		bool AddKeepAlives (TypeDefinition type)
 		{
 			bool changed = false;
 			foreach (MethodDefinition method in type.Methods) {
 				if (!method.CustomAttributes.Any (a => a.AttributeType.FullName == "Android.Runtime.RegisterAttribute"))
 					continue;
-				ILProcessor processor = method.Body.GetILProcessor ();
-				ModuleDefinition module = method.DeclaringType.Module;
-				MethodDefinition methodKeepAlive = Context.GetMethod ("mscorlib", "System.GC", "KeepAlive", new string [] { "System.Object" });
-				Instruction end = method.Body.Instructions.Last ();
+
+				if (method.Parameters.Count == 0)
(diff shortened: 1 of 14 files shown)


jonpryor:
Related:

  * https://github.com/xamarin/java.interop/commit/1f21f38c4f1b5d833565156da570853ffc58ae92
  * https://github.com/xamarin/java.interop/issues/719

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
```diff
@@ -0,0 +1,103 @@
...
+			foreach (MethodDefinition method in type.Methods) {
+				if (!method.CustomAttributes.Any (a => a.AttributeType.FullName == "Android.Runtime.RegisterAttribute"))
+					continue;
+				ILProcessor processor = method.Body.GetILProcessor ();
+				ModuleDefinition module = method.DeclaringType.Module;
+				MethodDefinition methodKeepAlive = Context.GetMethod ("mscorlib", "System.GC", "KeepAlive", new string [] { "System.Object" });
+				Instruction end = method.Body.Instructions.Last ();
+				if (end.Previous.OpCode == OpCodes.Endfinally)
+					end = end.Previous;
+				for (int i = 0; i < method.Parameters.Count; i++) {
```
radekdoulik, line 80: We should check whether the call is not already present.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
```diff
@@ -0,0 +1,103 @@
...
+			if (AddKeepAlives (assembly)) {
+				AssemblyAction action = Annotations.HasAction (assembly) ? Annotations.GetAction (assembly) : AssemblyAction.Skip;
+				if (action == AssemblyAction.Skip || action == AssemblyAction.Copy || action == AssemblyAction.Delete)
+					Annotations.SetAction (assembly, AssemblyAction.Save);
+			}
+		}
+
+		internal bool AddKeepAlives (AssemblyDefinition assembly)
+		{
+			if (assembly.Name.Name != "Mono.Android" && !assembly.MainModule.HasTypeReference ("Java.Lang.Object"))
```
radekdoulik, line 36: @jonpryor, do we actually ever want to process `Mono.Android` assembly? I think when this step is run, the `Mono.Android` coming from the installation should be already "fixed".

src/Xamarin.Android.Build.Tasks/Xamarin.Android.CSharp.targets
```diff
@@ -29,6 +29,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
         <TargetFrameworkIdentifier>MonoAndroid</TargetFrameworkIdentifier>
         <TargetFrameworkVersion Condition="'$(TargetFrameworkVersion)' == ''">v5.0</TargetFrameworkVersion>
         <LangVersion Condition=" '$(LangVersion)' == '' ">8.0</LangVersion>
+        <AndroidAddKeepAlives Condition="'$(AndroidAddKeepAlives)' == ''">True</AndroidAddKeepAlives>
```
jonathanpeppers, line 32: Do we really want to do this for all `Debug` builds? What is the build time impact? I'm thinking of how many methods would get changed in support library or AndroidX assemblies.

Maybe we can set `$(AndroidAddKeepAlives)` to `true` when `$(AndroidIncludeDebugSymbols)` is `false`? Then it would be `Release` builds only by default.

radekdoulik: It is very rare to happen, so I think it should be fine to disable for Debug. That leads to producing different code for Debug and Release, hopefully the XF test we run on emulator should be enough to test that, for regressions, during our CI builds.

PR #5282, opened 2020-11-10
Build with net5.0 TFM instead of netcoreapp3.1

radekdoulik:
Build with net5.0 TFM instead of netcoreapp3.1

 That would allow use of new NET5 API like the new dynamic dependency
 [attribute](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.dynamicdependencyattribute?view=net-5.0).

 Also replace NET5_LINKER define with NET5_0.


radekdoulik:
This PR is draft as it needs newer msbuild 16.8, which will be available in the next VS 2019 release.

It also needs https://github.com/xamarin/java.interop/pull/747

pjcollins:
I think we can close this now that https://github.com/xamarin/xamarin-android/pull/5891 has been merged.

PR #5283, opened 2020-11-10
[Xamarin.Android.Build.Tasks] ignore .csproj.user file in $(MSBuildAllProjects)

jonathanpeppers:
I noticed that changing the selected device or emulator in Visual
Studio 2019 16.8 on Windows was causing many MSBuild targets to run
again in the IDE. The time taken was similar to a `Rebuild`!

Looking closer, I noticed:

    Building target "_CompileJava" completely.
    Input file "App42.csproj.user" is newer than output file "obj\Debug\_javac.stamp".

`.csproj.user` is where the IDE stores the selected device or
emulator. The contents of the file:

    <Project ToolsVersion="Current" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
      <PropertyGroup>
        <SelectedDevice>pixel_2_pie_9_0_-_api_28</SelectedDevice>
        <DefaultDevice>pixel_2_pie_9_0_-_api_28</DefaultDevice>
      </PropertyGroup>
    </Project>

The reason this triggers many targets is the file is actually imported:

https://github.com/dotnet/msbuild/blob/7452552ce911efdca6aea5a189c409f083db7bc7/src/Tasks/Microsoft.Common.CurrentVersion.targets#L31

And so this file is automatically included in `$(MSBuildAllProjects)`,
which is used by many `Inputs` of Xamarin.Android MSBuild targets.

I first tried editing `$(MSBuildAllProjects)`, but this property
appears to be readonly, since it is a built-in property in MSBuild.
Attempts at setting it were ignored.

To solve this, lets create a new `@(_AndroidMSBuildAllProjects)` item
group to be used instead that excludes the `.csproj.user` file.

I was able to reproduce this issue in a new test in
`IncrementalBuildTest`.

After this goes in, there is one place in the xamarin/monodroid repo
that needs to use `@(_AndroidMSBuildAllProjects)` instead of
`$(MSBuildAllProjects)`.

### DIFF
--- a/Documentation/guides/MSBuildBestPractices.md
+++ b/Documentation/guides/MSBuildBestPractices.md
@@ -308,6 +308,11 @@ will re-run the target completely. `$(MSBuildAllProjects)` is a list
 of every MSBuild file imported during a build, MSBuild automatically
 evaluates `$(MSBuildAllProjects)` since [MSBuild 16.0][allprojects].
 
+> NOTE: You might consider using `@(_AndroidMSBuildAllProjects)`
+> instead of `$(MSBuildAllProjects)` when working on the
+> Xamarin.Android MSBuild targets. We have excluded the `*.csproj.user`
+> file for performance reasons.
+
 One pitfall, is this `_GenerateDocumentation` example *must* touch the
 timestamps on all files in `Outputs` -- regardless if they were
 actually changed. Otherwise, your target can get into a state where it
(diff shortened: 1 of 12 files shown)


jonathanpeppers:
I manually tested a build from here, and everything seems to work switching from x86 emulator to x86_64 to arm64-v8a device.

I also edited the `.csproj` to make sure the targets that should run, still do.

dellis1972 (approved):

PR #5285, opened 2020-11-10
[GitHub] Update CODEOWNERS

brendanzagaeski:


brendanzagaeski:
Now that the team as a whole is familiar with how to include release notes and add or update localizable strings, my individual review on those items is no longer needed.

PR #5293, opened 2020-11-13
[ci] Install .NET 5 stable and stable VS

pjcollins:
All build and test jobs have been updated to install both .NET Core
3.1.201 and .NET 5.0.100.  Eventually, we should be able to remove the
3.1.201 installation after we migrate everything targeting .NET Core to
.NET 5.  Windows jobs will now update the Visual Studio 2019 instance
on disk to the latest available stable version. The base image has been
updated to speed up update performance.  See below for more info:

https://github.com/xamarin/monodroid/wiki/Azure-Pipelines-and-DevOps#windows-scale-set-agent-pool-management

There are currently multiple issues with .NET 5 and `xabuild`.  First
off, the system Mono that we are using contains a version of MSBuild
that is not compatible with .NET 5 targets, which attempt to use a
static property method that does not exist in Mono's MSBuild:

    /Users/builder/azdo/_work/_tool/dotnet/sdk/5.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.TargetFrameworkInference.targets(54,5): error MSB4186: Invalid static method invocation syntax: "[MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')". Method '[MSBuild]::GetTargetFrameworkIdentifier' not found. Static method invocation should be of the form: $([FullTypeName]::Method()), e.g. $([System.IO.Path]::Combine(`a`, `b`)). Check that all parameters are defined, are of the correct type, and are specified in the right order.

Once this issue is resolved (or in the case of our Windows builds), we
hit the second issue:

    C:\Program Files\dotnet\sdk\5.0.100\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.TargetFrameworkInference.targets(54,5): error MSB4184: The expression "[MSBuild]::GetTargetFrameworkIdentifier(netstandard2.0)" cannot be evaluated. MSB0001: Internal MSBuild Error: A required NuGet assembly was not found. Expected Path: C:\Users\Peter\AppData\Local\Temp\ktbpeg2z.b31

If we decide that xabuild should support .NET 5+ in the future, we'll
need to copy a version of `NuGet.Frameworks.dll` next to
`MSBuildExeTempPath` to work around this issue.

### DIFF
--- a/build-tools/automation/azure-pipelines-oss.yaml
+++ b/build-tools/automation/azure-pipelines-oss.yaml
@@ -27,6 +27,7 @@ variables:
   EXTRA_MSBUILD_ARGS: /p:AutoProvision=True /p:AutoProvisionUsesSudo=True /p:IgnoreMaxMonoVersion=False
   PREPARE_FLAGS: PREPARE_CI=1 PREPARE_CI_PR=1
   DotNetCoreVersion: 3.1.201
+  DotNet5Version: 5.0.100
   GitHub.Token: $(github--pat--vs-mobiletools-engineering-service2)
 
 stages:
@@ -51,8 +52,12 @@ stages:
 
     - template: yaml-templates/use-dot-net.yaml
       parameters:
-        version: $(DotNetCoreVersion)
+        version: $(DotNet5Version)
         remove_dotnet: true
+
+    - template: yaml-templates/use-dot-net.yaml
+      parameters:
+        version: $(DotNetCoreVersion)
 
     - bash: |
         keychains=`security list-keychains`
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -52,10 +52,11 @@ variables:
   TestAssembliesArtifactName: test-assemblies
   NUnitConsoleVersion: 3.11.1
   DotNetCoreVersion: 3.1.201
+  DotNet5Version: 5.0.100
   HostedMacMojave: Hosted Mac Internal Mojave
   HostedMac: Hosted Mac Internal
   HostedWinVS2019: Hosted Windows 2019 with VS2019
-  VSEngWinVS2019: Xamarin-Android-Win2019
+  VSEngWinVS2019: Xamarin-Android-Win2019-1120
   # Run all tests if:
   # - User who queued the job requested it (They set XA.RunAllTests to true)
   # - This is the master integration branch (Pipeline defaults XA.RunAllTests to true)
@@ -137,8 +138,12 @@ stages:
 
     - template: yaml-templates/use-dot-net.yaml
       parameters:
-        version: $(DotNetCoreVersion)
+        version: $(DotNet5Version)
         remove_dotnet: true
+
+    - template: yaml-templates/use-dot-net.yaml
+      parameters:
+        version: $(DotNetCoreVersion)
 
     - template: install-certificates.yml@yaml
       parameters:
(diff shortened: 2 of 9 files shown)


radekdoulik (approved):

jonathanpeppers:
I think I'm seeing this with 16.8:
```
> xabuild .\Xamarin.Android-Tests.sln
...
error MSB4184: The expression "[MSBuild]::GetTargetFrameworkIdentifier
(netstandard2.0)" cannot be evaluated. MSB0001: Internal MSBuild Error: A required NuGet assembly was not found.
```
I think might be similar to what is happening here?

tools/xabuild/XABuild.cs
```diff
@@ -146,6 +146,13 @@ static void CreateConfig (XABuildPaths paths)
 			}
 
 			Directory.CreateDirectory (paths.MSBuildTempPath);
+			// Copy NuGet.Frameworks.dll from output directory to MSBuildTempPath to appease:
+			// https://github.com/dotnet/msbuild/blob/755d4d1e3d2a89f72f659fc3d7d2933cab619828/src/Build/Utilities/NuGetFrameworkWrapper.cs#L32
+			var nuGetFrameworksPath = Path.Combine (AppContext.BaseDirectory, "NuGet.Frameworks.dll");
+			if (!File.Exists (nuGetFrameworksPath))
+				throw new Exception ($"Could not find NuGet assembly {nuGetFrameworksPath} required by MSBuild.");
+
+			File.Copy (nuGetFrameworksPath, Path.Combine (paths.MSBuildTempPath, Path.GetFileName (nuGetFrameworksPath)));
```
jonathanpeppers, line 155: Yeah, I don't see another way to workaround this...

Instead of `throw` can this `Console.WriteLine()` the error and make the exit code 1?

jonathanpeppers (approved):
I like the latest workaround, to ignore .NET 5 and higher. 👍

pjcollins:
Full test run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4238000&view=logs&j=cac0e8d3-0ef5-5d2b-b57e-e8fde7204df3

pjcollins:
New test run for commit 06c6bde now that other test issues have been resolved: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4241493&view=results

pjcollins:
Full test run for commit 38c0b88: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4242239&view=results

I've ran a few full test runs against the new VM image and the windows jobs seem to be working reliably. The changes in the last commit are minimal and were made to avoid the VS update during the designer job, as the hosted pool it runs on should always be up to date. In addition to updating the base image used by the windows pool, I've updated the VM "size" to use the same SKU as the Microsoft hosted Windows agents - [Standard_DS4_v2](https://docs.microsoft.com/en-us/azure/virtual-machines/dv2-dsv2-series#dsv2-series). The hope is that this change will give us better and more consistent disk read/write speeds than the previously used burstable series, but the overall job timings seem pretty comparable so far. I haven't seen any job run for longer than 90 minutes on these new VMs yet however, which seems pretty promising.

dellis1972 (approved):

jonathanpeppers (approved):

pjcollins:
Full (hopefully last) test run for e3d6e85: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4243938&view=results

PR #5297, opened 2020-11-16
[ci] Restore custom use-dot-net task

pjcollins:
Context: https://github.com/xamarin/xamarin-android/pull/5293
Context: https://github.com/xamarin/xamarin-android/blob/f5680e783fe45d4f536f8199372ff504ebfdc393/build-tools/automation/yaml-templates/use-dot-net.yaml

PR #5293 introduced an incompatibility with xabuild, and it appears to
be causing issues with other stable builds.  The [UseDotNet][1] task has
a caching feature to improve "installation" times for certain versions,
however this can cause issues for our non-hosted build machines.

       /Users/builder/azdo/_work/_tool/dotnet/sdk/5.0.100/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.TargetFrameworkInference.targets(54,5): error MSB4186: Invalid static method invocation syntax: "[MSBuild]::GetTargetFrameworkIdentifier('$(TargetFramework)')". Method '[MSBuild]::GetTargetFrameworkIdentifier' not found. Static method invocation should be of the form: $([FullTypeName]::Method()), e.g. $([System.IO.Path]::Combine(`a`, `b`)). Check that all parameters are defined, are of the correct type, and are specified in the right order. [/Users/builder/azdo/_work/1/s/xamarin-android/external/Java.Interop/src/Java.Interop.GenericMarshaler/Java.Interop.GenericMarshaler.csproj]

We should continue to use the custom `use-dot-net` task that Jonathan
Peppers added for us during earlier .NET 5 preview work.  This template
was removed in PR #5191 but in hindsight it should not have been.

[1]: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/tool/dotnet-core-tool-installer?view=azure-devops

### DIFF
--- a/build-tools/automation/yaml-templates/use-dot-net.yaml
+++ b/build-tools/automation/yaml-templates/use-dot-net.yaml
@@ -1,0 +1,34 @@
+# This template enables installation of stable and master/nighly builds of .NET.
+# We prefer this over the UseDotNet task so that we can always clean up old/unstable versions on disk.
+
+parameters:
+  version: $(DotNetCoreVersion)
+  remove_dotnet: false
+
+steps:
+
+  - powershell: |
+      $ErrorActionPreference = 'Stop'
+      $ProgressPreference = 'SilentlyContinue'
+      $DotNetRoot = "$env:ProgramFiles\dotnet\"
+      if ("${{ parameters.remove_dotnet }}" -eq $true) {
+        Remove-Item -Recurse $DotNetRoot -Verbose
+      }
+      Invoke-WebRequest -Uri "https://dot.net/v1/dotnet-install.ps1" -OutFile dotnet-install.ps1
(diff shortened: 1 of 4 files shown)


build-tools/automation/yaml-templates/use-dot-net.yaml
```diff
@@ -0,0 +1,33 @@
+# This template enables installing master/nighly builds of .NET 5.
```
jonathanpeppers, line 1: You might update this to say .NET 6 if you change it again.

dellis1972 (approved):

build-tools/automation/yaml-templates/use-dot-net.yaml
pjcollins: I updated this to be version agnostic and to help explain why we'd want to use it over `UseDotNet`.

jonathanpeppers (approved):

PR #5302, opened 2020-11-17
[xaprepare] Fix for newer homebrew on macOs

grendello:
The latest version of homebrew introduced a breaking change in the
way it handles certain combinations of its command-line arguments.
Namely, what affects us, is the combination `--versions -1` to
limit the number of package's versions to a single line. Right
now it is not allowed and causes the following failure:

    Error: Invalid usage: Options --versions and -1 are mutually exclusive.
     [MISSING]
    Checking automake                                   Usage: brew list, ls [options] [formula|cask]

Instead of passing `-1`, do the processing ourselves.

### DIFF
--- a/build-tools/xaprepare/xaprepare/Application/HomebrewProgram.MacOS.cs
+++ b/build-tools/xaprepare/xaprepare/Application/HomebrewProgram.MacOS.cs
@@ -6,6 +6,8 @@ namespace Xamarin.Android.Prepare
 {
 	class HomebrewProgram : Program
 	{
+		static readonly char[] lineSplit = new [] { '\n' };
+
 		string cachedVersionOutput;
 		bool brewNeedsSudo = false;
 		bool multipleVersionsPickle = false;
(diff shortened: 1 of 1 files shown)


jonpryor:
I'm somewhat confused; the "processing ourselves" aspect is that we just take the first line of output.

On an older Homebrew (2.5.6), I only get one line of output *anyway*:

```bash
% brew ls --versions wget    
wget 1.20.3_2
```

I'm thus not sure why we needed `brew ls --versions -1 {Name}` in the first place?  I *suppose* that if a package had multiple versions installed, that would change things, but I don't have multiple versions of a single package installed, so I can't see if that changes anything locally.  If that *were* the case, then a "do the processing ourselves" approach which just takes the first line of output doesn't seem sufficient either.

Why can't we just drop `"-1"`?  Why is the "extra processing" necessary?

grendello:
> I'm somewhat confused; the "processing ourselves" aspect is that we just take the first line of output.
> 
> On an older Homebrew (2.5.6), I only get one line of output _anyway_:
> 
> ```shell
> % brew ls --versions wget    
> wget 1.20.3_2
> ```
> 
> I'm thus not sure why we needed `brew ls --versions -1 {Name}` in the first place? I _suppose_ that if a package had multiple versions installed, that would change things, but I don't have multiple versions of a single package installed, so I can't see if that changes anything locally. If that _were_ the case, then a "do the processing ourselves" approach which just takes the first line of output doesn't seem sufficient either.
> 
> Why can't we just drop `"-1"`? Why is the "extra processing" necessary?

It's a precaution. Brew can have more than one version of a given package installed, the `--versions` command will report them sorted in descending order. The old `-1` switch made it return just a single line, which is what we need to properly parse the version. With that switch being incompatible now, we need to make sure we process only the first line or version parsing will fail most likely should there be more than one version reported.

PR #5304, opened 2020-11-17
[Xamarin.Android.Build.Tasks] Add support for Proguard `mapping.txt` file

### LINKED ISSUE #5186: [Bug] Warning in Google Console when publishing an app from Visual Studio
### Description
When I published my app to the new (beta) Google console, i got two warnings, one of them was "This App Bundle contains Java/Kotlin code, which might be obfuscated." 

I found this issue that describes my situation: https://github.com/apache/cordova-android/issues/1008

From what I read it seems that the publishing procedure inside visual studio needs to be updated to upload some additional file so this warning goes away.

dellis1972:
Fixes #5186

When uploading a package to the Google Play Store users are seeing
the following warning.

    This App Bundle contains Java/Kotlin code, which might be obfuscated.

To fix this users need to provide a `mapping.txt` file which contains
the Java class mappings from their plain versions to the ones which
are obfuscated. By default we do not obfuscate the java code but the
warning still shows.

To fix this issue we have a new property `$(AndroidProguardMappingFile)`
which defaults to `$(OutputPath)mapping.txt`. This file will be produced
as part of the build process. In order for this mapping file to be
generated the following lines needed to be added to the `proguard_xamarin.cfg`
file.

    -keepattributes SourceFile
    -keepattributes LineNumberTable

### DIFF
--- a/Documentation/guides/building-apps/build-properties.md
+++ b/Documentation/guides/building-apps/build-properties.md
@@ -806,6 +806,17 @@ with build environments that have FIPS compliance enforced.
 
 Added in Xamarin.Android 10.1.
 
+## AndroidProguardMappingFile
+
+Specifies the `-printmapping` proguard rule for `r8`. This will
+mean the `mapping.txt` file will be produced in the `$(OutputPath)`
+folder. This file can then be used when uploading packages to the
+Google Play Store.
+
+Default value is `$(OutputPath)mapping.txt`
+
+Added in Xamarin.Android 11.2
+
 ## AndroidR8IgnoreWarnings
 
 Specifies
--- a/Documentation/release-notes/5304.md
+++ b/Documentation/release-notes/5304.md
@@ -1,0 +1,9 @@
+### Build and deployment performance
+
+  * [GitHub PR 5304](https://github.com/xamarin/xamarin-android/pull/5304):
+    Add support for producing a proguard `mapping.txt` file to the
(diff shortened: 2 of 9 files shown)


jonathanpeppers (commented):
Should we update one test to check `mapping.txt` exists?

jonathanpeppers (approved):
I restarted CI, looks good to me otherwise.

EmilAlipiev:
thats is awesome. is it already in any release we can test. By the way, it is said that Xamarin.Android r8 doesnt do obfuscation, is it correct?

dellis1972:
@EmilAlipiev it should be in d16-9 Preview 2 (I think). Generally we turn off r8 obfuscation because it stops our C#->Java intro layer working see https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Resources/proguard_xamarin.cfg#L3

EmilAlipiev:
This seems to be working in the latest releases. when i locally build, i can see that mapping.txt file is generated but in the AzureDevops i am not able to see the file under the artifacts. Do i have to something special in my Yaml file?

dellis1972:
@EmilAlipiev It takes a while for the latest release to get on to DevOps (assuming you are using the system installed XA on DevOps). As they have to update the image, tests it etc. 

Perhaps it might be better to use a tool like [boots](https://github.com/jonathanpeppers/boots) to install the specific version you want?

PR #5310, opened 2020-11-18
[Xamarin.Android.Build.Tasks] Use Proguard config files from .aar files.

### LINKED ISSUE #3752: Proguard configs should be included from .aar's
.AAR files can include `proguard.txt` configuration files.

These are not currently being seen and included in the proguard/r8 usage during the build.

Probably `ResolveLibraryProjectImports` needs to be responsible for looking for these after extracting the .aar's and adding them to the `@(ProguardConfiguration)` group.

We currently ship proguard.txt files inside some nuget packages with a .targets file which includes them in the `ProguardConfiguration` item group, so it would be important to test what happens if duplicate config files are passed to proguard/r8 if this work is done (since there will be nuget packages with them we cannot change).

dellis1972:
Fixes #3752

Add support for using `proguard.txt` file from `aar` files. 
Also add support for including `proguard.txt` files in our generated `aar` files when we build a Library project
in .net 6. This will merge all `ProguardConfiguration` Items into one `proguard.txt` file and include that in the 
root of the `aar`. 

Update the Unit Tests to include test for this functionality.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
@@ -45,6 +45,8 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <AndroidNativeLibrary Include="**\*.so"  Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
     <JavaSourceJar        Include="**\*-source.jar;**\*-sources.jar;**\*-src.jar"   Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
     <AndroidJavaSource    Include="**\*.java" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <ProguardConfiguration Include="**\proguard.cfg" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <ProguardConfiguration Include="**\proguard.txt" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
   </ItemGroup>
 
 </Project>
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
@@ -49,6 +49,7 @@ projects.
       <_CreateAarInputs Include="@(AndroidJavaLibrary)" />
       <_CreateAarInputs Include="@(EmbeddedJar)" />
       <_CreateAarInputs Include="@(EmbeddedNativeLibrary)" />
+      <_CreateAarInputs Include="@(ProguardConfiguration)" />
     </ItemGroup>
     <Hash
         ItemsToHash="@(_CreateAarInputs)"
@@ -80,6 +81,7 @@ projects.
         AndroidEnvironment="@(AndroidEnvironment)"
         JarFiles="@(AndroidJavaLibrary);@(EmbeddedJar)"
         NativeLibraries="@(EmbeddedNativeLibrary)"
+        ProguardConfigurationFiles="@(ProguardConfiguration)"
         OutputFile="$(_AarOutputPath)"
     />
     <ItemGroup>
--- a/src/Xamarin.Android.Build.Tasks/Tasks/CreateAar.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/CreateAar.cs
@@ -23,6 +23,8 @@ namespace Xamarin.Android.Tasks
 
 		public ITaskItem [] NativeLibraries { get; set; }
 
+		public ITaskItem [] ProguardConfigurationFiles { get; set; }
+
 		[Required]
 		public string AssetDirectory { get; set; }
 
@@ -102,6 +104,13 @@ namespace Xamarin.Android.Tasks
 						aar.AddStream (File.OpenRead (lib.ItemSpec), archivePath);
 						existingEntries.Remove (archivePath);
 					}
+				}
+				if (ProguardConfigurationFiles != null) {
+					var sb = new StringBuilder ();
+					foreach (var file in ProguardConfigurationFiles) {
+						sb.AppendLine (File.ReadAllText (file.ItemSpec));
+					}
+					aar.AddEntry ("proguard.txt", sb.ToString (), Files.UTF8withoutBOM);
 				}
 				foreach (var entry in existingEntries) {
 					Log.LogDebugMessage ($"Removing {entry} as it is not longer required.");
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ResolveLibraryProjectImports.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ResolveLibraryProjectImports.cs
@@ -59,6 +59,9 @@ namespace Xamarin.Android.Tasks
 		[Output]
 		public ITaskItem [] ResolvedResourceDirectoryStamps { get; set; }
 
+		[Output]
+		public ITaskItem [] ProguardConfigFiles { get; set; }
+
 		internal const string OriginalFile = "OriginalFile";
 		internal const string AndroidSkipResourceProcessing = "AndroidSkipResourceProcessing";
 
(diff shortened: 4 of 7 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
```diff
@@ -39,6 +39,7 @@ projects.
       <AndroidJavaLibrary    Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.jar' and '%(AndroidLibrary.Bind)' != 'true' " />
       <EmbeddedJar           Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.jar' and '%(AndroidLibrary.Bind)' == 'true' " />
       <EmbeddedNativeLibrary Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.so' " />
+      <ProguardConfiguration Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.cfg' " />
```
dellis1972, line 42: @jonathanpeppers I'm not sure what the default file filter should be under net 6. `*.cfg` or just `proguard.txt`.

jonathanpeppers: The defaults here are what split the `@(AndroidLibrary)` item group into the 6 other item groups.

I think you're wanting to make `*.cfg` automatically added to `@(ProguardConfiguration)`? I think we can do `**\*.cfg` and `**\proguard.txt` as wildcards in `AutoImports.props`?

Just keep in mind every .NET project type could import this file, so you must check `$(EnableDefaultAndroidItems)` for new `<ItemGroup/>` elements:

https://github.com/xamarin/xamarin-android/blob/6e3e3831afa958ab8126e268f7a128321f75dbd6/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props#L19-L36

src/Xamarin.Android.Build.Tasks/Tasks/CreateAar.cs
```diff
@@ -102,6 +104,15 @@ public override bool RunTask ()
 						existingEntries.Remove (archivePath);
 					}
 				}
+				if (ProguardConfigurationFiles != null) {
+					var sb = new StringBuilder ();
+					foreach (var file in ProguardConfigurationFiles) {
+						string[] lines = File.ReadAllLines (file.ItemSpec);
```
jonathanpeppers, line 110: Could this use `File.OpenText()` and `ReadLine()` instead of creating a `string[]`? Then it could add directly to the `StringBuilder`.

jonathanpeppers (commented):
The change here look good.

Do we need to be worried about proguard files getting included twice? I think some of the AndroidX/Support libraries extracted the proguard files and added them from .targets.

It seems like that might be OK, does it hurt anything?

dellis1972:
> The changes here look good.
> 
> Do we need to be worried about proguard files getting included twice? I think some of the AndroidX/Support libraries extracted the proguard files and added them from .targets.
> 
> It seems like that might be OK, does it hurt anything?

If we end up with duplicate commands from multiple pro guard files that doesn't seem to cause any problems on the app side.

dellis1972:
@jonathanpeppers only the designer test is failing here. I think this one is good to go. 

The only annoying bit I can see is if we reference C from B, B gets the proguard.txt from C in its .aar. As a result when the App A references B, it will also get C and will then have two of the same file during the R8/D8 compilation. 

Should that be something I should take out do you think? It doesn't seem to cause any problems  at the moment.

dellis1972:
I'm gonna rework this a bit so we don't include the additional proguard file if we are building a library project. The idea being that it will be included by the Reference by default in the final app.

jonathanpeppers (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
```diff
@@ -45,6 +45,8 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <AndroidNativeLibrary Include="**\*.so"  Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
     <JavaSourceJar        Include="**\*-source.jar;**\*-sources.jar;**\*-src.jar"   Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
     <AndroidJavaSource    Include="**\*.java" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <ProguardConfiguration Include="**\*.cfg" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
```
jonpryor, line 48: I wonder/fear if this is too "loose"; are these files frequently auto-generated and there are no reliable basename patterns?  What happens if "some other" `.cfg` gets globbed?  [fileformat.info](https://fileinfo.com/extension/cfg) lists ~8 different `.cfg` file types, and ProGuard *isn't* listed.

dellis1972: Pity we don't recommend a particular file in the docs https://docs.microsoft.com/en-us/xamarin/android/deploy-test/release-prep/proguard?tabs=windows#customizing-proguard. We just recommend ` a .cfg file`.

src/Xamarin.Android.Build.Tasks/Tasks/ResolveLibraryProjectImports.cs
```diff
@@ -431,6 +442,11 @@ public override bool RunTask ()
 					resolvedAssetDirectories.Add (new TaskItem (Path.GetFullPath (assetsDir), new Dictionary<string, string> {
 						{ OriginalFile, aarFullPath },
 					}));
+				if (File.Exists (proguardFile) && IsApplication) {
```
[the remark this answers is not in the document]
dellis1972: The main reason I did this was so that when we build a library project we don't even call proguard/r8 so its not really needed. But we do use it for Apps, so restricting it to apps only seemed logical.

jonpryor:
In-progress commit message:

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/3752

`.aar` and `.jar` files [can contain `proguard.txt` files][0]

> **Library dependencies**
> *Location*: AAR libraries: `<library-dir>/proguard.txt`
> *Location*: JAR libraries: `<library-dir>/META-INF/proguard/`
> *Description*: If an AAR library is published with its own ProGuard
> rules file, and you include that AAR as a compile-time dependency,
> R8 automatically applies its rules when compiling your project.

Update the `@(ProguardConfiguration)` item group to include
`proguard.cfg` and `proguard.txt` files in the project directory, and
for Library project builds add the contents of all
`@(ProguardConfiguration)` files into a `proguard.txt` entry within
the generated `.aar` file (fcd7cf8f).

[0]: https://developer.android.com/studio/build/shrink-code
```

jonpryor:
The Android docs state that `.jar` files *also* have such a mechanism: https://developer.android.com/studio/build/shrink-code

> **Library dependencies**
> *Location*: AAR libraries: `<library-dir>/proguard.txt`
> *Location*: JAR libraries: `<library-dir>/META-INF/proguard/`
> *Description*: If an AAR library is published with its own ProGuard
> rules file, and you include that AAR as a compile-time dependency,
> R8 automatically applies its rules when compiling your project.

Do we handle `mylib.jar!META-INF/proguard/proguard.txt` files?  (It doesn't look like it?)

*Should* we support `mylib.jar!META-INF/proguard/proguard.txt` files? (Maybe?)

Is there an easy way to find such a thing "in the wild"?

dellis1972:
> The Android docs state that `.jar` files _also_ have such a mechanism: https://developer.android.com/studio/build/shrink-code
> 
> > **Library dependencies**
> > _Location_: AAR libraries: `<library-dir>/proguard.txt`
> > _Location_: JAR libraries: `<library-dir>/META-INF/proguard/`
> > _Description_: If an AAR library is published with its own ProGuard
> > rules file, and you include that AAR as a compile-time dependency,
> > R8 automatically applies its rules when compiling your project.
> 
> Do we handle `mylib.jar!META-INF/proguard/proguard.txt` files? (It doesn't look like it?)
> 
> _Should_ we support `mylib.jar!META-INF/proguard/proguard.txt` files? (Maybe?)
> 
> Is there an easy way to find such a thing "in the wild"?

Well one issue there is we do not currently extract the contents of the `.jar` files into the `lp` directory. We do for `.aar` becuae we need to reference the `.jar` files within. So the `proguard.txt` will get extracted then, we do not do this for `.jar`. 

So if we want to support that we would need to add another step to look for and extract the proguard files from that `META-INF` folder in the .jar itself.

PR #5311, opened 2020-11-18
[Xamarin.Andrid.Build.Tests] Add MultiUser unit tests

dellis1972:
This commit adds support for running multi user tests on device.
It makes use of various adb commands to create test users and
remove them.

`shell pm create-user --guest {name}` will create a guest user of a
specific name.

`shell pm remove-user --guest {userId}` will remove a userId.

The userId can be retrieved via `shell pm list users`. This command
will return results as follows

    Users:
	    UserInfo{0:Owner:c13} running
	    UserInfo{10:guest1:414}

We can then use a RegEx to pull out the required `userId` for a specific
user.

Finally `shell am switch-user {userId}` can be used to switch between
users on the device.

Emulators seem to create the main user with the name "Owner" so that should
be the default name we use for the main user. In the unit tests if we
cannot find a user named "Owner" when looking for "Owner" we will default
to a `userId` of `0`. This always means the default user. This is to handle
the case where we are testing locally on devices where the main user is
not called "Owner".

### DIFF
--- a/.external
+++ b/.external
@@ -1,2 +1,2 @@
-xamarin/monodroid:master@cfa21d57b22a9c5bc51d7720a0f830e166381000
+xamarin/monodroid:master@f38dbc7abe462887c4d4b6e99affd0f2eb31ab62
 mono/mono:2020-02@ac596375c762c6b8dbe3c802f0ce626004eab51c
--- a/.external
+++ b/.external
@@ -1,2 +1,2 @@
-xamarin/monodroid:master@f38dbc7abe462887c4d4b6e99affd0f2eb31ab62
+xamarin/monodroid:master@27736a7ffc48d606ab45598f761e873f8572f46a
 mono/mono:2020-02@ac596375c762c6b8dbe3c802f0ce626004eab51c
--- a/.external
+++ b/.external
@@ -1,2 +1,2 @@
-xamarin/monodroid:master@0eef889156e92ff0d47e62477f10798b6d5b5c29
+xamarin/monodroid:master@5bb29971bebc1b8f7ef9aacd8e68aebac43302d2
 mono/mono:2020-02@ac596375c762c6b8dbe3c802f0ce626004eab51c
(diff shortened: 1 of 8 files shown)


tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
```diff
@@ -251,36 +251,60 @@ public override void OnCreate ()
...
+		public void ApplicationRunsWithDebuggerAndBreaks (bool embedAssemblies, string fastDevType, bool allowDeltaInstall, string username)
 		{
 			AssertCommercialBuild ();
 			AssertHasDevices ();
+
+			int userId = GetUserId (username);
+			if (SwitchUser (username)) {
+				WaitFor (5);
+				ClickButton ("", "android:id/button1", "Yes continue");
+			}
```
jonathanpeppers, line 306: Maybe instead of `Owner` we should just allow `username` to be blank. So we know these tests will work if you don't set `$(AndroidDeviceUserId)` at all?

.external
```diff
@@ -1,2 +1,2 @@
-xamarin/monodroid:master@0eef889156e92ff0d47e62477f10798b6d5b5c29
+xamarin/monodroid:master@5bb29971bebc1b8f7ef9aacd8e68aebac43302d2
```
jonathanpeppers, line 1: Maybe we should document `$(AndroidDeviceUserId)` along with this bump.

dellis1972: yup, will be doing that tomorrow.

tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
dellis1972: good idea

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
```diff
@@ -13,11 +13,14 @@ namespace Xamarin.Android.Build.Tests
 {
 	public class DeviceTest: BaseTest
 	{
+		public static string GuestUserName = "guest1";
```
jonpryor, line 16: Shouldn't this be `const` or `readonly`?  I can't imagine any reason to need to be able to *change* this value.

jonathanpeppers (approved):
Changes look good, if we are OK with the CI results.

PR #5313, opened 2020-11-18
[ci] Use unique storage URLs for installers

pjcollins:
Commit b864f93d appears to have broken the GitHub status for the .NET 6
installers.  These changes resulted in us attempting to use the same
blob storage location for both legacy and .NET 6 installers, and the
step which ran last (often legacy, as signing takes a while) would write
over the previous status.  Adding `net6` to the .NET 6 installer blob
storage path should fix this.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1182,7 +1182,7 @@ stages:
       parameters:
         ArtifactsDirectory: $(System.DefaultWorkingDirectory)\installer-artifacts
         Azure.ContainerName: $(Azure.Container.Name)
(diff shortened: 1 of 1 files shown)


pjcollins:
When complete this build should create and post aritfacts statuses for both installer types, and should be used to confirm the fix: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4236920&view=results

jonathanpeppers (approved):
The Github statuses I see currently seem like they worked.

The `Queue Vsix Signing` step is ongoing, so I'm not sure if that will affect anything.

pjcollins:
All three statuses have now posted with live links, I'll merge this.

PR #5316, opened 2020-11-18
[One .NET] move Properties\AndroidManifest.xml to project root

jonathanpeppers:
In "legacy" Xamarin.Android, projects default to:

    Properties\AndroidManifest.xml
    Properties\AssemblyInfo.cs
    Foo.csproj

Since `AssemblyInfo.cs` is generally not used in short-form MSBuild
projects, `AndroidManifest.xml` is the only file in the `Properties`
folder in .NET 6.

Let's move the default location for `AndroidManifest.xml` in .NET 6 to
clean things up. Xamarin.iOS also puts the `Info.plist` in the top
directory, so this aligns both platforms better.

To help with migration the old path will continue to work by default
if the file is found:

    <AndroidManifest Condition=" '$(AndroidManifest)' == '' and Exists ('Properties\AndroidManifest.xml') and !Exists ('AndroidManifest.xml') ">Properties\AndroidManifest.xml</AndroidManifest>
    <AndroidManifest Condition=" '$(AndroidManifest)' == '' ">AndroidManifest.xml</AndroidManifest>

We will default to simply `AndroidManifest.xml` otherwise and make
sure we pick `AndroidManifest.xml` if both are found.

I updated `XASdkProject` to use the new path. MSBuild tests that use
`XamarinAndroidApplicationProject` under a `dotnet` context will
continue to use `Properties\AndroidManifest.xml`, so we have test
coverage of both.


dellis1972 (approved):
Looks good. Just wondering if we should add a warning for those using the `Properties\AndroidManifest.xml` ?

pjcollins (approved):

jonathanpeppers:
> wondering if we should add a warning for those using the Properties\AndroidManifest.xml ?

@dellis1972 I think we should just support both, since it is just a one-line check. If we want to drop it someday in the future, we could add a warning later?

radekdoulik (approved):
LGTM

Maybe one small observation for discussion, if someone gets confused and adds manifest in both locations, shouldn't we warn about it or at least prefer the new location?

jonathanpeppers:
> Maybe one small observation for discussion, if someone gets confused and adds manifest in both locations, shouldn't we warn about it or at least prefer the new location?

I should change this check to:
```diff
--<AndroidManifest Condition=" '$(AndroidManifest)' == '' and Exists ('Properties\AndroidManifest.xml') ">Properties\AndroidManifest.xml</AndroidManifest>
++<AndroidManifest Condition=" '$(AndroidManifest)' == '' and Exists ('Properties\AndroidManifest.xml') and !Exists('AndroidManifest.xml') ">Properties\AndroidManifest.xml</AndroidManifest>
```

PR #5317, opened 2020-11-19
[Xamarin.Android.Build.Tasks] Link Resource Nested Types from assemblies

dellis1972:
The idea is to remove the `Resource` class NestedTypes from the final
linked assemblies in release mode. We do this by examining all the
locations where the fields are used and replacing those IL calls with a
call which specifies the actual resource id directly. This will remove
the need to call `global::Android.Runtime.ResourceIdManager.UpdateIdValues();`
completely. It will also allow us to completely remove the `Resource`
class sub classes. Due to an odd linker failure we cannot completely
remove the `Resource` class itself. However it will be completely empty.

As mentioned the call to `UpdateIdValues` will no  longer be required.
This should reduce startup time slightly. One thing to note is this will
only be true for Release apps when enabled. Debug apps will still
call `UpdateIdValues`. This is due to the time is takes to run
the Linker Step. Applying this to a Debug build would effect the  build
times during development.

A new MSBuild property `AndroidLinkResources` has been added to control
this feature. It is turned off by default for now.

This seems to save between 100-280kb for an apk. This is dependent on
how many assemblies are Android assemblies and contain or use Resources.

Xamarin.Forms base app apkdiff results

```
  -          10 assemblies/Mono.Android.dll
  -       4,001 assemblies/Xamarin.Essentials.dll
  -      45,426 assemblies/Xamarin.Forms.Platform.Android.dll
  -     102,839 assemblies/formstest.Android.dll
Summary:
  +           0 Other entries 0.00% (of 887,808)
  +           0 Dalvik executables 0.00% (of 3,341,152)
  -     152,276 Assemblies -3.23% (of 4,709,077)
  +           0 Shared libraries 0.00% (of 41,693,796)
  -     151,552 Package size difference -0.73% (of 20,820,695)
```

Maps sample apkdiff results

```
  -           4 lib/armeabi-v7a/libxamarin-app.so
  -          14 assemblies/Mono.Android.dll
  -       4,757 assemblies/Xamarin.Essentials.dll
  -      46,051 assemblies/Xamarin.Forms.Platform.dll
  -      47,586 assemblies/Xamarin.Forms.Platform.Android.dll
  -      48,470 assemblies/Xamarin.Forms.Maps.Android.dll
  -     135,589 assemblies/FormsMapsSample.Android.dll
Summary:
  +           0 Other entries 0.00% (of 1,137,209)
  +           0 Dalvik executables 0.00% (of 2,519,972)
  -     282,467 Assemblies -5.51% (of 5,123,833)
  -           4 Shared libraries -0.00% (of 41,843,596)
  -     282,624 Package size difference -1.34% (of 21,075,664)
```

### DIFF
--- a/src/Microsoft.Android.Sdk.ILLink/SetupStep.cs
+++ b/src/Microsoft.Android.Sdk.ILLink/SetupStep.cs
@@ -65,6 +65,11 @@ namespace Microsoft.Android.Sdk.ILLink
 			if (Context.TryGetCustomData ("AddKeepAlivesStep", out addKeepAlivesStep) && bool.TryParse (addKeepAlivesStep, out var bv) && bv)
 				InsertAfter (new AddKeepAlivesStep (cache), "CleanStep");
 
+			string androidLinkResources;
+			if (Context.TryGetCustomData ("AndroidLinkResources", out androidLinkResources) && bool.TryParse (androidLinkResources, out var linkResources) && linkResources) {
+				InsertAfter (new RemoveResourceDesignerStep (),  "CleanStep");
+				InsertAfter (new GetAssembliesStep (), "CleanStep");
+			}
 			InsertAfter (new StripEmbeddedLibraries (),  "CleanStep");
 		}
 
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
@@ -1,0 +1,210 @@
+using Mono.Cecil;
+using Mono.Linker;
+using Mono.Linker.Steps;
+using System;
+using System.Linq;
+using Xamarin.Android.Tasks;
+using System.Collections.Generic;
+using Mono.Cecil.Cil;
+using System.Text.RegularExpressions;
+
+namespace MonoDroid.Tuner
+{
+	public class RemoveResourceDesignerStep : BaseStep
+	{
+		TypeDefinition mainDesigner = null;
+		AssemblyDefinition mainAssembly = null;
+		CustomAttribute mainDesignerAttribute;
+		Dictionary<string, int> designerConstants;
+		Regex opCodeRegex = new Regex (@"([\w]+): ([\w]+) ([\w.]+) ([\w:./]+)");
+
+		protected override void Process ()
+		{
+			// resolve the MainAssembly Resource designer TypeDefinition
+			AndroidLinkConfiguration config = AndroidLinkConfiguration.GetInstance (Context);
+			if (config == null)
+				return;
+			foreach(var asm in config.Assemblies) {
+				if (FindResourceDesigner (asm, mainApplication: true, designer: out mainDesigner, designerAttribute: out mainDesignerAttribute)) {
+					mainAssembly = asm;
+				 	break;
+				}
+			}
+			if (mainDesigner == null) {
+				Context.LogMessage ($"  Main Designer not found.");
+				return;
+			}
+			Context.LogMessage ($"  Main Designer found {mainDesigner.FullName}.");
+			designerConstants = BuildResourceDesignerFieldLookup (mainDesigner);
+		}
+
+		protected override void EndProcess ()
+		{
+			if (mainDesigner != null) {
+				Context.LogMessage ($"  Setting Action on {mainAssembly.Name} to Save.");
+				Annotations.SetAction (mainAssembly, AssemblyAction.Save);
+			}
+		}
+
+		bool FindResourceDesigner (AssemblyDefinition assembly, bool mainApplication, out TypeDefinition designer, out CustomAttribute designerAttribute)
+		{
+			string designerFullName = null;
+			designer = null;
+			designerAttribute = null;
+			foreach (CustomAttribute attribute in assembly.CustomAttributes)
+			{
+				if (attribute.AttributeType.FullName == "Android.Runtime.ResourceDesignerAttribute")
+				{
+					designerAttribute = attribute;
+					if (attribute.HasProperties)
+					{
+						foreach (var p in attribute.Properties)
+						{
+							if (p.Name == "IsApplication" && (bool)p.Argument.Value == (mainApplication ? mainApplication : (bool)p.Argument.Value))
+							{
+								designerFullName = attribute.ConstructorArguments[0].Value.ToString ();
+								break;
+							}
+						}
+					}
+					break;
+
+				}
+			}
+			if (string.IsNullOrEmpty(designerFullName))
+				return false;
+
+			foreach (ModuleDefinition module in assembly.Modules)
+			{
+				foreach (TypeDefinition type in module.Types)
+				{
+					if (type.FullName == designerFullName)
+					{
+						designer = type;
+						return true;
+					}
+				}
+			}
+			return false;
+		}
+
+		Dictionary<string, int> BuildResourceDesignerFieldLookup (TypeDefinition type)
+		{
+			var output = new Dictionary<string, int> ();
+			foreach (TypeDefinition definition in type.NestedTypes)
+			{
+				foreach (FieldDefinition field in definition.Fields)
+				{
+					string key = $"{definition.Name}::{field.Name}";
+					if (!output.ContainsKey (key))
+						output.Add(key, int.Parse (field.Constant?.ToString () ?? "0"));
+				}
+			}
+			return output;
+		}
+
+		void ClearDesignerClass (TypeDefinition designer)
+		{
+			Context.LogMessage ($"    TryRemoving {designer.FullName}");
+			designer.NestedTypes.Clear ();
+			designer.Methods.Clear ();
+			designer.Fields.Clear ();
+			designer.Properties.Clear ();
+			designer.CustomAttributes.Clear ();
+			designer.Interfaces.Clear ();
+			designer.Events.Clear ();
+		}
+
+		void FixBody (MethodBody body, TypeDefinition localDesigner)
+		{
+			Dictionary<Instruction, int> instructions = new Dictionary<Instruction, int>();
+			var processor = body.GetILProcessor ();
+			string designerFullName = $"{localDesigner.FullName}/";
+			foreach (var i in body.Instructions)
+			{
+				string line = i.ToString ();
+				if (line.Contains (designerFullName) && !instructions.ContainsKey (i))
+				{
+					var match = opCodeRegex.Match (line);
+					if (match.Success && match.Groups.Count == 5) {
+						string key = match.Groups[4].Value.Replace (designerFullName, string.Empty);
+						if (designerConstants.ContainsKey (key) && !instructions.ContainsKey (i))
+							instructions.Add(i, designerConstants [key]);
+					}
+				}
+			}
+			if (instructions.Count > 0)
+				Context.LogMessage ($"    Fixing up {body.Method.FullName}");
+			foreach (var i in instructions)
+			{
+				var newCode = Extensions.CreateLoadArraySizeOrOffsetInstruction (i.Value);
+				Context.LogMessage ($"      Replacing {i.Key}");
+				Context.LogMessage ($"      With {newCode}");
+				processor.Replace(i.Key, newCode);
+			}
+		}
+
+		void FixType (TypeDefinition type, TypeDefinition localDesigner)
+		{
+			foreach (MethodDefinition method in type.Methods)
+			{
+				if (!method.HasBody)
+					continue;
+				FixBody (method.Body, localDesigner);
+			}
+			foreach (PropertyDefinition property in type.Properties)
+			{
+				if (property.GetMethod != null && property.GetMethod.HasBody)
+				{
+					FixBody (property.GetMethod.Body, localDesigner);
+				}
+				if (property.SetMethod != null && property.SetMethod.HasBody)
+				{
+					FixBody (property.SetMethod.Body, localDesigner);
+				}
+			}
+			foreach (TypeDefinition nestedType in type.NestedTypes)
+			{
+				FixType (nestedType, localDesigner);
+			}
+		}
+
+		protected override void ProcessAssembly (AssemblyDefinition assembly)
+		{
+			if (mainDesigner == null)
+				return;
+			var fileName = assembly.Name.Name + ".dll";
+			if (MonoAndroidHelper.IsFrameworkAssembly (fileName))
+				return;
+
+			Context.LogMessage ($"  Fixing up {assembly.Name.Name}");
+			TypeDefinition localDesigner = null;
+			CustomAttribute designerAttribute;
+			if (assembly != mainAssembly) {
+				Context.LogMessage ($"   {assembly.Name.Name} is not the main assembly. ");
+				if (!FindResourceDesigner (assembly, mainApplication: false, designer: out localDesigner, designerAttribute: out designerAttribute)) {
+					Context.LogMessage ($"   {assembly.Name.Name} does not have a designer file.");
+					return;
+				}
+			} else {
+				Context.LogMessage ($"   {assembly.Name.Name} is the main assembly. ");
+				localDesigner = mainDesigner;
+				designerAttribute = mainDesignerAttribute;
+			}
+
+			Context.LogMessage ($"   {assembly.Name.Name} has designer {localDesigner.FullName}.");
+
+			foreach (var mod in assembly.Modules) {
+				foreach (var type in mod.Types) {
+					if (type == localDesigner)
+						continue;
+					FixType (type, localDesigner);
+				}
+			}
+			ClearDesignerClass (localDesigner);
+			if (designerAttribute != null) {
+				assembly.CustomAttributes.Remove (designerAttribute);
+			}
+		}
+	}
+}
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -860,6 +860,7 @@ because xbuild doesn't support framework reference assemblies.
 <Target Name="_CreatePropertiesCache" DependsOnTargets="_SetupDesignTimeBuildForBuild;_SetLatestTargetFrameworkVersion;_ResolveMonoAndroidSdks">
 	<PropertyGroup>
 		<AndroidAddKeepAlives Condition="'$(AndroidAddKeepAlives)' == '' And '$(AndroidIncludeDebugSymbols)' != 'True'">True</AndroidAddKeepAlives>
+		<AndroidLinkResources Condition="'$(AndroidLinkResources)' == '' And '$(AndroidIncludeDebugSymbols)' != 'True'">False</AndroidLinkResources>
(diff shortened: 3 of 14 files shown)


src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/StripEmbeddedLibraries.cs
```diff
@@ -4,9 +4,192 @@
...
 using System.Linq;
 using Xamarin.Android.Tasks;
+using System.Collections.Generic;
+using Mono.Cecil.Cil;
+using System.Text.RegularExpressions;
 
 namespace MonoDroid.Tuner
 {
+#if !NET5_LINKER
+	public class RemoveResourceDesigner : BaseStep
```
dellis1972, line 14: Need to split this into a separate file.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/StripEmbeddedLibraries.cs
```diff
@@ -4,9 +4,192 @@
...
+	{
+		TypeDefinition mainDesigner = null;
+		AssemblyDefinition mainAssembly = null;
+		Dictionary<string, int> designerConstants;
+		Regex opCodeRegex = new Regex (@"([\w]+): ([\w]+) ([\w.]+) ([\w:./]+)");
+
+		protected override void Process ()
+		{
+			// resolve the MainAssembly Resource designer TypeDefinition
+			foreach(var asm in Context.GetAssemblies ()) {
```
dellis1972, line 24: .net 5/6 linker api does not have `GetAssemblies`. 
At this point we need to find the "Main" Resource class for the application (i.e IsApplication = true). We need to do this before we process any other assemblies as this class has the final resource id's we need to replace in the references.

src/Microsoft.Android.Sdk.ILLink/SetupStep.cs
```diff
@@ -53,6 +53,7 @@ protected override void Process ()
 				InsertAfter (new GenerateProguardConfiguration (proguardPath),  "CleanStep");
 
 			InsertAfter (new StripEmbeddedLibraries (),  "CleanStep");
+			//InsertAfter (new RemoveResourceDesigner (),  "CleanStep");
```
dellis1972, line 56: when we finally get this working in .net 5/6 not sure if this is the best place to insert this step.

dellis1972:
@radekdoulik this is a work in progress, but if you have any advice on how to do this more efficiently please feel free to comment.

src/Microsoft.Android.Sdk.ILLink/SetupStep.cs
radekdoulik: if we want it at the same place as in the "old" linker, move it one line up.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/StripEmbeddedLibraries.cs
radekdoulik: We can workaround the lack of `GetAssemblies ()` the same way as @rolfbjarne does here: https://github.com/xamarin/xamarin-macios/blob/main/tools/dotnet-linker/Steps/CollectAssembliesStep.cs
Context: https://github.com/mono/linker/issues/1455

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/StripEmbeddedLibraries.cs
```diff
@@ -4,9 +4,192 @@
...
+						if (designerConstants.ContainsKey (key))
+							instructions.Add(i, designerConstants[key]);
+					}
+				}
+			}
+			if (instructions.Count > 0)
+				Context.LogMessage ($"    Fixing up {body.Method.FullName}");
+			foreach (var i in instructions)
+			{
+				var newCode = processor.Create(OpCodes.Ldc_I4, i.Value);
```
radekdoulik, line 132: This can be optimized a bit by using [CreateLoadArraySizeOrOffsetInstruction](https://github.com/xamarin/xamarin-android/blob/d471b4a4308310afb8fa3fa1047003a7caedd57f/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MonoDroidMarkStep.cs#L318) plus maybe some refactoring to move it outside of `MonoDroidMarkStep`, so that we can build it easily for .NET5/6.

dellis1972: @radekdoulik not sure I understand what you mean by "move it outside of MonoDroidMarkStep" ? can you explain in more detail?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/StripEmbeddedLibraries.cs
dellis1972: We had a `AndroidLinkContext` class already, so I assume I will need to add an Assemblies list to that. Then add a new step which will populate it. Then just cast the `_context` as a `AndroidLinkContext` in the ResourceDesignerStep.

dellis1972: So AndroidLinkContext is not available in .net5/6. That CollectAssembliesStep in macOS/ios has allot of dependencies. Is there a simpler way other than copying all of that stuff over?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/StripEmbeddedLibraries.cs
radekdoulik: The `MonoDroidMarkStep` is not built on .NET5/6, so if we would like to use the code in step used in .NET5/6, we need to move it to a helper class or some other place.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
```diff
@@ -0,0 +1,223 @@
...
+
+		public static AndroidLinkConfiguration GetInstance (LinkContext context)
+		{
+			if (!configurations.TryGetValue (context, out AndroidLinkConfiguration config)) {
+				configurations.Add (context, new AndroidLinkConfiguration ());
+			}
+			return config;
+		}
+	}
+	public class GetAssembliesStep : BaseStep
```
radekdoulik, line 26: Please put these classes in `GetAssembliesStep.cs` and `AndroidLinkConfiguration.cs`.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/StripEmbeddedLibraries.cs
radekdoulik: I think this was resolved last week in the meeting, right?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
dellis1972: Yup, I know this was just me testing stuff

jonathanpeppers (commented):
Might want to wait on: https://github.com/xamarin/Xamarin.Forms/pull/13176

Is there a way to make a test for this?

* Xamarin.Forms app using `<Image Source="foo" />` where `foo` is `AndroidResource`
* Launch the app, verify the image is displayed?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/StripEmbeddedLibraries.cs
```diff
@@ -4,6 +4,9 @@
 using System;
 using System.Linq;
 using Xamarin.Android.Tasks;
+using System.Collections.Generic;
+using Mono.Cecil.Cil;
+using System.Text.RegularExpressions;
```
jonathanpeppers, line 9: Maybe this was leftover and can be removed now?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
```diff
@@ -0,0 +1,211 @@
...
+using System;
+using System.Linq;
+using Xamarin.Android.Tasks;
+using System.Collections.Generic;
+using Mono.Cecil.Cil;
+using System.Text.RegularExpressions;
+
+namespace MonoDroid.Tuner
+{
+	public class RemoveResourceDesignerStep : BaseStep
```
jonathanpeppers, line 13: An idea that could be done in a future iteration.

Could this step also modify this class:

https://github.com/xamarin/xamarin-android/blob/7ba62bac5a25eb783a6da9c2b6a989a820695a1f/src/Mono.Android/Android.Runtime/ResourceIdManager.cs#L6-L9

If `UpdateIdValues()` was a completely empty method, some reflection could be skipped at startup.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -859,7 +859,8 @@ because xbuild doesn't support framework reference assemblies.
 
 <Target Name="_CreatePropertiesCache" DependsOnTargets="_SetupDesignTimeBuildForBuild;_SetLatestTargetFrameworkVersion;_ResolveMonoAndroidSdks">
 	<PropertyGroup>
-		<AndroidAddKeepAlives Condition="'$(AndroidAddKeepAlives)' == '' And '$(AndroidIncludeDebugSymbols)' != 'True'">True</AndroidAddKeepAlives>
+	  <AndroidAddKeepAlives Condition="'$(AndroidAddKeepAlives)' == '' And '$(AndroidIncludeDebugSymbols)' != 'True'">True</AndroidAddKeepAlives>
+    <AndroidLinkResources Condition="'$(AndroidLinkResources)' == '' And '$(AndroidIncludeDebugSymbols)' != 'True'">False</AndroidLinkResources>
```
jonathanpeppers, line 863: The indentation here is odd, maybe could just make it match the other lines in this file.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
dellis1972: Depending on when the new Step runs, it might strip out that class already since the Designer is the thing that calls`UpdateIdValues`. But its worth doing in the future yes :)

jonpryor:
@jonathanpeppers [wrote](https://github.com/xamarin/xamarin-android/pull/5317#pullrequestreview-582553248):

> Might want to wait on: [xamarin/Xamarin.Forms#13176](https://github.com/xamarin/Xamarin.Forms/pull/13176)

xamarin/Xamarin.Forms#13176 has been merged.  How do we know when we can rely on it?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
jonpryor: @dellis1972 [wrote](https://github.com/xamarin/xamarin-android/pull/5317#discussion_r575295508):

> the Designer is the thing that calls `UpdateIdValues`.

This is incorrect.  All (current) `Resource.designer.cs` code calls `UpdateIdValues`:

https://github.com/xamarin/xamarin-android/blob/214dd76d9441c9974775f0b29feec0d5d8d1097f/src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs#L185-L192

https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Expected/GenerateDesignerFileExpected.cs

(which contains *21* references to `UpdateIdValues()`.

Any use of a `Resource.Descriptor.Name` field will cause `UpdateIdValues()` to be invoked.

Now, if after this "link resource nested types" change operates, the static constructor is *also* removed, meaning that `UpdateIdValues()` is likewise not invoked anymore, that would be cool.

If that's the case, though, that should *also* be mentioned in the PR/commit message.  That's a possibly important (and huge) semantic change.  (Not necessarily a "break", but certainly something we'll need to know about.)

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
```diff
@@ -0,0 +1,211 @@
...
+						output.Add(key, int.Parse (field.Constant?.ToString () ?? "0"));
+				}
+			}
+			return output;
+		}
+
+		void ClearDesignerClass (TypeDefinition designer)
+		{
+			Context.LogMessage ($"    TryRemoving {designer.FullName}");
+			designer.NestedTypes.Clear ();
```
jonpryor, line 109: This removes nested types, and thus would remove their static constructors…

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
```diff
@@ -0,0 +1,211 @@
...
+			designer.Methods.Clear ();
```
jonpryor, line 110: This will presumably remove the static constructor on the `Resource` type itself?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
```diff
@@ -0,0 +1,211 @@
...
+			designer.Fields.Clear ();
+			designer.Properties.Clear ();
+			designer.CustomAttributes.Clear ();
+			designer.GenericParameters.Clear ();
```
jonpryor, line 114: This feels weird.  If a detected `Resource.designer.cs` type contains generic parameters…perhaps it's *not* something we should be updating or touching?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/RemoveResourceDesignerStep.cs
dellis1972: Yes it would. They would no longer be needed. All there references in All libraries to those types will have been replaced with the actual values of the Id's. There is not need to call `global::Android.Runtime.ResourceIdManager.UpdateIdValues();` any more.

jonathanpeppers:
> xamarin/Xamarin.Forms#13176 has been merged. How do we know when we can rely on it?

@jonpryor the `[Obsolete]` messages say:
```csharp
[Obsolete("GetDrawableByName(string) is obsolete as of version 4.8. "
			+ "Please use GetDrawableId(string, context) instead.")]
```

I see the change in if I download & ilspy Xamarin.Forms [5.0.0.2012](https://www.nuget.org/packages/Xamarin.Forms/5.0.0.2012) but not [4.8.0.1821](https://www.nuget.org/packages/Xamarin.Forms/4.8.0.1821).

So maybe this is only in Xamarin.Forms 5.0 and higher.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AndroidLinkContext.cs
```diff
@@ -12,5 +13,6 @@ public AndroidLinkContext (Pipeline pipeline, AssemblyResolver resolver, ReaderP
 			: base (pipeline, resolver, readerParameters, factory) {}
 
 		public bool PreserveJniMarshalMethods { get; set; }
+
```
jonathanpeppers, line 16: I think there are maybe some leftover changes in this file, you could remove.

jonathanpeppers (approved):
Otherwise, LGTM 👍

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AndroidLinkContext.cs
dellis1972: I had to fix it :)

jonpryor:
Squash-and-merge:

Summary:

```
[Xamarin.Android.Build.Tasks] Add $(AndroidLinkResources) (#5317)
```

commit message:

```markdown
The generated `Resource` class in `Resource.designer.cs` can be quite
large.

When doing "Release" builds, update the linker to *remove* the
`Resource` class and its nested types.  We do this by examining all
the locations where the fields are used and replacing those IL calls
with a call which specifies the actual resource id value directly.

This will remove the need to call
`global::Android.Runtime.ResourceIdManager.UpdateIdValues();` from
the static constructors in the `Resource` class and its nested types,
which will improve app launch times (as `UpdateIdValues()` involves
Reflection, which is not "fast").

Due to an odd linker failure we cannot completely remove the
`Resource` class itself.  It *will* be completely empty.

"Debug" builds will *not* remove the `Resource` type, and will retain
the call to `UpdateIdValues()`, in order to ensure that app rebuild
and redeploy times are as fast as possible.

The new `$(AndroidLinkResources)` MSBuild property has been added to
control this feature.  It is disabled by default for now.

`$(AndroidLinkResources)`=True appears to save between 100-280kb for
an `.apk`.  This is dependent on how many assemblies are Android
assemblies and contain or use Resources.

Xamarin.Forms base app `apkdiff` results

	  -          10 assemblies/Mono.Android.dll
	  -       4,001 assemblies/Xamarin.Essentials.dll
	  -      45,426 assemblies/Xamarin.Forms.Platform.Android.dll
	  -     102,839 assemblies/formstest.Android.dll
	Summary:
	  +           0 Other entries 0.00% (of 887,808)
	  +           0 Dalvik executables 0.00% (of 3,341,152)
	  -     152,276 Assemblies -3.23% (of 4,709,077)
	  +           0 Shared libraries 0.00% (of 41,693,796)
	  -     151,552 Package size difference -0.73% (of 20,820,695)

Maps sample `apkdiff` results

	  -           4 lib/armeabi-v7a/libxamarin-app.so
	  -          14 assemblies/Mono.Android.dll
	  -       4,757 assemblies/Xamarin.Essentials.dll
	  -      46,051 assemblies/Xamarin.Forms.Platform.dll
	  -      47,586 assemblies/Xamarin.Forms.Platform.Android.dll
	  -      48,470 assemblies/Xamarin.Forms.Maps.Android.dll
	  -     135,589 assemblies/FormsMapsSample.Android.dll
	Summary:
	  +           0 Other entries 0.00% (of 1,137,209)
	  +           0 Dalvik executables 0.00% (of 2,519,972)
	  -     282,467 Assemblies -5.51% (of 5,123,833)
	  -           4 Shared libraries -0.00% (of 41,843,596)
	  -     282,624 Package size difference -1.34% (of 21,075,664)
```

PR #5318, opened 2020-11-19
[Xamarin.Android.Build.Tasks] Make Proguard mapping.txt opt-in

pjcollins:
Context: https://github.com/xamarin/xamarin-android/pull/5304

The Proguard mapping.txt PR was not quite ready for prime time and has
seemingly caused some new proguard test failures and an .apk size
regression in our Xamarin.Forms baseline.  Let's disable this by default
for now until we can fix these issues.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
@@ -56,11 +56,11 @@ namespace Xamarin.Android.Build.Tests
 			};
 			proj.SetProperty (proj.ReleaseProperties, KnownProperties.AndroidDexTool, "d8");
 			proj.SetProperty (proj.ReleaseProperties, KnownProperties.AndroidLinkTool, "r8");
+			// Projects must provide a $(AndroidProguardMappingFile) value to opt in
+			proj.SetProperty (proj.ReleaseProperties, "AndroidProguardMappingFile", @"$(OutputPath)\mapping.txt");
 
 			using (var b = CreateApkBuilder ()) {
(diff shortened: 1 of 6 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
```diff
@@ -56,9 +56,12 @@ public void CheckProguardMappingFileExists ()
 			};
 			proj.SetProperty (proj.ReleaseProperties, KnownProperties.AndroidDexTool, "d8");
 			proj.SetProperty (proj.ReleaseProperties, KnownProperties.AndroidLinkTool, "r8");
+
 			using (var b = CreateApkBuilder ()) {
-				Assert.IsTrue (b.Build (proj), "build should have succeeded.");
 				string mappingFile = Path.Combine (Root, b.ProjectDirectory, proj.OutputPath, "mapping.txt");
+				// Projects must provide a $(AndroidProguardMappingFile) value to opt in
+				proj.SetProperty (proj.ReleaseProperties, "AndroidProguardMappingFile", mappingFile);
```
jonathanpeppers, line 63: I think we should make the `.csproj` in this test use `<AndroidProguardMappingFile>$(OutputPath)mapping.txt</AndroidProguardMappingFile>`, and then make sure we have this example in the release notes:

https://github.com/xamarin/xamarin-android/blob/55faee0fe9e489bc03cc8ac08140169ddac4c5b5/Documentation/release-notes/5304.md

pjcollins: Good call, updated

dellis1972 (approved):

jonathanpeppers (approved):

jonathanpeppers:
The designer is failing on NuGet restore, but I think we can merge without this check.

PR #5325, opened 2020-11-23
[Xamarin.Android.Build.Tasks] Add Support for AndroidManifest.xml overlays

### LINKED ISSUE #5312: Add Support for AndroidManifest.xml overlays
Users would like the ability to merge AndroidManifest.xml files with additional debugging settings etc.
See https://github.com/xamarin/xamarin-android/issues/3428#issuecomment-729629251

It looks like I had the foresight to add a ManifestOverlayFiles property to the ManifestMerger Task, but neglected to hook it up in the targets. This is the exact property users would need.

Adding a new build action like `AndroidManifestOverlay` might do the trick. This can then be passed in to the manifest merger as an overlay on top of the other files.

Then Users might be able to do something like this.

<ItemGroup>
   <AndridManifestOverlay Include="AndroidManifestDev.xml" Condition=" '$(Configuration)' != 'Release' " />
</ItemGroup>

dellis1972:
Fixes #5312

The manifest merger tool supports providing additional AndroidManifest.xml
files which will `overlay` on top of the final file. This commit adds a
new Build Action `AndroidManifestOverlay` which can be used to provide
these files to the manifest merger.

Users can now use these overlay files to alter the manifest during
build time. This can be for adding new permissions or even removing
ones that are not needed. It will also allow for different manifest
files to be generated for debug/release configurations. This can be
done by conditionally including `overlay` files depending on the
`$(Configuration)`.

```
<ItemGroup>
  <AndroidManifestOverlay Include="DebugPermissions.xml" Condition=" '$(Configuration)' == 'Debug' " />
</ItemGroup>
```

### DIFF
--- a/Documentation/guides/building-apps/build-items.md
+++ b/Documentation/guides/building-apps/build-items.md
@@ -83,6 +83,36 @@ which tests to enable and disable.
 See the [lint documentation](https://developer.android.com/studio/write/lint)
 for more details.
 
+## AndroidManifestOverlay
+
+The build action `AndroidManifestOverlay` can we used to provide additional
+`AndroidManifest.xml` files to the [Manifest Merger]([~/android/deploy-test/building-apps/build-properties.md#](https://developer.android.com/studio/build/manifest-merge)) tool.
+Files with this build action will be passed to the Manifest Merger along with
+the main `AndroidManifest.xml` file and any additional manifest files from
+references. These will then be merged into the final manifest.
+
+You can use this build action to provide additional changes and settings to
+your app depending on your build configuration. For example if you need to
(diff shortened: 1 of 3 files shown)


Documentation/guides/building-apps/build-items.md
```diff
@@ -83,6 +83,36 @@ which tests to enable and disable.
 See the [lint documentation](https://developer.android.com/studio/write/lint)
 for more details.
 
+## AndroidManifestOverlay
+
+The build action `AndroidManifestOverlay` can we used to provide additional
```
radekdoulik, line 88: Probably a typo  `...can be used...` ('s/we/be')?

radekdoulik (approved):

Issue #4810, opened 2020-06-12
Support for Dynamic Feature Modules


fruitedev:
Hello,

I know there is support for AppBundles which makes things easier, but i couldn't find whether Dynamic Feature Modules is supported or not. Support of [Dynamic Feature Modules](https://developer.android.com/studio/projects/dynamic-delivery#customize_delivery) will be really really helpful for us so please share any plan you have.

Thanks,

jonathanpeppers:
@fruitedev I remember there might have been some limitations when I first implemented app bundles:

https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/app-bundles.md#what-are-app-bundles

I remember something about needing to support "Instant Apps" for dynamic modules to work, which has a hard file-size limit. The Mono runtime was over the limit.

We should still look into this again. Anyone interested, please give a thumbs up above and that will help us know the demand, thanks!

johnthiriet:
@jonathanpeppers Let's consider that the module I want to load dynamically is an aar archive with some jar dependencies and I have a binding over this one. Would that make the process easier on your side ? I mean I do want to load this aar dynamically since it's pretty heavy.

evansmj:
Now that Apple has introduced App Clips, clients are now more interested in Android Instant Apps.  Using app bundles may not be so possible due to the Mono runtime limit and that is fine, but if there was a way we could at least attach natively written instant app apk's as feature modules to the Xamarin.Android app that would be great.

johnthiriet:
Still really really interested in the dynamic delivery for our applications. We are open for a quick call to explain more about why if you want.
@jonathanpeppers @davidortinau.

dellis1972:
I found that a community member has already bound the PlayCore library, https://github.com/PatGet/XamarinPlayCoreUpdater. This might have enough support to allow you to deliver a natively written apk as a dynamic feature and keep the main app as a Xamarin App.

johnthiriet:
@dellis1972 Hum that might be but actually "publishing" this module to the play store will maybe be tricky if even possible as the application is in Xamarin. Even by creating a native application just for packaging this module I think it will not work well.
Have you heard about someone doing it ?

dellis1972:
@johnthiriet  I haven't heard of anyone doing it yet. The developer it seems only used it for in app updates.

dellis1972:
I did some investigation last week as to what might be needed to support this feature. 
[...]

PR #5327, opened 2020-11-24
[Xamarin.Android.Build.Tasks] Allow users to specify modules for apk sets.

dellis1972:
Context #4810

When using app bundles users install the application from an `apkset`.
At this time this only contains the base application. However we might
in the future support dynamic features in some fashion. In this case
developers should be able to pick which features they want installed.

This commit adds an `AndroidInstallModules` ItemGroup which will allow
developers to specify which modules to install. By default it will be
empty. This means all modules which are to be installed on first install
will be installed. So there is no change in the current behavior.

This commit also adds support for passing additional arguments to 
`bundle tool` via the new `AndroidBundleToolExtraArgs` property.

### DIFF
--- a/Documentation/guides/building-apps/build-properties.md
+++ b/Documentation/guides/building-apps/build-properties.md
@@ -204,6 +204,8 @@ Specifies additional
 command-line options to pass to the **bundletool** command when
 build app bundles.
 
+This property was added in Xamarin.Android 11.3.
+
 ## AndroidClassParser
 
 A string property which controls how
--- a/Documentation/guides/building-apps/build-properties.md
+++ b/Documentation/guides/building-apps/build-properties.md
@@ -198,6 +198,12 @@ Added in Xamarin.Android 10.3.
 
 [bundle-config-format]: https://developer.android.com/studio/build/building-cmdline#bundleconfig
 
+## AndroidBundleToolExtraArgs
+
+Specifies additional
+command-line options to pass to the **bundletool** command when
+build app bundles.
+
 ## AndroidClassParser
 
 A string property which controls how
--- a/src/Xamarin.Android.Build.Tasks/Tasks/InstallApkSet.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/InstallApkSet.cs
@@ -7,7 +7,7 @@ namespace Xamarin.Android.Tasks
 {
 	/// <summary>
 	/// Invokes `bundletool` to install an APK set to an attached device
-	/// 
+	///
 	/// Usage: bundletool install-apks --apks=foo.apks
 	/// </summary>
 	public class InstallApkSet : BundleToolAdbTask
@@ -16,6 +16,8 @@ namespace Xamarin.Android.Tasks
 
 		[Required]
 		public string ApkSet { get; set; }
+
+		public string[] Modules  { get; set; }
 
 		internal override CommandLineBuilder GetCommandLineBuilder ()
 		{
@@ -28,8 +30,11 @@ namespace Xamarin.Android.Tasks
 			// --modules: List of modules to be installed, or "_ALL_" for all modules.
 			// Defaults to modules installed during first install, i.e. not on-demand.
 			// Xamarin.Android won't support on-demand modules yet.
-			cmd.AppendSwitchIfNotNull ("--modules ", "_ALL_");
-			
+			if (Modules == null)
+				cmd.AppendSwitchIfNotNull ("--modules ", "_ALL_");
+			else
+				cmd.AppendSwitchIfNotNull ("--modules ", $"\"{string.Join ("\",\"", Modules)}\"");
+
 			return cmd;
 		}
 	}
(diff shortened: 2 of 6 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/InstallApkSet.cs
```diff
@@ -28,8 +30,11 @@ internal override CommandLineBuilder GetCommandLineBuilder ()
 			// --modules: List of modules to be installed, or "_ALL_" for all modules.
 			// Defaults to modules installed during first install, i.e. not on-demand.
 			// Xamarin.Android won't support on-demand modules yet.
-			cmd.AppendSwitchIfNotNull ("--modules ", "_ALL_");
-			
+			if (Modules == null)
+				cmd.AppendSwitchIfNotNull ("--modules ", "_ALL_");
+			else
+				cmd.AppendSwitchIfNotNull ("--modules", string.Join (",", Modules));
```
jonathanpeppers, line 36: Do we need to put quotes if the paths have a space in them?

dellis1972: Probably lol

src/Xamarin.Android.Build.Tasks/Tasks/InstallApkSet.cs
```diff
@@ -28,8 +30,11 @@ internal override CommandLineBuilder GetCommandLineBuilder ()
 			// --modules: List of modules to be installed, or "_ALL_" for all modules.
 			// Defaults to modules installed during first install, i.e. not on-demand.
 			// Xamarin.Android won't support on-demand modules yet.
-			cmd.AppendSwitchIfNotNull ("--modules ", "_ALL_");
-			
+			if (Modules == null)
+				cmd.AppendSwitchIfNotNull ("--modules ", "_ALL_");
+			else
+				cmd.AppendSwitchIfNotNull ("--modules", string.Join ("\",\"", Modules));
```
jonathanpeppers, line 36: I don't think this would put the leading & ending `"`, so you would get:
```
--modules foo","bar
```

jonathanpeppers: Or does `AppendSwitchIfNotNull` do that for you?

jonathanpeppers: Also this might be needed:
```suggestion
				cmd.AppendSwitchIfNotNull ("--modules ", string.Join ("\",\"", Modules));
```

Documentation/guides/building-apps/build-properties.md
```diff
@@ -633,6 +639,11 @@ APK root directory. The format of the path is `lib\ARCH\wrap.sh` where
 + `x86_64`
 + `x86`
 
+## AndroidInstallModules
```
jonpryor, line 642: This should go into `build-items.md`, not `build-properties.md`.  It's an item group, not a property.

src/Xamarin.Android.Build.Tasks/Tasks/InstallApkSet.cs
```diff
@@ -28,8 +30,11 @@ internal override CommandLineBuilder GetCommandLineBuilder ()
 			// --modules: List of modules to be installed, or "_ALL_" for all modules.
 			// Defaults to modules installed during first install, i.e. not on-demand.
 			// Xamarin.Android won't support on-demand modules yet.
-			cmd.AppendSwitchIfNotNull ("--modules ", "_ALL_");
-			
+			if (Modules == null)
+				cmd.AppendSwitchIfNotNull ("--modules ", "_ALL_");
+			else
+				cmd.AppendSwitchIfNotNull ("--modules", $"\"{string.Join ("\",\"", Modules)}\"");
```
jonathanpeppers, line 36: This switch would be:
```
--modules"foo","bar"
```

I don't know if it needs a space like the `--modules _ALL_` switch above?
```suggestion
				cmd.AppendSwitchIfNotNull ("--modules ", $"\"{string.Join ("\",\"", Modules)}\"");
```

jonpryor:
Some "public facing" documentation would be useful, not so much so that we can have something in `Documentation`, but more to validate that how this works is appropriate.  (It'll become part of the commit message!)

For example, this approach involves using the `@(AndroidInstallModules)` item group.  Item groups can't be specified on the command line.  Is this a problem?  Is there an expectation to be able to specify modules on a `msbuild /t:Install` command line, or is editing the `.csproj` the preferred experience?

dellis1972:
@jonpryor currently this new `AndroidInstallModules` is not very useful. Its the beginning of a much larger piece of work outlined in https://github.com/xamarin/xamarin-android/issues/4810#issuecomment-737229807. I can document the new ItemGroup, but users will currently have no use for it since we need to do allot of the other stuff on the list at https://github.com/xamarin/xamarin-android/issues/4810#issuecomment-737229807 to make it useful.

dellis1972:
@jonpryor I had added the docs for the ItemGroup here https://github.com/xamarin/xamarin-android/pull/5327/files#diff-db5645f8a1e864de401ac4a261afe73af36e10043135012bebc8cacfeb3820d0R149. There isn't much more to say since uses can't use this feature yet unless they build a module completely manually (and process is not not documented).

jonathanpeppers (approved):

PR #5340, opened 2020-11-30
[Xamarin.Android.Build.Tasks] Do not add READ_EXTERNAL_STORAGE permission

### LINKED ISSUE #5334: VS 2019 16.9.0. Prev 1.0 Removed Read/Write External Storage Permissions - but obj/manifest still generates them
I just download a new version from Google Play. In this build, I had removed the above permissions because I'd decided I didn't need them anymore. I then downloaded to a Pixel3a (Android 10) and made a test purchase and then ran the program. It didn't ask for any permissions. Did the same with a Samsung S20 and it did ask for permission. Checked my manifest again and it was fine. Checked the properties page and they are both unchecked.
Checked both generated obj/manifests and both Debug and Release still have both permissions.

How can I force the generated obj/manifests to drop those permissions? I can't actually remember when I ran any of my apps without those permissions, so I thought all that was required was either uncheck on the Properties page or just delete them from the manifest directly.

Doesn't explain why the Pixel3a didn't ask for them though!!!

dellis1972:
Fixes #5334

We are still adding the `android.permission.READ_EXTERNAL_STORAGE` permission
when we don't embed assemblies into the apk. Since 71dc7185 we no longer need
this permission since we fast deploy to the applications internal directory.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs
@@ -402,9 +402,6 @@ namespace Xamarin.Android.Tasks {
 			if (Debug || NeedsInternet)
 				AddInternetPermissionForDebugger ();
 
-			if (!embed)
-				AddFastDeployPermissions ();
-
 			// If the manifest has android:installLocation, but we are targeting
 			// API 7 or lower, remove it for the user and show a warning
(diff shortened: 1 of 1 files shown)


radekdoulik (approved):

jonathanpeppers (approved):
I restarted a few tests.

`CheckIncludedAssemblies` is going to fail until we fix it on master.

PR #5341, opened 2020-11-30
[illink] Support `LinkerDumpDependencies`

radekdoulik:
Support dependencies dump for ILLink with existing
`LinkerDumpDependencies` property.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -85,7 +85,10 @@ namespace Xamarin.Android.Build.Tests
 			using (var b = BuildHelper.CreateApkBuilder (Path.Combine ("temp", TestName))) {
 				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
 
-				var depsFile = GetLinkedPath (b, true, "linker-dependencies.xml.gz");
+				var depsFilename = "linker-dependencies.xml.gz";
+				var depsFile = Builder.UseDotNet
+					? GetLinkedPath (b, true, depsFilename)
+					: Path.Combine (proj.Root, b.ProjectDirectory, depsFilename);
 				FileAssert.Exists (depsFile);
 			}
 		}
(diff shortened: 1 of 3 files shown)


jonathanpeppers (commented):
Should we add a test for this feature somewhere here:

https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/LinkerTests.cs

Basically just turn it on and check the `*.gz` file exists.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -76,6 +84,12 @@ public void BuildReleaseArm64 ([Values (false, true)] bool forms)
 			// use BuildHelper.CreateApkBuilder so that the test directory is not removed in tearup
 			using (var b = BuildHelper.CreateApkBuilder (Path.Combine ("temp", TestName))) {
 				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
+
+				var depsFilename = "linker-dependencies.xml.gz";
+				var depsFile = Builder.UseDotNet
+					? GetLinkedPath (b, true, depsFilename)
+					: Path.Combine (proj.Root, b.ProjectDirectory, depsFilename);
```
jonathanpeppers, line 91: Should the file show up in `$(OutputPath)`? I guess it was showing up in the same directory as the `.csproj` before .NET 6.

radekdoulik: It is now written to the same directory as the linked assemblies, which kind of makes sense as they are usually also used, when analyzing linker behavior. The dependencies are not published with or within the apk, so I think the current location is fine. Or is there a reason to put it to output path I don't see?

jonathanpeppers: I don't see how our users would ever find it...

They are used to looking in `bin` for `.apk` or `.dll` files -- you don't really go looking in `obj` for anything.

We should either document where you can find the file, or move it to `$(OutputPath)`? I think either would be fine.

PR #5349, opened 2020-12-01
[One .NET] improve default MSBuild item groups

jonathanpeppers:
When implementing #5348, I noticed that we could add a wildcard to
make Java bindings simpler:

This way you can just put `.xml` files in the `Transforms` folder so
they will be picked up automatically.

We also need to include some additional wildcards:

    <AndroidLibrary       Include="**\*.jar" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
    <AndroidLibrary       Include="**\*.aar" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />

To make sure we exclude folders like `bin` and `obj`, we can rely on
`$(DefaultItemExcludes)` and `$(DefaultExcludesInProjectFolder)` as
the dotnet/sdk does:

https://github.com/dotnet/sdk/blob/3e9e1d1e3082cab14593b54562c956b155881658/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.DefaultItems.props#L37

I updated a test that uses a `Metadata.xml`, and it now relies on the
new defaults.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
@@ -16,9 +16,9 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
 -->
 <Project>
 
-  <!-- Default Resource file inclusion -->
-  <!-- https://developer.android.com/guide/topics/resources/providing-resources -->
   <ItemGroup Condition=" '$(EnableDefaultAndroidItems)' == 'true' ">
+    <!-- Default Resource file inclusion -->
+    <!-- https://developer.android.com/guide/topics/resources/providing-resources -->
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.xml" />
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.axml" />
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.png" />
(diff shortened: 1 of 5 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
```diff
@@ -35,4 +35,9 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <AndroidAsset Include="$(MonoAndroidAssetsPrefix)\**\*" Exclude="$(MonoAndroidAssetsPrefix)\**\.*\**" />
   </ItemGroup>
 
+  <!-- Default files for bindings -->
+  <ItemGroup Condition=" '$(EnableDefaultAndroidItems)' == 'true' ">
+    <TransformFile Include="Transforms\*.xml" />
+  </ItemGroup>
```
jonathanpeppers, line 41: Thinking about `.jar` files, in "legacy" Xamarin.Android you would do:

    <EmbeddedJar Include="Jars\foo.jar" />

However, in .NET 6 you need to specify `%(Bind)` if you actually want a binding:

    <AndroidLibrary Include="Jars\foo.jar" />
    <AndroidLibrary Include="Jars\bar.jar" Bind="true" />

I don't think we should wildcard these with `**\*.jar`, because bindings would have to do this anyway:

    <AndroidLibrary Update="Jars\bar.jar" Bind="true" />

I think we should just require `@(AndroidLibrary)` to be listed, but we can discuss.

radekdoulik: > However, in .NET 6 you need to specify %(Bind) if you actually want a binding:

What is the reason for that? In what scenario one would add jar without setting %Bind? (I am just curious, would like to improve my understanding of how bindings work)

dellis1972: allot of .jar libraries have additional dependencies which the user probably won't want to bind. I don't have an example but we have seen this before. In .aar files for example you can get the main .jar , but then a bunch of internal ones that need to be shipped, but don't actually need binding since they are not exposed in the main API. So allowing the user to "opt in" by using the `Bind` attribute seemed like a better approach than just binding everything by default.

That said, we could use the `ItemDefinitionGroup` to define the default value for `Bind` and have that depend on a property like `AndroidBindAll` so that users can switch binding of everything on if they want to...


```
<PropertyGroup>
   <AndroidBindAll Condition=" '$(AndroidBindAll)' == '' ">false</AndroidBindAll> 
</PropertyGroup>
<ItemDefinitionGroup>
		< AndroidLibrary Bind="$(AndroidBindAll)" />
</ItemDefinitionGroup>
```

jonathanpeppers: @dellis1972 do you think we should consider naming it `$(AndroidBindingProject)` that would default `%(Bind)` to `true`?

MSBuild.Sdk.Extras has `$(IsBindingProject)`:

https://github.com/xamarin/AndroidX/blob/ed9b698a7293be82f819d9bdd2aa6f7021e47540/source/AndroidXProject.cshtml#L5

Since we have a separate template, this might make sense?

dellis1972: I'm ok with `AndroidBindingProject` It was just a thought. If we do this, it will bind all by default, and users will have to turn if off. Is that the way we want to go ? or do the opposite.

jonathanpeppers: I think it makes sense to make `$(AndroidBindingProject)` `false` by default and only set in the binding project template in #5348.

jonathanpeppers: As discussed with @jonpryor today, I think we will go with a different approach. We don't need a new `$(AndroidBindingProject)` property.

In this PR, I should make some changes:

1. `%(Bind)` should just default to `true`. Even though it may cause build errors, it is the most useful option. Users won't know how to turn it on if it isn't on by default.
1. Wildcard include: `**\*.jar`, `**\*.aar`, `**\*.so`. Make sure to exclude `bin`/`obj`!

For future changes I filed: https://github.com/xamarin/xamarin-android/issues/5357

dellis1972 (approved):

Issue #4977, opened 2020-08-04
"warning XA0113: Google Play requires that new applications and updates must use a TargetFrameworkVersion of v9.0 (API level 28) or above." is outdated as of August 3, 2020


brendanzagaeski:
### Steps to reproduce

1. Create a new Android application.
2. Set the `TargetFrameworkVersion` MSBuild property to `v8.1` or lower.
3. Build the project. 

### Expected behavior

Now that a new minimum of Android 10 (API level 29) for new apps on Google Play has [just come into effect as of August 3, 2020](https://support.google.com/googleplay/android-developer/answer/113469#targetsdk), the XA0113 warning would ideally say something like:

```
warning XA0113: Google Play requires that new applications must use a TargetFrameworkVersion of v10.0 (API level 29) and updates must use a TargetFrameworkVersion of v9.0 (API level 28) or above. You are currently targeting v8.1 (API level 27).
```

In fact, assuming the Visual Studio 2019 release cadence remains similar for version 16.8 as for the last few releases, it might even make sense to update the message in the d16-8 branch directly to account for the November 2 change for app updates too:

```
warning XA0113: Google Play requires that new applications and updates must use a TargetFrameworkVersion of v10.0 (API level 29) or above. You are currently targeting v8.1 (API level 27).
```

### Actual behavior

The current XA0113 warning is now outdated:

```
warning XA0113: Google Play requires that new applications and updates must use a TargetFrameworkVersion of v9.0 (API level 28) or above. You are currently targeting v8.1 (API level 27).
```

### Version information

xamarin-android/master:

https://github.com/xamarin/xamarin-android/blob/396aca1c13a0d8049ef3ab6c4b3e97e79cbec21b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx#L240-L241

PR #5351, opened 2020-12-01
[Xamarin.Android.Build.Tasks] bump XA0113 warning to 29

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/4977
Context: https://support.google.com/googleplay/android-developer/answer/9859152?visit_id=637424522037563363-1219346479&rd=1#targetsdk

We currently warn if `$(TargetFrameworkVersion)` is less than 28:

    Google Play requires that new applications and updates must use a TargetFrameworkVersion of v9.0 (API level 28) or above.

Since November 2, 2020, Google Play requires API 29, so we need to
update this warning.


dellis1972 (approved):
This looks good. 

I do wonder though, if we should change the text so that we pass in the values for the expected  TargetFramework  and API Level... That way we don't have to re-translate the text every time we bump this.

jonathanpeppers:
> I do wonder though, if we should change the text so that we pass in the values for the expected TargetFramework and API Level...

Yes, I think this should probably use additional `string.Format` arguments if this is going to change every year.

dellis1972 (approved):

icalderond:
This solution work for me. Thanks

PR #5354, opened 2020-12-02
Remove custom types marking for types which XA does not own for netcore

marek-safar:
This is or will be handled by BCL libraries and XA tooling should try
not to interfere with more fine tuned handling done by BCL.

Simplifies `[Preserve]` attribute implementation and removes
old code to preserve serialization, which doesn't work
well with new .NET5/6 runtime libraries. (the old serialization
code: https://github.com/xamarin/xamarin-android/blob/master/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs#L57-L82)

The serialization is currently [broken](https://github.com/dotnet/runtime/issues/45559)
and will be fixed in runtime libs. Part of `CustomLinkDescriptionPreserve`
test is thus disabled.

Also the attribute removal is now handled in illink.

The serialization fix leads to great reduction of assemblies size in simple
XA test app by cca 1MB, apkdiff before/after:

    Summary:
      -   1,199,411 Assemblies -55.66% (of 2,154,984)
      -   1,206,333 Package size difference -13.37% (of 9,024,291)


src/Microsoft.Android.Sdk.ILLink/ApplyPreserveAttributeBase.cs
```diff
@@ -138,28 +138,12 @@ void TryApplyPreserveAttribute (TypeDefinition type)
...
-
-				attrs.Add (attribute);
-				if (remote_attribute)
-					attributes.RemoveAt (i);
-			}
-
-			return attrs;
+			yield return provider.CustomAttributes.Where (a => {
+				var type = a.Constructor.DeclaringType;
+				return type.Namespace = "Android.Runtime" && type.Name == "PreserveAttribute";
```
radekdoulik, line 145: This should not check the namespace, to keep the current behavior.

marek-safar: I could change the but isn't this the current code https://github.com/xamarin/xamarin-android/blob/9cdaf8414614d9c9c9be79e65867ace258a33683/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs#L50-L55 ?

radekdoulik: There's a default branch, which uses `PreserveAttribute` in any namespace
https://github.com/xamarin/xamarin-android/blob/9cdaf8414614d9c9c9be79e65867ace258a33683/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs#L83-L87

marek-safar: ok, fixed

tests/MSBuildDeviceIntegration/Tests/InstallTests.cs
```diff
@@ -217,7 +217,11 @@ public void InstallWithoutSharedRuntime ()
 				//	"The Shared Runtime should not have been installed.");
 				var directorylist = GetContentFromAllOverrideDirectories (proj.PackageName);
 				StringAssert.Contains ($"{proj.ProjectName}.dll", directorylist, $"{proj.ProjectName}.dll should exist in the .__override__ directory.");
-				StringAssert.Contains ($"System.dll", directorylist, $"System.dll should exist in the .__override__ directory.");
+				if (Builder.UseDotNet)
+					StringAssert.Contains ($"System.Private.CoreLib.dll", directorylist, $"System.Private.CoreLib.dll should exist in the .__override__ directory.");
```
jonathanpeppers, line 221: If this test fails, you might check for a different assembly like `System.Linq.dll`. I think `System.Private.CoreLib.dll` can be in a sub-directory like `armeabi-v7a` or `x86`.

marek-safar: System.Private.CoreLib.dll is probably only reliable marker. I hope `System.Linq.dll` will be gone too next week ;-)

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
```diff
@@ -400,8 +403,9 @@ public class LinkModeFullClass {
 			}, logcatPath, 90), "Linker test app did not run successfully.");
 
 			var logcatOutput = File.ReadAllText (logcatPath);
+			var logcatOutputCut = Substring(Math.Min(logcatOutput.Length - 3000, logcatOutput), logcatOutput.Length);
```
radekdoulik, line 406: This doesn't build. I am not sure what is the intention here, cut the begining or end? And which `[FAIL]` we want to ignore?

marek-safar: Accidental change

jonpryor:
@radekdoulik: please address the merge conflicts.

Additionally, you had mentioned that this change would "implicitly" break apps requiring serialization, e.g. https://github.com/dotnet/runtime/issues/45559

Please write up a description/explanation for the non-obvious relation here.

radekdoulik (approved):

PR #5360, opened 2020-12-02
Pull GetUninitializedObject from RuntimeHelpers type on netcore

marek-safar:


src/Mono.Android/Android.Runtime/ConstructorBuilder.cs
```diff
@@ -6,8 +6,13 @@
 using System.Threading;
 
 namespace Android.Runtime {
-	internal class ConstructorBuilder {
-		static MethodInfo newobject = typeof (System.Runtime.Serialization.FormatterServices).GetMethod ("GetUninitializedObject", BindingFlags.Public | BindingFlags.Static)!;
+	internal static class ConstructorBuilder {
+		static readonly MethodInfo newobject =
+#if NETCOREAPP3_1_OR_GREATER
+			typ##eof (System.Runtime.CompilerServices.RuntimeHelpers).GetMethod ("GetUninitializedObject", BindingFlags.Public | BindingFlags.Static)!;
```
radekdoulik, line 12: `typ##eof` is typo or intentional?

marek-safar: yep, checking if the define works and it does not :-(

Issue #5167, opened 2020-09-30
Review preserve lists


radekdoulik:
Review the preserve lists for illink/net5.

We should also try to use the new illink related [attributes](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis?view=net-5.0) where possible.

radekdoulik:
Also check whether System.ServiceModel is linker safe.

jonathanpeppers:
I think we did some auditing on our preserve lists.

Closing this, we can track this one instead going forward: https://github.com/xamarin/xamarin-android/issues/5652

PR #5363, opened 2020-12-03
[illink] Do not always preserve CharSequence

radekdoulik:
Part of https://github.com/xamarin/xamarin-android/issues/5167

Added dynamic dependency where needed.

Removed API from `BuildReleaseArm64False` test:

    apkdiff before.apk after.apk
    ...
    Type Android.Runtime.CharSequence
      -             Method public static SIGERR ToLocalJniHandle (System.Collections.Generic.IEnumerable`1<char>)
    ...


jonpryor:
What's the `illinkanalyzer` command that gave you the output provided in the PR message?

src/Mono.Android.Export/CallbackCode.cs
```diff
@@ -26,6 +27,7 @@ class DynamicInvokeTypeInfo
 		// NewArray<T>(T[])
 		static readonly MethodInfo jnienv_newarray                        = GetTArrayToIntPtr<int> (JNIEnv.NewArray<int>);
 		static readonly MethodInfo jnienv_getarray                        = GetIntPtrToTArray<int> (JNIEnv.GetArray<int>);
+		[DynamicDependency ("ToLocalJniHandle", "Android.Runtime.CharSequence", "Mono.Android")]
```
jonpryor, line 30: I don't understand; the following line is a delegate creation for `CharSequence.ToLocalJniHandle()`, so why is this `[DynamicDependency]` needed at all?  Related, if this *is* needed, then why isn't *every other* `static readonly` field here similarly attributed?

radekdoulik:
> What's the illinkanalyzer command that gave you the output provided in the PR message?

It is part of the default apkdiff output.

PR #5364, opened 2020-12-03
[illink] Do not preserve ConstructorBuilder

radekdoulik:
Part of https://github.com/xamarin/xamarin-android/issues/5167

The type is not accessed by reflection. It can also be made static,
but that is already part of https://github.com/xamarin/xamarin-android/pull/5360

Removed API from BuildReleaseArm64False test:

    apkdiff before.apk after.apk
    ...
    Type Android.Runtime.ConstructorBuilder
      -             Method public SIGERR .ctor ()
    ...


jonpryor:
What's the `illinkanalyzer` command that gave you the output provided in the PR message?

jonpryor:
Do we need to preserve `Android.Runtime.ConstructorBuilder` in "legacy" Xamarin.Android as well?  I'm not sure why we needed to preserve it in the first place, and xamarin/monodroid@11529d47e does not provide enlightenment.

radekdoulik:
> Do we need to preserve Android.Runtime.ConstructorBuilder in "legacy" Xamarin.Android as well?

I guess we can. I could try to use relevant parts in the "current/legacy" preserve files as well, once I am done with .NET5/6

jonpryor:
Squash-and-merge summary:

```
[One .NET] Don't always preserve Android.Runtime.ConstructorBuilder (#5364)
```

Squash-and-merge body:

```markdown
Context: https://github.com/xamarin/xamarin-android/issues/5167

The `Android.Runtime.ConstructorBuilder` type is not accessed
by reflection, and does not need to be fully preserved by the linker.
```

PR #5370, opened 2020-12-04
[build] Replace Jenkins-based signing step with Azure DevOps step

mjbond-msft:
As part of the transition off Jenkins-based operations, transition to using the Azure DevOps-based signing pipeline. The Azure DevOps solution is functionally equivalent to the Jenkins-based solution.

This change queues the signing job here:
https://dev.azure.com/devdiv/DevDiv/_build?definitionId=13878

Because the PR build does not include signing, a special manual build was launched (and passed) as follows:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4275293&view=results

Which in turn launched the following Azure DevOps signing job
https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=4275526&view=results

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1237,7 +1237,7 @@ stages:
     workspace:
       clean: all
     steps:
-
+      # Signing job is run here: https://dev.azure.com/devdiv/DevDiv/_build?definitionId=13878
       # https://github.com/huserben/TfsExtensions/blob/master/BuildTasks/overview.md
     - task: benjhuser.tfs-extensions-build-tasks.trigger-build-task.TriggerBuild@3
       displayName: 'xamarin vsix codesign - run Azure DevOps job'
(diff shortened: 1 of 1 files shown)


plequere-ms (approved):

pjcollins (approved):
I think this looks good to me. For the commit message, could we also provide a link to the YAML for the signing pipeline that is being queued?

mjbond-msft:
I included the following link to the signing job in the PR description and as a comment on the signing step https://github.com/xamarin/xamarin-android/pull/5370/commits/5e3450c7669ccfc5e151d5df5a9614eae02b4822
https://dev.azure.com/devdiv/DevDiv/_build?definitionId=13878

cadsit (approved):

dellis1972 (approved):

mjbond-msft:
> I think this looks good to me. For the commit message, could we also provide a link to the YAML for the signing pipeline that is being queued?

Here's the link to the signing pipeline YAML file: https://github.com/xamarin/release-scripts/blob/full-esrp-integration/code-signing/src/com/xamarin/codesign/sign-from-github.yaml

PR #5375, opened 2020-12-04
Revert "[Mono.Android] Add [Category] for IDE completion of string props

jonathanpeppers:
This reverts commit d548ec9ca5746e7ef2f4bd5f0ea11bd58b09ae3b.

Decoration of `ApplicationAttribute.Label` with:

    [System.ComponentModel.Category ("@string")]
    public string?                Label                   {get; set;}

Allowed an experimental feature in Visual Studio to offer `@string`
`AndroidResource` completion when changing the `Label` property in C#.

Unfortunately, the presence of this attribute causes the linker to
preserve various dependencies of `System.ComponentModel` in .NET 6.

It appears that this completion feature never shipped in the IDE, so
we can completely remove this. If we ever want to bring the feature
back, we should create a new `internal` attribute to be used instead.

Results of this change:

    > apkdiff before.apk after.apk
    -          56 lib/armeabi-v7a/libxamarin-app.so
    Section size difference
    -           8 .rel.dyn
    -          12 .data
    -       6,144 .bss
    -          77 assemblies/Mono.Android.dll
    -       3,194 assemblies/System.ComponentModel.Primitives.dll *1
    Summary:
    +           0 Other entries 0.00% (of 6,484)
    +           0 Dalvik executables 0.00% (of 317,508)
    -       3,271 Assemblies -0.15% (of 2,139,234)
    -          56 Shared libraries -0.00% (of 20,674,428)
    -       3,468 Package size difference -0.04% (of 9,485,287)

### DIFF
--- a/Documentation/release-notes/5375.md
+++ b/Documentation/release-notes/5375.md
@@ -1,0 +1,25 @@
+### Application size improvements
+
+  * [GitHub PR 5375](https://github.com/xamarin/xamarin-android/pull/5375):
+    Removed decoration of `System.ComponentModel.CategoryAttribute` to
+    improve linker behavior. This reduced the .NET assembly size of a
+    .NET 6 application by 3,271 bytes.
+
+`CategoryAttribute` was removed from properties on types such as
+`Android.App.ActivityAttribute`:
+
+```diff
+-[Category ("@drawable;@mipmap")]
+public string? Icon { get; set; }
+-[Category ("@string")]
+public string? Label { get; set; }
+```
+
+A full list of types changed:
+
+* `Android.App.ActivityAttribute`
+* `Android.App.ApplicationAttribute`
+* `Android.App.InstrumentationAttribute`
+* `Android.App.ServiceAttribute`
+* `Android.Content.BroadcastReceiverAttribute`
+* `Android.Content.ContentProviderAttribute`
--- a/src/Xamarin.Android.NamingCustomAttributes/Android.App/ActivityAttribute.cs
+++ b/src/Xamarin.Android.NamingCustomAttributes/Android.App/ActivityAttribute.cs
@@ -1,5 +1,4 @@
 using System;
-using System.ComponentModel;
 
 using Android.Content.PM;
 using Android.Views;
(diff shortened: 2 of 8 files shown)


jonathanpeppers:
Some conversation about the removal (sorry private link):

https://teams.microsoft.com/l/message/19:01de2e9c80d74444a4545eb4e1ad50e4@thread.skype/1607102592390?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=42121604-a67e-4f9a-9f3f-57c3c6d7ccd3&parentMessageId=1607102592390&teamName=Xamarin%20Platform%20Engineering&channelName=Visual%20Studio%20Extensions&createdTime=1607102592390

radekdoulik (approved):
LGTM. If we need to revive that inb future, it might be enough to let linker remove the attribute during linking.

radekdoulik (changes requested):
OTOH, now I noticed the attribute is public and so check API needs update.

I also noticed it is used on mac/ios, looks like in different context through. So I think it should be fine to remove it in XA. Might it be worth add mention it in the release notes?

jonathanpeppers:
I checked the before/after size on the app from this test, `android.21-arm` + `Release`:

https://github.com/xamarin/xamarin-android/blob/e346f8468e7727e27f2efadf5046c68e9cc5b33c/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs#L335

Not a huge difference:
```
> apkdiff before.apk after.apk
  -          56 lib/armeabi-v7a/libxamarin-app.so
                  Section size difference
    -           8 .rel.dyn
    -          12 .data
    -       6,144 .bss
  -          77 assemblies/Mono.Android.dll
  -       3,194 assemblies/System.ComponentModel.Primitives.dll *1
Summary:
  +           0 Other entries 0.00% (of 6,484)
  +           0 Dalvik executables 0.00% (of 317,508)
  -       3,271 Assemblies -0.15% (of 2,139,234)
  -          56 Shared libraries -0.00% (of 20,674,428)
  -       3,468 Package size difference -0.04% (of 9,485,287)
```
_NOTE: not sure why the shared libraries changed_

dellis1972:
I could have sworn I've seen this feature working in VSForMac... but perhaps not.

jonathanpeppers:
> I could have sworn I've seen this feature working in VSForMac... but perhaps not.

@dellis1972 no, it doesn't seem to work in VS Mac either...

radekdoulik (approved):

PR #5385, opened 2020-12-08
[CI] Migrate to ESRP macOS notarization

pjcollins:
Context: https://github.com/xamarin/release-scripts/blob/ce33e71cfe75f4ad098aa5f66c475e60582ccafb/notarize.ps1

We have received a request to migrate away from our Xamarin Notarization
logic and instead use a new ESRP Notarization script.  To do this we
need to first install the required MicroBuild plugin, and then call the
new `notarize.ps1` script.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1204,48 +1204,48 @@ stages:
       parameters:
         UnsignedPkgPath: $(XA.Unsigned.Pkg)
 
-  - powershell: |
-      # We require at a minimum version 2.8.0 of the azure cli installed for the ESRP signing
-      # First, check for what version is currently installed on the machine
-      & brew link azure-cli
-      $Versions = & az --version 
-      $AzureVersion = $Versions[0]
-      $AzureVersionFields = $AzureVersion.split(" ", [StringSplitOptions]::RemoveEmptyEntries)
-      $CurrentVersion = New-Object -TypeName System.Version -ArgumentList $AzureVersionFields[1]
-      $Version280 = New-Object -TypeName System.Version -ArgumentList "2.8.0"
+    - powershell: |
+        # We require at a minimum version 2.8.0 of the azure cli installed for the ESRP signing
+        # First, check for what version is currently installed on the machine
+        & brew link azure-cli
+        $Versions = & az --version 
+        $AzureVersion = $Versions[0]
+        $AzureVersionFields = $AzureVersion.split(" ", [StringSplitOptions]::RemoveEmptyEntries)
+        $CurrentVersion = New-Object -TypeName System.Version -ArgumentList $AzureVersionFields[1]
+        $Version280 = New-Object -TypeName System.Version -ArgumentList "2.8.0"
 
-      Write-Host "Current Version of azure-cli is $CurrentVersion"
+        Write-Host "Current Version of azure-cli is $CurrentVersion"
 
-      # Check if version 2.8 is higher than the current version.
-      # If it is then we need to do work and download and install the 2.8.0 version of the azure cli in
-      # a really really freaked out non-intuitive way using brew. Why is this so hard to do???
-      # https://stackoverflow.com/questions/62785290/installing-previous-versions-of-a-formula-with-brew-extract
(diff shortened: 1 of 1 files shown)


pjcollins:
@trevors20 @cadsit it looks like calling this new notarization pwsh script succeeded here - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4283121&view=logs&j=b53b0f16-cbff-5f88-33c0-fe0867b30359&t=3c384e6e-7310-5131-4f7c-7de9943cceb5

However notarization validation failed. Does anything stand out to either of you that may be missing in how I'm calling this script?

trevors20:
@pjcollins , I took a look.  You've uncovered a bug with the install signing plugin task where it failed but it did not fail the pipeline for some reason.  I'll log a bug for that.  But, in the meantime, it failed because it looks like the TeamName variable is not set.  Would you be able to set the TeamName variable in your pipeline and try again?

pjcollins:
Thanks for the catch @trevors20, I just kicked off a new test build here https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4283686&view=results

trevors20:
@pjcollins , it looks like it is failing to download the signing nuget package due to unauthorized.  Which is weird because I thought a "yaml" pipeline always had the system.accesstoken available.  But, it looks like it is undefined for some reason:  https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4283686&view=logs&j=b53b0f16-cbff-5f88-33c0-fe0867b30359&t=eca37269-4fd8-5f2b-fa6a-3864999d1b2b&l=100

On the build pipeline, it needs to allow the "OAuth" token to be used by scripts.  I know how to do that through the UI.  But, I'll try and see if I can find the yaml equivalent.  Unless you know already?

trevors20:
@pjcollins , that's interesting.  Looks like in yaml it still needs to be defined by yourself.  From this article: https://stackoverflow.com/questions/52837980/how-to-allow-scripts-to-access-oauth-token-from-yaml-builds

It looks like there are two ways to do this.
1. pipeline level:

variables:
  system_accesstoken: $(System.AccessToken)

2. script level

- powershell: script.ps1
  env:
    SYSTEM_ACCESSTOKEN: $(System.AccessToken)

I'll update the OneNote for the yaml portion.
Since it is in a Task, would you be able to try #1 and see what happens?  The weird thing is that we are calling "tl.getVariable('System.AccessToken')" in typescript in the task.  Technically, it should have been there.  

But, in the output of the Task environment, it says it is "undefined".

pjcollins:
Thanks for the investigation @trevors20, I think the `env:` parameter should work on any task so I've updated the PR to try that. New test run - https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=4284298&view=results

build-tools/automation/azure-pipelines.yaml
```diff
@@ -1197,11 +1198,18 @@ stages:
...
         UnsignedPkgPath: $(XA.Unsigned.Pkg)
 
+    - task: MicroBuildSigningPlugin@3
+      displayName: Install Signing Plugin
+      inputs:
+        signType: real
+        azureSubscription: MicroBuild Signing Task (DevDiv)
+      env:
+        TeamName: XamarinAndroid
+        SYSTEM_ACCESSTOKEN: $(system.accesstoken)
```
trevors20, line 1208: case sensitivity for $(system.accesstoken)?  Might need to do $(System.AccessToken)

pjcollins:
Task level variables didn't seem to work, new test job for commit dca4063: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=4284715&view=results

pjcollins:
It looks like the TeamName variable needs to be set on the stage/job level, and the OAUTH token needs to be passed directly to the task using the env parameter. New test run for commit fc31877: https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=4287122&view=results

pjcollins (commented):
This should be good to go, the new notarization tool looks to be working as expected.

trevors20:
I would have recommended checking this in in the early morning tomorrow so that if anything should happen people would be in the office to take a look.

PR #5392, opened 2020-12-09
Fix d16-8 Build

jonpryor:
"Environmental changes" broke the d16-8 build, with fixes on master.

CP the fixes so d16-8 can build on CI.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1126,3 +1126,28 @@ stages:
   - template: yaml-templates\run-timezoneinfo-tests.yaml
     parameters:
       node_id: 3
+
+  # Check - "Xamarin.Android (Finalize Installers Queue Vsix Signing)"
+  - job: queue_vsix_signing
+    displayName: Queue Vsix Signing
+    dependsOn: notarize_pkg_upload_storage
+    pool: VSEng-MicroBuildVS2019
+    timeoutInMinutes: 90
+    cancelTimeoutInMinutes: 1
+    workspace:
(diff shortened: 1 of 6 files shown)


radekdoulik (commented):
LGTM, it has triggered only 2 checks though, which doesn't seem right?

I will add HomeBrew fix to see whether it would trigger more checks.

radekdoulik:
Looks like the Homebrew fix might not be needed in d16-8 as it is not running `xaprep` near the fix location.

dellis1972 (approved):

build-tools/automation/azure-pipelines.yaml
```diff
@@ -1126,3 +1126,28 @@ stages:
   - template: yaml-templates\run-timezoneinfo-tests.yaml
     parameters:
       node_id: 3
+
+  # Check - "Xamarin.Android (Finalize Installers Queue Vsix Signing)"
+  - job: queue_vsix_signing
```
pjcollins, line 1131: We've moved things around a bit in master, but the Jenkins sign job does exist on d16-8:
https://github.com/xamarin/xamarin-android/blob/d16-8/build-tools/automation/azure-pipelines.yaml#L435

We should replace that job with the changes below.

mjbond-msft: I made the appropriate fixups per https://github.com/xamarin/xamarin-android/pull/5392/commits/cbca46c14eea0ed6ae5814e14db6fd8848046f6f

pjcollins (approved):

dellis1972 (approved):

PR #5394, opened 2020-12-09
[tests] check PerformanceTest for .NET 6

jonathanpeppers:
I cherry-picked some commits to see how close we will get to the exact
same build performance in .NET 6.

For now I triggered every `PerformanceTest` to fail, so I can compare
.NET 6 `.binlog` files against the "legacy" Xamarin.Android ones.


jonathanpeppers:
Don't need this anymore.

PR #5395, opened 2020-12-09
[One .NET] exclude "legacy" MSBuild targets from .NET 6 packages

jonathanpeppers:
We don't need to package `Xamarin.Android.Legacy.targets`, as well as
the following `.targets` that are imported at the bottom of "legacy"
`.csproj` files:

* `Xamarin.Android.Bindings.targets`
* `Xamarin.Android.CSharp.targets`
* `Xamarin.Android.FSharp.targets`
* `Xamarin.Android.VisualBasic.targets`

Let's go ahead and exclude all of these files to be sure that
everything continues to work.


dellis1972 (approved):

pjcollins (approved):
Looks good to me if all the tests are happy.

jonathanpeppers:
Test failures look OK:

* `CheckIncludedAssemblies` should be fixed in #5399 
* `DotNetLibraryAarChanges` needs a 1 second delay, I'll look into that separately.

PR #5397, opened 2020-12-09
[One .NET] proposal for "Single Project"

jonathanpeppers:
One theme for Xamarin in .NET 6 is to simplify the cross-platform
experience between iOS and Android. We would like to provide
cross-platform alternatives to `AndroidManifest.xml` and `Info.plist`.
The eventual goal would be to remove the need for these files for
simple apps and project templates. Developers would only need to
create them when a less widely used feature is needed. We would need
to create item templates for `AndroidManifest.xml` and `Info.plist`
when (and if) they are removed from project templates.

The concepts for a Xamarin "single project" go much beyond this. This
document is a starting point for what needs to go in the .NET Android
and Apple SDKs, while much of the remaining features would be in the
Xamarin.Forms/MAUI MSBuild targets.

This is a spec for the first phase that would introduce a few MSBuild
properties for the application title, id, and version. A
cross-platform `Directory.Build.props` could be:

    <Project>
      <PropertyGroup>
        <ApplicationTitle>Hello!</ApplicationTitle>
        <ApplicationId>com.companyname.hello</ApplicationId>
        <ApplicationVersion>1.0.0</ApplicationVersion>
      </PropertyGroup>
    </Project>

When this is implemented, we decide if we want to move ahead with the
remaining MSBuild properties that would be needed to completely omit
`Info.plist` and `AndroidManifest.xml`.

### DIFF
--- a/Documentation/proposals/OneDotNetSingleProject.md
+++ b/Documentation/proposals/OneDotNetSingleProject.md
@@ -22,14 +22,11 @@ Xamarin.Android and Xamarin.iOS/Mac SDKs:
 * `$(ApplicationVersion)` maps to `android:versionName` or
   `CFBundleVersion`. This is a version string that must be incremented
   for each iOS App Store or TestFlight submission.
-* `$(ApplicationVersionCode)` maps to `android:versionCode` (_Android
+* `$(AndroidVersionCode)` maps to `android:versionCode` (_Android
   only)_. This is unfortunately an integer and must be incremented for
-  each Google Play submission. If omitted, the Android SDK can use
-  `$(ApplicationVersion)` for values up to 128.128.128.128. We can
-  pack most version strings into a 32-bit integer by default.
-* `$(ApplicationDisplayVersion)` maps to `CFBundleShortVersionString`
-  (_iOS only)_. This can default to `$(ApplicationVersion)` when
-  blank.
+  each Google Play submission.
+* `$(AppleShortVersion)` maps to `CFBundleShortVersionString` (_iOS
+  only)_. This can default to `$(ApplicationVersion)` when blank.
 * `$(ApplicationTitle)` maps to `/application/@android:title` or
   `CFBundleDisplayName`
 
@@ -71,8 +68,10 @@ The full list of defaults might be something like:
 ```xml
 <PropertyGroup>
   <ApplicationVersion Condition=" '$(ApplicationVersion)' == '' ">1.0</ApplicationVersion>
-  <!-- $(ApplicationDisplayVersion) is iOS only -->
-  <ApplicationDisplayVersion Condition=" '$(ApplicationDisplayVersion)' == '' ">$(ApplicationVersion)</ApplicationVersion>
+  <!-- Android only -->
+  <AndroidVersionCode Condition=" '$(AndroidVersionCode)' == '' ">1</AndroidVersionCode>
+  <!-- Apple platforms only -->
+  <AppleShortVersion Condition=" '$(AppleShortVersion)' == '' ">$(AppleShortVersion)</ApplicationVersion>
   <AssemblyVersion Condition=" '$(AssemblyVersion)' == '' ">$(ApplicationVersion)</AssemblyVersion>
   <FileVersion Condition=" '$(FileVersion)' == '' ">$(ApplicationVersion)</FileVersion>
 </PropertyGroup>
@@ -90,6 +89,7 @@ The default Android project template would include:
     <ApplicationTitle>@string/application_title</ApplicationTitle>
     <ApplicationId>com.companyname.myapp</ApplicationId>
     <ApplicationVersion>1.0</ApplicationVersion>
+    <AndroidVersionCode>1</AndroidVersionCode>
   </PropertyGroup>
 </Project>
 
@@ -98,9 +98,6 @@ The default Android project template would include:
     <string name="application_title">MyApp</string>
 </resources>
 ```
-
-`$(ApplicationVersionCode)` is computed by a math expression to
-translate versions up to 128.128.128.128 to a 32-bit integer.
 
 Removed from `AndroidManifest.xml` in the project template:
 
@@ -127,8 +124,8 @@ The default iOS project template would include:
 </Project>
 ```
 
-`$(ApplicationDisplayVersion)` can default to `$(ApplicationVersion)`
-when blank.
+`$(AppleShortVersion)` can default to `$(ApplicationVersion)` when
+blank.
 
 Removed from `Info.plist` in the project template:
 
@@ -158,14 +155,14 @@ with:
   <PropertyGroup>
     <ApplicationTitle>Hello!</ApplicationTitle>
     <ApplicationId>com.companyname.hello</ApplicationId>
-    <!-- Increment to 1.0.1, 1.0.2, etc. on each release -->
     <ApplicationVersion>1.0.0</ApplicationVersion>
+    <AndroidVersionCode>1</AndroidVersionCode>
   </PropertyGroup>
 </Project>
 ```
 
-In this project, a developer would only need to increment
-`$(ApplicationVersion)` for each public release.
+In this project, a developer would increment `$(ApplicationVersion)`
+and `$(AndroidVersionCode)` for each public release.
 
 For our long-term vision, we could one day have a single project that
 multi-targets:
@@ -177,6 +174,7 @@ multi-targets:
     <ApplicationTitle>Hello!</ApplicationTitle>
     <ApplicationId>com.companyname.hello</ApplicationId>
     <ApplicationVersion>1.0.0</ApplicationVersion>
+    <AndroidVersionCode>1</AndroidVersionCode>
   </PropertyGroup>
 </Project>
 ```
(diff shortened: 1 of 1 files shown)


Documentation/proposals/OneDotNetMSBuildProperties.md
```diff
@@ -0,0 +1,178 @@
...
+
+`$(ApplicationVersionCode)` is computed by a math expression to
+translate versions up to 128.128.128.128 to a 32-bit integer.
+
+Removed from `AndroidManifest.xml` in the project template:
+
+* `/manifest/@android:versionCode="1"`
+* `/manifest/@android:versionName="1.0"`
+* `/manifest/@package="com.companyname.myapp"`
+* `/application/@android:label="MyApp"`
```
jonathanpeppers, line 81: Are we OK that the removal of these will make this property page blank? (until there is IDE support)

image

Documentation/proposals/OneDotNetMSBuildProperties.md
```diff
@@ -0,0 +1,178 @@
...
+* `$(ApplicationId)` maps to `/manifest/@package` and
+  `CFBundleIdentifier`
+* `$(ApplicationVersion)` maps to `android:versionName` or
+  `CFBundleVersion`. This is a version string that must be incremented
+  for each iOS App Store or TestFlight submission.
+* `$(ApplicationVersionCode)` maps to `android:versionCode` (_Android
+  only)_. This is unfortunately an integer and must be incremented for
+  each Google Play submission. If omitted, the Android SDK can use
+  `$(ApplicationVersion)` for values up to 128.128.128.128. We can
+  pack most version strings into a 32-bit integer by default.
```
jonathanpeppers, line 24: Do we need to worry if users put a build number as the 4th number, where they could easily have an `$(ApplicationVersion)` like 1.0.1.1234?

Documentation/proposals/OneDotNetMSBuildProperties.md
```diff
@@ -0,0 +1,178 @@
+# One .NET: New MSBuild Properties
```
jonpryor, line 1: Should this title include "Single App" or "Unified" or some other noun to suggest what's special about *these* properties vs. all the other properties we already support?

Documentation/proposals/OneDotNetMSBuildProperties.md
jonpryor: Do we know what the time delay would be between SDK support for these properties *and release* vs. IDE support?  If the IDE supports the new semantics before .NET 6 P1, it'll be fine.  If the IDE won't catch up until P2, things could be "weird".

Documentation/proposals/OneDotNetMSBuildProperties.md
dellis1972: Personally I would not allow the reuse of `ApplicationVersion` for `ApplicationVersionCode`. Users should provide both, especially as google play has specific restrictions around these fields. We don't want to get into a situation where a user cannot upload an update because our version scheme changed (this has happened before). 

That said is this an opportunity to implement a feature where we pick up the `ApplicationVersion` from the assembly version...

Documentation/proposals/OneDotNetMSBuildProperties.md
```diff
@@ -0,0 +1,178 @@
...
+beyond `$(ApplicationTitle)`, `$(ApplicationId)`,
+`$(ApplicationVersion)`, `$(ApplicationVersionCode)`, and
+`$(ApplicationDisplayVersion)`.
+
+This is a list of additional properties that cover most of the
+property pages in Visual Studio:
+
+* `$(iOSMinimumOSVersion)` _iOS only_
+* `$(iOSLaunchScreen)` _iOS only_
+* `$(iOSDeviceFamily)` _iOS only_ - `iPhone`, `iPad`, or `Universal`
```
rolfbjarne, line 153: We'll need `tvOS*`, `MacCatalyst*` and `macOS*` variants for some of these as well.

rolfbjarne (commented):
Does this apply to other platforms as well (UWP, Blazor, etc)? I'm not sure this is the right repo for the spec in that case.

Documentation/proposals/OneDotNetMSBuildProperties.md
jonathanpeppers: My initial thought was to require `$(ApplicationVersionCode)` to be set, and not try to pack versions like `1.0.0` into an `int`. @jonpryor was thinking we could make the properties identical for iOS/Android if we did it this way, but maybe we shouldn't do it?

> pick up the `ApplicationVersion` from the assembly version

There are 3 other properties that we might consider using:
* `$(Version)` used for NuGet package version
* `$(AssemblyVersion)` and `$(FileVersion)` into the generated `AssemblyInfo.cs`.

Maybe `$(ApplicationVersion)` should also set `$(FileVersion)` and `$(AssemblyVersion)` when they are blank? That probably makes sense.

I don't think we need to use `$(Version)` as you don't pack an app into a NuGet package. Maybe someone would try that, though?

Documentation/proposals/OneDotNetMSBuildProperties.md
jonathanpeppers: @Redth can we use the term "Single Project" or probably not?

Documentation/proposals/OneDotNetMSBuildProperties.md
```diff
@@ -0,0 +1,178 @@
...
+  for each iOS App Store or TestFlight submission.
+* `$(ApplicationVersionCode)` maps to `android:versionCode` (_Android
+  only)_. This is unfortunately an integer and must be incremented for
+  each Google Play submission. If omitted, the Android SDK can use
+  `$(ApplicationVersion)` for values up to 128.128.128.128. We can
+  pack most version strings into a 32-bit integer by default.
+* `$(ApplicationDisplayVersion)` maps to `CFBundleShortVersionString`
+  (_iOS only)_. This can default to `$(ApplicationVersion)` when
+  blank.
+* `$(ApplicationTitle)` maps to `/application/@android:title` or
```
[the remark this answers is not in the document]
jonathanpeppers: On Android, you could easily put a string resource name for `$(ApplicationTitle)`. We should probably put this in the Android project template?

For iOS, I think you would need to set [`LSHasLocalizedDisplayName](https://developer.apple.com/library/archive/documentation/General/Conceptual/MOSXAppProgrammingGuide/BuildTimeConfiguration/BuildTimeConfiguration.html) in the `Info.plist` and omit `$(ApplicationTitle)`.

I will update the doc to include this info, thanks!

Documentation/proposals/OneDotNetMSBuildProperties.md
jonathanpeppers: @Redth do we need to worry about Windows (UWP/WPF/etc.)? I will need to update this to consider tvOS, Catalyst, and macOS.

Documentation/proposals/OneDotNetMSBuildProperties.md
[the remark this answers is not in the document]
AmrAlSayed0: Currently, on Android, we set the app name by setting ```android:label``` on the main launcher [Activity](https://developer.android.com/guide/topics/manifest/activity-element#label) or the [Application](https://developer.android.com/guide/topics/manifest/application-element#label). We could also use C# classes with attributes to set those. Usually, it is set to a resource like so ```android:label="@string/app_name"```. ```@string/app_name``` is included in the template.

On iOS, we have localization strings like so
image
and inside, this
image

Documentation/proposals/OneDotNetMSBuildProperties.md
jonathanpeppers: I renamed some of properties to be `$(Apple*)` instead of `$(iOS*)`, Most of this is the "future" section, so we don't have to come up with the final names yet.

Documentation/proposals/OneDotNetMSBuildProperties.md
jonathanpeppers: I don't think we know yet.

We could always implement the new MSBuild properties, but only add them to the project templates when the IDE-side lands.

Documentation/proposals/OneDotNetMSBuildProperties.md
dellis1972: I think we should keep it separate and not pack. Remember the versionCode in the manifest also has some post processing via https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/building-apps/build-properties.md#androidversioncodepattern . 

The default pattern is `{abi}{versionCode:D5}` so we use the number in the manifest (e.g 1) and then combine that with other bits to produce a code based on the google recommendations https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/building-apps/~/android/deploy-test/building-apps/abi-specific-apks.md

jonathanpeppers: I changed this to be `$(AndroidVersionCode)`. We wouldn't try to pack the `$(ApplicationVersion)` into an integer.

Documentation/proposals/OneDotNetSingleProject.md
```diff
@@ -0,0 +1,243 @@
...
+* `$(ApplicationId)` maps to `/manifest/@package` and
+  `CFBundleIdentifier`
+* `$(ApplicationVersion)` maps to `android:versionName` or
+  `CFBundleVersion`. This is a version string that must be incremented
+  for each iOS App Store or TestFlight submission.
+* `$(AndroidVersionCode)` maps to `android:versionCode` (_Android
+  only)_. This is unfortunately an integer and must be incremented for
+  each Google Play submission.
+* `$(AppleShortVersion)` maps to `CFBundleShortVersionString` (_iOS
+  only)_. This can default to `$(ApplicationVersion)` when blank.
```
jonathanpeppers, line 29: I renamed this in the latest iteration.

dellis1972 (approved):

rolfbjarne (approved):

PR #5399, opened 2020-12-10
[illink] Update CheckIncludedAssemblies test

radekdoulik:
Two more assemblies are linked away, that is probably result of multiple
linker related fixes going in together yesterday.


dellis1972 (approved):

jonathanpeppers (approved):
Do we know how we keep missing these?

I checked this code and it seems correct:

https://github.com/xamarin/xamarin-android/blob/7b51fd1b45f7f315df402ba333e141a49c8a3ca7/build-tools/xaprepare/xaprepare/Steps/Step_DetermineAzurePipelinesTestJobs.cs#L88-L90

radekdoulik:
> Do we know how we keep missing these?

I think it is result of multiple illink PR's going in together, eliminating some of the API dependencies, which were not removed by one PR, but as the effect of multiple ones. I don't know what happened exactly as that would require quite some time to inspect.

The PRs run the CheckIncludedAssemblies test, so that should be fine.

PR #5406, opened 2020-12-11
[Xamarin.ProjectTools] log ExitCode when MSBuild exits

jonathanpeppers:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4291650&view=ms.vss-test-web.build-test-results-tab&runId=17600706&resultId=100198&paneView=attachments

We have occasionally seen failures under `dotnet build` such as:

    Project should have built.
    Expected: True
    But was:  False

But then the diagnostic MSBuild log says:

    Build succeeded.
        0 Warning(s)
        0 Error(s)

Looking at the logic in `Builder.BuildInternal()`, it seems like this
means the `ExitCode` was non-zero. There was no `Build Timed Out!` log
message either.

We should log the exit code, which might help diagnose the issue.


dellis1972 (approved):

jonathanpeppers:
No test failed, but I think we can merge this to help with a future PR.

PR #5414, opened 2020-12-14
Bump to apksigner from platform-tools 30.0.5

jonathanpeppers:
Context: https://android.googlesource.com/platform/tools/apksig/+/refs/tags/platform-tools-30.0.5
Context: https://developer.android.com/studio/releases/platform-tools#3005_november_2020

Bump to platform/tools/apksig/platform-tools-30.0.5@96e0c65e

As part of this, the `apksig` library has taken a dependency on:

* https://mvnrepository.com/artifact/org.conscrypt/conscrypt-openjdk/2.5.1

We will also need to exclude some files in the `META-INF` directory,
since this dependency is signed.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,7 +5,7 @@
 [submodule "external/apksig"]
     path = external/apksig
     url = https://android.googlesource.com/platform/tools/apksig
-    branch = platform-tools-30.0.3
+    branch = platform-tools-30.0.5
 [submodule "external/debugger-libs"]
     path = external/debugger-libs
     url = https://github.com/mono/debugger-libs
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
This bump seems to produce broken `.apk` files:
```
> adb install -r .\samples\HelloWorld\bin\Debug\com.xamarin.android.helloworld-Signed.apk
Performing Incremental Install
Serving...
Unknown command: install-incremental
Install command complete in 20 ms
```
Using an API 29 emulator ^^

jonathanpeppers:
After more research, it seems like the `apksigner` from platform-tools 30.0.5 requires changes to `adb`. You must have the `adb` from `platform-tools` 30.0.5...

It's also worth noting is that platform-tools 30.0.5 is where the breakage occurs, there were no changes between 30.0.0 and 30.0.4:
```
> git diff platform-tools-30.0.0...platform-tools-30.0.4
```
Empty ^^

jonathanpeppers:
We'll revisit this when the next version of `platform-tools` comes out.

Currently, we are using the `apksigner` from July and not the November one:

https://developer.android.com/studio/releases/platform-tools#3004_july_2020

So we are only 1 release behind.

PR #5426, opened 2020-12-15
[build] fix Inputs/Outputs for monodroid.csproj

jonathanpeppers:
On macOS, I noticed this happening on every build:

    $ msbuild src/monodroid -clp:performancesummary
    ...
     2934 ms  _BuildHostRuntimes
    41768 ms  _BuildAndroidRuntimes

The original cause was:

    Building target "_BuildAndroidRuntimes" completely.
    Output file "obj/Debug//armeabi-v7a-Debug/libxamarin-app.so" does not exist.

I updated the `Outputs` to use the correct file path:

```diff
-@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Debug\libxamarin-app.so');
-@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Release\libxamarin-app.so')
+@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\Debug\libxamarin-app.so');
+@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\Release\libxamarin-app.so')
```

Next, I was still hitting the problem, but instead hit:

    Building target "_BuildAndroidRuntimes" completely.
    Input file "jni/*.cc" does not exist.

You can't actually use wildcards in `Inputs`? I made a new
`_FindMonoDroidSources` MSBuild target that created a new
`@(_MonoDroidSources)` item group. I used this in place of wildcards.

The `_BuildHostRuntimes` MSBuild target had a typo which caused:

    Building target "_BuildHostRuntimes" completely.
    Input file "host-Darwin-Debug/CMakeCache.txt" does not exist.

The `Inputs` needed to change:

```diff
-$(IntermediatePath)
+$(IntermediateOutputPath)
```

It also needed to use `@(_MonoDroidSources)` as `Inputs`.

Now things seem to work correctly:

    Skipping target "_BuildAndroidRuntimes" because all outputs are up-to-date with respect to the input files.
    Skipping target "_BuildHostRuntimes" because all outputs are up-to-date with respect to the input files.

This saves ~45s off of incremental builds for me on macOS. The savings
on Windows would not be as high, as it does not run the "checked"
builds at all.

### DIFF
--- a/src/monodroid/monodroid.targets
+++ b/src/monodroid/monodroid.targets
@@ -118,11 +118,24 @@
     </ItemGroup>
   </Target>
 
+  <Target Name="_ConfigureRuntimesInputs">
+    <ItemGroup>
+      <_ConfigureRuntimesInputs  Include="CMakeLists.txt" />
+      <_ConfigureRuntimesInputs  Include="..\..\build-tools\scripts\cmake-common.props" />
+      <_ConfigureRuntimesInputs  Include="..\..\build-tools\scripts\Ndk.targets" />
+      <_ConfigureRuntimesOutputs Include="@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Debug\CMakeCache.txt')" />
+      <_ConfigureRuntimesOutputs Include="@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Release\CMakeCache.txt')" />
+      <_ConfigureRuntimesOutputs Include="@(_HostRuntime->'$(IntermediateOutputPath)%(OutputDirectory)-Debug\CMakeCache.txt')" />
+      <_ConfigureRuntimesOutputs Include="@(_HostRuntime->'$(IntermediateOutputPath)%(OutputDirectory)-Release\CMakeCache.txt')" />
+    </ItemGroup>
+  </Target>
+
   <Target Name="_ConfigureRuntimes"
-      DependsOnTargets="_PrepareConfigureRuntimeCommands"
-      Inputs="CMakeLists.txt;..\..\build-tools\scripts\cmake-common.props;..\..\build-tools\scripts\Ndk.targets"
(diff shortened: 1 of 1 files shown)


src/monodroid/monodroid.targets
```diff
@@ -125,9 +125,16 @@
...
+    </ItemGroup>
+  </Target>
+
   <Target Name="_BuildAndroidRuntimes"
-      Inputs="@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Debug\CMakeCache.txt');@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Release\CMakeCache.txt');jni\*.cc;jni\*.h;jni\*.hh;jni\**\*.c;;..\..\build-tools\scripts\Ndk.targets"
-      Outputs="@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android.debug.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android-checked+asan.debug.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android-checked+ubsan.debug.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android.release.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android-checked+asan.release.so');@(AndroidSupportedTargetJitAbi->'$(OutputPath)\%(Identity)\libmono-android-checked+ubsan.release.so');@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Debug\libxamarin-app.so');@(AndroidSupportedTargetJitAbi->'$(IntermediateOutputPath)\%(Identity)-Release\libxamarin-app.so')">
```
jonpryor, line 130: The line width here -- 789 columns! -- makes me think we should pull this out into a new item group, ditto the `Inputs`.

jonathanpeppers: This target is using batching at the target level...

I don't think we can actually do this or this batching won't work:

https://github.com/xamarin/xamarin-android/blob/087c7f43a1a60a6c92934d0a57c902e381877df3/src/monodroid/monodroid.targets#L133

If we created `@(_BuildAndroidRuntimesInputs)` and `@(_BuildAndroidRuntimesOutput)`, then there would not be a way for `%(Identity)` to be used.

I don't know how to clean this up without rewriting... I didn't really want to go that far.

jonathanpeppers: Nevermind, I think I just got it to work.

Looking at the target, I thought it was using [partial builds](https://github.com/xamarin/xamarin-android/blob/master/Documentation/guides/MSBuildBestPractices.md#building-partially), but it isn't. These targets don't really need to build partially, as if you change a `.cc` file all of the architectures have to rebuild anyway.

PR #5434, opened 2020-12-16
Remove brew/openssl workaround

radekdoulik:
The homebrew/openssl related workaround/fix started failing recently with:

    Error: No available formula or cask with the name "openssl@1.0.2t".

When running:

    brew uninstall openssl@1.0.2t && rm -rf /usr/local/etc/openssl && rm -rf /usr/local/etc/openssl@1.1

@pjcollins thinks the original issue was fixed meanwhile, so we can remove it,
instead of fixing it by forcing the uninstall.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -405,7 +405,7 @@ stages:
       parameters:
         configuration: $(ApkTestConfiguration)
 
-    - script: 'brew uninstall openssl@1.0.2t && rm -rf /usr/local/etc/openssl && rm -rf /usr/local/etc/openssl@1.1'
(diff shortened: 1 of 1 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -405,7 +405,7 @@ stages:
       parameters:
         configuration: $(ApkTestConfiguration)
 
-    - script: 'brew uninstall openssl@1.0.2t && rm -rf /usr/local/etc/openssl && rm -rf /usr/local/etc/openssl@1.1'
+    - script: 'brew uninstall -f openssl@1.0.2t && rm -rf /usr/local/etc/openssl && rm -rf /usr/local/etc/openssl@1.1'
```
pjcollins, line 408: I believe the issue that we needed this workaround for was fixed last week, I think we can remove this line entirely now instead.

dellis1972 (approved):
Looks ok providing it works :)

Issue #5400, opened 2020-12-10
[illink] Try to avoid using SLE in JniRuntime/ManagedPeer


radekdoulik:
Possible savings, measured by not calling ManagedPeer.Init (), which leads to calling static ctor with registration of methods using SLE and that is pulling a lot of code/API in during linking:
```
Size difference in bytes ([*1] apk1 only, [*2] apk2 only):
  -          27 assemblies/Mono.Android.dll
  -          30 assemblies/System.Collections.Concurrent.dll
  -         989 assemblies/System.Linq.dll
  -       2,930 assemblies/Java.Interop.dll
  -       4,514 assemblies/System.Collections.dll *1
  -       4,874 assemblies/System.ObjectModel.dll *1
  -       7,987 assemblies/System.Private.CoreLib.dll
  -     115,284 assemblies/System.Linq.Expressions.dll *1
Summary:
  -     136,635 Assemblies -14.87% (of 918,689)
  -     139,506 Package size difference -1.79% (of 7,780,819)
```

I think we don't use `com.xamarin.java_interop.ManagedPeer` anywhere (yet?), so we hopefully don't need to register its methods.

I will look into that and either try to remove it or find out whether it would be at least possible to replace SLE usage.

radekdoulik:
Turned out the Java.Interop tests are using ManagedPeer, so we ended with different solution. We add abstract `ActivatePeer` method to `JniValueManager` and have different implementations for desktop `MonoRuntimeValueManager` and mobile `AndroidValueManager`. The mobile is using existing code from XA's `Java.Interop.TypeManager`, which uses `System.Reflection.Emit`.

The SRE code is a lot more linker friendly than the SLE code and results in substantial size savings.

PR #5444, opened 2020-12-18
 [illink] Don't initialize managed peer

radekdoulik:
Context: #5400
Context: xamarin/java.interop#761

Generate `XAConfig.props` file in Java.Interop to define `XA_JI_EXCLUDE`
compilation constant.

The generated file contains properties set for Java.Interop by XA. So
far it is only setting one property, and thus doesn't necessarily be
generated. I still wanted to be written by `xaprep`, plus it might be
handy in future for other configuration.

The currently set `JavaInteropDefineConstants` property is used
to disable `ManagedPeer` initialization.

We don't really use `ManagedPeer` today and it is duplicating
functionality of `ConstructorBuilder` class. The unwanted effect of
having it in JI is that it is pulling in `System.Linq.Expressions`, which
greatly increase assemblies size after linking.

The size reduction for simple XA app is cca 135kbytes of compressed
size.

      -          27 assemblies/Mono.Android.dll
      -          30 assemblies/System.Collections.Concurrent.dll
      -         989 assemblies/System.Linq.dll
      -       2,930 assemblies/Java.Interop.dll
      -       4,514 assemblies/System.Collections.dll *1
      -       4,874 assemblies/System.ObjectModel.dll *1
      -       7,987 assemblies/System.Private.CoreLib.dll
      -     115,284 assemblies/System.Linq.Expressions.dll *1
    Summary:
      -     136,635 Assemblies -14.87% (of 918,689)
      -     139,506 Package size difference -1.79% (of 7,780,819)

### DIFF
--- a/build-tools/xaprepare/xaprepare/Application/GeneratedJavaInteropConfigPropsFile.cs
+++ b/build-tools/xaprepare/xaprepare/Application/GeneratedJavaInteropConfigPropsFile.cs
@@ -6,7 +6,7 @@ namespace Xamarin.Android.Prepare
 	{
 		const string Content = @"<?xml version=""1.0"" encoding=""utf-8""?>
 <!-- This is a GENERATED FILE -->
-<!-- See build-tools/xaprepare/xaprepare/Application/GeneratedJavaInteropConfigPropsFile.cs -->
+<!-- See https://github.com/xamarin/xamarin-android/tree/master/build-tools/xaprepare/xaprepare/Application/GeneratedJavaInteropConfigPropsFile.cs -->
 <Project xmlns=""http://schemas.microsoft.com/developer/msbuild/2003"">
   <PropertyGroup>
     <JavaInteropDefineConstants>XA_JI_EXCLUDE</JavaInteropDefineConstants>
--- a/build-tools/xaprepare/xaprepare/Application/GeneratedJavaInteropConfigPropsFile.cs
+++ b/build-tools/xaprepare/xaprepare/Application/GeneratedJavaInteropConfigPropsFile.cs
@@ -20,15 +20,13 @@ namespace Xamarin.Android.Prepare
 
 		public override void Generate (Context context)
 		{
-			string dir = Path.GetDirectoryName (OutputPath);
+			var dir = Path.GetDirectoryName (OutputPath);
 			if (!Directory.Exists (dir))
 				Directory.CreateDirectory (dir);
 
-			using (var fs = File.Open (OutputPath, FileMode.Create)) {
-				using (var sw = new StreamWriter (fs)) {
-					sw.Write (Content);
-				}
-			}
+			using var fs = File.Open (OutputPath, FileMode.Create);
+			using var sw = new StreamWriter (fs);
+			sw.Write (Content);
 		}
 	}
 }
--- a/build-tools/xaprepare/xaprepare/Application/GeneratedJavaInteropConfigPropsFile.cs
+++ b/build-tools/xaprepare/xaprepare/Application/GeneratedJavaInteropConfigPropsFile.cs
@@ -1,7 +1,4 @@
-using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 
 namespace Xamarin.Android.Prepare
 {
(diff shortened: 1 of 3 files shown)


grendello (approved):

jonpryor:
Current `Mono.Android-Tests` on-device `.apk` execution fails:

```
01-15 11:35:33.224  3468  3532 E mono    : Java.Lang.LinkageError: No implementation found for void com.xamarin.java_interop.ManagedPeer.construct(java.lang.Object, java.lang.String, java.lang.String, java.lang.Object[]) (tried Java_com_xamarin_java_1interop_ManagedPeer_construct and Java_com_xamarin_java_1interop_ManagedPeer_construct__Ljava_lang_Object_2Ljava_lang_String_2Ljava_lang_String_2_3Ljava_lang_Object_2)
```

This happens because commit 130905e1612f1c3f41b6b233056cae2753270630 added the Java.Interop unit tests to the Mono.Android-Tests app, and the Java.Interop unit tests *do* use `ManagedPeer`.

There are thus two paths forward:

1. "Partially revert" 130905e1612f1c3f41b6b233056cae2753270630 and *stop* running at least *some* of the Java.Interop unit tests under Xamarin.Android, or
2. Re-introduce `ManagedPeer`, and find a different way of removing the `System.Linq.Expressions.dll` usage.

I'm of course partial to (2).

How do we remove the `System.Linq.Expressions.dll` usage? *An* idea is to:

1. Add a `JniRuntime.JniValueManager.ActivatePeer()` method
2. Update `ManagedPeer.Construct()` to call `ActivatePeer()`
3. "Appropriate glue" to make it all work.

`JniRuntime.JniValueManager.ActivatePeer()`, in [`JniRuntime.JniValueManager`](https://github.com/xamarin/java.interop/blob/a8f68e562827e9ae6b2e12edf7445b1a3112d6fd/src/Java.Interop/Java.Interop/JniRuntime.JniValueManager.cs) could resemble:

```csharp
namespace Java.Interop {
    partial class JniRuntime {
        partial abstract class JniValueManager {
            public abstract void ActivatePeer (JniObjectReference reference, ConstructorInfo peerConstructor, object[]? constructorArgumentValues);
        }
    }
}
```

Within the Java.Interop repo, [`MonoRuntimeValueManager.cs`](https://github.com/xamarin/java.interop/blob/a8f68e562827e9ae6b2e12edf7445b1a3112d6fd/src/Java.Runtime.Environment/Java.Interop/MonoRuntimeValueManager.cs) would need to be updated to override `JniRuntime.JniValueManager.ActivatePeer()`, and would contain what is currently in [`ManagedPeer.Construct()` lines 107-120](https://github.com/xamarin/java.interop/blob/a8f68e562827e9ae6b2e12edf7445b1a3112d6fd/src/Java.Interop/Java.Interop/ManagedPeer.cs#L107-L120). (Note: LINQ use should be removed from lines 96-98.)

[`AndroidValueManager`](https://github.com/xamarin/xamarin-android/blob/49e570d8c93398643a84f5da251752e2bd8ccb0f/src/Mono.Android/Android.Runtime/AndroidRuntime.cs#L428) will need to be similarly updated, but would instead implement `JniValueManager.ActivatePeer()` by using [`TypeManager.n_Activate()`](https://github.com/xamarin/xamarin-android/blob/49e570d8c93398643a84f5da251752e2bd8ccb0f/src/Mono.Android/Java.Interop/TypeManager.cs#L169-L179).

Once `JniValueManager.ActivatePeer()` exists, `ManagedPeer.Construct()` can be updated to *remove* [`ManagedPeer.Construct()` lines 107-120](https://github.com/xamarin/java.interop/blob/a8f68e562827e9ae6b2e12edf7445b1a3112d6fd/src/Java.Interop/Java.Interop/ManagedPeer.cs#L107-L120), thus removing the `System.Linq.Expressions` usage from that type, and the [`JreRuntime` constructor can remove the `#if XA_JI_EXCLUDE` block](https://github.com/xamarin/java.interop/blob/a8f68e562827e9ae6b2e12edf7445b1a3112d6fd/src/Java.Interop/Java.Interop/JniRuntime.cs#L251).  Lines 107-120 would be replaced with:

```csharp
JniEnvironment.Runtime.ValueManager.ActivatePeer (new JniObjectReference (n_self), ctor, pvalues);
```

This should allow `ManagedPeer` to be *usable* on Xamarin.Android, *without* pulling in the large System.Linq.Expressions dependency.

radekdoulik:
Superseded by https://github.com/xamarin/xamarin-android/pull/5530

PR #5450, opened 2020-12-28
[illink] Do not preserve GeneratedEnumAttribute

radekdoulik:
Context: https://github.com/xamarin/xamarin-android/issues/5167

The attribute itself doesn't need to be preserved.

I didn't find any usage of the attribute instances during runtime, so
this change removes them during linking.

apk size comparison, BuildReleaseArm64False test:

    > apkdiff -f -e dll$ before.apk after.apk
      -          39 assemblies/Mono.Android.dll
        +         132 Resource Android.ILLink.ILLink.LinkAttributes.xml
        -             Type Android.Runtime.GeneratedEnumAttribute
    Summary:
      -          39 Assemblies -0.01% (of 749,078)

Note that the attribute instances removal doesn't work yet, we need
net6 illink for that.

### DIFF
--- a/src/Microsoft.Android.Sdk.ILLink/PreserveLists/Mono.Android.xml
+++ b/src/Microsoft.Android.Sdk.ILLink/PreserveLists/Mono.Android.xml
@@ -3,7 +3,6 @@
 	<assembly fullname="Mono.Android">
 		<type fullname="Android.Runtime.AnnotationAttribute" />
 		<type fullname="Android.Runtime.GeneratedDummyHost" />
-		<type fullname="Android.Runtime.GeneratedEnumAttribute" />
 		<type fullname="Android.Runtime.IJavaObject" />
 		<type fullname="Android.Runtime.InputStreamAdapter" />
 		<type fullname="Android.Runtime.InputStreamInvoker" />
--- a/src/Mono.Android/ILLink/ILLink.LinkAttributes.xml
+++ b/src/Mono.Android/ILLink/ILLink.LinkAttributes.xml
@@ -6,5 +6,8 @@
     <type fullname="Android.Runtime.IntDefinitionAttribute">
       <attribute internal="RemoveAttributeInstances" />
     </type>
+    <type fullname="Android.Runtime.GeneratedEnumAttribute">
+      <attribute internal="RemoveAttributeInstances" />
+    </type>
   </assembly>
 </linker>


radekdoulik:
@jpobst , is it OK to remove the attribute instances during linking? I tried to dig the history and this commit says they are not used yet. Is it still the case? https://github.com/xamarin/monodroid/commit/60961175f5aeb8503d6930f727eab62503b83b65

jonpryor:
@radekdoulik: what do you mean by "attribute instances"? I don't understand the question.

Generated code *does* use `[GeneratedEnum]`:

```bash
% grep -r '\[.*\.GeneratedEnum\]' src/Mono.Android/obj/Debug/monoandroid10/android-30 | head -10
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Views.Accessibility.AccessibilityEvent.cs:		public static unsafe string? EventTypeToString ([global::Android.Runtime.GeneratedEnum] Android.Views.Accessibility.EventTypes eventType)
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Views.Accessibility.AccessibilityEvent.cs:		[return:global::Android.Runtime.GeneratedEnum]
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Views.Accessibility.AccessibilityEvent.cs:		public static unsafe Android.Views.Accessibility.AccessibilityEvent? Obtain ([global::Android.Runtime.GeneratedEnum] Android.Views.Accessibility.EventTypes eventType)
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Views.Accessibility.AccessibilityEvent.cs:		public unsafe void SetAction ([global::Android.Runtime.GeneratedEnum] Android.AccessibilityServices.GlobalAction action)
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Views.Accessibility.AccessibilityEvent.cs:		public unsafe void WriteToParcel (Android.OS.Parcel? parcel, [global::Android.Runtime.GeneratedEnum] Android.OS.ParcelableWriteFlags flags)
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Text.Method.MetaKeyKeyListener.cs:		public static unsafe void ClearMetaKeyState (Android.Text.IEditable? content, [global::Android.Runtime.GeneratedEnum] Android.Views.MetaKeyStates states)
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Text.Method.MetaKeyKeyListener.cs:		public virtual unsafe void ClearMetaKeyState (Android.Views.View? view, Android.Text.IEditable? content, [global::Android.Runtime.GeneratedEnum] Android.Views.MetaKeyStates states)
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Text.Method.MetaKeyKeyListener.cs:		public virtual unsafe long ClearMetaKeyState ([global::Android.Runtime.GeneratedEnum] Android.Views.MetaKeyStates state, [global::Android.Runtime.GeneratedEnum] Android.Text.Method.MetaStates which)
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Text.Method.MetaKeyKeyListener.cs:		public static unsafe int GetMetaState (Java.Lang.ICharSequence? text, [global::Android.Runtime.GeneratedEnum] Android.Text.Method.MetaStates meta)
src/Mono.Android/obj/Debug/monoandroid10/android-30/mcw/Android.Text.Method.MetaKeyKeyListener.cs:		public static int GetMetaState (string? text, [global::Android.Runtime.GeneratedEnum] Android.Text.Method.MetaStates meta)
```

I think it should be fine to *remove* the `[GeneratedEnum]` attribute during `.apk` linking.

radekdoulik:
> @radekdoulik: what do you mean by "attribute instances"? I don't understand the question.

I mean the instances in the assemblies like: 
```
  .method public hidebysig newslot virtual final
          instance void  Require(valuetype Org.XmlPull.V1.XmlPullParserNode 'type',
                                 string namespace,
                                 string name) cil managed
  {
    .param [1]
    .custom instance void Android.Runtime.GeneratedEnumAttribute::.ctor() = ( 01 00 00 00 )
    // Code size       161 (0xa1)
    .maxstack  4
    .locals init (native int V_0,
             native int V_1,
             valuetype Android.Runtime.JValue* V_2,
             bool V_3)
...
```
And whether they are used during runtime, which I think are not. The above mentioned commit message says they are not used yet, so I wanted to be sure we didn't start using them later. I didn't find any place, so hopefully that's the case.

jonpryor:
`[GeneratedEnumAttribute]` is *not* used at *run-time*.  It is used at [*binding* time](https://github.com/xamarin/java.interop/blob/3894cd76f71f618949c8542f0edd95762e22881f/tools/generator/Java.Interop.Tools.Generator.Importers/CecilApiImporter.cs#L240-L241).

It should be safe to remove them during linking.  They're not needed for JCW creation.

PR #5458, opened 2021-01-04
[One .NET] new MSBuild properties for "single project"

jonathanpeppers:
Implementation of:

https://github.com/xamarin/xamarin-android/blob/051cad7972033b1b9c8b9fbea3be1b90c1c23fa9/Documentation/proposals/OneDotNetSingleProject.md

This adds the ability for a .NET 6 Android project to look like:

      <PropertyGroup>
        <ApplicationTitle>@string/application_title</ApplicationTitle>
        <ApplicationId>com.companyname.myapp</ApplicationId>
        <ApplicationVersion>1.0</ApplicationVersion>
        <AndroidVersionCode>1</AndroidVersionCode>
      </PropertyGroup>
    </Project>

Where iOS can eventually use the same names:

      <PropertyGroup>
        <TargetFramework>net6.0-ios</TargetFramework>
        <ApplicationTitle>MyApp</ApplicationTitle>
        <ApplicationId>com.companyname.myapp</ApplicationId>
        <ApplicationVersion>1.0</ApplicationVersion>
      </PropertyGroup>
    </Project>

To go along with this:

* Added `SingleProjectTest`
* Updated documentation

Some minor cleanup:

* Removed many calls to the `<GetAndroidPackageName/>` MSBuild task,
  so we can rely on a `_GetAndroidPackageName` MSBuild *target*
  instead. This way we have a single place where `$(_AndroidPackage)`
  is set.
* Removed a `Could not determine package name.` error message that did
  not seem possible. xamarin-android-tools would throw
  `ArgumentException`.
* Reworked the `<GetAndroidPackageName/>` MSBuild task, so the logic
  was a bit clearer. It also needs to consider `PackageName` as input.
* Renamed `ManifestDocument.ApplicationName` and
  `GenerateJavaStubs.ApplicationName` to `ApplicationLabel`.
* `${applicationId}` was being replaced with the contents of
  `ApplicationName`! The only reason this worked was because it was
  always blank, because it wasn't passed in from the MSBuild targets.
  The default was: `ApplicationName = ApplicationName ?? PackageName`.
* Removed usage of `ApplicationName` in `<Aapt/>` and `<Aapt2Link/>`.
  We shouldn't need to specify `application:label` in a temporary
  `AndroidManifest.xml` file.


Documentation/guides/building-apps/build-properties.md
```diff
@@ -1160,6 +1174,36 @@ string.
 
 Added in Xamarin.Android 7.2.
 
+## ApplicationId
+
+An MSBuild property that can be used as an alternative to
+`/manifest/@package` in the `AndroidManifest.xml`
+file. To opt into this feature you must also enable
+`<GenerateApplicationManifest>true</GenerateApplicationManifest>`.
+This will be the default going forward in .NET 6.
```
dellis1972, line 1183: Should we include a note that states this should NOT point to an android resource value e.g @strings/packageName. 
While aapt2 it seems lets you do this, the app will fail to run when installed.

Documentation/guides/building-apps/build-properties.md
```diff
@@ -1098,6 +1098,20 @@ than `aapt`.
 
 Added in Xamarin.Android 8.1.
 
+## AndroidVersionCode
+
+An MSBuild property that can be used as an alternative to
+`/manifest/@android:versionCode` in the `AndroidManifest.xml` file. To
+opt into this feature you must also enable
+`<GenerateApplicationManifest>true</GenerateApplicationManifest>`.
+This will be the default going forward in .NET 6.
```
dellis1972, line 1107: We should make a note that this needs to be an integer value.

src/Xamarin.Android.Build.Tasks/Tasks/GetAndroidPackageName.cs
```diff
@@ -47,29 +45,28 @@ public class GetAndroidPackageName : AndroidTask
...
+				PackageName = AndroidAppManifest.CanonicalizePackageName (PackageName);
+			} else if (!string.IsNullOrEmpty (ManifestFile) && File.Exists (ManifestFile)) {
+				using var stream = File.OpenRead (ManifestFile);
+				using var reader = XmlReader.Create (stream);
+				if (reader.MoveToContent () == XmlNodeType.Element) {
+					var package = reader.GetAttribute ("package");
+					if (!string.IsNullOrEmpty (package)) {
+						PackageName = AndroidAppManifest.CanonicalizePackageName (package);
+					}
+				}
```
dellis1972, line 58: what happens if PackageName is empty and the ManifestFile does not exist? Is that even possible?
We no longer fall back to the AssemblyName?

jonathanpeppers: A few lines lower if `PackageName` is still empty, it does that check last.

I was trying to make `AndroidAppManifest.CanonicalizePackageName` get called less, because it seems to do a lot, including a `Regex`:

https://github.com/xamarin/xamarin-android-tools/blob/26d65d95943b18bb9575d75c8c60ce36180a0db9/src/Xamarin.Android.Tools.AndroidSdk/AndroidAppManifest.cs#L54-L59

Documentation/guides/building-apps/build-properties.md
jonathanpeppers: I added a comment to all of these like:

> See the [Android documentation][manifest-element] for further details
about the requirements for `/manifest/@package`.

[manifest-element]: https://developer.android.com/guide/topics/manifest/manifest-element

Google has the full list of rules.

dellis1972 (approved):

jonpryor:
`<ApplicationTitle/>` concerns me: the hope/expectation is that this is for a "single project" which builds either/both Android & iOS apps, right?  How does that work when the *values* are entirely different?

```xml
<!-- Android -->
<ApplicationTitle>@string/application_title</ApplicationTitle>

<!-- iOS -->
<ApplicationTitle>MyApp</ApplicationTitle>
```

I'm not entirely sure how to square this circle?  [iOS localizes `CFBundleDisplayName` within `InfoPlist.strings`](https://stackoverflow.com/questions/2871990/how-to-localize-bundle-display-name-in-iphone-app); presumably `$(ApplicationTitle)` is just the default (English) `CFBundleDisplayName` value?

I suppose for Android we could have an `$(ApplicationTitleResource)` property which takes priority over `$(ApplicationTitle)`, but if `$(ApplicationTitleResource)` isn't set then `AndroidManifest.xml` can "inline" the `$(ApplicationTitle)` value, which would allow consistent behavior:

```xml
<!-- Android & iOS, default value -->
<ApplicationTitle>My Nifty App!</ApplicationTitle>

<!-- Android, optional override -->
<ApplicationTitleResource>@string/application_title</ApplicationTitle>
```

jonathanpeppers:
@rolfbjarne or @Redth do you have an opinion about `$(ApplicationTitle)` when it is localized? (see [comment above](https://github.com/xamarin/xamarin-android/pull/5458#issuecomment-762970920))

Another option besides introducing another MSBuild property, would be to omit `$(ApplicationTitle)` for iOS when it is localized:

```xml
<ApplicationTitle Condition=" '$(TargetPlatformIdentifier)' == 'android' ">@string/application_title</ApplicationTitle>
```

If you have `CFBundleDisplayName` defined in a `.strings` file, I think you can just remove it from the `Info.plist`?

jonathanpeppers:
@jonpryor I remembered I actually put this in [the spec](https://github.com/xamarin/xamarin-android/blob/469ada900643579a60c78854d9605d1c896d2dc8/Documentation/proposals/OneDotNetSingleProject.md#localization), but we can change this if there a better idea:

## Localization

`$(ApplicationTitle)` can easily be localized on Android by using an
**AndroidResource** as the value.

However, on iOS you would need to specify
[`LSHasLocalizedDisplayName`][0] in the `Info.plist` and provide
`CFBundleName` or `CFBundleDisplayName` in a `.strings` file for each
supported language. For our first implementation, you would omit
`$(ApplicationTitle)` from the `.csproj` file if you need to localize
it.

One can imagine supporting a `.resx` key via a new
`$(LocalizedApplicationTitle)` property. This would likely need to be
implemented in Xamarin.Forms/MAUI MSBuild tasks as a way to provide a
single `.resx` file to be translated to the appropriate format for iOS
and Android. This is a consideration for the future.

[0]: https://developer.apple.com/library/archive/documentation/General/Conceptual/MOSXAppProgrammingGuide/BuildTimeConfiguration/BuildTimeConfiguration.html

jonathanpeppers:
iOS side of this: https://github.com/xamarin/xamarin-macios/pull/10545

jonpryor:
It occurs to me that if `$(GenerateApplicationManifest)` will be True in .NET 6, then our `dotnet new` templates (e346f8468e7727e27f2efadf5046c68e9cc5b33c) should likewise be updated accordingly, e.g. *remove*:

  * https://github.com/xamarin/xamarin-android/blob/772a07273f6720d77ae6e01245f42da9e6f82e62/src/Microsoft.Android.Templates/android/AndroidManifest.xml

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/SingleProjectTest.cs
```diff
@@ -0,0 +1,54 @@
...
+	{
+		[Test]
+		public void AndroidManifestProperties ()
+		{
+			var packageName = "com.xamarin.singleproject";
+			var applicationLabel = "My Sweet App";
+			var versionName = "2.1";
+			var versionCode = "42";
+			var proj = new XamarinAndroidApplicationProject ();
+			proj.AndroidManifest = proj.AndroidManifest
```
jonpryor, line 22: Likewise, if the default `dotnet new android` template no longer contains an `AndroidManifest.xml` -- and why should it now? -- then these `.Replace()` calls don't make sense.  Instead, `project.AndroidManifest` should be `null` (or some other value to prevent it from being created).

jonpryor: Even in the case of "legacy" Xamarin.Android, if we're setting `$(GenerateApplicationManifest)`=True, then we shouldn't need `AndroidManifest.xml` at all, right?

jonathanpeppers: If anyone comes across this, we haven't completely removed `AndroidManifest.xml` yet.

The big feature missing to do this is app icons.

PR #5460, opened 2021-01-05
Bump to xamarin/sqlite/3.34.0@c9eba82c

grendello:
Context: https://sqlite.org/releaselog/3_32_3.html
Context: https://sqlite.org/releaselog/3_33_0.html
Context: https://sqlite.org/releaselog/3_34_0.html

Changes: https://github.com/xamarin/sqlite/compare/3.32.2..3.34.0

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -38,7 +38,7 @@
 [submodule "external/sqlite"]
     path = external/sqlite
     url = https://github.com/xamarin/sqlite.git
-    branch = 3.32.1
+    branch = 3.34.0
 [submodule "external/xamarin-android-tools"]
     path = external/xamarin-android-tools
     url = https://github.com/xamarin/xamarin-android-tools


radekdoulik (approved):

jonpryor:
While you were away, the default branch for xamarin/sqlite was changed from `master` to `main`:

  * main: https://github.com/xamarin/sqlite/commits/main
  * master: https://github.com/xamarin/sqlite/commits/master

`main` was updated on December 15 and used to create the https://github.com/xamarin/sqlite/tree/3.32.2 branch.

(I would have done 3.34 at the time, but Google's SQLite patch didn't apply to 3.34 at the time, so…)

@grendello: Would it be possible to update things to use sqlite/main instead of sqlite/master?

jonpryor (changes requested):
https://github.com/xamarin/xamarin-android/pull/5460#issuecomment-755854160

grendello:
@jonpryor the upstream patch still doesn't apply, I fixed it by hand :)

grendello:
@jonpryor I removed the "master" based branch and replaced with one rebased on "main" - it DID rewrite history for `xamarin/sqlite` though (shouldn't be a problem since it's only us who access it and the 3.34.0 branch was referenced just by this PR)

jonpryor:
@grendello  wrote:

>  I fixed it by hand

You are a braver man than I.

grendello:
> @grendello wrote:
> 
> > I fixed it by hand
> 
> You are a braver man than I.

It was only a very small fragment, offset by a couple hundred lines

Issue #5438, opened 2020-12-17
[One .NET] split up Android SDK pack into additional packs


jonathanpeppers:
We are thinking this would help with the install footprint.

Looking in the current pack, there are a couple options:

![image](https://user-images.githubusercontent.com/840039/102523082-b4d5a000-405c-11eb-8409-90f7298aa388.png)

These could be split up into a pack per-RID. If you build an app for `arm64-v8a;x86`, you would only need 2 of these packs installed.

Also we might consider putting `bundletool.jar` in its own pack:

![image](https://user-images.githubusercontent.com/840039/102523268-e6e70200-405c-11eb-80d9-29cc2b1aa220.png)

You would only need this installed when using `$(AndroidPackageFormat)=aab`.

`manifestmerger.jar` could also be split up based on if `$(AndroidManifestMerger)` is `manifestmerger.jar`. Although this may not matter as AndroidX packages require it. *Every* app will effectively use it.

Helpful links:
* https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workloads.md#composition
* https://github.com/dotnet/designs/pull/120/files

jonathanpeppers:
We have made the changes for `bundletool` here. We could split up things further per target device ABI, punting this to .NET 7.

jonathanpeppers:
This is no longer a goal -- splitting up packs. On Windows, it actually increases the install time to have more packs, as you can only install one `.msi` at a time.

We also can't split up on RIDs anymore, as projects build all 4 RIDs by default.

PR #5465, opened 2021-01-05
[One .NET] put bundletool.jar in separate workload pack

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/issues/5438

This only implements half of #5438.

`bundletool.jar` is quite large:

    24755974 bundletool.jar

Since this isn't needed for every build, only when
`$(AndroidPackageFormat)` is `aab`, we can put `bundletool.jar` in its
own "workload" pack.

The way this works is for `WorkloadManifest.targets` to define an
optional Sdk import:

        Condition=" '$(AndroidPackageFormat)' == 'aab' " />

If the SDK is missing, for a build using `AndroidPackageFormat=aab`
you get the error:

    Microsoft.NET.Workload.Android\WorkloadManifest.targets(4,33): error MSB4236: The SDK 'Microsoft.Android.Sdk.BundleTool' specified could not be found.

Our .NET 6 installers will include `Microsoft.Android.Sdk.BundleTool`
and install it to the correct location to avoid this error.

Down the road, `Microsoft.Android.Sdk.BundleTool` would be acquired by
a future `dotnet workload` command that does not exist yet.

The resulting size of the two Sdk packs are:

    22881359 Microsoft.Android.Sdk.BundleTool.11.0.200.nupkg
    49970851 Microsoft.Android.Sdk.win-x64.11.0.200.nupkg

### DIFF
--- a/build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
+++ b/build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
@@ -35,6 +35,7 @@
       <_FilesToCopy Include="$(DotNetPreviewPath)\sdk-manifests\$(DotNetPreviewVersionBand)\Microsoft.NET.Workload.Android\**\*" />
       <_FilesToCopy Include="$(DotNetPreviewPath)\packs\Microsoft.Android.Ref\**\*" />
       <_FilesToCopy Include="$(DotNetPreviewPath)\packs\Microsoft.Android.Sdk.osx-x64\**\*" />
+      <_FilesToCopy Include="$(DotNetPreviewPath)\packs\Microsoft.Android.Sdk.BundleTool\**\*" />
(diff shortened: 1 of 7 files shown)


jonathanpeppers:
I manually downloaded the `.msi` and `.pkg` installers, and the `Microsoft.Android.Sdk.BundleTool` pack is included.

pjcollins (approved):
LGTM 👍. I was wondering about the import ordering of `Microsoft.Android.Sdk.BundleTool/Sdk/Sdk.targets` happening after `Xamarin.Android.Common.targets`, but it looks like the `$(AndroidBundleToolJarPath)` property is set during the `_ResolveMonoAndroidSdks` target (rather than being statically set somewhere). The value from the new `BundleTool` targets should take precedence.

PR #771, opened 2017-08-22
[enumification] fix NotificationVisibility enumification.

atsushieno:
It was getLockscreenVisibility, not getLockScreenVisibility...

----

Do not merge it yet, let's see if run-api-compatibility-check actually reports the ABI breakage. It didn't for me...

PR #770, opened 2017-08-22
Bump to Java.Interop/master/33436346

jonpryor:
Fixes: https://bugzilla.xamarin.com/show_bug.cgi?id=57828

PR #768, opened 2017-08-22
[Xamarin.Android.Build.Tasks] Make use we use UTC Dates for ResolveLibraryProjectImports.cs

dellis1972:
We were not using UTC for allot of the date time comparisons in
the `ResolveLibraryProjectImports.cs` task. We really should be
since it means we should not hit any issues with LocalTimes.

PR #5468, opened 2021-01-06
Bump to xamarin/Java.Interop/master@fdc200cc

jonpryor:
Fixes: https://github.com/xamarin/java.interop/issues/767

Changes: https://github.com/xamarin/java.interop/compare/7574f166008bb45c0df97315aae7907ac25f8602...fdc200cca5676660dd8f732f4d7eef7716daa3df

  * xamarin/java.interop@fdc200cc: [Xamarin.Android.Tools.Bytecode] Relax _ApiXml check (#772)
  * xamarin/java.interop@f1b93653: [generator] Change generated code to not emit CA1305 warning. (#771)
  * xamarin/java.interop@2244407d: [generator] Ensure DIM from assembly refs are correctly marked (#770)
  * xamarin/java.interop@da73d6a5: [Java.Interop] Prevent premature collection w/ JniInstance* (#768)

Commit a7413a23 added support for invoking `java-source-utils.jar`
on `@(JavaSourceJar)` to extract Javadoc comments and translate them
into C# XML Documentation Comments.

What this can *also* do is provide correct parameter names.
As of commit a7413a23, the `BindingBuildTest.JavaSourceJar()`
integration test would emit the warning:

	obj/Debug/generated/src/Com.Xamarin.Android.Test.Msbuildtest.JavaSourceJarTest.cs(75,20):
	warning CS1572: XML comment has a param tag for 'name', but there is no parameter by that name

Commit xamarin/java.interop@fdc200cc allows `java-source-utils.jar`
output to be used with `class-parse`, allowing
`@(_JavaSourceJavadocXml)` files -- the output of
`java-source-utils.jar` -- to be included in
`@(_AndroidDocumentationPath)`.

This allows `@(JavaSourceJar)` files to provide parameter names
within bindings, removing the CS1572 warning, and making for
better overall bindings.

We can see the benefits of this change in
`tests/CodeGen-Binding/Xamarin.Android.JcwGen-Tests/BindingTests.cs`,
which required changes because the parameter names in the Java
`DataListener.onDataReceived()` method could now be determined;
previously they couldn't, resulting in the `P0`/`P1`/etc. names.
With the provision of `@(JavaSourceJar)` -- a7413a23 updated
`Xamarin.Android.McwGen-Tests.csproj` to have `@(JavaSourceJar)` --
the parameter names can now be determined, improving the binding.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Bindings.ClassParse.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Bindings.ClassParse.targets
@@ -19,9 +19,7 @@ This file is only used by binding projects.
       Outputs="$(ApiOutputFile)">
 
     <ItemGroup>
-      <!-- TODO: add when `ClassPath` supports `<api api-source="java-source-utils">`
       <_AndroidDocumentationPath Include="@(_JavaSourceJavadocXml)" />
-        -->
       <_AndroidDocumentationPath Include="@(JavaDocIndex->'%(RootDir)\%(Directory)')" />
       <_AndroidDocumentationPath Include="$(JavaDocPaths)" />
       <_AndroidDocumentationPath Include="$(Java7DocPaths)" />
--- a/tests/CodeGen-Binding/Xamarin.Android.JcwGen-Tests/BindingTests.cs
+++ b/tests/CodeGen-Binding/Xamarin.Android.JcwGen-Tests/BindingTests.cs
@@ -66,14 +66,14 @@ namespace Xamarin.Android.JcwGenTests {
 		{
 			using (var dh = new Com.Xamarin.Android.DataHandler ()) {
 				EventHandler<Com.Xamarin.Android.DataEventArgs> h = (o, e) => {
-					Assert.AreEqual ("fromNode", e.P0);
-					Assert.AreEqual ("fromChannel", e.P1);
-					Assert.AreEqual ("payloadType", e.P2);
-					for (int i = 0; i < e.P3.Length; ++i) {
-						for (int j = 0; j < e.P3 [i].Length; ++j) {
+					Assert.AreEqual ("fromNode", e.FromNode);
+					Assert.AreEqual ("fromChannel", e.FromChannel);
+					Assert.AreEqual ("payloadType", e.PayloadType);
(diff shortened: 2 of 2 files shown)


radekdoulik (approved):

dellis1972 (approved):
This looks ok. I think we need some release notes though. Will this impact users like it did our test? i.e will they need to update code to get around build errors?

jonpryor:
> Will this impact users like it did our test?

This shouldn't actually impact anybody.  (Rephrased; it fixes a regression!)

The whole point to `@(JavaSourceJar)` was to do exactly what is happening here: parse the `.java` files within `@(JavaSourceJar)` into "something", which can then be used to get better parameter names.

Before *JDK 11 support*, before 380e95e340168f68b1ae3eb43e0a71b05bc6d80b, this was done in the [`_GenerateJavaDocFromSourceJars`](https://github.com/xamarin/xamarin-android/blob/a111f6704d640481a6a85234ec4aa86ddbc18c3a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Bindings.Documentation.targets#L32-L51) target: we'd extract `@(JavaSourceJar)`, run `javadoc` on the sources, pray that it worked -- it did for `BindingTests.JavaSourceJar()`! -- and if it did work, we'd process the resulting HTML to extract parameter names.

The problem is that JDK 11 changed the HTML output format, breaking our parsing, so to "move forward" we *disabled* `@(JavaSourceJar)` when running under JDK 11.

…except we didn't *just* disable under JDK 11; we disabled *everywhere*: https://github.com/xamarin/xamarin-android/blob/380e95e340168f68b1ae3eb43e0a71b05bc6d80b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Bindings.Documentation.targets#L19-L21

^^ *always* evaluates to `""` (unset), because `$(_JdkVersion)` isn't set at the time that property group is evaluated.

So before JDK 11 support, we'd attempt to process `@(JavaSourceJar)`.  *After* JDK 11 support, we didn't at all, even though we *intended* to.

Thus, in migrating away from `javadoc` to `java-source-utils.jar`, we fixed this inadvertent regression, which apparently nobody noticed.

(…and probably nobody noticed because [only *36* people on GitHub use it](https://github.com/search?q=javasourcejar+extension%3Acsproj&type=Code&ref=advsearch&l=&l=).)

jonathanpeppers (approved):

PR #5475, opened 2021-01-07
Update NDK to r22

grendello:
The most important NDK changes:
  * GNU binutils are deprecated (but still used)
  * LLVM 11 is used for the toolchain

XA changes:
  * All the binutils tools are listed in a single location, `Configurables.Defaults`, now
  * Host NDK binutils are installed by xaprepare instead of by Xamarin.Android.Build.Tools
  * `UBSAN` checked builds require `RTTI` and exceptions now


src/monodroid/CMakeLists.txt
```diff
@@ -307,9 +309,9 @@ if(ENABLE_NDK AND (ENABLE_CLANG_UBSAN OR ENABLE_CLANG_ASAN))
   string(STRIP "${CMAKE_REQUIRED_FLAGS}" CMAKE_REQUIRED_FLAGS)
 endif()
 
-check_include_file_cxx("linux/netlink.h" HAVE_LINUX_NETLINK_H)
-check_include_file_cxx("linux/rtnetlink.h" HAVE_LINUX_RTNETLINK_H)
-check_include_file_cxx("linux/if_arp.h" HAVE_LINUX_IF_ARP_H)
+check_include_file("linux/netlink.h" HAVE_LINUX_NETLINK_H)
+check_include_file("linux/rtnetlink.h" HAVE_LINUX_RTNETLINK_H)
+check_include_file("linux/if_arp.h" HAVE_LINUX_IF_ARP_H)
```
radekdoulik, line 314: Just a nitpick, excessive space.

radekdoulik (approved):
LGTM once green.

radekdoulik:
Looks like it is now failing with:
```
CMake Error at CMakeLists.txt:85 (message):
  Please set the MINGW_DEPENDENCIES_ROOT_DIR variable on command line
  (-DMINGW_DEPENDENCIES_ROOT_DIR=PATH)
   Called from: [1]	C:/a/1/s/src/monodroid/CMakeLists.txt
```

grendello:
There's also the cmake policy warning which is treated as an error on Windows (dunno why). I'm about to look into those issues

build-tools/xaprepare/xaprepare/Steps/Step_Get_Windows_Binutils.cs
```diff
@@ -79,26 +79,32 @@ public Step_Get_Windows_Binutils ()
 		protected override async Task<bool> Execute (Context context)
 		{
 			string ndkVersion = BuildAndroidPlatforms.AndroidNdkVersion;
+			string baseArchivePath = $"android-ndk-r{ndkVersion}/toolchains/llvm/prebuilt/windows-x86_64/bin";
 
-			var neededFiles = new HashSet<string> (StringComparer.OrdinalIgnoreCase) {
```
jonpryor, line 83: I don't understand why the `HashSet<string>` was replaced with `Dictionary<string, string>`.  Nor do I understand what the "value" is supposed to be…

build-tools/xaprepare/xaprepare/Steps/Step_Get_Windows_Binutils.cs
```diff
@@ -79,26 +79,32 @@ public Step_Get_Windows_Binutils ()
...
-				$"android-ndk-r{ndkVersion}/toolchains/llvm/prebuilt/windows-x86_64/bin/arm-linux-androideabi-strip.exe",
-				$"android-ndk-r{ndkVersion}/toolchains/llvm/prebuilt/windows-x86_64/bin/x86_64-linux-android-as.exe",
-				$"android-ndk-r{ndkVersion}/toolchains/llvm/prebuilt/windows-x86_64/bin/x86_64-linux-android-ld.exe",
-				$"android-ndk-r{ndkVersion}/toolchains/llvm/prebuilt/windows-x86_64/bin/x86_64-linux-android-strip.exe",
-				$"android-ndk-r{ndkVersion}/toolchains/llvm/prebuilt/windows-x86_64/bin/aarch64-linux-android-as.exe",
-				$"android-ndk-r{ndkVersion}/toolchains/llvm/prebuilt/windows-x86_64/bin/aarch64-linux-android-ld.exe",
-				$"android-ndk-r{ndkVersion}/toolchains/llvm/prebuilt/windows-x86_64/bin/aarch64-linux-android-strip.exe",
-				$"android-ndk-r{ndkVersion}/toolchains/llvm/prebuilt/windows-x86_64/bin/libwinpthread-1.dll",
+			var neededFiles = new Dictionary<string, string> (StringComparer.OrdinalIgnoreCase) {
+				{ $"{baseArchivePath}/libwinpthread-1.dll", String.Empty },
```
jonpryor, line 85: …considering that the "value" is `string.Empty`.

build-tools/xaprepare/xaprepare/Steps/Step_Get_Windows_Binutils.cs
jonpryor: ah, "value" is "destination path".

grendello: Key is the original file name, Value is an optional target file name. Used to move `ld.gold` to `ld`

build-tools/xaprepare/xaprepare/Steps/Step_Get_Windows_Binutils.cs
grendello: It's empty because we're not renaming this particular file

jonpryor:
We may need to hold off on this PR?  There are lots of failing unit tests a'la https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4360715&view=ms.vss-test-web.build-test-results-tab&runId=18012466&resultId=100080&paneView=attachments

```
Should have no MSBuild warnings.
Expected: True
But was:  False
```

```
…/BuildHasNoWarningsFalseTrueTrueapk/Properties/AndroidManifest.xml :
warning XA4216: AndroidManifest.xml //uses-sdk/@android:minSdkVersion '19' is less than API-21, this configuration is not supported.
```

It appears that NDK r22 bumps the minSdkVersion value to API-21 -- likely because it no longer has a `platforms/android-19` directory -- which introducing all these warnings.

I thus fear that we can't *really* bump to NDK r22 until we bump our minimum to API-21…which is where things get complicated.  .NET 6 will have API-21 as a minimum, but we plan on concurrently shipping "legacy" installers for some period of overlapping time, and both the "legacy" and .NET 6 installers will be coming out of this repo.

How do we sanely support that?

grendello:
We also need to wait for NDK to be fixed for us before we can switch. Regarding supporting both net6 and "legacy", the only option would be to enable building with both r21 and r22 - probably different pipelines running in parallel. It will definitely require some thought and work. I'm going to leave this PR open as a draft and we can come back to it when NDK is fixed. In the meantime we can discuss how to solve the net6 vs legacy conundrum.

grendello:
Most of the issues are fixed, there's only one of `AOT + LLVM`. Since NDK r22 removed `$arch-ld` replacing it with `$arch-ld.{bfd,gold}`, Mono is not able to link the AOT-d code since it hardcodes `ld` as the linker name :( If @marek-safar  and @lambdageek  OK it, I will next work on fixing that bit in Mono next week.

grendello:
I fixed all AOT failures, but they won't work in this PR until https://github.com/mono/mono/pull/20816 is in and we can update Mono to one with the fix.

grendello:
We are ready for NDK r22 :) Now we need to wait for new NDK release which fixes the detection issue

PR #5477, opened 2021-01-07
[Xamarin.Android.Build.Tasks] fix build error for class library

jonathanpeppers:
@PureWeen ran into an error building this project in Xamarin.Forms:

https://github.com/xamarin/Xamarin.Forms/blob/5.0.0/Stubs/Xamarin.Forms.Platform.Android/Xamarin.Forms.Platform.Android%20(Forwarders).csproj

    Xamarin.Android.Legacy.targets(467,5): error MSB3375: The file "obj\Debug\monoandroid10.0\__AndroidLibraryProjects__.zip" does not exist.

I could reproduce with VS 2019 16.9, but not 16.8.

The file did not actually exist?

    > ls 'C:\src\Xamarin.Forms\Stubs\Xamarin.Forms.Platform.Android\obj\Debug\monoandroid10.0\__AndroidLibraryProjects__.zip'
    ls : Cannot find path 'C:\src\Xamarin.Forms\Stubs\Xamarin.Forms.Platform.Android\obj\Debug\monoandroid10.0\__AndroidLibraryProjects__.zip' because it does not exist.

If you have a class library with 0 `@(AndroidResource)` files and use
AndroidX, this seems to trigger the issue. I could create an MSBuild
test that hits the issue.

Using 16.8, the `__AndroidLibraryProjects__.zip` file contained *only*
directories:

    library_project_imports\assets
    library_project_imports\bin
    library_project_imports\java
    library_project_imports\res

So I think what's happening is the changes in 17e13bc4 caused
directories to not be included in `.zip` files. And so the `.zip`
isn't created at all!

Why are the `@(AndroidJavaLibrary)` items not added?

In the example:

https://github.com/xamarin/AndroidX/blob/1679307092d8aea27d0a6af40f103305f5331075/source/AndroidXTargets.cshtml#L51-L53

    <AndroidJavaLibrary Include="$(MSBuildThisFileDirectory)..\..\jar\com.foo.jar">

This would hit the following code:

    if (AndroidJavaSources != null)
        foreach (var item in AndroidJavaSources)
            MonoAndroidHelper.CopyIfChanged (item.ItemSpec, Path.Combine (outDirInfo.FullName, item.ItemSpec));

If `ItemSpec` is a full path, then `Path.Combine()` will just return
the full path:

https://docs.microsoft.com/dotnet/api/system.io.path.combine#System_IO_Path_Combine_System_String_System_String_

So then the code effectively copies the file to itself, such as:

    MonoAndroidHelper.CopyIfChanged (item.ItemSpec, item.ItemSpec);

This does *nothing*, it is a no-op...

We don't want to actually *fix* the underlying problem, or we would
introduce possible duplicate Java libraries. If a NuGet package
depends on AndroidX, and starts redistributing these `.jar` files, any
Xamarin.Android application project consuming the package would get an
error from d8/r8 saying there are duplicate Java types.

To solve the problem for now, let's just add `AlwaysCreate="true"`,
this is an easy fix for the breakage in 16.9.

Future considerations:

* We need to fix the AndroidX NuGet packages.
* For .NET 6 projects, we probably should look into improving error
  messages for duplicate Java libraries.

### DIFF
--- a/Documentation/release-notes/5477.md
+++ b/Documentation/release-notes/5477.md
@@ -1,0 +1,5 @@
+### Application and library build and deployment
+
+  * [GitHub PR 5477](https://github.com/xamarin/xamarin-android/pull/5477):
+    *error MSB3375: The file `__AndroidLibraryProjects__.zip` does not exist.*
+    could occur when building certain Xamarin.Android class library projects.
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -1793,6 +1793,17 @@ namespace App1
 			}
 		}
 
+		[Test]
+		public void AndroidXClassLibraryNoResources ()
+		{
+			var proj = new XamarinAndroidLibraryProject ();
+			proj.AndroidResources.Clear ();
+			proj.PackageReferences.Add (KnownPackages.AndroidXLegacySupportV4);
+			using (var b = CreateDllBuilder ()) {
+				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
+			}
+		}
+
 #pragma warning disable 414
 		static object [] BuildApplicationWithJavaSourceChecks = new object [] {
 			new object[] {
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
@@ -440,7 +440,7 @@ projects. .NET 5 projects will not import this file.
         OutputDirectory="$(IntermediateOutputPath)$(_NativeLibraryImportsDirectoryName)"
         EmbeddedNativeLibraries="@(EmbeddedNativeLibrary)"
     />
-    <Touch Files="$(IntermediateOutputPath)__AndroidNativeLibraries__.zip" />
+    <Touch Files="$(IntermediateOutputPath)__AndroidNativeLibraries__.zip" AlwaysCreate="true" />
     <ItemGroup>
       <FileWrites Include="$(IntermediateOutputPath)__AndroidNativeLibraries__.zip" />
       <EmbeddedResource Include="$(IntermediateOutputPath)__AndroidNativeLibraries__.zip">
@@ -464,7 +464,7 @@ projects. .NET 5 projects will not import this file.
         FlatArchivesDirectory="$(_AndroidLibraryFlatArchivesDirectory)"
         RemovedAndroidResourceFiles="@(_AndroidResourceDestRemovedFiles)"
     />
-    <Touch Files="$(IntermediateOutputPath)__AndroidLibraryProjects__.zip" />
+    <Touch Files="$(IntermediateOutputPath)__AndroidLibraryProjects__.zip" AlwaysCreate="true" />
     <ItemGroup>
       <FileWrites Include="$(IntermediateOutputPath)__AndroidLibraryProjects__.zip" />
       <EmbeddedResource Include="$(IntermediateOutputPath)__AndroidLibraryProjects__.zip">


dellis1972 (approved):

radekdoulik (commented):
Do we even need the `__AndroidLibraryProjects__.zip` archive in this case?

It looks like the problem happens in `_CreateManagedLibraryResourceArchive` target, where we are adding it as an embedded resource. Would it be OK just not add it instead of creating "empty" archive and reduce the apk size a bit?

dellis1972:
> Do we even need the `__AndroidLibraryProjects__.zip` archive in this case?
> 
> It looks like the problem happens in `_CreateManagedLibraryResourceArchive` target, where we are adding it as an embedded resource. Would it be OK just not add it instead of creating "empty" archive and reduce the apk size a bit?

This is a good fix, however I agree it would be nice to allow this file to be optional.

jonathanpeppers:
The AndroidX NuGet packages seem to trigger this, because of:

    AndroidJavaLibraries
        C:\Users\user\.nuget\packages\xamarin.androidx.annotation\1.1.0.8\buildTransitive\monoandroid90\..\..\jar\androidx.annotation.annotation.jar
        C:\Users\user\.nuget\packages\xamarin.androidx.collection\1.1.0.6\buildTransitive\monoandroid90\..\..\jar\androidx.collection.collection.jar
        C:\Users\user\.nuget\packages\xamarin.androidx.lifecycle.common\2.2.0.4\buildTransitive\monoandroid90\..\..\jar\androidx.lifecycle.lifecycle-common.jar
        C:\Users\user\.nuget\packages\xamarin.androidx.arch.core.common\2.1.0.7\buildTransitive\monoandroid90\..\..\jar\androidx.arch.core.core-common.jar

But these aren't even added! It just causes the target to run, and cause this problem.

Without AndroidX, the target is skipped completely:

    Skipping target "_CreateManagedLibraryResourceArchive" because it has no inputs.
    Though the target has declared its inputs, the input specification only references empty properties and/or empty item lists.

I will look into what is going on with the AndroidX NuGets. I feel like they should be only added as `@(AndroidJavaLibrary)` for "app" projects and not class libraries.

jonpryor:
This PR & associated issue confuse me.

See also? https://discord.com/channels/732297728826277939/732297837953679412/798366053033181195

What we have here is a fix for the build failure, but the fix *appears to me* to only be solving the *symptoms*.

The underlying condition doesn't make sense to me.

[This comment](https://github.com/xamarin/xamarin-android/pull/5477#issuecomment-756787120) suggests that the "problem" is that if `@(AndroidJavaLibrary)` contains an entry that contains `..\`, then the referenced file is *skipped*

> ```
> AndroidJavaLibraries
>     C:\Users\user\.nuget\packages\xamarin.androidx.annotation\1.1.0.8\buildTransitive\monoandroid90\..\..\jar\androidx.annotation.annotation.jar
> …
> ```

This is *bananas*.  *Even if true*, the inability to use `..\` to pull in associated resources is *wrong*, particularly in a world of multi-targeting!  Do we *want* to have a separate copy of `androidx.annotation.annotation.jar` *per Target Framework Moniker within the NuGet package*?  (No!)

Plus, I don't personally understand *how* it could be true: the `<CreateManagedLibraryResourceArchive/>` *copies* the contents of `@(AndroidJavaLibrary)` into `OutputDirectory`:

https://github.com/xamarin/xamarin-android/blob/597e7169b069eff781d8f29ddb7746b719d7f781/src/Xamarin.Android.Build.Tasks/Tasks/CreateManagedLibraryResourceArchive.cs#L93-L95

…except now I see how it *can* be true: line 95 has a [`Path.Combine()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=net-5.0#System_IO_Path_Combine_System_String_System_String_), which returns the *second parameter* if the second parameter is an absolute path.

In this case we *know* that the second parameter will be an absolute path, meaning line 95 is equivalent to:

```csharp
MonoAndroidHelper.CopyIfChanged (item.ItemSpec, item.ItemSpec);
```

which is a no-op; the file isn't copied.

Because the `@(AndroidJavaLibrary)` files are never copied into `OutputDirectory`, no files are added to `__AndroidLibraryProjects.__.zip`, and because no files are added, we get the MSB3375.

jonpryor:
I feel like we need a return to "first principals" to better understand what the fix should be.

In particular, what are the semantics for `<CreateManagedLibraryResourceArchive/>`?  Sure, it creates `__AndroidLibraryProjects__.zip`, but what are the semantics of `__AndroidLibraryProjects__.zip`?

  * Contains resource case maps: 628d6cb869c872c34d17b014724a43e8d451fb93
  * Files in `@(AndroidJavaSource)` are copied into `library_project_imports\%(Identity)`
  * Files in `@(AndroidJavaLibrary)` are copied into `library_project_imports\%(Identity)`
  * Files in `@(AndroidAsset)` are copied into `assets`
  * Files in `@(AndroidResource)` are copied into `res`
  * Files in `@(_AndroidResourceDestRemovedFiles)` are removed from `__AndroidLibraryProjects__.zip`

Yet we now have a scenario where not all `@(AndroidJavaLibrary)` (and `@(AndroidJavaSource)`!) entries are copied into `__AndroidLibraryProjects__.zip`: specifically, if `%(Identity)` is a full path.

Meaning that if a lib has:

```xml
<ItemGroup>
  <AndroidJavaLibrary Include="$(MSBuildThisFileDirectory)\Jars\foo.jar" />
</ItemGroup>
```

then `foo.jar` won't be included into `__AndroidLibraryProjects__.zip`, and thus won't be copied into the final application.

It seems "sensible" that *even if* `@(AndroidJavaLibrary)` *and* `@(AndroidJavaSource)` use full paths, they should still be copied into `__AndroidLibraryProjects__.zip`.  (It's certainly more "understandable" if this were the case.)

There are two problems with this "sensible" fix:

1. What path should be used for the copy?  The file basename?  (File basename may be sensible for `@(AndroidJavaLibrary)`, but won't be useful for `@(AndroidJavaSource)`.  *Is* there a "sensible" path inference algorithm?  Perhaps borrow `MonoAndroidHelper.GetRelativePathForAndroidAsset()` somehow?)
2. @jonathanpeppers believes that this "sensible fix" may *break* apps: https://discord.com/channels/732297728826277939/732297837953679412/798374263525605416

To wit:

> We have packages attempting to add these files right now, consuming a library that depends on AndroidX would end up with the jar twice
> So we fix it, will probably break someone too

Is there a way to "robustly" deal with the "same" library being included twice?

jonpryor:
Personally I like the idea of following .NET 6 + `<CreateAar/>`, and "hashing" the filename:

https://github.com/xamarin/xamarin-android/blob/597e7169b069eff781d8f29ddb7746b719d7f781/src/Xamarin.Android.Build.Tasks/Tasks/CreateAar.cs#L88

This arguably makes it *harder* to "robustly" deal with the same library being included twice, though, as the "same library" may have *multiple different filenames* across different dependency graphs.

However, this should be a current problem *already* in a .NET 6 contexts, as `@(AndroidJavaLibrary)` is used with `<CreateAar/>`!

https://github.com/xamarin/xamarin-android/blob/597e7169b069eff781d8f29ddb7746b719d7f781/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets#L76

(Also, `@(AndroidJavaSource)` doesn't appear to be a problem, as we can't find any code that *consumes* any `.java` files within `__AndroidLibraryProjects__.zip`, and thus it's moot.)

jonathanpeppers:
As discussed on our call today, we decided to simply add `AlwaysCreate="true"`.

I updated the commit message a bit, but @jonpryor might have more to add when we merge. 👍

aritchie:
@jonathanpeppers This issue is appearing in the 16.9 release.

Error	XACML7024	Xamarin.Tools.Zip.ZipIOException: File C:\REDACT	\obj\Debug\monoandroid10.0\__AndroidLibraryProjects__.zip is not a ZIP archive
   at Xamarin.Tools.Zip.ZipArchive.Open(String path, FileMode mode, String defaultExtractionDir, Boolean strictConsistencyChecks, IPlatformOptions options) in /Users/runner/work/1/s/ZipArchive.cs:line 273
   at Xamarin.Android.Tasks.CreateManagedLibraryResourceArchive.<>c__DisplayClass39_0.<RunTask>b__0(String f)
   at Xamarin.Android.Tools.Files.ArchiveZipUpdate(String target, Action`1 archiver)
   at Xamarin.Android.Tasks.CreateManagedLibraryResourceArchive.RunTask()
   at Xamarin.Android.Tasks.AndroidTask.Execute()	REDACT	C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\Xamarin.Android.Legacy.targets	456	

I have verified that AlwaysCreate="true" is in the Xamarin.Android.Legacy.targets.  The SDK style project if that matters.  The zip file is present at the path indicated, but is zero bytes.

jonathanpeppers:
It looks like your `$(IntermediateOutputPath)` is missing a trailing slash? 

    C:\REDACT \obj\Debug\monoandroid10.0_AndroidLibraryProjects_.zip

I think this is a different problem, can you file a new issue with a [diagnostic MSBuild log](https://docs.microsoft.com/xamarin/android/troubleshooting/troubleshooting#diagnostic-msbuild-output)? Multiple people tried the fix so far and it solved the issue for them.

aritchie:
I'm pretty sure it was removed by the github md due to the double underscore.  It is visible in the raw.  

I'll get you a build log today though.  Here is a fork that I started that was giving me this issue in the meantime: https://github.com/aritchie/aiforms.settingsview

aritchie:
@jonathanpeppers sorry for the delayed response.  I have more information and I do believe there is something worth investigating here.  The project did have compile issues of its own, however, it was producing a 0 byte zip file which was attempting to be used as shown in the exception above.  The main issue here is that the zip file exception was masking the actual issues of the underlying project.  I had to comb through the detailed build logs to actually fix all of the project issues.  Once I did that, the zip file issue also went away since the file now contained content.

Issue #5474, opened 2021-01-07
JDK 11 broke `@(JavaDocJar)` & related support


jonpryor:
Context: https://github.com/xamarin/xamarin-android/pull/5468#issuecomment-756057167

Commit 380e95e340168f68b1ae3eb43e0a71b05bc6d80b inadvertently broken `@(JavaSourceJar)` and `@(JavaDocJar)` support.

### Steps to Reproduce

1. Try the repro at Issue #2745:
  a. Create a new Binding project.
  b. Download [`glide-3.7.0.jar`](https://github.com/bumptech/glide/releases/download/v3.7.0/glide-3.7.0.jar), set its Build action to `@(EmbeddedJar)`.
  c. Download [`glide-3.7.0-javadoc.jar`](https://github.com/bumptech/glide/releases/download/v3.7.0/glide-3.7.0-javadoc.jar), set its Build action to `@(JavaDocJar)`
3. Build the project.

[Scratch.Gxa2745.zip](https://github.com/xamarin/xamarin-android/files/5783292/Scratch.Gxa2745.zip)

### Expected Behavior

The build should *use* the `@(JavaDocJar)` file.  (Not necessarily "build successfully", but "do something with" the `@(JavaDocJar)` files.

### Actual Behavior

The `@(JavaDocJar)` file isn't used.

Issue #2745 makes for an interesting comparison point: if the file were used, the build would *fail* with an MSB3375, unless the underlying bug was fixed (PR #4406?!), which hasn't happened.

### Discussion

However, this is a *regression* introduced by the introduction of JDK 11 support in 380e95e340168f68b1ae3eb43e0a71b05bc6d80b, due to buggy JDK 1.8 detection.

If we build the project, we get binding-related errors:

```bash
$ msbuild /v:diag
…
BINDINGSGENERATOR : warning BG8604: top ancestor SupportRequestManagerFragment not found for nested type Com.Bumptech.Glide.Manager.SupportRequestManagerFragment._1.
…
obj/Debug/generated/src/Com.Bumptech.Glide.Load.Resource.Gif.GifResourceEncoder.cs(10,70): error CS0535: 'GifResourceEncoder' does not implement interface member 'IEncoder.
Encode(Object, Stream)' 
```

Sure, it errors, but at least we got to the point of building the binding!

However, if we "manually work around" the broken JDK 1.8 detection support by setting the `$(_JavadocSupported)` property to True, we in fact *reproduce* Issue #2745:

```bash
[...]

Issue #2745, opened 2019-02-17
[Xamarin.Android.Build.Tasks] build error JavaSourceJar xxxx.stamp not exist


daddycoding:
### Steps to Reproduce

1. I tried to create a binding library from  [glide](https://github.com/bumptech/glide/releases/tag/v3.7.0)
download glide-3.7.0-javadoc.jar  glide-3.7.0.jar   and set action build to JavaDocJar
2. build binding library
3. 

### Expected Behavior
build success and binding library has  original Java parameter
### Actual Behavior
build error xxxx.stamp not exist 
### Log File
Xamarin.Android.Bindings.targets   
`<Touch Files="@(JavaDocJar->'$(IntermediateOutputPath)javadocs\%(FileName).stamp')" /> `
maybe miss the attribute  AlwaysCreate="True"

Daddoon:
This bug is still present on Visual Studio 2019 Preview 4.
I'm unable to create the JavaDoc, expect if i set AlwaysCreate="True" on the Xamarin.Android.Bindings.targets.

As it's from Visual Studio installation, it's not ideal at all.

PR #5479, opened 2021-01-07
[Xamarin.Android.Build.Tasks] Fix @(JavaDocJar)

jonpryor:
Fixes: https://github.com/xamarin/xamarin-android/issues/2745
Fixes: https://github.com/xamarin/xamarin-android/issues/5474

Issue #2745 noted that `@(JavaDocJar)` didn't work; attempting to use
it would result in an MSB3375 error:

	…/Xamarin.Android.Bindings.Documentation.targets(31,5): error MSB3375: The file "obj/Debug/javadocs/glide-3.7.0-javadoc.stamp" does not exist.

This issue was subsequently hidden by 380e95e3, which disabled the
`_ExtractJavaDocJars` target unless JDK 1.8 was being used.

There were two problems with 380e95e3:

 1. The `_ExtractJavaDocJars` target didn't need to be disabled!
    It just unzips the `.jar`; it does not require JDK 1.8.

 2. The check for JDK 1.8 was bad, and was *never* True.

The bad JDK 1.8 check was later removed in a7413a23, which "fixed"
(2), but a7413a23 continued to disable `_ExtractJavaDocJars`,
preventing `@(JavaDocJar)` from being used.

The only documentation-related functionality that needed to be
disabled because of JDK 11 was `@(JavaSourceJar)`, which got a better
fix in commits a7413a23 and 0e95ec7b.

Fix the `_ExtractJavaDocJars` target so that it properly runs when
there are any `@(JavaDocJar)` items, then fix the `_ExtractJavaDocJars`
target so that it doesn't emit an MSB3375 error.

Note: `@(JavaDocJar)` is still problematic: it involves parsing
Javadoc HTML, which contains numerous "dialects" (it's why the JDK 11
commit skipped `@(JavaSourceJar)` support: trying to update our HTML
scrapers was "too fiddly").

Deprecate `@(JavaDocJar)`, and remove
`Xamarin.Android.Bindings.Documentation.targets` from the .NET 6
installation packages.  Support for `$(_UseLegacyJavadocImport)`=False
won't exist for .NET 6; `@(JavaSourceJar)` support in .NET t will
*only* use `java-source-utils.jar` (a7413a23).

Update the `BindingBuildTest.cs` infrastructure so that instead of
having `.jar` files encoded in Base64, we instead use
`@(EmbeddedResource)`s for the `.jar` files.  A new `UpdateResources`
target will update the appropriate `.jar` files.


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/InlineData.cs
```diff
@@ -84,5 +34,18 @@ public static string ResxWithContents (string contents)
...
+		{
+			using var s = typeof (InlineData).Assembly.GetManifestResourceStream (name);
+			var data    = new List<byte> ();
+			var buf     = new byte[1024];
+			int r;
+			while ((r = s.Read (buf, 0, buf.Length)) > 0) {
+				for (int i = 0; i < r; ++i)
+					data.Add (buf [i]);
+			}
+			return data.ToArray ();
```
jonathanpeppers, line 48: Could this do something like this instead?

```csharp
var buf = new byte [s.Length];
s.Read (buf, 0, buf.Length);
return buf;
```
I think a `List<T>` would potentially resize the underlying array a couple times.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/InlineData.cs
```diff
@@ -8,60 +9,9 @@ namespace Xamarin.Android.Build.Tests
...
-6R/ys9K1KP+GWnoMqd9+MBgVIQ+IqR7afEiu81pNeDbD0j92ttv3dQVy0ZD+t0pP/h+fkYN6fvV1gCnvKKaoG6XMKMfmn
-8GeqKBHpqiFYIRiGpmv0SvaGqW8sthER7rzHkY28oDusMuoLvqAWDZ2+grt8hn6UhPGAv1NxB9DWcWbIk4QYxeyGEc3Rc
-BAJJXc0qmjw4eTvZ9QSwcINzBZxakBAAC1AgAAUEsBAhQAFAAICAgA74Q/TAAAAAACAAAAAAAAAAkABAAAAAAAAAAAAAA
-AAAAAAE1FVEEtSU5GL/7KAABQSwECFAAUAAgICADvhD9Mlldz8EQAAABFAAAAFAAAAAAAAAAAAAAAAAA9AAAATUVUQS1J
-TkYvTUFOSUZFU1QuTUZQSwECFAAUAAgICACzhD9MNzBZxakBAAC1AgAAPAAAAAAAAAAAAAAAAADDAAAAY29tL3hhbWFya
-W4vYW5kcm9pZC90ZXN0L21zYnVpbGR0ZXN0L0phdmFTb3VyY2VKYXJUZXN0LmNsYXNzUEsFBgAAAAADAAMA5wAAANYCAA
-AAAA==";
+		public  static  byte[]  JavaSourceJarTestJar        = GetResourceData ("javasourcejartest.jar");
+		public  static  byte[]  JavaSourceJarSourcesTestJar = GetResourceData ("javasourcejartest-sources.jar");
+		public  static  byte[]  JavaSourceJarJavadocTestJar = GetResourceData ("javasourcejartest-javadoc.jar");
```
jonathanpeppers, line 14: Should we make these properties, so they don't load the files unless the property is called? I'm thinking if I Debug a single test, it might load all these files.

It could do something like:
```csharp
static byte[] javaSourceJarJavadocTestJar;
public  static  byte[]  JavaSourceJarJavadocTestJar => javaSourceJarJavadocTestJar ??= GetResourceData ("javasourcejartest-javadoc.jar");
```

jonpryor: Do we need to worry about thread safety?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/InlineData.cs
jonpryor: That *may* work, but I don't know what the semantics of the stream returned by `GetManifestResourceStream()` is.  Is it possible that for "large" files `Stream.Read()` won't read in everything?

jonathanpeppers: Maybe this one then?
```csharp
using var s = typeof (InlineData).Assembly.GetManifestResourceStream (name);
using var m = new MemoryStream ();
s.CopyTo (m);
return m.ToArray ();
```

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/InlineData.cs
jonathanpeppers: Yes, I guess many of the tests use NUnit's parallel feature.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/InlineData.cs
radekdoulik: We might want to pass stream length to memory stream constructor if we want to avoid resizes during the process.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/InlineData.cs
```diff
@@ -8,60 +9,11 @@ namespace Xamarin.Android.Build.Tests
 {
 	static class InlineData
```
dellis1972, line 10: Bit of a nit pick, but should this still be called InLineData? cos its not really inline its embedded ?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/InlineData.cs
```diff
@@ -84,5 +36,13 @@ public static string ResxWithContents (string contents)
 		{
 			return Resx.Replace ("<!--contents-->", contents);
 		}
+
+		static byte[] GetResourceData (string name)
+		{
+			using var s = typeof (InlineData).Assembly.GetManifestResourceStream (name);
+			var data    = new byte [s.Length];
+			s.Read (data, 0, data.Length);
```
radekdoulik, line 44: This can return after reading less than data.Length bytes. https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=net-5.0

jonpryor: @radekdoulik: lol: https://github.com/xamarin/xamarin-android/pull/5479#discussion_r553655939

build-tools/installers/create-installers.targets
```diff
@@ -239,7 +239,7 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.targets" ExcludeFromAndroidNETSdk="true" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.ClassParse.targets" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.Core.targets" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.Documentation.targets" />
+    <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.Documentation.targets" ExcludeFromAndroidNETSdk="true" />
```
jonathanpeppers, line 242: Ok, this file wasn't even being imported in .NET 6. So this is safe to remove.

It was only imported for "legacy" binding projects:

https://github.com/xamarin/xamarin-android/blob/a7413a2389886082c3d3422c50a7e6cc84f43d8f/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Bindings.targets#L154

jonathanpeppers (approved):

dellis1972 (approved):

PR #5482, opened 2021-01-08
[build] Set up new Windows lane using `dotnet build`

jpobst:
Creates a new Windows build lane that Prepares and builds `Xamarin.Android.sln` using `dotnet build` instead of using .NET Framework.

Currently this is only implemented for Windows.  Mac needs the makefiles adjusted to call `dotnet` instead of `mono`.

Changes needed:
- Run `dotnet xaprepare.dll` instead of `xaprepare.exe`.
- `xabuild` removed from `Xamarin.Android.sln` and compiled as a separate step using .Net Framework/Mono.  This has issues due to its extensive "creative" use of MSBuild.
- Some fixups to `create-vsix`.  By default .Net Core only serializes string resources.  Added MSBuild prop and needed NuGet to support other resource types.
- By default .Net Core contains fewer encoding standards.  Add `System.Text.Encoding.CodePages` to `xaprepare` to add support for `iso-8859-4` for `Kajabity`

Future work needed:
- This is everything needed to build Xamarin.Android, but not everything needed by CI.  We still call:
  - `msbuild create-vsix`
  - `msbuild Xamarin.Android.sln /t:PackDotNet` (requires .NET 6)
  - `msbuild xabuild`
  - `xabuild Xamarin.Android-Tests.sln`
- Some executables that we build and run like `api-xml-adjuster` are still running their `.exe` form.


radekdoulik (approved):

jonathanpeppers (commented):
I think the main thing missing here is to update these docs:

* https://github.com/xamarin/xamarin-android/blob/master/Documentation/building/windows/instructions.md
* https://github.com/xamarin/xamarin-android/blob/master/Documentation/building/unix/instructions.md

They probably both need a short section saying to build `xabuild.csproj`, at least until we don't need that anymore.

grendello (approved):

jpobst:
Good call.  I updated the Windows directions.  The Unix directions shouldn't need it, since I added building `xabuild` into the `make` target.

pjcollins:
Is the goal here to replace the existing Windows Build + Test lane eventually? I am not sure if we should be running both of these jobs on every PR and CI build. Can we delete the existing Windows Build + Test job as part of these changes? If not, should we move it to the Nightly pipeline until we're ready to delete it?

jpobst:
I think the goal is to replace the .NET Framework build pretty quickly.  Once `dotnet` is supported for all platforms I think the next step will be to stop supporting building with Mono. (And stop supporting .NET Framework even though it will probably still work.)

Once we have stopped building with Mono, we can start targeting `NET5.0` or `NET6.0` instead of `NETCOREAPP3.1`.

Hopefully this transition will happen quick enough that temporarily running the extra lane doesn't really matter.

PR #5485, opened 2021-01-11
[Mono.Android] Add UpdateExternalDocumentation target

jonpryor:
Fixes: https://github.com/xamarin/xamarin-android/issues/5200

Context: a7413a2389886082c3d3422c50a7e6cc84f43d8f
Context: https://github.com/xamarin/android-api-docs/pull/23

Add a new `UpdateExternalDocumentation` target which:

 1. *Rebuilds* `src/Mono.Android` with
    `$(IncludeAndroidJavadoc)`=True.  This produces a
    `Mono.Anroid.xml` file containing imported API-30 Javadoc.

 2. Run `mdoc update --import Mono.Android.xml`, updating the
    [**mdoc**(5) documentation][0] within
    `external/android-api-docs/docs/Mono.Android/en`

Usage:

	msbuild /t:UpdateExternalDocumentation src/Mono.Android/Mono.Android.csproj

This process takes ~65 minutes on my machine.

Note: This uses the [mdoc NuGet package][1].

[0]: http://docs.go-mono.com/?link=man%3amdoc(5)
[1]: https://www.nuget.org/packages/mdoc/


src/Mono.Android/Mono.Android.targets
```diff
@@ -260,4 +268,26 @@
       <FileWrites Include="$(IntermediateOutputPath)CheckApiCompatibility.stamp" />
     </ItemGroup>
   </Target>
+
+  <Target Name="UpdateExternalDocumentation">
+    <Exec Command="git clean -xdf obj" WorkingDirectory="$(MSBuildThisFileDirectory)" />
+    <Exec
+        Command="time msbuild /restore /p:IncludeAndroidJavadoc=True /bl"
```
radekdoulik, line 275: `time` is not available on Windows. Would it make sense to change it to use MSBuild task or at least not use `time` conditionally on Windows? (same also few lines down)

jonpryor: `time` has been removed.

src/Mono.Android/Mono.Android.targets
```diff
@@ -260,4 +268,29 @@
...
+    <PropertyGroup>
+      <_Binlog>UpdateExternalDocumentation-$([System.DateTime]::Now.ToString ("yyyyMMddTHHmmss")).binlog</_Binlog>
+    </PropertyGroup>
+    <Exec
+        Command="msbuild /restore /p:IncludeAndroidJavadoc=True /bl:$(_Binlog)"
+        IgnoreStandardErrorWarningFormat="True"
+        WorkingDirectory="$(MSBuildThisFileDirectory)"
+    />
+    <PropertyGroup>
+      <_Mdoc Condition=" '$(Pkgmdoc)' != '' ">"$(Pkgmdoc)/tools/mdoc.exe"</_Mdoc>
```
jonpryor, line 283: @jonathanpeppers : any idea how to make this *work*.  Specifically, how to have the `$(Pkgmdoc)` MSBuild variable, set because of `%(PackageReference.GeneratePathProperty)` on the `mdoc` package, *actually exist* in the "outer" build?

Or is there some way to have the `UpdateExternalDocumentation` run as part of the "inner" build when invoking the target from the command-line?

```
msbuild /t:UpdateExternalDocumentation src/Mono.Android/Mono.Android.csproj
```

jonpryor:
More context: https://review.docs.microsoft.com/en-us/engineering/projects/reference/dotnet/mdoc

src/Mono.Android/Mono.Android.targets
jonathanpeppers: Unless you want to do this during `Build`, I think you'll have to make 2 targets for this to work...

Didn't test, but something like:
```xml
<Target Name="UpdateExternalDocumentation">
  <ItemGroup>
    <_UEDTargetFramework Include="$(TargetFrameworks)" />
  </ItemGroup>
  <MSBuild Projects="$(MSBuildThisProjectFile)" Targets="_UpdateExternalDocumentation" Properties="TargetFramework=%(_UEDTargetFramework.Identity)" />
 </Target>
 <Target Name="_UpdateExternalDocumentation">
   <!--Do the actual work here-->
 </Target>
 ```
 
 The "outer" one will need to call the `<MSBuild/>` task (batched), passing each value in `$(TargetFrameworks)` as `$(TargetFramework)`.

Comment (jonathanpeppers) [reply src/Mono.Android/Mono.Android.targets:283]: The inner target `_UpdateExternalDocumentation` might also need `DependsOnTargets="ResolveAssemblyReferences"` (or some target that populates `$(Pkg*)`).

Comment (jonpryor) [reply src/Mono.Android/Mono.Android.targets:292]: Got it working.  The inner target does *not* need any dependencies.  However:

1. it can't use be `<MSBuild Projects="$(MSBuildThisProjectFile)"/>`.  It needs to use the `.csproj` file, e.g. `<MSBuild Projects="$(MSBuildThisFileDirectory)Mono.Android.csproj" />.
2. I only want to run *one* inner build, so there's no need to have `@(_UEDTargetFramework)` or use `%(_UEDTargetFramework.Identity)`; just using `Properties="TargetFramework=monoandroid10"` suffices.
3. No `DependsOnTargets` is needed.

This in turn means that PR #5491 doesn't need to "worry about" installing the `mdoc` NuGet package into a top-level `xamarin-android/packages` folder, as we'll properly resolve the `mdoc` NuGet installation location via `$(Pkgmdoc)`.

File: src/Mono.Android/Mono.Android.targets [hunk 3]
```diff
@@ -260,4 +268,37 @@
...
+    <MSBuild Projects="$(MSBuildThisFileDirectory)Mono.Android.csproj"
+        Properties="TargetFramework=monoandroid10"
+        Targets="_UpdateExternalDocumentation"
+    />
+  </Target>
+  <Target Name="_UpdateExternalDocumentation">
+    <Exec
+        Command="git clean -xdf obj"
+        WorkingDirectory="$(MSBuildThisFileDirectory)"
+    />
```
jonathanpeppers, line 282: Could this use a regular `<RemoveDir/>`?

src/Mono.Android/Mono.Android.targets
```diff
@@ -260,4 +268,37 @@
...
+    <PropertyGroup>
+      <_Binlog>UpdateExternalDocumentation-$([System.DateTime]::Now.ToString ("yyyyMMddTHHmmss")).binlog</_Binlog>
+    </PropertyGroup>
+    <Exec
+        Command="msbuild /restore /p:IncludeAndroidJavadoc=True /bl:$(_Binlog)"
+        IgnoreStandardErrorWarningFormat="True"
+        WorkingDirectory="$(MSBuildThisFileDirectory)"
+    />
```
jonathanpeppers, line 290: Could this use `<MSBuild/>`? or does it actually need to run in a new process?

src/Mono.Android/Mono.Android.targets
pjcollins: I updated this to use removdir.

src/Mono.Android/Mono.Android.targets
pjcollins: I think the intent here was to make sure we always capture a unique binlog from the output. I've kept this for non windows platform, but updated this target to use <MSBuild/> on Windows so that it can be run there if desired.

jonathanpeppers (approved):

src/Mono.Android/Mono.Android.targets
```diff
@@ -260,4 +268,52 @@
...
+        Properties="TargetFramework=monoandroid10"
+        Targets="_UpdateExternalDocumentation;_RunMdoc"
+    />
+  </Target>
+  <Target Name="_UpdateExternalDocumentation">
+    <RemoveDir Directories="$(BaseIntermediateOutputPath)" />
+    <PropertyGroup>
+      <_Binlog>UpdateExternalDocumentation-$([System.DateTime]::Now.ToString ("yyyyMMddTHHmmss")).binlog</_Binlog>
+    </PropertyGroup>
+    <Exec
```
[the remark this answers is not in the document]
pjcollins: I mentioned this briefly in the comment above -- using Exec over MSBuild allows us to automatically capture a binlog for each Mono.Android build, I can move it to use MSBuild for both if we want though

PR #5488, opened 2021-01-12
[Xamarin.Android.Build.Tasks] cast with `as` for GetRegisteredTaskObject

jonathanpeppers:
Context: https://github.com/jonathanpeppers/Xamarin.Legacy.Sdk

When working on an experiment that enables multi-targeting such as:

  <Project Sdk="Xamarin.Legacy.Sdk">
    <PropertyGroup>
      <TargetFrameworks>monoandroid11.0;net6.0-android</TargetFrameworks>
    </PropertyGroup>
  </Project>

I got the build error:

    Xamarin.Android.Aapt2.targets(123,3): error MSB4018: The "Aapt2Compile" task failed unexpectedly.
    System.InvalidCastException: [A]Xamarin.Android.Tasks.Aapt2Daemon cannot be cast to [B]Xamarin.Android.Tasks.Aapt2Daemon. Type A originates from 'Xamarin.Android.Build.Tasks, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' at location 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\Xamarin.Android.Build.Tasks.dll'. Type B originates from 'Xamarin.Android.Build.Tasks, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' in the context 'Default' at location 'C:\Program Files\dotnet\packs\Microsoft.Android.Sdk.win-x64\11.0.200-ci.master.22\tools\Xamarin.Android.Build.Tasks.dll'.
      at Xamarin.Android.Tasks.Aapt2Daemon.GetInstance(IBuildEngine4 engine, String aapt2, Int32 numberOfInstances, Int32 initalNumberOfDaemons, Boolean registerInDomain)
      at Xamarin.Android.Tasks.Aapt2.Execute()
      at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
      at Microsoft.Build.BackEnd.TaskBuilder.ExecuteInstantiatedTask(ITaskExecutionHost taskExecutionHost, TaskLoggingContext taskLoggingContext, TaskHost taskHost, ItemBucket bucket, TaskExecutionMode howToExecuteTask)

This fails because we are doing an explicit cast instead of using the
`as` keyword. An `Aapt2Daemon` instance from the VS location of
`Xamarin.Android.Build.Tasks.dll` cannot be used from the .NET 6
location of the assembly.

I also added some extension methods that include the current
`Assembly.Location` in the key:

* `RegisterTaskObjectAssemblyLocal()`
* `GetRegisteredTaskObjectAssemblyLocal()`
* `UnregisterTaskObjectAssemblyLocal()`

I audited all `GetRegisteredTaskObject()` calls in this repo, and
added comments for the three tasks the usage seemed appropriate.

These all would be able to be shared between assemblies:

* `<ResolveAndroidTooling/>` uses the path to `aapt2` as a key and
  stores `System.Version` in the value.
* `<ResolveJdkJvmPath/>` uses `JavaSdkPath` as the key and stores the
  string `JdkJvmPath` in the value.
* `<ValidateJavaVersion/>` uses the path to java/javac as the key and
  stores a string in the value.

### DIFF
--- a/Documentation/guides/MSBuildBestPractices.md
+++ b/Documentation/guides/MSBuildBestPractices.md
@@ -462,11 +462,17 @@ Instead we can use an API provided by MSBuild:
 string key = "foo";
 BuildEngine4.RegisterTaskObject (key, "bar", RegisteredTaskObjectLifetime.Build, allowEarlyCollection: false);
 // To retrieve a value
-string value = (string)BuildEngine4.GetRegisteredTaskObject (key, RegisteredTaskObjectLifetime.Build);
+string value = BuildEngine4.GetRegisteredTaskObject (key, RegisteredTaskObjectLifetime.Build) as string;
 ```
 
 Or if you want to cache across multiple-builds, use
 `RegisteredTaskObjectLifetime.AppDomain` instead.
+
+> NOTE!
+> Use `as` for casts to avoid any unexpected `InvalidCastException`'s.
+> Use the `RegisterTaskObjectAssemblyLocal()` or
+> `GetRegisteredTaskObjectAssemblyLocal()` extension methods within
+> the Xamarin.Android codebase.
 
 To *test* and validate your MSBuild task's use of
 `RegisteredTaskObjectLifetime.AppDomain` you have two choices:
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -289,7 +289,7 @@ namespace Xamarin.Android.Tasks
 			if (compress) {
 				string key = CompressedAssemblyInfo.GetKey (ProjectFullPath);
 				Log.LogDebugMessage ($"Retrieving assembly compression info with key '{key}'");
-				compressedAssembliesInfo = BuildEngine4.UnregisterTaskObject (key, RegisteredTaskObjectLifetime.Build) as IDictionary<string, CompressedAssemblyInfo>;
+				compressedAssembliesInfo = BuildEngine4.UnregisterTaskObjectAssemblyLocal<IDictionary<string, CompressedAssemblyInfo>> (key, RegisteredTaskObjectLifetime.Build);
 				if (compressedAssembliesInfo == null)
 					throw new InvalidOperationException ($"Assembly compression info not found for key '{key}'. Compression will not be performed.");
 			}
--- a/src/Xamarin.Android.Build.Tasks/Tasks/FilterAssemblies.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/FilterAssemblies.cs
@@ -21,7 +21,6 @@ namespace Xamarin.Android.Tasks
 
 		const string TargetFrameworkIdentifier = "MonoAndroid";
 		const RegisteredTaskObjectLifetime Lifetime = RegisteredTaskObjectLifetime.AppDomain;
-		const bool AllowEarlyCollection = false;
 
 		public ITaskItem [] InputAssemblies { get; set; }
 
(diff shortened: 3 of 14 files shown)


radekdoulik (commented):
The change looks OK to avoid the exception.

Do I understand it correctly, that the `GetRegisteredTaskObject` is only used for caching, and so when we get `null` there, it would just mean the object (daemon, cachedMap) will be created? Which is hopefully what we want, so that we don't mix objects from 2 TFM builds.

jonathanpeppers:
> Do I understand it correctly, that the GetRegisteredTaskObject is only used for caching, and so when we get null there, it would just mean the object (daemon, cachedMap) will be created?

Yes, the places this is used would act as if it was the first call (nothing was cached). It would be unfortunate if a build somehow was switching between legacy/.NET 6 a lot, as it could render the cache useless.

In this example I was building a `samples.sln` that had two multi-targeted projects. It seems to happen every time in this case.

dellis1972 (approved):

jonpryor:
> This fails because we are doing an explicit cast instead

This doesn't quite make sense to me, and the *NOTE* that is added could likewise use some elaboration.

Is it fair/correct to say that when multi targeting is enabled -- `$(TargetFrameworks)`=`monoandroid11.0;net6.0-android` -- then *two different builds* of `Xamarin.Android.Build.Tasks.dll` are loaded: one build from the "legacy" net472/monoandroid11.0 environment, and one build for the net6.0 environment?

And it's this "'same' assembly built two different ways loaded at the same time" scenario which causes the `InvalidCastException`?

If so, then a *different* fix would be to instead *change the key* so that the key contains the target framework moniker.  Instead of:

```csharp
var cachedMap = (Dictionary<string, HashSet<string>>)engine?.GetRegisteredTaskObject (mapFile, RegisteredTaskObjectLifetime.Build)
```

It could instead be:

```csharp
// TODO
var tfm = "TODO";
var key = $"CustomMapFile{mapFile}@{tfm}";
var cachedMap = (Dictionary<string, HashSet<string>>)engine?.GetRegisteredTaskObject (key, RegisteredTaskObjectLifetime.Build);
```

"encoding" the target framework moniker -- or some "synonym" (full path to the calling assembly?) into the key could allow each separate "build" of `Xamarin.Android.Build.Tasks.dll` to use a different "key", which:

1. Would allow the current casts to work as-is, and
2. Would allow caching to work, addressing your [earlier comment](https://github.com/xamarin/xamarin-android/pull/5488#issuecomment-758825589):

    > It would be unfortunate if a build somehow was switching between legacy/.NET 6 a lot, as it could render the cache useless.

If we could "detect" this "switching" and encode it into the key, the cache would *not* be useless, *and* would allow the current casts to work.

Which could then be wrapped into a handy extension method and used to update *every* `GetRegisteredTaskObject()` invocation, because at present *all of them* are suspect.  This patch updates a `Dictionary`; why is `DeleteFileAfterBuild` special (`<WriteLockFile/>`), or `string` (`<ResolveAndroidTooling/>`, etc.).  (Or is it the `HashSet<string>` in `Dictionary<string, HashSet<string>>` which causes the problem?

jonathanpeppers:
> Which could then be wrapped into a handy extension method and used to update every GetRegisteredTaskObject() invocation, because at present all of them are suspect. This patch updates a Dictionary; why is DeleteFileAfterBuild special (<WriteLockFile/>), or string (<ResolveAndroidTooling/>, etc.). (Or is it the HashSet<string> in Dictionary<string, HashSet<string>> which causes the problem?

There is really only a problem if the type in the value comes from one of our assemblies. BCL types like `string`, `Dictionary`, `Version` would work fine.

I added extension methods that will put `Assembly.Location` in the key, because that is what causes the problem. 3 places that I thought should not use the new extension method, I added a comment. These were places that could share the value across both assemblies.

PR #5489, opened 2021-01-12
[tests] Test binary serialization

radekdoulik:
As we are adding `[Serializable]` attribute to `Java.Lang.Object` now,
test the binary serialization too.

### DIFF
--- a/tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
+++ b/tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
@@ -435,6 +435,10 @@ namespace Library1 {
 			proj.References.Add (new BuildItem.Reference ("System.Runtime.Serialization"));
 
 			if (Builder.UseDotNet) {
+				// serialization is broken on net6 when the app is linked. enable again once it is fixed
+				if (isRelease)
+					return;
+
 				proj.References.Add (new BuildItem.Reference ("System.Runtime.Serialization.Json"));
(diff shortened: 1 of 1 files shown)


jonathanpeppers (approved):

pjcollins (approved):

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
```diff
@@ -469,8 +467,45 @@ void TestJsonDeserializationCreatesJavaHandle ()
...
+				throw new InvalidOperationException (""BinaryFormatter deserialization of Name"");
+			if (p2.Age != 900)
+				throw new InvalidOperationException (""BinaryFormatter deserialization of Age"");
+			if (p2.Handle == IntPtr.Zero)
+				throw new InvalidOperationException (""Failed to instantiate new Java instance for Person!"");
+
+			Console.WriteLine ($""BinaryFormatter Person deserialized OK"");
+		}
+
+		void TestJsonDeserializationCreatesJavaHandle ()
```
jonpryor, line 498: This added test is failing under One .NET: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4365170&view=ms.vss-test-web.build-test-results-tab&runId=18051660&resultId=100089&paneView=attachments

radekdoulik: Nice. The net6 is more strict about unsafe binary serialization. I have added the binder and also disabled it in release as it fails with illink as linking of serialization is still broken on net6.

PR #5492, opened 2021-01-13
[Xamarin.Android.Build.Tests] Stop using PdbOnly for some unit tests.

dellis1972:
Since aab69af we have depreciated the use of Full and PdbOnly DebugTypes.
We should also update the unit tests which use those specific DebugTypes
to use the `Portable` type instead.


jonathanpeppers (commented):
This looks fine. Is there a way these tests could just omit `DebugType` completely? Just use default values?

PR #5493, opened 2021-01-13
[Xamarin.Android.Build.Tasks] Only run one _UpdateAndroidResources

dellis1972:
Lets try to see if we can make it so that `_UpdateAndroidResources`
will run only once. If there is a call to `UpdateAndroidResources`
and one is already running we should skip that call. So lets see
if we can do that.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -476,7 +476,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 </Target>
 
 <Target Name="UpdateGeneratedFiles"
-		DependsOnTargets="ResolveAssemblyReferences;_RemoveLegacyDesigner;UpdateAndroidResources"
+    DependsOnTargets="ResolveAssemblyReferences;_RemoveLegacyDesigner;UpdateAndroidResources"
 	>
 </Target>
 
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -969,10 +969,25 @@ because xbuild doesn't support framework reference assemblies.
 />
 
 <Target Name="UpdateAndroidResources"
-    Condition=" '$(DesignTimeBuild)' != 'True' Or '$(DeferredBuildSupported)' != 'True' Or '$(DeferredBuild)' == 'True' "
+    Condition=" ('$(DesignTimeBuild)' != 'True' Or '$(DeferredBuildSupported)' != 'True' Or '$(DeferredBuild)' == 'True') And !Exists ('$(_AndroidStampDirectory)_UpdateAndroidResources.stamp')"
     DependsOnTargets="_SetupDesignTimeBuildForCompile;_ManagedUpdateAndroidResgen;_UpdateAndroidResources"
 />
 
+<Target Name="_BeforeUpdateAndroidResources"
+    BeforeTargets="_SetupDesignTimeBuildForCompile"
```
dellis1972, line 977: note we use before and after targets in these new ones because we do want these to run even if there is an error.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -969,10 +969,25 @@ because xbuild doesn't support framework reference assemblies.
...
+    Condition="!Exists ('$(_AndroidStampDirectory)_UpdateAndroidResources.stamp')"
+>
+  <Touch Files="$(_AndroidStampDirectory)_UpdateAndroidResources.stamp" AlwaysCreate="True" />
+  <ItemGroup>
+    <FileWrites Include="$(_AndroidStampDirectory)_UpdateAndroidResources.stamp" />
+  </ItemGroup>
+</Target>
+
+<Target Name="_AfterUpdateAndroidResources" AfterTargets="UpdateAndroidResources">
```
jonpryor, line 986: Does an "AfterTargets" execute if the specified target doesn't execute?

`UpdateAndroidResources` is updated with a `Condition` so that it "doesn't run" if `_UpdateAndroidResources.stamp` exists.  Thus, if `UpdateAndroidResources` is currently executing -- and thus `_UpdateAndroidResources.stamp` exists -- and then a separate process/whatever *also* runs the `UpdateAndroidResources` target, which is a no-op because `_UpdateAndroidResources.stamp`, so the 2nd `UpdateAndroidResource` doesn't run -- will the separate "no-op" process _also_ run the `_AfterUpdateAndroidResources` target, thus deleting `_UpdateAndroidResources.stamp`, which will prevent this "file lock" solution from scaling to a *third* process running `UpdateAndroidResources`?

We might need to have "per-invocation" stamp files, and have `_AfterUpdateAndroidResources` *only* delete the file it created.  (Which might be differently "weird" if someone *kills* the msbuild process while it's running…?)

dellis1972: Yup this was my concern. But we also need to handle the situation where the underlying Aapt2 call Errors, in which case none of the dependencies will get run unless we use AfterTargets. So I'm not 100% sure this is going to work.

PR #5486, opened 2021-01-11
migration-to-sdk-short-style-projects-02-20210111

moljac:
Migrated:

*   check-boot-times 
*   conjure-xamarin-android-cecil
*   remap-assembly-ref
*   create-pkg

PR #5502, opened 2021-01-14
migration-sdk-style-prjsv build-tools\create-pkg

moljac:
create-pkg migrated


jonathanpeppers (commented):
This one needs the changes I commented about here:

https://github.com/xamarin/xamarin-android/pull/5486#pullrequestreview-568368846

It should be `Microsoft.Build.NoTargets` and shouldn't remove the import of `Configuration.props`.

PR #5509, opened 2021-01-15
Installing Azure CLI more robust in case az is not there at all.

trevors20:
If az is actually not even on the machine in the first place, it will fail the powershell script outright.

Instead, tell Powershell to continue and we check whether the command succeeded.  If it succeeded, do what we did before.  Otherwise, use a version of 0.0 to force us to install 2.8 az CLI.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/xamarin/xamarin-android?pullRequest=5509) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/xamarin/xamarin-android?pullRequest=5509) before we can accept your contribution.<br/><br/>:x: trevors20 [sign now](https://cla.dotnetfoundation.org/xamarin/xamarin-android?pullRequest=5509)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/xamarin/xamarin-android?pullRequest=5509) it.</sub>

RickKra (approved):

build-tools/automation/azure-pipelines.yaml
```diff
@@ -1208,11 +1208,23 @@ stages:
...
+            $AzureVersion = $Versions[0]
+            $AzureVersionFields = $AzureVersion.split(" ", [StringSplitOptions]::RemoveEmptyEntries)
+            $CurrentVersion = New-Object -TypeName System.Version -ArgumentList $AzureVersionFields[1]
+        }
+        else 
+        {
+            $CurrentVersion = New-Object -TypeName System.Version -ArgumentList "0.0"
+        }
+        
         $Version280 = New-Object -TypeName System.Version -ArgumentList "2.8.0"
```
jonpryor, line 1228: @trevors20: it occurs to me that *after* this point, we should "reset" [`$ErrorActionPreference`](https://docs.microsoft.com/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#erroractionpreference) back to `Stop` (which I assume is the default), so that if the subsequent `brew` commands fail (e.g. `brew` isn't installed either!), the step properly fails with an error, instead of blithely continuing execution…

jonpryor:
Squash-and-merge commit message for review:

```markdown
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4371179&view=logs&j=b53b0f16-cbff-5f88-33c0-fe0867b30359&t=c30f02ea-1494-5068-4c73-b6672a8ee06d

Sometimes, the **Finalize Installers** >
**Notarize and Upload to Storage** >
**Install Azure CLI >= 2.8.0** step fails:

	/usr/local/bin/pwsh -NoLogo -NoProfile -NonInteractive -Command . '/Users/builder/azdo/_work/_temp/b3001397-e37f-4340-93ab-f34e8e213a8e.ps1'
	& : The term 'az' is not recognized as the name of a cmdlet, function, script file, or operable program.
	Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
	At /Users/builder/azdo/_work/_temp/b3001397-e37f-4340-93ab-f34e8e213a8e.ps1:4 char:3
	+ & az --version
	+   ~~
	+ CategoryInfo          : ObjectNotFound: (az:String) [], ParentContainsErrorRecordException
	+ FullyQualifiedErrorId : CommandNotFoundException

The step fails because the `az` command isn't installed!

Update the step to set [`$ErrorActionPreference`][0] to `Continue`,
which prevents `az --version` from causing the step to fail entirely.
Instead, the process exit value is stored in `$?`.

Check the `az --version` exit value, and if it's an error, set
the `$CurrentVersion` value to `new Version("0.0")`.  This allows the
normal `$Version280 -gt $CurrentVersion` check to execute, which
will cause azure-cli to be installed.

[0]: https://docs.microsoft.com/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.1#erroractionpreference
```

build-tools/automation/azure-pipelines.yaml
trevors20: Great idea.  Let me do that.

mjbond-msft:
is there a reason you are not using the `az-client-update` template from `yaml-templates` in this case? that solution is now live. recommend applying this change their so all consumers can benefit
https://github.com/xamarin/yaml-templates/blob/main/azure-tools/az-client-update.yml

mjbond-msft (changes requested):
is there a reason you are not using the `az-client-update` template from `yaml-templates` in this case? that solution is now live. recommend applying this change their so all consumers can benefit
https://github.com/xamarin/yaml-templates/blob/main/azure-tools/az-client-update.yml

build-tools/automation/azure-pipelines.yaml
```diff
@@ -1205,48 +1205,10 @@ stages:
...
-            # link this version to start using it
-            & brew link azure-cli@2.8.0
-        }
-        else 
-        {
-            Write-Host "Version $CurrentVersion of azure-cli is higher than 2.8.0.  Nothin to do..."
-        }
-      displayName: 'Install Azure CLI >= 2.8.0'
+    # ESRP signing requires minimum azure client version 2.8.0
+    - template: azure-tools/az-client-update.yml@yaml
```
RickKra, line 1209: Where is this file? I'm not seeing it in the PR.

trevors20: Ya.  If you view the whole file you'll see something like this at the top of the file:
resources:
  repositories:
  - repository: yaml
    type: github
    name: xamarin/yaml-templates
    ref: refs/heads/main
    endpoint: xamarin

As part of the pipeline, they download the xamarin/yaml-templates repo on the fly.  Then, you can access those templates via something like "azure-tools/az-client-update.yml@yaml".  That way if there are any updates to any tasks in that repo, everyone will get a fix or whatever automatically when checked in.

I am moving these updates to the centralized repo with this PR: https://github.com/xamarin/yaml-templates/pull/99

Sound good?

RickKra: Sounds good.

RickKra (approved):

PR #5528, opened 2021-01-20
[Xamarin.Android.Build.Tasks] fix @(AndroidLibrary) in legacy Xamarin.Android

jonathanpeppers:
I was testing the new `@(AndroidLibrary)` item group, and noticed some
logic was backwards:

    <EmbeddedJar          Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.jar' and '%(AndroidLibrary.Bind)' != 'true' " />
    <EmbeddedReferenceJar Include="@(AndroidLibrary)" Condition=" '%(AndroidLibrary.Extension)' == '.jar' and '%(AndroidLibrary.Bind)' == 'true' " />

`@(EmbeddedJar)` should be used when `%(Bind)` is `true`! Because it
is using `@(EmbeddedReferenceJar)` instead, we are not currently
generating C# bindings by default.

We had a test for this item group, but it didn't actually check the
output assembly has a C# class from the `@(AndroidLibrary)`.

I updated the test and fixed the logic here.


dellis1972 (approved):

radekdoulik (approved):

jonathanpeppers:
Hopefully the `APK instrumentation` will be fixed by #5531 as this times out after 30 minutes: 

```
Error: Process '/usr/local/bin/brew "install" "xamarin/xamarin-android-windeps/mingw-zlib"' timed out after 00:30:00
Error: Installation of xamarin/xamarin-android-windeps/mingw-zlib failed
```

Going to merge.

PR #5529, opened 2021-01-21
[build] Enable signing for .NET 6 artifacts

pjcollins:
Context: https://github.com/xamarin/yaml-templates/tree/564100f2aa923ee4db4b69cc45a622822643309b/sign-artifacts
Context: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/650/Enabling-Signing-in-Your-Projects

The structure of our YAML pipeline has been updated to enable signing
for all relevant .NET 6 artifacts.  The new pipeline hierarchy is
described below:

    mac_build
        ...
        sign nupkgs

            create net6 .pkg
                download signed .nupkgs
                push signed .nupkgs to feed
                harden/sign mach-o binaries
                create .pkg
                sign .pkg
                notarize .pkg
                publish .pkg

                create net6 .msi
                    download signed .nupkgs
                    download notarized .pkg
                    create and sign .msi
                    upload .msi and .pkg to blob storage

The `dotnet_installers` stage now starts with a templated job which will
sign all .nupkg files created by our `mac_build` stage.  These signed
packages are then downloaded by a new macOS job that will push them to a
feed and use them to create, sign, and notarize the .NET 6 .pkg file.
Finally, the job which previously created our .NET 6 .msi has been
updated to wait for the result of the .NET 6 .pkg creation job.  This
job will now consume the signed .nupkg files to produce a signed .msi,
and upload both the signed .pkg and .msi files to blob storage.

None of the `MicroBuild` tasks/targets will run during local builds, so
we do not need to worry about additional environment provisioning when
creating these installers locally.  In order to test locally, the MicroBuild
signing NuGet needs to be installed locally and either the 
`MicroBuildOverridePluginDirectory` or `MicroBuildPluginDirectory` 
properties must be set and point to the folder where the signing NuGet
was installed. See the "Enabling signing in your projects" link above
for more information.

The new SignList.xml file is used by the .nupkg signing job.  The content
of this file controls which certs are used to sign which files.  Anything that
was signed through an alternate source should be added to the `@(Skip)`
ItemGroup.  Any third party file that needs signing should be added to the
`@(ThirdParty)` ItemGroup.


pjcollins:
Commit 5f79468 has successfully signed artifacts - https://dl.internalx.com/vsts-devdiv/Xamarin.Android/public/net6/4389120/sign-net6-nupkg/5f794681a045dad295119f3ea422beb1032b727a/artifacts.json. This should almost be ready to go once https://github.com/xamarin/yaml-templates/pull/96 is merged.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -1099,10 +1095,114 @@ stages:
   displayName: .NET 6 Preview Installers
   dependsOn: mac_build
   jobs:
+  # Check - "Xamarin.Android (.NET 6 Preview Installers Sign NuGets))"
```
jonathanpeppers, line 1098: I think you might have `))` on accident.

build-tools/create-dotnet-msi/create-dotnet-msi.csproj
```diff
@@ -9,6 +9,13 @@
   <Import Project="..\..\Configuration.props" />
   <Import Project="..\scripts\XAVersionInfo.targets" />
 
+  <ItemGroup>
+    <PackageReference Include="Microsoft.VisualStudioEng.MicroBuild.Core" Version="0.4.1">
```
jonathanpeppers, line 13: Since the `0.4.1` version is in a few places, should we put a `$(MicroBuildVersion)` property in `Configuration.props`?

build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
```diff
@@ -78,6 +85,10 @@
     <ItemGroup>
       <FileWrites Include="$(PkgProductOutputPath)" />
       <None Include="$(PkgProductOutputPath)" CopyToOutputDirectory="PreserveNewest" Link="Microsoft.NET.Workload.Android-$(AndroidPackVersionLong).pkg" />
+      <FilesToSign Include="$(OutputPath)\Microsoft.NET.Workload.Android-$(AndroidPackVersionLong).pkg">
```
jonathanpeppers, line 88: `$(OutputPath)` should always end with `\`, so this might be putting two `\\`.

build-tools/create-packs/Directory.Build.props
```diff
@@ -0,0 +1,12 @@
+<Project> 
+  <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../'))" />
```
jonathanpeppers, line 2: I think you could put this instead and it would be clearer:

```suggestion
  <Import Project="..\..\Directory.Build.props'))" />
```

`<Import/>` should look relative to the current file.

build-tools/installers/create-installers.targets
```diff
@@ -432,7 +432,6 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Installer.Common.dll" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Installer.Common.pdb" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Installer.Common.targets" />
-    <_MSBuildFiles Include="$(MSBuildSrcDir)\K4os.Compression.LZ4.dll" />
```
jonathanpeppers, line 435: Were the two assemblies here listed twice?

build-tools/scripts/DotNet.targets
```diff
@@ -26,6 +26,12 @@
         Projects="$(_Root)build-tools\create-dotnet-msi\create-dotnet-msi.csproj"
         Properties="Configuration=$(Configuration)"
     />
+    <MSBuild
+        Condition=" $([MSBuild]::IsOSPlatform('osx')) "
+        Targets="Restore;Build"
+        Projects="$(_Root)build-tools\create-dotnet-pkg\sign-dotnet-pkg.proj"
+        Properties="Configuration=$(Configuration)"
+    />
```
jonathanpeppers, line 34: Does this need to check some MSBuild property to skip it? Would it attempt to sign packages on Mac locally otherwise?

build-tools/automation/yaml-templates/install-signing-plugin.yaml
```diff
@@ -0,0 +1,57 @@
+parameters:
```
jonathanpeppers, line 1: Is it worth putting `-plugin` in the file name? I know they named it `MicroBuildSigningPlugin`, but it seems like they picked a weird name for a `.yaml` template.

jonathanpeppers (commented):
Otherwise looks great! I tested the Windows `.msi` and it is signed!

build-tools/installers/create-installers.targets
pjcollins: Yeah, I had an initial take where I was marking third party items in this file and found a couple duplicates that should be cleaned up as part of that.

build-tools/create-packs/Directory.Build.props
pjcollins: I was following the recommendation here: https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019#use-case-multi-level-merging

But I agree this is probably more complex than needed, since we know which "outer" file we want to import.

build-tools/scripts/DotNet.targets
pjcollins: I'll add some info to the description about this. None of the MicroBuild stuff will run unless the `MicroBuildOverridePluginDirectory` or `MicroBuildPluginDirectory` properties are set and pointing to a valid directory. Building this project locally will only codesign and harden the mach-o files that are going in the .pkg file, which can be done without specifying a signing identity (no Apple certs required on disk).

build-tools/automation/yaml-templates/install-signing-plugin.yaml
pjcollins: I'll update these names to `[install][remove]-microbuild-tooling`, which will hopefully provide better clarity on what the templates are doing.

build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
pjcollins: I am not sure if this is being added automatically, but it currently doesn't contain a trailing slash: https://github.com/xamarin/xamarin-android/blob/master/build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj#L5

I could update the rest of the project file to do this but it's maybe not worth the diff?

build-tools/automation/yaml-templates/install-signing-plugin.yaml
```diff
@@ -0,0 +1,57 @@
...
+  signType: $(MicroBuildSignType)
+  condition: succeeded()
+
+steps:
+- powershell: |
+    # We require at a minimum version 2.8.0 of the azure cli installed for the ESRP signing
```
trevors20: FYI. As of yesterday, this step was put into the templates repo.  If you wanted to start using these steps from there.

azure-tools/az-client-update.yml

trevors20 (approved):

build-tools/automation/yaml-templates/install-signing-plugin.yaml
pjcollins: Good to know, I've updated this.

build-tools/create-dotnet-pkg/create-dotnet-pkg.csproj
jonathanpeppers: Ok maybe only Microsoft.NET.Sdk adds the `\` and not Microsoft.Build.NoTargets.

If you don't see `\\` in the log, this should be fine.

pjcollins:
@trevors20 The `az-client-update` template hit an issue on this latest run - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4397974&view=logs&jobId=a62e275d-46d5-5c3b-5a47-c4e00ae7427d&j=a62e275d-46d5-5c3b-5a47-c4e00ae7427d&t=58b91e55-0185-54d7-8ced-025ce63f4b7f.

```
Error: No similarly named formulae found.
Error: No available formula or cask with the name "builder/azure-cli/azure-cli@2.8.0".
==> Searching for a previously deleted formula (in the last month)...
Error: No previously deleted formula found.
brew link azure-cli@2.8.0
Error: No such keg: /usr/local/Cellar/azure-cli@2.8.0
```


It looks like this is also being worked on in https://github.com/xamarin/xamarin-android/pull/5509, maybe I should wait for that to land first?

pjcollins:
Unsigned PR artifacts from commit c90b97b:
https://dl.internalx.com/vsts-devdiv/Xamarin.Android-PR/public/net6/4404211/merge/329563c7657077b99ed10bfd7a5848ea8751b8b7/artifacts.json

build-tools/automation/azure-pipelines.yaml
```diff
@@ -67,6 +67,11 @@ variables:
   NUnit.NumberOfTestWorkers: 4
   GitHub.Token: $(github--pat--vs-mobiletools-engineering-service2)
   CONVERT_JAVADOC_TO_XMLDOC: $[ne(variables['Build.DefinitionName'], 'Xamarin.Android-PR')]
+  ${{ if eq(variables['Build.DefinitionName'], 'Xamarin.Android-PR') }}:
+    MicroBuildSignType: Test
+  ${{ if and(eq(variables['Build.DefinitionName'], 'Xamarin.Android'), ne(variables['Build.Reason'], 'PullRequest')) }}:
+    MicroBuildSignType: Real
```
jonathanpeppers, line 73: Is there a way to put an `else` to make sure `$(MicroBuildSignType)` is set to `Test` if it were blank?

No idea if this syntax works:
```yaml
 ${{ if and(eq(variables['Build.DefinitionName'], 'Xamarin.Android'), ne(variables['Build.Reason'], 'PullRequest')) }}:
    MicroBuildSignType: Real
 ${{ else }}
    MicroBuildSignType: Test
```

jonathanpeppers (approved):
Otherwise LGTM 👍

build-tools/automation/azure-pipelines.yaml
pjcollins: I am fairly sure there's no way to do this inline in yaml, at least it's not documented anywhere that I can find:
https://docs.microsoft.com/en-us/azure/devops/pipelines/process/expressions?view=azure-devops#conditionally-assign-a-variable

jonathanpeppers: Should we put an `if` that has the expression inverted?
```yaml
 ${{ if and(eq(variables['Build.DefinitionName'], 'Xamarin.Android'), ne(variables['Build.Reason'], 'PullRequest')) }}:
    MicroBuildSignType: Real
 ${{ if or(ne(variables['Build.DefinitionName'], 'Xamarin.Android'), eq(variables['Build.Reason'], 'PullRequest')) }}:
    MicroBuildSignType: Test
```

pjcollins: I've made this change, seems reasonable to make sure some value will be set here if we introduce any new pipelines using this file in the future.

PR #5531, opened 2021-01-21
[xaprepare] remove dependency on the windep homebrew recipe on mac

grendello:
We've been seeing the attempts to install the recipe time out on our CI
servers:

    Error: Process '/usr/local/bin/brew "install" "xamarin/xamarin-android-windeps/mingw-zlib"' timed out after 00:30:00
    Error: Installation of xamarin/xamarin-android-windeps/mingw-zlib failed

Since the recipe existed only for the benefit of cross-building Mono for
Windows on macOS using mingw and we no longer do that, we can remove the
recipe.

### DIFF
--- a/build-tools/xaprepare/xaprepare/ConfigAndData/Dependencies/MacOS.cs
+++ b/build-tools/xaprepare/xaprepare/ConfigAndData/Dependencies/MacOS.cs
@@ -24,7 +24,6 @@ namespace Xamarin.Android.Prepare
 
 			new HomebrewProgram ("ninja"),
 			new HomebrewProgram ("p7zip", "7za"),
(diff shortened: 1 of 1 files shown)


jonathanpeppers (approved):
LGTM, if it builds 👍

radekdoulik (approved):

radekdoulik:
Now failing with

    Error: Installation of mingw-w64 failed

I guess we don't need it as well, right?

grendello:
> Now failing with
> 
> ```
> Error: Installation of mingw-w64 failed
> ```
> 
> I guess we don't need it as well, right?

It's unrelated. Homebrew doesn't support macOS 10.13 on which the build runs, thus this error. And we do need `mingw64` - we build our Windows stuff with it :)

PR #5536, opened 2021-01-22
[tools] Add a tool to decompress XA-compressed assemblies

grendello:
The utility is named `decompress-assemblies` and takes as its
parameters a list of files to decompress.  The files might be either
`.apk` archives or `.dll` files.

I case of `.apk`, managed assemblies stored in the `assemblies/`
directory inside the APK will be read and decompressed.

The uncompressed result is stored in the current directory with the
original assembly name modified to contain the `uncompressed-` prefix,
in order to prevent accidental overwriting of assemblies found in the
current directory.

### DIFF
--- a/tools/decompress-assemblies/decompress-assemblies.csproj
+++ b/tools/decompress-assemblies/decompress-assemblies.csproj
@@ -1,0 +1,30 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">  
+  <PropertyGroup>
+    <Company>Microsoft Corporation</Company>
+    <Copyright>2021 Microsoft Corporation</Copyright>
+    <Version>0.0.1</Version>
+    <TargetFramework>net472</TargetFramework>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <RootNamespace>Xamarin.Android.Tools.DecompressAssemblies</RootNamespace>
+    <AssemblyName>decompress-assemblies</AssemblyName>
+    <OutputPath>../../bin/$(Configuration)/bin</OutputPath>
+    <OutputType>Exe</OutputType>
+    <LibZipSharpBundleAllNativeLibraries>true</LibZipSharpBundleAllNativeLibraries>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <Import Project="..\..\Configuration.props" />
+
+  <ItemGroup>
+    <PackageReference Include="Xamarin.LibZipSharp" Version="$(LibZipSharpVersion)" />
+    <PackageReference Include="K4os.Compression.LZ4" Version="$(LZ4PackageVersion)" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="..\scripts\decompress-assemblies">
+      <Link>decompress-assemblies</Link>
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+</Project>
--- a/tools/decompress-assemblies/main.cs
+++ b/tools/decompress-assemblies/main.cs
@@ -1,0 +1,139 @@
+using System;
+using System.Buffers;
+using System.IO;
+
+using K4os.Compression.LZ4;
+using Xamarin.Tools.Zip;
+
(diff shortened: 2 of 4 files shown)


tools/decompress-assemblies/decompress-assemblies.csproj
```diff
@@ -0,0 +1,31 @@
...
+    <Copyright>2021 Microsoft Corporation</Copyright>
+    <Version>0.0.1</Version>
+    <TargetFramework>net472</TargetFramework>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <RootNamespace>Xamarin.Android.Tools.DecompressAssemblies</RootNamespace>
+    <AssemblyName>decompress-assemblies</AssemblyName>
+    <OutputPath>../../bin/$(Configuration)/bin</OutputPath>
+    <OutputType>Exe</OutputType>
+    <LibZipSharpBundleAllNativeLibraries>true</LibZipSharpBundleAllNativeLibraries>
+    <LangVersion>$(LangVersion)</LangVersion>
```
jonathanpeppers, line 14: I don't think you need to set `$(LangVersion)` with `$(LangVersion)`.

This imports `Configuration.props`, so I think it would get the value you want;

https://github.com/xamarin/xamarin-android/blob/1877dd5ebee90452c6ab2cc33a58498a83ea7c1f/Configuration.props#L38

tools/decompress-assemblies/main.cs
```diff
@@ -0,0 +1,126 @@
...
+			foreach (string file in args) {
+				string ext = Path.GetExtension (file);
+				if (String.Compare (".dll", ext, StringComparison.OrdinalIgnoreCase) == 0) {
+					if (!UncompressDLL (file)) {
+						haveErrors = true;
+					}
+					continue;
+				}
+
+				if (String.Compare (".apk", ext, StringComparison.OrdinalIgnoreCase) == 0) {
```
radekdoulik, line 115: This can be handy to extend to handle `.aab` files as well.

jonpryor:
Should we add this tool to the installer?

jpobst:
I would suggest the opposite: remove it from `xamarin-android`, put it in a separate repo, and make it available as a `dotnet` tool like we do for `apk-diff`.  (https://github.com/radekdoulik/apkdiff)

I think the last thing we want to do it add another one-off tool to our maintenance burden.

jonpryor:
> I think the last thing we want to do it add another one-off tool to our maintenance burden.

There's two related considerations, which also apply to PR #5541: 

1. These utilities are as much for *our own* use as for anyone else.  *We* need them for diagnostic purposes.
2. Where the utilities are located doesn't change the fact that we need to maintain them, as per (1).

I think the more important question is one of "tying": if "something internal" changes in this repo, will the tool need to be updated?

In the case of `decompress-assemblies` or `tmt` (PR #5541), the answer is "Yes!".  If the compression format changes, the the tool needs an update.  If the typemap format changes, then the tool needs an update.

In *both* cases, having the utility in the same tree makes it *easier* to review the changes and ensure that the tool continues to work.

Compare to `apkdiff`, which doesn't rely as much on xamarin-android internals.

tools/decompress-assemblies/main.cs
```diff
@@ -0,0 +1,126 @@
...
+			using (var fs = File.Open (filePath, FileMode.Open, FileAccess.Read)) {
+				return UncompressDLL (fs, filePath);
+			}
+		}
+
+		static bool UncompressFromAPK (string filePath)
+		{
+			using (ZipArchive apk = ZipArchive.Open (filePath, FileMode.Open)) {
+				foreach (ZipEntry entry in apk) {
+					if (!entry.FullName.StartsWith ("assemblies/", StringComparison.Ordinal)) {
```
dellis1972, line 80: I think if we changed this one line to be `Contains` rather than `StartsWith` this tool could support `aab` files as well.
(based on my knowledge of aab files). 
@jonathanpeppers my understanding is the assemblies in `aab` will be in `root/assemblies` correct?

grendello: @dellis1972 good call! Support for AAB added :)

tools/decompress-assemblies/main.cs
grendello: Support for AAB added :)

dellis1972 (approved):

jonathanpeppers (approved):

PR #5539, opened 2021-01-22
[One .NET] AOT support

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/6052

Helpful reading:

* https://github.com/dotnet/runtime/blob/15dec9a2aa5a4236d6ba70de2e9c146867b9d2e0/src/tasks/AotCompilerTask/MonoAOTCompiler.cs
* https://github.com/dotnet/runtime/blob/15dec9a2aa5a4236d6ba70de2e9c146867b9d2e0/src/mono/netcore/nuget/Microsoft.NET.Runtime.MonoAOTCompiler.Task/README.md

To make this work, I moved the existing `<Aot/>` MSBuild task calls to
a new `_AndroidAot` MSBuild target in `Xamarin.Android.Legacy.targets`.

In the .NET 6 targets, there is a *different* `_AndroidAot` MSBuild
target that runs the new `<MonoAOTCompiler/>` MSBuild task. The
`_AndroidAot` target runs per `$(RuntimeIdentifier)` after the linker
completes. Native libraries are added to the
`@(ResolvedFileToPublish)` item group to be included in the final
`.apk`.

To follow convention with Blazor WASM:

https://devblogs.microsoft.com/aspnet/asp-net-core-updates-in-net-6-preview-4/#blazor-webassembly-ahead-of-time-aot-compilation

The `$(RunAOTCompilation)` MSBuild property enables AOT. To help with
backwards compatibility with "legacy" Xamarin.Android, I kept the
`$(AotAssemblies)` MSBuild property intact.

Unfortunately, we have to manually import things to support
`$(AotAssemblies)`:

    <ImportGroup Condition=" '$(MonoAOTCompilerTasksAssemblyPath)' == '' and '$(AotAssemblies)' == 'true' ">
      <Import Project="Sdk.props" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task" />
      <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x86" />
      <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x64" />
      <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm" />
      <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm64" />
    </ImportGroup>

Since, the default Mono workload does not support `$(AotAssemblies)`:

https://github.com/dotnet/runtime/blob/69711860262e44458bbe276393ea3eb9f7a2192a/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.targets.in#L20-L25

I think this is reasonable for now.

## Limitations ##

* Profiled AOT is not implemented yet:

https://github.com/xamarin/xamarin-android/issues/6053

* `$(EnableLLVM)` fails when locating `opt`:

https://github.com/dotnet/runtime/issues/56386

* `AndroidClientHandler` usage causes runtime crash:

https://github.com/dotnet/runtime/issues/56315

* Use of folder names like `Build AndÜmläüts` fails:

https://github.com/dotnet/runtime/issues/56163

## Results ##

All tests were running on a Pixel 5.

Using the HelloAndroid app:

https://github.com/dotnet/maui-samples/tree/main/HelloAndroid

Defaults to two architectures: arm64 and x86

Average of 10 runs with `-c Release` and no AOT:

    Activity: Displayed     00:00:00.308

Apk size: 8367969

Average of 10 runs with `-c Release -p:RunAOTCompilation=true`:

    Activity: Displayed     00:00:00.209

Apk size: 12082123

Using the HelloMaui app:

https://github.com/dotnet/maui-samples/tree/main/HelloMaui

Defaults to two architectures: arm64 and x86

Average of 10 runs with `-c Release` and no AOT:

    Activity: Displayed     00:00:01.117

Apk size: 16272964

Average of 10 runs with `-c Release -p:RunAOTCompilation=true`:

    Activity: Displayed     00:00:00.568

Apk size: 42869016

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
@@ -1,0 +1,87 @@
+<!--
+***********************************************************************************************
+Microsoft.Android.Sdk.Aot.targets
+
+.NET 6 AOT support. You can find "legacy" Xamarin.Android AOT support
+in Xamarin.Android.Legacy.targets.
+
+For <MonoAOTCompiler/> usage, see:
+* https://github.com/dotnet/runtime/blob/15dec9a2aa5a4236d6ba70de2e9c146867b9d2e0/src/tasks/AotCompilerTask/MonoAOTCompiler.cs
+* https://github.com/dotnet/runtime/blob/15dec9a2aa5a4236d6ba70de2e9c146867b9d2e0/src/mono/netcore/nuget/Microsoft.NET.Runtime.MonoAOTCompiler.Task/README.md
+
+These targets are running within the _ComputeFilesToPublishForRuntimeIdentifiers target.
+They run in a context of an inner build with a single $(RuntimeIdentifier).
+
+***********************************************************************************************
+-->
+<Project>
+
+  <!--
+    NOTE: currently, the only way to allow $(AotAssemblies) in
+    .csproj files is to import these in the Android workload
+    when $(MonoAOTCompilerTasksAssemblyPath) is blank:
+    https://github.com/dotnet/runtime/blob/69711860262e44458bbe276393ea3eb9f7a2192a/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.targets.in#L20-L25
+  -->
+  <ImportGroup Condition=" '$(MonoAOTCompilerTasksAssemblyPath)' == '' and '$(AotAssemblies)' == 'true' ">
+    <Import Project="Sdk.props" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task" />
+    <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x86" />
+    <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x64" />
+    <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm" />
+    <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm64" />
+  </ImportGroup>
+
+  <UsingTask TaskName="Xamarin.Android.Tasks.GetAotArguments" AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
+
+  <Target Name="_AndroidAotInputs">
+    <ItemGroup>
+      <_AndroidAotInputs Include="@(ResolvedFileToPublish)" Condition=" '%(Extension)' == '.dll' " />
+    </ItemGroup>
+  </Target>
+
+  <Target Name="_AndroidAot"
+      Condition=" '$(AotAssemblies)' == 'true' and '$(RuntimeIdentifier)' != '' "
+      DependsOnTargets="_AndroidAotInputs"
+      Inputs="@(_AndroidAotInputs)"
+      Outputs="$(_AndroidStampDirectory)_AndroidAot.stamp">
+    <GetAotArguments
+        AndroidAotMode="$(AndroidAotMode)"
+        AndroidNdkDirectory="$(_AndroidNdkDirectory)"
+        AndroidBinUtilsDirectory="$(AndroidBinUtilsDirectory)"
+        AndroidApiLevel="$(_AndroidApiLevel)"
+        MinimumSupportedApiLevel="$(AndroidMinimumSupportedApiLevel)"
+        AndroidSequencePointsMode="$(_SequencePointsMode)"
+        AotAdditionalArguments="$(AndroidAotAdditionalArguments)"
+        AotOutputDirectory="$(_AndroidAotBinDirectory)"
+        RuntimeIdentifier="$(RuntimeIdentifier)"
+        EnableLLVM="$(EnableLLVM)"
+        Profiles="@(_AotProfiles)">
+      <Output PropertyName="_AotArguments" TaskParameter="Arguments" />
+      <Output PropertyName="_LLVMPath" TaskParameter="LLVMPath" />
+    </GetAotArguments>
+    <ItemGroup>
+      <_MonoAOTAssemblies Include="@(_AndroidAotInputs->'%(FullPath)')" AotArguments="$(_AotArguments)" />
+    </ItemGroup>
+    <MakeDir Directories="$(IntermediateOutputPath)aot\" />
+    <MonoAOTCompiler
+        Assemblies="@(_MonoAOTAssemblies)"
+        CompilerBinaryPath="@(MonoAotCrossCompiler->WithMetadataValue('RuntimeIdentifier', '$(RuntimeIdentifier)'))"
+        DisableParallelAot="$(_DisableParallelAot)"
+        LibraryFormat="So"
+        Mode="$(AndroidAotMode)"
+        OutputDir="$(IntermediateOutputPath)aot\"
+        OutputType="Library"
+        UseAotDataFile="false"
+        UseLLVM="$(EnableLLVM)"
+        LLVMPath="$(_LLVMPath)">
+      <Output TaskParameter="CompiledAssemblies" ItemName="_AotCompiledAssemblies" />
+      <Output TaskParameter="FileWrites"         ItemName="FileWrites" />
+    </MonoAOTCompiler>
+    <Touch Files="$(_AndroidStampDirectory)_AndroidAot.stamp" AlwaysCreate="true" />
+    <ItemGroup>
+      <ResolvedFileToPublish
+          Include="@(_AotCompiledAssemblies->'%(LibraryFile)')"
+          ArchiveFileName="libaot-$([System.IO.Path]::GetFileNameWithoutExtension('%(_AotCompiledAssemblies.LibraryFile)')).so"
+      />
+    </ItemGroup>
+  </Target>
+</Project>
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
@@ -1,0 +1,272 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Threading.Tasks;
+using Microsoft.Android.Build.Tasks;
+using Microsoft.Build.Framework;
+using Xamarin.Android.Tools;
+
+namespace Xamarin.Android.Tasks
+{
+	/// <summary>
+	/// The <Aot/> task subclasses this in "legacy" Xamarin.Android.
+	/// Called directly in .NET 5 to populate %(AotArguments) metadata.
+	/// </summary>
+	public class GetAotArguments : AndroidAsyncTask
+	{
+		public override string TaskPrefix => "GAOT";
+
+		[Required]
+		public string AndroidApiLevel { get; set; }
+
+		[Required]
+		public string AndroidAotMode { get; set; }
+
+		[Required]
+		public string AotOutputDirectory { get; set; }
+
+		[Required]
+		public string AndroidBinUtilsDirectory { get; set; }
+
+		/// <summary>
+		/// Will be blank in .NET 6+
+		/// </summary>
+		public string ManifestFile { get; set; }
+
+		/// <summary>
+		/// $(AndroidMinimumSupportedApiLevel) in .NET 6+
+		/// </summary>
+		public string MinimumSupportedApiLevel { get; set; }
+
+		public string RuntimeIdentifier { get; set; }
+
+		public string AndroidNdkDirectory { get; set; }
+
+		public bool EnableLLVM { get; set; }
+
+		public string AndroidSequencePointsMode { get; set; }
+
+		public ITaskItem [] Profiles { get; set; }
+
+		public string AotAdditionalArguments { get; set; }
+
+		[Output]
+		public string Arguments { get; set; }
+
+		[Output]
+		public string LLVMPath { get; set; }
+
+		protected AotMode AotMode;
+		protected SequencePointsMode SequencePointsMode;
+		protected string SdkBinDirectory;
+
+		public static bool GetAndroidAotMode(string androidAotMode, out AotMode aotMode)
+		{
+			aotMode = AotMode.Normal;
+
+			switch ((androidAotMode ?? string.Empty).ToLowerInvariant().Trim())
+			{
+			case "":
+			case "none":
+				aotMode = AotMode.None;
+				return true;
+			case "normal":
+				aotMode = AotMode.Normal;
+				return true;
+			case "hybrid":
+				aotMode = AotMode.Hybrid;
+				return true;
+			case "full":
+				aotMode = AotMode.Full;
+				return true;
+			case "interpreter":
+				// We don't do anything here for this mode, this is just to set the flag for the XA
+				// runtime to initialize Mono in the interpreter "AOT" mode.
+				aotMode = AotMode.Interp;
+				return true;
+			}
+
+			return false;
+		}
+
+		public static bool TryGetSequencePointsMode (string value, out SequencePointsMode mode)
+		{
+			mode = SequencePointsMode.None;
+			switch ((value ?? string.Empty).ToLowerInvariant ().Trim ()) {
+			case "none":
+				mode = SequencePointsMode.None;
+				return true;
+			case "normal":
+				mode = SequencePointsMode.Normal;
+				return true;
+			case "offline":
+				mode = SequencePointsMode.Offline;
+				return true;
+			}
+			return false;
+		}
+
+		public override Task RunTaskAsync ()
+		{
+			NdkTools? ndk = NdkTools.Create (AndroidNdkDirectory, Log);
+			if (ndk == null) {
+				return Task.CompletedTask; // NdkTools.Create will log appropriate error
+			}
+
+			bool hasValidAotMode = GetAndroidAotMode (AndroidAotMode, out AotMode);
+			if (!hasValidAotMode) {
+				LogCodedError ("XA3002", Properties.Resources.XA3002, AndroidAotMode);
+				return Task.CompletedTask;
+			}
+
+			if (AotMode == AotMode.Interp) {
+				LogDebugMessage ("Interpreter AOT mode enabled");
+				return Task.CompletedTask;
+			}
+
+			TryGetSequencePointsMode (AndroidSequencePointsMode, out SequencePointsMode);
+
+			SdkBinDirectory = MonoAndroidHelper.GetOSBinPath ();
+
+			var abi = AndroidRidAbiHelper.RuntimeIdentifierToAbi (RuntimeIdentifier);
+			if (string.IsNullOrEmpty (abi)) {
+				Log.LogCodedError ("XA0035", Properties.Resources.XA0035, RuntimeIdentifier);
+				return Task.CompletedTask;
+			}
+
+			(_, string outdir, string mtriple, AndroidTargetArch arch) = GetAbiSettings (abi);
+			string toolPrefix = GetToolPrefix (ndk, arch, out _);
+
+			Arguments = string.Join (",", GetAotOptions (outdir, mtriple, toolPrefix));
+			if (EnableLLVM)
+				LLVMPath = Path.GetDirectoryName (toolPrefix);
+			return Task.CompletedTask;
+		}
+
+		protected string GetToolPrefix (NdkTools ndk, AndroidTargetArch arch, out int level)
+		{
+			level = 0;
+			return EnableLLVM
+				? ndk.GetNdkToolPrefixForAOT (arch, level = GetNdkApiLevel (ndk, arch))
+				: Path.Combine (AndroidBinUtilsDirectory, $"{ndk.GetArchDirName (arch)}-");
+		}
+
+		int GetNdkApiLevel (NdkTools ndk, AndroidTargetArch arch)
+		{
+			AndroidAppManifest manifest = null;
+			if (!string.IsNullOrEmpty (ManifestFile)) {
+				manifest = AndroidAppManifest.Load (ManifestFile, MonoAndroidHelper.SupportedVersions);
+			}
+
+			int level;
+			if (manifest?.MinSdkVersion != null) {
+				level       = manifest.MinSdkVersion.Value;
+			} else if (int.TryParse (MinimumSupportedApiLevel, out level)) {
+				// level already set
+			} else if (int.TryParse (AndroidApiLevel, out level)) {
+				// level already set
+			} else {
+				// Probably not ideal!
+				level       = MonoAndroidHelper.SupportedVersions.MaxStableVersion.ApiLevel;
+			}
+
+			// Some Android API levels do not exist on the NDK level. Workaround this my mapping them to the
+			// most appropriate API level that does exist.
+			if (level == 6 || level == 7) level = 5;
+			else if (level == 10) level = 9;
+			else if (level == 11) level = 12;
+			else if (level == 20) level = 19;
+			else if (level == 22) level = 21;
+			else if (level == 23) level = 21;
+
+			// API levels below level 21 do not provide support for 64-bit architectures.
+			if (ndk.IsNdk64BitArch (arch) && level < 21) {
+				level = 21;
+			}
+
+			// We perform a downwards API level lookup search since we might not have hardcoded the correct API
+			// mapping above and we do not want to crash needlessly.
+			for (; level >= 5; level--) {
+				try {
+					ndk.GetDirectoryPath (NdkToolchainDir.PlatformLib, arch, level);
+					break;
+				} catch (InvalidOperationException ex) {
+					// Path not found, continue searching...
+					continue;
+				}
+			}
+
+			return level;
+		}
+
+		protected (string aotCompiler, string outdir, string mtriple, AndroidTargetArch arch) GetAbiSettings (string abi)
+		{
+			switch (abi) {
+				case "armeabi-v7a":
+					return (
+						Path.Combine (SdkBinDirectory, "cross-arm"),
+						Path.Combine (AotOutputDirectory, "armeabi-v7a"),
+						"armv7-linux-gnueabi",
+						AndroidTargetArch.Arm
+					);
+
+				case "arm64":
+				case "arm64-v8a":
+				case "aarch64":
+					return (
+						Path.Combine (SdkBinDirectory, "cross-arm64"),
+						Path.Combine (AotOutputDirectory, "arm64-v8a"),
+						"aarch64-linux-android",
+						AndroidTargetArch.Arm64
+					);
+
+				case "x86":
+					return (
+						Path.Combine (SdkBinDirectory, "cross-x86"),
+						Path.Combine (AotOutputDirectory, "x86"),
+						"i686-linux-android",
+						AndroidTargetArch.X86
+					);
+
+				case "x86_64":
+					return (
+						Path.Combine (SdkBinDirectory, "cross-x86_64"),
+						Path.Combine (AotOutputDirectory, "x86_64"),
+						"x86_64-linux-android",
+						AndroidTargetArch.X86_64
+					);
+
+				// case "mips":
+				default:
+					throw new Exception ("Unsupported Android target architecture ABI: " + abi);
+			}
+		}
+
+		/// <summary>
+		/// Returns a list of parameters to pass to the --aot switch
+		/// </summary>
+		protected List<string> GetAotOptions (string outdir, string mtriple, string toolPrefix)
+		{
+			List<string> aotOptions = new List<string> ();
+
+			if (Profiles != null && Profiles.Length > 0) {
+				aotOptions.Add ("profile-only");
+				foreach (var p in Profiles) {
+					var fp = Path.GetFullPath (p.ItemSpec);
+					aotOptions.Add ($"profile={fp}");
+				}
+			}
+			if (!string.IsNullOrEmpty (AotAdditionalArguments))
+				aotOptions.Add (AotAdditionalArguments);
+			if (SequencePointsMode == SequencePointsMode.Offline)
+				aotOptions.Add ($"msym-dir={outdir}");
+			if (AotMode != AotMode.Normal)
+				aotOptions.Add (AotMode.ToString ().ToLowerInvariant ());
+
+			aotOptions.Add ("asmwriter");
+			aotOptions.Add ($"mtriple={mtriple}");
+			aotOptions.Add ($"tool-prefix={toolPrefix}");
+			return aotOptions;
+		}
+	}
+}
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
@@ -50,7 +50,7 @@ namespace Xamarin.Android.Build.Tests
 			Directory.Delete (SdkWithSpacesPath, recursive: true);
 		}
 
-		[Test, Category ("SmokeTests")]
+		[Test, Category ("SmokeTests"), Category ("ProfiledAOT")]
 		public void BuildBasicApplicationReleaseProfiledAot ()
 		{
 			var proj = new XamarinAndroidApplicationProject () {
@@ -65,7 +65,7 @@ namespace Xamarin.Android.Build.Tests
 			}
 		}
 
-		[Test, Category ("SmokeTests")]
+		[Test, Category ("SmokeTests"), Category ("ProfiledAOT")]
 		public void BuildBasicApplicationReleaseWithCustomAotProfile ()
 		{
 			var proj = new XamarinAndroidApplicationProject () {
@@ -88,7 +88,7 @@ namespace Xamarin.Android.Build.Tests
 			}
 		}
 
-		[Test]
+		[Test, Category ("ProfiledAOT")]
 		public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
 		{
 			var proj = new XamarinAndroidApplicationProject () {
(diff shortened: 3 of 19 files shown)


pjcollins:
Should we be repacking the content of the `Microsoft.NET.Runtime.MonoAOTCompiler.Task` and `Microsoft.NetCore.App.Runtime.AOT*` .nupkgs in our SDK pack? I was thinking that the AOT components would be included as workload packs; declared in `WorkloadManifest.json`, and conditionally imported via `WorkloadManifest.targets` based on project settings.

Here's a rough example/outline:
 
WorkloadManifest.json
```json
  "packs": {
    ...
    "Microsoft.NET.Runtime.MonoAOTCompiler.Task": {
      "kind": "sdk",
      "version": "$(MonoAOTCompilerVersion)"
    },
    "Microsoft.NetCore.App.Runtime.AOT.arm": {
      "kind": "sdk",
      "version": "$(MonoAOTCompilerVersion)",
      "alias-to": {
        "osx-x64": "Microsoft.NetCore.App.Runtime.AOT.osx-x64.Cross.android-arm",
        "win-x64": "Microsoft.NetCore.App.Runtime.AOT.win-x64.Cross.android-arm"
      }
    },
    "Microsoft.NetCore.App.Runtime.AOT.arm64": {
      "kind": "sdk",
      "version": "$(MonoAOTCompilerVersion)",
      "alias-to": {
        "osx-x64": "Microsoft.NetCore.App.Runtime.AOT.osx-x64.Cross.android-arm64",
        "win-x64": "Microsoft.NetCore.App.Runtime.AOT.win-x64.Cross.android-arm64"
      }
    },
    ...
  }
```

WorkloadManifest.targets
```xml
  ...
  <Import Project="Sdk.targets" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task"
      Condition=" '$(AotAssemblies)' == 'true' " />
  <Import Project="Sdk.targets" Sdk="Microsoft.NetCore.App.Runtime.AOT.arm"
      Condition=" '$(AotAssemblies)' == 'true' and '$(RuntimeIdentifier)' == 'android.21-arm' " />
  ...
```

However maybe this won't work with the current version of these packages? I think for this to work these packs would need to include that `Sdk\Sdk.[props][targets]` file that we can import in our `WorkloadManifest.targets`, which contain properties pointing to the relevant tooling.

In the case of `Microsoft.NetCore.App.Runtime.AOT*`, it looks like we'd need an `Sdk` file containing at least this:

```xml
  <PropertyGroup>
    <AotCompilerBinaryPath>$(MSBuildThisFileDirectory)..\tools\mono-aot-cross</AotCompilerBinaryPath>
  </PropertyGroup>
```

In the case of `Microsoft.NET.Runtime.MonoAOTCompiler.Task`, this package does have a `build\Microsoft.NET.Runtime.MonoAOTCompiler.Task.props`:

```xml
  <PropertyGroup>
    <MonoAOTCompilerTaskAssemblyPath>$(MSBuildThisFileDirectory)..\tools\net5.0\MonoAOTCompiler.dll</MonoAOTCompilerTaskAssemblyPath>
  </PropertyGroup>
```

However this file may not be in the right location that would allow us to import it via our `WorkloadManifest.targets` file?

The `Microsoft.NetCore.App.Runtime.AOT*` _may_ also need to be renamed for the aliasing mentioned in the .json above to work? I am not sure.

@akoeplinger what are your thoughts on this? Should we be able to consume these .nupkgs as workload packs? I think this is the only way to conditionally install the AOT pieces that each individual project needs, as per the workload design?

jonathanpeppers:
Yes, I think we could easily make `Microsoft.NET.Runtime.MonoAOTCompiler.Task` a workload pack as things are currently. We can use `<Import Project="Sdk.targets" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task" />` to use it.

It would be helpful if the `Microsoft.NetCore.App.Runtime.AOT*` packs had a `.targets` file with some way to locate the command-line tooling within. However, it would need to be based per-RID, so maybe they would have to be item groups?

```
<ItemGroup>
  <AotCompilerTool Include="$(MSBuildThisFileDirectory)..\tools\mono-aot-cross" RuntimeIdentifier="android-arm" />
</ItemGroup>
```

Each pack would have a different value for `%(RuntimeIdentifier)`?

akoeplinger:
> @akoeplinger what are your thoughts on this? Should we be able to consume these .nupkgs as workload packs? It looks like the Microsoft.NetCore.App.Runtime.AOT* components are packaged as "runtime" packs, and should perhaps be "sdk" packs? I think this is the only way to conditionally install the AOT pieces that each individual Android project may need, as per the workload design.

TBH I don't have a lot of context around what the best approach is, so if you or @jonathanpeppers think we should do something differently then sure :)

jonathanpeppers:
I reworked this to use all of the MonoAOTCompiler packages as SDK workload packs.

Building app fails as before, but here is an updated log: [msbuild.binlog.zip](https://github.com/xamarin/xamarin-android/files/5889621/msbuild.binlog.zip)

👀 
```
    547976 Warning(s)
    10 Error(s)

Time Elapsed 00:47:07.24
```

akoeplinger:
@jonathanpeppers it looks like that binlog is corrupt, I can't open it. Would you mind creating a new one?

jonathanpeppers:
@akoeplinger did you try the latest one: https://github.com/xamarin/xamarin-android/pull/5539#issuecomment-769325110 I can open it with the Windows viewer at least.

I think the first one was partial because the process was killed.

akoeplinger:
@jonathanpeppers weird, for some reason downloading the .zip on macOS makes it corrupt. I can open when I download it on Windows, taking a look now :)

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
```diff
@@ -24,6 +29,26 @@
...
+      "version": "@MONO_AOT_VERSION@"
+    },
+    "Microsoft.NetCore.App.Runtime.AOT.osx-x64.Cross.android-x64": {
+      "kind": "sdk",
+      "version": "@MONO_AOT_VERSION@"
+    },
+    "Microsoft.NetCore.App.Runtime.AOT.osx-x64.Cross.android-x86": {
+      "kind": "sdk",
+      "version": "@MONO_AOT_VERSION@"
+    },
```
marek-safar, line 51: This should use `alias-to` instead

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
```diff
@@ -24,6 +29,26 @@
       "kind": "sdk",
       "version": "@SDK_PACK_VERSION@"
     },
+    "Microsoft.NET.Runtime.MonoAOTCompiler.Task": {
+      "kind": "sdk",
+      "version": "@MONO_AOT_VERSION@"
+    },
```
marek-safar, line 35: It's missing runtime packs as well

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
```diff
@@ -24,6 +29,26 @@
...
+    "Microsoft.NetCore.App.Runtime.AOT.osx-x64.Cross.android-arm": {
+      "kind": "sdk",
+      "version": "@MONO_AOT_VERSION@"
```
marek-safar, line 38: I think this versioning is wrong. This manifest file will ship with SDK so the versioning will be done by darc

src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
```diff
@@ -0,0 +1,253 @@
...
+using Microsoft.Build.Framework;
+using Xamarin.Android.Tools;
+
+namespace Xamarin.Android.Tasks
+{
+	/// <summary>
+	/// The <Aot/> task subclasses this in "legacy" Xamarin.Android.
+	/// Called directly in .NET 5 to populate %(AotArguments) metadata.
+	/// </summary>
+	public class GetAotArguments : AndroidAsyncTask
```
marek-safar, line 14: This looks redundant, it should leverage AOT tasks from dotnet/runtime

build-tools/create-packs/Directory.Build.targets
```diff
@@ -98,6 +106,7 @@
       <_WLTemplates Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nupkgs\Microsoft.Android.Templates.*.nupkg" />
       <!-- Runtime packs are not yet supported by workloads -->
       <!-- <_WLPacks Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nupkgs\Microsoft.Android.Runtime.*.nupkg" /> -->
+      <_AotPacks Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nupkgs\microsoft.netcore.app.runtime.aot.*.cross.*.nupkg" />
```
marek-safar, line 109: Why is this neede?

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.targets
```diff
@@ -3,4 +3,6 @@
       Condition=" '$(TargetPlatformIdentifier)' == 'android' " />
   <Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk.BundleTool"
       Condition=" '$(AndroidPackageFormat)' == 'aab' " />
+  <Import Project="..\build\Microsoft.NET.Runtime.MonoAOTCompiler.Task.props" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task"
```
marek-safar, line 6: this is wrong, it's will be installed on user machine

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
jonathanpeppers: I heard from Mikayla that we should be able to drop runtime packs in  `dotnet/packs` -- If that works, I'll do that in another PR and include them in our installer.

Right now you have to have a `NuGet.config` file to get runtime packs, which is not ideal.

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
jonathanpeppers: We don't have darc in this repo at all yet. Putting the version number in `Configuration.props` follows our current pattern until that is in place.

@pjcollins will be looking into darc, I think. We can start with getting it to bump the dotnet/sdk version.

src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
jonathanpeppers: This is used to populate `%(AotArguments)`, which is set on every assembly. It looks like it is the only way pass in paths to the Android NDK, etc. such as:

```xml
<AotArguments>tool-prefix=/some/path;mtriple=arm64-v8a</AotArguments>
```

This is reusing a small amount of code from our current `<Aot/>` task for this.

build-tools/create-packs/Directory.Build.targets
jonathanpeppers: For our local build, this downloads the pack and places it in `~/android-toolchain/dotnet/packs`.

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.targets
jonathanpeppers: When we get the changes here: https://github.com/dotnet/runtime/pull/48021

This might eventually be:

```xml
<Import Project="Sdk.props" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task" Condition=" '$(AotAssemblies)' == 'true' " />
```

_Something_ has to import this file so we get `$(MonoAOTCompilerTaskAssemblyPath)` and `<UsingTask/>` declared.

Another way for that to happen would be if there was a MonoAOT Workload that imports the file above. Is that something I should be using here instead?

If there is a MonoAOTCompiler.Workload with a `WorkloadManifest.targets` file, it could import the right file and everything will work.

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
marek-safar: What? What do you want to hack the system yourself? I believe they should be included in the manifest as everything else

/cc @mhutch

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
```diff
@@ -0,0 +1,71 @@
...
+        AndroidAotMode="$(AndroidAotMode)"
+        AndroidNdkDirectory="$(_AndroidNdkDirectory)"
+        AndroidBinUtilsDirectory="$(AndroidBinUtilsDirectory)"
+        AndroidApiLevel="$(_AndroidApiLevel)"
+        ManifestFile="$(IntermediateOutputPath)android\AndroidManifest.xml"
+        AndroidSequencePointsMode="$(_SequencePointsMode)"
+        AotAdditionalArguments="$(AndroidAotAdditionalArguments)"
+        AotOutputDirectory="$(_AndroidAotBinDirectory)"
+        RuntimeIdentifier="$(RuntimeIdentifier)"
+        EnableLLVM="$(EnableLLVM)"
```
marek-safar, line 44: IIRC we ship LLVM only mode

/cc @akoeplinger

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
```diff
@@ -0,0 +1,71 @@
...
+      <_AotAssemblies Include="@(_ShrunkAssemblies->'%(FullPath)')" AotArguments="$(_AotArguments)" />
+    </ItemGroup>
+    <MonoAOTCompiler
+        AotModulesTableLanguage="$(_AotModulesTableLanguage)"
+        AotModulesTablePath="$(_AotModulesTablePath)"
+        AotProfilePath="$(_AotProfilePath)"
+        Assemblies="@(_AotAssemblies)"
+        CompilerBinaryPath="%(_AotCompilerPaths.FullPath)"
+        DisableParallelAot="$(_DisableParallelAot)"
+        LLVMPath="$(_LLVMPath)"
```
marek-safar, line 60: Why is this needed?

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
```diff
@@ -0,0 +1,71 @@
...
+        Mode="$(AndroidAotMode)"
```
marek-safar, line 61: Why is this android specific setting?

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
marek-safar: I think there is bigger problem than that. Let me think about the right setup

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
jonathanpeppers: This was our existing MSBuild property in Xamarin.Android. If there is a new one we should use instead, I would like to make both old & new work for some period of time.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
jonathanpeppers: Some of these I was listing out every available parameter:
```
     TODO:
      * I set every [Input] with a non-existent property.
```
This is why this is a draft still. 😄

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.targets
marek-safar: I think we could do that

@radical do you agree we should do it that way across all workload packs?

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
```diff
@@ -7,7 +7,12 @@
         "Microsoft.Android.Sdk",
```
marek-safar, line 7: I just noticed, the name is not correct. It should be `sdk-manifest.json`, right?

jonathanpeppers: Maybe that `sdk-manifest.json` is in a spec somewhere, but renaming the file seems to break:

```
> mv .\bin\dotnet\sdk-manifests\6.0.100\microsoft.net.workload.android\WorkloadManifest.json .\bin\dotnet\sdk-manifests\6.0.100\microsoft.net.workload.android\sdk-manifest.json
> .\bin\dotnet\dotnet.exe build .\samples\HelloMaui\HelloMaui.Android\HelloMaui.Android.csproj
...
```

I think `WorkloadManifest.json` is what is actually implemented.

This repo might be helpful if someone needs an example of a workload: https://github.com/jonathanpeppers/maui-workload

marek-safar: I see. On that note are you building the package as DotnetPlatform type ?

jonathanpeppers: It looks like we need to put `PackageType=DotnetPlatform` or `PackageType=MSBuildSdk` in our own packages. I'll check on that.

jonathanpeppers: Fixed the ones in xamarin-android here: https://github.com/xamarin/xamarin-android/pull/5636

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
jonathanpeppers: I found out recently that we can't use runtime packs in `dotnet/packs`, see: https://github.com/dotnet/sdk/issues/14044

I tried this out here, but it doesn't seem like it works yet: https://github.com/jonathanpeppers/maui-workload/commit/6bd1b0e80074c2263d0e9e49603b1815b557857c

jonathanpeppers:
To test this currently, you would need: [MonoAOTCompiler.zip](https://github.com/xamarin/xamarin-android/files/6804264/MonoAOTCompiler.zip)

Copy `MonoAOTCompiler.dll` to `~\android-toolchain\dotnet\packs\Microsoft.NET.Runtime.MonoAOTCompiler.Task\6.0.0-preview.7.*\tasks\net6.0`. This will be needed until the changes land in dotnet/runtime and flow to us.

jonpryor:
Some of the tests are failing, e.g. [BuildBasicApplicationAndAotProfileIt](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5004716&view=ms.vss-test-web.build-test-results-tab&runId=23798926&resultId=100100&paneView=attachments), because they're invoking `aprofutil`, which we *removed* from the net6 installers in commit 8ad7f520e7f7e01b77cc262e3577b93f7a5f52e7:

```
Task "Exec" (TaskId:50)
  Task Parameter:Command="/Users/runner/Library/Android/dotnet/packs/Microsoft.Android.Sdk.Darwin/30.0.100-ci.pr.gh5539.94/tools/Darwin/aprofutil"  -s -v -p 9260 -o "custom.aprof" (TaskId:50)
  "/Users/runner/Library/Android/dotnet/packs/Microsoft.Android.Sdk.Darwin/30.0.100-ci.pr.gh5539.94/tools/Darwin/aprofutil"  -s -v -p 9260 -o "custom.aprof" (TaskId:50)
  /var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/tmp775e33529bfa4d3fbd9e131a1d463088.exec.cmd: line 2: /Users/runner/Library/Android/dotnet/packs/Microsoft.Android.Sdk.Darwin/30.0.100-ci.pr.gh5539.94/tools/Darwin/aprofutil: No such file or directory (TaskId:50)
```

jonathanpeppers:
After adding some new categories, I realized this breaks legacy due to:
```
Target Name=_AndroidAot Project=UnnamedProject.csproj
    Skipping target "_AndroidAot" because it has no outputs.
    Though the target has declared its outputs, the output specification only references empty properties and/or empty item lists.
```
Looking into it...

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
jonathanpeppers: Turning on `$(EnableLLVM)` does not appear to work quite yet: https://github.com/dotnet/runtime/issues/56386

We could make it the default value when it's working, if it makes sense to do so.

jonathanpeppers:
This only has a couple runtime test failures:
```
07-27 15:40:11.583  3847  3867 I NUnit   : 1) System.NetTests.NetworkInterfacesTest.DotNetInterfacesShouldEqualJavaInterfaces (Mono.Android.NET-Tests)
07-27 15:40:11.583  3847  3867 I NUnit   : System.Net.NetworkInformation.NetworkInformationException : No such file or directory
07-27 15:40:11.583  3847  3867 I NUnit   :    at System.Net.NetworkInformation.LinuxNetworkInterface.GetLinuxNetworkInterfaces() in System.Net.NetworkInformation.dll:token 0x600009b+0x40
07-27 15:40:11.583  3847  3867 I NUnit   :    at System.Net.NetworkInformation.NetworkInterfacePal.GetAllNetworkInterfaces() in System.Net.NetworkInformation.dll:token 0x6000098+0x0
07-27 15:40:11.583  3847  3867 I NUnit   :    at System.Net.NetworkInformation.NetworkInterface.GetAllNetworkInterfaces() in System.Net.NetworkInformation.dll:token 0x600004e+0x0
07-27 15:40:11.583  3847  3867 I NUnit   :    at System.NetTests.NetworkInterfacesTest.DotNetInterfacesShouldEqualJavaInterfaces() in Mono.Android.NET-Tests.dll:token 0x6000133+0x0
07-27 15:40:11.583  3847  3867 I NUnit   :    at System.Reflection.RuntimeMethodInfo.Invoke(Object , BindingFlags , Binder , Object[] , CultureInfo ) in System.Private.CoreLib.dll:token 0x600283b+0x6a
```
Then:
```
07-27 15:40:11.583  3847  3867 I NUnit   : 2) SystemTests.AppDomainTest.DateTime_Now_Works (Mono.Android.NET-Tests)
07-27 15:40:11.583  3847  3867 I NUnit   : System.PlatformNotSupportedException : PlatformNotSupported_AppDomains
07-27 15:40:11.583  3847  3867 I NUnit   :    at System.AppDomain.CreateDomain(String ) in System.Private.CoreLib.dll:token 0x6000656+0x13
07-27 15:40:11.583  3847  3867 I NUnit   :    at SystemTests.AppDomainTest.DateTime_Now_Works() in Mono.Android.NET-Tests.dll:token 0x600007b+0xa
07-27 15:40:11.583  3847  3867 I NUnit   :    at System.Reflection.RuntimeMethodInfo.Invoke(Object , BindingFlags , Binder , Object[] , CultureInfo ) in System.Private.CoreLib.dll:token 0x600283b+0x6a
```

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5024995&view=results

I have a mistake where the test results file is wrong, looking into that:

```
##[warning]No test result files matching TestResult-Mono.Android.NET_Tests-Release-Aot.xml were found.
```

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -418,17 +425,20 @@ public void HybridAOT ([Values ("armeabi-v7a;arm64-v8a", "armeabi-v7a", "arm64-v
 		}
 
 		[Test]
-		public void NoSymbolsArgShouldReduceAppSize ([Values (true, false)] bool enableHybridAot)
+		[Category ("LLVM")]
+		public void NoSymbolsArgShouldReduceAppSize ([Values ("", "Hybrid")] string androidAotMode)
```
jonpryor, line 429: Should it be `"Normal"` instead of `""`?

jonathanpeppers: I think the test was previously leaving it blank, so I kept the behavior the same:
```diff
--			if (enableHybridAot)
--				proj.SetProperty ("AndroidAotMode", "Hybrid");
++			if (!string.IsNullOrEmpty (androidAotMode))
++				proj.SetProperty ("AndroidAotMode", androidAotMode);
```

jonpryor:
For review:

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/6052

Helpful reading:

  * https://github.com/dotnet/runtime/blob/15dec9a2aa5a4236d6ba70de2e9c146867b9d2e0/src/tasks/AotCompilerTask/MonoAOTCompiler.cs
  * https://github.com/dotnet/runtime/blob/15dec9a2aa5a4236d6ba70de2e9c146867b9d2e0/src/mono/netcore/nuget/Microsoft.NET.Runtime.MonoAOTCompiler.Task/README.md

To make this work, I moved the existing `<Aot/>` MSBuild task calls to
a new `_AndroidAot` MSBuild target in `Xamarin.Android.Legacy.targets`.

In the .NET 6 targets, there is a *different* `_AndroidAot` MSBuild
target that runs the new `<MonoAOTCompiler/>` MSBuild task. The
`_AndroidAot` target runs once per `$(RuntimeIdentifier)`, after the
linker completes.

Native libraries are added to the `@(ResolvedFileToPublish)` item
group to be included in the final `.apk`.

To follow [convention with Blazor WASM][0], the `$(RunAOTCompilation)`
MSBuild property enables AOT.

To help with backwards compatibility with "legacy" Xamarin.Android, I
kept the `$(AotAssemblies)` MSBuild property intact.

Unfortunately, we have to manually import things to support
`$(AotAssemblies)`:

	<ImportGroup Condition=" '$(MonoAOTCompilerTasksAssemblyPath)' == '' and '$(AotAssemblies)' == 'true' ">
	  <Import Project="Sdk.props" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task" />
	  <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x86" />
	  <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x64" />
	  <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm" />
	  <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm64" />
	</ImportGroup>

as the [default Mono workload does not support `$(AotAssemblies)`][1].
I think this is reasonable for now.


~~ Limitations ~~

Profiled AOT is not implemented yet:

  * https://github.com/xamarin/xamarin-android/issues/6053

`$(EnableLLVM)` fails when locating `opt`:

  * https://github.com/dotnet/runtime/issues/56386

`$(AndroidClientHandler)` usage causes runtime crash:

  * https://github.com/dotnet/runtime/issues/56315

Use of folder names like `Build AndÜmläüts` fails:

  * https://github.com/dotnet/runtime/issues/56163


~~ Results ~~

All tests:

 1. Were running on a [Google Pixel 5][2], and
 2. Enabled two architectures, arm64 and x86, and
 3. **JIT time** was average of 10 runs with `-c Release`, with the
    `Activity: Displayed` time, and
 4. **AOT time** was average of 10 runs with
    `-c Release -p:RunAOTCompilation=true` with the
    `Activity: Displayed` time.
 5. Δ values are (AOT / JIT)*100.

| Test                |      JIT time |      AOT time |  Δ time |  JIT apk size |  AOT apk size | Δ size |
| ------------------- | ------------: | ------------: | ------: | ------------: | ------------: | -----: |
| [HelloAndroid][3]   |  00:00:00.308 |  00:00:00.209 |     68% |     8,367,969 |    12,082,123 | 144.4% |
| [HelloMaui][4]      |  00:00:01.117 |  00:00:00.568 |     51% |    16,272,964 |    42,869,016 | 263.4% |

[0]: https://devblogs.microsoft.com/aspnet/asp-net-core-updates-in-net-6-preview-4/#blazor-webassembly-ahead-of-time-aot-compilation
[1]: https://github.com/dotnet/runtime/blob/69711860262e44458bbe276393ea3eb9f7a2192a/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.targets.in#L20-L25
[2]: https://store.google.com/us/product/pixel_5_specs?hl=en-US
[3]: https://github.com/dotnet/maui-samples/tree/714460431541f40570e91225e8ba4bc1fe08025f/HelloAndroid
[4]: https://github.com/dotnet/maui-samples/tree/714460431541f40570e91225e8ba4bc1fe08025f/HelloMaui
```

jonpryor:
Fixes: https://github.com/xamarin/xamarin-android/issues/6052

Helpful reading:

  * https://github.com/dotnet/runtime/blob/15dec9a2aa5a4236d6ba70de2e9c146867b9d2e0/src/tasks/AotCompilerTask/MonoAOTCompiler.cs
  * https://github.com/dotnet/runtime/blob/15dec9a2aa5a4236d6ba70de2e9c146867b9d2e0/src/mono/netcore/nuget/Microsoft.NET.Runtime.MonoAOTCompiler.Task/README.md

To make this work, I moved the existing `<Aot/>` MSBuild task calls to
a new `_AndroidAot` MSBuild target in `Xamarin.Android.Legacy.targets`.

In the .NET 6 targets, there is a *different* `_AndroidAot` MSBuild
target that runs the new `<MonoAOTCompiler/>` MSBuild task. The
`_AndroidAot` target runs once per `$(RuntimeIdentifier)`, after the
linker completes.

Native libraries are added to the `@(ResolvedFileToPublish)` item
group to be included in the final `.apk`.

To follow [convention with Blazor WASM][0], the `$(RunAOTCompilation)`
MSBuild property enables AOT.

To help with backwards compatibility with "legacy" Xamarin.Android, I
kept the `$(AotAssemblies)` MSBuild property intact.

Unfortunately, we have to manually import things to support
`$(AotAssemblies)`:

	<ImportGroup Condition=" '$(MonoAOTCompilerTasksAssemblyPath)' == '' and '$(AotAssemblies)' == 'true' ">
	  <Import Project="Sdk.props" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task" />
	  <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x86" />
	  <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x64" />
	  <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm" />
	  <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm64" />
	</ImportGroup>

as the [default Mono workload does not support `$(AotAssemblies)`][1].
I think this is reasonable for now.


~~ Limitations ~~

Profiled AOT is not implemented yet:

  * https://github.com/xamarin/xamarin-android/issues/6053

`$(EnableLLVM)` fails when locating `opt`:

  * https://github.com/dotnet/runtime/issues/56386

`$(AndroidClientHandler)` usage causes runtime crash:

  * https://github.com/dotnet/runtime/issues/56315

Use of folder names like `Build AndÜmläüts` fails:

  * https://github.com/dotnet/runtime/issues/56163


~~ Results ~~

All tests:

 1. Were running on a [Google Pixel 5][2], and
 2. Enabled two architectures, arm64 and x86, and
 3. **JIT time** was average of 10 runs with `-c Release`, with the
    `Activity: Displayed` time, and
 4. **AOT time** was average of 10 runs with
    `-c Release -p:RunAOTCompilation=true` with the
    `Activity: Displayed` time.
 5. Δ values are (AOT / JIT)*100.

| Test                |      JIT time |      AOT time |  Δ time |  JIT apk size |  AOT apk size | Δ size |
| ------------------- | ------------: | ------------: | ------: | ------------: | ------------: | -----: |
| [HelloAndroid][3]   |  00:00:00.308 |  00:00:00.209 |     68% |     8,367,969 |    12,082,123 | 144.4% |
| [HelloMaui][4]      |  00:00:01.117 |  00:00:00.568 |     51% |    16,272,964 |    42,869,016 | 263.4% |

[0]: https://devblogs.microsoft.com/aspnet/asp-net-core-updates-in-net-6-preview-4/#blazor-webassembly-ahead-of-time-aot-compilation
[1]: https://github.com/dotnet/runtime/blob/69711860262e44458bbe276393ea3eb9f7a2192a/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.targets.in#L20-L25
[2]: https://store.google.com/us/product/pixel_5_specs?hl=en-US
[3]: https://github.com/dotnet/maui-samples/tree/714460431541f40570e91225e8ba4bc1fe08025f/HelloAndroid
[4]: https://github.com/dotnet/maui-samples/tree/714460431541f40570e91225e8ba4bc1fe08025f/HelloMaui

PR #5541, opened 2021-01-25
[tools] TypeMap dumper utility

grendello:
Utility to read and dump to files/screen the typemap contents found in
`libxamarin-android.so`

### DIFF
--- a/tools/tmt/Loader.cs
+++ b/tools/tmt/Loader.cs
@@ -1,0 +1,322 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+using Xamarin.Tools.Zip;
+
+namespace tmt
+{
+	class Loader
+	{
+		const string xamarinApp = "libxamarin-app.so";
+		const string sharedLibsDirName = "app_shared_libraries";
+		const string typemapIndexFileName = "typemap.index";
+
+		static readonly string TypemapIndexRelPath = Path.Combine ("android", "typemaps", typemapIndexFileName);
+		static readonly List<string> archDirs = new List<string> {
+			"arm64-v8a",
+			"armeabi-v7a",
+			"x86",
+			"x86_64",
+		};
+
+		static readonly char dirSep = Path.DirectorySeparatorChar;
+
+		AndroidArch archFilter;
+		bool loadOnlyFirst;
+
+		public Loader (AndroidArch archFilter, bool loadOnlyFirst)
+		{
+			this.archFilter = archFilter;
+			this.loadOnlyFirst = loadOnlyFirst;
+		}
+
+		public List<ITypemap> TryLoad (string path)
+		{
+			if (File.Exists (path)) {
+				return LoadFromFile (path);
+			}
+
+			if (!Directory.Exists (path)) {
+				Log.Info ($"Directory or file '{path}' not found.");
+				return new List<ITypemap> ();
+			}
+
+			string fsPath = Path.Combine (path, typemapIndexFileName);
+			if (File.Exists (fsPath)) {
+				return LoadFromFile (fsPath);
+			}
+
+			fsPath = Path.Combine (path, xamarinApp);
+			if (File.Exists (fsPath)) {
+				return LoadFromFile (fsPath);
+			}
+
+			if (PathEndsWith (path, sharedLibsDirName)) {
+				return LoadFromSharedLibsDir (path);
+			}
+
+			// typemap.index must come first, since we always create `libxamarin-app.so`, but it will be empty if
+			// FastDev is used, which would cause an error
+			fsPath = Path.Combine (path, TypemapIndexRelPath);
+			if (File.Exists (fsPath)) {
+				return LoadFromFile (fsPath);
+			}
+
+			fsPath = Path.Combine (path, sharedLibsDirName);
+			if (Directory.Exists (fsPath)) {
+				return LoadFromSharedLibsDir (fsPath);
+			}
+
+			if (PathEndsWith (path, "obj")) {
+				return LoadFromObjDir (path);
+			}
+
+			fsPath = Path.Combine (path, "obj");
+			if (Directory.Exists (fsPath)) {
+				return LoadFromObjDir (fsPath);
+			}
+
+			Log.Warning ($"Cannot determine what to load from {path}");
+			return new List<ITypemap> ();
+		}
+
+		List<ITypemap> LoadFromObjDir (string objDirPath)
+		{
+			string fsPath;
+
+			// The reason for two loops here is that we cannot guarantee the order in which subdirectories will be
+			// reported and we must always load FastDev typemap before we attempt to load one from `libxamarin-app.so`,
+			// since the latter would be empty and error out in case FastDev is used
+			foreach (string subdir in Directory.EnumerateDirectories (objDirPath)) {
+				fsPath = Path.Combine (subdir, TypemapIndexRelPath);
+				if (File.Exists (fsPath)) {
+					// We don't need to search more - typemap.index exists only in the obj/Debug directory
+					return LoadFromFile (fsPath);
+				}
+			}
+
+			var ret = new List<ITypemap> ();
+			foreach (string subdir in Directory.EnumerateDirectories (objDirPath)) {
+				fsPath = Path.Combine (subdir, sharedLibsDirName);
+				if (!Directory.Exists (fsPath)) {
+					continue;
+				}
+
+				ret.AddRange (LoadFromSharedLibsDir (fsPath));
+			}
+
+			return ret;
+		}
+
+		List<ITypemap> LoadFromSharedLibsDir (string dirPath)
+		{
+			var ret = new List<ITypemap> ();
+			foreach (string archDir in archDirs) {
+				string fsPath = Path.Combine (dirPath, archDir, xamarinApp);
+				if (File.Exists (fsPath)) {
+					ITypemap? dsoTypemap = LoadDSOFromFilesystem (fsPath);
+					if (dsoTypemap != null) {
+						ret.Add (dsoTypemap);
+						if (loadOnlyFirst) {
+							break;
+						}
+					}
+				}
+			}
+
+			return ret;
+		}
+
+		List<ITypemap> LoadFromFile (string filePath)
+		{
+			if (String.Compare (Path.GetFileName (filePath), "typemap.index", StringComparison.OrdinalIgnoreCase) == 0) {
+				return CreateListAndReturn (LoadFromFastDevTypemap (filePath));
+			}
+
+			string ext = Path.GetExtension (filePath);
+			if (String.Compare (ext, ".typemap", StringComparison.OrdinalIgnoreCase) == 0) {
+				return CreateListAndReturn (LoadFromFastDevTypemap (filePath));
+			}
+
+			if (String.Compare (ext, ".so", StringComparison.OrdinalIgnoreCase) == 0) {
+				return CreateListAndReturn (LoadDSOFromFilesystem (filePath));
+			}
+
+			if (String.Compare (ext, ".apk", StringComparison.OrdinalIgnoreCase) == 0) {
+				return LoadFromAPK (filePath, "assemblies/");
+			}
+
+			if (String.Compare (ext, ".aab", StringComparison.OrdinalIgnoreCase) == 0) {
+				return LoadFromAPK (filePath, "base/root/assemblies/");
+			}
+
+			Log.Info ($"Unsupported file extension '{ext}', unable to load");
+			return CreateListAndReturn (null);
+
+			List<ITypemap> CreateListAndReturn (ITypemap? tm)
+			{
+				if (tm == null) {
+					return new List<ITypemap> ();
+				}
+
+				return new List<ITypemap> {
+					tm
+				};
+			}
+		}
+
+		ITypemap? LoadFromFastDevTypemap (string filePath)
+		{
+			ITypemap tm = new FastDevTypeMap ();
+
+			var fs = File.Open (filePath, FileMode.Open, FileAccess.Read);
+			try {
+				if (tm.CanLoad (fs, filePath)) {
+					return tm;
+				}
+			} catch {
+				DisposeStream (fs);
+				throw;
+			}
+
+			DisposeStream (fs);
+			return null;
+
+			void DisposeStream (FileStream fs)
+			{
+				fs.Close ();
+				fs.Dispose ();
+			}
+		}
+
+		List<ITypemap> LoadFromAPK (string filePath, string assemblyEntryPrefix)
+		{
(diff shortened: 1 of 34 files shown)


tools/tmt/Loader.cs
```diff
@@ -0,0 +1,262 @@
...
+					dsoTypemap
+				};
+			}
+
+			if (String.Compare (ext, ".apk", StringComparison.OrdinalIgnoreCase) == 0) {
+				return LoadFromAPK (filePath);
+			}
+
+			if (String.Compare (ext, ".aab", StringComparison.OrdinalIgnoreCase) == 0) {
+				throw new NotImplementedException ("AAB");
```
dellis1972, line 123: abb should be pretty similar to apk I think. I suspect the only difference is the path which has a `root` folder which then contains the `lib` folders.

grendello: Yeah, I need to take a look at one :)

radekdoulik (approved):
LGTM

It would be nice if it can later provide a library (in form of nuget) to read type maps, so that it can be used as PackageReference in other tools. Here I have apkdiff in mind, where it can show changes in the typemaps.

radekdoulik:
There's this error on CI:

    NU1605: Detected package downgrade: K4os.Compression.LZ4 from 1.2.6 to 1.1.11. Reference the package directly from the project to select a different version.

It looks like the CI just didn't pickup last commit?

tools/tmt/tmt.csproj
```diff
@@ -0,0 +1,24 @@
...
+  <PropertyGroup>
+    <Company>Microsoft Corporation</Company>
+    <Copyright>2020 Microsoft Corporation</Copyright>
+    <Version>0.0.1</Version>
+    <TargetFrameworks>netcoreapp3.1</TargetFrameworks>
+    <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
+    <OutputPath>../../bin/$(Configuration)/bin/tmt</OutputPath>
+    <OutputType>Exe</OutputType>
+    <LibZipSharpBundleAllNativeLibraries>true</LibZipSharpBundleAllNativeLibraries>
+    <LangVersion>$(LangVersion)</LangVersion>
```
jonathanpeppers, line 12: I think you can leave this line out if you like.

dellis1972:
> There's this error on CI:
> 
> ```
> NU1605: Detected package downgrade: K4os.Compression.LZ4 from 1.2.6 to 1.1.11. Reference the package directly from the project to select a different version.
> ```
> 
> It looks like the CI just didn't pickup last commit?

Nope this is because monodroid is referencing the older package. There is a PR to upgrade monodroid that will need to be included in this one.

grendello:
> LGTM
> 
> It would be nice if it can later provide a library (in form of nuget) to read type maps, so that it can be used as PackageReference in other tools. Here I have apkdiff in mind, where it can show changes in the typemaps.

That's a good idea :) I will work on it at some point - we would probably need to move the reading code to a separate repository.

tools/tmt/tmt.csproj
grendello: I keep copying it over and over... :)

.external
```diff
@@ -1,2 +1,2 @@
-xamarin/monodroid:master@ad19471b8478f9f8028c9a3517ad3ca7bca5fb4a
+xamarin/monodroid:master@390c048caaddfc6549ef94d6a80f71e5aa9416bc
```
dellis1972, line 1: can you change this to this commit d9fcaa019c5e2fb14384d47040bbe7f0b8e7695c?  its the head of master. It should also fix a few failing tests.

grendello: Done

grendello:
Working on some improvements and support for FastDev typemap files

tools/tmt/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+By default the utility reads `libxamarin-app.so` for any architecture
+it finds in the location indicated on command line.  The `-a` option
+can be used to limit the set of architectures by specifying a
+comma-separated list of architecture names from the following set
+(architecture names are case-insensitive): `ARM`, `ARM64`, `X86` and
+`X86_64`
+
+# Command line options
+
```
jonpryor, line 109: This should use a triple-backtick "block" so that it renders as intended.

Please provide an example invocation, e.g. on `samples/HelloWorld` output.

grendello: Done

dellis1972 (approved):

jonathanpeppers (approved):

tools/tmt/README.md
```diff
@@ -0,0 +1,283 @@
...
+### Dump all typemaps in a project directory
+``` shell
+# 
+# Alternative invocations with the same result:
+#
+# bin/Debug/bin/tmt/tmt samples/HelloWorld/obj
+# bin/Debug/bin/tmt/tmt samples/HelloWorld/obj/Debug
+#
+$ bin/Debug/bin/tmt/tmt samples/HelloWorld/
```
jonpryor, line 148: I built this branch, and the `bin/Debug/bin/tmt/tmt` script does not exist.

In order to run the app, I need to use `dotnet bin/Debug/bin/tmt/tmt.dll`, which is fine and works (as far as it goes), but doesn't match this doc.

Additionally, `tools/tmt/run.sh` doesn't look like it's supposed to be copied into `bin/Debug/bin/tmt/tmt`, as that contains an `msbuild /r` invocation…

Is something missing from this PR which is still on your local tree?

tools/tmt/README.md
```diff
@@ -0,0 +1,283 @@
...
+                               default action when regex filtering is used,
+                               otherwise it defaults to `false`
+  -g, --generate-files       Generate report files. If regex filtering is used,
+                               this setting defaults to `false`, it is `true`
+                               otherwise.
+
+  -v, --verbose              Show debug messages
+```
+
+## Sample invocations
```
jonpryor, line 138: Part of the request for sample invocations was to see what output was generated, which also isn't described anywhere.  (This is why I built the branch, on the hopes I could see what it was like and consider updating the commit message/`README.md`… then found that the `tmt` script didn't exist.)

Now that I have the output, I see that it's a "tab-delimited format":

```
Java Type Name	Managed Type Name	Generic?	Duplicate?
android/accessibilityservice/AccessibilityButtonController	Android.AccessibilityServices.AccessibilityButtonController, Mono.Android		
```

This is good, as far as it goes, but there is no good description of what the "Generic?" column is for, and it doesn't appear to be used in the output for `samples/HelloWorld`.

The `Duplicate?` column is presumably for when the same Java type is mapped to multiple C# types, a'la 66f72064e00499e3c87c57def3f7ca2cad07e1be, but even then I'm not quite sure how to interpret it?

```
android/accessibilityservice/AccessibilityGestureEvent	Android.AccessibilityServices.AccessibilityGestureEvent, Mono.Android		
android/accessibilityservice/AccessibilityService	Android.AccessibilityServices.AccessibilityService, Mono.Android		aliased
```

`android/accessibilityservice/AccessibilityService` has an `aliased` entry, but what's the *non*-aliased entry? Or does `aliased` mean that there is more than one C# type, *period*, and indicates nothing about *preferred* mapping?

Looking at the managed-to-Java mapping, this "no preference" interpretation appears correct, as only one entry is listed Java-to-managed (above):

```
Managed Type Name	Java Type Name	Generic?	Duplicate?
Android.AccessibilityServices.AccessibilityService, Mono.Android	android/accessibilityservice/AccessibilityService		
…
Android.AccessibilityServices.AccessibilityServiceInvoker, Mono.Android	android/accessibilityservice/AccessibilityService		
```

TL;DR: these files are also part of the usage pattern of this app.  They should be likewise described.  I'm happy to help with that description, but I can't easily fix the "missing `tmt` script" problem.

jonpryor: Oh, and of course the Release format differs, so that'll need an explanation as well:

```
Java Type Name	Managed Type Name	MVID	Token ID
…
java/lang/Object	Java.Lang.Object, Mono.Android	dc97aced-a52d-4110-bb03-b57c6368c440	33554768 (0x02000150)
```

jonpryor: I wonder if these headers should use `-` instead of space ` ` within the description, to make it easier to visually distinguish the tabs, e.g.

```
Java-Type-Name	Managed-Type-Name	MVID	Token-ID
```

tools/tmt/README.md
grendello: `tmt` is a netcore app and it produces `bin/Debug/tmt/tmt` as a native executable, it's not a script:

```shell
$ ls -l bin/Debug/bin/tmt/tmt
-rwxr-xr-x 1 grendel grendel 86584 Feb  3 22:20 bin/Debug/bin/tmt/tmt

$ file bin/Debug/bin/tmt/tmt
bin/Debug/bin/tmt/tmt: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=22531e6f94df2fbec39dd3d1ef003bcdcc0288df, stripped
```

Not sure why you don't have it in your tree.

`run.sh` will be removed, it's merely a shortcut I use for testing to avoid typing the same commands all over again.

tools/tmt/README.md
grendello: > Part of the request for sample invocations was to see what output was generated, which also isn't described anywhere. (This is why I built the branch, on the hopes I could see what it was like and consider updating the commit message/`README.md`… then found that the `tmt` script didn't exist.)
> 
Not sure why your `dotnet` doesn't produce the `tmt` binary, I would assume some sort of consistency when building the same project with the same tool...

> ```
> Java Type Name	Managed Type Name	Generic?	Duplicate?
> android/accessibilityservice/AccessibilityButtonController	Android.AccessibilityServices.AccessibilityButtonController, Mono.Android		
> ```
> 
> This is good, as far as it goes, but there is no good description of what the "Generic?" column is for, and it doesn't appear to be used in the output for `samples/HelloWorld`.
> 
Where would I put the description? The column headers should be short and I think that, considering we're looking at the listing of C# and Java types, the word would be clear to mean a `Generic type`. As for the column being empty in `samples/HelloWorld` output... well, can't do much about it, can I? :)

> The `Duplicate?` column is presumably for when the same Java type is mapped to multiple C# types, a'la [66f7206](https://github.com/xamarin/xamarin-android/commit/66f72064e00499e3c87c57def3f7ca2cad07e1be), but even then I'm not quite sure how to interpret it?
> 
There's not much to interpret here - it's what you presumed originally. Again, the assumption is that the person using the tool knows what typemaps are and how to interpret the word in the context.

> ```
> android/accessibilityservice/AccessibilityGestureEvent	Android.AccessibilityServices.AccessibilityGestureEvent, Mono.Android		
> android/accessibilityservice/AccessibilityService	Android.AccessibilityServices.AccessibilityService, Mono.Android		aliased
> ```
> 
> `android/accessibilityservice/AccessibilityService` has an `aliased` entry, but what's the _non_-aliased entry? Or does `aliased` mean that there is more than one C# type, _period_, and indicates nothing about _preferred_ mapping?
> 
Jon, what are we trying to achieve here? Do you want the output of the tool to explain how typemaps work and what they are? I hardly think this is the correct place to do it. The format and the workings are documented elsewhere and I **really** think that the utility is not for general use by every developer using `Xamarin.Android`, but for forensic examination by those who need to dive deep enough to have to learn what typemaps are (i.e. mostly the XA team) - and if somebody goes that deep to the details of what we put in the APK, I'm sure they will know what "generic", "duplicate" and "aliased" are. If you don't like these terms, please tell me what to put there as I think they're clear enough.

> Looking at the managed-to-Java mapping, this "no preference" interpretation appears correct, as only one entry is listed Java-to-managed (above):
> 
> ```
> Managed Type Name	Java Type Name	Generic?	Duplicate?
> Android.AccessibilityServices.AccessibilityService, Mono.Android	android/accessibilityservice/AccessibilityService		
> …
> Android.AccessibilityServices.AccessibilityServiceInvoker, Mono.Android	android/accessibilityservice/AccessibilityService		
> ```
> 
> TL;DR: these files are also part of the usage pattern of this app. They should be likewise described. I'm happy to help with that description, but I can't easily fix the "missing `tmt` script" problem.

I can't fix that either as I don't know why `dotnet` doesn't produce the binary on your system but creates it on mine. I will instead provide an actual script which will work in both cases.

jonpryor: > Do you want the output of the tool to explain how typemaps work and what they are? 

No, not the *output of the tool*;  The `README.md` of the tool.  Which is why this comment thread started on the `README.md`.

And again, I'd be happy to write up that description, I just wasn't sure what some of the values were supposed to mean.

For example, my initial Hello World build did no linking, and thus was *huge* (yay?), with 6062 Java-to-managed entries and 5376 Managed-to-Java entries.

This should be ~everything in `Mono.Android.dll`, right? (Perhaps not?  There are 10848 TypeDefinitions within `Mono.Android.dll`, though not every typedef is a distinct binding, so "half-ish" might be a sensible value.)

So why don't any of these entries appear to produce output in the `Is-Generic-Type?` column?  Or am I missing something?  I thought we needed the `generic` column for e.g. `Android.Runtime.JavaList<T>`?

…which oddly isn't present?!

```
% grep JavaList x/typemap-v1-Debug-X86_64-managed.txt
Android.Runtime.JavaList, Mono.Android	java/util/ArrayList		duplicate entry
```

There *is* a generic type present, but it doesn't have anything in the `Is-Generic-Type?` column:

```
% grep JavaArray x/typemap-v1-Debug-X86_64-managed.txt
Android.Runtime.JavaArray`1, Mono.Android	mono/android/runtime/JavaArray		
```

Why doesn't `JavaArray<T>` have a value for `Is-Generic-Type`?

For that matter, why is `JavaArray<T>` the only generic type in here?

```
% grep '`' x/typemap-v1-Debug-X86_64-managed.txt
Android.Runtime.JavaArray`1, Mono.Android	mono/android/runtime/JavaArray		
```

At this point I start wondering if something is wrong with my setup…?

tools/tmt/XamarinAppReleaseDSO.cs
```diff
@@ -0,0 +1,480 @@
...
+				if (m.duplicate_map != null) {
+					foreach (TypeMapModuleEntry entry in m.duplicate_map) {
+						if (entry.type_token_id == tokenID) {
+							return (true, m, false, true);
+						}
+					}
+				}
+
+				Log.Error ($"Module {m.assembly_name} ({m.module_uuid}) at index {moduleIndex} doesn't contain an entry for managed type with token ID {tokenID}");
+				return (false, null, false, false);
```
jonpryor, line 341: …and *this* is presumably why I don't see any "generic" types (perhaps): `isGeneric` is always false.

Though I think this is only for the Release typemap, not the debug typemap, so…? ¯\_(ツ)_/¯

PR #5548, opened 2021-01-26
[Xamarin.Android.Build.Tasks] _AddMultiDexDependencyJars only runs for apps

jonathanpeppers:
Context: https://github.com/xamarin/AndroidX/pull/247

When porting AndroidX to .NET 6, I noticed something odd... I was
getting `.aar` files in the build output that appeared to contain the
contents of `android-support-multidex.jar`.

Then I noticed:

https://github.com/xamarin/AndroidX/blob/16b02ae51980f7bab7be395d44f9998eea9bf32f/source/AndroidXProject.cshtml#L18

The AndroidX binding projects are setting `$(AndroidEnableMultiDex)`?

This is surely not intentional to be set in a class library, but this
causes `_AddMultiDexDependencyJars` to run:

    <Target Name="_AddMultiDexDependencyJars">
      
      <ItemGroup Condition=" '$(AndroidEnableMultiDex)' == 'True' AND '$(AndroidMultiDexSupportJar)' == '' ">
        <AndroidJavaLibrary Include="$(MonoAndroidToolsDirectory)\android-support-multidex.jar" />
      </ItemGroup>
    </Target>

If this was added to `@(AndroidJavaLibrary)`, then the `<CreateAar/>`
MSBuild task would happily package it in a .NET 6 class library!

I think the solution here is that the `_AddMultiDexDependencyJars`
MSBuild target shouldn't actually run unless `$(AndroidApplication)`
is `true`.

I updated an existing test for this scenario.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -175,9 +175,30 @@ namespace Xamarin.Android.Build.Tests
 			//NOTE: these properties should not affect class libraries at all
 			proj.SetProperty ("AndroidPackageFormat", "aab");
 			proj.SetProperty ("AotAssemblies", "true");
+			proj.SetProperty ("AndroidEnableMultiDex", "true");
 			using (var b = CreateDllBuilder ()) {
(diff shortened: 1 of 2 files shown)


dellis1972 (approved):

jonathanpeppers:
The couple test failures look like they are related to FastDev, which hopefully #5544 will help.

Going to merge this one.

PR #5551, opened 2021-01-27
[MSBuildDeviceIntegration] Give each test a unique package name.

dellis1972:
One of the problems we might see on CI is when two tests run at the same
time on the same emulator. This can cause collisions and tests to fail
because the app was changed while the test was running.
Currently ALL of the tests user `UnnamedProject.UnnamedProject` as a
package name. This means they overwrite each other.

The commit introduces a new system which will
auto generate a package name based on the test name itself.
We do this by using the `[CallerMethod]` attribute to find out
which method called the constructor of the
`XamarinAndroidApplicationProject`. This is then used to construct
a unique (ish) package name in the form

```
com.xamarin.<TestMethodName>
``` 

This should stop collisions. As part of the process we also register each
packageName used with a static cache. We then uninstall those from the 
device when the `OnTimeTearDown` method is called. This means the 
device will be cleaned as groups of tests are run.


tests/MSBuildDeviceIntegration/Tests/AotProfileTests.cs
```diff
@@ -22,7 +22,10 @@ public void BuildBasicApplicationAndAotProfileIt ()
 		{
 			AssertHasDevices ();
 
-			var proj = new XamarinAndroidApplicationProject () { IsRelease = true };
+			var proj = new XamarinAndroidApplicationProject () {
+				IsRelease = true,
+				PackageName = GeneratePackageName (),
```
radekdoulik, line 27: When do we want to set this? For all "ondevice" tests?

dellis1972: I would say yes for all on Device tests, unless you manually specify a package name yourself. There are some on device tests which use a specific packageName.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
```diff
@@ -43,12 +43,30 @@ protected override void CleanupTest ()
...
+			if (TestPackageName.TryGetValue (TestContext.CurrentContext.Test.ID, out string packageName))
+				RunAdbCommand ($"uninstall {packageName}");
+		}
+
+		protected static string GeneratePackageName ()
+		{
+			if (TestPackageName.TryGetValue (TestContext.CurrentContext.Test.ID, out string packageName))
+				return packageName;
+
+			packageName = $"com.xamarin.test{TestContext.CurrentContext.Test.ID.Replace ("-", "")}";
```
jonathanpeppers, line 65: What is the value of `Test.ID` compared to `Test.Name`? Is it a GUID?

jonpryor:
There's one current test failure in [`Xamarin.Android.Build.Tests.BuildTest.MissingOrgApacheHttpClient("dx")`](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4466115&view=ms.vss-test-web.build-test-results-tab&runId=18962176&resultId=100179&paneView=attachments):

```
…/NuGet.targets(128,5): error : Could not find file '/Users/runner/.nuget/packages/xamarin.googleplayservices.base/117.1.0/1cev9rpg.52e'. [/Users/runner/work/1/s/bin/TestRelease/temp/MissingOrgApacheHttpClientdx/UnnamedProject.csproj]
```

This "feels like" a possible "concurrent NuGet restore" scenario, and the declaring test fixture doesn't prevent that: 

https://github.com/xamarin/xamarin-android/blob/214dd76d9441c9974775f0b29feec0d5d8d1097f/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L23

Restarting the tests.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
```diff
@@ -43,12 +43,30 @@ protected override void CleanupTest ()
...
 
+		protected static void UnInstallTestApp ()
+		{
+			if (TestPackageName.TryGetValue (TestContext.CurrentContext.Test.ID, out string packageName))
+				RunAdbCommand ($"uninstall {packageName}");
+		}
+
+		protected static string GeneratePackageName ()
+		{
+			if (TestPackageName.TryGetValue (TestContext.CurrentContext.Test.ID, out string packageName))
```
jonpryor, line 62: Since `TestPackage` is a `ConcurrentDictionary<TKey, TValue>`, shouldn't this use [`.GetOrAdd()`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd?view=net-5.0#System_Collections_Concurrent_ConcurrentDictionary_2_GetOrAdd__0_System_Func__0__1__)?

```csharp
return TestPackageName.GetOrAdd (TestContext.CurrentContext.Test.ID, key => {
    return $"com.xamarin.test{key.Replace ("-", "")}";
});
```

tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
```diff
@@ -39,6 +39,7 @@ public void ApplicationRunsWithoutDebugger ([Values (false, true)] bool isReleas
 
 			var proj = new XamarinFormsAndroidApplicationProject () {
 				IsRelease = isRelease,
+				PackageName = GeneratePackageName (),
```
jonpryor, line 42: Given how many places this is set, shouldn't we just change the *default* value?

https://github.com/xamarin/xamarin-android/blob/214dd76d9441c9974775f0b29feec0d5d8d1097f/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidApplicationProject.cs#L70

Aside: ^^ is wrong, in that `PackageName` *cannot* (reasonably) be set before the `XamarinAndroidApplicationProject` constructor executes (short of using Reflection to construct an uninitialized object, set the property, and the invoke the constructor…), and thus it'll always hit the `string.Format ("{0}.{0}", ProjectName)` expression.

Couldn't this instead just become:

```csharp
PackageName = GeneratePackageName ();
```

Is there really a problem if *every* project gets this new behavior?

dellis1972: The main reason I wanted to avoid this was the new method has a dependency on Unit. `Xamarin.ProjectTools` currently has not such dependency. I didn't want to add it unless we had a really good reason too.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
dellis1972: Didn't even know that method existing lol. Will make that change.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
dellis1972: We end up with id's like `com.xamarin.test01022`. My main concern with using the `Test.Name` is it includes all the arguments and even with replacing invalid chars we'd end up with something like 
```
com.xamarin.ApplicationRunsWithDebuggerAndBreaksTrueAssembliesFalsenull
```
which I suppose isn't that bad.

tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
dellis1972: I'm rewording the code

PR #5559, opened 2021-01-28
[build-tools/scripts] fixes for nunit3-console.cmd

jonathanpeppers:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4414611&view=logs&j=674b6335-1d71-5535-2f20-11f5708a0b9d&t=358dacbe-f0d4-5f52-9155-e5dfefeddb0c&l=67

Our Windows NUnit tests do not appear to be running at all:

    Microsoft Windows [Version 10.0.17763.1282]
    (c) 2018 Microsoft Corporation. All rights reserved.

    C:\a\2\s>

It just seems to launch a `cmd` prompt and continue on?

I could reproduce this locally, which I could fix by using `cmd /c`:

https://docs.microsoft.com/windows-server/administration/windows-commands/cmd

Then I hit the next problem:

    > & cmd /c 'C:\src\xamarin-android\build-tools/scripts/nunit3-console.cmd' bin\TestDebug\net472\Xamarin.Android.Build.Tests.dll --test foo
    The system cannot find the path specified.

Reviewing `nunit3-console.cmd`, it seems to parse this output:

    > nuget locals -list global-packages
    global-packages: C:\Users\jopepper\.nuget\packages\

It splits on the `:`, but unfortunately `C:\` contains `:`. The second
item also has a leading space. This logic didn't work for me locally.

I could update the script to split on the space instead and things
work properly now.

### DIFF
--- a/build-tools/automation/yaml-templates/run-nunit-tests.yaml
+++ b/build-tools/automation/yaml-templates/run-nunit-tests.yaml
@@ -25,7 +25,7 @@ steps:
            & $env:USERPROFILE\android-toolchain\dotnet\dotnet.exe test ${{ parameters.testAssembly }} --results-directory . --logger "trx;LogFileName=${{ parameters.testResultsFile }}" ${{ parameters.dotNetTestExtraArgs }} -- NUnit.NumberOfTestWorkers=${{ parameters.workers }}
         } else {
            Write-Host '##vso[task.setvariable variable=TestResultsFormat]NUnit'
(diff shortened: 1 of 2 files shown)


grendello (approved):

radekdoulik (approved):
`for /f "tokens=1,2" %a in ('"%NUGET_PATH%" locals -list global-packages') do set PACKAGES_PATH=%b` works OK for me locally too

Issue #5562, opened 2021-01-28
Crash in LinkAssembliesNoShrink when using AddKeepAlivesStep


radekdoulik:
Context: https://github.com/xamarin/GooglePlayServicesComponents/pull/419#issuecomment-769411572

`AddKeepAlivesStep` uses `Context` to get `System.GC.KeepAlive` method. When called from `LinkAssembliesNoShrink`, the context is not available.

We need to create new class based on `AddKeepAlivesStep` and provide method (or assembly lookup) to the base class, similar to how we do it for `FixAbstractMethodsStep`. This will allow us to avoid using `Context` there.

We should also add test in XA to check this scenario.

mehmetavsar:
Would you suggest any workaround for this?

PR #5563, opened 2021-01-28
Fix LinkAssembliesNoShrink/AndroidAddKeepAlives crash

radekdoulik:
Context: https://github.com/xamarin/xamarin-android/issues/5562

Add `AddKeepAlivesStep` class to `LinkAssembliesNoShrink` task. This class overrides linker's `AddKeepAlivesStep` and provides corlib assembly without accessing linker context. It also provides logging for the step.

Test more scenarios with `AndroidAddKeepAlives`. Two of these scenarios led to a crash in the linker step before the fix - `Debug` configuration with `AndroidAddKeepAlives` enabled and `Release` configuration with linker mode set to `None`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
@@ -103,22 +103,49 @@ namespace MonoDroid.Tuner
 					if (method.Parameters [i].ParameterType.IsValueType || method.Parameters [i].ParameterType.FullName == "System.String")
 						continue;
 
-					changed = true;
-
 					if (methodKeepAlive == null)
-						methodKeepAlive = Context.GetMethod (
-#if NETCOREAPP
-							"System.Private.CoreLib",
-#else
-							"mscorlib",
-#endif
-							"System.GC", "KeepAlive", new string [] { "System.Object" });
+						methodKeepAlive = GetKeepAliveMethod ();
+
+					if (methodKeepAlive == null) {
+						LogMessage ("Unable to add KeepAlive call, did not find System.GC.KeepAlive method.");
+						break;
+					}
 
 					processor.InsertBefore (end, GetLoadArgumentInstruction (method.IsStatic ? i : i + 1, method.Parameters [i]));
 					processor.InsertBefore (end, Instruction.Create (OpCodes.Call, module.ImportReference (methodKeepAlive)));
+					changed = true;
 				}
 			}
 			return changed;
+		}
+
+		protected virtual AssemblyDefinition GetCorlibAssembly ()
+		{
+			return Context.GetAssembly (
+#if NETCOREAPP
+							"System.Private.CoreLib"
+#else
+							"mscorlib"
+#endif
+				);
+		}
+
+		MethodDefinition GetKeepAliveMethod ()
+		{
+			var corlibAssembly = GetCorlibAssembly ();
+			if (corlibAssembly == null)
+				return null;
+
+			var gcType = Extensions.GetType (corlibAssembly, "System.GC");
+			if (gcType == null)
+				return null;
+
+			return Extensions.GetMethod (gcType, "KeepAlive", new string [] { "System.Object" });
+		}
+
+		public virtual void LogMessage (string message)
+		{
+			Context.LogMessage (message);
 		}
 
 		// Adapted from src/Mono.Android.Export/Mono.CodeGeneration/CodeArgumentReference.cs
--- a/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
@@ -55,7 +55,7 @@ namespace Xamarin.Android.Tasks
 				// Set up the FixAbstractMethodsStep
 				var step1 = new FixAbstractMethodsStep (resolver, new TypeDefinitionCache (), Log);
 				// Set up the AddKeepAlivesStep
-				var step2 = new MonoDroid.Tuner.AddKeepAlivesStep (new TypeDefinitionCache ());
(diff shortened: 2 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/LinkerTests.cs
```diff
@@ -289,11 +289,16 @@ public AttributedButtonStub (Context context) : base (context)
...
 		}
 
+		// mode
+		//   0 .. Debug configuration
+		//   1 .. Debug configuration, AndroidAddKeepAlives=true
+		//   2 .. Release configuration
+		//   3 .. Release configuration, AndroidLinkMode=None
 		[Test]
-		public void AndroidAddKeepAlives ()
+		public void AndroidAddKeepAlives ([Values (0, 1, 2, 3)] int mode)
```
jonpryor, line 298: I would prefer:

```csharp
public void AndroidAddKeepAlives ([Values ("Debug", "Release")] string configuration, [Values ("", "True")] string keepAlives, [Values ("", "None")] string linkMode)
```

This reduces my mental requirements -- no "magic numbers" -- and also builds the combinatorial for all those properties.

radekdoulik: Do we really want to test 16 combinations instead of 4 important ones?

dellis1972: If you want to only have 4 configurations then you should use `TestCaseSource`

```
#pragma warning disable 414
    static object [] KeepAlivesTestCases = new object [] {
        new object [] {
             /* configuration*/ "Release",
             /* keepAlives */ "True",
            /* linkMode */ "",
        },
        new object [] {
             /* configuration*/ "Release",
             /* keepAlives */ "False",
            /* linkMode */ "",
        }
    }
#pragma warning restore 414
    [Test]
    [TestCaseSource(nameof(KeepAlivesTestCases))]
    public void AndroidAddKeepAlives (string configuration, string keepAlives, string linkMode)
```

This is a common pattern in our unit tests https://github.com/xamarin/xamarin-android/blob/7b51fd1b45f7f315df402ba333e141a49c8a3ca7/tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs#L120

src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
```diff
@@ -117,5 +117,28 @@ public override void LogMessage (string message)
...
+			public AddKeepAlivesStep (DirectoryAssemblyResolver resolver, TypeDefinitionCache cache, TaskLoggingHelper logger)
+				: base (cache)
+			{
+				this.resolver = resolver;
+				this.logger = logger;
+			}
+
+			protected override AssemblyDefinition GetCorlibAssembly ()
+			{
+				return resolver.GetAssembly ("mscorlib.dll");
```
jonathanpeppers, line 135: When this task runs in .NET 6 does it need to look for `System.Private.CoreLib`?

We can't use `#if` here, this assembly is always `netstandard2.0`. You might need to pass `$(UsingAndroidNETSdk)` into this task, which is `true` in .NET 6 and blank otherwise.

radekdoulik: Good tip about the `netstandard2.0`. It took me some time to notice that :-) I have added detection of `System.Private.CoreLib` during the resolver preparation, that should work too.

jonathanpeppers (approved):

src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs
```diff
@@ -44,18 +44,22 @@ public override bool RunTask ()
 				DeterministicMvid = Deterministic,
 			};
 
+			var hasSystemPrivateCorelib = false;
 			using (var resolver = new DirectoryAssemblyResolver (this.CreateTaskLogger (), loadDebugSymbols: true, loadReaderParameters: readerParameters)) {
 				// Add SearchDirectories with ResolvedAssemblies
 				foreach (var assembly in ResolvedAssemblies) {
 					var path = Path.GetFullPath (Path.GetDirectoryName (assembly.ItemSpec));
+					if (Path.GetFileName (assembly.ItemSpec) == "System.Private.CoreLib.dll")
```
jonpryor, line 52: Should this instead use `string.Equals(…, StringComparison.OrdinalIgnoreCase)`?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
```diff
@@ -103,24 +103,51 @@ bool AddKeepAlives (TypeDefinition type)
...
-#if NETCOREAPP
-							"System.Private.CoreLib",
-#else
-							"mscorlib",
-#endif
-							"System.GC", "KeepAlive", new string [] { "System.Object" });
+						methodKeepAlive = GetKeepAliveMethod ();
+
+					if (methodKeepAlive == null) {
+						LogMessage ("Unable to add KeepAlive call, did not find System.GC.KeepAlive method.");
```
jonpryor, line 110: @dellis1972: *Should* this either be a `LogDebugMessage()` or a `LogWarning()` with new warning code, or is a "normal" `LogMessage()` acceptable?

jonpryor: Never mind, wrong `LogMessage()`…

dellis1972: If its not actionable by the user, the `LogDebugMessage` might be better. `LogMessage` has a cost since it will emit in any build except quiet. Where the LogDebug* works only for diagnostic builds (and bin logs)

PR #5567, opened 2021-01-29
[Xamarin.Android.Build.Tasks] Remove Seppuku.java

dellis1972:
The `Seppuku` Broadcast Receiver was only ever used on devices which
were below sdk 14. Since our minimum sdk is now 19, we can safely
remove this class from our system.


jonathanpeppers:
Should we also change this throw for older API levels?

https://github.com/xamarin/androidtools/blob/b639c1ee85191193453b0bf436e816619737a2d9/Xamarin.AndroidTools/Devices/AndroidDeviceExtensions.cs#L63-L69

dellis1972:
> Should we also change this to throw an exception for older API levels?
> 
> https://github.com/xamarin/androidtools/blob/b639c1ee85191193453b0bf436e816619737a2d9/Xamarin.AndroidTools/Devices/AndroidDeviceExtensions.cs#L63-L69

We don't even support anything less than API 19 now. So I'm not sure that fallback is even needed anymore, we should remove it and the `if` statement as well I think.

jonathanpeppers (approved):

PR #5568, opened 2021-01-29
[ci] Move TimeZoneInfo tests to Nightly pipeline

pjcollins:
The broad set of TimeZoneInfo deployment tests don't need to be running
against every commit and PR build.  They should be moved to the Nightly
build and test pipeline that was introduced in commit 916d24b.  This 
will reduce PR build execution times and the overall machine time cost
of PR/CI builds.

### DIFF
--- a/build-tools/automation/azure-pipelines-nightly.yaml
+++ b/build-tools/automation/azure-pipelines-nightly.yaml
@@ -43,6 +43,7 @@ variables:
   HostedMacImage: macOS-10.15
   VSEngWinVS2019: Xamarin-Android-Win2019-1120
   GitHub.Token: $(github--pat--vs-mobiletools-engineering-service2)
+  NUnit.NumberOfTestWorkers: 4
 
 stages:
 - stage: mac_build
--- a/build-tools/automation/azure-pipelines-nightly.yaml
+++ b/build-tools/automation/azure-pipelines-nightly.yaml
@@ -71,6 +71,9 @@ stages:
 - stage: test
   displayName: Test
   dependsOn: mac_build
+  variables:
+    - group: Xamarin-Secrets
+    - group: xamops-azdev-secrets
   jobs:
   - job: emulator_tests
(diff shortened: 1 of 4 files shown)


dellis1972:
@pjcollins  I'm working on this one https://github.com/xamarin/xamarin-android/pull/5566 can we hold off merging this until I've figured out why they are failing please :)

pjcollins:
@dellis1972 👍  no problem. Have you tried `NonParallelizable` over there? I am not sure if there is a difference between that and the `SingleThreaded` attribute on the fixture but figured it might be worth a shot.

pjcollins:
Nightly run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4457891&view=results

dellis1972:
I'm wondering if we should do the nightly for full timezone tests, but have a limited set of timezone tests run during a PR? Maybe 10  or something? rather than the full 900 odd. Just in case we do break something in say a mono bump.

pjcollins:
We can manually queue the nightly run against any PR that we want additional test coverage for, particularly mono bumps. I also think we have some coverage in Mono.Android-Tests that may be sufficient? Those tests don't change the system time zone though. I'm also not entirely sure how we'd select the set of timezones that should be ran against mono bumps, do you have any recommendations for that? Maybe it's easiest to continue to have these tests be nightly and manual as needed.

grendello (approved):

dellis1972:
> We can manually queue the nightly run against any PR that we want additional test coverage for, particularly mono bumps. I also think we have some coverage in Mono.Android-Tests that may be sufficient? Those tests don't change the system time zone though. I'm also not entirely sure how we'd select the set of timezones that should be ran against mono bumps, do you have any recommendations for that? Maybe it's easiest to continue to have these tests be nightly and manual as needed.

Ok we can just run the tests manually I think that should be good.

dellis1972 (approved):

PR #5578, opened 2021-02-01
[monodroid] fix for Fast Dev & extractNativeLibs="false"

jonathanpeppers:
When using `$(EmbedAssembliesIntoApk)=false` and
`/application/@extractNativeLibs="false"`, apps crash on startup with:

    D debug-app-helper: Checking if libmonodroid was unpacked to /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonodroid.so
    D debug-app-helper: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonodroid.so not found, assuming application/android:extractNativeLibs == false
    I debug-app-helper: Setting up for DSO lookup directly in the APK
    D debug-app-helper: Added APK DSO lookup location: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/base.apk!/lib/arm64-v8a
    W debug-app-helper: Using runtime path: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64
    W debug-app-helper: checking directory: `/data/user/0/UnnamedProject.UnnamedProject/files/.__override__/lib`
    W debug-app-helper: directory does not exist: `/data/user/0/UnnamedProject.UnnamedProject/files/.__override__/lib`
    W debug-app-helper: Trying to load sgen from: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonosgen-64bit-2.0.so
    W debug-app-helper: Checking whether Mono runtime exists at: /system/lib64/libmonosgen-2.0.so
    F debug-app-helper: Cannot find 'libmonosgen-2.0.so'. Looked in the following locations:
    F debug-app-helper:   /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/base.apk!/lib/arm64-v8a
    F debug-app-helper: Do you have a shared runtime build of your app with AndroidManifest.xml android:minSdkVersion < 10 while running on a 64-bit Android 5.0 target? This combination is not supported.
    F debug-app-helper: Please either set android:minSdkVersion >= 10 or use a build without the shared runtime (like default Release configuration).

This is only occurring on `master`, not `d16-9`.

It appears `get_libmonosgen_path()` needs to check
`androidSystem.is_embedded_dso_mode_enabled()` and return early for
`Debug` builds using Fast Deployment:

    if (androidSystem.is_embedded_dso_mode_enabled ()) {
        return SharedConstants::MONO_SGEN_SO;
    }

This was working for `Release` builds.

I think a combination of changes surfaced this bug:

* Fast Dev vNext was introduced, removing the Mono Shared runtime.
  Previously, the `libmonosgen-2.0.so` would be located from the
  shared runtime.
* We accidentally lost usage of `java_runtime_fastdev.jar`.
  Fixed in: https://github.com/xamarin/monodroid/commit/aae6670c
* `BuildConfig.java` now reports the appropriate value for
  `mono.android.BuildConfig.Debug`.

### DIFF
--- a/src/monodroid/jni/debug-app-helper.cc
+++ b/src/monodroid/jni/debug-app-helper.cc
@@ -29,7 +29,7 @@ void log_fatal (LogCategories category, const char *format, ...);
 
 static void copy_file_to_internal_location (char *to_dir, char *from_dir, char *file);
 static void copy_native_libraries_to_internal_location ();
-static char* get_libmonosgen_path ();
+static const char* get_libmonosgen_path ();
 
 bool maybe_load_library (const char *path);
 
@@ -87,7 +87,7 @@ Java_mono_android_DebugRuntime_init (JNIEnv *env, [[maybe_unused]] jclass klass,
 		log_warn (LOG_DEFAULT, "Using runtime path: %s", androidSystem.get_runtime_libdir ());
 	}
 
-	char *monosgen_path = get_libmonosgen_path ();
+	const char *monosgen_path = get_libmonosgen_path ();
 	void *monosgen = dlopen (monosgen_path, RTLD_LAZY | RTLD_GLOBAL);
 	if (monosgen == nullptr) {
 		log_fatal (LOG_DEFAULT, "Failed to dlopen Mono runtime from %s: %s", monosgen_path, dlerror ());
@@ -196,7 +196,7 @@ runtime_exists (const char *dir, char*& libmonoso)
 	return false;
 }
 
-static char*
+static const char*
 get_libmonosgen_path ()
 {
 	char *libmonoso;
@@ -207,7 +207,7 @@ get_libmonosgen_path ()
 	copy_native_libraries_to_internal_location ();
 
 	if (androidSystem.is_embedded_dso_mode_enabled ()) {
-		return (char*) SharedConstants::MONO_SGEN_SO;
+		return SharedConstants::MONO_SGEN_SO;
 	}
 
 	for (size_t i = 0; i < BasicAndroidSystem::MAX_OVERRIDES; ++i) {
--- a/src/monodroid/jni/debug-app-helper.cc
+++ b/src/monodroid/jni/debug-app-helper.cc
@@ -200,26 +200,25 @@ static char*
 get_libmonosgen_path ()
 {
 	char *libmonoso;
-	bool embedded_dso_mode_enabled = androidSystem.is_embedded_dso_mode_enabled ();
 
 	// Android 5 includes some restrictions on loading dynamic libraries via dlopen() from
 	// external storage locations so we need to file copy the shared object to an internal
 	// storage location before loading it.
 	copy_native_libraries_to_internal_location ();
 
-	if (!embedded_dso_mode_enabled) {
-		for (size_t i = 0; i < BasicAndroidSystem::MAX_OVERRIDES; ++i) {
-			if (runtime_exists (BasicAndroidSystem::override_dirs [i], libmonoso)) {
-				return libmonoso;
-			}
+	if (androidSystem.is_embedded_dso_mode_enabled ()) {
+		return (char*) SharedConstants::MONO_SGEN_SO;
+	}
+
+	for (size_t i = 0; i < BasicAndroidSystem::MAX_OVERRIDES; ++i) {
+		if (runtime_exists (BasicAndroidSystem::override_dirs [i], libmonoso)) {
+			return libmonoso;
 		}
 	}
 
-	if (!embedded_dso_mode_enabled) {
-		for (size_t i = 0; i < BasicAndroidSystem::app_lib_directories_size; i++) {
-			if (runtime_exists (BasicAndroidSystem::app_lib_directories [i], libmonoso)) {
-				return libmonoso;
-			}
+	for (size_t i = 0; i < BasicAndroidSystem::app_lib_directories_size; i++) {
+		if (runtime_exists (BasicAndroidSystem::app_lib_directories [i], libmonoso)) {
+			return libmonoso;
 		}
 	}
 
@@ -265,12 +264,10 @@ get_libmonosgen_path ()
 		return libmonoso;
 	log_fatal (LOG_DEFAULT, "Cannot find '%s'. Looked in the following locations:", SharedConstants::MONO_SGEN_SO);
 
-	if (!embedded_dso_mode_enabled) {
-		for (size_t i = 0; i < BasicAndroidSystem::MAX_OVERRIDES; ++i) {
-			if (BasicAndroidSystem::override_dirs [i] == nullptr)
-				continue;
-			log_fatal (LOG_DEFAULT, "  %s", BasicAndroidSystem::override_dirs [i]);
-		}
+	for (size_t i = 0; i < BasicAndroidSystem::MAX_OVERRIDES; ++i) {
+		if (BasicAndroidSystem::override_dirs [i] == nullptr)
+			continue;
+		log_fatal (LOG_DEFAULT, "  %s", BasicAndroidSystem::override_dirs [i]);
 	}
 
 	for (size_t i = 0; i < BasicAndroidSystem::app_lib_directories_size; i++) {


src/monodroid/jni/debug-app-helper.cc
```diff
@@ -200,26 +200,25 @@ static char*
...
 	// storage location before loading it.
 	copy_native_libraries_to_internal_location ();
 
-	if (!embedded_dso_mode_enabled) {
-		for (size_t i = 0; i < BasicAndroidSystem::MAX_OVERRIDES; ++i) {
-			if (runtime_exists (BasicAndroidSystem::override_dirs [i], libmonoso)) {
-				return libmonoso;
-			}
+	if (androidSystem.is_embedded_dso_mode_enabled ()) {
+		return (char*) SharedConstants::MONO_SGEN_SO;
```
jonathanpeppers, line 210: I don't actually know if this cast is right.

grendello: It's not needed - the constant has the `char[]` type, which is the same as `char*`

grendello (approved):

src/monodroid/jni/debug-app-helper.cc
jonathanpeppers: If I remove it, I get:
```
debug-app-helper.cc(210,10): error GF586CA7D: cannot initialize return object of type 'char *' with an lvalue of type 'const char [19]'
```

grendello: ugh, const, right. Change the return value of the function to `const char*` (on lines 31 and 90 too), that should do it

src/monodroid/jni/debug-app-helper.cc
```diff
@@ -196,30 +196,29 @@ runtime_exists (const char *dir, char*& libmonoso)
...
 	// storage location before loading it.
 	copy_native_libraries_to_internal_location ();
 
-	if (!embedded_dso_mode_enabled) {
-		for (size_t i = 0; i < BasicAndroidSystem::MAX_OVERRIDES; ++i) {
-			if (runtime_exists (BasicAndroidSystem::override_dirs [i], libmonoso)) {
-				return libmonoso;
-			}
+	if (androidSystem.is_embedded_dso_mode_enabled ()) {
+		return SharedConstants::MONO_SGEN_SO;
```
jonpryor, line 210: This may *appear* to work, but I don't understand *why* it works.  If we're in "embedded DSO mode" (feb9ea2a53,  6be4bdbcfd9f2fbe6267aceb934394bcdd0f13b4, bf657e815b200e57098725c0ad963d70e881ea69, …, 95ca1025f0, …?), then we need to *change* the path used for **dlopen**(3) to include the path to the `.apk`, a `!`, and then the path *within* the `.apk`.  See also: https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md

> Note that in API level 23 and above dlopen(3) will open a library from any zip file, not just your APK. Just give dlopen(3) a path of the form “my_zip_file.zip!/libs/libstuff.so”.

Thus, by my reading, *when* `androidSystem.is_embedded_dso_mode_enabled()`==true, we *can't* "just" `dlopen("libmonosgen-2.0.so")`, right…?

Unless we're implicitly depending on "Correct soname/path handling (Available in API level >= 23)" (same google doc link above):

> The dynamic linker now understands the difference between a library’s soname and its path…
> For example, `dlopen("/this/directory/does/not/exist/libc.so", RTLD_NOW)` would find `/system/lib/libc.so` because it’s already loaded.

*Because* we load `libmonosgen-2.0.so` very early in Java bootstrap, an "unpathed" `dlopen("libmonosgen-2.0.so")` would find the already loaded file, which is good.

…except that we currently support down to API-19, and with .NET 6 we'll support down to API-21, *both* of which are < API-23, which adds "Correct soname/path handling".

Thus, I *strongly suspect* that if you tried this codepath on an API-21 device/emulator, it would *not* work.

jonpryor:
The PR message states:

> I think a combination of changes surfaced this bug:

I don't think this "root cause" is quite sufficient, or at least it's not sufficient for my mind.  The `java_runtime_fastdev.jar` changes only impact which `MultiDexLoader.java` is used.  I do not see how the `MultiDexLoader` type would impact which paths are checked for `libmonosgen-2.0.so`.  This feels like a red herring.  

Ditto the `BuildConfig.java` change: I don't see how/why that would impact which locations are checked for `libmonosgen-2.0.so`, *especially* since commit bf657e815b200e57098725c0ad963d70e881ea69, which made the entire lookup process "more dynamic".

The only explanation which makes sense is the removal of the shared runtime in 2315b76abc732f4e925e48a4654af4844376f0a7 on 2021-Jan-6, which means we should see this on *every* build which runs `DebuggingTest.ApplicationRunsWithoutDebugger()` (in the "MSBuild With Emulator" stage) in current history (~2 weeks)…and we *don't* see that.

Which suggests that 2315b76abc732f4e925e48a4654af4844376f0a7 *in and of itself* didn't cause the crash.

What *did*?

(Is it strictly necessary to do a root-cause analysis to find which commit introduced the crash?  No; I'm just curious, as it makes for good commit message fodder.  Plus, [as mentioned elsewhere](https://github.com/xamarin/xamarin-android/pull/5578#discussion_r568276166), the fix suggested here likely won't work on API-21 devices, and thus won't work for our current requirements anyway. 🙁)

jonpryor:
With a "quick" glance through master builds, I see that `DebuggingTest.ApplicationRunsWithoutDebugger()` fails in 4bf39c343074282737828019b77f5a8b0858d9b9 and 81fee40b813e90968254d1e464c6b6dea1ddeb0f, but *doesn't* fail in commit 42150254a9e4b42c18cc3703763fd2b088e4fb29.

This appears to make 81fee40b813e90968254d1e464c6b6dea1ddeb0f the likely culprit, specifically:

https://github.com/xamarin/monodroid/commit/aae6670c6ff4b68d36fb976fd6dec67a2a6d43d7

which *isn't* on the d16-9 branch, and thus *would* explain why this fails *recently* on master (only since last Friday) but not on d16-9.  (Additionally, there are *no* differences in `src/monodroid` between the d16-9 & master branches, further suggesting that the *cause* of the fix is not within `src/monodroid`.  This doesn't mean that the fix can't/shouldn't be in `src/monodroid`, but it does "raise the bar", somewhat.)

src/monodroid/jni/debug-app-helper.cc
grendello: @jonpryor embedded DSOs aren't available in API 21, they were added starting from API 23 - together with the correct search path handling for `dlopen`. So `androidSystem.is_embedded_dso_mode_enabled()` will always return `false` in API levels older than 23 and things will work as expected.

grendello:
> 
> (Is it strictly necessary to do a root-cause analysis to find which commit introduced the crash? No; I'm just curious, as it makes for good commit message fodder. Plus, [as mentioned elsewhere](https://github.com/xamarin/xamarin-android/pull/5578#discussion_r568276166), the fix suggested here likely won't work on API-21 devices, and thus won't work for our current requirements anyway. slightly_frowning_face)

The fix here doesn't change anything [for API-21](https://github.com/xamarin/xamarin-android/pull/5578#discussion_r568388127) devices, they will keep working as before.

src/monodroid/jni/debug-app-helper.cc
jonpryor: ooh, good point.

jonpryor:
…but all xamarin/monodroid@aae6670 does is alter the `java_runtime.jar`/`java_runtime_fastdev.jar` built into the app…

Meaning the different `.jar` file *is* responsible.  This still doesn't make sense to me.

Aside? https://github.com/xamarin/monodroid/pull/1160#issuecomment-771567977

jonpryor:
> This still doesn't make sense to me.

Let's explore that.  The crashing app should be using `java_runtime_fastdev.jar`, which means it's uses:

  * https://github.com/xamarin/xamarin-android/blob/28b72ecf587cbbde47db8312fe0daa71fcf3fad4/src/java-runtime/java/mono/android/debug/BuildConfig.java
  * https://github.com/xamarin/xamarin-android/blob/28b72ecf587cbbde47db8312fe0daa71fcf3fad4/src-ThirdParty/bazel/java/mono/android/debug/MultiDexLoader.java

I *hope* that `MultiDexLoader.java` isn't responsible, so looking at `BuildConfig.java`, it only has a single member: `BuildConfig.Debug`.  Where is that used?

In `MonoPackageManager.java`: https://github.com/xamarin/xamarin-android/blob/28b72ecf587cbbde47db8312fe0daa71fcf3fad4/src/java-runtime/java/mono/android/MonoPackageManager.java#L74-L79

which in turn means that when fastdev is used, there *isn't* a `System.loadLibrary("monosgen-2.0")`; instead, `xamarin-debug-app-helper` is responsible, which is where `get_libmonosgen_path()` comes in: https://github.com/xamarin/xamarin-android/blob/28b72ecf587cbbde47db8312fe0daa71fcf3fad4/src/monodroid/jni/debug-app-helper.cc#L90

So now I have an explanation:

1. xamarin/monodroid@aae6670 caused us to use the correct `java_runtime_fastdev.jar` file, in which `BuildConfig.Debug`==true
2. When `BuildConfig.Debug`==true, we use `xamarin-app-debug-helper` & `DebugRuntime.init()` to find & load mono
3. mono couldn't be found, so app crash.

This still leaves point (3) though: *why* couldn't mono be found? Specifically, *where is it* that we're not looking?

```
Checking if libmonodroid was unpacked to /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonodroid.so
/data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonodroid.so not found, assuming application/android:extractNativeLibs == false
Setting up for DSO lookup directly in the APK
Added APK DSO lookup location: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/base.apk!/lib/arm64-v8a
```

OK, we assume extractNativeLibs==false, and thus we should be looking for files within the `.apk`.

```
…
Trying to load sgen from: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonosgen-64bit-2.0.so
```

Yet, after saying we'll look within the `.apk`, we look on-disk *anyway*.  Though this is for `libmonosgen-64bit-2.0.so`, not `libmonosgen-2.0.so`…

This fails.

```
Cannot find 'libmonosgen-2.0.so'. Looked in the following locations:
  /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/base.apk!/lib/arm64-v8a
```

The log message implies that it looked at a location which *was not actually checked*.  There was no `Trying to load sgen from` message where it tried to open `…/base.apk!/lib/arm64-v8a/libmonosgen-2.0.so`.

From which I infer that there *is* a `…/base.apk!/lib/arm64-v8a/libmonosgen-2.0.so`, we just never checked it, even though we said we would.

*This* doesn't make sense either. :-(

src/monodroid/jni/debug-app-helper.cc
jonpryor: Actually, no, I don't get it. if embedded DSO mode is enabled *and* we're using `java_runtime_fastdev.jar`, in which case there is no `System.loadLibrary("monosgen-2.0")`, then *how* does `dlopen("libmonosgen-2.0.so")` *work*?

…and if we're using `java_runtime.jar`, then *we won't get here*, because this code is in `xamarin-debug-app-helper`, which isn't used when `BuildConfig.Debug`==false.

Apparently this works; we don't see `DebuggingTest.ApplicationRunsWithoutDebugger()` failing on CI for this PR.

I still don't understand *why* it works. :-(

jonathanpeppers:
So just a general question: why do we need a `get_libmonosgen_path()` method at all anymore?

If there is no shared runtime, can we load it the same way as all other native libraries?

grendello:
> 
> The log message implies that it looked at a location which _was not actually checked_. There was no `Trying to load sgen from` message where it tried to open `…/base.apk!/lib/arm64-v8a/libmonosgen-2.0.so`.
> 
> From which I infer that there _is_ a `…/base.apk!/lib/arm64-v8a/libmonosgen-2.0.so`, we just never checked it, even though we said we would.
> 
> _This_ doesn't make sense either. :-(

We cannot fstat files inside the APK, not from this code. And that path is a fallback to use what shared runtime had, IIRC? So removing the shared runtime is what, eventually, caused the error.

grendello:
> So just a general question: why do we need a `get_libmonosgen_path()` method at all anymore?
> 
> If there is no shared runtime, can we load it the same way as all other native libraries?

The shared libraries can be uploaded to the override directories, which is what we search in `get_libmonosgen_path` - unless that changed with fastdev v2?

jonathanpeppers:
> The shared libraries can be uploaded to the override directories, which is what we search in get_libmonosgen_path - unless that changed with fastdev v2?

Yes, but _all_ native libraries are in the same place now. We should be able to load them the same?

jonathanpeppers:
I tried to test these changes on an API 19 emulator, but ran into: https://github.com/xamarin/monodroid/issues/1161

src/monodroid/jni/debug-app-helper.cc
grendello: `dlopen` searches the equivalent of `LD_LIBRARY_PATH` which automatically includes the `myapp.apk!/libs/[ARCH]` location, so if you do `dlopen("somelib.so")` it will find it if it's there.

jonathanpeppers:
This change seems to work on an API 19 emulator, the app starts:
```
I/ActivityManager( 1596): Displayed UnnamedProject.UnnamedProject/unnamedproject.unnamedproject.MainActivity: +1s386ms
```
So was there actually a problem with API < 21 here?

grendello:
> This change seems to work on an API 19 emulator, the app starts:
> 
> ```
> I/ActivityManager( 1596): Displayed UnnamedProject.UnnamedProject/unnamedproject.unnamedproject.MainActivity: +1s386ms
> ```
> 
> So was there actually a problem with API < 21 here?

No, there shouldn't have been - API <23 doesn't support embedded DSOs at all

jonpryor:
Commit message for review:

```markdown
Starting with commit 81fee40b, commercial builds which use
`$(EmbedAssembliesIntoApk)=false` and
`/application/@extractNativeLibs="false"`, will abort on startup with:

	D debug-app-helper: Checking if libmonodroid was unpacked to /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonodroid.so
	D debug-app-helper: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonodroid.so not found, assuming application/android:extractNativeLibs == false
	I debug-app-helper: Setting up for DSO lookup directly in the APK
	D debug-app-helper: Added APK DSO lookup location: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/base.apk!/lib/arm64-v8a
	W debug-app-helper: Using runtime path: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64
	W debug-app-helper: checking directory: `/data/user/0/UnnamedProject.UnnamedProject/files/.__override__/lib`
	W debug-app-helper: directory does not exist: `/data/user/0/UnnamedProject.UnnamedProject/files/.__override__/lib`
	W debug-app-helper: Trying to load sgen from: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonosgen-64bit-2.0.so
	W debug-app-helper: Checking whether Mono runtime exists at: /system/lib64/libmonosgen-2.0.so
	F debug-app-helper: Cannot find 'libmonosgen-2.0.so'. Looked in the following locations:
	F debug-app-helper:   /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/base.apk!/lib/arm64-v8a
	F debug-app-helper: Do you have a shared runtime build of your app with AndroidManifest.xml android:minSdkVersion < 10 while running on a 64-bit Android 5.0 target? This combination is not supported.
	F debug-app-helper: Please either set android:minSdkVersion >= 10 or use a build without the shared runtime (like default Release configuration).

The "cause" of the abort is:

 1. Commit [xamarin/monodroid@aae6670][0] updates the build so that
    the *intended* `java_runtime_fastdev.jar` is built into the app
    when `$(EmbedAssembliesIntoApk)`=False.

    `java_runtime_fastdev.jar` has a `mono.android.BuildConfig` type
    which has `BuildConfig.Debug`=true.

    In `java_runtime.jar`, `mono.android.BuildConfig.Debug`=false.

    Commit 81fee40b "brought in" xamarin/monodroid@aae6670.

 2. With `java_runtime_fastdev.jar` now used, `BuildConfig.Debug`
    would now be true in "Debug" configuration builds.  This in turn
    meant that [we'd now use][1] `libxamarin-debug-app-helper.so`
    (6be4bdbc).

 3. `DebugRuntime.init()`, within `libxamarin-debug-app-helper.so`,
    is responsible for finding and loading `libmonsogen-2.0.so`.
    `DebugRuntime.init()` determines that
    `/application/@extractNativeLibs`=false, as `libmonodroid.so`
    doesn't exist within [`ApplicationInfo.nativeLibraryDir`][2]:

        …/libmonodroid.so not found, assuming application/android:extractNativeLibs == false

 4. Despite believing that `@extractNativeLibs`=false,
    `DebugRuntime.init()` tried to load an on-disk filename:

        Trying to load sgen from: /data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/lib/arm64/libmonosgen-64bit-2.0.so

    Aside: it tries to load `libmonsgen-64bit-2.0.so` because the
    previous Fast Deployment infrastructure would create such a thing.
    This lookup is now *vestigial*; there is no need to try to find
    `SharedConstants::MONO_SGEN_ARCH_SO`.  (TODO: cleanup?).

    This attempt to load `libmonosgen-64bit-2.0.so` fails.

 5. Next `DebugRuntime.init()` tries to load from a system location:

        Checking whether Mono runtime exists at: /system/lib64/libmonosgen-2.0.so

    This attempt also fails.

 6. All known filesystem locations have been exhausted and
    `libmonosgen-2.0.so` hasn't been found.  `DebugRuntime.init()`
    aborts.

Notably absent from any of the locations checked: the location which
actually contains `libmonosgen-2.0.so`, the path *within* the `.apk`:

	/data/app/~~5pGpNKwo58itttVZIz8-GA==/UnnamedProject.UnnamedProject-Lh_MOHcs1J3EajxFou0uBg==/base.apk!/lib/arm64-v8a/libmonosgen-2.0.so

Fix `get_libmonosgen_path()` -- called by `DebugRuntime.init()` -- 
so that when `/application/@extractNativeLibs`=false, we perform *no*
filesystem checks, and instead (effectively) perform:

	void *monosgen = dlopen ("libmonosgen-2.0.so");

This works because [`/application/@extractNativeLibs`][3] was added
in API-23, and API-23 includes this [**dlopen**(3) change][4]:

> ## Changes to library dependency resolution
> 
> Until it was fixed in JB-MR2 [API-18 --ed], Android didn‘t include
> the application library directory on the dynamic linker’s search path.

meaning that with API-18+, the application's library directory,
*and in API-23* native libraries within the `.apk` itself, will be
searched when attempting to load a library.

Thus, `dlopen("libmonosgen-2.0.so")` will appropriately search the
`.apk` when `/application/@extractNativeLibs`=false.

[0]: https://github.com/xamarin/monodroid/commit/aae6670c6
[1]: https://github.com/xamarin/xamarin-android/blob/28b72ecf587cbbde47db8312fe0daa71fcf3fad4/src/java-runtime/java/mono/android/MonoPackageManager.java#L74-L79
[2]: https://developer.android.com/reference/android/content/pm/ApplicationInfo#nativeLibraryDir
[3]: https://developer.android.com/guide/topics/manifest/application-element#extractNativeLibs
[4]: https://android.googlesource.com/platform/bionic/+/master/android-changes-for-ndk-developers.md
```

PR #5581, opened 2021-02-02
Test removing *.NamingCustomAttributes.shproj.

jpobst:


Xamarin.Android.sln
```diff
@@ -1,53 +1,49 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27428.2037
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30914.41
 MinimumVisualStudioVersion = 10.0.40219.1
```
jonathanpeppers, line 5: It seems like the whole `.sln` file got upgraded.

Since, no one understands the syntax of this file, could we remove the couple lines that list the `.shproj` so the diff isn't as huge?

Might be better to update the `.sln` in a separate PR.

jpobst: Agreed, this was just a test PR!

PR #5585, opened 2021-02-02
[ci] Add more Mono.Android-Tests data to App Insights

pjcollins:
The AppBundle, MultiDex, and .NET 6 variants of Mono.Android-Tests will
now write their timing data to the same .csv file that is used for all
other Mono.Android-Tests variants.  This will allow us to compare
timing data from any Mono.Android-Tests run in the same chart.

See the [App Insights wiki][0] for more information on how we visualize
the startup timing that is being sent to App Insights.

[0]: https://github.com/xamarin/monodroid/wiki/Azure-Pipelines-and-DevOps#startup-timing-visualizations-using-app-insights


radekdoulik (approved):

jonpryor:
How will this new data be visualized?  Will they be new graphs?  Or will the new data be "merged" with existing graphs?

pjcollins:
The new data will automatically show up in the "Mono.Android-Tests" chart when this is merged. We can break up that chart into more pieces if it gets too noisy.

jonathanpeppers (approved):
The changes seem OK, I guess we'll have to merge to see if the graph is ok.

PR #5586, opened 2021-02-02
[tests] add MSBuild test for Xamarin.Legacy.Sdk

jonathanpeppers:
Context: https://github.com/xamarin/Xamarin.Legacy.Sdk

Xamarin.Legacy.Sdk is a way for .NET 6 projects to target older target
frameworks such as `monoandroid11.0`:

    <Project Sdk="Xamarin.Legacy.Sdk">
      <PropertyGroup>
        <TargetFrameworks>monoandroid11.0;net6.0-android</TargetFrameworks>
      </PropertyGroup>
    </Project>

This adds a test that creates a `.nupkg` file verifying that
assemblies for both .NET 6 and "legacy" Xamarin.Android can be
produced.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
@@ -490,6 +490,35 @@ namespace Xamarin.Android.Build.Tests
 			}
 		}
 
+		[Test]
+		public void XamarinLegacySdk ()
+		{
+			var proj = new XASdkProject (outputType: "Library") {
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
It appears a fix is needed:
```
_LegacyExtensionsPath = C:\Program Files (x86)\Microsoft Visual Studio\2019\EnterpriseMSBuild
```
It looks like `$(VsInstallRoot)` may/may not have a trailing slash, which breaks this logic

https://github.com/xamarin/Xamarin.Legacy.Sdk/blob/43391b2665dd48287bc8715122a162a9cd3b2cfa/src/Xamarin.Legacy.Sdk/Sdk/Xamarin.Legacy.Android.targets#L9-L16

radekdoulik (approved):

dellis1972 (approved):

PR #5588, opened 2021-02-03
[build] Sync solution files with what VS2019 16.8 generates.

jpobst:
Our solution files no longer match what VS2019 generates when it saves the file.  Although either version of the solution files work, they cause the local tree to be dirty, and can obscure actual intended changes to the solution file.

This commit loads the solution files in VS2019 16.8 and saves them.

The main changes are project types changing from `FAE04EC0-301F-11D3-BF4B-00C04F79EFBC` to `9A19103F-16F7-4668-BE54-9A1E7A4F7556`.  The former is the project type for long form C# projects and the latter is the project type for SDK style projects.

There are also some additional items VS wants to specify regarding Shared Projects.


jonathanpeppers (approved):
I think this is a good change, sometimes VS keeps changing this file for me, too.

I don't know how we figure out what `FAE04EC0-301F-11D3-BF4B-00C04F79EFBC` vs `9A19103F-16F7-4668-BE54-9A1E7A4F7556` means, though.

PR #5599, opened 2021-02-08
Try to fix a build error in xabuild on Windows

dellis1972:


jonathanpeppers:
@dellis1972 I figured this out friday, I think: https://github.com/xamarin/xamarin-android/pull/5597

PR #5604, opened 2021-02-09
[illink] Use latest illink

radekdoulik:
Use the latest `6.0.0-alpha.1.21109.1` version of `Microsoft.NET.ILLink`
linker SDK.

Target `net6.0` when building our custom steps.

We cannot build it with `msbuild` on Mac as we don't have version 16.8
or newer. So instead build the `Microsoft.Android.Sdk.ILLink` project
with `dotnet build`, just before packaging our net6 nugets.


src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
```diff
@@ -1,17 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
```
jonathanpeppers, line 1: Since this file isn't in a `.sln` anymore, should we make a new `.sln` file next to this project?

I think you could probably open a `Microsoft.Android.Sdk.ILLink.sln` if you have a system .NET 6 installed.

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
```diff
@@ -1,17 +1,14 @@
...
   <Import Project="..\..\Configuration.props" />
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <DefineConstants>NET5_LINKER</DefineConstants>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <OutputPath>$(XAInstallPrefix)xbuild\Xamarin\Android\</OutputPath>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.NET.ILLink" Version="5.0.0-preview.3.20227.3" />
```
jonathanpeppers, line 10: Is it worth putting this version in `Configuration.props` so it is next to `$(DotNetPreviewVersion)`?

We should probably bump this number when bumping the build of .NET 6?

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
radekdoulik: I think it is not needed. I usually just open the .csproj in  VS when I need UI.

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
radekdoulik: Makes sense, updated.

build-tools/create-packs/Directory.Build.targets
```diff
@@ -70,8 +70,12 @@
     </ItemGroup>
   </Target>
 
+  <Target Name="BuildILLinkCustomStep">
+    <Exec Command="$(AndroidToolchainDirectory)\dotnet\dotnet build /p:Configuration=$(Configuration) $(MSBuildThisFileDirectory)..\..\src\Microsoft.Android.Sdk.ILLink\Microsoft.Android.Sdk.ILLink.csproj" />
```
jonathanpeppers, line 74: This could use `$(DotNetPreviewTool)`, like the others below.

I think you might also need `&quot;` around the path to the `.csproj`. If there were a space in the path it would break.

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
```diff
@@ -1,17 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\Configuration.props" />
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
dellis1972, line 4: silly question, but does this need to be net6.0? or can it be netstandard2.1?

radekdoulik: it needs to be at least net5.0 IIRC

radekdoulik:
```
src\Microsoft.Android.Sdk.ILLink\Microsoft.Android.Sdk.ILLink.csproj : error NU1202: Package Microsoft.NET.ILLink 6.0.0-alpha.1.211
09.1 is not compatible with netstandard2.1 (.NETStandard,Version=v2.1). Package Microsoft.NET.ILLink 6.0.0-alpha.1.21109.1 supports: net5.0 (.NETCoreApp,Version=v
5.0)
```

dellis1972: ok , was just asking :)

dellis1972 (approved):

PR #5605, opened 2021-02-09
Default InvariantGlobalization in Xamarin Android

eerhardt:
This allows for unnecessary code to be trimmed by default in an app. See https://github.com/dotnet/runtime/pull/47999.


jonathanpeppers (approved):

jonpryor:
@eerhardt : why isn't `$(InvariantGlobalization)`=False the default in the first place?  What are the ramifications of this change?

eerhardt:
> why isn't $(InvariantGlobalization)=False the default in the first place? 

The behavior of `$(InvariantGlobalization)=False` is the default behavior of the runtime when `$(InvariantGlobalization)` isn't explicitly set. But we don't explicitly set this in the core SDK because this would not allow an end user set the environment variable to control this behavior, since the `runtimeconfig.json` value overrides the environment variable. In the core SDK, we only want the `runtimeconfig.json` setting set when the developer explicitly sets it.

> What are the ramifications of this change?

For app models that are trimmed (Blazor WASM, Xamarin), this setting allows the code that is only used when `$(InvariantGlobalization)=true` to be trimmed from the app when it isn't being used. This isn't a substantial amount of code, but for Blazor WASM, it was 10 KB uncompressed and 3 KB .br compressed.

radekdoulik (approved):

radekdoulik:
Added `do-not-merge` to postpone until I measure the difference for the description message.

radekdoulik:
Ah, but there will be no difference yet. We should bump net6 version too.

radekdoulik:
I tried to measure it with `6.0.0-preview.1.21109.8`, but that didn't contain the runtime change yet. So let's merge it without that information.

PR #796, opened 2017-08-30
[Xamarin.Andriod.Build.Tasks] Pulling in native libraries from incorrect supported architectures

nwestfall:
If building only for x86 or armeabi, x86_64 and armeabi-v7a native libaries are still pulled in when building the APK because the names are being checked using 'contains'.  This causes issues when deploying to x86_64 device, but only building for x86 because android will start the VM for x86_64 and get a runtime error because it will be missing the libmonodroid.so.

See bug https://bugzilla.xamarin.com/show_bug.cgi?id=47607
Test project: https://github.com/nwestfall/XamarinMultidexTest

PR #5608, opened 2021-02-10
Bump to xamarin/Java.Interop/main@b46598a2; packageSources

jonpryor:
Context: https://medium.com/@alex.birsan/dependency-confusion-4a5d60fec610
Context: https://azure.microsoft.com/en-us/resources/3-ways-to-mitigate-risk-using-private-package-feeds/
Context: https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/12676/ncident-help-for-Substitution-attack-risk-from-multiple-package-feeds

Changes: https://github.com/xamarin/Java.Interop/compare/ee7b6bbe382bf408cc1a991e6149819889ad8c6c...b46598a254c20060b107312564e0ec0aee9e33d6

  * xamarin/Java.Interop@b46598a2: Bump to xamarin/xamarin-android-tools/main@479931ce; packageSources (#796)

There is a Package Substitution Attack inherent in NuGet, whereby
if multiple package sources provide packages with the same name,
it is *indeterminate* which package source will provide the package.

For example, consider the [`XliffTasks` package][0], currently
provided from the [`dotnet-eng`][1] feed, and *not* present in the
NuGet.org feed.  If a "hostile attacker" submits an `XliffTasks`
package to NuGet.org, then we don't know, and cannot control, whether
the build will use the "hostile" `XliffTasks` package from NuGet.org
or the "desired" package from `dotnet-eng`.

There are two ways to prevent this attack:

 1. Use `//packageSources/clear` and have *only one*
    `//packageSources/add` entry in `NuGet.config`

 2. Use `//packageSources/clear` and *fully trust* every
    `//packageSources/add` entry in `NuGet.config`.
    `NuGet.org` *cannot* be a trusted source, nor can any feed
    location which allows "anyone" to add new packages, nor can
    a feed which itself contains [upstream sources][2].

As the `XliffTasks` package is *not* in `NuGet.org`, option (1)
isn't an option.  Go with option (2), using the existing
`dotnet-eng` source and the new *trusted* [`dotnet-public`][3]
package source.

[0]: https://github.com/dotnet/xliff-tasks
[1]: https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-eng
[2]: https://docs.microsoft.com/en-us/azure/devops/artifacts/concepts/upstream-sources?view=azure-devops
[3]: https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-public

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -781,6 +781,11 @@ stages:
         git submodule update -q --init --recursive
       displayName: Clone and update UITools
 
+    - task: NuGetAuthenticate@0
+      displayName: authenticate with azure artifacts
+      inputs:
+        forceReinstallCredentialProvider: true
+
     - task: provisionator@2
       displayName: provision designer dependencies
       inputs:
@@ -855,6 +860,11 @@ stages:
         git checkout $branchName
         git submodule update -q --init --recursive
       displayName: Clone and update UITools
+
+    - task: NuGetAuthenticate@0
+      displayName: authenticate with azure artifacts
+      inputs:
+        forceReinstallCredentialProvider: true
 
     - task: provisionator@2
       displayName: provision designer dependencies
--- a/NuGet.config
+++ b/NuGet.config
@@ -3,8 +3,8 @@
   <packageSources>
     <clear />
     <!-- ensure only the sources defined below are used -->
+    <add key="dotnet-public" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-public/nuget/v3/index.json" protocolVersion="3" />
     <add key="dotnet-eng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" protocolVersion="3" />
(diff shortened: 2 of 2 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -781,6 +781,11 @@ stages:
         git submodule update -q --init --recursive
       displayName: Clone and update UITools
 
+    - task: NuGetAuthenticate@0
```
radekdoulik, line 784: What are the conditions to use the **NuGetAuthenticate** task? In other words, when should we use that task and when it is not needed?

pjcollins: This was needed to access the package source used in android-sdk-installer: https://github.com/xamarin/android-sdk-installer/commit/e2f249ea010ad266861acb21890b65374cfc6797#diff-b54ee602750f150b3d4986e0334cc3ec901d47510560e26e792e8d294ab6f37aR5

@joj can maybe help provide more information about why it's now needed, I am assuming this new source is more restricted some how.

pjcollins (approved):

PR #5617, opened 2021-02-11
[illink] Set LinkAttributes logical name

radekdoulik:
Context: https://github.com/mono/linker/blob/main/docs/data-formats.md#custom-attributes-annotations-format

Change the logical name of attributes embedded resource to
`ILLink.LinkAttributes.xml` instead of current
`Android.ILLink.ILLink.LinkAttributes.xml`


jonathanpeppers (approved):
Is there a way to know we have this correct now?

Does it preserve more/less classes we could add a test for?

grendello (approved):

radekdoulik:
> 
> 
> Is there a way to know we have this correct now?
> 
> Does it preserve more/less classes we could add a test for?

I didn't see any changes in the apk. I think we might be still missing something. I am going to look closer at it in my next type registrations PR where I am adding substitutions xml file.

PR #5620, opened 2021-02-12
[One .NET] drop the .21 from android RIDs

jonathanpeppers:
When looking at the .NET runtime packs for iOS/Android:

    Microsoft.Android.Runtime.android.21-arm
    Microsoft.Android.Runtime.android.21-arm64
    Microsoft.Android.Runtime.android.21-x86
    Microsoft.Android.Runtime.android.21-x64
    Microsoft.iOS.Runtime.ios-arm
    Microsoft.iOS.Runtime.ios-arm64
    Microsoft.iOS.Runtime.ios-x86
    Microsoft.iOS.Runtime.ios-x64
    Microsoft.NETCore.App.Runtime.android-arm
    Microsoft.NETCore.App.Runtime.android-arm64
    Microsoft.NETCore.App.Runtime.android-x86
    Microsoft.NETCore.App.Runtime.android-x64
    Microsoft.NETCore.App.Runtime.ios-arm
    Microsoft.NETCore.App.Runtime.ios-arm64
    Microsoft.NETCore.App.Runtime.ios-x86
    Microsoft.NETCore.App.Runtime.ios-x64

It seems like we should not be putting the `.21` in the names of our
Android runtime packs.

In fact, this command does not work currently:

    > dotnet build HelloAndroid -r android-arm64
    ...
    Microsoft.NET.Sdk.FrameworkReferenceResolution.targets(389,5): error NETSDK1082:
    There was no runtime pack for Microsoft.Android available for the specified RuntimeIdentifier 'android-arm64'.

For the above command to work, we need to drop the `.21` from our
runtime pack names.

This repo appears to be the only place `.21` is hardcoded. The code in
`xamarin-android-tools` looks like it will continue to work:

https://github.com/xamarin/xamarin-android-tools/blob/479931ce1449b39c0165315fc103e7e12812f34b/src/Microsoft.Android.Build.BaseTasks/AndroidRidAbiHelper.cs#L74-L91

After these changes, any of the following commands work:

    > dotnet build HelloAndroid -r android-arm64
    > dotnet build HelloAndroid -r android.21-arm64
    # Or any API level in between
    > dotnet build HelloAndroid -r android.30-arm64

### DIFF
--- a/Documentation/guides/OneDotNet.md
+++ b/Documentation/guides/OneDotNet.md
@@ -96,7 +96,7 @@ Instead use .NET's concept of [runtime identifiers][rids]:
 ```xml
 <PropertyGroup>
   <!-- Used going forward in .NET -->
-  <RuntimeIdentifiers>android.21-arm;android.21-arm64;android.21-x86;android.21-x64</RuntimeIdentifiers>
+  <RuntimeIdentifiers>android-arm;android-arm64;android-x86;android-x64</RuntimeIdentifiers>
 </PropertyGroup>
 ```
 
--- a/Documentation/guides/messages/xa0035.md
+++ b/Documentation/guides/messages/xa0035.md
@@ -20,7 +20,7 @@ property are valid:
 
 ```xml
 <PropertyGroup>
-  <RuntimeIdentifiers>android.21-arm;android.21-arm64;android.21-x86;android.21-x64</RuntimeIdentifiers>
+  <RuntimeIdentifiers>android-arm;android-arm64;android-x86;android-x64</RuntimeIdentifiers>
 </PropertyGroup>
 ```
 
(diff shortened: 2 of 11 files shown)


dellis1972:
didn't the `21` refer to the min api level?

jonathanpeppers:
We have a `Microsoft.Android.Sdk.SupportedPlatforms.targets` file that defines:
```
<AndroidSdkSupportedTargetPlatformVersion Include="21" />
<!-- all the numbers in between -->
<AndroidSdkSupportedTargetPlatformVersion Include="30" />
```
If we drop the `.21` from our package names, it can use this file properly. If I try to use `android.20-arm`, it doesn't work.

I think we were somehow lucky the `.21` was working before?

jonathanpeppers:
Going to try this one as a non-fork: #5621

Issue #5515, opened 2021-01-18
[illink] Try to not preserve `__TypeRegistrations` classes


radekdoulik:
These should hopefully be not needed anymore as the new typemap should be sufficient.

radekdoulik:
@grendello does typemap handle all the managed to java type lookups? I wonder whether we can remove this part of code as well https://github.com/xamarin/xamarin-android/blob/7db5be8a56779fbcf0dc7f683d2831910b420d31/src/Mono.Android/Java.Interop/TypeManager.cs#L222-L237 Plus all the code related to package registration.

grendello:
@radekdoulik Typemap handles all the types handed to it by the type scanner (from Java.Interop) and what it cannot find is handled in the managed code in a [fallback path](https://github.com/xamarin/xamarin-android/blob/7db5be8a56779fbcf0dc7f683d2831910b420d31/src/Mono.Android/Android.Runtime/JNIEnv.cs#L697-L737). It feels like removing the type registrations should work?

radekdoulik:
Done.

https://github.com/xamarin/xamarin-android/pull/5516
https://github.com/xamarin/xamarin-android/pull/5629

PR #5629, opened 2021-02-15
[illink] Type registrations fallback switch

radekdoulik:
Context: https://github.com/xamarin/xamarin-android/issues/5515
Context: https://github.com/dotnet/runtime/blob/master/docs/workflow/trimming/feature-switches.md#adding-new-feature-switch

Add new feature switch `XATypeRegistrationFallback` for type
registrations fallback. The switch default value is `false` for linked
net6 apps.

We can remove the fallback by the linker, because all the java to
managed type names lookup is now handled by typemap, so we should not
need this fallback anymore - at least not on device.

On desktop we still need it for Designer, which uses unlinked
assemblies and thus the fallback still works.

Also add lazy initialization of `packageLookup` field. That allows
linker to link the field away. That also improves startup performance,
as we don't create the dictionary in the static constructor.

apk size savings, `BuildReleaseArm64False/net6` test:

      -         221 assemblies/System.Private.CoreLib.dll
        Type System.String
          -          13 Method public string ToUpperInvariant ()
          -          20 Method public int LastIndexOf (char)
        Type System.Char
          -           8 Method public static char ToUpperInvariant (char)
        Type System.Globalization.TextInfo
          -         214 Method static string ToUpperAsciiInvariant (string)
          -          36 Method static char ToUpperInvariant (char)
          -          37 Method public string ToUpper (string)
          -          22 Method static char ToUpperAsciiInvariant (char)
      -      21,506 assemblies/Mono.Android.dll
        -             Type Java.Interop.__TypeRegistrations
        Type Java.Interop.TypeManager
          -             Field static System.Collections.Generic.Dictionary`2<string,System.Collections.Generic.List`1<System.Converter`2<string,System.Type>>> packageLookup
          +             Field static readonly bool <TypeRegistrationFallbackIsEnabled>k__BackingField
          +             Property bool TypeRegistrationFallbackIsEnabled { get; }
          -         184 Method static System.Type GetJavaToManagedType (string)
          -         110 Method public static void RegisterPackage (string, System.Converter`2<string,System.Type>)
          -         218 Method public static void RegisterPackages (string[], System.Converter`2<string,System.Type>[])
          +           3 Method static bool get_TypeRegistrationFallbackIsEnabled ()
          +          31 Method static bool InitializeTypeRegistrationFallbackIsEnabled ()
        Type Java.Interop.Tools.TypeNameMappings.JavaNativeTypeManager
          -          86 Method public static string ToCliType (string)
          -          98 Method static string ToCliTypePart (string)
          -          67 Method static string ToPascalCase (string, int)
    Summary:
      -      21,727 Assemblies -3.05% (of 712,159)
      -     122,880 Uncompressed assemblies -8.30% (of 1,480,704)

### DIFF
--- a/src/Mono.Android/Java.Interop/TypeManager.cs
+++ b/src/Mono.Android/Java.Interop/TypeManager.cs
@@ -212,9 +212,9 @@ namespace Java.Interop {
 		[MethodImplAttribute(MethodImplOptions.InternalCall)]
 		static extern Type monodroid_typemap_java_to_managed (string java_type_name);
 
-		private static bool TypeRegistrationFallbackIsEnabled { get; } = InitializeTypeRegistrationFallbackIsEnabled ();
-		private static bool InitializeTypeRegistrationFallbackIsEnabled () =>
-		    AppContext.TryGetSwitch ("Java.Interop.TypeManager.TypeRegistrationFallbackIsEnabled", out bool isEnabled) ? isEnabled : true;
+		static bool TypeRegistrationFallbackIsEnabled { get; } = InitializeTypeRegistrationFallbackIsEnabled ();
+		static bool InitializeTypeRegistrationFallbackIsEnabled () =>
+		    !AppContext.TryGetSwitch ("Java.Interop.TypeManager.TypeRegistrationFallbackIsEnabled", out bool isEnabled) || isEnabled;
 
 		internal static Type? GetJavaToManagedType (string class_name)
 		{
@@ -370,7 +370,7 @@ namespace Java.Interop {
 		static void LazyInitPackageLookup ()
 		{
 			if (packageLookup == null)
-				packageLookup = new Dictionary<string, List<Converter<string, Type?>>> ();
+				packageLookup = new Dictionary<string, List<Converter<string, Type?>>> (StringComparer.Ordinal);
 		}
 
 		public static void RegisterPackage (string package, Converter<string, Type> lookup)
--- a/src/Mono.Android/Java.Interop/TypeManager.cs
+++ b/src/Mono.Android/Java.Interop/TypeManager.cs
@@ -212,6 +212,10 @@ namespace Java.Interop {
 		[MethodImplAttribute(MethodImplOptions.InternalCall)]
 		static extern Type monodroid_typemap_java_to_managed (string java_type_name);
 
+		private static bool TypeRegistrationFallbackIsEnabled { get; } = InitializeTypeRegistrationFallbackIsEnabled ();
+		private static bool InitializeTypeRegistrationFallbackIsEnabled () =>
+		    AppContext.TryGetSwitch ("Java.Interop.TypeManager.TypeRegistrationFallbackIsEnabled", out bool isEnabled) ? isEnabled : true;
+
 		internal static Type? GetJavaToManagedType (string class_name)
 		{
 			Type? type = monodroid_typemap_java_to_managed (class_name);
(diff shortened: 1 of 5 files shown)


src/Mono.Android/Java.Interop/TypeManager.cs
```diff
@@ -353,10 +365,18 @@ public static void RegisterType (string java_class, Type t)
 			}
 		}
 
-		static Dictionary<string, List<Converter<string, Type?>>> packageLookup = new Dictionary<string, List<Converter<string, Type?>>> ();
+		static Dictionary<string, List<Converter<string, Type?>>>? packageLookup;
+
+		static void LazyInitPackageLookup ()
+		{
+			if (packageLookup == null)
+				packageLookup = new Dictionary<string, List<Converter<string, Type?>>> ();
```
grendello, line 373: I think the dictionary should use the `Ordinal` comparer (or `OrdinalIgnoreCase` if we expect the package name can differ in case) instead of the default culture-aware one.

src/Mono.Android/Java.Interop/TypeManager.cs
```diff
@@ -212,6 +212,10 @@ static Exception CreateJavaLocationException ()
 		[MethodImplAttribute(MethodImplOptions.InternalCall)]
 		static extern Type monodroid_typemap_java_to_managed (string java_type_name);
 
+		private static bool TypeRegistrationFallbackIsEnabled { get; } = InitializeTypeRegistrationFallbackIsEnabled ();
+		private static bool InitializeTypeRegistrationFallbackIsEnabled () =>
```
grendello, line 216: Can you drop the `private` keyword please? :) This is the default visibility, there's no need to specify it explicitely

grendello (changes requested):

src/Mono.Android/Java.Interop/TypeManager.cs
radekdoulik: This was just refactored, shoudn't harm to use the comparer though, so updated.

src/Mono.Android/Java.Interop/TypeManager.cs
radekdoulik: Updated.

grendello (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
```diff
@@ -22,13 +22,18 @@ This file contains the .NET 5-specific targets to customize ILLink
...
 
   <Target Name="_PrepareLinking"
       Condition=" '$(PublishTrimmed)' == 'true' "
       AfterTargets="ComputeResolvedFilesToPublishList"
       DependsOnTargets="GetReferenceAssemblyPaths;_CreatePropertiesCache">
     <ItemGroup>
+      <RuntimeHostConfigurationOption Include="Java.Interop.TypeManager.TypeRegistrationFallbackIsEnabled"
+          Condition="'$(XATypeRegistrationFallback)' != ''"
+          Value="$(XATypeRegistrationFallback)"
+          Trim="true" />
```
jonathanpeppers, line 36: Where is this value actually stored? Is it a `*.runtimeconfig.json` file?

I don't see how that would get loaded at runtime, as we didn't do anything to put it in the `.apk` -- I don't think the runtime would know how to load it from the `.apk` either.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/LinkerTests.cs
```diff
@@ -394,5 +394,28 @@ public unsafe bool MyMethod (Android.OS.IBinder windowToken, [global::Android.Ru
 				}
 			}
 		}
+
+		[Test]
+		public void TypeRegistrationsFallback ([Values (true, false)] bool enabled)
+		{
+			if (!Builder.UseDotNet)
+				return;
```
jonathanpeppers, line 402: If you make other changes here, you might put `Assert.Ignore("Test only valid on .NET 6")` or something.

radekdoulik: Updated.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
radekdoulik: The feature switch value can be indeed set in a `runtimeconfig.json` file if I understand it correctly. In our case we care only about the linker, where it is passed as `--feature Java.Interop.TypeManager.TypeRegistrationFallbackIsEnabled false` option during the build.

PR #5634, opened 2021-02-17
Make apk size decrease regression as well

radekdoulik:
To make us have the reference files up-to-date.

Use new apkdiff's `--descrease-is-regression` option is used.

Added documentation, explain how to obtain the reference files. Also added
new simple powershell script to update the reference files locally.

Lower the thresholds to catch smaller regressions.

Also mention the need of Release XA builds. Context: https://github.com/xamarin/xamarin-android/issues/5675

### DIFF
--- a/Documentation/project-docs/ApkSizeRegressionChecks.md
+++ b/Documentation/project-docs/ApkSizeRegressionChecks.md
@@ -1,15 +1,18 @@
 # apk size regression checks
 
 We are checking the apk sizes for regression in CI PR builds.
-The `BuildReleaseArm64` test is used as the builds target both
-legacy monodroid and NET6 frameworks. It also builds simple XA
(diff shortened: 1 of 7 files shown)


radekdoulik:
I need to update the reference files again, it looks like the net6 was using packages from different build.

dellis1972:
@radekdoulik can we make sure those .apkdesc are archived as part of the build on ci (if they are not already)? It might make it easier to update if we can just download the required file from the ci system.

radekdoulik:
> @radekdoulik can we make sure those .apkdesc are archived as part of the build on ci (if they are not already)? It might make it easier to update if we can just download the required file from the ci system.

They are already part of the test results archive. I plan to update the documentation about that and also about the local builds requiring Release configuration (context https://github.com/xamarin/xamarin-android/issues/5675)

Issue #5633, opened 2021-02-17
AndroidUseInterpreter should work in net6


joj:
In .net4 AndroidUseInterpreter used to switch the runtime directory with one that includes an "-interpreter" suffix. That would automatically make the runtime use the interpreter. For net6, the runtime includes both versions in the package,
~~and the way to switch interpreter on is to add this environment variable:~~
~~`debug.mono.runtime_args=--interp=-inline`~~

The parameter is at the moment unchanged for net6, meaning that it adds a prefix that doesn't work anymore, and it doesn't do the environment, which is needed. The ask is to switch to that and then make sure interpreter works (that may require some help from mono runtime).

Thanks! 
J

jonathanpeppers:
This would be adding some new logic to:

https://github.com/xamarin/xamarin-android/blob/214dd76d9441c9974775f0b29feec0d5d8d1097f/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1503-L1518

Basically emit some extra `@(AndroidEnvironment)` lines for .NET 6 when `$(AndroidUseInterpreter)` is `true`.

An additional work item would be to standardize on an MSBuild property for both iOS/Android, as `$(MTouchExtraArgs)` is used for enabling the interpreter at the moment.

lambdageek:
I tried `AndroidUseInterpreter` and it "works" in that the runtime tries to start up with interp support, but it breaks because it passes `MONO_AOT_MODE_LAST` which used to work with old mono, but not net6 mono.

I'm working on a PR to update to use `MONO_AOT_MODE_INTERP_ONLY` (on top of #5632)

lambdageek:
One thing we need to figure out for hot reload:  we need to pass `mono_ee_interp_init("-inline")` some time shortly after `mono_jit_set_aot_mode (mode);` when hot reload is enabled.

jonathanpeppers:
@lambdageek do we still need to do something on the MSBuild side when `$(AndroidUseInterpreter)` is `true`?

Do we need `debug.mono.runtime_args=--interp=-inline` anymore?

joj:
Can we make AndroidUseInterpreter opt out (meaning, default to true when debugging)? What do you think? That will enable these new functionalities by default, and allow users to turn it off if it's causing problems.

lambdageek:
There's two orthogonal issues:
1. Is AndroidUseInterpreter usable with net6 now? Yes.
[...]

PR #5635, opened 2021-02-18
[monodroid] Use MONO_AOT_MODE_INTERP_ONLY when available

lambdageek:
This depends on a fix to mono/mono#18893 (make interp only mode distinct from `MONO_AOT_MODE_LAST`) which was done in mono/mono#20159 but only for .NET 6, not previous Mono releases.

Fixes https://github.com/xamarin/xamarin-android/issues/5633

### DIFF
--- a/src/monodroid/jni/android-system.cc
+++ b/src/monodroid/jni/android-system.cc
@@ -660,9 +660,9 @@ AndroidSystem::setup_environment ()
 #if !defined (NET6)
 				aotMode = MonoAotMode::MONO_AOT_MODE_LAST;
 				aot_mode_last_is_interpreter = true;
-#else
+#else   // defined (NET6)
 				aotMode = MonoAotMode::MONO_AOT_MODE_INTERP_ONLY;
-#endif
+#endif  // !defined (NET6)
 				break;
 
 			default:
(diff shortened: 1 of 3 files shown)


lambdageek:
Haven't tried it yet.  still need to read the documentation on how to run the changes locally.  But it feels right.

To test it, I woudl run this code:
```csharp
  private static string OnInterp ()    {
        if (RuntimeFeature.IsDynamicCodeSupported) {
        if (RuntimeFeature.IsDynamicCodeCompiled) {
            return "JIT";
        } else {
            return "Interp";
        }
        } else {
        return "FullAOT";
        }
    }
```

grendello (approved):

lambdageek:
The "MSBuild Emulator Tests MSBuild With Emulator - macOS - One .NET" lane is failing `InstallWithoutSharedRuntime` and `MonoAndroidExportReferencedAppStartsFalseAssembliesDexes` which I don't think are related to this PR.

Not sure I understand what the "Smoke Tests APK Instrumentation - macOS" failure is saying.

grendello:
`InstallWithoutSharedRuntime` was known to be failing in #5665 and I don't think either of the failures here are due to this here PR. We can merge, @jonpryor?

jonpryor:
The crash in **Smoke Tests** > **APK Instrumentation** for `Mono.Android_Tests-Bundle` is concerning.  I've restarted those tests.

PR #5640, opened 2021-02-19
[One .NET] fix support for uppercase file names in class libraries

jonathanpeppers:
Context: https://github.com/xamarin/Xamarin.Forms/tree/main-handler/src/Platform.Renderers/src/Xamarin.Forms.Platform.Android/Resources/anim

When porting some projects to .NET in Xamarin.Forms/MAUI, given a file
name such as:

    Resources/anim/EnterFromRight.xml

The `Resource.designer.cs` file contained:

    public static int enterfromright = 2130771997;

However, in "legacy" Xamarin.Android the casing was preserved:

    public static int EnterFromRight = 2130771969;

I was able to reproduce this in `XASdkTests`. The problem was the new
`_ResolveAars` MSBuild target that resolves `.aar` files from
dependencies was not running in class libraries, only application
projects. My original thinking was that class libraries don't need
`@(AndroidAarLibrary)` defined, but they *do* for
`Resource.designer.cs` generation.

The only other consideration was to set `%(Pack)` to `false` by
default, considering the case:

    LibraryA.csproj -> <ProjectReference Include="..\LibraryB\LibraryB.csproj" />
    dotnet pack LibraryA.csproj -> LibraryA.nupkg
    dotnet pack LibraryB.csproj -> LibraryB.nupkg

We only want `LibraryB`'s `.aar` file in `LibraryB.nupkg`,
`LibraryA.nupkg` should not include it by default.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
@@ -18,7 +18,7 @@ projects.
     <_AarOutputPath>$(OutputPath)$(TargetName).aar</_AarOutputPath>
   </PropertyGroup>
 
-  <Target Name="_ResolveAars" Condition=" '$(AndroidApplication)' == 'true' ">
+  <Target Name="_ResolveAars">
     <ItemGroup>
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
/cc @PureWeen I think you were running into this.

Note that I think you should probably rename this anyway to avoid (our) bugs & be closer to standard Android naming. I would rename anything from `.axml` to `.xml` too: 

https://github.com/jonathanpeppers/Xamarin.Forms/commit/7627fd8c4891f31363b0d4ae12257ce06be6415b

dellis1972 (approved):

PR #5641, opened 2021-02-19
[monodroid] Refactor CMake scripts

grendello:
The main goal of this PR is to make our CMake scripts and macros, used to
build the native runtime and sqlite, better organized and the monodroid
build configuration more maintainable.

With the near future addition of NET6 support for our runtime builds, we
will doubling the number of runtime configurations we build and thus the
current way of configuring the runtime build via `monodroid.targets`
will become a tangle of CMake options with many repetitions of the same
fragments, making the whole setup fragile and tedious to maintain.

This commit addresses the issue by teaching `xaprepare` how to generate
the CMake commands from a single source (`build-tools/xaprepare/xaprepare/ConfigAndData/CmakeBuilds.cs`).

In addition, this commit introduces a Unix shell script,
`build-tools/scripts/build-monodroid`, which allows one to (re)build any
configuration of the `monodroid` runtime without having to run
msbuild (which would build ALL of the configurations, forcing the
developer to wait longer than necessary). The script uses a file
generated by `xaprepare` from the same `CmakeBuilds.cs` source, which
contains all the necessary commands to build any runtime configuration.
Running the script without any parameters will show a simple help
screen:

    Usage: build-monodroid [OPTIONS] <TARGET> [<TARGET> ...]

    where OPTIONS is one or more of:

      -o|--output=DIR     set output directory instead of the default monodroid 'obj' one
      -r|--rebuild        rebuild from scratch (this involves removing the output directory,
                          after prompting for confirmation)
      -f|--force          skip directory deletion prompt on rebuilds
      -h|--help           show this help screen

    Each <TARGET> is a name of the monodroid runtime configuration to build. Format:

       {ARCH}-[net6-][CHECKER-]{CONFIGURATION}

    where strings in lower case (e.g. 'net6') are to be used verbatim and the ones in
    upper case (e.g. 'ARCH') are placeholders as defined below (alternative forms for
    each value are specified in parentheses). All values are case-insensitive:

       ARCH is one of: armeabi-v7a (arm32, armv7), arm64-v8a (arm64, armv8), x86, x86_64 (x64), host, win32, win64
       CHECKER is one of: asan, ubsan
       CONFIGURATION is one of: release, debug

    Examples:

      build-monodroid arm64-release
      build-monodroid x86-net6-release host-debug
      build-monodroid x64-ubsan-release x86-net6-asan-debug

This commit also updates CMake to version 3.18.1, which allows us to use
a few improvements in the CMake script language to better support linker
options, specify compiler options in a more uniform manner and test more
reliably the supported compiler/linker arguments.


src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
```diff
@@ -17,7 +17,8 @@
       "version": "@SDK_PACK_VERSION@",
       "alias-to": {
         "osx-x64": "Microsoft.Android.Sdk.osx-x64",
-        "win-x64": "Microsoft.Android.Sdk.win-x64"
+        "win-x64": "Microsoft.Android.Sdk.win-x64",
+        "linux-x64": "Microsoft.Android.Sdk.linux-x64"
```
jonathanpeppers, line 21: I don't think we should add this line unless we are also pushing the `Microsoft.Android.Sdk.linux-x64` pack to our NuGet feed.

It doesn't look like we do yet: https://dev.azure.com/azure-public/vside/_packaging?_a=feed&feed=xamarin-impl%40Local

build-tools/xaprepare/xaprepare/xaprepare.targets
```diff
@@ -75,9 +80,10 @@
       <Replacement Include="@MonoRequiredMinimumVersion@=$(MonoRequiredMinimumVersion)" />
       <Replacement Include="@MonoRequiredMaximumVersion@=$(MonoRequiredMaximumVersion)" />
       <Replacement Include="@MonoSourceFullPath@=$(MonoSourceFullPath)" />
+      <Replacement Include="@NinjaPath@=$(NinjaPath)" />
       <Replacement Include="@Pkg7-Zip_CommandLine@=$(Pkg7-Zip_CommandLine)" />
       <Replacement Include="@PkgXamarin_LibZipSharp@=$(PkgXamarin_LibZipSharp)" />
-      <Replacement Include="@ProductVersion@=$(ProductVersion)" />
+      <Replacement Include="@ProductVersion@=$(ProdquctVersion)" />
```
radekdoulik, line 86: A typo?

jonpryor:
All the `Mono.Android_Tests` test runs in the Smoketest are failing, e.g. https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4490193&view=logs&j=cd6eb242-2e57-569b-620a-eaaded8e9a51&t=dfeb5e61-1ddf-5766-1b6f-cd254c8f44fa

Which looks like the BCL+`NetworkInterfaces` issue we discussed earlier today, and suggested that the new emulator version might be responsible.

However, "current master 9697749952ced01f22cbe29390d9877d83330bb8" has a fully green smoketest run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4490308&view=results

BCL tests are also fully green on 9697749952ced01f22cbe29390d9877d83330bb8.

I'm thus slightly dubious about the `NetworkInterface`-related errors in *this* PR being due to the emulator.

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
```diff
@@ -17,7 +17,7 @@
       "version": "@SDK_PACK_VERSION@",
       "alias-to": {
         "osx-x64": "Microsoft.Android.Sdk.osx-x64",
-        "win-x64": "Microsoft.Android.Sdk.win-x64"
+        "win-x64": "Microsoft.Android.Sdk.win-x64",
```
jonpryor, line 20: This results in invalid JSON (JSON doesn't like trailing commas), which may or may not implicate "other" things.  For sanity, we should have only valid JSON.

jonpryor:
> I'm thus slightly dubious about the `NetworkInterface`-related errors in this PR being due to the emulator.

Continuing on this, something is *wrong* with this PR.

Case in point:

  * Good: https://artprodscussu1.artifacts.visualstudio.com/A011b8bdf-6d56-4f87-be0d-0092136884d9/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2RldmRpdi9wcm9qZWN0SWQvMGJkYmM1OTAtYTA2Mi00YzNmLWIwZjYtOTM4M2Y2Nzg2NWVlL2J1aWxkSWQvNDQ5MDAyNy9hcnRpZmFjdE5hbWUvVGVzdCtSZXN1bHRzKy0rQVBLK0luc3RydW1lbnRhdGlvbistK21hY09T0/content?format=file&subPath=%2FDefault%2FMono.Android_Tests-Signed.apk
  * Bad: https://artprodscussu1.artifacts.visualstudio.com/A011b8bdf-6d56-4f87-be0d-0092136884d9/0bdbc590-a062-4c3f-b0f6-9383f67865ee/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL2RldmRpdi9wcm9qZWN0SWQvMGJkYmM1OTAtYTA2Mi00YzNmLWIwZjYtOTM4M2Y2Nzg2NWVlL2J1aWxkSWQvNDQ5MDE5My9hcnRpZmFjdE5hbWUvVGVzdCtSZXN1bHRzKy0rQVBLK0luc3RydW1lbnRhdGlvbistK21hY09T0/content?format=file&subPath=%2FDefault%2FMono.Android_Tests-Signed.apk

Those are `.apk` files.

Download, and grab `libmonodroid.so`:

```
% unzip Mono.Android_Tests-Signed.apk lib/x86/libmonodroid.so
```

Disassemble:

```
% $HOME/android-toolchain/ndk/toolchains/llvm/prebuilt/darwin-x86_64/bin/i686-linux-android-objdump -D -s  lib/x86/libmonodroid.so > a.s
```

Then grep the contents of `a.s`.  The "good" version contains multiple mentions of `netlink`:

```
% grep netlink a.s | wc -l
     152
```

The "bad" version…doesn't:

```
% grep netlink a.s
 5a40 006e6574 6c696e6b 00706173 73696e67  .netlink.passing
```

The bad version contains *only one* mention of `netlink`, *not* as a function.

*This* is why all the `NetworkInterface`-related tests are failing: the CMake cleanup broke something.

PR #5644, opened 2021-02-19
[Xamarin.Android.Build.Tasks] bump to System.Collections.Immutable 1.7.1

jonathanpeppers:
I keep getting this error in VS Windows:

    Detected package downgrade: System.Collections.Immutable from 1.7.1 to 1.7.0. Reference the package directly from the project to select a different version.
        Xamarin.Android.Build.Tasks -> Microsoft.Android.Build.BaseTasks -> System.Collections.Immutable (>= 1.7.1)
        Xamarin.Android.Build.Tasks -> System.Collections.Immutable (>= 1.7.0)

But command-line it seems to work fine?

It does seem like this package is behind compared to one used in
`xamarin-android-tools`, so let's just update it.

I also went ahead and created `$(SystemCollectionsImmutableVersion)`
in `Directory.Build.props` as this NuGet package is used in a few
places.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
@@ -50,7 +50,7 @@
     <PackageReference Include="NuGet.Protocol" Version="$(NuGetApiPackageVersion)" />
     <PackageReference Include="NuGet.Versioning" Version="$(NuGetApiPackageVersion)" />
     <PackageReference Include="System.CodeDom" Version="4.7.0" />
-    <PackageReference Include="System.Collections.Immutable" Version="1.7.0" />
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj
```diff
@@ -50,7 +50,7 @@
     <PackageReference Include="NuGet.Protocol" Version="$(NuGetApiPackageVersion)" />
     <PackageReference Include="NuGet.Versioning" Version="$(NuGetApiPackageVersion)" />
     <PackageReference Include="System.CodeDom" Version="4.7.0" />
-    <PackageReference Include="System.Collections.Immutable" Version="1.7.0" />
+    <PackageReference Include="System.Collections.Immutable" Version="1.7.1" />
```
radekdoulik, line 53: It might be worth to use property for it and put it in `Directory.Build.props` and update the other `PackageReference`s?

radekdoulik (approved):

PR #5646, opened 2021-02-19
Set UseNativeHttpHandler project default

steveisok:
In https://github.com/dotnet/runtime/pull/47083, we are introducing the UseNativeHttpHandler feature switch.  This change sets the default to true for Android projects.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
@@ -62,7 +62,7 @@
     <GenerateApplicationManifest Condition=" '$(GenerateApplicationManifest)' == '' ">true</GenerateApplicationManifest>
 
     <!-- Enable/Disable default ruuntime features-->
-    <UseNativeHttpHandler>true</UseNativeHttpHandler>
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -60,6 +60,9 @@
     <AndroidManifest Condition=" '$(AndroidManifest)' == '' and Exists ('Properties\AndroidManifest.xml') and !Exists ('AndroidManifest.xml') ">Properties\AndroidManifest.xml</AndroidManifest>
     <AndroidManifest Condition=" '$(AndroidManifest)' == '' ">AndroidManifest.xml</AndroidManifest>
     <GenerateApplicationManifest Condition=" '$(GenerateApplicationManifest)' == '' ">true</GenerateApplicationManifest>
+
+    <!-- Enable/Disable default ruuntime features-->
+    <UseNativeHttpHandler>true</UseNativeHttpHandler>
```
jonathanpeppers, line 65: I think this should be:

```suggestion
    <UseNativeHttpHandler Condition=" '$(UseNativeHttpHandler)' == '' ">true</UseNativeHttpHandler>
```

So that users can override it in their `.csproj` file.

dellis1972 (approved):

radekdoulik (approved):
LGTM.

We now set the other feature switches in `Microsoft.Android.Sdk.ILLink.targets`, the `Microsoft.Android.Sdk.DefaultProperties.targets` looks like better place though. I think I will move the rest of them there.

PR #5653, opened 2021-02-23
[Mono.Android] Bind API-S Developer Preview 1.

jpobst:
Bindings for Android API-S DP1.

https://android-developers.googleblog.com/2021/02/android-12-dp1.html

Breaking changes:
- 2 new methods were added to the `Deprecated` interface.  Given that this is an annotation, this should not be used as a regular interface and should not break anything: https://developer.android.com/reference/java/lang/Deprecated.


jonathanpeppers (commented):
When this goes in, can you add `monoandroid12.0` here?

https://github.com/xamarin/xamarin-android/blob/e2a5a12b26db732253a6e54cc39c98dc07ec3f14/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.NuGet.targets#L14-L33

This is the workaround until we get:

https://github.com/dotnet/designs/blob/main/accepted/2021/net6.0-tfms/net6.0-tfms.md

grendello (approved):

jonathanpeppers (approved):

PR #5665, opened 2021-02-26
[monodroid] Build the Xamarin.Android runtime for net6

grendello:
This commit makes changes to build the native Xamarin.Android runtime
against Mono VM shipped from the `dotnet/runtime` repository in
nuget packages which also contain header files necessary to embed
Mono in Xamarin.Android.

Xamarin.Android runtime is now built separately for the "legacy"
target, using Mono from the SDK archives, and for net6. This
change required introduction of new target directories in
`$XA_INSTALL_ROOT/lib`:

  * arm64-v8a-net6 armeabi-v7a-net6 x86_64-net6 x86-net6

These directories now contain only the shared libraries together
comprising the Xamarin.Android runtime:

  * libmono-android.*.so libxamarin-debug-app-helper*.so
  * libxamarin-app.so

Mono runtime libraries are not copied from their respective nuget
packages. Compared to the "legacy" builds, we are missing the
`libMonoPosixHelper.so` library. The helper DSO will be eventually
placed in a `Mono.Posix` nuget, once we are able to separate
`Mono.Posix` files from the Mono sources and place them in a
separate repository.

`xaprepare` is modified to generate MSBuild and cmake fragments
responsible for building net6 runtime binaries.

Native runtime now doesn't export any p/invoke entry points,
instead it uses the new `PINVOKE_OVERRIDE` mechanism to handle the
p/invoke dispatch internally. This allows us to not only hide the
public symbols, but also remove the `libxa-internal-api.so`
library completely.


grendello:
I won't be working on more TODO items from the list in the OP - the PR is already big enough. If the PR is green, let's please commit it.

ThirdPartyNotices.txt
```diff
@@ -14,7 +14,9 @@ The attached notices are provided for information only.
 1. android/platform/tools/base (https://android.googlesource.com/platform/tools/base/+/d41d662dbf89f9b60ca6256415a059c0107749b8/sdk-common/NOTICE)
 2. bazelbuild/bazel (https://github.com/bazelbuild/bazel/)
 3. force-net/crc32.net (https://github.com/force-net/Crc32.NET)
-4. nunit/nunitlite (https://github.com/nunit/nunitlite/)
+4. google/bionic (https://android.googlesource.com/platform/bionic/)
+5. nunit/nunitlite (https://github.com/nunit/nunitlite/)
+6. tessil/robin-map (https://github.com/Tessil/robin-map)
```
jonpryor, line 19: This seems odd: the in-repro `ThirdPartyNotices.txt` only needs to contain stuff *within this repo*, e.g. files in `src-ThirdParty`.  A *git submodule* doesn't need to be listed in this file; instead, it needs to be listed in the *generated* `bin/$(Configuration)/lib/xamarin.android/ThirdPartyNotices.txt` file.

src/monodroid/jni/cxx-abi/bionic_futex.hh
```diff
@@ -0,0 +1,84 @@
+/*
```
jonpryor, line 1: *This* file should likely be in `src-ThirdParty/ndk`, or some other "meaningful" `src-ThirdParty` subdirectory.

src/monodroid/jni/mono_android_Runtime.h
```diff
@@ -39,6 +39,7 @@ JNIEXPORT void JNICALL Java_mono_android_Runtime_register
 JNIEXPORT void JNICALL Java_mono_android_Runtime_notifyTimeZoneChanged
   (JNIEnv *, jclass);
 
+#if !defined (ANDROID)
```
jonpryor, line 42: I assume this was a *manual* addition?  This is usually a generated file, so if this is a manual addition that's…problematic?  It might be better to wrap the `#include "mono_android_Runtime.h"` instead.

jonpryor: …though it's only the Desktop-related members here.  Ugh.

jonpryor: might be a good opportunity to introduce the long-considered `DesignerRuntime` class, have the `native` methods declared *there*, and update the current Designer-centric `Runtime` methods to instead invoke the `DesignerRuntime` methods.  This means we'd have a *new* `mono_android_DesignerRuntime.h` file, which can be appropriately "guarded".

src/monodroid/jni/monodroid-glue-internal.hh
```diff
@@ -10,10 +11,65 @@
...
+		using pinvoke_library_map = tsl::robin_map<
+			std::string,
+			pinvoke_api_map_ptr,
+			std::hash<std::string>,
+			std::equal_to<std::string>,
+			std::allocator<std::pair<std::string, pinvoke_api_map_ptr>>,
+			true
+		>;
+
+		static constexpr pinvoke_library_map::size_type LIBRARY_MAP_INITIAL_BUCKET_COUNT = 1;
```
jonpryor, line 70: Do we really want an initial bucket size of `1`? :-)

src/monodroid/jni/monodroid-glue-internal.hh
```diff
@@ -10,10 +11,65 @@
...
+#endif
```
jonpryor, line 71: Convention (if not always followed): `#endif`/`#elif`/etc. should have a comment to mention what symbols they're related to:

```cpp
#endif  /* defined (NET6) */
```

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -99,11 +98,17 @@ using namespace xamarin::android::internal;
 // implementation details as it would prevent mkbundle from working
 #include "mkbundle-api.h"
 
+#if !defined (NET6)
 #include "config.include"
+#endif
 #include "machine.config.include"
```
jonpryor, line 104: Do we still need `machine.config.xml`?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1125,14 +1121,77 @@ MonodroidRuntime::monodroid_dlopen_log_and_return (void *handle, char **err, con
...
+	}
+
+	constexpr char DSO_EXTENSION[] = ".so";
+	constexpr size_t DSO_EXTENSION_SIZE = sizeof(DSO_EXTENSION);
+
+	if (utils.ends_with (name, DSO_EXTENSION)) {
+		return monodroid_dlopen (name, flags, err);
+	}
+
+	size_t len = strlen (name) + DSO_EXTENSION_SIZE; // includes the trailing \0
```
jonpryor, line 1188: Do we check/how do we prevent. `strlen(name)` from being `std::numeric_limits<size_t>::max`, causing the addition to be an overflow?

Should this use one of our "checked addition" macros?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1635,6 +1695,20 @@ MonodroidRuntime::Java_mono_android_Runtime_initInternal (JNIEnv *env, jclass kl
...
+#if defined (NET6)
+	{
+		PInvokeOverrideFn pinvoke_override_cb = monodroid_pinvoke_override;
+		char ptr_str[20];
+		snprintf (ptr_str, sizeof(ptr_str), "%p", pinvoke_override_cb);
+
+		const char* property_keys[] { "PINVOKE_OVERRIDE" };
+		const char* property_values[] { ptr_str };
+
+		log_warn (LOG_DEFAULT, "Initializing .NET6 Mono VM (override callback at %p, passed as %s", pinvoke_override_cb, ptr_str);
```
jonpryor, line 1707: Do we want this printed in *Release* builds?  Or should this be `log_debug()`?

src/monodroid/jni/pinvoke-override-api.cc
```diff
@@ -0,0 +1,709 @@
...
+{
+        intptr_t rv = env->CallStaticIntMethod (monodroidRuntime.get_java_class_System (), monodroidRuntime.get_java_class_method_System_identityHashCode (), v);
+        return (void*) rv;
+}
+
+static void*
+_monodroid_timezone_get_default_id ()
+{
+	JNIEnv *env          = osBridge.ensure_jnienv ();
+	jmethodID getDefault = env->GetStaticMethodID (monodroidRuntime.get_java_class_TimeZone (), "getDefault", "()Ljava/util/TimeZone;");
```
jonpryor, line 175: We want to look these `jmethodID`s up every time, instead of caching them?

src/monodroid/jni/pinvoke-override-api.cc
```diff
@@ -0,0 +1,709 @@
...
+	jmethodID getID      = env->GetMethodID (monodroidRuntime.get_java_class_TimeZone (), "getID",      "()Ljava/lang/String;");
+	jobject d            = env->CallStaticObjectMethod (monodroidRuntime.get_java_class_TimeZone (), getDefault);
+	jstring id           = reinterpret_cast<jstring> (env->CallObjectMethod (d, getID));
+	const char *mutf8    = env->GetStringUTFChars (id, nullptr);
+	char *def_id         = strdup (mutf8);
```
jonpryor, line 180: This should null-check `mutf8`?

src/monodroid/jni/pinvoke-override-api.cc
```diff
@@ -0,0 +1,709 @@
...
+
+static void*
+monodroid_get_dylib (void)
+{
+	return nullptr;
+}
+
+#define PINVOKE_SYMBOL(_sym_) { #_sym_, reinterpret_cast<void*>(&_sym_) }
+
+MonodroidRuntime::pinvoke_api_map MonodroidRuntime::xa_pinvoke_map = {
```
jonpryor, line 339: Does this table need to be sorted?  If so, that should be commented.

src/monodroid/jni/monodroid-glue.cc
grendello: it contains a bunch of settings regarding pub tokens, remoting configuration etc - not **sure** we need it, but it might be needed for some things around the runtime

src/monodroid/jni/monodroid-glue.cc
grendello: Eventually `log_debug`, I'd rather keep it `log_warn` for now - until we are ready for final release

src/monodroid/jni/pinvoke-override-api.cc
grendello: Yes, we did that to speed up startup. Caching it at the later stage would probably require locks. We can implement that, but not in this PR.

src/monodroid/jni/pinvoke-override-api.cc
grendello: No, it doesn't need to be sorted. `robin-map` is an equivalent to `std::unordered_map`, I sorted it just for convenience.

src/monodroid/jni/monodroid-glue-internal.hh
grendello: Yep, it avoids an expensive call to guard locks around a static variable inside one of the `robin-map` functions

src/monodroid/jni/mono_android_Runtime.h
grendello: Definitely not material for this PR :)

jonpryor:
WIP commit message;

```markdown
Context: https://github.com/dotnet/runtime/issues/48416
Context: https://github.com/dotnet/runtime/pull/44505
Context: https://github.com/xamarin/xamarin-android/issues/5537

Update the `src/monodroid` build system to build the native
Xamarin.Android runtime libraries (e.g. `libmono-android.debug.so`)
against the MonoVM shipped from the `dotnet/runtime` repository in
the `Microsoft.NETCore.App.Runtime.*` NuGet packages.  These NuGet
packages also contain the header files needed to access and use
MonoVM's embedding API.

We thus ~double the number of libraries built by `src/monodroid`,
with one set of builds against the "legacy" Mono, using Mono from the
SDK archives, and another set of builds against .NET 6 MonoVM.

This change requires the introduction of new target directories in
`$(XAInstallPrefix)xbuild/Xamarin/Android/lib` (e.g.
`bin/Debug/lib/xamarin.android/xbuild/Xamarin/Android/lib`):

  * `arm64-v8a-net6`
  * `armeabi-v7a-net6`
  * `x86_64-net6`
  * `x86-net6`

These directories now contain only the shared libraries together
comprising the Xamarin.Android native runtime libraries:

  * `libmono-android.*.so`
  * `libxamarin-debug-app-helper*.so`
  * `libxamarin-app.so`

Mono runtime libraries are *not* copied from their respective NuGet
packages; the .NET 6 NuGet packages are required to obtain them.
Compared to the "legacy" builds, we are missing the 
libMonoPosixHelper.so` library.  This helper lib will eventually be
placed in a `Mono.Posix` NuGet, once we are able to separate
`Mono.Posix` files from the Mono sources and place them in a
[separate repository][0].

`xaprepare` is modified to generate MSBuild and cmake fragments
responsible for building net6 runtime binaries.

The `libmono-android.*.so` native libraries no longer need to export
any entry points for P/Invoke purposes.  Instead, the new
`PINVOKE_OVERRIDE` [runtime property][1] is used to hook into the
P/Invoke lookup process.  This allows us to not only hide the public
symbols, but also to remove the `libxa-internal-api.so`
library (d583b7c2) completely.

[0]: https://github.com/mono/mono.posix
[1]: https://docs.microsoft.com/en-us/dotnet/core/tutorials/netcore-hosting#step-3---prepare-runtime-properties
```

PR #5682, opened 2021-03-03
[Xamarin.Android.Build.Tasks] Add properties holding the Final VersionCode/Name.

dellis1972:
Fixes https://work.azdo.io/1005105

Users sometime want to use the values of `versionCode` and/or `versionName`
from the `AndroidManifest.xml` in their post build events. The problem
is if they are using things like `AndroidVersionCodePattern`, these values
are calculated during the build.

Users can make use of Post Build Tasks and `XmlPeek` to manually get
these values, but it would be nice to have some official properties which
they can use to get the final values.


dellis1972:
@jonpryor So this is still draft, but I could use a review to make sure this idea is good or not.

PR #5685, opened 2021-03-03
[ci] Use JDK 1.8 for legacy MSBuild test jobs

pjcollins:
The legacy MSBuild test jobs have been updated to use JDK 1.8 in order
to retain some JDK 1.8 test coverage.  JDK 1.8 is still installed by VS
and VS Mac, and it is likely what the vast majority of our customers
use today.

The Xamarin.ProjectTools.Builder and Xamarin.ProjectTools.DotNetCLI
classes are used for all Xamarin.Android.Build tests that need to invoke
msbuild or dotnet build.  Both of these wrapper classes will append
`/p:JavaSdkDirectory=$PATH` to all msbuild/dotnet build invocations, and
they use the same logic from Xamarin.ProjectTools.AndroidSdkResolver to
look up the JDK path.  See 16e971f29fc39639a9a245811e10bf9fd3c5047e for
more details.

This look up first checks to see if `$(JavaSdkDirectory)` is set in
Configuration.OperatingSystem.props.  This file is generated by an
xaprepare step that does not run on our test jobs, so the look up will
then check the value of `$JI_JAVA_HOME`. Overriding this environment
variable should ensure that all of our legacy MSBuild test jobs run
against JDK 1.8.

A few lint tests have been disabled when running against JDK 1.8. See 
https://github.com/xamarin/xamarin-android/issues/5698 for more info.

The `FixLintOnWindows` workaround has been removed. The xaprepare
tool installs the cmdline-tools Android SDK package, and the versions
of lint included in this package do not have this issue.  Our test
environments will never attempt to use the old version of lint found in
the tools folder.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Text;
 
 namespace Xamarin.ProjectTools
 {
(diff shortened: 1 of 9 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
```diff
@@ -60,6 +61,30 @@ public static string GetJavaSdkPath ()
...
+		{
+			if (string.IsNullOrEmpty (JavaSdkVersionString)) {
+				var javaPath = Path.Combine (GetJavaSdkPath (), "bin", "java");
+				var psi = new ProcessStartInfo (javaPath, "-version") {
+					CreateNoWindow = true,
+					RedirectStandardOutput = true,
+					RedirectStandardError = true,
+					WindowStyle = ProcessWindowStyle.Hidden,
+					UseShellExecute = false,
+				};
```
jonathanpeppers, line 76: I've seen `java -version` randomly take 2 seconds sometimes. Can we cache the result, so this would only be called once?

pjcollins: The `JavaSdkVersionString` static field used here should prevent subsequent calls from having to invoke java

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs
```diff
@@ -391,7 +391,7 @@ protected bool BuildInternal (string projectOrSolution, string target, string []
 							stdout.Set ();
 					};
 					p.StartInfo = psi;
-					Console.WriteLine ($"{psi.FileName} {psi.Arguments}");
+					Console.WriteLine ($"{psi.FileName} {psi.Arguments} {string.Join (" ", File.ReadAllLines (responseFile))}");
```
jonathanpeppers, line 394: I think the response file would get zipped up in the test results. Did you need this to figure out a problem?

pjcollins: I added this to check the JavaSdkDirectory value being used, but kept it because I thought some of the extra output was helpful and figured it likely has a negligible performance impact.

pjcollins: I ended up backing this out as it's not necessary.

PR #5692, opened 2021-03-04
[monodroid] Handle Mono log messages in XA runtime

grendello:
Mono produces a lot of messages at various levels of severity which, if
enabled, can quickly fill the logcat buffer also contributing to a
higher signal-to-noise ratio for those users who don't need the
messages.  In order to put the amount of logging under control, this
commit implements logging handlers which take over the logging
responsibility from Mono and put it on Xamarin.Android.  Three log
handlers are registered:

  * Mono tracing+logging
  * `stdout`
  * `stderr`

At the same time, the `debug.mono.log` system property gains two new
options, `mono_log_mask` and `mono_log_level`.  Both of them require a
value (described below), following the equals sign.

`mono_log_mask` corresponds to the `MONO_LOG_MASK` environment
variable (described in detail in the `mono(1)` manpage) and it accepts a
comma-separated list of categories for which to enable logging.  The
currently known categories are:

  * `asm` (assembly loader)
  * `type`
  * `dll` (native library loader)
  * `gc` (garbage collector)
  * `cfg` (config file loader)
  * `aot` (precompiler)
  * `security` (e.g. Moonlight CoreCLR support)
  * `threadpool` (thread pool generic)
  * `io-selector` (async socket operations)
  * `io-layer` (all of I/O layer - processes, files, sockets, events,
     semaphores, mutexes and  handles), or more finely grained:
    * `io-layer-process`
    * `io-layer-file`
    * `io-layer-socket`
    * `io-layer-event`
    * `io-layer-semaphore`
    * `io-layer-mutex`
    * `io-layer-handle`
  * `all`

By default Xamarin.Android disables all Mono logging, unless the
`assembly` or `gc` categories are enabled in the `debug.mono.log` system
property.  In such instance, `assembly` enables the `dll` and `asm` Mono
categories whicle `gc` enables the Mono `gc` category.

`mono_log_level` corresponds to the `MONO_LOG_LEVEL` environment
variable and accepts as its value one of the following log levels:

  * `error`
  * `critical`
  * `warning`
  * `message`
  * `info`
  * `debug`

`mono_log_level` defaults to `error`.

Additionally, this commit adds some more version information optionally
logged at the application startup:

  * NDK version Xamarin.Android was built with
  * NDK API level Xamarin.Android was built for
  * Machine architecture for which Xamarin.Android was built
  * Configuration for which Xamarin.Android was built (`Release` or
    `Debug`)
  * Date on which Xamarin.Android was built

Currently, we don't do any processing or filtering on the Mono log
messages, but it is possible to add such ability (e.g. log to a
different logcat buffer than the default one, log to file, ignore
certain messages etc) in the future, if needed.

### DIFF
--- a/src/monodroid/jni/monodroid-glue.cc
+++ b/src/monodroid/jni/monodroid-glue.cc
@@ -30,6 +30,7 @@
 #include <mono/metadata/mono-config.h>
 #include <mono/metadata/mono-debug.h>
 #include <mono/utils/mono-dl-fallback.h>
+#include <mono/utils/mono-logger.h>
 
 #include "mono_android_Runtime.h"
 
@@ -80,6 +81,7 @@
 #include "xamarin-app.hh"
 #include "timing.hh"
 #include "xa-internal-api-impl.hh"
+#include "build-info.hh"
 
 #ifndef WINDOWS
 #include "xamarin_getifaddrs.h"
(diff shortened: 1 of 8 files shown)


src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1760,6 +1852,21 @@ MonodroidRuntime::Java_mono_android_Runtime_initInternal (JNIEnv *env, jclass kl
...
+	} else {
+		// `mono_log_level` cannot be freed here, Mono VM stores it internally
+		mono_trace_set_level_string (mono_log_level);
+	}
+
+	setup_mono_tracing (mono_log_mask);
+
+#if defined (NET6)
+	install_logging_handlers ();
+#endif
```
jonpryor, line 1867: Ending comments please? :-)

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1881,13 +1988,33 @@ MonodroidRuntime::Java_mono_android_Runtime_initInternal (JNIEnv *env, jclass kl
 	/* the first assembly is used to initialize the AppDomain name */
 	create_and_initialize_domain (env, klass, runtimeApks, assemblies, nullptr, assembliesPaths, loader, /*is_root_domain:*/ true, /*force_preload_assemblies:*/ false);
 
+#if defined (ANDROID) && !defined (NET6)
+	// Mono from mono/mono has a bug which requires us to install the handlers after `mono_init_jit_version` is called
+	install_logging_handlers ();
+#endif
```
jonpryor, line 1994: Ditto.

jonpryor:
If `mono_log_mask` corresponds to `MONO_LOG_MASK` in **mono**(1), why not just call it `MONO_LOG_MASK`?  This should make copy/paste easier?

grendello:
> If `mono_log_mask` corresponds to `MONO_LOG_MASK` in **mono**(1), why not just call it `MONO_LOG_MASK`? This should make copy/paste easier?

It would stand out from the other `debug.mono.log` parameters and typing with shift being held down is annoying :)

radekdoulik:
For the record, adding apk size changes:
```
  +       7,348 lib/x86/libmonodroid.so
                  Section size difference
    +       5,488 .text
    +         240 .rodata
    +         143 .dynstr
    +          80 .dynsym
    +          80 .plt
    +          76 .eh_frame
    +          40 .hash
    +          40 .rel.plt
    +          20 .got.plt
    +          16 .rel.dyn
    +          16 .eh_frame_hdr
    +          10 .gnu.version
    +           8 .got
  +       5,068 lib/armeabi-v7a/libmonodroid.so
                  Section size difference
    +       3,000 .text
    +         168 .rodata
    +         143 .dynstr
    +          80 .dynsym
    +          80 .plt
    +          40 .hash
    +          40 .rel.plt
    +          20 .got.plt
    +          16 .rel.dyn
    +          12 .ARM.extab
    +          10 .gnu.version
    +           8 .ARM.exidx
    +           8 .got
  +           0 lib/armeabi-v7a/libmonodroid_bundle_app.so
                  Symbol size difference
    +           7 assembly_data_Mono_Android_dll
  +           0 lib/x86/libmonodroid_bundle_app.so
                  Symbol size difference
    +           7 assembly_data_Mono_Android_dll
Summary:
  +           0 Other entries 0.00% (of 1,274,559)
  +           0 Dalvik executables 0.00% (of 2,774,432)
  +      12,416 Shared libraries 0.06% (of 22,121,708)
  +       4,096 Package size difference 0.03% (of 16,168,950)
```

radekdoulik (commented):
I have updated `Xamarin.Forms_Performance_Integration-Signed-Release-Bundle.apkdesc` reference file.

We have accumulated smaller changes from last update of the file and
when this PR added 12k of shared libraries, it triggered the regression
check.

The 12k looks appropriate for this change.

Only thing to consider is whether we want it in release libraries? I guess yes, just wanted to mention it in case it is not intended.

Issue #5622, opened 2021-02-12
[illink] New feature switch to disable usage of S.C.WriteLine in our Logger


radekdoulik:
We use `WriteLine` method as fallback in our `Android.Runtime.Logger` in case liblog is not present (originally intended for desktop use). https://github.com/xamarin/xamarin-android/blob/39576f237961fb1f0d05a57bb988ed742620c465/src/Mono.Android/Android.Runtime/Logger.cs#L58-L59

It pulls in `AndroidLogPrint` in simple XA apps.

```
--- Raw dependencies: 'AndroidLogPrint' --------------------------------

--- Method:System.Void Interop/Logcat::AndroidLogPrint(Interop/Logcat/LogLevel,System.String,System.String) dependencies ---
Dependency #1
        Method:System.Void Interop/Logcat::AndroidLogPrint(Interop/Logcat/LogLevel,System.String,System.String)
        | Method:System.Void System.LogcatStream::Write(System.ReadOnlySpan`1<System.Byte>) [1 deps]
        | TypeDef:System.LogcatStream:System.Console.dll [3 deps]
        | Method:System.Void System.LogcatStream::.ctor() [1 deps]
        | Method:System.IO.Stream System.ConsolePal::OpenStandardOutput() [1 deps]
        | Method:System.IO.Stream System.Console::OpenStandardOutput() [1 deps]
        | Method:System.IO.TextWriter System.Console::<get_Out>g__EnsureInitialized|25_0() [1 deps]
        | Method:System.IO.TextWriter System.Console::get_Out() [1 deps]
        | Method:System.Void System.Console::WriteLine(System.String,System.Object,System.Object,System.Object) [1 deps]
        | Method:System.Void Android.Runtime.Logger::Log(Android.Runtime.LogLevel,System.String,System.String) [10 deps]
        | TypeDef:Android.Runtime.Logger:Mono.Android.dll [16 deps]
        | Other:C:\Users\rodo\android-toolchain\dotnet\packs\Microsoft.Android.Sdk.win-x64\11.0.200-ci.master.104\targets\..\PreserveLists\Mono.Android.xml

Found 1 matches
```

If we hide that behind feature switch, whole `System.Console.dll` or at least part of it will be linked away for Release apps.

marek-safar:
That code should be just deleted. The Console.WriteLine calls the exact same method https://github.com/dotnet/runtime/blob/master/src/libraries/System.Console/src/System/ConsolePal.Android.cs#L19

[...]

PR #5712, opened 2021-03-08
[illink] Do not fallback to S.Console in Logger

radekdoulik:
Context: https://github.com/xamarin/xamarin-android/issues/5622

Do not fallback to `System.Console.WriteLine` in our logger on devices.
We still need that for designer. That saves another 2.83% of assembly
size in simple XA `BuildReleaseArm64` test app.

Lets use the type registration fallback feature switch to hide
the fallback and rename it to `VSAndroidDesigner` as both
of these fallbacks are only used by the designer. Also move
it to `AndroidEnvironment` class.

Update the `CheckIncludedAssemblies` test, we just got rid of another
assembly :-)

Size savings, simple XA/net6 `BuildReleaseArm64` test:

    Summary:
      -         202 Other entries -0.35% (of 58,410)
      +           0 Dalvik executables 0.00% (of 316,728)
      -      19,471 Assemblies -2.83% (of 688,072)
      -          64 Shared libraries -0.00% (of 5,147,624)
      -      37,888 Uncompressed assemblies -2.82% (of 1,345,024)
      -      20,555 Package size difference -0.71% (of 2,905,966)

XForms/XA/net6 `BuildReleaseArm64` test:

    Summary:
      +           0 Other entries 0.00% (of 917,033)
      +           0 Dalvik executables 0.00% (of 3,455,720)
      +         177 Assemblies 0.00% (of 5,650,122)
      +           0 Shared libraries 0.00% (of 5,222,816)
      -       1,024 Uncompressed assemblies -0.01% (of 12,650,496)
      +           0 Package size difference 0.00% (of 9,886,918)

### DIFF
--- a/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
+++ b/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
@@ -358,7 +358,7 @@ namespace Android.Runtime {
 		internal static bool VSAndroidDesignerIsEnabled { get; } = InitializeVSAndroidDesignerIsEnabled ();
 
 		static bool InitializeVSAndroidDesignerIsEnabled () =>
-		    !AppContext.TryGetSwitch ("Java.Interop.TypeManager.VSAndroidDesignerIsEnabled", out bool isEnabled) || isEnabled;
+		    !AppContext.TryGetSwitch ("Xamarin.Android.VSAndroidDesigner.IsSupported", out bool isEnabled) || isEnabled;
 
 		class _Proxy : IWebProxy {
 			readonly ProxySelector selector = ProxySelector.Default!;
--- a/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
+++ b/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
@@ -355,6 +355,11 @@ namespace Android.Runtime {
 			return Activator.CreateInstance (handlerType);
 		}
 
+		internal static bool VSAndroidDesignerIsEnabled { get; } = InitializeVSAndroidDesignerIsEnabled ();
+
+		static bool InitializeVSAndroidDesignerIsEnabled () =>
+		    !AppContext.TryGetSwitch ("Java.Interop.TypeManager.VSAndroidDesignerIsEnabled", out bool isEnabled) || isEnabled;
+
 		class _Proxy : IWebProxy {
 			readonly ProxySelector selector = ProxySelector.Default!;
 
--- a/src/Mono.Android/Android.Runtime/Logger.cs
+++ b/src/Mono.Android/Android.Runtime/Logger.cs
@@ -55,7 +55,7 @@ namespace Android.Runtime {
 						hasNoLibLog = true;
 					}
 				}
-				if (hasNoLibLog)
+				if (AndroidEnvironment.VSAndroidDesignerIsEnabled && hasNoLibLog)
 					System.Console.WriteLine ("[{0}] {1}: {2}", level, appname, line);
 			}
 		}
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
@@ -74,7 +74,7 @@
     <UseNativeHttpHandler Condition="'$(UseNativeHttpHandler)' == ''">true</UseNativeHttpHandler>
 
     <!-- XA feature switches defaults -->
-    <XATypeRegistrationFallback Condition="'$(XATypeRegistrationFallback)' == ''">false</XATypeRegistrationFallback>
+    <VSAndroidDesigner Condition="'$(VSAndroidDesigner)' == ''">false</VSAndroidDesigner>
 
   </PropertyGroup>
   <PropertyGroup Condition=" '$(AndroidApplication)' == 'true' and '$(GenerateApplicationManifest)' == 'true' ">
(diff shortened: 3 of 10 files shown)


src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
```diff
@@ -355,6 +355,11 @@ static IWebProxy GetDefaultProxy ()
 			return Activator.CreateInstance (handlerType);
 		}
 
+		internal static bool VSAndroidDesignerIsEnabled { get; } = InitializeVSAndroidDesignerIsEnabled ();
+
+		static bool InitializeVSAndroidDesignerIsEnabled () =>
+		    !AppContext.TryGetSwitch ("Java.Interop.TypeManager.VSAndroidDesignerIsEnabled", out bool isEnabled) || isEnabled;
```
jonpryor, line 361: Naming quibble? Should this instead be `Xamarin.Android.AndroidDesigner.IsSupported`?

Quasi-rationale:

  * This is specific to Xamarin.Android, not Java.Interop, so a `Java.Interop` prefix doesn't quite make sense.
  * While this is used by `TypeManager`, it's no longer *limited* to the TypeManager, so I'm not sure that belongs.
  * A [quick perusal of `dotnet/runtime`](https://github.com/dotnet/runtime/search?q=AppContext+TryGetSwitch&type=code) shows existing switch names such as `System.Diagnostics.Tracing.EventSource.IsSupported`, `System.StartupHookProvider.IsSupported`, `System.Resources.UseSystemResourceKeys`, `System.Diagnostics.Tracing.EventSource.IsSupported`, etc. Thus suffixes of `Supported` or `Enabled`, or prefixes of `Use`, appear to be "common".

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -74,7 +74,7 @@
     <UseNativeHttpHandler Condition="'$(UseNativeHttpHandler)' == ''">true</UseNativeHttpHandler>
 
     <!-- XA feature switches defaults -->
-    <XATypeRegistrationFallback Condition="'$(XATypeRegistrationFallback)' == ''">false</XATypeRegistrationFallback>
+    <VSAndroidDesigner Condition="'$(VSAndroidDesigner)' == ''">false</VSAndroidDesigner>
```
jonathanpeppers, line 77: So will the designer have to change anything on their side to enable this property?

src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
radekdoulik: Ah, I missed that one out, when renaming. Changed it to use `IsSupported` suffix.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
radekdoulik: They would need to set the `$(VSAndroidDesigner)` to `true`. Or alternatively we might set it to false only for release configuration?

jonathanpeppers: The designer works fine for Release builds, so I don't think we can do it there.

For now, maybe you could file a quick issue on what they'll need to do? I'll find the right person take a look later on.

radekdoulik: I looked closer at it and the msbuild property only affects linking. So designer should run fine with `Debug` configuration.

To disable the fallback during runtime, it would need to be explicitely set in code by calling `AppContext.SetSwitch` or in the application configuration file. So designer should be fine.

src/Mono.Android/Android.Runtime/Logger.cs
```diff
@@ -55,7 +55,7 @@ internal static class Logger {
 						hasNoLibLog = true;
 					}
 				}
-				if (hasNoLibLog)
+				if (AndroidEnvironment.VSAndroidDesignerIsEnabled && hasNoLibLog)
```
jonathanpeppers, line 58: Is the `VSAndroidDesignerIsEnabled` check slow at all? Should it be second in this `if`-statement?

jonathanpeppers (approved):
Otherwise LGTM 👍

src/Mono.Android/Android.Runtime/Logger.cs
radekdoulik: The property just returns the backing field, so it is fast. I prefer `if (AndroidEnvironment.VSAndroidDesignerIsEnabled && hasNoLibLog)` as in release it produces shorter body, linker removes the `&& hasNoLibLog` part of the test.

Like this:
```
...
  catch [System.Private.CoreLib]System.DllNotFoundException 
  {
    IL_0042:  pop
    IL_0043:  ldc.i4.1
    IL_0044:  stsfld     bool Android.Runtime.Logger::hasNoLibLog
    IL_0049:  leave.s    IL_004b
  }  // end handler
  IL_004b:  call       bool Android.Runtime.AndroidEnvironment::get_VSAndroidDesignerIsEnabled()
  IL_0050:  brfalse.s  IL_0052
  IL_0052:  ldloc.1
  IL_0053:  ldc.i4.1
  IL_0054:  add
  IL_0055:  stloc.1
  IL_0056:  ldloc.1
  IL_0057:  ldloc.0
  IL_0058:  ldlen
  IL_0059:  conv.i4
  IL_005a:  blt.s      IL_0023
  IL_005c:  ret
} // end of method Logger::Log
```

In case of `if (hasNoLibLog && AndroidEnvironment.VSAndroidDesignerIsEnabled)` I think the linker has to keep the 1st part of test in place, leading to longer body.

jonathanpeppers: Oh, I see it has:
```csharp
internal static bool VSAndroidDesignerIsEnabled { get; } = InitializeVSAndroidDesignerIsEnabled ();
```
And not:
```csharp
internal static bool VSAndroidDesignerIsEnabled => InitializeVSAndroidDesignerIsEnabled ();
```

Issue #4717, opened 2020-05-23
[Bug] When using out array parameter in AIDL, the code generator is not working properly


MrAlbin:
I*m using Android Studio to buld an Android service and VisualStudio 2019 to build an App that uses this Service (have to do it this way).
There seems to be an issue when using `out` arrays like `int[]` or `byte[]`.

### Steps to Reproduce

1. AIDL file with a function like
`int GetByte(out byte[] myByte);`
2. Implement a Java-Service using Android-Studio that implements AIDL interface and write 1 byte to the out parameter `myByte`
3. Implement a Xamarin App that uses the Service and call `GetByte`.

### Expected Behavior
The App shoud receive the byte correctly.

### Actual Behavior
It always results in an Exception:

> java.lang.RuntimeException: bad array lengths
	at android.os.Parcel.readByteArray(Parcel.java:2233)

To me it looks like the code generator has a bug. I compared the generated Java code (Android Studio) and the corresponding C# code (Visual Studio):

Java:
```
@Override public int GetByte(byte[] myByte) throws android.os.RemoteException
{
	android.os.Parcel _data = android.os.Parcel.obtain();
	android.os.Parcel _reply = android.os.Parcel.obtain();
	int _result;
	try {
		_data.writeInterfaceToken(DESCRIPTOR);
		if ((count==null)) {
			_data.writeInt(-1);
		}
		else {
			_data.writeInt(count.length);
		}
		boolean _status = mRemote.transact(Stub.TRANSACTION_GetByte, _data, _reply, 0);
		if (!_status && getDefaultImpl() != null) {
			return getDefaultImpl().GetByte(count);
		}
		_reply.readException();
		_result = _reply.readInt();
		_reply.readByteArray(count);
	}
  	finally {
  	        _reply.recycle();
 	         _data.recycle();
 	}
	return _result;
}
```

C#
```
public int SIS_GetByte (byte [] myByte)
{
	global::Android.OS.Parcel __data = global::Android.OS.Parcel.Obtain ();

	global::Android.OS.Parcel __reply = global::Android.OS.Parcel.Obtain ();
	int __result = default (int);

	try {
		__data.WriteInterfaceToken (descriptor);
		remote.Transact (IScaleInterfaceServiceStub.TransactionSIS_GetByte, __data, __reply, 0);
		__reply.ReadException ();
		__result = __reply.ReadInt ();
		__reply.ReadByteArray (count);

	} finally {
		__reply.Recycle ();
[...]

PR #5726, opened 2021-03-10
[Xamarin..Android.Tools.Aidl] Handle out and inout Parameters in CodeGenerator properly

### LINKED ISSUE #5079: [Bug] When using inout parameter in AIDL, the code generator is not working properly
### Steps to Reproduce

1. Add a function with an `inout` parameter to an AIDL file:
for example:
`int GetStringArray(inout String[] stringArray);`
2. Implement the function in your Service.
3. Compile.

### Expected Behavior
The code generator should produce:
```
...
protected override bool OnTransact (int code, global::Android.OS.Parcel data, global::Android.OS.Parcel reply, int flags)
{
	switch (code) {
	...
	case TransactionGetStringArray: {
		data.EnforceInterface (descriptor);
		String [] arg0 = default (String []);
		arg0 = data.CreateStringArray ();
		var result = this.GetStringArray (arg0);
		reply.WriteNoException ();
		reply.WriteInt (result);
		reply.WriteStringArray (arg0);
		}
	...
	}
...
}
```

### Actual Behavior
The code generator produces:
```
...
protected override bool OnTransact (int code, global::Android.OS.Parcel data, global::Android.OS.Parcel reply, int flags)
{
	switch (code) {
	...
	case TransactionGetStringArray: {
		data.EnforceInterface (descriptor);
		String [] arg0 = default (String []);
		arg0 = data.CreateStringArray ();
		var result = this.GetStringArray (arg0);
		reply.WriteNoException ();
		reply.WriteInt (result);
		data.WriteStringArray (arg0);
		}
	...
	}
...
}
```

What goes wrong:
It should not be `data.WriteStringArray (arg0);` but `reply.WriteStringArray (arg0);`.

### Version Information

Microsoft Visual Studio Professional 2019
Version 16.7.2
VisualStudio.16.Release/16.7.2+30413.136
Microsoft .NET Framework
Version 4.8.03752

Installierte Version: Professional

Visual C++ 2019   00435-60000-00000-AA836
Microsoft Visual C++ 2019

Xamarin   16.7.000.440 (d16-7@358f3c6)
Visual Studio-Erweiterung, um Entwicklung für Xamarin.iOS und Xamarin.Android zu ermöglichen.

Xamarin Designer   16.7.0.495 (remotes/origin/d16-7@79c0c522c)
Visual Studio-Erweiterung zum Aktivieren der Xamarin Designer-Tools in Visual Studio.

Xamarin Templates   16.7.85 (1bcbbdf)
Templates for building iOS, Android, and Windows apps with Xamarin and Xamarin.Forms.

Xamarin.Android SDK   11.0.2.0 (d16-7/025fde9)
    Mono: 83105ba
    Java.Interop: xamarin/java.interop/d16-7@1f3388a
    ProGuard: Guardsquare/proguard/proguard6.2.2@ebe9000
    SQLite: xamarin/sqlite/3.32.1@1a3276b
    Xamarin.Android Tools: xamarin/xamarin-android-tools/d16-7@017078f

Xamarin.iOS and Xamarin.Mac SDK   13.20.2.2 (817b6f72a)
Xamarin.iOS and Xamarin.Mac Reference Assemblies and MSBuild support.

mfranke-moba:
Fixes #5079 and #4717

The CodeGenerator responible for producing C# Code for AIDL files so far had the problem that function parameters of some array type (like int[]) marked with out or inout were not handled properly.
For such parameters the legth of the array has to be written to the data parcelable in the Proxy function. See https://cs.android.com/android/platform/superproject/+/master:system/tools/aidl/generate_java_binder.cpp (lines 646 - 656).
Now just like in the linked sources something like:
```
if {arrarParam == null} {
    _data.writeInt(-1);
}
else {
    _data.writeInt(arrayParam.lenght)
}
```
will be produced.

The counterpart which is the corresponding Stub function will contain:
```
int _argX_length = data.readInt();
if ((_argX_length<0)) {
    _argX = null;
}
else {
    _argX = new byte[_argX_length];
}
```
See https://cs.android.com/android/platform/superproject/+/master:system/tools/aidl/generate_java_binder.cpp (line 486 and lines 358 - 372)
This is comparable to the java code the Android Studio produces.

Now also return values will be written to _result instead of _data which was wrong before.

The improvements of this commit show for interprocess communication between different Apps/Services. Now also Communication betwen Xamarin and Java Apps/Services works for array parameters marked with out/inout.

I really hope this bugfix will find its way into the nex release of Visual Studio.


jonpryor:
@mfranke-moba: how "interested" are you in a "larger project"? :-D

*If* you're interested, `src/Xamarin.Android.Tools.Aidl` *needs* unit tests.  I know of "a" pattern for writing unit tests for Irony grammars; see e.g. https://github.com/xamarin/Java.Interop/commit/7574f166008bb45c0df97315aae7907ac25f8602 and https://github.com/xamarin/java.interop/blob/94c0c70970cd30b9a86ec1607a5394f8c9a1bb8e/tests/Java.Interop.Tools.JavaSource-Tests/SourceJavadocToXmldocGrammar.InlineTagsBnfTermsTests.cs.

The question is whether or now *you* are interested in writing such a unit test framework.  We'll get around to it…some day…

But in the meantime, `src/Xamarin.Android.Tools.Aidl` desperately could use some unit tests, so that we can "feel better" about future fixes and improvements to this assembly.

mfranke-moba:
I already noticed there are almost no unit tests for `src/Xamarin.Android.Tools.Aidl`. This may one point where I could contribute something.
Besides the issue with out/inout arrays I also noticed that Parcelables do not work. Some unit test illustrating this problem might be a good starting point helping to fix this as well.

src/Xamarin.Android.Tools.Aidl/CSharpCodeGenerator.cs
```diff
@@ -277,6 +277,20 @@ protected override bool OnTransact (int code, global::Android.OS.Parcel data, gl
 					w.WriteLine ("\t\t\t\t{0} {1} = default ({0});", ToOutputTypeName (name_cache.ToCSharp (a.Type)), "arg" + i);
 					if (a.Modifier == null || a.Modifier.Contains ("in"))
 						w.WriteLine ("\t\t\t\t{0}", GetCreateStatements (a.Type, "data", "arg" + i));
+					else if (a.Modifier != null && a.Modifier.Contains ("out")) {
+						if (a.Type.ArrayDimension > 0) {
```
jonpryor, line 281: This code block appears to only support `a.Type.ArrayDimension` being `0` (no array)` or `1` (single-dimension array).  Could it also be `2` or more, e.g. for rectangular arrays such as `int[][]`?

It may be "safer" to `switch (a.Type.ArrayDimension)` and e.g. throw if it's an unsupported value.

src/Xamarin.Android.Tools.Aidl/CSharpCodeGenerator.cs
```diff
@@ -613,7 +630,15 @@ string GetWriteStatements (TypeName type, string parcel, string arg, string parc
 					return String.Format ("{1}.WriteStrongBinder (((({0} != null)) ? ({0}.AsBinder ()) : (null)));", arg, parcel);
 			}
 		}
-		
+
+		string GetWriteOutStatements (TypeName type, string parcel, string arg)
+		{
+			if (type.ArrayDimension > 0) {
```
jonpryor, line 636: …ditto here: is `int[][]` possible?

A quick gander at the [AIDL documentation](https://developer.android.com/guide/components/aidl) suggests it *isn't* possible.  However, it *does* suggest that `List`:

> must be one of the supported data types in this list or one of the other AIDL-generated interfaces or parcelables you've declared

which implies that a `List<List<List<String>>>` may be viable…and I have no idea what we'd do with that.  (Wouldn't impact `type.ArrayDimension`, though!)

mfranke-moba: The way I understand the AIDL dokumentation is that multidimensional arrays like `int[][]` are not supported. At the moment the `CSharpCodeGenerator` or maybe the `AidlParser` just ignores higher dimensions and int[][] is treated like int[]. However throwing some NotSupportedException might not be wrong.

For `List` I think - but not 100% sure - you can put any supported object in. And one object can be a List which can contain other Lists and so on. I don't know why (maybe a bug) but the Android Studio does not support parametrization like `List<List<List<String>>>>`, just `List<String>`. So you can use `List` and put Lists in but `List<List>>` does not work!?

The CSharpCodeGenerator might not cover all cases for Lists correct and should definetely be unit tested also in this perspect.

jonpryor:
Just to keep things "in the loop", PR #5756 adds unit tests for this assembly.  Once PR #5756 is merged, we'll update this PR (#5726) to merge with to-be-current main, and update the unit test expected outputs accordingly.

jonpryor:
With the merging of PR #5756, I've merged main into the [mfranke-moba/out_parameter_handling branch](https://github.com/mfranke-moba/xamarin-android/tree/out_parameter_handling) and updated the expected test output from 81d37a12e07ea797208dc05fc917d0be721f30bd so that the unit tests pass.

I have not in any way validated the the "changes required for unit tests to pass" are in any way *correct*…

src/Xamarin.Android.Tools.Aidl/CSharpCodeGenerator.cs
jonpryor: Answer: no, AIDL doesn't (currently) support mult-dimensional arrays.

src/Xamarin.Android.Tools.Aidl/CSharpCodeGenerator.cs
```diff
@@ -613,7 +630,15 @@ string GetWriteStatements (TypeName type, string parcel, string arg, string parc
...
+				return "if (" + arg + " == null) { " + parcel + ".WriteInt(-1); } else { " + parcel + ".WriteInt(" + arg + ".Length); }";
```
jonpryor, line 637: Given that `.WriteInt(-1)` doesn't show up anywhere in our current expected test output, we're going to need a new unit test which ensures this code path is hit.

src/Xamarin.Android.Tools.Aidl/CSharpCodeGenerator.cs
```diff
@@ -289,8 +303,8 @@ protected override bool OnTransact (int code, global::Android.OS.Parcel data, gl
 					w.WriteLine ("\t\t\t\t{0}", GetWriteStatements (method.ReturnType, "reply", "result", "global::Android.OS.ParcelableWriteFlags.ReturnValue"));
 				for (int i = 0; method.Arguments != null && i < method.Arguments.Length; i++) {
 					var a = method.Arguments [i];
-					if (a.Modifier == null || a.Modifier.Contains ("out"))
```
jonpryor, line 292: I assume this change/removal is why `data.WriteLong (arg0);` is removed in the unit test diff.

I *think* this is correct?

src/Xamarin.Android.Tools.Aidl/CSharpCodeGenerator.cs
jonpryor: @mfranke-moba : please add a unit test which causes this line to be triggered.

mfranke-moba: This is already covered by `Test11`. The `WriteInt(-1)` will be written for out array parameters like `out byte[]`. In this case no byte array is written with `WriteXyzArray` like for `in byte[]` or `inout byte[]` as no data has to be transmitted, just the size.

src/Xamarin.Android.Tools.Aidl/CSharpCodeGenerator.cs
mfranke-moba: This change was necessary because before for primitive input parameters like in `void Authenticate (long operationId)` the input parameter after function execution was written back to the `reply` Parcel which is only reqired for out and inout parameters.

src/Xamarin.Android.Tools.Aidl/CSharpCodeGenerator.cs
jonpryor: Ah, I had missed that. :-(

PR #5731, opened 2021-03-11
[build] Use arcade dependency management tooling

pjcollins:
Context: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md

The dotnet core engineering group has some dependency management tooling
(known as `darc`) that we'd like to adopt.  Integrating this tooling in
to the build system will make it easier to stay up to date with the
latest .NET 6 SDK changes.

Many dotnet repos use a [publishing workflow][0] that will push build
artifact data to a central location known as the "Build Asset Registry".
This data includes a "[Channel][1]" association, which is the key to
dependency updating.  Local updates and automatic update "Subscriptions"
compare the version files in a given repo against the product versions
avalable in the channel that you are interested in.

We hope to be able to publish Xamarin SDK build information to this
central registry in the near future, so that other products can take a
dependency on us as needed (MAUI for instance).

The `darc` tool looks for four different files in a repo when adding a
dependency or when checking for an update:

 * eng/Version.Details.xml
 * eng/Versions.props
 * nuget.config
 * global.json

Both of the `Version` files present in the `eng` folder are updated when
a new dependency is available.

To work with `darc` locally you'll need to install the global tool, join
the `arcade-contrib` GitHub team, and configure your [auth settings][2].

To add a new dependency, use the [`darc add-dependency`][3] command:

    darc add-dependency -n Microsoft.NetCore.App.Runtime.android-arm64 -t product -v 6.0.0-preview.2.21154.6 -r https://github.com/dotnet/runtime

To update all dependencies, use the [`darc update-dependencies`][4]
command:

    darc update-dependencies --channel ".NET 6"

To configure automatic updates, use the [`darc add-subscription`][5]
command to enroll a target repo/branch into updates from a particular
channel:

     darc add-subscription --channel ".NET 6" --source-repo https://github.com/dotnet/installer --target-repo https://github.com/xamarin/xamarin-android --target-branch main --update-frequency everyWeek --standard-automerge

Once a subscription is configured, [a pull request][6] will be created
automatically by the dotnet Maestro bot when dependency updates are
available.

---

This PR also contains a bump to .NET 6.0.100-preview.3.21168.18.
Changes: https://github.com/dotnet/installer/compare/19e22a76...823c1dfce

The .NET 6 .apkdesc files have been updated accordingly. It seems that
`System.Private.CoreLib.dll` grew in size, however reductions in native
libraries and other files results in overall smaller package sizes.

Simple XA:

```diff
-"PackageSize": 2889606
+"PackageSize": 2877318
```

XF/XA:

```diff
-"PackageSize": 8746124
+"PackageSize": 8733836
```

[0]: https://github.com/dotnet/arcade/blob/681511f2f63a3563494f1f27904b2842abef6b35/Documentation/CorePackages/Publishing.md#what-is-v3-publishing-how-is-it-different-from-v2
[1]: https://github.com/dotnet/arcade/blob/main/Documentation/BranchesChannelsAndSubscriptions.md#branches-channels-and-subscriptions-explained
[2]: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md#setting-up-your-darc-client
[3]: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md#add-dependency
[4]: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md#update-dependencies
[5]: https://github.com/dotnet/arcade/blob/ea609b8e036359934332480de9336d98fcbb3f91/Documentation/Darc.md#add-subscription
[6]: https://github.com/xamarin/xamarin-android/pull/5744

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,12 +1,12 @@
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.Dotnet.Sdk.Internal" Version="6.0.100-preview.3.21167.10">
+    <Dependency Name="Microsoft.Dotnet.Sdk.Internal" Version="6.0.100-preview.3.21168.18">
       <Uri>https://github.com/dotnet/installer</Uri>
-      <Sha>18a633b84588c419939a55ee9504727eeb2e9714</Sha>
+      <Sha>823c1dfceb916ff19de20c7ddccab0ea145508a3</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.NET.ILLink" Version="6.0.100-preview.2.21167.4">
+    <Dependency Name="Microsoft.NET.ILLink" Version="6.0.100-preview.2.21168.1">
       <Uri>https://github.com/mono/linker</Uri>
-      <Sha>75b1c3003bc12139c3a918705da5a4c12e7cb37b</Sha>
+      <Sha>9eff8c6ce48cb15cafb8eef4aaf853675b05f9d9</Sha>
     </Dependency>
   </ProductDependencies>
 </Dependencies>
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -1,16 +1,12 @@
 <Dependencies>
   <ProductDependencies>
-    <Dependency Name="Microsoft.Dotnet.Sdk.Internal" Version="6.0.100-preview.3.21160.15">
+    <Dependency Name="Microsoft.Dotnet.Sdk.Internal" Version="6.0.100-preview.3.21167.10">
       <Uri>https://github.com/dotnet/installer</Uri>
-      <Sha>a28ceebdcd4601c69f86d4ec1341ad95295f5020</Sha>
+      <Sha>18a633b84588c419939a55ee9504727eeb2e9714</Sha>
     </Dependency>
-    <Dependency Name="Microsoft.NETCore.App.Runtime.android-arm64" Version="6.0.0-preview.3.21160.15">
-      <Uri>https://github.com/dotnet/runtime</Uri>
-      <Sha>b38ea6a4f067b571a007b2ac69f3e4431cb100e7</Sha>
-    </Dependency>
-    <Dependency Name="Microsoft.NET.ILLink" Version="6.0.100-preview.2.21160.1">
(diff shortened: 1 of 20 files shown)


eng/Version.Details.xml
```diff
@@ -0,0 +1,12 @@
+<Dependencies>
```
jonathanpeppers, line 1: I think @jonpryor wasn't thrilled about a new `eng` folder.

Is there any way to put these in `build-tools`? If not, we probably have to put them here.

dotnet/maui already has an `eng` folder, so it will be fine.

jonathanpeppers (approved):
Otherwise, this looks great. I like how we don't have to lookup & hardcode the runtime packs version anymore.

grendello (approved):

eng/Version.Details.xml
pjcollins: I haven't found a way to provide alternate paths for files to `darc`, so unless we find the source and PR into that tool to check multiple paths we'll have to keep `eng`.

jonathanpeppers: There is this `$(RepositoryEngineeringDir)`:

https://github.com/dotnet/arcade/blob/6cc4c1e9e23d5e65e88a8a57216b3d91e9b3d8db/src/Microsoft.DotNet.Arcade.Sdk/tools/SourceBuild/SourceBuildArcadeTools.targets#L20

https://github.com/dotnet/arcade/blob/6cc4c1e9e23d5e65e88a8a57216b3d91e9b3d8db/src/Microsoft.DotNet.Arcade.Sdk/tools/RepoLayout.props#L44

But if we put our own value for `$(RepositoryEngineeringDir)` in `Directory.Build.props`, I don't know if the tooling would know to parse that value and use the correct value for `$(RepositoryEngineeringDir)`.

pjcollins: I think we'd need to PR something here (and perhaps elsewhere) - https://github.com/dotnet/arcade-services/blob/bb98f0d13ac51eb81131b5ef382ada12fd84e8e0/src/Microsoft.DotNet.Darc/src/DarcLib/Helpers/VersionFiles.cs#L16

dellis1972 (approved):

Issue #5707, opened 2021-03-07
Aapt2ToolPath is not being searched for in android-sdk\build-tools


jojorne:
### Steps to Reproduce

1. Just go to "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\Android" and there you will find an outdated version of the "aapt2.exe";
2. Rename it to "aapt2.ex_" and you will see that when you try to compile a project it doesn't look in "android-sdk\build-tools".

### Expected Behavior

Try to search for "aapt2.exe" in the "android-sdk\build-tools" directory **_first_**!

### Actual Behavior
When you try to compile a project it uses an outdated version of "aapt2.exe" found in "C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\Android".

### Version Information
Visual Studio 16.9.0
AAPT2 version:
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community\MSBuild\Xamarin\Android>aapt2 version
Android Asset Packaging Tool (aapt) 2.19-6503028
C:\Program Files (x86)\Android\android-sdk\build-tools\30.0.2>aapt2 version
Android Asset Packaging Tool (aapt) 2.19-6739378

jojorne:
PS: The old version in "MSBuild\Xamarin\Android" is crashing on my pc too. Replacing it with the one in "android-sdk\build-tools" fixs it.

dellis1972:
@jojorne 

There is a reason why we search in our directory by default. We have had issues in the past where versions of `aapt2.exe` were broken in the sdk for the features we are using, or the user had an older version which did not support the features we need. So we decided to ship our own `known` working version. 

If you want to completely override our default behaviour you should be able use the following in the first `PropertyGroup` section of your csproj file

```
<Aapt2ToolPath>$(AndroidSdkPath)build-tools\30.0.2</Aapt2ToolPath>
```

That should override the path. In the mean time we will look at updating the version of `aapt2` we ship. 
It would be interesting to see what crash you are getting from the version we ship.

jojorne:
@dellis1972

When compiling any Android project like MonoGame or Xamarin.Forms, AAPT2 will crash with:
Unhandled exception at 0x012D966D in aapt2.exe: 0xC000001D: Illegal Instruction.
Unhandled exception at 0x0148966D in aapt2.exe: 0xC000001D: Illegal Instruction.
[...]

PR #5736, opened 2021-03-12
Bump to aapt2 4.1.2-6503028

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5707

Will wait to see if this version actually *fixes* the issue, or we might need to use an even newer version of `aapt2`.


jonathanpeppers:
Let's hold off on this and bump `aapt2` closer to 16.10's release.

PR #5740, opened 2021-03-14
One directory higher then the immediate prefix of mono worked for me..

explorest:
I'm on: 
Ubuntu 20.04  
Microsoft (R) Build Engine version 16.6.0 for Mono. 
Mono JIT compiler version 6.12.0.122 (tarball Mon Feb 22 17:33:28 UTC 2021)
All installed with defaults. The Xamarin.Android source also built with defaults.

My mono is at /usr/bin/mono.   Doing "make install prefix=/usr/bin"  and then building the sample HelloWorld with MSBuild was giving the following error:
/home/..../xamarin-android/samples/HelloWorld/HelloWorld.csproj(78,3): error MSB4019: The imported project "/usr/lib/mono/xbuild/Xamarin/Android/Xamarin.Android.CSharp.targets" was not found. Confirm that the expression in the Import declaration "/usr/lib/mono/xbuild/Xamarin/Android/Xamarin.Android.CSharp.targets" is correct, and that the file exists on disk.

Took me a long time to track the source of the error.  Changing make install's prefix from /usr/bin to /usr fixed the error and msbuild successfully built the sample HelloWorld without any errors or warnings.

### DIFF
--- a/Documentation/workflow/UsingYourBuild.md
+++ b/Documentation/workflow/UsingYourBuild.md
@@ -20,7 +20,7 @@ the `$(prefix)` make variable:
 
 In order for `xbuild` or `msbuild` within `$PATH` to resolve the
 Xamarin.Android assemblies, you must install Xamarin.Android into the
-same prefix as mono.
+same prefix as mono; e.g. if mono is at /usr/bin/mono then prefix=/usr.
 
 
 ## Windows System-Wide Installation


jonpryor:
While updating the docs is a good idea, should we also update the `make install` target to "verify" the `$(prefix)` value?

At minimum, we could verify that `$(prefix)` *doesn't* end with `/bin`:

```diff
diff --git a/Makefile b/Makefile
index 9cd9c22c7..eaf90fcc1 100644
--- a/Makefile
+++ b/Makefile
@@ -113,6 +113,10 @@ install::
 		echo "run 'make all' before you execute 'make install'!"; \
 		exit 1; \
 	fi
+	@if echo "$(prefix)" | grep '/bin$$' 2>&1 >/dev/null ; then \
+		echo "Bad 'prefix' value.  Did you mean 'prefix=$(shell echo $(prefix) | sed 's,/bin$$,,g' )'?" ; \
+		exit 1; \
+	fi
 	-mkdir -p "$(prefix)/bin"
 	-mkdir -p "$(prefix)/lib/mono/xbuild-frameworks"
 	-mkdir -p "$(prefix)/lib/xamarin.android"
```

Which would result in:

```bash
% make install prefix=/usr/local/bin
Bad 'prefix' value.  Did you mean 'prefix=/usr/local'?
make: *** [install] Error 1
```

grendello:
I wouldn't make the prefix check failure an error, rather a warning - somebody may have a legitimate prefix ending with `/bin` (e.g. `/opt/bin`). A check that might be useful (but also a warning IMO) is a comparison between `$(prefix)` and Mono install location.

grendello (approved):

explorest:
The error message I gave in the original post above is not specific to prefixes ending in "bin".  The same EXACT error message happens when I "`make install`"  with  the `prefix=/usr/local `or any prefix other than `/usr`.  Only using` /usr` fixes it.

I did some investigating. I think I have got to the bottom of the issue. The source of the problem is not the prefix itself.   It is coming from using "`$ msbuild`" directly to build the sample ,  rather than using the provided wrapper script `bin/Debug/bin/xabuild`.  

When using msbuild directly --- the predefined MSBuild property `"MSBuildExtensionsPath"` used in  line 78 in `/samples/HelloWorld/HelloWorld.csproj`  expands to a predetermined path ( /usr/lib/mono/xbuild in my case )  ---  **regardless of the prefix used in "`make install`".**  ---  resulting in failure to locate things and hence the error message in the original post.  It is only when we use the prefix `/usr`  in the `make install `  making the install target create directories under `/usr` that the predetermined value of  the `MSBuildExtensionsPath` property in the .csproj happens to be the correct value !

Using the wrapper script `bin/Debug/bin/xabuild` on command line does not give these errors  -- regardless of the prefix value. Looking inside the script shows that it fixes `MSBuildExtensionsPath` among other things!  phewwwww :)

PR #5745, opened 2021-03-15
Avoids to remove the launch profiles capability when the _KeepLaunchProfiles property is true

tondat:
- this property will be defined on single projects to use launch profiles capability required on VisualStudio
https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1291512

When a single-project is loaded on VisualStudio, to enable the switch between target frameworks the LaunchProfiles capability need to be enabled. 

By default, it is enabled but this capability is removed by our SDKs because the HEAD projects (Android and iOS) does not use this IDE feature to display devices on the start button.

The _KeepLaunchProfiles property avoids that removal in order to have LaunchProfiles enabled on VisualStudio.

MAUI targets adds _KeepLaunchProfiles with the value true to enable it for single-project.


jonathanpeppers (commented):
Can you explain in the PR text or commit message what it's for, same on this one: https://github.com/xamarin/xamarin-macios/pull/10868

Needs to briefly explain what the issue is and what will need to happen in the Maui repo.

tondat:
> Can you explain in the PR text or commit message what it's for, same on this one: [xamarin/xamarin-macios#10868](https://github.com/xamarin/xamarin-macios/pull/10868)
> 
> Needs to briefly explain what the issue is and what will need to happen in the Maui repo.

done!

dellis1972 (approved):

jonathanpeppers (approved):

PR #5747, opened 2021-03-15
[One .NET] rename to Microsoft.Android.Sdk.$(HostOS)

jonathanpeppers:
When reviewing our `WorkloadManifest.json`, we realized that 32-bit
Windows was not listed.

However, 32-bit Windows should already work because:

* We ship 32 & 64-bit `libzip.dll`
* `aapt2.exe` is 32-bit

`Microsoft.Android.Sdk` needs to support both 32 & 64-bit, because you
can run `MSBuild.exe` in either:

    "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\MSBuild.exe"
    "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Current\Bin\amd64\MSBuild.exe"

By default MSBuild is 32-bit on Windows, as well as inside Visual Studio.

After reviewing the names of packages, using RIDs does not really make
sense. The `osx-x64` and `win-x64` packages are actually Mac & Windows
packages, regardless of the architecture.

Let's rename the packages to match the existing `$(HostOS)` variable
we have:

* Microsoft.Android.Sdk.Windows
* Microsoft.Android.Sdk.Darwin
* Microsoft.Android.Sdk.Linux

This cleans up things a bit, so we can drop the `$(AndroidHostRID)`
MSBuild property and just use `$(HostOS)` instead.

### DIFF
--- a/Documentation/building/unix/instructions.md
+++ b/Documentation/building/unix/instructions.md
@@ -90,8 +90,8 @@ packages with:
 Several `.nupkg` files will be output in `./bin/BuildDebug/nuget-unsigned`,
 but this is only part of the story. Your local
 `~/android-toolchain/dotnet/packs` directory will be populated with a
-local Android "workload" in `Microsoft.Android.Sdk.osx-x64` or
-`Microsoft.Android.Sdk.linux-x64` matching your operating system.
+local Android "workload" in `Microsoft.Android.Sdk.$(HostOS)` matching
+your operating system.
 
 To use the Android workload, you will need a `NuGet.config`:
 
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
@@ -16,10 +16,11 @@
       "kind": "sdk",
       "version": "@SDK_PACK_VERSION@",
       "alias-to": {
-        "osx-x64": "Microsoft.Android.Sdk.osx-x64",
-        "osx-arm64": "Microsoft.Android.Sdk.osx-x64",
-        "win-x64": "Microsoft.Android.Sdk.win-x64",
-        "linux-x64": "Microsoft.Android.Sdk.linux-x64"
+        "osx-x64": "Microsoft.Android.Sdk.Darwin",
+        "osx-arm64": "Microsoft.Android.Sdk.Darwin",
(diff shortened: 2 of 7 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
```diff
@@ -18,6 +18,7 @@
       "alias-to": {
         "osx-x64": "Microsoft.Android.Sdk.osx-x64",
         "osx-arm64": "Microsoft.Android.Sdk.osx-x64",
+        "win-x86": "Microsoft.Android.Sdk.win-x64",
         "win-x64": "Microsoft.Android.Sdk.win-x64",
         "linux-x64": "Microsoft.Android.Sdk.linux-x64"
```
jonathanpeppers, line 23: Should we instead rename these to:
* Microsoft.Android.Sdk.Windows
* Microsoft.Android.Sdk.MacOS
* Microsoft.Android.Sdk.Linux

They aren't going to map to RIDs, because we probably want the Mac & Windows packages to work on multiple architectures.

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Workload.Android/WorkloadManifest.in.json
pjcollins: I think renaming these as suggested makes sense 👍

jonathanpeppers: I used the name `Microsoft.Android.Sdk.Darwin`, so it matches `$(HostOS)`.

We'll see what breaks.

dellis1972 (approved):

Documentation/building/unix/instructions.md
```diff
@@ -91,7 +91,7 @@ Several `.nupkg` files will be output in `./bin/BuildDebug/nuget-unsigned`,
 but this is only part of the story. Your local
 `~/android-toolchain/dotnet/packs` directory will be populated with a
 local Android "workload" in `Microsoft.Android.Sdk.osx-x64` or
```
pjcollins, line 93: We can remove this (now duplicated) `Microsoft.Android.Sdk.osx-x64` mention

pjcollins (approved):

PR #5748, opened 2021-03-16
Update to new linker custom steps API

sbomer:
This updates the custom steps to use the new API which runs logic on mark (https://github.com/mono/linker/pull/1774). As part of this, I've removed the workaround that loads all referenced assemblies up-front and simplified some of the linker msbuild targets. Some of the MonoDroid.Tuner steps were duplicated and changed to implement the new IMarkHandler interface, to avoid touching the MonoDroid.Tuner code.

@radekdoulik PTAL - in particular I would appreciate your opinion on whether the below assumptions are reasonable. Please also add anyone else who you think should review the change.

Notes from https://github.com/mono/linker/pull/1774#issuecomment-783563526:
In my analysis of the custom steps, most of them "just work" if we call them only for marked members, because they ultimately either:

- Just call AddPreserved* to conditionally keep members of types (which with the new API will just happen when the type is marked)
- In the case of FixAbstractMethodsStep, inject missing interface implementations which will only be kept if they are marked for some other reason.

Most of the steps have been updated in a straightforward way based on the above.

The exceptions are:

- ApplyPreserveAttribute: this step globally marks members with PreserveAttribute.
  - The step is only active for non-SDK link assemblies. Usually we root non-SDK assemblies, but it will be a problem if linking all assemblies.
  - For now, I updated it to use the new custom step API on assembly mark - so it will scan for the attribute in all marked assemblies - but we should investigate whether this is good enough.
- PreserveExportedTypes: similar to the above, this globally marks members with Export*Attribute. It's used for java interop in cases where the java methods aren't referenced statically - like from xml, or for special serialization methods.
  - It's only active for non-SDK assemblies, so like above it will be a problem only if linking all assemblies.
  - Like above, I've made it scan marked assemblies.
- PreserveApplications: globally scans for ApplicationAttribute on types/assemblies, and sets the TypePreserve for any types referenced by the attribute.
  - This step technically does a global scan, but it's likely to work on marktype/markassembly - since presumably, ApplicationAttribute is only used on types/assembies that are already kept.
  - I've updated it to use the new customstep API based on this assumption.

edit: per conversation with @jonathanpeppers the plan is to ~~move all of the steps (including the exceptions I mentioned above) to use the new custom steps API, and get rid of the workaround that loads references up-front, even if there is a small chance that they will miss PreserveAttribute and similar in assemblies which are unused (never marked by the linker).~~ (done in this PR)


sbomer:
@jonathanpeppers could you help me trigger a ci run for this change?

jonathanpeppers:
Looks like `dotnet-install.sh` failed: `curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to dot.net:443`

Will just try again.

src/Microsoft.Android.Sdk.ILLink/FixAbstractMethodsHandler.cs
```diff
@@ -0,0 +1,293 @@
...
+using Java.Interop.Tools.Cecil;
+
+using Mono.Linker;
+using Mono.Linker.Steps;
+
+using Microsoft.Android.Sdk.ILLink;
+
+namespace Microsoft.Android.Sdk.ILLink
+{
+	public class FixAbstractMethodsHandler : IMarkHandler
```
jonathanpeppers, line 16: How different are `FixAbstractMethodsHandler` and `FixAbstractMethodsStep`? Would it be possible to keep a single `.cs` file and implement `IMarkHandler`? If so, could we continue sharing code with all the steps?

One reason this might be better is we run `FixAbstractMethodsStep` during `Debug` builds (without the linker):

https://github.com/xamarin/xamarin-android/blob/952e67be871b517208aa62ad6ae6679043669d8f/src/Xamarin.Android.Build.Tasks/Tasks/LinkAssembliesNoShrink.cs#L61

It could get confusing if a .NET 6 Debug build uses a different class for this than a Release build.

We were able to make `<LinkAssembliesNoShrink/>` fully incremental, so incremental builds it only runs on assemblies that change. We invoke some linker steps directly to make this work.

sbomer: Not that different. I changed it to use ifdefs, and most of the code is shared, but there are a few places where it impacts readability. The step can't implement both IMarkHandler and IStep, otherwise the linker doesn't know what kind of custom step this is.

sbomer:
Getting failures like:
```
/home/AzDevOps/.dotnet/sdk/5.0.103/NuGet.targets(131,5): error : Unable to load the service index for source https://devdiv.pkgs.visualstudio.com/_packaging/xamarin-xvs/nuget/v3/index.json. [/agent/_work/1/s/xamarin-android/external/monodroid/external/android-sdk-installer/Xamarin.Installer.Build.Tasks/Xamarin.Installer.Build.Tasks.csproj] [/agent/_work/1/s/xamarin-android/external/monodroid/monodroid.proj]
/home/AzDevOps/.dotnet/sdk/5.0.103/NuGet.targets(131,5): error : Response status code does not indicate success: 403 (Forbidden - User '6d3b3c1a-123d-454c-a78b-b4a426164711' lacks permission to complete this action. You need to have 'ReadPackages'. (DevOps Activity ID: 892BABAE-AEC5-40AC-8C60-E5509582972E)). [/agent/_work/1/s/xamarin-android/external/monodroid/external/android-sdk-installer/Xamarin.Installer.Build.Tasks/Xamarin.Installer.Build.Tasks.csproj] [/agent/_work/1/s/xamarin-android/external/monodroid/monodroid.proj]
/home/AzDevOps/.dotnet/sdk/5.0.103/NuGet.targets(131,5): error : Unable to load the service index for source https://devdiv.pkgs.visualstudio.com/_packaging/xamarin-xvs/nuget/v3/index.json. [/agent/_work/1/s/xamarin-android/external/monodroid/external/android-sdk-installer/Xamarin.Installer.Build.Tasks/Xamarin.Installer.Build.Tasks.csproj] [/agent/_work/1/s/xamarin-android/external/monodroid/monodroid.proj]
/home/AzDevOps/.dotnet/sdk/5.0.103/NuGet.targets(131,5): error : Response status code does not indicate success: 403 (Forbidden - User '6d3b3c1a-123d-454c-a78b-b4a426164711' lacks permission to complete this action. You need to have 'ReadPackages'. (DevOps Activity ID: 892BABAE-AEC5-40AC-8C60-E5509582972E)). [/agent/_work/1/s/xamarin-android/external/monodroid/external/android-sdk-installer/Xamarin.Installer.Build.Tasks/Xamarin.Installer.Build.Tasks.csproj] [/agent/_work/1/s/xamarin-android/external/monodroid/monodroid.proj]
/agent/_work/1/s/xamarin-android/external/monodroid/monodroid.proj(66,5): error MSB3073: The command "dotnet build -p:Configuration=Release -p:DebuggingToolsOutputDirectory=/agent/_work/1/s/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android -v:n -p:OutputPath=/agent/_work/1/s/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android -p:PublicIniParser=true -bl:"/agent/_work/1/s/xamarin-android/external/monodroid/bin/BuildRelease/sdkinstaller.binlog" "/agent/_work/1/s/xamarin-android/external/monodroid/external/android-sdk-installer\Xamarin.Installer.Build.Tasks\Xamarin.Installer.Build.Tasks.csproj"" exited with code 1.
    0 Warning(s)
    5 Error(s)
```

@jonathanpeppers is there a way I could get permissions to trigger the pipelines myself?

jonathanpeppers (approved):
I manually tested the changes here with the projects in [dotnet/net6-mobile-samples](https://github.com/dotnet/net6-mobile-samples).

@sbomer I think the couple CI failures here we can ignore. The test phases all look good.

@dellis1972 or @jonpryor care to review?

sbomer:
Thanks @jonathanpeppers!

src/Microsoft.Android.Sdk.ILLink/SetupStep.cs
```diff
@@ -26,36 +27,35 @@ class SetupStep : BaseStep
...
-		static MethodInfo getReferencedAssembliesMethod = typeof (LinkContext).GetMethod ("GetReferencedAssemblies", BindingFlags.Public | BindingFlags.Instance);
+		List<IMarkHandler> MarkHandlers {
+			get {
+				if (_markHandlers == null) {
+					var pipeline = typeof (LinkContext).GetProperty ("Pipeline").GetGetMethod ().Invoke (Context, null);
+					_markHandlers = (List<IMarkHandler>) pipeline.GetType ().GetProperty ("MarkHandlers").GetValue (pipeline);
+				}
+				return _markHandlers;
+			}
+		}
```
marek-safar, line 38: @sbomer I think this should be avoided and use a supported registration mechanism

src/Microsoft.Android.Sdk.ILLink/SubStepDispatcher.cs
```diff
@@ -1,10 +1,14 @@
 using System;
 using System.Collections.Generic;
 using System.Text;
+using Mono.Linker.Steps;
 
 namespace Microsoft.Android.Sdk.ILLink
 {
-	public class SubStepDispatcher : Mono.Linker.Steps.SubStepsDispatcher
+	public class SubStepDispatcher : MarkSubStepsDispatcher
```
marek-safar, line 8: Why is this needed?

sbomer: Because the dispatcher only accepts a list of steps in its protected constructor.

src/Microsoft.Android.Sdk.ILLink/SetupStep.cs
sbomer: I will look into updating it to use multiple custom step arguments instead of SetupStep

sbomer: This is possible, but messy because separate custom steps don't have a great way to share state. Either:
1. Steps with shared state need to be combined into one (which imposes ordering constraints and isn't great for code organization), or
2. Shared state needs to be static, which hits this bug: https://github.com/mono/linker/issues/1314.

It might be better to leave as-is, and make some of the Pipeline methods part of the public API for custom steps.
If we do want to do 1 or 2, I'd rather do it in a separate PR.

marek-safar: I'd like to see the reflection code removed as it's very fragile.

marek-safar: The change which adds public API for resolve cache landed and the API is available on LinkerContext.

@sbomer could you update the PR to use it instead of the one from xamarin-android?

sbomer: Done, PTAL. The LinkContextExtensions logic is copied and adjusted from https://github.com/xamarin/java.interop/tree/main/src/Java.Interop.Tools.Cecil/Java.Interop.Tools.Cecil. Another option if we want to avoid duplication would be to introduce a dependency on illink in java.interop - please let me know what would be best.

jonathanpeppers: It looks like this project links in a few files from `Java.Interop.Tools.Cecil`, can you add a new file here:

https://github.com/xamarin/xamarin-android/blob/af7f7f5da475c5cb8d2d725d77082ba7915bbc7e/src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj#L43-L46

We probably should make this a regular project reference, but you don't have to do that here -- we can do it later.

sbomer:
Just FYI the latest [changes](https://github.com/xamarin/xamarin-android/pull/5748/commits/dac30f3731514092b983c7b113b026e5be7ad02d) depend on https://github.com/xamarin/xamarin-android/pull/5870 so I expect to see test failures.

src/Microsoft.Android.Sdk.ILLink/SetupStep.cs
sbomer: I'm not sure I understand - are you saying you'd prefer the LinkContextExtensions to live in the Java.Interop repo? The way I have it now, they live in the Microsoft.Android.Sdk.ILLink project so they are already included as compile items.

jonathanpeppers: I guess I'm not following why we need two versions of the extension methods.

Could we create a `TypeDefinitionCache` in .NET 6, or did that cause a problem? We could create one at the step level, if we don't want to share it between steps.

sbomer: Ah, sorry. The shared `TypeDefinitionCache` doesn't work well with .NET 6, and the new extension methods use a cache that's built-in to the linker instead.

More detail:
The `TypeDefinitionCache` shared across steps worked fine before this change, but only because the steps were injected into the pipeline using reflection (with one `--custom-step` to bootstrap and set up the others). The "supported" API only allows adding steps using the `--custom-step` argument, which doesn't play nicely with shared state (https://github.com/mono/linker/issues/1314).

A step-level cache would still work, but could reduce the advantages of caching. @marek-safar instead added a shared cache to the linker itself in https://github.com/mono/linker/pull/1979, and suggested we use that one here, so I added those extension methods to use the built-in shared cache. This doesn't solve the shared state problem in general, but this was the only shared state being used in xamarin-android.

Other alternatives would be:
- Leave it as-is (continue using a shared `TypeDefinitionCache` instead of the built-in one, using reflection to inject the steps)
- Introduce a supported way to share state between steps (see https://github.com/mono/linker/issues/1953) and use a shared `TypeDefinitionCache`
- Avoid using reflection but forego the shared cache - only have one `TypeDefinitionCache` per step.

@marek-safar wanted the reflection to go away, and wasn't sure that the `TypeDefinitionCache` was a compelling reason to extend the custom step API since it would be useful as a shared cache. Now that the shared cache is available I don't see much of an advantage to having a cache per step. Let me know what you think!

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
```diff
@@ -0,0 +1,161 @@
...
+			while ((t = context.GetBaseType (t)) != null) {
+				yield return t;
+			}
+		}
+
+		public static bool IsAssignableFrom (this LinkContext context, TypeReference type, TypeReference c)
+		{
+			if (type.FullName == c.FullName)
+				return true;
+			var d = context.ResolveTypeDefinition (c);
```
marek-safar, line 136: This should use `TryResolveTypeDefinition`

src/Microsoft.Android.Sdk.ILLink/PreserveSubStepDispatcher.cs
```diff
@@ -0,0 +1,19 @@
...
+using System.Text;
+using Mono.Linker.Steps;
+using Mono.Tuner;
+
+namespace Microsoft.Android.Sdk.ILLink
+{
+	public class PreserveSubStepDispatcher : MarkSubStepsDispatcher
+	{
+		public PreserveSubStepDispatcher ()
+			: base (new List<ISubStep> () {
```
marek-safar, line 12: nit: use `new[]`

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -297,6 +339,19 @@ void AddNewExceptionMethod (TypeDefinition type, MethodDefinition method)
...
 		}
 
+		bool markedAbstractMethodErrorType;
+
+		void MarkAbstractMethodErrorType () {
+			if (markedAbstractMethodErrorType)
+				return;
+			markedAbstractMethodErrorType = true;
+				
+			var td = AbstractMethodErrorConstructor.DeclaringType.Resolve ();
```
marek-safar, line 349: should use context.ResolveTypeDefinition ()

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -297,6 +339,19 @@ void AddNewExceptionMethod (TypeDefinition type, MethodDefinition method)
...
+			Annotations.Mark (td);
+			Annotations.SetPreserve (td, TypePreserve.Nothing);
+			Annotations.AddPreservedMethod (td, AbstractMethodErrorConstructor.Resolve ());
```
marek-safar, line 352: Should use context.ResolveMethodDefinition () and check for null return value

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -19,43 +19,80 @@ namespace MonoDroid.Tuner
...
 #endif
 
+			return assembly.MainModule.HasTypeReference ("Java.Lang.Object");
+		}
+
+		void UpdateAssemblyAction (AssemblyDefinition assembly)
+		{
+			AssemblyAction action = Annotations.HasAction (assembly) ? Annotations.GetAction (assembly) : AssemblyAction.Skip;
+			if (action == AssemblyAction.Skip || action == AssemblyAction.Copy || action == AssemblyAction.Delete)
+				Annotations.SetAction (assembly, AssemblyAction.Save);
```
marek-safar, line 64: Do we really want to do this? I don't think we should change copy action to save not alone skip action to save.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -19,43 +19,80 @@ namespace MonoDroid.Tuner
 	/// <summary>
 	/// NOTE: this step is subclassed so it can be called directly from Xamarin.Android.Build.Tasks
 	/// </summary>
-	public class FixAbstractMethodsStep : BaseStep
+	public class FixAbstractMethodsStep
+#if NET5_LINKER
```
marek-safar, line 23: I'm wondering if it'd be better to split this to netcore version and keep the old one untouched

src/Microsoft.Android.Sdk.ILLink/SetupStep.cs
marek-safar: I don't think custom steps should have their own type/method/fields resolving caching. It's redundant and it also usually does not work well with the linker settings which control the resolving process

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
sbomer: Seems the Resolve calls were unnecessary to begin with, so I just changed the type of AbstractwMethodErrorConstructor to MethodDefinition (same above).

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
sbomer: I don't have full context on how we can get into this situation - it looks to me like it was intended to inject methods into non-SDK assemblies which may be copy. I could get rid of skip/delete, but I'm trying to minimize functional changes here.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
sbomer: I did this originally but changed it in response to https://github.com/xamarin/xamarin-android/pull/5748#discussion_r599262551.

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
```diff
@@ -0,0 +1,161 @@
...
+using Mono.Collections.Generic;
+using Mono.Linker;
+using System.Collections.Generic;
+
+namespace Microsoft.Android.Sdk.ILLink
+{
+    public static class LinkContextExtensions
+    {
+        public static MethodDefinition GetBaseDefinition (this LinkContext context, MethodDefinition method)
+		{
```
jonathanpeppers, line 13: There is some weird indentation here. You might run `dotnet format` on this file, so it will follow our `.editorconfig`.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
marek-safar: It'd be fine to fix this in follow-up PR but I think this needs to be addressed because we changed the meaning of "copy" action in net6. It also does not make much sense to be to update code in deleted assembly.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
```diff
@@ -13,6 +13,12 @@ This file contains the .NET 5-specific targets to customize ILLink
       Condition=" '$(PublishTrimmed)' == 'true' "
       AfterTargets="ComputeResolvedFilesToPublishList"
       DependsOnTargets="GetReferenceAssemblyPaths;_CreatePropertiesCache">
+    <PropertyGroup>
+      <_TrimmerDumpDependencies Condition=" '$(LinkerDumpDependencies)' == 'true' ">true</_TrimmerDumpDependencies>
+      <!-- make the output verbose to see what the linker is doing. FIXME: make dependent upon verbosity level -->
+      <_ExtraTrimmerArgs>$(_ExtraTrimmerArgs) --verbose</_ExtraTrimmerArgs>
```
marek-safar, line 19: The PR should not land with --verbose

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
sbomer: I got rid of the skip/delete cases because those don't make any sense to me. If anyone can provide more context on the copy case I'd be happy to address it in a follow-up.

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
sbomer: Oops, fixed. Thanks

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
sbomer: This is unrelated to these changes. As I mentioned in https://github.com/mono/linker/issues/1030#issuecomment-606292840, illink can control the importance of its messages, but it doesn't have a way to react to MSBuild's verbosity, so that FIXME is a bit misleading - I clarified the comment. The way to get rid of this would be to make it the default, or to add an extra task argument, say `TrimmerVerbosity`.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -19,43 +19,80 @@ namespace MonoDroid.Tuner
...
 #if !NET5_LINKER
 			CheckAppDomainUsageUnconditional (assembly, (string msg) => Context.LogMessage (MessageImportance.High, msg));
 #endif
 
+			return assembly.MainModule.HasTypeReference ("Java.Lang.Object");
+		}
+
+		void UpdateAssemblyAction (AssemblyDefinition assembly)
+		{
+			AssemblyAction action = Annotations.HasAction (assembly) ? Annotations.GetAction (assembly) : AssemblyAction.Skip;
```
marek-safar, line 62: Is Annotations.HasAction needed when you it's already doing same check at https://github.com/xamarin/xamarin-android/pull/5748/files#diff-cad3f2ceb954b4631b88ea1bf3228a3c04085149724a0cf544d6c6d987a46d2bR47 ?

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
marek-safar: I'm probably missing something but setting to verbose will start printing hundreds of linker diagnostic details. Why should we do that for Android by default?

sbomer: I'm not sure why it was added exactly, maybe someone thought those diagnostics were useful. To clarify, with --verbose those diagnostics are emitted to the linker's stdout with Low importance, but this is captured by MSBuild. MSBuild's default console logger won't print Low importance messages unless you pass -v:detailed or higher. So it's up to us whether we want them to be (optionally) there. If you prefer I can also remove --verbose here, then the diagnostics won't be printed for any MSBuild verbosity.

marek-safar: I'd much prefer to handle this at runtime SDK side and not by every linker consumer.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
sbomer: Fixed, thanks!

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
sbomer: I removed the argument and filed https://github.com/mono/linker/issues/2014 for it

sbomer:
@jonathanpeppers would you mind re-triggering ci? I am hitting issues with apkdiff after rebasing and want to check if it shows up in ci as well.

sbomer:
@jonathanpeppers could you trigger it again please?

sbomer:
Removing the designer was failing because some of the steps actually still share state, so I had to revert some of the changes I made there. Now only two steps are injected via reflection, and they can be removed when the linker fix https://github.com/mono/linker/pull/2019 flows.

@jonathanpeppers could you trigger it again please?

sbomer:
@jonathanpeppers a typo slipped in 😐. Could you trigger again please?

sbomer:
@jonathanpeppers I think this is good to go (finally). The remaining failures seem unrelated to my changes. The lint ones look like the ones you're fixing in https://github.com/xamarin/xamarin-android/pull/5920, and `RunWithInterpreterEnabled` is failing the same way you were seeing in https://github.com/xamarin/xamarin-android/pull/5920#issuecomment-838709958. When I run these tests locally they pass.

Screen Shot 2021-05-11 at 15 16 16

jonathanpeppers (commented):
@dellis1972 or @jonpryor would you be able to give this a review?

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ILLink.targets
```diff
@@ -13,49 +13,49 @@ This file contains the .NET 5-specific targets to customize ILLink
       Condition=" '$(PublishTrimmed)' == 'true' "
       AfterTargets="ComputeResolvedFilesToPublishList"
       DependsOnTargets="GetReferenceAssemblyPaths;_CreatePropertiesCache">
+    <PropertyGroup>
+      <_TrimmerDumpDependencies Condition=" '$(LinkerDumpDependencies)' == 'true' ">true</_TrimmerDumpDependencies>
+      <_CustomStepDll>$(MSBuildThisFileDirectory)..\tools\Microsoft.Android.Sdk.ILLink.dll</_CustomStepDll>
```
dellis1972, line 18: This is a tiny thing and I know its a "non public" property, but could this have a more "Android" name? `_CustomStepDll` seems pretty generic and it could possibly clash with user declared properties... maybe. 

Perhaps `_AndoridLinkerCustomStepAssembly` might be better?

sbomer: Thanks, fixed!

dellis1972 (approved):

sbomer:
@jonathanpeppers if you're satisfied with the change would you mind merging for me? (I don't have permissions)

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
```diff
@@ -0,0 +1,161 @@
...
+			if (method.IsStatic || method.IsNewSlot || !method.IsVirtual)
+				return method;
+
+			foreach (var baseType in context.GetBaseTypes (method.DeclaringType)) {
+				foreach (var m in baseType.Methods) {
+					if (!m.IsConstructor &&
+							m.Name == method.Name &&
+							(m.IsVirtual || m.IsAbstract) &&
+							context.AreParametersCompatibleWith (m.Parameters, method.Parameters)) {
```
jonpryor, line 22: I do not believe that this check is sufficient; see also commit f96fcf93e157472072576bcc0a8698302899e8cf.

In particular, for interface members, the `.override` IL directive is used to help mark method overrides, and IL permits *contravariance*:

```cil
// Metadata version: v4.0.30319
.assembly extern mscorlib
{
  .ver 4:0:0:0
}
.assembly override
{
}


.class interface private abstract auto ansi IExample
{
  .method public hidebysig newslot abstract virtual 
          instance void  IM(string s) cil managed
  {
  } // end of method IExample::IM

} // end of class IExample

.class private auto ansi beforefieldinit Base
       extends [mscorlib]System.Object
       implements IExample
{
  .method private hidebysig newslot virtual final 
          instance void  IExample.IM(string s) cil managed
  {
    .override IExample::IM
    ldstr      "Base::IM"
    call       void [mscorlib]System.Console::WriteLine(string)
    ret
  } // end of method Base::IExample.IM

  .method public hidebysig specialname rtspecialname 
          instance void  .ctor() cil managed
  {
    ldarg.0
    call       instance void [mscorlib]System.Object::.ctor()
    ret
  } // end of method Base::.ctor

} // end of class Base

.class private auto ansi beforefieldinit Derived
       extends Base
       implements IExample
{
  .method private hidebysig newslot virtual final 
          instance void  IExample.IM(object s) cil managed
  {
    .override method instance void IExample::IM(string s)
    ldstr      "Derived::IM"
    call       void [mscorlib]System.Console::WriteLine(string)
    ret
  } // end of method Derived::IExample.IM

  .method public hidebysig specialname rtspecialname 
          instance void  .ctor() cil managed
  {
    ldarg.0
    call       instance void Base::.ctor()
    ret
  } // end of method Derived::.ctor

} // end of class Derived

.class private auto ansi beforefieldinit App
       extends [mscorlib]System.Object
{
  .method public hidebysig static void  Main() cil managed
  {
    .entrypoint
    .maxstack  2
    .locals init (class Base V_0)
    nop
    newobj     instance void Derived::.ctor()
    stloc.0
    ldloc.0
    ldnull
    callvirt   instance void IExample::IM(string)
    nop
    ret
  } // end of method App::Main

} // end of class App
```

This is equivalent to:

```csharp
interface IExample {
    void IM(string s);
}
class Base : IExample {
    void IExample::IM(string s) { Console.WriteLine("Base::IM"); }
}
class Derived : Base, IExample {
    // not valid C#; parameter contravariance
    void IExample::IM(object s) { Console.WriteLine("Derived::IM"); }
}
class App {
    public static void Main() {
        Base b = new Derived();
        ((IExample)b).IM(null);
    }
}
```

This is valid IL, but `context.AreParametersCompatibleWith()` won't view this as a method override.

*Instead*, you should use the `MethodDefinition.Overrides` collection: https://github.com/xamarin/xamarin-android/blob/92efdccb686454be32d742b8ca9f6a166a7c1808/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs#L160-L170

Unfortunately, `MethodDefinition.Overrides` only helps with interface methods.

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
```diff
@@ -0,0 +1,161 @@
...
+using Mono.Cecil;
+using Mono.Collections.Generic;
+using Mono.Linker;
+using System.Collections.Generic;
+
+namespace Microsoft.Android.Sdk.ILLink
+{
+	public static class LinkContextExtensions
+	{
+		public static MethodDefinition GetBaseDefinition (this LinkContext context, MethodDefinition method)
```
jonpryor, line 12: I find it somewhat "disturbing"/"odd"/"unfortunate" that this would be adding a *different* `GetBaseDefinition()` method in the context of the linker, with a *different* algorithm for comparing method return types (or lack thereof!) and parameter types!  See also:

  * https://github.com/xamarin/java.interop/blob/412e974b9a65c67eab9c5fb5ed330a09abe66710/src/Java.Interop.Tools.Cecil/Java.Interop.Tools.Cecil/MethodDefinitionRocks.cs#L16-L32
  * https://github.com/xamarin/xamarin-android/blob/92efdccb686454be32d742b8ca9f6a166a7c1808/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs#L109-L158

The method here looks very similar to the one in Java.Interop -- which I now see is *incomplete*, because it doesn't check `MethodDefinition.HasOverrides`.

Which leads one to ask, *how many **different** (and potentially incomplete/wrong) implementations do we have to find the "base" method of an override?!*

I sense a Mono.Cecil feature request…

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
```diff
@@ -41,9 +41,6 @@
 
     <!--Other Xamarin.Android / Java.Interop files-->
     <Compile Include="..\..\external\Java.Interop\src\Java.Interop.Tools.Cecil\Java.Interop.Tools.Cecil\CustomAttributeProviderRocks.cs" Link="Java.Interop\CustomAttributeProviderRocks.cs" />
-    <Compile Include="..\..\external\Java.Interop\src\Java.Interop.Tools.Cecil\Java.Interop.Tools.Cecil\MethodDefinitionRocks.cs" Link="Java.Interop\MethodDefinitionRocks.cs" />
-    <Compile Include="..\..\external\Java.Interop\src\Java.Interop.Tools.Cecil\Java.Interop.Tools.Cecil\TypeDefinitionCache.cs" Link="Java.Interop\TypeDefinitionCache.cs" />
```
jonpryor, line 45: Why do we want to remove `TypeDefinitionCache`?  It's inclusion had a nice performance win; see also xamarin/java.interop@b81cfbb9ab8efa647a3bfcc2b2b97a5f1b1fa71e.

Now, perhaps it's "not needed" because `LinkContext` does the same thing…but does it?

Also, where *is* `LinkContext` defined?  The [`LinkContext.cs` in mono/linker](https://github.com/mono/linker/blob/a621b9a630950026f5508c8089a59c0e825bd6e0/src/linker/Linker/LinkContext.cs) doesn't contain an `AreParametersCompatibleWith()` method, e.g. used here: https://github.com/xamarin/xamarin-android/pull/5748/files#diff-973bf7b778beba59537f406bfc5abcacba0e5d76aac82068a2d15908e8fd5db8R22

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/Extensions.cs
```diff
@@ -11,6 +11,10 @@
 using Java.Interop;
 using Java.Interop.Tools.Cecil;
 
+#if NET5_LINKER
+using Microsoft.Android.Sdk.ILLink;
+#endif
```
jonpryor, line 16: (Attempted) convention: please use a comment on `#else`/`#endif` with the same expression as the `#if` they correspond to, e.g.

```csharp
#if NET5_LINKER
…
#else   // !NET5_LINKER
…
#endif  // !NET5_LINKER
```

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -19,43 +19,79 @@ namespace MonoDroid.Tuner
 	/// <summary>
 	/// NOTE: this step is subclassed so it can be called directly from Xamarin.Android.Build.Tasks
 	/// </summary>
-	public class FixAbstractMethodsStep : BaseStep
+	public class FixAbstractMethodsStep
+#if NET5_LINKER
+	: IMarkHandler
+	{
```
jonpryor, line 25: This structure is likely to "confuse"certain editors that "count" braces w/o knowing about `#if` precedence.

Please keep braces either fully "in" or fully "out", e.g.

```csharp
public class FixAbstractmethodsStep :
#if NET5_LINKER
    IMarkHandler
#else   // !NET5_LINKER
    BaseStep
#endif  // !NET5_LINKER
{

#if NET5_LINKER
    protected LinkContext Context { get; private set; }
    protected AnnotationStore Annotations => Context?.Annotations;
    // ...
#endif  // !NET5_LINKER
}
```

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -297,6 +338,20 @@ void AddNewExceptionMethod (TypeDefinition type, MethodDefinition method)
 			}
 		}
 
+		bool markedAbstractMethodErrorType;
+
+		void MarkAbstractMethodErrorType () {
```
jonpryor, line 343: Convention: `{` on new line.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/PreserveApplications.cs
```diff
@@ -12,26 +12,64 @@
...
+#else
+	: BaseSubStep
+#endif
+	{
+
+#if NET5_LINKER
+		LinkContext context;
+		AnnotationStore Annotations => context?.Annotations;
+
+		public void Initialize (LinkContext context, MarkContext markContext)
```
jonpryor, line 27: Given all the new `Initialize()` additions, *and* that these are all very similar, and that all the *removals* are for `BaseStep`/`BaseSubStep`, might it be "cleaner" -- in terms of `diff` size -- to *re-introduce* these types?

```csharp
#if NET5_LINKER

public class BaseStep : IMarkHandler {
    LinkContext context;
    protected AnnotationStore Annotations => context?.Annotations;

    public virtual void Initialize (LinkContext context, MarkContext markContext)
    {
        this.context = context;
    }
}

public class BaseSubStep : BaseStep {
}

#endif  // !NET5_LINKER
```

This would allow us to have:

```csharp
public class PreserveApplications : BaseSubStep // no change!
{
    // .NET 5 specific; just override Initialize()
#if NET5_LINKER
    public override void Initialize (LinkContext context, MarkContext markContext)
    {
        base.Initialize (context, markContext);
        markContext.RegisterMarkAssemblyAction(…);
        markContext.RegisterMarkTypeAction(…);
    }
#endif  // !NET5_LINKER
}
```

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
sbomer: See https://github.com/xamarin/xamarin-android/pull/5748#discussion_r606634212 for context - I unresolved it in case you want to continue the discussion there.

The idea was to move to a similar cache that's built-in to the linker. `AreParametersCompatibleWith` is defined in `LinkContextExtensions.cs` (copied from java.interop's [MethodDefinitionRocks](https://github.com/xamarin/java.interop/blob/main/src/Java.Interop.Tools.Cecil/Java.Interop.Tools.Cecil/MethodDefinitionRocks.cs#L54) and adapted to use the built-in cache).

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
sbomer: Agreed, in an ideal world the type hierarchy logicwith `Resolve` caching would be built-in to cecil. This version of `GetBaseDefinition` exists just to use a cache that's built-in to the linker in .NET 6. (It's actually copied from the java.interop version). :grimacing:

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
sbomer: I see what you're saying - I am wondering whether it's more important to be consistent with java.interop's `GetBaseDefinition` or to be correct in this case. I will leave it as-is until we decide what to do with the duplicated logic.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/PreserveApplications.cs
sbomer: I addressed part of this by adding a `BaseMarkHandler` class with the shared members. Some of the .NET6 steps still need to be old-style `BaseSubStep`s though - for example `ApplyPreserveAttribute`. To merge the base classes, the new `BaseSubStep` would need to implement `ISubStep` and `IMarkHandler`This would let us get rid of this pattern in a lot of places:
```csharp
#if !NET5_LINKER
	override
#endif  // !NET5_LINKER
```

But there are also some downsides:
- it would duplicate https://github.com/mono/linker/blob/main/src/linker/Linker.Steps/BaseSubStep.cs
- it will be less obvious whether a given derived type of `BaseSubStep` is behaving like an `ISubStep` or an `IMarkHandler` (it depends how it is set up in the pipeline)

Let me know what you think!

Also, the old `BaseStep` is still used (for example `AddKeepAlivesStep`), and it can't implement both `IStep` and `IMarkHandler` otherwise the linker won't know which interface to use for the `--custom-step` argument (I think the failure mode is that it would treat all `BaseStep`s as `IMarkHandler`s). So I would leave `BaseStep` as-is, but it's worth considering for `BaseSubStep`.

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
jonpryor: Also, re-asking:

Where *is* `LinkContext` defined?  The [`LinkContext.cs` in mono/linker](https://github.com/mono/linker/blob/a621b9a630950026f5508c8089a59c0e825bd6e0/src/linker/Linker/LinkContext.cs) doesn't contain an `AreParametersCompatibleWith()` method, e.g. used here: https://github.com/xamarin/xamarin-android/pull/5748/files#diff-973bf7b778beba59537f406bfc5abcacba0e5d76aac82068a2d15908e8fd5db8R22

Furthermore, I can't find `AreParametersCompatibleWith()` on *any* mono/linker branch:

```
$ git grep AreParametersCompatibleWith $(git rev-list --all)
# no matches
```

This is on a just-fetched repo, with HEAD at mono/linker@e069cd8d.

How does this PR even *build*?

/me checks PR status, realizes that it hasn't even been built as of the most recent commits to the branch. :thinking:

jonpryor: > Furthermore, I can't find `AreParametersCompatibleWith()` on any mono/linker branch:

:rofl: it's the extension method in this PR!

https://github.com/xamarin/xamarin-android/pull/5748/files#diff-973bf7b778beba59537f406bfc5abcacba0e5d76aac82068a2d15908e8fd5db8R30

/me hangs head in shame.

jonpryor: @sbomer: I wrote this up: https://github.com/xamarin/java.interop/pull/842

The idea being that if you also update `LinkContext` to likewise implement `IMetadataResolver`, most/all of `LinkContextExtensions.cs` won't be needed, removing "one" of the algorithmic duplications.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -19,43 +19,79 @@ namespace MonoDroid.Tuner
...
+	public class FixAbstractMethodsStep :
+#if NET5_LINKER
+	BaseMarkHandler
+#else   // !NET5_LINKER
+	BaseStep
+#endif  // !NET5_LINKER
 	{
+
+#if NET5_LINKER
+		public void Initialize (LinkContext context, MarkContext markContext)
```
jonpryor, line 31: Shouldn't this be `override`?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/MarkJavaObjects.cs
```diff
@@ -2,20 +2,39 @@
...
+#if NET5_LINKER
+	BaseMarkHandler
+#else   // !NET5_LINKER
+	BaseSubStep
+#endif  // !NET5_LINKER
+	{
 		Dictionary<ModuleDefinition, Dictionary<string, TypeDefinition>> module_types = new Dictionary<ModuleDefinition, Dictionary<string, TypeDefinition>> ();
 
+#if NET5_LINKER
+		public void Initialize (LinkContext context, MarkContext markContext)
```
jonpryor, line 22: This should likewise be `override`.

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
sbomer: @jonpryor Thanks, I like this idea a lot. To make sure I understand how this fits together, is this an accurate summary?
- java.interop will use `TypeDefinitionCache` as the `IMetadataResolver`
- the xamarin-android linker steps will use `LinkContext` as the `IMetadataResolver`
  - which uses the linker's built-in caching instead of `TypeDefinitionCache`
  - `LinkContext` doesn't take any extra ctor arguments

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
sbomer: Good catch, thanks!

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
sbomer: > it's the extension method in this PR!

😆 Sorry, I should've linked to it in my answer!

> realizes that it hasn't even been built as of the most recent commits to the branch.

I just have to remember to ping @jonathanpeppers each time I want to trigger ci

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/PreserveApplications.cs
jonpryor: @sbomer: I like `BaseMarkHandler` as a compromise.

That said, I'm still not fully clear on everything.  Why does `ApplyPreserveAttribute` need to be an old-style `BaseSubStep`?  Presumably because of `PreserveSubStepDispatcher`'s `new ISubStep[]{new ApplyPreserveAttribute()}`, but…why is `PreserveSubStepDispatcher` required?  Why can't `ApplyPreserveAttribute` be an `IMarkHandler`?

Ditto `AddKeepAlivesStep`.

Related: what is the "long term evolution" of mono/linker?  Will `ISubStep` & related stick around, or are those going away?

Why can't both `IStep` and `IMarkHandler` be implemented?  You mention:

> the linker won't know which interface to use for the `--custom-step` argument (I think the failure mode is that it would treat all `BaseStep`s as `IMarkHandler`s).

(1) *Is* that a failure mode?  It sounds useful.  (2) How is that a problem?

Another problem for me is that I don't understand the difference between `IStep` and `IMarkHandler`; I *thought* `IMarkHandler` was the replacement for `IStep`/etc., so if `IMarkHandler` is invoked, and we can use `IMarkHandler` (…can we? See also above questions about `ApplyPreserveAttribute` & `AddKeepAlivesStep`), then I don't see an actual problem here.

---

I suspect that the "answer" to my questions lies in the original PR message:

> The exceptions are:
> 
> * ApplyPreserveAttribute: …

The problem is that I don't understand the underlying context.  (Docs? :smile:). mono/linker#1774 mentions "run logic on mark".  How many "steps" are there in the linker, what is their ordering, etc.? When is "mark" vs. "IStep", and how do they relate?  Why *can't* "everything" become IMarkHandler?  Why are there exceptions?

I don't understand. :frowning_face:

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
jonpryor: @sbomer wrote:

> To make sure I understand how this fits together…

There are two similar-yet-distinct answers here.

Option 1, proposed by @jonathanpeppers [on Discord](https://discord.com/channels/732297728826277939/732297837953679412/844308968243593216):

> Could we just subclass `TypeDefinitionCache` and override `Resolve()` so it uses the `LinkContext`?

```csharp
partial class LinkContextTypeDefinitionCache : TypeDefinitionCache {
    LinkContext context;

    public LinkContextTypeDefinitionCache (LinkContext context)
    {
        this.context = context;
    }

    public override TypeDefinition Resolve (TypeReference typeReference)
    {
         return context.ResolveTypeDefinition (typeReference);
    }
}
```

This would require *no* changes to Java.Interop.

Option 2 is to merge https://github.com/xamarin/java.interop/pull/842 and then either:

1. Update `LinkContext` to implement `IMetadataResolver`, or
2. Add a new `LinkContextMetadataResolver` type, along the lines of the above `LinkContextTypeDefinitionCache` type, which implements `IMetadataResolver` in terms of `LinkContext`.

Either/both of these would remove the need for `LinkContextExtensions.cs`, and should allow using the existing `GetBaseDefinition()` extension method from Java.Interop without any changes, and would remove the need for `LinkContextExtensions.cs`.  (I think.)

Is there a preference between (1) or (2)?  Of course, *Both* is a viable answer: (1) could be used while waiting for (2) to be completed, and if we choose the same name -- `LinkContextMetadataResolver`? -- then future "integration" changes are reduced in scope.

jonpryor: @sbomer: related question: are you actively on Discord?  Your nick appears to be there, but I don't see any messages from you in recent history.  Or, are you a *different* `@sbomer` than is on our Discord? https://discord.gg/95Pww5tQ

sbomer: Yup that's me! I haven't been active on it, but I will start checking it more frequently.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/PreserveApplications.cs
sbomer: > I don't understand the difference between IStep and IMarkHandler; I thought IMarkHandler was the replacement for IStep

The linker will call one `IStep` at a time, in the pipeline order, each operating on the set of assemblies loaded so far. `IMarkHandler` instead can register actions that get called, during MarkStep, whenever certain things are marked - it's designed for logic that should run over marked things instead of the entire set of loaded assemblies.

> Why does ApplyPreserveAttribute need to be an old-style BaseSubStep? [...] Why can't ApplyPreserveAttribute be an IMarkHandler?

That step needs to scan code for attributes, including _unmarked_ code. So as a compromise, the proposed behavior is to scan for attributes in all marked assemblies. To do this it could be an `IMarkHandler` but it would have to contain the logic to scan over all members in an assembly. That's where the `PreserveSubStepDispatcher` comes in. It's derived from [`MarkSubStepsDispatcher`](https://github.com/mono/linker/blob/8f1fc2560ce846fb4b8c85fc398d054ba58039ab/src/linker/Linker.Steps/MarkSubStepsDispatcher.cs#L13-L19) which handles the scanning part, and allows substeps including `ApplyPreserveAttribute` to only contain the logic to process members in the marked assembly.

> Ditto AddKeepAlivesStep.

This one isn't triggered by marking - instead it runs for everything that's left over after the Mark+Sweep, so it's just a plain old step:
https://github.com/xamarin/xamarin-android/blob/cf84e1b9a172a1e31d7eabf2f5eecd57ab57f301/src/Microsoft.Android.Sdk.ILLink/SetupStep.cs#L66

> what is the "long term evolution" of mono/linker? Will ISubStep & related stick around, or are those going away?

I don't see any reason for `ISubStep`/`IStep` to go away; they're still used in lots of places, here and in xamarin-macios. I might not be the best person to comment on the _long_ term evolution, but for .NET 6 our main goal has been to provide correctness guarantees with static analysis warnings. The new APIs make it possible to correctly handle assemblies that are dynamically loaded (discovered by the linker only during MarkStep), and also allow custom steps to avoid scanning the world for operations that only need to touch marked things. /cc @marek-safar

> Why can't both IStep and IMarkHandler be implemented?

>> I think the failure mode is that it would treat all BaseSteps as IMarkHandlers

> (1) Is that a failure mode? It sounds useful.

The old-style BaseSteps weren't designed to be triggered by mark, and probably wouldn't do the right thing - `IStep` is a different beast than `IMarkHandler`. If there are any specific old-style BaseSteps that could be changed to run for marked members, we could make them `IMarkHandlers`, but any that need to go after CleanStep would be unchanged.

> How many "steps" are there in the linker, what is their ordering

This in particular is something I've been trying to untangle - the ordering is important in a lot of cases but it's not obvious when. This comment should give the current ordering for built-in steps: https://github.com/mono/linker/blob/main/src/linker/Linker/Driver.cs#L714.

I hope that helps a little bit - I'll update the linker docs to make some of this easier to discover in the future.

sbomer:
@jonathanpeppers mind triggering again please?

src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj
sbomer: Thanks for clarifying. I like the combination of the options. I implemented 1 (with the name `LinkContextMetadataResolver`) here  for now. I also have a PR out for option 2.1, so we can see what @marek-safar thinks of implementing `IMetadataResolver` directly on `LinkContext`.

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
sbomer: Removed in the latest iteration

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
sbomer: @jonpryor now that I've removed the copy of these extension methods - is it worth fixing this in java.interop?

jonpryor:
PR #5945 pulls in a Java.Interop with the `IMetadataResolver` API changes.

jonpryor:
Current state of unit test affairs:

  * **MSBuild With Emulator - macOS - Legacy** > **Run check-boot-times**: failure is somewhat expected, should be fixed by 85d6938c5c177aa32dd820195adc0b1e877d53c6.  Ignorable.
  * **APKs Legacy - macOS** > ** run VSAndroidApp**: failure is somewhat expected; should be fixed by 85d6938c5c177aa32dd820195adc0b1e877d53c6.  Ignorable.
  * **MSBuildDeviceIntegration On Device - macOS - Legacy** > [**JsonDeserializationCreatesJavaHandle(False)**](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4799810&view=ms.vss-test-web.build-test-results-tab&runId=22022748&paneView=attachments&resultId=100029): failed due to an XA0010.  Ignorable.
  * **MSBuildDeviceIntegration On Device - macOS - One .NET** > [**LocalizedAssemblies_ShouldBeFastDeployed**](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4799810&view=ms.vss-test-web.build-test-results-tab&runId=22020358&paneView=attachments&resultId=100088): failed due to an XA0010.  Ignorable.
  * **MSBuildDeviceIntegration On Device - macOS - One .NET** > **JsonDeserializationCreatesJavaHandle(False)**: should be fixed by 85d6938c5c177aa32dd820195adc0b1e877d53c6.  Ignorable.
  * **MSBuildDeviceIntegration On Device - macOS - One .NET** > **RunWithInterpreterEnabled(True)**: should be fixed by 85d6938c5c177aa32dd820195adc0b1e877d53c6.  Ignorable.

I've restarted **MSBuild Legacy - Windows-2** and **MSBuild One .NET - Windows-3** suites, as thy previously because they couldn't download test infrastructure.

src/Microsoft.Android.Sdk.ILLink/BaseMarkHandler.cs
```diff
@@ -0,0 +1,18 @@
+using Mono.Linker.Steps;
+using Microsoft.Android.Sdk.ILLink;
+
+namespace Mono.Linker
+{
+	public class BaseMarkHandler : IMarkHandler
+	{
+		protected LinkContext Context;
+		protected AnnotationStore Annotations => Context?.Annotations;
+		protected LinkContextMetadataResolver resolver;
```
jonpryor, line 10: Should this field instead be of type `IMetadataResolver`?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AddKeepAlivesStep.cs
```diff
@@ -10,17 +10,30 @@
...
+using Microsoft.Android.Sdk.ILLink;
+#endif  // NET5_LINKER
 
 namespace MonoDroid.Tuner
 {
 	public class AddKeepAlivesStep : BaseStep
 	{
+
+#if NET5_LINKER
+		LinkContextMetadataResolver cache;
```
jonpryor, line 23: If we change `cache` to be an `IMetadataResolver`, that will reduce what the `#if NET5_LINKER` needs to "protect"/differentiate.  `AddKeepAliveStep()` could likewise us an `IMetadataResolver cache` field.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -19,43 +19,79 @@ namespace MonoDroid.Tuner
...
 	/// </summary>
-	public class FixAbstractMethodsStep : BaseStep
+	public class FixAbstractMethodsStep :
+#if NET5_LINKER
+	BaseMarkHandler
+#else   // !NET5_LINKER
+	BaseStep
+#endif  // !NET5_LINKER
 	{
-		readonly TypeDefinitionCache cache;
```
jonpryor, line 24: Same comment as with `AddKeepAlivesStep`: this should instead become `IMetadataResolver cache`.  Then the `FixAbstractMethodStep()` constructor wouldn't need *any* changes.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
```diff
@@ -19,43 +19,79 @@ namespace MonoDroid.Tuner
...
 			if (IsProductOrSdkAssembly (assembly))
-				return;
+				return false;
 
 #if !NET5_LINKER
 			CheckAppDomainUsageUnconditional (assembly, (string msg) => Context.LogMessage (MessageImportance.High, msg));
-#endif
+#endif  // !NET5_LINKER
+
+			return assembly.MainModule.HasTypeReference ("Java.Lang.Object");
```
jonpryor, line 57: *This* concerns me, and I don't remember seeing it in a previous iteration of this PR. See also:

  * https://github.com/xamarin/xamarin-android/pull/5965
  * https://github.com/xamarin/xamarin-android/pull/4225#issuecomment-583139797

There is *no* requirement that an assembly which contains types which *need* to have abstract methods "fixed up" *also* contain a type reference to `Java.Lang.Object`, much less a reference to `Mono.Android.dll`!  Such is the glory of transitive references.

I'm fairly sure this check will *break* existing code.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
sbomer: Fixed. (I had originally changed the field name to `resolver` in this case, because I didn't want the field on `BaseMarkHandler` to be named `cache`, but it sounds like you prefer to reduce the size of the diff.)

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
sbomer: Your concern makes sense. This is an existing check (and was also present in previous iterations) - I moved it because this step now runs for marked types, instead of for all assemblies. I just filed https://github.com/xamarin/xamarin-android/issues/5968 about it - could this be addressed separately? I'm trying to minimize functional changes in this PR.

src/Microsoft.Android.Sdk.ILLink/BaseMarkHandler.cs
sbomer: Makes sense, fixed.

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
jonpryor: @sbomer asked:

> is it worth fixing this in java.interop?

Probably yes?  The problem is that I considered doing this fix as part of xamarin/java.interop@2573dc8c84fd4eb68e75bcae73912c26f4942356, but wasn't quite sure how to actually *do* it, based on the "scope" and information available.

In particular, `FixAbstractMethodStep.IsInOverrides()` (linked above) *knows* that it has an interface method `iMethod` and a type method `tMethod`.  It thus makes sense to ask "is `iMethod` one of the methods that `tMethod` overrides?

`FixAbstractMethodStep` can do this because it likewise knows when it's dealing with an interface, and thus *has* `iMethod`.

`MethodDefinitionRocks` in Java.Interop has *none* of that additional context, at least not structured in the same way as `FixAbstractMethodStep` thinks about things.

Consequently, while we (probably?) should fix this in Java.Interop, it wasn't a "trivial minimal thought" fix, so I skipped it.

I should file a bug about that, though…?

That said, the primary/original use case for `MethodDefinitionRocks` is in [`JavaCallableWrapperGenerator.cs`](https://github.com/xamarin/java.interop/blob/main/src/Java.Interop.Tools.JavaCallableWrappers/Java.Interop.Tools.JavaCallableWrappers/JavaCallableWrapperGenerator.cs), which is responsible for generating a Java "proxy" for a managed type, and it works by grabbing the closure of all interface methods *plus* all "override" methods to find the set of methods with a `[Register]` attribute, and if there's overlap between these approaches -- because a base class implements an interface via virtual method, so this lookup would find the interface method and virtual method separately -- then that's fine, we'll appropriately skip the "duplicate".

Thus, *in the context of Java.Interop*, this isn't a problem.

*In the context of the linker*… I have no idea.

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
jonpryor: Sorry, I didn't realize this check was originally there; I didn't see the movement from `FixAbstractMethodsUnconditional()`. :frowning_face:

jonpryor:
Proposed commit message:

```markdown
Context: https://github.com/mono/linker/pull/1774
Context: https://github.com/mono/linker/pull/1774#issuecomment-783563526
Context: https://github.com/xamarin/xamarin-android/pull/5870
Context: https://github.com/xamarin/xamarin-android/pull/5878

Update the custom linker steps to use the new `IMarkHandler` API
which runs logic during "MarkStep".

(See [this list][0] of pipeline steps for additional context.)

As part of this, I've removed the workaround that loads all referenced
assemblies up-front and simplified some of the linker MSBuild targets.
Some of the `MonoDroid.Tuner` steps were duplicated and changed to
implement the new `IMarkHandler` interface, to avoid touching the
`MonoDroid.Tuner` code.

In my analysis of the custom steps, most of them "just work" if we
call them only for marked members, because they ultimately either:

  - Just call `AddPreserved*()` to conditionally keep members of types
    (which with the new API will just happen when the type is marked)

  - In the case of `FixAbstractMethodsStep()`, inject missing interface
    implementations which will only be kept if they are marked for some
    other reason.

Most of the steps have been updated in a straightforward way based on
the above.

The exceptions are:

  - `AddKeepAlivesStep` needs to run on all code that survived
    linking, and can run as a normal step.

  - `ApplyPreserveAttribute`: this step globally marks members with
    `PreserveAttribute`.
    - The step is only active for non-SDK link assemblies.  Usually we
      root non-SDK assemblies, but it will be a problem if linking all
      assemblies.
    - For now, I updated it to use the new custom step API on assembly
      mark -- so it will scan for the attribute in all marked
      assemblies -- but we should investigate whether this is good
      enough.
    - This can't be migrated to use `IMarkHandler` because it needs
      to scan code for attributes, including unmarked code.

  - `PreserveExportedTypes`: similar to the above, this globally marks
    members with `Export*Attribute`.  It's used for java interop in
    cases where the java methods aren't referenced statically, like
    from xml, or for special serialization methods.
    - It's only active for non-SDK assemblies, so like above it will
      be a problem only if linking all assemblies.
    - Like above, I've made it scan marked assemblies.

  - `PreserveApplications`: globally scans for `ApplicationAttribute`
    on types/assemblies, and sets the `TypePreserve` annotation for
    any types referenced by the attribute.
    - This step technically does a global scan, but it's likely to work
      on "mark type"/"mark assembly", as `ApplicationAttribute` is only
      used on types/assembies that are already kept.
    - I've updated it to use the new `IMarkHandler` API.

TODO: the longer-term plan is to move all of the steps (including the
exceptions I mentioned above) to use the new custom steps API, and get
rid of the workaround that loads references up-front, even if there is
a small chance that they will miss `PreserveAttribute` and similar in
assemblies which are unused (never marked by the linker).

[0]: https://github.com/mono/linker/blob/main/src/linker/Linker/Driver.cs#L714-L730
```

src/Microsoft.Android.Sdk.ILLink/LinkContextExtensions.cs
sbomer: I don't know how important this is for the android linker steps either. Not sure if this is helpful, but I filed https://github.com/xamarin/xamarin-android/issues/5974 to track these concerns. I know the built-in linker steps have had issues related to this as well, so it is probably worth trying to get this right in the linker, with plenty of tests, etc. and then being consistent about it in xamarin-android.

sbomer:
@jonpryor thank you! I would remove the TODO - that work was done in this PR (I updated the original PR description to reflect this).

Could you also add:
```
Context: https://github.com/mono/linker/issues/1953
Context: https://github.com/xamarin/java.interop/pull/842
Context: https://github.com/mono/linker/pull/1979

This includes an update to remove the use of a `TypeDefinitionCache` shared
across steps, to allow passing individual custom steps as separate command-line
arguments. This cache is replaced by the use of a cache built in to the linker,
adapted to match the `IMetadataResolver` interface which is implemented on
`LinkContextMetadataResolver`.

TODO: Replace `LinkContextMetadataResolver` with `IMetadataResolver implemented
directly on `LinkContext` in https://github.com/mono/linker/pull/2045
TODO: Get rid of all reflection in SetupStep with the fix for shared state
in https://github.com/mono/linker/pull/2019
```

(Feel free to adjust wording or formatting)

jonpryor:
Indeed, commit da536fcf1b663fd71adee40c06466547a04cc6ed bumped our .NET 6 build, which includes a mono/linker bump, which includes mono/linker#2045…. Explains why my changes before the merge *also* failed to build, while this PR was otherwise green yesterday…

sbomer:
@jonpryor I addressed the remaining TODOs in the latest changes.

PR #5755, opened 2021-03-17
OpenTK Deprecation

dellis1972:
We are going to stop shipping OpenTK-1.0 in a future update. In order to give users who are using 
that assembly a chance to migrate we will deprecate it first. 
We will raise a `XA4313` warning if we detect that OpenTK-1.0 is in the `@(Refereces)` item group. 

A replacement Nuget Package has already been created at https://www.nuget.org/packages/Xamarin.Legacy.OpenTK/
Users seeing this warning should remove the reference to `OpenTK-1.0` and replace it will the following

```
<PackageReference Include="Xamarin.Legacy.OpenTK" Version="0.0.1-alpha" />
```

The version number will change in the future. 

The vast majority of changes in this commit are to stop the OpenTK java types being included in `mono-android.jar`. 
From this point on `OpenTK` will be treated like all other Android Class Libraries, the generated java code will be 
written to the `$(IntermediateOutputPath)\lp` directory along with all the other references.


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
```diff
@@ -190,7 +190,6 @@
     <!-- These files are build after Xamarin.Android.Build.Tasks but MUST be included in Profile.g.cs -->
     <_SharedRuntimeAssemblies Include="$(_SharedRuntimeBuildPath)$(AndroidFrameworkVersion)\Mono.Android.dll" />
     <_SharedRuntimeAssemblies Include="$(_SharedRuntimeBuildPath)$(AndroidFrameworkVersion)\OpenTK.dll" />
```
jonpryor, line 192: We can likely remove this line as well.

src/Xamarin.Android.Build.Tasks/Linker/Mobile.Tuner/MobileProfile.cs
```diff
@@ -54,7 +54,6 @@ public abstract class MobileProfile : Profile {
 			"Mono.Posix",
 			"Mono.Security",
 			"OpenTK",
```
jonpryor, line 56: We can likely remove this line as well.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -858,7 +858,15 @@ because xbuild doesn't support framework reference assemblies.
 	<AndroidUseManagedDesignTimeResourceGenerator Condition=" '$(AndroidUseManagedDesignTimeResourceGenerator)' == '' " >True</AndroidUseManagedDesignTimeResourceGenerator>
 </PropertyGroup>
 
-<Target Name="_CreatePropertiesCache" DependsOnTargets="_SetupDesignTimeBuildForBuild;_SetLatestTargetFrameworkVersion;_ResolveMonoAndroidSdks">
+<Target Name="_CheckForObsoleteFrameworkAssemblies">
```
jonathanpeppers, line 861: Can we put this target in `Xamarin.Legacy.targets`? So it won't run for .NET 6?

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -723,6 +723,15 @@ The following are literal names and should not be translated: ABI, 'libs/armeabi
     <value>Referencing the Android Wear application project '{0}' from an Android application project is deprecated and will no longer be supported in a future version of Xamarin.Android. Remove the Android Wear application project reference from the Android application project and distribute the Wear application as a standalone application instead.</value>
     <comment>The following are literal names and should not be translated: Android Wear, Android, Wear.
 {0} - The referenced Android Wear project.</comment>
+  </data>
+  <data name="XA4313" xml:space="preserve">
+    <value>'{0}' has been deprecated. You should replace it with the '{1}' Nuget Package.
```
jonathanpeppers, line 728: Some suggestions on the warning message:
```suggestion
    <value>The built-in '{0}' reference has been deprecated.
Remove the '{0}' reference from your project and add the '{1}' NuGet Package instead.
```

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets
```diff
@@ -22,6 +22,7 @@ projects, these properties are set in Xamarin.Android.Legacy.targets.
       _CleanIntermediateIfNeeded;
       _CheckProjectItems;
       _CheckForContent;
+      _CheckForObsoleteFrameworkAssemblies;
```
jonpryor, line 25: This is causing tests to fail under One.NET, e.g. https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4679171&view=ms.vss-test-web.build-test-results-tab&runId=20951028&resultId=100036&paneView=attachments

```
/Users/runner/Library/Android/dotnet/sdk/6.0.100-preview.4.21215.1/Microsoft.Common.CurrentVersion.targets(882,7):
error MSB4057: The target "_CheckForObsoleteFrameworkAssemblies" does not exist in the project. 
[/Users/runner/work/1/s/bin/TestRelease/temp/DuplicateRJavaOutput/UnnamedProject.csproj]
```

`_CheckForObsoleteFrameworkAssemblies` was only defined in `Xamarin.Android.Legacy.targets`: https://github.com/xamarin/xamarin-android/pull/5755/files#diff-47d20207182aad2c752c0fa397c77bd67457c3cda762b716be47e8d45c1ccda5R230

Perhaps `Microsoft.Android.Sdk.BuildOrder.targets` shouldn't be modified at all?

Documentation/guides/messages/xa4313.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Xamarin.Android error/warning XA4313
+description: XA4313 error/warning code
+ms.date: 03/19/2021
+---
+# Xamarin.Android error/warning XA4313
+
+The specified Framework assembly has been deprecated.
+You need up upgrade to use the specified NuGet Package.
```
akoeplinger, line 9:
```suggestion
You need an upgrade to use the specified NuGet package.
```

Documentation/guides/messages/xa4313.md
```diff
@@ -0,0 +1,20 @@
...
+
+For example, to upgade `OpenTK-1.0` remove
```
akoeplinger, line 11:
```suggestion
For example, to upgrade `OpenTK-1.0` remove
```

Documentation/guides/messages/xa4313.md
```diff
@@ -0,0 +1,20 @@
...
+
+    '<Reference Include="OpenTK-1.0" />'
+
+from your csproj then add
+
+    '<PackageReference Include="Xamarin.Legacy.OpenTK" Version="1.0.0" />'
```
akoeplinger, line 17: the version should probably be `0.0.1-alpha` ?

dellis1972: I'll probably push a release 1.0.0 before we ship this.

jonathanpeppers (commented):
Should we remove these, too?

https://github.com/xamarin/xamarin-android/blob/7f32ef06011dfcc4d2f54a6a3fc84cb3ddc7bc50/src/Xamarin.Android.Build.Tasks/Resources/proguard_xamarin.cfg#L13-L16

dellis1972:
> Should we remove these, too?
> 
> https://github.com/xamarin/xamarin-android/blob/7f32ef06011dfcc4d2f54a6a3fc84cb3ddc7bc50/src/Xamarin.Android.Build.Tasks/Resources/proguard_xamarin.cfg#L13-L16

I'm not sure why those are there. I suspect they will still be needed even with the Nuget.

jonpryor:
@dellis1972 [wrote](https://github.com/xamarin/xamarin-android/pull/5755#issuecomment-825662061):

> > Should we remove these, too?
> > https://github.com/xamarin/xamarin-android/blob/7f32ef06011dfcc4d2f54a6a3fc84cb3ddc7bc50/src/Xamarin.Android.Build.Tasks/Resources/proguard_xamarin.cfg#L13-L16
> 
> I'm not sure why those are there. I suspect they will still be needed even with the Nuget.

They need to be present *somewhere*, even with the NuGet, and even under .NET 6.

That doesn't mean that the xamarin-android repo is the appropriate place fo the ProGuard config info.  If possible, can the OpenTK-related statements within `progurad_xamarin.cfg` be *copied into* the `Xamarin.Legacy.OpenTK` package?  It won't be needed for Legacy SDK, but we *should* want it for .NET 6.

PR #5760, opened 2021-03-19
[Xamarin.Android.Build.Tasks] Xamarin.Android deployment fails unless project is cleaned first

dellis1972:
Fixes https://work.azdo.io/1288593

This is a weird bug which I cannot reproduce locally. However reading
the code this code path is possible.

When trying to do incremental builds when using the `AndroidPackageFormat`
`aab` we get the following error

     [BT : 1.4.0] error : Module 'base' is missing mandatory file 'manifest/AndroidManifest.xml'.

The problem is the base.zip doesn't have ANY AndroidManifest.xml in it...
But why?

Well what seems to happen is in this case the input file `packaged_resources`
and the output `base.zip` are unchanged. `packaged_resources` has
an `AndroidManifest.xml` file and `base.zip` has a `manifest/AndroidManifest.xml`
file. Because it was renamed from `AndroidManifest.xml` to `manifest/AndroidManifest.xml`
on a previous build. This is so that our `base.zip` confirms to the expected
folder structure for an `aab`. So this is by design.

However, in this case because BOTH files are up to date, we DONT add
the `AndroidManifest.xml` file from `packaged_resources` to the `base.zip`.
A compounding problem is because no physical file `manifest/AndroidManifest.xml`
exists, that file is then removed from the `base.zip`.
As a result the `FixupArchive` which does the renaming of the `AndroidManifest.xml`
for `aab` files has nothing to do. Because its source file does not exist.

So we end up with a `base.zip` without ANY `AndroidManifest.xml` file in it.

The fix in this case is to NEVER remove an `AndroidManifest.xml` file from
the base.zip or apk. It is a protected file and should always exist.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -158,8 +158,10 @@ namespace Xamarin.Android.Tasks
 							}
 							Log.LogDebugMessage ($"Deregistering item {entryName}");
 							existingEntries.Remove (entryName);
-							if (lastWriteInput <= lastWriteOutput)
+							if (lastWriteInput <= lastWriteOutput) {
+								Log.LogDebugMessage ($"Skipping to next item. {lastWriteInput} <= {lastWriteOutput}.");
 								continue;
+							}
 							if (apk.Archive.ContainsEntry (entryName)) {
 								ZipEntry e = apk.Archive.ReadEntry (entryName);
 								// check the CRC values as the ModifiedDate is always 01/01/1980 in the aapt generated file.
@@ -266,6 +268,9 @@ namespace Xamarin.Android.Tasks
 				}
 				// Clean up Removed files.
 				foreach (var entry in existingEntries) {
+					// never remove an AndroidManifest. It may be renamed when using aab.
+					if (string.Compare (Path.GetFileName (entry), "AndroidManifest.xml", StringComparison.OrdinalIgnoreCase) == 0)
+						continue;
 					Log.LogDebugMessage ($"Removing {entry} as it is not longer required.");
 					apk.Archive.DeleteEntry (entry);
 				}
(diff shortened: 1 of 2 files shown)


jonathanpeppers (approved):

jonpryor:
```markdown
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1288593

This is a weird bug which I cannot reproduce locally.  However,
reading the code this code path is possible:

When trying to do incremental builds when using
`$(AndroidPackageFormat)`=aab, we get the following error:

	[BT : 1.4.0] error : Module 'base' is missing mandatory file 'manifest/AndroidManifest.xml'.

The problem is that `base.zip` doesn't have *any*
`AndroidManifest.xml` in it!  But why?

What appears to happen is:

 1. During "initial build", `packaged_resources` is created, and
    contains `AndroidManifest.xml`.  `base.zip` is created based on
    the contents of `packaged_resources`, renaming
    `AndroidManifest.xml` to `manifest/AndroidManifest.xml`, so that
    `base.zip` has the expected folder structure for `aab` files.
    (This "copy and rename" is done by `<FixupArchive/>`.)

    Both `packaged_resources` and `base.zip` have the same timestamp.

 2. On a subsequent incremental build, neither `packaged_resources`
    nor `base.zip` is changed (because they didn't need to, e.g. no
    Android Resources were added), and the `_BuildApkEmbed` target is
    executed.

 3. The `_BuildApkEmbed` target runs the `<BuildApk/>` task, which
    "cleans up removed files" within `base.zip`.  Because there is no
    corresponding file "on disk" for `manifest/AndroidManifest.xml`
    within `base.zip`, the `manifest/AndroidManifest.xml` file is
    *removed* from `base.zip`.

 4. `_BuildApkEmbed` later calls `<BuildAppBundle/>`, which calls
    `bundletool.jar`, and fails with the above error message:

        [BT : 1.4.0] error : Module 'base' is missing mandatory file 'manifest/AndroidManifest.xml'.

The fix is to update the `<BuildApk/>` task to *never remove* files
named `AndroidManifest.xml`.  It is a protected file and should
always exist.
```

PR #5775, opened 2021-03-25
[runtime] Add Support for `/data/local/tmp/` override path.

dellis1972:
We are getting allot of reports from users that the new fast deployment
system is not working. The decision made to not support
fast dev on devices which do not support `run-as` might well have been
incorrect.

So we need a backup path. The only known writeable area on a device is
`/data/local/tmp`. So the idea will be that IF the device does NOT support
`run-as` we will fall back to `/data/local/tmp`.

This backup path will not be as quick since we will be using the standard
`adb push` commands. Also it will have the disadvantage of the directory
NOT being deleted when the app is removed from the device. So files will
be left in place until the OS decides to clean them up. So it is not ideal
but it might help customers who need to develop on older hardware.

- [ ] monodroid Fast Deployment changes.
- [ ] Add Unit test to test the backup path.

### DIFF
--- a/src/java-runtime/java/mono/android/DebugRuntime.java
+++ b/src/java-runtime/java/mono/android/DebugRuntime.java
@@ -4,5 +4,5 @@ public class DebugRuntime {
 	private DebugRuntime ()
 	{}
 
-	public static native void init (String[] apks, String runtimeLibDir, String[] appDirs);
+	public static native void init (String[] apks, String runtimeLibDir, String[] appDirs, String packageName);
 }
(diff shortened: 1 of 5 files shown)


Syed-RI:
Eagerly waiting on the fix! 🙏🏼

dellis1972:
Closing this PR. Its no longer needed thanks to https://github.com/xamarin/monodroid/commit/e57630bd3a790e5772ae030e517326d34f38431d

Syed-RI:
@dellis1972 given the monodroid is private repo can you please share a tl;dr for us mere mortals please?

dellis1972:
Well there might be some older devices that do not support run-as. But given our minimum API level is something like 21 now (not 10 when this PR was opened), its more than likely `run-as` is supported. 

The PR in question as a more resilient fast deployment system which has a simplified "backup" deployment which copies data to `/data/local/tmp` before placing it in the right place. It should make deployment more reliable if we hit a problem.

Issue #5751, opened 2021-03-16
Update linker defaults for .NET 6


sbomer:
In https://github.com/dotnet/sdk/pull/16327, the .NET SDK is changing the linker defaults to `TrimMode=link`, with `SuppressTrimAnalysisWarnings=false`. Xamarin-android should consider doing the following when this change flows:
- Default `SuppressTrimAnalysisWarnings` to true to avoid producing warnings
- Avoid setting `TrimMode` in https://github.com/xamarin/xamarin-android/blob/363935b4589b12ada06f6a0e584b547bb78178ff/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets#L52

PR #5781, opened 2021-03-25
[One .NET] default $(SuppressTrimAnalysisWarnings) to true

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5751

After bumping to .NET 6 Preview 3, we get hundreds of ILLINK warnings.
This is because of:

https://github.com/dotnet/sdk/pull/16327

We need to default `$(SuppressTrimAnalysisWarnings)` to ignore these
warnings for now.

We should continue to use `TrimMode=link`, as that is the behavior
Xamarin developers get today.


dellis1972 (approved):

marek-safar:
> We should continue to use TrimMode=link, as that is the behavior Xamarin developers get today.

As explained in the issue that code is now dead because it's exactly what sdk does for everyone (see https://github.com/dotnet/sdk/blob/main/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.ILLink.targets#L193)

jonathanpeppers:
> As explained in the issue that code is now dead because

I see now @marek-safar, thanks. I'll remove `TrimMode=link` when I fix: https://github.com/xamarin/xamarin-android/issues/5856

PR #5785, opened 2021-03-26
[MSBuildDeviceIntegration] Add Xamarin.Android unit test which uses ALL available Languages.

dellis1972:
Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1301607

Add a unit test which tests ALL the languages we can to make sure we can Fast Deploy.


dellis1972:
Needs https://github.com/xamarin/monodroid/pull/1189 to make the test pass

PR #5790, opened 2021-03-29
Disable assembly preload by default

grendello:
Disable the current default behavior of Xamarin.Android runtime to
preload all the assemblies shipped in the apk on application startup.
This behavior has been the default in order to support older versions of
`Xamarin.Forms`, sacrificing startup performance.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -290,7 +290,8 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidMakeBundleKeepTemporaryFiles Condition=" '$(AndroidMakeBundleKeepTemporaryFiles)' == '' ">False</AndroidMakeBundleKeepTemporaryFiles>
 
 	<!-- If true it will cause all the assemblies in the apk to be preloaded on startup time -->
-        <_AndroidEnablePreloadAssembliesDefault>True</_AndroidEnablePreloadAssembliesDefault>
+	<_AndroidEnablePreloadAssembliesDefault Condition=" '$(UsingAndroidNETSdk)' == 'true' ">False</_AndroidEnablePreloadAssembliesDefault>
+	<_AndroidEnablePreloadAssembliesDefault Condition=" '$(UsingAndroidNETSdk)' != 'true' ">True</_AndroidEnablePreloadAssembliesDefault>
 	<AndroidEnablePreloadAssemblies Condition=" '$(AndroidEnablePreloadAssemblies)' == '' ">$(_AndroidEnablePreloadAssembliesDefault)</AndroidEnablePreloadAssemblies>
 	<_NativeAssemblySourceDir>$(IntermediateOutputPath)android\</_NativeAssemblySourceDir>
 	<_AndroidUseNewTypemaps>True</_AndroidUseNewTypemaps>
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -3932,7 +3932,11 @@ namespace UnnamedProject
 				Assert.IsTrue (b.Build (proj), "build should have succeeded.");
 				var environment = b.Output.GetIntermediaryPath (Path.Combine ("__environment__.txt"));
 				FileAssert.Exists (environment);
-				Assert.AreEqual ($"__XA_PACKAGE_NAMING_POLICY__={packageNamingPolicy}", File.ReadAllText (environment).Trim ());
+				if (Builder.UseDotNet) {
+					Assert.AreEqual ($"__XA_PACKAGE_NAMING_POLICY__={packageNamingPolicy}{Environment.NewLine}mono.enable_assembly_preload=0", File.ReadAllText (environment).Trim ());
+				} else {
+					Assert.AreEqual ($"__XA_PACKAGE_NAMING_POLICY__={packageNamingPolicy}", File.ReadAllText (environment).Trim ());
+				}
 			}
 		}
 


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -290,7 +290,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidMakeBundleKeepTemporaryFiles Condition=" '$(AndroidMakeBundleKeepTemporaryFiles)' == '' ">False</AndroidMakeBundleKeepTemporaryFiles>
 
 	<!-- If true it will cause all the assemblies in the apk to be preloaded on startup time -->
-        <_AndroidEnablePreloadAssembliesDefault>True</_AndroidEnablePreloadAssembliesDefault>
+        <_AndroidEnablePreloadAssembliesDefault>False</_AndroidEnablePreloadAssembliesDefault>
 	<AndroidEnablePreloadAssemblies Condition=" '$(AndroidEnablePreloadAssemblies)' == '' ">$(_AndroidEnablePreloadAssembliesDefault)</AndroidEnablePreloadAssemblies>
```
jonathanpeppers, line 294: Maybe we should only do this for .NET 6, so instead something like:
```xml
<AndroidEnablePreloadAssemblies Condition=" '$(AndroidEnablePreloadAssemblies)' == '' and '$(UsingAndroidNETSdk)' == 'true' ">false</AndroidEnablePreloadAssemblies>
<AndroidEnablePreloadAssemblies Condition=" '$(AndroidEnablePreloadAssemblies)' == '' and '$(UsingAndroidNETSdk)' != 'true' ">true</AndroidEnablePreloadAssemblies>
```

grendello: And drop `_AndroidEnablePreloadAssembliesDefault` completely?

grendello: A couple other places use `$(_AndroidEnablePreloadAssembliesDefault)` so I instead added the condition to `_AndroidEnablePreloadAssembliesDefault` itself

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -290,7 +290,8 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidMakeBundleKeepTemporaryFiles Condition=" '$(AndroidMakeBundleKeepTemporaryFiles)' == '' ">False</AndroidMakeBundleKeepTemporaryFiles>
 
 	<!-- If true it will cause all the assemblies in the apk to be preloaded on startup time -->
-        <_AndroidEnablePreloadAssembliesDefault>True</_AndroidEnablePreloadAssembliesDefault>
+        <_AndroidEnablePreloadAssembliesDefault Condition=" '$(UsingAndroidNETSdk)' == 'true' ">False</_AndroidEnablePreloadAssembliesDefault>
+        <_AndroidEnablePreloadAssembliesDefault Condition=" '$(UsingAndroidNETSdk)' == 'false' Or '$(UsingAndroidNETSdk)' == '' ">True</_AndroidEnablePreloadAssembliesDefault>
```
jonpryor, line 294: I would suggest using `!= 'true`, which will nicely match both `false` and empty string and `/p:UsingAndroidNETSDK=ThisIsASeriouslyWeirdValue`.

jonpryor:
We now have unit test failures, e.g. [`PackageNamingPolicy("LowercaseMD5")`](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4602258&view=ms.vss-test-web.build-test-results-tab&runId=20180394&resultId=100190&paneView=debug):

```
Expected string length 41 but was 72. Strings differ at index 41.
Expected: "...E_NAMING_POLICY__=LowercaseMD5"
But was:  "...E_NAMING_POLICY__=LowercaseMD5\nmono.enable_assembly_preload=0"
--------------------------------------------^
```

Appears to be this assert: https://github.com/xamarin/xamarin-android/blob/c9e2d7540393f25d854014d158b4eb7c66a2b7c2/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L3935

Is `$(AndroidEnablePreloadAssemblies)` supposed to impact `__environment__.txt`?  (I forget.)

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -290,7 +290,8 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidMakeBundleKeepTemporaryFiles Condition=" '$(AndroidMakeBundleKeepTemporaryFiles)' == '' ">False</AndroidMakeBundleKeepTemporaryFiles>
 
 	<!-- If true it will cause all the assemblies in the apk to be preloaded on startup time -->
-        <_AndroidEnablePreloadAssembliesDefault>True</_AndroidEnablePreloadAssembliesDefault>
+        <_AndroidEnablePreloadAssembliesDefault Condition=" '$(UsingAndroidNETSdk)' == 'true' ">False</_AndroidEnablePreloadAssembliesDefault>
+        <_AndroidEnablePreloadAssembliesDefault Condition=" '$(UsingAndroidNETSdk)' != 'true' ">True</_AndroidEnablePreloadAssembliesDefault>
```
jonpryor, line 294: Since this PR will require additional changes anyway…

Could you please indent this line to match the following line, removing some of the leading whitespace?

grendello:
> 
> Appears to be this assert:
> 
> https://github.com/xamarin/xamarin-android/blob/c9e2d7540393f25d854014d158b4eb7c66a2b7c2/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L3935
> 
> Is `$(AndroidEnablePreloadAssemblies)` supposed to impact `__environment__.txt`? (I forget.)

Yep, here: https://github.com/xamarin/xamarin-android/blob/c9e2d7540393f25d854014d158b4eb7c66a2b7c2/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1511-L1518

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
grendello: Done

jonathanpeppers (approved):

PR #5799, opened 2021-03-30
[build] Add Microsoft.DotNet.ApiCompat dependency

pjcollins:
Adds the Microsoft.DotNet.ApiCompat package to our dotnet dependency
management files so that it can be automatically updated as needed
in the future.  This package version can also be used for
Microsoft.DotNet.GenAPI, as both nugets are produced by the same build.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -9,4 +9,10 @@
       <Sha>26d9440e6683b44c4db5f45b25e9c857a6563bb8</Sha>
     </Dependency>
   </ProductDependencies>
+  <ToolsetDependencies>
+    <Dependency Name="Microsoft.DotNet.ApiCompat" Version="6.0.0-beta.21179.7">
+      <Uri>https://github.com/dotnet/arcade</Uri>
+      <Sha>fd5f55c64d48b7894516cc841fba1253b2e79ffd</Sha>
+    </Dependency>
+  </ToolsetDependencies>
 </Dependencies>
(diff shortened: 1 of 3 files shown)


jonpryor:
New PR build at: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4614104&view=results

eng/Version.Details.xml
```diff
@@ -9,4 +9,10 @@
       <Sha>26d9440e6683b44c4db5f45b25e9c857a6563bb8</Sha>
     </Dependency>
   </ProductDependencies>
+  <ToolsetDependencies>
```
jonathanpeppers, line 12: What's the difference between `<ToolsetDependencies>` and `<ProductDependencies>`?

jonathanpeppers (approved):

eng/Version.Details.xml
pjcollins: I don't think there's any functional difference, the description from arcade says this:

> ### Toolset vs. Product Dependencies
> There are two types of dependencies: 'Product' and 'Toolset'. Choosing between them involves answering the question:
>
> "Does my repository repackage the input dependency's binaries, or information about those binaries in the outputs it creates for the product?" If so, that dependency is 'Product', otherwise, it's 'Toolset'.

https://github.com/dotnet/arcade/blob/6cc4c1e9e23d5e65e88a8a57216b3d91e9b3d8db/Documentation/Darc.md#toolset-vs-product-dependencies

jonathanpeppers: So things from dotnet/arcade or the dotnet-eng NuGet feed are definitely "toolset", that makes sense. 👍

PR #5802, opened 2021-03-30
[Mono.Android-Tests] Update ignored .NET 6 tests

pjcollins:
A few tests which were previously failing have been fixed and can now
be enabled.  Additionally, now that dotnet/runtime has removed the
incomplete OpenSSL support on Android, we can enable more tests
that used various System.Security.Cryptography APIs directly or
indirectly.

There are still some SSL related test failures that are being ignored
that are generally being tracked in https://github.com/dotnet/runtime/issues/45741.


jonathanpeppers (approved):
It does seem like the enabled tests are running and passing:

image

PR #5805, opened 2021-03-31
[azdo] restore with MSBuild instead of NuGet

jonathanpeppers:
This should allow us to get `.binlog` files for diagnosing when things
go wrong. `nuget restore -Verbosity detailed` is not as detailed as we
would like.

Use `$(RestoreConfigFile)`, which maps to the `-ConfigFile` switch for
NuGet:

https://docs.microsoft.com/nuget/reference/msbuild-targets

We no longer need to run the `NuGetToolInstaller` yaml task either.

### DIFF
--- a/build-tools/automation/yaml-templates/setup-test-environment.yaml
+++ b/build-tools/automation/yaml-templates/setup-test-environment.yaml
@@ -55,22 +55,19 @@ steps:
   displayName: install test dependencies
   condition: and(succeeded(), eq(variables['agent.os'], 'Windows_NT'))
 
-# Restore solutions for Xamarin.Android.Tools.sln, Xamarin.Android.sln, and Xamarin.Android-Tests.sln
-- task: NuGetToolInstaller@0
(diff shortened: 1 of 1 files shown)


dellis1972 (approved):

pjcollins (approved):
My only thought is that maybe we should use dotnet here instead? This probably needs testing but something like this may work:

```yaml
- task: DotNetCoreCLI@2
  inputs:
    command: restore
    projects: ${{ parameters.xaSourcePath }}/Xamarin.Android.Build.Tasks.sln
    restoreArguments: >-
      -c ${{ parameters.configuration }} -p:RestoreConfigFile=${{ parameters.xaSourcePath }}/NuGet.config
      -bl:${{ parameters.xaSourcePath }}/bin/Test${{ parameters.configuration }}/restore-Xamarin.Android.Build.Tasks.binlog
    verbosityRestore: diagnostic
```

I am happy with these changes either way though.

PR #5811, opened 2021-04-01
Revert "[One .NET] $(AndroidEnablePreloadAssemblies)=False by default"

jonathanpeppers:
This reverts commit d13d0f972a4eb663251b2a6b5a934c59a6789db1.

I also updated a test that was triggering an issue with
`$(AndroidEnablePreloadAssemblies)`. It was originally crashing with:

    Java.Lang.RuntimeException: Unable to start activity ComponentInfo{com.xamarin.classlibrarymainlauncherruns/com.xamarin.classlibrarymainlauncherruns.MainActivity}: android.view.InflateException: Binary XML file line #1 in com.xamarin.classlibrarymainlauncherruns:layout/foo: Binary XML file line #1: You must supply a layout_width attribute.
    ---> Android.Views.InflateException: Binary XML file line #1 in com.xamarin.classlibrarymainlauncherruns:layout/foo: Binary XML file line #1: You must supply a layout_width attribute.
    ---> Java.Lang.UnsupportedOperationException: Binary XML file line #1: You must supply a layout_width attribute.
    --- End of managed Java.Lang.UnsupportedOperationException stack trace ---
    java.lang.UnsupportedOperationException: Binary XML file line #1: You must supply a layout_width attribute.

I fixed the invalid XML in `foo.xml`, so the test will get a *better*
crash going forward:

    Process: com.xamarin.classlibrarymainlauncherruns, PID: 2993
    android.runtime.JavaProxyThrowable: System.NullReferenceException: Object reference not set to an instance of an object
        at MyLibrary.MainActivity.OnCreate(Bundle bundle)
        at Android.App.Activity.n_OnCreate_Landroid_os_Bundle_(IntPtr jnienv, IntPtr native__this, IntPtr native_savedInstanceState)
        at com.xamarin.classlibrarymainlauncherruns.MainActivity.n_onCreate(Native Method)
        at com.xamarin.classlibrarymainlauncherruns.MainActivity.onCreate(MainActivity.java:29)
        at android.app.Activity.performCreate(Activity.java:8000)
        at android.app.Activity.performCreate(Activity.java:7984)
        at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1309)
        at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3422)
        at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3601)
        at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:85)
        at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
        at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2066)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:223)
        at android.app.ActivityThread.main(ActivityThread.java:7656)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
        at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)

This way we won't be chasing invalid XML going forward.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -3932,11 +3932,7 @@ namespace UnnamedProject
 				Assert.IsTrue (b.Build (proj), "build should have succeeded.");
 				var environment = b.Output.GetIntermediaryPath (Path.Combine ("__environment__.txt"));
 				FileAssert.Exists (environment);
-				if (Builder.UseDotNet) {
-					Assert.AreEqual ($"__XA_PACKAGE_NAMING_POLICY__={packageNamingPolicy}{Environment.NewLine}mono.enable_assembly_preload=0", File.ReadAllText (environment).Trim ());
-				} else {
-					Assert.AreEqual ($"__XA_PACKAGE_NAMING_POLICY__={packageNamingPolicy}", File.ReadAllText (environment).Trim ());
-				}
+				Assert.AreEqual ($"__XA_PACKAGE_NAMING_POLICY__={packageNamingPolicy}", File.ReadAllText (environment).Trim ());
 			}
 		}
 
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -290,8 +290,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidMakeBundleKeepTemporaryFiles Condition=" '$(AndroidMakeBundleKeepTemporaryFiles)' == '' ">False</AndroidMakeBundleKeepTemporaryFiles>
 
 	<!-- If true it will cause all the assemblies in the apk to be preloaded on startup time -->
-	<_AndroidEnablePreloadAssembliesDefault Condition=" '$(UsingAndroidNETSdk)' == 'true' ">False</_AndroidEnablePreloadAssembliesDefault>
-	<_AndroidEnablePreloadAssembliesDefault Condition=" '$(UsingAndroidNETSdk)' != 'true' ">True</_AndroidEnablePreloadAssembliesDefault>
+        <_AndroidEnablePreloadAssembliesDefault>True</_AndroidEnablePreloadAssembliesDefault>
 	<AndroidEnablePreloadAssemblies Condition=" '$(AndroidEnablePreloadAssemblies)' == '' ">$(_AndroidEnablePreloadAssembliesDefault)</AndroidEnablePreloadAssemblies>
 	<_NativeAssemblySourceDir>$(IntermediateOutputPath)android\</_NativeAssemblySourceDir>
 	<_AndroidUseNewTypemaps>True</_AndroidUseNewTypemaps>
--- a/tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
+++ b/tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
@@ -102,7 +102,13 @@ namespace Xamarin.Android.Build.Tests
 			// Remove the default MainActivity.cs & AndroidResources
 			app.AndroidResources.Clear ();
 			app.AndroidResources.Add (new AndroidItem.AndroidResource ("Resources\\layout\\foo.xml") {
-				TextContent = () => "<?xml version=\"1.0\" encoding=\"utf-8\" ?><LinearLayout xmlns:android=\"http://schemas.android.com/apk/res/android\" />"
+				TextContent = () =>
+@"<?xml version=""1.0"" encoding=""utf-8""?>
+<LinearLayout
+  xmlns:android=""http://schemas.android.com/apk/res/android""
+  android:layout_width=""fill_parent""
+  android:layout_height=""wrap_content""
+/>"
 			});
 			app.Sources.Remove (app.GetItem ("MainActivity.cs"));
 


jonpryor:
https://discord.com/channels/732297728826277939/732297837953679412/827199058279923712

> do we want to spend any time figuring out how that test ever worked (works?!) in the first place?
> feels like we might have a "false positive" scenario here

jonathanpeppers:
@jonpryor yes, I think the actual problem here was quite different. I updated the PR description.

grendello (approved):

jonpryor:
@grendello: don't be too quick to mark this as "good".  Between this PR, d13d0f972a4eb663251b2a6b5a934c59a6789db1, and decfbccf3e, I think this bears additional thought.

For starters, in decfbccf3e, why does `GeneratePackageManagerJava.EnablePreloadAssembliesDefault` accept the `$(_AndroidEnablePreloadAssembliesDefault)` property, and *not* `$(AndroidEnablePreloadAssemblies)`?  Does that actually make sense?

https://github.com/xamarin/xamarin-android/blob/b595e11bf4d4dff99ca09866a79f609645d889b9/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1565

`GeneratePackageManagerJava.EnablePreloadAssembliesDefault` is what actually sets the required bits in `libxamarin-app.so`:

  * https://github.com/xamarin/xamarin-android/blob/b595e11bf4d4dff99ca09866a79f609645d889b9/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs#L163
  * https://github.com/xamarin/xamarin-android/blob/b595e11bf4d4dff99ca09866a79f609645d889b9/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs#L274

Thus, I think the *actual* fix should instead be:

```diff
diff --git a/external/Java.Interop b/external/Java.Interop
--- a/external/Java.Interop
+++ b/external/Java.Interop
@@ -1 +1 @@
-Subproject commit a3de91efdaec04b3927b755a07018e8d26cfdc71
+Subproject commit a3de91efdaec04b3927b755a07018e8d26cfdc71-dirty
diff --git a/external/android-api-docs b/external/android-api-docs
--- a/external/android-api-docs
+++ b/external/android-api-docs
@@ -1 +1 @@
-Subproject commit 6012e04ae54cdc26174dbfe0ca32063193101b32
+Subproject commit 6012e04ae54cdc26174dbfe0ca32063193101b32-dirty
diff --git a/external/apksig b/external/apksig
--- a/external/apksig
+++ b/external/apksig
@@ -1 +1 @@
-Subproject commit 264d99ce7922a8952c6d4f06c90888e9c75b95fc
+Subproject commit 264d99ce7922a8952c6d4f06c90888e9c75b95fc-dirty
diff --git a/external/nrefactory b/external/nrefactory
--- a/external/nrefactory
+++ b/external/nrefactory
@@ -1 +1 @@
-Subproject commit 495405a6ffb5eb6147de754b3a0ed84805e21544
+Subproject commit 495405a6ffb5eb6147de754b3a0ed84805e21544-dirty
diff --git a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Designer.targets b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Designer.targets
index 7eee40439..f47d8e9d9 100644
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Designer.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Designer.targets
@@ -137,7 +137,7 @@ Copyright (C) 2016 Xamarin. All rights reserved.
       IsBundledApplication="$(BundleAssemblies)"
       SupportedAbis="$(_BuildTargetAbis)"
       AndroidPackageName="$(_AndroidPackage)"
-      EnablePreloadAssembliesDefault="$(_AndroidEnablePreloadAssembliesDefault)"
+      EnablePreloadAssembliesDefault="$(AndroidEnablePreloadAssemblies)"
       InstantRunEnabled="$(_InstantRunEnabled)">
     <Output TaskParameter="BuildId" PropertyName="_XamarinBuildId" />
   </GeneratePackageManagerJava>
diff --git a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
index a9fcc31f1..6ed26ed22 100644
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -1562,7 +1562,7 @@ because xbuild doesn't support framework reference assemblies.
     IsBundledApplication="$(BundleAssemblies)"
     SupportedAbis="@(_BuildTargetAbis)"
     AndroidPackageName="$(_AndroidPackage)"
-    EnablePreloadAssembliesDefault="$(_AndroidEnablePreloadAssembliesDefault)"
+    EnablePreloadAssembliesDefault="$(AndroidEnablePreloadAssemblies)"
     PackageNamingPolicy="$(AndroidPackageNamingPolicy)"
     BoundExceptionType="$(AndroidBoundExceptionType)"
     InstantRunEnabled="$(_InstantRunEnabled)"
```

@grendello: does ^^ seem better/more consistent?

jonathanpeppers:
It seems like this is the target that actually turns the setting on or off:
https://github.com/xamarin/xamarin-android/blob/b595e11bf4d4dff99ca09866a79f609645d889b9/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1509-L1524

`<GeneratePackageManagerJava/>` needs to know the default value the runtime expects.

Is that correct @grendello?

grendello:
> It seems like this is the target that actually turns the setting on or off:
> 
> https://github.com/xamarin/xamarin-android/blob/b595e11bf4d4dff99ca09866a79f609645d889b9/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1509-L1524
> 
> `<GeneratePackageManagerJava/>` needs to know the default value the runtime expects.
> 
> Is that correct @grendello?

Yes, that is correct. We need to know both the default **and** the actual setting. That's why we pass `$(_AndroidEnablePreloadAssembliesDefault)` to `GeneratePackageManagerJava` and not `$(AndroidEnablePreloadAssemblies)`. This is to ascertain that the **default** setting is actually the default when runtime starts and `$(AndroidEnablePreloadAssemblies)` is absent or empty

jonpryor:
On a related point, I think there's a bug here: https://github.com/xamarin/xamarin-android/blob/b595e11bf4d4dff99ca09866a79f609645d889b9/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs#L210

shouldn't this line use `val == 1`, not `idx == 1`?   `idx` is the index within `lineToWrite` for where `=` is, which (1) will *never* be 1, and (2) doesn't make sense when setting `usesAssemblyPreload`.

jonpryor:
@jonathanpeppers [wrote](https://github.com/xamarin/xamarin-android/pull/5811#issuecomment-812575967):

> `<GeneratePackageManagerJava/>` needs to know the default value the runtime expects.

I'm not sure I understand this?  As-is, this implies that we have "two sources of knowledge":

 1. The build system, presumably via `$(AndroidEnablePreloadAssemblies)`, and
 2. The runtime, presumably somewhere within `src/monodroid`?

This doesn't make sense to me, particularly given that the `<GeneratePackageManagerJava/>` task *also* uses `ApplicationConfigNativeAssemblyGenerator`, to create the `environment.{abi}.s` files, compiled into into `libxamarin-app.so`.

As far as I know/can easily find, `src/monodroid` *doens't* have "it's own" default value for `uses_assembly_preload`.  The *only* value for `uses_assembly_preload` comes from `libxamarin-app.so`, which in turn comes from `<GenreratePackageManagerJava/>`, which in turn is coming from  -- currently! -- `$(_AndroidEnablePreloadAssembliesDefault)`.

jonpryor:
In a pre-d13d0f972a4eb663251b2a6b5a934c59a6789db1 world, how it looks like it worked was:

1. `$(_AndroidEnablePreloadAssembliesDefault)` was always True.
2. If an app did *not* override `$(AndroidEnablePreloadAssemblies)`, then `<GeneratePackageManagerJava EnablePreloadAssembliesDefault="True" />`, and `libxamarin-app.so` has `uses_assembly_preload=1`.
3. If an app *did* override `$(AndroidEnablePreloadAssemblies)`=False, then we still execute `<GeneratePackageManagerJava EnablePreloadAssembliesDefault="True" />`, and `<GeneratePackageManagerJava/>` *also* parses `__environment__.txt`, which sets `usesAssemblyPreload`=false.  Thus, `libxamarin-app.so` has uses_assembly_preload=0`.

From a documented customer perspective, it worked.  (Yay.)

"However", if you knew the internals, [the other issue](https://github.com/xamarin/xamarin-android/pull/5811#issuecomment-812631902) arises:  What *should* happen if you add an `@(AndroidEnvironment)` file which contains the line:

    mono.enable_assembly_preload=1

The result is that `uses_assembly_preload=0`!  (This is due to using `idx==1` and not `val==1`.)

However, this is internal; we never documented `mono.enable_assembly_preload`, only `$(AndroidEnablePreloadAssemblies)`, so this "weird mismatch" is *confusing*, but not "important".  (This should still be fixed, to reduce confusion.)

jonpryor:
In a post- d13d0f972a4eb663251b2a6b5a934c59a6789db1 world, *with* .NET 6:

 1. `$(_AndroidEnablePreloadAssembliesDefault)` is False by default.
 2. `$(AndroidEnablePreloadAssemblies)` is False by default.
 3. If an app did *not* override `$(AndroidEnablePreloadAssemblies)`, then `<GeneratePackageManagerJava EnablePreloadAssembliesDefault="False" />`, and `libxamarin-app.so` has `uses_assembly_preload=0`.
 4. If an app *did* override `$(AndroidEnablePreloadAssemblies)`=True, then we still execute <GeneratePackageManagerJava EnablePreloadAssembliesDefault="False" />`, and `<GeneratePackageManagerJava/>` also parses `__environment__.txt` -- which *doesn't* set `mono.enable_assembly_preload=1` -- and `libxamarin-app.so` *still* has `uses_assembly_preload=0`.

    This is a (the?) bug: when using .NET 6, `$(AndroidEnablePreloadAssemblies)`=True can't be expressed.

jonpryor:
I still don't understand how `ClassLibraryMainLauncherRuns` enters the picture, but I do see two related issues:

 1. in .NET 6, `$(AndroidEnablePreloadAssemblies)`=True doesn't set `uses_assembly_preload=1`.
 2. If an environment file ever specifies `mono.enable_assembly_preload=1`, this isn't parsed properly.

jonathanpeppers:
`ClassLibraryMainLauncherRuns` is only relevant in that it is failing on main since d13d0f972a4eb663251b2a6b5a934c59a6789db1.

After fixing `foo.xml`, I get the `NullReferenceException` because `MyLibrary.dll` does not seem to be loaded at the time `ResourceManagerId.UpdateIdValues()` runs.

https://github.com/xamarin/xamarin-android/blob/b595e11bf4d4dff99ca09866a79f609645d889b9/src/Mono.Android/Android.Runtime/ResourceIdManager.cs#L9

jonathanpeppers:
Either way, we can fix this after my vacation next week -- https://github.com/xamarin/xamarin-android/commit/d13d0f972a4eb663251b2a6b5a934c59a6789db1 isn't on our .NET 6 Preview 3 release branch.

jonpryor:
@jonathanpeppers [wrote](https://github.com/xamarin/xamarin-android/pull/5811#issuecomment-812722920):

>  I get the `NullReferenceException` because `MyLibrary.dll` does not seem to be loaded at the time `ResourceManagerId.UpdateIdValues()` runs.

Shouldn't 9e6ce03ca2d72f5415a2a1650185ba46113cc3dd fix that, and likewise avoid the call to `ResourceManagerId.UpdateIdValues()`?  Or is the `NullReferenceException` happening in *debug* builds, where 9e6ce03ca2d72f5415a2a1650185ba46113cc3dd doesn't run?

If so, should we consider some form of 9e6ce03ca2d72f5415a2a1650185ba46113cc3dd on *every* build?

jonathanpeppers:
@jonpryor as we discussed on our call today, I changed this to revert https://github.com/xamarin/xamarin-android/commit/d13d0f972a4eb663251b2a6b5a934c59a6789db1, and then also updated 1 test.

PR #5834, opened 2021-04-12
Move `monovm_initalize` property handling to a separate class

grendello:
Putting it in a separate class allows for nicer compile-time checking
for array overflow without the need to pollute code in
`monodroid-glue.cc`.

### DIFF
--- a/src/monodroid/jni/monodroid-glue.cc
+++ b/src/monodroid/jni/monodroid-glue.cc
@@ -82,6 +82,7 @@
 #include "timing.hh"
 #include "xa-internal-api-impl.hh"
 #include "build-info.hh"
+#include "monovm-properties.hh"
 
 #ifndef WINDOWS
 #include "xamarin_getifaddrs.h"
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -216,6 +219,19 @@ void AddEnvironment ()
 						continue;
 					}
 
+					if (lineToWrite.StartsWith ("System.Globalization.Invariant=", StringComparison.Ordinal)) {
```
jonpryor, line 222: These should use the "real" environment variable names: `DOTNET_SYSTEM_GLOBALIZATION_INVARIANT`, `DOTNET_SYSTEM_GLOBALIZATION_USENLS`, `DOTNET_SYSTEM_GLOBALIZATION_APPLOCALICU`:

https://github.com/dotnet/runtime/blob/ce63fb0a4145891760f9530674e93f237dc627c5/src/mono/mono/mini/monovm.c#L185-L194

Rationale: `System.Globalization.Invariant` can't be a Unix environment variable, so possible one-off testing may be hindered (and is inconsistent).

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -216,6 +219,19 @@ void AddEnvironment ()
...
+					}
+					if (lineToWrite.StartsWith ("DOTNET_SYSTEM_GLOBALIZATION_USENLS=", StringComparison.Ordinal)) {
+						haveDotnetSystemGlobalizationUseNls = IsEntryValueBooleanTrue (lineToWrite);
+						continue;
+					}
+					if (lineToWrite.StartsWith ("DOTNET_SYSTEM_GLOBALIZATION_APPLOCALICU=", StringComparison.Ordinal)) {
+						haveDotnetSystemGlobalizationAppLocalIcu = IsEntryValueBooleanTrue (lineToWrite);
+						continue;
+					}
+
```
marek-safar, line 234: Why are you adding this?

grendello: How otherwise these properties can be passed to MonoVM?

marek-safar: They should not be passed in, the code reading them will be removed.

grendello: ok, I'm putting this PR on hold till we know more

grendello:
Context: https://github.com/dotnet/runtime/commit/7fa839a727e510f2b1b104eafb2c1fb339249756

PR #5835, opened 2021-04-12
[One .NET] fix $(PackageTargetFallback) and $(TargetPlatformVersion)

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5819

A project using Maui and `TargetFramework=net6.0-android` can hit the
warning:

    warning CA1416: This call site is reachable on all platforms. 'Class1' is only supported on: 'android' 30.0 and later.

I found the project had:

    TargetPlatformVersion = 30

Projects that emit no warnings have `TargetPlatformVersion = 30.0`.

Update the `<GenerateSupportedPlatforms/>` MSBuild task to always
append `.0`:

    version.ApiLevel.ToString ("0.0", CultureInfo.InvariantCulture)

This solves the warning, but we still get an error in other projects
in the solution:

    error CS1061: 'Class1<TVirtualView, TNativeView>' does not contain a definition for 'Context' and no accessible extension method 'Context' accepting a first argument of type 'Class1<TVirtualView, TNativeView>' could be found (are you missing a using directive or an assembly reference?)

The build used:

    ~\.nuget\packages\microsoft.maui.core\6.0.100-preview.3.269\lib\netstandard2.1\Microsoft.Maui.dll

Instead of:

    ~\.nuget\packages\microsoft.maui.core\6.0.100-preview.3.269\lib\net6.0-android30.0\Microsoft.Maui.dll

I found this project uses `TargetFramework=net6.0-android30.0`, which
resulted in a weird value for `$(PackageTargetFallback)`:

    PackageTargetFallback =
        net6.0-android30.0.0;
        monoandroid12.0;
        monoandroid11.0;
        monoandroid10.0;
        monoandroid90;
        monoandroid81;
        monoandroid80;
        monoandroid70;
        monoandroid60;
        monoandroid50;

To fix this, I conditionally added the `.0` and the project builds
successfully.

We will hopefully be able to remove usage of
`$(PackageTargetFallback)` soon, as it was a workaround until NuGet
has official support for .NET 6 & MonoAndroid target frameworks.

I added a new test that covers these cases. I also added a new
`AssertHasNoWarnings()` extension method to simplify checking for
warnings. I also updated a couple tests that no longer emit warnings.

### DIFF
--- a/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
+++ b/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateSupportedPlatforms.cs
@@ -1,4 +1,5 @@
 using System.IO;
+using System.Globalization;
 using System.Linq;
 using System.Xml;
 using Microsoft.Build.Framework;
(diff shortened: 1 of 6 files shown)


dellis1972 (approved):

jonathanpeppers:
`ClassLibraryMainLauncherRuns` should be fixed in 522d7fb61f3669d85d077ba4f2889dbe8c9c8ac3, so I think this is green.

dellis1972 (approved):

Issue #5701, opened 2021-03-05
Investigate and fix MSBuild smoke test issues on Linux


pjcollins:
Context: https://github.com/xamarin/xamarin-android/pull/5681#issuecomment-790017891

We should extend the test coverage of our Linux build and smoke test job.

jonpryor:
Fixed by? https://github.com/xamarin/xamarin-android/commit/1b71da0cd3a44861b006ac41c77a24eedf066281

pjcollins:
Commit 1b71da0cd3a44861b006ac41c77a24eedf066281 moved Linux builds to a new 1ES Managed machine pool, but didn't change build/test scope at all. Extending test coverage for Linux should still be looked at eventually if we want to promote the build artifacts being created there.

pjcollins:
The smoke test issues mentioned in #5681 are fixed by #6585, and I will plan to separate out Linux tests into a separate job.

PR #5873, opened 2021-04-26
[build] Copy templates to lowercase path on Linux

pjcollins:
Context: https://github.com/dotnet/sdk/issues/16946

Copy our template workload pack to a lowercase destination so that it
can be detected and used when building and testing on Linux.

### DIFF
--- a/build-tools/create-packs/Directory.Build.targets
+++ b/build-tools/create-packs/Directory.Build.targets
@@ -128,7 +128,7 @@
     <MakeDir Directories="$(DotNetPreviewPath)template-packs" />
     <!-- Copy template packs to lowercase destination when running on Linux -->
     <Copy Condition="'$(HostOS)' != 'Linux'" SourceFiles="@(_WLTemplates)" DestinationFolder="$(DotNetPreviewPath)template-packs" />
-    <Copy Condition="'$(HostOS)' == 'Linux'" SourceFiles="@(_WLTemplates)" DestinationFiles="@(_WLTemplates->'%(Filename)'->ToLower()->'$(DotNetPreviewPath)template-packs\%(Identity)')" />
+    <Copy Condition="'$(HostOS)' == 'Linux'" SourceFiles="@(_WLTemplates)" DestinationFiles="@(_WLTemplates->'%(Filename)%(Extension)'->ToLower()->'$(DotNetPreviewPath)template-packs\%(Identity)')" />
     <ItemGroup>
       <_UnixExecutables Include="$(DotNetPreviewPath)packs\Microsoft.Android.Sdk.*\*\tools\$(HostOS)\**\*.*" />
       <_FilesToTouch Include="$(DotNetPreviewPath)sdk-manifests\$(DotNetPreviewVersionBand)\Microsoft.NET.Workload.Android\**" />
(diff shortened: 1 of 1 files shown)


grendello (approved):

build-tools/create-packs/Directory.Build.targets
```diff
@@ -126,7 +126,9 @@
         DestinationFolder="$(DotNetPreviewPath)packs\$([System.String]::Copy('%(_WLPacks.Filename)').Replace('.$(_WLPackVersion)', ''))\$(_WLPackVersion)"
     />
     <MakeDir Directories="$(DotNetPreviewPath)template-packs" />
-    <Copy SourceFiles="@(_WLTemplates)" DestinationFolder="$(DotNetPreviewPath)template-packs" />
+    <!-- Copy template packs to lowercase destination when running on Linux -->
+    <Copy Condition="'$(HostOS)' != 'Linux'" SourceFiles="@(_WLTemplates)" DestinationFolder="$(DotNetPreviewPath)template-packs" />
+    <Copy Condition="'$(HostOS)' == 'Linux'" SourceFiles="@(_WLTemplates)" DestinationFiles="@(_WLTemplates->'%(Filename)'->ToLower()->'$(DotNetPreviewPath)template-packs\%(Identity)')" />
```
jonathanpeppers, line 131: I think we could always do the lower case, even the .NET 5 ones are lower:

image

pjcollins: Sounds good, we can always copy to a lowercase location. I am also hoping to wait for a response on the issue or to chat about this quickly on Thursday to see if we should change the template pack ID / version (here and in macios) instead. If there won't be SDK changes coming that will process the packs regardless of casing then we should probably make that change instead.

jonpryor: Could someone explain that `DestinationFiles` expression to me?  I've never seen *multiple* `->` constructs before.  Does the rightmost `%(Identity)` apply to the "intermediate" list resulting from `@(_WLTemplates->'%(Filename)'->ToLower())`, or is it `%(_WLTemplates.Identity)`?

pjcollins: The `%(Identity)` in this case will be the product of the transform plus the item function, so the value of
  `@(_WLTemplates->'%(Filename)'->ToLower())` best summarizes it.

WonyoungChoi, line 131: It seems that the ".nupkg" extension is missing.
```suggestion
    <Copy Condition="'$(HostOS)' == 'Linux'" SourceFiles="@(_WLTemplates)" DestinationFiles="@(_WLTemplates->'%(Filename)%(Extension)'->ToLower()->'$(DotNetPreviewPath)template-packs\%(Identity)')" />
```

pjcollins: Yeah we do need to revisit test scope on Linux in the future, I think we still have a handful of failures to work through in even the smoke test category (#5701)

pjcollins:
Assuming this works everywhere I think we can merge it temporarily, or we can wait for a potential resolution to https://github.com/dotnet/sdk/issues/16946.

jonathanpeppers (approved):

Issue #5855, opened 2021-04-20
[NET6] libxamarin-debug-app-helper.so should not be packaged for Release builds


grendello:
`libxamarin-debug-app-helper.so` is a helper DSO used by Debug builds only, however `dotnet build -c Release` for NET6 places it in the apk:

```shell
net6-samples/HelloMaui/bin/Release/net6.0-android $ zipinfo com.microsoft.hellomaui-Signed.apk | grep helper
-rwxr-xr-x  6.3 unx    37104 b- defX 21-Apr-20 16:45 lib/arm64-v8a/libxamarin-debug-app-helper.so
-rwxr-xr-x  6.3 unx    31796 b- defX 21-Apr-20 16:45 lib/x86/libxamarin-debug-app-helper.so
```

The DSO comes from the `microsoft.android.runtime.android-{ARCH}` nuget packages which contains all the native libraries from Xamarin.Android, for both Release and Debug builds.

grendello:
It appears this issue also applies to `Java.Interop.dll` which is placed in Debug build packages, while it should only be present in the Release ones.

dellis1972:
My guess is `libxamarin-debug-app-helper.so` should not even be in the runtime pack. Its probably just being included because its the default behaviour of the new build system to include all native libs from the pack. 

My understanding about Java.Interop was the problem is its a DEBUG build of the assembly in the pack not a RELEASE one? As a result we are hitting asserts that we don't normally hit. That does raise the question on how we are getting a DEBUG release of that dll in the runtime pack...

grendello:
> My guess is `libxamarin-debug-app-helper.so` should not even be in the runtime pack. Its probably just being included because its the default behaviour of the new build system to include all native libs from the pack.
> 

If the same runtime pack is used to build debug apps, we need to have the DSO in there?

> My understanding about Java.Interop was the problem is its a DEBUG build of the assembly in the pack not a RELEASE one? As a result we are hitting asserts that we don't normally hit. That does raise the question on how we are getting a DEBUG release of that dll in the runtime pack...

Yeah, it's supposed to be used only in Debug builds, so I guess there's no non-debug build of it

jonathanpeppers:
I think we can make this task exclude `libxamarin-debug-app-helper.so` based on `$(AndroidIncludeDebugSymbols)`:

[...]

PR #5874, opened 2021-04-26
[One .NET] exclude libxamarin-debug-app-helper.so from Release builds

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5855

Currently, in .NET 6, any `.so` file in our runtime pack is included
in apps by default. This is the nature of how native libraries coming
from NuGet packages work. They just get added based on the file path
and `$(RuntimeIdentifier)` of the project.

We already have a special case for `libmono-android.release.so` and
`libmono-android.debug.so` that uses `$(AndroidIncludeDebugSymbols)`
as a way to know which native library is needed.

We can do the same thing here and exclude
`libxamarin-debug-app-helper.so`, which should only be included in
`Debug` builds.

This saves ~37K for all Release apps:

    -37,104 lib/arm64-v8a/libxamarin-debug-app-helper.so *1

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ProcessNativeLibraries.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ProcessNativeLibraries.cs
@@ -57,6 +57,10 @@ namespace Xamarin.Android.Tasks
 							continue;
 						library.SetMetadata ("ArchiveFileName", "libmonodroid.so");
 					}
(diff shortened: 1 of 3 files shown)


grendello (approved):

jonathanpeppers:
The one failure seems unrelated:
```
error XA0010: No available device.
```

PR #5884, opened 2021-04-28
[Xamarin.Android.Build.Tasks] Build should update designtime/Resource.designer.cs

jonathanpeppers:
When testing .NET 6 in the latest Internal Preview builds of Visual
Studio, we are now in a place where the Android designer is working
and rendering layouts.

Unfortunately if you add a new `android:id="@+id/foo"`, Intellisense
does not update to show that `Resource.Id.foo` is available.

After enabling build logging with `Project System Tools`:

https://marketplace.visualstudio.com/items?itemName=VisualStudioProductTeam.ProjectSystemTools

I found the following is happening:

* I hit Ctrl+S to save the `.xml` file
* `UpdateGeneratedFiles` runs, which updates
  `obj/Debug/*/designtime/Resource.designer.cs`. This runs in a
  design-time context where we use the `ManagedResourceParser`, it
  uses an `R.txt` if found on disk. In this case, it is an old `R.txt`
  from a previous build.
* `UpdateAndroidResources` runs, which updates `R.txt` and
  `obj/Debug/*/Resource.designer.cs`. It runs in a context where we
  run `aapt2` instead of the `ManagedResourceParser`.

Unfortunately, at this point the `designtime/Resource.designer.cs` is
out of date and does not have the new `android:id` available.

This problem seems to happen in "legacy" Xamarin.Android as well.
However, doing a regular build appears to update Intellisense. I think
developers are currently doing a regular build here, which is not
ideal? While in .NET 6, doing a `Build` does *not* update the
Intellisense. The only way I could get .NET 6 to update was to force a
design-time build by changing a random setting in `Project Options`.

To solve the issue, we can make both `UpdateAndroidResources` and
regular `Build` update the `designtime/Resource.designer.cs` file.
These files should stay in sync if we have the proper values from
`aapt2` anyway. I think this will solve the problem in .NET 6 as well
as improve the experience in "legacy" Xamarin.Android.

Down the road, I think we should rethink how design-time builds work,
in general here.

I updated some tests around this scenario to be sure the
`designtime/Resource.designer.cs` get updated with the same contents.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesigner.cs
@@ -18,6 +18,8 @@ namespace Xamarin.Android.Tasks
 		[Required]
 		public string NetResgenOutputFile { get; set; }
 
+		public string DesignTimeOutputFile { get; set; }
+
 		public string JavaResgenInputFile { get; set; }
 
 		public string RTxtFile { get; set; }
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
I want to manually test this in .NET 6 and legacy, when installers are produced.

jonathanpeppers:
After manually testing with VS 16.10 Internal Preview, I think this improves the behavior in "legacy" Xamarin.Android as well as .NET 6:

1. Add/remove/update an `android:id`
2. Ctrl+S
3. Alt+Tab to C# and and I have updated intellisense for the `Resource.Id` class.

I think "legacy" Xamarin.Android was only working before, because a `Build` would solve the issue. The legacy project system seems like it uses results from `Build` for Intellisense, but the new project system fully relies on design-time build.

dellis1972 (approved):
This looks good.

PR #5885, opened 2021-04-28
[One .NET] Make apk size checks more lenient

pjcollins:
Context: https://github.com/radekdoulik/apkdiff/pull/2

New percentage changed based checks have been added to apkdiff to allow
for greater flexibility when determining whether or not a size change
is problematic.  The .NET 6 .apk content and size checks have been
updated to use percentage based comparisons, and as a result have been
made more lenient.  This should reduce the amount of times we need to
manually update .apkdesc files.  These checks can be tightened as we get
closer to a stable release.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -279,9 +279,7 @@ stages:
         testResultsFiles: TestResult-*.xml
         testRunTitle: Java Interop Tests - Windows Build Tree
 
-    - script: dotnet tool update apkdiff --version 0.0.13 --add-source https://api.nuget.org/v3/index.json -g
-      displayName: install apkdiff dotnet tool
-      continueOnError: true
+    - template: yaml-templates\install-apkdiff.yaml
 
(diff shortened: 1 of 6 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -279,7 +279,7 @@ stages:
         testResultsFiles: TestResult-*.xml
         testRunTitle: Java Interop Tests - Windows Build Tree
 
-    - script: dotnet tool update apkdiff -g
+    - script: dotnet tool update apkdiff --version 0.0.13 --add-source https://api.nuget.org/v3/index.json -g
       displayName: install apkdiff dotnet tool
       continueOnError: true
```
jonathanpeppers, line 284: Should we make a `install-apkdiff.yaml` template?

Then we'd have the version in one place, too.

pjcollins: Good call, done.

jonathanpeppers (approved):
This looks good now assuming the build is green 👍

PR #5888, opened 2021-04-29
[aapt2] Use the aapt2 from `build-tools_r$(XABuildToolsVersion)-{os}.zip`

dellis1972:
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1288832
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1287930

Commit c9e2d754 had to revert the version of `aapt2` we were shipping
due to issues with older PC's. Specifically a certain type of AMD
processor the AMD Phenom II. That commit simply reverted back to the last
known working version.

This made us think about if getting `aapt2` from maven was the best
idea. We are not 100% sure if the builds on maven even match those
being shipped in the google sdk. So the idea now is to use the one that
google ships with its sdk. However since we will control which version
of the build tools we want to ship we will not be relying on the one
from the users SDK folder. This will make sure we always ship a known
working version of `aapt2` with our product.

The `XABuildToolsVersion` value will be sourced from [Configuration.props](https://github.com/xamarin/xamarin-android/blob/main/Configuration.props#L120).
the `os` value will be `windows`, `linux` or `macosx`.
At the time of this commit the `XABuildToolsVersion` value was 30.0.2.

### DIFF
--- a/Configuration.props
+++ b/Configuration.props
@@ -115,8 +115,11 @@
       For some reason, the URL for platform-tools/build-tools 30.0.2 is prefixed with what appears to be a GIT commit hash or some other checksum...
       Linux packages don't have any prefix, but this forces us to have *some* mechanism to handle this...
     -->
-    <XABuildToolsPackagePrefix Condition=" '$(HostOS)' == 'Darwin' ">5a6ceea22103d8dec989aefcef309949c0c42f1d.</XABuildToolsPackagePrefix>
-    <XABuildToolsPackagePrefix Condition=" '$(HostOS)' == 'Windows' ">efbaa277338195608aa4e3dbd43927e97f60218c.</XABuildToolsPackagePrefix>
+    <XABuildToolsPackagePrefixMacOS>5a6ceea22103d8dec989aefcef309949c0c42f1d.</XABuildToolsPackagePrefixMacOS>
+    <XABuildToolsPackagePrefixWindows>efbaa277338195608aa4e3dbd43927e97f60218c.</XABuildToolsPackagePrefixWindows>
+    <XABuildToolsPackagePrefixLinux></XABuildToolsPackagePrefixLinux>
+    <XABuildToolsPackagePrefix Condition=" '$(HostOS)' == 'Darwin' ">$(XABuildToolsPackagePrefixMacOS)</XABuildToolsPackagePrefix>
+    <XABuildToolsPackagePrefix Condition=" '$(HostOS)' == 'Windows' ">$(XABuildToolsPackagePrefixWindows)</XABuildToolsPackagePrefix>
     <XABuildToolsVersion>30.0.2</XABuildToolsVersion>
     <XABuildToolsFolder Condition="'$(XABuildToolsFolder)' == ''">30.0.2</XABuildToolsFolder>
     <XAPlatformToolsPackagePrefix Condition=" '$(HostOS)' == 'Darwin' ">b2be9c80582174e645d3736daa0d44d8610b38a8.</XAPlatformToolsPackagePrefix>
--- a/Documentation/release-notes/5888.md
+++ b/Documentation/release-notes/5888.md
@@ -1,0 +1,46 @@
+#### Application build and deployment
+
+- [GitHub PR #5888](https://github.com/xamarin/xamarin-android/pull/5888):
+  [aapt2] Use the aapt2 from build-tools_r$(XABuildToolsVersion)-{os}.zip.
+  Due to issues with the `aapt2` builds from maven, we are now sourcing
+  the version of `aapt2` we ship from the SDK build-tools on our build machines.
+
+  If you are experiencing endless builds or `aapt2` crashes during a build you
+  can try using `aapt2` from maven.
+
+  1. Windows users, download: https://dl.google.com/dl/android/maven2/com/android/tools/build/aapt2/4.0.0-6051327/aapt2-4.0.0-6051327-windows.jar
+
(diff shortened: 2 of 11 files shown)


grendello (approved):

jonpryor:

```markdown
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1288832
Context: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1287930

Commit c9e2d754 had to revert the version of `aapt2` we were shipping
due to issues with older PC's, specifically due to running `aapt2` on
the AMD Phenom II.  Commit c9e2d754 simply reverted back to the last
known working version.

This made us reconsider whether obtaining `aapt2` from maven was the
best idea: we are not 100% sure if the builds on maven even match
those being shipped in the Google Android SDK Build-tools package.

The new idea is to instead use the one that Google ships with the
Android SDK, in the Build-tools package.  However, since we will
control which version of the build tools we want to ship we will
not be relying on the one from the users SDK folder.  This will ensure
we always ship a known working version of `aapt2` with our product.

If `aapt2` continues to fail on certain customer machines, then we
can direct them to download the `aapt2` used in c9e2d754 and override
the `$(Aapt2ToolExe)` and `$(Aapt2ToolPath)` MSBuild properties.
```

@dellis1972 : *Question*: What is (was?) the URL used to obtain the previous `aapt2` binaries?

@dellis1972: Please add a `Documentation/release-notes` blurb for this change, and as part of the "blurb", provide the above URL(s) and required `.csproj` changes in order to use the older `aapt2` binaries.

dellis1972:
The url is in the diff https://github.com/xamarin/xamarin-android/pull/5888/files#diff-faecf4026a4a80638e99539d9c1bac1e2d99bb35806505aac68fa58e735cc7a7L31

`https://dl.google.com/dl/android/maven2/com/android/tools/build/aapt2/$(Aapt2Version)/aapt2-$(Aapt2Version)-%(_Aapt2Download.Platform).jar`

I have added the release notes (keep forgetting about those).

jonpryor:
@dellis1972  [asked](https://github.com/xamarin/xamarin-android/pull/5888#issuecomment-839826843)

> I'm not sure what you are asking for here.

Assume you're on "older" hardware and need to use the older `aapt2`?  You need to downgrade.  How do you do so?

We should write that down, 'cause we'll likely need to write it down *anyway*.

Thus:

 1. Windows users, download: https://dl.google.com/dl/android/maven2/com/android/tools/build/aapt2/4.0.0-6051327/aapt2-4.0.0-6051327-windows.jar
 
    macOS users, download: https://dl.google.com/dl/android/maven2/com/android/tools/build/aapt2/4.0.0-6051327/aapt2-4.0.0-6051327-osx.jar

2. Extract the `aapt2-*.jar` file into a "well-known" directory; this location will be used in step (3):
  * Windows:

        cd %USERPROFILE%
        mkdir xa-tools
        cd xa-tools
        jar xf %USERPROFILE%\Downloads\aapt2-4.0.0-6051327-windows.jar

  * macOS:

        cd
        mkdir xa-tools
        cd xa-tools
        jar xf ~/Downloads/aapt2-4.0.0-6051327-osx.jar

3. In the same directory as your App's `.csproj`, add or update `Directory.Build.props` to contain:

    ```xml
    <Project>

      <PropertyGroup Condition=" '$(OS)' == 'Windows_NT' ">
        <Aapt2ToolPath>$(USERPROFILE)\xa-tools</Aapt2ToolPath>
        <Aapt2ToolExe>aapt2.exe</Aapt2ToolExe>
      </PropertyGroup>

      <PropertyGroup Condition=" '$(OS)' != 'Windows_NT' ">
        <Aapt2ToolPath>$(HOME)\xa-tools</Aapt2ToolPath>
        <Aapt2ToolExe>aapt2</Aapt2ToolExe>
      </PropertyGroup>

    </Project>
    ```

dellis1972:
@jonpryor I'm not sure downgrading to the one off maven is the right thing. The users who hit this problem found that using the aapt2 from the sdk fixed the issue. This is one of the reasons we decided to source the version we ship from the sdk. 

I will add instructions in the release notes on how to use a custom aapt2 path, but I don't think we should be recommending maven builds to anyone as they seem to be unreliable.

jonpryor:
@dellis1972  wrote:

> I'm not sure downgrading to the one off maven is the right thing. The users who hit this problem found that using the aapt2 from the sdk fixed the issue.

This is in part why we need to discuss which concrete instructions, if any, to provide.  My thinking was that whatever `aapt2` we shipped in d16-8 is "fine", in that nobody complained until d16-9, and thus instructions to install the `aapt2` from d16-8 should be "fine", even if that download is from maven.  It's what we previously shipped, right?

Closely related is your blurb in the release notes:

> build-tools_r$(XABuildToolsVersion)-{os}.zip

What is a customer supposed to do with that?  `{XABuildToolsVersion}` and `{os}` are "greek" and possibly unknowable without additional investigations (where would they look?) or questions (read: new issues).

This is in part why [my suggested steps](https://github.com/xamarin/xamarin-android/pull/5888#issuecomment-840667716) included full URLs, no templating, specific to developer OS: to make the instructions easier to understand, with less "thought" required.

Thus:

1. Do we want to be fully explicit about URLs to download, above and beyond the "template" currently provided?
2. If (1) is Yes, which URLs should we provide?  The maven URLs used in d16-8, which *should* work?  A URL to the Android SDK packages, which we've never fully tested prior to this PR?  Something else?

PR #5890, opened 2021-04-29
First Pass on Dynamic Asset Features

dellis1972:
Context #4810 

This is the first implementation for supporting building Dynamic Feature Assets modules for Android. 
This idea is to have these "features" as standard android library projects. The `ProjectReference` will 
need to use the following

```xml
<ItemGroup>
    <ProjectReference Include="Features\AssetsFeature\AssetsFeature.csproj">
         <ReferenceOutputAssembly>false</ReferenceOutputAssembly>
         <AndroidDynamicFeature>true</AndroidDynamicFeature>
    </ProjectReference>
  </ItemGroup>
```

We set `ReferenceOutputAssembly` to `false` to ensure that the assembly for the feature is NOT included
in the final aab file.  The new meta Data `AndroidDynamicFeature` allows the build system to pick up 
project references which are "features".

As part of the final packaging step of the main app we will gather up all the `ProjectRefernce` items which have
`AndroidDynamicFeature` set to `true` (and maybe `ReferenceOutputAssembly` set to `false`). 
This will be done by the `_BuildDynamicFeatures` which will run just after `_CreateBaseApk`. 

It will call `_GetDynamicFeatureOutputs` for each `ProjectReference` which will collect the `output` files
for each feature. It will then call the `BuildDynamicFeature` target via the `MSBuild` task for each
`ProjectReference`.  The `BuildDynamicFeature` is the target responsible for collecting all the assets and 
packaging them using `aapt2` up into a zip. Once all the `BuildDynamicFeature` calls are complete the created
`zip` files will be added to the `AndroidAppBundleModules` and then included in the final `aab` file. 

It might seem odd that the feature projects are built after the main app. However this is required because the 
feature needs to use the `packaged_resources` file as an input to `aapt2` when building the feature. This is 
why the `_BuildDynamicFeatures` happens AFTER `_CreateBaseApk`. It is only at that point that the final 
`packaged_resources` file exists. 

One of the very weird things is that the feature zip needs to be built using the `aapt2` `--static-lib` flag. As a result
we need to call `aapt2 convert` on the final zip. This is because it is in the `apk` `binary` format and needs to be converted over to the `aab` `proto` format. So there is a new `Aapt2Convert` task which handles that job. It also makes sure the 
`AndroidManifest.xml` file is in the right place when converting to `proto` format.

A basic project example using .net 6 for a feature would look like this.

```xml
  <PropertyGroup>
  </PropertyGroup>
</Project>
``` 

As you can see it is just a normal library project. At this time is CANNOT contain any `AndroidResource` items such as drawables or layouts. It must only contain `AndroidAsset` items. So we probably should have a new template for a `Dynamic Feature` which just creates the `csproj` and the `Assets` folder. 

One sticking point is probably the `AndroidManifest.xml` file which we need for a `feature`. There is a sample

```xml
        xmlns:tools="http://schemas.android.com/tools"
        xmlns:dist="http://schemas.android.com/apk/distribution"
        android:versionCode="1"
        android:versionName="1.0"
        package="com.companyname.DynamicAssetsExample"
        featureSplit="assetsfeature"
        android:isFeatureSplit="true">
  <dist:module dist:title="@string/assetsfeature" dist:instant="false">
    <dist:delivery>
      <dist:on-demand />
    </dist:delivery>
    <dist:fusing dist:include="false" />
  </dist:module>
  <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="30" />
  <application android:hasCode="false" tools:replace="android:hasCode" />
</manifest>
```
The interesting parts are all the additional `dist` elements. What we can probably do is auto generate this 
during the `BuildDynamicFeature`. However we need to think carefully about this since if we plan to have 
code and `Activities` in the feature at some point, those will also need to end up in the `AndroidManifest.xml`.

### DIFF
--- a/Documentation/guides/DynamicFeatures.md
+++ b/Documentation/guides/DynamicFeatures.md
@@ -1,0 +1,461 @@
+# Android Dynamic Feature Delivery
+
+_NOTE: This document is very likely to change, as the the implementation
+of Dynamic Features matures._
+
+Android Dynamic Features are a way of splitting up your application
+into smaller parts which users can download on-demand. This can be
+useful in scenarios such as Games or apps with low usage parts.
+An example would be a Support module which does not get used very
+often, rather than including this code in the main app it could be
+a dynamic feature which is downloaded when the user requests support.
+Alternatively think of Game levels in a game. On first install only
+the first few levels are installed to save download size. But later
+the additional levels can be downloaded while the user is not playing
+the game.
+
+Dynamic Features can ONLY be used when targeting the Google Play
+Store and using `PackageFormat` set to `aab`. Using `apk` will
+NOT work.
+
+## Xamarin Dynamic Asset/Feature Delivery
+
+The Xamarin implementation of this will start by focusing on asset
+delivery. This is where the `AndroidAsset` items can be split out
+into a separate module for downloading later.
+
+_NOTE: In the future we plan to support feature or code delivery
+however there are significant challenges to supporting that in our
+runtime. So this will be done after the initial rollout of asset
+delivery._
+
+The new Dynamic Feature Module will be based around a normal
+Xamarin Android class library. The kind you get when calling
+
+```
+dotnet new androidlib
+```
+
+Although we plan to provide a template. The new Module will generally
+be nothing different from a normal class library project. The
+current restrictions are that it only contains `AndroidAsset` items.
+No `AndroidResource` items can be included and any C# code in the
+library will NOT end up in the application.
+
+The one additional file is a special `AndroidManifest.xml` file which
+will be generated as part of the module build process. This manifest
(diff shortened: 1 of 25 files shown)


dellis1972:
Depends on https://github.com/xamarin/xamarin-android/pull/5926

dellis1972:
I think we are going to need a specific template for this project type. The Reason being is that if you include any `AndroidResource` files we get an odd build error from bundletool. I haven't figured out how to make resources in features work yet. There must be some special aapt2 magic involved to get it working.

dellis1972:
I'm not sure we should use `ProjectReference` now. It makes things a bit tricky.

dellis1972:
Closing this PR as https://github.com/dotnet/android/pull/8631 was merged.

PR #5898, opened 2021-04-30
[One .NET] rename workload to Microsoft.NET.Sdk.Android.Manifest-6.0.100

jonathanpeppers:
Context: https://github.com/dotnet/designs/pull/188/files#diff-8fcaa29d8e6f00b34b3cb1830d93f33e75f04424780a66a3c658c7021048e74fR125

The `$(PackageId)` of our workload `.nupkg` needs to be:

    Microsoft.NET.Sdk.Android.Manifest-6.0.100

While the `$(PackageVersion)` remains the same as before.

The layout on disk will change to:

    dotnet\sdk-manifests\6.0.100\Microsoft.NET.Sdk.Android\
        WorkloadManifest.json
        WorkloadManifest.targets

Note that `.Manifest` and `-6.0.100` are not in the folder name on disk.

At the same time, let's also update the `version` in
`WorkloadManifest.json` so it contains the proper version for our
workload. This used to not be possible because `version` was a `long`,
but it now is a `string` where we can put our version.


jonathanpeppers:
@jonpryor on the weird filename:

    Successfully created package 'C:\src\xamarin-android\bin\BuildDebug\nuget-unsigned\Microsoft.NET.Sdk.Android.Manifest-6.0.100.11.0.200-ci.microsoft-net-sdk-android-manifest.255.nupkg'.

We were discussing how weird `6.0.100.11.0.200` looks. It appears that NuGet basically does `$(PackageId).$(PackageVersion)`, and we were wanting to put `-` as a delimiter instead of `.`.

It appears there isn't an MSBuild property for the filename for NuGet:

image

The only option that seems even close is `$(OutputFileNamesWithoutVersion)`, but I don't think we should remove the version from the file name.

Relevant source code:

https://github.com/NuGet/NuGet.Client/blob/c529bd75d247430ed19b1713041de0b5b240ef0d/src/NuGet.Core/NuGet.Build.Tasks.Pack/NuGet.Build.Tasks.Pack.targets#L220-L281

I think for now, we go with this as-is? The file name is weird, but no one should be using this file directly.

jonathanpeppers:
Filed an issue about the NuGet file name here: https://github.com/NuGet/Home/issues/10834

PR #5909, opened 2021-05-04
[One .NET] remove EnableWorkloadResolver.sentinel

jonathanpeppers:
This feature flag is now on by default in .NET 6.0.100-preview.4.21253.14 and higher.

We should stop generating `EnableWorkloadResolver.sentinel`, as it strongly ties our installers to a specific version of the .NET SDK.


jonathanpeppers:
I think we actually need a newer preview 5 build for this:
```
```

PR #2153, opened 2018-09-06
[Xamarin.Android.Build.Tasks] Generate JNI marshal methods

radekdoulik:
Implementation of https://github.com/xamarin/xamarin-android/issues/2138

Also fixes https://github.com/xamarin/java.interop/issues/387,
https://github.com/xamarin/java.interop/issues/388

Conditionally generate JNI marshal methods for user and relevant SDK
assemblies. It is enabled by setting
`AndroidGenerateJniMarshalMethods` to *True*.

This is done by `_GenerateJniMarshalMethods` target, which runs
jnimarshalmethod-gen.exe with XA's mono. It is possible to pass
additional arguments to the tool by setting
`AndroidGenerateJniMarshalMethodsAdditionalArguments` property.

The target also passes the path to the JVM shared library to the tool
using `--jvm` option. The path comes from the updated `ResolveSdks`
task. The task will be optimized to improve performance by caching the
path.

The commit also removes the `NotImplementedException` throws from the
`Android.Runtime.AndroidValueManager`. The reason for that is that
this is now used from the generated marshal methods, which handle
primitive arrays. Namely `AddPeer (IJavaPeerable value)` is used. That
is something to further examine.

The `RunJavaInteropTests` was moved from `_RunParallelTestTarget` item
group to `_RunTestTarget` group, because it overwrites the
`Java.Runtime.Environment.dll.config`, which is required by
`jnimarshalmethod-gen.exe` to run. And so we need to run these tests after
apk tests. That should be fixed in future, hopefully by adding and/or
fixing the Inputs/Outputs of the relevant targets.

Enable JNI marshal methods in Mono.Android runtime test and
Xamarin.Forms test in Release and Release/AOT configurations.

Add custom `JniCustomValueMarshaler`'s for `Android.Graphics.Color`
and for `IJavaObject` based objects, which don't implement
`IJavaPeerable`.

The generated `Color` class marshaler example:

    using Android.Graphics;
    using Java.Interop;
    using System;
    
    public static int n_getColor (IntPtr __jnienv, IntPtr __this)
    {
    	JniTransition jniTransition = new JniTransition (__jnienv);
    	JniRuntime runtime = default(JniRuntime);
    	try {
    		runtime = JniEnvironment.Runtime;
    		JniRuntime.JniValueManager valueManager = runtime.ValueManager;
    		valueManager.WaitForGCBridgeProcessing ();
    		MyPaint value = valueManager.GetValue<MyPaint> (__this);
    		Color color = value.Color;
[...]

PR #5919, opened 2021-05-06
[Xamarin.Android.Build.Tasks] Prefer supported JDK versions

jonpryor:
Context: https://devdiv.visualstudio.com/DevDiv/_releaseProgress?releaseId=1029208&_a=release-environment-extension&environmentId=5321844&extensionId=ms.vss-test-web.test-result-in-release-environment-editor-tab&runId=21424746&resultId=100001&paneView=attachments

Context: https://github.com/xamarin/xamarin-android-tools/commit/237642cee7f95e37e87b26231b47c74796166594
Context: https://github.com/xamarin/xamarin-android-tools/commit/dca30d9d3aae6c2a209349ea94d6b42443e0d906
Context: https://github.com/xamarin/xamarin-android-tools/commit/d92fc3e3a27e8240551baa813b15d6bf006a5620
Context: https://github.com/xamarin/xamarin-android-tools/commit/2241489b31a985bfb6249b63ab93eb3746891aac

xamarin/xamarin-android-tools was updated to check for JDK locations
in a variety of locations, checking for these locations (1) *after*
a "preferred" JDK location, but (2) *before* `$JAVA_HOME` and the
Visual Studio-installed "microsoft_dist_openjdk_1.8.0.25" package.
Consequently, when running on an environment in which multiple JDKs
are installed, *and* an *unsupported version* of a JDK is present,
it's possible that build failures would result, as Xamarin.Android
attempted to use an unsupported version.

This happened during one of our internal CI runs:

	Project templates should have zero errors
	Expected: <empty>
	But was:  < "
	[Line]: 248
	[Description]: Building with JDK version 14.0.2 is not supported. Please install JDK version 11.0. See https://aka.ms/xamarin/jdk9-errors (XA0030)
	[File]: Xamarin.Android.Legacy.targets

This particular machine had AdoptOpenJDK 14 installed, in addition
to other probed-for JDK locations, but because it was found first --
`JdkInfo.GetKnownSystemJdkInfos()` doesn't know or care about
Xamarin.Android JDK supported versions, and we're trying to deprecate
the "microsoft_dist_openjdk*" package, and thus it shouldn't be
preferred *anyway*… -- the Xamarin.Android build attempted to use
JDK 14, when JDK 14 isn't supported.

Fix this scenario by improving the `<ResolveJdkJvmPath/>` task to
use `$(MinimumSupportedJavaVersion)` and
`$(LatestSupportedJavaVersion)` to constrain which JDK we use, using
the first "known system JDK" which is greater than or equal to
`$(MinimumSupportedJavaVersion)` and less than or equal to
`$(LatestSupportedJavaVersion)`.

The `$(JavaSdkDirectory)` MSBuild property is still preferred, if set,
followed by the "preferred JDK location", if set.  It's only if
neither of these locations is set that we'll check for all known
system JDKs, at which point we now filter them.


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Tooling.targets
```diff
@@ -82,6 +82,8 @@ projects.
     </ResolveSdks>
     <ResolveJdkJvmPath
         JavaSdkPath="$(_JavaSdkDirectory)"
+        MinimumSupportedJavaVersion="$(MinimumSupportedJavaVersion)"
+        LatestSupportedJavaVersion="$(LatestSupportedJavaVersion)"
         Condition=" '$(DesignTimeBuild)' != 'True' And '$(_AndroidIsBindingProject)' != 'True' And '$(AndroidGenerateJniMarshalMethods)' == 'True' And '$(JdkJvmPath)' == '' ">
```
jonathanpeppers, line 87: How was a project template even hitting this? `$(AndroidGenerateJniMarshalMethods)` would not be `True`, right?

jonpryor: :thinking:

Maybe this also needs to be in VS as well?

jonpryor: We can't find any place within VS that would need fixing.

However, I *did* find a codepath wherein `<ResolveSdks/>` calls `AndroidSdkInfo`, which eventually calls `JdkInfo.GetKnownSystemJdkInfos()` if/when the JDK path provided to the `AndroidSdkInfo` constructor isn't valid, and *this* is likely responsible for the integration test failure originally reported.

jonpryor:
Thanks to @GouriKumari for bringing this scenario to our attention.

src/Xamarin.Android.Build.Tasks/Tasks/ResolveSdksTask.cs
```diff
@@ -81,6 +87,11 @@ public override bool RunTask ()
 			MonoAndroidLibPath  = MonoAndroidHelper.GetOSLibPath () + Path.DirectorySeparatorChar;
 			AndroidBinUtilsPath = MonoAndroidBinPath + "ndk" + Path.DirectorySeparatorChar;
 
+			var minVersion      = Version.Parse (MinimumSupportedJavaVersion);
+			var maxVersion      = Version.Parse (LatestSupportedJavaVersion);
+
+			JavaSdkPath         = MonoAndroidHelper.GetJdkInfo (this.CreateTaskLogger (), JavaSdkPath, minVersion, maxVersion)?.HomePath;
```
jonathanpeppers, line 93: This task gets called per Android project on every build. So I could see this adding incremental build time, especially in a larger solution. I think I've seen the `JdkInfo` lookup take 100ms before.

Is there a way we could pass the min/max JDK version through here instead:

https://github.com/xamarin/xamarin-android/blob/57e1a37c443c90f3487a93a312fbfef3441d2769/src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs#L116

If we want something easier to backport, maybe hardcoding something in xamarin-android-tools would work, too?

jonathanpeppers: Here is where I saw JDK lookups taking time: https://github.com/xamarin/xamarin-android/commit/e624434344ed54f39ed174d4c3a46a6cb06a7302

jonpryor: 1. Should we *merge* `<ResolveJdkJvmPath/>` and `<ResolveSdks/>`, "reverting" *and also* adding the `BuildEngine4.RegisterTaskObject()` invocations to `<ResolveSdks/>`?
2. I saw e624434 in passing, and I *wonder* what exactly was being measured?  (Aside: *why* does `JdkInfo` take 47ms to construct?!  Is it the `Lazy<T>` construction, used to parse `-XshowSettings:properties -version` output?)

Not that I'm saying you didn't see that improvement.  I'm wondering about the *environment* of the measurement.  In particular, was `msbuild /p:JavaSdkDirectory=…` used to capture that timing?

If `$(JavaSdkDirectory)` *isn't* set -- which I assume is the default and more common occurrence -- then we'll almost certainly "always" hit https://github.com/xamarin/xamarin-android-tools/blob/683f37508b56c76c24b3287a5687743438625341/src/Xamarin.Android.Tools.AndroidSdk/Sdks/AndroidSdkBase.cs#L69

```csharp
JavaSdkPath    = GetValidPath (ValidateJavaSdkLocation,     javaSdkPath,    () => PreferedJavaSdkPath,      () => GetJavaSdkPaths ());
```

`javaSdkPath` will be null, necessitating that we first try `() => PreferedJavaSdkPath`.

For most customer machines, this should return a valid value.  This (presumably) is a net-win, and would be a reason to look into updating `AndroidSdkInfo` to have version constraints.

However, if `() => PreferedJavaSdkPath` fails -- which presumably is the case on this CI machine -- then we hit the `() => GetJavaSdkPaths()` codepath, which is likely why we're finding and attempting to use an unsupported JDK version.

---

Which I suppose leaves two questions:

 1. Would `BuildEngine4.RegisterTaskObject()` within `<ResolveSdks/>` fix the problem?
 2. Do we know what about `JdkJvmPath` lookup is slow?  The directory traversal?

In retrospect, the "odd" thing about the pre-e624434 world order is that `ResolveSdks.GetJvmPath()` even existed in the first place, as constructed; once we have a JDK path, why would `GetJvmPath()` need to call `JdkInfo.GetKnownSystemJdkInfos()` again?

Looks like that was always the case, though:

https://github.com/xamarin/xamarin-android/commit/1a2eb953af76b76b2443b744894523ddfe239ecf#diff-ec49e6de47ef6cb5b59095d242a56acce5830d39402bf075722fce7dd1326612R105

Also odd that I kinda/sorta mentioned the "oddness" of having a `JdkInfo.GetKnownSystemJdkInfos()` invocation, then forgot about how odd that was: https://github.com/xamarin/xamarin-android/pull/2153#discussion_r215961078

I don't see why, given a valid JDK path, `JdkJvmPath` lookup should be *so slow*; via `csharp`:

```csharp
using Xamarin.Android.Tools;
using System.Diagnostics;
using System.IO
var validJdkPath=Path.Combine(Environment.GetEnvironmentVariable("HOME"), "Library/Developer/Xamarin/jdk/microsoft_dist_openjdk_1.8.0.25");
var s = Stopwatch.StartNew(); var p = new JdkInfo(validJdkPath).JdkJvmPath; s.Stop();
s.ElapsedMilliseconds
// 26 on first invocation, 0 on subsequent iterations.  Behold the JIT!
```

So maybe `JdkJvmPath` *is* slow, *because* of JIT overheads (ugh), and no amount of caching can fix things when we're concerned about design-time builds on the *first* build.  It doesn't matter if subsequent builds are faster (yay JIT!), when we care about the Design-Time build, which won't need `JdkJvmPath`…

But what about the JDK itself?

```csharp
using Xamarin.Android.Tools;
using System.Diagnostics;
using System.IO
var s = Stopwatch.StartNew(); var p = new AndroidSdkInfo().JavaSdkPath; s.Stop();
s.ElapsedMilliseconds;
print(p);
```

This is "differently fun"; if I edit `$HOME/.config/xbuild/monodroid-config.xml` and *remove* the `<java-sdk/>` element -- removing the "preferred JDK" location -- then the above takes 81ms to run, while if I have a `<java-sdk/>` element -- providing a "preferred JDK" location -- then the above takes 55ms to run, a savings of 26ms when the preferred JDK location is set and exists.

jonathanpeppers: When I was testing https://github.com/xamarin/xamarin-android/commit/e624434344ed54f39ed174d4c3a46a6cb06a7302, I probably didn't have `JavaSdkDirectory` set at all and was testing from the command-line.

Let me just test this change tomorrow on Windows/.NET framework with a Release build and see how long it takes.

Were your numbers above a `Release` build?

jonathanpeppers (approved):
I just tried a `Release` build of Xamarin.Android on Windows, and the changes here don't seem to be any worse with default settings:
```
> .\bin\Release\bin\xabuild .\samples\HelloWorld\HelloWorld.csproj -bl
...
Task ResolveSdks 20ms
...
Took: 00:00:00.0033017
...
Task ResolveSdks 3ms
...
Took: 00:00:00.0008511
```
So it would only be slower in the case this new `catch` block runs, which should be rare?

My timing was:
```diff
+                       Stopwatch w = new Stopwatch();
+                       w.Start();
                        var minVersion      = Version.Parse (MinimumSupportedJavaVersion);
                        var maxVersion      = Version.Parse (LatestSupportedJavaVersion);

                        JavaSdkPath         = MonoAndroidHelper.GetJdkInfo (this.CreateTaskLogger (), JavaSdkPath, minVersion, maxVersion)?.HomePath;
+                       w.Stop();
+                       Log.LogDebugMessage("Took: " + w.Elapsed);
```

jonathanpeppers:
Oh, maybe these tests fail now, though:

image

jonathanpeppers:
I think different ones are failing now (or I didn't notice them before):

image

jonpryor:
@jonathanpeppers [commented](https://github.com/xamarin/xamarin-android/pull/5919#issuecomment-834791510):

> I think different ones are failing now (or I didn't notice them before):

This was because `ResolveSdksTaskTests.cs` directly used `<ResolveSdks/>`, and wasn't setting the newly added (and now required) `MinimumSupportedJavaVersion` and `LatestSupportedJavaVersion` properties.

The unit test has been updated.

dellis1972 (approved):

PR #5920, opened 2021-05-06
Bump to cmdline-tools 4.0, fix lint errors

jonathanpeppers:
Context: https://dl-ssl.google.com/android/repository/repository2-1.xml

We were seeing some lint-related MSBuild tests fail with:

    Lint
        Parameters
            ToolPath = /Users/runner/Library/Android/sdk/cmdline-tools/latest/bin
        ...
        LintVersion: 4.2.0
        Environment Variables passed to tool:
        JAVA_HOME=/Users/runner/Library/Android/jdk-11/
        ...
        Errors
            obj/Debug/android/AndroidManifest.xml(12,89): error XA0103:  MainActivity must extend android.app.Activity [Instantiatable] [/Users/runner/work/1/s/bin/TestRelease/temp/CheckLintConfigMerging/UnnamedProject.csproj]
            obj/Debug/android/AndroidManifest.xml(18,28): error XA0103:  MonoRuntimeProvider must extend android.content.ContentProvider [Instantiatable] [/Users/runner/work/1/s/bin/TestRelease/temp/CheckLintConfigMerging/UnnamedProject.csproj]
        2 errors, 3 warnings
        The command exited with code -1.

We suspect the Hosted Agent pool was updated to install
`cmdline-tools` 4.0 in ~/Library/Android/sdk/cmdline-tools/latest and
cause this problem:

https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md#android

After updating locally to `cmdline-tools` 4.0 I can reproduce the issue.

I also updated the `<Lint/>` MSBuild task to ignore the
`Instantiatable` error for now.

After this change, emulators failed to start:

    Task CreateAndroidEmulator
    ...
    build-tools/scripts/TestApks.targets(52,5): Config file for AVD 'XamarinAndroidTestRunner29-x86_64' not found at /Users/runner/.android/avd/XamarinAndroidTestRunner29-x86_64.avd/config.ini [/Users/runner/work/1/s/tests/Mono.Android-Tests/Mono.Android-Tests.csproj]
    build-tools/scripts/TestApks.targets(52,5): AVD 'XamarinAndroidTestRunner29-x86_64' will use default emulator settings (memory and data partition size) [/Users/runner/work/1/s/tests/Mono.Android-Tests/Mono.Android-Tests.csproj]
    ...
    Task StartAndroidEmulator
    ...
    build-tools/scripts/TestApks.targets(65,5): Emulator failed to start: emulator: ERROR: Unknown AVD name [XamarinAndroidTestRunner29-x86_64], use -list-avds to see valid list. [/Users/runner/work/1/s/tests/Mono.Android-Tests/Mono.Android-Tests.csproj]

It appears the code in the `<CreateAndroidEmulator/>` MSBuild task
that attempts to pass `\n` to stdin is no longer working?

`avdmanager` prompts with:

    Do you wish to create a custom hardware profile? [no]

I found that if you pass `--device pixel_4`, for example, then
`avdmanager` does not prompt at all. I think we should select a
default device, so we don't have to work with stdin at all.

I converted `<CreateAndroidEmulator/>` to a regular `ToolTask` and
passed `--device`. The command appears to work properly now.

### DIFF
--- a/Configuration.props
+++ b/Configuration.props
@@ -159,8 +159,8 @@
     <AndroidToolPath Condition=" '$(AndroidToolPath)' == '' ">$(AndroidSdkFullPath)\tools</AndroidToolPath>
     <AndroidToolsBinPath Condition=" '$(AndroidToolsBinPath)' == '' ">$(AndroidToolPath)\bin</AndroidToolsBinPath>
     <AndroidToolExe Condition=" '$(AndroidToolExe)' == '' ">android</AndroidToolExe>
-    <CommandLineToolsFolder Condition=" '$(CommandLineToolsFolder)' == '' ">1.0</CommandLineToolsFolder>
-    <CommandLineToolsVersion Condition=" '$(CommandLineToolsVersion)' == '' ">6200805_latest</CommandLineToolsVersion>
+    <CommandLineToolsFolder Condition=" '$(CommandLineToolsFolder)' == '' ">4.0</CommandLineToolsFolder>
+    <CommandLineToolsVersion Condition=" '$(CommandLineToolsVersion)' == '' ">7302050_latest</CommandLineToolsVersion>
     <CommandLineToolsBinPath Condition=" '$(CommandLineToolsBinPath)' == '' ">$(AndroidSdkFullPath)\cmdline-tools\$(CommandLineToolsFolder)\bin</CommandLineToolsBinPath>
     <EmulatorVersion Condition=" '$(EmulatorVersion)' == '' ">7033400</EmulatorVersion>
     <EmulatorPkgRevision Condition=" '$(EmulatorPkgRevision)' == '' ">30.3.5</EmulatorPkgRevision>
--- a/Configuration.props
+++ b/Configuration.props
@@ -155,6 +155,7 @@
   <PropertyGroup>
     <AdbToolPath Condition=" '$(AdbToolPath)' == '' ">$(AndroidSdkFullPath)\platform-tools\</AdbToolPath>
     <AdbToolExe Condition=" '$(AdbToolExe)' == '' ">adb</AdbToolExe>
+    <AvdManagerHome Condition=" '$(AvdManagerHome)' == '' ">$(AndroidToolchainDirectory)</AvdManagerHome>
     <AvdManagerToolExe Condition=" '$(AvdManagerToolExe)' == '' ">avdmanager</AvdManagerToolExe>
     <AndroidToolPath Condition=" '$(AndroidToolPath)' == '' ">$(AndroidSdkFullPath)\tools</AndroidToolPath>
     <AndroidToolsBinPath Condition=" '$(AndroidToolsBinPath)' == '' ">$(AndroidToolPath)\bin</AndroidToolsBinPath>
--- a/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CreateAndroidEmulator.cs
+++ b/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CreateAndroidEmulator.cs
@@ -13,7 +13,11 @@ namespace Xamarin.Android.Tools.BootstrapTasks
 	{
 		public                  string          SdkVersion      {get; set;}
 		public                  string          AndroidAbi      {get; set;}
-		public                  string          AndroidSdkHome  {get; set;}
+		/// <summary>
+		/// Specifies $ANDROID_SDK_HOME. This is not the path to the Android SDK, but a root folder that contains the `.android` folder.
+		/// </summary>
+		[Required]
+		public                  string          AvdManagerHome  {get; set;}
 		public                  string          JavaSdkHome     {get; set;}
 
 		public                  string          TargetId        {get; set;}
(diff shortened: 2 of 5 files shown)


dellis1972 (approved):

jonpryor:
What's somewhat "odd" here is…our CI should be installing our own test dependencies:

https://github.com/xamarin/xamarin-android/blob/4ed6a222ed07551734b28525dd269d50ac107ea8/build-tools/automation/yaml-templates/setup-test-environment.yaml#L48

Which includes Android SDK & NDK:

https://github.com/xamarin/xamarin-android/blob/4ed6a222ed07551734b28525dd269d50ac107ea8/build-tools/xaprepare/xaprepare/Scenarios/Scenario_AndroidTestDependencies.cs#L17

which in turn includes cmdline-tools (seemingly?):

https://github.com/xamarin/xamarin-android/blob/4ed6a222ed07551734b28525dd269d50ac107ea8/build-tools/xaprepare/xaprepare/Steps/Step_Android_SDK_NDK.cs#L110

Given all that, why is CI trying to use the *system* cmdline-tools package at all?

```
_AndroidToolsDirectory=/Users/runner/Library/Android/sdk/cmdline-tools/latest
```

jonathanpeppers:
@pjcollins can comment, but I think what's happening here is we use `~/Library/Android/sdk/` to speed up our build. The build machine's image has an Android SDK there already. That way we aren't installing API level 30, for example, on every build, because it is already there.

The only probably that would arise would be something like `cmdline-tools` where a newer one appears, and we start using it.

jonathanpeppers:
I think there is something else broken here, it makes our CI unable to start emulators:
```
error : Emulator failed to start: emulator: ERROR: Unknown AVD name [XamarinAndroidTestRunner29-x86_64], use -list-avds to see valid list.
```

Configuration.props
```diff
@@ -159,8 +159,8 @@
     <AndroidToolPath Condition=" '$(AndroidToolPath)' == '' ">$(AndroidSdkFullPath)\tools</AndroidToolPath>
     <AndroidToolsBinPath Condition=" '$(AndroidToolsBinPath)' == '' ">$(AndroidToolPath)\bin</AndroidToolsBinPath>
     <AndroidToolExe Condition=" '$(AndroidToolExe)' == '' ">android</AndroidToolExe>
-    <CommandLineToolsFolder Condition=" '$(CommandLineToolsFolder)' == '' ">1.0</CommandLineToolsFolder>
-    <CommandLineToolsVersion Condition=" '$(CommandLineToolsVersion)' == '' ">6200805_latest</CommandLineToolsVersion>
+    <CommandLineToolsFolder Condition=" '$(CommandLineToolsFolder)' == '' ">4.0</CommandLineToolsFolder>
```
pjcollins, line 162: One piece I noticed recently was that we have a mismatch here and in the version the auto sdk tooling uses: https://github.com/xamarin/xamarin-android-tools/blob/26d65d95943b18bb9575d75c8c60ce36180a0db9/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.Versions.props#L13.  We can probably update xaprepare to read the folder version from xamarin-android-tools instead, but I think that can be done separately if we want.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CreateAndroidEmulator.cs
```diff
@@ -65,127 +85,5 @@ string GetAndroidPath ()
...
-
-			using (p) {
-				p.StartInfo = psi;
-				p.Start ();
-				p.BeginErrorReadLine ();
-				stdin = p.StandardInput;
-
-				// Relying on HasExited is racy, but we have no choice here since we need to tell
-				// avdmanager that we want to proceed in creation of the AVD by answering its question
-				// on whether to create a custom hardware profile with "Enter"...
```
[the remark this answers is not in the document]
jonathanpeppers: I found out how to make it stop asking the question:

> I found that if you pass `--device pixel_4`, for example, then
`avdmanager` does not prompt at all. I think we should select a
default device, so we don't have to work with stdin at all.

dellis1972 (approved):

pjcollins (approved):

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CreateAndroidEmulator.cs
```diff
@@ -1,55 +1,75 @@
...
+			} else {
+				ulong diskSize;
+				if (!UInt64.TryParse (DataPartitionSizeMB, out diskSize))
+					Log.LogError ($"Invalid data partition size '{DataPartitionSizeMB}' - must be a positive integer value expressing size in megabytes");
+
+				ulong ramSize;
+				if (!UInt64.TryParse (RamSizeMB, out ramSize))
+					Log.LogError ($"Invalid RAM size '{RamSizeMB}' - must be a positive integer value expressing size in megabytes");
+
+				File.AppendAllLines (configPath, new string [] {
```
jonpryor, line 51: What keeps `configPath` from having these values appended to it every time this task runs?   How does `avdmanager` behave if/when `configPath` contains the "same" values appended to the end of the file?

jonathanpeppers: Hmm, yeah it looks like it would always append... I guess it always was like this?

I will look into fixing this tomorrow, can probably just read every line of the file and only append if there isn't a match.

jonathanpeppers: I reworked the task, so it only writes the file when needed.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CreateAndroidEmulator.cs
```diff
@@ -1,190 +1,155 @@
...
+			if (Log.HasLoggedErrors)
+				return false;
 
-			ulong diskSize;
-			if (!UInt64.TryParse (DataPartitionSizeMB, out diskSize))
-				Log.LogError ($"Invalid data partition size '{DataPartitionSizeMB}' - must be a positive integer value expressing size in megabytes");
+			var configPaths = new [] {
+				Path.Combine (Environment.GetFolderPath (Environment.SpecialFolder.Personal), ".android", "avd", $"{ImageName}.avd", "config.ini"),
+				Path.Combine (AndroidSdkHome, ".android", "avd", $"{ImageName}.avd", "config.ini")
+			};
```
jonathanpeppers, line 53: I found the first path worked on macOS, but on Windows the file is in `%USERPROFILE%\android-toolchain\sdk\.android\`. Not sure why the path is different?

PR #5934, opened 2021-05-17
[monodroid] A handful of reliability + cosmetic changes

grendello:
Drop all use of the `assert(3)` C macro in favor of macros which aren't
turned off when the code is build with `NDEBUG`.  This is especially
important when macros test runtime conditions that depend on third party
input.

Stop using C numerical type limit constants in favor of the standard C++
`std::numeric_class<>` API

Replace the remaining instances of `UNUSED_ARG` with the standard C++
attribute `[[maybe_unused]]`

Disable `mkbundle` related code in NET6


src/monodroid/jni/cpp-util.hh
```diff
@@ -1,7 +1,41 @@
...
+		return;
+	}
+
+	va_list ap;
+
+	va_start (ap, fmt);
+#if defined (ANDROID)
+	__android_log_vprint (ANDROID_LOG_FATAL, "monodroid", fmt, ap);
+#else
+	vfprintf (stderr, fmt, ap);
```
jonpryor, line 29: Shouldn't this also call `fprintf (stderr, "\n")`?  Very often these messages will *not* have a trailing `\n`, and stderr is *line buffered* by default (IIRC), so for sanity don't we want to emit a final `\n`?

src/monodroid/jni/jni-wrappers.hh
```diff
@@ -19,21 +20,21 @@ namespace xamarin::android
 			: env (_env),
 			  jstr (nullptr)
 		{
-			assert (_env);
+			 abort_if_invalid_pointer_argument (_env);
```
jonpryor, line 23: Leading space?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -848,10 +829,15 @@ MonoDomain*
 MonodroidRuntime::create_domain (JNIEnv *env, jstring_array_wrapper &runtimeApks, bool is_root_domain)
 {
 	size_t user_assemblies_count   = 0;
+#if defined (NET6)
+	constexpr bool have_mono_mkbundle_init = false;
+#else
```
jonpryor, line 834: Missing `// !defined (NET6)`

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -848,10 +829,15 @@ MonoDomain*
...
+	bool have_mono_mkbundle_init = mono_mkbundle_init != nullptr;
+#endif
```
jonpryor, line 836: Missing `// !defined (NET6)`

jonpryor (changes requested):

PR #5942, opened 2021-05-19
[build] add dotnet/runtime dependency through dotnet/installer

jonathanpeppers:
Context: https://github.com/dotnet/installer/blob/f442964a34125c76240b1bf2d2c53fdb00d53e41/eng/Version.Details.xml#L20-L23

We can start tracking dotnet/runtime as a "coherent dependency" of
dotnet/installer. This is a bit simpler than manually parsing:

    dotnet\sdk\*\Microsoft.NETCoreSdk.BundledVersions.props

The benefit of doing this, is when maestro comes along to bump
dotnet/installer, we'll get a diff for dotnet/runtime as well.

We can also start using the `$(MicrosoftNETCoreAppRefPackageVersion)`
MSBuild property throughout our build as needed. We'll need to do this
if we want to provision `Microsoft.NET.Workload.Mono.ToolChain` to a
specific version.

An example if we updated things *now*:

    > darc update-dependencies --channel ".NET 6"
    Looking up latest build of https://github.com/dotnet/installer on .NET 6
    Looking up latest build of https://github.com/dotnet/arcade on .NET 6
    Updating 'Microsoft.Dotnet.Sdk.Internal': '6.0.100-preview.5.21266.3' => '6.0.100-preview.6.21269.6' (from build '20210519.6' of 'https://github.com/dotnet/installer')
    Checking for coherency updates...
    Using 'Strict' coherency mode. If this fails, a second attempt utilizing 'Legacy' Coherency mode will be made.
    Updating 'Microsoft.NET.ILLink.Tasks': '6.0.100-preview.5.21264.1' => '6.0.100-preview.5.21268.2' to ensure coherency with Microsoft.Dotnet.Sdk.Internal@6.0.100-preview.6.21269.6
    Updating 'Microsoft.NETCore.App.Ref': '6.0.0-preview.5.21265.5' => '6.0.0-preview.6.21269.1' to ensure coherency with Microsoft.Dotnet.Sdk.Internal@6.0.100-preview.6.21269.6
    Local dependencies updated from channel '.NET 6'.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -5,6 +5,7 @@
     <MicrosoftNETILLinkTasksPackageVersion>6.0.100-preview.5.21264.1</MicrosoftNETILLinkTasksPackageVersion>
     <MicrosoftDotNetApiCompatPackageVersion>5.0.0-beta.20181.7</MicrosoftDotNetApiCompatPackageVersion>
     <MicrosoftDotNetBuildTasksFeedPackageVersion>6.0.0-beta.21212.6</MicrosoftDotNetBuildTasksFeedPackageVersion>
+    <MicrosoftNETCoreAppRefPackageVersion>6.0.0-preview.5.21265.5</MicrosoftNETCoreAppRefPackageVersion>
   </PropertyGroup>
   <PropertyGroup>
     <!-- Trim all characters after first `-` or `+` is encountered. -->
--- a/build-tools/xaprepare/xaprepare/Application/Context.cs
+++ b/build-tools/xaprepare/xaprepare/Application/Context.cs
@@ -359,27 +359,6 @@ namespace Xamarin.Android.Prepare
 			}
 		}
 
-		string _bundledPreviewRuntimePackVersion;
-		public string BundledPreviewRuntimePackVersion {
-			get {
-				if (string.IsNullOrEmpty (_bundledPreviewRuntimePackVersion)) {
-					var dotnetPath = Properties.GetRequiredValue (KnownProperties.DotNetPreviewPath);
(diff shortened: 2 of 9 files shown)


grendello (approved):

eng/Versions.props
```diff
@@ -5,6 +5,7 @@
     <MicrosoftNETILLinkTasksPackageVersion>6.0.100-preview.5.21264.1</MicrosoftNETILLinkTasksPackageVersion>
     <MicrosoftDotNetApiCompatPackageVersion>5.0.0-beta.20181.7</MicrosoftDotNetApiCompatPackageVersion>
     <MicrosoftDotNetBuildTasksFeedPackageVersion>6.0.0-beta.21212.6</MicrosoftDotNetBuildTasksFeedPackageVersion>
+    <MicrosoftNETCoreAppRefPackageVersion>6.0.0-preview.5.21265.5</MicrosoftNETCoreAppRefPackageVersion>
```
jonathanpeppers, line 8: This should be the same dotnet/runtime version as main:
```
> cat ~\android-toolchain\dotnet\sdk\6.0.100-preview.5.21266.3\Microsoft.NETCoreSdk.BundledVersions.props | grep BundledNETCoreAppPackageVersion

    <BundledNETCoreAppPackageVersion>6.0.0-preview.5.21265.5</BundledNETCoreAppPackageVersion>
```

pjcollins (approved):

jonathanpeppers:
The only test failure of concern is `DotNetDebug`, but this test is passing locally for me.

Merging this, so we'll have a proper Maestro bump on Monday. I'll revisit if that test continues to cause an issue.

Issue #5944, opened 2021-05-20
One .NET: Remove BinaryFormatters support


jonpryor:
`BinaryFormatter` has been deprecated, and is no longer part of the mobile profile in .NET 6; see also: https://github.com/dotnet/designs/blob/8bff552b2bf5d9a36d39db58d106e2b2d34a510a/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md

As of https://discordapp.com/channels/732297728826277939/732297837953679412/844981772445024306 this change has caused our `JsonDeserializationCreatesJavaHandle` unit test to start failing: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4770105&view=ms.vss-test-web.build-test-results-tab&runId=21773130&resultId=100082&paneView=debug

```
Output did contain TestJsonDeserializationCreatesJavaHandle!
Expected: False
But was:  True
```

`logcat-failed.txt` contains:

```
--------- beginning of crash
05-18 03:12:13.988 10832 10832 E AndroidRuntime: FATAL EXCEPTION: main
05-18 03:12:13.988 10832 10832 E AndroidRuntime: Process: com.xamarin.jsondeserializationcreatesjavahandle, PID: 10832
05-18 03:12:13.988 10832 10832 E AndroidRuntime: android.runtime.JavaProxyThrowable: System.PlatformNotSupportedException: BinaryFormatter serialization and deserialization are disabled within this application. See https://aka.ms/binaryformatter for more information.
05-18 03:12:13.988 10832 10832 E AndroidRuntime:    at System.Runtime.Serialization.Formatters.Binary.BinaryFormatter.Serialize(Stream serializationStream, Object graph) in System.Runtime.Serialization.Formatters.dll:token 0x600013c+0x5
05-18 03:12:13.988 10832 10832 E AndroidRuntime:    at UnnamedProject.MainActivity.TestBinaryDeserialization(Person p) in UnnamedProject.dll:token 0x6000003+0xd
05-18 03:12:13.988 10832 10832 E AndroidRuntime:    at UnnamedProject.MainActivity.TestJsonDeserializationCreatesJavaHandle() in UnnamedProject.dll:token 0x6000004+0x1d
05-18 03:12:13.988 10832 10832 E AndroidRuntime:    at UnnamedProject.MainActivity.OnCreate(Bundle bundle) in UnnamedProject.dll:token 0x6000001+0x4b
[...]

PR #5947, opened 2021-05-21
[ci] Fix smoke and emulator test issues

pjcollins:
Fixes: https://github.com/xamarin/xamarin-android/issues/5944
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4791430&view=logs&jobId=9e58dffc-b78a-5028-45f4-4d99f4c2da6c

The TestBinaryDeserialization test has been disabled for .NET 6 as it
is no longer supported due to security reasons.

The "Run check-boot-times" test has seemingly been failing since commit
92efdccb6, which introduced a new version of avdmanager and changed the
path where our AVD files are stored.  The corresponding task and tool
have been fixed to know about this new AVD home location.  The create
and start Android emulator tasks have also been updated for uniformity.
All of these tasks now set [`$(ANDROID_PREFS_ROOT)`][0] rather than the
now deprecated `$(ANDROID_SDK_HOME)` variable.

I have also seen some intermittent test failures on various PR builds in
some of our InstallAndRun emulator tests, and I have bumped the timeout
for those tests.  The RunWithInterpreterEnabled test has been reworked
a bit to ensure that the required logging properties are set.

The nuget.config file that was added to the samples folder in commit
4201dc18 has been removed, as it cause problems with release builds.  A
new vscode task has been added to generate the config file as needed
for local testing instead.

[0]: https://developer.android.com/studio/command-line/variables.html

### DIFF
--- a/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CreateAndroidEmulator.cs
+++ b/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CreateAndroidEmulator.cs
@@ -14,7 +14,7 @@ namespace Xamarin.Android.Tools.BootstrapTasks
 		public                  string          SdkVersion      {get; set;}
 		public                  string          AndroidAbi      {get; set;}
 		/// <summary>
-		/// Specifies $ANDROID_SDK_HOME. This is not the path to the Android SDK, but a root folder that contains the `.android` folder.
+		/// Specifies $ANDROID_PREFS_ROOT. This is not the path to the Android SDK, but a root folder that contains the `.android` folder.
 		/// </summary>
 		[Required]
 		public                  string          AvdManagerHome  {get; set;}
@@ -47,7 +47,7 @@ namespace Xamarin.Android.Tools.BootstrapTasks
 			}
 
 			var env = new List<string> ();
-			env.Add ($"ANDROID_SDK_HOME={AvdManagerHome}");
+			env.Add ($"ANDROID_PREFS_ROOT={AvdManagerHome}");
 			if (!string.IsNullOrEmpty (JavaSdkHome)) {
 				env.Add ($"JAVA_HOME={JavaSdkHome}");
 			}
--- a/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/RunAndroidEmulatorCheckBootTimes.cs
+++ b/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/RunAndroidEmulatorCheckBootTimes.cs
@@ -2,6 +2,7 @@ using System;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text;
 using System.Threading;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
@@ -12,12 +13,16 @@ namespace Xamarin.Android.Tools.BootstrapTasks
 {
 	public class RunAndroidEmulatorCheckBootTimes : Task
 	{
+		public string AndroidSdkDirectory { get; set; }
+		public string AvdManagerHome { get; set; }
 		public string CheckBootTimesPath { get; set; }
 		public string DeviceName { get; set; } = "XamarinAndroidTestRunner64";
 
 		public override bool Execute ()
 		{
 			Log.LogMessage (MessageImportance.High, $"Task {nameof (RunAndroidEmulatorCheckBootTimes)}");
+			Log.LogMessage (MessageImportance.High, $"     {nameof (AndroidSdkDirectory)}: {AndroidSdkDirectory}");
+			Log.LogMessage (MessageImportance.High, $"     {nameof (AvdManagerHome)}: {AvdManagerHome}");
 			Log.LogMessage (MessageImportance.High, $"     {nameof (CheckBootTimesPath)}: {CheckBootTimesPath}");
 			Log.LogMessage (MessageImportance.High, $"     {nameof (DeviceName)}: {DeviceName}");
 
@@ -32,9 +37,16 @@ namespace Xamarin.Android.Tools.BootstrapTasks
 			var timeoutInMS = (int) TimeSpan.FromMinutes (15).TotalMilliseconds;
 			bool finishAsExpected = false;
 			bool processTimedOut = false;
+			var args = new StringBuilder ($"--devicename {DeviceName} --verbose");
+			if (!string.IsNullOrWhiteSpace (AndroidSdkDirectory)) {
+				args.Append ($" --sdkpath \"{AndroidSdkDirectory.TrimEnd ('\\')}\"");
+			}
+			if (!string.IsNullOrWhiteSpace (AvdManagerHome)) {
+				args.Append ($" --avdhome \"{AvdManagerHome.TrimEnd ('\\')}\"");
+			}
 			var success = RunProcess (
 				fileInfo.FullName,
-				$"--devicename {DeviceName} --verbose",
+				args.ToString (),
 				timeoutInMS,
 				(string data, ManualResetEvent mre) => {
 					Log.LogMessage (MessageImportance.High, $"CheckBootTimes ({DateTime.UtcNow}): {data}");
--- a/build-tools/check-boot-times/Program.cs
+++ b/build-tools/check-boot-times/Program.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.CommandLine;
 using System.Diagnostics;
@@ -13,6 +13,7 @@ namespace Xamarin.Android.Tools
 	{
 		static string sdkPath;
 		static string adbPath;
+		static string avdHome;
 		static string avdManagerPath;
 		static string emulatorPath;
 		static string sdkManagerPath;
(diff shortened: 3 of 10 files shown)


samples/NuGet.config
```diff
@@ -1,7 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
dellis1972, line 1: @pjcollins Why was this file deleted? Its kinda needed in order to run the dotnet 6 samples in vs code.

pjcollins: I replaced this with a task that generates the file instead, it was causing issues in CI since the Debug source doesn't exist in our Release test jobs. @dellis1972 could you test this https://github.com/xamarin/xamarin-android/pull/5947/files#diff-7d76d7533653c23b753fc7ce638cf64bdb5e419927d276af836d3a03fdf1745aR115 to see if it works for you?

jonpryor:
Peter believes that this PR will also address:  https://discord.com/channels/732297728826277939/732297837953679412/846380822905749515

> would you be able to look into the "flakiness" of the **run VSAndroidApp** step? e.g.
> ```
> NuGet.targets(131,5): error : The local source '/Users/runner/work/1/s/bin/BuildDebug/nuget-unsigned' doesn't exist. [/Users/runner/work/1/s/samples/VSAndroidApp/VSAndroidApp.csproj]
> ```

samples/NuGet.config
dellis1972: ok , I didn't know `tests/Mono.Android-Tests/Directory.Build.targets` even existed lol. That looks like it should work.

build-tools/scripts/TestApks.targets
```diff
@@ -346,6 +346,8 @@
   <Target Name="CheckBootTimes"
       DependsOnTargets="AcquireAndroidTarget;ReleaseAndroidTarget">
     <RunAndroidEmulatorCheckBootTimes
+        AndroidSdkDirectory="$(AndroidSdkDirectory)"
+        AvdHome="$(AvdManagerHome)\.android\avd"
```
jonathanpeppers, line 350: If you compare to the `<CreateAndroidEmulator/>` task:

https://github.com/xamarin/xamarin-android/blob/02ab8f723b6b8e327a3d8a6fd99192598ca94668/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CreateAndroidEmulator.cs#L16-L20

https://github.com/xamarin/xamarin-android/blob/02ab8f723b6b8e327a3d8a6fd99192598ca94668/build-tools/scripts/TestApks.targets#L55

Is there something we could do to make both of these tasks more similar?

pjcollins: I've updated all three tasks to now use `$(ANDROID_PREFS_ROOT)`.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CreateAndroidEmulator.cs
```diff
@@ -47,7 +47,7 @@ public override bool Execute ()
 			}
 
 			var env = new List<string> ();
-			env.Add ($"ANDROID_SDK_HOME={AvdManagerHome}");
+			env.Add ($"ANDROID_PREFS_ROOT={AvdManagerHome}");
```
jonpryor, line 50: do we have a link to docs on `$ANDROID_PREFS_ROOT`?

pjcollins: It's indirectly mentioned in the link in the description above: https://developer.android.com/studio/command-line/variables

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/RunAndroidEmulatorCheckBootTimes.cs
```diff
@@ -32,9 +36,16 @@ public override bool Execute ()
 			var timeoutInMS = (int) TimeSpan.FromMinutes (15).TotalMilliseconds;
 			bool finishAsExpected = false;
 			bool processTimedOut = false;
+			string args = $"--devicename {DeviceName} --verbose";
+			if (!string.IsNullOrWhiteSpace (AndroidSdkDirectory)) {
+				args += $" --sdkpath {AndroidSdkDirectory}";
```
jonpryor, line 41: For Great Sanity, this should probably be quoted somehow? Presumably:

```csharp
args += $" --sdkpath \"{AndroidSdkDirectory.TrimEnd ('\\')}\"";
```

Likewise `AvdManagerHome`, below.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/RunAndroidEmulatorCheckBootTimes.cs
```diff
@@ -32,9 +36,16 @@ public override bool Execute ()
 			var timeoutInMS = (int) TimeSpan.FromMinutes (15).TotalMilliseconds;
 			bool finishAsExpected = false;
 			bool processTimedOut = false;
+			string args = $"--devicename {DeviceName} --verbose";
```
jonpryor, line 39: `args` should become a `System.Text.StringBuilder`.

build-tools/check-boot-times/Program.cs
```diff
@@ -397,6 +401,20 @@ static Task<bool> RunProcess (string filename, string arguments, int timeout, Fu
...
 
+					var sdkRootKey = "ANDROID_SDK_ROOT";
+					var prefsRootKey = "ANDROID_PREFS_ROOT";
+					if (process.StartInfo.EnvironmentVariables.ContainsKey (sdkRootKey))
+						process.StartInfo.EnvironmentVariables [sdkRootKey] = sdkPath;
+					else
+						process.StartInfo.EnvironmentVariables.Add (sdkRootKey, sdkPath);
+
+					if (!string.IsNullOrWhiteSpace (avdHome)) {
+						if (process.StartInfo.EnvironmentVariables.ContainsKey (prefsRootKey))
```
jonpryor, line 412: This check isn't necessary; you can just always set it, even if it doesn't already exist:

```csharp
process.StartInfo.EnvironmentVariables [prefsRootKey] = avdHome;
```

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
```diff
@@ -566,16 +561,22 @@ public void RunWithInterpreterEnabled ([Values (false, true)] bool isRelease)
...
-			var logcatFilePath = Path.Combine (Root, builder.ProjectDirectory, "logcat.log");
-			var didStart = WaitForActivityToStart (proj.PackageName, "MainActivity", logcatFilePath, 30);
-			RunAdbCommand ("shell setprop debug.mono.log \"\"");
+			var didStart = WaitForActivityToStart (proj.PackageName, "MainActivity", "startup-logcat.log", 120);
+			RunAdbCommand ("shell setprop debug.mono.log \"''\"");
+			logProp = RunAdbCommand ("shell getprop debug.mono.log")?.Trim ();
+			Assert.AreEqual (logProp, string.Empty, "The debug.mono.log prop was not unset correctly.");
 			Assert.True (didStart, "Activity should have started.");
-			Assert.IsTrue (File.ReadAllText (logcatFilePath).Contains ("Enabling Mono Interpreter"), "logcat output did not contain 'Enabling Mono Interpreter'.");
+			Assert.True (MonitorAdbLogcat ((line) => {
```
jonpryor, line 577: Style: `line` doesn't *need* the extra parenthesis; `line => …` is fine.

jonpryor: For my understanding, is the reason this assert was frequently failing because the test didn't wait long enough between `RunAdbCommand()` and the `File.ReadAllText()` for the "Enabling Mono Interpreter" message to *reliably* exist in the log file?

If that's the case, this appears to be the only instance of this erroneous pattern; all other matches of:

```
% git grep 'File.ReadAllText.*logcat'
build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/RenameTestCases.cs:                               var logcat      = File.ReadAllText (logcatPath);
tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs:                     Assert.IsTrue (ObfuscatedStackRegex.IsMatch (File.ReadAllText (logcatFilePath)), "Original logcat output did not contain obfuscated crash info.");
tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs:                     var logcatOutput = File.ReadAllText (logcatPath);
tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs:                     Assert.IsTrue (File.ReadAllText (logcatFilePath).Contains ("Enabling Mono Interpreter"), "logcat output did not contain 'Enabling Mono Interpreter'.");
```

appear to occur *after* a `MonitorAdbLogcat()` invocation.

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
```diff
@@ -566,16 +561,22 @@ public void RunWithInterpreterEnabled ([Values (false, true)] bool isRelease)
...
+				return line.Contains ("Enabling Mono Interpreter");
+			}, Path.Combine (Root, builder.ProjectDirectory, "interpreter-logcat.log"), 120), "logcat output did not contain 'Enabling Mono Interpreter'.");
```
jonpryor, line 579: While I realize this "pattern" is used elsewhere in this file, *I* found this line confusing to read, as the `120)` appears to be "free floating", and I no longer see what it is applied to.

Thus, could you split things up for clarity?

```csharp
Func<string, bool> checkForInterpeter = line => {
    return line.Contains ("Enabling Mono Interpreter");
};
Assert.IsTrue (
    condition: MonitorAdbLogcat (
        action: checkForInterpreter,
        logcatFilePath: Path.Combine (Root, builder.ProjectDirectory, "interpreter-logcat.log"),
        timeout: 120),
    message: "logcat output did not contain 'Enabling Mono Interpreter'.");
```

pjcollins:
I think the failure in the latest run can be ignored. I've seen this intermittent issue before and we can probably try to address it separately eventually.
>   /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.Debugging.targets(604,5): error XA0010: Could not retrieve the Android ABI for the attached device or emulator. Please reconnect the device or restart the emulator, and try again. [/Users/runner/work/1/s/bin/TestRelease/temp/ProfilerLogOptionsShouldCreateMlpdFileslogsample/UnnamedProject.csproj]

jonathanpeppers (approved):

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
pjcollins: I am not entirely sure what the root cause was, in some cases I think there was a timing issue in reading the file, or a timing issue in app launching. I could never reproduce this one locally and only saw it on CI.

PR #5953, opened 2021-05-24
[Xamarin.Android.Build.Tasks] Allow package to be defined with AndroidManifestPlaceholders

### LINKED ISSUE #5951: AndroidManifestPlaceholders do not work when used for manifest->package
### General feature description since there is no docu for it.

This is what it does on Android:
[https://developer.android.com/studio/build/manifest-build-variables](https://developer.android.com/studio/build/manifest-build-variables)

It works in Xamarin Android by adding
```xml
<PropertyGroup> 
<AndroidManifestPlaceholders>PlaceHolderName=PlaceHolderValue</AndroidManifestPlaceholders> 
</PropertyGroup>
```
to the Xamarin Android csproj
and using it like this in the manifest:
`<application android:label="${PlaceHolderName}"></application>`

### Steps to Reproduce

1. Add AndroidManifestPlaceholders in the Xamarin Android csproj as explained above.
2. In the Manifest file use the place holder for manifest->package. like this
```xml
          package="${PlaceHolderName}">
```
3. Build and deploy it to a device/emulator
4. look at the apk file name. it looks like this "x__PlaceHolderName_.x__PlaceHolderName_-Signed.apk". I assume this is because the Placeholder is not replaced and when building the apk the not allowed character are converted to other characters. $->x and { -> _.

### Expected Behavior
The place holder should be replaced as expected.
```xml
          package="${PlaceHolderName}">
```
should become
```xml
          package="PlaceHolderValue">
```
and the resulting apk should be called PlaceHolderValue-Signed.apk

### Actual Behavior
described in the steps.

### Version Information
Xamarin.Android SDK   11.2.2.1 (d16-9/877f572)
    Mono: 5e9cb6d
    Java.Interop: xamarin/java.interop/d16-9@54f8c24
    ProGuard: Guardsquare/proguard/v7.0.1@912d149
    SQLite: xamarin/sqlite/3.34.1@daff8f4
    Xamarin.Android Tools: xamarin/xamarin-android-tools/d16-9@d210f11

dellis1972:
Fixes #5951
  Related: commit b423ff6.
  
  Our current `AndroidManifestPlaceholders` implementation would allow users
  to change any part of the manifest with the exception of the `package`
  element on the `manifest` itself. What was actually happening was the
  `package` was being changed but it was happening after  the
  `GetAndroidPackageName` had already been called. So the internal MSBuild
  property `$(_AndroidPackage)` did not reflect the desired value.
  
  So lets update `GetAndroidPackageName` to use the `AndroidManifestPlaceholders`.
  Note: This will only effect the `$(_AndroidPackage)` value when the
  manifest is being used. If the user is using the `GenerateApplicationManifest`
  option, the `$(_AndroidPackage)` will be controlled from the `$(ApplicationId)`
  MSBuild property.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs
@@ -501,18 +501,25 @@ namespace Bug12935
 		}
 
 		[Test]
-		public void ManifestPlaceholders ()
+		public void ManifestPlaceholders ([Values ("legacy", "manifestmerger.jar")] string manifestMerger)
 		{
 			var proj = new XamarinAndroidApplicationProject () {
 				IsRelease = true,
+				ManifestMerger = manifestMerger,
+				JavaPackageName = "com.foo.bar",
 			};
-			proj.AndroidManifest = proj.AndroidManifest.Replace ("application android:label=\"${PROJECT_NAME}\"", "application android:label=\"${ph1}\" x='${ph2}' ");
-			proj.SetProperty ("AndroidManifestPlaceholders", "ph2=a=b\\c;ph1=val1");
-			using (var builder = CreateApkBuilder (Path.Combine ("temp", TestContext.CurrentContext.Test.Name), false, false)) {
+			proj.AndroidManifest = proj.AndroidManifest.
+				Replace ("application android:label=\"${PROJECT_NAME}\"", "application android:label=\"${ph1}\" x='${ph2}' ").
+				Replace ("package=\"${PACKAGENAME}\"", "package=\"${Package}\"");
+			proj.SetProperty ("AndroidManifestPlaceholders", "ph2=a=b\\c;ph1=val1;Package=com.foo.bar");
+			using (var builder = CreateApkBuilder ()) {
 				builder.Build (proj);
 				var manifest = builder.Output.GetIntermediaryAsText (Root, Path.Combine ("android", "AndroidManifest.xml"));
 				Assert.IsTrue (manifest.Contains (" android:label=\"val1\""), "#1");
 				Assert.IsTrue (manifest.Contains (" x=\"a=b\\c\"".Replace ('\\', Path.DirectorySeparatorChar)), "#2");
+				Assert.IsTrue (manifest.Contains ("package=\"com.foo.bar\""), "PackageName should have been replaced with 'com.foo.bar'");
+				var apk = Path.Combine (Root, builder.ProjectDirectory, proj.OutputPath, $"com.foo.bar.apk");
+				FileAssert.Exists (apk, $"'{apk}' should have been created.");
 			}
 		}
 
--- a/src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs
@@ -258,6 +258,11 @@ namespace Xamarin.Android.Tasks {
 			} else {
 				PackageName = manifest_package;
 			}
+			if (PackageName.Contains ("${")) {
+				// placeholder detected
+				PackageName = ReplacePlaceholders (Placeholders, PackageName);
+				manifest.SetAttributeValue ("package", PackageName);
+			}
 
 			manifest.SetAttributeValue (XNamespace.Xmlns + "android", "http://schemas.android.com/apk/res/android");
 
@@ -960,13 +965,7 @@ namespace Xamarin.Android.Tasks {
 			}
 			if (!string.IsNullOrEmpty (PackageName))
 				s = s.Replace ("${applicationId}", PackageName);
-			if (Placeholders != null)
-				foreach (var entry in Placeholders.Select (e => e.Split (new char [] {'='}, 2, StringSplitOptions.None))) {
-					if (entry.Length == 2)
-						s = s.Replace ("${" + entry [0] + "}", entry [1]);
-					else
-						logCodedWarning ("XA1010", string.Format (Properties.Resources.XA1010, string.Join (";", Placeholders)));
-				}
+			s = ReplacePlaceholders (Placeholders, s, logCodedWarning);
 			stream.Write (s);
 		}
 
@@ -999,6 +998,22 @@ namespace Xamarin.Android.Tasks {
 			default:
 				throw new ArgumentOutOfRangeException ("abi", "unsupported ABI");
 			}
+		}
+
+		internal static string ReplacePlaceholders (string [] placeholders, string text, Action<string, string> logCodedWarning = null)
+		{
+			string result = text;
+			if (placeholders == null)
+				return result;
+			foreach (var entry in placeholders.Select (e => e.Split (new char [] {'='}, 2, StringSplitOptions.None))) {
+				if (entry.Length == 2)
+					result = result.Replace ("${" + entry [0] + "}", entry [1]);
+				else {
+					if (logCodedWarning != null)
+						logCodedWarning ("XA1010", string.Format (Properties.Resources.XA1010, string.Join (";", placeholders)));
+				}
+			}
+			return result;
 		}
 
 		public void SetAbi (string abi)
--- a/Documentation/release-notes/5951.md
+++ b/Documentation/release-notes/5951.md
@@ -1,0 +1,5 @@
+#### Application build and deployment
+
+- [GitHub Issue #5851](https://github.com/xamarin/xamarin-android/issues/5851):
+  Add support for using `AndroidManifestPlaceholders` to set the `package`
+  value in the `AndroidManifest.xml`.
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GetAndroidPackageName.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GetAndroidPackageName.cs
@@ -1,21 +1,21 @@
-// 
+//
 // GetAndroidPackageName.cs
-//  
+//
 // Author:
 //       Jonathan Pobst <monkey@jpobst.com>
-// 
+//
 // Copyright (c) 2010 Novell, Inc. (http://www.novell.com)
-// 
+//
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
 // in the Software without restriction, including without limitation the rights
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is
 // furnished to do so, subject to the following conditions:
-// 
+//
 // The above copyright notice and this permission notice shall be included in
 // all copies or substantial portions of the Software.
-// 
+//
 // THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 // IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
 // FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
@@ -23,8 +23,9 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-
+using System;
 using System.IO;
+using System.Linq;
 using System.Xml;
 using Microsoft.Build.Framework;
 using Xamarin.Android.Tools;
@@ -41,6 +42,8 @@ namespace Xamarin.Android.Tasks
 		[Required]
 		public string AssemblyName { get; set; }
 
+		public string [] ManifestPlaceholders { get; set; }
+
 		[Output]
 		public string PackageName { get; set; }
 
@@ -54,6 +57,7 @@ namespace Xamarin.Android.Tasks
 				if (reader.MoveToContent () == XmlNodeType.Element) {
 					var package = reader.GetAttribute ("package");
 					if (!string.IsNullOrEmpty (package)) {
+						package = ManifestDocument.ReplacePlaceholders (ManifestPlaceholders, package);
 						PackageName = AndroidAppManifest.CanonicalizePackageName (package);
 					}
 				}
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -509,6 +509,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
   <GetAndroidPackageName
       ManifestFile="$(_AndroidManifestAbs)"
       AssemblyName="$(AssemblyName)"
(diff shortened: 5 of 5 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidApplicationProject.cs
```diff
@@ -72,7 +72,7 @@ public XamarinAndroidApplicationProject (string debugConfigurationName = "Debug"
 			JavaPackageName = JavaPackageName ?? PackageName.ToLowerInvariant ();
 
 			OtherBuildItems.Add (new BuildItem.NoActionResource ("Properties\\AndroidManifest.xml") {
-				TextContent = ProcessManifestTemplate,
+				TextContent = () => { return ProcessManifestTemplate (); },
```
jonathanpeppers, line 75: Do we need this change?

jonathanpeppers (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidApplicationProject.cs
dellis1972: Yes. Without this change the `${PACKAGENAME}` value in the default manifest gets replaced with the default PackageName property. As a result the code I have to replace the package with the placeholder never works.

jonathanpeppers: I think the code is the same, the first one would be the same as doing:

```csharp
TextContent = new Func<string> (ProcessManifestTemplate)
```

I've heard putting `() => ProcessManifestTemplate()` prevents a delegate object from being created, but that performance shouldn't be needed here.

dellis1972: The new test where I replace the code in the manifest with a Placeholder did not work with the old `TestContent` code. I'll take it out and see if I can figure out why the replacement was not happening.

jonpryor:
There's a conceptual problem with this PR: it doesn't *fully* fix issue #5951.  Issue #5951 directs us to:

  * https://developer.android.com/studio/build/manifest-build-variables

which shows examples of how the "manifest build variables" can be used *anywhere*/everywhere within `AndroidManifest.xml`, e.g.

```xml
<intent-filter … >
    <data android:scheme="https" android:host="${hostName}" … />
    …
</intent-filter>
```

where `hostName` is a "manifest build variable".

*This* PR *only* updates the value of `/manifest/package`, and nothing else.  It is thus not an "android manifest placeholder", but rather an "android package-name placeholder".

---

Additionally, as a "slight change" to the original request, instead of an `$(AndroidManifestPlaceholders)` property, should it instead be handled via an `@(AndroidManifestPlaceholder)` *item group*?  This would make it easier for NuGet packages to influence the generated `AndroidManifest.xml`, and is otherwise easier to understand than the property-requiring "append" syntax:

```xml
<PropertyGroup>
  <AndroidManifestPlaceholders>$(AndroidManifestPlaceholders);NewValue=Here</AndroidManifestPlaceholders>
</PropertyGroup>
```

jonpryor:
Related: commit b423ff68a6d11230bb493c3cb0633ff2a9c37b20.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidApplicationProject.cs
dellis1972: I backed out this change and it still worked... hey ho. Must have been a typo in my replacement code or something when I was testing.

jonpryor:
Not sure what's up with the `Collision` test failure: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4797205&view=ms.vss-test-web.build-test-results-tab&runId=21992302&resultId=100160&paneView=debug

```
Execute() failed!
Expected: True
But was:  False
```

Re-running those tests…

jonpryor:
While writing up the commit message, ran into a couple "thinko's".

> What was actually happening was the
> `package` was being changed but it was happening after the
> `GetAndroidPackageName` had already been called.

I don't think that's *quite* true; I can only find two tasks that would meaningfully use `$(AndoridManifestPlaceholders)`: `<GenerateJavaStubs/>` and `<ManifestMerger/>`.

`<GenerateJavaStubs/>` is given the package name as a parameter, and thus does hit the described "timing" problem, but even if there were no timing problem, `<GenerateJavaStubs/>` uses [`ManifestDocument`](https://github.com/xamarin/xamarin-android/blob/9add575ba0e25903b6894f40fcd53e6d51785fae/src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs), and `ManifestDocument` *never* uses `ManifestDocument.Placeholders` to update `ManifestDocument.PackageName`.  The result is that the package name would never be updated in any file on-disk in this codepath.

Then there's `<ManifestMerger/>`, which *doesn't run by default*; it only runs when `$(AndroidManifestMerger)`=`manifestmerger.jar`, which is *never* the case.

TODO: Update the .NET 6 default properties so that `$(AndroidManifestMerger)` is set to `manifestmerger.jar`.

Furthermore, even if/when `$(AndroidManifestMerger)` is `manifestmerger.jar`, it still doesn't work, because `<ManifestMerger/>`s input of `obj/Debug/AndroidManifest.xml` has *already been updated* to have a `/manifest/@package` value that doesn't contain placeholders…by `<GenerateJavaStubs/>`.

As far as I can tell, *prior to this PR*, `/manifest/@package` can *never* participate in placeholder substitution.

jonathanpeppers:
> Then there's `<ManifestMerger/>`, which *doesn't run by default*; it only runs when `$(AndroidManifestMerger)`=`manifestmerger.jar`, which is *never* the case.

Just FYI any project using AndroidX turns on `manifestmerger.jar`:

https://github.com/xamarin/AndroidX/blob/d0dcede8cdfb5c05b2c6f12b659609d88b50a819/source/AndroidXTargets.cshtml#L32-L34

I think they did this because some AndroidX libraries required correct merging for them to work properly.

jonpryor:
That background out of the way, I now don't understand how this PR *works*:  `<GenerateJavaStubs/>` calls `ManifestDocument.Merge()`: https://github.com/xamarin/xamarin-android/blob/main/src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs#L290

`<GenerateJavaStubs/>` also writes out the "final" `AndroidManifest.xml` file, and thus is *really* important.

`ManifestDocument.Merge()`, meanwhile: https://github.com/xamarin/xamarin-android/blob/9add575ba0e25903b6894f40fcd53e6d51785fae/src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs#L246-L260

`ManifestDocument.PackageName` is set by `<GenerateJavaStubs/>` as part of `ManifestDocument` construction, and thus ultimately comes from `<GetAndroidPackageName/>`, which is updated by this PR.  So far so good.

However, if `AndroidManifest.xml` *has* `/manifest/@package` value -- which it does for this bug! -- then line 259 *replaces* `ManifestDocument.PackageName` with the `/manifest/@package` value.  `PackageName` is never updated to take `ManifestDocument.Placeholders` into consideration, so my *hypothesis* is that the *actual* `AndroidManifest.xml`'s `/manifest/@package` value is *still* "wrong", not containing appropriately updated values.

@dellis1972: please apply the following patch to your xamarin-android repo:

```diff
diff --git a/samples/HelloWorld/HelloWorld.csproj b/samples/HelloWorld/HelloWorld.csproj
index 2b5391ee2..71898b696 100644
--- a/samples/HelloWorld/HelloWorld.csproj
+++ b/samples/HelloWorld/HelloWorld.csproj
@@ -23,6 +23,7 @@
   />
   <PropertyGroup Condition="Exists('..\..\Configuration.props')">
     <TargetFrameworkVersion>$(AndroidFrameworkVersion)</TargetFrameworkVersion>
+    <AndroidManifestPlaceholders>hello=a;world=b</AndroidManifestPlaceholders>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
diff --git a/samples/HelloWorld/Properties/AndroidManifest.xml b/samples/HelloWorld/Properties/AndroidManifest.xml
index 3561f38df..52efdc5ae 100644
--- a/samples/HelloWorld/Properties/AndroidManifest.xml
+++ b/samples/HelloWorld/Properties/AndroidManifest.xml
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0" package="com.xamarin.android.helloworld">
+<manifest xmlns:android="http://schemas.android.com/apk/res/android" android:versionCode="1" android:versionName="1.0" package="com.xamarin.android.${hello}.${world}">
 	<uses-sdk android:minSdkVersion="21" />
 	<application android:allowBackup="true" android:icon="@mipmap/icon" android:label="@string/app_name" android:extractNativeLibs="false">
 	</application>
```

Then please build `samples/HelloWorld/HelloWorld.csproj` using your local build with this PR.

Based on the unit test, this *should* produce a `com.xamarin.android.a.b-Signed.apk` file.

Once built, please read the generated file `obj//Debug/AndroidManifest.xml`.  I *think* that this file will instead contain `package="com.xamarin.android.x__hello_.x__world_`, *not* `package="com.xamarin.android.a.b`.  (Meaning that this PR only really changes the *filename produced*, but not the actual intermediate `AndroidManifest.xml` files.)

If my hypothesis is correct, I believe that the appropriate fix would be to instead update `AndroidAppManifest.CanonicalizePackageName()` to take a `string[] placeholders` parameter, and have *both* `<GetAndroidPackageName/>` and `ManifestDocument.Merge()` call `AndroidAppManifest.CanonicalizePackageName(PackageName, placeholders)`.  (The current `foreach` over the placeholders would likewise move into `AndroidAppManifest.CanonicalizePackageName()`.)

jonpryor:
WIP commit message:

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/5951

Context: b423ff68a6d11230bb493c3cb0633ff2a9c37b20

Our current [`$(AndroidManifestPlaceholders)`][0] implementation,
added in commit b423ff68, allows users to use placeholders in any
part of `AndroidManifest.xml` *except* for the `/manifest/@package`
attribute value.

This is the case because `/manifest/@package` is read and
"canonicalized" before `$(AndroidManifestPlaceholders)` processing
occurs, and "canonicalization" *prevents* it from participating in
`$(AndroidManifestPlaceholders)` processing.

Consequently, if `/manifest/@package` contains placeholders:

	<manifest
	    xmlns:android="http://schemas.android.com/apk/res/android"
	    package="com.example.${placeholder}">

then after the build the generated `AndroidManifest.xml` -- often in
`obj/$(Configuration)/AndroidManifest.xml` -- will contain:

	<manifest
	    xmlns:android="http://schemas.android.com/apk/res/android"
	    package="com.example.x__placeholder_">

and the resulting `.apk` will likewise be
`com.example.x__placeholder_.apk`.

Fix this by updating `AndroidAppManifest.CanonicalizePackageName()`
to take the placeholders, so that both the `<GetAndroidPackageName/>`
and `<GenerateJavaStubs/>` tasks can produce the correct package name.

[0]: https://docs.microsoft.com/en-us/xamarin/android/deploy-test/building-apps/build-properties#androidmanifestplaceholders
```

jonpryor:
@dellis1972 : this could also use a release-notes blurb. :-)

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs
```diff
@@ -506,13 +506,17 @@ public void ManifestPlaceholders ()
...
+			proj.AndroidManifest = proj.AndroidManifest.
+				Replace ("application android:label=\"${PROJECT_NAME}\"", "application android:label=\"${ph1}\" x='${ph2}' ").
+				Replace ("package=\"${PACKAGENAME}\"", "package=\"${Package}\"");
+			proj.SetProperty ("AndroidManifestPlaceholders", "ph2=a=b\\c;ph1=val1;Package=com.foo.bar");
 			using (var builder = CreateApkBuilder (Path.Combine ("temp", TestContext.CurrentContext.Test.Name), false, false)) {
 				builder.Build (proj);
 				var manifest = builder.Output.GetIntermediaryAsText (Root, Path.Combine ("android", "AndroidManifest.xml"));
 				Assert.IsTrue (manifest.Contains (" android:label=\"val1\""), "#1");
 				Assert.IsTrue (manifest.Contains (" x=\"a=b\\c\"".Replace ('\\', Path.DirectorySeparatorChar)), "#2");
+				var apk = Path.Combine (Root, builder.ProjectDirectory, proj.OutputPath, $"com.foo.bar.apk");
```
jonpryor, line 518: Should also add:

```csharp
Assert.IsTrue (manifest.Contains (" package=\"com.foo.bar\""), "Package should be `com.foo.bar`.");
```

dellis1972:
@jonpryor I don't think we should mess about with `AndroidAppManifest.CanonicalizePackageName`. That method is in [androidtools](https://github.com/xamarin/androidtools/blob/d3b582246b060a4e70ad68bea9e7c0291ff23495/Xamarin.AndroidTools/AndroidAppManifest.cs#L56).

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs
```diff
@@ -501,18 +501,26 @@ public void VersionCodeTests (bool seperateApk, string abis, string versionCode,
...
+			proj.SetProperty ("AndroidManifestPlaceholders", "ph2=a=b\\c;ph1=val1;Package=com.foo.bar");
+			using (var builder = CreateApkBuilder ()) {
 				builder.Build (proj);
 				var manifest = builder.Output.GetIntermediaryAsText (Root, Path.Combine ("android", "AndroidManifest.xml"));
 				Assert.IsTrue (manifest.Contains (" android:label=\"val1\""), "#1");
 				Assert.IsTrue (manifest.Contains (" x=\"a=b\\c\"".Replace ('\\', Path.DirectorySeparatorChar)), "#2");
+				Assert.IsTrue (manifest.Contains ("package=\"com.foo.bar\""), "PackageName should have been replaced with 'com.foo.bar'");
+				var apk = Path.Combine (Root, builder.ProjectDirectory, proj.OutputPath, $"com.foo.bar.apk");
+				FileAssert.Exists (apk, $"'{apk}' should have been created.");
+				Assert.Fail ();
```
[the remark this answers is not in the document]
dellis1972: doh that was for debugging

src/Xamarin.Android.Build.Tasks/Utilities/ManifestDocument.cs
```diff
@@ -1001,6 +1000,21 @@ static int GetAbiCode (string abi)
 			}
 		}
 
+		internal static string ReplacePlaceolders (string [] placeholders, string text, Action<string, string> logCodedWarning = null) {
```
jonpryor, line 1003: `{` should go onto a new line.

PR #2899, opened 2019-03-29
[Xamarin.Android.Build.Tasks] Support Filtering Assemblies

dellis1972:
Context #2895 

As part of the AndroidX migration system we need to
be able to skip over the older support library
assemblies for certain build steps.

The following metadata is supported on the `ReferencePath`
and `ReferenceDependencyPaths` ItemGroups.

	AndroidSkipAddToPackage
	AndroidSkipResourceProcessing
	AndroidSkipJavaStubGeneration
	AndroidSkipResourceExtraction

`AndroidSkipResourceProcessing` is the renamed
`SkipAndroidResourceProcessing` metadata item.
This is used to skip case conversion on those
resources.

`AndroidSkipResourceExtraction` will stop the
resources in that library being extracted into
the `IntermediateOutputPath`.

`AndroidSkipJavaStubGeneration` when `true` this
item will stop java code being generated for the
library in quetion.

`AndroidSkipAddToPackage` is used by `BuildApk`
task to skip over the library and NOT add it to
the final APK.

To use these new metadata items you can define your
own `ItemGroup` as follows

	<ItemGroup>
		<_MyStuff Include="Xamarin.Android.Support.v4">
			<AndroidSkipAddToPackage>True</AndroidSkipAddToPackage>
		</_MyStuff>
	</ItemGroup>

Then add this new `ItemGroup` to the `AndroidCustomMetaDataForReferences`
like so.

	<ItemGroup>
		<AndroidCustomMetaDataForReferences Include="@(_MyStuff)" />
	</ItemGroup>

The `AndroidCustomMetaDataForReferences` ItemGroup will then
be processed by the `_AddAndroidCustomMetaData` target. This
will add the new metadata to the `ReferencePath`
and `ReferenceDependencyPaths` ItemGroups.

PR #5955, opened 2021-05-24
[runtime] Less reflection

grendello:
Try to remove as much reflection (both managed and Java) from the native
startup sequence.

Instead of using full reflection for managed classes, this commit
implements class and field loading using token IDs.  The IDs are
gathered at the time when typemaps are generated and then stored in the
generated assembly code, so that the native runtime can simply pass them
to Mono in order to load the needed class and its members.

Additionally, I noticed that the default Java time zone is no longer
requested during startup, so I moved the class lookup to a lazy code
path, protected with a mutex.


src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
```diff
@@ -126,21 +126,28 @@ void Run (DirectoryAssemblyResolver res)
...
+			var tokenIdCollection = new NativeRuntimeTokenIdCollection (Log);
 			foreach (var assembly in ResolvedAssemblies) {
 				bool value;
+				string fileName = Path.GetFileName (assembly.ItemSpec);
+
+				if (String.Compare ("Java.Interop.dll", fileName, StringComparison.CurrentCultureIgnoreCase) == 0) {
+				    tokenIdCollection.ProcessJavaInterop (assembly.ItemSpec);
+				}
+
 				if (bool.TryParse (assembly.GetMetadata (AndroidSkipJavaStubGeneration), out value) && value) {
```
jonpryor, line 138: (The joys of time and broader context…)

@dellis1972 : should commit 21561c3b11a0b7763e8bf2ed9f718e6a8f2446ce / PR #2899 have *also* set `%(AndroidSkipJavaStubGeneration)` on `Java.Interop.dll` and `Mono.Android.Export.dll`?  (And, *at the time*, `OpenTK-1.0.dll`?  Though that's needed *now* as of f909adb146c0ee87dafad2417e301a14afb79a33 …)

src/monodroid/jni/monodroid-glue-internal.hh
```diff
@@ -155,8 +155,12 @@ namespace xamarin::android::internal
 			return java_System_identityHashCode;
 		}
 
-		jclass get_java_class_TimeZone () const
+		jclass get_java_class_TimeZone ()
 		{
+			if (XA_UNLIKELY (java_TimeZone == nullptr)) {
+				look_up_java_timezone ();
```
jonpryor, line 161: Style: this should be `lookup_java_timezone()`.

Rationale: consistency:

```bash
% git grep look_up src/monodroid | wc -l
       0
% git grep lookup_ src/monodroid | wc -l
      16
```

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -980,14 +981,14 @@ MonodroidRuntime::init_android_runtime (MonoDomain *domain, JNIEnv *env, jclass
 		lookup_bridge_info (domain, image, &osBridge.get_java_gc_bridge_type (i), &osBridge.get_java_gc_bridge_info (i));
 	}
 
-	// TODO: try looking up the method by its token
-	MonoClass *runtime = utils.monodroid_get_class_from_image (domain, image, "Android.Runtime", "JNIEnv");
-	MonoMethod *method = mono_class_get_method_from_name (runtime, "Initialize", 1);
+	// Android.Runtime.JNIEnv
+	MonoClass *runtime = utils.monodroid_get_class (domain, image, managed_token_ids.android_runtime_jnienv);
+	abort_unless (runtime != nullptr, "INTERNAL ERROR: Unable to find class Android.Runtime.JNIEnv in the Mono.Android assembly");
```
jonpryor, line 986: We should prefer an existing convention when one exists.  In this case, this is an assembly-qualified name, thus:

```cc
"INTERNAL ERROR: Unable to find type `Android.Runtime.JNIEnv, Mono.Android`!"
```

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -98,6 +99,7 @@ using namespace xamarin::android::internal;
 // implementation details as it would prevent mkbundle from working
 #include "mkbundle-api.h"
 
+std::mutex MonodroidRuntime::java_reflection_lock;
```
jonpryor, line 102: "naming & scope & semantics, oh my!": JNI *itself* is thread-safe.  (The target Java class?  Possibly not.  Our usage?  Possibly not.  JNI *itself* certainly is thread-safe.)

My problem with the *name* `java_reflection_lock` is that it *implies* that we need to lock "all" JNI access, or all "Java reflection" access (whatever that would be?).  This isn't the case; it's only used to protect *setting* our `MonodroidRuntime::java_TimeZone` field.

We could go "very granular" and have a `java_TimeZone_lock` field, or we could have a `lazy_jni_lookup_lock` lock for use with `MonodroidRuntime::java_TimeZone` and other "lazily looked up" fields in the future, or some other better name.

`java_reflection_lock` is too easily misinterpreted.

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -980,14 +981,14 @@ MonodroidRuntime::init_android_runtime (MonoDomain *domain, JNIEnv *env, jclass
...
+	MonoClass *runtime = utils.monodroid_get_class (domain, image, managed_token_ids.android_runtime_jnienv);
+	abort_unless (runtime != nullptr, "INTERNAL ERROR: Unable to find class Android.Runtime.JNIEnv in the Mono.Android assembly");
+	jnienv = runtime;
 
-	if (method == nullptr) {
-		log_fatal (LOG_DEFAULT, "INTERNAL ERROR: Unable to find Android.Runtime.JNIEnv.Initialize!");
-		exit (FATAL_EXIT_MISSING_INIT);
-	}
+	// Android.Runtime.JNIEnv.Initialize
+	MonoMethod *method = mono_get_method (image, managed_token_ids.android_runtime_jnienv_initialize, runtime);
```
jonpryor, line 990: This method is highly repetitive: lookup `method` by token, assert `method` is not null, store "somewhere", continue.

Is there some way we could turn this into a *loop*?

```c
MonoMethod *jnienv_initialize;

struct MethodLookupInfo {
    MonoMethod **target;
    const char *description;
    uint32_t token_id;
} methods[] = {
    { &jnienv_initialize, "Android.Runtime.JNIEnv.Initialize", managed_token_ids.android_runtime_jnienv_initialize },
    { &registerType, "Android.Runtime.RegisterJniNatives", managed_token_ids.android_runtime_jnienv_registerjninatives },
    // …
};

for (int i = 0; i < sizeof(methods)/sizeof(methods[0]); ++i) {
    *methods[i].target = mono_get_method (image, methods[i].token_id, runtime);
    if (*methods [i].target != nullptr)
        continue;
    abort (methods[i].description);
}
```

src/monodroid/jni/monodroid-glue.cc
grendello: > "naming & scope & semantics, oh my!": JNI _itself_ is thread-safe. (The target Java class? Possibly not. Our usage? Possibly not. JNI _itself_ certainly is thread-safe.)
> 
The lock protects our class field, we can't avoid using it, alas.

> My problem with the _name_ `java_reflection_lock` is that it _implies_ that we need to lock "all" JNI access, or all "Java reflection" access (whatever that would be?). This isn't the case; it's only used to protect _setting_ our `MonodroidRuntime::java_TimeZone` field.
> 
> We could go "very granular" and have a `java_TimeZone_lock` field, or we could have a `lazy_jni_lookup_lock` lock for use with `MonodroidRuntime::java_TimeZone` and other "lazily looked up" fields in the future, or some other better name.
> 
> `java_reflection_lock` is too easily misinterpreted.

I'll rename it

src/monodroid/jni/monodroid-glue.cc
grendello: I took a different approach, especially that I realized that we mustn't use token IDs in desktop builds, see the updated code.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
dellis1972: They probably should have been included at that time yes.

PR #5956, opened 2021-05-24
Create SECURITY.md

jpobst:
Copied and adjusted from https://github.com/dotnet/runtime/blob/main/SECURITY.md.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,9 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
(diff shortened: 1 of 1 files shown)


jonpryor:
As with dotnet/runtime, we can also have a **Supported Versions** section, which can link to: https://dotnet.microsoft.com/platform/support/policy/xamarin

```markdown
## Supported Versions

The Xamarin.Android support policy, including supported versions, can be found at the [Xamarin Support Policy](https://dotnet.microsoft.com/platform/support/policy/xamarin).
```

PR #5967, opened 2021-05-27
[ci] Install credential provider when signing

pjcollins:
Context:  https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4803805&view=logs&j=3ce66085-846b-53dd-5323-3e00eb12b9a6&t=c7de918a-2762-5750-800e-a440a28cd7e6

We've been seeing some .NET 6 .msi signing attempts fail when installing
the MicroBuild signing tooling with the following:

    ##[error]The path 'Microsoft.NuGet.CredentialProvider.zip' either does not exist or is not a valid file system path.

Installing the NuGet credential provider will hopefully address this.

### DIFF
--- a/build-tools/automation/yaml-templates/install-microbuild-tooling.yaml
+++ b/build-tools/automation/yaml-templates/install-microbuild-tooling.yaml
@@ -3,6 +3,12 @@ parameters:
   condition: succeeded()
 
 steps:
+- task: NuGetAuthenticate@0
+  displayName: authenticate with azure artifacts
+  inputs:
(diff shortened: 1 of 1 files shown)


pjcollins:
Test build triggered here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4809630&view=results

pjcollins:
The test build worked, though the issue that this is attempting to address does look to be intermittent. I think we can merge this and wait to see if the issue pops up again in main/release branches.

jonathanpeppers (approved):

PR #5986, opened 2021-06-02
[tests] remove SingleThreadedAttribute

jonathanpeppers:
We occasionally hit test failures in the
`InstallAndRunTests.InterpreterEnabled` test. Running the test locally
by itself seems to work fine.

Then I noticed the class was decorated with the `[SingleThreaded]`
attribute and not `[NonParallelizable]`. `[SingleThreaded]`'s summary
says:

> Marks a test fixture as requiring all child tests to be run on the
> same thread as the OneTimeSetUp and OneTimeTearDown. A flag in the
> NUnit.Framework.Internal.TestExecutionContext is set forcing all
> child tests to be run sequentially on the current thread. Any
> NUnit.Framework.ParallelScope setting is ignored.

How is this test class not failing *more* often than it is? I found
that we set an assembly-level attribute in `DeploymentTest.cs`:

    [assembly: NonParallelizable]

But from the description of `[SingleThreaded]`, it seems like this
setting would be ignored.

To (hopefully) make things better:

* Move `[assembly: NonParallelizable]` to an easier to find
  `AssemblyInfo.cs` file.
* Remove all usage of `[NonParallelizable]` in other files in the
  `MSBuildDeviceIntegration` project.
* Remove all usage of `[SingleThreaded]`.
* Fix a couple places we had a `UsesDevices` category, as `UsesDevice`
  appears to be the category used more often.
* Add `[TestFixture]` in a few places that it was missing.

### DIFF
--- a/tests/MSBuildDeviceIntegration/AssemblyInfo.cs
+++ b/tests/MSBuildDeviceIntegration/AssemblyInfo.cs
@@ -1,0 +1,3 @@
+using NUnit.Framework;
+
+[assembly: NonParallelizable]
--- a/tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
+++ b/tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
@@ -12,7 +12,8 @@ using System.Collections.Generic;
 
 namespace Xamarin.Android.Build.Tests
 {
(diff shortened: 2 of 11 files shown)


jonathanpeppers:
This PR is green, it's incredible!

PR #5987, opened 2021-06-03
Update VSCode Support

dellis1972:
Update the extensions list to use the correct id for the vscode csharp extension.

Remove the `nxunitExplorer.nunit` setting as it was causing the unit tests not to load.
This is because the extension is expecting a CIL image, but we were passing a batch
file. As a result none of the unit tests would load in the IDE.

Add some documentation around using VSCode to run the unit tests and debug the 
sample apps. This includes a link to a patched version of the `testrun.exe` which the 
`nxunit.explorer` extension uses. There is a bug in the extension where tests will 
parameters will NOT run or debug. The link patched version of the files is included 
in the docs. A PR to the upstream extension has also been sent.

Added the `dotnet-adapter` extension which lets us run the tests under a `dotnet`
context.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,10 +1,11 @@
 {
     "recommendations": [
-      "ms-vscode.csharp",
+      "ms-dotnettools.csharp",
       "ms-vscode.cpptools",
       "ms-vscode.mono-debug",
-      "wghats.vscode-nxunit-test-adapter",
       "visualstudioexptteam.vscodeintellicode",
       "streetsidesoftware.code-spell-checker",
+      "wghats.vscode-nxunit-test-adapter",
+      "derivitec-ltd.vscode-dotnet-adapter",
     ]
 }
--- a/Documentation/guides/vscode-support.md
+++ b/Documentation/guides/vscode-support.md
@@ -1,82 +1,88 @@
 # VSCode Support
 
-Xamarin.Android itself can be developed within VSCode. There is
-a workspace included in the repo `Xamarin.Android.code-workspace`.
+Xamarin.Android itself can be developed within
+[Visual Studio Code (VSCode)](https://code.visualstudio.com/).
+There is a workspace included in the repo `Xamarin.Android.code-workspace`.
 The required extensions should be installed when you open the
 workspace.
 
 ## Building Xamarin.Android
 
-Open the `Xamarin.Android.code-workspace` in vscode. Then use the
+Open the `Xamarin.Android.code-workspace` in VSCode. Then use the
 Build Command Pallette (Ctrl+Shift+B in Windows, Cmd+Shift+B on Mac)
 to list the available build commands.
 
-Select "Build All Xamarin.Android". You will then be presented with a
-list of options.
+Select **Build All Xamarin.Android**. You will then be presented with a
+list of options:
 
-* Prepare - Installs the required Dependencies.
-* PrepareExternal - Installs the required Commercial Dependencies (Xamarin Team Members Only)
-* Build - Build Xamarin. Android.
-* Pack - Create the Nuget Packages.
-* Everything - Calls Prepare, Build and Pack.
+* `Prepare` - Installs the required Dependencies.
+* `PrepareExternal` - Installs the required Commercial Dependencies (Xamarin Team Members Only)
+* `Build` - Build Xamarin.Android.
+* `Pack` - Create the NuGet Packages.
+* `Everything` - Calls Prepare, Build and Pack.
 
 The normal order is `Prepare`, `Build` then `Pack`. This will result in
 a usable copy of Xamarin.Android. You can now use it to build apps
 and run the unit tests.
 
-Note: PrepareExternal is for internal Xamarin Team members only, this sets up
+Note: `PrepareExternal` is for internal Xamarin Team members only, this sets up
 the commercial parts of the repository. Trying to use this command when you
 do not have access to the required repositories will result in a failure.
 
 ## Running/Debugging Unit Tests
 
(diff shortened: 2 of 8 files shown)


Documentation/guides/vscode-support.md
```diff
@@ -0,0 +1,86 @@
+# VSCode Support
+
+Xamarin.Android itself can be developed within VSCode. There is
+a workspace included in the repo `Xamarin.Android.code-workspace`.
+The required extensions should be installed when you open the
+workspace.
+
+## Building Xamarin.Android
+
+Follow the instructions [here]() to build Xamarin.Android on the
```
jonpryor, line 10: TODO: url target for `here`?

Documentation/guides/vscode-support.md
```diff
@@ -0,0 +1,86 @@
...
+
+The setting you need to set is `nxunitExplorer.nunit`. This will need
+to be done in your `Preferences->Settings` in VSCode. This will save
+the setting globally so it will be available when ever you open
+Xamarin.Android.
+
+Alternatively you can add something like this to the `Xamarin.Android.code-workspace`.
+
+```json
+"nxunitExplorer.nunit": "/Users/<user>/.nuget/packages/nunit.consolerunner/3.11.1/tools/nunit3-console.exe"
```
jonpryor, line 30: Can [environment variables](https://code.visualstudio.com/docs/editor/variables-reference#_environment-variables) be used?

```json
"nxunitExplorer.nunit": "${env:HOME}/.nuget/packages/nunit.consolerunner/3.11.1/tools/nunit3-console.exe"
```

Assuming that works at all, the problem with it will be that it's specific to Linux & macOS, not Windows.  Though maybe *both* could be used, in a "fallback" affair?

```json
"nxunitExplorer.nunit": "${env:USERPROFILE}${env:HOME}/.nuget/packages/nunit.consolerunner/3.11.1/tools/nunit3-console.exe"
```

Documentation/guides/vscode-support.md
```diff
@@ -0,0 +1,86 @@
...
+
+Once that is setup you can to go the `Testing` tab in VSCode and browse/run/debug
+the unit tests.
+
+Note: There is a slight problem with the nxunitExplorer in that it does not escape
+the names of the unit tests it runs. As a result the unit tests which contain
+arguments do not run within VSCode. However a work around is to run the top
+level fixture rather than the single test.
+
+Alternatively you can download the patched executable from [here](https://github.com/dellis1972/vscode-nxunit-test-adapter/releases/tag/Patch2) and replace
```
jonpryor, line 44: May want to mention the upstream PR? https://github.com/prash-wghats/vscode-nxunit-test-adapter/pull/15

Though it's somewhat disturbing that prash-wghats/vscode-nxunit-test-adapter hasn't been updated since *2019*.

This is a strong sign that a different package should be suggested.  You want to add yours? ;-)

Documentation/guides/vscode-support.md
```diff
@@ -0,0 +1,86 @@
...
+### Running the Sample
+
+Xamarin.Android provides a few Debug configurations in VSCode to debug the
+sample applications `HelloWorld` and `VSAndroidApp`. You can select the
+`Debug Sample` Run and Debug config to run and debug the samples under
+`mono` or you can use `Debug Sample Under DotNet` to debug this under the
+.net 6 system.
+
+In order to use the `Debug Sample Under DotNet` you will need to have built
+Xamarin.Android for .net 6. See the windows [documentation](../building/windows/instructions.md) or
```
jonpryor, line 57: I think you should "expand" the link text, having `[windows documentation](…)` and `[\*nix documentation](…)`.

where.cmd
```diff
@@ -0,0 +1,2 @@
+@echo off
+CALL "C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -property installationPath
```
dellis1972, line 2: @jonathanpeppers So I took your advice, this seems to work nicely. Had to put it in a seperate .cmd file though as the
`IN ('')` cmd in the `build-all.cmd` file error when we put this directly in the `IN` command. 
I've tested it on my VM and it works nicely.

jonathanpeppers (commented):
Whoops I left some comments pending yesterday...

image

jonathanpeppers:
Only other question, is should we use `build.cmd/sh` on CI? Is there a step we could replace to call them.

.vscode/extensions.json
```diff
@@ -1,10 +1,11 @@
 {
     "recommendations": [
-      "ms-vscode.csharp",
+      "ms-dotnettools.csharp",
       "ms-vscode.cpptools",
       "ms-vscode.mono-debug",
-      "wghats.vscode-nxunit-test-adapter",
       "visualstudioexptteam.vscodeintellicode",
       "streetsidesoftware.code-spell-checker",
+      "wghats.vscode-nxunit-test-adapter",
```
jonpryor, line 8: Why reorder this?  Is order important?

.vscode/extensions.json
```diff
@@ -1,10 +1,11 @@
...
+      "derivitec-ltd.vscode-dotnet-adapter",
```
jonpryor, line 9: Are trailing commas allowed in `extensions.json`?

.vscode/extensions.json
dellis1972: Order was not important. I just decided to put all the unit test adapters at the bottom.

.vscode/extensions.json
dellis1972: It seems they are cos it works for me 🤷

PR #5991, opened 2021-06-03
Bump to xamarin/monodroid/main@1f2ce156

jonathanpeppers:
Changes: https://github.com/xamarin/monodroid/compare/76c04cd1...1f2ce156

* [tools/msbuild] Missing translations for XA0135
* Bump to xamarin/xamarin-android@032d840, xamarin/androidtools@355d015
* [build] fix dotnet tool install command
* [tools/fastdev] Rework Unix timestamp calculation code in xamarin.find
* [tools/fastdev] Add error checking when writing data to disk.
* [tools/msbuild] <GetPrimaryCpuAbi/> selects backup RIDs for .NET 6
* Bump android-sdk-installer to use Mono.Unix
* [tools/msbuild] only run _GetPrimaryCpuAbi for Fast Dev

Other changes:

* `InstantRunNativeLibrary` will need to set the ABI in the `.csproj`
  in order for the test to pass on .NET 6. It's use of
  `BaseTest.DeviceAbi` would not match for an x86_64 emulator.

### DIFF
--- a/.external
+++ b/.external
@@ -1,2 +1,2 @@
-xamarin/monodroid:main@76c04cd15eca7afc269a6d26296e9d2db6f79be2
+xamarin/monodroid:main@1f2ce156245ef1bf63ec8014882d283b3224216b
 mono/mono:2020-02@c633fe923832f0c3db3c4e6aa98e5592bf5a06e7
--- a/tests/MSBuildDeviceIntegration/Tests/InstantRunTest.cs
+++ b/tests/MSBuildDeviceIntegration/Tests/InstantRunTest.cs
@@ -297,6 +297,10 @@ namespace Xamarin.Android.Build.Tests
 					nativeLib,
 				},
 			};
+			if (Builder.UseDotNet) {
+				//NOTE: in .NET 6 by default an x86_64 emulator would fall back to x86 if we don't set this.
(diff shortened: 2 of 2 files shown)


jonathanpeppers:
Maybe don't merge this one yet, I found an issue testing this:
```
NotifySync SkipCopyFile obj\Debug\net6.0-android\android\assets\HelloAndroid.dll=>HelloAndroid.dll file is up to date. [0ms]
NotifySync CopyFile obj\Debug\net6.0-android\android\assets\HelloAndroid.pdb. [17ms]
```
I added a line like this to my `MainActivity.OnCreate()`:
```
System.Console.WriteLine("WOOT!");
```
But I didn't get the change.

jonathanpeppers:
The issue in https://github.com/xamarin/xamarin-android/pull/5991#issuecomment-854216388 should be working now.

I believe I was only seeing the problem in an earlier incarnation of https://github.com/xamarin/monodroid/pull/1205.

jonathanpeppers:
Latest issue here, is we should only be doing this block when Fast Deployment is enabled...

https://github.com/xamarin/monodroid/blob/48843fcb2c224fb2d5d2e65299c25abeb592445a/tools/msbuild/Xamarin.Android.Common.Debugging.targets#L408-L413

PR #5993, opened 2021-06-04
Try to detect when the emulator is not working

dellis1972:
We have seen in the past issues where the emulator seems to disappear during Unit testing. 
I don't have access to the error message which we got since those old builds have since been removed. 

Our existing setup checked that we had a running emulator when the test fixture was started. So this check
only happened once. If anything happens to the emulator after that check the unit tests would still think
that is was running. Also there is no recovery mechanism. All the code to start the emulator is handled by 
the pipelines build scripts and the `AcquireAndroidTarget`. 

So we need to rethink this a bit. The pipelines will still call `AcquireAndroidTarget` to start up the emulator 
before all the tests. But we need to introduce some additional checks to make sure the emulator is functioning
and has not crashed during the test run. The introduction of a new `CheckDevice` method which will run
BEFORE every device test will do a simple `echo OK` to the adb shell. This will allow us to check to see if 
the emulator is responding. If that fails there is a new `RestartDevice` method which will make use of the 
`AcquireAndroidTarget` to make sure the emulator starts up.  It does this by making use of a simple
`Emulator.csproj` which is just there to import the appropriate targets. We can then use the existing unit
test `Builder` infrastructure to run the required target and start the emulator. 

This may well increase our build times since we need to do the additional checks per unit test, but should
make the overall system more reliable.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
@@ -17,12 +17,49 @@ namespace Xamarin.Android.Build.Tests
 	{
 		public const string GuestUserName = "guest1";
 
+		protected bool HasDevices {
+			get {
+				string output = RunAdbCommand ("shell echo OK");
+				return output.Contains ("OK");
+			}
+		}
+
+		/// <summary>
+		/// Checks if there is a device available
+		/// * Defaults to Assert.Fail ()
+		/// </summary>
+		public void AssertHasDevices (bool fail = true)
+		{
+			if (!HasDevices) {
+				var message = "This test requires an attached device or emulator.";
+				if (fail) {
+					Assert.Fail (message);
+				} else {
+					Assert.Ignore (message);
+				}
+			}
+		}
+
+		protected string DeviceAbi => SetUp.DeviceAbi;
+
+		protected int DeviceSdkVersion => SetUp.DeviceSdkVersion;
+
 		[OneTimeSetUp]
 		public void DeviceSetup ()
 		{
 			SetAdbLogcatBufferSize (64);
 			RunAdbCommand ("logcat -c");
 			CreateGuestUser (GuestUserName);
+		}
+
+		[SetUp]
+		public void CheckDevice ()
+		{
+			if (!HasDevices) {
+				// something went wrong with the emulator.
+				// lets restart it.
+				RestartDevice (Path.Combine (Root, "Emulator.csproj"));
+			}
 		}
 
 		[TearDown]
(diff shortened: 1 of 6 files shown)


dellis1972:
Well this didn't work :'(

jonathanpeppers (approved):
This is green!

jonpryor:
TODO: commit message. :-)

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
```diff
@@ -61,6 +98,16 @@ protected static void UnInstallTestApp ()
 			}
  		}
 
+		public static void RestartDevice (string project)
+		{
+			TestContext.Out.WriteLine ($"Trying to restart Emulator");
+			// shell out to msbuild and start the emulator again
+			using (var builder = new Builder ()) {
+				var out1 = RunProcessWithExitCode (builder.BuildTool, $"{project} /restore /t:AcquireAndroidTarget", timeoutInSeconds: 120);
```
jonpryor, line 106: @dellis1972 : should `{project}` be *quoted* in case it's in a directory that contains spaces? e.g. `$"\"{project}\" /restore …"`?

jonpryor:
Squash-and-merge for review:

Summary:

```
[ci] Add emulator checks; Bump to xamarin/monodroid/main@33e542fa (#5993)
```

Body:

```markdown
Changes: https://github.com/xamarin/monodroid/compare/36536006a74d9fa9b08b14301f73b2e7da93d75d...33e542fac95f1ac4e56d3965d2f336a1d6a5f98e

  * xamarin/monodroid@33e542fac: [tools/fastdev] More Diagnostic logging. (#1212)
  * xamarin/monodroid@a0e1eb4d9: [tools/msbuild] Change `AndroidFastDeploymentToolVersion` to use `XamarinAndroidVersion` (#1210)

We have historically observed seen in the emulator "disappear" during
unit test execution, which causes "spurious" test failures.
(I don't have access to the error message which we got since those
old builds have since been removed.)

Our existing setup checked that we had a running emulator when the
test fixture was started.  This check only happened once.  If anything
happened to the emulator after that check, then the unit tests would
still think that is was running.  There was also no recovery mechanism.
All the code to start the emulator is handled by the pipelines build
scripts and the `AcquireAndroidTarget` target.

Rethink this a bit.  The pipelines will still call the
`AcquireAndroidTarget` MSBuild target to start up the emulator before
all the tests, but we need to introduce some additional checks to
ensure that the emulator is functioning and has not crashed during a
prior test run.

Add a new `DeviceTest.CheckDevice()` method which runs BEFORE every
device test, which does a simple `adb shell echo OK` to the adb shell.
This will allow us to check to see if the emulator is responding.
If that fails, there is a new `DeviceTest.RestartDevice()` method
which will make use of the  `AcquireAndroidTarget` MSBuild target to
ensure the emulator starts up.  It does this by making use of a simple
`Emulator.csproj` which is just there to import the appropriate
targets.  We can then use the existing unit test `Builder`
infrastructure to run the required target and start the emulator. 

This may well increase our build times since we need to do the
additional checks per unit test, but should make the overall system
more reliable.
```

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/DeviceTest.cs
dellis1972: Well its not a path. We just pass the actual name of the project because it is in the same directory, so I don't think we need to

jonpryor: This test doesn't create a project filename that contains spaces? https://github.com/xamarin/xamarin-android/blob/054a00c27563835d82d8b9eefc3c9dfdb8e1472e/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs#L2498

Though that test isn't an on-device test, so it's not specifically relevant, just "example" relevant…

PR #6009, opened 2021-06-07
[One .NET] *.runtimeconfig.json support

jonathanpeppers:
WIP, I implemented the MSBuild side partially.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
@@ -151,7 +151,7 @@ namespace Xamarin.Android.Build.Tests
 						ret.jni_add_native_method_registration_attribute_present = ConvertFieldToBool ("jni_add_native_method_registration_attribute_present", envFile, i, field [1]);
 						break;
 
-					case 7:
+					case 7: // have_runtime_config_blob: bool / .byte
 						AssertFieldType (envFile, ".byte", field [0], i);
 						ret.have_runtime_config_blob = ConvertFieldToBool ("have_runtime_config_blob", envFile, i, field [1]);
 						break;
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
@@ -25,13 +25,14 @@ namespace Xamarin.Android.Build.Tests
 			public bool   broken_exception_transitions;
 			public bool   instant_run_enabled;
 			public bool   jni_add_native_method_registration_attribute_present;
+			public bool   have_runtime_config_blob;
 			public byte   bound_stream_io_exception_type;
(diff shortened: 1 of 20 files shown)


dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
```diff
@@ -150,27 +151,32 @@ static ApplicationConfig ReadApplicationConfig (string envFile)
 						ret.jni_add_native_method_registration_attribute_present = ConvertFieldToBool ("jni_add_native_method_registration_attribute_present", envFile, i, field [1]);
 						break;
 
-					case 7: // bound_stream_io_exception_type: byte / .byte
+					case 7:
```
jonpryor, line 154: Two questions:

1. Elsewhere, there's a comment on the `case` line saying what field it's for.  It's missing here.  Should there be a comment?
2. Why insert as "field #7" instead of appending as "field #12"?

jonpryor: Why insert as "field #7"? So that all the `bool` fields are together in e.g. https://github.com/xamarin/xamarin-android/pull/6009/files#diff-920537864b4d4fbcacfad5ebfdcfd73470855176b83f4c4592d604a7057b7d3dR108

jonathanpeppers:
The new tests here look OK:
image

It looks like something is going on with some build machines, where they have a newer Mono:
```
Could not load file or assembly 'Microsoft.NET.StringTools, Version=1.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies.
```

jonpryor:
Commit message:

```markdown
Context: https://github.com/dotnet/runtime/blob/d95bfea59ed7d19e9b1db096c9d332005989296b/docs/design/mono/mobile-runtimeconfig-json.md

.NET allows including a [`runtimeconfig.json`][0] file in an
application, which can be used to control runtime configuration
options, provide "backing data" to [`AppContext.GetData()`][1],
control configuration values, and more.

Update the Xamarin.Android-for-.NET 6 build process to call the new
[`<RuntimeConfigParser/>` task][2], which will process the
`runtimeconfig.json` files into an efficient binary blob, which will
be added to the `.apk` as `assemblies/rc.bin`.

During process startup, `assemblies/rc.bin` will be proved to
`monovm_runtimeconfig_initialize()`, which will parse `rc.bin`.

Certain `runtimeconfig.json` properties *will not* be emitted into
`rc.bin`, as these properties are required for proper app operation,
such as `PINVOKE_OVERRIDE` (cf84e1b9, 0cd890bd).
The `@(_RuntimeConfigReservedProperties)` item group lists these
properties; the `<RuntimeConfigParser/>` will skip them when
generating `rc.bin`.

Co-authored-by: Marek Habersack <grendel@twistedcode.net>

[0]: https://docs.microsoft.com/en-us/dotnet/core/run-time-config/#runtimeconfigjson
[1]: https://docs.microsoft.com/en-us/dotnet/api/system.appcontext.getdata?view=net-5.0
[2]: https://github.com/dotnet/runtime/blob/01b7e73cd378145264a7cb7a09365b41ed42b240/src/tasks/RuntimeConfigParser/RuntimeConfigParser.cs
```

PR #6022, opened 2021-06-14
Add EventPipe tracing support

grendello:
`EventPipe` is a dotnet mechanism to implement profiling/tracing of
.NET5+ applications.  It replaces the Mono logging profiler which no
longer works.

This commit adds support of packaging the Mono components required for
EventPipe to work in the application APK as well as runtime support for
initializing and enabling the profiler.

Profiler is enabled by setting the `debug.mono.profile` Android system
property to the appropriate value (e.g. `127.0.0.1:9000,suspend`) which
is then exported by the XA runtime as an environment
variable (`DOTNET_DiagnosticPorts`) which, in turn, is read by the
EventPipe component in order to start profiling.
    
Details of the process are documented in the
`Documentation/guides/tracing.md` file.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
@@ -169,7 +169,7 @@ _ResolveAssemblies MSBuild target.
       <_ResolvedNativeLibraries Include="@(ResolvedFileToPublish)" Condition=" '%(ResolvedFileToPublish.Extension)' == '.so' " />
     </ItemGroup>
     <ItemGroup>
-      <_MonoComponent Condition=" '$(AndroidEnableProfiler)' != '' And '$(AndroidEnableProfiler)' != 'false' " Include="diagnostics_tracing" />
+      <_MonoComponent Condition=" '$(AndroidEnableProfiler)' == 'true' " Include="diagnostics_tracing" />
       <_MonoComponent Condition=" '$(AndroidUseInterpreter)' == 'true' " Include="hot_reload" />
     </ItemGroup>
     <ProcessNativeLibraries
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
@@ -168,8 +168,13 @@ _ResolveAssemblies MSBuild target.
     <ItemGroup>
       <_ResolvedNativeLibraries Include="@(ResolvedFileToPublish)" Condition=" '%(ResolvedFileToPublish.Extension)' == '.so' " />
     </ItemGroup>
+    <ItemGroup>
+      <_MonoComponent Condition=" '$(AndroidEnableProfiler)' != '' And '$(AndroidEnableProfiler)' != 'false' " Include="diagnostics_tracing" />
+      <_MonoComponent Condition=" '$(AndroidUseInterpreter)' == 'true' " Include="hot_reload" />
+    </ItemGroup>
     <ProcessNativeLibraries
         InputLibraries="@(_ResolvedNativeLibraries)"
+        Components="@(_MonoComponent->Distinct())"
         IncludeDebugSymbols="$(AndroidIncludeDebugSymbols)">
       <Output TaskParameter="OutputLibraries" ItemName="FrameworkNativeLibrary" />
     </ProcessNativeLibraries>
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
@@ -86,7 +86,7 @@
     <VSAndroidDesigner Condition="'$(VSAndroidDesigner)' == ''">false</VSAndroidDesigner>
 
     <!-- profiler won't work without internet permission, we must thus force it -->
-    <AndroidNeedsInternetPermission Condition=" '$(AndroidEnableProfiler)' != '' And '$(AndroidEnableProfiler)' != 'false' ">True</AndroidNeedsInternetPermission>
+    <AndroidNeedsInternetPermission Condition=" '$(AndroidEnableProfiler)' == 'true' ">True</AndroidNeedsInternetPermission>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(AndroidApplication)' == 'true' and '$(GenerateApplicationManifest)' == 'true' ">
     <ApplicationVersion Condition=" '$(ApplicationVersion)' == '' ">1.0</ApplicationVersion>
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
@@ -25,6 +25,9 @@
     -->
     <_GetChildProjectCopyToPublishDirectoryItems>false</_GetChildProjectCopyToPublishDirectoryItems>
     <UseMonoRuntime Condition=" '$(UseMonoRuntime)' == '' ">true</UseMonoRuntime>
+
+    <!-- Mono components -->
+    <AndroidEnableProfiler Condition=" '$(AndroidEnableProfiler)' == ''">false</AndroidEnableProfiler>
   </PropertyGroup>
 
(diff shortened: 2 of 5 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
```diff
@@ -14,6 +14,12 @@ _ResolveAssemblies MSBuild target.
   <UsingTask TaskName="Xamarin.Android.Tasks.ProcessNativeLibraries" AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
   <UsingTask TaskName="Xamarin.Android.Tasks.StripNativeLibraries" AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
 
+  <PropertyGroup>
+    <AndroidEnableProfiler Condition=" '$(AndroidEnableProfiler)' == ''">false</AndroidEnableProfiler>
+    <AndroidEnableHotReload Condition=" '$(AndroidEnableHotReload)' == '' and '$(Configuration)' == 'Debug' ">true</AndroidEnableHotReload>
```
dellis1972, line 19: now that I think about it do we really need a 'true` value for this? If this is controlled by the IDE then it will be responsible for passing in the value right? I which case we should just always default to `false` if `AndroidEnableHotReload` is not already set.

src/Xamarin.Android.Build.Tasks/Tasks/ProcessNativeLibraries.cs
```diff
@@ -13,19 +13,25 @@ namespace Xamarin.Android.Tasks
 	/// </summary>
 	public class ProcessNativeLibraries : AndroidTask
 	{
+		const string MonoComponentPrefix = "libmono-component-";
+
 		public override string TaskPrefix => "PRNL";
 
-		static readonly HashSet<string> DebugNativeLibraries = new HashSet<string> {
+		static readonly HashSet<string> DebugNativeLibraries = new HashSet<string> (StringComparer.OrdinalIgnoreCase) {
```
dellis1972, line 20: We can probably remove this if we use the `RemoveDuplicates ` task on the ItemGroup before we pass it in.

https://github.com/xamarin/monodroid/blob/main/tools/msbuild/Xamarin.Android.Common.Debugging.targets#L600

https://docs.microsoft.com/en-us/visualstudio/msbuild/removeduplicates-task?view=vs-2019

grendello: This particular set is used to exclude debug native libraries from the Release builds, so I think we should leave this in

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
grendello: Is hot reload usable outside the IDE?

dellis1972: I dunno something to ask. It is a Dev Loop feature so I would guess it kinda needs an IDE.

src/Xamarin.Android.Build.Tasks/Tasks/ProcessNativeLibraries.cs
dellis1972: Sorry I meant the `_MonoComponents` one.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
jonathanpeppers: I'd also move these settings to `Microsoft.Android.Sdk.DefaultProperties.targets`.

src/Xamarin.Android.Build.Tasks/Tasks/ProcessNativeLibraries.cs
jonathanpeppers: You can also do `@(_MonoComponents->Distinct())` when passing it in to a task. This should also preserve the ordering if that matters.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -26,6 +26,10 @@
     -->
     <_GetChildProjectCopyToPublishDirectoryItems>false</_GetChildProjectCopyToPublishDirectoryItems>
     <UseMonoRuntime Condition=" '$(UseMonoRuntime)' == '' ">true</UseMonoRuntime>
+
+    <!-- Mono components -->
+    <AndroidEnableProfiler Condition=" '$(AndroidEnableProfiler)' == ''">false</AndroidEnableProfiler>
```
jonpryor, line 31: The `$(AndroidIncludeDebugSymbols)`  initialization should also be updated to take `$(AndroidEnableProfiler)` into consideration: https://github.com/xamarin/xamarin-android/blob/7e9f595f23ebd5b53385b16e3890bc90f95f6fef/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L302-L323

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
```diff
@@ -168,8 +168,13 @@ _ResolveAssemblies MSBuild target.
     <ItemGroup>
       <_ResolvedNativeLibraries Include="@(ResolvedFileToPublish)" Condition=" '%(ResolvedFileToPublish.Extension)' == '.so' " />
     </ItemGroup>
+    <ItemGroup>
+      <_MonoComponent Condition=" '$(AndroidEnableProfiler)' != 'false' " Include="diagnostics_tracing" />
+      <_MonoComponent Condition=" '$(AndroidEnableHotReload)' != 'false' " Include="hot_reload" />
```
jonathanpeppers, line 173: Maybe we don't need new MSBuild property for hot reload yet, maybe we can do:
```suggestion
      <_MonoComponent Condition=" '$(AndroidIncludeDebugSymbols)' == 'true' and '$(AndroidUseInterpreter)' == 'true' " Include="hot_reload" />
```
See: https://github.com/xamarin/xamarin-android/pull/6027

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
grendello: What do you mean @jonpryor? The profiler is independent on whether or not the build is a debug one, it can be used in both after all

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
```diff
@@ -168,8 +168,13 @@ _ResolveAssemblies MSBuild target.
     <ItemGroup>
       <_ResolvedNativeLibraries Include="@(ResolvedFileToPublish)" Condition=" '%(ResolvedFileToPublish.Extension)' == '.so' " />
     </ItemGroup>
+    <ItemGroup>
+      <_MonoComponent Condition=" '$(AndroidEnableProfiler)' != '' And '$(AndroidEnableProfiler)' != 'false' " Include="diagnostics_tracing" />
```
jonathanpeppers, line 172: Could simplify this line:
```suggestion
      <_MonoComponent Condition=" '$(AndroidEnableProfiler)' == 'true' " Include="diagnostics_tracing" />
```

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -82,6 +85,8 @@
     <!-- XA feature switches defaults -->
     <VSAndroidDesigner Condition="'$(VSAndroidDesigner)' == ''">false</VSAndroidDesigner>
 
+    <!-- profiler won't work without internet permission, we must thus force it -->
+    <AndroidNeedsInternetPermission Condition=" '$(AndroidEnableProfiler)' != '' And '$(AndroidEnableProfiler)' != 'false' ">True</AndroidNeedsInternetPermission>
```
jonathanpeppers, line 89:
```suggestion
    <AndroidNeedsInternetPermission Condition=" '$(AndroidEnableProfiler)' == 'true' ">True</AndroidNeedsInternetPermission>
```

jonathanpeppers (approved):
Otherwise LGTM, the test failures look the same as what's failing on `main` right now.

dellis1972 (approved):

PR #6025, opened 2021-06-15
[One .NET] normalize case for $(TargetPlatformIdentifier)

jonathanpeppers:
Context: https://github.com/dotnet/designs/blob/1a79ddc4dc8df38f645d2b5517cdfa330747511e/accepted/2020/net5/net5.md#valid-platforms
Context: https://github.com/xamarin/xamarin-macios/commit/e35bb9c47913fc8ef6c23a89a73e219f75daa448

The spec for `$(TargetPlatformIdentifier)` mentions that each workload
should normalize the casing.

xamarin-macios already does this for `iOS`, `macOS`, `tvOS`, etc. We
should normalize to `Android`.

The place I set this is only imported by `WorkloadManifest.targets`
when MSBuild's case insensitive string comparison matches:

    <Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk"

So we can simply set this early on in our .NET 6 `.targets`:

    <TargetPlatformIdentifier>Android</TargetPlatformIdentifier>

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.targets
@@ -2,6 +2,8 @@
 
   <PropertyGroup>
     <MicrosoftAndroidSdkTargetsImported>true</MicrosoftAndroidSdkTargetsImported>
+    <!-- Normalize casing -->
+    <TargetPlatformIdentifier>Android</TargetPlatformIdentifier>
     <_XamarinAndroidBuildTasksAssembly>..\tools\Xamarin.Android.Build.Tasks.dll</_XamarinAndroidBuildTasksAssembly>
     <UsingAndroidNETSdk>true</UsingAndroidNETSdk>
     <!-- Enable nuget package conflict resolution -->


dellis1972 (approved):

jonpryor:
I'm confused about the usage of "normalize" here, especially considering that we're not *doing* anything in this PR.  We're setting a property value.  How does that "normalize" casing?  What (how?) is it set when we don't normalize?  Is the `$(TargetPlatformIdentifier)` value also used in file names somewhere?

dellis1972:
I think this is more about *forcing* the `TargetPlatformIdentifier` to be a particular casing. So that even if a user puts in `android` or `ANDROID` for the `TargetPlatformIdentifier` it will still work? I assume @jonathanpeppers  is that right?

jonathanpeppers:
The [spec](https://github.com/dotnet/designs/blob/1a79ddc4dc8df38f645d2b5517cdfa330747511e/accepted/2020/net5/net5.md#valid-platforms) says:

> Comparison of platform identifiers should be case insensitive, but normalized by the MSBuild logic that supports that platform. So with a `TargetFramework` of `net5.0-android`, NuGet should parse out the platform as `Android`. The core .NET SDK will then set the `TargetPlatformIdentifier` to `android`. The Android workload should then normalize the casing of the `TargetPlatformIdentifier` property to `Android`. Similarly, the iOS Workload and Windows targets should normalize the `TargetPlatformIdentifier` to `iOS` and `Windows`.

When you use `net6.0-android` the value comes through to our targets as `android`, and we just need to replace it with `Android`.

I think they plan on using `$(TargetPlatformIdentifier)` for other user-facing things, and they want the casing to be correct. Things like the IDE might use it?

Issue #5801, opened 2021-03-30
ClientWebSocket.ConnectAsync crashes on .NET 6 


pjcollins:
Our [ClientWebSocket test](https://github.com/xamarin/xamarin-android/blob/7e4c16c86b4877dfde26de24b8fadec10ff125dc/tests/Mono.Android-Tests/System.Net/WebSocketTests.cs#L13) is failing on .NET 6. I'm not sure if this is an issue for dotnet/runtime or us to address, so I'll file it here now for further investigation.

### Steps to Reproduce

```csharp
var clientWebSocket = new ClientWebSocket();
clientWebSocket.ConnectAsync(new Uri("ws://echo.websocket.org"), CancellationToken.None).GetAwaiter().GetResult();
Console.WriteLine("WS State: " + clientWebSocket.State);
```

### Log File

https://gist.githubusercontent.com/pjcollins/6e08f0e3b1681dd7a4fbc93e26443a6a/raw/e084f740cd487776832d0908cf08280f1daae26b/gistfile1.txt

PR #6028, opened 2021-06-16
[One .NET] enable WebSocketTests

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/5801

Web sockets appear to be working in .NET 6 Preview 5 and higher. I
think we can simply enable these tests now.


jonathanpeppers:
Yeah this test is passing now:

image

pjcollins (approved):

PR #6030, opened 2021-06-17
[CI] Add nuget to msi conversion and VS insert stage

pjcollins:
Context: https://github.com/xamarin/yaml-templates/pull/117

A new CI job and stage have been added to facilitate the Visual Studio
installer insertion process.

The first job named "Convert NuGet to MSI" will convert signed packages
to .msi installers, and generate Visual Studio manifests for the .msi
installers.  The .msi files are signed and converted back to .nupkg
files for the Windows command line install scenario.  These packages are
uploaded as pipeline artifacts named `vsdrop-signed`, and 
`vs-msi-nugets`.

The new `vs-workload.props` file contains version information and other
metadata used by the convert job to generate a Visual Studio manifest.

The `ShortNames` item in this file is used to avoid path length issues
that can arise due to the way VS caches some components when building.
The .msi conversion tooling uses this to replace strings in the VS
component file names and IDs generated for our workload packs.  In
some cases the semver label generated for our packs can be long, so
we replace the full semver version with a more predictable four part
version (30.0.100.x).

The new "VS Insertion" stage starts with a [manual validation task][0].
This task will pause execution and wait for someone to click a "Resume"
or "Reject" button that will appear on the pipeline UI.  This task is
configured to be rejected after waiting for two days, but it can be
manually re-ran any time after being rejected to trigger a VS insertion
for an older build.

If the manual validation task is approved, a VS Drop will be created
containing all .NET 6 .msi files.  This Drop URL can then be used to
update our component versions in Visual Studio.  This last piece is
currently manual as we will initially be introducing new components,
however we should be able to automate VS PR creation in the future.

[0]: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/manual-validation?view=azure-devops&tabs=yaml


jonathanpeppers (approved):
The changes here look good, is there a full run somewhere we could look at?

jonathanpeppers:
Ok, I found: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4992206&view=logs&jobId=17cc7b13-5a34-5a4d-2856-05e26b260054

jonathanpeppers (commented):
Should we be creating the arm64 `.msi`'s? I don't think XA works on Windows arm64.

image

pjcollins:
> Should we be creating the arm64 .msi's? I don't think XA works on Windows arm64.

VS doesn't support these either, and they aren't included in the `vsdrop-signed` artifact and `.vsman` manifest file.  The new .nupkg artifact isn't being used anywhere yet, and any changes to provide an option to not generate the arm64 packs should probably land in arcade (and yaml-templates), and can be done in the future I think.

PR #6033, opened 2021-06-21
try to use jdk 11 for device tests

dellis1972:
Since 5da92bd we are seeing the on device tests fail with the following error.

    error JAVAC0000: java.lang.AssertionError: annotationType() :  unrecognized Attribute name MODULE (class com.sun.tools.javac.util.SharedNameTable$NameImpl)

This suggests that AP 31 requires JDK 11 in order to compile.
So lets bump the version which we use for the on device tests for now,
until we can figure out what do actually do.


jonathanpeppers (approved):
I kind of think we should merge this to get CI green.

Then revisit next week what we should do?

PR #6037, opened 2021-06-22
Bump LibZipSharp to 2.0.0-alpha7

grendello:
Context: https://github.com/xamarin/LibZipSharp/commit/fcb9323d9d95b06051f67d8d73ba371a4029722d
Context: https://libzip.org/news/release-1.8.0.html

2.0.0-alpha7 changes:

  * bump libzip version to 1.8.0
  * add support for ZSTD compression
  * properly hide all the native library symbols except for the libzip ones

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -26,7 +26,7 @@
 
   <!-- Common <PackageReference/> versions -->
   <PropertyGroup>
-    <LibZipSharpVersion>2.0.0-alpha6</LibZipSharpVersion>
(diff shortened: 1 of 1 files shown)


dellis1972:
@grendello don't forget we need to do this in the submodules as well xamarin-android-tools I think

grendello:
> @grendello don't forget we need to do this in the submodules as well xamarin-android-tools I think

rats, forgot, thank you for reminding :)

dellis1972 (approved):

grendello:
it seems only half of the pipeline ran? There are usually 28 checks instead of 14

PR #6045, opened 2021-06-23
[One .NET] dotnet workload install android or android-aot

jonathanpeppers:
Context: https://github.com/dotnet/maui/tree/ef4518bf502ffa6760625c4cb53fa0cb015bc6d1/src/Workload#net-maui-workload-ids

Right now to install the Android workload you would run:

    dotnet workload install microsoft-android-sdk-full

Let's shorten this to be:

    dotnet workload install android

We can also create a new workload that extends `android`, and includes
AOT support:

    dotnet workload install android-aot

To test this behavior on our CI, I reworked the `ExtractWorkloadPacks`
to run `dotnet workload install` commands. This should give us more
confidence that our workload can be installed -- and our MSBuild tests
will consume the installed workload.

Other changes:

* Added our `Microsoft.Android.Runtime.*` packs to the
  `WorkloadManifest.json`. This was mostly omitted by mistake;
  however, runtime packs aren't resolved either due to:
  https://github.com/dotnet/sdk/issues/14044

* Call `setup-test-environment.yaml` in `dotnet_create_msi` job

* Delete `dotnet/metadata` folder in `Step_InstallDotNetPreview`

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1397,10 +1397,9 @@ stages:
     - ${{ if eq(variables['MicroBuildSignType'], 'Real') }}:
       - group: Publish-Build-Assets
     steps:
-    - checkout: self
-      submodules: recursive
-
-    - template: yaml-templates/use-dot-net.yaml
+    - template: yaml-templates/setup-test-environment.yaml
+      parameters:
+        configuration: $(XA.Build.Configuration)
 
     - task: DownloadPipelineArtifact@2
       inputs:
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1291,19 +1291,12 @@ stages:
     workspace:
       clean: all
     steps:
-    - checkout: self
-      submodules: recursive
+    - template: yaml-templates/setup-test-environment.yaml
+      parameters:
+        xaSourcePath: $(System.DefaultWorkingDirectory)/xamarin-android
 
     - checkout: release_scripts
       clean: true
-
-    - template: yaml-templates/use-dot-net.yaml
-      parameters:
-        remove_dotnet: true
-
-    - template: yaml-templates/use-dot-net.yaml
-      parameters:
-        version: $(DotNetCoreVersion)
 
     - task: DownloadPipelineArtifact@2
       inputs:
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
@@ -1,15 +1,23 @@
 {
   "version": "@WORKLOAD_VERSION@",
   "workloads": {
-    "microsoft-android-sdk-full": {
-      "description": "Android SDK",
+    "android": {
+      "description": "Microsoft Android SDK for .NET",
       "packs": [
         "Microsoft.Android.Sdk",
         "Microsoft.Android.Sdk.BundleTool",
         "Microsoft.Android.Ref",
+        "Microsoft.Android.Runtime.android-arm",
+        "Microsoft.Android.Runtime.android-arm64",
+        "Microsoft.Android.Runtime.android-x86",
+        "Microsoft.Android.Runtime.android-x64",
         "Microsoft.Android.Templates"
       ],
       "extends" : [ "microsoft-net-runtime-android" ]
+    },
+    "android-aot": {
+      "description": "Microsoft Android SDK for .NET with AOT support",
+      "extends" : [ "android", "microsoft-net-runtime-android-aot" ]
     }
   },
   "packs": {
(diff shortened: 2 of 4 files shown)


mhutch:
I know it's in another manifest, but resumably `microsoft-net-runtime-android-aot` should be renamed to `android-aot` as well?

jonathanpeppers:
@mhutch `microsoft-net-runtime-android-aot` is abstract, so users wouldn't see it:

https://github.com/dotnet/runtime/blob/333c4e72b83a08288db929ac56c0e336c936b655/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.json.in#L29-L30

I can add an `android-aot` workload that extends it? Then we can put a description for it in this repo as well.

grendello (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
```diff
@@ -1,15 +1,29 @@
...
+        "Microsoft.Android.Runtime.android-x86",
+        "Microsoft.Android.Runtime.android-x64",
         "Microsoft.Android.Templates"
       ],
       "extends" : [ "microsoft-net-runtime-android" ]
+    },
+    "android-full": {
+      "description": "Microsoft Android SDK for .NET with AOT support",
+      "packs": [
+        "Microsoft.Android.Sdk.BundleTool"
```
dellis1972, line 20: @jonathanpeppers so my guess is that bundle tool will not be installed by default. But if a user wants to deploy to google store they will need it? Do they need to do anything manually to install bundle tool or will it be installed automatically when they try to use it? (I assume this is the google bundle tool thingy).

dellis1972: The other thing is is this right to bundle this with aot? So it you want to deploy to google store, you need to download the entire AOT system as well regardless of it you use it or not?

jonathanpeppers: We could make an `android-bundletool` workload, so then if you had `<AndroidPackageFormat>aab</AndroidPackageFormat>` the error would say:

```
error NETSDK1147: To build this project, the following workloads must be installed: android-bundletool
error NETSDK1147: To install these workloads, run the following command: dotnet workload install android-bundletool
```

I split out this pack originally because `bundletool.jar` is 25 MB by itself:

https://github.com/google/bundletool/releases

dellis1972: I'm just thinking of the most common workflow, which I would suspect is publishing to the google store. 
So given google is making that a requirement soon , I guess my question is should we be splitting out bundle tool at all? 

@jonpryor any thoughts on this one?

pjcollins: Maybe we could go the other way with naming? Perhaps `android` could represent what is currently the `microsoft-android-sdk-full` workload, and we could add a new `android-minimal` that is the most lightweight option we can construct?

build-tools/automation/azure-pipelines.yaml
```diff
@@ -1397,10 +1397,9 @@ stages:
     - ${{ if eq(variables['MicroBuildSignType'], 'Real') }}:
       - group: Publish-Build-Assets
     steps:
-    - checkout: self
-      submodules: recursive
-
-    - template: yaml-templates/use-dot-net.yaml
+    - template: yaml-templates/setup-test-environment.yaml
```
pjcollins, line 1400: Was this only needed to bring in the new `metadata` folder cleaning changes in xaprepare? The extra .nupkg artifact download/extraction, .vsix installation, and Android SDK dependency provisioning added by this template are likely worth skipping if possible. Maybe we could also add the `metadata` folder removal to the `DeleteExtractedWorkloadPacks` target if it's all that is needed?

jonathanpeppers: I thought one build from before `~\android-toolchain\dotnet\` was not provisioned fully, but I can't find it now.

I reverted this change, and we can see what it does.

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
jonathanpeppers: I'll show these options to @jonpryor tomorrow, and hopefully we can get a decision on what to do.

I agree that if bundletool is *required* for Google Play soon, we should just put it in the box. Maybe it wouldn't even need to be its own pack?

build-tools/automation/azure-pipelines.yaml
jonathanpeppers: @pjcollins the error is:
```
 '"C:\Users\AzDevOps\android-toolchain\dotnet\dotnet"' is not recognized as an internal or external command,
```
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4927545&view=logs&j=3ce66085-846b-53dd-5323-3e00eb12b9a6&t=f1788e61-d854-556d-ca50-2c29dae4b93f&l=44

pjcollins: Oh, I overlooked the usage of our android-toolchain dotnet install. I think we'll need to add that install to both the .msi and .pkg creation jobs, since there is nothing that would guarantee that the right version is already in place. Maybe it's worth adding a new  `Scenario_InstallDotNetPreview.cs` scenario to xaprepare that will only run the dotnet preview install step. Or we could continue to use `setup-test-environment.yaml` which is probably easier. I don't think it matters much either way since these jobs should be going away in the not too distant future?

jonathanpeppers:
Discussed with @jonpryor and I think the plan is to have two top-level workloads:

* `android`: what we have today, includes `bundletool`
* `android-aot`: extends `android` and includes AOT support

So the `maui` workload can chose to extend `android` if needed. You can install `android-aot` if you want everything.

dellis1972 (approved):

pjcollins (approved):

PR #6057, opened 2021-06-29
[Xamarin.Android.Build.Tasks] Fix Assembly Ordering in PackageManager

dellis1972:
Context:

  * https://discord.com/channels/732297728826277939/732297965019988138/859449393472471082
  * ...through...
  * https://discord.com/channels/732297728826277939/732297965019988138/859454102999400488

The ordering of `MonoPackageManager_Resources.Assemblies` is very important to process startup, e.g. see 9c043786, wherein we decide that we *must* preload the *first* assembly listed within `MonoPackageManager_Resources.Assemblies`, as that's The One that has the correct resource IDs.

The problem is threefold:

 1. The order of `@(_ResolvedUserAssemblies)` is poorly defined/undefined.
 2. `@(_ResolvedUserAssemblies)` contains *paths* which includes directory names.
 3. The observed ordering of `@(_ResolvedUserAssemblies)` *differs* between "legacy" MSBuild and `dotnet build`.

See also: https://github.com/xamarin/xamarin-android/blob/52d6880e74337db3fa245ea82b2a02b5016efece/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs#L95-L109

The code in `GeneratePackageManagerJava.cs` line 97 was *always* faulty: `assemblies` contains *paths*, not filenames, and thus `fileNameEq (a.ItemSpec, mainFileName)` will *always* return false (e.g. `a.ItemSpec` would be `obj/Debug/whatever/MainAssembly.dll`, while `mainFileName` will be `MainAssembly.dll`, which will never be equal).

It Just Happened™ to work on Legacy because `@(_ResolvedUserAssemblies)` *always* contained the main/"app" assembly as the first assembly.  This isn't the case on .NET 6 `dotnet build`.

The "giveaway" is the contents of `MonoPackageManager_Resources`:

```java
public class MonoPackageManager_Resources {
        public static String[] Assemblies = new String[]{
                /* We need to ensure that "HelloAndroid.dll" comes first in this list. */
                "Xamarin.AndroidX.Activity.dll",
                // …
        };
}
```

`HelloAndroid.dll` isn't `Xamarin.AndroidX.Activity.dll`.

So let's rework the `GeneratePackageManagerJava` to force the issue and always put the `MainAssembly` first in the list.
We also upgrade the filter to use the filenames rather then the entire path when checking for the main assembly. 

Unit tests have also been added to ensure that the output is what we expect.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
@@ -3,7 +3,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Linq;
 using System.Reflection;
 using System.Text;
 using Microsoft.Build.Framework;
@@ -89,13 +88,7 @@ namespace Xamarin.Android.Tasks
 			var doc = AndroidAppManifest.Load (Manifest, MonoAndroidHelper.SupportedVersions);
 			int minApiVersion = doc.MinSdkVersion == null ? 4 : (int) doc.MinSdkVersion;
 			// We need to include any special assemblies in the Assemblies list
-			var assemblies = ResolvedUserAssemblies
-				.Concat (MonoAndroidHelper.GetFrameworkAssembliesToTreatAsUserAssemblies (ResolvedAssemblies))
-				.ToList ();
 			var mainFileName = Path.GetFileName (MainAssembly);
-			var mainAssembly = new List<ITaskItem> () { new TaskItem (mainFileName) };
-			Func<string,string,bool> fileNameEq = (a,b) => a.Equals (b, StringComparison.OrdinalIgnoreCase);
-			assemblies = mainAssembly.Concat (assemblies.Where (a => !fileNameEq (Path.GetFileName (a.ItemSpec), mainFileName))).ToList ();
 
 			using (var pkgmgr = MemoryStreamPool.Shared.CreateStreamWriter ()) {
 				pkgmgr.WriteLine ("package mono;");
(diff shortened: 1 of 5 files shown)


jonpryor:
Context:

  * https://discord.com/channels/732297728826277939/732297965019988138/859449393472471082
  * ...through...
  * https://discord.com/channels/732297728826277939/732297965019988138/859454102999400488

The ordering of `MonoPackageManager_Resources.Assemblies` is very important to process startup, e.g. see 9c0437866c7308794283b76125483adecdea9067, wherein we decide that we *must* preload the *first* assembly listed within `MonoPackageManager_Resources.Assemblies`, as that's The One that has the correct resource IDs.

The problem is threefold:

 1. The order of `@(_ResolvedUserAssemblies)` is poorly defined/undefined.
 2. `@(_ResolvedUserAssemblies)` contains *paths* which includes directory names.
 3. The observed ordering of `@(_ResolvedUserAssemblies)` *differs* between "legacy" MSBuild and `dotnet build`.

See also: https://github.com/xamarin/xamarin-android/blob/52d6880e74337db3fa245ea82b2a02b5016efece/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs#L95-L109

The code in `GeneratePackageManagerJava.cs` line 97 was *always* faulty: `assemblies` contains *paths*, not filenames, and thus `fileNameEq (a.ItemSpec, mainFileName)` will *always* return false (e.g. `a.ItemSpec` would be `obj/Debug/whatever/MainAssembly.dll`, while `mainFileName` will be `MainAssembly.dll`, which will never be equal).

It Just Happened™ to work on Legacy because `@(_ResolvedUserAssemblies)` *always* contained the main/"app" assembly as the first assembly.  This isn't the case on .NET 6 `dotnet build`.

The "giveaway" is the contents of `MonoPackageManager_Resources`:

```java
public class MonoPackageManager_Resources {
        public static String[] Assemblies = new String[]{
                /* We need to ensure that "HelloAndroid.dll" comes first in this list. */
                "Xamarin.AndroidX.Activity.dll",
                // …
        };
}
```

`HelloAndroid.dll` isn't `Xamarin.AndroidX.Activity.dll`.

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -93,8 +93,9 @@ public override bool RunTask ()
 				.Concat (MonoAndroidHelper.GetFrameworkAssembliesToTreatAsUserAssemblies (ResolvedAssemblies))
 				.ToList ();
 			var mainFileName = Path.GetFileName (MainAssembly);
+			var mainAssembly = new List<ITaskItem> () { new TaskItem (mainFileName) };
 			Func<string,string,bool> fileNameEq = (a,b) => a.Equals (b, StringComparison.OrdinalIgnoreCase);
-			assemblies = assemblies.Where (a => fileNameEq (a.ItemSpec, mainFileName)).Concat (assemblies.Where (a => !fileNameEq (a.ItemSpec, mainFileName))).ToList ();
+			assemblies = mainAssembly.Concat (assemblies.Where (a => !fileNameEq (Path.GetFileName (a.ItemSpec), mainFileName))).ToList ();
```
jonathanpeppers, line 98: Since we're changing this, should we take a moment and remove the System.Linq?

Maybe it could just be two `foreach` loops, that write as they go? Print all the matches for `mainFileName`, then the rest in a second pass?

dellis1972: Good call

jonathanpeppers (approved):

grendello (approved):

PR #6061, opened 2021-06-30
[ci] Use new 1ES Hosted Ubuntu Pools

pjcollins:
Context: https://www.1eswiki.com/wiki/1ES_hosted_AzureDevOps_Agents

We've been tasked with migrating away from our scale set agent pools to
a new 1ES service.  There are a couple of benefits to using 1ES agent
pools.  We should now be able to use customizable versions of the
[Microsoft-hosted agent images][0] which come preloaded with some of our
dependencies.  These images are also updated regularly which should
hopefully reduce maintenance costs.  Overall, the process for creating,
updating, and rolling out image changes to our scale set agent pools
should now be dramatically simplified over [the existing workflow][1].

Our internal and OSS Linux builds will now use the same Ubuntu 20.04
base image, so setup tasks can be shared.  A couple of workarounds
have been added to avoid some open issues in these new images.

[0]: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml
[1]: https://github.com/xamarin/monodroid/wiki/Azure-Pipelines-and-DevOps#windows-scale-set-agent-pool-management

### DIFF
--- a/build-tools/automation/azure-pipelines-oss.yaml
+++ b/build-tools/automation/azure-pipelines-oss.yaml
@@ -163,7 +163,7 @@ stages:
   jobs:
   - job: linux_build_package
     displayName: Linux Build
-    pool: Xamarin-Android-Ubuntu-Public
+    pool: android-public-ubuntu-vmss
     timeoutInMinutes: 180
     cancelTimeoutInMinutes: 5
     workspace:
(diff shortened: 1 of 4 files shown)


pjcollins:
Successful public build here: https://dev.azure.com/xamarin/public/_build/results?buildId=43137&view=logs&s=47748e8d-4396-5896-11b5-4960e3f92372

PR #6064, opened 2021-06-30
[debugger] Componentize the debugger, only link debugger when it's a debug app.

thaystg:
The default value for release app will be false, so debugger will not be available.
The default value for release app will be true, so debugger will be available.

Related to https://github.com/dotnet/runtime/pull/54887

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
@@ -32,9 +32,11 @@
 
   <!--  User-facing configuration-specific defaults -->
   <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+    <AndroidEnableDebugger Condition=" '$(AndroidEnableDebugger)' == ''">true</AndroidEnableDebugger>
     <AndroidLinkMode Condition=" '$(AndroidLinkMode)' == '' ">None</AndroidLinkMode>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <AndroidEnableDebugger Condition=" '$(AndroidEnableDebugger)' == ''">false</AndroidEnableDebugger>
     <EmbedAssembliesIntoApk Condition=" '$(EmbedAssembliesIntoApk)' == '' ">true</EmbedAssembliesIntoApk>
     <AndroidManagedSymbols Condition=" '$(AndroidManagedSymbols)' == '' ">true</AndroidManagedSymbols>
   </PropertyGroup>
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -32,9 +32,11 @@
 
   <!--  User-facing configuration-specific defaults -->
   <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+    <AndroidEnableDebugger Condition=" '$(AndroidEnableDebugger)' == ''">true</AndroidEnableDebugger>
     <AndroidLinkMode Condition=" '$(AndroidLinkMode)' == '' ">None</AndroidLinkMode>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
+    <AndroidEnableDebugger Condition=" '$(AndroidEnableDebugger)' == ''">false</AndroidEnableDebugger>
```
jonathanpeppers, line 39: There is another property `$(AndroidUseDebugRuntime)`:

https://github.com/xamarin/xamarin-android/blob/3b30862b6e511d4d496f1c1c3cd6976d12d403ed/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets#L66-L70

It's based on `$(Optimize)`:

https://github.com/xamarin/xamarin-android/blob/3b30862b6e511d4d496f1c1c3cd6976d12d403ed/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L196-L199

Should we use that instead of making a new property?

thaystg: If you prefer to reuse something I think that we should reuse: AndroidIncludeDebugSymbols

thaystg: What do you think about it?

jonathanpeppers: Yes, `$(AndroidIncludeDebugSymbols)` seems right, because it also looks at `$(DebugType)`:

https://github.com/xamarin/xamarin-android/blob/3b30862b6e511d4d496f1c1c3cd6976d12d403ed/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L302-L323

thaystg: Done :) 
Thanks!

grendello (approved):

Issue #4657, opened 2020-05-06
[.NET 6] Binding projects should default enable C# 8 features


jonpryor:
`Mono.Android.dll` enables some (currently experimental) features which require C#8:

https://github.com/xamarin/xamarin-android/blob/7dab4ee432dcd1e51c52dfe579023fccaf09c471/src/Mono.Android/Mono.Android.targets#L118-L119

Specifically:

  * `--lang-features=nullable-reference-types`: Nullable reference types
  * `--lang-features=default-interface-methods`: Bind Java interface default methods as C# interface default members
  * `--lang-features=nested-interface-types,`: C# interfaces can contain nested types.
  * `--lang-features=interface-constants`: C# interfaces can contain constants.

One of those also binds Java interface static methods as C#8 interface static methods.

We should enable *all* of these by default in .NET 6 projects.

jpobst:
Default is good, but note that some of these create API breaking changes which would hit users migrating their existing bindings to .NET 5.  Each feature will need a `true`/`false` property to disable it, and will need to be documented in a migration guide.

PR #6066, opened 2021-07-01
[Xamarin.Android.Build.Tasks] Set .NET6 defaults for bindings projects.

jpobst:
Fixes: #4657.

With .NET 6 we want to turn on the following binding project enhancements added with C# 8 by default:
- Interface constants
- Interface static methods
- Interface default methods
- Interface nested types

However, each of these features previously had a workaround that we performed so the code could be used by .NET.  This means that users who are updating their existing bindings may find API breaks as we now use the new method instead of the workaround.  Thus each of these needs a way for the user to opt out of the new behavior to preserve API compatibility.

## New MSBuild Properties

- **`<AndroidBoundInterfacesContainConstants>`** - Controls whether constants on interfaces will be supported, or the old workaround of creating a `IMyInterfaceConsts` class will be used.
- **`<AndroidBoundInterfacesContainTypes>`** - Controls whether types nested in interfaces will be supported, or the old workaround of creating a non-nested type like `IMyInterfaceMyNestedClass`.
- **`<AndroidBoundInterfacesContainStaticAndDefaultInterfaceMethods>`** - Controls whether default and static members on interfaces will be supported, or the old workaround of creating a sibling class containing static members like `abstract class MyInterface`.

## .NET 6

For .NET 6, the above properties all default to `true` if not explicitly set by the user.

## Legacy

For legacy, the above properties all default to `false`.  The "legacy" way of enabling these features was a single preview flag **`<_EnableInterfaceMembers>`**.  We will continue to support this flag for legacy only, and it will enable all the above flags, unless a user has explicitly set one of the flags.

## AndroidCodegenTarget

.NET 6 binding projects will only support `XAJavaInterop1`, so other values are ignored.

### DIFF
--- a/Documentation/guides/OneDotNet.md
+++ b/Documentation/guides/OneDotNet.md
@@ -22,37 +22,11 @@ See the [Target Framework Names in .NET 5][net5spec] spec for details.
 
 [net5spec]: https://github.com/dotnet/designs/blob/5e921a9dc8ecce33b3195dcdb6f10ef56ef8b9d7/accepted/2020/net5/net5.md
 
-## Consolidation of binding projects
+## Bindings projects
 
-In .NET 6, there will no longer be a concept of a [binding
-project][binding] as a separate project type. Any of the MSBuild item
-groups or build actions that currently work in binding projects will
-be supported through a .NET 6 Android application or library.
+Changes for bindings projects are documented [here][binding].
 
-For example, a binding library would be identical to a class library:
-
-```xml
-<Project Sdk="Microsoft.NET.Sdk">
-  <PropertyGroup>
-    <TargetFramework>net6.0-android</TargetFramework>
-  </PropertyGroup>
-</Project>
-```
-
-Along with the file structure:
-
-    Transforms/
-        Metadata.xml
-    foo.jar
-
-`Transforms\*.xml` files are automatically included as a
-`@(TransformFile)` item, and `.jar` files are automatically included
-as a `@(AndroidLibrary)` item.
-
-This will bind C# types for the Java types found in `foo.jar` using
-the metadata fixups from `Transforms\Metadata.xml`.
-
-[binding]: https://docs.microsoft.com/xamarin/android/platform/binding-java-library/
+[binding]: OneDotNetBindingProjects.md
 
 ## .NET Configuration Files
 
--- a/Documentation/guides/OneDotNetBindingProjects.md
+++ b/Documentation/guides/OneDotNetBindingProjects.md
@@ -107,7 +107,7 @@ public interface IFoo {
 
 However this means we no longer generate the alternative class that existing code may depend on.
 
-Setting `<EnableBindingInterfaceConstants>false</EnableBindingInterfaceConstants>` will revert to the legacy behavior.
+Setting `<AndroidEnableBindingInterfaceConstants>false</AndroidEnableBindingInterfaceConstants>` will revert to the legacy behavior.
 
 ### Nested Interface Types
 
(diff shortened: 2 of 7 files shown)


dellis1972 (changes requested):
This looks good :) 
We do need to document the three new properties in https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/building-apps/build-properties.md though as they are public. This is in addition to the docs you added here already.

dellis1972 (approved):

jonathanpeppers (approved):

jonpryor:
Time for our favorite past-time, [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding) names!

The names in play: `$(AndroidEnableBindingInterfaceConstants)`, `$(AndroidEnableBindingNestedInterfaceTypes)`, `$(AndroidEnableBindingStaticAndDefaultInterfaceMethods)`.

The semantics are a tad odd with `$(AndroidEnableBindingInterfaceConstants)`, as the implication is that "interface constants" *weren't* previously bound.  But they were! This just changes *where* they're bound, from an "interface alternative" type that isn't the interface, to the interface itself; see also xamarin/java.interop@105d5443.

`$(AndroidEnableBindingStaticAndDefaultInterfaceMethods)` is in a *somewhat* similar boat, in that the "interface alternative" type could contain static methods, yet the name implies that they weren't supported at all.

Thus, I'm not fond of these names.  The actual semantic changes aren't hinted at; it's impossible to know what they actually mean without reading the docs.

I would instead suggest:

  * `$(AndroidBoundInterfacesContainConstants)` instead of `$(AndroidEnableBindingInterfaceConstants)`
  * `$(AndroidBoundInterfacesContainTypes)` instead of `$(AndroidEnableBindingNestedInterfaceTypes)`
  * `$(AndroidBoundInterfacesContainStaticAndDefaultInterfaceMethods)` instead of `$(AndroidEnableBindingStaticAndDefaultInterfaceMethods)`

Though I do wonder if we need *this* much granularity in a "public-facing" manner.  The real question is, are we emitting C#7-compatible interfaces, in which interfaces only contain properties, events and method declarations, or are we emitting C#8-required interfaces, in which interfaces can *also* contain fields, static methods, methods with bodies, nested types, etc.

Thus, should we "unify" them into a single property, e.g. `$(AndroidBoundInterfacesMimicJava)`? ("Mimic Java" because, as far as most are concerned, Java supported these language features before C#, and enabling these features cause our bindings to more closely mirror the original Java declarations.)

jpobst:
Capturing some discussion:
```
3:24 PM] jpobst: i think the name changes are fine, i believe the granularity is helpful if you want to move forward to C# 8, but might need to disable parts for backwards compatbility
[3:44 PM] jonathanpeppers: Users probably wouldn't ever turn these off, would they?
They just get opted into them for .NET 6
[3:44 PM] jpobst: correct, unless they are migrating and need to maintain backwards compat
[3:44 PM] jonathanpeppers: Maybe the other names could just be private (as an escape hatch), and we have a single $(AndroidBoundInterfacesMimicJava) property that covers most cases
[3:51 PM] jpobst: so like for all the migration documentation: https://github.com/xamarin/xamarin-android/pull/6066/files#diff-d3fc71da91b99201132a14e3e1179d3b2fcb4aa49f920793246a2b661fb3beacR73, would we not mention that you can turn off individual features? or would we document the private flags?
[3:52 PM] jonp: indeed, good question.
[3:54 PM] jonp: @jpobst: alternate take: instead of controlling whether e.g. IFoo.Bar is emitted, it should "always" be emitted in a net6 context.  Instead, the question is whether the Foo.Bar "interface alternative" is generated.
[3:54 PM] jonp: could we have it so that both can be done, with Foo.Bar emitted for compat purposes, and IFoo.Bar emitted because It's Good And Proper
[3:54 PM] jpobst: i think each of these features has a different answer for that
[3:55 PM] jpobst: like constants, yes, but nested types, no
[3:55 PM] jonp: hm, indeed
[3:56 PM] jonp: which suggests we do want the fine-grained control
```

PR #6071, opened 2021-07-05
[NET6] Remove as much AppDomain use as possible

grendello:
NET6+ no longer supports multiple AppDomains, there's always only a
single "root" AppDomain.  This allows us to drop bits of code which deal
with obtaining and setting the current AppDomain in various context.

This commit removes almost all uses of AppDomains from the native
Xamarin.Android runtime.  Unfortunately, it requires sometimes messy use
of the C++ preprocessor, but wherever it is possible and not overly
ugly, I try to separate the NET6 and "legacy" code completely, for
easier and less noisy future removal of the latter.

### DIFF
--- a/src/monodroid/jni/osbridge.hh
+++ b/src/monodroid/jni/osbridge.hh
@@ -117,7 +117,9 @@ namespace xamarin::android::internal
 		void initialize_on_onload (JavaVM *vm, JNIEnv *env);
 		void initialize_on_runtime_init (JNIEnv *env, jclass runtimeClass);
 		void add_monodroid_domain (MonoDomain *domain);
+#if !defined (NET6)
 		void remove_monodroid_domain (MonoDomain *domain);
+#endif // ndef NET6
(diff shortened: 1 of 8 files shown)


src/monodroid/jni/osbridge.hh
```diff
@@ -117,7 +117,9 @@ namespace xamarin::android::internal
 		void initialize_on_onload (JavaVM *vm, JNIEnv *env);
 		void initialize_on_runtime_init (JNIEnv *env, jclass runtimeClass);
 		void add_monodroid_domain (MonoDomain *domain);
+#if !defined (NET6)
```
jonpryor, line 120: What this brings to mind is that `Runtime.destroyContexts()` is "bad" ("weird"?), in that it is a Java method which exists "everywhere", but is only implemented on Desktop.  I assume that if anyone tried to invoke this on Android they'd get an exception, as `monodroid-glue-designer.cc` isn't part of the non-Desktop build.

:-(

PR #6073, opened 2021-07-06
Add support for NDK r23.

grendello:
Context: https://github.com/android/ndk/wiki/Changelog-r23#changelog

The most important changes for Xamarin.Android:

NDK r23 removes GNU Binutils and so we need to switch to using our
bundled copy of them.

Upstream changes:

  * Includes Android 12 APIs.
  * Updated LLVM to clang-r416183b, based on LLVM 12 development.
    * [Issue 1047]: Fixes crash when using ASan with the CFI unwinder.
    * [Issue 1096]: Includes support for [Polly]. Enable by adding `-mllvm -polly`
      to your cflags.
    * [Issue 1230]: LLVM's libunwind is now used instead of libgcc for all
      architectures rather than just 32-bit Arm.
    * [Issue 1231]: LLVM's libclang_rt.builtins is now used instead of libgcc.
    * [Issue 1406]: Fixes crash with Neon intrinsic.
  * Vulkan validation layer source and binaries are no longer shipped in the NDK.
    The latest are now posted directly to [GitHub](https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases).
  * Vulkan tools source is also removed, specifically vulkan_wrapper.
    It should be downloaded upstream from [GitHub](https://github.com/KhronosGroup/Vulkan-Tools).
  * The toolchain file (android.toolchain.cmake) is refactored to base on CMake's
    integrated Android support. This new toolchain file will be enabled by default
    for CMake 3.21 and newer. No user side change is expected. But if anything goes
    wrong, please file a bug and set `ANDROID_USE_LEGACY_TOOLCHAIN_FILE=ON` to
    restore the legacy behavior.
      * When using the new behavior (when using CMake 3.21+ and not explicitly
        selecting the legacy toolchain), **default build flags may change**. One
        of the primary goals was to reduce the behavior differences between our
        toolchain and CMake, and CMake's default flags do not always match the
        legacy toolchain file. Most notably, if using `CMAKE_BUILD_TYPE=Release`,
        your optimization type will likely be `-O3` instead of `-O2` or `-Oz`. See
        [Issue 1536] for more information.
  * [Issue 929]: `find_library` now prefers shared libraries from the sysroot over
    static libraries.
  * [Issue 1390]: ndk-build now warns when building a static executable with the
    wrong API level.
  * [Issue 1452]: `NDK_ANALYZE=1` now sets `APP_CLANG_TIDY=true` rather than using
    scan-build. clang-tidy performs all the same checks by default, and scan-build
    was no longer working. See the bug for more details, but no user-side changes
    should be needed.

[Issue 929]: https://github.com/android/ndk/issues/929
[Issue 1047]: https://github.com/android/ndk/issues/1047
[Issue 1096]: https://github.com/android/ndk/issues/1096
[Issue 1230]: https://github.com/android/ndk/issues/1230
[Issue 1231]: https://github.com/android/ndk/issues/1231
[Issue 1390]: https://github.com/android/ndk/issues/1390
[Issue 1406]: https://github.com/android/ndk/issues/1406
[Issue 1452]: https://github.com/android/ndk/issues/1452
[Issue 1536]: https://github.com/android/ndk/issues/1536
[Polly]: https://polly.llvm.org/

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
@@ -42,6 +42,11 @@ namespace Xamarin.Android.Build.Tests
 				private set;
 			}
 
+			public static string OSBinDirectory {
+				get;
+				private set;
+			}
+
 			static SetUp ()
 			{
 #if NETCOREAPP
@@ -54,6 +59,16 @@ namespace Xamarin.Android.Build.Tests
 					CommercialBuildAvailable = File.Exists (Path.Combine (builder.AndroidMSBuildDirectory, "Xamarin.Android.Common.Debugging.targets"));
 					AndroidMSBuildDirectory = builder.AndroidMSBuildDirectory;
 				}
+
+				string osSubdirName;
+				if (TestEnvironment.IsLinux) {
+					osSubdirName = "Linux";
+				} else if (TestEnvironment.IsMacOS) {
+					osSubdirName = "Darwin";
+				} else {
+					osSubdirName = String.Empty;
+				}
+				OSBinDirectory = Path.Combine (AndroidMSBuildDirectory, osSubdirName);
 			}
 
 			[OneTimeSetUp]
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleLegacy.apkdesc
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleLegacy.apkdesc
@@ -2,25 +2,25 @@
   "Comment": null,
   "Entries": {
     "AndroidManifest.xml": {
-      "Size": 2584
+      "Size": 2604
     },
     "assemblies/Java.Interop.dll": {
       "Size": 67759
     },
     "assemblies/Mono.Android.dll": {
-      "Size": 251450
+      "Size": 251399
     },
     "assemblies/mscorlib.dll": {
       "Size": 769324
     },
     "assemblies/System.Core.dll": {
-      "Size": 28186
+      "Size": 28187
     },
     "assemblies/System.dll": {
       "Size": 9171
     },
     "assemblies/UnnamedProject.dll": {
-      "Size": 2867
+      "Size": 2866
     },
     "classes.dex": {
       "Size": 345208
@@ -32,13 +32,13 @@
       "Size": 707024
     },
     "lib/arm64-v8a/libmonodroid.so": {
-      "Size": 282936
+      "Size": 286736
     },
     "lib/arm64-v8a/libmonosgen-2.0.so": {
       "Size": 4037584
     },
     "lib/arm64-v8a/libxa-internal-api.so": {
-      "Size": 65496
+      "Size": 65624
     },
     "lib/arm64-v8a/libxamarin-app.so": {
       "Size": 20032
@@ -74,5 +74,5 @@
       "Size": 1724
     }
   },
-  "PackageSize": 3978964
+  "PackageSize": 3983060
 }
--- a/Documentation/guides/building-apps/build-properties.md
+++ b/Documentation/guides/building-apps/build-properties.md
@@ -129,14 +129,14 @@ Added in Xamarin.Android 6.1.
 
 A path to a directory containing
 the Android [binutils][binutils] such as `ld`, the native linker,
-and `as`, the native assembler. These tools are part of the Android
-NDK and are also included in the Xamarin.Android installation.
+and `as`, the native assembler. These tools are included in the
+Xamarin.Android installation.
 
-The default value is `$(MonoAndroidBinDirectory)\ndk\`.
+The default value is `$(MonoAndroidBinDirectory)\binutils\`.
 
 Added in Xamarin.Android 10.0.
 
-[binutils]: https://android.googlesource.com/toolchain/binutils/
+[binutils]: https://github.com/xamarin/xamarin-android-binutils/
 
 ## AndroidBoundExceptionType
 
--- a/build-tools/installers/create-installers.targets
+++ b/build-tools/installers/create-installers.targets
@@ -310,18 +310,18 @@
     <_MSBuildFilesWin Include="$(MSBuildSrcDir)\lib64\libZipSharpNative.pdb" />
     <_MSBuildFilesWin Include="$(MSBuildSrcDir)\proguard\bin\proguard.bat" />
     <_MSBuildFilesWin Include="$(MSBuildSrcDir)\aapt2.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\aarch64-linux-android-as.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\aarch64-linux-android-ld.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\aarch64-linux-android-strip.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\arm-linux-androideabi-as.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\arm-linux-androideabi-ld.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\arm-linux-androideabi-strip.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\i686-linux-android-as.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\i686-linux-android-ld.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\i686-linux-android-strip.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\x86_64-linux-android-as.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\x86_64-linux-android-ld.exe" />
-    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\ndk\x86_64-linux-android-strip.exe" />
+    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\binutils\aarch64-linux-android-as.exe" />
+    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\binutils\aarch64-linux-android-ld.exe" />
+    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\binutils\aarch64-linux-android-strip.exe" />
+    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\binutils\arm-linux-androideabi-as.exe" />
+    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\binutils\arm-linux-androideabi-ld.exe" />
+    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\binutils\arm-linux-androideabi-strip.exe" />
+    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\binutils\i686-linux-android-as.exe" />
+    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\binutils\i686-linux-android-ld.exe" />
+    <_MSBuildFilesWin Include="$(MSBuildSrcDir)\binutils\i686-linux-android-strip.exe" />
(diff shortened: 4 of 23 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleLegacy.apkdesc
```diff
@@ -4,75 +4,75 @@
...
-    "lib/arm64-v8a/libmonosgen-2.0.so": {
-      "Size": 4050176
-    },
     "lib/arm64-v8a/libmono-native.so": {
       "Size": 707024
     },
-    "META-INF/ANDROIDD.SF": {
-      "Size": 2225
+    "lib/arm64-v8a/libmonodroid.so": {
+      "Size": 281352
```
jonpryor, line 35: Why did `libmonodroid.so` increase in size by ~12.5KB?  Are there newer/better/more "security checks" in the generated code?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleLegacy.apkdesc
```diff
@@ -4,75 +4,75 @@
...
+    },
+    "lib/arm64-v8a/libmonosgen-2.0.so": {
+      "Size": 4037584
```
jonpryor, line 38: For that matter, how did `libmonosgen-2.0.so` *shrink* by ~12KB?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleLegacy.apkdesc
grendello: It's hard to tell. We don't fully strip the DSOs, perhaps a newer format of debug/symbol info increased the size. It might also have to do with a newer version of libc++ and the bits we use from it.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleLegacy.apkdesc
grendello: `libmonosgen-2.0.so` is not built by this NDK. Runtime uses its own version of it (I think it's 19c?)

jonathanpeppers: A past dotnet/installer bump could have made `libmonosgen-2.0.so` smaller, but wasn't enough to fail the test.

jonpryor: particularly since the test really only cares about the overall `.apk` size, not the sizes of entries within the `.apk`, iirc.

jonathanpeppers: It can fail if a single file changes enough: https://github.com/xamarin/xamarin-android/pull/6118#discussion_r682906288

```
Error: apkdiff: File 'assemblies/UnnamedProject.dll' has changed by 358 bytes (10.14 %). This exceeds the threshold of 5.00 %.
```

jonpryor:
```markdown
[build] Add support for NDK r23. (#6073)

Context: https://github.com/android/ndk/wiki/Changelog-r23#changelog

The most important NDK r23 changes for Xamarin.Android:

NDK r23 removes GNU Binutils.  We need to switch to using our bundled
copy of them, as per 27967cab.

Upstream NDK r23changes:

  * Includes Android 12/API-31 APIs.

  * Updated LLVM to clang-r416183b, based on LLVM 12 development.
    * [Issue 1047]: Fixes crash when using ASan with the CFI unwinder.
    * [Issue 1096]: Includes support for [Polly].
      Enable by adding `-mllvm -polly` to your cflags.
    * [Issue 1230]: LLVM's `libunwind` is now used instead of `libgcc`
      for all architectures rather than just 32-bit ARM.
    * [Issue 1231]: LLVM's `libclang_rt.builtins` is now used instead
      of `libgcc`.
    * [Issue 1406]: Fixes crash with Neon intrinsic.

  * Vulkan validation layer source and binaries are no longer shipped
    in the NDK.  The latest are now posted directly to
    [KhronosGroup/Vulkan-ValidationLayers].

  * Vulkan tools source is also removed, specifically `vulkan_wrapper`.
    It should be downloaded upstream from [KhronosGroup/Vulkan-Tools].

  * Refactored the toolchain file `android.toolchain.cmake`, basing
    it on CMake's integrated Android support.  This new toolchain file
    will be enabled by default for CMake 3.21 and newer.
    No user side change is expected.  But if anything goes wrong,
    please file a bug and set `ANDROID_USE_LEGACY_TOOLCHAIN_FILE=ON`
    to restore the legacy behavior.
      * When using the new behavior (when using CMake 3.21+ and not
        explicitly selecting the legacy toolchain),
	**default build flags may change**.  One of the primary goals
	was to reduce the behavior differences between our toolchain
	and CMake, and CMake's default flags do not always match the
        legacy toolchain file.  Most notably, if using
	`CMAKE_BUILD_TYPE=Release`, your optimization type will likely
	be `-O3` instead of `-O2` or `-Oz`.
	See [Issue 1536] for more information.

  * [Issue 929]: `find_library` now prefers shared libraries from the
    sysroot over static libraries.

  * [Issue 1390]: `ndk-build` now warns when building a static
    executable with the wrong API level.

  * [Issue 1452]: `NDK_ANALYZE=1` now sets `APP_CLANG_TIDY=true` rather
    than using `scan-build`.  `clang-tidy` performs all the same checks
    by default, and `scan-build` was no longer working.
    See [Issue 1452] for more details; no user-side changes should
    be needed.

[Issue 929]: https://github.com/android/ndk/issues/929
[Issue 1047]: https://github.com/android/ndk/issues/1047
[Issue 1096]: https://github.com/android/ndk/issues/1096
[Issue 1230]: https://github.com/android/ndk/issues/1230
[Issue 1231]: https://github.com/android/ndk/issues/1231
[Issue 1390]: https://github.com/android/ndk/issues/1390
[Issue 1406]: https://github.com/android/ndk/issues/1406
[Issue 1452]: https://github.com/android/ndk/issues/1452
[Issue 1536]: https://github.com/android/ndk/issues/1536
[Polly]: https://polly.llvm.org/
[KhronosGroup/Vulkan-ValidationLayers]: https://github.com/KhronosGroup/Vulkan-ValidationLayers/releases
[KhronosGroup/Vulkan-Tools]: https://github.com/KhronosGroup/Vulkan-Tools
```

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
```diff
@@ -75,6 +79,67 @@ public void BeforeAllTests ()
 				} catch (Exception ex) {
 					Console.Error.WriteLine ("Failed to determine whether there is Android target emulator or not: " + ex);
 				}
+
+				CreateBinutilsSymlink ();
```
pjcollins, line 83: Why is this symlink creation now required? I'm assuming there were some test failures that popped up that required this addition?

grendello: It's required to locate the in-tree binutils which only exist in our "installation" directory in the local build, they don't exist in e.g. `bin/TestRelease/blah`. The installation directory is relative to `XABT.dll` location, which will be in `bin/TestRelease/blah` but the `binutils` subdir won't be there - that's why we need the symlink

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
```diff
@@ -75,6 +79,67 @@ public void BeforeAllTests ()
...
+
+				// We first try the in-tree location...
+				string binutilsDirFullPath = Path.Combine (XABuildPaths.PrefixDirectory, "lib", "xamarin.android", "xbuild", "Xamarin", "Android", hostNdkDir);
+				if (!Directory.Exists (binutilsDirFullPath)) {
+					// ...since it failed, we'll try to find the system location
+					binutilsDirFullPath = Path.Combine (TestEnvironment.MonoAndroidToolsDirectory, hostNdkDir);
+					if (!Directory.Exists (binutilsDirFullPath)) {
+						throw new InvalidOperationException ($"Host NDK directory does not exist, {binutilsDirFullPath}");
+					}
+				}
```
pjcollins: If we do restore this, you can use Setup.AndroidMSBuildDirectory here instead (it will do the same local build tree check first): https://github.com/xamarin/xamarin-android/blob/75a0743ed078d0cbdb07fc05ff91165147520ab2/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs#L40

```suggestion
				string binutilsDirFullPath = Path.Combine (AndroidMSBuildDirectory, hostNdkDir);
				if (!Directory.Exists (binutilsDirFullPath)) {
					throw new InvalidOperationException ($"Host NDK directory does not exist, {binutilsDirFullPath}");
				}
```

jonathanpeppers:
Ok, with me reverting the change to `BaseTest.cs`, `NdkUtilTests.TestNdkUtil()` fails on macOS, but not Windows?

```
System.InvalidOperationException : Required tool '/Users/runner/work/1/s/bin/TestRelease/net6.0/Darwin/binutils/i686-linux-android-as' not found

Stack trace
   at Xamarin.Android.Tasks.NdkTools.GetExecutablePath(String toolPath, Boolean mustExist)
   at Xamarin.Android.Tasks.NdkToolsWithClangNoBinutils.GetEmbeddedToolPath(NdkToolKind kind, AndroidTargetArch arch)
   at Xamarin.Android.Tasks.NdkToolsWithClangNoBinutils.GetToolPath(NdkToolKind kind, AndroidTargetArch arch, Int32 apiLevel)
   at Xamarin.Android.Build.Tests.Tasks.NdkUtilTests.TestNdkUtil() in /Users/builder/azdo/_work/2/s/xamarin-android/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/NdkUtilTests.cs:line 52
```

But it seems like just this one test?

Could we make `GetOSBinPath()` something that could be passed in during a test:

https://github.com/grendello/xamarin-android/blob/a85c0f126b40444eb7e310a8ea8b1cb4d336b5e4/src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClangNoBinutils.cs#L51

Something like?
```csharp
ndk.OSBinPath = Path.Combine (XABuildPaths.PrefixDirectory, "lib", "xamarin.android", "xbuild", "Xamarin", "Android");
```
And if it is blank, it could call `GetOSBinPath()` by default?

tests/MSBuildDeviceIntegration/MSBuildDeviceIntegration.csproj
```diff
@@ -17,6 +17,7 @@
     <Compile Include="..\..\src\Xamarin.Android.Build.Tasks\Tests\Xamarin.Android.Build.Tests\Utilities\*.cs">
       <Link>Utilities\%(FileName).cs</Link>
     </Compile>
+    <Compile Include="..\..\tools\xabuild\SymbolicLink.cs" />
```
jonathanpeppers, line 20: If we keep `BaseTest.cs` unchanged, you could remove this one, too. I missed it.

jonpryor:
The `.apkdesc` files need to be updated again; from [BuildReleaseArm64(False)](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5148289&view=ms.vss-test-web.build-test-results-tab&runId=25238156&resultId=100032&paneView=debug):

```
apkdiff regression test failed with exit code: 3
context: https://github.com/xamarin/xamarin-android/blob/main/Documentation/project-docs/ApkSizeRegressionChecks.md
stdOut: Saving apk description to '/Users/runner/work/1/s/bin/TestRelease/BuildReleaseArm64SimpleLegacy.apkdesc'
…
Error: apkdiff: PackageSize increase 12,288 is 7,168 bytes more than the threshold 5,120. apk1 size: 3,970,772 bytes, apk2 size: 3,983,060 bytes.
```

pjcollins (approved):

jonathanpeppers (approved):

Issue #5996, opened 2021-06-04
Audit static field usage in Xamarin.Android.Build.Tasks


jonpryor:
Related: https://github.com/xamarin/xamarin-android/issues/5964#issuecomment-854273002

Question: How "safe" is it to have *mutable* (non-`const`, non-`readonly`) `static` fields in `Xamarin.Android.Build.Tasks.dll` which are used by MSBuild tasks?

The primary cause of this question is from investigating Issue #5964, in which the (thread-static!) mutable static field `NdkUtil.usingClangNDK` was inconsistently initialized, resulting in "not entirely obvious" runtime behavior.

…and it occurs to me that, *in the context of MSBuild*, I'm not sure how safe it is to have mutable static fields in the first place!  Continuing with `NdkUtil.usingClangNDK`, it's set based on the NDK version, as determined by `$(AndroidNdkDirectory)`/etc.  *For a given build*, this won't change, but *within the IDE*, it very well *could* change, e.g. by the customer changing the Android SDK directory.

It "feels" very "dubious" to have potentially changeable information stored in `static` fields.

The "obvious" replacement? Instance fields in new classes which can be cached in e.g. [`IBuildEngine4.GetRegisteredTaskObject()`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.ibuildengine4.getregisteredtaskobject?view=msbuild-16-netcore), which -- to my mind, anyway -- allows understanding of the relation of the data to the overall build process.

There are (currently) 27 fields that are `static` mutable fields:

```
% git grep '\<static\>.*;' src/Xamarin.Android.Build.Tasks | grep -v readonly | grep -v extern | grep -v Tests/
src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/AndroidLinkConfiguration.cs:		static ConditionalWeakTable<LinkContext, AndroidLinkConfiguration> configurations = new ConditionalWeakTable<LinkContext, AndroidLinkConfiguration> ();
src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs:		static List<ICustomAttributeProvider> srs_data_contract = new List<ICustomAttributeProvider> ();
src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/ApplyPreserveAttribute.cs:		static List<ICustomAttributeProvider> xml_serialization = new List<ICustomAttributeProvider> ();
[...]

Issue #5964, opened 2021-05-27
Could not AOT the Assemblies - Visual Studio 16.10


EP01:
### Steps to Reproduce

1. Update to VS 16.10.0.
2. Try to archive a Xamarin Android app with AOT + LLVM enabled. I used the default Xamarin Forms template to reproduce easily by just changing those two parameters in Release config and pressing Archive. I've attached this.
3. Fails

[App2.zip](https://github.com/xamarin/xamarin-android/files/6554017/App2.zip)

dellis1972:
This looks like a duplicate of https://github.com/xamarin/xamarin-android/issues/5764

EP01:
Sorry @dellis1972, I missed that because it was raised in March, I assumed it would have been amongst the more recent issues.

dellis1972:
@EP01 no problem :D

inforithmics:
@dellis1972 Can  you reopen this issue. Because https://github.com/xamarin/xamarin-android/issues/5764 is a different issue.

I discovered that llvm with Visual Studio 2019 doesn't work 16.10. 
The reason is that the Android sdk path has a space and then the ndk doesn't work see comments in the other issue.

Works:
C:\Users\**username**\AppData\Local\Android\sdk
Doesn't work:
C:\program Files (x86)\Android\android-sdk

And altough azure cloud builds start failing because of this.

AndreyBespamyatnov:
The same issue for AzureDevOps builds:

```2021-05-31T21:49:31.2414613Z   [aot-compiler stderr] C:\Program Files (x86)\Android\android-sdk\ndk\21.4.7075529\toolchains\llvm\prebuilt\windows-x86_64\bin\arm-linux-androideabi-ld.EXE: error: cannot open Files: No such file or directory
2021-05-31T21:49:31.2416389Z   [aot-compiler stderr] C:\Program Files (x86)\Android\android-sdk\ndk\21.4.7075529\toolchains\llvm\prebuilt\windows-x86_64\bin\arm-linux-androideabi-ld.EXE: error: cannot open (x86)\Android\android-sdk\ndk\21.4.7075529\toolchains\llvm\prebuilt\windows-x86_64\lib\gcc\arm-linux-androideabi\4.9.x: No such file or directory
[...]

PR #6083, opened 2021-07-12
[build] Reimplement NdkUtils

grendello:
Context: https://github.com/xamarin/xamarin-android/issues/5996
Context: https://github.com/xamarin/xamarin-android/issues/5964
Context: https://github.com/xamarin/xamarin-android/issues/5964#issuecomment-854273002

The `NdkUtils` class used by Xamarin.Andrid.Build.Tasks to find tooling
shipped with the Android NDK, has grown increasingly complicated over
the years due to a number of incompatibilities between various versions
of the NDK. The code became hard to follow and untidy. This commit
attempts to address the issue by replacing the single static `NdkUtils`
class with a hierarchy of dynamically instantiated classes rooted in a
new base class, `NdkTools`.

`NdkUtils` had to be initialized for each thread that needed to access
its methods, which led to various issues with concurrency and lack of
proper initialization since the initialization had to be done wherever
`NdkUtils` was first accessed, meaning that any task using it had to do
it.

`NdkTools` doesn't require such initialization, instead it provides a
factory method called `Create` which takes path to the NDK as its
parameter and returns an instance of `NdkTools` child class (or `null`
if an error occurs) which the can be safely used by the caller.  Callers
need not concern themselves with what is the actual type of the returned
instance, they access only methods and properties defined in the
`NdkTools` base abstract class.

The hierarchy of `NdkTools` derivatives is structured and named after
the breaking changes in the NDK.  For instance, NDK versions before 16
used the GNU compilers, while release 16 and above use the clang
compilers - this is reflected in existence of two classes derived from
`NdkTools`, `NoClang` for NDKs older than r16 and `WithClang` for the
newer ones. The other breaking changes are the addition of unified
headers in r19, removal of the `platforms` directory in r22 and removal
of GNU Binutils in r23.

NDK r23 is recognized in this commit but it is NOT supported. Support
for r23 is being worked on in PR #6073 which will be merged once r23 is
out of beta.


src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkToolKind.cs
```diff
@@ -0,0 +1,11 @@
+namespace Xamarin.Android.Tasks
+{
+	public enum NdkToolKind
+	{
+		Assembler,
+		CompilerC,
+		CompilerCPlusPlus,
+		Linker,
+		Strip
```
jonpryor, line 9: Please use trailing commas.

jonpryor:
Context: https://github.com/xamarin/xamarin-android/issues/5996
Context: https://github.com/xamarin/xamarin-android/issues/5964
Context: https://github.com/xamarin/xamarin-android/issues/5964#issuecomment-854273002

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkToolKind.cs
```diff
@@ -0,0 +1,11 @@
+namespace Xamarin.Android.Tasks
+{
+	public enum NdkToolKind
+	{
+		Assembler,
+		CompilerC,
```
jonpryor, line 6: Given issues reported "elsewhere" (don't want to find a ref) wherein on Windows the `clang.cmd` script actually invokes `clang++.exe`, is there some way that we could "just" deal with C++ code and not C code?  In effect, *not have* an `NdkToolKind.CompilerC` enum value, and just use `NdkToolKind.CompilerCPlusPlus` everywhere?

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
```diff
@@ -0,0 +1,391 @@
...
+		// We could call `MonoAndroidHelper.RefreshSupportedVersions` but since this NdkTools
+		// instance already knows the location of NDK, it would be just a waste of time.
+		protected string HostPlatform => GetNdkHostPlatform ();
+		protected bool IsWindows => OS.IsWindows;
+
+		protected TaskLoggingHelper? Log { get; }
+
+		protected NdkTools (string androidNdkPath, NdkVersion version, TaskLoggingHelper? log = null)
+		{
+			if (String.IsNullOrEmpty (androidNdkPath)) {
```
jonpryor, line 52: Should this also `Directory.Exists(androidNdkPath)`?

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
```diff
@@ -0,0 +1,391 @@
...
+				throw new ArgumentException ("must be a non-empty string", nameof (androidNdkPath));
+			}
+
+			NdkRootDirectory = androidNdkPath;
+			Version = version;
+		}
+
+		public static NdkTools? Create (string androidNdkPath, TaskLoggingHelper? log = null)
+		{
+			if (String.IsNullOrEmpty (androidNdkPath)) {
```
jonpryor, line 62: Should this also call `Directory.Exists(androidNdkPath)`?

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
```diff
@@ -0,0 +1,391 @@
...
+			}
+
+			if (!Directory.Exists (androidNdkPath)) {
+				log?.LogCodedError ("XA5104", Properties.Resources.XA5104);
+				return null;
+			}
+
+			NdkVersion? version = ReadVersion (androidNdkPath, log);
+			if (version == null) {
+				return null;
```
jonpryor, line 74: *May* want to add a `log?.LogDebugMessage("Could not read version information")` or something, if/when this starts returning null…

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
```diff
@@ -0,0 +1,391 @@
...
+			}
+
+			if (version.Main.Major < 14) {
+				throw new InvalidOperationException ($"Unsupported NDK version {version}");
```
jonpryor, line 78: This should also report an XA5104 error.  Sure, we "found" an NDK error, but as-is if someone tries to use NDK r13 (?! insanity ?!), the build will fail with the InvalidOperationException, which isn't as useful as an IOE + XA5104.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkToolKind.cs
grendello: `mkbundle` requires a C compiler (or we'd have to change it to pass `-x C` to the C++ compiler). And the `clang.cmd` invokes clang++ only in NDK 19 and 19b, 19c fixes it:

```cmd
@echo off
setlocal
call :find_bin
if "%1" == "-cc1" goto :L

set "_BIN_DIR=" && %_BIN_DIR%clang.exe --target=armv7a-linux-androideabi16 -fno-addrsig %*
if ERRORLEVEL 1 exit /b 1
goto :done

:L
rem Target is already an argument.
set "_BIN_DIR=" && %_BIN_DIR%clang.exe %*
if ERRORLEVEL 1 exit /b 1
goto :done

:find_bin
rem Accommodate a quoted arg0, e.g.: "clang"
rem https://github.com/android-ndk/ndk/issues/616
set _BIN_DIR=%~dp0
exit /b

:done
```

We already hack around the above CMD (due to quoting issues), I don't want to introduce another hack to make `mkbundle` work with a C++ compiler :)

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
grendello: No, that check is made in `NdkTools.Create` which is the only supported way to create instances of the NdkTools derivatives.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkToolKind.cs
jonpryor: …Meaning this is a place that we can cleanup when we drop support for the Legacy SDK, and thus also drop support for `mkbundle` entirely.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
```diff
@@ -0,0 +1,391 @@
...
+			return Path.Combine (GetUnifiedHeadersDirPath (NdkRootDirectory), GetArchTriple (arch));
+		}
+
+		protected string UnifiedHeaders_GetPlatformIncludeDirPath (AndroidTargetArch arch, int apiLevel)
+		{
+			return GetUnifiedHeadersDirPath (NdkRootDirectory);
+		}
+
+		const string platformLinux64 = "linux-x86_64";
+                const string platformLinux32 = "linux-x86";
```
jonpryor, line 306: Indentation starting here is wrong, and appears to be using spaces instead of tabs.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
```diff
@@ -0,0 +1,391 @@
...
+                                throw new InvalidOperationException ($"Unable to determine host NDK platform");
+                        }
+
+                        throw new InvalidOperationException ($"Unsupported OS");
+
+                        bool HasPrebuiltDir (string name)
+                        {
+                                return Directory.Exists (Path.Combine (NdkRootDirectory, "prebuilt", name));
+                        }
+                }
```
jonpryor, line 362: End of "wrong indentation" section.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
jonpryor: This is kinda/sorta done, actually, as `ReadVersion()` contains some logging.

The "problem"/"concern" is one of "scope" & expectations: since I saw `LogCodedError()` previously in this method, I *expected* a similar log here.  Perhaps my expectation is wrong.

Additionally, `ReadVersion()` has a code path which could throw an exception, which would result in XA5104 *not* being raised, which could be differently underirable.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
```diff
@@ -0,0 +1,391 @@
...
+			string? ver = null;
+			foreach (string l in File.ReadAllLines (sourcePropertiesPath)) {
+				string line = l.Trim ();
+				if (!line.StartsWith ("Pkg.Revision", StringComparison.Ordinal)) {
+					continue;
+				}
+
+				string[] parts = line.Split (splitChars, 2);
+				if (parts.Length != 2) {
+					throw new InvalidOperationException ($"Invalid NDK version format in '{sourcePropertiesPath}'.");
```
jonpryor, line 382: May be better if this is a `log?.LogDebugMessage()` + `return null`.  Should also emit an XA5104, which could be done here, or could instead be done within `Create()`; see also? https://github.com/xamarin/xamarin-android/pull/6083/files#r669778870

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NoClang.cs
```diff
@@ -0,0 +1,99 @@
...
+		public override bool ValidateNdkPlatform (Action<string> logMessage, Action<string, string> logError, AndroidTargetArch arch, bool enableLLVM)
+		{
+			// Check that we have a compatible NDK version for the targeted ABIs.
+			if (IsNdk64BitArch (arch) && Version.Main.Major < 10) {
+				logMessage (
+					"The detected Android NDK version is incompatible with the targeted 64-bit architecture, " +
+					"please upgrade to NDK r10 or newer.");
+			}
+
+			// NDK r10d is buggy and cannot link x86_64 ABI shared libraries because they are 32-bits.
```
jonpryor, line 29: If NDK r14 is our minimum supported version, do we need to care about NDK r10?  We could probably just remove this code entirely.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
```diff
@@ -0,0 +1,391 @@
...
+		// We can't use MonoAndroidHelper.AndroidSdk.AndroidNdkHostPlatform here since it's
+		// not initialized while running tests and attempts to use cause a NREX
+		// We could call `MonoAndroidHelper.RefreshSupportedVersions` but since this NdkTools
+		// instance already knows the location of NDK, it would be just a waste of time.
+		protected string HostPlatform => GetNdkHostPlatform ();
+		protected bool IsWindows => OS.IsWindows;
+
+		protected TaskLoggingHelper? Log { get; }
+
+		protected NdkTools (string androidNdkPath, NdkVersion version, TaskLoggingHelper? log = null)
```
jonpryor, line 50: Shouldn't the `Log` property be set to `log`?

Relatedly, do we actually want `log` to be a `TaskLoggingHelper` and not an `Action<TraceLevel, string>`, as is done elsewhere?  That way if `log` is null, we can default it to `Console.WriteLine()`s as a "worst case we need to log *something*" scenario.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClang.cs
```diff
@@ -0,0 +1,128 @@
...
+			string toolchainDir = GetToolchainDir ();
+			string clangBaseDir = Path.Combine (toolchainDir, "lib64", "clang");
+
+			if (!Directory.Exists (clangBaseDir)) {
+				throw new InvalidOperationException ($"Clang toolchain directory '{clangBaseDir}' not found");
+			}
+
+			// There should be just one subdir - clang version - but it's better to be safe than sorry...
+			foreach (string dir in Directory.EnumerateDirectories (clangBaseDir)) {
+				if (dir[0] == '.') {
```
jonpryor, line 79: Is there a reason we need to explicitly skip over hidden dirs?

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClangNoPlatforms.cs
```diff
@@ -0,0 +1,86 @@
...
+		{
+			// NDK r22 and newer no longer have a single platforms dir.  The API level directories are now found in per-arch
+			// subdirectories under the toolchain directory. We need to examine all of them and compose a list of unique
+			// API levels (since they are repeated in each per-arch subdirectory, but not all architectures have the
+			// same set of API levels)
+			var apiLevels = new HashSet<int> ();
+			string sysrootLibDir = GetToolchainLibDir ();
+			foreach (AndroidTargetArch targetArch in Enum.GetValues (typeof (AndroidTargetArch))) {
+				if (targetArch == AndroidTargetArch.None ||
+				    targetArch == AndroidTargetArch.Other ||
```
jonpryor, line 31: Please indent nested expression *two* tab stops, not spaces:

```csharp
if (targetArch == AndroidTargetArch.None ||
[tab][tab]targetArch == AndroidTargetArch.Other ||
[tab][tab]targetArch == AndroidTargetArch.Mips)
[tab]continue;
```

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClangNoPlatforms.cs
```diff
@@ -0,0 +1,86 @@
...
+
+		public override IEnumerable<int> GetSupportedPlatforms ()
+		{
+			// NDK r22 and newer no longer have a single platforms dir.  The API level directories are now found in per-arch
+			// subdirectories under the toolchain directory. We need to examine all of them and compose a list of unique
+			// API levels (since they are repeated in each per-arch subdirectory, but not all architectures have the
+			// same set of API levels)
+			var apiLevels = new HashSet<int> ();
+			string sysrootLibDir = GetToolchainLibDir ();
+			foreach (AndroidTargetArch targetArch in Enum.GetValues (typeof (AndroidTargetArch))) {
```
jonpryor, line 29: This feels odd; why use `Enum.GetValues()` instead of `new[]{AndroidTargetArch.Arm, …}`?

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClangWithPlatforms.cs
```diff
@@ -0,0 +1,70 @@
...
+
+		public override string GetCompilerTargetParameters (AndroidTargetArch arch, int apiLevel, bool forCPlusPlus = false)
+		{
+			if (!NeedClangWorkaround) {
+				return base.GetCompilerTargetParameters (arch, apiLevel, forCPlusPlus);
+			}
+
+			var sb = new StringBuilder ();
+			sb.Append ("--target=").Append (GetCompilerTriple (arch)).Append (apiLevel)
+			  .Append (" -fno-addrsig");
```
jonpryor, line 57: Spaces here should be a tab.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NoClang.cs
grendello: I think we should still keep the validation code, should anyone have NDK r10 installed. I will update the message to recommend at least r14 though

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
grendello: log is `null` only in some tests and I want to have the full logger available, I've always disliked this bit of the old code as logging of non-error messages was unnecessarily hard.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClang.cs
grendello: It's not about hidden directories but `.` and `..` entries. We don't want to travel up, as we know there's nothing there for us to find, and we do know that the current dir also doesn't have what we want - this way we save creation of two strings and at least two `stat` calls. Skipping of the hidden directories is just a bonus (likely irrelevant as it's unlikely any will be found there).

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClangNoPlatforms.cs
grendello: Old code, copy pasta

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
```diff
@@ -0,0 +1,397 @@
...
+		{
+			return ValidateNdkPlatform ((m) => Log?.LogMessage (m), (c, m) => Log?.LogCodedError (c, m), arch, enableLLVM);
+		}
+
+		public string GetArchDirName (AndroidTargetArch arch)
+		{
+			return GetArchTriple (arch);;
+		}
+
+		public virtual IEnumerable<int> GetSupportedPlatforms ()
```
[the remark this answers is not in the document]
grendello: This stuff is new for NDK r22, and there are no non-int subdirectories there, even in NDK r23-beta5:

```bash
android-ndk-r23-beta5/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/aarch64-linux-android$ find -type d
.
./21
./22
./23
./24
./26
./27
./28
./29
./30
./31
```

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkToolKind.cs
grendello: When that happens we can drop the compiler support completely, as AOT (and us, as well) use only the assembler, the linker and strip

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
grendello: No more exceptions thrown there in the latest update.

grendello: TBH, I'd much prefer to keep throwing exceptions (also from `.Create`) to have a sorta-kinda RAII-like behavior, and some code somewhere else capture the exceptions and turn them into XA5104, but I don't think there's any central location where that could be done

jonpryor: The problem is that the exceptions *are* caught from a central location:

https://github.com/xamarin/xamarin-android-tools/blob/623332da80e1ff2e1a567258c8df8a2a87558aba/src/Microsoft.Android.Build.BaseTasks/AndroidTask.cs#L16-L21

The problem is that they're not transformed into any "controllable" error code; they just go to `Log.LogUnhandledException()`, which makes for a "less than stellar" error experience.

I'm fine with exceptions, as-such.  But I'd *also* like `Log.LogCodedError()`s before the exception, when sane/possible.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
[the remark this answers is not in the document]
jonpryor: …and API-31 isn't even stable yet.  Guess `int`s are fine, at least for now.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NoClang.cs
jonpryor: Sorry; I misread this: https://github.com/xamarin/xamarin-android/pull/6083/files#diff-7c0f17654397609ff55091f911023c78c6325652f1c4839fa8a9cdead0772bf4R77

I thought ^^ was an "early error check + return", but it isn't.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClang.cs
jonpryor: `Directory.EnumerateDirectories()` doesn't return `.` or `..`, at least not for me:

```
System.IO.Directory.EnumerateDirectories("/tmp");
// { "/tmp/powerlog", "/tmp/com.apple.launchd.FsyjqY8Rry", "/tmp/.dotnet" }
```

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClangNoBinutils.cs
```diff
@@ -0,0 +1,42 @@
...
+					return base.GetToolPath (kind, arch, apiLevel);
+			}
+		}
+
+		string GetEmbeddedToolPath (NdkToolKind kind, AndroidTargetArch arch)
+		{
+			string toolName = GetToolName (kind);
+			string triple = GetArchTriple (arch);
+
+			return $"[TODO]/{triple}-{toolName}";
```
jonpryor, line 39: Not sure how I missed this last time, but… `[TODO]`?!

Won't this just fail everywhere?  Maybe an exception should be thrown?  (How could this possibly work?)

jonpryor: This is present for @grendello 's sanity when running tests locally.

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
jonpryor: As discussed [on discord](https://discord.com/channels/732297728826277939/732297837953679412/864968631582064671), the first sentence was overlooked:

> Shouldn't the `Log` property be set to `log`?

The `Log` property isn't set by any codepath.

This constructor should be updated to include:

```csharp
Log = log;
```

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/NdkTools.cs
jonpryor: @grendello  [wrote](https://discord.com/channels/732297728826277939/732297837953679412/864974432242368523):

> I tried this
> exceptions **and*8 a logcodederror
> breaks the tests

*Weird*. :-/

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClang.cs
grendello: Mono must have changed the behavior since that code was implemented, it no longer returns the `.` or `..` entries

src/Xamarin.Android.Build.Tasks/Utilities/NdkTools/WithClangNoBinutils.cs
grendello: The class can't be constructed, so this overload won't ever run. The class will be usable when PR #6073 is in and NDK r23 is released.

PR #6084, opened 2021-07-12
[Xamarin.Android.Build.Tasks] require JDK 11 for API 31

jonathanpeppers:
API 31 builds fail when using JDK 1.8 with:

    /Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1728,3):
    error JAVAC0000: java.lang.AssertionError: annotationType() : unrecognized Attribute name MODULE (class com.sun.tools.javac.util.SharedNameTable$NameImpl)

To solve this issue, we must require JDK 11 for API 31.

There are two code paths here for "legacy" Xamarin.Android and .NET 6.
Legacy is straightforward, as we already had a pattern for this.

In .NET 6, I checked the `$(TargetPlatformVersion)` property for now
and require JDK 11 in that case. When API 31 is stable, we can
probably remove this code and just set this value by default:

    <MinimumSupportedJavaVersion Condition=" '$(MinimumSupportedJavaVersion)' == '' ">11.0</MinimumSupportedJavaVersion>

I updated some tests in this area as well.

### DIFF
--- a/Documentation/release-notes/6084.md
+++ b/Documentation/release-notes/6084.md
@@ -1,0 +1,8 @@
+#### Application build and deployment
+
+- [GitHub Pull Request #6084](https://github.com/xamarin/xamarin-android/pull/6084):
+  Android API 31 and a `$(TargetFrameworkVersion)` of `v11.0.99`
+  requires JDK 11. This can be obtained manually by installing
+  [Microsoft OpenJDK 11][ms-openjdk].
+
+[ms-openjdk]: https://docs.microsoft.com/java/openjdk/download
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Tooling.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Tooling.targets
@@ -23,6 +23,7 @@ called for "legacy" projects in Xamarin.Android.Legacy.targets.
         JavaSdkPath="$(_JavaSdkDirectory)"
         JavaToolExe="$(JavaToolExe)"
         JavacToolExe="$(JavacToolExe)"
+        TargetPlatformVersion="$(TargetPlatformVersion)"
         LatestSupportedJavaVersion="$(LatestSupportedJavaVersion)"
         MinimumSupportedJavaVersion="$(MinimumSupportedJavaVersion)">
       <Output TaskParameter="JdkVersion"                PropertyName="_JdkVersion" />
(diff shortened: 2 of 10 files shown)


jonpryor:
We're starting to wonder if we properly understand what's going on here. The [current PR build](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4975120&view=ms.vss-test-web.build-test-results-tab&runId=23503086&resultId=100019&paneView=debug), which requires JDK 11 for API-31+, has *lots* of unit test failures, most of which are *because* of the new check, [e.g. LatestTFV_OldTargetSdkVersion](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=4975120&view=ms.vss-test-web.build-test-results-tab&runId=23499372&resultId=100181&paneView=attachments):

```
/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Legacy.targets(239,5):
error XA0031: Java SDK 11.0 or above is required when using $(TargetFrameworkVersion) v11.0.99. 
```

These tests are *currently* building now, meaning *most* of our tests are currently able to target v11.0.99 ("API-S") ***and build***.  They're building *now*, without incurring the `JAVAC0000` error about `unrecognized Attribute name MODULE`.

There's something important that we're not understanding here. :-(

More investigation is required.

jonathanpeppers:
The tests that fail in #6094 are Xamarin.Forms 4.0 apps that use Android Support libraries.

So I tried doing a Blank Android App in VS 2022, then changed `TargetFrameworkVersion=v11.0.99`. It *also* failed with the error.

The problem went away when I removed all AndroidX packages. So perhaps, the precense of Android Support or AndroidX causes the issue? Many of our tests just use plain Android.

jonathanpeppers:
Ugh, need to fix the conflict

jonpryor:
```markdown
API-31 builds fail when using JDK 1.8 and using AndroidX with:

	/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1728,3):
	error JAVAC0000: java.lang.AssertionError: annotationType() : unrecognized Attribute name MODULE (class com.sun.tools.javac.util.SharedNameTable$NameImpl)

There are two plausible solutions to better handle this:

 1. Only require JDK 11 when API-31 and AndroidX are being used, or
 2. Always required JDK 11 for API-31+

As we believe *most* apps will be using AndroidX, there is little
point to attempting so support (1); (2) it is!

We must require JDK 11 for API-31+.

There are two code paths here: one for "legacy" Xamarin.Android, and
one for .NET 6.

Legacy is straightforward, as we already had a pattern for this, we
just update the `<ValidateJavaVersion/>` task.

For .NET 6, I checked the `$(TargetPlatformVersion)` property for now
and require JDK 11 in that case.  When API-31 is stable, we can
probably remove this code and just set this value by default:

	<MinimumSupportedJavaVersion Condition=" '$(MinimumSupportedJavaVersion)' == '' ">11.0</MinimumSupportedJavaVersion>

For our unit tests -- most of which would be solution (1) (API-31
and *no* AndroidX use), but many of which started to fail because we
went with solution (2) -- we did one of two things:

 1. Use a different `$(TargetFrameworkVersion)` and/or remove
    `$(TargetFrameworkVersion)` entirely, or

 2. Use a new `BuildTest.AssertTargetFrameworkVersionSupported()`
    method which `Assert.Ignore()`s the test if it would not
    pass otherwise.
```

PR #6087, opened 2021-07-14
[One .NET] select defaults for App Bundles

### LINKED ISSUE #6059: [One .NET] select defaults for App Bundles
August 2021, Android App Bundles will be required for new apps on Google Play:

https://android-developers.googleblog.com/2020/11/new-android-app-bundle-and-target-api.html

In .NET 6, we should:

1. Release builds with a blank `$(AndroidPackageFormat)` should default to `aab`.
2. Come up with a new property that generates `.apk` files in addition to `.aab`. Enable this by default for release builds.

For no. 2 to work, we should be able to generate the `.aab` as usual, then add a separate step for generating a "fat" `.apk` file for all architectures. This way apps could submit builds to other stores than Google Play, which might not support App Bundles yet.

dellis1972:
Fixes #6059

Users will probably want to target more than one App Store. Google is now requiring the `aab` format for 
Google Play Store uploads. Unfortunately this package format is not compatible with other stores. 
Users can build their app twice producing an `aab` for one build and `apk` for another, but we should
try to make this a bit easier. 

`bundle-tool` has the ability to create a universal apk from the `aab` file. So lets make use of that to 
generate one along side the `aab`. We are introducing a new property `AndroidPackageFormats`.
Under .net 6 this will have a value of `aab;apk` by default for Release builds, for Legacy it will be 
empty. 

If `AndroidPackageFormats` is specified and `AndroidPackageFormat` is empty we will use the 
values in `AndroidPackageFormats` to populate `AndroidPackageFormat`. If none of these values
are provided the `AndroidPackageFormat` will still default to `apk` as it has always done. 

The following setting will produce both an `aab` and an `apk`.
```
<PropertyGroup>
  <AndroidPackageFormats>aab;apk</AndroidPackageFormats>
</PropertyGroup>
```

This will produce just an `aab`

```
<PropertyGroup>
  <AndroidPackageFormats>aab</AndroidPackageFormats>
</PropertyGroup>
```

and this will produce just an `apk`

```
<PropertyGroup>
  <AndroidPackageFormats>apk</AndroidPackageFormats>
</PropertyGroup>
```

For .net 6 users this will be enabled by default for Release builds. For legacy users they will need
to define the property manually.

### DIFF
--- a/Documentation/guides/building-apps/build-properties.md
+++ b/Documentation/guides/building-apps/build-properties.md
@@ -172,8 +172,8 @@ Added in Xamarin.Android 10.2.
 ## AndroidBoundInterfacesContainConstants
 
 A boolean property that
-determines whether binding constants on interfaces will be supported, 
-or the workaround of creating an `IMyInterfaceConsts` class 
+determines whether binding constants on interfaces will be supported,
+or the workaround of creating an `IMyInterfaceConsts` class
 will be used.
 
 Defaults to `True` in .NET 6 and `False` for legacy.
@@ -181,8 +181,8 @@ Defaults to `True` in .NET 6 and `False` for legacy.
 ## AndroidBoundInterfacesContainStaticAndDefaultInterfaceMethods
 
 A boolean property that
-whether default and static members on interfaces will be supported, 
-or  old workaround of creating a sibling class containing static 
+whether default and static members on interfaces will be supported,
+or  old workaround of creating a sibling class containing static
 members like `abstract class MyInterface`.
 
 Defaults to `True` in .NET 6 and `False` for legacy.
@@ -190,7 +190,7 @@ Defaults to `True` in .NET 6 and `False` for legacy.
 ## AndroidBoundInterfacesContainTypes
 
 A boolean property that
-whether types nested in interfaces will be supported, or the workaround 
+whether types nested in interfaces will be supported, or the workaround
 of creating a non-nested type like `IMyInterfaceMyNestedClass`.
 
 Defaults to `True` in .NET 6 and `False` for legacy.
@@ -913,6 +913,43 @@ properties are set, which are required for Android App Bundles:
 
 [apk]: https://en.wikipedia.org/wiki/Android_application_package
 [bundle]: https://developer.android.com/platform/technology/app-bundle
+
+This property will be deprecated for .net 6. Users should switch over to
+the newer [`AndroidPackageFormats`](~/android/deploy-test/building-apps/build-properties.md#androidpackageformats).
+
+## AndroidPackageFormats
+
+A semi-colon delimited property with valid values of `apk` and `aab`.
+This indicates if you want to package the Android application as
+an [APK file][apk] or [Android App Bundle][bundle]. App Bundles
+are a new format for `Release` builds that are intended for
+submission on Google Play.
+
+When building a Release build you might want to generate both
+and `aab` and an `apk` for distribution to various stores.
+
+Setting `AndroidPackageFormats` to `aab;apk` will result in both
+being generated. Setting `AndroidPackageFormats` to either `aab`
+or `apk` will generate only one file.
+
+For .net 6 `AndroidPackageFormats` will be set to `aab;apk` for
+`Release` builds only. It is recommended that you continue to use
+just `apk` for debugging.
+
+For Legacy Xamarin.Android this value currently defaults to `""`.
+As a result Legacy Xamarin.Android will NOT by default produce
+both as part of a release build. If a user wants to produce both
+outputs they will need to define the following in their `Release`
+configuration.
+
+```
+<AndroidPackageFormats>aab;apk</AndroidPackageFormats>
+```
+
+You will also need to remove the existing `AndroidPackageFormat` for
+that configuration if you have it.
+
+Added in Xamarin.Android 11.5.
 
 ## AndroidPackageNamingPolicy
 
--- a/build-tools/automation/yaml-templates/apk-instrumentation.yaml
+++ b/build-tools/automation/yaml-templates/apk-instrumentation.yaml
@@ -19,7 +19,7 @@ steps:
       solution: ${{ parameters.project }}
       configuration: ${{ parameters.configuration }}
       msbuildArguments: >-
-        /restore
+        /restore /v:diag
         /t:AcquireAndroidTarget,SignAndroidPackage,DeployTest${{ parameters.packageType }}s,CheckAndRecordApkSizes,RunTestApks,UndeployTestApks,RenameApkTestCases,ReportComponentFailures
         /bl:$(System.DefaultWorkingDirectory)/bin/Test${{ parameters.configuration }}/run-${{ parameters.testName }}.binlog
         ${{ parameters.extraBuildArgs }}
@@ -34,7 +34,7 @@ steps:
       arguments: >-
         -t:AcquireAndroidTarget,SignAndroidPackage,DeployTest${{ parameters.packageType }}s,CheckAndRecordApkSizes,RunTestApks,UndeployTestApks,RenameApkTestCases,ReportComponentFailures
         -bl:$(System.DefaultWorkingDirectory)/bin/Test${{ parameters.configuration }}/run-${{ parameters.testName }}.binlog
-        -v:n -c ${{ parameters.configuration }} ${{ parameters.extraBuildArgs }}
+        -v:d -c ${{ parameters.configuration }} ${{ parameters.extraBuildArgs }}
       condition: ${{ parameters.condition }}
 
 - script: >
--- a/build-tools/automation/yaml-templates/apk-instrumentation.yaml
+++ b/build-tools/automation/yaml-templates/apk-instrumentation.yaml
@@ -19,7 +19,7 @@ steps:
       solution: ${{ parameters.project }}
       configuration: ${{ parameters.configuration }}
       msbuildArguments: >-
-        /restore /v:diag
+        /restore
         /t:AcquireAndroidTarget,SignAndroidPackage,DeployTest${{ parameters.packageType }}s,CheckAndRecordApkSizes,RunTestApks,UndeployTestApks,RenameApkTestCases,ReportComponentFailures
         /bl:$(System.DefaultWorkingDirectory)/bin/Test${{ parameters.configuration }}/run-${{ parameters.testName }}.binlog
         ${{ parameters.extraBuildArgs }}
@@ -34,7 +34,7 @@ steps:
       arguments: >-
         -t:AcquireAndroidTarget,SignAndroidPackage,DeployTest${{ parameters.packageType }}s,CheckAndRecordApkSizes,RunTestApks,UndeployTestApks,RenameApkTestCases,ReportComponentFailures
         -bl:$(System.DefaultWorkingDirectory)/bin/Test${{ parameters.configuration }}/run-${{ parameters.testName }}.binlog
-        -v:d -c ${{ parameters.configuration }} ${{ parameters.extraBuildArgs }}
+        -v:n -c ${{ parameters.configuration }} ${{ parameters.extraBuildArgs }}
       condition: ${{ parameters.condition }}
 
 - script: >
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
@@ -18,7 +18,7 @@
     <!-- mono-symbolicate is not supported -->
     <MonoSymbolArchive>false</MonoSymbolArchive>
     <!--
-      Disable @(Content) from referenced projects 
+      Disable @(Content) from referenced projects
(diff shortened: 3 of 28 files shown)


dellis1972:
Ok , huge number of test failures because we are producing an `aab` for release under .net 6.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -43,6 +43,8 @@
   <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
     <EmbedAssembliesIntoApk Condition=" '$(EmbedAssembliesIntoApk)' == '' ">true</EmbedAssembliesIntoApk>
     <AndroidManagedSymbols Condition=" '$(AndroidManagedSymbols)' == '' ">true</AndroidManagedSymbols>
+    <AndroidPackageFormat Condition=" '$(AndroidPackageFormat)' == '' " >aab</AndroidPackageFormat>
+    <AndroidAdditionalPackageFormats Condition=" '$(AndroidAdditionalPackageFormats)' == '' " >apk</AndroidAdditionalPackageFormats>
```
jonathanpeppers, line 47: Should this maybe also check:
```suggestion
    <AndroidAdditionalPackageFormats Condition=" '$(AndroidPackageFormat)' == 'aab' and '$(AndroidAdditionalPackageFormats)' == '' " >apk</AndroidAdditionalPackageFormats>
```

jonpryor:
Conceptually, `$(AndroidPackageFormat)` + `$(AndroidAdditionalPackageFormats)` feels "weird": it's inconsistent with other existing properties, such as `$(TargetFrameworks)`, `$(RuntimeIdentifiers)`.

Is there a reason to not go with `$(AndroidPackageFormats)` (plural)?  Is it easier for the IDE to have the two separate properties?

jonpryor:
@jonpryor [asked](https://github.com/xamarin/xamarin-android/pull/6087#issuecomment-880755331):

> Is there a reason to not go with $(AndroidPackageFormats) (plural)?

@dellis1972 [addressed this](https://github.com/xamarin/xamarin-android/issues/6059#issuecomment-880656838) on #6059:

> Changing over completely to AndroidPackageFormats would have resulted in a rather large PR.

I wonder how much larger it would be, though…. Could we make it work with an "inner build"-like system?

dellis1972:
> Conceptually, `$(AndroidPackageFormat)` + `$(AndroidAdditionalPackageFormats)` feels "weird": it's inconsistent with other existing properties, such as `$(TargetFrameworks)`, `$(RuntimeIdentifiers)`.
> Is there a reason to not go with `$(AndroidPackageFormats)` (plural)? Is it easier for the IDE to have the two separate properties?

We have many MSBuild Properties which rely on an it being one or the other for example (and this is just a small example).

```
<AndroidUseAapt2            Condition=" '$(AndroidPackageFormat)' == 'aab' ">True</AndroidUseAapt2>
<AndroidUseApkSigner        Condition=" '$(AndroidPackageFormat)' == 'aab' ">False</AndroidUseApkSigner>
<AndroidCreatePackagePerAbi Condition=" '$(AndroidCreatePackagePerAbi)' == 'aab' ">False</AndroidCreatePackagePerAbi>
...

<_BuildApkEmbedOutputs Condition=" '$(AndroidPackageFormat)' == 'aab' ">
<_BuildApkEmbedOutputs Condition=" '$(AndroidPackageFormat)' != 'aab' ">
```

Also take a look at our current signing system. https://github.com/xamarin/xamarin-android/blob/main/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L2287

Its pretty messy. and depends on if we are signing with ApkSigner or not. `aab` files MUST use ApkSigner, `apk` MUST not. 

Now I'm not saying that we don't need to rework this because its clear that we do. However, the approach in this PR seemed like the most straight forward given the time constraints for this feature. Tacking on the additional apk generation for this specific case i.e when the user is in release mode and are generating an `aab`.

As we talked about on the meeting, we could just make `AndroidAdditionalPackageFormats` internal. So not something users would use, but it would be automatic if they are using .net 6 AND `aab` And `Release`.


> I wonder how much larger it would be, though…. Could we make it work with an "inner build"-like system?

my concern with inner builds are performance, the fact that we (in .net 6) have to do an inner build for each abi already, adding yet another one for packaging/signing, then in the future more for Dynamic Features. I'd rather avoid those if possible. But it might well be our only option.

So I guess the question really is, how quickly do we want this? Should we do it fast or right?

dellis1972:
Oh and all the unit tests are going to have to change as well. Most are hardcoded to look for `{proj.PackageName}.apk` most of the time.

dellis1972:
Interesting, I just found out that this might be a thing in .net 6

```xml
  <PropertyGroup Condition="'$(Configuration)' == 'AmazonRelease'">
    <ConfigurationGroup>Release</ConfigurationGroup>
    <AndroidPackageFormat>apk</AndroidPackageFormat>
  </PropertyGroup>
```

jonpryor:
@dellis1972: what I think we should do is "split the difference", somewhat:

 1. Document and "add" a new `$(AndroidPackageFormats)` property, which is a `;`-delimited list of values to use among `apk` or `aab`.  Order isn't important, and thus the following values are supported:

      * `apk`
      * `aab`
      * `apk;aab`
      * `aab;apk`

 2. Instead of "properly" supporting `$(AndroidPackageFormats)`, we process it and convert it into `$(AndroidPackageFormat)` + `$(_AndroidAdditionalPackageFormats)`.

      * `apk` => `$(AndroidPackageFormat)`=apk, `$(_AndroidAdditionalPackageFormats)`=''
      * `aab` => `$(AndroidPackageFormat)`=aab, `$(_AndroidAdditionalPackageFormats)`=''
      * `apk;aab` => `$(AndroidPackageFormat)`=aab, `$(_AndroidAdditionalPackageFormats)`=apk
      * `aab;apk` => `$(AndroidPackageFormat)`=aab, `$(_AndroidAdditionalPackageFormats)`=apk

This should allow a more "future-proof" "public API" in the form of `$(AndroidPackageFormats)`, while allowing us to begin supporting `$(AndroidPackageFormats)` in a faster manner.

Documentation/guides/building-apps/build-properties.md
```diff
@@ -908,6 +908,32 @@ properties are set, which are required for Android App Bundles:
...
+
+Setting `AndroidPackageFormats` to `aab;apk` will result in both
+being generated. Setting `AndroidPackageFormats` to either `aab`
+or `apk` will generate only one file.
+
+For .net 6 `AndroidPackageFormats` will be set to `aab;apk` for
+`Release` builds only. It is recommended that you continue to use
+just `apk` for debugging.
+
+For Legacy Xamarin.Android this value currently defaults to `apk`.
```
jonpryor, line 933: This implies that this new property is supported/works on Legacy.  Is this the case?  If not, we should explicitly mention that.

Documentation/guides/building-apps/build-properties.md
```diff
@@ -908,6 +908,32 @@ properties are set, which are required for Android App Bundles:
...
+
+Added in Xamarin.Android 11.5.
```
jonpryor, line 935: @jonathanpeppers: What do we call the final .NET 6 Android for .NET (or whatever) version?  Assuming that this new property is only supported in .NET6+, what should we say here?

jonathanpeppers: I think we should put `.NET 6 Preview 7`, if this is where the fix would land.

This currently is working in legacy as well, though, right? I don't see what it would hurt if it was added to legacy, as it's just an extra target that runs for a new MSBuild property that didn't exist before.

dellis1972: @jonpryor is 11.5 the right version? I can never remember where to look for the next release number.
Also do we need to start putting in the .net 6 version as well or instead of?

jonathanpeppers: I think 11.5 is right because main is:

https://github.com/xamarin/xamarin-android/blob/21751b5bc4292be4b23dd58577cb6cf9f01d82bb/Directory.Build.props#L16

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.targets
```diff
@@ -2,7 +2,7 @@
   <Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk"
       Condition=" '$(TargetPlatformIdentifier)' == 'android' " />
   <Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk.BundleTool"
-      Condition=" '$(AndroidPackageFormat)' == 'aab' " />
+      Condition=" '$(AndroidPackageFormat)' == 'aab' Or $(AndroidPackageFormats.Contains ('aab')) " />
```
jonathanpeppers, line 5: I was talking about this with @jonpryor, I think you can just completely remove the `Condition` here and always import it.

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.targets
```diff
@@ -1,8 +1,7 @@
 <Project>
   <Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk"
       Condition=" '$(TargetPlatformIdentifier)' == 'android' " />
-  <Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk.BundleTool"
-      Condition=" '$(AndroidPackageFormat)' == 'aab' " />
+  <Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk.BundleTool" />
```
jonathanpeppers, line 4: Sorry, I guess we still need to check `Condition=" '$(TargetPlatformIdentifier)' == 'android' "` here. It should match the line above.

Otherwise we'd import this for a .NET 6 console app.

dellis1972:
doh. 

So .. we have a slight problem here. bundle-tool does NOT support the `env:password` option, only `pass:` and `file:`. 
But `jarsigner` does support both. So with this new system we can build the `aab` but then when creating a new `apk` from that with an `env:` style password we get this from bundletool

```
[BT : 1.4.0] error : Passwords must be prefixed with "pass:" or "file:".
```

So we effectively can't use `env:` anymore.

grendello (approved):

build-tools/automation/yaml-templates/apk-instrumentation.yaml
```diff
@@ -19,7 +19,7 @@ steps:
       solution: ${{ parameters.project }}
       configuration: ${{ parameters.configuration }}
       msbuildArguments: >-
-        /restore
+        /restore /v:diag
```
jonathanpeppers, line 22: Did we need `/v:diag` here? It's already saving a `.binlog`.

build-tools/automation/yaml-templates/apk-instrumentation.yaml
```diff
@@ -34,7 +34,7 @@ steps:
       arguments: >-
         -t:AcquireAndroidTarget,SignAndroidPackage,DeployTest${{ parameters.packageType }}s,CheckAndRecordApkSizes,RunTestApks,UndeployTestApks,RenameApkTestCases,ReportComponentFailures
         -bl:$(System.DefaultWorkingDirectory)/bin/Test${{ parameters.configuration }}/run-${{ parameters.testName }}.binlog
-        -v:n -c ${{ parameters.configuration }} ${{ parameters.extraBuildArgs }}
+        -v:d -c ${{ parameters.configuration }} ${{ parameters.extraBuildArgs }}
```
jonathanpeppers, line 37: Same here, this one is changing `Normal` to `Detailed`, but it has a `.binlog`.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
```diff
@@ -127,10 +127,12 @@ public static int GetMaxInstalledPlatform ()
 			foreach (var dir in Directory.EnumerateDirectories (Path.Combine (sdkPath, "platforms"))) {
 				int version;
 				string v = Path.GetFileName (dir).Replace ("android-", "");
+				Console.WriteLine ($"GetMaxInstalledPlatform: Parsing {v}");
 				if (!int.TryParse (v, out version))
 					continue;
 				if (version < maxInstalled)
 					continue;
+				Console.WriteLine ($"GetMaxInstalledPlatform: Setting maxInstalled to {version}");
```
jonathanpeppers, line 135: Maybe these should do `TestContext.Out.WriteLine()`, then we'd have the output in a file?

build-tools/automation/yaml-templates/apk-instrumentation.yaml
dellis1972: We should probably attach the binlog as a test attachment I think. At the moment if a test fails we have to download the entire set of test results as a zip , then find the failing test directory then look at the binlog.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
dellis1972: the ProjectTools assembly doesn't reference NUnit. Which is why I used Console here.

jonathanpeppers (approved):

PR #6088, opened 2021-07-14
[docs] add some tips for .NET 6 Mono runtime packs

jonathanpeppers:
This gives a few examples on how to test a specific dotnet/runtime build in a .NET 6 Android application.

### DIFF
--- a/Documentation/workflow/DevelopmentTips.md
+++ b/Documentation/workflow/DevelopmentTips.md
@@ -373,3 +373,69 @@ steps rely on having the Android NDK installed.
         (gdb) set sysroot /tmp/gdb-symbols
         (gdb) set solib-search-path /tmp/gdb-symbols
         (gdb) target remote :50999
+
+# .NET 6 Tips
+
+## Finding Mono runtime packs
+
+`main` builds of the Mono runtime packs are on the following NuGet
+feed, such as this `nuget.config`:
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="dotnet6" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json" />
+  </packageSources>
+</configuration>
+```
+
+Internal MS users can view packages on Azure DevOps here:
+
+* https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet6
+
+To find the commit of a given package version, locate the `.nuspec`
+file inside the package, and look for:
+
+```xml
+<repository type="git" url="https://github.com/dotnet/runtime" commit="7bd472498e690e9421df86d5a9d728faa939742c" />
+```
+
+## Testing different Mono runtime pack versions
+
+One common scenario that comes up -- how does one test a specific
+dotnet/runtime build along with a .NET 6 Android application?
+
+The simplest way would be to copy individual files on top of the NuGet
+cache, such as:
+
+* `~/.nuget/packages/microsoft.netcore.app.runtime.mono.android-arm/`
+* `~/.nuget/packages/microsoft.netcore.app.runtime.mono.android-arm64/`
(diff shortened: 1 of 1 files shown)


Documentation/workflow/DevelopmentTips.md
```diff
@@ -373,3 +373,69 @@ steps rely on having the Android NDK installed.
...
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <add key="dotnet6" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json" />
+  </packageSources>
+</configuration>
+```
+
+Internal MS users can view packages on Azure DevOps here:
```
jonpryor, line 393: This URL is public, in that I can access it from an Incognito window without logging in.

jonpryor: and we should write out "Microsoft", not use "MS". :-)

Documentation/workflow/DevelopmentTips.md
```diff
@@ -373,3 +373,69 @@ steps rely on having the Android NDK installed.
...
+
+* https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet6
+
+To find the commit of a given package version, locate the `.nuspec`
```
jonpryor, line 397: I don't understand how to get from the package listing above to the `.nuspec` file "inside the package".  Does the package need to be Downloaded first?

Documentation/workflow/DevelopmentTips.md
```diff
@@ -373,3 +373,69 @@ steps rely on having the Android NDK installed.
...
+```xml
+<repository type="git" url="https://github.com/dotnet/runtime" commit="7bd472498e690e9421df86d5a9d728faa939742c" />
+```
+
+## Testing different Mono runtime pack versions
+
+One common scenario that comes up -- how does one test a specific
+dotnet/runtime build along with a .NET 6 Android application?
+
+The simplest way would be to copy individual files on top of the NuGet
```
jonpryor, line 409: I'm not sure I'd call this "simple", as there are *lots* of files in there, most of which is inscrutable.  My `$HOME/.nuget/packages/microsoft.netcore.app.runtime.mono.android-arm` contains 7 directories; which one should I update?  How do I figure that out?  (Answer: by reading an existing build log, and updating whichever version was implicated.  Requires reading build logs!)

`$HOME/.nuget/packages/microsoft.netcore.app.runtime.mono.android-arm/6.0.0-preview.7.21352.16` contains 230 files as well.  How is anyone supposed to know which files to update?

This isn't "simple".  This is the "advanced" "I think i know what I'm doing, maybe" scenario.  It just *looks* simple.

Documentation/workflow/DevelopmentTips.md
```diff
@@ -373,3 +373,69 @@ steps rely on having the Android NDK installed.
...
+cache, such as:
+
+* `~/.nuget/packages/microsoft.netcore.app.runtime.mono.android-arm/`
+* `~/.nuget/packages/microsoft.netcore.app.runtime.mono.android-arm64/`
+* `~/.nuget/packages/microsoft.netcore.app.runtime.mono.android-x86/`
+* `~/.nuget/packages/microsoft.netcore.app.runtime.mono.android-x64/`
+
+A second way is to add this MSBuild target to your Android `.csproj` file:
```
jonpryor, line 417: *This* is the simple scenario. :-)

Documentation/workflow/DevelopmentTips.md
jonathanpeppers: I didn't realize that site is public -- so took that out.

Documentation/workflow/DevelopmentTips.md
jonathanpeppers: I wrote a few more sentences on this.

jonathanpeppers:
@jonpryor the latest changes should be easier to follow now.

PR #6094, opened 2021-07-16
Revert "[tests] try to use jdk 11 for device tests"

jonathanpeppers:
This reverts commit d14f4a2349f6be660ab6a109a1065ffb08fb12dd.

Wanting to see the error we saw for: https://github.com/xamarin/xamarin-android/pull/6084

```
/Library/Frameworks/Mono.framework/External/xbuild/Xamarin/Android/Xamarin.Android.Common.targets(1728,3):
error JAVAC0000: java.lang.AssertionError: annotationType() : unrecognized Attribute name MODULE (class com.sun.tools.javac.util.SharedNameTable$NameImpl)
```

What is the actual case this fails? Do we even need #6084?

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1033,7 +1033,7 @@ stages:
     parameters:
       job_name: mac_msbuilddevice_tests
       job_suffix: Legacy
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
Don't think we need this anymore.

PR #6104, opened 2021-07-20
Use `mono_unhandled_exception` for NET6

grendello:
Context: https://github.com/dotnet/runtime/pull/55904#discussion_r672577807
Context: https://github.com/xamarin/xamarin-android/pull/4927#issuecomment-875864999
Context: https://github.com/xamarin/xamarin-android/pull/4927#issuecomment-875876255

Xamarin.Android has been using the `AppDomain.DoUnhandledException` API
since the dawn of time to propagate uncaught Java exceptions to the
managed world.  However, said API (and AppDomains) are gone from the
NET6 MonoVM runtime and we need to switch to something else - the
`mono_unhandled_exception` native API.

This commit introduces an internal call,
`JNIEnv.monodroid_unhandled_exception`, which is used instead of the
older mechanism when targetting NET6 and which calls the native API
mentioned above.

Add a device integration test which makes sure the uncaught exceptions
are propagated as required.

### DIFF
--- a/src/Mono.Android/Android.Runtime/JNIEnv.cs
+++ b/src/Mono.Android/Android.Runtime/JNIEnv.cs
@@ -242,7 +242,9 @@ namespace Android.Runtime {
 		}
 
 		static Action<Exception> mono_unhandled_exception = null!;
+#if !NETCOREAPP
 		static Action<AppDomain, UnhandledExceptionEventArgs> AppDomain_DoUnhandledException = null!;
+#endif // ndef NETCOREAPP
 
 		static void Initialize ()
 		{
@@ -253,6 +255,7 @@ namespace Android.Runtime {
 					mono_unhandled_exception = (Action<Exception>) Delegate.CreateDelegate (typeof(Action<Exception>), mono_UnhandledException);
 			}
 
+#if !NETCOREAPP
 			if (AppDomain_DoUnhandledException == null) {
 				var ad_due = typeof (AppDomain)
 					.GetMethod ("DoUnhandledException",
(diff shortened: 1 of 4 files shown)


src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -265,8 +268,14 @@ static void Initialize ()
 						typeof (Action<AppDomain, UnhandledExceptionEventArgs>), ad_due);
 				}
 			}
+#endif
 		}
 
+#if NETCOREAPP
+		[DllImport (AndroidRuntime.InternalDllName, CallingConvention = CallingConvention.Cdecl)]
+		extern static void monodroid_unhandled_exception (Exception javaException);
```
jonpryor, line 276: If passing a managed type like this, doesn't this need to be an internal call, not a DllImport?

grendello: I think it's treated as an `IntPtr` in this case, it appears to work with the test app

jonpryor: @vargaz thinks it should be an icall: https://discord.com/channels/732297728826277939/732297965019988138/867104696488427580

It might "work", but I don't think it's *guaranteed* to work.

grendello: Done

src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -242,7 +242,9 @@ static void ManualJavaObjectDispose (Java.Lang.Object obj)
 		}
 
 		static Action<Exception> mono_unhandled_exception = null!;
+#if !NETCOREAPP
 		static Action<AppDomain, UnhandledExceptionEventArgs> AppDomain_DoUnhandledException = null!;
+#endif
```
jonpryor, line 247: Please comment on `#else` and `#endif`.

dellis1972 (approved):

PR #6106, opened 2021-07-20
[debugger] Calling embedded api and avoid changing System.Diagnostics.Debugger class.

thaystg:
Waiting this to be merged https://github.com/dotnet/runtime/pull/56071

### DIFF
--- a/src/Mono.Android/Android.Runtime/JNIEnv.cs
+++ b/src/Mono.Android/Android.Runtime/JNIEnv.cs
@@ -241,19 +241,19 @@ namespace Android.Runtime {
 			GC.SuppressFinalize (obj);
 		}
 
-		static Action<Exception> mono_unhandled_exception = null!;
+#if NETCOREAPP
+		internal static Action<Exception> mono_unhandled_exception = monodroid_debugger_unhandled_exception;
+#else  // NETCOREAPP
+		internal static Action<Exception> mono_unhandled_exception = null!;
+#endif  // NETCOREAPP
+
 		static Action<AppDomain, UnhandledExceptionEventArgs> AppDomain_DoUnhandledException = null!;
 
 		static void Initialize ()
 		{
 			if (mono_unhandled_exception == null) {
-#if NETCOREAPP
-				var mono_UnhandledException = typeof (Android.Runtime.JNIEnv)
-					.GetMethod ("monodroid_debugger_unhandled_exception", BindingFlags.NonPublic | BindingFlags.Static);
-#else
 				var mono_UnhandledException = typeof (System.Diagnostics.Debugger)
 					.GetMethod ("Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
-#endif
 				if (mono_UnhandledException != null)
 					mono_unhandled_exception = (Action<Exception>) Delegate.CreateDelegate (typeof(Action<Exception>), mono_UnhandledException);
 			}
--- a/src/Mono.Android/Android.Runtime/JNIEnv.cs
+++ b/src/Mono.Android/Android.Runtime/JNIEnv.cs
@@ -247,8 +247,13 @@ namespace Android.Runtime {
 		static void Initialize ()
 		{
 			if (mono_unhandled_exception == null) {
+#if NETCOREAPP
 				var mono_UnhandledException = typeof (Android.Runtime.JNIEnv)
 					.GetMethod ("monodroid_debugger_unhandled_exception", BindingFlags.NonPublic | BindingFlags.Static);
+#else
+				var mono_UnhandledException = typeof (System.Diagnostics.Debugger)
+					.GetMethod ("Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
+#endif
 				if (mono_UnhandledException != null)
 					mono_unhandled_exception = (Action<Exception>) Delegate.CreateDelegate (typeof(Action<Exception>), mono_UnhandledException);
 			}
@@ -682,8 +687,10 @@ namespace Android.Runtime {
 		[MethodImplAttribute(MethodImplOptions.InternalCall)]
 		static extern unsafe IntPtr monodroid_typemap_managed_to_java (Type type, byte* mvid);
 
+#if NETCOREAPP
 		[MethodImplAttribute(MethodImplOptions.InternalCall)]
 		static extern unsafe void monodroid_debugger_unhandled_exception (Exception e);
+#endif
 
 		internal static void LogTypemapTrace (StackTrace st)
 		{
--- a/src/Mono.Android/Android.Runtime/JNIEnv.cs
+++ b/src/Mono.Android/Android.Runtime/JNIEnv.cs
@@ -247,8 +247,8 @@ namespace Android.Runtime {
 		static void Initialize ()
 		{
 			if (mono_unhandled_exception == null) {
-				var mono_UnhandledException = typeof (System.Diagnostics.Debugger)
-					.GetMethod ("Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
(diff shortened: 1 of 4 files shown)


src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -247,8 +247,13 @@ static void ManualJavaObjectDispose (Java.Lang.Object obj)
 		static void Initialize ()
 		{
 			if (mono_unhandled_exception == null) {
+#if NETCOREAPP
```
jonpryor, line 250: This feels "wasteful", in that we're using Reflection to create a delegate for a method which is directly convertible to the desired type *without* using Reflection.

Thus, I'd prefer:

```csharp
#if NETCOREAPP
				mono_unhandled_exception = monodroid_debugger_unhandled_exception;
#else   // NETCOREAPP
				var mono_UnhandledException = typeof (System.Diagnostics.Debugger)
					.GetMethod ("Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
				if (mono_UnhandledException != null)
					mono_unhandled_exception = (Action<Exception>) Delegate.CreateDelegate (typeof(Action<Exception>), mono_UnhandledException);
#endif  // NETCOREAPP
```

src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -247,8 +247,13 @@ static void ManualJavaObjectDispose (Java.Lang.Object obj)
...
+				var mono_UnhandledException = typeof (Android.Runtime.JNIEnv)
+					.GetMethod ("monodroid_debugger_unhandled_exception", BindingFlags.NonPublic | BindingFlags.Static);
+#else
 				var mono_UnhandledException = typeof (System.Diagnostics.Debugger)
 					.GetMethod ("Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
+#endif
```
jonpryor, line 256: Code convention: `#else` and `#endif` statements should have a comment "describing" the condition they're associated with, e.g.

```csharp
#endif  // NETCOREAPP
```

src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
```diff
@@ -20,6 +20,12 @@ static void get_runtime_types ()
 				"Mono_UnhandledException", BindingFlags.NonPublic | BindingFlags.Static);
 			if (mono_unhandled_exception_method == null)
 				AndroidEnvironment.FailFast ("Cannot find System.Diagnostics.Debugger.Mono_UnhandledException");
+#endif
+#if NETCOREAPP
+			mono_unhandled_exception_method = typeof (Android.Runtime.JNIEnv).GetMethod (
```
jonpryor, line 25: Instead of duplicating this logic, we should instead make `JNIEnv.mono_unhandled_exception` *`internal`*, so that we can directly access it here:

```csharp
mono_unhandled_exception_method = JNIEnv.mono_unhandled_exception.Method;
```

src/Mono.Android/Android.Runtime/JNIEnv.cs
thaystg: Fixed :)

src/Mono.Android/Android.Runtime/JNIEnv.cs
thaystg: Fixed :)

src/Mono.Android/Android.Runtime/JNINativeWrapper.cs
thaystg: Fixed :)

grendello (approved):

thaystg:
@jonpryor now that it's already merged on dotnet/runtime preview 7, I think this is ready for review. Do I need to bump runtime version anywhere to get the exported function working?

jonpryor:
@thaystg [asked](https://github.com/xamarin/xamarin-android/pull/6106#issuecomment-884924663):

> Do I need to bump runtime version anywhere to get the exported function working?

Kinda; dotnet-maestro does bumps, e.g. PR #6110, but PR #6110 (currently) bumps to dotnet/runtime/main@bd35632, which *doesn't* include `mono_debugger_agent_unhandled_exception()`, added in dotnet/runtime/main@3746672c5b78f0aacf43ee1d170c3ca88c34ce91.

Additionally, PR #6110 has introduced issues, e.g. satellite assembly handling has changed (?!): https://github.com/xamarin/xamarin-android/pull/6110#issuecomment-885086836

I don't know when we can bump main/rc1.

Which leaves .NET 6 Preview 7; we have a PR to bump that as well: PR #6112.  PR #6112 *fails to build* for not currently understood reasons, *and* PR #6112 currently bumps to commit dotnet/runtime@ ba08d9a305e2e0debeed1c96b3137b44305466b8, which doesn't include dotnet/runtime@743bd894c40b00b93e805db9d62664c915e0c0ee.

@thaystg: Would you be able to investigate why PR #6112 is failing to build? Once we get #6112 building and bumped to something that includes dotnet/runtime@743bd894c, we could re-target *this* PR (#6106) to target the release/6.0.1xx-preview7 branch.

jonpryor:
latest PR build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5012864&view=results

thaystg:
> latest PR build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5012864&view=results

I think the build is finished and passed, can we merge?
Should I do something manually to backport to preview 7?

jonpryor:
Commit message:

```markdown
[One .NET] Use Mono embedding API for unhandled exceptions (#6106)

Context: https://github.com/dotnet/runtime/pull/56071
Context: https://github.com/xamarin/xamarin-android/pull/4877
Context: https://github.com/xamarin/xamarin-android/pull/4927#discussion_r490457925
Context: https://github.com/xamarin/xamarin-android/pull/4927#issuecomment-875864999

Context: https://github.com/xamarin/monodroid/commit/3e9de5a51bd46263b08365ef18bed1ae472122d8
Context: https://github.com/xamarin/monodroid/commit/b0f85970102d43bab9cd860a8e8884d136d766b3
Context: https://github.com/xamarin/monodroid/commit/12a012e00b4533d586ef31ced33351b63c9de883

What should happen when an exception is thrown and a debugger is
attached?

This is in fact a loaded question: there's what Xamarin.Android
(Legacy) *has* done, vs. what .NET 6 Android for .NET does, vs.
what "should" happen.

What "should" happen is easiest:

 1. We should behave like a "normal" Desktop .NET app when a debugger
    is attached, AND

 2. We shouldn't corrupt JVM state.

Unfortunately, (1)+(2) is currently not possible, in part because
Java doesn't have an equivalent to Windows' [two attempt][0] debugger
notification infrastructure.
See xamarin/xamarin-android#4877 for details.

What Legacy Xamarin.Android does is also detailed in
xamarin/xamarin-android#4877, and relies on the
`Debugger.Mono_UnhandledException()` method in order to alert an
attached debugger that there is an exception to show to the user.

However, `Debugger.Mono_UnhandledException()` never made it to the
`dotnet/runtime` repo.  It never existed there.

PR dotnet/runtime#56071 added a new
`mono_debugger_agent_unhandled_exception()` Mono embedding API which
is equivalent to `Debugger.Mono_UnhandledException()` for use with
.NET 6 + MonoVM.

Update `src/Mono.Android` and `src/monodroid` so that
`mono_debugger_agent_unhandled_exception()` is used to alert the
debugger that an exception has been thrown at a JNI boundary.
This should allow .NET 6 + Android to have equivalent exception
handling semantics as legacy Xamarin.Android.

[0]: https://docs.microsoft.com/en-us/windows/win32/debug/debugger-exception-handling
```

Issue #6107, opened 2021-07-20
[One .NET] Missing AndroidSdkSupportedTargetPlatformVersion values


jpobst:
The file `Microsoft.Android.Sdk.SupportedPlatforms.targets` should include `<AndroidSdkSupportedTargetPlatformVersion>` versions for every API level we support (`21` - `30`).  However in Preview 6, it only includes `30`:

```xml
<Project>
  <PropertyGroup>
    <TargetPlatformSupported>true</TargetPlatformSupported>
    <TargetPlatformVersion Condition=" '$(TargetPlatformVersion)' == '' ">30.0</TargetPlatformVersion>
  </PropertyGroup>
  <ItemGroup>
    <AndroidSdkSupportedTargetPlatformVersion Include="30.0" />
    <SdkSupportedTargetPlatformVersion Include="@(AndroidSdkSupportedTargetPlatformVersion)" />
  </ItemGroup>
</Project>
```

This means a user cannot currently use eg: `net6.0-android28.0` as a `<TargetFramework>`:

```
error NETSDK1140: 28.0 is not a valid TargetPlatformVersion for android. Valid versions include:
error NETSDK1140: 30.0
```

@jonathanpeppers says this is a regression and that it used to work.

jonathanpeppers:
The issue appears to be that at the time this runs:

https://github.com/xamarin/xamarin-android/blob/6f9bc6b20927eb64800e4b42cf43f8e52b0b1a73/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets#L220-L224

Only a single `AndroidApiInfo.xml` exists:

![image](https://user-images.githubusercontent.com/840039/126506061-de600431-b7b1-4cc9-a969-c5e628046394.png)

jpobst:
Note that there is currently some discussion that maybe we don't want `net6.0-android28.0` to exist after all that we should resolve before we fix this:

https://github.com/dotnet/sdk/issues/19075#issuecomment-883628444

jonpryor:
I'm reminded of our `compileSdkVersion` project: https://github.com/xamarin/xamarin-android/projects/3

…which now needs to be updated for our .NET 6 world order.

Here's how I think things should play out:

[...]

PR #6111, opened 2021-07-21
[One .NET] implement $(SupportedPlatformOSVersion) for minSdkVersion

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/issues/6107#issuecomment-884391615

In .NET 6, we are wanting `AndroidManifest.xml` in project templates
to no longer contain:

    <uses-sdk android:minSdkVersion="21" android:targetSdkVersion="30" />

This is because we can define these values such as:

    <TargetFramework>net6.0-android30</TargetFramework>
    <SupportedPlatformOSVersion>21.0</SupportedPlatformOSVersion>

There is not really a reason to put this information in two places.
`$(SupportedPlatformOSVersion)` is also required for enabling C#
compiler warnings such as:

    MainActivity.cs(28,4): warning CA1416: This call site is reachable on: 'Android' 21.0 and later. 'View.AccessibilityTraversalAfter.get' is only supported on: 'android' 22.0 and later.

This is an example of calling an API from 22, with a
`$(SupportedPlatformOSVersion)` for 21.

If `<uses-sdk/>` is still *used* in `AndroidManifest.xml` in a
project, it will be preferred as the final source of truth.

One other improvement is that `$(SupportedPlatformOSVersion)` is
expected to include `.0`, or you get the warning:

    UnnamedProject.AssemblyInfo.cs(21,12): warning CA1418: Version '21' is not valid for platform 'Android'. Use a version with 2-4 parts for this platform.

Android unique in that this value is always an integer. In order for
users to be able to specify `21`, I added a check in
`Microsoft.Android.Sdk.DefaultProperties.targets` so that we append
`.0` when needed.

I added some tests around this scenario.

### DIFF
--- a/src/Microsoft.Android.Templates/android-bindinglib/.template.config/template.json
+++ b/src/Microsoft.Android.Templates/android-bindinglib/.template.config/template.json
@@ -15,5 +15,14 @@
   "primaryOutputs": [
     { "path": "AndroidBinding1.csproj" }
   ],
+  "symbols": {
+    "supportedOSVersion": {
+      "type": "parameter",
+      "description": "Overrides $(SupportedOSPlatformVersion) in the project",
+      "datatype": "string",
+      "replaces": "SUPPORTED_OS_PLATFORM_VERSION",
+      "defaultValue": "21"
+    }
+  },
   "defaultName": "AndroidBinding1"
 }
--- a/src/Microsoft.Android.Templates/android-bindinglib/AndroidBinding1.csproj
+++ b/src/Microsoft.Android.Templates/android-bindinglib/AndroidBinding1.csproj
@@ -1,6 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net6.0-android</TargetFramework>
+    <SupportedOSPlatformVersion>SUPPORTED_OS_PLATFORM_VERSION</SupportedOSPlatformVersion>
     <RootNamespace Condition="'$(name)' != '$(name{-VALUE-FORMS-}safe_namespace)'">AndroidBinding1</RootNamespace>
   </PropertyGroup>
 </Project>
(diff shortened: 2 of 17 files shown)


dellis1972 (approved):
looks ok

jonathanpeppers:
This broke a lot of tests:
```
(_CheckForSupportedOSPlatformVersionHigherThanTargetPlatformVersion target) -> 
/Users/runner/Library/Android/dotnet/sdk/6.0.100-rc.1.21372.10/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.TargetFrameworkInference.targets(185,18):
error MSB4184: The expression "[MSBuild]::VersionGreaterThan(.0, 30.0)" cannot be evaluated. Version string was not in a correct format. 
```

I'm a bit confused on why I didn't see it locally, looking into it...

jonathanpeppers:
The problem before was I needed to check if `$(SupportedOSPlatformVersion)`was blank:
```xml
<!-- $(SupportedOSPlatformVersion) must be '21.0', but we should support integer values like '21' -->
<SupportedOSPlatformVersion Condition=" '$(SupportedOSPlatformVersion)' != '' and !$(SupportedOSPlatformVersion.Contains('.')) ">$(SupportedOSPlatformVersion).0</SupportedOSPlatformVersion>
```

jonathanpeppers:
`TargetFrameworkVersion` should always be 6 for .NET 6. You can see the design about this here:

https://github.com/dotnet/designs/blob/main/accepted/2021/net6.0-tfms/net6.0-tfms.md

For Android class libraries, you can put `net6.0-android31` or `net6.0-android32` if you really need this. If it is an Android application, I'd recommend putting `net6.0-android` and let the Android workload choose the default for you.

In the future, you might consider opening an issue instead of commenting on an old PR or comment in the [#android channel on Discord](https://aka.ms/dotnet-discord). You might get a quicker response that way, thanks!

PR #6117, opened 2021-07-23
[tests] we don't need to check if Clean deletes directories

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/6112

The .NET 6 Preview 7 bump had a test failure:

    CleanBasicBindingLibrary("class-parse")
    obj/Release should have no directories.
    Expected: <empty>
    But was:  < "/Users/runner/work/1/s/bin/TestRelease/temp/CleanBasicBindingLibraryclass-parse/obj/Release/refint" >

`refint` is a new directory, see:

* https://github.com/dotnet/sdk/pull/19114/commits/e424c0eae2dcbdb71ff60014c6bdad82ddb3175d
* https://github.com/dotnet/msbuild/blob/9e576281e638d60701ca34411e2483bed01e35c7/src/Tasks/Microsoft.Common.CurrentVersion.targets#L397

Clean doesn't actually delete directories, no need to test for this.

The test already is checking if any files exist, and that should be
sufficient.


dellis1972 (approved):

jonathanpeppers:
The changed test still passes, going to merge.

image

Issue #6076, opened 2021-07-07
Android project templates use latest C# language idioms


mhutch:
In .NET 6, the included C# project templates will use the latest language idioms.

See the [parent Epic](https://github.com/dotnet/templating/issues/3359) for details.

Depends on:

- [ ] https://github.com/dotnet/sdk/pull/18459
- [x] https://github.com/xamarin/xamarin-android/issues/6075

Issue #6075, opened 2021-07-07
Add Android default global usings in MSBuild targets


mhutch:
Following the pattern in https://github.com/dotnet/sdk/pull/18459, the Android SDK targets should contain default global usings that can be disabled by an MSBuild property.

jonpryor:
As per https://github.com/xamarin/xamarin-android/pull/6118#issuecomment-886863501, I'm starting to wonder if *anything* should be added to `@(Import)` for Android apps.  It feels like anything we add would be likely to introduce type resolution conflicts for MAUI and BCL types.

mhutch:
@jonpryor if there are type conflicts between system types and android types we can make aliasing choices and emit those too.

PR #6118, opened 2021-07-23
[One .NET] support latest C# 10 language features

jonathanpeppers:
Context: https://github.com/dotnet/sdk/issues/19521
Fixes: https://github.com/xamarin/xamarin-android/issues/6075
Fixes: https://github.com/xamarin/xamarin-android/issues/6076

We need to make two sets of changes for C# 10:

1. Support "global usings". Our .NET 6 templates should have no
   `using` statements at the top of `.cs` files.
2. Use `$(Nullable)` `enable` by default in project templates.

To test this, our .NET 6 MSBuild tests use `Nullable=enable` and
`ImplicitUsings=enable` by default and do not include `using`
statements in `.cs` files.

I've made a new `MainActivity.cs` for our .NET 6 MSBuild tests. The
"legacy" Xamarin.Android tests will use the original file.

Our default `global using` are:

```csharp
global using global::Android.App;
global using global::Android.Widget;
global using Bundle = global::Android.OS.Bundle;
````

The last one is intentionally not bringing in `Android.OS`, because
`Android.OS.Environment` would conflict with `System.Environment`.

So `AutoImport.props` should become:

```xml
<ItemGroup Condition=" '$(TargetPlatformIdentifier)' == 'android' and ('$(ImplicitUsings)' == 'true' or '$(ImplicitUsings)' == 'enable') ">
  <Using Include="Android.App" />
  <Using Include="Android.Widget" />
  <Using Include="Android.OS.Bundle" Alias="Bundle" />
</ItemGroup>
```

So these items are present at the time `.csproj` files are evaluated.

Any templates will add:

```xml
<Nullable>enable</Nullable>
<ImplicitUsings>enable</ImplicitUsings>
```

If users want to configure these settings, they can remove
`$(ImplicitUsings)` from the `.csproj` completely or remove specific
`@(Using)` items:

```xml
<ItemGroup>
  <Using Remove="Android.App" />
</ItemGroup>
```

### DIFF
--- a/src/Microsoft.Android.Templates/android-bindinglib/AndroidBinding1.csproj
+++ b/src/Microsoft.Android.Templates/android-bindinglib/AndroidBinding1.csproj
@@ -3,5 +3,7 @@
     <TargetFramework>net6.0-android</TargetFramework>
     <SupportedOSPlatformVersion>SUPPORTED_OS_PLATFORM_VERSION</SupportedOSPlatformVersion>
     <RootNamespace Condition="'$(name)' != '$(name{-VALUE-FORMS-}safe_namespace)'">AndroidBinding1</RootNamespace>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
   </PropertyGroup>
 </Project>
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
@@ -205,6 +205,7 @@ namespace Xamarin.Android.Build.Tests
 			Assert.IsTrue (dotnet.New ("android-activity"), "`dotnet new android-activity` should succeed");
 			Assert.IsTrue (dotnet.New ("android-layout", Path.Combine (dotnet.ProjectDirectory, "Resources", "layout")), "`dotnet new android-layout` should succeed");
 			Assert.IsTrue (dotnet.Build (), "`dotnet build` should succeed");
+			dotnet.AssertHasNoWarnings ();
 		}
 
 		[Test]
@@ -519,7 +520,7 @@ namespace Xamarin.Android.Build.Tests
 			};
 			// Call AccessibilityTraversalAfter from API level 22
 			// https://developer.android.com/reference/android/view/View#getAccessibilityTraversalAfter()
-			proj.MainActivity = proj.DefaultMainActivity.Replace ("button.Click", "button.AccessibilityTraversalAfter.ToString ();\nbutton.Click");
+			proj.MainActivity = proj.DefaultMainActivity.Replace ("button!.Click", "button!.AccessibilityTraversalAfter.ToString ();\nbutton!.Click");
 
 			var dotnet = CreateDotNetBuilder (proj);
 			Assert.IsTrue (dotnet.Build (), "`dotnet build` should succeed");
@@ -614,7 +615,7 @@ namespace Xamarin.Android.Build.Tests
 		public void XamarinLegacySdk ()
 		{
 			var proj = new XASdkProject (outputType: "Library") {
-				Sdk = "Xamarin.Legacy.Sdk/0.1.0-alpha2",
+				Sdk = "Xamarin.Legacy.Sdk/0.1.0-alpha4",
 				Sources = {
 					new AndroidItem.AndroidLibrary ("javaclasses.jar") {
 						BinaryContent = () => ResourceData.JavaSourceJarTestJar,
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
@@ -23,7 +23,7 @@
       "Size": 2681
     },
     "assemblies/UnnamedProject.dll": {
-      "Size": 3186
+      "Size": 3535
     },
     "classes.dex": {
       "Size": 345240
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Xamarin.ProjectTools.csproj
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Xamarin.ProjectTools.csproj
@@ -11,72 +11,7 @@
   <Import Project="..\..\..\..\external\xamarin-android-tools\src\Microsoft.Android.Build.BaseTasks\MSBuildReferences.projitems" />
   <ItemGroup>
     <Compile Remove="Resources\**\*.cs" />
-    <EmbeddedResource Include="Resources\AndroidX\Tabbar.xml">
-      <LogicalName>Xamarin.ProjectTools.Resources.AndroidX.Tabbar.xml</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\AndroidX\Toolbar.xml">
-      <LogicalName>Xamarin.ProjectTools.Resources.AndroidX.Toolbar.xml</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\Wear\LayoutMain.axml">
-      <LogicalName>Xamarin.ProjectTools.Resources.Wear.LayoutMain.axml</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\Wear\LayoutRectMain.axml">
-      <LogicalName>Xamarin.ProjectTools.Resources.Wear.LayoutRectMain.axml</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\Wear\LayoutRoundMain.axml">
-      <LogicalName>Xamarin.ProjectTools.Resources.Wear.LayoutRoundMain.axml</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\Wear\MainActivity.cs">
-      <LogicalName>Xamarin.ProjectTools.Resources.Wear.MainActivity.cs</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\Wear\Strings.xml">
-      <LogicalName>Xamarin.ProjectTools.Resources.Wear.Strings.xml</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\Base\AndroidManifest.xml">
-      <LogicalName>Xamarin.ProjectTools.Resources.Base.AndroidManifest.xml</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\Base\Icon.png">
-      <LogicalName>Xamarin.ProjectTools.Resources.Base.Icon.png</LogicalName>
-    </EmbeddedResource>
-    <EmbeddedResource Include="Resources\Base\LayoutMain.axml">
-      <LogicalName>Xamarin.ProjectTools.Resources.Base.LayoutMain.axml</LogicalName>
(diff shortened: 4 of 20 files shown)


jonathanpeppers:
This is a draft, because I suspect we might get test failures on this one.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ImplicitNamespaceImports.targets
```diff
@@ -0,0 +1,22 @@
...
+<Project>
+  <!--
+    There's a master switch, $(DisableImplicitNamespaceImports),
+    which, if set, will prevent all implicit namespace imports.
+    This is done automatically, we don't have to check
+    $(DisableImplicitNamespaceImports) here.
+  -->
+  <ItemGroup Condition=" '$(DisableImplicitNamespaceImports_Android)' != 'true' ">
+    <Import Include="Android.App" />
+    <Import Include="Android.OS" />
```
jonpryor, line 18: Thus begins the question: *Should* `Android.OS` be in the default namespace list?

It has been in our default template, *but*…. What's been *lacking* in our default templates? `using System`!

Unfortunately, both `System` and `Android.OS` contain an `Environment` type.

I *assume* that if we have both `System` and `Android.OS` "in scope" at the same time, any "unqualified" use of `Environment` will result in a C# error, as it'll be ambiguous.

I also have to assume that `System` is in the default `@(Import)` list.

Thus, the original question: *Should* `Android.OS` be in `@(Import)`?

If we *don't* include it, then we'll need to update our templates to fully qualify `Android.OS.Bundle`.  I think that would be "fine".

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ImplicitNamespaceImports.targets
```diff
@@ -0,0 +1,22 @@
...
+***********************************************************************************************
+-->
+<Project>
+  <!--
+    There's a master switch, $(DisableImplicitNamespaceImports),
+    which, if set, will prevent all implicit namespace imports.
+    This is done automatically, we don't have to check
+    $(DisableImplicitNamespaceImports) here.
+  -->
+  <ItemGroup Condition=" '$(DisableImplicitNamespaceImports_Android)' != 'true' ">
```
jonpryor, line 16: Is there *also* a `$(DisableImplicitNamespaceImports)` property?  If so, should this group be conditional on `'$(DisableImplicitNamespaceImports)' != 'true'` as well?  Or will `Microsoft.Android.Sdk.ImplicitNamespaceImports.targets` not be `<Import/>`ed when `'$(DisableImplicitNamespaceImports)' != 'true'`?

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ImplicitNamespaceImports.targets
```diff
@@ -0,0 +1,22 @@
...
+    <Import Include="Android.App" />
+    <Import Include="Android.OS" />
+    <Import Include="Android.Runtime" />
```
jonpryor, line 19: Another one that I'm not sure should be part of the default namespace set.  Sure, our default templates have used it for ages, but should it really be in the default `@(Import)` list?

I don't think excluding this namespace would break any of the templates in this PR.  Which suggests it's not *that* important.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Xamarin.ProjectTools.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
jonpryor, line 1: BOM removal?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Xamarin.ProjectTools.csproj
```diff
@@ -9,72 +9,7 @@
   <Import Project="..\..\..\..\external\xamarin-android-tools\src\Microsoft.Android.Build.BaseTasks\MSBuildReferences.projitems" />
   <ItemGroup>
     <Compile Remove="Resources\**\*.cs" />
-    <EmbeddedResource Include="Resources\AndroidX\Tabbar.xml">
```
jonpryor, line 12: I'm not sure it's "safe" to remove all of these, largely because of the `%(LogicalName)` overrides.

(That, and I can never remember the semantics for the default `%(LogicalName)` value…)

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ImplicitNamespaceImports.targets
jonathanpeppers: The target that uses the `@(Import)` item group doesn't run if you set `$(DisableImplicitNamespaceImports)`:

https://github.com/dotnet/sdk/blob/93b432bd8e73bb68aa4621db8a967d62e6f0cc23/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.GenerateImplicitNamespaceImports.targets#L35-L38

Here is the commit that implements this:

https://github.com/dotnet/sdk/commit/85943958e5a01d6bbd39a947aec2a4a37b8ba7cf

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Xamarin.ProjectTools.csproj
jonathanpeppers: Is our `.editorconfig` wrong? That's what caused this:

https://github.com/xamarin/xamarin-android/blob/e4debf72015549e6e19cbfb247a47fead6ce7a01/.editorconfig#L7-L15

We could change `.csproj` to `utf8-bom`, but I don't know what it's supposed to be actually.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Xamarin.ProjectTools.csproj
jonathanpeppers: The `%(LogicalName)` were all listing the default value over and over... that is why I removed them.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.ImplicitNamespaceImports.targets
jonathanpeppers: Yes, I think we should remove `Android.OS` and `Android.Runtime`.

So the list is going to be:
```xml
<Import Include="Android.App" />
<Import Include="Android.Widget" />
```

I thought about `Android.Views` -- but that could cause some upstream complications in dotnet/maui. Our template doesn't even have it *now*, so we probably don't need it.

jonathanpeppers:
One complication that came up in a test:
```
error CS0104: 'Application' is an ambiguous reference between 'Android.App.Application' and 'Xamarin.Forms.Application'
```

So I suspect that dotnet/maui is going to set `$(DisableImplicitNamespaceImports_Android)`. Because it has to? There are so many types that will conflict.

jonpryor:
@jonathanpeppers  [wrote](https://github.com/xamarin/xamarin-android/pull/6118#issuecomment-885915457):

> One complication that came up in a test:
> 
>     error CS0104: 'Application' is an ambiguous reference between 'Android.App.Application' and 'Xamarin.Forms.Application'

This suggests that we might not want `Android.App` in the default `@(Import)` *either*, since *lots* of our customers will be using MAUI.

Will `Android.Widget` conflict with anything?

The "better answer" here *may* be to just not add *anything* to `@(Import)`, as anything we add will likely conflict with MAUI…

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
```diff
@@ -606,7 +607,7 @@ void CreateEmptyFile (params string [] paths)
 		public void XamarinLegacySdk ()
 		{
 			var proj = new XASdkProject (outputType: "Library") {
-				Sdk = "Xamarin.Legacy.Sdk/0.1.0-alpha2",
+				Sdk = "Xamarin.Legacy.Sdk/0.1.0-alpha3",
```
jonathanpeppers, line 610: I had to make Xamarin.Legacy.Sdk support C# 10, so I shipped a release with this fix.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
```diff
@@ -23,7 +23,7 @@
       "Size": 2676
     },
     "assemblies/UnnamedProject.dll": {
-      "Size": 3171
+      "Size": 3535
```
jonathanpeppers, line 26: Adding `Nullable=enable`, increased this assembly's size enough to fail the test...
```
Error: apkdiff: File 'assemblies/UnnamedProject.dll' has changed by 358 bytes (10.14 %). This exceeds the threshold of 5.00 %.
```

jonathanpeppers:
We need to wait on this, because there are changes to the .NET SDK coming:

1. `@(Import)` will be renamed to `@(Using)`
2. `$(DisableImplicitNamespaceImports)=false` will be `$(ImplicitUsings)=true`. Projects will have to opt into it, so we will put `$(ImplicitUsings)=true` in our project templates.
3. They might support our `global using Bundle = global::Android.OS.Bundle;` use case.

See: https://github.com/dotnet/sdk/issues/19535#issuecomment-893878263

I'd rather wait for the final implementation to land before we opt into the changes here.

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
```diff
@@ -217,7 +217,7 @@ public class Class1 {
 			};
 			proj.SetAndroidSupportedAbis ("armeabi-v7a", "arm64-v8a", "x86", "x86_64");
 			proj.SetProperty (proj.ReleaseProperties, "MonoSymbolArchive", "True");
-			proj.MainActivity = proj.DefaultMainActivity.Replace ("//${AFTER_ONCREATE}",
+			proj.MainActivity = proj.DefaultMainActivity.Replace ("//${AFTER_FORMS_INIT}",
```
jonathanpeppers, line 220: I found this test was not using a Forms-based `MainActivity.cs` before, but it was a Xamarin.Forms app! So we need to use `${AFTER_FORMS_INIT}` now and the line number changed.

src/Microsoft.Android.Templates/android-bindinglib/AndroidBinding1.csproj
```diff
@@ -3,5 +3,7 @@
     <TargetFramework>net6.0-android</TargetFramework>
     <SupportedOSPlatformVersion>SUPPORTED_OS_PLATFORM_VERSION</SupportedOSPlatformVersion>
     <RootNamespace Condition="'$(name)' != '$(name{-VALUE-FORMS-}safe_namespace)'">AndroidBinding1</RootNamespace>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
```
jonpryor, line 7: Should we update `AndroidBinding1.csproj` to *also* set `$(AndroidGenerateResourceDesigner)`=False, as per #6224?

PR #6120, opened 2021-07-23
[CI] Use NuGet to MSI conversion v2

pjcollins:
Context: https://github.com/xamarin/yaml-templates/pull/121

Uses new versions of the nuget-msi-convert templates to support creation
of .msi files for workload manifest packs via the `GenerateManifestMsi`
task available in the `Microsoft.DotNet.Build.Tasks.Workloads` arcade
sdk.


jonathanpeppers (approved):
This looks good if CI is green.

PR #6134, opened 2021-07-27
[CI] Remove temporary .NET 6 installer jobs

pjcollins:
With `dotnet workload install` and the soon to come Visual Studio based
acqusition paths, we should no longer need the .pkg and .msi installers
for our .NET 6 content.  Signing and notarization for these installers
takes a decent amount of time in our official build pipeline.  We can
save some turnaround time by removing these jobs as they should no
longer be needed.  The source used to create these installers is not
being removed in case it's ever needed in the future.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1250,11 +1250,11 @@ stages:
 
     - template: yaml-templates/fail-on-issue.yaml
 
-- stage: dotnet_installers
-  displayName: .NET 6 Preview Installers
+- stage: dotnet_prepare_release
+  displayName: .NET 6 Prepare Release
   dependsOn: mac_build
   jobs:
-  # Check - "Xamarin.Android (.NET 6 Preview Installers Sign NuGets)"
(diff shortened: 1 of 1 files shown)


pjcollins:
Test run (sets both PushXAPackage vars to false): https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5026710&view=results

jonathanpeppers (commented):
Can we delete these, too?

* https://github.com/xamarin/xamarin-android/tree/main/build-tools/create-dotnet-msi
* https://github.com/xamarin/xamarin-android/tree/main/build-tools/create-dotnet-pkg
* https://github.com/xamarin/xamarin-android/blob/main/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/GenerateWixFile.cs

pjcollins:
> Can we delete these, too?

I figured it made sense to keep those sources around a little longer if ever needed them for whatever reason, maybe we could delete after GA in November? I am happy to delete them in this PR as well if we want though.

jonathanpeppers (approved):
Maybe we should just merge this as-is, because it will speed up the time it takes to produce builds.

PR #6135, opened 2021-07-28
Bump LibZipSharp, Mono.Unix and xamarin-android-tools.

dellis1972:


grendello (approved):

jonathanpeppers:
I guess we need to bump monodroid first?
```
(_ResolveXamarinAndroidTools target) -> 
  C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\Xamarin.Android.Common.Debugging.targets(109,2): error MSB4062: The "Xamarin.Android.Tasks.ResolveXamarinAndroidTools" task could not be loaded from the assembly C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\MSBuild\Xamarin\Android\Xamarin.Android.Build.Debugging.Tasks.dll. Could not load file or assembly 'Microsoft.Android.Build.BaseTasks, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. 

    1 Warning(s)
    1 Error(s)
```

dellis1972:
I'm including a test monodroid bump to see if this passes without strong naming. Since we need to strong name java.interop etc.

jonathanpeppers:
@dellis1972 I think it's working now.

So we'll have to merge monodroid (it will be broken temporarily), and then come back to this?

dellis1972:
> @dellis1972 I think it's working now.
> 
> So we'll have to merge monodroid (it will be broken temporarily), and then come back to this?

Correct.

PR #6139, opened 2021-07-28
[One .NET] rework $(ApplicationVersion) and $(ApplicationDisplayVersion)

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/1662

We previously were mapping new MSBuild properties to values in
`AndroidManifest.xml`:

* `$(AndroidVersionCode)` -> `android:versionCode`
* `$(ApplicationVersion)` -> `android:versionName`

To be more consistent with other platforms, we're reworking this to be:

* `$(ApplicationVersion)` -> `android:versionCode`
* `$(ApplicationDisplayVersion)` -> `android:versionName`

If we force `$(ApplicationVersion)` to be an integer, you can use the
same values on all platforms dotnet/maui supports. We no longer need
an Android-specific `$(AndroidVersionCode)`.

We also need to consider the `$(Version)` property that influences
`$(AssemblyVersion)`, `$(FileVersion)`, and `$(InformationalVersion)`.
We can simply set `$(Version)` when `$(ApplicationDisplayVersion)` is
set to influence *all* the version values.

We also needed to use @jonpryor's clever trick to detect when
`$(ApplicationDisplayVersion)` is not a valid `System.Version`:

    <Version Condition=" $([System.Version]::TryParse ('$(ApplicationDisplayVersion)', $([System.Version]::Parse('1.0')))) ">$(ApplicationDisplayVersion)</Version>

I updated tests and documentation around this scenario. I also updated
the `dotnet new android` project template.

There will need to be future changes in xamarin/xamarin-macios for
Apple platforms and dotnet/maui for WinUI.

### DIFF
--- a/Documentation/guides/OneDotNetSingleProject.md
+++ b/Documentation/guides/OneDotNetSingleProject.md
@@ -19,25 +19,27 @@ Xamarin.Android and Xamarin.iOS/Mac SDKs:
 
 * `$(ApplicationId)` maps to `/manifest/@package` and
   `CFBundleIdentifier`
-* `$(ApplicationVersion)` maps to `android:versionName` or
-  `CFBundleVersion`. This is a version string that must be incremented
-  for each iOS App Store or TestFlight submission.
-* `$(AndroidVersionCode)` maps to `android:versionCode` (_Android
-  only)_. This is unfortunately an integer and must be incremented for
-  each Google Play submission.
-* `$(AppleShortVersion)` maps to `CFBundleShortVersionString` (_iOS
-  only)_. This can default to `$(ApplicationVersion)` when blank.
+* `$(ApplicationVersion)` maps to `android:versionCode` or
+  [`CFBundleVersion`][CFBundleVersion]. This is required to be an integer on Android and
+  less than 10000 on iOS. This value must be incremented for each
+  Google Play or App Store / TestFlight submission.
+* `$(ApplicationDisplayVersion)` maps to `android:versionName` or
+  [`CFBundleShortVersionString`][CFBundleShortVersionString]. This can
+  default to `$(ApplicationVersion)` when blank.
 * `$(ApplicationTitle)` maps to `/application/@android:title` or
   `CFBundleDisplayName`
+
+[CFBundleVersion]: https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364
+[CFBundleShortVersionString]: https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-111349
 
 The final value that is generated in the `Info.plist` or
 `AndroidManifest.xml` can be overridden at different times. The final
 source of truth is determined in order of:
 
-1. `Info.plist` or `AndroidManifest.xml` in the iOS/Android head project.
-2. iOS/Android head `.csproj` defines the MSBuild properties
-3. _(To be implemented in MAUI/Forms)_ set in a shared `.csproj`.
-4. The properties set by MSBuild via other means such as
+1. `Info.plist` or `AndroidManifest.xml` in the iOS/Android project.
+2. iOS/Android `.csproj` defines the MSBuild properties. This could
+   also be done in a .NET MAUI "Single Project".
+3. The properties set by MSBuild via other means such as
    `Directory.Build.props`, etc.
 
 Even if we did not complete the goal of complete removal of
@@ -57,7 +59,7 @@ property to disable the behavior.
 In most cases, developers would only use `$(GenerateApplicationManifest)`
 if they want to try the new features in "legacy" Xamarin.
 
-## AssemblyVersion and FileVersion
+## Version, AssemblyVersion, FileVersion, and InformationalVersion
 
 Since we are adding *more* version properties, we should consider
 adding defaults to consolidate the assembly-level attributes when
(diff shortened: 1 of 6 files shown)


jonathanpeppers:
This needs support for a `$(Version)` property, see: https://github.com/dotnet/maui/issues/1662#issuecomment-890016404

This is already built into the dotnet/sdk, and we just missed that this exists.

dellis1972:
This looks good so far. I like how we still raise `XA0003` if the value is not an `int`.

Documentation/guides/OneDotNetSingleProject.md
```diff
@@ -67,16 +69,22 @@ The full list of defaults might be something like:
...
 <PropertyGroup>
-  <ApplicationVersion Condition=" '$(ApplicationVersion)' == '' ">1.0</ApplicationVersion>
-  <!-- Android only -->
-  <AndroidVersionCode Condition=" '$(AndroidVersionCode)' == '' ">1</AndroidVersionCode>
-  <!-- Apple platforms only -->
-  <AppleShortVersion Condition=" '$(AppleShortVersion)' == '' ">$(ApplicationVersion)</ApplicationVersion>
-  <AssemblyVersion Condition=" '$(AssemblyVersion)' == '' ">$(ApplicationVersion)</AssemblyVersion>
-  <FileVersion Condition=" '$(FileVersion)' == '' ">$(ApplicationVersion)</FileVersion>
+  <ApplicationVersion Condition=" '$(ApplicationVersion)' == '' ">1</ApplicationVersion>
+  <Version Condition=" '$(ApplicationDisplayVersion)' != '' ">$(ApplicationDisplayVersion)</Version>
```
dellis1972, line 73: Should we check `Version` is empty before overwriting it? Users might be setting it via some build process like GitInfo.

jonathanpeppers: It's always set to `1.0.0` by the dotnet/sdk, so we don't know if users set it. We could maybe check if it equals `1.0.0`?

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/SingleProjectTest.cs
```diff
@@ -11,13 +12,25 @@ namespace Xamarin.Android.Build.Tests
...
+			new object [] {
+				/* versionName */ "2.1",
+				/* versionCode */ "42",
+				/* errorCode */   "",
+			},
+			new object [] {
+				/* versionName */ "1.0.0",
+				/* versionCode */ "1.0.0",
+				/* errorCode */   "XA0003",
+			},
```
jonathanpeppers, line 25: I need to add two test cases here:
* `3.1.3a1` - iOS number
* `6.0.100-preview.7` - semver

Just to see what happens, and if we need to do anything for these.

The iOS number comes from:
> While developing a new version of your app, you can include a suffix after the number that is being updated; for example 3.1.3a1. The character in the suffix represents the stage of development for the new version. For example, you can represent development, alpha, beta, and final candidate, by d, a, b, and fc. The final number in the suffix is the build version, which cannot be 0 and cannot exceed 255. When you release the new version of your app, remove the suffix.

jonathanpeppers: This one fails pretty badly...
```
~\android-toolchain\dotnet\sdk\6.0.100-rc.2.21456.2\NuGet.targets(130,5): '3.1.3a1' is not a valid version string. (Parameter 'value')
```

jonathanpeppers: I can't find anything sensible to do here, `$([System.Version]::Parse())` or `$([MSBuild]:VersionGreaterThanOrEquals())` throw during evaluation.

I think adding the test case is fine for now, and you would get the above error from `NuGet.targets`?

dellis1972: Hmm 🤷 I'm not sure what the best option is. Especially since the error is in Nuget. Could we have a task that runs before `Restore` that checks and gives a more actionable "android"/"iOS" error message?

jonathanpeppers: Well, the same issue happens if you put `<Version>3.1.3a1</Version>` in a plain .NET class library.

So the error message needs to be improved in NuGet? So it helps all project types?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/SingleProjectTest.cs
```diff
@@ -11,13 +12,37 @@ namespace Xamarin.Android.Build.Tests
...
+				/* versionCode */  "1.0.0",
+				/* errorMessage */ "XA0003",
+			},
+			new object [] {
+				/* versionName */  "3.1.3a1",
+				/* versionCode */  "42",
+				// NOTE: NuGet.targets has no error code
+				// NuGet.targets(130,5): '3.1.3a1' is not a valid version string. (Parameter 'value')
+				/* errorMessage */ "not a valid version string",
+			},
```
jonathanpeppers, line 32: This test case fails on legacy, need to fix this.

jonpryor:
Through a bit of "inspired *insanity*", behold! https://gist.github.com/jonpryor/5d4a58216f8e7e7e49f52651d9bb018b

```xml
  <PropertyGroup>
    <InvalidVersion>1.2.3a4</InvalidVersion>
    <IsValid Condition="$([System.Version]::TryParse(
      '$(InvalidVersion)',
      $([System.Version]::Parse('1.0'))
    ))">True</IsValid>
  </PropertyGroup>
```

That works, and doesn't throw an exception or otherwise generate an error, which allows us to do *some* degree of parameter validation/sanity checking if/when `$(ApplicationDisplayVersion)` contains an invalid Version value.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/SingleProjectTest.cs
```diff
@@ -11,13 +13,35 @@ namespace Xamarin.Android.Build.Tests
...
+			new object [] {
+				/* versionName */  "1.0.0",
+				/* versionCode */  "1.0.0",
+				/* errorMessage */ "XA0003",
+			},
+			new object [] {
+				/* versionName */  "3.1.3a1",
+				/* versionCode */  "42",
+				/* errorMessage */ "",
+			},
```
jonathanpeppers, line 31: @jonpryor I used your trick and this test case passes now.

PR #6155, opened 2021-07-30
[One .NET] add missing API documentation

jonathanpeppers:
Right now in .NET 6, you don't get any `<summary>` documentation on
Android APIs.

This adds files like `Mono.Android.xml` and `Java.Interop.xml` to the
Microsoft.Android.Ref pack to solve this issue.

I also fixed a place in `Mono.Android.csproj` that used `net6.0`, so
this will continue to work in a future .NET 7.

### DIFF
--- a/src/Mono.Android/Mono.Android.csproj
+++ b/src/Mono.Android/Mono.Android.csproj
@@ -25,7 +25,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <IncludeAndroidJavadoc Condition=" '$(IncludeAndroidJavadoc)' == '' And '$(CONVERT_JAVADOC_TO_XMLDOC)' == 'true' And '$(AndroidFrameworkVersion)' == '$(AndroidLatestStableFrameworkVersion)'">True</IncludeAndroidJavadoc>
+    <IncludeAndroidJavadoc Condition=" '$(IncludeAndroidJavadoc)' == '' And '$(CONVERT_JAVADOC_TO_XMLDOC)' == 'true' And ('$(AndroidFrameworkVersion)' == '$(AndroidLatestStableFrameworkVersion)' or '$(TargetFramework)' != 'monoandroid10')">True</IncludeAndroidJavadoc>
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
Test run where I forced `Mono.Android.xml` to be produced:

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5040789&view=artifacts&pathAsName=false&type=publishedArtifacts

image

Now I'll remove ca49ba6, and we can merge this.

src/Mono.Android/Mono.Android.csproj
```diff
@@ -25,7 +25,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <IncludeAndroidJavadoc Condition=" '$(IncludeAndroidJavadoc)' == '' And '$(CONVERT_JAVADOC_TO_XMLDOC)' == 'true' And '$(AndroidFrameworkVersion)' == '$(AndroidLatestStableFrameworkVersion)'">True</IncludeAndroidJavadoc>
+    <IncludeAndroidJavadoc Condition=" '$(IncludeAndroidJavadoc)' == '' And '$(CONVERT_JAVADOC_TO_XMLDOC)' == 'true' And ('$(AndroidFrameworkVersion)' == '$(AndroidLatestStableFrameworkVersion)' or '$(TargetFramework)' != 'monoandroid10')">True</IncludeAndroidJavadoc>
```
pjcollins, line 28: I'm not sure if the `!= monoandroid10` condition is needed here? We should only be building the net6.0 version of Mono.Android.dll when `$(AndroidFrameworkVersion)' == '$(AndroidLatestStableFrameworkVersion)'`. Probably doesn't hurt to have this either way though, it may be useful in the future if/when we produce a Ref pack for a preview API level?

jonathanpeppers: I was probably seeing that `$(CONVERT_JAVADOC_TO_XMLDOC)` was blank, and we don't need it *now*.

But since we would need this when we have an API 32 "preview" for `net6.0-android32`, maybe we should leave it?

pjcollins (approved):

PR #6157, opened 2021-07-30
[One .NET] support $(EnableLLVM) with AOT

jonathanpeppers:
Fixed some issues that allows `$(EnableLLVM)` to work:

* `$(_LLVMPath)` needs to point to Mono's AOT pack that contains `opt`
  and `llc`.
* We need to fill out `ld-name` and `ld-flags`.
* `ld-flags` needs some special workarounds in order for them to be
  passed in correctly on .NET 6.
    * Use space as a delimiter
    * Escape spaces in paths
* Added `temp-path` parameter. This prevents the warning on Windows:

```
'rm' is not recognized as an internal or external command
```

I enabled tests around .NET 6 and LLVM.

## Results ##

All tests:

 1. Were running on a [Google Pixel 5][0], and
 2. Enabled two architectures, arm64 and x86, and
 3. **AOT time** was average of 10 runs with `-c Release
    -p:RunAOTCompilation=true`, with the `Activity: Displayed` time
 4. **AOT+LLVM time** was average of 10 runs with `-c Release
    -p:RunAOTCompilation=true -p:EnableLLVM=true` with the
    `Activity: Displayed` time.

| Test                |      AOT time | AOT+LLVM time |  AOT apk size |  AOT+LLVM apk size |
| ------------------- | ------------: | ------------: | ------------: | -----------------: |
| [HelloAndroid][1]   |  00:00:00.238 |  00:00:00.226 |    12,073,931 |         12,602,315 |
| [HelloMaui][2]      |  00:00:00.624 |  00:00:00.596 |    43,167,801 |         47,546,425 |

[0]: https://store.google.com/us/product/pixel_5_specs?hl=en-US
[1]: https://github.com/dotnet/maui-samples/tree/714460431541f40570e91225e8ba4bc1fe08025f/HelloAndroid
[2]: https://github.com/dotnet/maui-samples/tree/714460431541f40570e91225e8ba4bc1fe08025f/HelloMaui

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Aot.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Aot.cs
@@ -2,15 +2,12 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
-using System.Linq;
 using System.Text;
 using System.Threading;
 using Microsoft.Build.Framework;
 using Microsoft.Build.Utilities;
 
-using Java.Interop.Tools.Diagnostics;
 using Xamarin.Android.Tools;
-using Xamarin.Build;
 using Microsoft.Android.Build.Tasks;
 
 namespace Xamarin.Android.Tasks
@@ -53,44 +50,6 @@ namespace Xamarin.Android.Tasks
 
 		[Output]
 		public string[] NativeLibrariesReferences { get; set; }
-
-		static string GetNdkToolchainLibraryDir (NdkTools ndk, string binDir, string archDir = null)
-		{
-			var baseDir = Path.GetFullPath(Path.Combine(binDir, ".."));
-
-			string libDir = Path.Combine (baseDir, "lib", "gcc");
-			if (!String.IsNullOrEmpty (archDir))
-				libDir = Path.Combine (libDir, archDir);
-
-			var gccLibDir = Directory.EnumerateDirectories (libDir).ToList();
-			gccLibDir.Sort();
-
-			var libPath = gccLibDir.LastOrDefault();
-			if (libPath == null) {
-				goto no_toolchain_error;
-			}
-
-			if (ndk.UsesClang)
-				return libPath;
-
-			gccLibDir = Directory.EnumerateDirectories(libPath).ToList();
-			gccLibDir.Sort();
-
-			libPath = gccLibDir.LastOrDefault();
-			if (libPath == null) {
-				goto no_toolchain_error;
-			}
-
-			return libPath;
-
-		  no_toolchain_error:
-			throw new Exception("Could not find a valid NDK compiler toolchain library path");
-		}
-
-		static string GetNdkToolchainLibraryDir (NdkTools ndk, string binDir, AndroidTargetArch arch)
-		{
-			return GetNdkToolchainLibraryDir (ndk, binDir, ndk.GetArchDirName (arch));
-		}
 
 		static string QuoteFileName(string fileName)
 		{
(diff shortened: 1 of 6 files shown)


jonathanpeppers:
This is failing to load the native libraries at runtime currently:
```
08-02 10:04:57.751 25521 25521 I monodroid-assembly: Trying to load shared library '/data/app/~~LyPhmF5JQkyho706qFJDWQ==/com.microsoft.net6.helloandroid-6Khj2PEudy92T_xZKNUHkw==/lib/arm64/libaot-Mono.Android.dll.so'
08-02 10:04:57.752 25521 25521 I monodroid-assembly: Failed to load shared library '/data/app/~~LyPhmF5JQkyho706qFJDWQ==/com.microsoft.net6.helloandroid-6Khj2PEudy92T_xZKNUHkw==/lib/arm64/libaot-Mono.Android.dll.so'. dlopen failed: cannot locate symbol "memset" referenced by "/data/app/~~LyPhmF5JQkyho706qFJDWQ==/com.microsoft.net6.helloandroid-6Khj2PEudy92T_xZKNUHkw==/lib/arm64/libaot-Mono.Android.dll.so"...
08-02 10:04:57.752 25521 25521 I monodroid-assembly: open_from_bundles: loaded assembly: 0xb400007bcc1744f0
```
And so it's just falling back to the JIT.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -150,8 +158,6 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
 		[Category ("DotNetIgnore")] // Not currently working, see: https://github.com/dotnet/runtime/issues/56163
```
dellis1972, line 158: Do we need to remove these now?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -230,8 +236,6 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
...
```
dellis1972, line 236: same here

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
jonathanpeppers: The tests that mention https://github.com/dotnet/runtime/issues/56163 don't work yet. I think we need changes in dotnet/runtime to fix those.

Folder names like `BuildAotApplicationAndBundle AndÜmläüts_arm64-v8a_False_True` fail, and I saw the `<MonoAOTCompiler/>` task output characters that looked like the encoding was messed up.

jonathanpeppers (commented):
Something in the latest changes broke legacy:
```
[aot-compiler stderr] '"opt"' is not recognized as an internal or external command, (TaskId:180)
```
I thought it was green, but then I changed some other things...

src/Xamarin.Android.Build.Tasks/Tasks/Aot.cs
```diff
@@ -232,20 +138,13 @@ IEnumerable<Config> GetAotConfigs (NdkTools ndk)
...
-						aotOptions.Add ($"ld-name={ldName}");
-					}
+					Directory.CreateDirectory (tempDir);
 
-					aotOptions.Add ($"ld-flags={ldFlags}");
+					var aotOptions = GetAotOptions (ndk, arch, level, outdir, mtriple, toolPrefix);
+					// NOTE: ordering seems to matter on Windows
+					aotOptions.Insert (0, $"outfile={outputFile}");
+					aotOptions.Insert (0, $"llvm-path={SdkBinDirectory}");
+					aotOptions.Insert (0, $"temp-path={tempDir}");
```
jonathanpeppers, line 147: The PR build is green (just seems like Github status didn't update).

The most recent thing that broke was the ordering of these parameters. It seems like the AOT compiler wants `llvm-path` before `ld-name` or `ld-flags`.

dellis1972:
@jonathanpeppers I would feel more comfortable if we ran the entire test suite on this (if we are not already) rather than relying on the targeted tests which we normally do. Messing with aot path quoting etc is tricky and can break in unexpected places.

jonathanpeppers:
@dellis1972 the test run should have every step, nothing skipped:

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5055201&view=results

If a file changes in `build-tools`, it runs everything.

dellis1972 (approved):

Issue #6108, opened 2021-07-21
aar app export?


jonpryor:
Context: https://developercommunity.visualstudio.com/t/C-library-integration-in-android/1445867

A customer would like to be able to "include" the contents of a Xamarin.Android app into an Android app, kinda/sorta like what https://github.com/mono/embeddinator-4000 was trying to support.

Supporting "`.aar` export" at the level of a *library* would be complicated.

Supporting "`.aar` export" at the level of an *app* should be "reasonably" straightforward (famous last words), in that it should be *possible* to emit a `.aar` that contains all (*linked*!) assemblies and native libraries and resources (oh my!), along with `AndroidManifest.xml`.

This would *not* permit creating a `.aar` and distributing it on e.g. Maven (there'd be duplicate `mscorlib.dll` files, duplicate `libmonosgen-2.0.so` files, etc., etc.).  It *would* allow *easier* use of C# in more contexts than is currently supported.

AlexSchuetz:
I have some some business-logic written in C# (netstandard2.0) that is used in various szenarios (xamarin-forms, unity, avalonia and ASP-NETcore).
The main goal was to make this logic available to as much platforms as possible without duplication written in a high-level OO-language.

Next and final step would be to make the library available for native android (and ios) applications as well.

I'd like to work on this issue, but I have no clue where to start. Any hints?

llzz19851985:
Is there a development plan for this feature?I hava a project need this feature too!

soma9017:
@llzz19851985 anyway to use the c# dll to use in native android application developments..? if yes please explain how to achieve this as i am also looking for something like this

PR #6159, opened 2021-08-02
Use more standard C++ features + performance improvements

grendello:
C++ features:

 * Use C++20 concepts to make compile-time correctness verification
   stronger
 * Use `std::vector` for dynamic heap buffers
 * Use `std::array` for static stack buffers
 * Use `std::unique_ptr` instead of the home-grown equivalent (also for
   C-allocated memory)
 * Use C++20 structure designators

Optimizations:

 * Add string routines (`ends_with` and `find_last`) optimized for the
   stack string types (derived from `string_base`)
 * Count assemblies at the build time and pre-allocate enough memory (in
   the form of `std::vector` instance) to store information about the
   bundled assemblies, thus avoiding 2 allocations per assembly (one
   `realloc` and one `malloc`) making the bundled assemblies loading
   algorithm closer to O(1) instead of the previous O(n*2)
 * Use character arrays instead of "plain" string literals to optimize
   some string operations.

These changes improve plain XA test app startup time by around 11%

### DIFF
--- a/src/monodroid/CMakeLists.txt
+++ b/src/monodroid/CMakeLists.txt
@@ -492,6 +492,7 @@ if(ANDROID)
       ${BIONIC_SOURCES_DIR}/cxa_guard.cc
       ${SOURCES_DIR}/cxx-abi/string.cc
       ${SOURCES_DIR}/cxx-abi/terminate.cc
+      ${SOURCES_DIR}/cxx-abi/vector.cc
       )
   endif()
 else()
--- a/src/monodroid/jni/embedded-assemblies-zip.cc
+++ b/src/monodroid/jni/embedded-assemblies-zip.cc
@@ -1,5 +1,7 @@
+#include <array>
 #include <cerrno>
 #include <cctype>
+#include <vector>
 #include <libgen.h>
 
 #include <mono/metadata/assembly.h>
@@ -9,6 +11,17 @@
 #include "globals.hh"
 
 using namespace xamarin::android::internal;
+
+// This type is needed when calling read(2) in a MinGW build, as it defines the `count` parameter as `unsigned int`
+// instead of `size_t` which then causes the following warning if we pass a value of type `size_t`:
+//
+//   warning: conversion from ‘size_t’ {aka ‘long long unsigned int’} to ‘unsigned int’ may change value [-Wconversion]
+//
+#if defined (WINDOWS)
+using read_count_type = unsigned int;
+#else
+using read_count_type = size_t;
+#endif
 
 void
 EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_should_register should_register)
@@ -32,19 +45,16 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
 		exit (FATAL_EXIT_NO_ASSEMBLIES);
 	}
 
-	// C++17 allows template parameter type inference, but alas, Apple's antiquated compiler does
-	// not support this particular part of the spec...
-	simple_pointer_guard<uint8_t[]>  buf (new uint8_t[cd_size]);
+	std::vector<uint8_t>  buf (cd_size);
 	const char           *prefix     = get_assemblies_prefix ();
-	size_t                prefix_len = strlen (prefix);
+	size_t                prefix_len = get_assemblies_prefix_length ();
 	size_t                buf_offset = 0;
 	uint16_t              compression_method;
 	uint32_t              local_header_offset;
 	uint32_t              data_offset;
 	uint32_t              file_size;
-	char                 *file_name;
 
-	ssize_t nread = read (fd, buf.get (), cd_size);
+	ssize_t nread = read (fd, buf.data (), static_cast<read_count_type>(buf.size ()));
 	if (static_cast<size_t>(nread) != cd_size) {
 		log_fatal (LOG_ASSEMBLY, "Failed to read Central Directory from the APK archive %s. %s (nread: %d; errno: %d)", apk_name, std::strerror (errno), nread, errno);
 		exit (FATAL_EXIT_NO_ASSEMBLIES);
@@ -54,6 +64,19 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
 #if defined (NET6)
 	bool runtime_config_blob_found = false;
 #endif // def NET6
+
+	bool bundled_assemblies_slow_path = false;
+	if (bundled_assemblies.empty ()) {
+		// We're called for the first time, let's allocate space for all the assemblies counted during build. resize()
+		// will allocate all the memory and zero it in one operation
+		bundled_assemblies.resize (application_config.number_of_assemblies_in_apk);
+	} else {
+		// We are probably registering some assemblies dynamically, possibly loading from another apk, we need to extend
+		// the storage instead of taking advantage of the pre-allocated slots.
+		bundled_assemblies_slow_path = true;
+	}
+
+	int64_t assembly_count = application_config.number_of_assemblies_in_apk;
 
 	// clang-tidy claims we have a leak in the loop:
 	//
@@ -66,13 +89,12 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
 	for (size_t i = 0; i < cd_entries; i++) {
 		entry_name.clear ();
 
-		bool result = zip_read_entry_info (buf.get (), cd_size, buf_offset, compression_method, local_header_offset, file_size, entry_name);
-		file_name = entry_name.get ();
+		bool result = zip_read_entry_info (buf, buf_offset, compression_method, local_header_offset, file_size, entry_name);
 
 #ifdef DEBUG
-		log_warn (LOG_ASSEMBLY, "%s entry: %s", apk_name, file_name == nullptr ? "unknown" : file_name);
+		log_warn (LOG_ASSEMBLY, "%s entry: %s", apk_name, entry_name.get () == nullptr ? "unknown" : entry_name.get ());
 #endif
-		if (!result || file_name == nullptr) {
+		if (!result || entry_name.empty ()) {
 			log_fatal (LOG_ASSEMBLY, "Failed to read Central Directory info for entry %u in APK file %s", i, apk_name);
 			exit (FATAL_EXIT_NO_ASSEMBLIES);
 		}
@@ -87,14 +109,14 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
 		if (compression_method != 0)
 			continue;
 
-		if (strncmp (prefix, file_name, prefix_len) != 0)
+		if (strncmp (prefix, entry_name.get (), prefix_len) != 0)
 			continue;
 
 #if defined (NET6)
 		if (application_config.have_runtime_config_blob && !runtime_config_blob_found) {
-			if (utils.ends_with (file_name, SharedConstants::RUNTIME_CONFIG_BLOB_NAME)) {
+			if (utils.ends_with (entry_name, SharedConstants::RUNTIME_CONFIG_BLOB_NAME)) {
 				runtime_config_blob_found = true;
-				runtime_config_blob_mmap = md_mmap_apk_file (fd, data_offset, file_size, file_name, apk_name);
+				runtime_config_blob_mmap = md_mmap_apk_file (fd, data_offset, file_size, entry_name.get (), apk_name);
 				continue;
 			}
 		}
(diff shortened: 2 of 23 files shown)


grendello:
Looks like Xcode we have on bots it's too old :(

jonathanpeppers:
@grendello we could update the version numbers in this script:

https://github.com/xamarin/xamarin-android/blob/main/build-tools/provisioning/xcode.csx

The problem is that Xcode 12.5 requires Big Sur. So it depends on if we have a pool we can use for that?

grendello:
> @grendello we could update the version numbers in this script:
> 
> https://github.com/xamarin/xamarin-android/blob/main/build-tools/provisioning/xcode.csx
> 
> The problem is that Xcode 12.5 requires Big Sur. So it depends on if we have a pool we can use for that?

Hm, I see. I can work around it and add a compiler feature check then. It's not a crucial part of the code.

pjcollins:
We are using three Android specific pools for our macOS builds that do not yet have any Big Sur machines in them. We should start looking at upgrading some machines in those pools to Big Sur, but if you can work around this requirement for now as mentioned that's probably best.

grendello:
> We are using three Android specific pools for our macOS builds that do not yet have any Big Sur machines in them. We should start looking at upgrading some machines in those pools to Big Sur, but if you can work around this requirement for now as mentioned that's probably best.

Just pushed a change that should make it work with the older Xcode, thanks :)

src/monodroid/CMakeLists.txt
```diff
@@ -492,6 +492,7 @@ if(ANDROID)
       ${BIONIC_SOURCES_DIR}/cxa_guard.cc
       ${SOURCES_DIR}/cxx-abi/string.cc
       ${SOURCES_DIR}/cxx-abi/terminate.cc
+      ${SOURCES_DIR}/cxx-abi/vector.cc
```
jonpryor, line 495: On a lark, I wanted to re-examine which public symbols we're exporting:

```
% $HOME/android-toolchain/ndk//toolchains/llvm/prebuilt/darwin-x86_64/bin/x86_64-linux-android-nm -D -C bin/Debug/lib/xamarin.android/xbuild/Xamarin/Android/lib/x86_64/libmono-android.debug.so
00000000000216b0 T JNI_OnLoad
0000000000021730 T Java_mono_android_Runtime_init
00000000000236e0 T Java_mono_android_Runtime_initInternal
0000000000029410 T Java_mono_android_Runtime_notifyTimeZoneChanged
0000000000023d70 T Java_mono_android_Runtime_propagateUncaughtException
0000000000023730 T Java_mono_android_Runtime_register
```

Good and understandable.

```
000000000003e480 T std::terminate()
0000000000014f80 T operator delete[](void*)
0000000000014f90 T operator delete[](void*, std::nothrow_t const&)
0000000000014fa0 T operator delete[](void*, unsigned long)
0000000000014f50 T operator delete(void*)
0000000000014f60 T operator delete(void*, std::nothrow_t const&)
0000000000014f70 T operator delete(void*, unsigned long)
0000000000014ef0 T operator new[](unsigned long)
0000000000014f30 T operator new[](unsigned long, std::nothrow_t const&)
0000000000014e90 T operator new(unsigned long)
0000000000014ed0 T operator new(unsigned long, std::nothrow_t const&)
```

Uh... is this still sensible or desirable?  `libmonodroid.so` is loaded by *Android*, so *we* don't control if it's loaded with `RTLD_GLOBAL` or `RTLD_LOCAL` set, and if `RTLD_GLOBAL` *is* used, I don't know what happens if a *real* C++ lib is also loaded in the process.

Is there a way to make these *weak* exports?

grendello: They shouldn't be visible, and they weren't previously. Apparently the switch to the newer NDK with the newer libc++ changed the visibility at some point. I am able to hide most of them instead of:

```
0000000000011d04 T operator delete[](void*)
0000000000011ce0 T operator delete(void*)
0000000000011c9c T operator new[](unsigned long)
0000000000011c58 T operator new(unsigned long)
```

They are declared by the compiler in each compilation unit, without having to include `<new>`, I'll check tomorrow if there's a way to change their visibility.

jonpryor: Makes me wonder if we need a "test" to verify which symbols are exported, so that we don't accidentally export symbols we don't want exported.

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -87,115 +109,119 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
...
-		if ((utils.ends_with (file_name, ".pdb") || utils.ends_with (file_name, ".mdb")) &&
+		if ((utils.ends_with (entry_name, ".pdb") || utils.ends_with (entry_name, ".mdb")) &&
 				register_debug_symbols &&
 				!entry_is_overridden &&
-				bundled_assemblies != nullptr) {
-			md_mmap_info map_info = md_mmap_apk_file(fd, data_offset, file_size, file_name, apk_name);
-			if (register_debug_symbols_for_assembly (file_name, (bundled_assemblies) [bundled_assemblies_count - 1], (const mono_byte*)map_info.area, static_cast<int>(file_size)))
+				bundled_assembly_index >= 1) {
+			md_mmap_info map_info = md_mmap_apk_file(fd, data_offset, file_size, entry_name.get (), apk_name);
+			if (register_debug_symbols_for_assembly (entry_name, (bundled_assemblies) [bundled_assembly_index - 1], (const mono_byte*)map_info.area, static_cast<int>(file_size)))
```
jonpryor, line 140: Why parens around `(bundled_assemblies)`?  (Yes, it *previously* had them, but it shouldn't need them *now*…)

I also wonder if this is a potential future "bug," *requiring* that the `.dll` come *before* the `.pdb`.  If we ever support "`.aar` export" (#6108), *we* will no longer be responsible for creating the `.apk`/`.aab` in all circumstances.  Should we use a more resilient system?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -87,115 +109,119 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
...
 				continue;
 		}
 
 #if !defined(NET6)
-		if (utils.ends_with (file_name, ".config") && bundled_assemblies != nullptr) {
-			char *assembly_name = strdup (basename (file_name));
+		if (utils.ends_with (entry_name, ".config") && !bundled_assemblies.empty ()) {
+			char *assembly_name = strdup (basename (entry_name.get ()));
```
jonpryor, line 146: I look at this, wonder "why `char*` and not `std::string`", look at the source to `mono_register_config_for_assembly()`, and see that mono doesn't duplicate the string, but just uses it as-is.  Meaning we *cannot* free this value, and thus we shouldn't use `std::string`.

Which makes me wonder if there's a more "type declarative" approach here, to say "we explicitly relinquish ownership of this value to mono"?

…and I think there is/could be: `std::unique_ptr::release`:

```cpp
c_unique_ptr<char[]> assembly_name {strdup (basename (entry_name.get ()))};
*strrchr (assembly_name.get (), '.') = '\0';
// …
mono_register_config_for_assembly (assembly_name.release (), (const char *) map_info.area);
```

"Pro": this is exception safe.  (Not that we expect or should otherwise be using exceptions, as who knows what all that'll break.  So possibly a con?)

"Con": perhaps more overhead because of the use of `c_unique_ptr`, though I'm not sure how much this would actually impact anything.

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -87,115 +109,119 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
...
 			continue;
 
 		if (entry_is_overridden)
 			continue;
 
-		size_t alloc_size = MULTIPLY_WITH_OVERFLOW_CHECK (size_t, sizeof(void*), bundled_assemblies_count + 1);
-		bundled_assemblies = reinterpret_cast<MonoBundledAssembly**> (utils.xrealloc (bundled_assemblies, alloc_size));
-		MonoBundledAssembly *cur = bundled_assemblies [bundled_assemblies_count] = reinterpret_cast<MonoBundledAssembly*> (utils.xcalloc (1, sizeof (MonoBundledAssembly)));
-		++bundled_assemblies_count;
+		assembly_count--;
```
jonpryor, line 163: I look at this and wonder…. This statement is "weird" even in-context; why is it being decremented?  It starts as `application_config.number_of_assemblies_in_apk`, which could be 0 for a "debug+fastdev" build…

but why decrement *first*?  Why not instead change the `if` below to:

```cpp
if (XA_UNLIKELY (assembly_count == 0 || …)
```

I don't understand why this decrement is needed in the first place.  Why this particular pattern?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -87,115 +109,119 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
...
 		unsigned int *psize = (unsigned int*) &cur->size;
 		*psize = static_cast<unsigned int>(file_size);
 
-		if (utils.should_log (LOG_ASSEMBLY)) {
+		if (XA_UNLIKELY (utils.should_log (LOG_ASSEMBLY))) {
 			const char *p = (const char*) cur->data;
 
-			char header[9];
-			for (size_t j = 0; j < sizeof(header)-1; ++j)
+			std::array<char, 9> header;
```
jonpryor, line 187: Sure, more C++20 use, but…really?  Not sure what `std::array` is buying us here.

jonpryor: Likely "obvious" answer: https://coders-corner.net/2018/06/16/stdarray-vs-c-style-array

Good for `<algorithm>` integration, range checking in Debug builds, etc.

Not used *here*, but maybe eventually?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1913,18 +1903,19 @@ MonodroidRuntime::setup_mono_tracing (char const* const& mono_log_mask)
 	constexpr size_t MASK_DLL_LEN = sizeof(MASK_DLL) - 1;
 	constexpr char   MASK_GC[] = "gc";
 	constexpr size_t MASK_GC_LEN = sizeof(MASK_GC) - 1;
+	constexpr char   COMMA[] = ",";
```
jonpryor, line 1906: What's this gain us?  I find the `","` more readable…

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: > Why parens around `(bundled_assemblies)`? (Yes, it _previously_ had them, but it shouldn't need them _now_…)
> 
Omission

> I also wonder if this is a potential future "bug," _requiring_ that the `.dll` come _before_ the `.pdb`. If we ever support "`.aar` export" (#6108), _we_ will no longer be responsible for creating the `.apk`/`.aab` in all circumstances. Should we use a more resilient system?

Yeah, it's something I've always disliked about this code. I'm going to deal with it in another PR as the fix isn't a one-liner.

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: Your observations are valid, and normally I'd agree that we should use both `std::string` and `std::unique_ptr` but they do carry an overhead, and in this code we're trying to gain as much performance as possible, thus `char*`. Regarding exceptions - I'd like to use them for "real" RAII but enabling them without linking `libc++` and `libc++abi` is not a quick task. Exceptions require a bit of infrastructure from both of these libraries, code we'd either have to implement or port from `libc++` sources. I looked at it briefly and it's something that would require a bit of time to implement.

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: Yes, it's that and it's for consistency around the code. Also, `std::array` is a zero-overhead abstraction over C style arrays so we don't sacrifice anything.

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: It's a "low level" marker. And the reason for decrementing is that it's the cheapest approach here. Your proposal of `assembly_count == 0` check means that we'd have to stop decrementing at 0, which means changing line 183 to

```cpp
if (assembly_count > 0)
  assembly_count--;
```

which means a tiny, and unnecessary, performance loss. And it is first decremented because vector indices are 0-based and the counter starts from the vector **size** (`application_config.number_of_assemblies_in_apk`) and decrementing it keeps it in "sync" with `bundled_assembly_index` which counts up from a (valid) index of 0, thus the decrement maintains the symmetry. While `assembly_count` isn't used for indexing, I need it to hit the low level marker (-1) at the same time `bundled_assembly_index` becomes larger than the vector size.

src/monodroid/jni/monodroid-glue.cc
grendello: I prefer `","` as well, but this is again done to gain a tiny bit of performance. Using a constant character array makes the compiler choose the `.append` overload which uses the string size computed at build time, while using a `","` means it treats the literal as a `char*` pointer and we need to use `strlen` in `append` to get its length. Also, C++ guidelines now suggest using either `const char*` or `constexpr char[]` for string literals instead of the "direct" literal, the reason being that the latter is a mutable pointer pointing to read-only memory and any attempt to write to that memory causes undefined behavior. This ambiguity should be avoided whenever possible.

src/monodroid/CMakeLists.txt
grendello: I made the for `delete` and `new` operators weak symbols, I can't figure out how to hide them right now (can't find the built-in declaration of the functions in the compiler)

src/monodroid/jni/monodroid-glue.cc
jonpryor: @grendello [wrote](https://github.com/xamarin/xamarin-android/pull/6159#discussion_r685758397)

> while using a `","` means it treats the literal as a `char*` pointer and we need to use `strlen` in append to get its length.

Wait, what?  For `log_mask.append (",")`, why would C++ prefer `append(const char *s)` over `const char (&s)[Size])`?

So I go about and test it:

```cpp
#include <iostream>

using namespace std;

void foo (const char* s)
{
    cout << "foo(const char*): " << s << endl;
}

template<size_t Size>
void foo (const char (&s)[Size])
{
    cout << "foo(const char (&s)[" << Size << "]): " << s << endl;
}

constexpr char LITERAL[] = "constexpr";

int main ()
{
    foo ("literal");
    foo (LITERAL);
}
```

and… it *always* prefers the `foo(const char*)` overload:

```
% c++ -std=c++20 overload.cc && ./a.out
foo(const char*): literal
foo(const char*): constexpr
```

The only way to cause the `foo (const char (&s)[Size])` overload to be used is if `foo(const char*)` *does not exist*.  When `foo(const char*)` doesn't exist, then *both* `foo("literal")` and `foo(LITERAL)` invocations hit `foo (const char (&s)[Size])`.

From this, I infer that using `COMMA` everywhere won't actually help: it's almost certainly still using the `append(const char*)` overload.

I thus further infer and ask: do we need `append(const char*)` in the first place?  Yes, we do, but perhaps we could rename it so that the template version is preferred?

grendello: So, I know what throws it off. It's the `const char*` in the overload. I modified your sample above so that the first overload has `char*` instead of `const char*` and here's the result:

```shell
$ ./t
foo(const char (&s)[8]): literal
foo(const char (&s)[10]): constexpr
```

That is.. surprising

jonpryor:
```markdown
[monodroid] Use more standard C++ features, performance improvements (#6159)

C++ features:

  * Use C++20 concepts to improve compile-time correctness
    verification
  * Use `std::vector` for dynamic heap buffers
  * Use `std::array` for static stack buffers
  * Use `std::unique_ptr` instead of the home-grown equivalent
    (also for C-allocated memory)
  * Use C++20 structure designators

Optimizations:

  * Add string routines (`ends_with()` and `find_last()`) optimized
    for the stack string types (derived from `string_base`)
  * Count assemblies at build time and pre-allocate enough memory (in
    the form of a `std::vector` instance) to store information about
    the bundled assemblies, thus avoiding 2 allocations per assembly
    (one `realloc()` and one `malloc()`) making the bundled assemblies
    loading algorithm closer to O(1) instead of the previous O(n*2).
  * Use character arrays instead of "plain" string literals to
    optimize some string operations.

These changes improve plain XA test app startup time by around 11%

Additionally, rename `string_segment::append(const char*)` to
`string_segment::append_c(const char*)`.  We [found that][0] the C++
compiler preferred the non-`template` `append(const char*)` overload
over the `template<size_t S> append(const char (&s)[S])` overload,
meaning `append("foo")` would hit the slower `strlen()`-based method.
Likewise create `assign_c(const char*)` and
`starts_with_c(const char*)` member functions.

Finally, [upon review][1] of the exported symbols, we found that many
symbols that we didn't want to export were being exported:

	% $HOME/android-toolchain/ndk/toolchains/llvm/prebuilt/darwin-x86_64/bin/x86_64-linux-android-nm -D -C bin/Debug/lib/xamarin.android/xbuild/Xamarin/Android/lib/x86_64/libmono-android.debug.so
	…
	000000000003e480 T std::terminate()
	0000000000014f80 T operator delete[](void*)
	0000000000014f90 T operator delete[](void*, std::nothrow_t const&)
	0000000000014fa0 T operator delete[](void*, unsigned long)
	0000000000014f50 T operator delete(void*)
	0000000000014f60 T operator delete(void*, std::nothrow_t const&)
	0000000000014f70 T operator delete(void*, unsigned long)
	0000000000014ef0 T operator new[](unsigned long)
	0000000000014f30 T operator new[](unsigned long, std::nothrow_t const&)
	0000000000014e90 T operator new(unsigned long)
	0000000000014ed0 T operator new(unsigned long, std::nothrow_t const&)

Update our declarations of these functions so that they are instead
*weak* symbols, or otherwise no longer exported:

	0000000000011e10 W operator delete[](void*)
	0000000000011dfc W operator delete(void*)
	0000000000011dec W operator new[](unsigned long)
	0000000000011da8 W operator new(unsigned long)

[0]: https://github.com/xamarin/xamarin-android/pull/6159#discussion_r686185106
[1]: https://github.com/xamarin/xamarin-android/pull/6159#discussion_r685533429
```

PR #6167, opened 2021-08-05
Bump Java.Interop to bring in StrongNamed assemblies.

dellis1972:


jonathanpeppers (approved):
The One .NET MSBuild + emulator tests timed out, but I think we could probably merge this without rerunning?

Legacy is going to have more issues with strong naming, because it runs under .NET framework.

PR #6168, opened 2021-08-05
Bump to MSbuild.StructuredLogger 2.1.500

jonathanpeppers:
Some of our `PerformanceTest`'s are currently ignored due to:

    Ignored : Xamarin.Android.Build.Tests.PerformanceTest.Build_No_Changes
    Test requires an updated MSBuild.StructuredLogger

We ignored these tests in 304e6989. This was failing due to changes in
the `.binlog` file format.

I think we can remove this with the latest MSBuild.StructuredLogger.


jonathanpeppers:
These tests are running and passing now:

image

dellis1972 (approved):

jonathanpeppers:
I think we can ignore the test timeout here (shouldn't be affected by this PR), going to merge.

PR #6174, opened 2021-08-09
Mark API-31 as stable.

jpobst:
Set API-S as API-31, v12.0, and stable.  Bump .NET 6 pack version to `31.0.100`.

Notes:
- API-31 requires JDK 11 to function, so bump Legacy CI lanes to use JDK 11 instead of JDK 1.8. (.NET6 lanes already used JDK 11.)
- Various tests updated to expect API-31 results.
- The source `.jar` used for API-31 documentation is not available yet, filed an issue to track: https://github.com/xamarin/xamarin-android/issues/6215.


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
```diff
@@ -504,7 +505,7 @@ public void DotNetBuild (string runtimeIdentifiers, bool isRelease, bool aot)
 		}
 
 		[Test]
-		public void SupportedOSPlatformVersion ([Values (21, 30)] int minSdkVersion)
+		public void SupportedOSPlatformVersion ([Values (21, 31)] int minSdkVersion)
```
jonathanpeppers, line 508: Maybe we should leave this one as API 30 and leave a comment why:
```
--min-api 31
```
This causes the latest version of d8 to print:
```
D8 An API level of 31 is not supported by this compiler. Please use an API level of 30 or earlier 
```
And it seems like we have the latest:
https://mvnrepository.com/artifact/com.android.tools/r8

jonathanpeppers (commented):
Go ahead and bump this to `31.0.100`:

https://github.com/xamarin/xamarin-android/blob/f7d32102cc3ada67aa12cc0beed2cf97226f41b3/Directory.Build.props#L24

Do we know if it will be Android 12? If we don't know yet, we can leave this one alone:

https://github.com/xamarin/xamarin-android/blob/f7d32102cc3ada67aa12cc0beed2cf97226f41b3/Directory.Build.props#L16

jonathanpeppers:
It doesn't look like 823fec5650ee21673f323d02a32092ae0110a626 fixed the issue with binding projects.

It's tough because these tests pass for me locally somehow...

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
jpobst: You marked this as resolved but did not make the change you suggested.  This seems to be the only remaining test failure on this PR.  Should we commit your suggestion?

jonathanpeppers: Somehow a commit got lost (or I didn't push it).

I still had the commit locally, I pushed it.

jonpryor:
@jonathanpeppers [asked](https://github.com/xamarin/xamarin-android/pull/6174#pullrequestreview-731877263):

> Do we know if it will be Android 12? If we don't know yet, we can leave this one alone:

The current plan will be to cherry-pick all the API-31 -related changes to the d16-11 branch.  d16-11 will becomes Xamarin.Android 12.0, and main will become 12.0.99.

dellis1972 (approved):

grendello (approved):

jonathanpeppers (approved):

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -552,17 +552,6 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 		DependsOnTargets="$(_OnResolveMonoAndroidSdks)">
 </Target>
 
-<!--
-  NOTE:
-  This target runs during Restore, and there is no $(RestoreDependsOn) property.
-  There appears to be no other way to do this other than use BeforeTargets.
--->
-<Target Name="_SetLatestTargetFrameworkVersionForPackageReference"
```
jonpryor, line 560: What's the rationale for this change?  What does it fix?  What could it break?

jonpryor: ah, moved to Legacy.  My bad.

jonathanpeppers: Details are in: https://github.com/xamarin/xamarin-android/pull/6174/commits/11289695e08b83429146e5648d81ab1453d9e0ab

Issue #6140, opened 2021-07-29
Don't set XA_TLS_PROVIDER env var for .NET 6 app


akoeplinger:
It's not used anymore and has no effect.

Probably also needs some update to this section in the docs:
https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/building-apps/build-properties.md#androidtlsprovider

PR #6182, opened 2021-08-12
[NET6] Don't export the XA_TLS_PROVIDER envvar

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/6140

`XA_TLS_PROVIDER` environment variable was used to set the TLS provider
for HttpClient in "legacy" Xamarin.Android.  NET6 and newer no longer
use that variable, so we can stop exporting it.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
@@ -72,6 +72,7 @@ namespace Xamarin.Android.Tasks
 		public string TlsProvider { get; set; }
 		public string AndroidSequencePointsMode { get; set; }
 		public bool EnableSGenConcurrent { get; set; }
+		public bool UsingAndroidNETSdk { get; set; }
 
 		[Output]
 		public string BuildId { get; set; }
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -160,8 +160,9 @@ static internal NativeAssemblerTargetProvider GetAssemblyTargetProvider (string
 		static readonly string[] defaultLogLevel = {"MONO_LOG_LEVEL", "info"};
 		static readonly string[] defaultMonoDebug = {"MONO_DEBUG", "gen-compact-seq-points"};
 		static readonly string[] defaultHttpMessageHandler = {"XA_HTTP_CLIENT_HANDLER_TYPE", "System.Net.Http.HttpClientHandler, System.Net.Http"};
+#if !NETCOREAPP
```
jonathanpeppers, line 163: We don't actually multitarget `Xamarin.Android.Build.Tasks.csproj`, it is just a `netstandard2.0` project right now...

The way to do this *now* would be to add a new property like:

https://github.com/xamarin/xamarin-android/blob/286683224ea9c38be3028dd03bf47b73f4da45f4/src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs#L53

And pass in `UsingAndroidNETSdk="$(UsingAndroidNETSdk)"`.

dellis1972:
Looks ok. Just rerunning a few jobs to see if they go green.

jonathanpeppers (approved):

dellis1972 (approved):

PR #6187, opened 2021-08-13
[One .NET] use rc.1 version number for main

jonathanpeppers:
Context: https://semvercompare.azurewebsites.net/?version=30.0.100-preview.7.110&version=30.0.100-ci.main.111

We have an issue currently where our version number on `main` is lower
than `preview.7`:

    30.0.100-ci.main.111 < 30.0.100-preview.7.110

With this setup, `dotnet workload install` cannot possibly update to
`ci.main` versions (it always grabs latest) -- unless we somehow
constructed a NuGet feed without `preview.7` builds.

The only way for people to use `ci.main` builds at present is with
`maui-check --main`, because it uses
`dotnet workload update --from-rollback-file` to manually specify
versions. We want `maui-check` to be *optional* if possible.

The solution for `ci.main` versions is:

* Switch `xamarin-android/main` to `rc.1` *now*.
* When we branch for `rc.1`, we update `main` to `rc.2`.

This follows the same pattern for dotnet/installer and other repos in
the dotnet organization. I updated `HowToBranch.md` accordingly.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -22,6 +22,7 @@
          * Bump first digit of the patch version for feature releases (and reset the first two digits to 0)
     -->
     <AndroidPackVersion>30.0.100</AndroidPackVersion>
+    <AndroidPackVersionSuffix>rc.1</AndroidPackVersionSuffix>
   </PropertyGroup>
 
   <!-- Common <PackageReference/> versions -->
--- a/Documentation/guides/HowToBranch.md
+++ b/Documentation/guides/HowToBranch.md
@@ -12,27 +12,28 @@ sequence of events would be:
 3. `xamarin-android` branches `release/6.0.1xx-preview42`. GitHub Web
    UI is fine for this.
 
-4. Manually make a commit to `release/6.0.1xx-preview42` such as
-  [df122518][2], so that `$(_AndroidPackLabel)` is
-  `preview.42.$(PackVersionCommitCount)`.
-
-5. Subscribe to Maestro updates for [dotnet/installer][1] `release/6.0.1xx-preview42`:
+4. Subscribe to Maestro updates for [dotnet/installer][1] `release/6.0.1xx-preview42`:
 
 ```bash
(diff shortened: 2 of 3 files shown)


pjcollins (approved):

jonpryor:
@jonathanpeppers : I think `HowToBranch.md` will also need to be updated to mention that the version band in `$(AndroidPackVersion)` will possibly need to be updated during branches as well.

How's this commit message?

```markdown
Context: https://semvercompare.azurewebsites.net/?version=30.0.100-preview.7.110&version=30.0.100-ci.main.111

We have an issue currently where our version number on `main` is
lower than `preview.7`:

	30.0.100-ci.main.111 < 30.0.100-preview.7.110

Furthermore, `main` will *always* be lower than `preview.7`, because
`c` (in `ci.main`) sorts before `p` (in `preview.7`).

With this setup, `dotnet workload install` cannot possibly update to
`ci.main` versions (it always grabs latest), unless we somehow
constructed a NuGet feed without `preview.7` builds.

The only way for people to use `ci.main` builds at present is with
`maui-check --main`, because it uses
`dotnet workload update --from-rollback-file` to manually specify
versions.

We want `maui-check` to be *optional* if possible.

To fix this conundrum:

 1. Change `$(_AndroidPackLabel)` so that on the `main` or `release/`
    branches, it instead uses `$(AndroidPackVersionSuffix)`.

 2. On `main`, `$(AndroidPackVersionSuffix)` will be `rc.1` *now*.
    This will cause net6 packages coming out of `main` to have a
    version number of e.g. `30.0.100-rc.1.111`, which will sort
    *after* `30.0.100-preview.7.110`.

 3. In the future when we create the actual rc.1 branch, we update
    `$(AndroidPackVersionSuffix)` on `main` to `rc.2`.

 4. When net6 is GA, the `$(AndroidPackVersion)` "version band" will
    be updated, and  `$(AndroidPackVersionSuffix)` on `main` will
    "roll around" to `preview.1`, resulting in a version number of
    e.g. `30.0.200-preview.1.10`.

This follows the same pattern for dotnet/installer and other repos in
the dotnet organization.  I updated `HowToBranch.md` accordingly.
```

PR #6188, opened 2021-08-16
Tune bundled assemblies

grendello:
Bundled assemblies are now stored on run time in an array pre-allocated
during build. This allows us to avoid any memory allocation during
startup. At the same time, a provision is made for adding more
assemblies than those counted and placed in the .apk during the build.
This may come handy when Xamarin.Android supports split
applications (with components placed in different APKs) and also as a
simple fail-safe feature should the build miscalculate something.

Debug information for assemblies is no longer registered when reading
the APK, but rather whenever an assembly load is requested by Mono.

Neither the assemblies nor their debug information are mmapped when
reading the APK anymore. This is done lazily when assembly load is
requested by Mono. Mapping of files after the startup phase is protected
with a POSIX semaphore.

Additionally, a handful of simple optimizations are implemented:

  * we no longer query the device's memory page size whenever we mmap a
    file, the value is instead queried and cached in the `Util`
    constructor.
  * current app domain is no longer queried when getting a Mono object's
    type under NET6
  * a number of log statements during startup no longer run by
    default

The changes reduce startup time for a simple (one control in a layout)
plain Xamarin.Android application by around 4ms measured for `Displayed`
time, and around 2ms when measured for native runtime startup.

For the `Hello Maui` sample from the `maui-samples` repository, the
startup time is reduced by ~200ms measured for `Displayed` time (from
~1.6s to ~1.4s), with the same speed up for the native runtime as in the
plain Xamarin.Android test above.

Performance was measured on a Pixel 3XL phone running Android 12 beta -
the `Displayed` measurements were found to be **very** unstable during
testing (with variance between runs reaching 600ms sometimes).

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
@@ -269,10 +269,38 @@ namespace Xamarin.Android.Tasks
 				throw new InvalidOperationException ($"Unsupported BoundExceptionType value '{BoundExceptionType}'");
 			}
 
+			int assemblyNameWidth = 0;
 			int assemblyCount = ResolvedAssemblies.Length;
+			Encoding assemblyNameEncoding = Encoding.UTF8;
+
+			Action<ITaskItem> updateNameWidth = (ITaskItem assembly) => {
+				string assemblyName = Path.GetFileName (assembly.ItemSpec);
+				int nameBytes = assemblyNameEncoding.GetBytes (assemblyName).Length;
+				if (nameBytes > assemblyNameWidth) {
+					assemblyNameWidth = nameBytes;
+				}
+			};
+
 			if (SatelliteAssemblies != null) {
 				assemblyCount += SatelliteAssemblies.Length;
+
+				foreach (ITaskItem assembly in SatelliteAssemblies) {
+					updateNameWidth (assembly);
+				}
 			}
+
+			foreach (var assembly in ResolvedAssemblies) {
+				updateNameWidth (assembly);
+			}
+
+			int abiNameLength = 0;
+			foreach (string abi in SupportedAbis) {
+				if (abi.Length <= abiNameLength) {
+					continue;
+				}
+				abiNameLength = abi.Length;
+			}
+			assemblyNameWidth += abiNameLength + 1; // room for '/'
 
 			bool haveRuntimeConfigBlob = !String.IsNullOrEmpty (RuntimeConfigBinFilePath) && File.Exists (RuntimeConfigBinFilePath);
 			var appConfState = BuildEngine4.GetRegisteredTaskObjectAssemblyLocal<ApplicationConfigTaskState> (ApplicationConfigTaskState.RegisterTaskObjectKey, RegisteredTaskObjectLifetime.Build);
@@ -295,6 +323,7 @@ namespace Xamarin.Android.Tasks
 					JniAddNativeMethodRegistrationAttributePresent = appConfState != null ? appConfState.JniAddNativeMethodRegistrationAttributePresent : false,
 					HaveRuntimeConfigBlob = haveRuntimeConfigBlob,
 					NumberOfAssembliesInApk = assemblyCount,
+					BundledAssemblyNameWidth = assemblyNameWidth + 1,
 				};
 
 				using (var sw = MemoryStreamPool.Shared.CreateStreamWriter ()) {
--- a/src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
@@ -11,6 +11,7 @@ namespace Xamarin.Android.Tasks
 		SortedDictionary <string, string> environmentVariables;
 		SortedDictionary <string, string> systemProperties;
 		uint stringCounter = 0;
+		uint bufferCounter = 0;
 
 		public bool IsBundledApp { get; set; }
 		public bool UsesMonoAOT { get; set; }
@@ -24,6 +25,7 @@ namespace Xamarin.Android.Tasks
 		public bool JniAddNativeMethodRegistrationAttributePresent { get; set; }
 		public bool HaveRuntimeConfigBlob { get; set; }
 		public int NumberOfAssembliesInApk { get; set; }
+		public int BundledAssemblyNameWidth { get; set; } // including the trailing NUL
 
 		public PackageNamingPolicy PackageNamingPolicy { get; set; }
 
@@ -90,6 +92,9 @@ namespace Xamarin.Android.Tasks
 				WriteCommentLine (output, "number_of_assemblies_in_apk");
 				size += WriteData (output, NumberOfAssembliesInApk);
 
+				WriteCommentLine (output, "bundled_assembly_name_width");
+				size += WriteData (output, BundledAssemblyNameWidth);
+
 				WriteCommentLine (output, "android_package_name");
 				size += WritePointer (output, MakeLocalLabel (stringLabel));
 
@@ -103,6 +108,57 @@ namespace Xamarin.Android.Tasks
 
 			WriteNameValueStringArray (output, "app_environment_variables", environmentVariables);
 			WriteNameValueStringArray (output, "app_system_properties", systemProperties);
+
+			WriteBundledAssemblies (output);
+		}
+
+		void WriteBundledAssemblies (StreamWriter output)
+		{
+			WriteCommentLine (output, $"Bundled assembly name buffers, all {BundledAssemblyNameWidth} bytes long");
+			WriteSection (output, ".bss.bundled_assembly_names", hasStrings: false, writable: true, nobits: true);
+
+			var name_labels = new List<string> ();
+			for (int i = 0; i < NumberOfAssembliesInApk; i++) {
+				string bufferLabel = GetBufferLabel ();
+				WriteBufferAllocation (output, bufferLabel, (uint)BundledAssemblyNameWidth);
+				name_labels.Add (bufferLabel);
+			}
+
+			string label = "bundled_assemblies";
+			WriteCommentLine (output, "Bundled assemblies data");
+			WriteDataSection (output, label);
+			WriteStructureSymbol (output, label, alignBits: TargetProvider.MapModulesAlignBits, isGlobal: true);
+
+			uint size = 0;
+			for (int i = 0; i < NumberOfAssembliesInApk; i++) {
+				size += WriteStructure (output, packed: false, structureWriter: () => WriteBundledAssembly (output, MakeLocalLabel (name_labels[i])));
+			}
+			WriteStructureSize (output, label, size);
+		}
+
+		uint WriteBundledAssembly (StreamWriter output, string nameLabel)
+		{
+			WriteCommentLine (output, "apk_fd");
+			uint size = WriteData (output, (int)-1);
+
+			WriteCommentLine (output, "data_offset");
+			size += WriteData (output, (uint)0);
+
+			WriteCommentLine (output, "data_size");
+			size += WriteData (output, (uint)0);
+
+			WriteCommentLine (output, "data");
+			size += WritePointer (output);
+
+			WriteCommentLine (output, "name_length");
+			size += WriteData (output, (uint)0);
+
+			WriteCommentLine (output, "name");
+			size += WritePointer (output, nameLabel);
+
+			output.WriteLine ();
+
+			return size;
 		}
 
 		void WriteNameValueStringArray (StreamWriter output, string label, SortedDictionary<string, string> entries)
@@ -148,6 +204,12 @@ namespace Xamarin.Android.Tasks
 		{
 			stringCounter++;
 			return $"env.str.{stringCounter}";
+		}
+
+		string GetBufferLabel ()
+		{
+			bufferCounter++;
+			return $"env.buf.{bufferCounter}";
 		}
 	};
 }
--- a/src/monodroid/jni/embedded-assemblies-zip.cc
+++ b/src/monodroid/jni/embedded-assemblies-zip.cc
@@ -9,6 +9,7 @@
 #include "embedded-assemblies.hh"
 #include "cpp-util.hh"
 #include "globals.hh"
+#include "xamarin-app.hh"
 
 using namespace xamarin::android::internal;
 
@@ -23,8 +24,18 @@ using read_count_type = unsigned int;
 using read_count_type = size_t;
 #endif
 
+force_inline bool
+EmbeddedAssemblies::is_debug_file (dynamic_local_string<SENSIBLE_PATH_MAX> const& name) noexcept
+{
+	return utils.ends_with (name, ".pdb")
+#if !defined (NET6)
+		|| utils.ends_with (name, ".mdb")
+#endif
+		;
+}
+
 void
-EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_should_register should_register)
+EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, [[maybe_unused]] monodroid_should_register should_register)
 {
 	uint32_t cd_offset;
 	uint32_t cd_size;
@@ -47,7 +58,7 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
 
 	std::vector<uint8_t>  buf (cd_size);
 	const char           *prefix     = get_assemblies_prefix ();
-	size_t                prefix_len = get_assemblies_prefix_length ();
+	uint32_t              prefix_len = get_assemblies_prefix_length ();
 	size_t                buf_offset = 0;
 	uint16_t              compression_method;
 	uint32_t              local_header_offset;
@@ -65,18 +76,8 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
 	bool runtime_config_blob_found = false;
 #endif // def NET6
 
-	bool bundled_assemblies_slow_path = false;
-	if (bundled_assemblies.empty ()) {
-		// We're called for the first time, let's allocate space for all the assemblies counted during build. resize()
-		// will allocate all the memory and zero it in one operation
-		bundled_assemblies.resize (application_config.number_of_assemblies_in_apk);
-	} else {
-		// We are probably registering some assemblies dynamically, possibly loading from another apk, we need to extend
-		// the storage instead of taking advantage of the pre-allocated slots.
-		bundled_assemblies_slow_path = true;
-	}
-
-	int64_t assembly_count = application_config.number_of_assemblies_in_apk;
+	bool bundled_assemblies_slow_path = bundled_assembly_index >= application_config.number_of_assemblies_in_apk;
+	uint32_t max_assembly_name_size = application_config.bundled_assembly_name_width - 1;
 
 	// clang-tidy claims we have a leak in the loop:
 	//
@@ -92,7 +93,7 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
 		bool result = zip_read_entry_info (buf, buf_offset, compression_method, local_header_offset, file_size, entry_name);
 
 #ifdef DEBUG
-		log_warn (LOG_ASSEMBLY, "%s entry: %s", apk_name, entry_name.get () == nullptr ? "unknown" : entry_name.get ());
+		log_info (LOG_ASSEMBLY, "%s entry: %s", apk_name, entry_name.get () == nullptr ? "unknown" : entry_name.get ());
 #endif
 		if (!result || entry_name.empty ()) {
 			log_fatal (LOG_ASSEMBLY, "Failed to read Central Directory info for entry %u in APK file %s", i, apk_name);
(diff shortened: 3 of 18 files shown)


src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -129,25 +120,31 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
...
+#endif
 
-		if ((utils.ends_with (entry_name, ".pdb") || utils.ends_with (entry_name, ".mdb")) &&
-				register_debug_symbols &&
-				!entry_is_overridden &&
-				bundled_assembly_index >= 1) {
-			md_mmap_info map_info = md_mmap_apk_file(fd, data_offset, file_size, entry_name.get (), apk_name);
-			if (register_debug_symbols_for_assembly (entry_name, bundled_assemblies [bundled_assembly_index - 1], (const mono_byte*)map_info.area, static_cast<int>(file_size)))
-				continue;
+		if (register_debug_symbols && !entry_is_overridden && (utils.ends_with (entry_name, ".pdb") || utils.ends_with (entry_name, ".mdb"))) {
```
jonathanpeppers, line 130: In .NET 6, `.mdb` files won't even make it to the `.apk` anymore. Is it worth making this legacy-only?

grendello: Yeah, I was wondering about that. If the `.mdb` files aren't a thing in NET6 I will definitely make that legacy only :)

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -295,6 +321,7 @@ void AddEnvironment ()
 					JniAddNativeMethodRegistrationAttributePresent = appConfState != null ? appConfState.JniAddNativeMethodRegistrationAttributePresent : false,
 					HaveRuntimeConfigBlob = haveRuntimeConfigBlob,
 					NumberOfAssembliesInApk = assemblyCount,
+					BundledAssemblyNameWidth = assemblyNameWidth + 1,
```
jonpryor, line 324: Why the `+1` here?  Terminating NUL? If so, should `ApplicationConfigNativeAssemblyGenerator` do that addition instead?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -65,18 +76,8 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
...
-		bundled_assemblies.resize (application_config.number_of_assemblies_in_apk);
-	} else {
-		// We are probably registering some assemblies dynamically, possibly loading from another apk, we need to extend
-		// the storage instead of taking advantage of the pre-allocated slots.
-		bundled_assemblies_slow_path = true;
-	}
-
-	int64_t assembly_count = application_config.number_of_assemblies_in_apk;
+	bool bundled_assemblies_slow_path = bundled_assembly_index >= application_config.number_of_assemblies_in_apk;
+	uint32_t max_assembly_name_size = application_config.bundled_assembly_name_width - 1;
```
jonpryor, line 80: This is "differently" amusing: we have to do `+1` to support the terminating NUL, yet when we hit native code we just `-1` that value.

Do we actually *need* the terminating NUL?

Probably, actually?

The `set_entry_data<false>()` method uses `max_assembly_name_size` as the `len` parameter to **strncpy**(3), and `strncpy()` won't read or write past `len`, it *feels like* there could be a situation where `XamarinAndroidBundledAssembly::name` isn't properly NUL-terminated?

src/monodroid/jni/embedded-assemblies-zip.cc
```diff
@@ -157,42 +164,59 @@ EmbeddedAssemblies::zip_load_entries (int fd, const char *apk_name, monodroid_sh
...
-		}
+template<bool NeedsNameAlloc>
+force_inline void
+EmbeddedAssemblies::set_entry_data (XamarinAndroidBundledAssembly &entry, int apk_fd, uint32_t data_offset, uint32_t data_size, uint32_t prefix_len, uint32_t max_name_size, dynamic_local_string<SENSIBLE_PATH_MAX> const& entry_name) noexcept
+{
+	entry.apk_fd = apk_fd;
+	if constexpr (NeedsNameAlloc) {
+		entry.name = utils.strdup_new (entry_name.get () + prefix_len);
+	} else {
+		strncpy (entry.name, entry_name.get () + prefix_len, max_name_size);
```
jonpryor, line 203: Should this also explicitly set `entry.name[max_name_size]=0` or `entry.name[max_name_size-1]=0`?

Given the `+1`/`-1` operations elsewhere, I *think* `libxamarin-app.so` will allocate the `XamarinAndroidBundledAssembly::name` entry to be of length `char[max_name_size+1]` (due to the `-1` in `libmonodroid.so`, and thus it'll come "pre-NUL-terminated", but it's taken me more mental effort than I'd like to admit to come to this conclusion.

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -269,10 +269,36 @@ void AddEnvironment ()
 				throw new InvalidOperationException ($"Unsupported BoundExceptionType value '{BoundExceptionType}'");
 			}
 
+			int assemblyNameWidth = 0;
 			int assemblyCount = ResolvedAssemblies.Length;
+
+			Action<ITaskItem> updateNameWidth = (ITaskItem assembly) => {
+				string assemblyName = Path.GetFileName (assembly.ItemSpec);
+				if (assemblyName.Length > assemblyNameWidth) {
+					assemblyNameWidth = assemblyName.Length;
```
jonpryor, line 278: I like the idea, but this isn't "Unicode"-aware: for better or worse, this is entirely valid:

```
% csc /out:😀.exe hw.cs
% mono 😀.exe               
Hello, world!
```

or for the `.csproj`-inclined:

```xml
  <PropertyGroup>
    <AssemblyName>😀</AssemblyName>
  </PropertyGroup>
```

On macOS, what *is* 😀?

```
% ls -l | tail -1 | hexdump -C
00000000  2d 72 77 2d 72 2d 2d 72  2d 2d 20 20 20 20 31 20  |-rw-r--r--    1 |
00000010  6a 6f 6e 20 20 73 74 61  66 66 20 20 20 20 20 20  |jon  staff      |
00000020  33 35 38 34 20 41 75 67  20 31 37 20 31 39 3a 33  |3584 Aug 17 19:3|
00000030  32 20 f0 9f 98 80 2e 65  78 65 0a                 |2 .....exe.|
```

i.e. 😀 is { f0 9f 98 80 }, which is *4* bytes, not 1.

You shouldn't assume that assembly names are restricted to ASCII.

Is this even possible in Xamarin.Android?  I don't know about *emoji*, but we *do* use umlauts within directory names: https://github.com/xamarin/xamarin-android/blob/f7d32102cc3ada67aa12cc0beed2cf97226f41b3/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs#L159

…which means we should improve that test (and others?) so that *assemblies* also contain umlauts.  Pity we missed that opportunity.

jonpryor: Related, in terms of Debug typemap generation: https://github.com/xamarin/java.interop/blob/ff2714200107fb616828b9d1013f69605791d2ba/src/Java.Interop.Tools.JavaCallableWrappers/Java.Interop.Tools.JavaCallableWrappers/TypeNameMapGenerator.cs#L229-L232

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
```diff
@@ -90,6 +92,9 @@ protected override void WriteSymbols (StreamWriter output)
 				WriteCommentLine (output, "number_of_assemblies_in_apk");
 				size += WriteData (output, NumberOfAssembliesInApk);
 
+				WriteCommentLine (output, "bundled_assembly_name_width");
+				size += WriteData (output, BundledAssemblyNameWidth);
```
jonpryor, line 96: e.g. perhaps *this* should contain `+1 /* NUL */`?

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
```diff
@@ -103,6 +108,57 @@ protected override void WriteSymbols (StreamWriter output)
...
+
+		void WriteBundledAssemblies (StreamWriter output)
+		{
+			WriteCommentLine (output, $"Bundled assembly name buffers, all {BundledAssemblyNameWidth} bytes long");
+			WriteSection (output, ".bss.bundled_assembly_names", hasStrings: false, writable: true, nobits: true);
+
+			var name_labels = new List<string> ();
+			for (int i = 0; i < NumberOfAssembliesInApk; i++) {
+				string bufferLabel = GetBufferLabel ();
+				WriteBufferAllocation (output, bufferLabel, (uint)BundledAssemblyNameWidth);
```
jonpryor, line 123: Ditto here `(uint) BundledAssemblyNameWidth+1 /* NUL */`?

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
grendello: No, `ApplicationConfigNativeAssemblyGenerator` doesn't calculate the value - its task is just to write it.

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: The NUL is just a failsafe, just as is `max_assembly_name_size`.

src/monodroid/jni/embedded-assemblies-zip.cc
grendello: `strncpy` copies **at most** `n` bytes (in this case `max_name_size`) which won't include the terminating NUL byte, and since this line is used only with pre-allocated buffers, the string is terminated properly even if the name takes the entire `max_name_size` characters. `strncpy` also stops copying at the first `\0` it encounters and copies it to the destination buffer.

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
grendello: Good point, I keep forgetting that people can do weird things (especially with emoji) :)

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
grendello: No, as said above, this class merely generates the native assembly - value calculation is not its concern.

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
grendello: Ditto :)

grendello:
The test failures don't appear to be caused by this PR. I managed to reproduce them locally using the tip of `main`. The WebSocket tests fail because they can't reach the `echo.websocket.org` server (using `ws://echo.websocket.org`) and the interpreter crash appears to be caused by an issue in the legacy Mono runtime:

```
08-18 12:39:20.475 25293 25312 F libc    : Fatal signal 4 (SIGILL), code 0 (SI_USER) in tid 25312 (Instrumentation), pid 25293 (o.Android_Tests)
08-18 12:39:20.548  3468  3533 D DeviceStateHelper: Audio mode: 0
08-18 12:39:20.586 25323 25323 I crash_dump32: obtaining output fd from tombstoned, type: kDebuggerdTombstoneProto
08-18 12:39:20.587   876   876 I tombstoned: received crash request for pid 25312
08-18 12:39:20.588 25323 25323 I crash_dump32: performing dump of process 25293 (target tid = 25312)
08-18 12:39:20.609 25323 25323 E DEBUG   : failed to read /proc/uptime: Permission denied
08-18 12:39:20.745 25323 25323 W unwind  : Failed to initialize DEX file support: dlopen failed: library "libdexfile.so" not found
08-18 12:39:20.860 25323 25323 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
08-18 12:39:20.860 25323 25323 F DEBUG   : Build fingerprint: 'google/crosshatch/crosshatch:12/SPB4.210715.011/7608474:user/release-keys'
08-18 12:39:20.860 25323 25323 F DEBUG   : Revision: 'MP1.0'
08-18 12:39:20.860 25323 25323 F DEBUG   : ABI: 'arm'
08-18 12:39:20.860 25323 25323 F DEBUG   : Timestamp: 2021-08-18 12:39:20.608873094+0200
08-18 12:39:20.860 25323 25323 F DEBUG   : Process uptime: 0s
08-18 12:39:20.860 25323 25323 F DEBUG   : Cmdline: Mono.Android_Tests
08-18 12:39:20.860 25323 25323 F DEBUG   : pid: 25293, tid: 25312, name: Instrumentation  >>> Mono.Android_Tests <<<
08-18 12:39:20.860 25323 25323 F DEBUG   : uid: 10517
08-18 12:39:20.860 25323 25323 F DEBUG   : signal 4 (SIGILL), code 0 (SI_USER), fault addr --------
08-18 12:39:20.860 25323 25323 F DEBUG   :     r0  00000000  r1  00000000  r2  00000030  r3  b9efbc50
08-18 12:39:20.860 25323 25323 F DEBUG   :     r4  b9ff6840  r5  b9c76326  r6  00000000  r7  bbe5d564
08-18 12:39:20.861 25323 25323 F DEBUG   :     r8  b9efbcfc  r9  b9ff6828  r10 e8055eb0  r11 b9efbc68
08-18 12:39:20.861 25323 25323 F DEBUG   :     ip  00000000  sp  b9efba90  lr  bbdaec80  pc  bbdae968
08-18 12:39:20.861 25323 25323 F DEBUG   : backtrace:
08-18 12:39:20.861 25323 25323 F DEBUG   :       #00 pc 0012b968  /data/app/~~ra-9GTek0k9wm1fjca-eLw==/Mono.Android_Tests-CI3wvNJ2tvljF1F-oWIDsw==/lib/arm/libmonosgen-2.0.so (interp_exec_method_full+48928)
```

I have no idea why the designer tests crash though... They shouldn't be affected by this PR, which is about code that runs on Android device.

grendello:
WebSocket tests will be disabled by [this PR](https://github.com/xamarin/xamarin-android/pull/6197)

jonpryor:
```markdown
When Embedded Assemblies are used (`$(EmbedAssembliesIntoApk)`=True),
assembly names are detected during `.apk` content traversal,
duplicated, and stored into `bundled_assemblies` and
`MonoBundledAssembly::name`.

Reduce memory allocations by updating `libxamarin-app.so` to contain
a `bundled_assembly_names` array, a'la:

	struct XamarinAndroidBundledAssembly {
	    int32_t     apk_fd;
	    uint32_t    data_offset;
	    uint32_t    data_size;
	    uint8_t    *data;
	    uint32_t    name_length;
	    char       *name;
	};
	constexpr size_t AssemblyNameWidth;
	static char assembly_name_1 [AssemblyNameWidth];
	static char assembly_name_2 [AssemblyNameWidth];
	// …

	extern "C" XamarinAndroidBundledAssembly bundled_assemblies[] = {
	    {
	        .apk_fd         = -1,
	        .data_offset    = 0,
	        .data_size      = 0,
	        .name_length	= 0,
	        .name           = assembly_name_1,
	    },
	    {
	        .apk_fd         = -1,
	        .data_offset    = 0,
	        .data_size      = 0,
	        .name_length	= 0,
	        .name           = assembly_name_2,
	    },
	    // …
	}

The `bundled_assemblies` array *doesn't* contain the assembly names
detected during packaging time.  Rather, it contains *buffers*, each
large enough to hold any assembly name.

This removes the need to allocate additional memory for assembly
names during process startup.  (We'll still need to allocate memory
for other entries within the `.apk`, such as `.pdb` files.)

Additionally, a provision is made for adding more assemblies than
those counted and placed in the .`apk` during the build.  This may
come handy when Xamarin.Android supports split applications (with
components placed in different `.apk` files), and also as a simple
fail-safe feature should the build miscalculate something.

Debug information for assemblies is no longer registered when reading
the APK, but rather whenever an assembly load is requested by Mono.

Neither the assemblies nor their debug information are mmapped when
reading the APK anymore.  This is done lazily when assembly load is
requested by Mono.  Mapping of files after the startup phase is
protected with a POSIX semaphore.

Additionally, a handful of simple optimizations are implemented:

  * We no longer query the device's memory page size whenever we
    `mmap()` a file; the value is instead queried and cached in the
    `Util` constructor.
  * The current AppDomain is no longer queried when getting a Mono
    object's type under .NET 6
  * Many log statements are no longer run by default during startup.

These changes reduce startup time for a simple plain
Xamarin.Android application (one control in a layout) by around 4ms
measured for `Displayed` time, and by around 2ms when measured for
native runtime startup.

For the `Hello Maui` sample from the `maui-samples` repository, the
startup time is reduced by ~200ms measured for `Displayed` time (from
~1.6s to ~1.4s), with the same speed up for the native runtime as in
the plain Xamarin.Android test above.

Performance was measured on a Pixel 3XL phone running Android 12 beta.
*Note*: the `Displayed` measurements were found to be **very** unstable
during testing on Android 12, with variance between runs reaching 600ms
sometimes.
```

src/monodroid/jni/embedded-assemblies.hh
```diff
@@ -81,6 +83,12 @@ namespace xamarin::android::internal {
 #endif // def NET6
 		MonoReflectionType* typemap_java_to_managed (MonoString *java_type);
 
+		void prepare_for_multiple_threads () noexcept
+		{
+			int ret = sem_init (&assembly_mmap_semaphore, 0 /* pshared */, 1 /* value */);
+			abort_unless (ret == 0, "Failed to initialize assembly mapping semaphore. %s", strerror (errno));
```
jonathanpeppers, line 89: I think this `abort` is happening during the Designer tests since this got merged?

I got this from their log, is it useful?
```
[2021-08-18 23:21:55.5] Renderer >> =================================================================
[2021-08-18 23:21:55.5] Renderer >> 	Native Crash Reporting
[2021-08-18 23:21:55.5] Renderer >> =================================================================
[2021-08-18 23:21:55.5] Renderer >> Got a abrt while executing native code. This usually indicates
[2021-08-18 23:21:55.5] Renderer >> a fatal error in the mono runtime or one of the native libraries 
[2021-08-18 23:21:55.5] Renderer >> used by your application.
[2021-08-18 23:21:55.5] Renderer >> =================================================================
[2021-08-18 23:21:55.5] Renderer >> 
[2021-08-18 23:21:55.5] Renderer >> =================================================================
[2021-08-18 23:21:55.5] Renderer >> 	Native stacktrace:
[2021-08-18 23:21:55.5] Renderer >> =================================================================
[2021-08-18 23:21:55.5] Renderer >> 	0x11d9349d8 - /Library/Frameworks/Xamarin.Android.framework/Versions/11.4.99.110/lib/xamarin.android/xbuild/Xamarin/Android/lib/host-Darwin/libmonosgen-2.0.dylib : mono_dump_native_crash_info
[2021-08-18 23:21:55.5] Renderer >> 	0x11d8cf38e - /Library/Frameworks/Xamarin.Android.framework/Versions/11.4.99.110/lib/xamarin.android/xbuild/Xamarin/Android/lib/host-Darwin/libmonosgen-2.0.dylib : mono_handle_native_crash
[2021-08-18 23:21:55.5] Renderer >> 	0x11d933fdf - /Library/Frameworks/Xamarin.Android.framework/Versions/11.4.99.110/lib/xamarin.android/xbuild/Xamarin/Android/lib/host-Darwin/libmonosgen-2.0.dylib : sigabrt_signal_handler
[2021-08-18 23:21:55.5] Renderer >> 	0x7fff2043dd7d - /usr/lib/system/libsystem_platform.dylib : _sigtramp
[2021-08-18 23:21:55.5] Renderer >> 	0x1060aa468 - Unknown
[2021-08-18 23:21:55.5] Renderer >> 	0x7fff2034d406 - /usr/lib/system/libsystem_c.dylib : abort
[2021-08-18 23:21:55.5] Renderer >> 	0x11d7f0cbd - /Library/Frameworks/Xamarin.Android.framework/Versions/11.4.99.110/lib/xamarin.android/xbuild/Xamarin/Android/lib/host-Darwin/libmono-android.debug.dylib : _ZL15do_abort_unlessbPKcz
[2021-08-18 23:21:55.5] Renderer >> 	0x11d7eda7a - /Library/Frameworks/Xamarin.Android.framework/Versions/11.4.99.110/lib/xamarin.android/xbuild/Xamarin/Android/lib/host-Darwin/libmono-android.debug.dylib : _ZN7xamarin7android8internal16MonodroidRuntime38Java_mono_android_Runtime_initInternalEP7JNIEnv_P7_jclassP8_jstringP13_jobjectArrayS8_SA_P8_jobjectSA_ih
[2021-08-18 23:21:55.5] Renderer >> 	0x11d7ed005 - /Library/Frameworks/Xamarin.Android.framework/Versions/11.4.99.110/lib/xamarin.android/xbuild/Xamarin/Android/lib/host-Darwin/libmono-android.debug.dylib : Java_mono_android_Runtime_init
[2021-08-18 23:21:55.5] Renderer >> 	0x1061b33a7 - Unknown
[2021-08-18 23:21:55.5] Renderer >> 	0x1061a2ffd - Unknown
[2021-08-18 23:21:55.5] Renderer >> 	0x1061a27d0 - Unknown
[2021-08-18 23:21:55.5] Renderer >> 
[2021-08-18 23:21:55.5] Renderer >> =================================================================
[2021-08-18 23:21:55.5] Renderer >> 	Basic Fault Address Reporting
[2021-08-18 23:21:55.5] Renderer >> =================================================================
[2021-08-18 23:21:55.5] Renderer >> Memory around native instruction pointer (0x7fff203c992e):0x7fff203c991e  ff ff c3 90 90 90 b8 48 01 00 02 49 89 ca 0f 05  .......H...I....
[2021-08-18 23:21:55.5] Renderer >> 0x7fff203c992e  73 08 48 89 c7 e9 a1 a1 ff ff c3 90 90 90 b8 53  s.H............S
[2021-08-18 23:21:55.5] Renderer >> 0x7fff203c993e  00 00 02 49 89 ca 0f 05 73 08 48 89 c7 e9 89 a1  ...I....s.H.....
[2021-08-18 23:21:55.5] Renderer >> 0x7fff203c994e  ff ff c3 90 90 90 b8 83 01 00 02 49 89 ca 0f 05  ...........I....
[2021-08-18 23:21:55.5] Renderer >> 
[2021-08-18 23:21:55.5] Renderer >> =================================================================
[2021-08-18 23:21:55.5] Renderer >> 	Managed Stacktrace:
[2021-08-18 23:21:55.5] Renderer >> =================================================================
[2021-08-18 23:21:55.5] Renderer >> =================================================================
```

grendello: It does help, thanks!

Issue #6058, opened 2021-06-30
[One .NET] split up Microsoft.Android.Runtime packs


jonathanpeppers:
We currently have 4 packs:

Microsoft.Android.Runtime.android-arm
Microsoft.Android.Runtime.android-arm64
Microsoft.Android.Runtime.android-x86
Microsoft.Android.Runtime.android-x64

These all contain duplicate assemblies:

![image](https://user-images.githubusercontent.com/840039/123993588-2b4f4500-d992-11eb-8250-0b4172830f99.png)

There might be a way to put these files in a common pack. This would save ~90mb on disk and ~20mb download size.

jonathanpeppers:
As seen in https://github.com/xamarin/xamarin-android/pull/6193#pullrequestreview-733380565, `Mono.Android.pdb` is about 3MB, so we'd save 9MB if we could move all managed code to a single pack somehow.

PR #6193, opened 2021-08-17
[One .NET] Include symbol files in runtime packs

pjcollins:
Adds the .pdb files associated with our framework assemblies to the .NET
runtime packs.  This will allow them to be fast deployed or included in
an .aab/.apk for Debug builds.  A couple of symbol related tests have
been updated to also check for Mono.Android.pdb.

I also noticed that the `SmokeTest` category was not running for the
`net6.0` MSBuildDeviceIntegration tests, as the emulator smoke test job
only runs against the `net472` test assembly.  This currently doesn't
matter as all `net6.0` smoke tests are ignored through other category
attributes, but this fix should be needed in the future.


pjcollins:
The test failures here look unrelated to me

jonathanpeppers (approved):
I restarted a couple test phases that had weird errors.

This seems like it adds about 3MB to each runtime pack:

image

So I'll comment on https://github.com/xamarin/xamarin-android/issues/6058.

jonathanpeppers:
A lot of tests failed, I'm going to try a rerun which should pick up the latest changes in main.

jonathanpeppers:
Only a legacy `ConcurrentGC` test failed, going to merge this.

PR #6208, opened 2021-08-19
[build] Fix release branch package versions

pjcollins:
I noticed our `release/6.0.1xx-rc1` branch was producing packs with a
strange version string:

    30.0.100-ci.release-6-0-1xx-rc1.14

The regex performed on `$(_AndroidPackBranch)` strips all `/` characters
from the branch name, causing the release branch condition specified in
`$(_AndroidPackBranch.StartsWith('release/'))` to never be true.

Fix this condition by using the unaltered `$(XAVersionBranch)` property
value.


pjcollins:
I tested this with a `release/*` branch locally,
before:
image

after:
image

jonathanpeppers (approved):

PR #6210, opened 2021-08-19
Speed up p/invoke override lookups

grendello:
Instead of using the `robin_map` hash map for the internal p/invokes,
incurring startup time penalty as the map has to be instantiated and
populated with `std::string` instances, use a plain array of structures
describing each p/invoke entry.  Each structure contains the p/invoke
function name, a pointer to that function as well as a hash calculated
on the build time using the xxHash algorithm.  The arrays are sorted on
the hash value so that during run time we can use a quick binary search
to locate the right function for the requested p/invoke.

There are two sets of p/invokes - one for internal Xamarin.Android
p/invokes (corresponding to the `xa-internal-api` and `java-interop`
"library" names) and another for the shared libraries currently shipped
with DotNet6.  The former are fully configured on build time, since we
can directly specify pointers to the functions implementing a p/invoke.
The latter, however, require more work on the run time since we cannot
set the pointer to a function during build.  Instead, on first request
for a given p/invoke, we load the relevant library and look up the
correct symbol, storing the pointer in the array afterwards.

Both the internal and dotnet library names are matched without using
string comparison.  Instead, a hash of the library name passed from Mono
is generated and compared against pre-generated hashes to determine
which table is to be used for lookups.

Both the tables and library name hashes are generated by the
`generate-pinvoke-tables` C++ program which also serves as a build-time
test of the xxHash implementation (it includes a handful of simple
compile-time tests).  Alas, currently this program can be built only on
the Linux CI bots, since neither Windows nor macOS machines there
support enough of C++20.  For this reason, the generated file is
committed to the repository instead of generating it each time
`src/monodroid` is built.  However, when running on CI Linux machines,
we do compile the utility, generate the tables and compare the generated
file to the one committed to repository.  If the files differ the build
will error out and let us examine the differences (both the new
generated file and the diff are part of the build status archive).  More
work needs to be done in order to compare the public interfaces found in
the dotnet shared libraries against the table stored in the generator.

The generated tables allow us to not use the "slow" `robin_map<>` hash
path during application startup (MAUI apps use the DotNet p/invokes
during that phase).  Additionally, the `robin_map<>` cache now uses the
xxHash hashes as well for faster lookups.

In addition, both the embedded assembly mmap code as well as the DotNet
p/invoke entry code now use the GCC/clang `__atomic` interfaces to
atomically set pointers, which avoids having to use any form of locks.

A new lock guard class, `StartupAwereLock`, is used to take a lock only
after the startup phase of Xamarin.Android is complete and it's possible
that multiple threads may be calling the runtime routines.

All of the above changes improve application startup time on a Pixel 3
XL device by:

  * Plain XA sample: ~17ms faster
  * Hello MAUI sample: ~100ms faster


src/monodroid/jni/pinvoke-tables.include
```diff
@@ -0,0 +1,1327 @@
...
+//
+// To regenerate run ../../../build-tools/scripts/generate-pinvoke-tables.sh on Linux or macOS
+// A compiler with support for C++20 ranges is required
+//
+
+#include <cstdint>
+
+#if INTPTR_MAX == INT64_MAX
+//64-bit internal p/invoke table
+static PinvokeEntry internal_pinvokes[] = {
```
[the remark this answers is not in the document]
grendello: Yeah, `const` doesn't matter here

src/monodroid/jni/generate-pinvoke-tables.cc
```diff
@@ -0,0 +1,859 @@
...
+#include <unordered_set>
+#include <vector>
+
+#include "xxhash.hh"
+
+namespace fs = std::filesystem;
+using namespace xamarin::android;
+
+const std::vector<std::string> internal_pinvoke_names = {
+	"create_public_directory",
```
jonpryor, line 31: This feels like a "better late than never" question, but… Why is `create_public_directory` exported?

It's not P/Invoked, afaict, so why is it exported?  What's using it?

src/monodroid/jni/xxhash.hh
```diff
@@ -0,0 +1,242 @@
+#if !defined (__XXHASH_HH)
+#define __XXHASH_HH
+
+//
+// Based on original code at https://github.com/ekpyron/xxhashct
```
jonpryor, line 5: Sounds like this code was copied from "elsewhere".

Please move it into `src-ThirdParty`, and update `build-tools/xaprepare/xaprepare/Scenarios/Scenario_ThirdPartyNotices.cs` & co. so that the in-tree `ThirdPartyNotices.txt` appropriately mentions this code.

src/monodroid/jni/generate-pinvoke-tables.cc
grendello: I think we should leave the review of exported APIs to some later date. I have no idea what uses this function, but there might be something somewhere that does...

src/monodroid/jni/xxhash.hh
grendello: This code is not copied from the above repository verbatim, I modified it quite a bit.

jonpryor:
```markdown
[monodroid] Speed up P/Invoke override lookups (#6210)

Commit 0cd890bd added use of [robin_map][0] to cache P/Invokes lookups.
This incurred a startup time penalty, as the map must be instantiated
and populated with `std::string` instances.

Continue using `robin_map` as a fallback cache, and introduce a
"primary" P/Invoke lookup data structure, an array of structures
describing each P/Invoke entry:

	struct PinvokeEntry {
	    hash_t     hash;
	    const char *name;
	    void       *func;
	};
	static PinvokeEntry internal_pinvokes[] = {
	    // …
	};
	static PinvokeEntry dotnet_pinvokes[] = …

`PinvokeEntry::hash` is a hash of `PinvokeEntry::name`, computed via
a modified [xxHash algorithm][1], in `src/monodroid/jni/xxhash.hh`.

THe `internal_pinvokes` & `dotnet_pinvokes` arrays are sorted on
`PinvokeEntry::hash`, allowing a binary search to be performed after
hashing the function name to find.

The `internal_pinvokes` array contains P/Invokes for the
`xa-internal-api` and `java-interop` library names, and are fully
configured at build-time, as we can directly reference the target
functions:

	PinvokeEntry internal_pinvokes[] = {
	    {0x1e035ea, "java_interop_jnienv_get_string_chars", reinterpret_cast<void*>(&java_interop_jnienv_get_string_chars)},
	    // …
	};

The `dotnet_pinvokes` array *doesn't* mention the function targets:

	static PinvokeEntry dotnet_pinvokes[] = {
	    {0xaf6b1c, "AndroidCryptoNative_RsaPrivateDecrypt", nullptr},
	    // …
	};

`PinvokeEntry::func` will be computed on first request for a given
P/Invoke, when we load the relevant library and look up the symbol,
storing the pointer in the array afterwards.

Both the internal and dotnet library names are matched without using
string comparison.  Instead, a hash of the library name passed from
Mono is generated and compared against pre-generated hashes to
determine which table is to be used for lookups.

Both the tables and library name hashes are generated by the new
`generate-pinvoke-tables` C++ app, which also serves as a build-time
test of the xxHash implementation (it includes a handful of simple
compile-time tests).  Alas, currently this program can be built only
on the Linux CI bots, since neither Windows nor macOS machines
support enough of C++20.  For this reason, the generated file is
committed to the repository instead of generating it each time
`src/monodroid` is built.  When building on CI Linux machines, we do
compile the utility, generate the tables, and compare the generated
file to the one committed to repository.  If the files differ, the
build will error out and let us examine the differences (both the new
generated file and the diff are part of the build status archive).

More work needs to be done in order to compare the public interfaces
found in the dotnet shared libraries against the table stored in the
generator.

The generated tables allow us to not use the "slow" `robin_map<>` hash
path during application startup (MAUI apps use the .NET 6+ P/Invokes
during that phase).  Additionally, the `robin_map<>` cache now uses
the xxHash hashes as well for faster lookups.

In addition, both the embedded assembly mmap code as well as the
.NET 6+ P/Invoke entry code now use the GCC/clang `__atomic` APIs to
atomically set pointers, which avoids having to use any form of locks.

A new lock guard class, `StartupAwereLock`, is used to take a lock
only after the startup phase of Xamarin.Android is complete and it's
possible that multiple threads may be calling the runtime routines.

All of the above changes improve application startup time on a
Pixel 3 XL device by:

  * Plain XA sample: ~17ms faster
  * Hello MAUI sample: ~100ms faster

[0]: https://github.com/Tessil/robin-map
[1]: https://github.com/ekpyron/xxhashct
```

Issue #5652, opened 2021-02-23
Make Android SDK trimming safe


marek-safar:
As part of the .NET6 effort we are making the .NET Runtime [trimming safe](https://github.com/dotnet/designs/blob/main/accepted/2020/linking-libraries.md#trimming-compatibility) allow developers to annotate their libraries to be also trimming safe. As Android SDK is one of the primary workloads which will always use linker and cares about the size it's important to resolve any linker warnings which the developers would experience when enabling linker analyzers.

The work to enable the linker analyzer is not yet simple and the effort to change that is tracked at https://github.com/dotnet/sdk/issues/15962

jonpryor:
Context: https://github.com/xamarin/xamarin-android/issues/5856#issuecomment-826841163
Context: https://github.com/xamarin/xamarin-android/pull/5878

Warnings that need to be fixed include:

```
ILLink warning IL2072: Java.Interop.JniRuntime.SetMarshalMemberBuilder(JniRuntime.CreationOptions): '#0' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicParameterlessConstructor' in call to 'System.Object System.Activator::CreateInstance(System.Type)'. The return value of method 'System.Reflection.Assembly.GetType(String)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
ILLink warning IL2026: Java.Interop.JniRuntime.SetMarshalMemberBuilder(JniRuntime.CreationOptions): Using method 'System.Reflection.Assembly.GetType(String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Types might be removed.
ILLink warning IL2067: Java.Interop.JniRuntime.JniTypeManager.FindAndCallRegisterMethod(Type,JniNativeMethodRegistrationArguments): 'type' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.NonPublicMethods' in call to 'System.Reflection.RuntimeReflectionExtensions.GetRuntimeMethods(Type)'. The parameter 'marshalType' of method 'Java.Interop.JniRuntime.JniTypeManager.FindAndCallRegisterMethod(Type,JniNativeMethodRegistrationArguments)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.
[...]

PR #6216, opened 2021-08-20
[One .NET] fix remaining Mono.Android.Export linker warning

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/issues/5652

If you build an app with:

    > dotnet build -c Release -p:SuppressTrimAnalysisWarnings=false

You get the linker warning:

    src\Mono.Android\Android.Runtime\AndroidRuntime.cs(296,5): warning IL2026: Android.Runtime.AndroidTypeManager.CreateDynamicCallback(MethodInfo): Using method 'System.Reflection.Assembly.GetType(String)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. Types might be removed.

We can simply suppress the warning:

    [UnconditionalSuppressMessage ("Trimming", "IL2026", Justification = "DynamicDependency should preserve the Create() method.")]

If we applied this, every app would preserve
`Mono.Android.Export.dll`, so we should not do this:

    [DynamicDependency ("Create", "Java.Interop.DynamicCallbackCodeGenerator", "Mono.Android.Export")]

This is already preserved because we have a `DynamicDependency` in
`ExportAttribute.cs`:

    [DynamicDependency (DynamicallyAccessedMemberTypes.All, "Java.Interop.DynamicCallbackCodeGenerator", "Mono.Android.Export")]

I also had to bring in the source for `UnconditionalSuppressMessage`
for `monoandroid10` or `netstandard2.0` assemblies. I also changed an
MSBuild `Condition`, so it will continue to work in .NET 7.


jonathanpeppers:
The .NET 6 Windows test is seeing `Mono.Android.Export.dll` get preserved by the linker:
```
 16,861 assemblies/Mono.Android.Export.dll *2
```
But it's weird that this only happens for the Windows "in-tree" tests, not on the test lanes that download the packages built on Mac and install them as a workload.

src/Xamarin.Android.NamingCustomAttributes/System.Runtime.CompilerServices/UnconditionalSuppressMessageAttribute.cs
```diff
@@ -0,0 +1,92 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
jonpryor, line 1: If this file was copied from elsewhere -- which I think it was -- then it should instead be in e.g. `src-ThirdParty/…`.

This was overlooked with `DynamicallyAccessedMemberTypes.cs` and `DynamicDependencyAttribute.cs`, so those should likewise move. :-(

jonathanpeppers: These should be moved around now.

PR #6221, opened 2021-08-23
[Linux] Properly detect Debian bookworm/sid

grendello:
The current Debian unstable distribution (codenamed "bookworm") no
longer contains any release version information in the `/etc/os-release`
file (the `VERSION` and `VERSION_ID` and `VERSION_CODENAME` fields were
removed) causing `xaprepare` to fail when running on this version of
Debian.

Fix the Debian/unstable version detection by looking for known release
codenames and faking the version number.


grendello:
Merging since `xaprepare` and build stage succeeded on all the platforms and the change affects only Linux.

Issue #6198, opened 2021-08-18
SDK Binding projects shouldn't have Resource.designer.cs


jpobst:
Using: 6.0.100-preview.7.21379.14

In Legacy binding projects, a `Resource.designer.cs` file is not created and no `Resource` class is added to the assembly.  In .NET 6, `Resource.designer.cs` is created and a `Resource` class is added, even if it's "empty".

I assume this class is not needed as we've never had it before.

This is causing a few issues for AndroidX/GPS:
* The `Resource` class is generated in the default namespace, which is the assembly name, which isn't actually used elsewhere and conflicts with a type name.  eg: `Xamarin.AndroidX.LiveData.Resource` and `Xamarin.AndroidX.LiveData` types.
* The GPS default namespaces are actually invalid, since they are based on the `aar` name, resulting in types like `com.google.android.gms.play-services-basement.Resource`.

While these issues are ultimately fixable, it feels like we don't intend to add the `Resource` class for binding projects.  (Maybe `$(OutputType) = 'Library'`?)

Repro projects:
[Legacy.zip](https://github.com/xamarin/xamarin-android/files/7008165/Legacy.zip)
[NET6.zip](https://github.com/xamarin/xamarin-android/files/7008166/NET6.zip)

Note the projects won't fully build because I didn't do the metadata or dependencies for them, but they get to the `csc` step, which means the `Resource.designer.cs` has been generated already.

jonathanpeppers:
.NET 6 doesn't have binding projects any more, so that seems like it's causing the issue here.

Do we need an MSBuild property force `Resource.designer.cs` to be disabled?

@dellis1972 I was thinking this might check if `@(AndroidResource)` is empty, and skip it:

https://github.com/xamarin/xamarin-android/blob/e811ecb5b7a6288b382a4b393e86db58e7a95e7d/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L1217-L1219

But is there some case where a NuGet package, `.jar`, `.aar`, etc. might put values in `Resource.designer.cs`?

dellis1972:
We might need to check both `@(AndroidResource)` and `@(AndroidAarLibrary)` or '@(AndroidLibrary)' where `Bind = true`.

jonathanpeppers:
So in the `NET6.zip` above, the `Resource.designer.cs` has values inside. It has an `.aar` file with resources.

I found I could delete this line:

[...]

PR #6224, opened 2021-08-23
[One .NET] add $(AndroidGenerateResourceDesigner) property

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/6198

When migrating Xamarin.Android binding projects to .NET 6, one
difference is that `Resource.designer.cs` will be generated by .NET 6
class libraries.

There is not currently a way to turn this off, if you pass:

    dotnet build -p:AndroidUseIntermediateDesignerFile=false

Then `Resources\Resource.designer.cs` will be generated instead. There
is no way to prevent this behavior, because `$(AndroidResgenFile)`
defaults to this value.

To solve cases where you *don't* want `Resource.designer.cs` at all:

1. Remove the `$(AndroidResgenFile)` default value.
2. Add a new `$(AndroidGenerateResourceDesigner)` MSBuild property,
   that seems like it will be a better name for .NET 6.
3. `$(AndroidUseIntermediateDesignerFile)` is based on
   `$(AndroidGenerateResourceDesigner)`.

I added a test for the new property.

### DIFF
--- a/Documentation/guides/OneDotNetBindingProjects.md
+++ b/Documentation/guides/OneDotNetBindingProjects.md
@@ -217,5 +217,19 @@ As the default for .NET 6 is `disable`, the same applies for Xamarin Android pro
 
 Use `enable` as shown above to enable NRT support.
 
-
 [11-0-release-notes]: https://docs.microsoft.com/en-us/xamarin/android/release-notes/11/11.0
+
+### `Resource.designer.cs`
+
+In Xamarin.Android, Java binding projects did not support generating a `Resource.designer.cs` file.
+Since binding projects are just class libraries in .NET 6, this file will be generated. This could
+be a breaking change when migrating existing projects.
+
+To disable `Resource.designer.cs` generation, set `$(AndroidGenerateResourceDesigner)` to `false`
+in your `.csproj`:
+
+```xml
+<PropertyGroup>
+  <AndroidGenerateResourceDesigner>false</AndroidGenerateResourceDesigner>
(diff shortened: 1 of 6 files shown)


grendello (approved):

jpobst (approved):
We should probably add it to this list of things Upgrade Assistant should do for converting bindings projects:

https://github.com/dotnet/upgrade-assistant/issues/657

Documentation/guides/OneDotNetBindingProjects.md
```diff
@@ -217,5 +217,19 @@ As the default for .NET 6 is `disable`, the same applies for Xamarin Android pro
...
 Use `enable` as shown above to enable NRT support.
 
-
 [11-0-release-notes]: https://docs.microsoft.com/en-us/xamarin/android/release-notes/11/11.0
+
+### `Resource.designer.cs`
+
+In Xamarin.Android, Java binding projects did not support generating a `Resource.designer.cs` file.
+Since binding projects are just class libraries in .NET 6, this file will be generated. This could
+be a breaking change when migrating existing projects.
```
jonpryor, line 226: One request for improvement: please provide an example error message.  What should customers be looking for in their migrated projects to know that `$(AndroidGenerateResourceDesigner)`=false should be set?

jonathanpeppers: I added two examples where `Resource.designer.cs` can break.

PR #6252, opened 2021-08-31
[Xamarin.Android.Build.Tasks] Improve XA0031 error

jonpryor:
Context: https://aka.ms/vs2019-and-jdk-11
Context: https://github.com/xamarin/xamarin-android/wiki/JDK-11-Warning

The XA0031 error will be generated when
`$(TargetFrameworkVersion)`=v12.0 and JDK 1.8 is detected, as JDK-11
is required for API-31; see also 8140991e, 26770284.

The problem is the XA0031 error message:

	error XA0031: Java SDK 11.0 or above is required when using $(TargetFrameworkVersion) v12.0.

The error message implies that error can be fixed by installing JDK-11.

Which is correct: the XA0031 error *can* be fixed by installing JDK-11.

The problem is that *once JDK-11 is installed*, the Android Designer
breaks, as the Android Designer, as shipped in Visual Studio 16.11 and
Visual Studio for Mac 8.10, only works with JDK 1.8.

This can easily result in a scenario where a customer wants to use
API-31, installs JDK-11, and then can no longer use the Designer.
(This could be mentioned in the Release Notes, but historically nobody
reads the release notes.)

Update the XA0031 error message so that this implicit dependency is
made explicit:

	error XA0031: Java SDK 11.0 or above is required when using $(TargetFrameworkVersion) v12.0.  Note: the Android Designer is incompatible with JDK-11: https://aka.ms/vs2019-and-jdk-11

<https://aka.ms/vs2019-and-jdk-11> provides more details.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -178,7 +178,7 @@ In this message, the phrase "should not be reached" means that this error messag
     <comment>The abbreviation "JDK" should not be translated.</comment>
   </data>
   <data name="XA0031" xml:space="preserve">
-    <value>Java SDK {0} or above is required when using {1}.</value>
+    <value>Java SDK {0} or above is required when using {1}.  Note: the Android Designer is incompatible with Java SDK 11.0: https://aka.ms/vs2019-and-jdk-11</value>
     <comment>
 {0} - The Java SDK version number
 {1} - The feature being used: `$(TargetFrameworkVersion) v10.0`</comment>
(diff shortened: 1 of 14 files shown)


src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -178,7 +178,7 @@ In this message, the phrase "should not be reached" means that this error messag
     <comment>The abbreviation "JDK" should not be translated.</comment>
   </data>
   <data name="XA0031" xml:space="preserve">
-    <value>Java SDK {0} or above is required when using {1}.</value>
+    <value>Java SDK {0} or above is required when using {1}.  Note: the Android Designer is incompatible with JDK-11: https://aka.ms/vs2019-and-jdk-11</value>
```
jonathanpeppers, line 181: The message before says `Java SDK 11.0`, should we use the same term in both sentences?
```suggestion
    <value>Java SDK {0} or above is required when using {1}.  Note: the Android Designer is incompatible with Java SDK 11.0: https://aka.ms/vs2019-and-jdk-11</value>
```

jpobst (approved):

PR #6253, opened 2021-09-01
[tests] Fix TargetFramework8_1_Requires_1_8_0

jonpryor:
Commit 658db25e inadvertently broke the
`ValidateJavaVersionTests.TargetFramework8_1_Requires_1_8_0()`
and `ValidateJavaVersionTests.BuildTools27_Requires_1_8_0()` tests,
by updating the XA0031 error message text.  Because these tests use
string.operator==() to compare the string contents, updating the
XA0031 message text broke the tests.

(This also means that the tests will fail if/when localizations are
ever actually used… 🤔)

Update these tests to instead use `string.StartsWith()`, as the *end*
of the XA0031 message was changed, not the beginning.


jonathanpeppers (commented):
This looks good if `TargetFramework8_1_Requires_1_8_0` and `BuildTools27_Requires_1_8_0` both pass.

PR #6264, opened 2021-09-03
[Mono.Android] make UnconditionalSuppressMessageAttribute internal

jonathanpeppers:
We have a copy of a few .NET 5.0 linker-related attributes for
`monoandroid10` and `netstandard2.0` usage of `Mono.Android.dll`.

In 8eb10dd4, I added a new one, but forgot to make it `internal` like
the rest. This could be problematic, as we don't want to expose it as
a public API.

### DIFF
--- a/src-ThirdParty/System.Runtime.CompilerServices/UnconditionalSuppressMessageAttribute.cs
+++ b/src-ThirdParty/System.Runtime.CompilerServices/UnconditionalSuppressMessageAttribute.cs
@@ -15,12 +15,7 @@ namespace System.Diagnostics.CodeAnalysis
     /// <see cref="ConditionalAttribute"/>. So it is always preserved in the compiled assembly.
     /// </remarks>
     [AttributeUsage(AttributeTargets.All, Inherited = false, AllowMultiple = true)]
-#if SYSTEM_PRIVATE_CORELIB
-    public
-#else
-    internal
-#endif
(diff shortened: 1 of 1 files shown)


jonpryor:
> This could be problematic, as we don't want to expose it as a public API

We *don't* expose it as public API.  It's only public if `SYSTEM_PRIVATE_CORELIB` is defined, and nothing defines it within our build system.  Therefore, it's `internal`.

Which just raises a different point: on the one hand, this change *is* more explicit, and makes it more obvious that the type isn't public, without requiring a `git grep SYSTEM_PRIVATE_CORELIB` to verify.

On the other hand, this change introduces changes with upstream.

I'm inclined to go with it, as we're unlikely to ever need to touch this file again.  My main point here is that commit 8eb10dd *did not* introduce new public API, and this PR *isn't necessary*.  It's a *good idea*, but not *necessary*.

jonathanpeppers:
It seems like the other attributes had `#if SYSTEM_PRIVATE_CORELIB` too, and we removed it?

https://github.com/dotnet/runtime/blob/ceb182461998ab51bf4f3205863855428c9f646c/src/libraries/System.Private.CoreLib/src/System/Diagnostics/CodeAnalysis/DynamicDependencyAttribute.cs#L18-L22

Or maybe it wasn't there when we copied the source.

PR #6273, opened 2021-09-08
Special Char Unit Test [WIP]

dellis1972:


jonpryor:
I *was* going to comment that the test works, so seemingly there isn't a problem, right?

…but then I thought about it: `tests/MSBuildDeviceIntegration` only appears to be run on macOS!

https://github.com/xamarin/xamarin-android/blob/fe8c8514fa59446e49eba6b52fd23747ca7c131f/build-tools/automation/azure-pipelines.yaml#L901-L906

https://github.com/xamarin/xamarin-android/blob/fe8c8514fa59446e49eba6b52fd23747ca7c131f/build-tools/automation/azure-pipelines.yaml#L1030-L1048

I don't think *any* of `tests/MSBuildDeviceIntegration` is run on Windows, is it?

Which means we have *no idea* if non-ASCII paths work on Windows at all, at least not via CI, right?

dellis1972:
I'll see about adding the SmokeTest run of MSBuildDeviceIntegration to the windows smoke test run

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
```diff
@@ -830,6 +830,7 @@ public class Foo {
 					rawToDelete,
 				},
 			};
+			libProj.SetProperty ("Deterministic", "true");
```
dellis1972, line 833: @jonathanpeppers I'm not sure why I've had to add this to this particular test. It was failing without it.

dellis1972:
Turns out @jonathanpeppers found AndroidStudio doesn't even let you create a project with these special characters. So I guess it doesn't work at all in Android Studio.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -59,6 +59,28 @@ public void BuildBasicApplication ([ValueSource (nameof (SupportedTargetFramewor
 			}
 		}
 
+		[Test]
+		[Category ("SmokeTests")]
+		public void SmokeTestBuildWithSpecialCharacters ([Values (false, true)] bool forms)
+		{
+			var testName = "テスト";
```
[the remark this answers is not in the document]
dellis1972: @jonpryor  ^ Given this ... should we even bother?

src/Xamarin.Android.Build.Tasks/Tasks/ResolveLibraryProjectImports.cs
```diff
@@ -422,6 +424,16 @@ public override bool RunTask ()
 			}
 		}
 
+		void CreateResourceArchive (string resDir, string outputFile)
```
jonathanpeppers, line 427: If we go with this, I think I'd want to know the build-time impace of creating `res.zip` then if `aapt2` command are any slower using the `.zip` instead of a directory that is already extracted.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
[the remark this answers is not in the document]
jonpryor: @dellis1972: i think that, to the extent that we *can* bother, we *should* bother.

Rephrased: what can/should be done to improve UX and clarity over what is supported?  Android Studio is, up-front, asserting that only ASCII is supported, and will prevent project creation.  This isn't "ideal", but at least is *up-front* and *clear*.

Visual Studio has no such current check.

Thus, for clarity, we have (at least?) two options:

 1. Update Visual Studio to likewise prevent non-ASCII paths.  *Note*: this might anger certain (important) groups. ;-)
 2. Update *and test* our toolchain so that *either* (a) it works, even though Google doesn't support it, or (b) it doesn't work, and we know specifically *why* it doesn't work, so that we can track it into the future.

Results of (2) may dictate (1).

Yes, even if we get this working, Google may "just break" us in the future, but:

 1. They might not break us!
 2. At least we tried.
 3. In trying, we can provide specific feedback to Google around what breaks/needs fixing.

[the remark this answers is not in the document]
grendello: android_studio_non_ascii

On Unix the error turns into a warning and I can create the project, so I still think it's worth to support this scenario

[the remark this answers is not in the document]
jonathanpeppers: I think as long as we can do it and not regress build times for everyone.

src/Xamarin.Android.Build.Tasks/Tasks/ResolveLibraryProjectImports.cs
dellis1972: This is interesting. 

### Old

```
2118 ms  ResolveLibraryProjectImports               1 calls
```

### Res.Zip

```
2387 ms  ResolveLibraryProjectImports               1 calls
```

So it is slightly slower to generate the `res.zip` file. This is on an AndroidX based project so its generating about 15 of those files. However this is not the full story. Lets take a look at what happens with the `aapt2 compile` calls.

### Old

```
581 ms  Aapt2Compile                               1 calls
```

### Res.Zip

```
366 ms  Aapt2Compile                               1 calls
```

It seems that not traversing the directory actually speeds up the `aapt2 compile` slightly 🤷

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Compile.cs
```diff
@@ -38,7 +38,7 @@ public class Aapt2Compile : Aapt2 {
 		protected override int GetRequiredDaemonInstances ()
 		{
 			return Math.Min ((ResourcesToCompile ?? ResourceDirectories).Length, DaemonMaxInstanceCount);
```
jonpryor, line 40: @dellis1972 : considering c626679763280937bac2dbd724d56f40fb11fbd8, shouldn't this have a check so that `GetRequiredDaemonInstances()` always returns at least 1?

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Compile.cs
```diff
@@ -65,9 +66,13 @@ void ProcessDirectory (ITaskItem item, object lockObject)
 			string fileOrDirectory = item.GetMetadata ("ResourceDirectory");
 			if (string.IsNullOrEmpty (fileOrDirectory) || !isDirectory)
 				fileOrDirectory = item.ItemSpec;
+			if (isDirectory && !Directory.Exists (fileOrDirectory)) {
+				LogWarning ($"Ignoring '{fileOrDirectory}' as it does not exist!");
```
jonpryor, line 70: Should this say `Ignoring directory …`, so that it's more explicit that it's a directory which doesn't exist, not a file?

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2Compile.cs
```diff
@@ -111,7 +122,7 @@ protected string[] GenerateCommandLineCommands (string fileOrDirectory, bool isD
 				cmd.Add (GetFullPath (ResourceSymbolsTextFile));
 			}
 			if (isDirectory) {
-				cmd.Add ("--dir");
+				cmd.Add (isArchive ? "--zip" : "--dir");
 				cmd.Add (GetFullPath (fileOrDirectory).TrimEnd ('\\'));
```
jonpryor, line 126: Considering that this PR is in the context of "Chinese character set support", and that this is a command-line argument to `aapt2`, should this *really* use `GetFullPath(fileOrDirectory)` at all anymore?

I'm kinda surprised that this works at all, in that the PRs are green…

src/Xamarin.Android.Build.Tasks/Tasks/AndroidApkSigner.cs
```diff
@@ -102,7 +102,7 @@ protected override string GenerateCommandLineCommands ()
 			if (!string.IsNullOrEmpty (AdditionalArguments))
 				cmd.AppendSwitch (AdditionalArguments);
 
-			cmd.AppendSwitchIfNotNull (" ", Path.GetFullPath (ApkToSign));
+			cmd.AppendSwitchIfNotNull (" ", ApkToSign);
```
jonpryor, line 105: …especially considering that we're removing `Path.GetFullPath()` *here*…

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidUpdateResourcesTest.cs
jonpryor: Frankly, we should use `$(Deterministic)`=true *everywhere*…

src/Xamarin.Android.Build.Tasks/Xamarin.Android.EmbeddedResource.targets
```diff
@@ -56,6 +56,10 @@ This file is used by all project types, including binding projects.
       <Output TaskParameter="ResolvedEnvironmentFiles" ItemName="LibraryEnvironments" />
       <Output TaskParameter="ResolvedResourceDirectoryStamps" ItemName="_LibraryResourceDirectoryStamps" />
     </ReadLibraryProjectImportsCache>
+    <ItemGroup>
+      <FileWrites Include="@(ResolvedResourceDirectories->'$(Identity)..\res.zip')"
```
jonpryor, line 60: Doesn't this assume/require that `%(Identity)` end with `\`?  What code *ensures* that's true?

As far as I can easily tell, `%(Identity)` will *never* end with `\`/`/`, as it appears to come from: https://github.com/xamarin/xamarin-android/blob/709be67be32b6734f74fafdb9c7b6c00421b1dee/src/Xamarin.Android.Build.Tasks/Tasks/ResolveLibraryProjectImports.cs#L195-L200

…and `Path.GetFullPath(dir)` *never* appends the directory separator char, unless `dir` already ends with it…

src/Xamarin.Android.Build.Tasks/Tasks/ResolveLibraryProjectImports.cs
```diff
@@ -406,13 +416,15 @@ public override bool RunTask ()
...
 				if (Directory.Exists (resDir)) {
+					CreateResourceArchive (resDir, resDirArchive);
 					var skipProcessing = aarFile.GetMetadata (AndroidSkipResourceProcessing);
 					if (string.IsNullOrEmpty (skipProcessing)) {
 						skipProcessing = "True";
 					}
 					resolvedResourceDirectories.Add (new TaskItem (Path.GetFullPath (resDir), new Dictionary<string, string> {
 						{ OriginalFile, aarFullPath },
 						{ AndroidSkipResourceProcessing, skipProcessing },
+						{ ResourceDirectoryArchive, Path.GetFullPath (resDirArchive)},
```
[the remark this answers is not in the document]
dellis1972: I tried that. `aapt2` errors because there are files like `classes.jar` etc in there. It only seems to like files which ONLY contain `res` files.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -23,6 +23,27 @@ namespace Xamarin.Android.Build.Tests
...
+
+			var rootPath = Path.Combine (Root, "temp", TestName);
+			var proj = forms ?
+				new XamarinFormsAndroidApplicationProject () :
+				new XamarinAndroidApplicationProject ();
```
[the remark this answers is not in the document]
jonpryor: The problem, as outlined [earlier](https://github.com/xamarin/xamarin-android/pull/6273#issuecomment-915683031), is that the MSBuildDeviceIntegration tests only run on macOS, which doesn't have a "non-ASCII characters" problem in the first place.

What we want/need is for MSBuildDevice integration to run on *Windows*.  The problem is we tried that in an earlier iteration of https://github.com/xamarin/xamarin-android/pull/6297, and hit https://github.com/xamarin/xamarin-android/pull/6297#issuecomment-920309384

> Instead, it (later) insta-fails, because of virtualization:
> 
>      [emulator stdout] emulator: ERROR: x86_64 emulation currently requires hardware acceleration! (TaskId:5)
>      C:\a\_work\1\s\build-tools\scripts\TestApks.targets(65,5): error : Emulator failed to start: emulator: ERROR: x86_64 emulation currently requires hardware acceleration! [C:\a\_work\1\s\tests\Mono.Android-Tests\Mono.Android-Tests.csproj]
>      [emulator stdout] CPU acceleration status: Hyper-V detected and Windows Hypervisor Platform is available. Please ensure both the "Hyper-V" and "Windows Hypervisor Platform" features enabled in "Turn Windows features on or off". (TaskId:5)
>      [emulator stdout] More info on configuring VM acceleration on Windows: (TaskId:5)
>      [emulator stdout] https://developer.android.com/studio/run/emulator-acceleration#vm-windows (TaskId:5)
>      [emulator stdout] General information on acceleration: https://developer.android.com/studio/run/emulator-acceleration. (TaskId:5)

@pjcollins had additional followup: https://github.com/xamarin/xamarin-android/pull/6297#issuecomment-920334650

> * Investigate Helix machine pools for Android emulator testing.
> * Revisit physical Windows machine usage. One possible target would be the shared pool we used to run QA tests on. Some of these machines have physical Android devices attached, but could still potentially be used: …
> * See if we can somehow enable virtualization on our 1ES Windows pool.

*In the meantime*, we're going to need to rely on manual testing for Windows (ugh).

jonathanpeppers (approved):

PR #6278, opened 2021-09-10
Downgrade binutils to 2.35.2

grendello:
Binutils 2.36 introduced [code][0] to deal extra long path names on
Windows which makes compilation (in our case using `as`) and linking
to break if the source/object files reside in a directory which has
special characters (in our case they were `テスト`):

    Xamarin.Android.Common.targets(1887,3): error XA3006: Could not compile native assembly file: typemaps.armeabi-v7a.s [SmokeTestBuildWithSpecialCharactersFalse\テスト\テスト.csproj]
      [aarch64-linux-android-as.EXE stderr] Assembler messages: (TaskId:187)
      [aarch64-linux-android-as.EXE stderr] Fatal error: can't create typemaps.arm64-v8a.o: Invalid argument (TaskId:187)

Downgrade binutils version to 2.35.2 which doesn't contain the offending
code block and will work for us, with relative paths. It will **still**
break with full paths if the path contains special characters, but since
our use case involves only relative paths, this is fine.

[0]: https://sourceware.org/git/?p=binutils-gdb.git;a=blob;f=bfd/bfdio.c;h=463b3879c52ba6beac47190f8eb0810b0c330e65;hb=HEAD#l124

### DIFF
--- a/build-tools/xaprepare/xaprepare/ConfigAndData/Configurables.cs
+++ b/build-tools/xaprepare/xaprepare/ConfigAndData/Configurables.cs
@@ -15,7 +15,7 @@ namespace Xamarin.Android.Prepare
 	//
 	partial class Configurables
 	{
-		const string BinutilsVersion                = "2.37-XA.1";
+		const string BinutilsVersion                = "2.35.2-XA.1";
 
 		const string MicrosoftOpenJDK11Version      = "11.0.10";
 		const string MicrosoftOpenJDK11Release      = "9.1";


jonpryor:
The offending commit appears to be: https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff;h=17d60030ae57fb53f5977fa6928d7d5006f42881

grendello:
Changes will be incorporated into @dellis1972's PR, closing this one

PR #6295, opened 2021-09-14
Bump to cmdline-tools 5.0.

jonpryor:
Testing?

### DIFF
--- a/build-tools/check-boot-times/Program.cs
+++ b/build-tools/check-boot-times/Program.cs
@@ -544,22 +544,22 @@ namespace Xamarin.Android.Tools
 					if (!string.IsNullOrWhiteSpace (sdkPath)) {
 						potentialLocations.AddRange (new []{
 							$"{sdkPath}/platform-tools",
-							$"{sdkPath}/cmdline-tools/4.0/bin",
+							$"{sdkPath}/cmdline-tools/5.0/bin",
 							$"{sdkPath}/cmdline-tools/latest/bin",
 							$"{sdkPath}/emulator",
 						});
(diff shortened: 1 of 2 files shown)


build-tools/check-boot-times/Program.cs
```diff
@@ -544,30 +544,30 @@ static string GetProgramPath (params string [] filenames)
 					if (!string.IsNullOrWhiteSpace (sdkPath)) {
 						potentialLocations.AddRange (new []{
 							$"{sdkPath}/platform-tools",
-							$"{sdkPath}/cmdline-tools/4.0/bin",
+							$"{sdkPath}/cmdline-tools/5.0/bin",
```
grendello, line 547: Maybe this is a good opportunity to use value of the `$(CommandLineToolsVersion)` property here [a'la xaprepare](https://github.com/xamarin/xamarin-android/blob/4364b4ca94759acf637b75f95d62a5384852e69f/build-tools/xaprepare/xaprepare/xaprepare.targets#L36-L41)?

jonpryor:
```markdown
Context: https://dl-ssl.google.com/android/repository/repository2-1.xml
Context: https://developer.android.com/studio/releases/cmdline-tools
Context: https://developer.android.com/studio/command-line#tools-sdk
```

build-tools/check-boot-times/Program.cs
jonpryor: In theory, yes.  In practice, I'm trying to get things ready for branching ASAP. :-)

grendello (approved):

Issue #6302, opened 2021-09-16
Failure to build/deploy/run with android:packageName and <ApplicationId>


mattleibow:
### Steps to Reproduce

1. Create an Android app with a package name in the manifest and the .csproj
2. Observe the build create an .apk and deploy it using the name in the csproj
3. Observe the deploy fail or runtime exception when the app actually has the package name in the android manifest

ID in Manifest: com.microsoft.maui
ID in Project: com.microsoft.maui.sample

![image](https://user-images.githubusercontent.com/1096616/133549034-ed99587f-78e3-4216-8d80-dd34ca6d9e4d.png)

[com.microsoft.maui.sample-Signed.zip](https://github.com/xamarin/xamarin-android/files/7174904/com.microsoft.maui.sample-Signed.zip)

### Expected Behavior
Deploy and run.

### Actual Behavior
Failure to deploy or broken run.

### Log File
[msbuild.zip](https://github.com/xamarin/xamarin-android/files/7174896/msbuild.zip)

dellis1972:
So I guess the first question is which packageId should take precedence ? csproj or AndroidManifest.xml?

jonathanpeppers:
This seems broken on our end:

![image](https://user-images.githubusercontent.com/840039/133677489-feb7bef7-304a-4731-88a9-623df1274663.png)

I have `package="com.companyname.bar"` in the `AndroidManifest.xml`

PR #6304, opened 2021-09-16
[One .NET] fix $(ApplicationId) and //manifest@package at the same time

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/6302

If you try to use both `$(ApplicationId)` and `@package` in
`AndroidManifest.xml` at the same time, you hit errors during
deployment:

    Xamarin.Android.Common.Debugging.targets(611,5): error XA0132: The package was not installed. Please check you do not have it installed under any other user.
    If the package does show up on the device, try manually uninstalling it then try again.
    You should be able to uninstall the app via the Settings app on the device.

From the spec, the `AndroidManifest.xml` is *supposed* to be preferred.

I found the `<GetAndroidPackageName/>` MSBuild task is accidentally
preferring the `$(ApplicationId)` MSBuild property. I reordered the
logic and only use the property when no value is found in the
`AndroidManifest.xml`.

I could reproduce the issue in a test, which now passes.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GetAndroidPackageName.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GetAndroidPackageName.cs
@@ -49,23 +49,23 @@ namespace Xamarin.Android.Tasks
 
 		public override bool RunTask ()
 		{
-			if (!string.IsNullOrEmpty (PackageName)) {
-				PackageName = AndroidAppManifest.CanonicalizePackageName (PackageName);
-			} else if (!string.IsNullOrEmpty (ManifestFile) && File.Exists (ManifestFile)) {
+			if (!string.IsNullOrEmpty (ManifestFile) && File.Exists (ManifestFile)) {
 				using var stream = File.OpenRead (ManifestFile);
 				using var reader = XmlReader.Create (stream);
 				if (reader.MoveToContent () == XmlNodeType.Element) {
 					var package = reader.GetAttribute ("package");
 					if (!string.IsNullOrEmpty (package)) {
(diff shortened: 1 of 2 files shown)


dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tasks/GetAndroidPackageName.cs
```diff
@@ -49,18 +49,20 @@ public class GetAndroidPackageName : AndroidTask
...
-			} else if (!string.IsNullOrEmpty (ManifestFile) && File.Exists (ManifestFile)) {
+			if (!string.IsNullOrEmpty (ManifestFile) && File.Exists (ManifestFile)) {
 				using var stream = File.OpenRead (ManifestFile);
 				using var reader = XmlReader.Create (stream);
 				if (reader.MoveToContent () == XmlNodeType.Element) {
 					var package = reader.GetAttribute ("package");
 					if (!string.IsNullOrEmpty (package)) {
 						package = ManifestDocument.ReplacePlaceholders (ManifestPlaceholders, package);
 						PackageName = AndroidAppManifest.CanonicalizePackageName (package);
+					} else if (!string.IsNullOrEmpty (PackageName)) {
```
jonpryor, line 60: Instead of this `else if` here, and the "duplicate" `PackageName = …` statement between the two blocks, why not *remove* this `else if`, and turn the `else if` on line 64 into an `if`?

jonathanpeppers: I moved it around, should be better now. I removed a `CanonicalizePackageName()` call as well, so it isn't called twice.

PR #6305, opened 2021-09-17
[Xamarin.Android.Build.Tests] Fix GetMaxInstalledPlatform threading.

dellis1972:
When we have unit tests running in parallel we sometimes get issues
where the method `GetMaxInstalledPlatform` returns a lower API level
than the max installed. This is because multiple tests are calling
the method at the same time. As a result they overwrite each other
as they are both scanning the directories. The results in some tests
getting a lower API level than expected becuase the `maxInstalled`
field has a value.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
@@ -117,11 +117,11 @@ namespace Xamarin.ProjectTools
 		}
 
 		static int? maxInstalled;
-		static object lockObject = new object ();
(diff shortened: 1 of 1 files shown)


jonathanpeppers (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/AndroidSdkResolver.cs
```diff
@@ -117,8 +117,16 @@ static string RunPathsTargets (string target)
 		}
 
 		static int? maxInstalled;
+		static object lockObject = new object ();
```
jonpryor, line 120: Should this have a name which better describes the "scope" that it's protecting?  Perhaps `maxInstalledLock`?

PR #6306, opened 2021-09-17
[Xamarin.Android.Build.Tasks] introduce @(AndroidResourceId)

jonathanpeppers:
We currently have an issue where a class library:

1. Includes AndroidX & Google Material
2. Includes at least one `@(AndroidResource)` and uses the ID from C#.
3. `Resource.designer.cs` has 2,700+ fields.

2,700 fields is not a huge problem for a single class library. But
then the problem compounds:

1. Library A has 2,700 fields. It is a dependency of many other
   libraries.
2. Library B, C, and D have a dependency on A, and have at least one
   `@(AndroidResource)` that is used from C#.
3. Application A uses Library A, B, C, and D. Application A sets 2,700
   X 4 fields at startup -- over 10,000 fields.

We saw a case where a .NET MAUI app with SkiaSharp and Telerik set
90,000 fields at startup.

The solution here isn't easy. We can't just set
`$(AndroidGenerateResourceDesigner)` to `false` because the values are
used in C#. What we need is a way for "advanced users" to list which
fields their class library uses.

I propose:

    <AndroidResourceId Include="Resource.Drawable.foo" />
    <AndroidResourceId Include="Resource.Id.bar" />

If `@(AndroidResourceId)` is blank, you get the existing behavior.

In these names you should be able to omit `Resource.` if you like,
just to make things less error-prone.

After this change goes in, I would update SkiaSharp and .NET MAUI to
opt into this behavior.


jonathanpeppers:
@dellis1972 and @jonpryor if you're generally OK with this idea, I need to write docs for the new item group.

src/Xamarin.Android.Build.Tasks/Utilities/ResourceParser.cs
```diff
@@ -48,10 +50,21 @@ internal string GetResourceName (string type, string name, Dictionary<string, st
 				return ResourceIdentifier.CreateValidIdentifier (mappedValue.Substring (mappedValue.LastIndexOf (Path.DirectorySeparatorChar) + 1));
 			}
 
-			Log.LogDebugMessage ("  - Not remapping resource: {0}.{1}", type, name);
-
```
jonathanpeppers, line 52: I saw this message ~2700 times in a build log. It is still logging the *other* case, which I think is sufficient.

jonpryor:
Is this the right solution though?  It's not particularly "scalable", especially if you need to reference dozens of ids.

Didn't one of us have a "rando idea" to rework how resources are bound, turning them into "assembly"-scoped constructs without duplication?

The problem is that Libraries B, C, D reference Library A, and thus get *copies* of *all* resource ids from A.  What if that *weren't* the case?  What if B's `Resource` class *only* contained resources declared in B, and C's `Resource` class only contained resources declared in C, etc.?  Remove the duplication!

The downside with this is that it would be a semantic break for existing code which just assumes that their `Resource` type contains everything from all referenced assemblies, so this would need to be opt-in, but if this were done, wouldn't this be a "more reasonable" solution to this problem?

jonathanpeppers:
Closing this in favor of a simpler workaround:

https://github.com/jonathanpeppers/CustomResourceDesigner

PR #6309, opened 2021-09-17
[build] Add standalone VS2022 support to vsix

pjcollins:
Context: https://docs.microsoft.com/en-us/visualstudio/extensibility/migration/update-visual-studio-extension?view=vs-2022#extensions-without-running-code

Our .vsix installer can't be installed "on top" of VS2022.  This is
something we'll want to continue to support for testing purposes, and
for our OSS installer.  Adding an `amd64` installation target and
updating VS SDK and Tools package reference versions should fix this.

The workaround from commit 4799ea2 has been removed as it no
longer appears to be needed.

### DIFF
--- a/build-tools/create-vsix/create-vsix.csproj
+++ b/build-tools/create-vsix/create-vsix.csproj
@@ -17,7 +17,6 @@
   </PropertyGroup>
 
   <Import Project="..\..\Configuration.props" />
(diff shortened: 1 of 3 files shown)


pjcollins:
Tested with a Windows built OSS installer, before:
VSIXInstaller_h1FHTk0FcJ

After:
image

I'll want to test install/uninstall for an official package still.

pjcollins:
It looks like the unsigned official package from this build is working, I was able to install it into and then revert it from VS 16.11 and VS 17 Int Preview.

dellis1972 (approved):

PR #6311, opened 2021-09-20
Embedded assemblies blob

grendello:
Let's see what breaks


src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -641,7 +698,7 @@ private void AddNativeLibraries (ArchiveFileList files, string [] supportedAbis)
 
 			NdkTools? ndk = NdkTools.Create (AndroidNdkDirectory, Log);
 			if (ndk == null) {
-				return; // NdkTools.Create will log appropriate error
+				return; // NdkTools.Create will Log appropriate error
```
jonpryor, line 701: not sure why this line changed?

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -323,7 +361,9 @@ void AddEnvironment ()
 					JniAddNativeMethodRegistrationAttributePresent = appConfState != null ? appConfState.JniAddNativeMethodRegistrationAttributePresent : false,
 					HaveRuntimeConfigBlob = haveRuntimeConfigBlob,
 					NumberOfAssembliesInApk = assemblyCount,
-					BundledAssemblyNameWidth = assemblyNameWidth + 1,
+					BundledAssemblyNameWidth = assemblyNameWidth,
+					NumberOfAssemblyBlobsInApk = 2, // Until feature APKs are a thing, we're going to have just two blobs in each app
```
jonpryor, line 365: …which implies two blobs.  What are the blobs?  Or is this related to the previous "blob ID" comment? https://github.com/xamarin/xamarin-android/pull/6311/files#diff-9b3577c08294fc97880076f163cbbde1e3fb46a0d20dde1a89cbf94ab932a1f7R347

> The very first blob (ID 0) **must** contain an index of all the assemblies included in the application

At this point in my review, I haven't seen anything about the *structure* of the blobs, and how it relates to the structure of `.aab` and `.apk` files…

jonpryor: Related: we need an "overview" of how this works and how all the pieces fit together, either in the commit message or in a `README.md`.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
grendello: Search and replace I bet

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
grendello: The structure is described [here](https://github.com/xamarin/xamarin-android/pull/6311/files#diff-920537864b4d4fbcacfad5ebfdcfd73470855176b83f4c4592d604a7057b7d3dR111-R131) and I'm going to add a doc once everything is certain to be correct (which it should be now, hopefully). I'm working on adding blob support to some utilities

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -168,6 +182,7 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
 			proj.SetProperty (KnownProperties.TargetFrameworkVersion, "v5.1");
 			proj.SetAndroidSupportedAbis (supportedAbis);
 			proj.SetProperty ("EnableLLVM", enableLLVM.ToString ());
+			proj.SetProperty ("AndroidUseAssembliesBlob", usesAssemblyBlobs.ToString ());
```
jonpryor, line 185: I think we want a "better" name than "blob".  I think I previously suggested "AssemblyContainer", though perhaps "Archive" could be a shorter replacement?

jonpryor, line 185: I think we want a "better" name than "blob".  I think I previously suggested "AssemblyContainer", though perhaps "Archive" could be a shorter replacement?

jonpryor: …though "Archive" might be too easily confused with the same noun used in the IDEs: https://devblogs.microsoft.com/xamarin/android-archiving-and-publishing-made-easy/

Naming things is hard. :-/

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
```diff
@@ -158,37 +160,47 @@ static ApplicationConfig ReadApplicationConfig (string envFile)
 						ret.have_runtime_config_blob = ConvertFieldToBool ("have_runtime_config_blob", envFile, i, field [1]);
 						break;
 
-					case 8: // bound_stream_io_exception_type: byte / .byte
+					case 8: // have_assemblies_blob: bool / .byte
```
jonpryor, line 163: Given that every time we update `libxamarin-app.so` this `switch` block is updated, and some (many?) of the `case` labels changed… is there a more "dummy proof" data structure that can be used here?  A `List<Action>` or something?

src/Xamarin.Android.Build.Tasks/Utilities/AssemblyBlob.cs
```diff
@@ -0,0 +1,391 @@
...
+namespace Xamarin.Android.Tasks
+{
+	abstract class AssemblyBlob
+	{
+		// The two constants below must match their counterparts in src/monodroid/jni/xamarin-app.hh
+		const uint BlobMagic = 0x41424158; // 'XABA', little-endian, must match the BUNDLED_ASSEMBLIES_BLOB_MAGIC native constant
+		const uint BlobVersion = 1; // Must match the BUNDLED_ASSEMBLIES_BLOB_VERSION native constant
+
+		// MUST be equal to the size of the BlobBundledAssembly struct in src/monodroid/jni/xamarin-app.hh
+		const uint BlobBundledAssemblyNativeStructSize = 6 * sizeof (uint);
```
jonpryor, line 19: (Insert screaming for [`create-native-map`](https://github.com/mono/mono-tools/blob/main/create-native-map/man/create-native-map.1) here…. Pity that's not packaged anymore.)

jonpryor: I suppose a more "in-tree" solution to ensuring that things are consistent would be to have `Xamarin.Android.Build.Tasks.csproj` have a `@(ProjectReference)` on `src/monodroid/monodroid.csproj`, and have `monodroid.csproj` write out a file containing the sizes of these structures?  Or vice-versa, have `monodroid.csproj` reference `Xamarin.Android.Build.Tasks.csproj`, and have XABT emit a `.h` file as part of its build?

src/monodroid/jni/shared-constants.hh
```diff
@@ -39,6 +43,18 @@ namespace xamarin::android::internal
 		static constexpr char MONO_SGEN_SO[]      = "monosgen-2.0";
 		static constexpr char MONO_SGEN_ARCH_SO[] = "monosgen-" __BITNESS__ "-2.0";
 #endif
+
+#if __arm__
+		static constexpr char android_abi[] = "armeabi_v7a";
```
jonpryor, line 48: Why can't this use `-`, so that `armeabi-v7a` is valid?

src/monodroid/jni/xamarin-app.hh
```diff
@@ -106,6 +108,86 @@ struct XamarinAndroidBundledAssembly final
 	char    *name;
 };
 
+//
+// Blob format
```
jonpryor, line 112: Format description could also use a "fleshed out" example, e.g. what structs, struct order, and struct values would one expect for `App.dll` & `mscorlib.dll`?  Where do satellite assemblies fit in?  Or do they just "fall out" because their "relative pathname" is hashed, and it's the hash that we care about?

Are there any checks for hash collisions, and an "appropriate error" when a collision is detected?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -264,13 +280,12 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
 					Assert.IsTrue (File.Exists (assemblies), "{0} libaot-UnnamedProject.dll.so does not exist", abi);
 					var apk = Path.Combine (Root, b.ProjectDirectory,
 						proj.OutputPath, $"{proj.PackageName}-Signed.apk");
+					var helper = new ArchiveAssemblyHelper (apk, usesAssemblyBlobs);
+					Assert.IsFalse (helper.Exists ("assemblies/UnnamedProject.dll"), $"UnnamedProject.dll should not be in the {proj.PackageName}-Signed.apk");
```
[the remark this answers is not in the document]
grendello: The difference is handled and hidden by `ArchiveAssemblyHelper`.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
grendello: Why complicate/obfuscate matters if they can remain simple? The `switch` is self-documenting and clear in its purpose. Adding a level of indirection wouldn't give us any advantage IMO except for a smaller occasional diff.

src/monodroid/jni/shared-constants.hh
grendello: Because `_` is what Android uses when generating split APKs: `split_config.armeabi_v7a.apk` is what they generate and we check that name against the list of APKs passed to runtime init

src/monodroid/jni/xamarin-app.hh
grendello: > Format description could also use a "fleshed out" example, e.g. what structs, struct order, and struct values would one expect for `App.dll` & `mscorlib.dll`? Where do satellite assemblies fit in? Or do they just "fall out" because their "relative pathname" is hashed, and it's the hash that we care about?
> 
> Are there any checks for hash collisions, and an "appropriate error" when a collision is detected?

I was focused on implementing the thing first instead of writing and rewriting documents... Once I'm sure that the format is final, I can spend time on documentation

grendello: > Format description could also use a "fleshed out" example, e.g. what structs, struct order, and struct values would one expect for `App.dll` & `mscorlib.dll`? Where do satellite assemblies fit in? Or do they just "fall out" because their "relative pathname" is hashed, and it's the hash that we care about?
> 
> Are there any checks for hash collisions, and an "appropriate error" when a collision is detected?

Yes there are, [here](https://github.com/xamarin/xamarin-android/pull/6311/files#diff-1b9c459e1ad04c80fa943403a2facf8fb0f726e5b086a004f59d731e3cb80598R124-R134) and [here](https://github.com/xamarin/xamarin-android/pull/6311/files#diff-1b9c459e1ad04c80fa943403a2facf8fb0f726e5b086a004f59d731e3cb80598R168-R177). This code generates the **global** index (contained only in a single blob, the blob with ID 0) spanning all of the blobs that are generated for the application

Documentation/project-docs/AssemblyBlobs.md
```diff
@@ -0,0 +1,214 @@
...
+    if `debug_data_offset` is `0`
+  - `config_data_offset`: offset of the assembly's config file data
+    from the  beginning of the blob file. A value of `0` indicates
+    there's no config file data for this assembly.
+  - `config_data_size`: number of bytes of config file data. Can be
+    `0` only if `config_data_offset` is `0`
+
+## Index blob
+
+Each application will contain exactly one blob with a global index -
```
jonpryor, line 162: This concerns me; https://github.com/xamarin/xamarin-android/pull/6311/files#r725237133

Documentation/project-docs/AssemblyBlobs.md
```diff
@@ -0,0 +1,214 @@
...
+### Hash table format
+
+Both tables share the same format, despite the hashes themselves being
+of different sizes.  This is done to make handling of the tables
+easier on the runtime.
+
+Each entry contains, among other fields, the assembly name hash.  The
+hash value is obtained using the
+[xxHash](https://cyan4973.github.io/xxHash/) algorithm and is
+calculated **without** including the `.dll` extension.  This is done
```
jonpryor, line 187: I would like an example that pertains to *satellite* assemblies: how/where is the path used?  Consider `es/libZipSharp.resources.dll` vs. `fr/libZipSharp.resources.dll`.

Documentation/project-docs/AssemblyBlobs.md
```diff
@@ -0,0 +1,214 @@
...
+        uint32_t mapping_index;
+        uint32_t local_blob_index;
+        uint32_t blob_id;
+    };
+
+Individual fields have the following meanings:
+
+ - `hash64`/`hash32`: the 32-bit or 64-bit hash of the assembly's name
+   **without** the `.dll` suffix
+ - `mapping_index`: index into a compile-time generated array of
```
jonpryor, line 209: I'm not sure what this means, but strongly implies that the App needs to "know about" Features.  This worries me?

Documentation/project-docs/AssemblyBlobs.md
```diff
@@ -0,0 +1,214 @@
...
+        uint32_t blob_id;
+    ;
+
+Individual fields have the following meanings:
+
+ - `magic`: has the value of 0x41424158 (`XABA`)
+ - `version`: a value increased every time blob format changes.
+ - `local_entry_count`: number of assemblies stored in this blob (also
+   the number of entries in the assembly descriptor table, see below)
+ - `global_entry_count`: number of entries in the index blob's (see
```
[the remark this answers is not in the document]
grendello: > I don't understand.
> 
> I think we need more nouns. :-/
> 
> Thus, a suggestion/clarification:
> 
> There are the following blob _files_:
> 
>     * App Architecture-Independent Assembly [& related] Blob (`assemblies.blob` in main app)
> 
>     * App Architecture-Specific Assembly Blobs (`assemblies.[ARCH].blob` in main app)
> 
>     * Feature Architecture-Independent Assembly [& related] Blob (zero or more) (`assemblies.blob` in feature)
> 
>     * Feature Architecture-Specific Assembly Blobs (zero or more) (`assemblies.[ARCH].blob` in feature)
> 
> 
> _Each_ of those blob files has an internal structure.

Which is identical for all the blobs, with the only difference being the global hash index which is **not** part of the general blob header.

> 
> We thus return to `BundledAssemblyBlobHeader::global_entry_count`: _what is it_, and for _where_?

A count of all the entries in all the blobs in the application together - a **global** count of assemblies, versus a **local** count in a single blob file.

> 
> For the App Architecture-Independent Blob, is it the count of _all_ of the above (all App + Feature blobs)? Just the "current" blob". Just the _App_ blobs? Other?

It's a global count for the entire app - all the blobs summed up. It's essentially the size of the pre-generated table in `libxamarin-app.so`

> 
> The "all the other blobs store `0` in this field" sentence fragment (below) is not helping.

Not sure what you mean, it's a format specification and it clearly states that this field is 0 for all the blobs except the index one.

> 
> I _imagine_ that it may be _possible_ to update a Feature without updating the corresponding App -- and, if supported, would be _beneficial_, as it means less data needs to be sent to devices -- so I _hope_ that the App Arch-Independent Blob's `global_entry_count` _doesn't_ include counts for _all related Feature blobs_.

I doubt it will be possible to update a Feature without updating the app.  With the AAB format being the required distribution format for the Play Store, I don't think it's possible to update the app piecemeal - how would it be done? By uploading a "mini-AAB" with just the feature update in it? I rather think the apps will be uploaded as a single AAB from which feature APKs will be created by Google, on the Play Store side. So the app will be rebuilt entirely and there's no issue with this global count.

Documentation/project-docs/AssemblyBlobs.md
grendello: I don't think it's going to be a problem

Documentation/project-docs/AssemblyBlobs.md
grendello: The path is treated as part of the name, and we reconstruct the name with culture [here](https://github.com/xamarin/xamarin-android/pull/6311/files#diff-3157a46af8a1b1c13c7d1ad16bcfe787f40a0a07e3e7251282500e2a8af5a7f2R475-R483), using the result to calculate the hash at the runtime that matches the one calculated on the build time.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/ArchiveAssemblyHelper.cs
```diff
@@ -0,0 +1,299 @@
...
+			".dll",
+			".config",
+			".pdb",
+			".mdb",
+		};
+
+		static readonly Dictionary<string, string> ArchToAbi = new Dictionary<string, string> (StringComparer.OrdinalIgnoreCase) {
+			{"x86", "x86"},
+			{"x86_64", "x86_64"},
+			{"armeabi_v7a", "armeabi-v7a"},
```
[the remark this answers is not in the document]
jonpryor: @grendello [wrote](https://github.com/xamarin/xamarin-android/pull/6311#discussion_r724038421):

> it's a mapping between "our" ABI names and Android runtime ABI names

Which still confuses me: where do "our" ABI names use `_`?

```
% git grep armeabi_v7
# no matches

% git grep armeabi-v7 | wc -l
     468
```

*Our* ***existing*** ABI names use `-`: armeabi-v7a, arm64-v8a, etc.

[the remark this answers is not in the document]
grendello: Our blobs use the same convention as Google, see [here](https://github.com/xamarin/xamarin-android/pull/6311/files#diff-31c02ad43b39cb7cc279860f5484429fb84075670ed3ea251978ad688ccfe172R105-R106).

jonathanpeppers (commented):
The designer tests fail with:
```
[2021-10-18 11:53:24.4] Renderer (error) >> ../../../jni/embedded-assemblies.hh:155 (ensure_valid_assembly_stores): Invalid or incomplete assembly store data
[2021-10-18 11:53:24.4] ERROR: : Layout inflation failed
Xamarin.AndroidDesigner.Java.RendererErrorMessageException: Connection is closed
  at Xamarin.AndroidDesigner.Java.JavaProcessConnection.GetResponse (Xamarin.AndroidDesigner.Java.BinaryMessage message) [0x00102] in <a1fe0e312d9347bea60d3d402d9d0db5>:0 
  at Xamarin.AndroidDesigner.AndroidRenderSession.InflateAsync () [0x0022f] in <a1fe0e312d9347bea60d3d402d9d0db5>:0 
=> Xamarin.AndroidDesigner.Tests.AndroidDesignerSessionCustomControlsTest.AssertCanReloadAfterEmptyBuildOutput
=> Xamarin.AndroidDesigner.Tests.AndroidDesignerSessionCustomControlsTest.AvailableThemesDotNotContainAppCompatBoilerplate
```

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5339839&view=logs&j=afc75dd9-ea98-591e-f4f4-0b6212ccdfde&t=e7c34002-0933-5b75-f765-dd65be44a067&l=784

I don't know if that is helpful at all?

src/monodroid/jni/application_dso_stub.cc
```diff
@@ -80,3 +82,33 @@ XamarinAndroidBundledAssembly bundled_assemblies[] = {
...
+		.config_data = nullptr,
+		.descriptor = nullptr,
+	},
+
+	{
+		.image_data = nullptr,
+		.debug_info_data = nullptr,
+		.config_data = nullptr,
+		.descriptor = nullptr,
+	}
```
jonpryor, line 99: Trailing `,`? :-)

src/Xamarin.Android.Build.Tasks/Tasks/RegisterAssemblyStoreState.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace Xamarin.Android.Tasks
+{
+	public class RegisterAssemblyStoreState : AndroidTask
+	{
+		public const string RegisteredObjectKey = ".:RegisterAssemblyStoreState_Key:.";
+
+		public override string TaskPrefix => "RASS";
+
+		[Required]
+		public bool UseAssemblyStore { get; set; }
```
jonathanpeppers, line 18: I think some of the issues with incremental builds are due to this task and usage of `RegisterTaskObject`.

Can we instead do something like?

1. Delete the `<RegisterAssemblyStoreState/>` task
2. Add `UseAssemblyStore` to `<BuildApk/>`. Don't make it required.
3. Sometime in the future we could update monodroid, but I don't think we even want this value used for Fast Deploy scenarios?

https://github.com/xamarin/monodroid/blob/9b3a37af1102f56ccaf449864f18a7d2a78e9682/tools/msbuild/Xamarin.Android.Common.Debugging.targets#L238-L268

Maybe we don't ever need step 3 if this isn't meant to work for Fast Deployment.

I think that might solve the bugs here and be less code.

grendello: I added this task today to avoid precisely the problem you're describing. `GenerateJavaStubs` registers an app config object which, before I added the above code, had a boolean property telling `BuildApk` whether to use assembly stores. It worked fine but not with `dotnet run` which appears to build the app twice - first the build and second the packaging. In the build phase, app config had `true` for the `UseAssemblyStore` flag, and that value was placed in the generated native assembly code - the runtime would expect to find assembly store blobs in the apk. However, in the packaging phase of `dotnet run` `GenerateJavaStubs` was skipped and so the app config object was never registered and `BuildApk` packaged **assemblies** in the apk instead of the blobs, resulting in the native runtime aborting execution when it could not find the blobs.

velocitysystems:
@grendello For verifying DLLs inside the blob, are there any good CLI tools (already written) to unpack the blob and DLLs?

jonathanpeppers:
This tool should be able to unpack/decompress them:

https://github.com/xamarin/xamarin-android/tree/main/tools/decompress-assemblies

velocitysystems:
Awesome. Thanks @jonathanpeppers!
I see this has a `net472` target. Would there be much work to port this to .NET 6?

jonathanpeppers:
I would just try it and see if it builds.

velocitysystems:
Thanks @jonathanpeppers. Did a quick port to a standalone .NET 6 console project and works perfectly!

PR #6322, opened 2021-09-21
[ci] Make logging less verbose

jpobst:
Context: https://twitter.com/KirillOsenkov/status/1440192351528374279

Our verbose logging is getting a little out of control.  The `make jenkins` log is ~870 MB and can no longer be viewed in the AzDO UI.

Remove `V=1` from our invocations, which removes `/v:diag`.

Timings aren't stable across build agents so numbers aren't very accurate, but there are definitely improvements:
- Log file of `make jenkins` went from 870 MB to 32 MB
- Build time of `make jenkins` went from 42:49 to 34:43

Full logs are still available in the `.binlog`'s we already produce and place in artifacts.

### DIFF
--- a/build-tools/automation/azure-pipelines-oss.yaml
+++ b/build-tools/automation/azure-pipelines-oss.yaml
@@ -91,26 +91,26 @@ stages:
 
     # Prepare and build everything
     - script: >
-        echo "make prepare-update-mono V=1 CONFIGURATION=$(XA.Build.Configuration) $(PREPARE_FLAGS) MSBUILD_ARGS='$(EXTRA_MSBUILD_ARGS)'" &&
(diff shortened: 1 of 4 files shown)


jonathanpeppers (commented):
It looks like this is in more `.yaml` files, should we updates this one, too?

https://github.com/xamarin/xamarin-android/blob/c3a21d9462ca275a2a47f6500ae3d7c970374b0e/build-tools/automation/azure-pipelines-oss.yaml#L94

pjcollins:
We maybe want to remove this from the JI prepare step as well: https://github.com/xamarin/xamarin-android/blob/6467cda58822b4405a8a105a3268bd5341230187/build-tools/xaprepare/xaprepare/Steps/Step_PrepareExternalJavaInterop.Unix.cs#L24

My only concern with this would be losing potential out on potential information we get from passing `--debug` to various `mono` invocations via `$MONO_OPTIONS`, though at a glance that usage does look pretty sparse so I think it would be fine to lose that as well.

pjcollins (approved):

jonathanpeppers (approved):

Issue #6308, opened 2021-09-17
[One .NET] fix gaps in incremental build performance


jonathanpeppers:
Two factors add a performance cost to .NET 6 builds that are currently not in "legacy" Xamarin.Android.

1. An additional `<MSBuild/>` call per RID evaluates the project again:

https://github.com/xamarin/xamarin-android/blob/d61aae7f120f9c39c2186908fb3cd791b3bde871/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets#L83-L88

2. The `<ProcessAssemblies/>` MSBuild task loads each assembly looking at MVIDs. If it finds assemblies with different MVIDs per architecture, those assemblies are put in sub-directories, like `assemblies/armeabi-v7a/System.Private.CoreLib.dll`.

Both of these together might add 500ms to *all builds* for a decent-sized project. If we fixed these gaps, I think we could get "Hello World" incremental build+deploy performance under 1 second.

We already have code that reduces builds down to a single RID when an emulator is running or device attached. We could potentially skip 1 and 2 if there is a single RID.

jonathanpeppers:
No.2 (`<ProcessAssemblies/>`) is implemented in https://github.com/xamarin/xamarin-android/pull/6323

No.1 (trying to eliminate `<MSBuild/>`) actually does not work:

https://github.com/xamarin/xamarin-android/compare/main...jonathanpeppers:use-calltarget-singlerid

I found that `RuntimeIdentifier` "singular" needs to be passed into the full build pipeline. Targets like `ProcessFrameworkReferences` that run really early during the build expect it to be set.

I think we should come back to the idea in No.1 if the dotnet/sdk adds a concept of multiple-RID builds as mentioned in:

https://github.com/dotnet/sdk/issues/14359

We can close this for .NET 6, I think.

PR #6323, opened 2021-09-21
[One .NET] add a single-RID "fast path" for <ProcessAssemblies/>

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/issues/6308
Context: https://github.com/xamarin/monodroid/commit/2e93b630e47701dae56f426fe5934846d571832b

When a device or emulator is selected during a build with Fast
Deployment enabled, a single `$(RuntimeIdentifier)` is selected. This
gives us a reasonable speedup for the build.

One place we can do more is the `<ProcessAssemblies/>` MSBuild task.
Add a "fast path", where we don't need to read the MVIDs from every
assembly. We can simply loop over the assemblies and set required item
metadata.

I also made various changes to refactor the `<ProcessAssemblies/>` so
there is more shared helper methods within the task, etc.

## Results ##

An incremental build of a `dotnet new android` project with no changes:

    Two RIDs:
    81 ms  ProcessAssemblies                          1 calls
    One RID:
    45 ms  ProcessAssemblies                          1 calls

An incremental build of a `dotnet new maui` project with no changes:

    Two RIDs:
    85 ms  ProcessAssemblies                          1 calls
    One RID:
    54 ms  ProcessAssemblies                          1 calls

This should save 30-40ms on every build, when you are debugging and
have a device or emulator selected.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
@@ -78,9 +78,6 @@ namespace Xamarin.Android.Tasks
 			return !Log.HasLoggedErrors;
 		}
 
-		/// <summary>
-		/// Called when RuntimeIdentifiers has one item
-		/// </summary>
 		void SetMetadataForAssemblies (List<ITaskItem> output, Dictionary<string, ITaskItem> symbols)
 		{
 			foreach (var assembly in InputAssemblies) {
@@ -93,9 +90,6 @@ namespace Xamarin.Android.Tasks
 			}
 		}
 
-		/// <summary>
-		/// Called when RuntimeIdentifiers has more than one item
-		/// </summary>
 		void DeduplicateAssemblies (List<ITaskItem> output, Dictionary<string, ITaskItem> symbols)
 		{
 			// Group by assembly file name
@@ -124,7 +118,7 @@ namespace Xamarin.Android.Tasks
 				// If we end up with more than 1 unique mvid, we need *all* assemblies
 				if (mvids.Count > 1) {
 					foreach (var assembly in group) {
-						var symbol = GetOrCreateSymbolItem(symbols, assembly);
+						var symbol = GetOrCreateSymbolItem (symbols, assembly);
 						SetDestinationSubDirectory (assembly, group.Key, symbol);
 						output.Add (assembly);
 					}
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
@@ -1,3 +1,4 @@
+#nullable enable
 using System;
 using System.Collections.Generic;
 using System.IO;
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
The designer tests keep failing on NuGet restore, I think we can ignore:
```
(Restore target) -> 
  /Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/NuGet.targets(131,5): error : Unable to load the service index for source https://pkgs.dev.azure.com/***/DevDiv/_packaging/Xamarin.UITooling/nuget/v3/index.json. [/Users/runner/work/1/s/UITools/Designer/Xamarin.Designer.Android/Xamarin.AndroidDesigner.sln]
/Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/NuGet.targets(131,5): error :   Response status code does not indicate success: 401 (Unauthorized). [/Users/runner/work/1/s/UITools/Designer/Xamarin.Designer.Android/Xamarin.AndroidDesigner.sln]
```

src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
```diff
@@ -46,82 +50,118 @@ public override bool RunTask ()
...
+				Log.LogDebugMessage ("Found a single RuntimeIdentifier");
+				SetMetadataForAssemblies (output, symbols);
+			}
+
+			OutputAssemblies = output.ToArray ();
+			ResolvedSymbols = symbols.Values.ToArray ();
+
+			// Set ShrunkAssemblies for _RemoveRegisterAttribute and <BuildApk/>
+			if (PublishTrimmed) {
+				ShrunkAssemblies = OutputAssemblies.Select (a => {
```
dellis1972, line 67: Do we want to use Linq here? We usually try to avoid it for performance reasons.

jonathanpeppers: I changed this one to a `foreach` loop.

There is another place in here that uses `.Where(..).GroupBy(...)`, which I think is a good usage of Linq.

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
```diff
@@ -46,82 +50,120 @@ public override bool RunTask ()
...
+					});
+				}
+				ShrunkAssemblies = shrunkAssemblies.ToArray ();
+			}
+
+			return !Log.HasLoggedErrors;
+		}
+
+		/// <summary>
+		/// Called when RuntimeIdentifiers has one item
```
jonpryor, line 82: Aside: not sure we need XML docs for *private* methods in general, never mind a class that doesn't otherwise contain XML documentation at all…

src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
```diff
@@ -46,82 +50,120 @@ public override bool RunTask ()
...
 				if (mvids.Count > 1) {
 					foreach (var assembly in group) {
-						var symbolPath = Path.ChangeExtension (assembly.ItemSpec, ".pdb");
-						if (!symbols.TryGetValue (symbolPath, out var symbol)) {
-							// Sometimes .pdb files are not included in @(ResolvedFileToPublish), so add them if they exist
-							if (File.Exists (symbolPath)) {
-								symbols [symbolPath] = symbol = new TaskItem (symbolPath);
-							}
-						}
+						var symbol = GetOrCreateSymbolItem(symbols, assembly);
```
jonpryor, line 127: Code convention: space ` ` before `(`.

src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs
```diff
@@ -46,82 +50,120 @@ public override bool RunTask ()
...
 					}
 				}
 			}
+		}
 
-			OutputAssemblies = output.ToArray ();
-			ResolvedSymbols = symbols.Values.ToArray ();
+		static bool IsFrameworkAssembly (ITaskItem assembly)
+		{
+			string packageId = assembly.GetMetadata ("NuGetPackageId") ?? "";
```
jonpryor, line 152: Documentation for [`ITaskItem.GetMetadata()`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.framework.itaskitem.getmetadata?view=msbuild-16-netcore) leaves a bit to be desired, but… will this return `null`, or `""`, when the specified metadata isn't found?

It might be "safer" to just use `string.IsNullOrEmpty()`.

PR #6328, opened 2021-09-22
[Xamarin.Android.Build.Tasks] <LinkAssembliesNoShrink/> can skip the main assembly

jonathanpeppers:
We found an incremental build in a `dotnet new android` app with C#
code changes has this duration:

    100 ms  LinkAssembliesNoShrink                     1 calls

It turns out we don't actually need to run `<LinkAssembliesNoShrink/>`
against the "main" assembly in .NET 6. After passing in
`$(TargetName)` and simply copying the file, this time improved to:

      9 ms  LinkAssembliesNoShrink                     1 calls

This change should improve incremental C# changes by ~91ms in `dotnet
new android` projects.

In "legacy" Xamarin.Android, it still seems possible for an
application project to reference a class library with a higher
`$(TargetFrameworkVersion)`. So we cannot apply this optimization for
Xamarin.Android.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
@@ -52,7 +52,7 @@ namespace MonoDroid.Tuner
 				return false;
 
 #if !NET5_LINKER
-			CheckAppDomainUsageUnconditional (assembly, (string msg) => Context.LogMessage (MessageImportance.High, msg));
+			CheckAppDomainUsage (assembly, (string msg) => Context.LogMessage (MessageImportance.High, msg));
 #endif  // !NET5_LINKER
 
 			return assembly.MainModule.HasTypeReference ("Java.Lang.Object");
@@ -86,22 +86,24 @@ namespace MonoDroid.Tuner
 			if (!CheckShouldProcessAssembly (assembly))
 				return;
 
-			if (FixAbstractMethodsUnconditional (assembly)) {
+			if (FixAbstractMethods (assembly)) {
 				Context.SafeReadSymbols (assembly);
 				UpdateAssemblyAction (assembly);
 				MarkAbstractMethodErrorType ();
 			}
 		}
 
-		internal void CheckAppDomainUsage (AssemblyDefinition assembly, Action<string> warn)
+		internal bool FixAbstractMethods (AssemblyDefinition assembly)
 		{
-			if (IsProductOrSdkAssembly (assembly))
-				return;
-
-			CheckAppDomainUsageUnconditional (assembly, warn);
+			bool changed = false;
+			foreach (var type in assembly.MainModule.Types) {
+				if (MightNeedFix (type))
+					changed |= FixAbstractMethods (type);
+			}
+			return changed;
 		}
 
-		void CheckAppDomainUsageUnconditional (AssemblyDefinition assembly, Action<string> warn)
+		internal void CheckAppDomainUsage (AssemblyDefinition assembly, Action<string> warn)
 		{
 			if (!assembly.MainModule.HasTypeReference ("System.AppDomain"))
 				return;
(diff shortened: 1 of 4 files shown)


dellis1972:
I'm not sure this will work. 

One of the things `LinkAssembliesNoShrink` does is to fix up the Abstract methods on the C# side of bound objects. These can be exposed in a number of ways. For example if a user uses a specific type of layout or custom layout in their UI code in the main app, this may result in these abstract methods being fixed up on the C# side in the main app assembly.  These is also code to fix up the KeepAlives on the Java Objects, again this can also happen in the main assembly. 

I don't think we can skip the main app assembly in this case. We'd need to have a unit test which hits this code path to make sure the app still runs if this is going to be implemented. I suspect it would crash when those abstract methods are hit.

jonathanpeppers:
The case of abstract methods, you can't actually make a test case for the changes in this PR:
1. A class library has an older TargetFrameworkVersion using `ICursor`, for example.
2. The main app has a newer TargetFrameworkVersion, and `ICursor` has additional methods in the newer API level.

So we shouldn't ever need to fix abstract methods in the main app?

I need to look at the failing `WarnAboutAppDomains()` test here, maybe we can only skip that part in .NET 6.

jonpryor:
@dellis1972 [said](https://github.com/xamarin/xamarin-android/pull/6328#issuecomment-925616392):

> These is also code to fix up the KeepAlives on the Java Objects, again this can also happen in the main assembly.

`GC.KeepAlive()` insertions (e88cfbcf9ff4f4e179a60f1e8af2bef3537a66b4) is for pre-existing binding assemblies, "pre-existing" meaning "before xamarin/Java.Interop@1f21f38c was merged".  The binding fix is in d16-8/Xamarin.Android 11.1.0.  This should not be a concern for net6.

> One of the things `LinkAssembliesNoShrink` does is to fix up the Abstract methods on the C# side of bound objects.

This is the more important case.  However, it's only truly needed when the *app*'s `$(TargetFrameworkVersion)` is greater than a referenced *library*'s `$(TargetFrameworkVersion)`.  If they're the same, it's not necessary: everything is consistent.  (Not that we can ever expect/rely on everything being the same…)

@jonathanpeppers [replied](https://github.com/xamarin/xamarin-android/pull/6328#issuecomment-925817245):

> So we shouldn't ever need to fix abstract methods in the main app?

Nominally correct.

The *only* way this can fail is if we *allow* the app assembly to use a `$(TargetFrameworkVersion)` *less than* any referenced assembly's `$(TargetFrameworkVersion)`, and in that case I'm not sure that `FixAbstractMethodsStep` will necessarily "save" us (see below).

On the one hand, we *do* emit an [XA0105 warning](https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/messages/xa0105.md) if this happens: https://github.com/xamarin/xamarin-android/blob/cee5e72764ab9fa2f9c78a651707b7efbf8732cc/src/Xamarin.Android.Build.Tasks/Tasks/ResolveAssemblies.cs#L119-L123

However, from a *quick* glance, that's *all* we do: *warn*.  We don't update the `$(TargetFrameworkVersion)` of the app to be that of the highest found `$(TargetFrameworkVersion)` of a referenced assembly, we don't error.

*In a .NET 6 context*, we're fine: .NET 6 only supports the latest API level.

*In a Legacy context*, this is somewhat worrying: what happens when the reference assembly references a type that doesn't exist in the `$(TargetFrameworkVersion)` used by the app?  What happens?  (Also, I *thought* we overrode the App's `$(TargetFrameworkVersion)` value to be >= that of any reference assembly?  But I can't find such code?)

jonpryor:
@jonathanpeppers wrote: https://discord.com/channels/732297728826277939/732297837953679412/892999577362399232

> yeah, this seems to be grabbing the older TFV
>
> .apk contents

Just a tad worrying…. We should probably turn the XA0105 warning into an error.

That would be for a separate PR.

jonathanpeppers:
@jonpryor I updated this, so the optimization only applies to .NET 6.

In other PRs, we can:
1. Make `XA0105` an error
2. Setup localization for `Microsoft.Android.Sdk.ILLink.csproj`, so we can emit `AppDomain` warnings for Release builds.

jonpryor:
@dellis1972: how's this look now?

jonpryor:
Commit message for review:

```markdown
Context: https://github.com/xamarin/xamarin-android/issues/6356
Context: https://github.com/xamarin/xamarin-android/issues/6357

We found an incremental build in a `dotnet new android` app with C#
code changes has this duration:

	100 ms  LinkAssembliesNoShrink                     1 calls

It turns out we don't actually need to run `<LinkAssembliesNoShrink/>`
against the "main" assembly in .NET 6.  After passing in
`$(TargetName)` and simply copying the file, this time improved to:

	  9 ms  LinkAssembliesNoShrink                     1 calls

This change should improve incremental C# changes by ~91ms in
`dotnet new android` projects.

In "Xammie" Xamarin.Android, it is possible for an application project
to reference a class library with a higher `$(TargetFrameworkVersion)`;
see [warning XA0105 docs][0].

We cannot apply this optimization for Xamarin.Android.

However, in thinking about and revisiting this code, we've made two
determinations for future work:

  * XA0105 being a warning is a terrible idea, as it can allow apps
    to crash via `TypeLoadException`.

    See xamarin/xamarin-android#6356.

  * .NET 6 builds don't emit [XA2000 warnings][1] when a referenced
    assembly references `AppDomain.CreateDomain()`.  This isn't
    desirable, as `AppDomain.CreateDomain()` is still part of the
    .NET Standard 2.0 API contract, and thus could still be used,
    even though it won't work.  *Something* should warn if
    `AppDomain.CreateDomain()` is referenced, to reduce the need for
    manual app testing.

    See xamarin/xamarin-android#6357.

[0]: https://github.com/xamarin/xamarin-android/blob/af60da7f13e613faa8ded5ef734e8c6d91445acc/Documentation/guides/messages/xa0105.md
[1]: https://github.com/xamarin/xamarin-android/blob/03d4e40d310255f80cc5ffb6475f03a02a561bc5/Documentation/guides/messages/xa2000.md
```

dellis1972 (approved):

PR #6331, opened 2021-09-23
[ci] Run Mac MSBuild tests on 6 nodes instead of 3.

jpobst:
In order to have sub-2 hour builds, we should endeavor to keep the `Mac Build` stage under an hour, and each individual Test job under an hour.

The Mac MSBuild tests take considerably longer than the Windows tests, breaking the 1 hour limit:

image

This is because they run on the public `Azure Pipelines` Mac images, which are slower than our private Android-specific Windows agents.  The good news is that there are an ~infinite number of them, which means we can simply split our tests across more agents.

However, our Android-specific Windows agents **are** limited in number, so we do not want to split the tests any further for them.  By doubling our Mac agents from 3 to 6, we can simplify the logic to keep Windows at 3 nodes.  Each Windows node runs 2 categories:

- Windows Node 1 runs `[Category ("Node-1")]` and `[Category ("Node-4")]`
- Windows Node 2 runs `[Category ("Node-2")]` and `[Category ("Node-5")]`
- Windows Node 3 runs `[Category ("Node-3")]` and `[Category ("Node-6")]`

The `BuildTest` test fixture had to be split into `BuildTest2` in order to split its test load across 2 agents.

With this PR, each test job is approximately under an hour.  There are some variances based on how long it takes to provision the dependencies that make it impossible to perfectly load balance.  We can tweak the balance in the future if needed.

image

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -954,11 +954,33 @@ stages:
       job_name: mac_msbuild_tests_3
       job_suffix: Legacy
       nunit_categories: '&& cat != SmokeTests'
+      
+  - template: yaml-templates\run-msbuild-mac-tests.yaml
+    parameters:
+      node_id: 4
+      job_name: mac_msbuild_tests_4
+      job_suffix: Legacy
+      nunit_categories: '&& cat != SmokeTests'
+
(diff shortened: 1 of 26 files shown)


jonathanpeppers (approved):
This looks good, like it should help quite a bit.

We need this same treatment on these, though:

image

pjcollins (approved):

PR #6348, opened 2021-09-29
[Xamarin.Android.Build.Tests] Ensure UseDotNet is set before NUnit evaluates test parameters.

jpobst:
We set `Builder.UseDotNet` for `Xamarin.Android.Build.Tests` tests in a static `SetUp ()` constructor in an NUnit `[SetUpFixture]`.  However this is not getting set before it is consumed by NUnit for test parameters.

For example:
```
// BuildTest.cs
public void BuildBasicApplication ([ValueSource (nameof (SupportedTargetFrameworks))] string tfv, [Values (true, false)] bool isRelease)
```

is being evaluated as `UseDotNet=false`, and thus running this test on all TFV's for .NET 6, when it should only be run for the latest TFV.

By changing `Builder.UseDotNet` to an evaluated property-getter, we can ensure that it is always correct for running on .NET 5+.

The net result of this change is that the `MSBuild - One .NET - Node-1` test suites run 24 fewer tests for both Windows and Mac.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/BaseTest.cs
@@ -49,12 +49,6 @@ namespace Xamarin.Android.Build.Tests
 
 			static SetUp ()
 			{
-#if NETCOREAPP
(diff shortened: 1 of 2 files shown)


dellis1972 (approved):

jonathanpeppers (approved):
We get different numbers of skipped tests, so I think this change is working:

image

Issue #6271, opened 2021-09-08
VS2019 16.11.2 Failing to build Android binding libraries


kristijan-boskovic:
After switching from Debug to Release in Visual Studio and trying to build the solution, machine in our office was not able to build Android binding libraries anymore (BoradRadio.FaceSdk and BroadRadio.RfidSdk). We tried switching back to Debug, cleaning&rebuilding, re-installing VS/Android Studio/All Android SDKs/Java JDKs, and nothing helped. During the build, /lang/something is always added to these castings within CreateDelegate, which leads to many build errors.

### Log File
[MyProjectOutput.log](https://github.com/xamarin/xamarin-android/files/7126882/MyProjectOutput.log)

dellis1972:
Are you able to provide a test project? 

Or show the code generated for `obj\Debug\generated\src\Android_serialport_api.SerialPort.cs` ? 
It might help us figure out why we are generating invalid C# code.

kristijan-boskovic:
Here is the test project that contains test Anroid binding library with files generated after failed build:
[Testapp.zip](https://github.com/xamarin/xamarin-android/files/7127985/Testapp.zip)

dellis1972:
The test app Binding project produces this error for me

```
obj/Debug/generated/src/Com.MV.Nfe.Version.cs(33,23): error CS0542: 'Version': member names cannot be the same as their enclosing type
```

Is that what you are seeing?

kristijan-boskovic:
Sorry, we forgot to add one <remove-node ... /> in the Metadata.xaml of the BindingLibrary. Here is a version that builds on all the company machines besides the affected one. Screenshot related, that's some of the generated output with errors I get. You should be able to find the generated files inside TestBindingLibrary\obj\Debug\generated\src.

[Testapp.zip](https://github.com/xamarin/xamarin-android/files/7128555/Testapp.zip)

![Errors](https://user-images.githubusercontent.com/20706387/132503308-4f2639bb-62a1-490e-84da-f0bf8cee25b3.png)

dellis1972:
@jpobst  any ideas on this one? If you look at the screen shot ^ we are generating delegates with method names which include `/` characters. 

I can't repo this issue on my Mac, and the NamespaceMapping.cs file does not contain any entries with a `/` in them.

jpobst:
[...]

PR #6368, opened 2021-10-05
[build] Enable globalization analyzer errors and fix issues.

jpobst:
Context: https://github.com/xamarin/xamarin-android/issues/6271

Making culture-aware string comparisons when not intended can result in very hard to diagnose bugs.  (See https://github.com/xamarin/xamarin-android/issues/6271)

Like we added in JI (https://github.com/xamarin/java.interop/pull/879), we should use the appropriate Roslyn code analyzers set to `Error` severity to ensure that we fix current issues and do not introduce any new issues in the future.

One complication with enabling these rules is that the .NET 6+ version of these rules are stricter and require overloads that do not exist in .NET Framework or .NET Standard to fix the violations. Enabling these rules in `.editorconfig` affects all `$(TargetFrameworkMoniker)`s; we will instead use `Configuration.props` to only enable them for non-.NET 6+ builds.

Also tweaks and reuses the existing Roslyn Analyzers logic to exclude external code, as many libraries that are used for testing (like `Mono.Debugg[ing|er].Soft)` have issues surfaced by these analyzers.

Note that some non-test comparisons on filenames/paths were switch from case-sensitive to case-insensitive compares.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
@@ -95,7 +95,7 @@ namespace Xamarin.Android.Tasks
 						if (!string.IsNullOrEmpty (ProjectDir)) {
 							var fullRelPath = Path.GetFullPath (rel).Normalize (NormalizationForm.FormC);
 							var fullProjectPath = Path.GetFullPath (ProjectDir).Normalize (NormalizationForm.FormC);
-							if (fullRelPath.StartsWith (fullProjectPath)) {
+							if (fullRelPath.StartsWith (fullProjectPath, StringComparison.Ordinal)) {
 								rel = fullRelPath.Replace (fullProjectPath, string.Empty);
 							}
 						}
@@ -109,7 +109,7 @@ namespace Xamarin.Android.Tasks
 
 				if (prefixes != null) {
 					foreach (var p in prefixes) {
-						if (rel.StartsWith (p))
+						if (rel.StartsWith (p, StringComparison.Ordinal))
 							rel = rel.Substring (p.Length);
 					}
 				}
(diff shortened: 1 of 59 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
```diff
@@ -95,7 +95,7 @@ public override bool RunTask ()
 						if (!string.IsNullOrEmpty (ProjectDir)) {
 							var fullRelPath = Path.GetFullPath (rel).Normalize (NormalizationForm.FormC);
 							var fullProjectPath = Path.GetFullPath (ProjectDir).Normalize (NormalizationForm.FormC);
-							if (fullRelPath.StartsWith (fullProjectPath)) {
+							if (fullRelPath.StartsWith (fullProjectPath, StringComparison.Ordinal)) {
```
jonpryor, line 98: Paths *generally* should use `StringComparison.OrdinalIgnoreCase`: https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings#choosing-a-stringcomparison-member-for-your-method-call

> | Data | Behavior | Corresponding System.StringComparison value |
> | -- | -- | -- |
> | File paths | A non-linguistic identifier, where case is irrelevant; especially data stored in most Windows system services. | OrdinalIgnoreCase |

src/Xamarin.Android.Build.Tasks/Tasks/AndroidComputeResPaths.cs
```diff
@@ -109,7 +109,7 @@ public override bool RunTask ()
 
 				if (prefixes != null) {
 					foreach (var p in prefixes) {
-						if (rel.StartsWith (p))
+						if (rel.StartsWith (p, StringComparison.Ordinal))
```
jonpryor, line 112: If `rel` is a path, this should use `OrdinalIgnoreCase`.

src/Xamarin.Android.Build.Tasks/Tasks/CheckDuplicateJavaLibraries.cs
```diff
@@ -22,7 +22,7 @@ public class CheckDuplicateJavaLibraries : AndroidTask
 
 		public override bool RunTask ()
 		{
-			var jarFiles = (JavaSourceFiles != null) ? JavaSourceFiles.Where (f => f.ItemSpec.EndsWith (".jar")) : null;
+			var jarFiles = (JavaSourceFiles != null) ? JavaSourceFiles.Where (f => f.ItemSpec.EndsWith (".jar", StringComparison.Ordinal)) : null;
```
jonpryor, line 25: Filename checks should use `.OrdinalIgnoreCase`.

grendello (approved):

PR #6370, opened 2021-10-06
[ProjectTools] Guard tests against a Mono environment variable setting bug.

jpobst:


jonathanpeppers (approved):
This is probably fine to fix these tests.

Did we try just calling `System.Environment.GetEnvironmentVariables ()` ourselves manually at test startup? It seems like it fails on the first call when two happen in parallel?

PR #6378, opened 2021-10-08
[ci] Install .NET Core 3.1.19 runtime to notarize

pjcollins:
It appears that we were previously trying to install a .NET Core 3 SDK
version as a runtime, using the latest runtime version should fix that.


pjcollins:
Missed this in review on the previous PR, testing here - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5308613&view=results

PR #6382, opened 2021-10-11
[ci] Fix nightly ci by comparing 'XAShouldAnalyzeAssembly' as string instead of bool

jpobst:
The nightly pipeline is failing with:
```
/Users/builder/azdo/_work/4/s/xamarin-android/build-tools/scripts/RoslynAnalyzers.projitems(3,14): error MSB4113: Specified condition " $(XAShouldAnalyzeAssembly) " evaluates to "" instead of a boolean. [/Users/builder/azdo/_work/4/s/xamarin-android
```

We should update this comparison to a string compare so that it still works if `$(XAShouldAnalyzeAssembly)` doesn't get set at all.

### DIFF
--- a/build-tools/scripts/RoslynAnalyzers.projitems
+++ b/build-tools/scripts/RoslynAnalyzers.projitems
@@ -1,6 +1,6 @@
 <Project>
 
-  <ItemGroup Condition=" $(XAShouldAnalyzeAssembly) == 'True' ">
+  <ItemGroup Condition=" '$(XAShouldAnalyzeAssembly)' == 'True' ">
     <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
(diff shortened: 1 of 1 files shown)


jonpryor:
```markdown
Context: ad5f6e85b09baf2644b6131667b9631b64ad0cf3

The nightly pipeline is failing with:

	/Users/builder/azdo/_work/4/s/xamarin-android/build-tools/scripts/RoslynAnalyzers.projitems(3,14):
	error MSB4113: Specified condition " $(XAShouldAnalyzeAssembly) " evaluates to "" instead of a boolean.

The cause of the problem is commit ad5f6e85, which implicitly
requires that `$(XAShouldAnalyzeAssembly)` have a "boolean" value,
i.e. `True` or `False`.  The empty string is not a valid boolean
value.

If `$(XAShouldAnalyzeAssembly)` isn't set -- as can be the case on
the nightly CI run -- then an MSB4113 error results.

Update the `<ItemGroup/>` within
`build-tools/scripts/RoslynAnalyzers.projitems` to use a "safe"
boolean expression:

	<ItemGroup Condition=" '$(XAShouldAnalyzeAssembly)' == 'True' ">

This will work reliably even if `$(XAShouldAnalyzeAssembly)` isn't set.
```

PR #6387, opened 2021-10-12
[ci] Use .NET 6 RC 2 and use shared variables

pjcollins:
Updates CI to use .NET 6.0.100-rc.2 instead of the much older preview 3
version that we've been building against.  Shared variables (including
`DotNet6Version`) have been moved to a new variables.yaml template file.

### DIFF
--- a/build-tools/automation/yaml-templates/variables.yaml
+++ b/build-tools/automation/yaml-templates/variables.yaml
@@ -1,0 +1,33 @@
+variables:
+- name: RunningOnCI
+  value: true
+- name: XA.Build.Configuration
+  value: Release
+- name: XA.Jdk8.Folder
+  value: jdk-1.8
+- name: XA.Jdk11.Folder
+  value: jdk-11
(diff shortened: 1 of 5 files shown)


jpobst (approved):

build-tools/automation/yaml-templates/variables.yaml
```diff
@@ -0,0 +1,33 @@
...
+- name: InstallerArtifactName
+  value: installers-unsigned
+- name: TestAssembliesArtifactName
+  value: test-assemblies
+- name: NUnitConsoleVersion
+  value: 3.11.1
+- name: NUnit.NumberOfTestWorkers
+  value: 4
+- name: DotNet6Version
+  value: 6.0.100-rc.2.21505.57
```
jonathanpeppers, line 21: If we are using public builds now, could we change `use-dot-net.yaml` to use the built-in `UseDotNet` task again?

This would be for a separate PR.

PR #6390, opened 2021-10-13
[build] Fixes needed to build Xamarin.Android.sln with 'dotnet build'

jpobst:
Some fixes that were required to get `Xamarin.Android.sln` to build with `dotnet build` on Mac:
```
dotnet build Xamarin.Android.sln -m:1
```

Largely this is just path separator fixes for instances where .NET's `msbuild` doesn't fix paths that Mono's `msbuild` does.

Additionally change `api-compat` to run on `dotnet` when building with `dotnet build`.  Mono allows you to `Process.Start ("blah.exe")` and it will automagically change it to `"mono blah.exe"` behind your back.  .NET doesn't do this, so we need to explicitly call `mono` or `dotnet` when running on .NET.  We went with `dotnet` as long as we were making changes anyways.

Note: This is not an exhaustive PR of needed changes, but combined with https://github.com/xamarin/xamarin-android/pull/6388 it's pretty close.

Also note: This does not attempt to eliminate Mono from the build, targets that `<Exec>` out to run `.exe` processes (like `api-merge.exe`) still use Mono.

### DIFF
--- a/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
+++ b/build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
@@ -159,7 +159,7 @@ namespace Xamarin.Android.Tools.BootstrapTasks
 			for (int i = 0; i < 3; i++) {
 				using (var genApiProcess = new Process ()) {
 
-					if (Environment.Version.Major > 5) {
+					if (Environment.Version.Major >= 5) {
 						var apiCompat = new FileInfo (Path.Combine (ApiCompatPath, "..", "netcoreapp2.1", "Microsoft.DotNet.ApiCompat.dll"));
 						genApiProcess.StartInfo.FileName = "dotnet";
 						genApiProcess.StartInfo.Arguments = $"\"{apiCompat}\" ";
--- a/build-tools/scripts/JavaCallableWrappers.targets
+++ b/build-tools/scripts/JavaCallableWrappers.targets
@@ -16,7 +16,7 @@
       <_LibDirs>-L "$(OutputPathAbs.TrimEnd('\'))"</_LibDirs>
       <_LibDirs Condition=" '$(TargetFramework)' != 'net6.0' ">$(_LibDirs) -L "$(OutputPathAbs)..\v1.0" -L "$(OutputPathAbs)..\v1.0\Facades"</_LibDirs>
       <_LibDirs Condition=" '$(TargetFramework)' == 'net6.0' ">$(_LibDirs) -L "$(_CorlibDir.TrimEnd('\'))"</_LibDirs>
-      <_Out>-o "$(MSBuildProjectDirectory)\$(IntermediateOutputPath)jcw/src"</_Out>
+      <_Out>-o "$(MSBuildProjectDirectory)/$(IntermediateOutputPath)jcw/src"</_Out>
     </PropertyGroup>
     <Exec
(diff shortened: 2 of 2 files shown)


dellis1972:
This seems to work

```xml
  <Target Name="Test">
    <Message Text="$([System.IO.Path]::DirectorySeparatorChar)" />
  </Target>
```

So we might be able to setup a property which contains `DirectorySeparatorChar` so its not all `/`

dellis1972:
This seems to work

```xml
  <Target Name="Test">
    <Message Text="$([System.IO.Path]::DirectorySeparatorChar)" />
  </Target>
```

So we might be able to setup a property which contains `DirectorySeparatorChar` so its not all `/`

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -161,8 +159,17 @@ void ValidateApiCompat (string contractPath, bool validateAgainstReference)
 			for (int i = 0; i < 3; i++) {
 				using (var genApiProcess = new Process ()) {
 
-					genApiProcess.StartInfo.FileName = apiCompat.FullName;
-					genApiProcess.StartInfo.Arguments = $"\"{contractAssembly.FullName}\" -i \"{TargetImplementationPath}\" --allow-default-interface-methods ";
+					if (Environment.Version.Major > 5) {
```
jonathanpeppers, line 162: Should this be?
```suggestion
					if (Environment.Version.Major >= 5) {
```

build-tools/scripts/JavaCallableWrappers.targets
```diff
@@ -16,7 +16,7 @@
       <_LibDirs>-L "$(OutputPathAbs.TrimEnd('\'))"</_LibDirs>
       <_LibDirs Condition=" '$(TargetFramework)' != 'net6.0' ">$(_LibDirs) -L "$(OutputPathAbs)..\v1.0" -L "$(OutputPathAbs)..\v1.0\Facades"</_LibDirs>
       <_LibDirs Condition=" '$(TargetFramework)' == 'net6.0' ">$(_LibDirs) -L "$(_CorlibDir.TrimEnd('\'))"</_LibDirs>
-      <_Out>-o "$(MSBuildProjectDirectory)\$(IntermediateOutputPath)jcw\src"</_Out>
+      <_Out>-o "$(MSBuildProjectDirectory)\$(IntermediateOutputPath)jcw/src"</_Out>
```
jonathanpeppers, line 19: Maybe it's worth fixing all the slashes in this one:

```suggestion
      <_Out>-o "$(MSBuildProjectDirectory)/$(IntermediateOutputPath)jcw/src"</_Out>
```

It's weird this property doesn't have a trailing slash...

https://docs.microsoft.com/visualstudio/msbuild/msbuild-reserved-and-well-known-properties

jonpryor:
Some questions:

 1. How "complete" is this?  "This is not an exhaustive PR of needed changes" implies it's not complete.
 2. Should we create a new build lane which uses `dotnet build` on macOS, a'la 214dd76d9441c9974775f0b29feec0d5d8d1097f, or should we just update & replace the current macOS build job when this is complete?
 3. Even if this isn't complete (1) and we don't want to create or replace an existing CI lane (2), should this be merged *anyway* until (1)+(2) can be done?

What's the expected "strategy" here?

jpobst:
I have it building locally with `dotnet build`, but it requires additional hacks that I'm not yet sure how to do properly.  So we cannot set up a new lane or replace the existing one ***yet***.

I think which strategy we eventually decide to do will come down to if we intend to continue supporting building with `mono msbuild`.  I suspect any build system we aren't testing in CI will quickly bit-rot.  If we are comfortable losing the ability to build with `mono` then we should switch the current lane to use `dotnet`.  If not, we will need to build both with 2 lanes.

These fixes seem rather trivial and compatible with the current system, so I think we should go ahead and commit them.

Issue #6353, opened 2021-09-30
[.NET 6] Deploy fails on x86_64 API 31 emulator


jonathanpeppers:
### Steps to Reproduce

1. Setup x86_64 API 31 emulator
2. `dotnet new android`
3. `dotnet build -t:Install`

Fails with:

```
C:\Program Files\dotnet\packs\Microsoft.Android.Sdk.Windows\31.0.101-preview.9.16\tools\Xamarin.Android.Common.Debugging.targets(611,5): error ADB0020: Mono.AndroidTools.IncompatibleCpuAbiExceptiopn: The package does not support the CPU architecture of this device.
   at Mono.AndroidTools.Internal.AdbOutputParsing.CheckInstallSuccess(String output, String packageName) in /Users/builder/azdo/_work/1/s/xamarin-android/external/monodroid/tools/msbuild/external/androidtools/Mono.AndroidTools/Internal/AdbOutputParsing.cs:line 349
   at Mono.AndroidTools.AndroidDevice.<>c__DisplayClass104_0.<InstallPackage>b__0(Task`1 t) in /Users/builder/azdo/_work/1/s/xamarin-android/external/monodroid/tools/msbuild/external/androidtools/Mono.AndroidTools/AndroidDevice.cs:line 804
   at System.Threading.Tasks.ContinuationTaskFromResultTask`1.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__271_0(Object obj)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
--- End of stack trace from previous location ---
   at AndroidDeviceExtensions.PushAndInstallPackageAsync(AndroidDevice device, PushAndInstallCommand command, CancellationToken token) in /Users/builder/azdo/_work/1/s/xamarin-android/external/monodroid/tools/msbuild/external/androidtools/Xamarin.AndroidTools/Devices/AndroidDeviceExtensions.cs:line 206
[...]

PR #6398, opened 2021-10-18
[One .NET] specify all RIDs by default

jonathanpeppers:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1413756
Fixes: https://github.com/xamarin/xamarin-android/issues/6353

The API 31 emulator no longer has 32-bit images and the x86_64 image
has dropped support for 32-bit architectures:

    > adb shell getprop | grep cpu
    [ro.product.cpu.abi]: [x86_64]
    [ro.product.cpu.abilist]: [x86_64,arm64-v8a]
    [ro.product.cpu.abilist32]: []
    [ro.product.cpu.abilist64]: [x86_64,arm64-v8a]

Compared to an API 30 x86_64 emulator:

    > adb shell getprop | grep cpu
    [ro.product.cpu.abi]: [x86_64]
    [ro.product.cpu.abilist]: [x86_64,x86,arm64-v8a,armeabi-v7a,armeabi]
    [ro.product.cpu.abilist32]: [x86,armeabi-v7a,armeabi]
    [ro.product.cpu.abilist64]: [x86_64,arm64-v8a]

The problem is our default RIDs are:

    <RuntimeIdentifiers Condition=" '$(RuntimeIdentifier)' == '' And '$(RuntimeIdentifiers)' == '' ">android-arm64;android-x86</RuntimeIdentifiers>

And so you hit this error when trying to deploy a `dotnet new android`
app on an API 31 x86_64 emulator:

    error ADB0020: Mono.AndroidTools.IncompatibleCpuAbiExceptiopn: The package does not support the CPU architecture of this device.

To workaround this, you can add `android-x64` to your list of
`$(RuntimeIdentifiers)`.

To solve this issue, we can default `$(RuntimeIdentifiers)` to all 4
architectures. We have code that will select a single architecture for
Debug builds using "Fast Deployment". It seems better to have a
default here that will always work, and the only drawback would be the
additional architectures for Release builds.

After this change, I discovered an issue introduced in 33a6d1e52:

    ILLink error IL1012: IL Linker has encountered an unexpected error. Please report the issue at https://github.com/mono/linker/issues [C:\a\_work\1\s\bin\TestRelease\temp\BuildProguard Enabled Project(1)Trued8r8\UnnamedProject.csproj]
    ...
    Unhandled exception. System.IO.IOException: The process cannot access the file 'C:\a\_work\1\s\bin\TestRelease\temp\BuildProguard Enabled Project(1)Trued8r8\obj\Release\proguard\proguard_project_references.cfg' because it is being used by another process.

When we build each RID in parallel, each inner build was attempting to
write to the same file. I changed the path for this file to be:

    $(IntermediateOutputPath)%(_RIDs.Identity)\proguard\proguard_project_references.cfg

And then set `$(_ProguardProjectConfiguration)` appropriately after
the inner builds complete. This also needs to actually be a file path,
I don't see how the value was working before:

    <_ProguardProjectConfiguration Condition=" '$(AndroidLinkTool)' != '' ">;_ProguardProjectConfiguration=$(IntermediateOutputPath)proguard\proguard_project_references.cfg</_ProguardProjectConfiguration>

Then another related issue:

    (_TouchAndroidLinkFlag target) ->
    Microsoft.Android.Sdk.ILLink.targets(114,5): error MSB3371: The file "obj\Release\net6.0-android\link.flag" cannot be created. The process cannot access the file 'C:\a\_work\2\s\bin\TestRelease\temp\DotNetBuildandroid-armandroid-arm64android-x86android-x64TrueTrue\obj\Release\net6.0-android\link.flag' because it is being used by another process.

I setup `$(_AndroidLinkFlag)` the same as
`$(_ProguardProjectConfiguration)` to solve this issue.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
@@ -28,7 +28,6 @@ _ResolveAssemblies MSBuild target.
     <OutputPath Condition=" '$(_OuterOutputPath)' != '' ">$(_OuterOutputPath)</OutputPath>
     <OutDir     Condition=" '$(_OuterOutputPath)' != '' ">$(_OuterOutputPath)</OutDir>
     <PublishDir>$(OutputPath)</PublishDir>
-    <_AndroidLinkFlag>$(_OuterIntermediateOutputPath)link.flag</_AndroidLinkFlag>
     <BuildDependsOn>_RemoveLegacyDesigner;$(BuildDependsOn)</BuildDependsOn>
     <!-- We don't want IncrementalClean to run here, or files get deleted -->
     <CoreBuildDependsOn>
@@ -69,7 +68,6 @@ _ResolveAssemblies MSBuild target.
       <_RIDs Include="$(RuntimeIdentifiers)" Condition=" '$(RuntimeIdentifiers)' != '' " />
     </ItemGroup>
     <PropertyGroup>
-      <_ProguardProjectConfiguration Condition=" '$(AndroidLinkTool)' != '' ">;_ProguardProjectConfiguration=$(IntermediateOutputPath)proguard\proguard_project_references.cfg</_ProguardProjectConfiguration>
       <_AdditionalProperties>
         _ComputeFilesToPublishForRuntimeIdentifiers=true
         ;AppendRuntimeIdentifierToOutputPath=true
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
Another related error:
```
(_TouchAndroidLinkFlag target) -> 
Microsoft.Android.Sdk.ILLink.targets(114,5): error MSB3371: The file "obj\Release\net6.0-android\link.flag" cannot be created. The process cannot access the file 'C:\a\_work\2\s\bin\TestRelease\temp\DotNetBuildandroid-armandroid-arm64android-x86android-x64TrueTrue\obj\Release\net6.0-android\link.flag' because it is being used by another process.
```

dellis1972 (approved):

PR #6400, opened 2021-10-18
[Xamarin.Android.Build.Tasks] Remove Java.Interop.dll reference.

jpobst:
Commit 3e6a6232d904acfc95017eeae58408787b0f8dc9
Context: https://discord.com/channels/732297728826277939/732297837953679412/898675755779768330

An unfortunate unexpected breakage occurred due to commit 3e6a6232:
the version of `Java.Interop.dll` changed -- which we *should* have
expected but completely overlooked & forgot -- which in turn meant
that *everything* which depends upon it -- which is everything --
started breaking due to the version change, e.g.

	error CS1705: Assembly 'Microsoft.Maui' with identity 'Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'
	uses 'Java.Interop, Version=0.1.6.0, Culture=neutral, PublicKeyToken=84e04ff9cfb79065'
	which has a higher version than referenced assembly
	'Java.Interop' with identity 'Java.Interop, Version=0.1.2.0, Culture=neutral, PublicKeyToken=84e04ff9cfb79065'

The primary motivation with 3e6a6232 was to improve the use of
*MSBuild* task assemblies, however `Java.Interop.dll` is *not* really a *MSBuild*
assembly.  It is simply a dependency pulled in for `Java.Interop.Export.dll` which was
itself pulled in for a single small static method `MarshalMemberBuilder.GetMarshalMethodName (...)`.

As such, let's remove both `Java.Interop.dll` and `Java.Interop.Export.dll` from `Xamarin.Android.Build.Tasks`
and simple copy the needed method.

This means we do not need to continuously update the version number of JI/JIE, which will allow
assemblies created with newer versions of XA to continue working on earlier releases where JI has a
version number of `0.1.0.0`.

Additionally, remove `Java.Interop.dll` and `Java.Interop.Export.dll` from the `lib\xamarin.android\xbuild\Xamarin\Android` directory in the installer.  `Java.Interop.dll` should only be consumed via the framework profiles.  `Java.Interop.Export.dll` was experientable and is not in active use.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/Extensions.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/Extensions.cs
@@ -329,6 +329,7 @@ namespace MonoDroid.Tuner {
 			return false;
 		}
 
+		// Keep in sync with: https://github.com/xamarin/java.interop/blob/8ccb8374d242490d8d1b032f2c8ca7a813fd40f3/src/Java.Interop.Export/Java.Interop/MarshalMemberBuilder.cs#L405-L421
 		public static string GetMarshalMethodName (string name, string signature)
 		{
 			if (name == null)
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/Extensions.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/Extensions.cs
@@ -316,7 +316,7 @@ namespace MonoDroid.Tuner {
 			if (marshalType == null || !marshalType.HasMethods)
 				return false;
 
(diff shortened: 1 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/Extensions.cs
```diff
@@ -329,6 +329,24 @@ public static bool TryGetMarshalMethod (this MethodDefinition method, string nat
 			return false;
 		}
 
+		public static string GetMarshalMethodName (string name, string signature)
```
jonpryor, line 332: Need comment:

```csharp
// Keep in sync with: https://github.com/xamarin/java.interop/blob/8ccb8374d242490d8d1b032f2c8ca7a813fd40f3/src/Java.Interop.Export/Java.Interop/MarshalMemberBuilder.cs#L405-L421
```

PR #6408, opened 2021-10-21
[build] bump xamarin-android/main to 31.0.200.preview.11

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/blob/d61aae7f120f9c39c2186908fb3cd791b3bde871/Documentation/guides/HowToBranch.md
Context: https://github.com/xamarin/xamarin-android/tree/release/6.0.1xx-preview10

Branching xamarin-android with the rest of .NET. `main` is now
31.0.200.preview.11.

Going forward, our version number will be slightly confusing:

1. .NET 6 RTM/GA == 31.0.101.preview.10
2. .NET 6.0.200 == 31.0.200.preview.11

We may need to make other changes when we start consuming 6.0.200 .NET
SDK builds.


jonathanpeppers:
FYI @rolfbjarne @dalexsoto @Redth when you branch this time, I think it is the right thing to change `main` to `.200`?

Directory.Build.props
```diff
@@ -21,8 +21,8 @@
          * Bump last two digits of the patch version for service releases.
          * Bump first digit of the patch version for feature releases (and reset the first two digits to 0)
     -->
-    <AndroidPackVersion>31.0.101</AndroidPackVersion>
```
pjcollins, line 24: Maybe we should continue to use `.101` until we are consuming 6.0.200 .NET SDK builds? Alternatively, we might even want to stay on .101 until GA, and reserve `.200` for our second stable release? I am not sure what's best here...

PR #6440, opened 2021-10-27
A handful of tiny improvements for the native runtime

grendello:
* Rewrite `BasicUtilities::monodroid_strsplit`
    Instead of reallocating the vector on each new token, pre-calculate
    the number of tokens and allocate enough memory for the entire
    vector at once.
  * Rewrite `init_logging_categories`
    Eliminate memory allocation by not using `monodroid_strsplit`,
    instead do all the processing using on-stack strings and their
    tokenizer.  Also, replace the `CATEGORY` preprocessor macro with a
    template function.

### DIFF
--- a/src/monodroid/jni/basic-utilities.cc
+++ b/src/monodroid/jni/basic-utilities.cc
@@ -284,69 +284,73 @@ BasicUtilities::monodroid_strfreev (char **str_array)
 char **
 BasicUtilities::monodroid_strsplit (const char *str, const char *delimiter, size_t max_tokens)
 {
-	const char *c;
-	char *token, **vector;
-	size_t size = 1;
-
-	if (strncmp (str, delimiter, strlen (delimiter)) == 0) {
-		vector = (char **)xmalloc (2 * sizeof(vector));
-		vector[0] = strdup ("");
-		size++;
-		str += strlen (delimiter);
-	} else {
-		vector = nullptr;
+	if (str == nullptr || *str == '\0') {
+		return static_cast<char**>(xcalloc (sizeof(char*), 1));
 	}
 
-	while (*str && !(max_tokens > 0 && size >= max_tokens)) {
-		c = str;
-		if (strncmp (str, delimiter, strlen (delimiter)) == 0) {
-			token = strdup ("");
-			str += strlen (delimiter);
+	const char *p_str = str;
+	size_t tokens_in_str = 0;
+	size_t delimiter_len = strlen (delimiter);
+
+	while (*p_str != '\0') {
+		size_t bytes = strspn (p_str, delimiter);
+		if (bytes == 0) {
+			bytes = 1;
 		} else {
-			while (*str && strncmp (str, delimiter, strlen (delimiter)) != 0) {
-				str++;
-			}
-
-			if (*str) {
-				size_t toklen = static_cast<size_t>((str - c));
-				size_t alloc_size = ADD_WITH_OVERFLOW_CHECK (size_t, toklen, 1);
-				token = new char [alloc_size];
-				strncpy (token, c, toklen);
-				token [toklen] = '\0';
-
-				/* Need to leave a trailing empty
-				 * token if the delimiter is the last
-				 * part of the string
-				 */
-				if (strcmp (str, delimiter) != 0) {
-					str += strlen (delimiter);
-				}
-			} else {
-				token = strdup (c);
-			}
+			tokens_in_str += bytes / delimiter_len;
 		}
 
-		add_to_vector (&vector, size, token);
-		size++;
+		p_str += bytes;
 	}
 
-	if (*str) {
-		if (strcmp (str, delimiter) == 0)
-			add_to_vector (&vector, size, strdup (""));
-		else {
-			/* Add the rest of the string as the last element */
-			add_to_vector (&vector, size, strdup (str));
+	size_t vector_size = (max_tokens > 0 && tokens_in_str >= max_tokens) ? max_tokens + 1 : tokens_in_str + 2; // Includes the terminating 'nullptr` entry
+
+	char **vector = static_cast<char**>(xmalloc (MULTIPLY_WITH_OVERFLOW_CHECK (size_t, sizeof(char*), vector_size)));
+	size_t vector_idx = 0;
+
+	while (*str != '\0' && !(max_tokens > 0 && vector_idx + 1 >= max_tokens)) {
+		const char *c = str;
+
+		if (strncmp (str, delimiter, delimiter_len) == 0) {
+			vector[vector_idx++] = strdup ("");
+			str += delimiter_len;
+			continue;
 		}
-		size++;
+
+		while (*str != '\0' && strncmp (str, delimiter, delimiter_len) != 0) {
+			str++;
+		}
+
+		if (*str == '\0') {
+			vector[vector_idx++] = strdup (c);
+			continue;
+		}
+
+		size_t toklen = static_cast<size_t>((str - c));
+		size_t alloc_size = ADD_WITH_OVERFLOW_CHECK (size_t, toklen, 1);
+		char *token = static_cast<char*>(xmalloc (alloc_size));
+		strncpy (token, c, toklen);
+		token [toklen] = '\0';
+		vector[vector_idx++] = token;
+
+		/* Need to leave a trailing empty
+		 * token if the delimiter is the last
+		 * part of the string
+		 */
+		if (strcmp (str, delimiter) != 0) {
+			str += delimiter_len;
+		}
 	}
 
-	if (vector == nullptr) {
-		vector = (char **) xmalloc (2 * sizeof (vector));
-		vector [0] = nullptr;
-	} else if (size > 0) {
-		add_to_vector (&vector, size, nullptr);
+	if (*str != '\0') {
+		if (strncmp (str, delimiter, delimiter_len) == 0) {
+			vector[vector_idx++] = strdup ("");
+		} else {
+			vector[vector_idx++] = strdup (str);
+		}
 	}
 
+	vector[vector_idx] = nullptr;
 	return vector;
 }
 
@@ -369,17 +373,4 @@ BasicUtilities::monodroid_strdup_vprintf (const char *format, va_list vargs)
 	int n = vasprintf (&ret, format, vargs);
 
 	return n == -1 ? nullptr : ret;
-}
-
-void
-BasicUtilities::add_to_vector (char ***vector, size_t size, char *token)
-{
-	if (*vector == nullptr) {
-		*vector = (char **)static_cast<char**>(xmalloc (size * sizeof(*vector)));
-	} else {
-		size_t alloc_size = MULTIPLY_WITH_OVERFLOW_CHECK (size_t, sizeof(*vector), size + 1);
-		*vector = static_cast<char**>(xrealloc (*vector, alloc_size));
-	}
-
-	(*vector)[size - 1] = token;
 }
--- a/src/monodroid/jni/logger.cc
+++ b/src/monodroid/jni/logger.cc
@@ -24,6 +24,7 @@
 	va_end ((_args_));
 
 using namespace xamarin::android;
+using namespace xamarin::android::internal;
 
 // Must match the same ordering as LogCategories
 static const char* log_names[] = {
@@ -99,10 +100,10 @@ open_file (LogCategories category, const char *path, const char *override_dir, c
 	return f;
 }
 
-static const char *gref_file = NULL;
-static const char *lref_file = NULL;
-static int light_gref  = 0;
-static int light_lref  = 0;
+static const char *gref_file = nullptr;
+static const char *lref_file = nullptr;
+static bool light_gref  = false;
+static bool light_lref  = false;
 
 void
 init_reference_logging (const char *override_dir)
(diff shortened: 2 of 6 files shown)


src/monodroid/jni/basic-utilities.cc
```diff
@@ -284,69 +284,73 @@ BasicUtilities::monodroid_strfreev (char **str_array)
...
-				 * part of the string
-				 */
-				if (strcmp (str, delimiter) != 0) {
-					str += strlen (delimiter);
-				}
-			} else {
-				token = strdup (c);
-			}
+	while (*p_str != '\0') {
+		size_t bytes = strspn (p_str, delimiter);
```
jonpryor, line 296: While the use of `strspn()` works for our codebase, that's only because our codebase is a "degenerate" case, and `delimiter` is always a single character.  Consider the [`strspn` docs](https://www.cplusplus.com/reference/cstring/strspn/) and example:

```cpp
  char strtext[] = "129th";
  char cset[] = "1234567890";

  i = strspn (strtext,cset);
  printf ("The initial number has %d digits.\n",i);

// Output:
// The initial number has 3 digits.
```

The second parameter is a "character set"; `strspn()` returns an index into `strtext` for the first character *not* in the "character set" specified by the second parameter.

So long as `delimiter` is only a single character, we have "semantic overlap".  As soon as `delimiter` becomes a *multi-character* delimiter, this will fail in unexpected ways, e.g.

```cpp
char **strs = monodroid_strsplit ("foo=>bar", "=>", 2);
// Reader (read: @jonpryor) expects strs[0][0]=="foo", strs[0][1]=="bar".
// Use of strspn() *probably* results in {"foo", ">bar"} or something otherwise unexpected.
```

I thus believe that use of `strspn()` means that this isn't a "drop-in" replacement.  It works *now*, but only by happenstance, and doesn't match the "understood" semantics of what "strsplit" *means*.

src/monodroid/jni/basic-utilities.cc
```diff
@@ -284,69 +284,73 @@ BasicUtilities::monodroid_strfreev (char **str_array)
 char **
```
jonpryor, line 284: If we're gonna "improve" things, should we consider changing the return type of `monodroid_strsplit()` from `char**` to `vector<string[max_tokens]>`?  (Is that even an express-able construct?)

src/monodroid/jni/basic-utilities.cc
jonpryor: I think I wrote too soon, in that this `while` loop is to "pre-count" the number of items to allocate.

That said, I'm not sure how `strspn()` actually interacts with the "pre-count" idea, and can't reasonably verify this algorithm.

src/monodroid/jni/logger.cc
```diff
@@ -113,93 +114,149 @@ init_reference_logging (const char *override_dir)
...
 		}
 	}
 }
 
+template<size_t NameSize>
+force_inline static bool
+set_category (const char (&name)[NameSize], string_segment& arg, unsigned int entry, bool arg_starts_with_name = false)
+{
+	if ((log_categories & entry) == entry) {
+		return false;
```
jonpryor, line 130: Not sure I understand this rationale: I think the `if` is "if this category is already being logged".  If the category *is* already being logged, shouldn't this return `true` so that we hit the `continue` in `init_logging_categories()`?

Consider:

```cpp
set_category ("default",  param, LOG_DEFAULT);
set_category ("default",  param, LOG_DEFAULT);
```

Shouldn't the 2nd invocation have the same return value as the first invocation?

src/monodroid/jni/basic-utilities.cc
grendello: Having had the same doubts about `strspn` as you, I wrote a test program to verify that it works. Here it is:

```cpp
#include <string>
#include <iostream>
#include <vector>
#include <cstdlib>
#include <cstring>
#include <cstdint>

void *xmalloc (size_t size)
{
	return malloc (size);
}

void *xrealloc (void *ptr, size_t size)
{
	return realloc (ptr, size);
}

void *xcalloc (size_t nmemb, size_t size)
{
	return calloc (nmemb, size);
}

void add_to_vector (char ***vector, size_t size, char *token)
{
	if (*vector == nullptr) {
		*vector = (char **)static_cast<char**>(xmalloc (size * sizeof(*vector)));
	} else {
		size_t alloc_size = sizeof(*vector) * (size + 1);
		*vector = static_cast<char**>(xrealloc (*vector, alloc_size));
	}

	(*vector)[size - 1] = token;
}

char **monodroid_strsplit (const char *str, const char *delimiter, size_t max_tokens)
{
	const char *c;
	char *token, **vector;
	size_t size = 1;

	if (strncmp (str, delimiter, strlen (delimiter)) == 0) {
		vector = (char **)xmalloc (2 * sizeof(vector));
		vector[0] = strdup ("");
		size++;
		str += strlen (delimiter);
	} else {
		vector = nullptr;
	}

	while (*str && !(max_tokens > 0 && size >= max_tokens)) {
		c = str;
		if (strncmp (str, delimiter, strlen (delimiter)) == 0) {
			token = strdup ("");
			str += strlen (delimiter);
		} else {
			while (*str && strncmp (str, delimiter, strlen (delimiter)) != 0) {
				str++;
			}

			if (*str) {
				size_t toklen = static_cast<size_t>((str - c));
				size_t alloc_size = toklen + 1;
				token = new char [alloc_size];
				strncpy (token, c, toklen);
				token [toklen] = '\0';

				/* Need to leave a trailing empty
				* token if the delimiter is the last
				* part of the string
				*/
				if (strcmp (str, delimiter) != 0) {
					str += strlen (delimiter);
				}
			} else {
				token = strdup (c);
			}
		}

		add_to_vector (&vector, size, token);
		size++;
	}

	if (*str) {
		if (strcmp (str, delimiter) == 0)
			add_to_vector (&vector, size, strdup (""));
		else {
			/* Add the rest of the string as the last element */
			add_to_vector (&vector, size, strdup (str));
		}
		size++;
	}

	if (vector == nullptr) {
		vector = (char **) xmalloc (2 * sizeof (vector));
		vector [0] = nullptr;
	} else if (size > 0) {
		vector[size - 1] = nullptr;
	}

	return vector;
}

char **new_monodroid_strsplit (const char *str, const char *delimiter, size_t max_tokens)
{
	if (str == nullptr || *str == '\0') {
		return static_cast<char**>(xcalloc (sizeof(char*), 1));
	}

	const char *p_str = str;
	size_t tokens_in_str = 0;
	size_t delimiter_len = strlen (delimiter);

	while (*p_str != '\0') {
		size_t bytes = strspn (p_str, delimiter);
		if (bytes == 0) {
			bytes = 1;
		} else {
			tokens_in_str += bytes / delimiter_len;
		}

		p_str += bytes;
	}

	size_t vector_size = (max_tokens > 0 && tokens_in_str >= max_tokens) ? max_tokens + 1 : tokens_in_str + 2; // Includes the terminating 'nullptr` entry
	char **vector = static_cast<char**>(xmalloc (sizeof(char*) * vector_size));
	size_t vector_idx = 0;

	while (*str != '\0' && !(max_tokens > 0 && vector_idx + 1 >= max_tokens)) {
		const char *c = str;

		if (strncmp (str, delimiter, delimiter_len) == 0) {
			vector[vector_idx++] = strdup ("");
			str += delimiter_len;
			continue;
		}

		while (*str != '\0' && strncmp (str, delimiter, delimiter_len) != 0) {
			str++;
		}

		if (*str == '\0') {
			vector[vector_idx++] = strdup (c);
			continue;
		}

		size_t toklen = static_cast<size_t>((str - c));
		size_t alloc_size = toklen + 1;
		char *token = static_cast<char*>(xmalloc (alloc_size));
		strncpy (token, c, toklen);
		token [toklen] = '\0';
		vector[vector_idx++] = token;

		/* Need to leave a trailing empty
		* token if the delimiter is the last
		* part of the string
		*/
		if (strcmp (str, delimiter) != 0) {
			str += delimiter_len;
		}
	}

	if (*str != '\0') {
		if (strncmp (str, delimiter, delimiter_len) == 0) {
			vector[vector_idx++] = strdup ("");
		} else {
			vector[vector_idx++] = strdup (str);
		}
	}

	vector[vector_idx] = nullptr;
	return vector;
}

struct Test
{
	std::string input_string;
	std::string delimiter;
	size_t max_tokens = 0;
	std::vector<std::string> expected_result;
	bool null_return = false;
};

static std::vector<Test> tests {
	{
		"No delimiters",
		","
	},

	{
		",delimiter at start",
		",",
	},

	{
		"delimiter at end,",
		",",
	},

	{
		",,consecutive delimiters at start",
		",",
	},

	{
		"consecutive delimiters at end,,",
		",",
	},

	{
		"some, delimiters, in, the, middle",
		",",
	},

	{
		"consecutive, delimiters,, in, the, middle",
		",",
	},

	{
		"max_tokens,less,than,actual,tokens",
		",",
		3,
	},

	{
		",",
		",",
	},

	{
		",,",
		",",
	},

	{
		"",
		",",
	},

	{
		"##one##two##three",
		"##"
	},

	{
		"####one##two##three",
		"##"
	},
};

void dump_vector (std::vector<std::string> const& vector)
{
	std::cout << "  Vector contents:\n";
	size_t i = 0;
	for (std::string const& s : vector) {
		std::cout << "    [" << i++ << "]: '" << s << "'\n";
	}
}

int main ()
{
	int ret = 0;

	for (Test& test : tests) {
		std::cout << "Input string: '" << test.input_string << "'; Tokens (max " << test.max_tokens << "):\n";
		char **result = monodroid_strsplit (test.input_string.c_str (), test.delimiter.c_str (), test.max_tokens);

		if (result == nullptr) {
			std::cout << "  result is nullptr\n\n";
			test.null_return = true;
			continue;
		}

		if (*result == nullptr) {
			std::cout << "  result is an empty vector\n\n";
			continue;
		}

		char **p = result;
		int i = 0;
		while (p != nullptr && *p != nullptr) {
			std::cout << "  [" << i++ << "]: '" << *p << "'\n";
			test.expected_result.emplace_back (*p);
			p++;
		}

		std::cout << "\n";
	}

	std::cout << "\nTesting new implementation against results from the old one\n\n";
	for (Test& test : tests) {
		std::cout << "Input string: '" << test.input_string << "'; Tokens (max " << test.max_tokens << "):\n";
		char **result = new_monodroid_strsplit (test.input_string.c_str (), test.delimiter.c_str (), test.max_tokens);

		if (result == nullptr) {
			if (!test.null_return) {
				std::cout << "  fail (should not return nullptr)\n";
				ret++;
			} else {
				std::cout << "  success\n";
			}
			continue;
		}

		if (*result == nullptr) {
			if (!test.expected_result.empty ()) {
				std::cout << "  fail (should not return an empty vector)\n";
			} else {
				std::cout << "  success\n";
			}
			continue;
		}

		std::vector<std::string> result_vector;
		char **p = result;
		int i = 0;
		while (p != nullptr && *p != nullptr) {
			//			std::cout << "  [" << i++ << "]: '" << *p << "'\n";
			result_vector.emplace_back (*p);
			p++;
		}

		if (test.expected_result.size () != result_vector.size ()) {
			std::cout << "  fail (mismatched vector size)\n";
			dump_vector (result_vector);
			ret++;
			continue;
		}

		bool found_mismatch = false;
		for (size_t i = 0; i < test.expected_result.size (); i++) {
			if (test.expected_result[i] == result_vector[i]) {
				continue;
			}
			std::cout << "  fail (strings differ at index " << i << ": '" << test.expected_result[i] << "' != '" << result_vector[i] << "')\n";
			found_mismatch = true;
			ret++;
		}

		if (found_mismatch) {
			continue;
		}

		std::cout << "  success\n";
	}

	return ret;
}
```

grendello: And, for completeness, here is the implementation of `strspn` from bionic (it's copied from OpenBSD):

```c
size_t
strspn(const char *s1, const char *s2)
{
        const char *p = s1, *spanp;
        char c, sc;

        /*
         * Skip any characters in s2, excluding the terminating \0.
         */
cont:
        c = *p++;
        for (spanp = s2; (sc = *spanp++) != 0;)
                if (sc == c)
                        goto cont;
        return (p - 1 - s1);
}
```

src/monodroid/jni/basic-utilities.cc
grendello: We can't do that since that would be an ABI break. `monodroid_strsplit` is a p/invoke, and also we pass its return value to Mono when [initializing the JIT](https://github.com/xamarin/xamarin-android/blob/617500c92d989c87641db7010595f296eeffcfde/src/monodroid/jni/monodroid-glue.cc#L823).

src/monodroid/jni/logger.cc
grendello: No, we can't do that. Consider that each category name is checked against each token in the property value being parsed. If we return `true` from here, then the category names that follow the call in the code won't be matched. Consider the value of `debug.mono.log` being `assembly,default,debugger` and this code:

```cpp
while (value.next_token (',', param)) {
  if (set_category ("assembly", param, LOG_ASSEMBLY)) {
    continue;
  }

  if (set_category ("default",  param, LOG_DEFAULT)) {
    continue;
  }

  if (set_category ("debugger", param, LOG_DEBUGGER))  {
    continue;
  }
```

`assembly` would be matched for the first token and neither `default` nor `debugger` would ever be matched if we returned `true` on line 130. In other words, the return value from `set_category` should be read as `found a match` instead of `category already set`. It's just a tiny optimization.

src/monodroid/jni/basic-utilities.cc
jonpryor: And for added completeness, how to build & run @grendello's [test app](https://github.com/xamarin/xamarin-android/pull/6440#discussion_r742622097):

    c++ -std=c++20 app.cpp && ./a.out

Additionally, if I uncomment line 318:

```cpp
std::cout << "  [" << i++ << "]: '" << *p << "'\n";
```

and add a new test case to `tests`:

```cpp
	{
		"one##two##three",
		"##",
		2,
	},
```

I get the desired output:

```
Input string: 'one##two##three'; Tokens (max 2):
  [0]: 'one'
  [1]: 'two##three'
  success
```

Looks good to me!

PR #6445, opened 2021-10-29
[Mono.Android] add DynamicDependency for AndroidMessageHandler

jonathanpeppers:
Fixes: https://github.com/dotnet/runtime/issues/60896

`dotnet new android` with `new HttpClient()` used crashes during a
`Release` build with:

    UNHANDLED EXCEPTION:
    Android.Runtime.JavaProxyThrowable: Exception_WasThrown, Android.Runtime.JavaProxyThrowable
    --- End of managed Android.Runtime.JavaProxyThrowable stack trace ---
    android.runtime.JavaProxyThrowable: System.Reflection.TargetInvocationException: Arg_TargetInvocationException
    ---> System.Reflection.TargetInvocationException: Arg_TargetInvocationException
    ---> System.MissingMethodException: Arg_NoDefCTor, Xamarin.Android.Net.AndroidMessageHandler
        at System.RuntimeType.CreateInstanceMono(Boolean , Boolean )
        at System.RuntimeType.CreateInstanceDefaultCtor(Boolean , Boolean )
        at System.Activator.CreateInstance(Type type, Boolean nonPublic, Boolean wrapExceptions)
        at System.Activator.CreateInstance(Type type, Boolean nonPublic)
        at System.Activator.CreateInstance(Type type)
        at Android.Runtime.AndroidEnvironment.GetHttpMessageHandler()

Building the app with `-p:SuppressTrimAnalysisWarnings=false
-p:TrimmerSingleWarn=false` you also get the linker warning:

    src/Mono.Android/Android.Runtime/AndroidEnvironment.cs(349,5): warning IL2057: Android.Runtime.AndroidEnvironment.GetHttpMessageHandler():
    Unrecognized value passed to the parameter 'typeName' of method 'System.Type.GetType(String,Boolean)'. It's not possible to guarantee the availability of the target type.

I was able to reproduce this issue in an existing linker test.

We simply need to decorate `AndroidEnvironment.GetHttpMessageHandler()`
with `[DynamicDependency]`. This fixes the failing test.

I also removed `AndroidMessageHandler` from "classic" Xamarin.Android,
where it should not be used.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64XFormsDotNet.apkdesc
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64XFormsDotNet.apkdesc
@@ -5,226 +5,226 @@
       "Size": 3568
     },
     "assemblies/FormsViewGroup.dll": {
-      "Size": 7236
+      "Size": 7251
     },
     "assemblies/Java.Interop.dll": {
-      "Size": 65412
+      "Size": 65415
     },
     "assemblies/Microsoft.Win32.Primitives.dll": {
-      "Size": 4860
+      "Size": 4826
     },
     "assemblies/Mono.Android.dll": {
-      "Size": 454745
+      "Size": 476882
     },
     "assemblies/mscorlib.dll": {
-      "Size": 4233
+      "Size": 4191
     },
     "assemblies/netstandard.dll": {
-      "Size": 5948
+      "Size": 5909
     },
     "assemblies/rc.bin": {
       "Size": 946
     },
     "assemblies/System.Collections.Concurrent.dll": {
-      "Size": 12647
+      "Size": 12611
     },
     "assemblies/System.Collections.dll": {
-      "Size": 20532
+      "Size": 20498
     },
     "assemblies/System.Collections.NonGeneric.dll": {
-      "Size": 9510
+      "Size": 9467
     },
     "assemblies/System.ComponentModel.dll": {
-      "Size": 2910
+      "Size": 2873
     },
     "assemblies/System.ComponentModel.Primitives.dll": {
-      "Size": 3503
+      "Size": 3468
     },
     "assemblies/System.ComponentModel.TypeConverter.dll": {
-      "Size": 7236
+      "Size": 7198
     },
     "assemblies/System.Console.dll": {
-      "Size": 7625
+      "Size": 7589
     },
     "assemblies/System.Core.dll": {
-      "Size": 2408
+      "Size": 2373
     },
     "assemblies/System.Diagnostics.DiagnosticSource.dll": {
-      "Size": 4030
+      "Size": 3991
     },
     "assemblies/System.Diagnostics.TraceSource.dll": {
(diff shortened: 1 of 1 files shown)


src/Mono.Android/Mono.Android.csproj
```diff
@@ -361,8 +361,8 @@
     <Compile Include="System.IO\AndroidExtensions.cs" />
     <Compile Include="System.Linq\Extensions.cs" />
     <Compile Condition=" '$(TargetFramework)' != 'monoandroid10' " Include="Xamarin.Android.Net\AndroidClientHandler.cs" />
+    <Compile Condition=" '$(TargetFramework)' != 'monoandroid10' " Include="Xamarin.Android.Net\AndroidMessageHandler.cs" />
```
jonathanpeppers, line 364: Is it OK to do this now? It broke the apicompat check:
```
/Users/builder/azdo/_work/1/s/xamarin-android/src/Mono.Android/Mono.Android.targets(259,5): error : TypesMustExist : Type 'Xamarin.Android.Net.AndroidMessageHandler' does not exist in the implementation but it does exist in the contract. 
```

jonpryor: Can `AndroidMessageHandler` *work* on Classic Xamarin.Android?  If it can, keeping it around would be useful for "migration support".  Or, how hard would it be to provide a useful Classic implementation?

jonathanpeppers: I mean you could probably do `new HttpClient(new AndroidMessageHandler())` in classic Xamarin.Android, but we don't have tests in place to know if this even works. I think we just accidentally added the type.

I would think for migration is most people will have `new HttpClient()` and expect the right native handler to be used depending on their project settings?

jonathanpeppers: I moved the tests around in https://github.com/xamarin/xamarin-android/pull/6445/commits/3c068980f6a1340d7a2531714f15ab45eca4c48d, and `AndroidMessageHandler` is working. Let's just add it back and see how it goes.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64XFormsDotNet.apkdesc
```diff
@@ -5,226 +5,226 @@
...
-      "Size": 65412
+      "Size": 65415
     },
     "assemblies/Microsoft.Win32.Primitives.dll": {
-      "Size": 4860
+      "Size": 4826
     },
     "assemblies/Mono.Android.dll": {
-      "Size": 454745
+      "Size": 476882
```
jonathanpeppers, line 17: I think this is correct? Now that `AndroidEnvironment.GetHttpMessageHandler()` preserves `AndroidMessageHandler`, the various `Java.Net.*` APIs are also preserved in `Mono.Android.dll`?

grendello: I think it's allright

grendello (approved):

dellis1972 (approved):

PR #6449, opened 2021-11-01
[illink] fix XA2000 warning, again

jonathanpeppers:
Context: https://github.com/dotnet/linker/issues/2333

In testing 10b97206, I found that the XA2000 warning didn't actually
appear when consuming a class library with `AppDomain` usage in .NET 6.

After some discussion on dotnet/linker#2333, the only way to log a
warning from a custom trimmer step is:

    Context.LogMessage (MessageContainer.CreateCustomWarningMessage (Context, msg, 6200, new MessageOrigin (), WarnVersion.ILLink5))

Where the code must be 6001 or higher, and it will automatically be
prefixed with `IL`. I mapped `XA2000`, to `IL6200`, to just have the
number *similar*? We can chooose something else, let me know.

Once the warning was working, I got *many* warnings for the same
assembly. It turns out we were emitting it in .NET 6 for every type!

    protected void ProcessType (TypeDefinition type)
    {
        var assembly = type.Module.Assembly;
        if (!CheckShouldProcessAssembly (assembly))
            return;
    //...
    bool CheckShouldProcessAssembly (AssemblyDefinition assembly)
    {
        if (!Annotations.HasAction (assembly))
            Annotations.SetAction (assembly, AssemblyAction.Skip);
        if (IsProductOrSdkAssembly (assembly))
            return false;
        // Called for each type!
        CheckAppDomainUsage (assembly, (string msg) =>

To solve this issue, I created a new `HashSet<string>` of assembly
names. This way we will only log this warning once per assembly.

I updated a test for this scenario.

### DIFF
--- a/Documentation/guides/messages/il6200.md
+++ b/Documentation/guides/messages/il6200.md
@@ -1,9 +1,9 @@
 ---
-title: Xamarin.Android warning XA2000
-description: XA2000 warning code
-ms.date: 1/13/2019
+title: Xamarin.Android warning IL6200
+description: IL6200 warning code
+ms.date: 11/1/2021
 ---
-# Xamarin.Android warning XA2000
+# Xamarin.Android warning IL6200
 
 ## Example messages
 
@@ -11,7 +11,7 @@ ms.date: 1/13/2019
 Use of AppDomain.CreateDomain() detected in assembly: {assembly}. .NET 6 will only support a single AppDomain, so this API will no longer be available in Xamarin.Android once .NET 6 is released.
 ```
 
-_Note: this error maps to [`IL6200`](il6200.md]) in .NET 6 and higher._
+_Note: this error maps to [`XA2000`](xa2000.md]) in Xamarin.Android._
 
 ## Solution
 
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixAbstractMethodsStep.cs
@@ -56,9 +56,9 @@ namespace MonoDroid.Tuner
 
 			CheckAppDomainUsage (assembly, (string msg) =>
 #if ILLINK
-				Context.LogMessage (msg)
+				Context.LogMessage (MessageContainer.CreateCustomWarningMessage (Context, msg, 6200, new MessageOrigin (), WarnVersion.ILLink5))
 #else   // !ILLINK
-				Context.LogMessage (MessageImportance.High, msg)
+				Context.LogMessage (MessageImportance.High, "warning XA2000: " + msg)
 #endif  // !ILLINK
 			);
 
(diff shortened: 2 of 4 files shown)


dellis1972 (approved):
Looks ok to me.

jonpryor:
@jonathanpeppers wrote:

> Once the warning was working, I got many warnings for the same
> assembly.

Could you please provide an example of the warnings you were getting?

jonathanpeppers:
Sorry yes, in this test:
https://github.com/xamarin/xamarin-android/blob/85a50c8099731971f8dd7c927e683e83f5462ea8/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/LinkerTests.cs#L181-L182

I got this warning 7 x 4 RIDs = 28 times:
```
ILLink warning IL6200: Use of AppDomain.CreateDomain() detected in assembly: Library, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null. .NET 6 and higher will only support a single AppDomain, so this API will no longer be available in Xamarin.Android once .NET 6 is released.
```

dellis1972 (approved):

jonpryor:
```markdown
Context: https://github.com/dotnet/linker/issues/2333

In testing 10b97206, I found that the XA2000 warning didn't actually
appear when consuming a class library with `AppDomain` usage in a
.NET 6 SDK for Android app.

After some discussion on dotnet/linker#2333, the only way to log a
warning from a custom trimmer step is:

	Context.LogMessage (
	    MessageContainer.CreateCustomWarningMessage (
	      Context, msg, 6200, new MessageOrigin (), WarnVersion.ILLink5))

Where the code must be 6001 or higher, and it will automatically be
prefixed with `IL`.  I mapped `XA2000`, to `IL6200`, to just have a
*similar* value.

Once the warning was working, I got *many* warnings for the same
assembly.  It turns out that, due to how
`FixAbstractMethodStep.ProcessType()` and
`FixAbstractMethodStep.CheckShouldProcessAssembly()` interacted,
if *any* method called `AppDomain.CreateDomain()`, then *every type*
in the assembly would elicit a warning for the assembly:

	partial class FixAbstractMethodStep {

	    // ProcessType() is called once per type in a given library
	    protected void ProcessType (TypeDefinition type)
	    {
	        var assembly = type.Module.Assembly;

		// We thus call CheckShouldProcessAssembly() once per type
	        if (!CheckShouldProcessAssembly (assembly))
	            return;
	        // …
	    }

	    bool CheckShouldProcessAssembly (AssemblyDefinition assembly)
	    {
	        // …
	        // We thus warn for each separate type!
	        CheckAppDomainUsage (assembly, (string msg) =>
	    }
	}

This would spam the build log; our [unit test][0] contains only one
call to `AppDomain.CreateDomain()`, but would result in *28* warnings,
as the library contained 7 types, and was built across 4 RIDs.

To solve this issue, I created a new `HashSet<string>` of assembly
names.  This way we will only log this warning once per assembly.

I updated a test for this scenario.

[0]: https://github.com/xamarin/xamarin-android/blob/85a50c8099731971f8dd7c927e683e83f5462ea8/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/LinkerTests.cs#L181-L211
```

PR #6450, opened 2021-11-01
Testing https://github.com/xamarin/java.interop/pull/904

jonathanpeppers:


jonathanpeppers:
Don't think we need this anymore.

PR #6460, opened 2021-11-03
[ci] Add pipeline for updating api docs

pjcollins:
Introduces a new `azure-pipelines-apidocs.yaml` pipeline that can be
used to update the Mono.Android docs in the xamarin/android-api-docs
repository.  We will now only build Mono.Android with generated JavaDoc
xml when running this new pipeline.  The pipeline will run the
`<UpdateExternalDocumentation/>` target and upload the contents of
/android-api-docs/docs/Mono.Android/en/ as a build artifact.  These
updated docs can be downloaded for manual review and pull request
creation.

The .NET 6 packs have been updated to include the same Mono.Android.xml
docs that are bundled into the traditional Windows .vsix installer.
This xml file is generated from the content of external/android-api-docs
when the `<AssembleApiDocs/>` installer target is ran.

### DIFF
--- a/build-tools/automation/azure-pipelines-apidocs.yaml
+++ b/build-tools/automation/azure-pipelines-apidocs.yaml
@@ -1,0 +1,68 @@
+# Pipeline for updating and uploading android-api-docs/docs/Mono.Android/en/*
+
+trigger: none
+pr: none
+
+# Global variables
+variables:
+- template: yaml-templates/variables.yaml
+
+stages:
+- stage: mac_build
+  displayName: Build
+  dependsOn: []
+  jobs:
+  - job: mac_build_update_docs
+    displayName: macOS - Update API Docs
+    pool:
+      vmImage: macOS-11
+    timeoutInMinutes: 120
+    workspace:
+      clean: all
+    steps:
+    - checkout: self
+      submodules: recursive
+
+    - script: echo "##vso[task.setvariable variable=JI_JAVA_HOME]$HOME/Library/Android/$(XA.Jdk11.Folder)"
+      displayName: set JI_JAVA_HOME
+
+    - template: yaml-templates/use-dot-net.yaml
+
+    - task: NuGetAuthenticate@0
+      displayName: authenticate with azure artifacts
+      inputs:
+        forceReinstallCredentialProvider: true
+
+    - script: make prepare-update-mono CONFIGURATION=$(XA.Build.Configuration) PREPARE_CI=1 PREPARE_AUTOPROVISION=1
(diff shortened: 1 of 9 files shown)


pjcollins:
Test run of new pipeline: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5402379&view=logs&j=18eaad8b-760d-59fe-74fb-add94ccecaed

build-tools/installers/create-installers.targets
```diff
@@ -52,8 +52,12 @@
   <Target Name="_GenerateMsxDocXmls"
       DependsOnTargets="_FindFrameworkDirs;_FindDocSourceFiles"
       Inputs="@(_MsxDocSourceFile)"
-      Outputs="@(_MsxDocAssembly->'$(_LatestStableFrameworkDir)%(Identity).xml')">
+      Outputs="@(_MsxDocAssembly->'$(_LatestStableFrameworkDir)%(Identity).xml');@(_MsxDocAssembly->'$(_MonoAndroidNETOutputDir)%(Identity).xml')">
     <Exec Command="$(ManagedRuntime) &quot;$(MSBuildSrcDir)mdoc.exe&quot; --debug export-msxdoc -o &quot;$(_LatestStableFrameworkDir)%(_MsxDocAssembly.Identity).xml&quot; &quot;%(_MsxDocAssembly.SourceDir)&quot;" />
+    <Copy
+        SourceFiles="$(_LatestStableFrameworkDir)%(_MsxDocAssembly.Identity).xml"
+        DestinationFolder="$(_MonoAndroidNETOutputDir)"
+    />
```
jonathanpeppers, line 60: You might want to `<Touch/>` all the `Outputs` here, as `<Copy/>` doesn't do that...

build-tools/automation/azure-pipelines-apidocs.yaml
```diff
@@ -0,0 +1,70 @@
+# Pipeline for updating and uploading android-api-docs/docs/Mono.Android/en/*
+# https://dev.azure.com/devdiv/DevDiv/_build?definitionId=15262
+
+trigger: none
+pr: none
```
jonathanpeppers, line 5: So is the plan to just run this every now and then? Should we make a document that says how to do it?

pjcollins: Yeah, my thinking was that every time we get a new binding version or fix something in the javadoc transformation we could use this to build a new set of docs. Eventually the pipeline could probably create a PR into android-api-docs, but I don't think that's necessary for now.

jonathanpeppers (approved):

PR #6470, opened 2021-11-08
[ci] override schedule for Xamarin.Android nightly

jonathanpeppers:
We noticed the Xamarin.Android nightly job is running against old
branches like:

* `release/6.0.1xx-preview3`
* `release/6.0.1xx-preview4`
* `release/6.0.1xx-preview5`
* `release/6.0.1xx-preview6`
* `release/6.0.1xx-preview7`

xamarin-android/main does *not* have a schedule setup for these
branches... Meaning, we would have to go update the trigger on each
branch?

Instead, let's define the schedule in Azure DevOps UI. I've updated
`HowToBranch.md` with the new instructions.

I removed the schedule from `azure-pipelines-nightly.yaml`. After this
is merged, we may have to cherry-pick the schedule removal to each
branch. After this is merged, we'll find out.

### DIFF
--- a/Documentation/guides/HowToBranch.md
+++ b/Documentation/guides/HowToBranch.md
@@ -30,11 +30,15 @@ See [eng/README.md][2] for details on `darc` commands.
    `$(AndroidPackVersionSuffix)` in `Directory.Build.props` is
    incremented to the *next* version: `preview.43`. You may also need
    to update `$(AndroidPackVersion)` if `main` needs to target a new
-   .NET version band. In the same PR, update `azure-pipelines-nightly.yaml`
-   to build the new release branch.
+   .NET version band.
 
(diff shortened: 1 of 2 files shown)


jpobst (approved):

pjcollins (approved):
This seems like it should be a good way to avoid having to go update all older branches:

> If your YAML pipeline has both YAML scheduled triggers and UI defined scheduled triggers, only the UI defined scheduled triggers are run. 

https://docs.microsoft.com/en-us/azure/devops/pipelines/process/scheduled-triggers?view=azure-devops&tabs=yaml#scheduled-triggers

jonathanpeppers:
The test failures shouldn't matter here, going to merge.

Issue #3932, opened 2019-11-19
Do not add AOT'ed assemblies with 0 methods added


radekdoulik:
When using the profiled AOT, the aot compiler creates the `.so` libraries even for assemblies with 0 methods emitted. Avoiding these should result in a bit smaller apk size (yay) and possibly also better performance as we will not be loading these.

We should either fix aot compiler to not do it or AOT only assemblies in profiles. Another possibility might be to check the verbose output to see how many methods were added.

I will try to first fix it in mono as I think it doesn't make sense to create these.

claudioredi:
@radekdoulik : We're having a problem on Android <= 4.3 where the amount of generated .so files is greater than 128 and this causes the app to not start. I found this mono log

**D/Mono: DllImport error loading library 'libmono-native.so': 'Cannot load library: soinfo_alloc(linker.cpp:287): too many libraries when loading "libmono-native.so"'.**

As far I understand, this indicates that we're generating more .so files than Android can handle. I wonder if once you fix this problem I get a more reasonable amount of libraries. (I currently have around 140 libaot-xxxx libraries)

radekdoulik:
Yes, if many of those doesn't contain any methods then this should help. It is usually the case with the profiled AOT.

FoggyFinder:
any news?

PR #6477, opened 2021-11-10
Don't package empty AOT-d DSOs

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/3932

Profiled AOT tends to produce a number of "empty" .so shared libraries
because the profile names only a handful of types from a handful of
assemblies and the AOT compiler in Mono currently doesn't support not
outputting the .so in such cases.

This has the unfortunate effect that an application may have a large
number of such empty .so libraries and each of them will have to be
opened and loaded before Mono can determine that there's no code in them
and the JIT has to compile the requested types and methods.  Loading
every .so takes a non-zero amount of time, especially if the shared
libraries aren't unpacked to the filesystem.  I've seen times between 10
and 30ms to load a single shared library, which can add up to a
non-trivial amount of time wasted loading empty libraries.

This commit adds code to examine and ignore such .so files which don't
include any executable code.  The checks are crafted so that only AOT-d
libraries are considered for rejection.

Time savings are around 3ms for a plain XA app and around 10ms for a
sample MAUI app on the Pixel 3 XL phone.  In the latter case 49 shared
libraries out of 120 total are rejected and not packaged, contributing
to (small, each empty DSO is between 7 and 25k) space savings in the APK
in addition to slightly better startup performance.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -772,7 +772,12 @@ namespace Xamarin.Android.Tasks
 				Log.LogCodedWarning ("XA4301", path, 0, Properties.Resources.XA4301, item.archivePath);
 				return;
 			}
-			files.Add (item);
+
+			if (!ELFHelper.IsEmptyAOTLibrary (Log, item.filePath)) {
+				files.Add (item);
+			} else {
+				Log.LogDebugMessage ($"{item.filePath} is an empty (no executable code) AOT assembly, not including it in the archive");
+			}
 		}
 
 		// This method is used only for internal warnings which will never be shown to the end user, therefore there's
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -36,6 +36,7 @@
     <MonoOptionsVersion>6.12.0.148</MonoOptionsVersion>
     <SystemCollectionsImmutableVersion>1.7.1</SystemCollectionsImmutableVersion>
     <XliffTasksVersion>1.0.0-beta.20420.1</XliffTasksVersion>
+    <ELFSharpVersion>2.13.1</ELFSharpVersion>
   </PropertyGroup>
 
   <!-- Properties to help us run managed assemblies on various runtimes.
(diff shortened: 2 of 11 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -772,7 +775,75 @@ void AddNativeLibrary (ArchiveFileList files, string path, string abi, string ar
...
-			files.Add (item);
+
+			if (!IsEmptyAOTLibrary (item.filePath)) {
+				files.Add (item);
+			} else {
+				Log.LogDebugMessage ($"{item.filePath} is an empty (no executable code) AOT assembly, not including it in the archive");
+			}
+		}
+
+		bool IsEmptyAOTLibrary (string path)
```
jonathanpeppers, line 786: The only thing I'd suggest is to put this in a new class? `<BuildApk/>` is many lines, could we have a static `ElfHelper.IsEmptyAOTLibrary()` method instead?

grendello: Sure, it's a good idea, I'll update the code.

dellis1972:
I'm wondering, do we put any Metadata on the AOT'd libraries which we could use to early out this whole process. It would allow us to completely skip the .so file which don't have the required metadata.

grendello:
> I'm wondering, do we put any Metadata on the AOT'd libraries which we could use to early out this whole process. It would allow us to completely skip the .so file which don't have the required metadata.

This would move the check to earlier than BuildApk, perhaps it would be a better idea to check after we AOT the so and add the Metadata?

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -8,6 +8,9 @@
 using System.Text;
 using System.Text.RegularExpressions;
 
+using ELFSharp;
+using ELFSharp.ELF;
+using ELFSharp.ELF.Sections;
```
jonathanpeppers, line 13: You can probably drop these now

jonathanpeppers (approved):
Can this PR close this one, too?

https://github.com/xamarin/xamarin-android/issues/3932

jonpryor:
Shouldn't there be changes to `tests/apk-sizes-reference/Xamarin.Forms_Performance_Integration-Signed-Release-Profiled-Aot.apkdesc`?  Seems hard to believe that e.g. `libaot-Plugin.Connectivity.Abstractions.dll.so` has any code in there, or `libaot-Xamarin.AndroidX.Lifecycle.Common.dll.so`, or…

grendello:
> Shouldn't there be changes to `tests/apk-sizes-reference/Xamarin.Forms_Performance_Integration-Signed-Release-Profiled-Aot.apkdesc`? Seems hard to believe that e.g. `libaot-Plugin.Connectivity.Abstractions.dll.so` has any code in there, or `libaot-Xamarin.AndroidX.Lifecycle.Common.dll.so`, or…

The assemblies are AOT-d according to a pre-cooked profile, and also the code that filters them is very conservative since it's better to let something through the crack than to not package a `.so` that's needed.

PR #6478, opened 2021-11-10
[ci] Build with stable .NET 6 release

pjcollins:
Bumps the system .NET 6 install from RC 2 to the official stable build.


jonathanpeppers (approved):
LGTM, if the build ends up green. (I hope it does!)

PR #6485, opened 2021-11-12
[Samples] Rework Samples so it compiles with dotnet.

dellis1972:
The HelloWorld sample was not working with `dotnet`.
This is related to the following issue https://github.com/dotnet/sdk/issues/21613.

So for now, lets put the main app in a SubDirectory so that it will
compile.

PR #6492, opened 2021-11-16
[.NET] Stop .aar files being picked up from sub projects.

dellis1972:
PR #6485 reworked the HelloWorld samples example to fix the following
issue.

```
error APT2144: invalid file path '/Users/.../Documents/Sandbox/Xamarin/designerrewrite/samples/HelloWorld/obj/Debug/net6.0-android/lp/1.stamp'
```

This was occurring because when we built the `HelloWorld` app it would
automatically pickup various aar files in the `HelloWorldLibrary` project.
This project was  in a sub directory of the `HelloWorld` app.

What we would end up with is the app would end up including both
`Debug` and `Release` .aar files from the sub directory (if both were built).
But in addition it would also run a search for other `.aar` files which
would then be added with the `AndroidSkipResourceProcessing` metadata
set to `false`. So we would end up with data such as

```
AndroidAarLibrary=HelloWorldLibrary/bin/Debug/HelloWorldLibrary.aar
            AndroidSkipResourceProcessing=true
            Pack=true
AndroidAarLibrary=HelloWorldLibrary/bin/Release/HelloWorldLibrary.aar
            AndroidSkipResourceProcessing=true
            Pack=true
AndroidAarLibrary=/Users/foo/path/to/xa/samples/HelloWorldLibrary/bin/Debug/HelloWorldLibrary.aar
            AndroidSkipResourceProcessing=false
            Pack=false
```

These are ALL the same library and they should only be included once.
So we have a few fixes for this. The first is to alter the `AutoImport`
code to ignore sub directories depending on the build `$(Configuration)`.
This is so when building `Debug` we do NOT include `Release` in the
`AutoImport`.

Next is to alter the `_ResolveAars` target to exclude the auto imported
items from the `.aar` libraries we find during the search. This has to
be done using the `FullPath` metadata as the search will result in paths
which are full paths, but auto import produces relative ones. This is why
the old `Exclude` code was not working as expected.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
@@ -18,6 +18,9 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
 
   <PropertyGroup>
     <_DefaultJavaSourceJarPattern>**\*-source.jar;**\*-sources.jar;**\*-src.jar</_DefaultJavaSourceJarPattern>
+    <!-- Ignore Resources and files in SubDirectories -->
+    <_DefaultLibraryItemExcludes Condition=" '$(Configuration)' == 'Debug' " >$(DefaultItemExcludes);**/bin/Release/**;**/obj/Release/**</_DefaultLibraryItemExcludes>
+    <_DefaultLibraryItemExcludes Condition=" '$(Configuration)' == 'Release' ">$(DefaultItemExcludes);**/bin/Debug/**;**/obj/Debug/**</_DefaultLibraryItemExcludes>
   </PropertyGroup>
 
   <ItemGroup Condition=" '$(TargetPlatformIdentifier)' == 'android' and ('$(ImplicitUsings)' == 'true' or '$(ImplicitUsings)' == 'enable') ">
@@ -45,7 +48,7 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <TransformFile Include="Transforms\**\*.xml" />
     <!-- Default Java or native libraries -->
     <AndroidLibrary       Include="**\*.jar" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder);$(_DefaultJavaSourceJarPattern)" />
-    <AndroidLibrary       Include="**\*.aar" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <AndroidLibrary       Include="**\*.aar" Exclude="$(_DefaultLibraryItemExcludes);$(DefaultExcludesInProjectFolder)" />
     <AndroidNativeLibrary Include="**\*.so"  Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
     <JavaSourceJar        Include="$(_DefaultJavaSourceJarPattern)"   Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
   </ItemGroup>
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
```diff
@@ -18,6 +18,9 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
 
   <PropertyGroup>
     <_DefaultJavaSourceJarPattern>**\*-source.jar;**\*-sources.jar;**\*-src.jar</_DefaultJavaSourceJarPattern>
+    <!-- Ignore Resources and files in SubDirectories -->
+    <_DefaultLibraryItemExcludes Condition=" '$(Configuration)' == 'Debug' " >$(DefaultItemExcludes);**/bin/Release/**;**/obj/Release/**</_DefaultLibraryItemExcludes>
+    <_DefaultLibraryItemExcludes Condition=" '$(Configuration)' == 'Release' ">$(DefaultItemExcludes);**/bin/Debug/**;**/obj/Debug/**</_DefaultLibraryItemExcludes>
```
jonathanpeppers, line 23: I'm not sure this is actually valid, because you can name `$(Configuration)` anything you want. A lot of times people have  names like `AppStore`, `GooglePlay`, `Amazon`, etc.

Will it still work if we only specify `bin` and `obj` here?

Also, does it do any file I/O if you put `**/somedir/` in an `Exclude`?

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
```diff
@@ -45,7 +48,7 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <TransformFile Include="Transforms\**\*.xml" />
     <!-- Default Java or native libraries -->
     <AndroidLibrary       Include="**\*.jar" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder);$(_DefaultJavaSourceJarPattern)" />
-    <AndroidLibrary       Include="**\*.aar" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
+    <AndroidLibrary       Include="**\*.aar" Exclude="$(_DefaultLibraryItemExcludes);$(DefaultExcludesInProjectFolder)" />
     <AndroidNativeLibrary Include="**\*.so"  Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
```
jonathanpeppers, line 52: Would we want the new `Exclude` applied to all file types? You might have an `.jar` file in `obj`, too.

jonathanpeppers (commented):
Overall, I'm not sure I follow the pattern of what other .NET SDKs do for this problem.

One example, I found:

https://github.com/dotnet/sdk/blob/a30e465a2e2ea4e2550f319a2dc088daaafe5649/src/RazorSdk/Sdk/Sdk.Razor.StaticAssets.ProjectSystem.props#L22-L24

Maybe we could ignore the `lp` directory? It seems like ignoring `bin` and `obj` should be in the dotnet/sdk, because that would apply for any project type?

jonpryor:
@dellis1972: did this fall through the cracks?  Is this worth resuscitating?

dellis1972:
@jonpryor we can, but I'm not sure how to resolve https://github.com/dotnet/android/pull/6492#discussion_r750330841. I'll convert this one to draft and put it in my work log.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
dellis1972: @jonathanpeppers I've thought about this a bit more and I think we can just totally ignore any `**/obj/**` and `**/bin/**` directories in all configurations? It won't cover people who change their intermediate and output folder names but its a decent default.

jonathanpeppers: Is that the same as what they have here already?
* https://github.com/dotnet/sdk/blob/33343859a1a9a1fa171a92af62634fa7d0851638/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.DefaultOutputPaths.targets#L121-L122

dellis1972: not quite, this is for projects in Subdirectories. So the `$(OutputPath)` only covers `bin/xxx` not `SomeSubProject/bin/xxx`. 

Generally if people do this when using the default auto import they are in a world of pain as it will include code from sub projects also.

jonathanpeppers: Should .NET fix this? Their `$(DefaultItemExcludes)` would also be wrong if you had a project in a subfolder?

dellis1972: dunno, this is specifically for .aar files.

PR #6494, opened 2021-11-16
Fix `mono_log_mask` parsing and add property docs

grendello:
From the `how did I miss it?!` department.

The main purpose of this commit is to fix parsing of the `mono_log_mask`
component of the `debug.mono.log` property.  The component is set to a
list of Mono logging categories (e.g. `asm`, `gc` or `dll`) which, so
far, were supposed to be a comma-separated list.  However...
`debug.mono.log` **itself** uses commas to separate components, so using
the following:

```shell
$ adb shell setprop debug.mono.log mono_log_mask=asm,gc,dll
```

would **not** set the `MONO_LOG_MASK` environment variable to the
expected value of `asm,gc,dll` but instead it would end up being just
`asm`, as the other values would be considered to be `debug.mono.log`
components instead.  To fix this, this commit changes the
`mono_log_mask` separator to `:`, so the above command line will now be:

```shell
$ adb shell setprop debug.mono.log mono_log_mask=asm:gc:dll
```

Additionally, document all the `debug.mono.*` properties and remove the
`DEBUG_MONO_GDBPORT_PROPERTY` constant, since it's not used anywhere.

Also, when `assembly` or `gc` logging is enabled, we now set the Mono
tracing level to `info` instead of the default `error`.  This makes
tracking assembly loader and garbage collector issues easier by showing
more info from the Mono runtime, not just from Xamarin.Android.

### DIFF
--- a/src/monodroid/jni/monodroid-glue.cc
+++ b/src/monodroid/jni/monodroid-glue.cc
@@ -1958,11 +1958,8 @@ MonodroidRuntime::get_my_location (bool remove_file_name)
 
 #if defined (ANDROID)
 force_inline void
-MonodroidRuntime::setup_mono_tracing (char const* const& mono_log_mask)
+MonodroidRuntime::setup_mono_tracing (char const* const& mono_log_mask, bool have_log_assembly, bool have_log_gc)
 {
-	bool have_log_assembly = (log_categories & LOG_ASSEMBLY) != 0;
-	bool have_log_gc = (log_categories & LOG_GC) != 0;
-
 	constexpr char   MASK_ASM[] = "asm";
 	constexpr size_t MASK_ASM_LEN = sizeof(MASK_ASM) - 1;
 	constexpr char   MASK_DLL[] = "dll";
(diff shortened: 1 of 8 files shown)


src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1991,14 +1988,17 @@ MonodroidRuntime::setup_mono_tracing (char const* const& mono_log_mask)
 		return;
 	}
 
+	dynamic_local_string<PROPERTY_VALUE_BUFFER_LEN> input_log_mask;
+	input_log_mask.assign_c (mono_log_mask);
+	delete[] mono_log_mask;
```
jonpryor, line 1993: This feels "wrong", in that we're deleting something in a scope that differs from where it was allocated.  Shouldn't `mono_log_mask` (and `mono_log_level`!) be `delete[]`d in `MonodroidRuntime::Java_mono_android_Runtime_initInternal()`?  That way if (when?) `Java_mono_android_Runtime_initInternal()` is updated to use `mono_log_mask` *after* `setup_mono_tracing()` is called, it won't "unexpectedly" introduce a "use after free" bug?

Related: when `mono_log_mask` *is* deleted, shouldn't we set it to `nullptr` so that we'll get a SIGSEGV instead of a "use after free" bug?

grendello: Well spotted, I'll fix that

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -2050,10 +2049,12 @@ MonodroidRuntime::Java_mono_android_Runtime_initInternal (JNIEnv *env, jclass kl
                                                           jobjectArray assembliesJava, jint apiLevel, jboolean isEmulator,
                                                           jboolean haveSplitApks)
 {
-	char *mono_log_mask = nullptr;
+	char *mono_log_mask_raw = nullptr;
 	char *mono_log_level = nullptr;
```
jonpryor, line 2053: Shouldn't `mono_log_level` likewise be `delete[]`d?

We do have a comment saying that we can't free `mono_log_level`, as mono keeps it: https://github.com/xamarin/xamarin-android/blob/main/src/monodroid/jni/monodroid-glue.cc#L2124-L2125

However, I don't see any evidence that mono *does* keep it, e.g.

https://github.com/mono/mono/blob/b32801a63cff0561d2a327c71986ce05254c4d8b/mono/utils/mono-logger.c#L257-L278

Presumably we should be able to free `mono_log_level`?

grendello: Hmm, indeed, I have no idea why I thought it was stored

PR #6495, opened 2021-11-16
Don't count assemblies to be skipped when packaging

grendello:
Fixes: https://developercommunity.visualstudio.com/t/Xamarin-Forms-Android-App-crashes-on-sta/1578587
Context: c9270261

c9270261 added support for assembly stores which, in turn, depend on
knowing the exact number of assemblies packaged in the APK in order to
properly perform binary search on the tables of assembly name hashes.
This information is recorded by the `GeneratePackageManagerJava` task
in the `libxamarin-app.so` library on the build time and then used by
the assembly store code on the run time.

The code responsible for counting failed to skip assemblies which had
the `AndroidSkipAddToPackage` metadata, in our case the old Android
Support Library ones, replaced by AndroidX, with the metadata set on the
old assemblies by the AndroidX migration code.

Properly ignore skipped assemblies when counting.

Additionally, improve the `assembly-store-reader` utility to show the
hash tables stored in the blob as well as check whether they are
properly sorted.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -25,6 +25,34 @@ namespace Xamarin.Android.Build.Tests
 	{
 		[Test]
 		[Category ("SmokeTests")]
+		public void CheckAssemblyCounts ([Values (false, true)] bool isRelease)
+		{
+			var proj = new XamarinAndroidApplicationProject () {
+				IsRelease = isRelease,
+				EmbedAssembliesIntoApk = true,
+			};
+
+			var abis = new [] { "armeabi-v7a", "x86" };
+			proj.SetAndroidSupportedAbis (abis);
+			proj.SetProperty (proj.ActiveConfigurationProperties, "AndroidUseAssemblyStore", "True");
+
+			using (var b = CreateApkBuilder ()) {
+				Assert.IsTrue (b.Build (proj), "Build should have succeeded.");
+				string objPath = Path.Combine (Root, b.ProjectDirectory, proj.IntermediateOutputPath);
+
+				List<string> envFiles = EnvironmentHelper.GatherEnvironmentFiles (objPath, String.Join (";", abis), true);
+				EnvironmentHelper.ApplicationConfig app_config = EnvironmentHelper.ReadApplicationConfig (envFiles);
+				Assert.That (app_config, Is.Not.Null, "application_config must be present in the environment files");
+
+				string apk = Path.Combine (Root, b.ProjectDirectory, proj.OutputPath, $"{proj.PackageName}-Signed.apk");
+				var helper = new ArchiveAssemblyHelper (apk, useAssemblyStores: true);
+
+				Assert.IsTrue (app_config.number_of_assemblies_in_apk == (uint)helper.GetNumberOfAssemblies (), "Assembly count must be equal between ApplicationConfig and the archive contents");
+			}
+		}
+
+		[Test]
+		[Category ("SmokeTests")]
 		public void SmokeTestBuildWithSpecialCharacters ([Values (false, true)] bool forms)
 		{
 			var testName = "テスト";
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -24,10 +24,11 @@ namespace Xamarin.Android.Build.Tests
 	public partial class BuildTest : BaseTest
 	{
 		[Test]
+		[NonParallelizable]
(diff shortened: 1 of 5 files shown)


jonathanpeppers (approved):
LGTM, do we need a test somehow?

grendello:
I could add a test which compares the number of assemblies in the blob and the one stored in app config

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -23,6 +23,34 @@ namespace Xamarin.Android.Build.Tests
...
 	public partial class BuildTest : BaseTest
 	{
+		[Test]
+		[Category ("SmokeTests")]
+		public void CheckAssemblyCounts ([Values (false, true)] bool isRelease)
+		{
+			var proj = new XamarinAndroidApplicationProject () {
+				IsRelease = isRelease,
+				EmbedAssembliesIntoApk = true,
+			};
```
jonathanpeppers, line 33: You could make this test like this one:

https://github.com/xamarin/xamarin-android/blob/c92702619f5fabcff0ed88e09160baf9edd70f41/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest2.cs#L320-L332

Then I think it would reproduce the original bug.

grendello: Ah, great! I wasn't sure how to reproduce that, thanks :)

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/ArchiveAssemblyHelper.cs
```diff
@@ -187,6 +187,27 @@ public List<string> ListArchiveContents (string storeEntryPrefix = DefaultAssemb
...
+		{
+			List<string> contents = ListArchiveContents (assembliesRootDir, forceRefresh);
+			var dlls = contents.Where (x => x.EndsWith (".dll", StringComparison.OrdinalIgnoreCase));
+
+			if (!countAbiAssembliesOnce) {
+				return dlls.Count ();
+			}
+
+			var cache = new HashSet<string> (StringComparer.OrdinalIgnoreCase);
+			return dlls.Where (x => {
```
jonpryor, line 200: Why `dlls.Where()` + `HashSet<string>` instead of [`dlls.Select(n => Path.GetFileName(n)).Distinct(StringComparer.OrdinalIgnoreCase)`](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.distinct?view=net-5.0#System_Linq_Enumerable_Distinct__1_System_Collections_Generic_IEnumerable___0__System_Collections_Generic_IEqualityComparer___0__)?

tools/assembly-store-reader/Program.cs
```diff
@@ -50,13 +52,56 @@ static void ShowAssemblyStoreInfo (string storePath)
...
+			if (indexStore == null) {
+				Console.WriteLine ("Index store not found in the set, not showing hash tables.");
+				return;
+			}
+
+			Console.WriteLine ("Hash tables:");
+			WriteHashTable ("32", indexStore.GlobalIndex32, "x08");
+			WriteHashTable ("64", indexStore.GlobalIndex64, "x016");
+
+			void WriteHashTable (string bitness, List<AssemblyStoreHashEntry> table, string hashFormat)
```
jonpryor, line 77: Should `bitness` really be a string and not an `int`?  (Crazy town: should it be an enum?)

tools/assembly-store-reader/Program.cs
```diff
@@ -50,13 +52,56 @@ static void ShowAssemblyStoreInfo (string storePath)
...
+			}
+
 			void WriteOptionalDataLine (string label, uint offset, uint size)
 			{
 				Console.Write ($"{infoIndent}{label}: ");
 				if (offset == 0) {
 					Console.WriteLine ("absent");
 				} else {
-					Console.WriteLine ("offset == {offset}; size == {size}");
+					Console.WriteLine ($"offset == {offset}; size == {size}");
```
jonpryor, line 104: Doh!

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/ArchiveAssemblyHelper.cs
grendello: I don't like LINQ too much and the `HashSet<string>` route felt faster (one pass over the set instead of two)

tools/assembly-store-reader/Program.cs
grendello: Nah, it's just a label :)

tools/assembly-store-reader/Program.cs
grendello: Indeed :)

PR #6496, opened 2021-11-16
[xaprepare] Fix Unicode detection when running on dotnet

grendello:
37dbc459 made `xaprepare` build and run using dotnet which
caused `xaprepare` to misdetect whether the current terminal
supports Unicode and, thus, not use emoji.  This cannot be! :D

The problem lies in the fact that dotnet sets `Console.OutputEncoding`
to an internal class wrapping `System.Encoding` and our check for
the known Unicode encoding types fails.  Before looking at the type
of `OutputEncoding`, check whether the encoding name contains the
word "Unicode".  This fixes the check on dotnet.


pjcollins (approved):

grendello:
Merging since `xaprepare` worked on all platforms

PR #6506, opened 2021-11-18
Stop invoking nuget.exe in favor of `dotnet nuget`

grendello:
Nuget is still downloaded because of X.A.B.T. tests which still use it


jonathanpeppers (approved):
This looks like it works now, does it still need to be a draft?

grendello:
> This looks like it works now, does it still need to be a draft?

There's still the usage in `ProjectTools` I wasn't sure how to fix

jonpryor:
Commit message?

```markdown
[build] Use `dotnet nuget`, not `nuget.exe` (#6506)

Instead of using a downloaded `nuget.exe` to restore
`Xamarin.Android.sln` during `make prepare` /
`msbuild /t:Prepare`, instead use `dotnet nuget`.

The `Xamarin.Android.Build.Tests.dll` unit test assembly still
requires `nuget.exe` in order to test `packages.config`-based apps.
Use NuGet v6.0.0 instead of v4.9.4.

Co-authored-by: Dean Ellis <dellis1972@googlemail.com>
```

PR #6507, opened 2021-11-18
[net6] Don't blindly load Mono components

grendello:
Whenever a NET6 Xamarin.Android app starts, MonoVM will probe for a
number of Mono components:

    11-16 14:23:24.805 10791 10791 I monodroid-assembly: Trying to load shared library '/data/app/~~5lwn5C40pkRva2-L4kilZA==/com.microsoft.net6.helloandroid-MrYUTOqX7yZPkmtLOPnuuw==/split_config.arm64_v8a.apk!/lib/arm64-v8a/libmono-component-debugger.so'
    11-16 14:23:24.806 10791 10791 I monodroid-assembly: Failed to load shared library '/data/app/~~5lwn5C40pkRva2-L4kilZA==/com.microsoft.net6.helloandroid-MrYUTOqX7yZPkmtLOPnuuw==/split_config.arm64_v8a.apk!/lib/arm64-v8a/libmono-component-debugger.so'.
                                                         dlopen failed: library "/data/app/~~5lwn5C40pkRva2-L4kilZA==/com.microsoft.net6.helloandroid-MrYUTOqX7yZPkmtLOPnuuw==/split_config.arm64_v8a.apk!/lib/arm64-v8a/libmono-component-debugger.so" not found

The above is repeated for, currently, 3 components.  Together, the
failed load attempts cost us around 1ms of startup time (on Pixel 3 XL,
the amount of time will depend on the speed of device's storage and CPU)

Since we know at the build time which components are included, we can
optimize the startup process by recording a flag indicating which
components can be loaded successfully.

Add a `mono_components_mask` field to `ApplicationConfig` which is set
to a bitmask indicating which components are packaged.  On application
startup, whenever `monodroid_dlopen` is called, we has the name of the
library passed to us by Mono and see if it matches one of the known
hashes for the various components.  If yes, we consult the mask stored
at build time and attempt to load the component only if its bit is set.

The above check is performed **only** during application startup since
that's when Mono probes for the components and it would be a waste of
time later in the application life.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
@@ -28,6 +28,8 @@ namespace Xamarin.Android.Tasks
 		[Required]
 		public ITaskItem[] ResolvedUserAssemblies { get; set; }
 
+		public ITaskItem[] MonoComponents { get; set; }
+
 		public ITaskItem[] SatelliteAssemblies { get; set; }
 
 		public bool UseAssemblyStore { get; set; }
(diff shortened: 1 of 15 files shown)


jonathanpeppers (approved):
The MSBuild changes LGTM 👍

PR #6515, opened 2021-11-20
Fix `monodroid_dlopen` to load arbitrary DSOs

grendello:
When the `monodroid_dlopen` is called by Mono, we search for libraries
in a handful of known locations (either app private library directory or
the base.apk archive, possibly override directories for Debug builds)
but if these fail to find the library, we give up and return `NULL`.
For that reason, a p/invoke into e.g. `libEGL.so` will "fail" in that we
will report the library as not found (since it's a system library, not
one private to the app) but the p/invoke may still work since a system
library may already be loaded, so `dlsym` will work despite the load
"failure".

Fix this by calling `AndroidSystem::load_dso` if we fail to locate the
shared library in the app directories, passing it the name sent to us by
Mono.

### DIFF
--- a/src/monodroid/jni/monodroid-glue.cc
+++ b/src/monodroid/jni/monodroid-glue.cc
@@ -1344,6 +1344,7 @@ MonodroidRuntime::monodroid_dlopen (const char *name, int flags, char **err)
 	}
 
 	if (!utils.ends_with (name, ".dll.so")) {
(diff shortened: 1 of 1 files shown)


jonpryor:
Reported at: https://discord.com/channels/732297728826277939/732297837953679412/911401847397429288

PR #6522, opened 2021-11-24
Hash names of all the packaged DSOs

grendello:
Use the hash to map any form of DSO name sent to us by Mono (we generate
hashes for several mutations) to find a cache entry for that DSO. Cache
includes the actual DSO name, whether it should be ignored when
requested (true for empty AOT DSOs) and the cached module handle.

This removes a bit of string processing from both startup and the
application run time.

This commit started with the goal to implement the above, but during 
implementation of the DSO cache I hit known limitations of the native
assembler generator code which, until this commit weren't a problem
(mostly related to hard-coded structure and array alignment).  So now
it includes a new implementation of the said generator, this time it 
fully implements the structure and symbol alignment calculation for
all the supported architectures.  Also added is the ability to generate
assembler code from managed structures, using reflection, without the need
of manually written code.  It also fixes previously unnoticed issue which
made typemap structures not aligned properly, which may have made them 
slightly slower than necessary

The implemented changes improve startup of both plain XA and MAUI applications,
measured on Pixel 3 XL running Android 12:

# Plain XA application

## AOT, Displayed time

| Before  | After   | Δ        | Notes                                 |
| ------- | ------- | -------- | ------------------------------------- |
| 252.400 | 245.000 | -2.93% ✓ | defaults; profiled AOT; 32-bit build  |
| 253.500 | 247.000 | -2.56% ✓ | defaults; profiled AOT; 64-bit build  |
| 251.000 | 247.100 | -1.55% ✓ | defaults; full AOT; 64-bit build      |
| 255.700 | 252.500 | -1.25% ✓ | defaults; full AOT; 32-bit build      |
| 257.000 | 256.400 | -0.23% ✓ | defaults; full AOT+LLVM; 64-bit build |
| 254.300 | 258.000 | +1.43% ✗ | defaults; full AOT+LLVM; 32-bit build |

## AOT, Total Initialization Time

| Before | After  | Δ        | Notes                                 |
| ------ | ------ | -------- | ------------------------------------- |
| 41.684 | 38.677 | -7.21% ✓ | defaults; profiled AOT; 64-bit build  |
| 41.357 | 38.727 | -6.36% ✓ | defaults; profiled AOT; 32-bit build  |
| 47.356 | 43.911 | -7.27% ✓ | defaults; full AOT; 32-bit build      |
| 47.256 | 44.356 | -6.14% ✓ | defaults; full AOT; 64-bit build      |
| 49.223 | 45.813 | -6.93% ✓ | defaults; full AOT+LLVM; 64-bit build |
| 48.755 | 46.098 | -5.45% ✓ | defaults; full AOT+LLVM; 32-bit build |

## No AOT, Displayed time

| Before  | After   | Δ        | Notes                                          |
| ------- | ------- | -------- | ---------------------------------------------- |
| 309.900 | 305.900 | -1.29% ✓ | preload disabled; no compression; 32-bit build |
| 317.900 | 309.900 | -2.52% ✓ | preload disabled; no compression; 64-bit build |
| 311.500 | 311.600 | +0.03% ✗ | preload enabled; 64-bit build                  |
| 314.800 | 311.600 | -1.02% ✓ | preload disabled; 64-bit build                 |
| 309.200 | 312.500 | +1.06% ✗ | preload enabled; no compression; 64-bit build  |
| 312.600 | 313.300 | +0.22% ✗ | preload disabled; 32-bit build                 |
| 314.100 | 315.300 | +0.38% ✗ | preload enabled; 32-bit build                  |
| 312.300 | 315.700 | +1.08% ✗ | preload enabled; no compression; 32-bit build  |

## No AOT, Total Init Time

| Before | After  | Δ        | Notes                                          |
| ------ | ------ | -------- | ---------------------------------------------- |
| 61.433 | 60.357 | -1.75% ✓ | preload disabled; no compression; 32-bit build |
| 60.245 | 60.521 | +0.46% ✗ | preload enabled; no compression; 32-bit build  |
| 60.216 | 60.801 | +0.96% ✗ | preload disabled; no compression; 64-bit build |
| 60.261 | 61.141 | +1.44% ✗ | preload enabled; no compression; 64-bit build  |
| 62.366 | 61.636 | -1.17% ✓ | preload enabled; 64-bit build                  |
| 61.924 | 61.776 | -0.24% ✓ | preload disabled; 64-bit build                 |
| 61.754 | 62.007 | +0.41% ✗ | preload enabled; 32-bit build                  |
| 62.529 | 62.792 | +0.42% ✗ | preload disabled; 32-bit build                 |

# MAUI Hello World

## AOT, Displayed time

| Before   | After    | Δ        | Notes                                 |
| -------- | -------- | -------- | ------------------------------------- |
| 868.900  | 831.700  | -4.28% ✓ | defaults; profiled AOT; 64-bit build  |
| 868.900  | 839.200  | -3.42% ✓ | defaults; profiled AOT; 32-bit build  |
| 901.900  | 868.800  | -3.67% ✓ | defaults; full AOT; 64-bit build      |
| 914.300  | 876.600  | -4.12% ✓ | defaults; full AOT+LLVM; 64-bit build |
| 904.300  | 877.500  | -2.96% ✓ | defaults; full AOT; 32-bit build      |
| 913.400  | 879.000  | -3.77% ✓ | defaults; full AOT+LLVM; 32-bit build |

## AOT, Total Init Time

| Before  | After   | Δ         | Notes                                 |
| ------- | ------- | --------- | ------------------------------------- |
| 114.953 | 95.573  | -16.86% ✓ | defaults; profiled AOT; 32-bit build  |
| 115.280 | 95.971  | -16.75% ✓ | defaults; profiled AOT; 64-bit build  |
| 205.505 | 172.190 | -16.21% ✓ | defaults; full AOT; 64-bit build      |
| 206.151 | 172.892 | -16.13% ✓ | defaults; full AOT; 32-bit build      |
| 215.617 | 181.895 | -15.64% ✓ | defaults; full AOT+LLVM; 64-bit build |
| 216.313 | 181.980 | -15.87% ✓ | defaults; full AOT+LLVM; 32-bit build |

## No AOT, Displayed time

| Before   | After    | Δ        | Notes                                          |
| -------- | -------- | -------- | ---------------------------------------------- |
| 1152.800 | 1145.200 | -0.66% ✓ | preload enabled; 64-bit build; no compression  |
| 1164.000 | 1153.100 | -0.94% ✓ | preload enabled; 64-bit build                  |
| 1141.500 | 1155.200 | +1.19% ✗ | preload disabled; 64-bit build; no compression |
| 1156.300 | 1159.500 | +0.28% ✗ | preload disabled; 64-bit build                 |
| 1230.500 | 1231.900 | +0.11% ✗ | preload enabled; 32-bit build; no compression  |
| 1235.600 | 1238.300 | +0.22% ✗ | preload disabled; 32-bit build; no compression |
| 1241.400 | 1241.900 | +0.04% ✗ | preload enabled; 32-bit build                  |
| 1240.500 | 1253.100 | +1.01% ✗ | preload disabled; 32-bit build                 |

## No AOT, Total Init Time

| Before | After  | Δ        | Notes                                          |
| ------ | ------ | -------- | ---------------------------------------------- |
| 73.570 | 73.827 | +0.35% ✗ | preload disabled; 64-bit build; no compression |
| 73.512 | 73.962 | +0.61% ✗ | preload enabled; 64-bit build; no compression  |
| 76.792 | 76.771 | -0.03% ✓ | preload enabled; 64-bit build                  |
| 77.382 | 77.109 | -0.35% ✓ | preload disabled; 64-bit build                 |
| 84.462 | 84.566 | +0.12% ✗ | preload disabled; 32-bit build; no compression |
| 84.051 | 84.725 | +0.80% ✗ | preload enabled; 32-bit build; no compression  |
| 88.051 | 88.015 | -0.04% ✓ | preload enabled; 32-bit build                  |
| 87.596 | 88.445 | +0.96% ✗ | preload disabled; 32-bit build                 |


src/monodroid/jni/application_dso_stub.cc
```diff
@@ -117,3 +119,22 @@ AssemblyStoreRuntimeData assembly_stores[] = {
 		.assemblies = nullptr,
 	},
 };
+
+constexpr char fake_dso_name[] = "libaot-Some.Assembly.dll.so";
```
jonpryor, line 123: Why have *any* `DSOCacheEntry` values in the "stub" build?  Why not just a zero-length array?  (Or does something require non-zero length arrays?)

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblyGenerator.cs
```diff
@@ -0,0 +1,1223 @@
...
+				WriteStringSection (Output, AutomaticStringSection);
+
+				foreach (LabeledStringSymbol sym in stringTable) {
+					WriteSymbol (Output, sym, local: true, ownSection: false);
+				}
+
+				Output.WriteLine ();
+			}
+
+			WriteDirective (".ident", QuoteString ($"Xamarin.Android {XABuildConfig.XamarinAndroidVersion}"));
```
jonpryor, line 227: Why do we need to start emitting `ident "Xamarin.Android 12.2.99"`?

What do we start doing when .NET 7+ is our primary distribution vehicle and not Classic Xamarin.Android? Presumably we'll stop bumping or otherwise remove `$(ProductVersion)` *eventually*, and then use `$(AndroidPackVersion)` instead.  (Presumably?). Should use use `$(ProductVersion)` for both .NET 6+ & Classic now, `$(AndroidPackVersion)` for both now, or selectively use one depending on which is being used?

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -142,20 +144,20 @@ public override bool RunTask ()
 			return !Log.HasLoggedErrors;
 		}
 
-		static internal NativeAssemblerTargetProvider GetAssemblyTargetProvider (string abi)
+		static internal AndroidTargetArch GetAndroidTargetArchForAbi (string abi)
```
jonpryor, line 147: This feels like an "odd" type to contain this method.  Wouldn't `NdkTools` make more sense?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
```diff
@@ -33,15 +34,16 @@ public sealed class ApplicationConfig
 			public uint   system_property_count;
 			public uint   number_of_assemblies_in_apk;
 			public uint   bundled_assembly_name_width;
-			public uint   number_of_assembly_blobs;
+			public uint   number_of_assembly_store_files;
```
jonpryor, line 37: Why rename "blobs" to "files"?  Clarity?  Other?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
```diff
@@ -33,15 +34,16 @@ public sealed class ApplicationConfig
...
 			public uint   mono_components_mask;
 			public string android_package_name;
 		};
-		const uint ApplicationConfigFieldCount = 18;
+		const uint ApplicationConfigFieldCount = 19;
 
 		static readonly object ndkInitLock = new object ();
 		static readonly char[] readElfFieldSeparator = new [] { ' ', '\t' };
-		static readonly Regex stringLabelRegex = new Regex ("^\\.L\\.env\\.str\\.[0-9]+:", RegexOptions.Compiled);
+		static readonly Regex stringLabelRegex = new Regex ("^\\.L\\.autostr\\.[0-9]+:", RegexOptions.Compiled);
```
jonpryor, line 46: I have no idea what this means or implies.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
```diff
@@ -521,9 +536,27 @@ static byte ConvertFieldToByte (string fieldName, string envFile, int fileLine,
...
-			Assert.IsTrue (Byte.TryParse (value, out fv), $"Field '{fieldName}' in {envFile}:{fileLine} is not a valid uint8_t value (not a valid integer)");
+			Assert.IsTrue (TryParseInteger (value, out fv), $"Field '{fieldName}' in {envFile}:{fileLine} is not a valid uint8_t value (not a valid integer)");
 
 			return fv;
 		}
+
+		static bool TryParseInteger (string value, out uint fv)
+		{
+			if (value.StartsWith ("0x", StringComparison.Ordinal)) {
+				return UInt32.TryParse (value.Substring (2), NumberStyles.AllowHexSpecifier, null, out fv);
```
jonpryor, line 547: Shouldn't this use `provider:CultureInfo.InvariantCulture` instead of `provider:null`? From the [`uint.Parse()` docs](https://docs.microsoft.com/en-us/dotnet/api/system.uint32.tryparse?view=net-6.0#system-uint32-tryparse(system-string-system-globalization-numberstyles-system-iformatprovider-system-uint32@)):

> If `provider` is `null`, the `NumberFormatInfo` object for the current culture is used.

Maybe this is more of a problem for floating-point values than integers, but I'm not sure we want a "current culture" integer parse here & elsewhere?

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfig.cs
```diff
@@ -0,0 +1,32 @@
+using System;
+
+namespace Xamarin.Android.Tasks
+{
+	// Order of fields and their types must correspond *exactly* to that in
```
jonpryor, line 5: May need to elaborate on the meaning of "exactly".  "Exactly" at the ABI level?  Other?

I'm not even sure what C++ `bool` *is* at the ABI level (i.e. what is `sizeof(bool)` in C++? I don't know offhand).  The header uses `uint32_t` while C# uses `uint`, which is a 32-bit int, so these are the same in terms of ABI, but only "semantically the same" at the API level.

Then there's `string android_package_name`, and I don't know what brain bending I should use to associate C# `string` with C++ `const char *`.

("ABI -ish"?)

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
```diff
@@ -16,8 +21,77 @@ enum MonoComponent
 		Tracing   = 0x04,
 	}
 
-	class ApplicationConfigNativeAssemblyGenerator : NativeAssemblyGenerator
+	class ApplicationConfigNativeAssemblyGenerator : NativeAssemblyComposer
```
jonpryor, line 24: Possible crazy talk, but should we change the suffix if this & other classes from `Generator` to `Composer` to match the new base class?  Why did the base class name change?

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
```diff
@@ -16,8 +21,77 @@ enum MonoComponent
...
+			public bool ignore;
+
+			[NativeAssemblerString (UsesDataProvider = true)]
+			public string name;
+			public IntPtr handle = IntPtr.Zero;
+		}
+
+		// Order of fields and their type must correspond *exactly* to that in
+		// src/monodroid/jni/xamarin-app.hh AssemblyStoreSingleAssemblyRuntimeData structure
+		sealed class AssemblyStoreSingleAssemblyRuntimeData
```
jonpryor, line 64: What is the *relationship* here?  Do these types need to be kept in-sync *manually*?  Is a tool involved?  Something else?

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
```diff
@@ -16,8 +21,77 @@ enum MonoComponent
...
+		// src/monodroid/jni/xamarin-app.hh XamarinAndroidBundledAssembly structure
+		sealed class XamarinAndroidBundledAssembly
+		{
+			public int    apk_fd;
+			public uint   data_offset;
+			public uint   data_size;
+			public IntPtr data;
+			public uint   name_length;
+
+			[NativeAssemblerString (PointerToSymbol = true)]
```
jonpryor, line 91: /me sees "pointer to symbol", and is not sure if he should flee, or *flee very quickly*.

What does this *mean*?

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
```diff
@@ -39,287 +113,246 @@ class ApplicationConfigNativeAssemblyGenerator : NativeAssemblyGenerator
...
+				bundled_assembly_name_width = (uint)BundledAssemblyNameWidth,
+				number_of_assembly_store_files = (uint)NumberOfAssemblyStoresInApks,
+				number_of_dso_cache_entries = dsoCacheEntries,
+				mono_components_mask = (uint)MonoComponents,
+				android_package_name = AndroidPackageName,
+			};
+
+			NativeAssemblyGenerator.StructureWriteContext structStatus = generator.StartStructure ();
+			generator.WriteStructure (structStatus, application_config);
+            generator.WriteSymbol (structStatus, "application_config", local: false);
```
jonpryor, line 173: Spaces and not tabs used for indentation on this line.

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
```diff
@@ -39,287 +113,246 @@ class ApplicationConfigNativeAssemblyGenerator : NativeAssemblyGenerator
...
-			List<string> name_labels = null;
+			List<NativeAssemblyGenerator.LabeledSymbol>? name_labels = null;
 			if (!HaveAssemblyStore) {
-				name_labels = new List<string> ();
+				name_labels = new List<NativeAssemblyGenerator.LabeledSymbol> ();
 				for (int i = 0; i < NumberOfAssembliesInApk; i++) {
-					string bufferLabel = GetBufferLabel ();
-					WriteBufferAllocation (output, bufferLabel, (uint)BundledAssemblyNameWidth);
-					name_labels.Add (bufferLabel);
+					name_labels.Add (generator.WriteEmptyBuffer ((uint)BundledAssemblyNameWidth, "env.buf"));
```
jonpryor, line 310: Another one for the naming $DEITYs, "write empty buffer"?  I'm not at all sure what the semantics are based on just looking at this line.

I wonder if `BeginBuffer(…)` would be better?  Unless this is asserting that the buffer is in fact empty, but why would we ever have known empty buffers?  (I have not yet read what `WriteEmptyBuffer()` *does*.  I'm not sure I'll understand what it does when I get there.)

src/Xamarin.Android.Build.Tasks/Utilities/ELFHelper.cs
```diff
@@ -7,18 +8,28 @@
 using Microsoft.Android.Build.Tasks;
 using Microsoft.Build.Utilities;
 
+using ELFSymbolType = global::ELFSharp.ELF.Sections.SymbolType;
+
 namespace Xamarin.Android.Tasks
 {
 	static class ELFHelper
 	{
+		static ConcurrentDictionary<string, bool> checkedDSOs = new ConcurrentDictionary<string, bool> (StringComparer.OrdinalIgnoreCase);
```
jonpryor, line 17: Do we *actually* want static mutable data?  We've previously had bugs due to static mutable data in this assembly: https://github.com/xamarin/xamarin-android/issues/5996

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/Arm32NativeAssemblyGenerator.cs
```diff
@@ -0,0 +1,94 @@
...
+
+			return GetNativeTypeSize<T> ();
+		}
+
+		protected override void ConfigureTypeMappings (Dictionary<Type, NativeType?> mapping)
+		{
+			base.ConfigureTypeMappings (mapping);
+
+			// Alignments and sizes as per https://github.com/ARM-software/abi-aa/blob/320a56971fdcba282b7001cf4b84abb4fd993131/aapcs32/aapcs32.rst#fundamental-data-types
+			ConfigureTypeMapping<short>	 (".short", size: 2, alignment: 2);
```
jonpryor, line 55: Where do the *names* come from? I don't see `.short` or `.long` in the referenced URL.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/ArmNativeAssemblyGenerator.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Collections.Generic;
+using System.IO;
+
+namespace Xamarin.Android.Tasks
+{
+	abstract class ArmNativeAssemblyGenerator : NativeAssemblyGenerator
+	{
+		protected abstract string ArchName { get; }
+
+        protected ArmNativeAssemblyGenerator (StreamWriter output, string fileName)
```
jonpryor, line 11: Spaces not tabs used for indentation.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblerContextDataProvider.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace Xamarin.Android.Tasks
+{
+	class NativeAssemblerStructContextDataProvider
+	{
+		public virtual string GetComment (object data, string fieldName)
+		{
+			return String.Empty;
+		}
+
+		public virtual uint GetMaxInlineWidth (object data, string fieldName)
```
jonpryor, line 12: What is an "inline" width?

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblerAttribute.cs
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Xamarin.Android.Tasks
+{
+	[AttributeUsage (AttributeTargets.Field | AttributeTargets.Property, Inherited = true)]
+	class NativeAssemblerAttribute : Attribute
+	{
+		public bool Ignore			 { get; set; }
+		public bool UsesDataProvider { get; set; }
+		public string? Comment		 { get; set; }
+		public string? Name			 { get; set; }
```
jonpryor, line 11: You should use spaces after non-whitespace characters to ensure appropriate vertical alignment.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblyGenerator.cs
```diff
@@ -0,0 +1,1223 @@
...
+
+using Xamarin.Android.Tools;
+
+namespace Xamarin.Android.Tasks
+{
+	abstract class NativeAssemblyGenerator
+	{
+		protected const string AutomaticStringSection = "autostr";
+
+		internal class NativeType
```
jonpryor, line 15: Given that the declaring class is `internal`, do these & other types need to *also* be `internal`, or should they just be `public`?

src/Xamarin.Android.Build.Tasks/Utilities/TypeMappingDebugNativeAssemblyGenerator.cs
```diff
@@ -28,115 +81,98 @@ public TypeMappingDebugNativeAssemblyGenerator (NativeAssemblerTargetProvider ta
...
 
-		void WriteSharedBits (StreamWriter output, bool haveJavaToManaged, bool haveManagedToJava)
+		void WriteSharedBits (NativeAssemblyGenerator generator, StreamWriter output, bool haveJavaToManaged, bool haveManagedToJava)
 		{
-			string label;
-
 			if (haveJavaToManaged) {
-				WriteCommentLine (output, "Java type names: START");
+				generator.WriteStringSection (output, "java_type_names");
+				generator.WriteCommentLine (output, "Java type names: START", useBlockComment: true);
```
jonpryor, line 160: Why does this block need to use "block comments" instead of line comments?  This method appears to be the only use of `useBlockComment: true`.

I also don't understand how you *terminate* a block comment.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/Arm32NativeAssemblyGenerator.cs
```diff
@@ -0,0 +1,94 @@
...
+			ConfigureTypeMapping<nuint>	 (".long",	size: 4, alignment: 4);
+			ConfigureTypeMapping<IntPtr> (".long",	size: 4, alignment: 4);
+		}
+
+		public override void WriteFileTop ()
+		{
+			base.WriteFileTop ();
+
+			WriteDirective (".syntax", "unified");
+			WriteDirectiveWithComment (".eabi_attribute", "Tag_conformance", 67, QuoteString ("2.09"));
```
jonpryor, line 73: Where do *these* values come from?

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblyGenerator.cs
```diff
@@ -0,0 +1,1223 @@
...
+				throw new InvalidOperationException ("Structure context doesn't refer to array of structures, did you use StartStructureArray()?");
+			}
+
+			var ret = StartStructure ();
+			status.ArrayOfStructures.Add (ret);
+
+			return ret;
+		}
+
+		public void WriteStructure<T> (StructureWriteContext status, T[] collection) where T: class
```
jonpryor, line 383: I find it slightly amusing that a "structure" must be a `class`, but I can't think of a better name either…

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/SymbolType.cs
```diff
@@ -0,0 +1,13 @@
+namespace Xamarin.Android.Tasks
+{
+	enum SymbolType
+	{
+		Function,
+		GnuIndirectFunction,
```
jonpryor, line 6: Why include enum members that aren't really *used* anywhere? `WriteSymbolType()` "supports" it, but I don't see anywhere that uses it.  Why have this member at all?

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblyGenerator.cs
```diff
@@ -0,0 +1,1223 @@
...
+
+		public void WriteSection (TextWriter writer, string name, SectionFlags flags = SectionFlags.None, SectionType type = SectionType.None, string? flagSpecificArgs = null, string? numberOrCustomFlag = null, string? numberOrCustomType = null)
+		{
+			writer.WriteLine ();
+			writer.Write ($"{Indent}.section{Indent}{name}");
+			if (flags != SectionFlags.None) {
+				writer.Write (", \"");
+
+				var list = new StringBuilder ();
+				if ((flags & SectionFlags.Allocatable) == SectionFlags.Allocatable) {
```
jonpryor, line 1019: Have you considered instead using `if (flags.HasFlag(SectionFlags.Allocatable)) {…}`?

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/SectionFlags.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+
+namespace Xamarin.Android.Tasks
+{
+	[Flags]
+	enum SectionFlags
+	{
+		None				   = 0,
+		Allocatable			   = 1 << 0,
```
jonpryor, line 9: Where'd these values come from?

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/X86NativeAssemblyGenerator.cs
```diff
@@ -0,0 +1,26 @@
...
+        protected X86NativeAssemblyGenerator (StreamWriter output, string fileName)
+			: base (output, fileName)
+		{}
+
+		protected override void ConfigureTypeMappings (Dictionary<Type, NativeType?> mapping)
+		{
+			base.ConfigureTypeMappings (mapping);
+
+			ConfigureTypeMapping<short>	 (".short", size: 2, alignment: 2);
+			ConfigureTypeMapping<ushort> (".short", size: 2, alignment: 2);
```
jonpryor, line 20: Tabs after non-whitespace characters redux: this diff looks nice if I copy the `.patch` into an editor with 4 space tabs.  It doesn't look as nice in the GitHub web UI, which uses 8 space tabs.

You should use *only spaces* on a line once a non-whitespace character has been used.

src/Xamarin.Android.Build.Tasks/Utilities/TypeMappingDebugNativeAssemblyGenerator.cs
jonpryor: Part of why I ask is that *support* for "block comments" appears to go *everywhere*, and it feels like things would get much simpler if you removed support for block comments, given that it's really only used in one place.

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -565,11 +565,18 @@ EmbeddedAssemblies::binary_search (const Key *key, const Entry *base, size_t nme
 		exit (FATAL_EXIT_MISSING_ASSEMBLY);
 	}
 
-	constexpr size_t size = sizeof(Entry);
+	[[maybe_unused]]
+	size_t size;
+
+	if constexpr (use_precalculated_size) {
+		size = precalculated_size;
+		log_warn (LOG_ASSEMBLY, "Pre-calculated entry size = %u", size);
```
jonpryor, line 573: Should this be `log_warn()` and not `log_info()`?

src/monodroid/jni/application_dso_stub.cc
grendello: I'm using disassembly of the stub to make sure our managed generator does the right job, for that I need it to have "valid" content.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblyGenerator.cs
grendello: I don't really care which version we use, I wanted to have `Xamarin.Android ${VERSION}` in there, it might be helpful in diagnosing issues to know which XA built the binary. We can put git commit hash there, whatever you think makes the most sense.

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
grendello: No, because it's not an NDK target arch, but the arch as used by the Android OS runtime. NDK uses different names for ARM targets, and the forms used here don't exist in NDK at all.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
grendello: I thought we wanted to drop the word `blob` usage. And I didn't want to use `stores` because it didn't sound right to me and since they are files... :)

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
grendello: The new assembly generator now automatically handles assigning unique labels to all strings used throughout the code. It then outputs them at the end of build in a single section. Strings used to have labels like `L.env.1.str` before (an artifact from the times when only environment variables were output, thus `env`) but now we have more strings which are NOT environment variable names or values, thus the change to `autostr` - for "automatically handled strings". `autostr` is also the name of section where the strings are kept.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
grendello: The only problem can be parsing the thousands separator, but considering that this is a hexadecimal number that we're parsing, it shouldn't matter either. The only thing `InvariantCulture` may give us is, perhaps, slightly faster parsing.

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfig.cs
grendello: This is all about declaration order of fields and their bit size. I will add a comment with the mapping.

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
grendello: Because the classes that derive from `NativeAssemblyComposer` no longer have any code that directly generates the assembler code. They **compose** the shape of the assembly source file, giving it structure, and call high-level methods on `NativeAssemblyGenerator` to generate actual code (or output strings, assign labels etc). They no longer deal with any low-level **assembler** code output directly. I didn't rename them because from the "user's" point of view (that is, the Tasks that call the `NativeAssemblyComposer` derivatives) the `ApplicationConfigNativeAssemblyGenerator` is **still** a generator that outputs assembly code. The base class of `ApplicationConfigNativeAssemblyGenerator` is a technicality that's of no interest to the caller, and the `Generator` suffix describes function of the class, not its place in some hierarchy which is an internal design detail.

The goal of the redesign was to make modifying the `*Generator` classes easier for people not familiar with native assembly.

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
grendello: Nothing has changed in this regard yet. Still maintained manually.

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
grendello: I can rename it to `PointerToNativeSymbolInELFBinary` if you think it's better :) I'm not sure there's a need to explain what a symbol in native code is, and what is a pointer to it. I'd love to assume a basic familiarity of underlying technologies of anyone modifying this code. What name would you like to see here?

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
grendello: I'm really not sure what you're asking for here. If one doesn't know what the buffer does in this context, then I don't want them to touch the code. This is the lowest level generator for a specific set of specific structures that describe the entirety of application config, and modifying the code **requires** understanding what the structures are on the native side. Once that is understood, the terms used will make sense. The buffer is in fact empty, because it just allocates space in the output DSO to be filled at the run time. In this case, the code is used when assembly stores are off and we read the bundled assemblies from the apk, filling in this structure:

```c++

struct XamarinAndroidBundledAssembly final
{
        int32_t  apk_fd;
        uint32_t data_offset;
        uint32_t data_size;
        uint8_t *data;
        uint32_t name_length;
        char    *name;
};
```

the buffer is for the assembly name (as suggested by `BundledAssemblyNameWidth` in the method invocation) and has to be there because `name` has to point to something, we don't allocate data directly. All that information is there, in source code, and I do expect anyone modifying this code to familiarize themselves with the native side...

I'm **really** not sure how do you want to explain all that in a method name?

src/Xamarin.Android.Build.Tasks/Utilities/ELFHelper.cs
grendello: `ConcurrentDictionary` is designed for this kind of sharing scenario, it should be fine.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/Arm32NativeAssemblyGenerator.cs
grendello: The URL describes alignments and sizes of native architectural types, not directive names of the assembler. Directive names are described in https://sourceware.org/binutils/docs-2.37/as/index.html.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblerContextDataProvider.cs
grendello: Get maximum width of data buffer allocated inline (that is not pointed to)

src/Xamarin.Android.Build.Tasks/Utilities/TypeMappingDebugNativeAssemblyGenerator.cs
grendello: Block comments look better sometimes - I care about that when I have to read the generated code, I want the code to be readable and nicely formatted. And the fact that they aren't used everywhere right now, doesn't mean they won't be used somewhere in the future when we add some code.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/Arm32NativeAssemblyGenerator.cs
grendello: From assembly generated by g++ and clang, somewhat described here https://sourceware.org/binutils/docs-2.37/as/ARM-Directives.html#ARM-Directives
Not all of them apply to data-only DSOs, but I want to match output of the compilers.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/SymbolType.cs
grendello: For completeness. We don't generate executable code but I didn't see harm in naming all the possible symbol (and section) types, if only to show that the code is backed by a thorough understanding of the topic.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblyGenerator.cs
grendello: Never liked it... an unnecessary function call. Will change.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/SectionFlags.cs
grendello: https://sourceware.org/binutils/docs-2.37/as/Section.html#Section

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblyGenerator.cs
jonpryor: I kinda like the `xamarin/xamarin-android@COMMIT-HASH` idea, actually… Especially since it's "product name agnostic", and thus avoids the entire question in the first place.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
jonpryor: TIL! Thanks!

src/Xamarin.Android.Build.Tasks/Utilities/ELFHelper.cs
jonpryor: Respectfully disagree?  `ConcurrentDictionary` is designed to be thread-safe.  That doesn't describe or answer interactions with AppDomains and the IDE and user behavior and caching.

For example -- untested, but this "feels" right -- assume you have a Solution with two projects, `App.dll` and `Lib.dll`.

Initially, `Lib.dll` is *empty*: no classes, just e.g. `@(EmbeddedResource)` or whatever.

AOT is enabled.

At this point, AOT'ing `Lib.dll` *should* result in `Lib.dll` having *no* AOT members, as it's empty.

Developer updates `Lib.dll` so that it now contains types & methods.  Now, AOT'ing `Lib.dll` *should* result in a non-empty `libaot-Lib.dll.so` file, as it contains code.

How does `ELFHelper.checkedDSOs` get updated in this scenario?

If you have a separate process per build, it's moot: you have an empty `ELFHelper.checkedDSOs` dict at the beginning of every build.

*Within the IDE*, ***I don't know*** what will happen.  Historically (?), `Xamarin.Android.Build.Tasks.dll` is kept "alive" once loaded; *I imagine* that *within the IDE*, `ELFHelper.checkedDSOs` will remain loaded once set, *possibly across projects*.  (Again, no idea.  @jonathanpeppers? @dellis1972?)  The lack of knowledge is concerning, regardless.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/Arm32NativeAssemblyGenerator.cs
jonpryor: Thank you for the link.  Pity it doesn't mention `.xword` though.  And `.long` says "see `.int`." `.int` doesn't mention how big it is.  In short, there's no way to independently verify that these directives are correct without having some knowledge of the target assembler, which isn't me. :-/

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblyGenerator.cs
grendello: I think it still should have some "product" name, not just the hash. `Xamarin.Android [HASH]`?

src/Xamarin.Android.Build.Tasks/Utilities/ELFHelper.cs
grendello: Hm, that scenario might be valid. I've never thought about the IDE in this context.

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/Arm32NativeAssemblyGenerator.cs
grendello: `.xword` is for 64-bit machines (https://sourceware.org/binutils/docs-2.37/as/AArch64-Directives.html#AArch64-Directives), 32-bit machines would use `.qword` (https://sourceware.org/binutils/docs-2.37/as/Quad.html#Quad) or two `.long` (like ARM does). `.int` is the native word size of the architecture. The documentation is lacking sometimes, so it does take some time to familiarize yourself with how all of this works. There are a lot of historical solutions that are kept because of backward compatibility (for instance comment characters on different platforms, directive names, ambiguities on per-platform basis - eg the `.align` directive and more)

src/Xamarin.Android.Build.Tasks/Utilities/ELFHelper.cs
grendello: I've decided to drop the caching, I like the simplicity of `ELFHelper.IsEmptyAOTLibrary` being static and the caching might not even make the builds that much faster (if at all, locking may make it potentially slower)

src/Xamarin.Android.Build.Tasks/Utilities/ApplicationConfigNativeAssemblyGenerator.cs
grendello: I took a slightly different approach, but similar:

```C#
enum AssemblerStringFormat
{
	/// <summary>
	/// Local string label is automatically generated for the string, this is the default behavior
	/// </summary>
	Automatic,

	/// <summary>
	/// String is output as an inline (no pointer) array/buffer filled directly with string data
	/// </summary>
	InlineArray,

	/// <summary>
	/// String is output as a pointer to symbol whose name is contents of the string being processed
	PointerToSymbol,
}

[AttributeUsage (AttributeTargets.Field | AttributeTargets.Property, Inherited = true)]
class NativeAssemblerStringAttribute : NativeAssemblerAttribute
{
	AssemblerStringFormat format;

	public bool Inline          => format == AssemblerStringFormat.InlineArray;
	public bool PadToMaxLength  { get; set; }
	public bool PointerToSymbol => format == AssemblerStringFormat.PointerToSymbol;

	public NativeAssemblerStringAttribute ()
	{
		format = AssemblerStringFormat.Automatic;
	}

	public NativeAssemblerStringAttribute (AssemblerStringFormat format)
	{
		this.format = format;
	}
}
```

src/Xamarin.Android.Build.Tasks/Utilities/NativeAssemblyGenerator/NativeAssemblyGenerator.cs
jonpryor: A "product name" of `xamarin/xamarin-android@` avoids the need (responsibility) to pick a name, and since I'm being incredibly wishy-washy here, abdicating that responsibility sounds awesome.

…or we just stick with `Xamarin.Android`. :-).

The hash remains useful, and should be extremely helpful for future issue investigations.

jonpryor:
Proposed commit message:

```markdown
[monodroid] Hash the names of all packaged DSOs (#6522)

Context: https://github.com/xamarin/xamarin-android/pull/6522
Context: https://sourceware.org/binutils/docs-2.37/as/index.html
Context: https://sourceware.org/binutils/docs-2.37/as/AArch64-Directives.html#AArch64-Directives
Context: https://sourceware.org/binutils/docs-2.37/as/ARM-Directives.html#ARM-Directives
Context: https://sourceware.org/binutils/docs-2.37/as/ARM-Directives.html#ARM-Directives
Context: https://sourceware.org/binutils/docs-2.37/as/Quad.html#Quad

Commit 000cf5a8 introduced hashing of native library names so that
we would only try to load specific Mono components which were
configured at app build time.

Expand this concept so that we support hashing any form of native
library name (DSO name) sent to use by Mono, and look for a cached
entry for that DSO.  The cache includes the actual native library
name, whether it should be ignored when requested (e.g. an empty AOT
`.so` file; see db161ae7 & df667b07), and the cached `dlopen()` value:

	struct DSOCacheEntry {
	    uint64_t    hash;
	    bool        ignore;
	    const char *name;
	    void       *handle;
	};
	DSOCacheEntry dso_cache[] = {…};

The `dso_cache` is computed at App build time, and stored in
`libxamarin-app.so`, and contains values sorted on
`DSOCacheEntry::hash`.

The use of `dso_cache` removes a bit of string processing from both
startup and the application run time, reducing app startup times in
some scenarios:

| Scenario                                      | Before ms |  After ms |        Δ |
| --------------------------------------------- | --------: | --------: | -------: |
| Plain Xamarin.Android (JIT, 64-bit)           |   311.500 |   311.600 | +0.03% ✗ |
| Plain Xamarin.Android (Profiled AOT, 64-bit)  |   253.500 |   247.000 | -2.56% ✓ |
| .NET MAUI Hello World (JIT, 64-bit)           |  1156.300 |  1159.500 | +0.28% ✗ |
| .NET MAUI Hello World (Profiled AOT, 64-bit)  |   868.900 |   831.700 | -4.28% ✓ |

(Times measured on a Pixel 3 XL.)

Above table is a subset of values from xamarin/xamarin-android#6522;
see original PR for complete table information.

TODO: Why would assembly preloading *improve* performance?
TODO: Why do the "default 64-bit JIT" times *increase*?

While implementing the above described `dso_cache` idea, I hit known
limitations of the native assembler generator code which, until this
commit, weren't a problem (mostly related to hard-coded structure and
array alignment).  Address these limitations by rewriting the assembly
generator.  It now fully implements the structure and symbol alignment
calculation for all the supported architectures.  Also added is the
ability to generate assembler code from managed structures, using
reflection, without the need of manually written code.  It also fixes
a previously unnoticed issue which made typemap structures not aligned
properly, which may have made them slightly slower than necessary
```

grendello:
Regarding the TODOs above, the differences in both mentioned cases are so small that I chalked it up to measurement error (especially for Displayed times which is inherently unreliable), but JIT may be affected by external factors on device as well. The tests are executed on a "live" device which does a lot in the background, any form of I/O may affect the tests performance. Over the time I've been working on performance, I learned that differences of around 1% can largely be ignored and accepted as "no change"

Issue #6524, opened 2021-11-25
class-parse has issue if the input jar references some unknown type


sunnyqeen:
### Error Message or Issue

We have an interface class that references other types not in the jar.
Use jar2xml way, the library and interface are generated correctly.
But after we target android 12, and use jdk11, jar2xml way is not working any more, VS gives errors.
error:
JARTOXML : error : Could not create the Java Virtual Machine.
JARTOXML : error : A fatal exception has occurred. Program will exit.

Then we switch to class-parse with more issues

We have an interface lib like

```
package test2;
import test1.TestClassA;

public class TestClassB {
  public interface BBB {
	  void DoJob1();
	  void DoJob2(TestClassA.AAA aaa);
  }
}
```

and another reference lib like

```
package test1;

public class TestClassA {
	public class AAA {
		
	}

}

```

Then the in generated C# dll,
Only method DoJob1() is there.
DoJob2 is stripped by using  class-parse.

![image](https://user-images.githubusercontent.com/649698/143452027-1f7fa7ff-69d5-4f32-9b73-6e480861b4ba.png)

(We check it using ilspy)

The example project

[lib2.zip](https://github.com/xamarin/xamarin-android/files/7603229/lib2.zip)

### Version Information
Microsoft Visual Studio Professional 2019
Version 16.11.7
VisualStudio.16.Release/16.11.7+31911.196
Microsoft .NET Framework
Version 4.8.04084

Installed Version: Professional

Visual C++ 2019   00435-20313-02959-AA247
Microsoft Visual C++ 2019

ASP.NET and Web Tools 2019   16.11.94.52318
ASP.NET and Web Tools 2019

Azure App Service Tools v3.0.0   16.11.94.52318
Azure App Service Tools v3.0.0

C# Tools   3.11.0-4.21403.6+ae1fff344d46976624e68ae17164e0607ab68b10
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.2.6 (master@34d6af2)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

IntelliCode Extension   1.0
IntelliCode Visual Studio Extension Detailed Info

Linux Core Dump Debugging   1.0.9.31829
Enables debugging of Linux core dumps.

[...]

PR #6531, opened 2021-11-29
[Xamarin.Android.Build.Tasks] Don't use -XX:-UseSplitVerifier

jonpryor:
Context: https://bugzilla.xamarin.com/show_bug.cgi?id=21096 [a 404 :-(]
Context: http://github.com/xamarin/monodroid/commit/df479c296720ee8fd2f19b12d2ff58d2b7474141
Context: https://github.com/xamarin/xamarin-android/issues/6524

(Behold the perils of *not* copying "important contextual bug info"
into commit messages, and the original bug was "private", and now
nobody can access the bug…)

Way back in 2014, we updated `<JarToXml/>` to *disable* use of the
[new type checker with StackMapTable attributes][0]:

	java -XX:-UseSplitVerifier -jar /path/to/jar2xml.jar …

This was presumably to work around a bug with [asm][1], which was
probably fixed by [bumping to asm 5.0.3][2].

Unfortunately, using `-XX:-UseSplitVerifier` *fails* on JDK-11:

	% path/to/jdk-11/java -XX:-UseSplitVerifier -jar /path/to/jar2xml.jar …
	Unrecognized VM option 'UseSplitVerifier'
	Error: Could not create the Java Virtual Machine.
	Error: A fatal exception has occurred. Program will exit.

This means it's currently impossible to use `jar2xml` with JDK-11.

Update the `<JarToXml/>` task to no longer use
`-XX:-UseSplitVerifier`.  It's no longer necessary as of
xamarin/jar2xml@d414d5b4, and it can't be used with JDK-11.

[0]: https://www.oracle.com/java/technologies/javase/vmoptions-jsp.html
[1]: https://asm.ow2.io/
[2]: https://github.com/xamarin/jar2xml/commit/d414d5b45e91e9ebb73639b163ab2bd418642f96

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/JarToXml.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/JarToXml.cs
@@ -91,9 +91,6 @@ namespace Xamarin.Android.Tasks
 			// N can be in the form of Nm or NGB (e.g 100m or 1GB ) 
 			cmd.AppendSwitchIfNotNull("-Xmx", JavaMaximumHeapSize);
 
-			// See https://bugzilla.xamarin.com/show_bug.cgi?id=21096
-			cmd.AppendSwitch ("-XX:-UseSplitVerifier");
-
 			// Arguments sent to java.exe
 			cmd.AppendSwitchIfNotNull ("-jar ", Path.Combine (MonoAndroidToolsDirectory, "jar2xml.jar"));
 


jonathanpeppers (approved):
So right now we have classic Xamarin.Android testing with JDK 1.8, and .NET 6 JDK 11.

This feature fell through the gap, because jar2xml isn't supported in .NET 6. Should we change anything to try to catch issues like this one?

jonpryor:
@jonathanpeppers [wrote](https://github.com/xamarin/xamarin-android/pull/6531#pullrequestreview-818393727:

> So right now we have classic Xamarin.Android testing with JDK 1.8, and .NET 6 JDK 11.
>
> This feature fell through the gap, because jar2xml isn't supported in .NET 6. Should we change anything to try to catch issues like this one?

We'd have to run the classic Xamarin.Android tests against JDK-11 in addition to JDK 1.8.  We didn't do this because we didn't want to increate our PR build times.

Sounds like something we should add to our nightly builds?

PR #6539, opened 2021-12-01
[Xamarin.Android.Build.Tasks] make $(AndroidNdkDirectory) optional

jonathanpeppers:
It is possible for us to not require the Android NDK to be installed,
but we have checks in place that would emit errors, such as:

    error XA5104: Could not locate the Android NDK. Please make sure the Android NDK is installed in the Android SDK Manager, or if using a custom NDK path, please ensure the $(AndroidNdkDirectory) MSBuild property is set to the custom path.

To make it possible to build with AOT without the NDK:

1. Make a `NullNdkTools` class to use when an NDK is not found.

2. Make the NDK only actually required when `$(EnableLLVM)` is `true`.

3. Update our MSBuild tests so they leave `$(AndroidNdkDirectory)`
   blank.

After these changes, I'm able to build full and profiled AOT applications without
an Android NDK installed.

After this is merged, we might be able to make `Release` builds in .NET 6 default
to use profiled AOT.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -694,8 +694,8 @@ namespace Xamarin.Android.Tasks
 				return;
 			}
 
-			NdkTools? ndk = NdkTools.Create (AndroidNdkDirectory, Log);
-			if (ndk == null) {
+			NdkTools ndk = NdkTools.Create (AndroidNdkDirectory, logErrors: false, log: Log);
+			if (Log.HasLoggedErrors) {
 				return; // NdkTools.Create will log appropriate error
 			}
 
--- a/src/Xamarin.Android.Build.Tasks/Tasks/MakeBundleNativeCodeExternal.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/MakeBundleNativeCodeExternal.cs
@@ -23,7 +23,6 @@ namespace Xamarin.Android.Tasks
 
 		const string BundleSharedLibraryName = "libmonodroid_bundle_app.so";
 
-		[Required]
 		public string AndroidNdkDirectory { get; set; }
 
 		[Required]
@@ -57,8 +56,8 @@ namespace Xamarin.Android.Tasks
 
 		public override bool RunTask ()
 		{
-			NdkTools? ndk = NdkTools.Create (AndroidNdkDirectory, Log);
-			if (ndk == null) {
+			NdkTools ndk = NdkTools.Create (AndroidNdkDirectory, logErrors: true, log: Log);
+			if (Log.HasLoggedErrors) {
 				return false; // NdkTools.Create will log appropriate error
 			}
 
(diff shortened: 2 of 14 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/MakeBundleNativeCodeExternal.cs
```diff
@@ -23,7 +23,6 @@ public class MakeBundleNativeCodeExternal : AndroidTask
 
 		const string BundleSharedLibraryName = "libmonodroid_bundle_app.so";
 
-		[Required]
 		public string AndroidNdkDirectory { get; set; }
```
jonathanpeppers, line 26: I removed `[Required]` because the `NdkTools.Create()` call below gives a much better error message:

    error XA5104: Could not locate the Android NDK. Please make sure the Android NDK is installed in the Android SDK Manager, or if using a custom NDK path, please ensure the $(AndroidNdkDirectory) MSBuild property is set to the custom path.

grendello (approved):

dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -694,10 +694,7 @@ private void AddNativeLibraries (ArchiveFileList files, string [] supportedAbis)
 				return;
 			}
 
-			NdkTools? ndk = NdkTools.Create (AndroidNdkDirectory, Log);
-			if (ndk == null) {
-				return; // NdkTools.Create will log appropriate error
-			}
+			NdkTools ndk = NdkTools.Create (AndroidNdkDirectory, logErrors: false, log: Log);
```
jonpryor, line 697: @jonathanpeppers: should we similarly check `Log.HasLoggedErrors` here as well?

jonathanpeppers: Ok, whoops I put this check back, so it should be closer to what we had before.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -258,6 +258,7 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
 				AotAssemblies = true,
 				PackageName = "com.xamarin.buildaotappandbundlewithspecialchars",
 			};
+			proj.SetProperty ("AndroidNdkDirectory", AndroidNdkPath);
```
jonpryor, line 261: Why is this property added, when it (apparently?) wasn't previously required?  Ditto elsewhere.

jonathanpeppers: `$(AndroidNdkDirectory)` is not passed into the MSBuild tests anymore, I deleted:

https://github.com/xamarin/xamarin-android/blob/5dcf2945e4cd7da50c5c8f76608e0d0266022b09/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/Builder.cs#L314-L317

And so we need to pass it in, for tests using LLVM or `$(BundleAssemblies)` now. We should also have some confidence that the other tests work now *without* `$(AndroidNdkDirectory)` passed in.

PR #6541, opened 2021-12-02
Deprecate AndroidClientHandler in .NET 6

### LINKED ISSUE #6526: Cannot create a custom AndroidClientHandler that derives from AndroidClientHandler
### Steps to Reproduce

0. Create a MAUI project in Visual Studio Preview
1. Create a class that derives from AndroidClientHandler in Xamarin.Android.Net
2. Instantiate this class and pass it to an HttpClient.
3. The construction of the derived class will fail when it calls base() since it cannot find the _nativeHandler field in AndroidClientHandler.

I don't see this happening in a Xamarin project, I guess because it builds with mono android 10?

```csharp
public class DerivedHandler : AndroidClientHandler
    {
        public DerivedHandler() : base()
        {

        }
    }
```

Using the class above fails in MAUI. Interestingly enough doing:

```csharp
public class DerivedHandler2 : DerivedHandler 
    {
        public DerivedHandler() : base()
        {

        }
    }
```

Gives you an error but this time it complains that DerivedHandler and not AndroidClientHandler doesn't have the _nativeHandler field. I am guessing it's this line of [AndroidClientHandler.cs#L318](https://github.com/xamarin/xamarin-android/blob/b04b9fda5c34b9edec10bc6fd65dd5b67fe8d52a/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs#L318), that creates the issue since it tries to find the _nativeHandler in the base clase which is different if I've inherited the class.

### Expected Behavior

Be able to call the base constructor from the derived class without issues.

### Actual Behavior

Base constructor fails in [AndroidClientHandler Line 70](https://github.com/xamarin/xamarin-android/blob/b04b9fda5c34b9edec10bc6fd65dd5b67fe8d52a/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs#L70) when it tries to call the GetUnderlyingHandler method.

### Version Information

Microsoft Visual Studio Community 2022
Version 17.1.0 Preview 1.0
VisualStudio.17.Preview/17.1.0-pre.1.0+31903.286
Microsoft .NET Framework
Version 4.8.04084

Installed Version: Community

Visual C++ 2022   00482-90000-00000-AA766
Microsoft Visual C++ 2022

ADL Tools Service Provider   1.0
This package contains services used by Data Lake tools

ASA Service Provider   1.0

ASP.NET and Web Tools 2019   17.1.125.8155
ASP.NET and Web Tools 2019

ASP.NET Web Frameworks and Tools 2019   17.1.125.8155
For additional information, visit https://www.asp.net/

Azure App Service Tools v3.0.0   17.1.125.8155
Azure App Service Tools v3.0.0

Azure Data Lake Tools for Visual Studio   2.6.4000.0
Microsoft Azure Data Lake Tools for Visual Studio

Azure Functions and Web Jobs Tools   17.1.125.8155
Azure Functions and Web Jobs Tools

Azure Stream Analytics Tools for Visual Studio   2.6.4000.0
Microsoft Azure Stream Analytics Tools for Visual Studio

C# Tools   4.1.0-1.21551.6+e4419d6f6792da7011c8589ba118f59d830ca72f
C# components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Common Azure Tools   1.10
Provides common services for use by Azure Mobile Services and Microsoft Azure Tools.

Extensibility Message Bus   1.2.6 (master@34d6af2)
Provides common messaging-based MEF services for loosely coupled Visual Studio extension components communication and integration.

Fabric.DiagnosticEvents   1.0
Fabric Diagnostic Events

Microsoft Azure Hive Query Language Service   2.6.4000.0
Language service for Hive query

Microsoft Azure Service Fabric Tools for Visual Studio   17.0
Microsoft Azure Service Fabric Tools for Visual Studio

Microsoft Azure Stream Analytics Language Service   2.6.4000.0
Language service for Azure Stream Analytics

Microsoft Azure Tools for Visual Studio   2.9
Support for Azure Cloud Services projects

Microsoft JVM Debugger   1.0
Provides support for connecting the Visual Studio debugger to JDWP compatible Java Virtual Machines

Microsoft Library Manager   2.1.134+45632ee938.RR
Install client-side libraries easily to any web project

Microsoft MI-Based Debugger   1.0
Provides support for connecting Visual Studio to MI compatible debuggers

Microsoft Visual C++ Wizards   1.0
Microsoft Visual C++ Wizards

Microsoft Visual Studio Tools for Containers   1.2
Develop, run, validate your ASP.NET Core applications in the target environment. F5 your application directly into a container with debugging, or CTRL + F5 to edit & refresh your app without having to rebuild the container.

Microsoft Visual Studio VC Package   1.0
Microsoft Visual Studio VC Package

Mono Debugging for Visual Studio   17.1.4 (dc2a48c)
Support for debugging Mono processes with Visual Studio.

NuGet Package Manager   6.1.0
NuGet Package Manager in Visual Studio. For more information about NuGet, visit https://docs.nuget.org/

ProjectServicesPackage Extension   1.0
ProjectServicesPackage Visual Studio Extension Detailed Info

Razor (ASP.NET Core)   17.0.0.2152601+724154d925d7d9d26ebf8a73a66d5219aa320400
Provides languages services for ASP.NET Core Razor.

SQL Server Data Tools   17.0.62110.20190
Microsoft SQL Server Data Tools

ToolWindowHostedEditor   1.0
Hosting json editor into a tool window

TypeScript Tools   17.0.1029.2001
TypeScript Tools for Microsoft Visual Studio

Visual Basic Tools   4.1.0-1.21551.6+e4419d6f6792da7011c8589ba118f59d830ca72f
Visual Basic components used in the IDE. Depending on your project type and settings, a different version of the compiler may be used.

Visual F# Tools   17.1.0-beta.21525.2+46af4a248255f4af2284883f48983fe7dd07a760
Microsoft Visual F# Tools

Visual Studio Code Debug Adapter Host Package   1.0
Interop layer for hosting Visual Studio Code debug adapters in Visual Studio

Visual Studio Container Tools Extensions   1.0
View, manage, and diagnose containers within Visual Studio.

Visual Studio IntelliCode   2.2
AI-assisted development for Visual Studio.

Visual Studio Tools for Containers   1.0
Visual Studio Tools for Containers

VisualStudio.DeviceLog   1.0
Information about my package

VisualStudio.Foo   1.0
Information about my package

VisualStudio.Mac   1.0
Mac Extension for Visual Studio

Xamarin   17.1.0.163 (d17-1-p1@42933de)
Visual Studio extension to enable development for Xamarin.iOS and Xamarin.Android.

Xamarin Designer   17.1.0.43 (remotes/origin/main@931878527)
Visual Studio extension to enable Xamarin Designer tools in Visual Studio.

Xamarin Templates   17.1.8 (590388a)
Templates for building iOS, Android, and Windows apps with Xamarin and Xamarin.Forms.

Xamarin.Android SDK   12.1.99.56 (main/43243b4)
    Mono: c633fe9
    Java.Interop: xamarin/java.interop/main@79744f61
    ProGuard: Guardsquare/proguard/v7.0.1@912d149
    SQLite: xamarin/sqlite/3.36.0@a575761
    Xamarin.Android Tools: xamarin/xamarin-android-tools/main@a5194e9

Xamarin.iOS and Xamarin.Mac SDK   15.3.0.41 (7b786db64)
Xamarin.iOS and Xamarin.Mac Reference Assemblies and MSBuild support.

simonrozsival:
After talking to @steveisok we decided to ~seal~ deprecate `AndroidClientHandler` in .NET 6. Users should prefer deriving from `AndroidMessageHandler`.

There are several changes in this PR:
- `AndroidMessageHandler` had `internal virtual` methods, these have to be changed to `protected virtual`
- The usage of `AndroidClientHandler` should be discouraged, so we added the `[Obsolete]` attribute
- There were many places in the codebase that refer to the `AndroidClientHandler` so these needed to be changed to `AndroidMessageHandler` but only for the .NET 6 variant
- There was a bug in `AndroidClientHandler.GetUnderlyingHandler()` which prevented users from extending the class. Since we're not sealing it after all, I fixed it.

Closes #6526


src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
```diff
@@ -58,7 +58,7 @@ namespace Xamarin.Android.Net
 	/// if they use a certificate with a fully validated trust chain) unless you store the CA certificates from your Android system in <see cref="TrustedCerts"/> along with
 	/// the self-signed certificate(s).</para>
 	/// </remarks>
-	public class AndroidClientHandler : HttpClientHandler
+	public sealed class AndroidClientHandler : HttpClientHandler
```
grendello, line 61: Thanks @simonrozsival for the PR, but I'm afraid it's not enough just to make the class sealed :( If you read the docstring above the class, it talks about a method to implement SSL certificate validation by deriving a class from `AndroidClientHandler`.  Could you make changes to address that scenario, also updating the docs as needed? Thanks!

grendello (changes requested):
Thanks @simonrozsival for the PR, but I'm afraid it's not enough just to make the class sealed :( If you read the docstring above the class, it talks about a method to implement SSL certificate validation by deriving a class from AndroidClientHandler. Could you make changes to address that scenario, also updating the docs as needed? Thanks!

steveisok:
> Thanks @simonrozsival for the PR, but I'm afraid it's not enough just to make the class sealed :( If you read the docstring above the class, it talks about a method to implement SSL certificate validation by deriving a class from AndroidClientHandler. Could you make changes to address that scenario, also updating the docs as needed? Thanks!

@grendello Given the changes in net6, `AndroidMessageHandler` is the class to derive from.

grendello:
@steveisok Yes, I know, but the doc string (at the very least) needs to be updated to reflect that change, so that it's not misleading.

simonrozsival:
Thanks for the feedback. I'll make those changes 👍

emorell96:
So before net6.0 if you wanted to do client certificates on Xamarin you would do:

```csharp
public class AndroidHttpsClientHandler : AndroidClientHandler
    {
        private SSLContext sslContext;
        private ITrustManager[] trustManagers;
        private IKeyManager[] keyManagers = null;

       

        public void Initialize()
        {
            trustManagers = GetTrustManagers();
            sslContext = GetSSLContext();
        }

        private SSLContext GetSSLContext()
        {
            string protocol;
            if (SslProtocols == SslProtocols.Tls11)
            {
                protocol = "TLSv1.1";
            }
            else if (SslProtocols == SslProtocols.Tls || SslProtocols == SslProtocols.Tls12)
            {
                protocol = "TLSv1.2";
            }
            else
            {
                throw new IOException("unsupported ssl protocol: " + SslProtocols.ToString());
            }
            SSLContext ctx = SSLContext.GetInstance(protocol);
            ctx.Init(keyManagers, trustManagers, null);
            return ctx;
        }

        public new SslProtocols SslProtocols { get; set; } = SslProtocols.Tls12;

        public void SetClientCertificate(byte[] pkcs12, char[] password)
        {
            keyManagers = GetKeyManagersFromClientCert(pkcs12, password);
            SSLContext newContext = GetSSLContext();
            sslContext = newContext;
        }

        private ITrustManager[] GetTrustManagers()
        {
            TrustManagerFactory trustManagerFactory = TrustManagerFactory.GetInstance(TrustManagerFactory.DefaultAlgorithm);
            trustManagerFactory.Init((KeyStore)null);
            return trustManagerFactory.GetTrustManagers();
        }

        private IKeyManager[] GetKeyManagersFromClientCert(byte[] pkcs12, char[] password)
        {
            if (pkcs12 != null)
            {
                using (MemoryStream memoryStream = new MemoryStream(pkcs12))
                {
                    KeyStore keyStore = KeyStore.GetInstance("pkcs12");
                    
                    keyStore.Load(memoryStream, password);
                    KeyManagerFactory kmf = KeyManagerFactory.GetInstance("x509");
                    kmf.Init(keyStore, password);
                    return kmf.GetKeyManagers();
                }
            }
            return null;
        }

        protected override SSLSocketFactory ConfigureCustomSSLSocketFactory(HttpsURLConnection connection)
        {
            SSLSocketFactory socketFactory = sslContext.SocketFactory;
            if (connection != null)
            {
                connection.SSLSocketFactory = socketFactory;
            }
            return socketFactory;
        }
    }

```

Where you would override ``ConfigureCustomSSLSocketFactory``, how would you do the same thing with ``AndroidMessageHandler``?

Thanks you!

samples/VSAndroidApp/VSAndroidApp.csproj
```diff
@@ -25,7 +25,7 @@
     <MonoAndroidAssetsPrefix>Assets</MonoAndroidAssetsPrefix>
     <AndroidEnableSGenConcurrent>true</AndroidEnableSGenConcurrent>
     <AndroidUseAapt2>true</AndroidUseAapt2>
-    <AndroidHttpClientHandlerType>Xamarin.Android.Net.AndroidClientHandler</AndroidHttpClientHandlerType>
+    <AndroidHttpClientHandlerType>Xamarin.Android.Net.AndroidMessageHandler</AndroidHttpClientHandlerType>
```
jonathanpeppers, line 28: This sample is meant to be the default template you get in VS for Xamarin.Android, not .NET 6.

So some concerns are:
1. Do we want this the default in Xamarin.Android? We'll have to update templates for this new value, they are in a different repo.
2. We'd also need the Xamarin extension in Visual Studio to update its UI on the properties pages.
3. Should we change the default value of `$(AndroidHttpClientHandlerType)` when it is blank?

These lead me to think we should make these changes in .NET 6 only and not affect "classic" Xamarin.Android?

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.Legacy.cs
```diff
@@ -56,6 +56,7 @@ namespace Xamarin.Android.Net
 	/// if they use a certificate with a fully validated trust chain) unless you store the CA certificates from your Android system in <see cref="TrustedCerts"/> along with
 	/// the self-signed certificate(s).</para>
 	/// </remarks>
+	[Obsolete("AndroidClientHandler has been deprecated. Use AndroidMessageHandler instead.")]
 	public class AndroidClientHandler : HttpClientHandler
```
jonathanpeppers, line 60: Should this be obsolete only in .NET 6? Or Xamarin.Android as well?

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
```diff
@@ -50,24 +50,26 @@ namespace Xamarin.Android.Net
...
+	/// the Java APIs, AndroidClientHandler defines a way to handle the situation. It can store the necessary certificates (either CA or server certificates)
+	/// in the <see cref="TrustedCerts"/> collection. If, however, the application requires finer control over the SSL configuration (e.g. it implements its own
+	/// TrustManager) then it should derive a custom class from <see cref="Xamarin.Android.Net.AndroidMessageHandler"/> instead of using AndroidClientHandler.
+	/// Note that the instance of AndroidClientHandler configured to accept an "invalid" certificate from the particular server will most likely fail to validate
+	/// certificates from other servers (even if they use a certificate with a fully validated trust chain) unless you store the CA certificates from your Android
+	/// system in <see cref="TrustedCerts"/> along with the self-signed certificate(s).</para>
 	/// </remarks>
-	public class AndroidClientHandler : HttpClientHandler
+	[Obsolete("AndroidClientHandler has been deprecated. Use AndroidMessageHandler instead.")]
+	public sealed class AndroidClientHandler : HttpClientHandler
```
jonathanpeppers, line 61: I agree with the concern here: https://github.com/xamarin/xamarin-android/pull/6541#issuecomment-985045429

If this is `sealed`, how do you implement "ssl pinning" in an app?

simonrozsival: All the original logic was transferred into `AndroidMessageHandler` so anything that was previously possible to achieve by extending the `AndroidClientHandler` will be possible by extending `AndroidMessageHandler`.

jonathanpeppers: So if there is a NuGet package out there now that subclasses `AndroidClientHandler` internally, then updating Xamarin.Android they would probably run into a scary runtime exception?

@jonpryor can we break the API like this in Xamarin.Android?

It seems like this break should be reasonable in .NET 6, as customer would expect breaks as they move over.

samples/VSAndroidApp/VSAndroidApp.csproj
steveisok: Yeah, I believe this change should only be for .NET6

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
steveisok: I think it would make sense to obsolete the type in .NET6 and keep things the same in Xamarin.Android.

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.Legacy.cs
simonrozsival: I removed the attribute from the Xamarin.Android version

samples/VSAndroidApp/VSAndroidApp.csproj
jonathanpeppers: Ok @simonrozsival, if you want to follow what we did elsewhere is use `#if MONOANDROID1_0` for Xamarin.Android, and `#if !MONOANDROID1_0` for .NET 6.

We could also exclude whole files from the compilation, but that probably won't apply here.

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
jonathanpeppers: @simonrozsival I think we would wrap the `[Obsolete]` attribute with:
```csharp
#if !MONOANDROID1_0
[Obsolete("AndroidClientHandler has been deprecated. Use AndroidMessageHandler instead.")]
#endif
```

simonrozsival: @jonathanpeppers The attribute is only used in `AndroidClientHandler.cs` (included only when `TargetFramework` is `!= monoandroid10` and not in `AndroidClientHandler.Legacy.cs` which is included for `monoandroid10` (https://github.com/xamarin/xamarin-android/blob/main/src/Mono.Android/Mono.Android.csproj#L350..L351). Shouldn't that be enough?

jonathanpeppers: Missed that, thanks.

simonrozsival:
The code is broken atm and I'm working on a fix locally. I'll let you know when it's finished so you can re-run the tests. I'll switch the PR to draft until it's ready.

simonrozsival:
@jonathanpeppers I was able to compile the code locally but running tests fails for some reason and I'm not able to get it to work. Could you please trigger the tests in CI for me again please? 😅

simonrozsival:
The `DisallowAutoRedirect` integration test fails with `AndroidMessageHandler` because it doesn't throw when the server returns a redirect response and `AllowAutoRedirect` is set to `false`. Based [on the docs](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.allowautoredirect?view=net-6.0#System_Net_Http_HttpClientHandler_AllowAutoRedirect) it seems to me that the handler wasn't supposed to throw an exception in that case and it was supposed to return the 3XX response 🤔

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
```diff
@@ -308,14 +306,19 @@ protected virtual Task SetupRequest (HttpRequestMessage request, HttpURLConnecti
...
+			return _underlyingHander.ConfigureCustomSSLSocketFactoryInternal (connection);
 		}
 
 		[DynamicDependency (DynamicallyAccessedMemberTypes.PublicParameterlessConstructor, typeof (AndroidMessageHandler))]
 		object GetUnderlyingHandler ()
 		{
 			var fieldName = "_nativeHandler";
-			var baseType = GetType ().BaseType;
+			var type = GetType ();
+			while (type != typeof (AndroidClientHandler)) {
```
steveisok, line 317: Does it make sense to also break when type == typeof(Object)?  Maybe throw as well when we don't have HttpClientHandler as the base type.

steveisok:
@jonathanpeppers not sure why tests fail b/c of `DisallowAutoRedirect` ?

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
simonrozsival: This is a member function of `AndroidClientHandler` and so it will always be called on `AndroidClientHandler` or classes derived from it. I assume that this loop always has to stop but I can definitely throw an exception if for any reason we reach `object`.

jonathanpeppers:
So this test has different behavior now using `AndroidMessageHandler`: https://github.com/xamarin/xamarin-android/blob/681887ebdbd192ce7ce1cd02221d4939599ba762/tests/Mono.Android-Tests/Xamarin.Android.Net/HttpClientIntegrationTests.cs#L926-L954

It seems like it failed on .NET 6 & Xamarin.Android, so it is related to the new `AndroidMessageHandler` type. Should we address this?

jonathanpeppers:
I wonder if the `AllowAutoRedirect` is not working as intended with `AndroidMessageHandler`?

steveisok:
> I wonder if the `AllowAutoRedirect` property is not working as intended with `AndroidMessageHandler`?

I would have expected the same behavior, but maybe the defaults are somehow different.

simonrozsival:
When I tried using the DLL in a test app, both client handlers throw an exception when I set `AllowAutoRedirect = false` (`HttpClient` calls `response.EnsureSuccessStatusCode()` from `GetStringAsync`). This must be something wrong with my test's code.

simonrozsival:
Perhaps this is because `HttpListener` doesn't intercept the HTTP requests handled by the native android implementation and it only intercepts the managed implementation? The `CreateListener` function relieas on `HttpListener` and if it doesn't intercept the response and modify the status code then the test can't fail.

There are only 2 tests which use `CreateListener` and aren't disabled or are missing the `[Test]` attribute (18 are disabled with `#if TODO ` and 1 is missing the attribute). Out of these two tests one _succeeds_ if the listener _doesn't intercept_ a request and the other one is `DisallowAutoRedirect` which fails.

If I'm right I would suggest disabling `DisallowAutoRedirect` for now. If `HttpListener` is supposed to work then maybe we should create a separate issue for it.

simonrozsival:
@jonathanpeppers do you know what could be the problem with the `HttpListener`? I don't have any experience working with them so I don't know if the problem with these tests is easy to fix or not.

Except for the failing tests (which could be disabled IMO) this PR is ready for review.

jonathanpeppers:
@jonpryor what do you think? since @grendello is out for the holidays.

It seems like this test is pretty old, predates OSS on Github.

simonrozsival:
The previous commit (https://github.com/xamarin/xamarin-android/pull/6541/commits/91ea475a73236233434734ae28f085b694dc9760) should fix the problem with the failing tests. It seems that the proper way to send a redirect response from `HttpListener` is to use the `Redirect` function (simply setting the properties individually leaves the connection in a weird state and the next HTTP request fails for some reason).

simonrozsival:
The failing tests seem unrelated to this PR.

jonathanpeppers (approved):
LGTM, I restarted a couple test lanes.

Maybe @jonpryor can review, since Marek is out?

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
```diff
@@ -345,7 +383,10 @@ string EncodeUrl (Uri url)
 			}
 		}
 
-		internal virtual async Task <Java.Net.Proxy?> GetJavaProxy (Uri destination, CancellationToken cancellationToken)
+		internal Task <HttpResponseMessage> SendAsyncInternal (HttpRequestMessage request, CancellationToken cancellationToken)
+			=> SendAsync (request, cancellationToken);
+
+		protected virtual async Task <Java.Net.Proxy?> GetJavaProxy (Uri destination, CancellationToken cancellationToken)
```
jonpryor, line 389: I don't understand changes "like this", in which an `internal virtual` method is turned into a `protected virtual` + `internal` method pair.  This increase "supported API surface area", without providing docs for the newly added members.

Is this for compatibility with Legacy `AndroidClientHandler`, which *does* have a `protected virtual GetJavaProxy` method?

A quick spot-check suggests that these changes *do* increase compatibility with Legacy `AndroidClientHandler`, which just "redirects" the question: why were these methods changed *from* `protected virtual` to `internal virtual` in 1e5bfa3ff0ae18ce9407677667f05a9bfba53478?  @steveisok?  Is there something we should be aware of here?

jonpryor: https://discord.com/channels/732297728826277939/732297837953679412/920360699711090768

steveisok: Beat me to it ;-)

simonrozsival: > without providing docs for the newly added members

Just a comment regarding the docs: I fixed all comments in `AndroidMessageHandler` (they were all copied from `AndroidClientHandler` and they often still contained references to `AndroidClientHandler`) so the docs of the protected functions will be correct. The `GetJavaProxy` fn doesn't have any docs in [AndroidClientHandler](https://docs.microsoft.com/en-us/dotnet/api/xamarin.android.net.androidclienthandler?view=xamarin-android-sdk-12) so I didn't try to come up with any new documentation for the function.

grendello (approved):

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
dellis1972: Might be a silly question, but do we also need to worry about never breaking out of the loop here?
I think having a throw if we reach `object` would be a good idea.

simonrozsival: If we reached the state where `type == typeof(object)`, then `type.BaseType` would return `null` and the next iteration would cause a null reference exception (not great but better than an infinite loop). I don't think it's possible to get to that state though as I described it the previous comment. Do you think there's some way to reach that state?

jonpryor:
work-in-progress Squash-and-merge commit message:

```markdown
[One .NET] Deprecate Xamarin.Android.Net.AndroidClientHandler (#6541)

Fixes: https://github.com/xamarin/xamarin-android/issues/6526

The .NET 6+ `AndroidClientHandler` type ("added" in 1e5bfa3f) cannot
be used as a base class.  Attempting to do so [results in][0] an
`InvalidOperationException`

	class DerivedHandler : Xamarin.Android.Net.AndroidClientHandler {
	}
	class DerivedHandler2 : DerivedHandler {
	}

	// …
	new DerivedHandler2();
	// throws InvalidOperationException:
	//   Field '_nativeHandler' is missing from type 'DerivedHandler'.

Fix this bug by updating `AndroidClientHandler.GetUnderlyingHandler()`
to check all base classes, not just the immediate base class, for the
`_nativeHandler` field.

Additionally, `[Obsolete]` the `AndroidClientHandler` type.

TODO: @steveisok: is there any elaboration on *why* we decided to
deprecate `AndroidClientHandler`?

.NET 6 users should instead inherit from
`Xamarin.Android.Net.AndroidMessageHandler`.

Administrivia:

  * `AndroidMessageHandler` had `internal virtual` methods.  These
    have to be changed to `protected virtual`.

  * There were many places in the codebase that refer to the
    `AndroidClientHandler`; these have been changed to refer to
    `AndroidMessageHandler` in .NET 6+ builds.


[0]: https://github.com/xamarin/xamarin-android/blob/e24ea21b77353f458633e781d2f02a97fa3c96ad/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs#L317-L322
```

src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs
```diff
@@ -308,14 +306,19 @@ protected virtual Task SetupRequest (HttpRequestMessage request, HttpURLConnecti
...
+			return _underlyingHander.ConfigureCustomSSLSocketFactoryInternal (connection);
 		}
 
 		[DynamicDependency (DynamicallyAccessedMemberTypes.PublicParameterlessConstructor, typeof (AndroidMessageHandler))]
 		object GetUnderlyingHandler ()
 		{
 			var fieldName = "_nativeHandler";
-			var baseType = GetType ().BaseType;
+			var type = GetType ();
+			// while (type != typeof (AndroidClientHandler)) {
```
jonpryor, line 317: When this `while` loops is re-enabled, I think a more "conventional" approach would be to instead do:

```csharp
FieldInfo? field = null;
for (var type = GetType (); type != null; type = type.BaseType) {
    field = type.GetField (fieldName, BindingFlags.Instance | BindingFlags.NonPublic);
    if (field != null)
        break;
}
if (field == null)
    throw new InvalidOperationException (…);
…
```

The benefit to this approach is that it's *conceptually simpler* (yay maintenance), and also means that *if someone knows what they're doing* (possible :rofl:), they can "simply" add a new `_nativeHandler` instance field and things will (more or less) Just Work™.

simonrozsival: I like this approach, thanks for the suggestion!

simonrozsival:
This commit fixed the stuck `Xamarin.Android-PR (Smoke Tests APKs .NET - macOS)` (interpreter): https://github.com/xamarin/xamarin-android/pull/6541/commits/142a0d5beef7bd332b07b9667188b74163a3b878

I'm not sure why but this code was causing it:
```
var obj = CreateHandler ();
if (obj is not HttpClientHandler h) {
	Assert.Ignore ($"{obj.GetType()} is not a HttpClientHandler.");
	return;
}
```

Now there's just `Xamarin.Android-PR (Linux Build and Smoke Test)` that's failing and it's because of `Error MSB3030: Could not copy the file "/mnt/vss/_work/1/s/xamarin-android/bin/Release/lib/xamarin.android/xbuild/Xamarin/Android//de/Xamarin.Android.Build.Debugging.Tasks.resources.dll" because it was not found.`. Does anyone know what could be the cause of that? Is that a known issue?

simonrozsival:
I think there's not much I can do about the failing test (it seems completely unrelated to my PR) and so I think this PR is done.

PR #6547, opened 2021-12-06
[One .NET] default to Profiled AOT for Release builds

jonathanpeppers:
To improve startup performance for `Release` builds:

1. Default `$(RunAOTCompilation)` to `true`

2. Default `$(AndroidEnableProfiledAot)` to `true`.

Going forward, you would have to explicitly turn these off if you want
them to be off.

We also can drop the `android-aot` workload, and just include the
dotnet/runtime packs in the `android` workload:

https://github.com/dotnet/runtime/blob/c7adf5598f27e59644030ddae4e382cbda74e41e/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.json.in#L29-L40

These packs are already inserted into Visual Studio, so we can rely on
them.

Since the new project options screen evaluates MSBuild, we don't
appear to need to do anything on the UI side:

image

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -132,6 +132,9 @@ namespace Xamarin.Android.Build.Tests
 				new XamarinAndroidApplicationProject ();
 			proj.ProjectName = testName;
 			proj.IsRelease = true;
+			// TODO: AOT fails https://github.com/xamarin/xamarin-android/issues/6685
+			// .NET 6 uses AOT by default for Release
+			proj.AotAssemblies = false;
 			if (forms) {
 				proj.PackageReferences.Clear ();
 				proj.PackageReferences.Add (KnownPackages.XamarinForms_4_7_0_1142);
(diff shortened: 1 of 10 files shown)


dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
```diff
@@ -132,6 +132,10 @@ public void SmokeTestBuildWithSpecialCharacters ([Values (false, true)] bool for
 				new XamarinAndroidApplicationProject ();
 			proj.ProjectName = testName;
 			proj.IsRelease = true;
+			if (Builder.UseDotNet) {
+				// See: https://github.com/dotnet/runtime/issues/56163
```
jonathanpeppers, line 136: This issue is closed, we might need a new one.

jonathanpeppers:
Ok, here is the plan to finish this up:
1. Get rid of the `android-aot` workload -- and put everything in the `android` workload.
2. Make sure the AOT compilers from dotnet/runtime get inserted into VS. (This may/may not already be happening, need to check)
3. Fix projected named `テスト`
4. Fix F# projects & satellite assemblies

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
```diff
@@ -614,6 +614,7 @@ public void RunWithInterpreterEnabled ([Values (false, true)] bool isRelease)
 
 			proj = new XamarinAndroidApplicationProject () {
 				IsRelease = isRelease,
+				AotAssemblies = false, // Release defaults to Profiled AOT for .NET 6
```
jonathanpeppers, line 617: Any of the tests with this comment, they relied on it being a non-AOT build to pass.

dellis1972 (approved):

PR #6551, opened 2021-12-07
[Xamarin.Android.Build.Tasks] Use class-parse.exe out of MSBuild process

jpobst:
Context: https://github.com/xamarin/java.interop/pull/499

There are potentially some conflicts with using `Xamarin.Android.Tools.Bytecode` directly in the MSBuild process, as it now imports `protobuf-net` which is a commonly used library.

To ensure it isn't running in-process, switch our targets to shell out to `class-parse[.exe/.dll]`.

Additionally add `class-parse.dll`, etc. to the .NET 6 install pack.  We are already shipping `class-parse.exe`, etc. in the Classic installer even though it wasn't used, so no additional installer changes are needed there.

As expected, moving this out-of-process is slightly slower, but still acceptable.

```
Before
292 ms  ClassParse                                 1 calls

After
425 ms  ClassParse                                 1 calls
```

Additionally adds `java-resolution-report.log` to `<FileWrites />` which was added in https://github.com/xamarin/java.interop/pull/849.

### DIFF
--- a/build-tools/create-packs/Microsoft.Android.Sdk.proj
+++ b/build-tools/create-packs/Microsoft.Android.Sdk.proj
@@ -69,6 +69,9 @@ core workload SDK packs imported by WorkloadManifest.targets.
       <_PackageFiles Include="$(XAInstallPrefix)xbuild\Xamarin\Android\Microsoft.Android.Sdk.ILLink.pdb" PackagePath="tools" />
       <_PackageFiles Include="$(XAInstallPrefix)xbuild\Xamarin\Android\%(_LocalizationLanguages.Identity)\Microsoft.Android.Sdk.ILLink.resources.dll" PackagePath="tools\%(_LocalizationLanguages.Identity)" />
       <_PackageFiles Include="$(ToolsSourceDir)**" PackagePath="tools" />
+      <_PackageFiles Include="$(NetCoreAppToolsSourceDir)class-parse.dll" PackagePath="tools" />
+      <_PackageFiles Include="$(NetCoreAppToolsSourceDir)class-parse.pdb" PackagePath="tools" />
+      <_PackageFiles Include="$(NetCoreAppToolsSourceDir)class-parse.runtimeconfig.json" PackagePath="tools" />
       <_PackageFiles Include="$(NetCoreAppToolsSourceDir)generator.dll" PackagePath="tools" />
       <_PackageFiles Include="$(NetCoreAppToolsSourceDir)generator.pdb" PackagePath="tools" />
       <_PackageFiles Include="$(NetCoreAppToolsSourceDir)generator.runtimeconfig.json" PackagePath="tools" />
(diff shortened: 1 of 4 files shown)


build-tools/create-packs/Microsoft.Android.Sdk.proj
```diff
@@ -69,6 +69,9 @@ core workload SDK packs imported by WorkloadManifest.targets.
       <_PackageFiles Include="$(XAInstallPrefix)xbuild\Xamarin\Android\Microsoft.Android.Sdk.ILLink.pdb" PackagePath="tools" />
       <_PackageFiles Include="$(XAInstallPrefix)xbuild\Xamarin\Android\%(_LocalizationLanguages.Identity)\Microsoft.Android.Sdk.ILLink.resources.dll" PackagePath="tools\%(_LocalizationLanguages.Identity)" />
       <_PackageFiles Include="$(ToolsSourceDir)**" PackagePath="tools" />
+      <_PackageFiles Include="$(NetCoreAppToolsSourceDir)class-parse.dll" PackagePath="tools" />
+      <_PackageFiles Include="$(NetCoreAppToolsSourceDir)class-parse.pdb" PackagePath="tools" />
+      <_PackageFiles Include="$(NetCoreAppToolsSourceDir)class-parse.runtimeconfig.json" PackagePath="tools" />
```
jonathanpeppers, line 74: For classic Xamarin.Android, is there also a `class-parse.exe` you need here:

https://github.com/xamarin/xamarin-android/blob/f98c4a8ad4b8703cf9bb39b383b44b1759b49acd/build-tools/installers/create-installers.targets#L130-L131

jpobst: I would have thought so, but we were already shipping `class-parse.exe` in Classic even though we weren't using it:

https://github.com/xamarin/xamarin-android/blob/f98c4a8ad4b8703cf9bb39b383b44b1759b49acd/build-tools/installers/create-installers.targets#L128-L129

src/Xamarin.Android.Build.Tasks/Tasks/ClassParse.cs
```diff
@@ -26,43 +20,33 @@ public class ClassParse : AndroidTask
...
-				};
-				foreach (var jar in SourceJars) {
-					if (Bytecode.ClassPath.IsJarFile (jar.ItemSpec)) {
-						classPath.Load (jar.ItemSpec);
-					}
-				}
-				classPath.SaveXmlDescription (output);
+			var cmd = GetCommandLineBuilder ();
+
+			var responseFile = Path.Combine (Path.GetDirectoryName (OutputFile), "class-parse.rsp");
```
jonathanpeppers, line 27: If `class-parse.exe` succeeds, should we delete this file immediately? Just leave it on disk if there is a non-zero exit code?

Otherwise, should something delete it when you run the `Clean` target?

jonathanpeppers (commented):
Could you measure the build time before & after?

I think you could build a binding with `-clp:performancesummary` or `-bl` and just see how long this task takes.

src/Xamarin.Android.Build.Tasks/Tasks/ClassParse.cs
jpobst: Good suggestions, add performance data and `<FileWrites />` for response file.

jonathanpeppers (approved):
Looks good if CI ends up green 👍

PR #6555, opened 2021-12-08
[build] set $DOTNET_MULTILEVEL_LOOKUP to 0

jonathanpeppers:
Context: https://docs.microsoft.com/dotnet/core/install/windows#download-and-manually-install

I have VS main installed, and my system has:

    > dotnet --version
    6.0.200-preview.21573.3

And so any built-in `~\android-toolchain\dotnet\dotnet.exe` commands
would actually run my system `dotnet` because it is newer than the
version our build provisions. In this case, my `C:\Program
Files\dotnet\dotnet.exe` was attempting to install my local build.

We can use `$DOTNET_MULTILEVEL_LOOKUP` to force
`~\android-toolchain\dotnet\dotnet.exe` to be always used. I couldn't
find many docs on this, but it is used in various dotnet repos, such as:

https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/dotnet.cmd#L16-L17

I also fixed the wildcard match for `@(_PreviewPacks)`, which matched
a commit distance of 132!

    Microsoft.Android.Ref.31.31.0.101-ci.pr.gh6555.132.nupkg

It is intended to only match:

    Microsoft.Android.Ref.32.*.nupkg

### DIFF
--- a/build-tools/create-packs/Directory.Build.targets
+++ b/build-tools/create-packs/Directory.Build.targets
@@ -119,7 +119,7 @@
     </PropertyGroup>
     <ItemGroup>
       <_NuGetSources Include="$(OutputPath.TrimEnd('\'))" />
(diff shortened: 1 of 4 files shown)


dellis1972 (approved):

jonathanpeppers:
image

I think this test started failing on main, going to merge.

PR #6556, opened 2021-12-08
[One .NET] fix AOT builds with different settings

jonathanpeppers:
Building a .NET 6 `Release` app with `UseInterpreter=true` and
`RunAOTCompilation=true` would fail with:

    Microsoft.Android.Sdk.Aot.targets(71,5): Unknown Mode value: Interpreter. 'Mode' must be one of: Normal,JustInterp,Full,FullInterp,Hybrid,LLVMOnly,LLVMOnlyInterp

Reviewing the code:

https://github.com/dotnet/runtime/blob/60edc0bdafce1849fec19e3ec4f766074de50fc3/src/tasks/AotCompilerTask/MonoAOTCompiler.cs#L1064-L1073

There does not appear to be a mode of `Normal` + `Interp`. So let's
just make AOT take precendence over `$(UseInterpreter)`. On Android
`$(UseInterpreter)` is a Debug-mode setting for Hot Reload.

Building a .NET 6 `Debug` app `RunAOTCompilation=true` would fail with:

    Microsoft.Android.Sdk.Aot.targets(49,5): error MSB4044: The "GetAotAssemblies" task was not given a value for the required parameter "AndroidApiLevel".

This is because the linker was skipped, particularly this target:

`_CreatePropertiesCache` doesn't run, and so we have some empty
properties in this case. We can simply add `_CreatePropertiesCache` to
the `_AndroidAot` MSBuild target's `DependsOnTargets` to solve this
issue.

### DIFF
--- a/Documentation/guides/messages/xa0119.md
+++ b/Documentation/guides/messages/xa0119.md
@@ -27,14 +27,21 @@ Remove the following options from `Debug` configurations:
 * App Bundles
   * `<AndroidPackageFormat>aab</AndroidPackageFormat>`
 
+Remove the following from `Release` configurations:
+
+* Hot Reload support
+  * `<UseInterpreter>true</UseInterpreter>`
+
 *DO* use the following options for `Debug` configurations:
 
 * `<AndroidLinkMode>None</AndroidLinkMode>`
 * `<EmbedAssembliesIntoApk>False</EmbedAssembliesIntoApk>`
+* `<UseInterpreter>true</UseInterpreter>`
 
 *DO* use the following options for `Release` configurations:
 
 * `<EmbedAssembliesIntoApk>True</EmbedAssembliesIntoApk>`
+* `<AotAssemblies>True</AotAssemblies>` or in .NET 6 `<RunAOTCompilation>True</RunAOTCompilation>`
 
 Consider submitting a [bug][bug] if you are getting one of these
 warnings under normal circumstances.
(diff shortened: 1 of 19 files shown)


jonathanpeppers:
I added:
```
warning XA0119: Using the interpreter and AOT at the same time is not recommended. Use the interpreter for hot reload support in Debug configurations and AOT for Release configurations.
```

jonpryor:
```markdown
Building a .NET 6 `Release` app with `UseInterpreter=true` and
`RunAOTCompilation=true` would fail with:

    Microsoft.Android.Sdk.Aot.targets(71,5): Unknown Mode value: Interpreter. 'Mode' must be one of: Normal,JustInterp,Full,FullInterp,Hybrid,LLVMOnly,LLVMOnlyInterp

Reviewing [the code][0]:

	public enum MonoAotMode
	{
	    Normal,
	    JustInterp,
	    Full,
	    FullInterp,
	    Hybrid,
	    LLVMOnly,
	    LLVMOnlyInterp
	}

There does not appear to be a mode for `Normal` + `Interp`.  So let's
instead have AOT take precedence over `$(UseInterpreter)`.
On Android `$(UseInterpreter)` is a Debug-mode setting for Hot Reload;
the *expectation* is that AOT won't be enabled in Debug configs as
Debug builds should be "fast" while AOT builds are *slow*, so if AOT
is explicitly enabled, *presumably* the developer doesn't want the
Interpreter.

To inform developers of this change, we now emit an XA0119 warning
when both `$(UseInterpreter)`=true and `$(RunAOTCompilation)`=true:

	dotnet build -p:UseInterpreter=true -p:RunAOTCompilation=true
	…
	warning XA0119: Disabling the interpreter;
	  using the interpreter and AOT at the same time is not supported.
	  Use the interpreter for hot reload support in Debug configurations and
	  AOT for Release configurations.

While testing this new priority order, we found that building a
.NET 6 `Debug` config app with `$(RunAOTCompilation)`=true would
fail with:

	Microsoft.Android.Sdk.Aot.targets(49,5): error MSB4044:
	The "GetAotAssemblies" task was not given a value for the required parameter "AndroidApiLevel".

This is because the linker was skipped, particularly this target:

The `_CreatePropertiesCache` target didn't run, and so we have some
empty properties in this case.

We can simply add `_CreatePropertiesCache` to the `_AndroidAot` MSBuild
target's `DependsOnTargets` to solve this issue.

[0]: https://github.com/dotnet/runtime/blob/60edc0bdafce1849fec19e3ec4f766074de50fc3/src/tasks/AotCompilerTask/MonoAOTCompiler.cs#L1064-L1073
```

Issue #6520, opened 2021-11-23
.NET SDK for Android + `$(EnableLLVM)`=True + `$(AotAssemblies)`=True + Windows: bad path quoting?


jonpryor:
### Steps to Reproduce

1. `dotnet new android -n android`
2. Edit `android\android.csproj`, and add:

   ```xml
      <PropertyGroup Condition=" '$(TargetFramework)' == 'net6.0-android' And '$(Configuration)' == 'Release' ">
        <EnableLLVM>True</EnableLLVM>
        <AotAssemblies>True</AotAssemblies>
      </PropertyGroup>
   ```

3. Build! `dotnet build -c Release -v:diag > b.txt`

### Expected Behavior

Success!

### Actual Behavior

Fails to build:

```
[...]

PR #6544, opened 2021-12-03
[tests] enable NoSymbolsArgShouldReduceAppSize test

jonathanpeppers:
Context: https://github.com/dotnet/runtime/issues/57800

dotnet/runtime#57800 is fixed.

PR #6560, opened 2021-12-09
Revert "[Xamarin.Android.Build.Tasks] $(AndroidNdkDirectory) is optio…

jonpryor:
…nal (#6539)"

This reverts commit c96d9f496675d177687273fb20bc65f6e5dc73cd.

Context? https://github.com/xamarin/xamarin-android/issues/6520

On CI, ever since c96d9f49 landed, the
`NoSymbolsArgShouldReduceAppSize` test has been failing.

Is this because of c96d9f49?

Try reverting c96d9f49, see if the test passes again.


jonpryor:
Conjecture:

The failing `NoSymbolsArgShouldReduceAppSize` test was merged on 2021-Dec-6 in #6544.

The change to make `$(AndroidNdkDirectory)` optional in #6539 was *also* merged on 2021-Dec-06, and was last updated on 2021-Nov-30, *before* #6544 was merged.  PR #6539 thus never ran the `NoSymbolsArgShouldReduceAppSize` test.

#6544 passed because our unit tests provision the NDK into `$HOME`, which is a directory that doesn't contain spaces.

#6539, in making the NDK "optional", *instead* uses NDK-related tooling that are located in the Xamarin.Android installation directory, which is within Visual Studio, which is underneath `C:\Program Files…`, which is a directory which contains spaces.

Thus, it looks like #6539 "broke" the `NoSymbolsArgShouldReduceAppSize` test, by causing it to use NDK-related tools in a directory containing spaces.  (Which is something that we should be testing for *anyway*!  And I thought was checked for as part of fd5f31cc89648957243f31ba8fd7af7fccba11e4!  Apparently that's still not quite enough.)

If this PR has `NoSymbolsArgShouldReduceAppSize` fail, this conjecture will be validated.

jonpryor:
*Assuming* that [the conjecture](https://github.com/xamarin/xamarin-android/pull/6560#issuecomment-990033229) is validated, what do we do?

 1. We could revert #6539, making the NDK required.  However, this only "hides" the problem; builds will still fail if the NDK installation directory contains spaces.
 2. We could revert #6544, disabling the `NoSymbolsArgShouldReduceAppSize` test.  We still have a problem if the NDK dir contains spaces, but at least we'll be green!
 3. Just leave the test failing until Issue #6520 is fixed.

If (3) doesn't take too long, that might be the better idea.

jonathanpeppers (changes requested):
I would go with option 2 and revert this one instead:

https://github.com/xamarin/xamarin-android/pull/6544

Option 3 will require a change in dotnet/runtime to fix, and it seems preferable for the `$(AndroidNdkDirectory)` to not be required even if it uncovered a separate issue.

jonpryor:
@jonpryor [wrote](https://github.com/xamarin/xamarin-android/pull/6560#issuecomment-990033229):

> If this PR has `NoSymbolsArgShouldReduceAppSize` pass, this conjecture will be validated.

`NoSymbolsArgShouldReduceAppSize` passed, which is the behavior seen with PR #6544.  The conjecture is validated.

We'll revert #6544 so that our CI can be green again, and properly fix it when a proper fix exists.

PR #6562, opened 2021-12-09
[One .NET] rework .NET 6 AOT support

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/6520
Context: https://github.com/dotnet/runtime/issues/56163
Context: https://github.com/dotnet/runtime/pull/62725

This fixes AOT builds for project names like `foo Ümläüts`, and
LLVM-related options are no longer space delimited.

Previously, `<MonoAOTCompiler/>` managed its own `WorkingDirectory` to
the location that `.dll` input files were located. The problem with
this, is that *other* paths like `temp-path`, etc. would then need to
be full paths. Full paths containing characters like `Ümläüts` would
break!

Additionally, `%(AotArguments)` metadata is split on `;` and joined on
`,`. And since `ld-flags` contains a `;`, we would lose it. So we
delimited by the space character, which also breaks if directories
contain a space!

To solve these issues:

* Added strongly-typed properties to the `<MonoAOTCompiler/>` MSBuild
  task. We no longer have to put as many things in `%(AotArguments)`
  on each assembly. Only assembly-specific settings go there now, like
  the path to AOT profiles. Right now the "main assembly" has no
  profile at all, and the rest get built-in profiles.

* Added a `WorkingDirectory` property, which we pass in
  `$(MSBuildProjectDirectory)` for the current project directory.

After these changes, I could remove many temporary comments around
dotnet/runtime#56163, as the tests work properly now.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
@@ -53,14 +53,17 @@ They run in a context of an inner build with a single $(RuntimeIdentifier).
         AndroidApiLevel="$(_AndroidApiLevel)"
         MinimumSupportedApiLevel="$(AndroidMinimumSupportedApiLevel)"
         AndroidSequencePointsMode="$(_SequencePointsMode)"
-        AotAdditionalArguments="$(AndroidAotAdditionalArguments)"
         TargetName="$(TargetName)"
         ResolvedAssemblies="@(_AndroidAotInputs)"
         AotOutputDirectory="$(_AndroidAotBinDirectory)"
         RuntimeIdentifier="$(RuntimeIdentifier)"
         EnableLLVM="$(EnableLLVM)"
-        UsingAndroidNETSdk="true"
         Profiles="@(AndroidAotProfile)">
+      <Output PropertyName="_Triple"     TaskParameter="Triple" />
+      <Output PropertyName="_ToolPrefix" TaskParameter="ToolPrefix" />
+      <Output PropertyName="_MsymPath"   TaskParameter="MsymPath" />
+      <Output PropertyName="_LdName"     TaskParameter="LdName" />
+      <Output PropertyName="_LdFlags"    TaskParameter="LdFlags" />
       <Output ItemName="_MonoAOTAssemblies" TaskParameter="ResolvedAssemblies" />
     </GetAotAssemblies>
     <PropertyGroup>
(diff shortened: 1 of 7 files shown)


pjcollins (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
```diff
@@ -79,7 +85,11 @@ They run in a context of an inner build with a single $(RuntimeIdentifier).
         OutputType="Library"
         UseAotDataFile="false"
         UseLLVM="$(EnableLLVM)"
-        LLVMPath="$(_LLVMPath)">
+        LLVMPath="$(_LLVMPath)"
+        LdName="$(_LdName)"
+        LdFlags="$(_LdFlags)"
+        WorkingDirectory="$(MSBuildProjectDirectory)"
```
dellis1972, line 91: Will this still work if the user has not ascii chars in its `Configuration` ? e.g `Débug` ?

jonathanpeppers: This just gets it on par with Xamarin.Android now, so it's going to be the same.

Right now, I would guess that it would break.

src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
```diff
@@ -51,13 +51,24 @@ public abstract class GetAotArguments : AndroidAsyncTask
...
 
-		public bool UsingAndroidNETSdk { get; set; }
-
-		public string AotAdditionalArguments { get; set; } = "";
-
 		[Required, Output]
 		public ITaskItem [] ResolvedAssemblies { get; set; } = Array.Empty<ITaskItem> ();
 
+		[Output]
+		public string? Triple { get; set; }
```
jonpryor, line 58: Naming suggestion: `TargetTriplet` (as per [autoconf docs](https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Specifying-Target-Triplets.html)) instead of `Triple`.  `Triple` is not necessarily "fully semantically meaningful."

jonathanpeppers: Neither `TargetTriplet` or `Triple` really mean anything to me -- so open to suggestions.

The `<MonoAOTCompiler/>` task uses `Triple`, which passes `mtriple` to the actual AOT compiler:

https://github.com/dotnet/runtime/blob/bdc6ad73834a13e80f131e402e883903c31ea172/src/tasks/AotCompilerTask/MonoAOTCompiler.cs#L573-L576

We could use a different name, but is that worse? We'd still be using `Triple` on the next task unless we also renamed it -- and also got that change backported.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
dellis1972: ok as long as we are on par with classic this should be good.

dellis1972 (approved):

PR #6577, opened 2021-12-15
[CI] Run macOS test jobs on macOS 11

pjcollins:
Now that macOS 12 Monterey is publicly available, it seems like a good
time to upgrade our macOS test machines to last years macOS 11 Big Sur
release.

### DIFF
--- a/build-tools/automation/yaml-templates/variables.yaml
+++ b/build-tools/automation/yaml-templates/variables.yaml
@@ -22,7 +22,7 @@ variables:
 - name: GitHub.Token
   value: $(github--pat--vs-mobiletools-engineering-service2)
 - name: HostedMacImage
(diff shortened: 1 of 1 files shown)


pjcollins:
We hit a couple timeouts in [the previous run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5555129&view=results), though I have seen a few jobs timing out in another PR. Running the pipeline again for more result to try better compare timing and reliability.

pjcollins:
The [previous run](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5555970&view=results) was green on the first try and had similar timing compared to other PRs. Let's try another for good measure.

jonathanpeppers (approved):

PR #6585, opened 2021-12-20
[WIP] More dotnet build support for unix

pjcollins:


Makefile
```diff
@@ -22,7 +22,7 @@ _PREPARE_CI_MODE_ARGS = $(_PREPARE_CI_MODE_PR_ARGS) -a
 _PREPARE_ARGS =
 
 all:
-	$(call MSBUILD_BINLOG,all,$(_SLN_BUILD)) /restore $(MSBUILD_FLAGS) $(SOLUTION)
+	$(call DOTNET_BINLOG,all) $(MSBUILD_FLAGS) $(SOLUTION) -m:1
```
jonathanpeppers, line 25: Should `DOTNET_BINLOG` add the `-m:1`, so this is set in one place?

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks.csproj
```diff
@@ -19,4 +19,28 @@
...
+    <Exec
+        Command="lipo @(_MonoUnixDylib->'&quot;%(FullPath)&quot;', ' ') -create -output &quot;$(_MonoUnixOutputDylib)&quot;"
+        Condition=" '$(HostOS)' == 'Darwin' "
+    />
+    <Copy
+        SourceFiles="@(_MonoUnixDylib)"
+        DestinationFolder="$(OutputPath)"
+        Condition=" '$(HostOS)' != 'Darwin' "
+    />
+  </Target>
```
jonathanpeppers, line 44: Seems like we have this twice now:

https://github.com/xamarin/xamarin-android/blob/fc12409a0690ceea491be8f0a05890972da70860/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets#L373-L388

Is there a way to put this target in its own file and import twice? If we used the same MSBuild properties?

pjcollins: This was my original thought, but the overhead didn't seem worth it compared to copying a few lines. I think we'd want this if we need to do something similar in a third or fourth place in the future though, so I will try to adjust this.

Makefile
pjcollins: I'm not sure if we want to set it for every `dotnet` invocation as we should only need it when building Xamarin.Android.sln.

I haven't found great documentation on this parameter though. Neither the command line docs (below) or [this guide](https://docs.microsoft.com/en-us/visualstudio/msbuild/building-multiple-projects-in-parallel-with-msbuild?view=vs-2022) mention whether or not this is a "solution only" flag, or if it also works on projects with multiple project references.

```bash
  -maxCpuCount[:n]   Specifies the maximum number of concurrent processes to
                     build with. If the switch is not used, the default
                     value used is 1. If the switch is used without a value
                     MSBuild will use up to the number of processors on the
                     computer. (Short form: -m[:n])
```

The command line docs aren't consistent with what I see locally either. The Xamarin.Android.sln build fails for me in lots of weird ways if `-m:1` isn't set (even though the docs suggest the default value should be 1, which I assume would be "no parallelization" ). I know we had to set this explicitly on the Windows .sln build in the past for similar reasons as well.

The overall mac build job does seem faster in general in this PR, but I haven't done any digging into how much time is saved by 1) skipping the OSS zip creation, 2) using dotnet build more, or even the potential of 3) having more parallelization in some build commands by default.

My inclination is to keep it mostly as is, and only specify the `-m` flag when we know something breaks without it.

jonathanpeppers (approved):
LGTM now, if the builds end up green. 👍

jonpryor:
We shouldn't want/need to update `xabuild`; instead, for `dotnet`-based builds we should use the `$DOTNETSDK_WORKLOAD_PACK_ROOTS` and `$DOTNETSDK_WORKLOAD_MANIFEST_ROOTS` environment variables:

  * https://github.com/dotnet/sdk/issues/18104
  * https://github.com/dotnet/sdk/pull/18823
  * https://github.com/dotnet/sdk/blob/9db8eacc7e48c85b5b6821684b25d42c2d9b345d/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/WorkloadResolver.cs#L51-L55
  * https://github.com/dotnet/sdk/blob/22c4860dcb2cf6b123dd641cc4a87a50380759d5/src/Common/EnvironmentVariableNames.cs#L9-L10

jonathanpeppers:
> MSBuild is still required to build xabuild.csproj and create-vsix.csproj.

I think this is saying you still need to build `xabuild` with MSBuild, and you can't use `dotnet build`.

I don't currently see why we would need `xabuild` at all in .NET 6. Only benefit would be to speed up our development loop to avoid `dotnet pack` and `dotnet workload install` commands?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.targets
```diff
@@ -316,8 +316,7 @@
   <Target Name="ILRepacker"
       BeforeTargets="CopyFilesToOutputDirectory"
       Inputs="$(MSBuildAllProjects);@(IntermediateAssembly);@(InputAssemblies)"
-      Outputs="$(IntermediateOutputPath)ILRepacker.stamp"
-      Condition=" '$(HostOS)' != 'Linux' ">
```
dellis1972, line 320: Can anyone remember the reason why we disabled this in the first place?

grendello: It used to crash on Linux under Mono

pjcollins:
Commit message draft:

```markdown
Bumps to mono/debugger-libs@c53676b
Changes: https://github.com/mono/debugger-libs/compare/08268b730a73cca0e3b3ac9c703f0b773a3be25b...c53676b08005c806fc170b7be5cda25e52fe9054

Many instances of `msbuild` have been replaced with `dotnet build` in
our make files.  The `-m:1` flag has been added to `Xamarin.Android.sln`
builds to disable parallelization.  We still run into issues running
`dotnet build` on the .sln when this flag is not used.  Additionally,
`msbuild` is still required when building `xabuild.csproj` and
`create-vsix.csproj`.

The `make jenkins` target has been updated to no longer create an OSS
.zip file to improve build speeds.  OSS installers can be built
separately as needed.

The `Xamarin.Android.Build.Tasks` build has been updated to run ILRepack
on Linux to fix [various issues][0].

A new `MonoUnixNative.targets` file has been added that can be used by
projects that produce an assembly containing MSBuild tasks that depend
on Mono.Unix.  The `<_MakeMonoUnixFatBinariesOSX/>` target will create
a multiarchitecture version of `libMono.Unix.dylib` in the projects
output directory when running on macOS.  It will also copy
`runtimes\linux-x64\native\libMono.Unix.so` and
`runtimes\osx-x64\native\libMono.Unix.dylib` to the output directory
when running on Windows or Linux.

[0]: https://github.com/xamarin/xamarin-android/pull/5681#issuecomment-790017891
```

dellis1972 (approved):

grendello (approved):

PR #6588, opened 2021-12-21
[build] specify path for self checkouts

jonathanpeppers:
We have a private mirror, for servicing security fixes, or testing
builds of .NET with security fixes.

By default, the private mirror, `xamarin-android-private` is checking
out into:

    Finishing: Checkout xamarin/xamarin-android-private@main to s/xamarin-android-private

Which causes the build to fail on a later step:

    Could not find a part of the path '/Users/builder/azdo/_work/7/s/xamarin-android/build-tools/provisioning/xcode.csx'

Specify the path when checking out `self` to solve this issue.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -125,6 +125,7 @@ stages:
     steps:
     - checkout: self
       submodules: recursive
+      path: s/xamarin-android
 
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
@pjcollins this is a commit on the private mirror, we should just bring over.

pjcollins (approved):

jonathanpeppers:
Going to merge this, the test failure seems unrelated.

Issue #6142, opened 2021-07-29
Bind library that uses Java 11 source code


4brunu:
Hi,

Since Android Studio Gradle Plugin, it's possible to use some Java 11 source code in all Android versions
https://developer.android.com/studio/releases/gradle-plugin#java-11

Is it possible to bind a native android library (.aar), that uses Java 11 source code?

I tried to create a native android library and deployed it in a sample project to Android 5.1.

```
public class JavaElevenAPI {

    public static String stringTest() {

        var arrInteger = new Integer[]{5, 9, 3, 6, 2, 4, 8, 7, 1};
        long cnt = Arrays.stream(arrInteger).filter(
                (@NotNull var a) -> (a != null && a > 5)).count();
        System.out.println(cnt);

        return Long.toString(cnt);
    }
}
```

Is Android Studio with the native library and the native app, everything worked as expected.

When binding the native library in Xamarin, it builds, but it cashes in runtime with the following error.

```
java.lang.NoSuchMethodError
No static method stream([Ljava/lang/Object;)Ljava/util/stream/Stream; in class Ljava/util/Arrays; or its super classes (declaration of 'java.util.Arrays' appears in /system/framework/core-libart.jar)
```

For this to work, it's necessary to use in conjunction with Core Library Dessugaring.
https://developer.android.com/studio/write/java8-support#library-desugaring

Since `coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'` doesn't exist in Xamarin, I added that library as a dependency. 

Is there any way to solve this?
Is this supported in Xamarin?

Thanks

jpobst:
Did adding `coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'` as a dependency fix the issue?

4brunu:
No.

But actually, now that you are mentioning this, I made some more tests, and I think the problem is related to `coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'` that work in some cases, like `java.time` but not with `streams`.
Java features, like java 10 `var`, works!

```
public class JavaElevenAPI {
    public static String stringTestVar() {
        // var - java 10 feature - works!
        // doesn't need coreLibraryDesugaring
        var arrInteger = new Integer[]{5, 9, 3, 6, 2, 4, 8, 7, 1};

        return Long.toString(arrInteger[3]);
    }

    public static String stringTestArrays() {
[...]

PR #6591, opened 2021-12-22
[Xamarin.Android.Build.Tasks] MAM Member Remapping?

jonpryor:
[Mono.Android] Type & Member Remapping Support

Context: https://github.com/xamarin/java.interop/pull/936
Context: https://github.com/xamarin/xamarin-android/issues/6142#issuecomment-889435599

Fixes: https://github.com/xamarin/java.interop/issues/867

~~ The Scenarios ~~

Our Java binding infrastructure involves looking up types and methods
via [JNI][0], and assumes that types and methods won't "move" in an
unexpected manner.  Methods which move from a subclass to a
superclass works transparently.  Methods which are moved to an
entirely different class causes us problems.

Case in point: [desugaring][0], which can *move* Java types to places
that our bindings don't expect.  For example, [`Arrays.stream(T[])`][1]
may be moved into a `DesugarArrays` class, or a default interface
method `Example.m()` may be moved into an `Example$-CC` type.
Java.Interop has not supported such changes, resulting in throwing a
`Java.Lang.NoSuchMethodError` on Android versions where methods are
not where we expect.

Additionally, the [InTune Mobile Application Management][3] team
needs a expanded type and member lookup mechanism in order to
simplify how they maintain their product.  Currently they make things
work by rewriting IL, which can be brittle.

~~ The Pieces ~~

To improve support for this, xamarin/java.interop#936 introduces
new `virtual` methods into `Java.Interop.JniRuntime.JniTypeManager`
which are called as part of type and member lookup, allowing
`AndroidTypeManager` to participate in the type and member resolution
process.

`AndroidTypeManager` in turn needs to know what types and members can
be remapped, and what they should be remapped to.  Some of these can
be algorithmic, such as pre-pending `Desugar` or appending `$-CC` for
the Desugar case.

The InTune use case involves a table, contained within the
[Microsoft.Intune.MAM.Remapper.Tasks NuGet package][4].

Update `src/Xamarin.Android.Build.Tasks` to add a new
`@(_AndroidMamMappingFile)` Build action.  Files with this build
action are expected to be JSON documents which follow the current
conventions of `remapping-config.json`, within the
`Microsoft.Intune.MAM.Remapper.Tasks` NuGet package.  This build
action is not externally supported; this is currently for testing
purposes.  `@(_AndroidMamMappingFile)` files are processed into an
XML intermediary as part of the App build, and bundled into the app
as an `@(AndroidAsset)` named `xa-internal/xa-mam-mapping.xml`:

    <replacements>
      <replace-type
          from="android/app/Activity"
          to="com/microsoft/intune/mam/client/app/MAMActivity" />
      <replace-method
          source-type="com/microsoft/intune/mam/client/app/MAMActivity"
          source-method-name="onCreate" source-method-signature="(Landroid/os/Bundle;)"
          target-type="com/microsoft/intune/mam/client/app/MAMActivity"
          target-method-name="onMAMCreate" target-method-instance-to-static="false" />
    </replacements>

`//replacements/replace-method` is structured with each attribute
corresponding to a member on the `JniRuntime.ReplacementMethodInfo`
struct, in xamarin/java.interop#936.

  * `//replace-method/@source-type` is
    `JniRuntime.ReplacementMethodInfo.SourceJniType`
  * `//replace-method/@source-method-name` is
    `JniRuntime.ReplacementMethodInfo.SourceJniMethodName`
  * `//replace-method/@source-method-signature` is
    `JniRuntime.ReplacementMethodInfo.SourceJniMethodSignature`
  * `//replace-method/@target-type` is
    `JniRuntime.ReplacementMethodInfo.TargetJniType`
  * `//replace-method/@target-method-name` is
    `JniRuntime.ReplacementMethodInfo.TargetJniMethodName`
  * `//replace-method/@target-method-signature` is
    `JniRuntime.ReplacementMethodInfo.SourceJniMethodSignature`
  * `//replace-method/@target-method-instance-to-static` is
    `JniRuntime.ReplacementMethodInfo.TargetJniMethodIsStatic`

During app startup on the Android device, the
`xa-internal/xa-mam-mapping.xml` asset is opened and provided to
`JNIEnv.Initialize()` as part of native app startup.

~~ Putting it all together ~~

This will only work on .NET 7+.

App project has a `@(_AndroidMamMappingFile)` file.  This item is
processed during App build, stored into the `.apk`, and read during
app startup on an Android device.

Given a Java binding such as:

    public partial class Activity {
        static readonly JniPeerMembers _members = new XAPeerMembers ("android/app/Activity", typeof (Activity));
    }

when the `JniPeerMembers` constructor runs, it will call
`JniEnvironment.Runtime.TypeManager.GetReplacementType("android/app/Activity")`.
If `@(_AndroidMamMappingFile)` uses the InTune `remapping-config.json`
file, then `com/microsoft/intune/mam/client/app/MAMActivity` will
be resolved via `JNIEnv::FindClass()`, *not* `android/app/Activity`.

*If `MAMActivity` can't be found*, e.g. you're testing this all out,
the app will ~immediately crash, as `MAMActivity` doesn't exist. 😅

If `MAMActivity` can be found, eventually `Activity.OnCreate()` will
need to be invoked:

	partial class Activity {
	    {
	        try {
	        } finally {
	        }
	    }
	}

`_members.InstanceMethods.InvokeVirtualVoidMethod()` will internally
make a call similar to:

	var r = JniEnvironment.Runtime.TypeManager.GetReplacementMethodInfo (
	    "android/app/Activity",
	    "onCreate",
	    "(Landroid/os/Bundle;)V"
	);

The data returned will be equivalent to:

	var r = new JniRuntime.ReplacementMethodInfo {
	    SourceJniType                   = "android/app/Activity",                               // from input parameter
	    SourceJniMethodName             = "onCreate",                                           // from input parameter
	    SourceJniMethodSignature        = "(Landroid/os/Bundle;)V",                             // from input parameter

	    TargetJniType                   = "com/microsoft/intune/mam/client/app/MAMActivity",    // from //replace-method/@target-type
	    TargetJniMethodName             = "onMAMCreate",                                        // from //replace-method/@target-method-name
	    SourceJniMethodSignature        = "(Landroid/os/Bundle;)V",                             // from input parameter, as signature didn't change
	    TargetJniMethodParameterCount   = 1,                                                    // computed based on signature
	    TargetJniMethodIsStatic         = false,                                                // from //replace-method/@target-method-instance-to-static
	}

This will allow `_members.InstanceMethods.InvokeVirtualVoidMethod()`
to instead resolve and invoke `MAMActivity.onMAMCreate()`.

[0]: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html
[1]: https://developer.android.com/studio/write/java8-support#library-desugaring
[2]: https://developer.android.com/reference/java/util/Arrays#stream(T[])
[3]: https://docs.microsoft.com/en-us/mem/intune/fundamentals/what-is-intune
[4]: https://www.nuget.org/packages/Microsoft.Intune.MAM.Remapper.Tasks/


src/Xamarin.Android.Build.Tasks/Utilities/MamJsonParser.cs
```diff
@@ -0,0 +1,215 @@
...
+					var o       => JavaToJniTypeSignature (o),
+				};
+				if (javaType == null) {
+					continue;
+				}
+				types.Add (javaType);
+			}
+			if (types.Count == 0)
+				return null;
+			return "(" + string.Join ("", types) + ")";
```
grendello, line 131: I wonder if using a `StringBuilder` here (and getting rid fo the `types` list) would be a bit faster?

src/Xamarin.Android.Build.Tasks/Utilities/MamJsonParser.cs
```diff
@@ -0,0 +1,215 @@
...
+		{
+			if (javaType == null) {
+				return null;
+			}
+			int arrayCount = 0;
+			while (javaType.EndsWith ("[]")) {
+				arrayCount++;
+				javaType = javaType.Substring(0, javaType.Length - "[]".Length);
+			}
+			return new string('[', arrayCount) + "L" + JavaToJniType (javaType) + ";";
```
grendello, line 149: Like wise here, a `StringBuilder` could be used (including as parameter to `JavaToJniType`) and instead of creating a substring, a `ReadOnlySpan` could be used on line 147, thus minimizing expensive string operations.

src/Xamarin.Android.Build.Tasks/Utilities/MamJsonParser.cs
jonpryor: Updated to use `StringBuilder`.

samples/HelloWorld/HelloWorld/HelloWorld.DotNet.csproj
```diff
@@ -10,4 +10,19 @@
...
+        GeneratePathProperty="True"
+        ReferenceOutputAssembly="False"
+    />
+  </ItemGroup>
+  <Target Name="_AddMamFiles"
+      BeforeTargets="_AddAndroidCustomMetaData">
+    <ItemGroup>
+      <_AndroidMamMappingFile Include="$(PkgMicrosoft_Intune_MAM_Remapper_Tasks)/content/MonoAndroid10/remapping-config.json" />
+    </ItemGroup>
+  </Target>
```
jonathanpeppers, line 27: Should we really put this in the Hello World project? Should it maybe be an MSBuild test in Xamarin.Android.Build.Tests?

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -275,13 +280,139 @@ protected override IEnumerable<string> GetSimpleReferences (Type type)
...
+			};
+		}
+
+		protected override string? GetReplacementTypeCore (string jniSimpleReference)
+		{
+			if (JNIEnv.ReplacementTypes == null) {
+				return null;
+			}
+			Logger.Log (LogLevel.Warn, "*jonp*", $"# jonp: looking for replacement type for `{jniSimpleReference}`");
+			Logger.Log (LogLevel.Warn, "*jonp*", new System.Diagnostics.StackTrace (true).ToString ());
```
jonathanpeppers, line 314: Probably still a lot of `jonp` print statements throughout.

src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -203,6 +215,14 @@ internal static unsafe void Initialize (JnienvInitializeArgs* args)
 			SynchronizationContext.SetSynchronizationContext (Android.App.Application.SynchronizationContext);
 #endif
 
+#if NET
+			if (args->mappingXml != IntPtr.Zero) {
+				var xml = Encoding.UTF8.GetString ((byte*) args->mappingXml, args->mappingXmlLen);
+				Logger.Log (LogLevel.Warn, "*jonp*", $"# jonp: mapping xml: len={args->mappingXmlLen}; {xml}");
+				(ReplacementTypes, ReplacementMethods) = MamXmlParser.ParseStrings (xml);
```
jonathanpeppers, line 222: Instead of using XML here at all, could we generate a type in the main assembly that has this data? So you would just use reflection to look for a couple well-known class and field names?

Maybe that is more work than we're willing to do right now.

src/Mono.Android/Android.Runtime/MamXmlParser.cs
```diff
@@ -0,0 +1,159 @@
...
+using ReplacementTypesDict              = System.Collections.Generic.Dictionary<string, string>;
+using ReplacementMethodsDictStrings     = System.Collections.Generic.Dictionary<string, string>;
+using ReplacementMethodsDictStructured  = System.Collections.Generic.Dictionary<
+	(string SourceType, string SourceName, string? SourceSignature),
+	(string? TargetType, string? TargetName, string? TargetSignature, int? ParamCount, bool IsStatic)
+>;
+
+namespace Android.Runtime {
+
+	class MamXmlParser {
```
jonathanpeppers, line 21: If we went with the C# code generation idea, maybe this class could write a `.cs` file instead of XML? We don't have any Roslyn source-generators yet, so that would probably be the simplest route.

src/Mono.Android/Mono.Android.csproj
```diff
@@ -22,6 +22,7 @@
     <Nullable>enable</Nullable>
     <ProduceReferenceAssembly>true</ProduceReferenceAssembly>
     <EnableSingleFileAnalyzer>true</EnableSingleFileAnalyzer>
+    <EnablePreviewFeatures>True</EnablePreviewFeatures>
```
jonathanpeppers, line 25: Probably won't be able to keep this line when we merge.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -432,6 +433,25 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
...
+      MappingFiles="@(_AndroidMamMappingFile)"
+      XmlMappingOutput="$(_AndroidMamMappingXml)"
+  />
+  <ItemGroup
+      Condition=" Exists ('$(_AndroidMamMappingXml)') ">
+    <AndroidAsset Include="$(_AndroidMamMappingXml)" LogicalName="xa-internal/xa-mam-mapping.xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <FileWrites Include="$(_AndroidMamMappingXml)" />
+  </ItemGroup>
```
jonathanpeppers, line 454: Could this chunk of MSBuild be in its own target? I'm thinking we'd want a `Condition` that skips the target:

```xml
Condition=" '@(_AndroidMamMappingFile->Count())' != '0' "
```

And then maybe the Inputs/Outputs would be:
```xml
Inputs="$(_AndroidBuildPropertiesCache);@(_AndroidMamMappingFile)"
Outputs="$(_AndroidMamMappingXml)"
```

src/java-runtime/java/mono/android/MonoPackageManager.java
```diff
@@ -162,4 +166,31 @@ public static String[] getDependencies ()
...
+				if (!asset.equals ("xa-mam-mapping.xml")) {
+					continue;
+				}
+				Log.d ("*jonp*", "# jonp: found `xa-mam-mapping.xml`; trying to load...");
+				InputStream s = manager.open ("xa-internal/xa-mam-mapping.xml");
+				Log.d ("*jonp*", "# jonp: s? " + (s != null));
+				byte[] contents = new byte[s.available ()];
+				int r = s.read (contents);
+				Log.d ("*jonp*", "# jonp: read " + r + " bytes from inputstream! expected " + contents.length + "!");
+				s.close ();
```
jonathanpeppers, line 188: Java doesn't have the C# `using` block, so is one supposed to call `close()` in a `try`-`finally` block?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -2422,7 +2432,7 @@ JNI_OnLoad (JavaVM *vm, void *reserved)
 JNIEXPORT void JNICALL
 Java_mono_android_Runtime_init (JNIEnv *env, jclass klass, jstring lang, jobjectArray runtimeApksJava,
                                 jstring runtimeNativeLibDir, jobjectArray appDirs, jobject loader,
-                                [[maybe_unused]] jobjectArray externalStorageDirs, jobjectArray assembliesJava, [[maybe_unused]] jstring packageName,
+                                [[maybe_unused]] jobjectArray externalStorageDirs, jobjectArray assembliesJava, jbyteArray mappingXml, jint mappingXmlLen, [[maybe_unused]] jstring packageName,
```
jonathanpeppers, line 2435: If you add new parameters here, does that break the Android designer? I think that is why we made `Java_mono_android_Runtime_initInternal `.

src/Mono.Android/Android.Runtime/JNIEnv.cs
jonpryor: See my [other comment](https://github.com/xamarin/xamarin-android/pull/6591#discussion_r870607684), but…. We don't *want* to "privilege" the "main" assembly ("what's a main assembly?"), because when we *have*, things *break*; see e.g. 807e66501257d2adee731df1ed501301e2083530, wherein an app crashed because the Activity launched *wasn't* in the "main" assembly.

The "best" fix is to put this data into `libxamarin-app.so`.  I think we can do that by e.g. .NET 7 P7.

src/Mono.Android/Mono.Android.csproj
jonpryor: We don't need it now, as I've updated Java.Interop to stop using the "preview features" attributes.  I just forgot about this.

samples/HelloWorld/HelloWorld/HelloWorld.DotNet.csproj
jonpryor: Beginnings of a "lesser" test just pushed, via introduction of a new `@(_AndroidRemapMembers)` build action.  This uses the same XML format that the InTune JSON is converted to, so we can test things by writing our XML format directly, instead of going through a JSON intermediary.

TODO: add an actual on-device unit test which does this.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
jonpryor: I wish i could figure out how to have a "nice, short, readable" `Outputs` value, instead of the "full path" I'm currently using, but otherwise… this has been done.  More targets, smaller targets.

src/java-runtime/java/mono/android/MonoPackageManager.java
jonpryor: Java now has a "try-with-resources" construct, akin to C# `using` blocks.  Code updated to use that.

tests/MSBuildDeviceIntegration/Tests/XASdkDeployTests.cs
```diff
@@ -156,6 +157,70 @@ public void DotNetDebug ()
 			WaitFor (2000);
 			Assert.IsTrue (breakpointHit, "Should have a breakpoint");
 		}
+
+		[Test]
+		public void TypeAndMemberRemapping ([Values (false, true)] bool isRelease)
```
jonpryor, line 162: @dellis1972 , @jonathanpeppers : please review this test for sanity. :-)

(I have no idea what I'm doing here. [dog-on-keyboard.gif])

tests/MSBuildDeviceIntegration/Tests/XASdkDeployTests.cs
```diff
@@ -156,6 +157,70 @@ public void DotNetDebug ()
...
+			proj.SetRuntimeIdentifier (DeviceAbi);
+			var relativeProjDir = Path.Combine ("temp", TestName);
+			var fullProjDir     = Path.Combine (Root, relativeProjDir);
+			TestOutputDirectories [TestContext.CurrentContext.Test.ID] = fullProjDir;
+			var files = proj.Save ();
+			proj.Populate (relativeProjDir, files);
+			proj.CopyNuGetConfig (relativeProjDir);
+			var dotnet = new DotNetCLI (proj, Path.Combine (fullProjDir, proj.ProjectFilePath));
+
+			Assert.IsTrue (dotnet.Build (), "`dotnet build` should succeed");
```
jonathanpeppers, line 205: The other `XASdkTests` class has some helper methods:

https://github.com/xamarin/xamarin-android/blob/1efa0cf46c9079fc06669a4434f597faf47504af/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs#L1071-L1094

Should we put these somewhere, so that `XASdkDeployTests` can use them?

Then you can do:
```csharp
var builder = CreateDotNetBuilder (proj);
Assert.IsTrue (builder.Build (), $`dotnet build` should succeed");
```

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -402,7 +404,8 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidCustomMetaDataForReferences Include="@(_AndroidAssemblySkipCases)" />
 </ItemGroup>
 
-<Target Name="_AddAndroidCustomMetaData">
+<Target Name="_AddAndroidCustomMetaData"
+    DependsOnTargets="_ConvertAndroidMamMappingFileToXml;_CollectAndroidRemapMembers">
```
dellis1972, line 408: I *think* (and @jonathanpeppers can correct me :) ) we tend to prefer that we explicitly define the dependencies in
`src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.BuildOrder.targets` for .NET and `src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets` for Legacy. We can do this in addition to `DependsOnTargets` but it does help when figuring out the build order if we have it in the other files too.

This is for .NET 7+ only right?

tests/MSBuildDeviceIntegration/Tests/XASdkDeployTests.cs
dellis1972: This test looks good. But making use of methods like `CreateDotNetBuilder` would make it a bit nicer :)

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
jonpryor: Correct, this is .NET 7+ only.

src/monodroid/jni/monodroid-glue.cc
jonpryor: Fixed.  Extra parameters removed from `Runtime.initInternal()`, designer tests now passing.

tests/MSBuildDeviceIntegration/Tests/XASdkDeployTests.cs
```diff
@@ -83,6 +84,57 @@ public void DotNetInstallAndRun (bool isRelease, bool xamarinForms, bool? publis
 			Assert.IsTrue(didLaunch, "Activity should have started.");
 		}
 
+		[Test]
+		public void TypeAndMemberRemapping ([Values (false, true)] bool isRelease)
```
jonpryor, line 88: This test never runs, for reasons I cannot currently explain or fathom.  It's in the downloaded `MSBuildDeviceIntegration.dll`.  It's not `[Values]`; `ApplicationRunsWithoutDebugger` *is* run.  It's not position in the file (moving it didn't change anything).

Meanwhile, I *can* run it locally, via the updated docs:

    ./dotnet-local.sh test bin/TestDebug/MSBuildDeviceIntegration/net6.0/MSBuildDeviceIntegration.dll --filter "Name~TypeAndMemberRemapping"

Fortunately, type and member remapping is *also* tested as part of `Java.Interop-Tests.dll`, included in `Mono.Android.NET-Tests.apk`, so I'm confident it works.  It would just be nice if this test *also* worked.  Or at least *ran*.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
```diff
@@ -88,9 +88,21 @@ public void CheckIncludedAssemblies ([Values (false, true)] bool usesAssemblySto
...
 					"Mono.Android.dll",
 					"rc.bin",
+					"System.Collections.dll",
+					"System.Collections.Concurrent.dll",
+					"System.Collections.NonGeneric.dll",
+					"System.Console.dll",
+					"System.IO.Compression.dll",
+					"System.Net.Http.dll",
+					"System.Net.Primitives.dll",
+					"System.Net.Requests.dll",
```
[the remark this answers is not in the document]
jonpryor: Yes; probably the new `XmlReader` use in `JNIEnv.Initialize()`.

Presumably these will "disappear" once we do the "non-prototype" approach of putting al this data into `libxamarin-app.so`…

src/Mono.Android/Android.Runtime/MamXmlParser.cs
```diff
@@ -0,0 +1,159 @@
...
+			var (types, methodsStructured) = ParseStructured (xml);
+
+			var methodsStrings  = new ReplacementMethodsDictStrings ();
+			foreach (var e in methodsStructured) {
+				var key     = $"{e.Key.SourceType}\t{e.Key.SourceName}\t{e.Key.SourceSignature}";
+				var value   = $"{e.Value.TargetType}\t{e.Value.TargetName}\t{e.Value.TargetSignature}\t{e.Value.ParamCount?.ToString () ?? ""}\t{(e.Value.IsStatic ? "true" : "false")}";
+				methodsStrings [key] = value;
+			}
+
+			return (types, methodsStrings);
```
grendello, line 43: It might be faster to use `out` parameters for the dictionaries, to avoid instantiating a tuple

src/monodroid/jni/monodroid-glue-internal.hh
```diff
@@ -174,7 +176,8 @@ namespace xamarin::android::internal
 		void Java_mono_android_Runtime_register (JNIEnv *env, jstring managedType, jclass nativeClass, jstring methods);
 		void Java_mono_android_Runtime_initInternal (JNIEnv *env, jclass klass, jstring lang, jobjectArray runtimeApksJava,
 		                                             jstring runtimeNativeLibDir, jobjectArray appDirs, jobject loader,
-		                                             jobjectArray assembliesJava, jint apiLevel, jboolean isEmulator,
+		                                             jobjectArray assembliesJava, jbyteArray mappingXml, jint mappingXmlLen,
+		                                             jint apiLevel, jboolean isEmulator,
```
grendello, line 180: I wish we could somehow kill this monster of a method... So many arguments :)

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1192,6 +1192,12 @@ MonodroidRuntime::init_android_runtime (
 		native_to_managed_index = internal_timing->start_event (TimingEventKind::NativeToManagedTransition);
 	}
 
+	if (mappingXml != nullptr && mappingXmlLen > 0) {
+		init.mappingXml             = env->GetByteArrayElements (mappingXml, nullptr);
+		init.mappingXmlLen          = mappingXmlLen;
+	}
+	log_warn (LOG_DEFAULT, "# jonp: mappingXml? len=%i, xml=%p", init.mappingXmlLen, init.mappingXml);
```
grendello, line 1199: Ditto logging

.gitmodules
```diff
@@ -12,8 +12,8 @@
     branch = main
 [submodule "external/Java.Interop"]
     path = external/Java.Interop
-    url = https://github.com/xamarin/java.interop.git
-    branch = main
+    url = https://github.com/jonpryor/java.interop.git
+    branch = jonp-member-remapping
```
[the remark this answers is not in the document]
jonpryor: Yes, when https://github.com/xamarin/java.interop/pull/936 is merged, but not until then.

Please review https://github.com/xamarin/java.interop/pull/936 ;-)

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -275,13 +280,149 @@ protected override IEnumerable<string> GetSimpleReferences (Type type)
...
+		protected override JniRuntime.ReplacementMethodInfo? GetReplacementMethodInfoCore (string jniSourceType, string jniMethodName, string jniMethodSignature)
+		{
+			if (JNIEnv.ReplacementMethods == null) {
+				return null;
+			}
+			Logger.Log (LogLevel.Warn, "*jonp*", $"# jonp: looking for replacement method for (\"{jniSourceType}\", \"{jniMethodName}\", \"{jniMethodSignature}\")");
+#if !STRUCTURED
+			if (!JNIEnv.ReplacementMethods.TryGetValue ($"{jniSourceType}\t{jniMethodName}\t{jniMethodSignature}", out var r) &&
+					!JNIEnv.ReplacementMethods.TryGetValue ($"{jniSourceType}\t{jniMethodName}\t{GetMethodSignatureWithoutReturnType ()}", out r) &&
+					!JNIEnv.ReplacementMethods.TryGetValue ($"{jniSourceType}\t{jniMethodName}\t", out r)) {
```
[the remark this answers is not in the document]
jonpryor: Would be interesting to profile that -- interpolated strings are *supposed* to be getting lots of perf love.

Fortunately, someone used Benchmark.net: https://blog.ladeak.net/posts/string-interpolation-stringbuilder

String interpolation isn't the worst, but it's not the best.  Using `StringBuilder.Append()` is the 2nd fastest.

For now, I'll add a new method which uses `StringBiulder`.  (Pity `string.Concat()` only takes up to 4 parameters!  5 would have been best.)

src/Mono.Android/Android.Runtime/MamXmlParser.cs
jonpryor: will be moot once this is all in `libxamarin-app.so`. :-)

src/Mono.Android/Android.Runtime/MamXmlParser.cs
```diff
@@ -0,0 +1,159 @@
...
+					break;
+				case "to":
+					to      = reader.Value;
+					break;
+				}
+			}
+			if (string.IsNullOrEmpty (from) || string.IsNullOrEmpty (to)) {
+				return;
+			}
+			replacementTypes [from] = to;
```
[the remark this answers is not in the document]
jonpryor: And chance an exception if there *are* duplicates?

Something to consider in future revisions.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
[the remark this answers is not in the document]
jonpryor: This week?  Yes.

Once we get all this re-hosted on `libxamarin-app.so`, no.

src/monodroid/jni/monodroid-glue.cc
jonpryor: For now, i'll leave this in so that there is *some* runtime indicator that mapping is being used, but i'll move it into the above `if` block so it's only printed *when used*.

src/Mono.Android/Android.Runtime/MamXmlParser.cs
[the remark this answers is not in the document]
grendello: Well, it depends on whether it matters or not. If it's not an error to overwrite an earlier value, then this code's fine :)

jonpryor:
```markdown
[Mono.Android] Type & Member Remapping Support

Fixes: https://github.com/xamarin/java.interop/issues/867

Context: https://github.com/xamarin/java.interop/commit/1f27ab552d03aeb74cdc6f8985fcffbfdb9a7ddf
Context: https://github.com/xamarin/xamarin-android/issues/6142#issuecomment-889435599
Context: https://github.com/xamarin/xamarin-android/issues/7020

Changes: https://github.com/xamarin/java.interop/compare/843f3c7817dc4bdae9ce69d04274f29fce574e09...1f27ab552d03aeb74cdc6f8985fcffbfdb9a7ddf

  * xamarin/java.interop@1f27ab55: [Java.Interop] Type & Member Remapping Support (#936)
  * xamarin/java.interop@02aa54e0: [Java.Interop.Tools.JavaCallableWrappers] marshal method decl types (#987)
  * xamarin/java.interop@e7bacc37: [ci] Update azure-pipelines.yaml to Pin .NET 6.0.202 (#986)
  * xamarin/java.interop@fb94d598: [Java.Interop.Tools.JavaCallableWrappers] Collect overriden methods (#985)
  * xamarin/java.interop@3fcce746: [Java.Interop.{Dynamic,Export}] Nullable Reference Type support (#980)

~~ The Scenarios ~~

Our Java binding infrastructure involves looking up types and methods
via [JNI][0], and assumes that types and methods won't "move" in an
unexpected manner.  Methods which move from a subclass to a
superclass works transparently.  Methods which are moved to an
entirely different class causes us problems.

Case in point: [desugaring][0], which can *move* Java types to places
that our bindings don't expect.  For example, [`Arrays.stream(T[])`][1]
may be moved into a `DesugarArrays` class, or a default interface
method `Example.m()` may be moved into an `Example$-CC` type.
Java.Interop has not supported such changes, resulting in throwing a
`Java.Lang.NoSuchMethodError` on Android versions where methods are
not where we expect.

Additionally, the [InTune Mobile Application Management][3] team
needs a expanded type and member lookup mechanism in order to
simplify how they maintain their product.  Currently they make things
work by rewriting IL, which can be brittle.


~~ Build actions ~~

To improve support for this, xamarin/java.interop#936 introduces
new `virtual` methods into `Java.Interop.JniRuntime.JniTypeManager`
which are called as part of type and member lookup, allowing
`AndroidTypeManager` to participate in the type and member resolution
process.

`AndroidTypeManager` in turn needs to know what types and members can
be remapped, and what they should be remapped to.  Some of these can
be algorithmic, such as pre-pending `Desugar` or appending `$-CC` for
the Desugar case.

The InTune use case involves a table, contained within the
[Microsoft.Intune.MAM.Remapper.Tasks NuGet package][4].

Update `src/Xamarin.Android.Build.Tasks` to add a new
`@(_AndroidRemapMembers)` Build action.  This build action is not
externally supported; it's to help test the feature.  Files with this
build action are XML files which control type and member remapping:

	<replacements>
	  <replace-type
	      from="android/app/Activity"
	      to="com/microsoft/intune/mam/client/app/MAMActivity" />
	  <replace-method
	      source-type="com/microsoft/intune/mam/client/app/MAMActivity"
	      source-method-name="onCreate" source-method-signature="(Landroid/os/Bundle;)V"
	      target-type="com/microsoft/intune/mam/client/app/MAMActivity"
	      target-method-name="onMAMCreate" target-method-instance-to-static="false" />
	</replacements>

`//replacements/replace-method` is structured with each attribute
corresponding to a member on the `JniRuntime.ReplacementMethodInfo`
structure, in xamarin/java.interop@1f27ab55.

  * `//replace-method/@source-type` is
    `JniRuntime.ReplacementMethodInfo.SourceJniType`
  * `//replace-method/@source-method-name` is
    `JniRuntime.ReplacementMethodInfo.SourceJniMethodName`
  * `//replace-method/@source-method-signature` is
    `JniRuntime.ReplacementMethodInfo.SourceJniMethodSignature`
  * `//replace-method/@target-type` is
    `JniRuntime.ReplacementMethodInfo.TargetJniType`
  * `//replace-method/@target-method-name` is
    `JniRuntime.ReplacementMethodInfo.TargetJniMethodName`
  * `//replace-method/@target-method-signature` is
    `JniRuntime.ReplacementMethodInfo.TargetJniMethodSignature`
    This attribute is optional.
  * `//replace-method/@target-method-parameter-count` is
    `JniRuntime.ReplacementMethodInfo.TargetJniMethodParameterCount`.
    This attribute is optional.
  * `//replace-method/@target-method-instance-to-static` is
    `JniRuntime.ReplacementMethodInfo.TargetJniMethodIsStatic`

`@source-type`, `@source-method-name`, and `@source-method-signature`
combined serve as a "key" for looking up the associated `@target-*`
information.

Update `src/Xamarin.Android.Build.Tasks` to add a new
`@(_AndroidMamMappingFile)` Build action.  This build action is not
externally supported; it's to help test the feature.  Files with this
build action are expected to be JSON documents which follow the
current conventions of `remapping-config.json`, within the
`Microsoft.Intune.MAM.Remapper.Tasks` NuGet package.  This build
action is not externally supported; this is currently for testing
purposes.  `@(_AndroidMamMappingFile)` files are processed at build
time into `@(_AndroidRemapMembers)` XML files.

During App builds, all `@(_AndroidRemapMembers)` files are merged
into an `@(AndrodAsset)` named `xa-internal/xa-mam-mapping.xml`.
This asset is opened and provided to `JNIEnv.Initialize()` as part of
native app startup.


~~ Putting it all together ~~

This will only work on .NET 7+.

App project has a `@(_AndroidRemapMembers)` file.  This item is
processed during App build, stored into the `.apk`, and read during
app startup on an Android device.

Given a Java binding such as:

	public partial class Activity {
	    static readonly JniPeerMembers _members = new XAPeerMembers ("android/app/Activity", typeof (Activity));
	}

when the `JniPeerMembers` constructor runs, it will call
`JniEnvironment.Runtime.TypeManager.GetReplacementType("android/app/Activity")`.
If `@(_AndroidRemapMembers)` is based on the InTune
`remapping-config.json` file, then `android/app/Activity` is mapped
to `com/microsoft/intune/mam/client/app/MAMActivity`, and
`JNIEnv::FindClass()` will be told to lookup `MAMActivity`, *not*
`Activity`.

*If `MAMActivity` can't be found*, e.g. you're testing this all out,
the app will ~immediately crash, as `MAMActivity` doesn't exist. 😅

If `MAMActivity` can be found, eventually `Activity.OnCreate()` will
need to be invoked:

	partial class Activity {
	    {
	        try {
	        } finally {
	        }
	    }
	}

`_members.InstanceMethods.InvokeVirtualVoidMethod()` will internally
make a call similar to:

	var r = JniEnvironment.Runtime.TypeManager.GetReplacementMethodInfo (
	    "com/microsoft/intune/mam/client/app/MAMActivity",
	    "onCreate",
	    "(Landroid/os/Bundle;)V"
	);

The data returned will be equivalent to:

	var r = new JniRuntime.ReplacementMethodInfo {
	    SourceJniType                   = "com/microsoft/intune/mam/client/app/MAMActivity",    // from input parameter
	    SourceJniMethodName             = "onCreate",                                           // from input parameter
	    SourceJniMethodSignature        = "(Landroid/os/Bundle;)V",                             // from input parameter

	    TargetJniType                   = "com/microsoft/intune/mam/client/app/MAMActivity",    // from //replace-method/@target-type
	    TargetJniMethodName             = "onMAMCreate",                                        // from //replace-method/@target-method-name
	    TargetJniMethodSignature        = "(Landroid/os/Bundle;)V",                             // from input parameter, as signature didn't change
	    TargetJniMethodParameterCount   = 1,                                                    // computed based on signature
	    TargetJniMethodIsStatic         = false,                                                // from //replace-method/@target-method-instance-to-static
	}

This will allow `_members.InstanceMethods.InvokeVirtualVoidMethod()`
to instead resolve and invoke `MAMActivity.onMAMCreate()`.

~~ Tools ~~

`tools/remap-mam-json-to-xml` is added, and will process the InTune
JSON file into `@(_AndroidRemapMembers)` XML:

	$ dotnet run --project tools/remap-mam-json-to-xml  -- \
	  $HOME/.nuget/packages/microsoft.intune.mam.remapper.tasks/0.1.4635.1/content/MonoAndroid10/remapping-config.json
	<replacements>…


~~ Unit Tests ~~

`@(_AndroidRemapMembers)` usage is required by
`Mono.Android.NET-Tests.apk`, as `Java.Interop-Tests.dll` exercises
the type and member remapping logic.


~~ Unrelated `gref+` logging fixes ~~

When `debug.mono.log` = `gref+`, the app could crash:

	signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0x79c045dead

This was likely because a constant string was provided to
`OSBridge::_write_stack_trace()`, which tried to write into the
constant string, promptly blowing things up.

Workaround: don't use `gref+` logging when a GC occurs?
(Fortunately, `gref+` logging isn't the default.)

Better workaround: Don't Do That™. Don't write to const strings.

~~ About `@(_AndroidRemapMembers)` Semantics… ~~

 1. Changing the Java hierarchy "requires" changing the managed
    hierarchy to mirror it.

    If we rename `Activity` to `RemapActivity` but *don't* change
    `MainActivity` to inherit the (bound!) `Example.RemapActivity`,
    the app *crashes*:

        JNI DETECTED ERROR IN APPLICATION: can't call void example.RemapActivity.onMyCreate(android.os.Bundle) on instance of example.MainActivity

    This can be "fixed" *without* changing the base class
    of `MainActivity` by instead changing the base class of the
    Java Callable Wrapper for `MainActivity` to `example.RemapActivity`.
    This can be done manually (just edit the files in `obj/…`!), but
    isn't really supported in "normal" xamarin-android usage
    (the next Clean will wipe your changes).

    Presumably InTune would make this Just Work by e.g. patching the
    `MainActivity.class` file.

 2. `/replacements/replace-type` interacts with
    `/replacements/replace-method`: at runtime, `//replace-type@from`
    *no longer exists*, meaning you ***cannot*** use that name
    in `//replace-method/@source-type` either!

    If `Activity` is remapped to `RemapActivity`, then *there is no*
    `Activity.onCreate()` method to similarly remap.  Instead, you
    need to specify `RemapActivity.onCreate()`.

    This warps the brain a bit.

    This:

        <replace-method
            source-type="example/RemapActivity"
            source-method-name="onCreate"
            target-type="example/RemapActivity"
            target-method-name="onMyCreate" target-method-instance-to-static="false" />

    not this:

        <replace-method
            source-type="android/app/Activity"
            source-method-name="onCreate"
            target-type="example/RemapActivity"
            target-method-name="onMyCreate" target-method-instance-to-static="false" />

 3. Don't intermix type renames with
    `/replace-method/@target-method-instance-to-static='true']`.
    It *can* be done, but also warps the brain.

    The deal with `@target-method-instance-to-static` is that it
    it changes the target method signature -- unless explicitly
    provided in `/replace-method/@target-method-signature` --
    so that the "source declaring type" is a prefix.

    Thus given

        <replace-method
            source-type="android/view/View"
            source-method-name="setOnClickListener"
            target-type="example/ViewHelper"
            target-method-name="mySetOnClickListener" target-method-instance-to-static="true" />

    we'll look for `ViewHelper.mySetOnClickListener(View, View.OnClickListener)`.

    If we renamed `View` to `MyView`, we would instead look for
    `ViewHelper.mySetOnClickListener(MyView, View.OnClickListener)`
    (note changed parameter type).

    This almost certainly *won't* work.


~~ InTune Integration Testing? ~~

For "more complete" InTune integration testing, one will want the
path to `remapping-config.json`, without hardcoding things.
This can be done with `%(PackageReference.GeneratePathProperty)`=True
and using `$(PkgMicrosoft_Intune_MAM_Remapper_Tasks)`:

	<ItemGroup>
	  <PackageReference
	      Include="Microsoft.Intune.MAM.Remapper.Tasks"
	      Version="0.1.4635.1"
	      IncludeAssets="none"
	      GeneratePathProperty="True"
	      ReferenceOutputAssembly="False"
	  />
	</ItemGroup>
	<Target Name="_AddMamFiles"
	    BeforeTargets="_AddAndroidCustomMetaData">
	  <ItemGroup>
	    <_AndroidMamMappingFile Include="$(PkgMicrosoft_Intune_MAM_Remapper_Tasks)/content/MonoAndroid10/remapping-config.json" />
	  </ItemGroup>
	</Target>

This is still fraught with some peril, as it likely also depends on
getting the right "inner" build, which may require using the plural
`$(TargetFrameworks)` property, not the singular `$(TargetFramework)`.
This might still be a useful start.

~~ TODO ~~

Optimize this mess:
https://github.com/xamarin/xamarin-android/issues/7020

[0]: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html
[1]: https://developer.android.com/studio/write/java8-support#library-desugaring
[2]: https://developer.android.com/reference/java/util/Arrays#stream(T[])
[3]: https://docs.microsoft.com/en-us/mem/intune/fundamentals/what-is-intune
[4]: https://www.nuget.org/packages/Microsoft.Intune.MAM.Remapper.Tasks/
```

PR #6594, opened 2021-12-22
[build] .yaml changes for private fork & NuGet feeds

jonathanpeppers:
This brings over more changes from `xamarin-android-private`.

With these changes, the private fork should be able to build `main` and push builds to an internal feed.

I also set `$(TeamName)`, which isn't used here -- but will be needed when consuming internal .NET builds on the private fork.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -46,20 +46,20 @@ variables:
   value: '& TestCategory != DotNetIgnore & TestCategory != HybridAOT & TestCategory != ProfiledAOT & TestCategory != MkBundle & TestCategory != MonoSymbolicate & TestCategory != PackagesConfig & TestCategory != StaticProject & TestCategory != Debugger & TestCategory != SystemApplication'
 - name: TeamName
   value: XamarinAndroid
-- ${{ if eq(variables['Build.DefinitionName'], 'Xamarin.Android Private') }}:
+- ${{ if eq(variables['Build.DefinitionName'], 'Xamarin.Android-Private') }}:
   - group: AzureDevOps-Artifact-Feeds-Pats
   - group: DotNet-MSRC-Storage
   - name: DotNetFeedCredential
     value: dotnet6-internal-dnceng-internal-feed
-- ${{ if ne(variables['Build.DefinitionName'], 'Xamarin.Android Private') }}:
+- ${{ if ne(variables['Build.DefinitionName'], 'Xamarin.Android-Private') }}:
   - name: DotNetFeedCredential
     value: dnceng-dotnet6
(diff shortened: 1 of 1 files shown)


jonathanpeppers:
Testing:
* [Public build](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5568984&view=results)
* [Private build](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5568988&view=results)

pjcollins:
I think we have a bug in this script for build pipelines which contain spaces in the name - https://github.com/xamarin/yaml-templates/blob/38e978a718e7f1d514cb6763ce30a3b888af3a69/upload-to-storage.yml#L46. The easiest fix would be to rename the pipeline to `Xamarin.Android-Private` (and update pipeline name checks in yaml accordingly).

jonathanpeppers:
Renamed to `Xamarin.Android-Private` new private build [here](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5569521&view=results)

jonathanpeppers:
I renamed it, now need to request new signing approval...

pjcollins:
Oof, I guess for some reason that approval list is keyed off actual name instead of the pipeline ID 😢

jonathanpeppers:
This is still waiting signing approval, maybe we'll get that soon in the new year.
```
##[error]Xamarin.Android-Private is not approved for signing in https://devdiv.visualstudio.com/. Please read https://devdiv.visualstudio.com/DevDiv/_wiki/wikis/DevDiv.wiki/650/MicroBuild-Signing
```

jonathanpeppers:
It seems like it might have gotten approved and it didn't notify me?

Running now: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5627305&view=results

jonathanpeppers:
The build on the private side appears to be working well enough to sign packages and run tests. 👍

build-tools/automation/azure-pipelines.yaml
```diff
@@ -44,12 +44,22 @@ variables:
   value: $[or(eq(variables['XA.RunAllTests'], true), eq(variables['IsMonoBranch'], true))]
 - name: DotNetNUnitCategories
   value: '& TestCategory != DotNetIgnore & TestCategory != HybridAOT & TestCategory != ProfiledAOT & TestCategory != MkBundle & TestCategory != MonoSymbolicate & TestCategory != PackagesConfig & TestCategory != StaticProject & TestCategory != Debugger & TestCategory != SystemApplication'
-- ${{ if and(eq(variables['Build.DefinitionName'], 'Xamarin.Android'), ne(variables['Build.Reason'], 'PullRequest')) }}:
+- name: TeamName
+  value: XamarinAndroid
```
pjcollins, line 48: I think we can cut this, it was moved to https://github.com/xamarin/xamarin-android/blob/main/build-tools/automation/yaml-templates/variables.yaml#L34

pjcollins (approved):
Other than the minor duplication above, LGTM

PR #6601, opened 2022-01-03
[Mono.Android] Bind Android 12L Beta 1.

jpobst:
Context: https://developer.android.com/about/versions/12/12L
Context: https://android-developers.googleblog.com/2021/10/12L-preview-large-screens.html

Android 12L Developer Beta 1 has been released.  The Android 12L

  * October: Developer Preview 1
  * December: Beta 1 (Final APIs)
  * January: Beta 2
  * February: Beta 3
  * Q1 2022: Final

[0]: https://developer.android.com/about/versions/12/12L/overview

Enumify API-32, set API-Sv2 as API-32, v12.1, and stable.  Bump .NET 6 pack version to `32.0.100`.


jonathanpeppers (commented):
Any MSBuild test now fails with the warning:
```
Xamarin.Android.Aapt2.targets(157,3): warning APT2000: aapt2 W 01-06 00:34:15 13393 106914 LoadedArsc.cpp:657] Unknown chunk type '200'.
```
And so every test checking there are no warnings fails.

I'll look into bumping build-tools that we use during our build, that looks to be where we get `aapt2` from. Hopefully that solves it?

jonathanpeppers:
Some of the tests failed with:
```
error NU1101: Unable to find package Microsoft.Android.Ref.31. No packages exist with this id in source(s): dotnet-eng, dotnet-public, xamarin.android util
```
I got these packages added to the `dotnet-public` feed:

* https://www.nuget.org/packages/Microsoft.Android.Ref.31/31.0.101-preview.11.117
* https://www.nuget.org/packages/Microsoft.Android.Runtime.31.android-x86/31.0.101-preview.11.117
* https://www.nuget.org/packages/Microsoft.Android.Runtime.31.android-x64/31.0.101-preview.11.117
* https://www.nuget.org/packages/Microsoft.Android.Runtime.31.android-arm/31.0.101-preview.11.117
* https://www.nuget.org/packages/Microsoft.Android.Runtime.31.android-arm64/31.0.101-preview.11.117

We'll see if those tests pass on a rerun now.

jonathanpeppers:
As for this warning:

```
D8 : warning : An API level of 31 is not supported by this compiler. Please use an API level of 30 or earlier 
```
Hoping https://github.com/xamarin/xamarin-android/pull/6620 might fix?

jonathanpeppers:
This is now looking "greener".

So the issue now is `build-tools` 32.0.0 no longer ships with `dx.jar`:
```
  COMPILETODALVIK : error : Unable to access jarfile C:\Users\cloudtest\android-toolchain\sdk\build-tools\32.0.0\\lib\dx.jar [C:\a\_work\1\a\TestRelease\01-07_19.50.01\temp\CheckClassesDexIsIncludeddx\UnnamedProject.csproj]
```
Unsure what to do here if we want to support `dx`...

jonathanpeppers:
Oh, this might not actually be green, there is a merge conflict, lol...

build-tools/xaprepare/xaprepare/Steps/Step_Get_Android_BuildTools.cs
```diff
@@ -21,6 +21,16 @@ public Step_Get_Android_BuildTools ()
...
 			packages.Add ((package: $"build-tools_r{XABuildToolsVersion}-windows.zip", prefix: XABuildToolsPackagePrefixWindows));
 			packages.Add ((package: $"build-tools_r{XABuildToolsVersion}-linux.zip", prefix: XABuildToolsPackagePrefixLinux));
+
+			// build-tools 30, for DX tests
+			string XABuildTools30Version     = Context.Instance.Properties [KnownProperties.XABuildTools30Version] ?? String.Empty;
+			string XABuildTools30PackagePrefixMacOS = Context.Instance.Properties [KnownProperties.XABuildTools30PackagePrefixMacOS] ?? string.Empty;
+			string XABuildTools30PackagePrefixWindows = Context.Instance.Properties [KnownProperties.XABuildTools30PackagePrefixWindows] ?? string.Empty;
+			string XABuildTools30PackagePrefixLinux = Context.Instance.Properties [KnownProperties.XABuildTools30PackagePrefixLinux] ?? string.Empty;
+
+			packages.Add ((package: $"build-tools_r{XABuildTools30Version}-macosx.zip", prefix: XABuildTools30PackagePrefixMacOS));
```
grendello, line 31: Shouldn't this block be skipped if `XABuildTools30Version` or any prefix it uses are `String.Empty`? If the properties are optional, then there's no need to introduce potential errors in the build later on.

jonathanpeppers: I'll add a `string.IsNullOrEmpty()` check. I was thinking we'd just delete all this code when we stop supporting DX, but that might get missed or something.

jonathanpeppers:
Ok this is now only failure 1 set of tests:
```
platforms/android-32 should be a dependency.
```
I think I need to make them check for 31 under .NET 6, and then hopefully this will be green.

build-tools/xaprepare/xaprepare/Steps/Step_Get_Android_BuildTools.cs
grendello: It'll be definitely safer this way

grendello (approved):

dellis1972 (approved):

tests/Mono.Android-Tests/Mono.Android-Tests.csproj
```diff
@@ -21,7 +21,7 @@
     <MandroidI18n>All</MandroidI18n>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\product.snk</AssemblyOriginatorKeyFile>
-    <TargetFrameworkVersion>v12.0</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v12.1</TargetFrameworkVersion>
```
pjcollins, line 24: Should this (and the two projects above) be using `$(AndroidLatestStableFrameworkVersion)` instead?

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
```diff
@@ -47,23 +47,23 @@
...
     },
     "Microsoft.Android.Runtime.31.android-x86": {
       "kind": "framework",
-      "version": "@WORKLOAD_VERSION@"
+      "version": "31.0.101-preview.11.117"
     },
     "Microsoft.Android.Runtime.31.android-x64": {
       "kind": "framework",
-      "version": "@WORKLOAD_VERSION@"
+      "version": "31.0.101-preview.11.117"
```
pjcollins, line 66: It seems the choice to keep API 31 as the .NET default is introducing some new complexity... It looks like we're locking the API 31 ref/runtime pack versions to the latest version available on nuget.org. I think this makes sense as we're now only building the net6.0 version of Mono.Android.dll for API 32.

On the .NET MSBuild testing side we may have a new gap in coverage though. I think the default test project will continue to target `net6.0-android31`, which will be running against ref/runtime packs that have already been published and not the ones produced by the build. The positive side of this is that we will ensure that any changes to the SDK continue to work with our previously produced API 31 packs. A potential downside is that we may now miss an issue caused by a change in Mono.Android.csproj. I'm not sure what the best solution for this is at the moment and I don't think it necessarily needs to be addressed in this PR, but it's worth further discussion.

jonathanpeppers: Yes, if we end up with 32 as the default in .NET 6, we wouldn't have to put the 31 packs in the workload.

This logic here, allows them to come from NuGet *or* the workload:

https://github.com/xamarin/xamarin-android/blob/5f66c36fc10ae2f159885fafe15b16b871378bd1/build-tools/create-packs/Microsoft.Android.Sdk.proj#L122-L142

I think we should wait and see what we end up doing.

build-tools/create-packs/Microsoft.Android.Sdk.proj
```diff
@@ -118,18 +118,23 @@ WARNING:  DO NOT MODIFY this file unless you are knowledgeable about MSBuild and
   <PropertyGroup>
     <AndroidNETSdkVersion>$(AndroidPackVersionLong)</AndroidNETSdkVersion>
     <XamarinAndroidVersion>$(AndroidPackVersionLong)</XamarinAndroidVersion>
-    <_AndroidRuntimePackId Condition=" '%24(TargetPlatformVersion)' != '$(AndroidLatestUnstableApiLevel).0' ">$(AndroidLatestStableApiLevel)</_AndroidRuntimePackId>
-    <_AndroidRuntimePackId Condition=" '%24(TargetPlatformVersion)' == '$(AndroidLatestUnstableApiLevel).0' ">$(AndroidLatestUnstableApiLevel)</_AndroidRuntimePackId>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '%24(TargetPlatformVersion)' == '31.0' ">
```
jonpryor, line 122: Should this be using `$(AndroidTargetDotNetApiLevel)` (and possibly other new properties)? Isn't that what `$(AndroidTargetDotNetApiLevel)` is for?

Related: instead of `$(AndroidTargetDotNetApiLevel)`, should we call it `$(AndroidDefaultTargetDotnetApiLevel)`?  This "emphasizes" the "default" nature of this value.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs
```diff
@@ -95,7 +95,7 @@ public void GetDependencyWhenBuildToolsAreMissingTest ()
 				builder.Target = "GetAndroidDependencies";
 				Assert.True (builder.Build (proj, parameters: parameters),
 					string.Format ("First Build should have succeeded"));
-				int apiLevel = Builder.UseDotNet ? AndroidSdkResolver.GetMaxInstalledPlatform () : 26;
+				int apiLevel = Builder.UseDotNet ? 31 : 26;
```
jonpryor, line 98: Should this "31" value come from "elsewhere"?  Perhaps using the `$(AndroidTargetDotNetApiLevel)` / `$(AndroidDefaultTargetDotnetApiLevel)` property?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
```diff
@@ -614,6 +614,7 @@ public void DotNetBuild (string runtimeIdentifiers, bool isRelease, bool aot, bo
 			XNamespace ns = "http://schemas.android.com/apk/res/android";
 			var uses_sdk = manifest.Root.Element ("uses-sdk");
 			Assert.AreEqual ("21", uses_sdk.Attribute (ns + "minSdkVersion").Value);
+			// TODO: Change this when 32 becomes the default for .NET 6+
```
jonpryor, line 617: We're *totally* going to forget about this.

build-tools/create-packs/Microsoft.Android.Sdk.proj
jonathanpeppers: In some of the other tests here, I'll try to use `$(AndroidDefaultTargetDotnetApiLevel)` so we don't have to change them again later.

In this case, however, the number 31 needs to remain, it will need to eternally pull this version from NuGet (or our workload):

https://github.com/xamarin/xamarin-android/blob/bedecc25f93a82ecaa8df5cb4e92424aa669125d/build-tools/create-packs/Microsoft.Android.Sdk.proj#L122-L125

I guess that over time, we'll have numbers listed for 32, 33, and so on. We can also update those packs at some point if ever needed (hopefully not ever?), we'd just list a different version number here.

jonpryor:
Proposed commit message:

```markdown
[Mono.Android] API-32 stabilization

Context: https://developer.android.com/about/versions/12/12L
Context: https://android-developers.googleblog.com/2021/12/beta-1-update-for-12l-feature-drop.html

Android 12L Developer Beta 1 has been released.

  * [API diff vs. API-31][0]
  * [API diff vs. Sv2 DP1][1]

The Android 12L Developer Preview Program Overview
[Timeline and updates][2] section suggests the following timeline
(which differs from ef011213):

  * October:    Developer Preview 1
  * December:   Beta 1; Final APIs (previously expected in Beta 2)
  * January:    Beta 2; Incremental beta update
  * February:   Beta 3; Incremental beta update
  * Q1 2022:    Final release

In particular, Beta 1 has "Final APIs".  We now have API-32!

Enumify API-32, and declare it stable.

As a change from "upstream Google", Classic Xamarin.Android uses
`$(TargetFrameworkVersion)`=v12.1 for API-32, as there is no way to
express "12L" in a `System.Version`.  We have decreed 12.1 as 12L.

.NET SDK for Android uses `$(TargetFramework)`=net6.0-android32.0.

However, we don't want to change the default API level for .NET 6
projects; the default will remain `net6.0-android31.0` (API-31),
using the [31.0.101-preview.11.117 binaries on NuGet.org][3].
This requires various changes to the unit test infrastructure and
.NET SDK workflow package creation.

Additionally, `d8` doesn't appear to support API-32 yet (?!);
building an app with an `AndroidManifest.xml` containing:

	<uses-sdk android:minSdkVersion="32" android:targetSdkVersion="32" />

would use `d8 --min-api 32`, which results in a warning:

	D8 : warning : An API level of 32 is not supported by this compiler.
	Please use an API level of 31 or earlier

Update the `XASdkTests.SupportedOSPlatformVersion()` test so that it
only uses API-31 and not API-32.  This avoids the above warning.

Finally, we had to update xaprepare to install the new
Android SDK Build-tools version 32.0.0, as `aapt` from previous
Build-tools versions would emit a warning when building for API-32:

	Xamarin.Android.Aapt2.targets(157,3): warning APT2000: aapt2 W 01-05 23:58:06  8352 43857 LoadedArsc.cpp:657] Unknown chunk type '200'.

Unfortunately, installing Build-tools 32.0.0 promptly broke unit
tests using `$(AndroidDexTool)`=dx, with:

	COMPILETODALVIK : error : Unable to access jarfile C:\Users\jopepper\android-toolchain\sdk\build-tools\32.0.0\\lib\dx.jar

as Build-tools 31 and later no longer contains `lib/dx.jar`.  Update
xaprepare to install *both* Build-tools 30.0.3 *and* 32.0.0, so that
tests which require `dx.jar` continue to run, and we can get *some*
test coverage against Build-tools 32.

[0]: https://developer.android.com/sdk/api_diff/32/changes
[1]: https://developer.android.com/sdk/api_diff/32-incr/changes
[2]: https://web.archive.org/web/20220113142518/https://developer.android.com/about/versions/12/12L/overview
[3]: https://www.nuget.org/packages/Microsoft.Android.Ref.31/31.0.101-preview.11.117
```

jonathanpeppers:
@jonpryor that old commit was true at the time, but should now be:

> But with `net6.0-android` (defaults to 31 now), you get:
> 
>     dotnet\packs\Microsoft.Android.Ref.31\*\ref\net6.0\Mono.Android.dll

Issue #6472, opened 2021-11-09
xamarin-android linux artifacts


rokups:
Now that .NET6 is released, maybe it would be a good time to release nightly artifacts of xamarin-android workload for linux, as discussed in https://github.com/dotnet/maui-samples/issues/61?

Thanks

fogzot:
Yes, please. I understand why MS does not want to support *running* MAUI on Linux but making it impossible to *develop* on it is clearly sending the message that Linux is a second class citizen in the dotnet exosystem.

redradist:
@jonpryor Does it mean that now I can develop Android application on Ubuntu without any hack ?
If yes, then is there some example ?

Cheesebaron:
Have you tried: `dotnet workload install android`?

rokups:
Something does not quite work. Looks like some package can not be found.

<details>

```
% sudo dotnet workload install android
[sudo] password for rk: 

Welcome to .NET 6.0!
---------------------
SDK Version: 6.0.102

----------------
Installed an ASP.NET Core HTTPS development certificate.
To trust the certificate run 'dotnet dev-certs https --trust' (Windows and macOS only).
Learn about HTTPS: https://aka.ms/dotnet-https
----------------
Write your first app: https://aka.ms/dotnet-hello-world
Find out what's new: https://aka.ms/dotnet-whats-new
Explore documentation: https://aka.ms/dotnet-docs
Report issues and find source on GitHub: https://github.com/dotnet/core
Use 'dotnet --help' to see available commands or visit: https://aka.ms/dotnet-cli
--------------------------------------------------------------------------------------

Skip NuGet package signing validation. NuGet signing validation is not available on Linux or macOS https://aka.ms/workloadskippackagevalidation .
Updated advertising manifest microsoft.net.sdk.maui.
Updated advertising manifest microsoft.net.sdk.ios.
Updated advertising manifest microsoft.net.sdk.maccatalyst.
Updated advertising manifest microsoft.net.workload.emscripten.
Updated advertising manifest microsoft.net.sdk.macos.
Updated advertising manifest microsoft.net.sdk.android.
Updated advertising manifest microsoft.net.sdk.tvos.
Updated advertising manifest microsoft.net.workload.mono.toolchain.
Installing workload manifest microsoft.net.sdk.maccatalyst version 15.0.101-preview.11.551…
Installing workload manifest microsoft.net.sdk.ios version 15.0.101-preview.11.551…
[...]

PR #6613, opened 2022-01-06
[ci] Sign and publish Microsoft.Android.Sdk.Linux

pjcollins:
Fixes: https://github.com/xamarin/xamarin-android/issues/6472

A `Sign Linux Archive` job has been added to the `Prepare .NET Release`
stage to conditionally Authenticode sign the Microsoft.Android.Sdk.Linux
package produced by the Linux build.  The "Real" signing condition that
was previously set on each job in this stage has been moved up to the
stage level for simplicity.

The signed version of Microsoft.Android.Sdk.Linux will now be pushed to
the dotnet6 feed and to Maestro on all official builds.  It will also be
pushed to NuGet.org when that manually triggered stage runs.


pjcollins:
Skipped a handful of steps here, but this looks like it should work as expected: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5597148&view=results

pjcollins:
There was an issue in .msi generation due to the branch name but the stage/job additions and dependency changes are behaving as expected: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5604581&view=results

build-tools/automation/azure-pipelines.yaml
```diff
@@ -1323,35 +1325,50 @@ stages:
...
-      condition: eq(variables['MicroBuildSignType'], 'Real')
 
- # Check - "Xamarin.Android (.NET 6 Prepare Release Push NuGets)"
+  # Check - "Xamarin.Android (Prepare .NET Release Push Internal)"
   - job: push_signed_nugets
-    displayName: Push NuGets
-    dependsOn: nuget_convert
-    condition: and(eq(dependencies.nuget_convert.result, 'Succeeded'), eq(variables['MicroBuildSignType'], 'Real'))
+    displayName: Push Internal
+    dependsOn:
```
jonpryor, line 1368: I'm not *quite* sure how to read the [docs](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema), but shouldn't the value to `dependsOn` be "things that *exist*"? Though I'm not sure how to check "existence", and I don't know where the previously depended on value `nuget_convert` comes from either, so I think I don't understand what this *does*?

Or it doesn't care about "namespaces"?  There does appear to be a `nuget_convert` job: https://github.com/xamarin/yaml-templates/blob/11019011d31d6f5030cddefba088783b4d0ef397/nuget-msi-convert/job/v1.yml#L65

I'm less certain about what the `sign_net_linux` job is, unless template parameter names can provide job names, i.e. `sign_net_linux` is from the `sign-artifacts/jobs/v2.yml@yaml` template invocation above?  (Which would be both awesome and horrifying?)

build-tools/create-packs/linux-README.md
```diff
@@ -0,0 +1 @@
+Please note that this package is not officially supported, and has not received extensive testing.
```
jonathanpeppers, line 1: Should we put this on the package description, too?
https://github.com/xamarin/xamarin-android/blob/df667b079fdccb9a263d5c12467eef7f2fa0ff55/build-tools/create-packs/Microsoft.Android.Sdk.proj#L15

jonathanpeppers (approved):
Otherwise this looks good. When I look at the different phases on the builds, things look like they make sense.

build-tools/automation/azure-pipelines.yaml
pjcollins: If the `dependsOn` parameter is used on a job, the value should contain one or more job names (not display names) that exist in the same stage of the pipeline. If it is used on a stage, the value should contain one or more stage names. 

The `push_signed_nugets` job has been updated to now depend on both the existing `nuget_convert` and new `sign_net_linux` jobs. As you deduced, the [`nuget_convert job`](https://github.com/xamarin/yaml-templates/blob/11019011d31d6f5030cddefba088783b4d0ef397/nuget-msi-convert/job/v1.yml#L65) is defined in yaml-templates.

YAML templates are quite flexible, and a job name can be specified as a parameter to a job that is declared in a template. This is the case for the shared [`sign-artifacts` job](https://github.com/xamarin/yaml-templates/blob/11019011d31d6f5030cddefba088783b4d0ef397/sign-artifacts/jobs/v2.yml#L33). Since we are now using this signing job template twice in the same stage, I have set the `name` parameter on both instances to ensure that we are using unique job names.

build-tools/create-packs/linux-README.md
pjcollins: I added this to the description as well.

Issue #5378, opened 2020-12-07
Debug symbols upload on Google Play Console for better crash information


EmilAlipiev:
When we upload an app bundle or apk, below warning is returned from google play console. Question is what kind of source symbols should we or can we upload here? I am guessing that it is it not pdb files :) Is there a way for Xamarin.forms applications for android to get symbols and better crash information?

`This App Bundle contains native code, and you've not uploaded debug symbols. We recommend that you upload a symbol file to make your crashes and ANRs easier to analyse and debug.`
 
![image](https://user-images.githubusercontent.com/10422347/101300920-7f98a900-3837-11eb-87c9-e7cce1c89fb5.png)

jonathanpeppers:
This was solved for obfuscated Java/Kotlin code here:

https://github.com/xamarin/xamarin-android/pull/5304

It doesn't look like we have solved this for native code, which would include AOT'd .NET assemblies.

tbrentlong:
I'm seeing the same issue when uploading a bundle build via VS2019.  Is there a file somewhere that can be uploaded?  I also still get the deobfuscation message noted in 5186 even though I'm not using a code shrinker.

jonathanpeppers:
@tbrentlong as mentioned in #5304, I think you would need to use `AndroidLinkMode=r8` to get the `mapping.txt` file.

Can you share a screenshot of the warning you get?

Eversor:
I'm getting the same issue here, 
![image](https://user-images.githubusercontent.com/733699/115208154-8fd16680-a0fc-11eb-958b-3246480516fa.png)
![image](https://user-images.githubusercontent.com/733699/115208403-d030e480-a0fc-11eb-8e07-31476f0ad078.png)

Microsoft Visual Studio Community 2019
Version 16.7.6
VisualStudio.16.Release/16.7.6+30611.23
Microsoft .NET Framework
Version 4.8.04084

Xamarin   16.7.000.456 (d16-7-6@f8d55b6)
Visual Studio extension to enable development for Xamarin.iOS and Xamarin.Android.

Xamarin.Android SDK   11.0.2.0 (d16-7/025fde9)
Xamarin.Android Reference Assemblies and MSBuild support.
    Mono: 83105ba
    Java.Interop: xamarin/java.interop/d16-7@1f3388a
    ProGuard: Guardsquare/proguard/proguard6.2.2@ebe9000
    SQLite: xamarin/sqlite/3.32.1@1a3276b
[...]

PR #6616, opened 2022-01-06
[Xamarin.Android.Build.Tasks] Add Support for bundletool meta-data

dellis1972:
Fixes: #5378

Context: 2dab2bc

When uploading an `.apk` or `.aab` file which contains native
libraries such as AOT'd .NET assemblies to the Google Play store,
the Google Play Console reports the following warning:

> This App Bundle contains native code, and you've not uploaded
> debug symbols.  We recommend that you upload a symbol file to make
> your crashes and ANRs easier to analyse and debug.

Commit 2dab2bc addresses a similar warning regarding obfuscated
Java/Kotlin code.

Improve support for this scenario by adding a new
`$(AndroidCreateProguardMappingFile)` MSBuild property which controls
the creation of the ProGuard mapping file.

When True (the default when r8 is used), a ProGuard mapping file will
be created and added to `.apk`/`.aab` build outputs.  This increases
the size of resulting applications by ~1% for our unit tests, and
will remove the above warning for Google Play Store submissions.

To make `$(AndroidCreateProguardMappingFile)`=True *work*, add
support for a new `@(AndroidAppBundleMetaDataFile)` item group.
`%(AndroidAppBundleMetaDataFile.Include)` values are those used by
the [`bundletool build-bundle --metadata-file`][0] option:

> `--metadata-file=target-bundle-path:local-file-path`
>
>   * `target-bundle-path` specifies a path relative to the root of
>      the app bundle where you would like the metadata file to be
>      packaged, and
>   * `local-file-path` specifies the path to the local metadata file
>     itself.

When `$(AndroidCreateProguardMappingFile)`=True, we automatically add:

	<AndroidAppBundleMetaDataFile Include="com.android.tools.build.obfuscation/proguard.map:$(AndroidProguardMappingFile" />

which causes the ProGuard mapping file to be included into the bundle.
This only applies to `.aab` package formats.

For `.apk` files the ProGuard mapping file will not be included in the
`.apk`,  as it doesn't not support it. However the file will be copied
to the  `$(OutputPath)` along with the `.apk`. It can be manually uploaded
to the Google Play Store from there.

[0]: https://developer.android.com/studio/build/building-cmdline#bundletool-build

### DIFF
--- a/Documentation/guides/building-apps/build-items.md
+++ b/Documentation/guides/building-apps/build-items.md
@@ -38,6 +38,26 @@ Used to provide an AOT profile, for use with profile-guided AOT.
 It can be also used from Visual Studio by setting the `AndroidAotProfile`
 build action to a file containing an AOT profile.
 
+## AndroidAppBundleMetaDataFile
+
+Specifies a file that will be included as metadata in the Android App Bundle.
+The format of the flag value is `<bundle-path>:<physical-file>` where
+`bundle-path` denotes the file location inside the App Bundle's metadata
+directory, and `physical-file` is an existing file containing the raw data
+to be stored.
+
+```xml
+<ItemGroup>
+  <AndroidAppBundleMetaDataFile
+    Include="com.android.tools.build.obfuscation/proguard.map:$(OutputPath)mapping.txt"
+  />
+</ItemGroup>
+```
+
+See [bundletool](https://developer.android.com/studio/build/building-cmdline#build_your_app_bundle_using_bundletool) documentation for more details.
+
+Added in Xamarin.Android 12.3.
+
 ## AndroidBoundLayout
 
 Indicates that the layout file is to have code-behind generated for it in case when
--- a/Documentation/guides/building-apps/build-properties.md
+++ b/Documentation/guides/building-apps/build-properties.md
@@ -300,6 +300,28 @@ should be created instead of having support for all ABIs in a single `.apk`.
 See also the [Building ABI-Specific APKs](~/android/deploy-test/building-apps/abi-specific-apks.md)
 guide.
 
+## AndroidCreateProguardMappingFile
+
+A boolean property that controls if a proguard mapping file is
+generated as part of the build process.
+
+Adding the following to your csproj will cause the file to be
+generated. This uses the [`AndroidProguardMappingFile`](#androidproguardmappingfile) property
+to control the location of the final mapping file.
+
+```
+<AndroidCreateProguardMappingFile>True</AndroidCreateProguardMappingFile>
+```
+
+Note that if you are using `.aab` files the mapping file with be
+automatically included in your package. So there is no need to upload
+it to the Google Play Store manually. If you are using `.apk` files
+you will need to manually upload the [`AndroidProguardMappingFile`](#androidproguardmappingfile).
+
+The default value is `True` when using `AndroidLinkTool=r8`.
+
+Added in Xamarin.Android 12.3.
+
 ## AndroidDebugKeyAlgorithm
 
 Specifies the default
@@ -987,7 +1009,18 @@ mean the `mapping.txt` file will be produced in the `$(OutputPath)`
 folder. This file can then be used when uploading packages to the
 Google Play Store.
 
-The default value is `$(OutputPath)mapping.txt`.
+By default this file is produced automatically when using `AndroidLinkTool=r8`
+and will generate the following file `$(OutputPath)mapping.txt`.
+
+If you do not want to generate this mapping file you can use the
+[`AndroidCreateProguardMappingFile`](#androidcreateproguardmappingfile) property to stop creating it .
+Add the following in your project
+
+```
+<AndroidCreateProguardMappingFile>False</AndroidCreateProguardMappingFile>
+```
+
+or use `-p:AndroidCreateProguardMappingFile=False` on the command line.
 
 This property was added in Xamarin.Android 11.2.
 
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
@@ -12,7 +12,7 @@ namespace Xamarin.Android.Tasks
 {
 	/// <summary>
 	/// Invokes `bundletool` to create an Android App Bundle (.aab file)
-	/// 
+	///
 	/// Usage: bundletool build-bundle --modules=base.zip --output=foo.aab --config=BundleConfig.json
 	/// </summary>
 	public class BuildAppBundle : BundleTool
@@ -63,8 +63,10 @@ namespace Xamarin.Android.Tasks
 		public string BaseZip { get; set; }
 
 		public string CustomBuildConfigFile { get; set; }
-		
+
 		public string [] Modules { get; set; }
+
+		public ITaskItem [] MetaDataFiles { get; set; }
 
 		[Required]
 		public string Output { get; set; }
@@ -130,6 +132,9 @@ namespace Xamarin.Android.Tasks
 			cmd.AppendSwitchIfNotNull ("--modules ", string.Join (",", modules));
 			cmd.AppendSwitchIfNotNull ("--output ", Output);
 			cmd.AppendSwitchIfNotNull ("--config ", temp);
+			foreach (var file in MetaDataFiles ?? Array.Empty<ITaskItem> ()) {
+				cmd.AppendSwitch ($"--metadata-file={file.ItemSpec}");
+			}
 			return cmd;
 		}
 	}
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -264,6 +264,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidEnableDesugar  Condition=" '$(AndroidEnableDesugar)' == '' And ('$(AndroidDexTool)' == 'd8' Or '$(AndroidLinkTool)' == 'r8') ">True</AndroidEnableDesugar>
 	<AndroidEnableDesugar  Condition=" '$(AndroidEnableDesugar)' == '' ">False</AndroidEnableDesugar>
 	<AndroidR8IgnoreWarnings    Condition=" '$(AndroidR8IgnoreWarnings)' == '' ">True</AndroidR8IgnoreWarnings>
+	<AndroidCreateProguardMappingFile Condition="'$(AndroidCreateProguardMappingFile)' == '' And '$(AndroidLinkTool)' == 'r8'">True</AndroidCreateProguardMappingFile>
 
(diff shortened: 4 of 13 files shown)


tests/apk-sizes-reference/Xamarin.Forms_Performance_Integration-Signed-Release-Bundle.apkdesc
```diff
@@ -5,7 +5,7 @@
       "Size": 3768
     },
     "classes.dex": {
-      "Size": 2802888
+      "Size": 2976868
```
jonathanpeppers, line 8: So is this why we didn't set `$(AndroidProguardMappingFile)` by default before? Because it regresses .apk size? This might be fine?

If someone didn't want this mapping, it doesn't look like there is a way to turn if off now?

dellis1972: It might well be the reason why we didn't set it. 
So perhaps we should leave it off by default still.

jonathanpeppers: For developers submitting to Google Play, it seems like they would probably rather have a 17kb larger `classes.dex` file and the warning in Google Play would go away?

Maybe we should just add a new boolean property to turn this on or off, and default to on?

dellis1972: Maybe this might work?

```
<Target Name="_SetProguardMappingFileProperty">
	<PropertyGroup>
		<AndroidProguardMappingFile Condition=" '$(AndroidLinkTool)' == 'r8' And '$(AndroidProguardMappingFile)' != 'None' ">$(OutputPath)mapping.txt</AndroidProguardMappingFile>
	</PropertyGroup>
</Target>
```

Then if a user wants to opt-out then can define 

```
< AndroidProguardMappingFile>None</AndroidProguardMappingFile>
```

or 

```
-p: AndroidProguardMappingFile=None
```

jonathanpeppers: Seems like that would work? Just couldn't name the file `None`, lol.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1825,6 +1826,12 @@ because xbuild doesn't support framework reference assemblies.
 	</_CompileDexDependsOn>
 </PropertyGroup>
 
+<Target Name="_SetProguardMappingFileProperty">
+	<PropertyGroup>
+		<AndroidProguardMappingFile Condition=" '$(AndroidLinkTool)' == 'r8' And '$(AndroidProguardMappingFile)' != 'None' ">$(OutputPath)mapping.txt</AndroidProguardMappingFile>
```
jonpryor, line 1831: I'm not sure these semantics quite make sense?

If someone builds with:

```
msbuild /t:SignAndroidPackage /p:AndroidProguardMappingFile=my-custom-mapping.txt
```

what is supposed to happen?  This isn't a "documented" scenario, but with a property name ending in "file" that doesn't imply a "boolean" value, this looks quasi-reasonable?  Except values on the command-line aren't actually modifiable, meaning `$(AndroidProguardMappingFile)` would remain `my-custom-mapping.txt`, and I'm not sure what that would *do*.

I think a "better" idea would be to use a "boolean" value, e.g. `$(AndroidCreateProguardMappingFile)` (=True or False, default to True), and use a "private" `$(_AndroidProguardMappingFile)` to contain the filename:

```xml
<AndroidCreateProguardMappingFile Condition=" '$(AndroidCreateProguardMappingFile)' == '' And '$(AndroidLinkTool)' == 'r8' ">True</AndroidCreateProguardMappingFile>
<_AndroidProguardMappingFile Condition=" '$(AndroidCreateProguardMappingFile)' == 'True' ">$(OutputPath)mapping.txt<_AndroidProguardMappingFile>
```

Documentation/guides/building-apps/build-items.md
```diff
@@ -38,6 +38,24 @@ Used to provide an AOT profile, for use with profile-guided AOT.
 It can be also used from Visual Studio by setting the `AndroidAotProfile`
 build action to a file containing an AOT profile.
 
+## AndroidAppBundleMetaDataFile
+
+Specifies a file that will be included as metadata in the Android App Bundle.
+The format of the flag value is '<bundle-path>:<physical-file>' where
```
jonpryor, line 44: This should use backticks `` ` `` not `'`, as Markdown "passes through" `<looks-like-an-html-tag-to-me>`.  If you view the Markdown preview for this page, you see:

>  The format of the flag value is ':' where 'bundle-path' denotes…

which is *not* what you intended. :-)

We may also want to link to the appropriate `bundletool build-bundle` documentation: https://developer.android.com/studio/build/building-cmdline

Documentation/guides/building-apps/build-properties.md
```diff
@@ -300,6 +300,28 @@ should be created instead of having support for all ABIs in a single `.apk`.
...
+```
+
+Note that if you are using `.aab` files the mapping file with be
+automatically included in your package. So there is no need to upload
+it to the Google Play Store manually. If you are using `.apk` files
+you will need to manually upload the `AndroidProguardMappingFile`.
+
+The default value is `False`.
+
+Added in Xamarin.Android 11.2.
```
jonpryor, line 323: Added in Xamarin.Android 12.3. :-)

Documentation/guides/building-apps/build-properties.md
```diff
@@ -300,6 +300,28 @@ should be created instead of having support for all ABIs in a single `.apk`.
 See also the [Building ABI-Specific APKs](~/android/deploy-test/building-apps/abi-specific-apks.md)
 guide.
 
+## AndroidCreateProguardMappingFile
+
+A boolean property that controls if a proguard mapping file is
+generated as part of the build process.
+
+Adding the following to your csproj will cause the file to be
+generated. This uses the `AndroidProguardMappingFile` property
```
jonpryor, line 309: We can now link to the relevant docs:

```markdown
This uses the [`AndroidProguardMappingFile`](#androidproguardmappingfile) property  …
```

Documentation/guides/building-apps/build-items.md
jonpryor: This doc section should also mention that when `$(AndroidProguardMappingFile)` isn't empty and `$(AndroidCreateProguardMappingFile)`=True (w/ appropriate doc links), we automatically create an item in this item group.

jonpryor:
Shouldn't we enable this by default for .NET 6+ apps?

src/Xamarin.Android.Build.Tasks/Tasks/BuildAppBundle.cs
```diff
@@ -63,9 +63,11 @@ public class BuildAppBundle : BundleTool
 		public string BaseZip { get; set; }
 
 		public string CustomBuildConfigFile { get; set; }
-		
+
 		public string [] Modules { get; set; }
 
+		public string [] MetaDataFiles { get; set; }
```
jonpryor, line 69: Convention question: should this be an `ITaskItem[]` and not a string, as it's coming from an item group?

dellis1972: It can be.... but the other parts of the Task use `string []` so I either go with what is already there or fix up the other code?

dellis1972:
> Shouldn't we enable this by default for .NET 6+ apps?

Enabling this causes the apk to get bigger (which is why the apkdiff desc files were updated in the first place). Is this something we want?

jonpryor:
The `RemoveEventHandlerResolution` failures are fixed by 599fffd408707677a465fe45c2a4f343389b5cad.

jonathanpeppers (approved):

jonpryor:
Context: https://discord.com/channels/732297728826277939/732297837953679412/933674595557466172

Enabling `$(AndroidCreateProguardMappingFile)`=True by default increases `.apk` size by ~1%.  We felt this was acceptable.

@dellis1972: That said, *where* is that 1% coming from?  I figured there'd be a `mapping.txt` entry somewhere in the `.apk`, but the updated `.apkdesc` files in this PR don't show the presence of a `mapping.txt`.  Where is the `mapping.txt` information kept?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1967,6 +1975,12 @@ because xbuild doesn't support framework reference assemblies.
       $(ApkFileIntermediate)
     </_BuildApkEmbedOutputs>
   </PropertyGroup>
+  <ItemGroup>
+    <AndroidAppBundleMetaDataFile
+        Condition=" '$(AndroidProguardMappingFile)' != '' And Exists ('$(AndroidProguardMappingFile)') "
```
jonpryor, line 1980: Can't `Exists()` be used with empty values?  Why can't this line just be:

```
Condition=" Exists ('$(AndroidProguardMappingFile)') "
```

jonpryor:
```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/5378

Context: 2dab2bc3725a3ca3d56c6767c60f0eecf6426f46

When uploading an `.apk` or `.aab` file which contains native
libraries such as AOT'd .NET assemblies to the Google Play store,
the Google Play Console reports the following warning:

> This App Bundle contains native code, and you've not uploaded
> debug symbols.  We recommend that you upload a symbol file to make
> your crashes and ANRs easier to analyse and debug.

Commit 2dab2bc3 addresses a similar warning regarding obfuscated
Java/Kotlin code.

Improve support for this scenario by adding a new
`$(AndroidCreateProguardMappingFile)` MSBuild property which controls
the creation of the ProGuard mapping file.

When True (the default when r8 is used), a ProGuard mapping file will
be created and added to `.apk`/`.aab` build outputs.  This increases
the size of resulting applications by ~1% for our unit tests, and
will remove the above warning for Google Play Store submissions.

To make `$(AndroidCreateProguardMappingFile)`=True *work*, add
support for a new `@(AndroidAppBundleMetaDataFile)` item group.
`%(AndroidAppBundleMetaDataFile.Include)` values are those used by
the [`bundletool build-bundle --metadata-file`][0] option:

> `--metadata-file=target-bundle-path:local-file-path`
> 
>   * `target-bundle-path` specifies a path relative to the root of
>      the app bundle where you would like the metadata file to be
>      packaged, and
>   * `local-file-path` specifies the path to the local metadata file
>     itself.

When `$(AndroidCreateProguardMappingFile)`=True, we automatically add:

	<AndroidAppBundleMetaDataFile Include="com.android.tools.build.obfuscation/proguard.map:$(AndroidProguardMappingFile" />

which causes the ProGuard mapping file to be included into the bundle.

[0]: https://developer.android.com/studio/build/building-cmdline#bundletool-build
```

dellis1972:
@jonpryor I made a few additions to your commit message and updated my commit to use it.

PR #6620, opened 2022-01-06
Bump to r8 3.0.73

jonathanpeppers:
Context: https://r8.googlesource.com/r8/+/refs/tags/3.0.73
Context: https://mvnrepository.com/artifact/com.android.tools/r8/3.0.73

Bump from r8 2.2.64 to 3.0.73.

PR #6621, opened 2022-01-06
Enable github dependabot for gradle files

jonathanpeppers:
Context: https://docs.github.com/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/
Context: https://docs.github.com/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates

To help us keep r8 and manifest-merger updated, we can configure
Github's dependabot to send us PRs weekly.

It appears we have to merge this file and "see if it works"?

I did this in another repo for NuGet packages here:

https://github.com/jonathanpeppers/boots/commit/b5f191d1fc1ccd7a024472245f6fb89ac77210e0

And the file seems to be what configures everything.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "weekly"


dellis1972 (approved):

jonathanpeppers:
Let's merge this after we get #6620, so the bot doesn't try to update r8.

We should see the bot attempt to update manifest-merger, which will be a good test.

PR #6627, opened 2022-01-08
Add missing restore instruction for building xabuild.csproj

lauxjpn:
It seems to be necessary to restore the packages before the first build of `xabuild.csproj` on Windows.


jonathanpeppers (approved):
Thanks! You can actually add `/restore` and do it in one call.

Issue #6615, opened 2022-01-06
Missing AOT'ed assemblies - Regression in VS 2022 17.1.0-pre2


tranb3r:
### Android application type

Classic Xamarin.Android (MonoAndroid12.0, etc.)

### Affected platform version

VS 2022 17.1.0-pre2

### Description

Using full aot, AOT'ed assemblies are missing in the apk.
This was working in previous previews of VS 2022 17.1.0.

### Steps to Reproduce

Enable full AOT and build an apk.
Inspect lib folder inside the built apk.

### Did you find any workaround?

No

dellis1972:
@tranb3r  do you have a full diagnostic build log for this issue? 
I just tried to build a project with Full AOT against on latest `main` branch and I get the AOT'd files in the apk. 

It might also be useful to provide the settings for the configuration you are building for so we can try to replicate your exact issue.

tranb3r:
Here are the logs.
[build.zip](https://github.com/xamarin/xamarin-android/files/7823340/build.zip)

build-ko.log is the one were aot'ed assemblies are missing.
build-ok.log is ok and was built on the same machine with VS 2022 17.0.4.

Thanks for investigating, and please ask if you need more information.

dellis1972:
@tranb3r ok this is a problem

```
obj\Release\120\aot\arm64-v8a\libaot-System.Text.Encoding.CodePages.dll.so is an empty (no executable code) AOT assembly, not including it in the archive 
```

@grendello  has there been any bugs recently in the code which skips aot'd .so files when adding them to the apk?

dellis1972:
@tranb3r are you able to provide us a zip containing the AOT'd  .so files? There is something about them which is tripping up our code which checks to see if they are empty.

tranb3r:
Here is a zip with the aot'ed so files:
[arm64-v8a.zip](https://github.com/xamarin/xamarin-android/files/7824311/arm64-v8a.zip)

tranb3r:
I've tested the ELFHelper.IsEmptyAOTLibrary method on my so files, the result is that all libs are considered as empty (which is obviously incorrect).

tranb3r:
I'm actually using these arguments:
`<AndroidAotAdditionalArguments>no-write-symbols,nodebug</AndroidAotAdditionalArguments>`
When I remove this from my csproj, there is no more issue. The no-write-symbols arg probably messes with your IsEmptyAOTLibrary code.
[...]

PR #6630, opened 2022-01-10
Don't skip symbol-less AOT assemblies

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/6615
Context: db161ae7eaad1acb344192614eab9ebf14896291

db161ae7 added a way to analyze and skip native shared libraries
produced by the Mono AOT compiler and skip packaging for those which
contain no executable code (a common case for Profiled AOT builds).
However, it turns out that by adding the following property to the
project configuration:

    <AndroidAotAdditionalArguments>no-write-symbols,nodebug</AndroidAotAdditionalArguments>

Can make the `ELFHelper` treat the shared library in question as empty,
even though it does contain executable, AOT-d, code.

The reason for this is the way Mono generates code when the above
options are in question.  It will output a blob of executable code
without any labels (which are turned to symbols when compiling and
linking) and then add just a single non-function symbol whose name ends
with the `_plt` prefix.  This symbol is essentially a jump table - a
collection of tiny snippets of code which jump into the executable code
blob mentioned above.  Since the `*_plt` symbol is **not** a
function (that is its symbol type is not FUNC), it was skipped by
ELFHelper:

    27: 0000000000078620 22480 NOTYPE  LOCAL  DEFAULT     6 mono_aot_Mono_Android_plt

Furthermore, since shared libraries built with the above options do not
contain any `FUNC` symbols, they were consequently rejected as empty and
not packaged, leading to issue6615.

Fix the problem by looking for non-empty symbols in executable sections
of the shared library that end in `_plt`.


dellis1972 (approved):

jonpryor:
@grendello : could this PR also modify or add a unit test which sets `$(AndroidAotAdditionalArguments)`?

PR #6631, opened 2022-01-10
[Xamarin.Android.Build.Tasks] Rename Unzip to UnzipToFolder

dellis1972:
MSBuild ships with an Unzip task already. This task has a different
API to our one. So we would rename the Task so that is does not clash
with the existing task.


jonathanpeppers (approved):
This looks good if tests end up green. 👍

PR #6635, opened 2022-01-11
[One .NET] fix GetAndroidDependencies target with --no-restore

jonathanpeppers:
I've noticed the following sometimes happens when a new .NET MAUI
project is created in VS 2022 on Windows:

    dotnet\sdk\6.0.200-preview.22055.18\Sdks\Microsoft.NET.Sdk\targets\Microsoft.PackageDependencyResolution.targets(267,5):
    error NETSDK1004: Assets file 'obj\project.assets.json' not found.
    Run a NuGet package restore to generate this file.

I seem to only get this for MAUI projects and only *sometimes*.
`dotnet new android` projects seem to always work fine.

When this error occurs, I think the dropdown fails to load the device
list. What you end up with is the play button that just says
`> Android Emulator`. After some amount of "fiddling", you can get the
IDE to load the device list.

What I think is happening is:

1. NuGet restore takes longer in MAUI projects than `dotnet new android`

2. Sometimes `GetAndroidDependencies` runs *before* NuGet restore, and
   that triggers the above error.

I could validate this hypothesis with:

    > dotnet new android
    > dotnet build -t:GetAndroidDependencies --no-restore

And I get the above error! I could also reproduce in a test.

This target works fine in a classic Xamarin.Android app:

    git clean -dxf .\samples\
    msbuild .\samples\HelloWorld\HelloWorld\HelloWorld.csproj -t:GetAndroidDependencies

(note that the absence of `-restore`, is the same as `--no-restore`)

After digging through `.binlog` files, I see
`$(_ResolveSdksDependsOnTargets)` depends on
`ResolveTargetingPackAssets`. This triggers the failure.

Reviewing another target:

https://github.com/dotnet/sdk/blob/22c4860dcb2cf6b123dd641cc4a87a50380759d5/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.FrameworkReferenceResolution.targets#L49-L63

`ProcessFrameworkReferences` runs a bit earlier, and does not appear
to fail with `--no-restore`. It also has a
`%(TargetingPack.PackageDirectory)` value we can use instead.

After this change, my earlier test passes. I think this might also
have a small improvement to the performance of the
`GetAndroidDependencies` target.

This change should improve the overall reliability of creating new
.NET MAUI projects in VS.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
@@ -899,6 +899,16 @@ public abstract class Foo<TVirtualView, TNativeView> : ViewHandler<TVirtualView,
 			Assert.IsTrue (builder.Build (), $"{proj.ProjectName} should succeed");
 		}
 
+		[Test]
+		public void GetAndroidDependencies()
+		{
+			var proj = new XASdkProject ();
+			var builder = CreateDotNetBuilder (proj);
+
+			// `dotnet build -t:GetAndroidDependencies --no-restore` should succeed
+			Assert.IsTrue (builder.Build (target: "GetAndroidDependencies", norestore: true), $"{proj.ProjectName} should succeed");
+		}
+
 		DotNetCLI CreateDotNetBuilder (string relativeProjectDir = null)
 		{
 			if (string.IsNullOrEmpty (relativeProjectDir)) {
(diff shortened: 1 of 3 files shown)


jonathanpeppers:
A couple issues:
1. Some projects hit a new warning: `warning XA5300:  Unable to determine Xamarin.Android version.`
2. Binding projects appear to fail? ` (_ExportJarToXml target) -> Xamarin.Android.Bindings.ClassParse.targets(36,5): error XABGN7009: System.InvalidOperationException: Sequence contains no matching element`

Will look into these.

pjcollins (approved):

dellis1972 (approved):

PR #6640, opened 2022-01-13
[build] Remove jna content from manifestmerger.jar

pjcollins:
Our .pkg notarization step has been failing since bumping manifestmerger
to 30.0.4.  After further investigation, this appears to be due to some
new macOS binaries:

    "issues": [
        {
          "severity": "error",
          "code": null,
          "path": "xamarin.android.pkg Contents/Payload/Library/Frameworks/Xamarin.Android.framework/Versions/12.2.99.37/lib/xamarin.android/xbuild/Xamarin/Android/manifestmerger.jar/com/sun/jna/darwin/libjnidispatch.jnilib",
          "message": "The binary is not signed.",
          "docUrl": null,
          "architecture": "i386"
        }
    ]

Rather than trying to sign/harden these files ourselves, we should be
able to remove them from manifestmerger.jar to fix this.

### DIFF
--- a/build-tools/installers/sign-pkg-content.proj
+++ b/build-tools/installers/sign-pkg-content.proj
@@ -43,24 +43,8 @@ ourself (using an empty signing identity) before passing these files to ESRP.
     <RemoveDir Directories="$(_BundleToolExtractLocation)" />
   </Target>
 
-  <!-- Remove the jna subdirectory in manifestmerger.jar as it should not be needed by our usage.
-    This simplifies signing/notarization and reduces our installation footprint slightly. -->
-  <Target Name="_RemoveJnaFromManifestMerger" >
-    <PropertyGroup>
-      <_ManifestMergerJar>$([System.IO.Path]::GetFullPath('$(MSBuildSrcDir)\manifestmerger.jar'))</_ManifestMergerJar>
-      <_ManifestMergerExtractLocation>$(MSBuildSrcDir)\mm-unzipped</_ManifestMergerExtractLocation>
-    </PropertyGroup>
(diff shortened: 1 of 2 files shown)


pjcollins:
This approach doesn't seem to have caused any new test failures. Notarization test is here - https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5625313&view=results.

build-tools/installers/sign-pkg-content.proj
```diff
@@ -43,8 +43,24 @@ ourself (using an empty signing identity) before passing these files to ESRP.
     <RemoveDir Directories="$(_BundleToolExtractLocation)" />
   </Target>
 
+  <!-- Remove the jna subdirectory in manifestmerger.jar as it should not be needed by our usage.
+    This simplifies signing/notarization and reduces our installation footprint slightly. -->
+  <Target Name="_RemoveJnaFromManifestMerger" >
```
jonathanpeppers, line 48: Maybe instead we should update:

https://github.com/xamarin/xamarin-android/blob/3e27d4b4c647090bcb5994d77e7de6ada2123879/src/manifestmerger/build.gradle#L41

To do something like: `'com/sun/jna/*/*jnidispatch*'`

pjcollins: Thanks, this is a much better approach.

pjcollins:
New notarization test: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5625448&view=results

pjcollins:
This has fixed our classic signing job, I'm going to merge it.

PR #6642, opened 2022-01-13
[build] Remove aapt2 from bundletool during build

pjcollins:
We've been removing the aapt2 executables included in bundletool.jar
during our CI builds on macOS only.  This is being done to simplify the
work needed for installer signing and notarization.  We should move this
logic to the project that builds bundletool.jar to establish consistent
behavior across build environments.


pjcollins:
Notarization test: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5626328&view=results

dellis1972 (approved):

PR #6643, opened 2022-01-13
[Xamarin.Android.Build.Tasks] Add support for '<AndroidNamespaceReplacement>'.

jpobst:
Context: https://github.com/xamarin/java.interop/issues/727

Add an MSBuild equivalent to the `<ns-replace>` metadata feature added in https://github.com/xamarin/java.interop/issues/727.

This takes namespace transforms like:
```xml
<ItemGroup>
  <AndroidNamespaceReplacement Include='Androidx' Replacement='AndroidX' />
  <AndroidNamespaceReplacement Include='Com' Replacement='' />
  <AndroidNamespaceReplacement Include='Com.Google.' Replacement='Google' />
</ItemGroup>
```

And places them into a generated `msbuild-metadata.xml` file that is passed to `generator`:

```xml
<metadata>
  <ns-replace source='Androidx' replacement='AndroidX' />
  <ns-replace source='Com' replacement='' />
  <ns-replace source='Com.Google.' replacement='Google' />
</metadata>
```

Today, this only supports `<ns-replace>` metadata, but it could be expanded in the future if we feel that allowing other metadata is desirable.  (I am personally skeptical 😁.)

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Xml;
 using System.Xml.Linq;
 using Microsoft.Build.Framework;
 using Microsoft.Android.Build.Tasks;
@@ -55,6 +56,7 @@ namespace Xamarin.Android.Tasks
 		public ITaskItem[] TransformFiles { get; set; }
 		public ITaskItem[] ReferencedManagedLibraries { get; set; }
 		public ITaskItem[] AnnotationsZipFiles { get; set; }
+		public ITaskItem[] NamespaceTransforms { get; set; }
 
 		public ITaskItem[] JavadocXml { get; set; }
 		public string JavadocVerbosity { get; set; }
@@ -113,6 +115,20 @@ namespace Xamarin.Android.Tasks
 					}
 				}
 
+
+			if (NamespaceTransforms?.Any () == true)
+				foreach (var nt in NamespaceTransforms) {
+					if (!nt.MetadataNames.OfType<string> ().Contains ("Replacement", StringComparer.OrdinalIgnoreCase))
+						Log.LogCodedError (
+							code: "XA4233",
+							message: Properties.Resources.XA4233,
+							messageArgs: new [] { nt.ItemSpec }
+						);
+				}
+
+			if (Log.HasLoggedErrors)
+				return false;
+
 			return base.RunTask ();
 		}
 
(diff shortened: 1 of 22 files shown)


jonpryor:
The `RemoveEventHandlerResolution` failures are fixed by 599fffd408707677a465fe45c2a4f343389b5cad.

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
```diff
@@ -126,6 +128,23 @@ protected override string GenerateCommandLineCommands ()
 		{
 			var cmd = GetCommandLineBuilder ();
 
+			if (NamespaceTransforms?.Any () == true) {
+				// ex: obj/Debug/generated/msbuild-metadata.xml
+				var transform_file = Path.Combine (OutputDirectory, "..", "msbuild-metadata.xml");
+
+				var xml = new XDocument ();
```
jonpryor, line 135: I'm not sure you need the `XDocument` at all?  `XElement.WriteTo()` also works…

jonpryor:
@jpobst: the new item group should also be documented in: https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/building-apps/build-items.md

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
```diff
@@ -126,6 +128,23 @@ protected override string GenerateCommandLineCommands ()
...
+				var root = new XElement ("metadata");
+				xml.Add (root);
+
+				foreach (var nt in NamespaceTransforms)
+					xml.Root.Add (new XElement ("ns-replace", new XAttribute ("source", nt.ItemSpec), new XAttribute ("replacement", nt.GetMetadata ("replacement"))));
```
jonathanpeppers, line 140: Should we emit a warning/error if `%(Replacement)` is blank? I'm just thinking if someone put `Replace="AndroidX"` on accident, they won't know what's wrong?

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
```diff
@@ -126,6 +128,23 @@ protected override string GenerateCommandLineCommands ()
 		{
 			var cmd = GetCommandLineBuilder ();
 
+			if (NamespaceTransforms?.Any () == true) {
```
jonathanpeppers, line 131: This is probably fine, but feels weird. Would this be just as good?
```suggestion
			if (NamespaceTransforms != null && NamespaceTransforms.Length > 0) {
```

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
jpobst: `%(Replacement)` being *blank* is valid and useful.  I guess the question is should we require them to provide the blank attribute `Replacement=""`?  Perhaps we should.

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
```diff
@@ -126,6 +128,23 @@ protected override string GenerateCommandLineCommands ()
...
+				// ex: obj/Debug/generated/msbuild-metadata.xml
+				var transform_file = Path.Combine (OutputDirectory, "..", "msbuild-metadata.xml");
```
dellis1972, line 133: Does this file need to be added to `FileWrites` ? It will be deleted on the next build otherwise.

jpobst: The entire generated output directory is deleted on every build:

https://github.com/xamarin/xamarin-android/blob/main/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Bindings.Core.targets#L53-L54

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
jpobst: Yes, those are equivalent.  I think `Any` and null-check operator is generally considered the modern "best practice".

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
jpobst: Updated.

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
jonathanpeppers: I generally avoid `using System.Linq;`, but there are some cases where `OrderBy()` is ok.

This is fine, though.

src/Xamarin.Android.Build.Tasks/Tasks/Generator.cs
jpobst: Added and documented new error.

PR #6654, opened 2022-01-19
[ci] Use Windows 2022 agent pools

pjcollins:
Updates our Microsoft hosted and VSEng scale set Windows pools to the
latest and greatest images.  The designer integration test job has been
disabled on Windows for some time, and it will continue to use the
`windows-2019` pool until it is re-enabled.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -14,7 +14,7 @@ resources:
   - repository: yaml
     type: github
     name: xamarin/yaml-templates
-    ref: refs/heads/dev/pjc/sign-use-nuget
+    ref: refs/heads/main
     endpoint: xamarin
   - repository: monodroid
     type: github
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -14,7 +14,7 @@ resources:
   - repository: yaml
     type: github
     name: xamarin/yaml-templates
-    ref: refs/heads/main
+    ref: refs/heads/dev/pjc/sign-use-nuget
     endpoint: xamarin
   - repository: monodroid
     type: github
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -1362,7 +1362,8 @@ stages:
 
   - job: push_to_nuget_org
     displayName: Push to NuGet.org
-    pool: $(VSEngMicroBuildPool)
+    pool:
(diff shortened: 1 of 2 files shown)


pjcollins:
Testing with https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5644645&view=results

build-tools/automation/azure-pipelines.yaml
```diff
@@ -14,7 +14,7 @@ resources:
   - repository: yaml
     type: github
     name: xamarin/yaml-templates
-    ref: refs/heads/main
+    ref: refs/heads/dev/pjc/sign-use-nuget
```
jonpryor, line 17: I assume this means that there is some "other" PR in xamarin/yaml-templates -- perhaps xamarin/yaml-templates#160? -- that should be merged *first*?

build-tools/automation/azure-pipelines.yaml
```diff
@@ -1081,7 +1081,8 @@ stages:
   - job: designer_integration_win
     condition: false
     displayName: Windows
-    pool: $(HostedWinVS2019)
```
jonpryor, line 1084: Not sure I understand this change.  Shouldn't we *want* to use a variable?

pjcollins: I didn't add a variable for this since it's now the only job using the 2019 pool. I also think we will want to update it to use the 2022 image defined in `$(HostedWinImage)` when we re-enable it.

build-tools/automation/azure-pipelines.yaml
pjcollins: That's correct, I should marked this as a draft after hitting the issue fixed by xamarin/yaml-templates#160. That PR has now been merged and the branch reference has been updated here.

jonpryor:
Restarted the failing **MSBuild Emulator Tests** jobs, to see if e.g. [`Install_CSharp_Change`](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5649587&view=ms.vss-test-web.build-test-results-tab&runId=31711158&resultId=100020&paneView=debug) continued failing:

```
Exceeded expected time of 4500ms, actual 4659.879ms
```

The `RemoveEventHandlerResolution` test also failed, but that's due to a branch name change, and fixed in 599fffd408707677a465fe45c2a4f343389b5cad.

pjcollins:
I think this should be good to go, the pool changes here are primarily for post release jobs which were all working in https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5646233&view=results.

PR #6656, opened 2022-01-20
[docs] improve the `dotnet trace` instructions

jonathanpeppers:
I simplified some of the steps, and included instructions for installing `dotnet-dsrouter`.

These instructions should be a little simpler to follow, for both Windows & Unix-like OS users.

/cc @eerhardt


eerhardt (approved):
LGTM. Thanks!

Issue #6480, opened 2021-11-11
Can't debug with VS 2022 17.1.0 Prev 1.0


gmck:
I can't debug at all. Hangs on the splash screen. A breakpoint in OnCreate of the MainActivity is never hit as I get the following in the debugger. 

```
System.InvalidProgramException
  Message=Invalid IL code in Mono.SystemDependencyProvider:Initialize (): IL_002e: endfinally
```
A release build runs fine, except for what was reported in #6479.

Tried on another machine, and get the same result. Both machines Windows 11

Reverted back to VS 2022 17.0.0 and can debug normally again.

lanarchyste:
I have the same error and behavior with Visual Studio for Mac 8.10.13 (build 11) and Xamarin.Android 12.1.99.75 (version that fixes the build with the exported attribute)

tspersian:
same vs 17.1.0 Preview 1.1 
android 7 (Galaxy A7)

steveisok:
@thaystg can you please take a look at this?

thaystg:
Can you please help me to reproduce it? What kind of project are you creating?

gmck:
@thaystg  _What kind of project are you creating?_

Seems a strange question! What type of answer are you looking for? My own app is a single activity/multi fragment app using the NavigationComponent, although I can't see how that relates to a debugging problem that now hangs on the splash screen of the app and never even gets to the OnCreate of the MainActivity. This wasn't a problem until 17.1.0 Prev 1.0. So it would seem to indicate that something changed between 17.0.0 and 17.1.0 Prev 1.0.  17.1.0 Prev 1.1 also has the same problem.

If the above is not what you are looking for could you please be more specific re the question of the kind of project?

thaystg:
Two questions:
1) If you create a new project from the templates list of VS, you can reproduce the error? What kind of project should I create to reproduce it?
2) If you start the app without debugging, I mean, the app in debug mode but started using this option:
![image](https://user-images.githubusercontent.com/4503299/142520234-92a35217-6b40-426c-b981-f04a281a6576.png)

Thanks a lot.

gmck:
@thaystg. Thanks for your reply. No, I can't reproduce it when I create say a Blank app from the templates.

If I start the app via Start Without Debugging - it just hangs on the splash screen. Hence, that was why I tried to debug it and got the error.  The debugger is telling me it is not my code e.g.

[...]

PR #6660, opened 2022-01-24
[Xamarin.Android.Build.Tasks] default $(AndroidUseAssemblyStore) to false for debug builds

jonathanpeppers:
Context: https://github.com/xamarin/QualityAssurance/tree/master/Manual/CrossPlatformDebugging
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1430409
Fixes: https://github.com/xamarin/xamarin-android/issues/6480

A `CrossPlatformDebugging` sample project in the QA repo fails to
debug with:

    System.InvalidProgramException
    Message=Invalid IL code in Mono.SystemDependencyProvider:Initialize (): IL_002e: endfinally

The problem goes away if you do one of:

* Turn *on* Fast Deployment
* Add `<AndroidUseAssemblyStore>false</AndroidUseAssemblyStore>`
* Change `$(DebugType)` to `portable`. (the project has `full`)

The problem appears to be:

1. `$(AndroidUseAssemblyStore)` defaults to `true` when you turn Fast
   Deployment *off*.

2. Assembly stores appear to have some issue when `DebugType=full` --
   that is where the crash occurs.

In this case let's address No. 1, as we don't *really* want the
assembly store to be used when debugging. I reordered default values
for MSBuild properties, so that `$(AndroidIncludeDebugSymbols)` is
taken into account.

I added parameters for a debugging test for `DebugType=full`, and I
updated another test that used to set `AndroidUseAssemblyStore=false`
that should not be needed.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest.cs
@@ -1035,7 +1035,6 @@ AAMMAAABzYW1wbGUvSGVsbG8uY2xhc3NQSwUGAAAAAAMAAwC9AAAA1gEAAAAA") });
 			var proj = new XamarinAndroidApplicationProject {
 				EmbedAssembliesIntoApk = true,
 			};
-			proj.SetProperty ("AndroidUseAssemblyStore", "False");
 			using (var b = CreateApkBuilder ()) {
 				var reference = new BuildItem.Reference ("PdbTestLibrary.dll") {
 					WebContentFileNameFromAzure = "PdbTestLibrary.dll"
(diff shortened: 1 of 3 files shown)


dellis1972 (approved):
Look ok to me

grendello (approved):

jonathanpeppers:
Some networking/SSL failures seem to be the only failures:

https://discord.com/channels/732297728826277939/732297837953679412/935538687968694402

Going to merge, I think.

MaxFelinger:
I cannot turn on Fast Deploy on new Emulators (API level 32 and 33). 
PS: I'm using a Mac Book Pro with an M1 Pro chip. I'm using Xamarin.Android 12.2.8.3 (Visual Studio for Mac 8.10.25 (Build 2)).

Error Message: 
  Xamarin.Android.Common.Debugging.targets(611, 5): [XA0129] Error deploying 'files/.__override__/nn/MyPackageName.resources.dll'.
Please set the 'EmbedAssembliesIntoApk' MSBuild property to 'true' to disable Fast Deployment in the Visual Studio project property pages, or edit the project file in a text editor.

PR #6661, opened 2022-01-24
Revert "[One .NET] fix GetAndroidDependencies target with --no-restore"

jonathanpeppers:
Context: https://github.com/dotnet/sdk/issues/23562

This reverts commit a21d1a7d.

When installing from Visual Studio, builds can hit the error:

    error XA0002: Could not find mono.android.jar

Running `dotnet workload install maui` over top, solves the issue!

Since, `%(TargetingPack.PackageDirectory)` appears to be blank when
installed from Visual Studio, I think we just have to revert this
change.


dellis1972 (approved):

jonathanpeppers:
Onlyl failures are networking test errors like: 
```
System.AggregateException: One or more errors occurred. (The SSL connection could not be established, see inner exception.) ---> System.Net.Http.HttpRequestException: The SSL connection could not be established, see inner exception.
```
Going to merge.

PR #6662, opened 2022-01-25
[Xamarin.Android.Build.Tasks] Support %(JavaSourceJar.DocRootUrl)

pjcollins:
Context: https://github.com/xamarin/java.interop/commit/13def0e105f876678256a53b2f362380cd4818a0

Bump to xamarin/java.interop@32635fd67bfc74fbb7d0370d1bc75bd13c838b61
Changes: https://github.com/xamarin/java.interop/compare/d3f0c5c6f634380b6156c98f57e98463adf0de52...32635fd67bfc74fbb7d0370d1bc75bd13c838b61
  * xamarin/java.interop@312fbf43: [jnienv-gen] Add possible C#9 function pointer backend
  * xamarin/java.interop@4cbc07a5: [Java.Interop.Tools.JavaSource] Replace TODO with To be added
  * xamarin/java.interop@0c90cf5c: [generator] Add support for <ns-replace> metadata.
  * xamarin/java.interop@13def0e1: [Java.Interop.Tools.JavaSource] Add {@docroot} support
  * xamarin/java.interop@88d6093c: [ci] Build without -m:1.
  * xamarin/java.interop@32635fd6: [Xamarin.Android.Tools.Bytecode] Extend api xml doc type detection

Adds `%(DocRootUrl)` to the set of metadata that is supported by
`@(JavaSourceJar)` items.  Fixes processing of `%(JavadocUrlPrefix)` and
`%(JavadocUrlStyle)` metadata values.

The `<_BuildAndroidJavadocXml/>` target has also been updated to use the
new `--doc-root-url` parameter supported by `java-source-utils.jar`.


src/Xamarin.Android.Build.Tasks/Tasks/JavaSourceUtils.cs
```diff
@@ -139,11 +140,15 @@ string CreateResponseFile ()
 			}
 			if (!string.IsNullOrEmpty (JavadocUrlPrefix)) {
 				AppendArg (response, "--doc-url-prefix");
-				AppendArg (response, Path.GetFullPath (JavadocUrlPrefix));
+				AppendArg (response, JavadocUrlPrefix);
```
jonathanpeppers, line 143: So several of these weren't even files? And we were doing `GetFullPath()`?

jonathanpeppers (approved):
LGTM

src/Xamarin.Android.Build.Tasks/Tasks/JavaSourceUtils.cs
pjcollins: Yeah, I think this may have just been an oversight initially.

Documentation/guides/building-apps/build-items.md
```diff
@@ -370,6 +370,11 @@ Starting in Xamarin.Android 11.3, the following MSBuild metadata is supported:
     online documentation.  Only one style is currently supported:
     `developer.android.com/reference@2020-Nov`.
 
+Starting in Xamarin.Android 12.1, the following MSBuild metadata is supported:
```
jonpryor, line 373: Visual Studio 17.1, as found in the d17-1 branch, is Xamarin.Android 12.2, and this feature won't be available until the following release, Xamarin.Android *12.3*. :-)

src/Mono.Android/Mono.Android.targets
```diff
@@ -285,6 +287,7 @@
     <PropertyGroup>
       <_Binlog>$(MSBuildThisFileDirectory)../../bin/Build$(Configuration)/UpdateExternalDocumentation-$([System.DateTime]::Now.ToString ("yyyyMMddTHHmmss")).binlog</_Binlog>
     </PropertyGroup>
+    <Exec Command="dotnet build -v:n -c $(Configuration) $(JavaInteropFullPath)/build-tools/jnienv-gen/jnienv-gen.csproj" />
```
jonpryor, line 290: I don't understand this change.  Why do we need it?  If we *do* need it, shouldn't we instead add `jnienv-gen.csproj` to `@(ProjectReference)` in `Mono.Android.csproj`?

Documentation/guides/building-apps/build-items.md
pjcollins: Oops I was thinking about the target framework version rather than product version 🤦

src/Mono.Android/Mono.Android.targets
pjcollins: This was needed when building from a fully clean state, I've added the project reference to Mono.Android.csproj instead to see if that will work.

pjcollins: I'm just going to back this change out for now as the proper fix may warrant a java.interop changes as well.

jonpryor:
Why does this add `src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Resources/javasourcejartest-javadoc.jar` and `…/javasourcejartest.jar` and `…/javasourcejartest-sources.jar`?

These are built and included by: https://github.com/xamarin/xamarin-android/blob/main/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Directory.Build.targets

pjcollins:
> Why does this add `src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Resources/javasourcejartest-javadoc.jar` and `…/javasourcejartest.jar` and `…/javasourcejartest-sources.jar`?

I ran the `UpdateResources` target locally to update these files after changing the java source, I had assumed that these resource updates were done manually as needed. I don't see anything else that calls the `UpdateResources` target in that file, but maybe I am missing something?

jonpryor:
I don't understand why the [`DotNetInstallAndRun(True,False,null)`](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5680265&view=ms.vss-test-web.build-test-results-tab&runId=32193276&resultId=100019&paneView=attachments) test is failing :-(

```
Running: /Users/runner/Library/Android/dotnet/dotnet run --project "/Users/runner/work/1/a/TestRelease/01-27_20.04.58/temp/DotNetInstallAndRunTrueFalsenull/UnnamedProject.csproj" --no-build /bl:"/Users/runner/work/1/a/TestRelease/01-27_20.04.58/temp/DotNetInstallAndRunTrueFalsenull/run.binlog"
Exit Code: 1
```

Additionally, `run.binlog` isn't an attachment.  Could we update the unit test infrastructure so that `run.binlog` is included?

jonpryor:
The [`Mono.Android.NET_Tests-Interpreter` crash](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5680265&view=ms.vss-test-web.build-test-results-tab&runId=32192988&paneView=debug&resultId=100000) is also concerning. :-(

jonpryor:
@pjcollins [wrote](https://github.com/xamarin/xamarin-android/pull/6662#issuecomment-1023752970):

> I ran the `UpdateResources` target locally to update these files after changing the java source, I had assumed that these resource updates were done manually as needed.

My mistake; I thought these `.jar` files were being *added*, but they were instead *updates*.  Apologies.  See also commit 16e0aa21d9:

> A new `UpdateResources` target will update the appropriate `.jar` files.

pjcollins:
I'm going to try to address some of the test failure reporting limitations mentioned here with https://github.com/xamarin/xamarin-android/pull/6681. I think this should be good to go otherwise.

PR #6665, opened 2022-01-25
Use custom validation callback for server certificates in HTTP handlers

simonrozsival:
The `AndroidClientHandler` and `AndroidMessageHandler` classes both have the `ServerCertificateCustomValidationCallback` property, which should be useful e.g. to allow running the Android app against a server with a self-signed SSL certificate during development, but the callback is never used. Unfortunatelly since .NET 6 the `System.Net.Http.SocketsHttpHandler` for Android doesn't support the use case anymore. That means that [the recommended way of connecting to local web server](https://docs.microsoft.com/en-us/xamarin/cross-platform/deploy-test/connect-to-local-web-services) won't work in MAUI.

This PR introduces an implementation of `IX509TrustManger` which wraps the default Java X509 trust manager and calls the user's callback on top of the default validation.

- The Legacy APK test works just fine.
- The .NET APK test doesn't even build and I'm not sure what to do about it. I'd be grateful for any pointers to setup of .NET 6 APK tests. - _I removed the .NET test as I couldn't get it to work._
- It turns out that `X509Chain` `Build` function doesn't work on Android, so I'm not calling it and I'm passing the chain to the callback directly. ~I'm leaving the TODO in the source for now and maybe somebody will have an idea what to do about it during reviews.~

Ref https://github.com/dotnet/runtime/issues/62966


src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
```diff
@@ -0,0 +1,97 @@
...
+using Android.Runtime;
+
+using Javax.Net.Ssl;
+
+using JavaCertificateException = Java.Security.Cert.CertificateException;
+using JavaX509Certificate = Java.Security.Cert.X509Certificate;
+
+namespace Xamarin.Android.Net
+{
+    [Register ("xamarin/android/net/X509TrustManagerWithValidationCallback")]
```
jonpryor, line 17: This repo uses tabs for indentation, not spaces.  Thank you.

jonpryor: What's odd is that [`.editorconfig`](https://github.com/xamarin/xamarin-android/blob/main/.editorconfig) *should* be formatting this file appropriately.  What editor are you using?  Does it support `.editorconfig` files?

simonrozsival: I'm using VS Code so it should pick up the config just fine 🤔 I'll convert the spaces to tabs, no problem. Thanks for pointing it out!

grendello (approved):

simonrozsival:
@grendello Do you think we could merge this PR now?

grendello:
@simonrozsival I'm fine with it, @jonpryor ?

steveisok:
@jonpryor Do you think we can take this change?

Eagle3386:
> @jonpryor Do you think we can take this change?

The check below says, @pjcollins is the pending reviewer, not @jonpryor - maybe change that? 🤔

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
```diff
@@ -157,7 +157,7 @@ public CookieContainer CookieContainer
 
 		public bool CheckCertificateRevocationList { get; set; } = false;
 
-		public Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> ServerCertificateCustomValidationCallback { get; set; }
+		public Func<HttpRequestMessage, X509Certificate2?, X509Chain?, SslPolicyErrors, bool>? ServerCertificateCustomValidationCallback { get; set; }
```
jonathanpeppers, line 160: If users never set this, would the changes here be a no-op? There is a lot of certificate loading here, and I'm wondering how that will impact performance.

jonathanpeppers (commented):
Do you have an idea of the performance impact here? Maybe we could setup a BenchmarkDotNet project and compare before & after with a simple GET & POST or something.

We don't have one in this repo yet, but I set something up in dotnet/maui:

https://github.com/dotnet/maui/tree/main/src/Core/tests/Benchmarks.Droid

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64XFormsDotNet.apkdesc
```diff
@@ -113,10 +119,16 @@
...
+    },
+    "assemblies/System.Security.Cryptography.Encoding.dll": {
+      "Size": 13705
     },
     "assemblies/System.Security.Cryptography.Primitives.dll": {
-      "Size": 2973
+      "Size": 7317
+    },
+    "assemblies/System.Security.Cryptography.X509Certificates.dll": {
+      "Size": 52650
```
jonathanpeppers, line 131: So I'm guessing an app using `HttpClient` will bring in these assemblies by default now? Is the linker preserving every property on `AndroidMessageHandler`?

Maybe we need to look into why that's happening.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -479,6 +479,15 @@ stages:
         artifactSource: bin/Test$(ApkTestConfiguration)/Xamarin.Android.EmbeddedDSO_Test-Signed.apk
         artifactFolder: Default
 
+    - template: yaml-templates/apk-instrumentation.yaml
+      parameters:
+        configuration: $(ApkTestConfiguration)
+        testName: Xamarin.Android.Net-Tests
```
jonpryor, line 485: @simonrozsival: Why add a new app instead of adding the new tests to the existing `tests/Mono.Android-Tests` app?  I don't understand the rationale.

tests/Xamarin.Android.Net-Tests/ServerCertificateCustomValidationCallbackTest.cs
```diff
@@ -0,0 +1,120 @@
...
+using Xamarin.Android.Net;
+
+using NUnit.Framework;
+
+namespace Xamarin.Android.Net.Tests
+{
+	[Category ("InetAccess")]
+	public abstract class ServerCertificateCustomValidationCallbackTest
+	{
+		protected abstract HttpMessageHandler CreateHandlerWithCallback(Func<HttpRequestMessage, X509Certificate2?, X509Chain?, SslPolicyErrors, bool> callback);
```
jonathanpeppers, line 16: @simonrozsival so I think the only change needed, is to move these tests to:

https://github.com/xamarin/xamarin-android/tree/main/tests/Mono.Android-Tests/Xamarin.Android.Net

Then they will run on CI, and we won't need a new project. Thanks!

jonathanpeppers (changes requested):

dellis1972 (approved):

build-tools/automation/azure-pipelines.yaml
simonrozsival: I started with a test in `tests/Mono.Android-Tests` but I wasn't able to get it woring. I was getting `Java.Lang.ClassNotFoundException : xamarin.android.net.X509TrustManagerWithValidationCallback` because a java wrapper for the class I added, `X509TrustManagerWithValidationCallback`, isn't being generated automatically. I got [some tips from you](https://discord.com/channels/732297728826277939/732297837953679412/936242135286575124) and I instead created an apk project which allowed me to run the test successfully.

I reverted the apk project. I think I somehow need to special-case the `X509TrustManagerWithValidationCallback` type so that it's built into `mono.android.jar` to make the test work inside of `tests/Mono.Android-Tests`.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64XFormsDotNet.apkdesc
simonrozsival: I thought that the linker would remove these assemblies when the callback isn't set but apparently that's not happening. I need to check the .dll that are produced in these tests and what's trimmed away and what's left in there.

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
akoeplinger: This will be a no-op now.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64XFormsDotNet.apkdesc
akoeplinger: I was able to figure out how to make the linker trim away these dependencies when the callback is not set, I pushed to the PR and @simonrozsival will update the .apkdesc files with the new values.

simonrozsival:
@jonathanpeppers @jonpryo I've moved the code into `tests/Mono.Android-Tests` but now the tests are failing because the new class can't be instantiated:
```
Java.Lang.ClassNotFoundException : xamarin.android.net.X509TrustManagerWithValidationCallback
----> Java.Lang.ClassNotFoundException : xamarin.android.net.X509TrustManagerWithValidationCallback
 at Java.Interop.JniEnvironment.StaticMethods.CallStaticObjectMethod(JniObjectReference , JniMethodInfo , JniArgumentValue* )
   at Android.Runtime.JNIEnv.FindClass(String )
   at Android.Runtime.JNIEnv.AllocObject(String )
   at Android.Runtime.JNIEnv.StartCreateInstance(String , String , JValue* )
   at Android.Runtime.JNIEnv.StartCreateInstance(String , String , JValue[] )
   at Xamarin.Android.Net.X509TrustManagerWithValidationCallback..ctor(IX509TrustManager , HttpRequestMessage , Func`5 )
   at Xamarin.Android.Net.X509TrustManagerWithValidationCallback.Helper.Inject(ITrustManager[] , HttpRequestMessage )
   at Xamarin.Android.Net.AndroidMessageHandler.SetupSSL(HttpsURLConnection , HttpRequestMessage )
   at Xamarin.Android.Net.AndroidMessageHandler.SetupRequestInternal(HttpRequestMessage , URLConnection )
   at Xamarin.Android.Net.AndroidMessageHandler.SendAsync(HttpRequestMessage , CancellationToken )
   at System.Net.Http.HttpClient.GetStringAsyncCore(HttpRequestMessage , CancellationToken )
   at NUnit.Framework.Internal.AsyncInvocationRegion.AsyncTaskInvocationRegion.WaitForPendingOperationsToComplete(Object invocationResult)
   at NUnit.Framework.Internal.Commands.TestMethodCommand.RunAsyncTestMethod(TestExecutionContext context)
  --- End of managed Java.Lang.ClassNotFoundException stack trace ---
java.lang.ClassNotFoundException: xamarin.android.net.X509TrustManagerWithValidationCallback
	at java.lang.Class.classForName(Native Method)
	at java.lang.Class.forName(Class.java:454)
	at crc643df67da7b13bb6b1.TestInstrumentation_1.n_onStart(Native Method)
	at crc643df67da7b13bb6b1.TestInstrumentation_1.onStart(TestInstrumentation_1.java:35)
	at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2189)
Caused by: java.lang.ClassNotFoundException: xamarin.android.net.X509TrustManagerWithValidationCallback
	... 5 more

--ClassNotFoundException

  --- End of managed Java.Lang.ClassNotFoundException stack trace ---
java.lang.ClassNotFoundException: xamarin.android.net.X509TrustManagerWithValidationCallback
	at java.lang.Class.classForName(Native Method)
	at java.lang.Class.forName(Class.java:454)
	at crc643df67da7b13bb6b1.TestInstrumentation_1.n_onStart(Native Method)
	at crc643df67da7b13bb6b1.TestInstrumentation_1.onStart(TestInstrumentation_1.java:35)
	at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2189)
```

I don't know how make it work. On Discord you said that the classes in Mono.Android are "special cased". I wasn't able to find a way to special case the `X509TrustManagerWithValidationCallback` on my own. Could you please help me finish this PR or point me in the right direction? I believe this is the last obstacle for this PR since @akoeplinger figured out how to get rid fo the excess file size when this feature isn't used.

lmageste:
Hi @simonrozsival and @jonathanpeppers 

I am consuming this change in the latest version of Visual Studio Preview 17.3. Preview 1.1 using Maui (.NET 6).

I can confirm that the test using the badssl website with self-signed certificate works and gets correctly bypassed, thanks for the change!

In fact, every common certificate problem they have in their website works with the validation callback bypassing cert errors, except for the [wrong host example](https://wrong.host.badssl.com/)

This scenario, along with all others, were working before, but this particular one is, unfortunately, a blocker for me. I believe you will be able to repro it on your side as well by attempting to access that URL.

Follows the exception I get after using the validation callback to bypass the certificate errors in the wrong host website from badssl:
System.Net.WebException
  Message=Hostname wrong.host.badssl.com not verified:
    certificate: sha1/1fBSv4IoEephjuel/BGL9TBSEgk=
    DN: CN=*.badssl.com
    subjectAltNames: [*.badssl.com, badssl.com]

(Please let me know if I should file another bug for this particular case)

simonrozsival:
Hi @Imageste! I think you need to extend `AndroidMessageHandler` with your own class and overwrite `GetSSLHostnameVerifier` and return your own implementation of `IHostnameVerifier` which will bypass that validation. Please try it and let me know if that helps or not.

lmageste:
Thanks @simonrozsival!

This works and I can communicate with the server with a wrong host certificate. However, it is not ideal since I need to use a platform-specific API (Android) in order to extend HostnameVerifier and establish communication.

Ideally, since my app targets multi-platforms, a universal solution for my http client library targeting a single framework would be the best (my http client communication library is universal and consumed by all of my platforms), which was how it worked before in the Android platform (when using mono android 8.1 framework) by simply extending the ServerCertificateCustomValidationCallback.

This is, although a bit cumbersome (since my client library needs to target android specifically for this workaround), an acceptable workaround for my case, however.

Thank again!

simonrozsival:
The certificate validation and hostname verification are two independent checks. I'm afraid that in your case you really need both and you'll have to have some platform-specific code in your app. I'm not sure why it wasn't necessary before. Is it possible that you used the managed implementation of the HTTP handler instead of the native one?

Eilon:
I started a discussion topic on how to connect from Android emulators to a local ASP.NET Web API running on Windows: https://github.com/dotnet/maui/discussions/8131

Please check that out and let us know if you have any feedback on any of the solutions presented.

PR #6692, opened 2022-02-01
[main] Add maestro dependencies for dotnet/maui

rmarinho:
Context: https://github.com/dotnet/maui/pull/4424

Maestro bumps in dotnet/maui are doing things like:

    -<Dependency Name="Microsoft.Extensions.Hosting" Version="6.0.2-mauipre.1.22069.5" CoherentParentDependency="Microsoft.AspNetCore.App.Runtime.win-x64">
    +<Dependency Name="Microsoft.Extensions.Hosting" Version="6.0.0" CoherentParentDependency="Microsoft.AspNetCore.App.Runtime.win-x64">

We have a dependency chain such as:

* dotnet/maui -> depends on
* xamarin/xamarin-android -> depends on
* dotnet/runtime

The issue being that `Microsoft.Extensions.Hosting` is not defined in
xamarin-android's `Version.Details.xml`. For Maestro to "know" about
parent dependencies, sometimes you have to list additional
dependencies in repositories that don't use them. In order for
dependencies to use them appropriately.

### DIFF
--- a/eng/Version.Details.xml
+++ b/eng/Version.Details.xml
@@ -12,6 +12,138 @@
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>059caa278b1e5a62d0d60c9d84bdd737b04e6da8</Sha>
     </Dependency>
+     <Dependency Name="Microsoft.Extensions.Primitives" Version="6.0.2-mauipre.1.22074.5">
+      <Uri>https://github.com/dotnet/runtime</Uri>
+      <Sha>63715256cc94963d608ef3d683d2529332c5140c</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.Extensions.Hosting" Version="6.0.2-mauipre.1.22074.5">
+      <Uri>https://github.com/dotnet/runtime</Uri>
+      <Sha>63715256cc94963d608ef3d683d2529332c5140c</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.Extensions.Hosting.Abstractions" Version="6.0.2-mauipre.1.22074.5">
+      <Uri>https://github.com/dotnet/runtime</Uri>
+      <Sha>63715256cc94963d608ef3d683d2529332c5140c</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.Extensions.DependencyInjection.Abstractions" Version="6.0.2-mauipre.1.22074.5">
+      <Uri>https://github.com/dotnet/runtime</Uri>
+      <Sha>63715256cc94963d608ef3d683d2529332c5140c</Sha>
+    </Dependency>
+    <Dependency Name="Microsoft.Extensions.DependencyInjection" Version="6.0.2-mauipre.1.22074.5">
+      <Uri>https://github.com/dotnet/runtime</Uri>
+      <Sha>63715256cc94963d608ef3d683d2529332c5140c</Sha>
+    </Dependency>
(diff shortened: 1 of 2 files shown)


eng/Version.Details.xml
```diff
@@ -12,6 +12,138 @@
       <Uri>https://github.com/dotnet/runtime</Uri>
       <Sha>059caa278b1e5a62d0d60c9d84bdd737b04e6da8</Sha>
     </Dependency>
+     <Dependency Name="Microsoft.Extensions.Primitives" Version="6.0.2-mauipre.1.22074.5">
```
jonathanpeppers, line 15: Before merging this one, we should find out if it needs `CoherentParentDependency`.

pjcollins:
I am wondering if MAUI should stop taking .NET SDK updates through Android at this point, since you have many more dependencies than we do? Would it make more sense to remove [this CoherentParentDependency](https://github.com/dotnet/maui/blob/6945e425da2e2f8a7d0873c98c26c3eb31cf7320/eng/Version.Details.xml#L3) instead and move these changes to the MAUI repo?

eng/Version.Details.xml
jonathanpeppers: This is correct, we shouldn't have `CoherentParentDependency` here.

jonathanpeppers:
> I am wondering if MAUI should stop taking .NET SDK updates through Android at this point, since you have many more dependencies than we do? Would it make more sense to remove this CoherentParentDependency instead and move these changes to the MAUI repo?

@rmarinho do you want to have your own subscription in dotnet/maui for dotnet/runtime?

Main issue is this is a lot of things we don't even use.

rmarinho:
but that way we would have updates from the runtime, and we can have an update of the runtime that doesn't work with android yet no ?

jonathanpeppers:
> but that way we would have updates from the runtime, and we can have an update of the runtime that doesn't work with android yet no ?

We have the same problem with iOS, as you can't put `CoherentParentDependency` for two repos. You might take a xamarin-android bump that would break iOS projects, etc.

We're trying to avoid xamarin-android Maestro bumps
* to go from this (so short!): https://github.com/xamarin/xamarin-android/pull/6669
* to looking like this: https://github.com/dotnet/maui/pull/4444

We have a meeting tomorrow, with all three teams, I'll bring this up.

jonathanpeppers:
Maybe a different question, is it possible to delete all these lines?

https://github.com/dotnet/maui/blob/b32d68d558589d68c97ab40b649f10b080a39f04/eng/Version.Details.xml#L42-L117

Then use `$(MicrosoftNETCoreAppRefPackageVersion)` for everything? I don't see why you need dotnet/runtime versions in here multiple times, but maybe I misunderstand something.

rmarinho:
i think this was a request from Blazor @Eilon  to make sure we test against new stuff and they don't break us

Eilon:
I don't recall _exactly_ what it was for, but it was something relating to ensuring we got coherent dependency versions from what I think was a diamond dependency. We needed to get WinForms+WPF+ASP.NET+.NETRuntime dependencies that all reasonably matched.

Now, given that .NET 6 has shipped, I don't think that matters so much, because all those dependencies I _think_ should be locked to 6.0 RTM (or 6.0.x patch?) and they would thus always be coherent with each other.

Of course, Android/iOS are not RTM so they will rev, but Blazor has no special requirements for those outside of what MAUI needs.

jonathanpeppers:
@rmarinho can we close this? I think we ended up removing these in dotnet/maui.

jonathanpeppers:
Let me know if we need this, and we can reopen, thanks.

Issue #6634, opened 2022-01-11
manifestmerger error message missing


jonathanpeppers:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

.NET 6

### Description

I saw this on the MAUI bump here: https://github.com/dotnet/maui/pull/4033

It got the error:
```
D : \agent\1\s\src\Essentials\samples\Samples\obj\Debug\net6.0-android\AndroidManifest.xml error :  [D:\agent\1\s\src\Essentials\samples\Samples\Essentials.Sample.csproj]
```

But when I look at the `.binlog`:

![image](https://user-images.githubusercontent.com/840039/148962919-d9c3fdab-ba18-4fc2-904c-e77765a4ab1f.png)

So I think the error message is just getting cut off?

### Steps to Reproduce

Something in this project causes the above error:

https://github.com/dotnet/maui/tree/main/src/Essentials/samples/Samples

### Did you find any workaround?

Still looking into it, I might try removing `<uses-sdk/>` to get rid of the error and use .NET MSBuild properties instead.

### Relevant log output

[Microsoft.Maui-net6-Debug-Build.zip](https://github.com/xamarin/xamarin-android/files/7847399/Microsoft.Maui-net6-Debug-Build.zip)

```
ManifestMerger
    Assembly = D:\agent\1\s\bin\dotnet\packs\Microsoft.Android.Sdk.Windows\31.0.200-preview.13.11\tools\Xamarin.Android.Build.Tasks.dll
    CommandLineArguments = C:\Program Files\Microsoft\jdk-11.0.12.7-hotspot\bin\java.exe -cp D:\agent\1\s\bin\dotnet\packs\Microsoft.Android.Sdk.Windows\31.0.200-preview.13.11\tools\manifestmerger.jar com.xamarin.manifestmerger.Main obj\Debug\net6.0-android\android\manifestmerger.rsp 
    Parameters
        ManifestMergerJarPath = D:\agent\1\s\bin\dotnet\packs\Microsoft.Android.Sdk.Windows\31.0.200-preview.13.11\tools\manifestmerger.jar
        ToolPath = C:\Program Files\Microsoft\jdk-11.0.12.7-hotspot\bin
        OutputManifestFile = obj\Debug\net6.0-android\android\AndroidManifest.xml
        AndroidManifest = obj\Debug\net6.0-android\AndroidManifest.xml
        LibraryManifestFiles
[...]

PR #6694, opened 2022-02-01
[Xamarin.Android.Build.Tasks] improve <ManifestMerger/> error messages

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/6634

Adding a `[Service]` with `[IntentFilter]` requires you to specify
`[Service(Exported=true)]` in API 30. Newer versions of
`manifestmerger.jar` enforce this at build time -- instead of
previously you would only hit the error during `adb install`.

Unfortunately, Xamarin.Android displays the error message poorly:

    D : \agent\1\s\src\Essentials\samples\Samples\obj\Debug\net6.0-android\AndroidManifest.xml error :  [D:\agent\1\s\src\Essentials\samples\Samples\Essentials.Sample.csproj]

The error text is literally `D`!

To fix this:

1. Make `JavaToolTask.CodeErrorRegEx` overridable.

2. Provide a `Regex` to match the error messages that
   `manifestmerger.jar` gives for `AndroidManifest.xml` files.

We now get appropriate error messages with line numbers:

    obj\Debug\AndroidManifest.xml(12,5): java.exe error AMM0000: android:exported needs to be explicitly specified for element <service#crc642ecd25190884e03d.TestService>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/JavaToolTask.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/JavaToolTask.cs
@@ -66,8 +66,8 @@ namespace Xamarin.Android.Tasks
 			at com.android.dx.command.Main.main(Main.java:106)
 		*/
 		const string ExceptionRegExString = @"(?<exception>java.lang.+):(?<error>.+)";
-		protected static readonly Regex CodeErrorRegEx = new Regex (CodeErrorRegExString, RegexOptions.Compiled);
-		protected static readonly Regex ExceptionRegEx = new Regex (ExceptionRegExString, RegexOptions.Compiled);
+		static readonly Regex codeErrorRegEx = new Regex (CodeErrorRegExString, RegexOptions.Compiled);
+		static readonly Regex exceptionRegEx = new Regex (ExceptionRegExString, RegexOptions.Compiled);
 		bool foundError = false;
 		List<string> errorLines = new List<string> ();
 		StringBuilder errorText = new StringBuilder ();
@@ -85,6 +85,10 @@ namespace Xamarin.Android.Tasks
 		protected override string ToolName {
 			get { return OS.IsWindows ? "java.exe" : "java"; }
 		}
+
+		protected virtual Regex CodeErrorRegEx => codeErrorRegEx;
+
+		protected virtual Regex ExceptionRegEx => exceptionRegEx;
 
 		protected override bool HandleTaskExecutionErrors ()
 		{
@@ -136,10 +140,8 @@ namespace Xamarin.Android.Tasks
 					errorText.Clear ();
 				}
 				file = match.Groups ["file"].Value;
-				line = int.Parse (match.Groups ["line"].Value) + 1;
 				var error = match.Groups ["error"].Value;
-				column = 0;
-
+				GetLineNumber (match.Groups ["line"].Value, out line, out column);
 				errorText.AppendLine (error);
 				return true;
 			} else if (exceptionMatch.Success) {
@@ -166,6 +168,12 @@ namespace Xamarin.Android.Tasks
 				errorText.AppendLine (singleLine);
 			}
 			return true;
+		}
+
+		protected virtual void GetLineNumber (string match, out int line, out int column)
+		{
+			line = int.Parse (match) + 1;
+			column = 0;
 		}
 
 		protected override void LogEventsFromTextOutput (string singleLine, MessageImportance messageImportance)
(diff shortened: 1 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/JavaToolTask.cs
```diff
@@ -136,10 +140,8 @@ bool ProcessOutput (string singleLine)
 					errorText.Clear ();
 				}
 				file = match.Groups ["file"].Value;
-				line = int.Parse (match.Groups ["line"].Value) + 1;
```
jonathanpeppers, line 139: @dellis1972 I'm a little confused about adding `+ 1` to the line number here, where it gets `- 1` later on:

https://github.com/xamarin/xamarin-android/blob/4a71aa6217dabc75b8616ee72440d85352868118/src/Xamarin.Android.Build.Tasks/Tasks/JavaToolTask.cs#L98

Do you know the history behind that? It also does similar with `column`.

Is it something about BOMs?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/ManifestTest.cs
```diff
@@ -898,5 +898,32 @@ class TestService : Service { public override Android.OS.IBinder OnBind (Android
...
+ using Android.Views;
+ [Service]
+[IntentFilter(new[] {{ ""{proj.PackageName}.PLAY"" }})]
+ class TestService : Service {{ public override Android.OS.IBinder OnBind (Android.Content.Intent intent) {{ return null; }} }}"
+			});
+
+			using var b = CreateDllBuilder ();
+			b.ThrowOnBuildFailure = false;
+			Assert.IsFalse (b.Build (proj), "Build should have failed");
+			Assert.IsTrue (b.LastBuildOutput.ContainsText ("AndroidManifest.xml(12,5): java.exe error AMM0000:"), "Should recieve AMM0000 error");
```
jonathanpeppers, line 925: Whoops, on mac it doesn't say `java.exe` just `java`

src/Xamarin.Android.Build.Tasks/Tasks/JavaToolTask.cs
dellis1972: If I remember rightly, it was to do with how it appears in the IDE. Since we do change the code to strip out BOMs etc when we copy the files to the directory where we are going to compile thngs, its probably related to that.

src/Xamarin.Android.Build.Tasks/Tasks/ManifestMerger.cs
```diff
@@ -32,9 +31,39 @@ public class ManifestMerger : JavaToolTask
 
 		public string ExtraArgs { get; set; }
 
+		/*
+		 * obj\Debug\AndroidManifest.xml:12:5-16:15 Error:
+		 * 	android:exported needs to be explicitly specified for element <service#crc642ecd25190884e03d.TestService>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.
+		 */
+		static readonly Regex manifestErrorRegEx = new Regex (@"(?<file>.+AndroidManifest\.xml):(?<line>\d+:\d+).+Error:(?<error>.+)?", RegexOptions.Compiled | RegexOptions.IgnoreCase);
```
dellis1972, line 38: will this also pick up standard Java errors that might be raised when running manifest merger?
There is a chance we could get both types of error from the tooling.

jonathanpeppers: Reviewing this again, this replaces this Regex:
https://github.com/xamarin/xamarin-android/blob/5f0a7da3d22c0968b39c73681ec0f459ebf3e311/src/Xamarin.Android.Build.Tasks/Tasks/JavaToolTask.cs#L69

Which (I think) shows errors in `*.java` files. We replace with the regex for `AndroidManifest.xml`.

This regex will also get used:
https://github.com/xamarin/xamarin-android/blob/5f0a7da3d22c0968b39c73681ec0f459ebf3e311/src/Xamarin.Android.Build.Tasks/Tasks/JavaToolTask.cs#L70

So we should still get messages if `manifestmerger` throws an Java exception, but I'm not sure how we test that?

dellis1972 (approved):

PR #6706, opened 2022-02-04
[build] Update Arcade dependencies

pjcollins:
Bump Microsoft.DotNet.ApiCompat and Microsoft.DotNet.Build.Tasks.Feed to
the latest package version available in the ".NET Eng - Latest" channel.

The ApiCompat package appears to have some behavioral changes, and
tests/api-compatibility has been updated accordingly.


jpobst (commented):
Looks like the changes to API definitions are largely fine:

- Removing `[Obsolete]` from a member is no longer reported as a break.
- It looks like now it looking at attributes applied to method parameters and it wasn't before.

The only ones I don't really understand is removing a few of these in `acceptable-breakages-v12.0.txt`:
```
CannotChangeAttribute : Attribute 'Android.Runtime.RequiresPermissionAttribute' on 'Android.Telephony
.PhoneStateListenerFlags Android.Telephony.PhoneStateListener.ListenCallDisconnectCauses' changed from
'[RequiresPermissionAttribute("android.Manifest.permission.READ_PRECISE_PHONE_STATE")]' in the contract
to '[RequiresPermissionAttribute("android.permission.READ_PRECISE_PHONE_STATE")]' in the implementation.
```

However it kept several identical ones on other members, so I'm inclined to believe that it is still working correctly.

As an aside for another day, I wonder if we should stop comparing the `Android.Runtime.StringDefAttribute` and `Android.Runtime.IntDefAttribute` attributes.  They seem to churn a lot and I don't know that we actually care about their values for API compatibility.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -165,7 +165,7 @@ void ValidateApiCompat (string contractPath, bool validateAgainstReference)
 				using (var genApiProcess = new Process ()) {
 
 					if (Environment.Version.Major >= 5) {
-						var apiCompat = new FileInfo (Path.Combine (ApiCompatPath, "..", "netcoreapp2.1", "Microsoft.DotNet.ApiCompat.dll"));
+						var apiCompat = new FileInfo (Path.Combine (ApiCompatPath, "..", "netcoreapp3.1", "Microsoft.DotNet.ApiCompat.dll"));
```
jonathanpeppers, line 168: I got xamarin-android to build on M1 the other day (first time!).

This is one of the apps that fails, maybe the new version will work?

jonathanpeppers (commented):
Overall, this looks good to me 👍 , but I don't know if I'm qualified to review the API diffs.

PR #6707, opened 2022-02-04
[Mono.Android] use Unsafe.As<T>() over Delegate.CreateDelegate()

jonathanpeppers:
In porting System.Reflection.Emit code in 32cff438, I was considering
our use of `Delegate.CreateDelegate()`. This appears to take up
~2.83ms, when using `dotnet trace` to profile. Can we just cast
instead?

"generator" from xamarin/java.interop generates `internal` delegates
for each Java binding assembly:

    _JniMarshal_PP_V

A direct cast wouldn't work, if for example,
`Xamarin.AndroidX.AppCompat.dll` had its own `_JniMarshal_PP_V`, we
can't cast to the one in `Mono.Android.dll`.

To solve this problem we can use `Unsafe.As<T>()` instead:

https://docs.microsoft.com/dotnet/api/system.runtime.compilerservices.unsafe.as?view=net-6.0

Testing a `dotnet new maui` app on a Pixel 5, a `Release` build
without AOT:

    Before:
    02-04 16:03:18.603  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s64ms
    02-04 16:03:20.789  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s37ms
    02-04 16:03:23.038  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s65ms
    02-04 16:03:25.233  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s58ms
    02-04 16:03:27.434  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s36ms
    02-04 16:03:29.622  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s43ms
    02-04 16:03:31.807  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s48ms
    02-04 16:03:33.998  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s57ms
    02-04 16:03:36.146  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s35ms
    02-04 16:03:38.357  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s44ms
    Average(ms): 1048.7
    Std Err(ms): 3.64554522671164
    Std Dev(ms): 11.5282262295637
    After:
    02-04 22:23:22.592  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s26ms
    02-04 22:23:24.839  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s25ms
    02-04 22:23:27.079  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s58ms
    02-04 22:23:29.302  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s57ms
    02-04 22:23:31.539  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s76ms
    02-04 22:23:33.732  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s48ms
    02-04 22:23:35.956  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s55ms
    02-04 22:23:38.136  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s34ms
    02-04 22:23:40.360  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s37ms
    02-04 22:23:42.619  1867  2226 I ActivityTaskManager: Displayed com.companyname.foo/crc64808a40cc7e533249.MainActivity: +1s50ms
    Average(ms): 1046.6
    Std Err(ms): 5.08636521605666
    Std Dev(ms): 16.0844990941935

It did seem to save ~2.1ms in this case?

### DIFF
--- a/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs
+++ b/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 
 namespace Android.Runtime
 {
@@ -18,7 +19,7 @@ namespace Android.Runtime
 		{
 			switch (delegateType.Name) {
 				case nameof (_JniMarshal_PP_V): {
-					_JniMarshal_PP_V callback = (_JniMarshal_PP_V) Delegate.CreateDelegate (typeof (_JniMarshal_PP_V), dlg.Target, dlg.Method);
+					_JniMarshal_PP_V callback = Unsafe.As<_JniMarshal_PP_V> (dlg);
 					_JniMarshal_PP_V result = (jnienv, klazz) => {
 						JNIEnv.WaitForBridgeProcessing ();
 						try {
@@ -31,7 +32,7 @@ namespace Android.Runtime
 					return result;
 				}
 				case nameof (_JniMarshal_PPI_V): {
-					_JniMarshal_PPI_V callback = (_JniMarshal_PPI_V) Delegate.CreateDelegate (typeof (_JniMarshal_PPI_V), dlg.Target, dlg.Method);
+					_JniMarshal_PPI_V callback = Unsafe.As<_JniMarshal_PPI_V> (dlg);
 					_JniMarshal_PPI_V result = (jnienv, klazz, p0) => {
 						JNIEnv.WaitForBridgeProcessing ();
 						try {
@@ -44,7 +45,7 @@ namespace Android.Runtime
 					return result;
 				}
 				case nameof (_JniMarshal_PPL_L): {
-					_JniMarshal_PPL_L callback = (_JniMarshal_PPL_L) Delegate.CreateDelegate (typeof (_JniMarshal_PPL_L), dlg.Target, dlg.Method);
+					_JniMarshal_PPL_L callback = Unsafe.As<_JniMarshal_PPL_L> (dlg);
 					_JniMarshal_PPL_L result = (jnienv, klazz, p0) => {
 						JNIEnv.WaitForBridgeProcessing ();
 						try {
@@ -57,7 +58,7 @@ namespace Android.Runtime
 					return result;
 				}
 				case nameof (_JniMarshal_PPL_V): {
-					_JniMarshal_PPL_V callback = (_JniMarshal_PPL_V) Delegate.CreateDelegate (typeof (_JniMarshal_PPL_V), dlg.Target, dlg.Method);
+					_JniMarshal_PPL_V callback = Unsafe.As<_JniMarshal_PPL_V> (dlg);
 					_JniMarshal_PPL_V result = (jnienv, klazz, p0) => {
 						JNIEnv.WaitForBridgeProcessing ();
 						try {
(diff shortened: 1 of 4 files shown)


eerhardt (approved):
If this passes CI, then it LGTM!

eerhardt:
It probably doesn't make a difference, but does the Ref.Emit fallback code still call `Delegate.CreateDelegate`? Should that change to a cast?

jonathanpeppers:
The SRE case, doesn't use `CreateDelegate()`, because it emits a dynamic method:

https://github.com/xamarin/xamarin-android/blob/16c0593ab25262b920a4688c9ca7fd2e9500a37e/src/Mono.Android/Android.Runtime/JNINativeWrapper.cs#L99

So, we just did something silly when porting the strongly-typed version.

src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs
```diff
@@ -18,7 +18,7 @@ private static Delegate CreateBuiltInDelegate (Delegate dlg, Type delegateType)
 		{
 			switch (delegateType.Name) {
 				case nameof (_JniMarshal_PP_V): {
-					_JniMarshal_PP_V callback = (_JniMarshal_PP_V) Delegate.CreateDelegate (typeof (_JniMarshal_PP_V), dlg.Target, dlg.Method);
+					_JniMarshal_PP_V callback = (_JniMarshal_PP_V) dlg;
```
jonathanpeppers, line 21: Some tests hit:
```
System.InvalidCastException : Specified cast is not valid.
```
Thinking about this... I think `_JniMarshal_PP_V` is actually private to each assembly, and so this might not actually work?

If you have a class library that binds a Java library for C#, it gets its own `internal` version of these delegates. So if this code is in `Mono.Android.dll`, it wouldn't be able to cast as-is.

jonathanpeppers: Will check if this works:
```csharp
var callback = Unsafe.As<_JniMarshal_PP_V> (dlg);
```
The signature should match, seems like it might work?

eerhardt: Another thought would be to do an `is _JniMarshal_PP_V` check, and if it isn't, then call `Delegate.CreateDelegate(typeof(_JniMarshal_PP_V))`.

jonathanpeppers:
I think this is overall working.

This adds a new assembly to `dotnet new android`: `1,184 assemblies/System.Runtime.CompilerServices.Unsafe.dll`

That seems like it is OK, we'll probably be using types in there for performance reasons anyway.

marek-safar:
> That seems like it is OK, we'll probably be using types in there for performance reasons anyway.

There is an effort for net7 to make this assembly be facade only so the file should go away.

dellis1972 (approved):

PR #6708, opened 2022-02-05
Replace string.Split in RegisterNativeMembers and use Spans instead

eerhardt:
The string.Split usage in this method was showing up in a profile of Maui startup on Android. Optimizing the method to reduce allocations.

cc @jonathanpeppers

### DIFF
--- a/src/Mono.Android/Android.Runtime/AndroidRuntime.cs
+++ b/src/Mono.Android/Android.Runtime/AndroidRuntime.cs
@@ -388,50 +388,90 @@ namespace Android.Runtime {
 					return;
 				}
 
-				string[] members = methods!.Split ('\n');
-				if (members.Length < 2) {
+				int newLineIndex = methods!.IndexOf ('\n');
+				if (newLineIndex == -1) {
 					if (jniAddNativeMethodRegistrationAttributePresent)
 						base.RegisterNativeMembers (nativeClass, type, methods);
 					return;
 				}
 
-				JniNativeMethodRegistration[] natives = new JniNativeMethodRegistration [members.Length-1];
-				for (int i = 0; i < members.Length; ++i) {
-					string method = members [i];
-					if (string.IsNullOrEmpty (method))
-						continue;
-					string[] toks = members [i].Split (new[]{':'}, 4);
-					Delegate callback;
-					if (toks [2] == "__export__") {
-						var mname = toks [0].Substring (2);
-						MethodInfo? minfo = null;
-						foreach (var mi in type.GetMethods (BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static))
-							if (mi.Name == mname && JavaNativeTypeManager.GetJniSignature (mi) == toks [1]) {
-								minfo = mi;
-								break;
-							}
+				List<JniNativeMethodRegistration> natives = new List<JniNativeMethodRegistration> (16); // an initial capacity since most classes have more than 4-8 methods
+				ReadOnlySpan<char> methodsSpan = methods;
 
-						if (minfo == null)
-							throw new InvalidOperationException (String.Format ("Specified managed method '{0}' was not found. Signature: {1}", mname, toks [1]));
-						callback = CreateDynamicCallback (minfo);
-					} else {
-						Type callbackDeclaringType = type;
-						if (toks.Length == 4) {
-							callbackDeclaringType = Type.GetType (toks [3], throwOnError: true)!;
+				MethodInfo []? typeMethods = null;
+				while (!methodsSpan.IsEmpty) {
+					ReadOnlySpan<char> methodLine = methodsSpan.Slice (0, newLineIndex != -1 ? newLineIndex : methodsSpan.Length);
+					if (!methodLine.IsEmpty) {
+						SplitMethodLine (methodLine,
+							out ReadOnlySpan<char> name,
+							out ReadOnlySpan<char> signature,
+							out ReadOnlySpan<char> callbackString,
+							out ReadOnlySpan<char> callbackDeclaringTypeString);
+						Delegate callback;
+						if (callbackString == "__export__") {
+							var mname = name.Slice (2);
+							MethodInfo? minfo = null;
+							typeMethods ??= type.GetMethods (BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static);
+							foreach (var mi in typeMethods)
+								if (mi.Name == mname && JavaNativeTypeManager.GetJniSignature (mi) == signature) {
+									minfo = mi;
+									break;
+								}
+
+							if (minfo == null)
+							    throw new InvalidOperationException (String.Format ("Specified managed method '{0}' was not found. Signature: {1}", mname.ToString (), signature.ToString ()));
+							callback = CreateDynamicCallback (minfo);
+						} else {
+							Type callbackDeclaringType = type;
+							if (!callbackDeclaringTypeString.IsEmpty) {
+								callbackDeclaringType = Type.GetType (callbackDeclaringTypeString.ToString (), throwOnError: true)!;
+							}
+							while (callbackDeclaringType.ContainsGenericParameters) {
+								callbackDeclaringType = callbackDeclaringType.BaseType!;
+							}
+							GetCallbackHandler connector = (GetCallbackHandler) Delegate.CreateDelegate (typeof (GetCallbackHandler),
+								callbackDeclaringType, callbackString.ToString ());
+							callback = connector ();
 						}
-						while (callbackDeclaringType.ContainsGenericParameters) {
-							callbackDeclaringType = callbackDeclaringType.BaseType!;
-						}
-						GetCallbackHandler connector = (GetCallbackHandler) Delegate.CreateDelegate (typeof (GetCallbackHandler),
-							callbackDeclaringType, toks [2]);
-						callback = connector ();
+						natives.Add (new JniNativeMethodRegistration (name.ToString (), signature.ToString (), callback));
 					}
-					natives [i] = new JniNativeMethodRegistration (toks [0], toks [1], callback);
+
+					if (newLineIndex == -1) {
+						methodsSpan = default;
+					} else {
+						methodsSpan = methodsSpan.Slice (newLineIndex + 1);
+						newLineIndex = methodsSpan.IndexOf ('\n');
+					}
 				}
 
-				JniEnvironment.Types.RegisterNatives (nativeClass.PeerReference, natives, natives.Length);
+				JniEnvironment.Types.RegisterNatives (nativeClass.PeerReference, natives.ToArray());
 			} catch (Exception e) {
 				JniEnvironment.Runtime.RaisePendingException (e);
+			}
+		}
+
+		private static void SplitMethodLine(
+			ReadOnlySpan<char> methodLine,
+			out ReadOnlySpan<char> name,
+			out ReadOnlySpan<char> signature,
+			out ReadOnlySpan<char> callback,
+			out ReadOnlySpan<char> callbackDeclaringType)
+		{
+			int colonIndex = methodLine.IndexOf(':');
+			name = methodLine.Slice(0, colonIndex);
+			methodLine = methodLine.Slice(colonIndex + 1);
+	
+			colonIndex = methodLine.IndexOf(':');
+			signature = methodLine.Slice(0, colonIndex);
+			methodLine = methodLine.Slice(colonIndex + 1);
+
+			colonIndex = methodLine.IndexOf(':');
+			callback = methodLine.Slice(0, colonIndex != -1 ? colonIndex : methodLine.Length);
+
+			if (colonIndex == -1) {
+				callbackDeclaringType = default;
+			} else {
+				callbackDeclaringType = methodLine.Slice (colonIndex + 1);
 			}
 		}
 	}


jonathanpeppers:
Next week, let's see if we can get your the right access, hopefully you can see some of the test failures like:
```
Mono.Android_Tests, Java.InteropTests.JnienvTest.CreateTypeWithExportedMethods 

Error message
Java.Lang.NoClassDefFoundError : crc64d83ea27a2cea9f79.ContainsExportedMethods
----> System.ArgumentException : Couldn't bind to method 'export'.

Stack trace
  at Java.Interop.JniEnvironment+Object.AllocObject (Java.Interop.JniObjectReference type) [0x0003c] in <bd66b127723142bcb4bc1da645083ea5>:0 
  at Java.Interop.JniType.AllocObject () [0x0000c] in <bd66b127723142bcb4bc1da645083ea5>:0 
  at Java.Interop.JniPeerMembers+JniInstanceMethods.StartCreateInstance (System.String constructorSignature, System.Type declaringType, Java.Interop.JniArgumentValue* parameters) [0x00044] in <bd66b127723142bcb4bc1da645083ea5>:0 
  at Java.Lang.Object..ctor () [0x00034] in <9bb9146e9c344bf29a3625889c068ffa>:0 
  at Java.InteropTests.ContainsExportedMethods..ctor () [0x00000] in <ef9f0845156945dda6edd8d0444353e9>:0 
  at Java.InteropTests.JnienvTest.CreateTypeWithExportedMethods () [0x00000] in <ef9f0845156945dda6edd8d0444353e9>:0 
  at System.Reflection.RuntimeMethodInfo.Invoke (System.Object obj, System.Reflection.BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) [0x0006a] in <764e62ab5c2e4b278350e7eaa396ac9b>:0 
  --- End of managed Java.Lang.NoClassDefFoundError stack trace ---
java.lang.NoClassDefFoundError: crc64d83ea27a2cea9f79.ContainsExportedMethods
	at crc64306a4245d1cd4a3b.TestInstrumentation_1.n_onStart(Native Method)
	at crc64306a4245d1cd4a3b.TestInstrumentation_1.onStart(TestInstrumentation_1.java:35)
	at android.app.Instrumentation$InstrumentationThread.run(Instrumentation.java:2189)
Caused by: android.runtime.JavaProxyThrowable: System.ArgumentException: Couldn't bind to method '__export__'.
  at System.Delegate.GetCandidateMethod (System.Type type, System.Type target, System.String method, System.Reflection.BindingFlags bflags, System.Boolean ignoreCase, System.Boolean throwOnBindFailure) [0x000f9] in <764e62ab5c2e4b278350e7eaa396ac9b>:0 
  at System.Delegate.CreateDelegate (System.Type type, System.Type target, System.String method, System.Boolean ignoreCase, System.Boolean throwOnBindFailure) [0x00014] in <764e62ab5c2e4b278350e7eaa396ac9b>:0 
  at System.Delegate.CreateDelegate (System.Type type, System.Type target, System.String method) [0x00000] in <764e62ab5c2e4b278350e7eaa396ac9b>:0 
  at Android.Runtime.AndroidTypeManager.RegisterNativeMembers (Java.Interop.JniType nativeClass, System.Type type, System.String methods) [0x001a2] in <9bb9146e9c344bf29a3625889c068ffa>:0 
	at mono.android.Runtime.register(Native Method)
	at crc64d83ea27a2cea9f79.ContainsExportedMethods.<clinit>(ContainsExportedMethods.java:15)
	at java.lang.Class.classForName(Native Method)
	at java.lang.Class.forName(Class.java:454)
	... 3 more

--ArgumentException
  at System.Delegate.GetCandidateMethod (System.Type type, System.Type target, System.String method, System.Reflection.BindingFlags bflags, System.Boolean ignoreCase, System.Boolean throwOnBindFailure) [0x000f9] in <764e62ab5c2e4b278350e7eaa396ac9b>:0 
  at System.Delegate.CreateDelegate (System.Type type, System.Type target, System.String method, System.Boolean ignoreCase, System.Boolean throwOnBindFailure) [0x00014] in <764e62ab5c2e4b278350e7eaa396ac9b>:0 
  at System.Delegate.CreateDelegate (System.Type type, System.Type target, System.String method) [0x00000] in <764e62ab5c2e4b278350e7eaa396ac9b>:0 
  at Android.Runtime.AndroidTypeManager.RegisterNativeMembers (Java.Interop.JniType nativeClass, System.Type type, System.String methods) [0x001a2] in <9bb9146e9c344bf29a3625889c068ffa>:0 

  at Java.Interop.JniEnvironment+StaticMethods.CallStaticObjectMethod (Java.Interop.JniObjectReference type, Java.Interop.JniMethodInfo method, Java.Interop.JniArgumentValue* args) [0x0006e] in <bd66b127723142bcb4bc1da645083ea5>:0 
  at Android.Runtime.JNIEnv.CallStaticObjectMethod (System.IntPtr jclass, System.IntPtr jmethod, Android.Runtime.JValue* parms) [0x
```
(end partially cut off)

jonathanpeppers (approved):
This looks like the test passed, I just fixed some formatting.

It also seems like classic Xamarin.Android still works -- can use `Span<T>` just fine, and no `.apk` size regressions over our threshold.

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -388,52 +388,92 @@ public override void RegisterNativeMembers (JniType nativeClass, Type type, stri
...
-					string[] toks = members [i].Split (new[]{':'}, 4);
-					Delegate callback;
-					if (toks [2] == "__export__") {
-						var mname = toks [0].Substring (2);
-						MethodInfo? minfo = null;
-						foreach (var mi in type.GetMethods (BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static))
-							if (mi.Name == mname && JavaNativeTypeManager.GetJniSignature (mi) == toks [1]) {
-								minfo = mi;
-								break;
+				List<JniNativeMethodRegistration> natives = new List<JniNativeMethodRegistration> (16); // an initial capacity since most classes have more than 4-8 methods
```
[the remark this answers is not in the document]
eerhardt: If you don't set a capacity, it starts at 4 and doubles every time it needs to grow. Which means for classes that have 9-16 methods, you allocate 2 "throw away" arrays and pay for copying twice.

I guess it all depends on what the "average" class contains.

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -388,52 +388,92 @@ public override void RegisterNativeMembers (JniType nativeClass, Type type, stri
...
+				ReadOnlySpan<char> methodsSpan = methods;
```
jonpryor, line 399: If we're going for `ReadOnlySpan<char>`, we could probably further improve things by:

 1. Update `JNIEnv.RegisterJniNatives()` to instead create the `ReadOnlySpan<char>`: https://github.com/xamarin/xamarin-android/blob/85bfb77dc4f405026ab8b1fd08f3cd30deac2249/src/Mono.Android/Android.Runtime/JNIEnv.cs#L139
    This would avoid a `string` creation.

    ```csharp
    ReadOnlySpan<char> methods = new ReadOnlySpan<char>((void*) methods_ptr, methods_len);
    androidRuntime!.TypeManager.RegisterNativeMembers (jniType, type, methods);
    ```

 2. Overload this `RegisterNativeMembers()` method to take `ReadOnlySpan<char>`, and have the override invoke the overload:

    ```csharp
    public override void RegisterNativeMembers (JniType nativeClass, Type type, string? methods) =>
        RegisterNativeMembers (nativeClass, type, new ReadOnlySpan<char>(methods));
    public void RegisterNativeMembers (JniType nativeClass, Type type, ReadOnlySpan<char> methods)
    {
        …
    }
    ```

This would introduce complications around `FastRegisterNativeMembers()`, but those can be handled.

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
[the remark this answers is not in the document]
jonpryor: Therein lies a good question: what *is* average?

However, I'm not sure we need to answer that.  This is all in-memory; accessing the Span should be fairly quick.  Which is to say, why can't we iterate over it twice?

Count the number of `\n`s, and that's how many entries you need.

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -388,52 +388,92 @@ public override void RegisterNativeMembers (JniType nativeClass, Type type, stri
...
+				while (!methodsSpan.IsEmpty) {
+					ReadOnlySpan<char> methodLine = methodsSpan.Slice (0, newLineIndex != -1 ? newLineIndex : methodsSpan.Length);
+					if (!methodLine.IsEmpty) {
+						SplitMethodLine (methodLine,
+							out ReadOnlySpan<char> name,
+							out ReadOnlySpan<char> signature,
+							out ReadOnlySpan<char> callbackString,
+							out ReadOnlySpan<char> callbackDeclaringTypeString);
+						Delegate callback;
+						if (callbackString.SequenceEqual ("__export__")) {
```
jonpryor, line 411: What's the perf impact of using `.SequenceEqual()` here, particularly in terms of going through interfaces instead of concrete types (interface method dispatch is nominally slower)?

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -388,52 +388,92 @@ public override void RegisterNativeMembers (JniType nativeClass, Type type, stri
...
+							var mname = name.Slice (2);
+							MethodInfo? minfo = null;
+							typeMethods ??= type.GetMethods (BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.Static);
```
jonpryor, line 414: This shouldn't need to look at Public or Instance methods.

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -388,52 +388,92 @@ public override void RegisterNativeMembers (JniType nativeClass, Type type, stri
...
+							foreach (var mi in typeMethods)
+								if (mname.SequenceEqual(mi.Name) && signature.SequenceEqual(JavaNativeTypeManager.GetJniSignature (mi))) {
```
jonpryor, line 416: Please don't add `JavaNativeTypeManager.GetJniSignature(mi)` to this codepath.  It's already reflection heavy, and this would only make it *worse*:

https://github.com/xamarin/java.interop/blob/f91b0770af630cadc270290e7d6e79637ba3f11c/src/Java.Interop.Tools.TypeNameMappings/Java.Interop.Tools.TypeNameMappings/JavaNativeTypeManager.cs#L363-L372

Only the name is required.

For further rationale, build e.g. `samples/HelloWorld/HelloWorld/HelloWorld.csproj`, and then view `samples/HelloWorld//HelloWorld/obj/Debug/android/src/example/MainActivity.java`:

```java
		__md_methods = 
			"n_onCreate:(Landroid/os/Bundle;)V:GetOnCreate_Landroid_os_Bundle_Handler\n" +
			"";
		mono.android.Runtime.register ("HelloWorld.MainActivity, HelloWorld", MainActivity.class, __md_methods);
```

The method this will lookup is `GetOnCreate_Landroid_os_Bundle_Handler`.  That's *very* unique, and not likely to cause conflicts.

*Furthermore*, the types are wrong; `signature` would be `(Landroid/os/Bundle;)V`, while `GetOnCreate_Landroid_os_Bundle_Handler` (generated in e.g. `src/Mono.Android/obj/Debug/monoandroid10/android-32/mcw/Android.App.Activity.cs`) is:

```csharp
		{
		}
```

which would imply a "signature" of `()V` (no parameters, returns `void`), which *will rarely match*.

jonpryor: Contextual related: https://github.com/xamarin/xamarin-android/wiki/Blueprint#java-type-registration

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
eerhardt: These are the existing binding flags being used. I'm not inclined to changing that with this PR. I believe something like that should be a separate change.

The only thing that I changed here is to cache the call to `type.GetMethods` into a local variable, since the `type` never changes. Calling `type.GetMethods` each time in the loop doesn't make sense.

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
eerhardt: The perf impact is identical to `==` on strings (albiet with a few special checks in `==` missing):

https://github.com/dotnet/runtime/blob/3c43ef6af7dd0a54e26ae23e4610f4937a7a7e4d/src/libraries/System.Private.CoreLib/src/System/String.Comparison.cs#L684-L709

Calls "EqualsHelper", which in turn just calls `SpanHelpers.SequenceEqual`:

https://github.com/dotnet/runtime/blob/3c43ef6af7dd0a54e26ae23e4610f4937a7a7e4d/src/libraries/System.Private.CoreLib/src/System/String.Comparison.cs#L24-L34

Where the SequenceEqual method being called here is:

https://github.com/dotnet/runtime/blob/3c43ef6af7dd0a54e26ae23e4610f4937a7a7e4d/src/libraries/System.Private.CoreLib/src/System/MemoryExtensions.cs#L456-L463

(note `IsBitwiseEquatable` is `true` for `char`).

There are no interfaces at work here. It is an extension method directly on `ReadOnlySpan<T>`.

jonpryor: That's…wow: https://github.com/dotnet/runtime/blob/3c43ef6af7dd0a54e26ae23e4610f4937a7a7e4d/src/libraries/System.Private.CoreLib/src/System/SpanHelpers.T.cs#L1152-L1241

eerhardt: That's actually not the path it will go down for strings or `ReadOnlySpan<char>`. It will go down this path (taking the SIMD approach), since `IsBitwiseEquatable` is `true` for `char`:

https://github.com/dotnet/runtime/blob/3c43ef6af7dd0a54e26ae23e4610f4937a7a7e4d/src/libraries/System.Private.CoreLib/src/System/SpanHelpers.Byte.cs#L1699-L1909

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64XFormsDotNet.apkdesc
```diff
@@ -5,22 +5,22 @@
...
-      "Size": 61695
+      "Size": 61902
     },
     "assemblies/Microsoft.Win32.Primitives.dll": {
-      "Size": 3834
+      "Size": 3833
     },
     "assemblies/Mono.Android.dll": {
-      "Size": 437730
+      "Size": 442644
```
jonathanpeppers, line 17: I think maybe we were near the limit here, and this change pushed it over the edge?

Main is currently passing, and it looks like we don't keep the build artifact of this file for passing tests.

jonpryor:
Draft commit message:

```markdown
[Mono.Android] Use Span<char> instead of string.Split() (#6708)

Profiling showed that `System.String.Split()` had a noticeable impact
on on `AndroidRuntime.RegisterNativeMembers()` performance.

TODO: what is the nature of the impact?  How would one characterize this?

Replace `string.Split()` usage with `Span<char>` usage.  This reduces
memory pressure, and improves performance of a startup codepath.

TODO: What is the nature of the improvement?  What is improved, and
by how much?

```

jonpryor:
@eerhardt: is there any way to fill in the "TODO"s in the [draft commit message](https://github.com/xamarin/xamarin-android/pull/6708#issuecomment-1034445290)?  How much does this improve things?

PR #6711, opened 2022-02-07
[xaprepare] Add Disk information to output

dellis1972:


dellis1972:
@grendello not sure if this is the best approach. Should we list all the disks? or just the disk we are working on?. 
Should it all be in its own section rather than in the operating system one?

grendello:
I think OS section is just fine. I would however limit the drives to the working drive plus `/tmp` (or wherever the temporary files are placed, the `$TMP` and `$TMPDIR` envvars might point to something else than `/tmp`), as some systems may have plenty of virtual filesystems mounted, which would only increase the signal to noise ratio.

build-tools/xaprepare/xaprepare/Application/Utilities.cs
```diff
@@ -446,6 +446,12 @@ public static void FormatSize (ulong dSize, out decimal value, out string unit)
 			value = SignificantDigits (value, 3);
 		}
 
+		public static string SizeToString (long dSize)
```
dellis1972, line 449: I used `long` here because the `DriveInfo` type uses `long` for its properties. I would rather cast to `ulong` in this method rather than at the call site.

dellis1972:
@grendello this should be good to go now.

dellis1972:
Failure does not look related 

```
error : java.lang.IllegalStateException :  Error retrieving device SDK version. Please try again.
```

grendello (approved):

PR #6724, opened 2022-02-09
Remove InvariantGlobalization workaround

eerhardt:
This is no longer needed since https://github.com/dotnet/runtime/pull/53453 fixed the issue this was working around.

Bringing back the original optimization in https://github.com/xamarin/xamarin-android/pull/5605.


jonathanpeppers (approved):
Yeah looks like da536fcf1b663fd71adee40c06466547a04cc6ed #5954 was just working around a bug, and we never followed up.

Will just need to check the CI results.

PR #6727, opened 2022-02-10
Default VerifyDependencyInjectionOpenGenericServiceTrimmability

eerhardt:
This feature switch ensures that DynamicallyAccessedMembers are applied correctly to open generic types used in Dependency Injection.

In the base SDK, this switch is enabled when PublishTrimmed=true. However, Android apps don't set PublishTrimmed=true in Debug builds. So devs are missing out on this validation.

Conversely, in published apps, we don't want to pay the cost of doing this validation. It was showing up in startup profiles. So turning the feature switch off in Release builds.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
@@ -98,6 +98,9 @@
     <_AggressiveAttributeTrimming Condition="'$(_AggressiveAttributeTrimming)' == ''">true</_AggressiveAttributeTrimming>		
     <NullabilityInfoContextSupport Condition="'$(NullabilityInfoContextSupport)' == ''">false</NullabilityInfoContextSupport>
     <BuiltInComInteropSupport Condition="'$(BuiltInComInteropSupport)' == ''">false</BuiltInComInteropSupport>    
+    <!-- Verify DI trimmability at development-time, but turn the validation off for Release builds. -->
+		<VerifyDependencyInjectionOpenGenericServiceTrimmability Condition="'$(VerifyDependencyInjectionOpenGenericServiceTrimmability)' == '' and '$(Configuration)' == 'Release'">false</VerifyDependencyInjectionOpenGenericServiceTrimmability>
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -98,6 +98,9 @@
     <_AggressiveAttributeTrimming Condition="'$(_AggressiveAttributeTrimming)' == ''">true</_AggressiveAttributeTrimming>		
     <NullabilityInfoContextSupport Condition="'$(NullabilityInfoContextSupport)' == ''">false</NullabilityInfoContextSupport>
     <BuiltInComInteropSupport Condition="'$(BuiltInComInteropSupport)' == ''">false</BuiltInComInteropSupport>    
+    <!-- Verify DI trimmability at development-time, but turn the validation off for Release builds. -->
+		<VerifyDependencyInjectionOpenGenericServiceTrimmability Condition="'$(VerifyDependencyInjectionOpenGenericServiceTrimmability)' == '' and '$(Configuration)' == 'Release'">false</VerifyDependencyInjectionOpenGenericServiceTrimmability>
```
jonathanpeppers, line 102: Should this maybe check if `$(PublishTrimmed)` is true instead of `$(Configuration)`?

A lot of times developer might make a new configuration named `AppStore` or something, and they would configure it like a `Release` build.

eerhardt: Yeah, that's probably a better property to switch on. Unless there is a better property.

eerhardt:
I've responded to the current feedback. PTAL.

jonathanpeppers (approved):
This looks ok, we can link to this in the final commit message:
* https://github.com/dotnet/runtime/blob/main/docs/workflow/trimming/feature-switches.md

It doesn't look like this one is on MS docs anywhere.

Still waiting to see if CI is green.

dellis1972:
> How do you normally handle the final commit message? For other repos I typically "Squash and merge" in the GH UI and adjust the commit message then. Is that what is normally done here?

Yes. But it's normally best to update the PR text to have the final commit message (or post the final commit message as a comment) so it can be reviewed. 
Plus the more detail in the commit message the better :)

PR #6728, opened 2022-02-10
[ci] Swap Maestro and artifact status ordering

pjcollins:
We've seen some failures in recent attempts to push package info to the
Maestro Build Asset Registry.  Fix this by using the correct OutputPath
when running `<PushManifestToBuildAssetRegistry/>`.


pjcollins:
Testing with https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5737322&view=results

pjcollins:
Looks like this fixed the manifest creation issue.

jonathanpeppers (approved):

Issue #6732, opened 2022-02-10
Incremental Release builds don't AOT the app


eerhardt:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

Maui main

### Description

When I build for Release a 2nd time, it doesn't look like the app is getting AOT'd because the app is going from 690ms to 2.3s for startup.

### Steps to Reproduce

1. `dotnet new maui` (tweak it to only have android TFM)
2. `dotnet build -c Release -t:Run -f net6.0-android -r android-x64 /bl`
3. app starts in 690ms on the device
4. `dotnet build -c Release -t:Run -f net6.0-android -r android-x64 /bl` (again)
5. app starts in 2.3 s on the device

### Did you find any workaround?

Deleting `bin` and `obj` folders and building again.

PR #6738, opened 2022-02-11
[One .NET] fix incremental AOT builds

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/6732

A problem occurs when you do:

    > dotnet new android
    > dotnet build -c Release -r android-arm64
    > dotnet build -c Release -r android-arm64

After the second build with no changes, MSBuild targets run that
shouldn't -- and you end up with an `.apk` that is missing all the
AOT'd assemblies!

What is happening is the `Inputs` to the `_RemoveRegisterAttribute`
have the wrong path of a file:

    Input file "obj\Release\net6.0-android\android-arm64\android-arm64\link.flag" does not exist.

Removing `-r android-arm64` from the above commands solve the problem!

The issue here being:

1. We have an "outer" build where normally `$(RuntimeIdentifiers)`
   (plural) triggers an "inner" build for each RID.

2. Each inner build runs ILLink & AOT compilation.

3. The outer build collects the resulting files and does other work.

At step 3, we were using the wrong path for `$(_AndroidLinkFlag)` --
but only when `$(RuntimeIdentifier)` (singular) is used.

I fixed this problem, and updated various tests to catch this in the
future.

I would ideally love for this inner/outer build situation to be
handled by the dotnet/sdk, but this is what we have for now.


jonathanpeppers:
`BuildIncrementalAot("arm64-v8a","",False,True)` test fails on classic Xamarin.Android, can fix this up next week.

dellis1972 (approved):

PR #6750, opened 2022-02-15
[Mono.Android.dll] Bind API-Tiramisu DP 1.

jpobst:
Context: https://developer.android.com/about/versions/13
Context: https://android-developers.googleblog.com/2022/02/first-preview-android-13.html

Android 13 Developer Preview 1 has been released.  The Android 13

  * Feb/Mar: Developer Previews
  * April/May: Unstable Betas
  * June/July: Stable Betas
  * ???: Final

### Acceptable Breakages

* `Android.Runtime.StringDefAttribute` and `Android.Runtime.IntDefAttribute` continue to churn as expected.
* Some members are no longer marked as `Obsolete`.
* The way Google specifies some `RequiresPermission` annotations changed, and we will need to figure out how we want to support them.  (#6775)  This shouldn't be required for early previews.

[0]: https://developer.android.com/about/versions/13/overview


jonathanpeppers (approved):
This looks like everything worked as intended, looks good!

image

grendello (approved):

dellis1972 (approved):

src/Mono.Android/metadata
```diff
@@ -1659,7 +1659,47 @@
...
+
+  <!-- Fix covariant return type -->
+  <attr api-since="33" path="/api/package[@name='java.security.cert']/class[@name='URICertStoreParameters']/method[@name='clone' and count(parameter)=0]" name="managedReturn">Java.Lang.Object</attr>
+
+  <!-- We have 2 methods with the same name and the same parameter name. One needs to be enumified but not the other.
+       Our methodmap.csv isn't capable of only choosing 1 of the methods, so it sets them both as the enum, making both have the
+       same signature. So we are going to set the enum transformation here instead, where we can specify a path that only affects a single method. -->
+  <attr path="/api/package[@name='android.content.pm']/class[@name='PackageManager']/method[@name='getActivityInfo' and count(parameter)=2 and parameter[1][@type='android.content.ComponentName'] and parameter[2][@type='int']]/parameter[@name='flags']" name="enumType">Android.Content.PM.PackageInfoFlags</attr>
```
jonpryor, line 1673: Reading things like this, I wonder if we could instead start using `//class/method/@jni-signature`, a'la:

```xml
<attr
    path="/api/package[@name='android.content.pm']/class[@name='PackageManager']/method[
      @name='getActivityInfo' and
      starts-with(@jni-signature, '(Landroid/content/ComponentName;I)')
    ]"
    name="enumType"
>Android.Content.PM.PackageInfoFlags</attr>
```

It would certainly be *shorter*…

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
```diff
@@ -15,6 +15,18 @@
       ],
       "platforms": [ "win-x64", "linux-x64", "osx-x64", "osx-arm64" ],
       "extends" : [ "microsoft-net-runtime-android", "microsoft-net-runtime-android-aot" ]
+    },
+    "android-33": {
+      "description": "Preview support for Android API-33.",
+      "packs": [
+        "Microsoft.Android.Ref.33",
```
jonpryor, line 22: @jonathanpeppers: can we/should we insert "Preview" into these package names?  Is that possible?

PR #6771, opened 2022-02-16
Add support for excluding some content items from check for content

pranavkm:


jonathanpeppers (approved):
This looks OK, I restarted the designer tests.

We should link to this in the final commit message: https://github.com/dotnet/maui/pull/4728

dellis1972:
Is there an example of a content file that the runtime will want to exclude? It might be helpful for the commit message for future reference.

jonathanpeppers:
I'll merge this shortly, I'll write something up here as I have example warnings from the .NET Podcast app.

PR #6773, opened 2022-02-16
[One .NET] default $(TargetPlatformVersion) to 32

jonathanpeppers:
Currently, if you build with `net6.0-android` and leave off the
number, it defaults to API 31. Let's bump this to API 32, Android 12L.

I also fixed the `Inputs` of the `_GenerateSupportedPlatforms` target,
so it will rerun when `Configuration.props` changes.

We will likely need to coordinate the API level installed by various
Visual Studio installers.

### DIFF
--- a/Configuration.props
+++ b/Configuration.props
@@ -26,8 +26,6 @@
     <AndroidFirstFrameworkVersion Condition="'$(AndroidFirstFrameworkVersion)' == ''">v4.4</AndroidFirstFrameworkVersion>
     <AndroidFirstApiLevel Condition="'$(AndroidFirstApiLevel)' == ''">19</AndroidFirstApiLevel>
     <AndroidJavaRuntimeApiLevel Condition="'$(AndroidJavaRuntimeApiLevel)' == ''">21</AndroidJavaRuntimeApiLevel>
-    <!-- The default API level used for $(TargetPlatformVersion) -->
-    <AndroidDefaultTargetDotnetApiLevel Condition="'$(AndroidDefaultTargetDotnetApiLevel)' == ''">31</AndroidDefaultTargetDotnetApiLevel>
     <!-- The min API level supported by Microsoft.Android.Sdk, should refactor/remove when this value is the same as $(AndroidFirstApiLevel) -->
     <AndroidMinimumDotNetApiLevel Condition="'$(AndroidMinimumDotNetApiLevel)' == ''">21</AndroidMinimumDotNetApiLevel>
     <AndroidFirstPlatformId Condition="'$(AndroidFirstPlatformId)' == ''">$(AndroidFirstApiLevel)</AndroidFirstPlatformId>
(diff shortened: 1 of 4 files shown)


jonpryor:
As per discussions with @pjcollins, this also should be done alongside a bump in xamarin-android-tools, changing [`$(AndroidSdkPlatformVersion)`](https://github.com/xamarin/xamarin-android-tools/blob/a7f4d30677d0ffd39996ad76784357e87ad8678d/src/Xamarin.Android.Tools.AndroidSdk/Xamarin.Android.Tools.Versions.props#L16) to `android-32`.

Which prompted me to do PR #6772, to bump other Android SDK versions to their current values.

Configuration.props
```diff
@@ -39,6 +37,8 @@
     <AndroidLatestUnstableApiLevel Condition="'$(AndroidLatestUnstableApiLevel)' == ''">32</AndroidLatestUnstableApiLevel>
     <AndroidLatestUnstablePlatformId Condition="'$(AndroidLatestUnstablePlatformId)' == ''">32</AndroidLatestUnstablePlatformId>
     <AndroidLatestUnstableFrameworkVersion Condition="'$(AndroidLatestUnstableFrameworkVersion)'==''">v12.1</AndroidLatestUnstableFrameworkVersion>
+    <!-- The default API level used for $(TargetPlatformVersion) in .NET 6+ -->
+    <AndroidDefaultTargetDotnetApiLevel Condition=" '$(AndroidDefaultTargetDotnetApiLevel)' == '' ">$(AndroidLatestStableApiLevel)</AndroidDefaultTargetDotnetApiLevel>
```
jonpryor, line 41: Not sure if/how this should interact with template packs, a'la 6eb11f159c692563200ffb88cddc7b69f761cace & https://github.com/xamarin/xamarin-android/blob/c80dfff7a3183e21d356d2c5835aa0821fd9bd90/src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json#L32-L35

Additionally, this feels potentially "brittle"; "API-33" (currently API-Tiramisu) will *eventually* become the `$(AndroidLatestStableApiLevel)` value, *probably* in the July-August timeframe.  Do we actually want `$(AndroidDefaultTargetDotnetApiLevel)` to update to API-33 at that time, automatically?

jonpryor:
PR #6772 has been merged.

Next up are:

  * https://github.com/xamarin/xamarin-android-tools/pull/156
  * https://github.com/xamarin/xamarin-android/pull/6774

Configuration.props
jonathanpeppers: We can probably drop the 31 pack from the workload. It should keep resolving from NuGet.org if we keep this in place:

https://github.com/xamarin/xamarin-android/blob/25e0eefbd20bb4778a4717e4ef587a080c82bca0/build-tools/create-packs/Microsoft.Android.Sdk.proj#L125-L128

There are usages of `net6.0-android3` on Github:

https://github.com/search?q=net6.0-android31+extension%3Acsproj&type=Code&ref=advsearch

PR #6785, opened 2022-02-22
Test single Mono.Android.dll runtime pack

pjcollins:


pjcollins:
I think we will eventually want .NET SDK changes as discussed in https://github.com/dotnet/sdk/issues/24077 to support this instead, so this is on hold for now.

PR #6797, opened 2022-02-25
Move dotnet sandbox to build output directory

pjcollins:
Moves the "sandbox" installation of `dotnet` that we use for testing
purposes into the repo at bin/$(Configuration)/dotnet.  While this could
lead to increased disk usage across some local build environments it
provides a handful of benefits.

Every time `make prepare` or `-t:Prepare` runs we remove this directory
and start fresh.  Having it in a single "global" location is not too
desirable when working out of multiple checkout locations (fork vs
non-fork for example), or when using multiple build configurations at
the same time.

This sandbox can also now more easily be removed with a `git clean` if
desirable, we had no "automatic" way to remove the ~/android-toolchain
installation location previously.

This move will also make it easier to eventually structure our .NET
build output in a way that will allow it to be tested without having to
pack and unpack .nupkg files.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -138,7 +138,7 @@
         {
           "label": "build-sample-under-dotnet",
           "type": "shell",
-          "command": "${env:HOME}/android-toolchain/dotnet/dotnet build ${input:project} -p:Configuration=${input:configuration} -t:${input:target}",
+          "command": "bin/${input:configuration}/dotnet/dotnet build ${input:project} -p:Configuration=${input:configuration} -t:${input:target}",
           "group": {
               "kind": "build",
               "isDefault": true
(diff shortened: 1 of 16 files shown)


grendello (approved):

dellis1972:
I like this.

dellis1972 (approved):

dellis1972:
Is it worth adding shell script/.cmd file in the project root which will use the `bin/<config>/dotnet/dotnet` ?
Just to make this a little easier for developers, so we can use `dotnet-xa` (or something) from the project root and it will run the one in `bin/<config>` ? 

Is it worth it? to save a few key strokes... 🤔

jonathanpeppers:
If we had scripts like `dotnet.ps1/cmd/sh`, then we could also do things like set `$DOTNET_MULTILEVEL_LOOKUP` to 0.

At some point down the road we could also use `$DOTNETSDK_WORKLOAD_PACK_ROOTS` or `$DOTNETSDK_WORKLOAD_MANIFEST_ROOTS` to point at a different folders for workload packs.

It would also make some of our samples & instructions easier?

pjcollins:
We can add on to this in future PRs to continue to make things easier, some top level scripts seem like they would be useful. Those changes could be done today without moving this dotnet install location though. The primary reason behind these changes is that I was getting annoyed with only having the one dotnet sandbox path when working out of a couple of checkouts at once. I think the pros outweigh the cons when it comes to moving this dotnet install "into the repo", but does anyone have any reasons why this may not be desirable (aside from a potential increase in disk space consumption)?

grendello:
> We can add on to this in future PRs to continue to make things easier, some top level scripts seem like they would be useful. Those changes could be done today without moving this dotnet install location though. The primary reason behind these changes is that I was getting annoyed with only having the one dotnet sandbox path when working out of a couple of checkouts at once. I think the pros outweigh the cons when it comes to moving this dotnet install "into the repo", but does anyone have any reasons why this may not be desirable (aside from a potential increase in disk space consumption)?

it's been annoying me for quite a while, too, since I use worktrees and switching between them frequently... Each time it would nuke the copy of dotnet we installed, and require to pack it again, install maui again... very time wasting. So I really like this PR! :)

moljac (approved):

PR #6799, opened 2022-02-28
[installer] add missing .pdb files to Xamarin.Android

jonathanpeppers:
Partially fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1467817

A warning is printed when building with `-p:MonoSymbolArchive=true`:

    Warning: Directory obj\Release\120\android\assets contains Java.Interop.dll but no debug symbols file was found.

It turns out we are missing `Java.Interop.pdb` from our installer?

I went ahead and also added `Mono.Posix.pdb` and
`System.EnterpriseServices.pdb`, since we appear to have these in
build output.


jonathanpeppers:
As I suspected this hit:
```
error MSB3030: Could not copy the file "/Users/builder/azdo/_work/2/s/xamarin-android/bin/Release/lib/xamarin.android/xbuild-frameworks/MonoAndroid/v1.0/Java.Interop.pdb" because it was not found. 
```

PR #6805, opened 2022-03-02
[Mono.Android] JNIEnv.FindClass() can use StaticMethods.CallStaticObjectMethod()

jonathanpeppers:
Code paths using `JNIEnv.CallStaticObjectMethod` are using the "old"
way of doing Java.Interop. One such case is found when profiling the
.NET Podcast app:

    15.09ms Mono.Android!Android.Runtime.JNIEnv.FindClass
     6.25ms Mono.Android!Android.Runtime.JNIEnv.CallStaticObjectMethod

This is called in one place by:

    6.16ms Mono.Android!Android.Content.Intent..ctor(Android.Content.Context,System.Type)

If we instead define on `Java.Lang.Class`:

    internal static JniPeerMembers Members => _members;

We could access `Java.Lang.Class.Members.StaticMethods`, however in
this case we get the method ID from:

    internal static unsafe void Initialize (JnienvInitializeArgs* args)
    {
        // ...
        mid_Class_forName = args->Class_forName;

If we use the `IntPtr` here to create:

    new JniMethodInfo (args->Class_forName, isStatic: true)

We can instead use:

    JniEnvironment.StaticMethods.CallStaticObjectMethod (Java.Lang.Class.Members.JniPeerType.PeerReference, mid_Class_forName, parameters)

With this change in place, I'm not really seeing a noticeable
performance difference, but it should help a small amount by
allocating less?

### DIFF
--- a/src/Mono.Android/Android.Runtime/JNIEnv.cs
+++ b/src/Mono.Android/Android.Runtime/JNIEnv.cs
@@ -574,7 +574,7 @@ namespace Android.Runtime {
 			}
 
 			IntPtr global_ref = NewGlobalRef (local_ref.Handle);
-			DeleteLocalRef (local_ref.Handle);
+			JniObjectReference.Dispose (ref local_ref);
 			return global_ref;
 		}
 
--- a/src/Mono.Android/Android.Runtime/JNIEnv.cs
+++ b/src/Mono.Android/Android.Runtime/JNIEnv.cs
@@ -43,7 +43,7 @@ namespace Android.Runtime {
 		static IntPtr java_vm;
 		static IntPtr load_class_id;
 		static IntPtr gref_class;
-		static IntPtr mid_Class_forName;
+		static JniMethodInfo? mid_Class_forName;
 		static int version;
 		static int androidSdkVersion;
 
@@ -162,7 +162,7 @@ namespace Android.Runtime {
 			java_class_loader = args->grefLoader;
 			load_class_id     = args->Loader_loadClass;
 			gref_class        = args->grefClass;
-			mid_Class_forName = args->Class_forName;
+			mid_Class_forName = new JniMethodInfo (args->Class_forName, isStatic: true);
 
 			if (args->localRefsAreIndirect == 1)
 				IdentityHash = v => _monodroid_get_identity_hash_code (Handle, v);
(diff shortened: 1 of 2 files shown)


src/Mono.Android/Android.Runtime/JNIEnv.cs
```diff
@@ -557,19 +557,24 @@ static unsafe IntPtr BinaryName (string classname)
...
+				};
+				local_ref = JniEnvironment.StaticMethods.CallStaticObjectMethod (Java.Lang.Class.Members.JniPeerType.PeerReference, mid_Class_forName!, parameters);
 			} finally {
 				DeleteLocalRef (native_str);
 			}
 
-			IntPtr global_ref = NewGlobalRef (local_ref);
-			DeleteLocalRef (local_ref);
+			IntPtr global_ref = NewGlobalRef (local_ref.Handle);
+			DeleteLocalRef (local_ref.Handle);
```
jonpryor, line 577: This should instead be:

```csharp
JniObjectReference.Dispose (ref local_ref);
```

jonpryor:
```markdown
Context: https://github.com/xamarin/monodroid/commit/9f59e196782f5ffa8c31471a629dd45cf2538e85

When `Java.Interop.dll` was first integrated into Xamarin.Android,
the integration resulted in *two separate* bindings for the JNIEnv
JNI structure.  xamarin/monodroid@9f59e196 replaced the original
`Android.Runtime.JNIEnv` binding with a set of methods which just
forwarded to the new `Java.Interop.dll`-based binding:

	partial class JNIEnv {

	    // Pre xamarin/monodroid@9f59e196 binding behavior
	    public static IntPtr CallObjectMethod (IntPtr jobject, IntPtr jmethod)
	    {    
	        if (jobject == IntPtr.Zero)
	            throw new ArgumentException ("'jobject' must not be IntPtr.Zero.", "jobject");
	        if (jmethod == IntPtr.Zero)
	            throw new ArgumentException ("'jmethod' must not be IntPtr.Zero.", "jmethod");

	        IntPtr tmp = Env.CallObjectMethod (Handle, jobject, jmethod);

	        Exception __e = AndroidEnvironment.GetExceptionForLastThrowable ();
	        if (__e != null)
	            ExceptionDispatchInfo.Capture (__e).Throw ();

	        return LogCreateLocalRef (tmp);
	    }

	    // Post xamarin/monodroid@9f59e196 binding behavior
	    public static unsafe short CallShortMethod (IntPtr jobject, IntPtr jmethod, JValue* parms)
	    {
	        return JniEnvironment.InstanceMethods.CallShortMethod (
	                new JniObjectReference (jobject),
	                new JniMethodInfo (jmethod, isStatic: false),
	                (JniArgumentValue*) parms);
	    }
	}

The problem with this "binding unification" is that all codepaths
using the original `JNIEnv.Call*Method()` methods can result in the
creation of "extra garbage" which the GC needs to deal with, e.g. the
`new JniMethodInfo(…)` expression in the above snippet.  These methods
are for *compatibility*, not performance.

We *should* migrate *off* those methods, but we haven't done so.

The legacy `JNIEnv.FindClass()` and `JNIEnv.CallStaticObjectMethod()`
methods show up in profiling the .NET Podcast app:

	15.09ms Mono.Android!Android.Runtime.JNIEnv.FindClass
	 6.25ms Mono.Android!Android.Runtime.JNIEnv.CallStaticObjectMethod

This is called in one place by:

	 6.16ms Mono.Android!Android.Content.Intent..ctor(Android.Content.Context,System.Type)

Update `JNIEnv.FindClass()` so that it doesn't use
`JNIEnv.CallStaticObjectMethod()`, but instead use
`Java.Interop.JniEnvironment.StaticMethods.CallStaticObjectMethod()`.
This in turn requires updating `Java.Lang.Class` to declare:

	internal static JniPeerMembers Members => _members;

so that the `jclass` instance for `java.lang.Class.class` can be
used as part of the static method invocation.

With this change in place, I'm don't really seen a noticeable
performance difference, but it should help a small amount by
reducing GC allocations.
```

PR #6818, opened 2022-03-08
[docs] improve instructions for obtaining methods.txt

jonathanpeppers:
When recording JIT times, we didn't have instructions to set
/manifest/application/@android:debuggable="true". This is a crucial
step for Release builds.

I also noted that `counters.txt` is not useful in .NET 6. It is
mostly empty text such as:

    ## Runtime.register: type=Microsoft.NetConf2021.Maui.MainApplication, Microsoft.NetConf2021.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
    ## Runtime.register: type=Microsoft.Maui.MauiApplication, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
    ## Runtime.register: type=Microsoft.Maui.MauiApplication+ActivityLifecycleCallbacks, Microsoft.Maui

It might be worth removing support for this in .NET 6, but let's just
comment that it isn't useful for now.

### DIFF
--- a/Documentation/guides/profiling.md
+++ b/Documentation/guides/profiling.md
@@ -163,7 +163,18 @@ selectively AOT'ing only the methods included in the profile.
 
 ## Profiling the JIT Compiler
 
-Before launching the app, run the `adb` command:
+If profiling a Release build, you'll need to edit your
+`AndroidManifest.xml` to make the application "debuggable":
+
+```xml
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+  <application android:debuggable="true" ...
+```
+
+Debug builds already set this value by default.
+
+Next, run the `adb` command:
 
     > adb shell setprop debug.mono.log timing,default
 
@@ -213,6 +224,8 @@ Mono:
     Methods JITted using mono JIT       : 921
     Methods JITted using LLVM           : 0
     Methods using the interpreter       : 0
+
+_NOTE: that `counters.txt` is not available in .NET 6 projects._
 
 `methods.txt` has the individual JIT times of each method:
 


jonathanpeppers:
@grendello should we think about getting rid of `counters.txt` in .NET 6? We probably have code that doing something and you end up with a mostly empty file?

Documentation/guides/profiling.md
```diff
@@ -214,6 +225,8 @@ Mono:
     Methods JITted using LLVM           : 0
     Methods using the interpreter       : 0
 
+_NOTE: that `counters.txt` is not available in .NET 6 projects._
```
grendello, line 228: The code that generates `counters.txt` is still active in NET6 and `dotnet/mono` runtime appears to still [support](https://github.com/dotnet/runtime/blob/a0a5fc9771f63aa31cb45b3577a7cea31527a2fb/src/mono/mono/mini/mini-runtime.c#L4949-L4994) [`mono_counters_dump`](https://github.com/dotnet/runtime/blob/a0a5fc9771f63aa31cb45b3577a7cea31527a2fb/src/mono/mono/utils/mono-counters.c#L619-L662) API we use (and it's not marked as obsolete), it should still work?

grendello:
> @grendello should we think about getting rid of `counters.txt` in .NET 6? We probably have code that is doing something and you end up with a mostly empty file?

From looking at the Mono runtime code, it should be gathering statistics as before if JIT stats are enabled.  We enable it [here](https://github.com/xamarin/xamarin-android/blob/703564fe335aa6c02b4994f15e059280d87ae5e7/src/monodroid/jni/monodroid-glue.cc#L2191) using the [following flags](https://github.com/xamarin/xamarin-android/blob/703564fe335aa6c02b4994f15e059280d87ae5e7/src/monodroid/jni/monodroid-glue-internal.hh#L159), so unless something was disabled/changed in dotnet/mono, it should still work.  But, if we can get the same stats using the tracing API, then it might make more sense for .NET 6+, agreed.

Documentation/guides/profiling.md
jonathanpeppers: I tried a Debug build and `-c Release -p:RunAOTCompilation=false` (to turn off AOT).

This appears to work fine in Xamarin.Android projects in either of those configurations, but not .NET 6?

grendello: Hm, can it be due to missing FS write permissions for the test app?

jonathanpeppers: No the file is there, it's just mostly blank like:
```
## Runtime.register: type=Microsoft.NetConf2021.Maui.MainApplication, Microsoft.NetConf2021.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
## Runtime.register: type=Microsoft.Maui.MauiApplication, Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
## Runtime.register: type=Microsoft.Maui.MauiApplication+ActivityLifecycleCallbacks, Microsoft.Maui
```
Each line has several empty lines in between.

grendello: So it appears the counters are disabled in dotnet/mono, even though code is there.

Issue #6840, opened 2022-03-17
Strip AOT shared libraries when packaging


grendello:
https://github.com/xamarin/xamarin-android/pull/6683 introduced LLVM-based toolchain and it appears that the debug data generated during build is much bigger than before.
We should probably strip AOT (and perhaps other) .so libraries on packaging time.

grendello:
Alternatively, we can pass `-S` via the `LdFlags` property of the `AotCompiler` task

jonathanpeppers:
Radek add this at one point:

https://github.com/xamarin/xamarin-android/blob/dccda9898170e7fe979fd7f3e1811061a9b19af7/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets#L212-L227

So AOT `.so` files aren't going through here?

grendello:
It appears it only affects framework libs? I think adding `-s` to `LdFlags` would be faster - we'd save running `strip` X times

PR #6683, opened 2022-01-28
Use new LLVM-based Xamarin.Android toolchain

grendello:
Context: https://github.com/xamarin/xamarin-android/issues/6685
Context: fd5f31cc89648957243f31ba8fd7af7fccba11e4

So far, `Xamarin.Android` has been using a GNU Binutils toolchain to compile and link native assembler code we generate
during application build.  Unfortunately, Binutils have a problem with certain filename encodings on Windows and so we
decided to switch to an LLVM-based toolchain which handles such file names without issues.

However, since Mono/dotnet AOT expects a GNU build chain, it was necessary to implement a GNU Assembler (`gas`) wrapper
around the LLVM `llvm-mc` assembler, so that command lines used by the Mono AOT compiler keep working fine.

Since LLVM utilities are multi-target by default and, unlike GNU Binutils, they don't need separate builds of every utility,
we provide GNU Binutils architecture-prefixed wrapper scripts which invoke their LLVM counterparts with appropriate parameters.

The following LLVM utilities are included:

 - `llvm-mc` (assembler)
 - `lld` (linker)
 - `llvm-strip`

PR #6842, opened 2022-03-17
Strip AOT shared libraries when linking

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/6840
Context: https://github.com/xamarin/xamarin-android/pull/6683

Pass the `-s` flag to the native linker in order to produce shared
AOT libraries without debug symbols.  Symbols are stripped unless
the `$(DebugSymbols)` property is set to `True`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -193,6 +193,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidAotMode Condition=" '$(AndroidAotMode)' == '' ">None</AndroidAotMode>
 	<AotAssemblies Condition=" '$(AndroidAotMode)' != '' And '$(AndroidAotMode)' != 'None' And '$(AndroidAotMode)' != 'Interpreter' ">True</AotAssemblies>
 	<AotAssemblies Condition=" '$(AotAssemblies)' == '' ">False</AotAssemblies>
+    <_AndroidAotStripLibraries Condition=" '$(_AndroidAotStripLibraries)' == '' And '$(AndroidIncludeDebugSymbols)' != 'true' ">True</_AndroidAotStripLibraries>
 
 	<AndroidUseDebugRuntime
 			Condition="'$(AndroidUseDebugRuntime)' == '' And '$(EmbedAssembliesIntoApk)' == 'True' And '$(Optimize)' == 'True' "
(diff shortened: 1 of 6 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -193,6 +193,7 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<AndroidAotMode Condition=" '$(AndroidAotMode)' == '' ">None</AndroidAotMode>
 	<AotAssemblies Condition=" '$(AndroidAotMode)' != '' And '$(AndroidAotMode)' != 'None' And '$(AndroidAotMode)' != 'Interpreter' ">True</AotAssemblies>
 	<AotAssemblies Condition=" '$(AotAssemblies)' == '' ">False</AotAssemblies>
+    <_AndroidAotStripLibraries Condition=" '$(_AndroidAotStripLibraries)' == '' And '$(AndroidIncludeDebugSymbols)' != 'true' ">True</_AndroidAotStripLibraries>
```
jonathanpeppers, line 196: One issue is `$(AndroidIncludeDebugSymbols)` is defined later in the file, ~line 310, so it would be blank here.

jonathanpeppers:
This seems pretty good!
```

 17,416 lib/armeabi-v7a/libaot-Mono.Security.dll.so *2
 11,356 lib/x86/libaot-Mono.Security.dll.so *2
Error: apkdiff: PackageSize decrease 3,413,474 is 3,364,322 bytes more than the threshold 49,152. apk1 size: 19,475,110 bytes, apk2 size: 16,061,636 bytes.
```
And this was a classic Xamarin.Andorid/Xamarin.Forms apps.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
dellis1972: Also indentation :D

grendello: > Also indentation :D

I still say we should unify indentation in MSBuild files :) `.editorconfig` says it should be two spaces, and that's what my editor used here... :) But the rest of the lines above is indented with tabs... :)

jonathanpeppers (approved):

dellis1972 (approved):

jonpryor:
```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/6840

Context: b21cbf943a19079773f9e762908939181a7f2180

Commit b21cbf94 contained a TODO:

> TODO: the Resulting `.apk` sizes also increase[d] unexpectedly, with
> `Xamarin.Forms_Performance_Integration-Signed-Release-Profiled-Aot.apkdesc`
> showing a 1.7MB increase in `.apk` size.  We believe that this is
> because of more verbose debug symbols.
> xamarin/xamarin-android#6840 will track this.

Update the `<GetAotArguments/>` MSBuild task to add the `-s` linker
flag to `GetAotArguments.LdFlags`.  This will cause the native linker
to produce shared AOT libraries without debug symbols.

Debug symbols are stripped unless the `$(DebugSymbols)`=True.

This fixes the size regression in
`Xamarin.Forms_Performance_Integration-Signed-Release-Profiled-Aot.apkdesc`,
shrinking PackageSize from 19,475,110 down to 16,061,636.
Compare to the pre-b21cbf94 PackageSize of 17,713,830: we're now
1.6MB *smaller* than b21cbf94!
```

PR #6844, opened 2022-03-21
DesignTime Builds should honour AndroidGenerateResourceDesigner.

dellis1972:
Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1440274

The design time build was not taking into account the `AndroidGenerateResourceDesigner` property value. 
As a result we would always get a `Resource.designer.cs` file generated in the designtime directory.

So lets turn off the designtime build if that property is set. 
Also note set a defaut for `AndroidGenerateResourceDesigner` in legacy so that it is always `true`. 
This is to get around some issues under legacy where `AndroidGenerateResourceDesigner` was not 
defined. This is because the designtime build tasks and targets are shared between both systems.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
@@ -12,6 +12,7 @@ projects. .NET 5 projects will not import this file.
 
   <PropertyGroup>
     <AndroidBoundExceptionType Condition=" '$(AndroidBoundExceptionType)' == '' ">Java</AndroidBoundExceptionType>
+    <AndroidGenerateResourceDesigner>True</AndroidGenerateResourceDesigner>
   </PropertyGroup>
 
   <PropertyGroup Condition=" '$(AndroidBuildApplicationPackage)' == 'True' And '$(AndroidApplication)' == 'True' ">
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
@@ -12,7 +12,7 @@ projects. .NET 5 projects will not import this file.
 
   <PropertyGroup>
     <AndroidBoundExceptionType Condition=" '$(AndroidBoundExceptionType)' == '' ">Java</AndroidBoundExceptionType>
-    <AndroidGenerateResourceDesigner>True</AndroidGenerateResourceDesigner>
+    <AndroidGenerateResourceDesigner Condition=" '$(AndroidGenerateResourceDesigner)' == '' ">True</AndroidGenerateResourceDesigner>
   </PropertyGroup>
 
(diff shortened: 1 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Xamarin.Android.Legacy.targets
```diff
@@ -12,6 +12,7 @@ projects. .NET 5 projects will not import this file.
 
   <PropertyGroup>
     <AndroidBoundExceptionType Condition=" '$(AndroidBoundExceptionType)' == '' ">Java</AndroidBoundExceptionType>
+    <AndroidGenerateResourceDesigner>True</AndroidGenerateResourceDesigner>
```
jonathanpeppers, line 15: I think this might need to check if it is blank, like it does in .NET 6:

https://github.com/xamarin/xamarin-android/blob/d4da1c252f45c4910abc1bd9e5be9ecf9dc683a0/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets#L13

This way you can set it to `false`.

dellis1972: We don't support this property under Legacy as per the [docs](https://github.com/xamarin/xamarin-android/blob/b21cbf943a19079773f9e762908939181a7f2180/Documentation/guides/building-apps/build-properties.md#androidgenerateresourcedesigner). So I think it should always default to `true`

jonathanpeppers: If you set to `false` in your `.csproj`, this would change it to `true`. I think we want to allow them to set it to `false`?

jonpryor:
```markdown
Fixes: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1440274

In some circumstances, when
[`$(AndroidGenerateResourceDesigner)`][0]=False and a custom
`Resource.Designer.cs` is provided in the project, loading a project
would cause Visual Studio IntelliSense to show an ambiguity with e.g.
`Resource.Layout.activity_main`, as the member exists in *both* the
custom `Resource.Designer.cs` *and* in the design-time-generated
`Resource.Designer.cs`.

This was because the design-time build was not taking
`$(AndroidGenerateResourceDesigner)` into consideration before
generating `Resource.designer.cs`.  Consequently, we would *always*
create a `Resource.designer.cs` file into the `designtime` directory.

Fix this by turning off the `designtime` build when
`$(AndroidGenerateResourceDesigner)`=False.

Additionally, update `Xamarin.Android.Legacy.targets` so that
`$(AndroidGenerateResourceDesigner)` is *always* set in Legacy builds,
with a default value of True.  This is to get around some issues under
legacy where `$(AndroidGenerateResourceDesigner)` was not defined.
This is because the `designtime` build tasks and targets are shared
between both Classic and .NET 6 build systems.

[0]: https://github.com/xamarin/xamarin-android/blob/4b33a6229c9babde795dea48d20ca86d8f9f9c97/Documentation/guides/building-apps/build-properties.md#androidgenerateresourcedesigner
```

PR #6845, opened 2022-03-21
Raise an error if the Resource starts with an Invalid Character.

dellis1972:
Fixes https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1333264/

Android Resources MUST start with a Letter or an underscore. They cannot
start with any other character. We should raise a specific error for this
so it is obvious to the user what the issue is.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/CheckForInvalidResourceFileNames.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/CheckForInvalidResourceFileNames.cs
@@ -30,6 +30,10 @@ namespace Xamarin.Android.Tasks {
 					resourceFile = resource.ItemSpec;
 				var fileName = Path.GetFileName (resourceFile);
 				var directory = Path.GetFileName (Path.GetDirectoryName (resourceFile));
+				char firstChar = fileName [0];
+				if (!(Char.IsLetter (firstChar) || firstChar != '_')) {
+					Log.LogCodedError ("APT0004", resource.ItemSpec, 0, Properties.Resources.APT0004, fileNameCheck);
+				}
 				if (directory.StartsWith ("values", StringComparison.OrdinalIgnoreCase)) {
 					var match = fileNameWithHyphenCheck.Match (fileName);
 					if (match.Success) {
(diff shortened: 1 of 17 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/CheckForInvalidResourceFileNames.cs
```diff
@@ -30,6 +30,10 @@ public override bool RunTask ()
 					resourceFile = resource.ItemSpec;
 				var fileName = Path.GetFileName (resourceFile);
 				var directory = Path.GetFileName (Path.GetDirectoryName (resourceFile));
+				char firstChar = fileName [0];
+				if (!(Char.IsLetter (firstChar) || firstChar != '_')) {
```
jonathanpeppers, line 34: I think maybe the logic is backwards checking against `_`, should it be?
```suggestion
				if (!Char.IsLetter (firstChar) && firstChar != '_') {
```

jonathanpeppers (commented):
Should we add one test that tries to create an `AndroidResource` that starts with a number?

src/Xamarin.Android.Build.Tasks/Tasks/CheckForInvalidResourceFileNames.cs
dellis1972: Doh, I thought I'd created a draft PR 🤦  Good catch though :D

src/Xamarin.Android.Build.Tasks/Tasks/CheckForInvalidResourceFileNames.cs
```diff
@@ -30,6 +30,10 @@ public override bool RunTask ()
 					resourceFile = resource.ItemSpec;
 				var fileName = Path.GetFileName (resourceFile);
 				var directory = Path.GetFileName (Path.GetDirectoryName (resourceFile));
+				char firstChar = fileName [0];
+				if (!Char.IsLetter (firstChar) && firstChar != '_') {
```
jonpryor, line 34: [`char.IsLetter()`](https://docs.microsoft.com/en-us/dotnet/api/system.char.isletter?view=net-6.0#system-char-isletter(system-char)) determines whether or not a character is a "member of any category of Unicode letter".  This would include things such as `é` and `の`, which will *not* be supported by `aapt`.

This should instead use a range check:

```csharp
('a' <= firstChar && firstChar <= 'z') ||
('A' <= firstChar && firstChar <= 'Z') ||
firstChar == '_'
```

jonathanpeppers (approved):

PR #6850, opened 2022-03-23
Add `dotnet-local` scripts to make using in tree dotnet easier.

dellis1972:
Lets add a couple of `dotnet-local` scripts which will allow us to
use the in-tree dotnet install easier. Also update the VSCode tasks
to make use of this new script.

### DIFF
--- a/dotnet-local.sh
+++ b/dotnet-local.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+if [ -d "bin/Release/dotnet" ]; then
+    ./bin/Release/dotnet/dotnet $@
+else
+    ./bin/Debug/dotnet/dotnet $@
+fi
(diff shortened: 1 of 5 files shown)


dotnet-local.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+if [ -d "bin/Release/dotnet" ]; then
```
grendello, line 2: I'd use `-x "bin/Release/dotnet/dotnet"` here, so that we know that the binary both exists and is executable

dotnet-local.sh
```diff
@@ -0,0 +1,6 @@
...
+    ./bin/Release/dotnet/dotnet $@
```
grendello, line 3: use `exec ./bin/Release/dotnet/dotnet "$@"`, which will replace `bash` process with `dotnet` and will properly quote arguments with spaces in them

dotnet-local.sh
```diff
@@ -0,0 +1,6 @@
...
+else
```
grendello, line 4: I'd add another `-x` check here and an `else` which would report an error if the `dotnet` binary doesn't exist or isn't executable

jonathanpeppers (approved):

PR #6851, opened 2022-03-24
[Xamarin.Android.Build.Tasks] remove Microsoft.AspNetCore.App.Runtime.linux workaround

jonathanpeppers:
Context: https://github.com/dotnet/sdk/issues/19891

In 3f052b5f, I added a workaround to exclude
`Microsoft.AspNetCore.App.Runtime.linux` packages from Android builds.
This was a problem where the `android` RID "inherited" from the
`linux` RID, and so we ended up with unintended packages restored.

This was fixed in https://github.com/dotnet/installer/pull/12702, so
we should remove this workaround.


dellis1972 (approved):

dellis1972:
The error is our old friend "java.util.zip.ZipException: invalid stored block lengths", it looks unrelated.

PR #6853, opened 2022-03-24
Generate LLVM IR instead of native assembler code

grendello:
Instead of continuing to generate native assembler code, as we've been doing for a while now, switch to
generating [LLVM IR](https://llvm.org/docs/LangRef.html) sources.  This has the advantage that LLVM takes
care for us of generating valid native assembler code, making sure it adheres to all the ABI requirements
of the target platform.  It also makes it easier, in future, to generate actual executable code instead of
just data as we do now.

Once the LLVM IR is generated, the `llc` utility from LLVM is used to generate the actual native assembler
code.  LLVM IR files are larger than the native assembler ones we've been generating so far, because of the
language's verbosity, but we generate less of them.

This PR covers generation of compressed assemblies, application environment and typemap sources.


jonathanpeppers (approved):
Everything in the `src/Xamarin.Android.Build.Tasks` folder looks OK to me. 👍

dellis1972 (approved):

PR #6856, opened 2022-03-25
Optionally force `res` directory creation for `.aar` libraries.

dellis1972:
Normally an `R.java` class is generated for each library. This would
allow the `R.*` classes to be access from Java without needing to get
them from C# first. However our `R.java` generation code only did this
for libraries which contain a `res` folder. This made sense at the time
as it helped reduce apk size by not generating an `R.java` when it did
not seem to be needed.

In Maui we have a few places where we want to be able to access the
AndroidResource identifiers from within custom Java Code. However this
library does NOT have any resources of its own, so we don't generate a
`R.java` file for it. This means we have to look up the code in C# then
pass that value into the custom Java code. These values then become
part of the API. Its not ideal.

So lets add support for adding additional metadata to `AarLibrary`
ItemGroup Items like we do for `References`. This new metadata item will
be `AndroidCustomMetaDataForAarLibrary`. This will allow us to add
additional metadata for `.aar` files we discover as part of the build
process.

In this case we want to support a new `AndroidForceCreateResourceDirectory`
metadata property. This will cause the `ResolveLibraryProjectImports`
task to create an empty `res` directory as part of its `.aar` extraction
process. This empty directory will be ignored when calling `aapt2` but
will allow the generation of the required `R.java` file. This will allow
us to access things like `R.styleable` in Java code for the library and
have it resolve to the correct value at runtime.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AndroidLibraries.targets
@@ -10,6 +10,7 @@ projects.
 -->
 <Project>
 
+  <UsingTask TaskName="Xamarin.Android.Tasks.AppendCustomMetadataToItemGroup" AssemblyFile="$(_XamarinAndroidBuildTasksAssembly)" />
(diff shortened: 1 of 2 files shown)


jonpryor:
I'm not sure why we *need* `%(AndroidForceCreateResourceDirectory)` metadata.  It exists to force creation of `R.java`.  However, *simply adding an `@(AndroidResource)`* will *also* force the creation of `R.java`, and it doesn't seem unreasonable to me that if a project needs access to Android resources now, it might need more of its own in the future, so a "hacky workaround" of "just add an `@(AndroidResource)`" doesn't seem "bad" to me.

PR #6857, opened 2022-03-25
Remove unnecessary background Task in AndroidMessageHandler

### LINKED ISSUE #6854: AndroidMessageHandler queues unnecessary Task in SetupRequest
### Android application type

### Affected platform version

Maui main

### Description

When profiling the [Maui .NET Podcasts app](https://github.com/microsoft/dotnet-podcasts/tree/main/src/Mobile), I'm seeing a few Threads getting spun up. Looking at one of them, this method sticks out:

https://github.com/xamarin/xamarin-android/blob/b3cb3a9e49bbffed4fd00f716ce04ab502396fae/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs#L851-L855

All AssertSelf does is: check that the current object is not disposed.

https://github.com/xamarin/xamarin-android/blob/b3cb3a9e49bbffed4fd00f716ce04ab502396fae/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs#L282-L287

There seems little reason to run this in a background Task. Instead, it should just be:

```C#
		protected virtual Task SetupRequest (HttpRequestMessage request, HttpURLConnection conn)
		{
			AssertSelf ();

			return Task.CompletedTask;
		}
```

This allows the method to return `Task`, but not do anything asynchronously.

### Steps to Reproduce

1. Run https://github.com/microsoft/dotnet-podcasts/tree/main/src/Mobile using the instructions located at https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/tracing.md.
2. Inspect how many Threads are being spawned in the speedscope.

### Expected results
Only necessary threads are spawned.

The ones that look necessary are:

1. The "main" thread
2. The TimerQueue is spawning a TimerThread (from Http preparing a cancellation source).
3. The PortableThreadPool is spawning a GateThread
4. AndroidMessageHandler.ConnectAsync is creating a background task (which seems appropriate)

### Actual results
It looks like 6 threads are getting spawned. See the below speed scope. The threads 3 & 6 look unnecessary. Thread 3 is the one that is getting spawned due to SetupRequest running an unnecessary background Task.

_No response_

### Relevant log output

[maui-app_20220324_165101.speedscope.zip](https://github.com/xamarin/xamarin-android/files/8346188/maui-app_20220324_165101.speedscope.zip)

eerhardt:
AndroidMessageHandler.SetupRequest returns a Task so overrides can be async. The base implementation checks the object hasn't been disposed. However, it is doing this check using Task.Run, which queues a work item to the ThreadPool. This is unnecessary and causes more threads to be spawned than what is needed.

The fix is to just check for disposed synchronously and return a completed Task.

Fix #6854

### DIFF
--- a/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.Legacy.cs
+++ b/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.Legacy.cs
@@ -761,8 +761,9 @@ namespace Xamarin.Android.Net
 		/// <param name="conn">Pre-configured connection instance</param>
 		protected virtual Task SetupRequest (HttpRequestMessage request, HttpURLConnection conn)
 		{
-			Action a = AssertSelf;
-			return Task.Run (a);
+			AssertSelf ();
+			
+			return Task.CompletedTask;
 		}
 
 		/// <summary>
--- a/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
+++ b/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
@@ -850,8 +850,9 @@ namespace Xamarin.Android.Net
 		/// <param name="conn">Pre-configured connection instance</param>
 		protected virtual Task SetupRequest (HttpRequestMessage request, HttpURLConnection conn)
 		{
-			Action a = AssertSelf;
-			return Task.Run (a);
+			AssertSelf ();
+			
+			return Task.CompletedTask;
 		}
 
 		internal Task SetupRequestInternal (HttpRequestMessage request, HttpURLConnection conn)


jonathanpeppers (approved):
This looks good to me, @grendello do you remember why it was like that?

I think this was the commit:

https://github.com/xamarin/xamarin-android/commit/c873c7ba6041d10def9206df6ca732a29accb11f

grendello (approved):

grendello:
> This looks good to me, @grendello do you remember why it was like that?
> 
> I think this was the commit:
> 
> [c873c7b](https://github.com/xamarin/xamarin-android/commit/c873c7ba6041d10def9206df6ca732a29accb11f)

Yep, that was it.

Issue #6775, opened 2022-02-17
Explore permission annotations in API-Tiramisu


jpobst:
We use Google's `annotations.zip` file included with each Android SDK Platform to create certain metadata in our bindings.

In API-Tiramisu, some `RequiresPermission` annotations changed from:

```xml
<item name="android.companion.CompanionDeviceManager void associate(android.companion.AssociationRequest, android.companion.CompanionDeviceManager.Callback, android.os.Handler)">
  <annotation name="androidx.annotation.RequiresPermission">
    <val name="value" val="&quot;android.permission.REQUEST_COMPANION_PROFILE_WATCH&quot;" />
    <val name="conditional" val="true" />
  </annotation>
</item>
```

to:

```xml
<item name="android.companion.CompanionDeviceManager void associate(android.companion.AssociationRequest, android.companion.CompanionDeviceManager.Callback, android.os.Handler)">
  <annotation name="androidx.annotation.RequiresPermission">
    <val name="anyOf" val="{&quot;android.permission.REQUEST_COMPANION_PROFILE_WATCH&quot;, &quot;android.permission.REQUEST_COMPANION_PROFILE_APP_STREAMING&quot;, &quot;android.permission.REQUEST_COMPANION_PROFILE_AUTOMOTIVE_PROJECTION&quot;}" />
    <val name="conditional" val="true" />
  </annotation>
</item>
```

We have 2 issues here:
- We need to fix our annotation parser to support this change.
- We use `[RequiresPermission]` to automatically add the permission to the Android manifest.  We do not have a concept of "add one of these 3 permissions" like the new annotation specifies.  We will need to decide the best way to handle this.

jpobst:
After investigation, we do no use `[RequiresPermission]` to generate manifest permissions.  I was thinking of `[assembly: UsesPermission(Android.Manifest.Permission.Internet)]`.  As such, these are not actually used by anything and it doesn't really matter if they are not correct.

PR #6868, opened 2022-03-28
[Mono.Android] API-Tiramisu Developer Preview 2

jpobst:
Context: https://developer.android.com/about/versions/13
Context: https://android-developers.googleblog.com/2022/03/second-preview-android-13.html

Android 13 Developer Preview 2 has been released.  The Android 13

  * Feb/Mar: Developer Previews
  * April/May: Unstable Betas
  * June/July: Stable Betas
  * ???: Final

### Acceptable Breakages

* `Android.Runtime.StringDefAttribute` and `Android.Runtime.IntDefAttribute` continue to churn as expected.
* Some members are no longer marked as `Obsolete`.
* The way Google specifies some `RequiresPermission` annotations changed, and we will need to figure out how we want to support them.  (#6775)  This shouldn't be required for early previews.
* `Android.Net.IpSecManager` was unsealed, making its members `virtual`, adding connectors to `[Register]`

[0]: https://developer.android.com/about/versions/13/overview


src/Mono.Android/metadata
```diff
@@ -1699,7 +1699,15 @@
   <!-- This method was already specified on a base interface, ignore to avoid "breakage" report. -->
   <remove-node api-since="33" path="/api/package[@name='java.util']/interface[@name='Deque']/method[@name='addAll' and count(parameter)=1 and parameter[1][@type='java.util.Collection&lt;? extends E&gt;']]" />
 
-	<!--
+  <!-- Rename Namespaces -->
+  <attr api-since="33" path="/api/package[@name='android.os.vibrator']" name="managedName">Android.OS.Vibrators</attr>
```
jonpryor, line 1703: Does `<ns-replace/>` support `api-since`?  *Should it*?

src/Mono.Android/metadata
```diff
@@ -1699,7 +1699,15 @@
...
+  <attr api-since="33" path="/api/package[@name='android.app.appsearch.observer']" name="managedName">Android.App.AppSearch.Observer</attr>
```
jonpryor, line 1704: I wonder if we should "preemptively pluralize" this namespace, in case they introduce a `.Observer` type.  It would also match `.AppSearch.Exceptions`.

src/Mono.Android/metadata
jpobst: It does not.  However `<ns-replace />` will not throw a warning if doesn't match anything, like `metadata` will.  😁

src/Mono.Android/metadata
jpobst: Seems reasonable.  Updated.

PR #6874, opened 2022-03-29
[xaprepare] Allow desired Android SDK platform levels to be selected.

jpobst:
Currently, we need all Android SDK platforms to perform the `api-merge` needed to build Xamarin.Android.  However, we do not need all of them on the test agents in order to run our tests.

Add an option to `xaprepare` called `--android-sdk-platforms` that can be used to specify which API levels are needed.

For test agents, request API levels `19,21,26,32` needed for tests to successfully run.  This cuts the time taken for the `xaprepare` step on a fresh Windows VM image from `19m:36s` to `5m:11s`.

A future enhancement would be to examine the tests that require API levels 19/21/26 to determine if these tests actually need to run against those levels, or if they could be updated to use the latest level.

Tests requiring old API levels:
26 - CheckMaxResWarningIsEmittedAsAWarning
21 - FailedBuildFew_ConflictingTextView, FailedBuildMany_ConflictingLinearLayout, SuccessfulBuildFew_DTB, etc.
19 - CheckTargetFrameworkVersion


pjcollins (approved):
The [AndroidToolchainComponentType.BuildDependency](https://github.com/xamarin/xamarin-android/blob/681887ebdbd192ce7ce1cd02221d4939599ba762/build-tools/xaprepare/xaprepare/Application/AndroidToolchainComponent.cs#L45) flag could also be used here, if it is added to the API levels that aren't required by tests the `AndroidTestDependencies` scenario would skip them by default.  The new argument that lets you select specific versions for _any_ xaprepare scenario also seems like a good addition.

grendello (approved):

PR #6880, opened 2022-03-30
[NET6] Call JNIEnv.Initialize directly, via a function pointer

grendello:
With the .NET6 Mono runtime we are able to use the
`[UnmanagedCallersOnly]` attribute to mark certain methods as directly
callable from native code.  This is faster than going through
`mono_runtime_invoke`, as there is nearly no marshaling of parameters
performed.  Since `Android.Runtime.JNIEnv.Initialize` is called only by
our native runtime, we can use this approach here.

Additionally, remove some reflection use by looking up and saving
managed tokens of the `Android.Runtime.JNIEnv` class as well as its two
methods, `Initialize` and `RegisterJniNatives`.  This saves time by
allowing us to not look for the above methoods via their names at
startup.

Time savings aren't spectacular in the overall startup time, but the
"native to managed" phase (essentially a call to `JNIEnv.Initialize`)
paired with the reflection savings gives us a speedup of 0.1ms here on
Pixel 6 Pro (savings might be bigger on slower devices).

### DIFF
--- a/src/monodroid/jni/application_dso_stub.cc
+++ b/src/monodroid/jni/application_dso_stub.cc
@@ -58,6 +58,9 @@ const ApplicationConfig application_config = {
 	.bundled_assembly_name_width = 0,
 	.number_of_assembly_store_files = 2,
 	.number_of_dso_cache_entries = 2,
+	.android_runtime_jnienv_class_token = 1,
+	.jnienv_initialize_method_token = 2,
+	.jnienv_registerjninatives_method_token = 3,
 	.mono_components_mask = MonoComponent::None,
 	.android_package_name = android_package_name,
 };
--- a/src/monodroid/jni/monodroid-glue.cc
+++ b/src/monodroid/jni/monodroid-glue.cc
@@ -33,6 +33,10 @@
 #include <mono/utils/mono-dl-fallback.h>
 #include <mono/utils/mono-logger.h>
 
+#if defined (NET6)
+#include <mono/metadata/mono-private-unstable.h>
+#endif
+
 #include "mono_android_Runtime.h"
 
 #if defined (DEBUG) && !defined (WINDOWS)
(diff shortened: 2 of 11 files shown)


jonpryor:
The fact that the Designer tests are passing on main while failing repeatedly on this PR is concerning:

  * https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5959594&view=logs&s=5d4d655b-31e2-594d-78dc-b1abbc6c4d01&j=291cfdd8-b55d-5aec-3c07-abeba850aaac
  * https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=5958428&view=ms.vss-test-web.build-test-results-tab&runId=36814984&resultId=100694&paneView=debug

Perusing the **Run Unit Tests** logs, this error message is present only in the output for this PR, but not main:

```
[2022-03-30 23:03:19.2] Renderer >> * Assertion at /Users/builder/jenkins/workspace/archive-mono/2020-02/android/release/mono/metadata/class.c:2834, condition `is_ok (error)' not met, function:mono_class_get, Unknown type token 0
```

So either we're really unlucky on this build, or this PR has introduced the above assertion.  Pity we don't have more context.

src/monodroid/jni/application_dso_stub.cc
```diff
@@ -58,6 +58,9 @@ const ApplicationConfig application_config = {
 	.bundled_assembly_name_width = 0,
 	.number_of_assembly_store_files = 2,
 	.number_of_dso_cache_entries = 2,
+	.android_runtime_jnienv_class_token = 1,
```
jonpryor, line 61: I wonder if these should be "*really* invalid" values, like 0 or -1.

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -1089,19 +1093,11 @@ MonodroidRuntime::init_android_runtime (
...
 
-	// TODO: try looking up the method by its token
-	MonoClass *runtime;
-#if defined (NET6)
-	runtime = mono_class_from_name (image, SharedConstants::ANDROID_RUNTIME_NS_NAME, SharedConstants::JNIENV_CLASS_NAME);
-#else
-	runtime = utils.monodroid_get_class_from_image (domain, image, SharedConstants::ANDROID_RUNTIME_NS_NAME, SharedConstants::JNIENV_CLASS_NAME);
-#endif
-	MonoMethod *method = mono_class_get_method_from_name (runtime, "Initialize", 1);
+	MonoClass *runtime = mono_class_get (image, application_config.android_runtime_jnienv_class_token);
```
jonpryor, line 1096: I wonder if this is what's crashing the Designer tests.

src/monodroid/jni/application_dso_stub.cc
grendello: Hm, might be. I forgot that `libxamarin-app.so` produced from this is used in the Android Designer context.

PR #6881, opened 2022-03-30
[ci] Add job to run after Xamarin.Android pipeline

pjcollins:
Context: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/pipeline-triggers

A new pipeline has been added to capture build information from the
Xamarin.Android pipeline, which can than be processed by old UI-style
release definitions.

The Xamarin.Android pipeline often has known or acceptable test failures
in the builds that we want to release, and old UI-style release
definitions will not run against partially failed builds.  This new
pipeline avoids that limitation as it will always be green.  The new
pipeline will also publish information about the Xamarin.Android build
that triggered it, which will allow an old UI-style release definition
to download artifacts from the Xamarin.Android build using the
[DownloadPipelineArtifact][1] task.

A [pipeline resource][0] controls the automatic triggering of this new
pipeline, and it will run once the `post_build` stage in the XA pipeline
completes successfully.

The classic VS insertions will look like this:

  1) A `Xamarin.Android` pipeline build finishes.
  2) A new build of `Xamarin.Android Release Trigger` will run automatically.
  3) This pipeline publishes a file containing `Xamarin.Android` build and pipeline ID information.
  4) The `Xamarin.Android - VS + VSM Release Insertion` release definition uses the
      build and pipeline ID to download the SBOM artifact in the corresponding `Xamarin.Android` build.

This new pipeline can also be used to "transfer" other build artifacts
from the `Xamarin.Android` pipeline to other release definitions in the
future if needed.

[0]: https://docs.microsoft.com/en-us/azure/devops/pipelines/process/resources?view=azure-devops&tabs=schema#define-a-pipelines-resource
[1]: https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/utility/download-pipeline-artifact?view=azure-devops#arguments

### DIFF
--- a/build-tools/automation/azure-pipelines-release-trigger.yaml
+++ b/build-tools/automation/azure-pipelines-release-trigger.yaml
@@ -7,8 +7,8 @@ pr: none
 
 resources:
   pipelines:
-  - pipeline: macios
-    source: xamarin-macios
+  - pipeline: xamarin-android
+    source: Xamarin.Android
     trigger:
       stages:
       - post_build
--- a/build-tools/automation/azure-pipelines-release-trigger.yaml
+++ b/build-tools/automation/azure-pipelines-release-trigger.yaml
@@ -2,11 +2,16 @@
 
 name: $(Build.SourceBranchName)-$(Build.SourceVersion)-$(Rev:r)
 
-trigger:
-- none
+trigger: none
+pr: none
 
-pr:
-- none
+resources:
+  pipelines:
+  - pipeline: macios
+    source: xamarin-macios
+    trigger:
+      stages:
+      - post_build
 
 jobs:
 - job: release_trigger
--- a/build-tools/automation/azure-pipelines-release-trigger.yaml
+++ b/build-tools/automation/azure-pipelines-release-trigger.yaml
@@ -24,22 +24,21 @@ jobs:
   - checkout: none
 
   - powershell: |
-      $json = @"
-      {
-          "triggeringbuild": {
-              "id": "$(Build.TriggeredBy.BuildId)",
-              "pipeline": "$(Build.TriggeredBy.DefinitionId)",
-              "project": "$(Build.TriggeredBy.ProjectID)"
+      $triggeringBuild = @{
+          "triggeringBuild" = @{
+              "id" = "$(Build.TriggeredBy.BuildId)"
+              "pipeline" = "$(Build.TriggeredBy.DefinitionId)"
+              "project" = "$(Build.TriggeredBy.ProjectID)"
           }
       }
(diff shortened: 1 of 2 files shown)


build-tools/automation/azure-pipelines-release-trigger.yaml
```diff
@@ -0,0 +1,45 @@
...
+  displayName: Upload Info from Triggering Build
+  timeoutInMinutes: 30
+  pool: VSEngSS-MicroBuild2022-1ES
+  workspace:
+    clean: all
+  steps:
+  - checkout: none
+
+  - powershell: |
+      $json = @"
```
mjbond-msft, line 27: Recommend using a dictionary instead and convert to json such as
```
$triggeringBuild = @{
    "triggeringBuild" = @{
        "id" = 5
    }
}

$json = $triggeringBuild | ConvertTo-Json -Depth 5
```

mjbond-msft: Plus assuming `Build.TriggeredBy.BuildId` is an int then also recommend leaving the quotes off as in above example

build-tools/automation/azure-pipelines-release-trigger.yaml
```diff
@@ -0,0 +1,45 @@
...
+      "@
+      New-Item -Path $(Build.StagingDirectory) -Name triggering-build.json -ItemType file -Value $json -Force
+      $jsonContent = Get-Content "$(Build.StagingDirectory)\triggering-build.json" | ConvertFrom-Json
+      Write-Host Triggering Build Info: $jsonContent.triggeringbuild.project - $jsonContent.triggeringbuild.pipeline - $jsonContent.triggeringbuild.id
+    displayName: write triggering-build.json
+
+  - task: PublishPipelineArtifact@1
+    displayName: upload triggering-build.json
+    inputs:
+      artifactName: triggering-build.json
```
mjbond-msft, line 44: `artifactName` seems a bit too specific here. Plus a bit odd having an extension name. Perhaps call it "build-info" or "triggering-build-info" in case you want include other stuff in the artifact down the road

build-tools/automation/azure-pipelines-release-trigger.yaml
pjcollins: Applied, and left the quotes to be safe.

build-tools/automation/azure-pipelines-release-trigger.yaml
pjcollins: Done

mjbond-msft (approved):

PR #6887, opened 2022-03-31
[ci] Fix variable names for release triggering build

pjcollins:
Context: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/resources-pipelines-pipeline?view=azure-pipelines#the-pipeline-resource-metadata-as-predefined-variables

Commit 7982895c added a new pipeline that can be used to trigger release
definitions, however the automatic trigger was not fully testable until
after it was merged.

Update the new pipeline to use the correct variable names that are set
when it is automatically triggered by the Xamarin.Android pipeline.

### DIFF
--- a/build-tools/automation/azure-pipelines-release-trigger.yaml
+++ b/build-tools/automation/azure-pipelines-release-trigger.yaml
@@ -26,9 +26,9 @@ jobs:
   - powershell: |
       $triggeringBuild = @{
           "triggeringBuild" = @{
-              "id" = "$(Build.TriggeredBy.BuildId)"
-              "pipeline" = "$(Build.TriggeredBy.DefinitionId)"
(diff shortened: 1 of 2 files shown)


mjbond-msft (approved):

pjcollins:
Context: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/resources-pipelines-pipeline?view=azure-pipelines#the-pipeline-resource-metadata-as-predefined-variables

Commit 7982895c added a new pipeline that can be used to trigger release
definitions, however the automatic trigger was not fully testable until
after it was merged.

Update the new pipeline to use the correct variable names that are set
when it is automatically triggered by the Xamarin.Android pipeline.

PR #6889, opened 2022-03-31
[tests] enable additional runtime tests for .NET 6

jonathanpeppers:
Context: https://github.com/dotnet/runtime/issues/35836

These tests were disabled back when some HTTP and SSL-related features weren't working yet in .NET 6.

Let's enable these now, since they appear to pass for me locally.


jonathanpeppers:
The previously ignored tests passed:

image

image

pjcollins (approved):

PR #6891, opened 2022-04-01
[tests] add linker test that makes HttpClient request

jonathanpeppers:
Context: https://github.com/dotnet/runtime/issues/67402

This is a new regression test that should catch a problem like dotnet/runtime#67402 going forward.

This code runs in a class library that gets trimmed. `Mono.Android-Tests` has many tests including usage of `HttpWebRequest`, that I believe avoids the problem.

### DIFF
--- a/tests/MSBuildDeviceIntegration/Resources/LinkDescTest/HttpClientTest.cs
+++ b/tests/MSBuildDeviceIntegration/Resources/LinkDescTest/HttpClientTest.cs
@@ -1,19 +1,24 @@
 using System;
+using System.Text;
 using System.Net.Http;
 
 public class HttpClientTest
 {
 	// [Test]
-	public static string NewHttpClient()
+	public static string Post ()
 	{
 		try
 		{
-			new HttpClient();
-			return $"[PASS] {nameof(NewHttpClient)}";
+			var client = new HttpClient ();
+			var data = new StringContent ("{\"foo\": \"bar\" }", Encoding.UTF8, "application/json");
+			var response = client.PostAsync ("https://httpbin.org/post", data).Result;
+			response.EnsureSuccessStatusCode ();
+			var json = response.Content.ReadAsStringAsync ().Result;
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
This test should fail, but I thought I would get it ready for when the above problem is fixed.

grendello:
Potentially related? https://github.com/xamarin/xamarin-android/issues/6895

tests/MSBuildDeviceIntegration/Resources/LinkDescTest/HttpClientTest.cs
```diff
@@ -4,16 +4,16 @@
 public class HttpClientTest
 {
 	// [Test]
-	public static string NewHttpClient()
+	public static string GetString()
```
eerhardt, line 7: Now that we know the issue in https://github.com/dotnet/runtime/issues/67402, it occurs to me that we should also have a "Post" or "Put" test as well. That way we ensure writing to the stream works, it takes a different enough code path that we should ensure it works.

See https://github.com/dotnet/runtime/issues/67402#issuecomment-1090305928

jonathanpeppers: I changed this to do a post and read the response back.

jonathanpeppers:
Looks like this test has an issue on main:

image

tests/MSBuildDeviceIntegration/Resources/LinkDescTest/HttpClientTest.cs
```diff
@@ -1,19 +1,24 @@
 using System;
+using System.Text;
 using System.Net.Http;
 
 public class HttpClientTest
 {
 	// [Test]
-	public static string NewHttpClient()
+	public static string Post()
```
jonpryor, line 8: If the line is going to be changed, please add a ` ` (space) between the method name and the `()`.

PR #6900, opened 2022-04-06
[build] Fix AOT no-write-symbols test

pjcollins:


pjcollins:
This change doesn't seem to be needed when using our own `ld` when LLVM is enabled (see https://github.com/xamarin/xamarin-android/pull/6901). If that approach is viable than this PR may not be needed.

Issue #6454, opened 2021-11-03
Linking OpenMP to NativeEmbeddedLibrary causes System.DllNotFoundException in Xamarin Forms Project since Android NDK 21e LTS


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Linking-OpenMP-to-NativeEmbeddedLibrary-/1507519)._

---
[severity:I'm unable to use this version] [regression] [worked-in:16.9.3]
Hi,

I use a C++ library in my Xamarin-Forms-Android-Project similar to the one described here:

[https://docs.microsoft.com/en-us/xamarin/cross-platform/cpp/](https://docs.microsoft.com/en-us/xamarin/cross-platform/cpp/).

My C++ -library uses OpenMP so I have to setup "-fopenmp" in additional linker dependencies.
Since I have upgraded from Visual Studio 16.9.3 and Android NDK 16b to Visual Studio 16.10.2, 16.10.3, 16.10.4 and also to 16.11.1 and Android NDK 21e LTS I get an System.DllNotFoundException to my NativeEmbeddedLibrary during the execution of my Xamarin Forms project (compilations runs fine) if i try to access functions of my C++ -library.

I have created a simplified sample project to reproduce the error: 

[https://github.com/DiebBlue/Cpp2XamarinTest](https://github.com/DiebBlue/Cpp2XamarinTest)

The  System.DllNotFoundException occurs, if "Cpp2XamarinTestCpp.Android" will be compiled with Linkeroption -fopenmp and a choosen Android NDK 21e in the cross-platform-settings -> C++ -> Android.

Please tell me, how can I fix this.

Greets Daniel

---
#### Feedback Bot on 8/18/2021, 11:57 PM: 

<p>We have directed your feedback to the appropriate engineering team for further evaluation. The team will review the feedback and notify you about the next steps.</p>

---
### Original Solutions
(no solutions)

dellis1972:
Can you provide a Diagnostic build log for the application and detailed `adb logcat` output so we can investigate further.
We tested the sample app and it works as expected. 

Do get the diagnostic output follow these instructions https://docs.microsoft.com/en-us/xamarin/android/troubleshooting/troubleshooting#diagnostic-msbuild-output

grendello:
[...]

PR #6901, opened 2022-04-06
[Xamarin.Android.Build.Tasks] Use binutils for Aot and LLVM

pjcollins:
It appears that an Android NDK installation is no longer needed when
using Aot with LLVM.  Projects which enable Aot and LLVM will no longer
attempt to use the NDK unless it is explicitly requested by setting
`$(AndroidNdkDirectory)` to a valid NDK path in the project file.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
@@ -75,6 +75,7 @@ namespace Xamarin.Android.Tasks
 		protected AotMode AotMode;
 		protected SequencePointsMode SequencePointsMode;
 		protected string SdkBinDirectory = "";
+		protected bool UseAndroidNdk => !string.IsNullOrWhiteSpace (AndroidNdkDirectory);
 
 		public static bool GetAndroidAotMode(string androidAotMode, out AotMode aotMode)
 		{
@@ -125,7 +126,7 @@ namespace Xamarin.Android.Tasks
 		protected string GetToolPrefix (NdkTools ndk, AndroidTargetArch arch, out int level)
 		{
 			level = 0;
-			return EnableLLVM
+			return UseAndroidNdk
 				? ndk.GetNdkToolPrefixForAOT (arch, level = GetNdkApiLevel (ndk, arch))
 				: Path.Combine (AndroidBinUtilsDirectory, $"{ndk.GetArchDirName (arch)}-");
 		}
@@ -230,7 +231,7 @@ namespace Xamarin.Android.Tasks
 				MsymPath = outdir;
 
 			string ldName;
-			if (EnableLLVM) {
+			if (UseAndroidNdk) {
 				ldName = ndk.GetToolPath (NdkToolKind.Linker, arch, level);
 				if (!string.IsNullOrEmpty (ldName)) {
 					ldName = Path.GetFileName (ldName);
@@ -250,11 +251,11 @@ namespace Xamarin.Android.Tasks
 			}
 		}
 
-		string GetLdFlags(NdkTools ndk, AndroidTargetArch arch, int level, string toolPrefix)
+		string GetLdFlags (NdkTools ndk, AndroidTargetArch arch, int level, string toolPrefix)
 		{
 			var toolchainPath = toolPrefix.Substring (0, toolPrefix.LastIndexOf (Path.DirectorySeparatorChar));
 			var ldFlags = new StringBuilder ();
-			if (EnableLLVM) {
+			if (UseAndroidNdk && EnableLLVM) {
 				string androidLibPath = string.Empty;
 				try {
 					androidLibPath = ndk.GetDirectoryPath (NdkToolchainDir.PlatformLib, arch, level);
@@ -306,7 +307,9 @@ namespace Xamarin.Android.Tasks
 				// Without the flag, `lld` will modify AOT-generated code in a way that the Mono runtime doesn't support. Until
 				// the runtime issue is fixed, we need to pass this flag then.
 				//
-				ldFlags.Append ("--no-relax");
+				if (!UseAndroidNdk) {
+					ldFlags.Append ("--no-relax");
+				}
 			}
 
 			if (StripLibraries) {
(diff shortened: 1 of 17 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
```diff
@@ -228,120 +226,23 @@ protected void GetAotOptions (NdkTools ndk, AndroidTargetArch arch, int level, s
...
-				libs.Add (Path.Combine (androidLibPath, "libc.so"));
-				libs.Add (Path.Combine (androidLibPath, "libm.so"));
-
-				ldFlags = $"\\\"{string.Join ("\\\";\\\"", libs)}\\\"";
-			}
-
-			if (!StripLibraries) {
-				return ldFlags;
-			}
-
```
pjcollins, line 300: @grendello what are your thoughts on this? Do you remember why we needed to pass these extra lib paths from the NDK, and do you think they would still be needed if we use the `ld` that we now bundle and ship?

I tested template project with AOT+LLVM on a few devices and emulators (including an armeabi Nexux 5 running Android 6.0), and I haven't hit any issues yet...

grendello: As far as I recall, the extra lib paths were for NDK versions which used `gcc`, as we needed to link private gcc libraries.  For the LLVM-based NDK we probably no longer need it, and our native toolchain doesn't need it too.  The only reason to keep them, I think, is to keep support for old NDK versions installed on user machines (we had an instance when somebody still used NDK r16). Other than that, I **think** they can be safely removed.

pjcollins: Should we even continue to allow usage of an arbitrary Android NDK for our AOT tasks if our installers now contain all of the required tools?

grendello: We have no influence on what version of NDK is installed on the user's machine, so we should support as many of them as possible

pjcollins: I guess that's my question, is there a strong reason to allow using an external NDK instead of our toolchain at this point in time?

grendello: AOT tasks should work just fine with our toolchain, it would be good to make that non-optional. Otherwise, we need to keep allowing NDK because of projects which include native code that requires the full C/C++ toolchain.

jonpryor: Rationale for why we may need to support older NDK tooling, e.g. https://github.com/xamarin/xamarin-android/issues/6454

grendello: By default, yes, we should use our embedded toolchain, ignoring `$(AndroidNdkDirectory)`. But we must make it possible for people to use their NDK if they need to because of compiler bugs, toolchain version requirements (for instance for reproducible builds). I can't find another issue we had when someone had to use NDK r16b because newer versions broke some code for them. For these reasons, even though I'd personally prefer to rely only on our embedded toolchain, we need to keep external NDK support in some form.

grendello (approved):

jonathanpeppers (approved):
After we merge this, is it worth making the lookup of the NDK in xamarin-android-tools optional?

I haven't measured, but surely some amount of milliseconds probing for NDKs, and we could skip that now?

jonathanpeppers:
Oh, I see the `do-not-merge` label now. Is there an issue?

pjcollins:
> Oh, I see the do-not-merge label now. Is there an issue?

I think there are still a couple of test failures using the latest Android NDK when LLVM is disabled, but I wanted to open this up for review on the general approach of the changes. I'm wondering if we should only allow AOT to use the Android NDK when LLVM is enabled, as that is how things work today?

> After we merge this, is it worth making the lookup of the NDK in xamarin-android-tools optional?

I'm not sure we want to do this quite yet, as `$(_AndroidNdkDirectory)` is still used by `BuildApk` (required when `$(_AndroidCheckedBuild)` is true), and `MakeBundleNativeCodeExternal`.

jonathanpeppers:
Right now I don't think you can use LLVM without having an NDK -- but I was thinking we could fix that? Like it should be possible to work?

If you didn't need the NDK, we could enable LLVM by default for `Release` builds to improve startup times.

pjcollins:
The latest test failures appear to be ignorable. I spent some more time testing this today but am blocked on LLVM configurations with .NET 7 due to a known crash. We may want to consider holding this until we can further test LLVM with .NET 7.

> Right now I don't think you can use LLVM without having an NDK -- but I was thinking we could fix that? Like it should be possible to work?

@jonathanpeppers do you happen to remember a scenario where LLVM relied on the NDK? When previously testing this against .NET 6 and classic LLVM was working for me without the NDK at least in some simple project template cases.

pjcollins:
There are a couple of bugs to revisit in the future once we have the latest 7.0.100 RC builds from runtime, but this should be good to go if CI looks good.

src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
```diff
@@ -306,7 +307,9 @@ string GetLdFlags(NdkTools ndk, AndroidTargetArch arch, int level, string toolPr
 				// Without the flag, `lld` will modify AOT-generated code in a way that the Mono runtime doesn't support. Until
 				// the runtime issue is fixed, we need to pass this flag then.
 				//
-				ldFlags.Append ("--no-relax");
+				if (!UseAndroidNdk) {
```
grendello, line 310: I wonder if (at some point, not necessarily in this PR) we could detect the version of LLVM in the NDK and append `--no-relax` if that version is >= 14

grendello (approved):

TomSoPolaris:
Hi, we're hoping this fixes https://github.com/xamarin/xamarin-android/issues/7352. Has this been released yet? I'm not good at knowing how issues make it into releases and get shipped. Would appreciate some insight.
We had to disable LLVM on our Android app store builds or else they'd insta-crash on launch.

PR #6904, opened 2022-04-07
[ci] Remove extra .NET SDKs on Windows and Linux

pjcollins:
Related: https://github.com/xamarin/xamarin-android/pull/6598

Our Windows and Linux pools are "stateful" and are only wiped fresh
every 7 or 3 days respectively.  This allows them to service jobs more
quickly when demand is high.  Now that we are starting to build and test
against .NET 7 previews we should enable .NET SDK clean up, as these
pools will be shared across .NET 6 and .NET 7 branches.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -151,6 +151,13 @@ stages:
       displayName: set JI_JAVA_HOME
 
     - template: yaml-templates\use-dot-net.yaml
+      parameters:
+        remove_dotnet: true
+
+    # xabuild still depends on .NET Core 3 or earlier
+    - template: yaml-templates\use-dot-net.yaml
+      parameters:
+        version: 3.1.417
 
     # Downgrade the XA .vsix installed into the instance of VS that we are building with so that we don't restore/build against a test version.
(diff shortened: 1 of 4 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -151,6 +151,13 @@ stages:
       displayName: set JI_JAVA_HOME
 
     - template: yaml-templates\use-dot-net.yaml
+      parameters:
+        remove_dotnet: true
+
+    # xabuild still depends on .NET Core 3 or earlier
```
jpobst, line 157: It might be worth using the `UseDotNet` task here, as I think it can pull from a cache?

jpobst (approved):
LGTM!

grendello (approved):

dellis1972 (approved):

build-tools/automation/azure-pipelines.yaml
pjcollins: It looks like this may only work if you install into `$(Agent.ToolsDirectory)`. I think I'll stick with our template as it also includes some retries, and this extra step only took 21 seconds in the PR run.

PR #6916, opened 2022-04-12
[ci] Sign vsix in separate job, add CodesignVerify task

pjcollins:
When looking at the [Xamarin.Android pipeline analytics][0] I noticed
that the Prepare Classic Release stage accounted for a very large
percentage of our overall failures.

image

To help address this I've moved the vsix signing step into a new job
that runs on a Windows agent, where the signing tasks are hopefully more
resilient.  This seems to have sped up signing performance dramatically
as well, as a step that would take upwards of an hour or more on macOS
now seems to be completing in less than 10 minutes on the Windows
MicroBuild agent pool.

I've also added the `MicroBuildCodesignVerify` task in a couple of
places to validate the signing status of the vsix and msi files we ship.

[0]: https://devdiv.visualstudio.com/DevDiv/_pipeline/analytics/stageawareoutcome?definitionId=11410&contextType=build

### DIFF
--- a/build-tools/installers/sign-content.proj
+++ b/build-tools/installers/sign-content.proj
@@ -30,7 +30,7 @@ ourself (using an empty signing identity) before passing these files to ESRP.
     </PackageReference>
   </ItemGroup>
 
-  <Target Name="_AddMachOEntitlements" >
+  <Target Name="AddMachOEntitlements" >
     <Exec Command="codesign -vvvv -f -s - -o runtime --entitlements &quot;%(_MSBuildFilesUnixSignAndHarden.EntitlementsPath)&quot; &quot;%(_MSBuildFilesUnixSignAndHarden.Identity)&quot;" />
     <ItemGroup>
       <FilesToSign Include="@(_MSBuildFilesUnixSign)">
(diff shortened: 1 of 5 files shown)


pjcollins:
Latest test run was successful: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6026033&view=logs&j=7ad108a5-ebcd-5edf-b52d-10cb155700f6

build-tools/installers/sign-content.proj
```diff
@@ -82,15 +82,20 @@ ourself (using an empty signing identity) before passing these files to ESRP.
...
-  <Target Name="_AddClassicInstallers" >
+  <Target Name="_SignVsix" >
+    <ItemGroup>
+      <FilesToSign Include="$(OutDir)\Xamarin.Android.Sdk-*.vsix">
+        <Authenticode>VsixSHA2</Authenticode>
+      </FilesToSign>
+    </ItemGroup>
+  </Target>
+
+  <Target Name="_SignPkg" >
```
jonathanpeppers, line 93: nit: It seems like we shouldn't prefix a target with `_` if we call it from the command-line. It feels like it is more "public" if we invoke them that way?

jonathanpeppers (approved):

build-tools/installers/sign-content.proj
pjcollins: Makes sense and should make it easier to type locally whenever needed, I removed the leading `_` from the targets in this file.

PR #6922, opened 2022-04-13
[tests] fix DotNetBuild tests

jonathanpeppers:
Some tests are failing on xamarin-android/main, such as:

    bin\Release\net6.0-android\com.xamarin.dotnetbuild-Signed.apk should contain assemblies/System.Linq.dll
    Expected: True
    But was:  False

Trying locally, this fails because:

1. It is a `Debug` build
2. Fast Deployment is off

These only fail on the Windows lane because it doesn't include our
internal bits from xamarin/monodroid. If Fast Deployment was used (as
on other test lanes), then this assertion would get skipped.

To solve this problem, let's just assert against a *different* file.
If `System.Linq.dll` is sometimes arhitecture-specific, let's just
assert `Mono.Android.dll` instead. The tests now pass.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
@@ -634,7 +634,7 @@ namespace Xamarin.Android.Build.Tests
 			var helper = new ArchiveAssemblyHelper (apkPath, usesAssemblyStore, rids);
 			helper.AssertContainsEntry ($"assemblies/{proj.ProjectName}.dll", shouldContainEntry: expectEmbeddedAssembies);
 			helper.AssertContainsEntry ($"assemblies/{proj.ProjectName}.pdb", shouldContainEntry: !CommercialBuildAvailable && !isRelease);
-			helper.AssertContainsEntry ($"assemblies/System.Linq.dll",        shouldContainEntry: expectEmbeddedAssembies);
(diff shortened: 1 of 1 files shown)


pjcollins:
I've also seen this intermittently on macOS test jobs that have the commercial bits, and it's puzzling to me that this doesn't fail consistently on the Windows build + smoke test either. How could `System.Linq.dll` _sometimes_ be architecture specific, and not consistently across multiple build/test attempts?

Maybe we need to take a closer look at https://github.com/xamarin/xamarin-android/blob/ee4322054292a8f6e33ca6c53fe9ce609fce3c95/src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs#L121-L171?

jonathanpeppers:
@pjcollins do you have an example where it failed on the other test lanes?

The Windows lane made sense to me, because I could reproduce it locally and `expectEmbeddedAssembies` is `true` in this test when monodroid isn't present.

pjcollins:
Here's some from the PR I just merged to main today https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6019245&view=ms.vss-test-web.build-test-results-tab&runId=37961400&resultId=100116&paneView=debug

jonathanpeppers:
@pjcollins there is something else weird here, but I think it's our test.

The cases where macOS failed, it seems like `usesAssemblyStore=false`. So how would this class know to check if the files exist in that case?

https://github.com/xamarin/xamarin-android/blob/968c4245d3bb24670f402eb207109ca837fdd195/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs#L634-L638

Let's look into this again if it fails like this later...

Issue #6895, opened 2022-04-05
Using HttpClient in a Xamarin Android Release app crashes


eerhardt:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

Maui main

### Description

I'm getting a crashing app when using HttpClient in a Release build. It appears that the `FromJniHandle` method is getting trimmed in `JavaList<T>`, causing the app to crash.

### Steps to Reproduce

1. `dotnet new android`
2. Write the following code in the MainActivity:

```C#
namespace XamarinAndroid;

[Activity(Label = "@string/app_name", MainLauncher = true)]
public class MainActivity : Activity
{
    protected override async void OnCreate(Bundle? savedInstanceState)
    {
        base.OnCreate(savedInstanceState);

        // Set our view from the "main" layout resource
        SetContentView(Resource.Layout.activity_main);

        HttpClient client = new HttpClient();
        client.BaseAddress = new Uri("https://www.bing.com");
        string results = await client.GetStringAsync("search?q=maui");


        AlertDialog.Builder dialog = new AlertDialog.Builder(this);
        var alert = dialog.Create();
        alert.SetTitle($"results Length: {results.Length}");

        alert.Show();
    }
}
```

3. Run the app in Release mode.

### Expected results

The app should work the same as it does in Debug mode.

### Actual results

The app crashes. See the below logcat output.

### Did you find any workaround?

You can manually preserve the method by adding a `<TrimmerRootDescriptor Include="ILLInk.Descriptors.xml" />` with contents:

```xml
<linker>
  <assembly fullname="Mono.Android">
    <type fullname="Android.Runtime.JavaList`1">
      <method name="FromJniHandle" />
    </type>
  </assembly>
</linker>
```

### Relevant log output

```shell
04-05 11:28:40.105  9171  9171 D AndroidRuntime: Shutting down VM
04-05 11:28:40.105  9171  9171 E AndroidRuntime: FATAL EXCEPTION: main
04-05 11:28:40.105  9171  9171 E AndroidRuntime: Process: com.companyname.XamarinAndroid, PID: 9171
04-05 11:28:40.105  9171  9171 E AndroidRuntime: android.runtime.JavaProxyThrowable: System.ArgumentNullException: ArgumentNull_Generic Arg_ParamName_Name, method
04-05 11:28:40.105  9171  9171 E AndroidRuntime:    at System.Delegate.CreateDelegate(Type , Object , MethodInfo , Boolean , Boolean )
[...]

PR #6923, opened 2022-04-13
[Mono.Android] preserve public methods for GetJniHandleConverterForType

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/6895

In some cases an app with a single `HttpClient` request can crash with:

    android.runtime.JavaProxyThrowable: System.ArgumentNullException: ArgumentNull_Generic Arg_ParamName_Name, method
    at System.Delegate.CreateDelegate(Type , Object , MethodInfo , Boolean , Boolean )
    at System.Delegate.CreateDelegate(Type , MethodInfo , Boolean )
    at System.Delegate.CreateDelegate(Type , MethodInfo )
    at Java.Interop.JavaConvert.GetJniHandleConverterForType(Type )
    at Java.Interop.JavaConvert.GetJniHandleConverter(Type )
    at Java.Interop.JavaConvert.FromJniHandle[IList`1](IntPtr , JniHandleOwnership , Boolean& )
    at Java.Interop.JavaConvert.FromJniHandle[IList`1](IntPtr , JniHandleOwnership )
    at Android.Runtime.JavaDictionary`2[[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Collections.Generic.IList`1[[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].Get(String )
    at Android.Runtime.JavaDictionary`2[[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.Collections.Generic.IList`1[[System.String, System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]], System.Private.CoreLib, Version=6.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]].get_Item(String )
    at Xamarin.Android.Net.AndroidMessageHandler.CopyHeaders(HttpURLConnection , HttpResponseMessage )
    at Xamarin.Android.Net.AndroidMessageHandler.DoProcessRequest(HttpRequestMessage , URL , HttpURLConnection , CancellationToken , RequestRedirectionState )
    at Xamarin.Android.Net.AndroidMessageHandler.SendAsync(HttpRequestMessage , CancellationToken )
    at System.Net.Http.HttpClient.GetStringAsyncCore(HttpRequestMessage , CancellationToken )
    at XamarinAndroid.MainActivity.OnCreate(Bundle savedInstanceState)
    at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object )
    at Android.App.SyncContext.<>c__DisplayClass2_0.<Post>b__0()
    at Java.Lang.Thread.RunnableImplementor.Run()
    at Java.Lang.IRunnableInvoker.n_Run(IntPtr , IntPtr )
    at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V(_JniMarshal_PP_V , IntPtr , IntPtr )
    at mono.java.lang.RunnableImplementor.n_run(Native Method)
    at mono.java.lang.RunnableImplementor.run(RunnableImplementor.java:30)
    at android.os.Handler.handleCallback(Handler.java:938)
    at android.os.Handler.dispatchMessage(Handler.java:99)
    at android.os.Looper.loopOnce(Looper.java:201)
    at android.os.Looper.loop(Looper.java:288)
    at android.app.ActivityThread.main(ActivityThread.java:7870)
    at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
    at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)

In this case, `JavaDictionary<,>.FromJniHandle()` was linked away, and
we access it via reflection:

https://github.com/xamarin/xamarin-android/blob/9f3ee589d9a8cf9eb0d9d0dbe6c6e6acc8fc7f96/src/Mono.Android/Java.Interop/JavaConvert.cs#L87-L92

This works in Xamarin.Android due to:

https://github.com/xamarin/xamarin-android/blob/0ab3db079d43d47ad1bc522e28e65204542fea60/src/Xamarin.Android.Build.Tasks/Linker/PreserveLists/Mono.Android.xml#L25-L28

We actually get a linker warning for this:

    src\Mono.Android\Java.Interop\JavaConvert.cs(89,4):
    warning IL2070: Java.Interop.JavaConvert.GetJniHandleConverterForType(Type): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicMethods' in call to 'System.Type.GetMethod(String,BindingFlags)'. The parameter 't' of method 'Java.Interop.JavaConvert.GetJniHandleConverterForType(Type)' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to.

Adding `[DynamicallyAccessedMembers]` preserves these methods, and the
warning goes away. This seems like a safe option, as we shouldn't need
to keep updating a list of types like we did in Xamarin.Android.

I tried updating `InstallAndRunTests.CustomLinkDescriptionPreserve()`,
but some usage of types is preserving `FromJniHandle()`. I couldn't
get the test to fail...

Long term, we should get the linker warnings to zero and make our
build fail if new warnings are ever introduced. Tracking this in:

https://github.com/xamarin/xamarin-android/issues/5652

### DIFF
--- a/src-ThirdParty/System.Diagnostics.CodeAnalysis/DynamicallyAccessedMembersAttribute.cs
+++ b/src-ThirdParty/System.Diagnostics.CodeAnalysis/DynamicallyAccessedMembersAttribute.cs
@@ -1,0 +1,51 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+
+namespace System.Diagnostics.CodeAnalysis
+{
+    /// <summary>
+    /// Indicates that certain members on a specified <see cref="Type"/> are accessed dynamically,
+    /// for example through <see cref="System.Reflection"/>.
+    /// </summary>
+    /// <remarks>
+    /// This allows tools to understand which members are being accessed during the execution
+    /// of a program.
+    ///
+    /// This attribute is valid on members whose type is <see cref="Type"/> or <see cref="string"/>.
+    ///
+    /// When this attribute is applied to a location of type <see cref="string"/>, the assumption is
+    /// that the string represents a fully qualified type name.
+    ///
+    /// When this attribute is applied to a class, interface, or struct, the members specified
+    /// can be accessed dynamically on <see cref="Type"/> instances returned from calling
+    /// <see cref="object.GetType"/> on instances of that class, interface, or struct.
+    ///
+    /// If the attribute is applied to a method it's treated as a special case and it implies
+    /// the attribute should be applied to the "this" parameter of the method. As such the attribute
+    /// should only be used on instance methods of types assignable to System.Type (or string, but no methods
+    /// will use it there).
+    /// </remarks>
+    [AttributeUsage(
+        AttributeTargets.Field | AttributeTargets.ReturnValue | AttributeTargets.GenericParameter |
+        AttributeTargets.Parameter | AttributeTargets.Property | AttributeTargets.Method |
+        AttributeTargets.Class | AttributeTargets.Interface | AttributeTargets.Struct,
+        Inherited = false)]
+    internal sealed class DynamicallyAccessedMembersAttribute : Attribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DynamicallyAccessedMembersAttribute"/> class
+        /// with the specified member types.
+        /// </summary>
+        /// <param name="memberTypes">The types of members dynamically accessed.</param>
+        public DynamicallyAccessedMembersAttribute(DynamicallyAccessedMemberTypes memberTypes)
+        {
+            MemberTypes = memberTypes;
+        }
+ 
+        /// <summary>
+        /// Gets the <see cref="DynamicallyAccessedMemberTypes"/> which specifies the type
+        /// of members dynamically accessed.
+        /// </summary>
+        public DynamicallyAccessedMemberTypes MemberTypes { get; }
+    }
+}
--- a/src/Mono.Android/Java.Interop/JavaConvert.cs
+++ b/src/Mono.Android/Java.Interop/JavaConvert.cs
@@ -84,7 +84,7 @@ namespace Java.Interop {
 			return null;
 		}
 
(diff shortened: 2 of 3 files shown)


jonpryor:
@jonathanpeppers : what I'd like is an elaboration on:

> Adding `[DynamicallyAccessedMembers]` preserves these methods

*How* does [`[DynamicallyAccessedMembers(DynamicallyAccessedMemberTypes.PublicMethods)]`](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.dynamicallyaccessedmembersattribute?view=net-6.0) cause `JavaDictionary<,>.FromJniHandle()` to be preserved?  Does it cause the linker to look at the `Type` values for all calls to `GetJniHandleConverterForType()`, thus finding `typeof(JavaDictionary<,>).MakeGenericType(target.GetGenericArguments())`?  (This *sounds* like it's the case, based on the docs, but that feels like some form of *wizardry*.)

jonathanpeppers:
@jonpryor yes, it's my understanding that in .NET 6 the linker has some static analysis that can tell which `System.Type` are passed into a method. There could be bugs there, but I've heard that if you got a warning and then solved the warning, that you're in good shape?

eerhardt:
> This sounds like it's the case, based on the docs, but that feels like some form of wizardry.

It sort of is (at least to me). But if you want to read deeper about it, check out:

* https://github.com/dotnet/linker/blob/main/docs/design/reflection-flow.md
* https://github.com/dotnet/designs/blob/main/accepted/2020/linking-libraries.md

Which has the basic designs for how these things work.

PR #6929, opened 2022-04-15
[Xamarin.Android.Build.Tasks] fix --runtime and r8 at the same time

jonathanpeppers:
If you do:

    dotnet new android
    dotnet build -c Release -r android-arm64 -p:AndroidLinkTool=r8

You'll get the warning:

    warning XA4304: ProGuard configuration file 'obj\Release\android-arm64\android-arm64\proguard\proguard_project_references.cfg' was not found.

That path has two `android-arm64` in it!

Additionally, I found this problem was hidden in our testing
infrastructure because we set `$(AppendRuntimeIdentifierToOutputPath)`
to `false` in several tests... After removing this, I could write a
test that reproduces this issue.

The fix for the issue is that `$(_InnerIntermediateOutputPath)` needs
to check if the outer build has a single `$(RuntimeIdentifier)`. If
so, we don't need to append a RID in that path.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetXamarinProject.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetXamarinProject.cs
@@ -29,7 +29,6 @@ namespace Xamarin.ProjectTools
 				SetProperty ("TargetFramework", "net6.0-android");
 				SetProperty ("EnableDefaultItems", "false");
 				SetProperty ("AppendTargetFrameworkToOutputPath", "false");
-				SetProperty ("AppendRuntimeIdentifierToOutputPath", "false");
 			} else {
 				AddReferences ("System"); // default
 				SetProperty ("Platform", "AnyCPU", "'$(Platform)' == ''");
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/DotNetXamarinProject.cs
```diff
@@ -29,7 +29,6 @@ protected DotNetXamarinProject (string debugConfigurationName = "Debug", string
 				SetProperty ("TargetFramework", "net6.0-android");
 				SetProperty ("EnableDefaultItems", "false");
 				SetProperty ("AppendTargetFrameworkToOutputPath", "false");
-				SetProperty ("AppendRuntimeIdentifierToOutputPath", "false");
```
jonathanpeppers, line 32: I suspect this will break random tests, and I'll need to update them. We'll see.

jonpryor:
```markdown
If you do:

	dotnet new android
	dotnet build -c Release -r android-arm64 -p:AndroidLinkTool=r8

You'll get the warning:

	warning XA4304: ProGuard configuration file
	  'obj\Release\android-arm64\android-arm64\proguard\proguard_project_references.cfg'
	  was not found.

That path has two `android-arm64`s in it!

Additionally, I found this problem was hidden by our testing
infrastructure because we set
`$(AppendRuntimeIdentifierToOutputPath)`=False in several tests.
After removing this, I could write a test that reproduces this issue.

The fix for the issue is that `$(_InnerIntermediateOutputPath)` needs
to check if the outer build has a single `$(RuntimeIdentifier)`.
If so, we don't need to append a RID in that path.

Updating the tests to no longer set
`$(AppendRuntimeIdentifierToOutputPath)` at all resulted in some unit
test failures, as the tests were asserting the existence of BCL
assemblies such as `System.Console.dll` in an ABI-invariant location,
and leaving `$(AppendRuntimeIdentifierToOutputPath)` unset would
result in these assemblies becoming ABI-specific.  Change the assert
to instead check `System.Private.CoreLib.dll`, as
`System.Private.CoreLib.dll` is *always* ABI-specific.
```

PR #6942, opened 2022-04-20
Go back to LLVM 13 toolchain

grendello:


grendello:
PR with the fix: https://github.com/xamarin/xamarin-android/pull/6946

PR #6946, opened 2022-04-20
[AOT] Fix for a compatibility issue between AOT and LLVM 14

grendello:
Context: https://releases.llvm.org/14.0.0/tools/lld/docs/ReleaseNotes.html#elf-improvements
Context: 16bbf6f0b0cd49f96f73197d28e8faf31aa9e499

After switching to LLVM 14 we noticed that all `arm64` apps which use
AOT, crash with the following assertion from the Mono runtime:

    * Assertion at /__w/1/s/src/mono/mono/mini/tramp-arm64.c:53, condition `((ins >> 24) & 0x1f) == 0x10' not met

After investigation, it turns out that one of the LLVM 14 changes for
the `aarch64` architecture in the `lld` linker was an optimization which
modifies generated code in the following fashion (`-` is the "good"
code, `+` is the "bad" one):

     -   13a44: 30 01 00 b0          adrp    x16, 0x38000 <.text+0x148>
     -   13a48: 10 e2 3e 91          add     x16, x16, #4024         // =4024
     +   13a44: 1f 20 03 d5          nop
     +   13a48: 90 ab 12 10          adr     x16, #152944

Mono AOT generates a pair of instructions, `adrp` and `add` and, at the
runtime, it expects to find the exact code that was generated (runtime
disassembles the binary code and looks for specific patterns), but
instead it finds a `nop` and an `adr` instructions, which eventually
triggers the above assertion.

Until Mono runtime code is fixed, we can work around this issue by
passing the `--no-relax` flag to the native linker.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
@@ -3,6 +3,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Text;
 using Java.Interop.Tools.Diagnostics;
 using Microsoft.Android.Build.Tasks;
 using Microsoft.Build.Framework;
@@ -252,7 +253,7 @@ namespace Xamarin.Android.Tasks
 		string GetLdFlags(NdkTools ndk, AndroidTargetArch arch, int level, string toolPrefix)
 		{
 			var toolchainPath = toolPrefix.Substring (0, toolPrefix.LastIndexOf (Path.DirectorySeparatorChar));
-			var ldFlags = string.Empty;
+			var ldFlags = new StringBuilder ();
 			if (EnableLLVM) {
 				string androidLibPath = string.Empty;
 				try {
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
```diff
@@ -291,19 +292,27 @@ string GetLdFlags(NdkTools ndk, AndroidTargetArch arch, int level, string toolPr
...
-				return StripFlag;
+			//
+			// This flag is needed for Mono AOT to work correctly with the LLVM 14 `lld` linker due to the following change:
+			//
+			//   The AArch64 port now supports adrp+ldr and adrp+add optimizations. --no-relax can suppress the optimization.
+			//
+			// Without the flag, `lld` will modify AOT-generated code in a way that the Mono runtime doesn't support. Until
+			// the runtime issue is fixed, we need to pass this flag then.
+			//
+			ldFlags.Append ("--no-relax");
```
jonathanpeppers, line 310: Do we need an MSBuild property to turn this off? It could be private like `$(_AndroidAotNoRelax)` that defaults to `true`.

grendello: @jonathanpeppers I wouldn't bother. We can just remove the code when we move to fixed Mono AOT

grendello:
The failing tests also fail in `main`, so I think the PR is ready

dellis1972 (approved):
This looks good

A thought... we do testing on an x86-64 Emulator. I am wondering if we should really start looking at running some tests on arm64 somehow... since that is what most devices will be running once an app hits the store.  Even if it is a custom qemu setup.

grendello:
> This looks good
> 
> A thought... we do testing on an x86-64 Emulator. I am wondering if we should really start looking at running some tests on arm64 somehow... since that is what most devices will be running once an app hits the store. Even if it is a custom qemu setup.

I would love to, but even though Android SDK ships with both the `aarch64` emulator and system images, I wasn't able to launch it on an `x86_64` host (AVD created with Android Studio's AVD Manager).

grendello:
> Will we get ARM64 build machines on Azure DevOps at some point? They work on Apple M1.

It would be great, unfortunately I have no ideea if we're getting ones :(

PR #6947, opened 2022-04-20
[build] Use variables for `net6.0` where possible

pjcollins:
Context: https://github.com/xamarin/xamarin-android/pull/6598

In preparation for .NET 7 we should remove as many hard-coded instances
of `net6.0` as possible.  I've cherry-picked a handful of @jonathanpeppers 
commits over to simplify that diff and to apply these generic
changes sooner than later.  New variables, MSBuild properties, and YAML
parameters are now used in place of explicit `net6.0` mentions.  Some
conditions have also been flipped to check against MonoAndroid rather
than net6.0.

### DIFF
--- a/eng/Versions.props
+++ b/eng/Versions.props
@@ -12,6 +12,7 @@
   <PropertyGroup>
     <!-- Match the first three version numbers and append 00 -->
     <DotNetPreviewVersionBand Condition=" '$(DotNetPreviewVersionBand)' == '' ">$([System.Text.RegularExpressions.Regex]::Match($(MicrosoftDotnetSdkInternalPackageVersion), `^\d+\.\d+\.\d`))00</DotNetPreviewVersionBand>
+    <DotNetSdkManifestsFolder>$(DotNetPreviewVersionBand)$([System.Text.RegularExpressions.Regex]::Match($(MicrosoftDotnetSdkInternalPackageVersion), `\-preview.\d+`))</DotNetSdkManifestsFolder>
     <!-- NOTE: sometimes we hardcode these when transitioning to new version bands -->
     <DotNetAndroidManifestVersionBand>$(DotNetPreviewVersionBand)</DotNetAndroidManifestVersionBand>
     <DotNetMonoManifestVersionBand>$(DotNetPreviewVersionBand)</DotNetMonoManifestVersionBand>
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
@@ -233,7 +233,7 @@ namespace Xamarin.Android.Build.Tests
 		[TestCaseSource (nameof (DotNetPackTargetFrameworks))]
 		public void DotNetPack (string dotnetVersion, string platform, int apiLevel)
 		{
-			var targetFramework = dotnetVersion + platform;
+			var targetFramework = $"{dotnetVersion}-{platform}";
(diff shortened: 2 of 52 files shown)


eng/Versions.props
```diff
@@ -12,6 +12,7 @@
   <PropertyGroup>
     <!-- Match the first three version numbers and append 00 -->
     <DotNetPreviewVersionBand Condition=" '$(DotNetPreviewVersionBand)' == '' ">$([System.Text.RegularExpressions.Regex]::Match($(MicrosoftDotnetSdkInternalPackageVersion), `^\d+\.\d+\.\d`))00</DotNetPreviewVersionBand>
+    <DotNetSdkManifestsFolder>$(DotNetPreviewVersionBand)$([System.Text.RegularExpressions.Regex]::Match($(MicrosoftDotnetSdkInternalPackageVersion), `\-preview.\d+`))</DotNetSdkManifestsFolder>
```
jonathanpeppers, line 15: So this expression is meant to match, `7.0.100-preview.4.22219.19`, and we want this to come up with `7.0.100-preview.4`.

What happens in `main` where we have `6.0.300-rtm.22214.7`? Should we fallback to `$(DotNetPreviewVersionBand)` if the Regex doesn't match?

pjcollins: Let me test this part, I'll add the fallback to fix this up as needed.

pjcollins: I think this works fine as is, if there is no `-preview.` string in `$(MicrosoftDotnetSdkInternalPackageVersion)` the match will be empty and `$(DotNetSdkManifestsFolder)` will just be `$(DotNetPreviewVersionBand)`.

grendello (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
```diff
@@ -264,8 +267,8 @@ public void DotNetPack (string targetFramework, int apiLevel)
 			var nupkgPath = Path.Combine (FullProjectDirectory, proj.OutputPath, "..", $"{proj.ProjectName}.1.0.0.nupkg");
 			FileAssert.Exists (nupkgPath);
 			using (var nupkg = ZipHelper.OpenZip (nupkgPath)) {
-				nupkg.AssertContainsEntry (nupkgPath, $"lib/net6.0-android{apiLevel}.0/{proj.ProjectName}.dll");
-				nupkg.AssertContainsEntry (nupkgPath, $"lib/net6.0-android{apiLevel}.0/{proj.ProjectName}.aar");
+				nupkg.AssertContainsEntry (nupkgPath, $"lib/{dotnetVersion}-android{apiLevel}.0/{proj.ProjectName}.dll");
```
dellis1972, line 270: Do we need to use the `{platform}` argment to replace the `-android` bits here?

jonathanpeppers: I think NuGet always puts `31.0` for all of `net6.0-android`, `net6.0-android31`, `net6.0-android31.0`.

jonathanpeppers (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
pjcollins: I think we can keep it like this in case we ever want to use a TF like `net6.0-android32.0` as a test case source. Using the API level instead of platform lets us always append that trailing `.0` without having to check for it first.

Issue #6626, opened 2022-01-08
Using the default TrimMode=link breaks FirstChanceException handler support and crashes app


lauxjpn:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

.NET 6.0.100

### Description

When registering an `AppDomain.CurrentDomain.FirstChanceException` handler, the handler works fine with non-trimmed builds.

With the default `TrimMode=link` option when using `PublishTrimmed=true`, the app crashes at runtime (presumable when the handler gets called for the first time).

I also tried the following, but the crash still happens at runtime:

```
<Target Name="ConfigureTrimming"
        BeforeTargets="PrepareForILLink">
    <ItemGroup>
        <ManagedAssemblyToLink Condition="'%(Filename)' == 'System.Runtime'">
            <IsTrimmable>false</IsTrimmable>
        </ManagedAssemblyToLink>
    </ItemGroup>
</Target>
```

---

I wanted to use a `FirstChanceException` handler to break into the code when an exception because of another trimmed method occurs, so that I could take a look at the call stack and the parameters of a `System.Delegate.CreateDelegate(Type type, Type target, String method)` call, so that I could investigate which target method got trimmed in https://github.com/xamarin/xamarin-android/issues/6612#issuecomment-1006145047.

Unfortunately, since the `TrimMode=link` also trims away the `FirstChanceException` handler support, I can't use it for that purpose.

### Steps to Reproduce

See description.

jonathanpeppers:
This example in a `dotnet new android` app works for me:
```csharp
[Activity(Label = "@string/app_name", MainLauncher = true)]
public class MainActivity : Activity
{
    protected override void OnCreate(Bundle? savedInstanceState)
    {
        base.OnCreate(savedInstanceState);

        // Set our view from the "main" layout resource
        SetContentView(Resource.Layout.activity_main);

        AppDomain.CurrentDomain.FirstChanceException += (sender, e) =>
        {
            Console.WriteLine("First change exception!");
        };
    }

    public override void OnBackPressed() => Task.Factory.StartNew(() => throw new Exception("foo"));
}
```

However, if I do this, then `FirstChanceException` doesn't fire:
```csharp
public override void OnBackPressed() => throw new Exception("foo");
```

jonathanpeppers:
I think this behavior is correct for Debug builds ^^

The problem is Release builds (that are trimmed) crash with:
```
[...]

PR #6953, opened 2022-04-21
Workaround FirstChanceExceptionEventArgs being trimmed

akoeplinger:
Fixes: https://github.com/xamarin/xamarin-android/issues/6626

### DIFF
--- a/src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Private.CoreLib.xml
+++ b/src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Private.CoreLib.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<linker>
+	<!-- Workaround for https://github.com/xamarin/xamarin-android/issues/6626, remove once https://github.com/dotnet/runtime/pull/68265 lands -->
+	<assembly fullname="System.Private.CoreLib">
+		<type fullname="System.Runtime.ExceptionServices.FirstChanceExceptionEventArgs">
+			<method signature="System.Void .ctor(System.Exception)" />
+		</type>
+	</assembly>
+</linker>


src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Private.CoreLib.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<linker>
+	<!-- Workaround for https://github.com/xamarin/xamarin-android/issues/6626, remove once https://github.com/dotnet/runtime/pull/68265 lands -->
+	<assembly fullname="System.Private.CoreLib">
+		<type fullname="System.Runtime.ExceptionServices.FirstChanceExceptionEventArgs">
+			<method signature="System.Void .ctor(System.Exception)" />
```
jonathanpeppers, line 6: This looks like the right names: https://source.dot.net/#System.Private.CoreLib/FirstChanceExceptionEventArgs.cs,9

jonathanpeppers (approved):

src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Private.CoreLib.xml
jonathanpeppers: I think it's:

https://github.com/dotnet/runtime/blob/6c7ce85e149eaaf4cc452a7008754d5c799db938/src/mono/mono/metadata/object.c#L94

jonpryor:
Given that dotnet/runtime#68235 has been merged and backported, do we *really* need this fix?

src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Private.CoreLib.xml
akoeplinger: Yes it's that object.c code from mono, nothing in the BCL references the ctor.

akoeplinger:
@jonpryor the backport won't be merged in time for the 6.0 release that you're shipping with, so this is to fix the issue until that backport reaches you.

jonpryor:
```markdown
[Microsoft.Android.Sdk.ILLink] Preserve FirstChanceExceptionEventArgs

Fixes: https://github.com/xamarin/xamarin-android/issues/6626

Context: https://github.com/dotnet/runtime/pull/68235

If you create a .NET 6 app whichs ubscribes to the
[`AppDomain.FirstChanceException`][0] event:

	partial class MainActivity {
	    protected override void OnCreate (Bundle? savedInstanceState)
	    {
	        AppDomain.CurrentDomain.FirstChanceException += (o, e) => {
	            Console.WriteLine ("First chance exception!");
	       };
	    }
	}

and has a way to cause an exception to be thrown:

	partial class MainActivity {
	    public override void OnBackPressed() => throw new Exception ("wee?");
	}

and is built in Release configuration with Trimming enabled,

then the app will crash when the exception is thrown (in the above
case, by launching the app and then pressing the Back button):

	E companyname.fo: * Assertion at /__w/1/s/src/mono/mono/metadata/object.c:4179, condition `ctor' not met

The cause of the crash is that the linker removes the
[`FirstChanceExceptionEventArgs(Exception)`][1] constructor.

This was fixed in dotnet/runtime#68235 by using appropriate
`[DynamicDependency]` attributes.  However, this runtime fix is
unlikely to be released before our next RC.

Add a workaround by updating
`src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Private.CoreLib.xml`
so that the linker won't remove the
`FirstChanceExceptionEventArgs(Exception)` constructor.

[0]: https://docs.microsoft.com/en-us/dotnet/api/system.appdomain.firstchanceexception?view=net-6.0
[1]: https://docs.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.firstchanceexceptioneventargs.-ctor?view=net-6.0
```

Issue #6949, opened 2022-04-20
The constructor of AndroidHttpHandler should be preserved together with AndroidMessageHandler


huoyaoyuan:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

.NET SDK 6.0.202, Runtime 6.0.4

### Description

`Xamarin.Android.Net.AndroidHttpHandler` was used for Xamarin android applications, and is set when using Visual Studio project setting UI. 
However, only the new `AndroidMessageHandler` is preserved:

https://github.com/xamarin/xamarin-android/blob/bacc3f0282b05978f1cde13f0ca49ecc58048efe/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs#L342-L348

When using `HttpClientHandler` with a migrated application, it crashes hardly in Debug configuration, and throws NullReferenceException in Release configuration. It's quite hard to debug for people unfamiliar with how http client handler works.

### Steps to Reproduce

Use `<AndroidHttpClientHandlerType>Xamarin.Android.Net.AndroidHttpHandler</AndroidHttpClientHandlerType>` in any minimal application, and use `HttpClientHandler`.

### Did you find any workaround?

I browsed the code path setting up native handler, and realized only `AndroidMessageHandler` would work.

jonathanpeppers:
@huoyaoyuan `AndroidClientHandler` is marked obsolete:
https://github.com/xamarin/xamarin-android/blob/bacc3f0282b05978f1cde13f0ca49ecc58048efe/src/Mono.Android/Xamarin.Android.Net/AndroidClientHandler.cs#L60-L61

Is there a reason you need to use it? Just so we understand the use-case.

For now, I'd workaround with a custom linker configuration:

https://docs.microsoft.com/en-us/xamarin/cross-platform/deploy-test/linker

```xml
<assembly fullname="Mono.Android">
    <type fullname="Xamarin.Android.Net.AndroidHttpHandler">
        <method signature="System.Void .ctor()" />
    </type>
</assembly>
```

We don't want to put `DynamicDependency`, because that would preserve it in every `net6.0-android` app using `HttpClient`.

huoyaoyuan:
> Is there a reason you need to use it? Just so we understand the use-case.

Converted legacy project specifies this handler. Supporting it would make migration easier.
It's also nice to include a sdk warning when the handler is specified, and update Visual Studio project UI to select the new one.

jonathanpeppers:
[...]

PR #6966, opened 2022-04-26
[One .NET] proper default value for $(UseNativeHttpHandler)

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/6949
Fixes: https://github.com/dotnet/maui/issues/6174

The linker feature switch `$(UseNativeHttpHandler)` defaults to:

    <UseNativeHttpHandler Condition="'$(UseNativeHttpHandler)' == ''">true</UseNativeHttpHandler

If you set `$(AndroidHttpClientHandlerType)` to something else other
than the default, like `System.Net.Http.HttpClientHandler`, you would
actually need to toggle this switch yourself for things to work
properly.

We should only default `$(UseNativeHttpHandler)` to `true` if:

* `$(AndroidHttpClientHandlerType)` is blank
* `$(AndroidHttpClientHandlerType)` is `Xamarin.Android.Net.AndroidMessageHandler`
* `$(AndroidHttpClientHandlerType)` is `Xamarin.Android.Net.AndroidClientHandler`

Otherwise, `$(UseNativeHttpHandler)` defaults to `false`.

I updated a test around this scenario, to verify the feature flag is
set in `runtimeconfig.json`.


jonathanpeppers:
Currently, unsure if this is the right fix.

dellis1972:
we should ask @jonpryor if this is the right thing to do cos I have no idea :D

jonathanpeppers:
There is a conversation in Discord, starting here: 
https://discord.com/channels/732297728826277939/732297837953679412/968574970668138516

There are a couple related issues, I'm not clear on what to fix yet.

PR #6970, opened 2022-04-29
Better timing infrastructure

grendello:
We rely on precise (nanosecond level) performance measurements and, as
we improve various components of the code, the precision requirements
increase - especially with regards to the amount of indeterminism.  One
such source of indeterminism are the `logcat` calls which we make as the
measurements are taken.  Such calls send the message to a system daemon
which needs to acquire locks in order to put messages in the `logcat`
buffer(s).  Each such call means that the operating system schedules our
process out and an undetermined amount of time can pass before control
comes back to our thread.

In order to avoid this particular issue, this commit implements a new
timing infrastructure which can log events without incurring the cost of
`logcat` until such time as the timing is done and the information can
be logged without affecting measurements.  This is done by introducing a
new timing mode, `timing=fast-bare`, which can be used in the
`debug.mono.log` system property instead of `timing` or `timing=bare`.
The mode is similar to `timing=bare` but it postpones until a broadcast
intent is sent to the application.  This can be done using the following
command:

    adb shell broadcast -a mono.android.app.DUMP_TIMING_DATA <PACKAGE_NAME>

At this point, all the gathered events are logged.  In addition, the new
infrastructure measures the time it spent initializing itself, which can
then be subtracted from the total time to get as close to "real world"
performance of the app as possible.

In testing, I've seen that this causes timing results to be more stable
as well as saving us up to 4% time for the `Displayed` measure.

### DIFF
--- a/Documentation/workflow/SystemProperties.md
+++ b/Documentation/workflow/SystemProperties.md
@@ -157,12 +157,73 @@ categories:
     Enable logging for messages related to network activity.
   * `timing=bare`
     Enable logging of native code performance information, without
-    logging method execution timing information to a file.
+    logging method execution timing information to a file.  Timed
+    events are logged to `logcat` immediately, which affects the
+    accuracy of measurements and introduces an element of
+    unpredictability to measurements.
+  * `timing=fast-bare`
+    Similar to `timing=bare` above, but the timings aren't logged
+    until the `mono.android.app.DUMP_TIMING_DATA` broadcast is sent to
+    the application.  This can be done with `adb shell am broadcast -a
+    mono.android.app.DUMP_TIMING_DATA [PACKAGE_NAME]` command.
   * `timing`
     Enable logging of native code performance information, including
     method execution timing which is written to a file named
     `methods.txt`.  `timing=bare` should be used in preference to this
     category.
+
+#### Timing events format
+
+In any of the timing modes, the logged messages have the following
+format:
+
+```
+<TAG>: [<S>/<E>] <MESSAGE>; elapsed s:ms::ns
+```
+
+Where:
+
+All lower case and punctuation characters are verbatim in every message
+
+`<TAG>` is always `monodroid-timing`
+
+`<S>` denotes a "stage":
+
+  * `0`: events before control is handed over to the managed runtime
+    (the "native init" stage)
+  * `1`: events after the above "native init" stage
+  * `2`: used only in the `timing=fast-bare` mode to mark the summary
+    information.
+
+`<E>` denotes an "event":
+
+  * For the `0` and `1` stages it's one of
+    * `0`: assembly decompression
+    * `1`: assembly load
+    * `2`: assembly preload
+    * `3`: initialization and start of debugging
+    * `4`: timing subsystem initialization
+    * `5`: java-to-managed type lookup
+    * `6`: managed-to-java type lookup
+    * `7`: Mono runtime initialization
+    * `8`: native to managed transition (a call to
+      `Android.Runtime.JNIEnv.Initialize`)
+    * `9`: time spent registering the runtime config blob on NET6+
+    * `10`: managed type registration
+    * `11`: total time spent initializing the native runtime
+    * `12`: unspecified event
+  * For the `2` stage it's one of:
+    * `1`: mode marker message logged at the very beginning of the app
+    * `2`: performance results "heading"
+    * `3`: no events logged message
+    * `4`: accumulated results "heading"
+    * `5`: total time spent loading assemblies
+    * `6`: total time spent performing java-to-managed type lookups
+    * `7`: total time spent performing managed-to-java type lookups
+
+The format is meant to make it easier for scripts/other software which
+look at the log to find timing events without having to rely on the
+actual wording of the event message.
 
 ### debug.mono.max_grefc
 
--- a/Documentation/workflow/SystemProperties.md
+++ b/Documentation/workflow/SystemProperties.md
@@ -225,6 +225,63 @@ The format is meant to make it easier for scripts/other software which
 look at the log to find timing events without having to rely on the
 actual wording of the event message.
 
+#### Reading the timing information
+
+`timing` and `timing=bare` modes write all the event messages to
+`logcat` immediately, so it is sufficient to either watch the `logcat`
+buffer in real time:
+
+```
+adb logcat
+```
+
+or dump the buffer it after the timing sequence is entirely recorded:
+
+```
+adb logcat -d > logcat.txt
+```
+
+The `timing=fast-bare` mode is slightly more involved in that in
+requires the broadcast to be sent before data is logged in `logcat`.
+The following sequence of commands can be placed in a script and used
+to obtain the information:
+
+```shell
+# Enable only timing log messages, to minimize impact on application
+# performance
+$ adb shell setprop debug.mono.log timing=fast-bare
+
+# Build and install the application
+$ dotnet build -t:Install -p:Configuration=Release
+
+# Increase logcat buffer size so that we don't miss any messages
+$ adb logcat -G 16M
+
+# Clear logcat buffer
+$ adb logcat -c
+
+# Start the activity and wait for the startup process to finish. This
+# is marked by Android by logging the `Displayed` message in logcat,
+# formatted similarly to:
+#
+#  ActivityTaskManager: Displayed PACKAGE_NAME/ACTIVITY_NAME: +398ms
+#
+$ adb shell am start -n PACKAGE_NAME/ACTIVITY_NAME -S -W
+
+# Tell the timing infrastructure to log all the gathered events to
+# logcat:
+$ adb shell am broadcast -a mono.android.app.DUMP_TIMING_DATA PACKAGE_NAME
+
+# Dump logcat contents into a file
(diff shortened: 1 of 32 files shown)


src/java-runtime/java/mono/android/MonoPackageManager.java
```diff
@@ -35,6 +35,11 @@ public static void LoadApplication (Context context, ApplicationInfo runtimePack
 				);
 				context.registerReceiver (new mono.android.app.NotifyTimeZoneChanges (), timezoneChangedFilter);
 
+				android.content.IntentFilter dumpTimingDataFilter = new android.content.IntentFilter (
+						"mono.android.app.DUMP_TIMING_DATA"
+				);
+				context.registerReceiver (new mono.android.app.DumpTimingData (), dumpTimingDataFilter);
```
jonathanpeppers, line 41: Is there any problem to put this in all `Release` apps? I know we probably want to *be able* to get the timings in `Release` mode, but you'd be able to run this broadcast on customers' apps on Google Play.

grendello: > Is there any problem to put this in all `Release` apps? I know we probably want to _be able_ to get the timings in `Release` mode, but you'd be able to run this broadcast on customers' apps on Google Play.

It would be completely harmless then, [see here](https://github.com/xamarin/xamarin-android/pull/6970/files#diff-c2b317c53337d77015d6dc79f6a41f5fab23908c44e83ad7ef16be79b66abeb7R2533).

dellis1972: This might also be useful for customers who want to profile their app I guess.

But I do wonder if there is a way we can enable this at built time somehow, so generate the java code which adds the Intent .

grendello: The only way would be to have a 3rd build for both the native runtime and the managed+java parts. We could probably put a runtime condition on the above code, but it would have to be either a system property check or a Bundle key lookup, neither of which is free. The native code would have to be compiled with appropriate options, OTOH, since the timing code is sprinkled all around our code base **and** inlined (which is responsible for the `libmonodroid.so` size increase). Personally, I would prefer to have a separate "timing" build (Release only, Debug can include the timing infrastructure) as it would have even less impact on the actual measurements, since we wouldn't need runtime checks whether timing is enabled or not. However, that would change the way we've been using this infrastructure since very long time ago, and it would also make the build process more (conceptually) complicated as we'd have to introduce a new "supported" configuration, say, `ReleaseTiming`. But, maybe, it's not such a bad idea? If it were implemented, the "standard" `Release` build could contain no timing code at all.

dellis1972: I was just wondering if we can put something like this in the manifest and remove the need to hardcode it?

```
<receiver android:name="mono.android.app.DumpTimingData">
    <intent-filter>
        <action android:name="mono.android.app.DUMP_TIMING_DATA"/>
    </intent-filter>
</receiver>
```

If this is the case we can expose an MSbuild property and have the generateJavaStubs or manifest merger inject the required receiver element.

Documentation/workflow/SystemProperties.md
```diff
@@ -157,13 +157,74 @@ categories:
...
     `methods.txt`.  `timing=bare` should be used in preference to this
     category.
 
+#### Timing events format
+
+In any of the timing modes, the logged messages have the following
+format:
+
+```
+<TAG>: [<S>/<E>] <MESSAGE>; elapsed s:ms::ns
```
jonpryor, line 181: What's the rationale for `<S>` and `<E>` being *numbers*, as opposed to some "more meaningful" descriptor?

If (when?) we add additional stages (perhaps not *that* plausible) or events (feels quite plausible), numbers will require updating, which in turn will mean that the output is dependent upon the version of xamarin-android being used, which feels…brittle.

Documentation/workflow/SystemProperties.md
```diff
@@ -157,13 +157,74 @@ categories:
...
+  * For the `2` stage it's one of:
+    * `1`: mode marker message logged at the very beginning of the app
+    * `2`: performance results "heading"
+    * `3`: no events logged message
+    * `4`: accumulated results "heading"
+    * `5`: total time spent loading assemblies
+    * `6`: total time spent performing java-to-managed type lookups
+    * `7`: total time spent performing managed-to-java type lookups
+
+The format is meant to make it easier for scripts/other software which
```
jonpryor, line 224: Numbers aren't needed for this.  String "prefixes" with some constraints (no spaces) would also satisfy this.

jonpryor:
It would be useful to have a "complete workflow" example, a'la:

```
adb logcat > log.txt
dotnet build -f net6.0-android -t:Run
# wait a bit…
adb shell broadcast -a mono.android.app.DUMP_TIMING_DATA <PACKAGE_NAME>
```

It may also be useful to add a `_DumpTimingData` target so that you can instead do:

```
dotnet build -f net6.0-android -t:_DumpTimingData
```

as the target can have access to the package name.

What data is written, and where to?  Is it written to `adb logcat`?  Is it written to the "host" console of `adb shell broadcast`?  Somewhere else?

What does the output look like?

Documentation/workflow/SystemProperties.md
grendello: > What's the rationale for `<S>` and `<E>` being _numbers_, as opposed to some "more meaningful" descriptor?

 * Numbers don't require coming up with unique names, they are just monotonically increased.
 * The output is shorter, leaving more room for the actual message (logcat messages have a limit of 1023 characters)
 * The tag is meant for script/software parsing, humans reading the output need to pay attention only to the actual
   message, thus there's no need for the tag to have meaningful names

> 
> If (when?) we add additional stages (perhaps not _that_ plausible) or events (feels quite plausible), numbers will require updating, which in turn will mean that the output is dependent upon the version of xamarin-android being used, which feels…brittle.

The new events will just next numbers in sequence. This is no more brittle than adding a new event name, which also makes output dependent on the version...

Documentation/workflow/SystemProperties.md
grendello: > Numbers aren't needed for this. String "prefixes" with some constraints (no spaces) would also satisfy this.

Yes, they would, but see my comment above.  Adding a meaning to the tag is redundant (for a human reader) with the message contents.

grendello:
> It would be useful to have a "complete workflow" example, a'la:
> 
> ```
> adb logcat > log.txt
> dotnet build -f net6.0-android -t:Run
> # wait a bit…
> adb shell broadcast -a mono.android.app.DUMP_TIMING_DATA <PACKAGE_NAME>
> ```
> 

I'll add the sample

> It may also be useful to add a `_DumpTimingData` target so that you can instead do:
> 
> ```
> dotnet build -f net6.0-android -t:_DumpTimingData
> ```
> 
> as the target can have access to the package name.

Not sure if it's that useful... For this to be useful you need to be in the same directory as the solution you're timing (so that the csproj/sln is found),
and also the whole msbuild machinery behind `dotnet build` will take more time than practical to issue a single command you can either quickly type in
the terminal or have in a shell script somewhere.

> 
> What data is written, and where to? Is it written to `adb logcat`? Is it written to the "host" console of `adb shell broadcast`? Somewhere else?
This is described in the documentation of the [`timing=fast-bare`](https://github.com/xamarin/xamarin-android/pull/6970/files#diff-5e9908be302e0d576f3190e339a4489c089ed70c1a7eb7b452c929be75c86970R165-R168) option
It states that the option is similar to `timing=bare` with the exception of the messages being delayed till the broadcast is received. `timing-bare`
docs state that the messages are written to `logcat`

> 
> What does the output look like?
This is described in [this section](https://github.com/xamarin/xamarin-android/pull/6970/files#diff-5e9908be302e0d576f3190e339a4489c089ed70c1a7eb7b452c929be75c86970R175)

src/java-runtime/java/mono/android/MonoPackageManager.java
dellis1972: OK so I tested the following.

Add the code below to a `Test.java` source file

```java
package mono.android.app;
import android.content.BroadcastReceiver;
import android.content.Intent;
import android.content.Context;
import android.util.Log;


public class Test extends BroadcastReceiver
{
    public Test ()
    {

    }
    @Override
    public void onReceive(Context context, Intent intent)
    {
        Log.d ("DEBUG", "Got a broadcast!!!! DEBUG!!!");
    }
}
```

Add the following to your csproj

```xml
<ItemGroup>
   <AndroidJavaSource Include="Test.java" />
</ItemGroup>
```

Add the following to your `application` element in your `AndroidManifest.xml`

```xml
	<receiver android:name="mono.android.app.Test">
		<intent-filter>
			<action android:name="mono.android.app.DUMP_TIMING_DATA"/>
		</intent-filter>
	</receiver>
```

Build and run your app (make sure the app is actually running).

Run 

```
 adb shell am broadcast -a mono.android.app.DUMP_TIMING_DATA <packagename>
```

and you should see the log message from the code above in `adb logcat`. 
This is with no C# code to register the reciever. So we should be able to generate this reciever at build time based on a MSBuild property that the end user can set. This will mean we don't need to always include the reciever in all builds.

PR #6971, opened 2022-04-29
[templates] add INTERNET permission

jonathanpeppers:
Context: https://github.com/dotnet/maui/issues/6576
Context: https://github.com/dotnet/maui/pull/6673

Debug builds on Android automatically add the `INTERNET` permission,
because otherwise debugging wouldn't work! This definitely adds some
confusion if you switch to a `Release` build and start getting
exceptions in your `HttpClient` calls.

We should just add the `INTERNET` premission to the Android templates,
as nearly every modern app will use the internet. Customers can simply
remove the permission if they don't want it.


pjcollins (approved):

jonathanpeppers:
These test failures should be unrelated:

image

Looking into these this afternoon ^^

PR #6972, opened 2022-04-29
[Mono.Android] custom validation callback for server certificates in HTTP handlers

jonathanpeppers:
Backport of: https://github.com/xamarin/xamarin-android/pull/6665
Context: https://github.com/dotnet/runtime/issues/62966

The `AndroidClientHandler` and `AndroidMessageHandler` classes both
have the `ServerCertificateCustomValidationCallback` property, which
should be useful e.g. to allow running the Android app against a
server with a self-signed SSL certificate during development, but the
callback is never used. Unfortunatelly since .NET 6 the
`System.Net.Http.SocketsHttpHandler` for Android doesn't support the
use case anymore. That means that [the recommended way of connecting
to local web server][0] won't work in MAUI.

This PR introduces an implementation of `IX509TrustManger` which wraps
the default Java X509 trust manager and calls the user's callback on
top of the default validation.

It turns out that `X509Chain` `Build` function doesn't work on
Android, so I'm not calling it and I'm passing the chain to the
callback directly.

Additionally, we need a default proguard rule due to:

https://github.com/xamarin/xamarin-android/blob/46002b49d8c0b7b1a17532a8e104b4d31afee7a6/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/GenerateProguardConfiguration.cs#L50-L57

    -keep class xamarin.android.net.X509TrustManagerWithValidationCallback { *; <init>(...); }

`Mono.Android.dll` is skipped during the
`GenerateProguardConfiguration` linker step. It might be worth
addressing this in a future PR.

[0]: https://docs.microsoft.com/en-us/xamarin/cross-platform/deploy-test/connect-to-local-web-services

### DIFF
--- a/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
+++ b/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
@@ -157,7 +157,15 @@ namespace Xamarin.Android.Net
 
 		public bool CheckCertificateRevocationList { get; set; } = false;
 
-		public Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> ServerCertificateCustomValidationCallback { get; set; }
+		X509TrustManagerWithValidationCallback.Helper? _callbackTrustManagerHelper = null;
+
+		public Func<HttpRequestMessage, X509Certificate2?, X509Chain?, SslPolicyErrors, bool>? ServerCertificateCustomValidationCallback
+		{
+			get => _callbackTrustManagerHelper?.Callback;
+			set {
+				_callbackTrustManagerHelper = value != null ? new X509TrustManagerWithValidationCallback.Helper (value) : null;
+			}
+		}
 
 		// See: https://developer.android.com/reference/javax/net/ssl/SSLSocket#protocols
 		public SslProtocols SslProtocols { get; set; } =
@@ -199,7 +207,7 @@ namespace Xamarin.Android.Net
 		/// If the website requires authentication, this property will contain data about each scheme supported
 		/// by the server after the response. Note that unauthorized request will return a valid response - you
 		/// need to check the status code and and (re)configure AndroidMessageHandler instance accordingly by providing
-		/// both the credentials and the authentication scheme by setting the <see cref="PreAuthenticationData"/> 
+		/// both the credentials and the authentication scheme by setting the <see cref="PreAuthenticationData"/>
 		/// property. If AndroidMessageHandler is not able to detect the kind of authentication scheme it will store an
 		/// instance of <see cref="AuthenticationData"/> with its <see cref="AuthenticationData.Scheme"/> property
 		/// set to <c>AuthenticationScheme.Unsupported</c> and the application will be responsible for providing an
@@ -939,7 +947,7 @@ namespace Xamarin.Android.Net
 
 			// SSL context must be set up as soon as possible, before adding any content or
 			// headers. Otherwise Java won't use the socket factory
-			SetupSSL (httpConnection as HttpsURLConnection);
+			SetupSSL (httpConnection as HttpsURLConnection, request);
 			if (request.Content != null)
 				AddHeaders (httpConnection, request.Content.Headers);
 			AddHeaders (httpConnection, request.Headers);
@@ -997,7 +1005,7 @@ namespace Xamarin.Android.Net
 		internal SSLSocketFactory? ConfigureCustomSSLSocketFactoryInternal (HttpsURLConnection connection)
 			=> ConfigureCustomSSLSocketFactoryInternal (connection);
 
-		void SetupSSL (HttpsURLConnection? httpsConnection)
+		void SetupSSL (HttpsURLConnection? httpsConnection, HttpRequestMessage requestMessage)
 		{
 			if (httpsConnection == null)
 				return;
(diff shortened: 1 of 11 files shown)


jonathanpeppers:
Going to mark this as draft until we see if something is wrong with the tests at:

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6089774&view=ms.vss-test-web.build-test-results-tab&runId=39320146&resultId=100166&paneView=debug

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
```diff
@@ -1017,35 +1025,48 @@ void SetupSSL (HttpsURLConnection? httpsConnection)
...
-				// chain
-				if (!gotCerts)
+				// If there are no trusted certs, no custom trust manager factory or custom certificate validation callback
+				// there is no point in changing the behavior of the default SSL socket factory
+				if (!gotCerts && _callbackTrustManagerHelper == null)
 					return;
 				
 				tmf = TrustManagerFactory.GetInstance (TrustManagerFactory.DefaultAlgorithm);
-				tmf?.Init (keyStore);
+				tmf?.Init (gotCerts ? keyStore : null); // only use the custom key store if the user defined any trusted certs
```
grendello, line 1040: Should it say `custom` instead of `trusted`? Certs defined by the user will most likely not be trusted in the official CA chain sense

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
```diff
@@ -1017,35 +1025,48 @@ void SetupSSL (HttpsURLConnection? httpsConnection)
...
+			context?.Init (kmf?.GetKeyManagers (), trustManagers, null);
 			httpsConnection.SSLSocketFactory = context?.SocketFactory;
+
+			KeyStore? InitializeKeyStore (out bool gotCerts)
+			{
+				var keyStore = KeyStore.GetInstance (KeyStore.DefaultType);
+				keyStore?.Load (null, null);
+				gotCerts = TrustedCerts?.Count > 0;
+
+				if (gotCerts) {
```
grendello, line 1059: This is an absolute nitpick on my part, but could you change the code to

```csharp
if (!gotCerts) {
  return keystore;
}

for (...) {
}
```

I tend to prefer to keep large blocks of code outside the `if` statement sub-blocks - in my opinion it makes code more readable :)

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
```diff
@@ -0,0 +1,96 @@
...
+using JavaCertificateException = Java.Security.Cert.CertificateException;
+using JavaX509Certificate = Java.Security.Cert.X509Certificate;
+
+namespace Xamarin.Android.Net
+{
+	internal sealed class X509TrustManagerWithValidationCallback : Java.Lang.Object, IX509TrustManager
+	{
+		internal sealed class Helper
+		{
+			public Func<HttpRequestMessage, X509Certificate2?, X509Chain?, SslPolicyErrors, bool> Callback { get; }
```
grendello, line 19: Should `Func<>` be perhaps replaced by a specific delegate to be used throughout this class but also in `AndroidMessageHandler`? It would make it easier to modify it in the future, should there be a need for that and also reduce the signal to noise ratio when reading the code

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
```diff
@@ -0,0 +1,96 @@
...
+				Callback = callback;
+			}
+
+			public ITrustManager[] Inject (
+				ITrustManager[]? trustManagers,
+				HttpRequestMessage requestMessage)
+			{
+				IX509TrustManager? x509TrustManager = trustManagers?.OfType<IX509TrustManager> ().FirstOrDefault ();
+				IEnumerable<ITrustManager> otherTrustManagers = trustManagers?.Where (manager => manager != x509TrustManager) ?? Enumerable.Empty<ITrustManager> ();
+				var trustManagerWithCallback = new X509TrustManagerWithValidationCallback (x509TrustManager, requestMessage, Callback);
```
grendello, line 32: I'm not very fond of LINQ code in frameworks or libraries, since it's not very performant. Could we use the boring loops here? :)

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
```diff
@@ -0,0 +1,96 @@
...
+				return otherTrustManagers.Prepend (trustManagerWithCallback).ToArray ();
+			}
+		}
+
+		private readonly IX509TrustManager? _internalTrustManager;
+		private readonly HttpRequestMessage _request;
+		private readonly Func<HttpRequestMessage, X509Certificate2?, X509Chain?, SslPolicyErrors, bool> _serverCertificateCustomValidationCallback;
+
+		private X509TrustManagerWithValidationCallback (
```
grendello, line 41: The `private` modifiers are superfluous since this is the default visibility of a class member.

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
```diff
@@ -0,0 +1,96 @@
...
+		}
+
+		public void CheckServerTrusted (JavaX509Certificate[] javaChain, string authType)
+		{
+			var sslPolicyErrors = SslPolicyErrors.None;
+			var certificates = ConvertCertificates (javaChain);
+
+			try {
+				_internalTrustManager?.CheckServerTrusted (javaChain, authType);
+			} catch (JavaCertificateException) {
```
grendello, line 58: I feel we should log this exception, not the whole trace, just the message, using the `Debug` severity so that it's not shown by default.

grendello (changes requested):

jonathanpeppers:
@grendello this is a backport of 48540d6933e2bbded3db6106d594a3949e9ac020 to the .NET 6 release branch (I think you approved). I mostly just fixed up the `.apkdesc` files, so the tests would pass here.

Do you want @simonrozsival to address these comments in `main` before we backport this?

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
dellis1972: I agree this is generally how we prefer stuff in XA, the "early out" option is best 👍

grendello:
> @grendello this is a backport of [48540d6](https://github.com/xamarin/xamarin-android/commit/48540d6933e2bbded3db6106d594a3949e9ac020) to the .NET 6 release branch (I think you approved). I mostly just fixed up the `.apkdesc` files, so the tests would pass here.
> 
> Do you want @simonrozsival to address these comments in `main` before we backport this?

DotNet follows a different style, and sinde my comments are mostly stylistic, hence my lack of comments then :)

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
grendello, line 1040:
```suggestion
				tmf?.Init (gotCerts ? keyStore : null); // only use the custom key store if the user defined any custom certs
```

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
```diff
@@ -0,0 +1,96 @@
...
+				HttpRequestMessage requestMessage)
+			{
+				IX509TrustManager? x509TrustManager = trustManagers?.OfType<IX509TrustManager> ().FirstOrDefault ();
+				IEnumerable<ITrustManager> otherTrustManagers = trustManagers?.Where (manager => manager != x509TrustManager) ?? Enumerable.Empty<ITrustManager> ();
+				var trustManagerWithCallback = new X509TrustManagerWithValidationCallback (x509TrustManager, requestMessage, Callback);
+				return otherTrustManagers.Prepend (trustManagerWithCallback).ToArray ();
+			}
+		}
+
+		private readonly IX509TrustManager? _internalTrustManager;
```
grendello, line 37:
```suggestion
		readonly IX509TrustManager? _internalTrustManager;
```

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
grendello, line 41:
```suggestion
		X509TrustManagerWithValidationCallback (
```

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
```diff
@@ -0,0 +1,96 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
```
grendello, line 6: using Android.Runtime;

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
```diff
@@ -0,0 +1,96 @@
...
+			}
+		}
+
+		public void CheckClientTrusted (JavaX509Certificate[] chain, string authType)
+			=> _internalTrustManager?.CheckClientTrusted (chain, authType);
+
+		public JavaX509Certificate[] GetAcceptedIssuers ()
+			=> _internalTrustManager?.GetAcceptedIssuers () ?? Array.Empty<JavaX509Certificate> ();
+
+		private static X509Chain CreateChain (X509Certificate2[] certificates)
```
grendello, line 80:
```suggestion
		static X509Chain CreateChain (X509Certificate2[] certificates)
```

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
```diff
@@ -0,0 +1,96 @@
...
+
+			chain.ChainPolicy.RevocationMode = X509RevocationMode.Online;
+			chain.ChainPolicy.RevocationFlag = X509RevocationFlag.ExcludeRoot;
+
+			chain.ChainPolicy.ExtraStore.AddRange (certificates);
+
+			return chain;
+		}
+
+		private static X509Certificate2[] ConvertCertificates (JavaX509Certificate[] certificates)
```
grendello, line 93:
```suggestion
		static X509Certificate2[] ConvertCertificates (JavaX509Certificate[] certificates)
```

src/Mono.Android/Xamarin.Android.Net/X509TrustManagerWithValidationCallback.cs
jonpryor: If exposed publicly from a public type, you don't necessarily want "easier to modify", as changing the delegate signature is a breaking change.  This should to be the same delegate type as the public `AndroidMessageHandler.ServerCertificateCustomValidationCallback` property, and thus is a type that *cannot* be changed.

PR #6982, opened 2022-05-04
[ci] Use new macOS signing certs

pjcollins:
Move to the new macOS signing certs which use a newer version of Xcode.
There should also be more machine capacity available for processing
signing requests using these certs.


pjcollins:
Testing this with https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6109290&view=results

PR #6985, opened 2022-05-05
[build] Exclude proguard files from .NET packs

pjcollins:
The `CreateMultiDexMainDexClassList` and `Proguard` tasks that rely on
`proguard.jar` are only used in classic Xamarin.Android.  We should not
need to ship proguard files in the .NET packs/installers.


jonathanpeppers (approved):
`Xamarin.Android.DX.targets` is not in .NET 6, so there isn't even a way to use these currently.

PR #6988, opened 2022-05-06
[One .NET] support building net6.0-android apps in .NET 7

jonathanpeppers:
We currently have a hard dependency between:

* `Xamarin.Android.Build.Tasks.dll`
* `libmonodroid.so`

The design for building a `net6.0-android` app with a .NET 7 SDK
results in us using a .NET 7 `Xamarin.Android.Builds.Tasks.dll` and a
.NET 6 `libmonodroid.so`. This crashes with:

    05-06 09:32:31.206  8713  8713 E AndroidRuntime: java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "java_name_width" referenced by "/data/app/~~8CHtXY8wK4g7s9VcOLTQEg==/com.microsoft.net6.helloandroid-cUUhwuQls7TRb6JdBsbPdg==/split_config.arm64_v8a.apk!/lib/arm64-v8a/libmonodroid.so"...

To solve this problem:

* Define `$(AndroidNet6Version)` to specify our .NET 6 GA `32.0.301`
  build that is released.

* Add an alias for a `Microsoft.Android.Sdk.$(HostOS).NET6` workload
  pack that loads a .NET 6 version of `Microsoft.Android.Sdk.$(HostOS)`.

* Include both the .NET 6 and .NET 7 versions of these packs in the
  workload.

* Check `$(TargetFrameworkVersion)` and import the .NET 7 or .NET 6
  MSBuild targets when appropriate.

* Update the `@(KnownFrameworkReference)` to the version of .NET 6
  that our .NET 7 packs know about.

* We have a matching .NET 6 `Xamarin.Android.Build.Tasks.dll` and
  `libmonodroid.so`!

This obviously increases our install footprint of
`Microsoft.Android.Sdk.$(HostOS)` by 2x. We might be able to address
this by splitting out some files into a
`Microsoft.Android.Sdk.Tooling` pack that can be shared between .NET 6
and .NET 7. Perhaps files like `bundletool.jar` and `r8.jar` could be
shared?

One hack I had to put in place was to avoid the crash:

    05-06 12:21:19.981 18868 18868 E AndroidRuntime: java.lang.UnsatisfiedLinkError: dlopen failed: cannot locate symbol "mono_opt_aot_lazy_assembly_load" referenced by "/data/app/~~Fp4gEr_9sxU1qU6PoI-v6Q==/com.companyname.foo-EKEtL67XJI-d0H17_3oz3g==/lib/arm64/libmonodroid.so"...

The .NET 7 SDK we are on, does not yet know about .NET runtime 6.0.5
(it uses 6.0.3), so I manually fixed this version number for now. We
should be able to remove this eventually when the .NET 7 SDK provides
6.0.5 by default.

A second hack I put in place:

* Make Microsoft.Android.Sdk a "framework" and not an "sdk" in
  `WorkloadManifest.json`.

This prevents two `AutoImport.props` from being imported... for the
.NET 7 one and the .NET 6 one.

This seems very weird, but this still works:

    <Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk" />

We can remove this if I condition everything in `AutoImport.props`
behind a current `$(TargetFrameworkVersion)`. We would then need to
ship this in a .NET 6 release.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
@@ -5,6 +5,7 @@
       "description": ".NET SDK Workload for building Android applications.",
       "packs": [
         "Microsoft.Android.Sdk",
+        "Microsoft.Android.Sdk.NET6",
         "Microsoft.Android.Ref.32",
         "Microsoft.Android.Runtime.32.android-arm",
         "Microsoft.Android.Runtime.32.android-arm64",
@@ -32,6 +33,17 @@
     "Microsoft.Android.Sdk": {
       "kind": "sdk",
       "version": "@WORKLOAD_VERSION@",
+      "alias-to": {
+        "osx-x64": "Microsoft.Android.Sdk.Darwin",
+        "osx-arm64": "Microsoft.Android.Sdk.Darwin",
+        "win-x86": "Microsoft.Android.Sdk.Windows",
+        "win-x64": "Microsoft.Android.Sdk.Windows",
+        "linux-x64": "Microsoft.Android.Sdk.Linux"
+      }
+    },
+    "Microsoft.Android.Sdk.NET6": {
+      "kind": "framework",
+      "version": "@NET6_VERSION@",
       "alias-to": {
         "osx-x64": "Microsoft.Android.Sdk.Darwin",
         "osx-arm64": "Microsoft.Android.Sdk.Darwin",
--- a/Configuration.props
+++ b/Configuration.props
@@ -47,6 +47,7 @@
     <DebugType Condition=" '$(DebugType)' == '' ">portable</DebugType>
     <Deterministic Condition=" '$(Deterministic)' == '' ">True</Deterministic>
     <LangVersion Condition=" '$(LangVersion)' == '' ">latest</LangVersion>
+    <AndroidNet6Version Condition=" '$(AndroidNet6Version)' == '' ">32.0.301</AndroidNet6Version>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(HostOS)' == '' ">
     <HostOS Condition="$([MSBuild]::IsOSPlatform('windows'))">Windows</HostOS>
--- a/build-tools/create-packs/Directory.Build.targets
+++ b/build-tools/create-packs/Directory.Build.targets
@@ -123,6 +123,8 @@
     </PropertyGroup>
     <ItemGroup>
       <_NuGetSources Include="$(OutputPath.TrimEnd('\'))" />
+      <!-- This allows us to install our Android .NET 6 packs -->
+      <_NuGetSources Include="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet6/nuget/v3/index.json" />
       <_PreviewPacks Condition=" '$(AndroidLatestStableApiLevel)' != '$(AndroidLatestUnstableApiLevel)' " Include="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nuget-unsigned\Microsoft.Android.Ref.$(AndroidLatestUnstableApiLevel).*.nupkg" />
       <_InstallArguments Include="android" />
       <_InstallArguments Include="android-$(AndroidLatestUnstableApiLevel)" Condition=" '@(_PreviewPacks->Count())' != '0' " />
--- a/build-tools/create-packs/Microsoft.NET.Sdk.Android.proj
+++ b/build-tools/create-packs/Microsoft.NET.Sdk.Android.proj
@@ -28,18 +28,24 @@ about the various Microsoft.Android workloads.
       DependsOnTargets="_GetDefaultPackageVersion;_GetLicense">
     <PropertyGroup>
       <WorkloadManifestJsonPath>$(OutputPath)workload-manifest\WorkloadManifest.json</WorkloadManifestJsonPath>
+      <WorkloadManifestTargetsPath>$(OutputPath)workload-manifest\WorkloadManifest.targets</WorkloadManifestTargetsPath>
     </PropertyGroup>
 
-    <MakeDir Directories="$([System.IO.Path]::GetDirectoryName ($(WorkloadManifestJsonPath)))" />
+    <MakeDir Directories="$(OutputPath)workload-manifest" />
     <ReplaceFileContents
         SourceFile="$(XamarinAndroidSourcePath)src\Xamarin.Android.Build.Tasks\Microsoft.NET.Sdk.Android\WorkloadManifest.in.json"
         DestinationFile="$(WorkloadManifestJsonPath)"
-        Replacements="@WORKLOAD_VERSION@=$(AndroidPackVersionLong)">
+        Replacements="@WORKLOAD_VERSION@=$(AndroidPackVersionLong);@NET6_VERSION@=$(AndroidNet6Version)">
+    </ReplaceFileContents>
+    <ReplaceFileContents
+        SourceFile="$(XamarinAndroidSourcePath)src\Xamarin.Android.Build.Tasks\Microsoft.NET.Sdk.Android\WorkloadManifest.in.targets"
+        DestinationFile="$(WorkloadManifestTargetsPath)"
+        Replacements="@NET6_VERSION@=$(AndroidNet6Version)">
     </ReplaceFileContents>
 
     <ItemGroup>
-      <_PackageFiles Include="$(XamarinAndroidSourcePath)src\Xamarin.Android.Build.Tasks\Microsoft.NET.Sdk.Android\WorkloadManifest.targets" PackagePath="data" />
       <_PackageFiles Include="$(WorkloadManifestJsonPath)" PackagePath="data" />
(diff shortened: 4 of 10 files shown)


jonathanpeppers:
TODO: add some `net6.0-android` runtime tests. We need to test these apps launch and do not crash.

/cc @dsplaisted @steveisok 

This PR gets `net6.0-android` apps running again, but one issue is the AOT packs.

`dotnet build -c Release` is currently using:

```
_MonoAOTCompilerPath = C:\src\xamarin-android\bin\Debug\dotnet\packs\Microsoft.NETCore.App.Runtime.AOT.win-x64.Cross.android-arm\7.0.0-preview.4.22172.7\Sdk\..\tools\mono-aot-cross.exe
```

Which is basically just imported by:
```xml
<Import Project="Sdk.props" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task" />
<Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x86" />
<Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x64" />
<Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm" />
<Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-arm64" />
```

Apps somehow still work with this mismatch? @lambdageek was saying this was just lucky and it might not work at all. So there is probably some other work needed in the Mono workload for these.

jonathanpeppers:
I requested to get some of the .NET 6 workload packs on the `dotnet-public` feed.

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
```diff
@@ -29,9 +30,20 @@
...
+      "alias-to": {
+        "osx-x64": "Microsoft.Android.Sdk.NET7.Darwin",
+        "osx-arm64": "Microsoft.Android.Sdk.NET7.Darwin",
+        "win-x86": "Microsoft.Android.Sdk.NET7.Windows",
+        "win-x64": "Microsoft.Android.Sdk.NET7.Windows",
+        "linux-x64": "Microsoft.Android.Sdk.NET7.Linux"
+      }
+    },
+    "Microsoft.Android.Sdk": {
+      "kind": "framework",
```
jonathanpeppers, line 45: I was getting two `AutoImport.props` one from .NET 6 and .NET 7, until I changed this to be a `framework`.

But we can still import this like:
```xml
<Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk" />
```

This feels weird to me, but gives the behavior we want? Basically we always use the .NET 7 `AutoImport.props`, but import the .NET 6 MSBuild targets for `net6.0-android` projects.

If we add new things in `AutoImport.props` that might break, we can condition them based on `TargetFrameworkVersion` being >= 7.0.

lambdageek:
> Apps somehow still work with this mismatch? @lambdageek was saying this was just lucky and it might not work at all. So there is probably some other work needed in the Mono workload for these.

Just to make sure I understand:

you're using the .NET 7 AOT compiler, to compile net6.0-android assemblies?  in particular, CoreLib, the BCL etc are all from the 6.0 runtime pack?

In that case here's what's happening:
* the AOT file format has a version
   https://github.com/dotnet/runtime/blob/f14aa5d03e8bbf7e7700e9cf021b91867fb7e775/src/mono/mono/mini/aot-runtime.h#L14
* as long as the AOT compiler (NET 7) and the AOT runtime (net6.0-android) agree on the file format version, the runtime will try to load AOT image. (it has to be an exact match, there's no forward or backward compat)
* once the image is loaded, as long as the image matches the expected MVID of the assembly and all MVIDs of all the referenced assemblies everything, and it was compiled against the 6.0 CoreLib, everything makes sense and we will attempt to run the AOT image.

So there are two failure modes:
1. (nice) we remember to bump the AOT file format version when we make breaking changes in .NET 7 and the net6.0-android runtime will refuse to load the AOT images and you will get a warning in the ADB log and slower performance when we fall back to the JIT
2. (bad) we make a breaking change between the AOT compiler and the AOT runtime and forget to bump the file format version. Things will appear to work but then there will be crashes in native code when some assumption somewhere is violated.

We're generally good about bumping the version when there are physical changes to the AOT file format (layout changes, we add or remove tables, etc).  I think we're probably less good about semantic changes (some helper function in the AOT runtime changes behavior, some ).

TL;DR it'll just work until one day it doesn't. if you're lucky it'll break in a nice way, if not, it'll break in a very hard to debug way.

jonathanpeppers:
@lambdageek right now the AOT packs are working (pure luck!), but @steveisok is figuring out how we can structure the Mono workload for this problem.

When we have this figured out, we should be able to use .NET 6 AOT packs for net6.0-android, and .NET 7 AOT packs for net7.0-android.

src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
```diff
@@ -40,6 +41,17 @@
...
+      "version": "@NET6_VERSION@",
+      "alias-to": {
+        "osx-x64": "Microsoft.Android.Sdk.Darwin",
+        "osx-arm64": "Microsoft.Android.Sdk.Darwin",
+        "win-x86": "Microsoft.Android.Sdk.Windows",
+        "win-x64": "Microsoft.Android.Sdk.Windows",
+        "linux-x64": "Microsoft.Android.Sdk.Linux"
+      }
+    },
```
jonathanpeppers, line 54: After looking through some of the workload docs/examples:
* https://github.com/dsplaisted/WorkloadVersioningSample/blob/main/workloads/sdk-manifests/6.0.200/DoomSlug.MonoToolChain.net6/WorkloadManifest.json#L22-L28
* https://github.com/dotnet/sdk/blob/main/documentation/general/workloads/cross-version-workloads.md

I reworked this to be simpler. The `Microsoft.Android.Sdk.$(HostOS)` pack is no longer renamed.

And we simply use these alias mappings instead.

jonpryor:
```markdown
We currently have a hard dependency between:

  * `Xamarin.Android.Build.Tasks.dll`
  * `libmonodroid.so`

The design for building a `net6.0-android` app with a .NET 7 SDK
results in us using a .NET 7 `Xamarin.Android.Builds.Tasks.dll` and a
.NET 6 `libmonodroid.so`.  This crashes with:

	E AndroidRuntime: java.lang.UnsatisfiedLinkError: dlopen failed:
	cannot locate symbol "java_name_width" referenced by
	"/data/app/~~8CHtXY8wK4g7s9VcOLTQEg==/com.microsoft.net6.helloandroid-cUUhwuQls7TRb6JdBsbPdg==/split_config.arm64_v8a.apk!/lib/arm64-v8a/libmonodroid.so"...

To solve this problem:

  * Define `$(AndroidNet6Version)` to specify our .NET 6 GA
    `32.0.301` build that is released.

  * Add an alias for a `Microsoft.Android.Sdk.$(HostOS).NET6`
    workload pack that loads a .NET 6 version of
    `Microsoft.Android.Sdk.$(HostOS)`.

  * Include both the .NET 6 and .NET 7 versions of these packs in the
    workload.

  * Check `$(TargetFrameworkVersion)` and import the .NET 7 or .NET 6
    MSBuild targets when appropriate.

  * Update `@(KnownFrameworkReference)` to the version of .NET 6
    that our .NET 7 packs know about.

  * We have a matching .NET 6 `Xamarin.Android.Build.Tasks.dll` and
    `libmonodroid.so`!

This obviously increases our install footprint of
`Microsoft.Android.Sdk.$(HostOS)` by 2x.  We might be able to address
this by splitting out some files into a
`Microsoft.Android.Sdk.Tooling` pack that can be shared between
.NET 6 and .NET 7.  (Perhaps files like `bundletool.jar` and `r8.jar`
could be shared?)

One hack I had to put in place was to avoid the crash:

	E AndroidRuntime: java.lang.UnsatisfiedLinkError: dlopen failed:
	cannot locate symbol "mono_opt_aot_lazy_assembly_load" referenced by
	"/data/app/~~Fp4gEr_9sxU1qU6PoI-v6Q==/com.companyname.foo-EKEtL67XJI-d0H17_3oz3g==/lib/arm64/libmonodroid.so"...

The .NET 7 SDK we are on does not yet know about .NET runtime 6.0.5
(it uses 6.0.3), so I manually fixed this version number for now.
We should be able to remove this eventually when the .NET 7 SDK
provides 6.0.5 by default.

A second hack I put in place:

Make `Microsoft.Android.Sdk.NET6` a "framework" and not an "sdk"
within `WorkloadManifest.json`.  This prevents *both* the .NET 6
`AutoImport.props` *and* .NET 7 `AutoImport.props` files from being
imported, causing duplication of item groups/etc.

This seems very weird, but this still works:

	<Import Project="Sdk.targets" Sdk="Microsoft.Android.Sdk" />

We can remove this if I condition everything in `AutoImport.props`
behind a current `$(TargetFrameworkVersion)`.  We would then need to
ship this in a .NET 6 release.
```

pjcollins (approved):

PR #6997, opened 2022-05-11
Fix TrimEnd for _AndroidTargetingPackId in Microsoft.Android.Sdk.BundledVersions.in.targets

akoeplinger:
We can't use `TrimEnd` here since it will use the `TrimEnd(char[])` overload so it'd trim **all** `.` and `0` chars from the end, not the whole string `.0`. 
This results in `30.0` being trimmed to just `3` instead of the expected `30`.

See https://github.com/dotnet/sdk/issues/25370 for more details, I first thought it's a dotnet SDK issue but it turned out to be an XA issue.

/cc @pjcollins


jonathanpeppers:
Was building an app with `net6.0-android30.0` enough to break this? At one point I thought we had a test for that, maybe we should add further parameters for the tests here:

https://github.com/xamarin/xamarin-android/blob/2a4c58c6162c18b86d6de9d08b27a06bef9af594/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs#L712

akoeplinger:
Yes just changing the TFM to `net6.0-android30.0` broke the build during restore.

pjcollins (approved):
Good find, we indeed missed this because net6.0-android30.0 is not supported by our SDK.

jonathanpeppers (approved):

PR #7001, opened 2022-05-13
[build] Refactor net6 references in src/monodroid

pjcollins:
Instances of `NET6` in src/monodroid and xaprepare have been removed or
replaced with `NET`.  This should hopefully improve readability and
maintainability as we begin work on .NET 7 and beyond.

Build and output directories for One .NET versions of monodroid.csproj
have been updated to use `$(AndroidRID)` as a unique identifier, rather
than `$(AndroidAbi)-net6`.

### DIFF
--- a/src/monodroid/jni/build-info.hh
+++ b/src/monodroid/jni/build-info.hh
@@ -17,13 +17,6 @@ namespace xamarin::android::internal
 		static constexpr char xa_version[] = XA_VERSION;
 		static constexpr char date[] = __DATE__;
 
-		static constexpr bool is_net6 =
-#if NET6
-			true;
-#else // def NET6
-			false;
-#endif // ndef NET6
-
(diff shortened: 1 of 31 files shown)


src/monodroid/jni/build-info.hh
```diff
@@ -17,13 +17,6 @@ namespace xamarin::android::internal
 		static constexpr char xa_version[] = XA_VERSION;
 		static constexpr char date[] = __DATE__;
 
-		static constexpr bool is_net6 =
```
pjcollins, line 20: @grendello I couldn't find references to this anywhere, should this be restored and renamed or is it not needed?

grendello: It can be removed. I think I used it at one point in code that's gone now, so no need to keep it around.

grendello (approved):

PR #7010, opened 2022-05-17
[build] Copy .NET build output to ref/runtime dirs

pjcollins:
Build steps have been updated to circumvent the need to pack and install
.nupkgs to test local build changes against .NET 7 +.

Projects which produce content required for our Ref and Runtime packs
have been updated to copy their outputs to a `packs` directory.  This
directory follows the structure that dotnet workload resolution expects.
The version of these local packs has been simplified compared to CI, it
will not include commit distance or any semver metadata.

A new `ConfigureLocalWorkload` target has been added to run a couple of
extra steps that are required to set up the local workload.  This
includes generating workload manifest, target, and FrameworkList files,
installing mono/runtime dependencies, and creating the templates pack.

A new `BuildDotNet` target has been added (primarily for Windows) which
will only build one API level and call this new target.  The `make all`
and `make jenkins` rules will also now run `ConfigureLocalWorkload` by
default.

The `dotnet-local` scripts have been updated to set the variables needed
to resolve `sdk-manifests` and `packs` content from a non default path.

After running `make` or the `BuildDotNet` target, the `dotnet-local`
scripts can be used to build and run a template:

    .\dotnet-local.cmd new android -o androidnet
    .\dotnet-local.cmd build -t:Run androidnet\androidnet.csproj

### DIFF
--- a/Configuration.props
+++ b/Configuration.props
@@ -65,7 +65,13 @@
     <_XABinRelativeInstallPrefix>lib\xamarin.android</_XABinRelativeInstallPrefix>
     <XAInstallPrefix Condition=" '$(XAInstallPrefix)' == '' ">$(MSBuildThisFileDirectory)bin\$(Configuration)\$(_XABinRelativeInstallPrefix)\</XAInstallPrefix>
     <_MonoAndroidNETOutputRoot>$(XAInstallPrefix)xbuild-frameworks\Microsoft.Android\</_MonoAndroidNETOutputRoot>
-    <MicrosoftAndroidSdkPackDir>$(BuildOutputDirectory)packs\$(MicrosoftAndroidSdkPackName)\$(AndroidPackVersion)\</MicrosoftAndroidSdkPackDir>
+    <_MonoAndroidNETDefaultOutDir>$(_MonoAndroidNETOutputRoot)$(AndroidApiLevel)\</_MonoAndroidNETDefaultOutDir>
+    <MicrosoftAndroidRefPackDir>$(BuildOutputDirectory)lib\packs\Microsoft.Android.Ref.$(AndroidApiLevel)\$(AndroidPackVersion)\ref\$(DotNetTargetFramework)\</MicrosoftAndroidRefPackDir>
+    <MicrosoftAndroidArmPackDir>$(BuildOutputDirectory)lib\packs\Microsoft.Android.Runtime.$(AndroidApiLevel).android-arm\$(AndroidPackVersion)\runtimes\android-arm\</MicrosoftAndroidArmPackDir>
+    <MicrosoftAndroidArm64PackDir>$(BuildOutputDirectory)lib\packs\Microsoft.Android.Runtime.$(AndroidApiLevel).android-arm64\$(AndroidPackVersion)\runtimes\android-arm64\</MicrosoftAndroidArm64PackDir>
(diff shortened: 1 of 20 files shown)


dellis1972:
If this does what I think it does ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️  😄

grendello (approved):

jonathanpeppers (approved):
I guess this requires `dotnet-local` now for the new env vars. In a random folder, I tried:
```
> C:\src\xamarin-android\bin\Debug\dotnet\dotnet.exe new android
No templates found matching: 'android'.

To list installed templates, run:
   dotnet new list
To search for the templates on NuGet.org, run:
   dotnet new search android

> C:\src\xamarin-android\dotnet-local.cmd new android
The template "Android Application" was created successfully.
```

But this seems to be faster than what it was doing before:
image

jonpryor:
```markdown
How does one test incremental changes to xamarin-android for .NET6+?
Previously, every change would require re-creating the .NET packs:

	# make changes to src/Xamarin.Android.Build.Tasks
	% dotnet build src/Xamarin.Android.Build.Tasks/Xamarin.Android.Build.Tasks.csproj

	# re-create .NET packs
	% make pack-dotnet

	# can now use `./dotnet-local.*` to test changes.

The problem with this is that `make pack-dotnet` can take multiple
minutes to complete.  This slows down local testing.

Update the build system to instead place build artifacts within the
`bin\$(Configuration)\lib\packs` directory, and update the
`dotnet-local` scripts to set:

  * The `DOTNETSDK_WORKLOAD_MANIFEST_ROOTS` environment variable to
    `bin\$(Configuration)\lib\sdk-manifests`, and the

  * The `DOTNETSDK_WORKLOAD_PACK_ROOTS` environment variable to
    `bin\$(Configuration)\lib`.

This simplifies the .NET 6 test workflow to:

	./dotnet-local build Xamarin.Android.sln
	# Use of `dotnet-local` now includes in-tree changes.


The `bin\$(Configuration)\lib\packs` directory contains the content
required for our Ref and Runtime packs, and follows the directory
structure that the `dotnet workload` resolution process expects.
This in-build-tree structure is simplified compare to the CI
structure; commit distance and semver metadata isn't present.

A new `ConfigureLocalWorkload` target has been added to run a couple
of extra steps that are required to set up the local workload.  This
includes generating workload manifest, target, and FrameworkList files,
installing mono/runtime dependencies, and creating the templates pack.
It is run as part of the `make all` and `make jenkins` Make targets,
and [TODO: how often should it be re-run?]

A new `BuildDotNet` target has been added (primarily for Windows) which
will only build one API level and call this new target [TODO: which 
new target?].
```

pjcollins:
Partially updated message:

```markdown
A new `ConfigureLocalWorkload` target has been added to run a couple
of extra steps that are required to set up the local workload.  This
includes generating workload manifest, target, and FrameworkList files,
installing mono/runtime dependencies, and creating the templates pack.
It is run as part of the `make all` and `make jenkins` Make targets,
as well as the new `BuildDotNet` target that has been added primarily
for Windows.

The `ConfigureLocalWorkload` target should only need to run once per
"clean build".  It is fine to run this target more frequently than that,
but it is only required when building from scratch or after one of the
following changes:

  * The structure of the android workload changes.
  * Files are added to either the Ref or Runtime packs.
  * Changes are made to the templates pack.
  * The dotnet/runtime dependency is updated.
```

PR #7018, opened 2022-05-19
[ci] Upgrade to macOS 12 Monterey and Xcode 13.4

pjcollins:
Context: https://github.com/xamarin/xamarin-android/commit/fb227bdfcc59d3d69017e1e53f403a97b9fefcf7

We recently ran into an issue in our official builds that only occurred
when using Xcode 13.2.  This issue didn't surface in our PR builds which
use Xcode 12.4:

    timing-internal.hh(433,4): error GD3D6EBB4: no matching function for call to 'calculate_interval'
        calculate_interval (event.start, event.end, interval, total_ns);

Our [PR build pool][0] was recently updated to include some Monterey
images.  We should be able to avoid issues like this in the future by
using the same macOS and Xcode versions for both PR builds and official
builds.

Includes a bump to xamarin/monodroid@ac6ecd3cb:
https://github.com/xamarin/monodroid/compare/4fa114311a0ab882c1de1ed265450747a7c34c25...ac6ecd3cb16d85948967f4968d1f35eb2849e7a5

[0]: https://devdiv.visualstudio.com/DevDiv/_settings/agentqueues?queueId=2206&view=agents


pjcollins:
Additional test run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6181196&view=results

jonathanpeppers (approved):

grendello (approved):

PR #7023, opened 2022-05-20
[workload] AutoImport.props supports future .NET versions

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/pull/6988

One problem we hit when trying to build `net6.0-android` projects from
a .NET 7 SDK & Android workload, was that we ended up importing *two*
`AutoImport.props` files.

These are imported for any pack in `WorkloadManifest.json` declared as
an SDK:

    "packs": {
      "Microsoft.Android.Sdk": {
        "kind": "sdk",

So the way this has to work:

* Any `<ItemGroup>` must be conditioned so that
  `$(TargetFrameworkVersion)` matches the .NET SDK version expected.

* Never use an MSBuild property in `AutoImport.props`. Properties are
  evaluated in the first pass and `$(TargetFrameworkVersion)` will be
  blank...

I removed `$(_DefaultJavaSourceJarPattern)` and just duplicated the
value where it was used.

I used `[MSBuild]::VersionEquals()` for all checks, because the value
is `v6.0` and the helper method properly handles this case for us.

When this change ships in a .NET 6 servicing release, we can then
update our .NET 7 packages to rely on this change. Currently, #6988 is
declaring the .NET 6 Microsoft.Android.Sdk as a "framework", which
hacks around the problem.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
@@ -16,17 +16,13 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
 -->
 <Project>
 
-  <PropertyGroup>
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
Ok unexpectedly this breaks `Xamarin.Legacy.Sdk`:

image

pjcollins (approved):

PR #7028, opened 2022-05-23
[Java.Interop] Bump to xamarin/java.interop@1bab47d

jpobst:
Bumps Java.Interop to target `net7.0`.

Changes: https://github.com/xamarin/java.interop/compare/1f27ab552d03aeb74cdc6f8985fcffbfdb9a7ddf...1bab47d6eda682de1e38ed289acb56ebda770c38

The `Configuration.Java.Interop.Override.props` file is now partially
generated by xaprepare, to avoid having to hard code duplicate version
and path values in it.

The Java.Interop prepare step has been moved out of xaprepare and into
a new MSBuild target that will be ran with the .NET 7 preview SDK that is
installed by xaprepare.

Extra Java.Interop tests have been removed from the "Windows Build and
Smoke Test" job, as these should already be running as part of Java
Interop CI.

The `dotnet --list-sdks` command was failing on macOS when installing
the latest .NET 7 Preview SDK, and it has been replaced by a step that
runs `dotnet --info`.


grendello (approved):

build-tools/installers/create-installers.targets
```diff
@@ -270,7 +270,7 @@
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Bindings.JarToXml.targets" ExcludeFromAndroidNETSdk="true" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Build.Tasks.dll" />
     <_MSBuildFiles Include="$(MSBuildSrcDir)\Xamarin.Android.Build.Tasks.pdb" />
-    <_MSBuildFiles Include="@(_LocalizationLanguages->'$(MSBuildSrcDir)\%(Identity)\Java.Interop.Localization.resources.dll')" />
+    <_MSBuildFiles Include="@(_LocalizationLanguages->'$(MSBuildSrcDir)\%(Identity)\Java.Interop.Localization.resources.dll')" ExcludeFromAndroidNETSdk="true" />
```
jonathanpeppers, line 273: Does this file not exist now?

pjcollins: I think this change is not strictly necessary, but I figured it would be best to pull these `Java.Interop.Localization.resources.dll` files from the same location that we take `Java.Interop.Localization.dll` and other JI tools from for the .NET SDK (e.g. https://github.com/xamarin/xamarin-android/pull/7028/files#diff-0134c03744a69c35881982fb2140cf780816fe186f0b6fac66477a8416d2fca6R82)

pjcollins:
I guess I missed a JI bump when testing here before that PR was merged... We may need https://github.com/xamarin/xamarin-android/pull/7038 to land before this.

PR #7032, opened 2022-05-24
[ci] Use nuget-msi-convert/job/v3.yml

pjcollins:

Use v3 of the .msi installer generation templates.  This new template
version adds support for a somewhat recent rework in dotnet/arcade, and
updates the manifest pack versioning schema.


pjcollins:
Testing with https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6195016&view=results

jonathanpeppers (approved):
From what I can tell these packages look OK:

image

We might not really be able to test these until we have a VS insertion.

PR #7054, opened 2022-05-31
[Mono.Android] Move API xml file generation to dedicated project.

jpobst:
*One* issue with running `dotnet build Xamarin.Android.sln` without `-m:1` is that it attempts to run our Android API extraction process multiple times simultaneously.  This results in file locking issues:

```
C:\code\xamarin-android\src\Mono.Android\Mono.Android.targets(409,5): "C:\code\xamarin-android\src\Mono.Android\..\..\bin\Debug\lib\xamarin.android\xbuild\Xamarin\Android\class-parse.exe C:\Users\jopobst\android-toolchain\sdk\platforms\android-26\android.jar -platform=26 -parameter-names="C:\code\xamarin-android\src\Mono.Android\..\..\src\Mono.Android\Profiles\api-26.params.txt" -o="C:\code\xamarin-android\src\Mono.Android\..\..\bin\BuildDebug\api\api-26.xml.class-parse"" failed with code: -532462766  Error output: 
Unhandled Exception: System.IO.IOException: The process cannot access the file 'C:\code\xamarin-android\bin\BuildDebug\api\api-26.xml.class-parse' because it is being used by another process.
   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
   at System.IO.StreamWriter.CreateFile(String path, Boolean append, Boolean checkHost)
   at System.IO.StreamWriter..ctor(String path, Boolean append, Encoding encoding, Int32 bufferSize, Boolean checkHost)
   at System.IO.StreamWriter..ctor(String path, Boolean append, Encoding encoding)
   at Xamarin.Android.Tools.App.Main(String[] args) in C:\code\xamarin-android\external\Java.Interop\tools\class-parse\Program.cs:line 68 [C:\code\xamarin-android\src\Mono.Android\Mono.Android.csproj]
```

This seems to be a result of running the process as an `AfterTargets="Build"` target in `api-xml-adjuster.csproj`.  By moving it to a `NoTargets` project, we can eliminate MSBuild trying to run it multiple times.  Thus, the process has been moved to the new `create-android-api.csproj` project.

Also, move the `api-merge` step from `Mono.Android.targets`.  This currently runs twice, once for `monoandroid10` and once for `net7.0`, however the output is not dependent on `TargetFramework`.  Instead of generating:
- `src\Mono.Android\obj\Debug\monoandroid10\android-32\mcw\api.xml`
- `src\Mono.Android\obj\Debug\net7.0\android-32\mcw\api.xml`

we will now generate:
- `bin\Build$(Configuration)\api\api-32.xml`

Calling `api-merge` only once saves ~25s from build time.

Note that `generator` writes intermediate files (eg: `api.xml.fixed`) to the directory containing `api.xml`, so `Mono.Android.targets` will copy the `api-X.xml` file to its eg: `obj\Debug\net7.0\android-32\mcw` directory to work from.  Without this, both `monoadroid10` and `net7.0` would write to the common location, causing possible conflicts.

Additionally, let VS2022 rewrite `Xamarin.Android.sln` to its preferred format in order to add the new project.

### DIFF
--- a/build-tools/create-android-api/create-android-api.csproj
+++ b/build-tools/create-android-api/create-android-api.csproj
@@ -1,13 +1,30 @@
-<Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-
+<Project Sdk="Microsoft.Build.NoTargets">
   <UsingTask AssemblyFile="$(BootstrapTasksAssembly)" TaskName="Xamarin.Android.Tools.BootstrapTasks.RunParallelCmds" />
-  <Import Project="..\..\bin\Build$(Configuration)\Mono.Android.Apis.projitems" Condition="Exists('..\..\bin\Build$(Configuration)\Mono.Android.Apis.projitems')" />
 
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+  </PropertyGroup>
+  
+  <Import Project="..\..\Configuration.props" />
+  <Import Project="..\..\bin\Build$(Configuration)\Mono.Android.Apis.projitems" Condition="Exists('..\..\bin\Build$(Configuration)\Mono.Android.Apis.projitems')"/>
+
+  <PropertyGroup>
+    <OutputPath>$(XAInstallPrefix)xbuild\Xamarin\Android</OutputPath>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\external\Java.Interop\tools\class-parse\class-parse.csproj" ReferenceOutputAssembly="False" />
+    <ProjectReference Include="..\api-xml-adjuster\api-xml-adjuster.csproj" ReferenceOutputAssembly="False" />
+  </ItemGroup>
+  
   <PropertyGroup>
     <_TopDir>$(MSBuildThisFileDirectory)..\..</_TopDir>
     <_OutputPath>$(_TopDir)\bin\Build$(Configuration)\</_OutputPath>
   </PropertyGroup>
 
+  <!-- First we need to figure out all the needed API levels.
+       This is done by looking at which levels are included in 'Mono.Android.Apis.projitems'
+       and looking for matching 'api-X.params.txt' files. -->
   <Target Name="_DefineApiFiles">
     <ItemGroup>
       <_Api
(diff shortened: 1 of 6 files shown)


src/Mono.Android/Mono.Android.targets
```diff
@@ -2,6 +2,7 @@
 <Project DefaultTargets="Build" ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <UsingTask AssemblyFile="$(PrepTasksAssembly)" TaskName="Xamarin.Android.BuildTools.PrepTasks.ReplaceFileContents" />
   <UsingTask AssemblyFile="$(BootstrapTasksAssembly)" TaskName="Xamarin.Android.Tools.BootstrapTasks.CheckApiCompatibility" />
+  <UsingTask AssemblyFile="$(BootstrapTasksAssembly)" TaskName="Xamarin.Android.Tools.BootstrapTasks.RunParallelCmds" />
```
jonathanpeppers, line 5: Was this one leftover from refactoring? I don't see `<RunParallelCmds/>` used in this file.

build-tools/create-android-api/create-android-api.csproj
```diff
@@ -1,32 +1,64 @@
...
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\external\Java.Interop\tools\class-parse\class-parse.csproj" ReferenceOutputAssembly="False" />
+    <ProjectReference Include="..\api-xml-adjuster\api-xml-adjuster.csproj" ReferenceOutputAssembly="False" />
+    <ProjectReference Include="..\api-merge\api-merge.csproj" ReferenceOutputAssembly="False" />
+  </ItemGroup>
+  
   <PropertyGroup>
     <_TopDir>$(MSBuildThisFileDirectory)..\..</_TopDir>
     <_OutputPath>$(_TopDir)\bin\Build$(Configuration)\</_OutputPath>
```
jonathanpeppers, line 23: What's the difference between `$(OutputPath)` and `$(_OutputPath)` in this project? Do we need both, or can we remove one?

build-tools/create-android-api/create-android-api.csproj
```diff
@@ -1,32 +1,64 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\external\Java.Interop\tools\class-parse\class-parse.csproj" ReferenceOutputAssembly="False" />
+    <ProjectReference Include="..\api-xml-adjuster\api-xml-adjuster.csproj" ReferenceOutputAssembly="False" />
+    <ProjectReference Include="..\api-merge\api-merge.csproj" ReferenceOutputAssembly="False" />
+  </ItemGroup>
+  
   <PropertyGroup>
     <_TopDir>$(MSBuildThisFileDirectory)..\..</_TopDir>
```
jonathanpeppers, line 22: If you like, you could remove `$(_TopDir)` and use this instead:

https://github.com/xamarin/xamarin-android/blob/e52c131543580641a8884866af2e6aa1e1e02898/Configuration.props#L115

build-tools/create-android-api/create-android-api.csproj
```diff
@@ -0,0 +1,131 @@
...
+  <!-- Next we need to run class-parse on each platform's 'android.jar' in order
+       to produce the 'api-X.xml.class-parse' files. -->
+  <Target Name="_ClassParse"
+      BeforeTargets="_AdjustApiXml"
+      DependsOnTargets="_DefineApiFiles"
+      Inputs="@(ApiFileDefinition->'%(ParameterDescription)')"
+      Outputs="@(ApiFileDefinition->'%(ClassParseXml)')">
+      
+    <PropertyGroup>
+      <ClassParse>$(XamarinAndroidSourcePath)\bin\$(Configuration)\lib\xamarin.android\xbuild\Xamarin\Android\class-parse.exe</ClassParse>
```
pjcollins, line 54: Are there any differences in the `net472` and `net7.0` versions of class-parse that would make us want to run this for each framework version instead? I guess eventually we will just use the net7.0 version?

pjcollins (approved):

Issue #7020, opened 2022-05-20
Optimize type & member remapping storage, lookup


jonpryor:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

-

### Description

Context: https://github.com/xamarin/xamarin-android/pull/6591

## Background
PR #6591 fixes https://github.com/xamarin/java.interop/issues/867, but it's largely a "prototype" with great cost to app size:

  * It uses XML as the intermediate format.
  * This XML is parsed during app startup, before most other managed init.
  * Which caused the `CheckIncludedAssemblies()` test to require updates, adding the assemblies `System.Collections.dll`, `System.Collections.Concurrent.dll`, `System.Collections.NonGeneric.dll`, `System.Console.dll`, `System.IO.Compression.dll`, `System.Net.Http.dll`, `System.Net.Primitives.dll`, `System.Net.Requests.dll`, `System.Private.Uri.dll`, `System.Private.Xml.dll`, `System.Security.Cryptography.dll`, and `System.Text.RegularExpressions.dll`
  * which in turn increase the `.apk` size; `BuildReleaseArm64SimpleDotNet.apkdesc` saw `PackageSize` increase from 2,959,252 bytes to 3,451,764 bytes, an increase of ~17%.

PR #6591 exists to allow further testing.  It is *not* the end state of how things should be.

## Proposal

Instead of using XML as an intermediary packaged and parsed within the app, the data should instead be stored within `libxamarin-app.so`, a'la typemap data.

We would add two new exports to `libmonodroid.so`:

```c
extern "C" const char *
_monodroid_lookup_replacement_type (const char *jniSimpleReference);

extern "C" const char *
_monodroid_lookup_replacement_method_info (const char *key);
```

Update/extend the `_CollectAndroidRemapMembers` target so that instead of creating `xa-remap-members.xml` as an `@(AndroidAsset)`, `xa-remap-members.xml` is instead provided as input into whatever creates `libxamarin-app.so`.

The assembly generator will read `<replace-type/>` to populate the table used by `_monodroid_lookup_replacement_type()`.  `//replace-type/@from` will be the key, while `//replace-type/@to` will be the value.

The assembly generator will read `<replace-method/>` to populate the table used by `_monodroid_lookup_replacement_method_info()`.  The `key` to search on will be a *tab* separated value of:

    {@source-type}\t{@source-method-name}\t{@source-method-signature}

The value will be a *tab* separated value of:

[...]

PR #7059, opened 2022-06-01
Mam remapping opt

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/7020

Optimize MAM remapping by generating native data at build time, 
to be included in `libxamarin-app.so` and eliminating the need
to parse XML and use managed dictionaries at run time when
remapping types/methods.


src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -317,123 +324,44 @@ protected override IEnumerable<string> GetSimpleReferences (Type type)
 			};
 		}
 
+		[DllImport (AndroidRuntime.InternalDllName, CallingConvention = CallingConvention.Cdecl)]
+		static extern string? _monodroid_lookup_replacement_type (string jniSimpleReference);
```
jonpryor, line 328: Unless `_monodroid_lookup_replacement_type()` is allocating memory with the "COM task memory allocator" (iirc `g_malloc()` & `g_free()` on Mono + Unixy platforms), this is Not Good™.   This should probably be `IntPtr`.

jonpryor: See also:

* https://www.mono-project.com/docs/advanced/pinvoke/#strings-as-return-values
* https://www.mono-project.com/docs/advanced/pinvoke/#classes-and-structures-as-return-values

grendello: > Unless `_monodroid_lookup_replacement_type()` is allocating memory with the "COM task memory allocator" (iirc `g_malloc()` & `g_free()` on Mono + Unixy platforms), this is Not Good™. This should probably be `IntPtr`.

Memory is in a read-only string section in ELF

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Utilities/EnvironmentHelper.cs
```diff
@@ -45,7 +45,7 @@ public sealed class ApplicationConfig
 			public bool   uses_assembly_preload;
 			public bool   is_a_bundled_app;
 			public bool   broken_exception_transitions;
-			public bool   instant_run_enabled;
+			public bool   instant_run_enabled ;
```
jonathanpeppers, line 48:
```suggestion
			public bool   instant_run_enabled;
```

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
```diff
@@ -116,5 +83,5 @@
       "Size": 1904
     }
   },
-  "PackageSize": 3492724
+  "PackageSize": 3012575
```
jonathanpeppers, line 86: Yay!

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -243,6 +243,13 @@ public override void DeleteWeakGlobalReference (ref JniObjectReference value)
 	}
 
 	class AndroidTypeManager : JniRuntime.JniTypeManager {
+		struct MAMReplacementMethod
```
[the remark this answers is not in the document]
grendello: It's a very good point, perhaps we should prefix the types with something more generic (I do want to have a prefix that identifies the structures as belonging to a particular feature). `MAM` -> `Remapping`, perhaps? @jonpryor?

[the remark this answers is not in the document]
jonpryor: @grendello: yes, let's go with `JniRemapping` or something as a prefix.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1620,6 +1624,41 @@ because xbuild doesn't support framework reference assemblies.
...
+        Inputs="$(MSBuildProjectFullPath)"
+        Outputs="@(_AndroidRemapAssemblySource)">
+  <GenerateJniRemappingNativeCode
+      OutputDirectory="$(_NativeAssemblySourceDir)"
+      GenerateEmptyCode="True"
+      SupportedAbis="@(_BuildTargetAbis)"
+   />
+</Target>
+
+<Target Name="_GenerateAndroidRemapNativeCode"
```
dellis1972, line 1650: Are we expecting these targets to run on every build? i.e no support for skipping if nothing changes between builds?

grendello: We require the sources they generate for all builds, even if the remapping data isn't there (libxamarin-app.so ABI requirements). However, if they already generated the LLVM IR code and there are no changes, then they don't need to regenerate the sources.

jonpryor:
```markdown
[Xamarin.Android.Build.Tasks, monodroid] Optimize member remapping (#7059)

Fixes: https://github.com/xamarin/xamarin-android/issues/7020

Context: f6f11a5a797cd8602854942dccb0f2b1db57c473

Commit f6f11a5a introduced type & member remapping.  The problem with
its approach was that it used XML as the representation format, which
needed to be parsed during process startup.  This would contribute to
app startup slowdowns if there were any remappings, but even if there
weren't any remappings, the minimum App size increased by ~490KB, due
to the added dependencies on `System.Private.Xml.dll` & more.

This app size increase is not ideal.

Remove most of the `.apk` size increases by moving the remapping info
into `libxamarin-app.so`.  If no remappings are used, then the size
increase to `libxamarin-app.so` is negligible, and the `.apk` size is
only 61KB larger than pre-f6f11a5a.


~~ API Changes ~~

`Android.Runtime.AndroidTypeManager` gains the following members:

	partial class AndroidTypeManager {
	    struct JniRemappingReplacementMethod {
	        public string target_type, target_name;
	        public bool   is_static;
	    }
	    static extern byte* _monodroid_lookup_replacement_type (
	            string jniSimpleReference
	    );
	    static extern JniRemappingReplacementMethod* _monodroid_lookup_replacement_method_info (
	            string jniSourceType,
	            string jniMethodName,
	            string jniMethodSignature
	    );
	}

`AndroidTypeManager._monodroid_lookup_replacement_type()` replaces
the `JNIEnv.ReplacementTypes` dictionary from f6f11a5a.

`AndroidTypeManager._monodroid_lookup_replacement_method_info()`
replaces the `JNIEnv.ReplacementMethods` dictionary from f6f11a5a.

Both `_monodroid_lookup_replacement_type()` and
`_monodroid_lookup_replacement_method_info()` are P/Invokes into
`libxamarin-app.so`.


~~ `libxamarin-app.so` Changes ~~

The contents of the `@(_AndroidRemapMembers)` item group are now
stored within `libxamarin-app.so`, with the following structure:

	const uint32_t                          jni_remapping_replacement_method_index_entry_count;
	const JniRemappingIndexTypeEntry        jni_remapping_method_replacement_index[];

	const uint32_t                          jni_remapping_replacement_type_count;
	const JniRemappingTypeReplacementEntry  jni_remapping_type_replacements[];

	struct JniRemappingString {
	    const uint32_t                  length;
	    const char                     *str;
	};
	struct JniRemappingReplacementMethod {
	    const char                     *target_type, *target_name;
	    const bool                      is_static;
	};
	struct JniRemappingIndexMethodEntry {
	    JniRemappingString              name, signature;
	    JniRemappingReplacementMethod   replacement
	};
	struct struct JniRemappingIndexTypeEntry {
	    JniRemappingString              name;
	    uint32_t                        method_count;
	    JniRemappingIndexMethodEntry   *methods;
	};
	struct JniRemappingTypeReplacementEntry {
	    JniRemappingString              name;
	    const char                     *replacement;
	};
	const char *
	_monodroid_lookup_replacement_type (const char *);

	const JniRemappingReplacementMethod*
	_monodroid_lookup_replacement_method_info (const char *jniSourceType, const char *jniMethodName, const char *jniMethodSignature);

Referring to the `<replacements/>` XML from f6f11a5a in
`@(_AndroidRemapMembers)`:

  * `//replace-type/@from` fills `JniRemappingTypeReplacementEntry::name`,
    `//replace-type/@to` fills `JniRemappingTypeReplacementEntry::replacement`,
    and `_monodroid_lookup_replacement_type()` performs a linear
    search over `jni_remapping_type_replacements`.

  * `//replace-method/@source-type` fills `JniRemappingIndexTypeEntry::name`,
    `//replace-method/@source-method-name` fills `JniRemappingIndexMethodEntry::name`,
    `//replace-method/@source-method-signature` fills `JniRemappingIndexMethodEntry::signature`,
    `//replace-method/@target-type` fills `JniRemappingReplacementMethod::target_type`,
    `//replace-method/@target-method-name` fills `JniRemappingReplacementMethod::target_name`,
    `//replace-method/@target-method-signature` and
    `//replace-method/@target-method-parameter-count` are *ignored*,
    `//replace-method/@target-method-instance-to-static` fills `JniRemappingReplacementMethod::is_static`,
    and `_monodroid_lookup_replacement_method_info()` performs a
    search over `jni_remapping_method_replacement_index` looking for
    entries with "matching" type names, method names, and method
    signatures, and once a match is found it returns a pointer to the
    `JniRemappingReplacementMethod` instance.

```

PR #7064, opened 2022-06-02
[build] Version .NET VS manifest file

pjcollins:
As part of supporting .NET 6 projects with our .NET 7 workload, we will
want to be able to insert both .NET 6 and .NET 7 workload packs into VS.

Historically, we have replaced the [VS manifest file][0] every time we
update our workload.  For .NET 7 we will likely want two VS manifests
side by side, one which contains our .NET 6 packs and one for .NET 7.

Add a version to the .NET VS manifest file to support this potential
side by side scenario.

[0]: https://devdiv.visualstudio.com/DevDiv/_git/VS?path=/.corext/Configs/dotnet-workloads-components.json&version=GBmain&line=10&lineEnd=11&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

### DIFF
--- a/build-tools/create-packs/vs-workload.in.props
+++ b/build-tools/create-packs/vs-workload.in.props
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
   <PropertyGroup>
-    <TargetName>Microsoft.NET.Sdk.Android.Workload</TargetName>
+    <TargetName>Microsoft.NET.Sdk.Android.Workload.@VSMAN_VERSION@</TargetName>
     <ManifestBuildVersion>@WORKLOAD_VERSION@</ManifestBuildVersion>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


build-tools/create-packs/vs-workload.in.props
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project>
   <PropertyGroup>
-    <TargetName>Microsoft.NET.Sdk.Android.Workload</TargetName>
+    <TargetName>Microsoft.NET.Sdk.Android.Workload.@VSMAN_VERSION@</TargetName>
```
jonathanpeppers, line 4: So this would be `Microsoft.NET.Sdk.Android.Workload.7.0`?

https://github.com/xamarin/xamarin-android/blob/e52c131543580641a8884866af2e6aa1e1e02898/Directory.Build.props#L4

That's probably fine, but do we need the `.0`? If this would never be seen by a customer, probably doesn't matter.

jonathanpeppers (approved):

build-tools/create-packs/vs-workload.in.props
jonpryor: I like `net7.0` as a suffix.

pjcollins: Agreed, it makes the `7.0` less arbitrary.

PR #7062, opened 2022-06-01
[build] Output MSBuild tools to .NET SDK pack path

pjcollins:
Bumps to xamarin/java.interop@d0ef9e32.
Changes: https://github.com/xamarin/java.interop/compare/1bab47d6eda682de1e38ed289acb56ebda770c38...d0ef9e3228472c9673451c924df50441c0258706

  * https://github.com/xamarin/java.interop/commit/d0ef9e32: [build] Use explicit netstandard2.0 protobuf ref

The vast majority of instances of the MSBuild tool output path:

    bin\$(Configuration)\lib\xamarin.android\xbuild\Xamarin\Android\

Have been replaced with a new .NET SDK pack path:

    bin\$(Configuration)\packs\Microsoft.Android.Sdk.$(HostName)\$(AndroidPackVersion)\tools

Eventually, the `$(DOTNETSDK_WORKLOAD_PACK_ROOTS)` variable can be set
when running `bin\$(Configuration)\dotnet\dotnet` to resolve our SDK
from this new path.  This will allow us to circumvent the need to pack
and install .nupkgs to test local build changes on .NET 7+.

Certain "classic" files, such as `net472` versions of Java.Interop
utilities, will still be output to the old MSBuild path.  Additionally,
the SDK content in the new .NET SDK path will be copied over to the old
MSBuild tool path every time `tools\xabuild\xabuild.csproj` is built.
This will allow `xabuild.exe` to continue to function for testing
"classic" build outputs.

PR #7068, opened 2022-06-03
[Xamarin.Android.sln] Fix issues preventing parallel builds.

jpobst:
There are several known issues that cause errors when attempting to `dotnet build Xamarin.Android.sln` without `-m:1`.  With these fixes, the build *seems* to be buildable without `-m:1`.  

Testing conducted is 4 consecutive CI builds without errors.  Note that due to the unpredictability of race-condition errors appearing, this is not a guarantee that all issues are fixed.  We will simply have to run with this and fix any additional errors as they appear.

Fixes:

- Use `create-android-api` to build `JNIEnv.g.cs` only once, since it is not TargetFramework dependent.  Prevents sharing violation when multiple `Mono.Android.csproj` builds attempt to create it simultaneously.
  - `System.IO.IOException: The process cannot access the file 'C:\a\_work\1\s\src\Mono.Android\Android.Runtime\JNIEnv.g.cs' because it is being used by another process.`
- `Microsoft.Android.Sdk.ILLink.csproj` needs a dependency on `Xamarin.Android.Build.Tasks.csproj` so that `Xamarin.Android.Build.Tasks\obj\$(Configuration)\Profile.g.cs` is available for compile.
  - `CSC error CS2001: Source file 'C:\code\xamarin-android\src\Microsoft.Android.Sdk.ILLink\..\Xamarin.Android.Build.Tasks\obj\Debug\Profile.g.cs' could not be found. [C:\code\xamarin-android\src\Microsoft.Android.Sdk.ILLink\Microsoft.Android.Sdk.ILLink.csproj]`
- Add a project reference to `jnienv-gen.csproj` from `Java.Interop-MonoAndroid.csproj`.  This ensures `jnienv-gen.exe` is available when building `Java.Interop-MonoAndroid.csproj`. 
  - https://github.com/xamarin/java.interop/pull/990
- Add project references between `apksigner`/`r8`/`manifestmerger`.  `gradlew` hits an issue when running multiple copies simultaneously.  While there isn't an actual dependency between these, it forces them to build sequentially.
  - `org.gradle.launcher.daemon.client.DaemonConnectionException: Timeout waiting to connect to the Gradle daemon.` and `The file lock is held by a different Gradle process.`

### DIFF
--- a/tools/javadoc2mdoc/javadoc2mdoc.csproj
+++ b/tools/javadoc2mdoc/javadoc2mdoc.csproj
@@ -5,6 +5,7 @@
     <OutputType>Exe</OutputType>
     <RootNamespace>Xamarin.Android.Tools.JavadocToMDoc</RootNamespace>
     <AssemblyName>javadoc-to-mdoc</AssemblyName>
+    <IntermediateOutputPath>$(BaseIntermediateOutputPath)$(Configuration)\$(TargetFramework)\</IntermediateOutputPath>
   </PropertyGroup>
   <Import Project="..\..\Configuration.props" />
   <PropertyGroup>
--- a/tools/javadoc2mdoc/javadoc2mdoc.csproj
+++ b/tools/javadoc2mdoc/javadoc2mdoc.csproj
@@ -5,7 +5,6 @@
     <OutputType>Exe</OutputType>
     <RootNamespace>Xamarin.Android.Tools.JavadocToMDoc</RootNamespace>
     <AssemblyName>javadoc-to-mdoc</AssemblyName>
-    <IntermediateOutputPath>$(BaseIntermediateOutputPath)$(Configuration)\$(TargetFramework)\</IntermediateOutputPath>
   </PropertyGroup>
   <Import Project="..\..\Configuration.props" />
   <PropertyGroup>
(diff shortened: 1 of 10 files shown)


tools/javadoc2mdoc/javadoc2mdoc.csproj
```diff
@@ -5,6 +5,7 @@
     <OutputType>Exe</OutputType>
     <RootNamespace>Xamarin.Android.Tools.JavadocToMDoc</RootNamespace>
     <AssemblyName>javadoc-to-mdoc</AssemblyName>
+    <IntermediateOutputPath>$(BaseIntermediateOutputPath)$(Configuration)\$(TargetFramework)\</IntermediateOutputPath>
```
jonathanpeppers, line 8: For this one could you remove this instead?
https://github.com/xamarin/xamarin-android/blob/d73db9d8b40a2530b04c0c1d5ecf02ce5f51aa44/tools/javadoc2mdoc/javadoc2mdoc.csproj#L4

jpobst: I don't know, I think it would append it to our `<OutputPath>` which would break stuff?

```
<OutputPath>$(XAInstallPrefix)xbuild\Xamarin\Android\</OutputPath>
```

pjcollins: I just removed the net6.0 build from this project as it will remain a legacy-only feature, so we may not need any change here.

jpobst: Good point, this has changed since I wrote this.  😁 

https://github.com/xamarin/xamarin-android/pull/7062

pjcollins (approved):

jonathanpeppers (approved):

PR #7075, opened 2022-06-07
[Mono.Android.Export] Fix build warnings.

jpobst:
Eliminate warnings from `Mono.Android.Export.csproj`.

- Fix 27 warnings of `warning CA1416: This call site is reachable on all platforms.` by specifying that this assembly relies on the `Android21.0` platform because it relies on `Mono.Android.dll`.
- Suppress 9 warnings of `warning CS0649: Field 'CodeTry.condition' is never assigned to, and will always have its default value null` and similar because this is "external" code we have imported from Mono, and I doubt we're going to fix them.

### DIFF
--- a/src/Mono.Android.Export/Properties/AssemblyInfo.cs
+++ b/src/Mono.Android.Export/Properties/AssemblyInfo.cs
@@ -1,5 +1,6 @@
 ﻿using System.Reflection;
 using System.Runtime.CompilerServices;
+using System.Runtime.Versioning;
 
 // Information about this assembly is defined by the following attributes. 
 // Change them to the values specific to your project.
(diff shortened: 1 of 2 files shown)


src/Mono.Android.Export/Properties/AssemblyInfo.cs
```diff
@@ -9,6 +10,7 @@
 [assembly: AssemblyTrademark ("Xamarin")]
 [assembly: AssemblyCulture ("")]
 [assembly: AssemblyMetadata ("IsTrimmable", "True")]
+[assembly: SupportedOSPlatform ("Android21.0")]
```
jonathanpeppers, line 13: I think it might be better to put something like this in the `.csproj`?
```xml
<SupportedOSPlatformVersion>$(AndroidMinimumDotNetApiLevel)</SupportedOSPlatformVersion>
```
I'm also not sure, but the MSBuild property might enable more build analyzers that if you just have it in `.cs` source files.

jonathanpeppers: Oh maybe you have to do it this way... The project is `net6.0` and not `net6.0-android`.

PR #7079, opened 2022-06-08
[docs] on how to install .NET 7/main builds

jonathanpeppers:


Documentation/guides/InstallBuildsFromMain.md
```diff
@@ -0,0 +1,78 @@
...
+If you get the error:
+
+```
+error XA5207: Could not find android.jar for API level 32.
+This means the Android SDK platform for API level 32 is not installed.
+Either install it in the Android SDK Manager (Tools > Android > Android SDK Manager...),
+or change the Xamarin.Android project to target an API version that is installed.
+```
+
+Open the `Android SDK Manager` as described, click the gear icon in the
```
jonpryor, line 48: Instead of/in addition to using the Android SDK Manager, we should suggest using the `InstallAndroidDependencies` target, which may require setting the `$(AndroidManifestType)` MSBuild property to `GoogleV2` (64b22fc94e453f77817b29e00afddff598336937):

```
# default SDK Manifest
dotnet build -t:InstallAndroidDependencies

# -or-

dotnet build -t:InstallAndroidDependencies -p:AndroidManifestType=GoogleV2
```

We may also want to mention setting `$(p:AcceptAndroidSDKLicenses)`=True.

Documentation/guides/InstallBuildsFromMain.md
```diff
@@ -0,0 +1,78 @@
...
+bottom right, and select `Repository > Full List`.
```
jonpryor, line 49: Convention used elsewhere is that "GUI items" should be **bolded**, thus:

```markdown
**Repository** > **Full List**.
```

PR #7087, opened 2022-06-10
[One .NET] add BuildAndStartTracing and FinishTracing targets

jonathanpeppers:
This allows you to do:

    $ dotnet build -t:BuildAndStartTracing -c Release -r android-arm64

Then when you are ready to collect the trace:

    $ dotnet build -t:FinishTracing

Resulting in a `.nettrace` file in the project directory.

I ran into a lot of difficulty trying to gracefully shutdown `dotnet
trace`...

1. We can't use `RedirectStandardInput`, because both tools check
   `Console.IsInputRedirected`:

https://github.com/dotnet/diagnostics/blob/0136710bb4c274dd877a965331c75bb34f617dbe/src/Tools/dotnet-dsrouter/DiagnosticsServerRouterCommands.cs#L98
https://github.com/dotnet/diagnostics/blob/0136710bb4c274dd877a965331c75bb34f617dbe/src/Tools/dotnet-trace/CommandLine/Commands/CollectCommand.cs#L79

2. We can't p/invoke `GenerateConsoleCtrlEvent()` on Windows, because
   it would call `FreeConsole()` on the MSBuild process. This breaks
   any long-lived MSBuild nodes until they are stopped manually.

https://stackoverflow.com/a/29274238

So the only thing that works, is to write a console app,
`console-ctrl`, that can call `FreeConsole()` on itself. This
mechanism appears to work, and not break MSBuild.

On non-Windows platforms, we can use `Syscall.kill (pid,
Signum.SIGINT)` from Mono.Unix instead.


jonathanpeppers:
TODO:
1. Test if it works at all on Mac
2. Write a unit test

PR #7096, opened 2022-06-14
[XABT] Support testing with dotnet packs override

pjcollins:
Adds support to the desktop MSBuild tests to find and use build output
in the `bin/$(Configuration)/lib/packs` folder.  This is an extension
of the changes in commits b1232a2b and c21e78ca which allowed local
builds to skip .nupkg packing and installing.

The `Xamarin.Android.Build.Tests` and `MSBuildDeviceIntegration` test
suites will now attempt to run against local build outputs by default,
and fall back to `dotnet/packs` or a legacy system install location if
a local build does not exist.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/TestEnvironment.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/TestEnvironment.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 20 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/TestEnvironment.cs
```diff
@@ -87,32 +83,103 @@ public static VisualStudioInstance GetVisualStudioInstance ()
...
+
+			return dirs.FirstOrDefault ();
+		}
+
+		public static bool UseLocalBuildOutput {
+			get {
+				var msbuildDir = Builder.UseDotNet ? Path.Combine (LocalDotNetAndroidSdkDirectory, "tools") : LocalMonoAndroidToolsDirectory;
+				return Directory.Exists (msbuildDir) && File.Exists (Path.Combine (msbuildDir, "Xamarin.Android.Build.Tasks.dll"));
+			}
+		}
```
jonathanpeppers, line 173: Is the only case this would normally be `false` when running on CI? The test lanes don't have a local build, they just have our packs installed?

Is it worth adding a `<summary>` mentioning this?

jonathanpeppers (approved):
Otherwise, this looks good to me. 👍

I noticed our MSBuild tests didn't work locally when using `BuildDotNet`, I had to use `PackDotNet`.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Common/TestEnvironment.cs
pjcollins: That is correct, this should only be false for tests running on CI. I updated the description and added a comment for this.

PR #7100, opened 2022-06-15
[Mono.Android] Generate docs for any API level

pjcollins:
The `_UpdateExternalDocumentation` target has been updated to allow API
docs to be updated against any API level.  Latest stable versions will
still be used by default, however the `$(AndroidApiLevel)`,
`$(AndroidPlatformId)`, and `$(AndroidFrameworkVersion)` properties can
now be set if needed.  This will let us update the current API 31 docs
without having to release an API 32 update.

The api docs pipeline has been updated to allow these properties to be
set at queue time, and it will now upload a `.diff` file rather than
the entire docs folder.

### DIFF
--- a/src/Mono.Android/Mono.Android.targets
+++ b/src/Mono.Android/Mono.Android.targets
@@ -75,14 +75,14 @@
   <!-- Pulls documentation from JavaDoc -->
   <PropertyGroup>
     <_JavaSourceUtilsJar>$(MicrosoftAndroidSdkOutDir)java-source-utils.jar</_JavaSourceUtilsJar>
-    <_AndroidStableSrcDir>$(AndroidSdkDirectory)\platforms\android-$(AndroidLatestStableApiLevel)\src</_AndroidStableSrcDir>
+    <_AndroidSrcDir>$(AndroidSdkDirectory)\platforms\android-$(AndroidApiLevel)\src</_AndroidSrcDir>
     <_AndroidJavadocXml>..\..\bin\Build$(Configuration)\android-javadoc.xml</_AndroidJavadocXml>
   </PropertyGroup>
   
   <Target Name="_BuildAndroidJavadocXml"
       Condition=" '$(IncludeAndroidJavadoc)' == 'True' "
       BeforeTargets="CoreCompile"
-      Inputs="$(MSBuildThisFile);$(_AndroidStableSrcDir)\source.properties;$(_JavaSourceUtilsJar)"
+      Inputs="$(MSBuildThisFile);$(_AndroidSrcDir)\source.properties;$(_JavaSourceUtilsJar)"
       Outputs="$(_AndroidJavadocXml)">
     <ItemGroup>
       <_Doclink Include="--doc-copyright" />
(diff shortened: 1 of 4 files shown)


pjcollins:
Docs pipeline test run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6289643&view=logs&j=18eaad8b-760d-59fe-74fb-add94ccecaed

pjcollins:
The following options are now available when queuing the API docs pipeline: 
Screen Shot 2022-06-15 at 3 43 39 PM

jonpryor:
The `SuccessfulAndroidXApp` test fails even after (one) re-run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6296228&view=ms.vss-test-web.build-test-results-tab&runId=44170128&resultId=100131&paneView=attachments

```
  /Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/NuGet.targets(131,5): error : Unable to resolve 'Xamarin.AndroidX.AppCompat (>= 1.4.1.1)' for 'MonoAndroid,Version=v12.0'. [/Users/runner/work/1/a/TestRelease/06-16_21.06.42/temp/CodeBehind/SuccessfulAndroidXApp/Release/project/CodeBehindBuildTests.csproj]
```

This is also happening on main for 3d5231ca3a094b85302b5dbd3cc074be37efa2c0: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6296554&view=ms.vss-test-web.build-test-results-tab&runId=44172514&resultId=100131&paneView=attachments

We can thus ignore this failure here.

src/Mono.Android/Mono.Android.targets
```diff
@@ -352,5 +360,15 @@
         WorkingDirectory="$(MSBuildThisFileDirectory)"
     />
   </Target>
-  
+
+  <Target Name="_GenerateApiDocsDiff">
```
jonpryor, line 364: What does this target do?  What's it for?  The commit message should mention it. :-)

pjcollins: The description briefly mentions it:

> The api docs pipeline has been updated to allow these properties to be
set at queue time, and it will now upload a .diff file rather than
the entire docs folder.

Previously the API docs pipeline would upload the updated doc files, and now it will upload this diff file instead.

jonpryor:
```markdown
The `UpdateExternalDocumentation` target (cee5e727) would use
Android SDK sources from the *latest **stable** API level* (a7413a23).
The problem is that frequently Android SDK source "drops" would be
*delayed*: API-32 ("Android 12L") was released on 2021-Dec-8
(6eb11f15), while the *API-32 source package* was released on
2022-05-04 (according to comments in [repository2-3.xml][0]), a
nearly *6 month* delay.

Thus, `xaprepare` has not yet been updated to download and install
the API-32 sources package (8bede23e).

Consequently, if you *do* try to update external documentation:

	msbuild -t:UpdateExternalDocumentation src/Mono.Android/Mono.Android.csproj

*no documentation will be imported*: the default API level is API-32,
for which we do not (yet) install the `sources-32_r01.zip` package,
which means there is nothing for the `UpdateExternalDocumentation`
target to do!

Update the `UpdateExternalDocumentation` target so that instead of
using `$(AndroidLatestStableApiLevel)`, it uses `$(AndroidApiLevel)`.
This allows us to explicitly build `src/Mono.Android` for APi-31
*and import* the API-31 docs:

	msbuild -t:UpdateExternalDocumentation src/Mono.Android/Mono.Android.csproj \
	  -p:AndroidApiLevel=31 -p:AndroidPlatformId=31 -p:AndroidFrameworkVersion=v12.0

The latest API level is still imported by default, if the properties
are not specified on the command-line.

This will allow us to update the current API-31 documentation without
needing to release an API-32 update.

The API docs pipeline (0503d636) has been updated to allow these
properties to be set at queue time.  This pipeline will use the new
`_GenerateApiDocsDiff` target to upload a `.diff` file rather than
the entire docs folder, which should reduce review requirements.

[0]: https://dl-ssl.google.com/android/repository/repository2-3.xml
```

Issue #7102, opened 2022-06-15
Xamarin Android: System.NullReferenceException on any exception thrown inside Task


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity.visualstudio.com/t/Xamarin-Android:-SystemNullReferenceExc/10069660)._

---
If we try to throw any exception inside a Task and not catch it, the original exception is lost and NullReferenceException is raised instead. If we try to catch the exception, it works properly.  
This behavior is not observed in VS2019, it works properly there with this same project.

Attached Screenshot provides all the necessary information.

Stack trace output:
  at Android.Runtime.JNINativeWrapper._unhandled_exception (System.Exception e) [0x0000e] in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:12 
  at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPL_V (_JniMarshal_PPL_V callback, System.IntPtr jnienv, System.IntPtr klazz, System.IntPtr p0) [0x0001d] in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:111 
  at (wrapper native-to-managed) Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPL_V(intptr,intptr,intptr)

---
#### Feedback Bot on 6/14/2022, 01:32 AM: 

(private comment, text removed)

---
### Original Solutions
(no solutions)

jpobst:
Possible duplicate of https://github.com/xamarin/xamarin-android/issues/7085?

jonathanpeppers:
Yeah, I think this will be fixed by: https://github.com/xamarin/xamarin-android/pull/7103

Issue #7085, opened 2022-06-10
The debugger isn't "breaking" correctly when there is an exception in VSMac 2022 / VSMac Preview 2022


xDaijobu:
### Android application type

Classic Xamarin.Android (MonoAndroid12.0, etc.)

### Affected platform version

VSMac & VSMac Preview

### Description

**net6.0-android ( VSMac & VSMac Preview )**
![image](https://user-images.githubusercontent.com/22674537/173105620-0a60e96e-712a-4b57-ac52-f3f172a3c6b5.png)

**Xamarin.Android ( VS Windows )**
![image](https://user-images.githubusercontent.com/22674537/173105215-1e7682cd-b802-49c9-ac40-d49b252a29c6.png)

**Xamarin.Android ( VSMac & VSMac Preview )**
![image](https://user-images.githubusercontent.com/22674537/173105300-0936da65-20c7-4bca-950a-9cb5ae894f11.png)
<img width="2554" alt="Bc6d36a4fbb6a4e8c9ed825676c86a77c637904472730329015_Screen_Shot_2022-06-10_at_15 35 26" src="https://user-images.githubusercontent.com/22674537/173104725-7870ab03-a830-4eb0-a510-9f599c7ae6ee.png">

source: ‘Mono.Android’
stacktrace:
```
[mono-rt] [ERROR] FATAL UNHANDLED EXCEPTION: System.NullReferenceException: Object reference not set to an instance of an object.

at Android.Runtime.JNINativeWrapper._unhandled_exception (System.Exception e) [0x0000e] in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:12
at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPL_V (_JniMarshal_PPL_V callback, System.IntPtr jnienv, System.IntPtr klazz, System.IntPtr p0) [0x0001d] in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:111
at (wrapper native-to-managed) Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPL_V(intptr,intptr,intptr)
```

### Steps to Reproduce

1. Create a Xamarin.Forms / Xamarin.Android project.
2. throw new Exception (“Fire!”);
3. run the project with VSMac / VSMac Preview

[...]

PR #7103, opened 2022-06-15
[Mono.Android] handle NRE in VS Mac when breaking on exceptions

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/7085
Fixes: https://github.com/xamarin/xamarin-android/issues/7102
Fixes: https://dev.azure.com/devdiv/DevDiv/_workitems/edit/1554705

In a new Xamarin.Android project, if you go to your `OnCreate()`
method and add a `throw new Exception("test")` VS Mac breaks on a
`NullReferenceException` such as:

    [mono-rt] [ERROR] FATAL UNHANDLED EXCEPTION: System.NullReferenceException: Object reference not set to an instance of an object.
    at Android.Runtime.JNINativeWrapper._unhandled_exception (System.Exception e) [0x0000e] in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:12
    at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPL_V (_JniMarshal_PPL_V callback, System.IntPtr jnienv, System.IntPtr klazz, System.IntPtr p0) [0x0001d] in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:111
    at (wrapper native-to-managed) Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPL_V(intptr,intptr,intptr)

This does not happen in VS Windows, only VS Mac.

After reviewing my changes in 32cff438, the System.Reflection.Emit
code path does a null check for `mono_unhandled_exception_method`:

    bool  filter = Debugger.IsAttached || !JNIEnv.PropagateExceptions;
    if (filter && JNIEnv.mono_unhandled_exception_method != null) {
        ig.BeginExceptFilterBlock ();

        ig.Emit (OpCodes.Call, JNIEnv.mono_unhandled_exception_method);
        ig.Emit (OpCodes.Ldc_I4_1);
        ig.BeginCatchBlock (null!);
    } else {
        ig.BeginCatchBlock (typeof (Exception));
    }

While the new "fast" code path, does not:

    static bool _unhandled_exception (Exception e)
    {
        if (Debugger.IsAttached || !JNIEnv.PropagateExceptions) {
            JNIEnv.mono_unhandled_exception (e);
            return false;
        }
        return true;
    }

Adding `JNIEnv.mono_unhandled_exception?.Invoke(e)` appears to solve
the problem when I build a local copy of `Mono.Android.dll` and test
it inside VS Mac. I see my exception break properly on the line I
threw the exception.

`mono_unhandled_exception` being null appears to be something
introduced in 2aff4e74. That commit's goal was to not lookup
`mono_unhandled_exception_method` at startup, but wait until an
exception is thrown. Unfortunately, `mono_unhandled_exception_method`
is null at the time that the code S.R.Emitted, so we've had this
behavior for a while!

Since we are looking at reworking this entire system with "marshal
methods" as in #7004, I think we should simply add the null check for
now. We should probably investigate the sequence of events during
startup & unhandled exceptions when the new system is in place.

### DIFF
--- a/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs
+++ b/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs
@@ -9,7 +9,7 @@ namespace Android.Runtime
 		static bool _unhandled_exception (Exception e)
 		{
 			if (Debugger.IsAttached || !JNIEnv.PropagateExceptions) {
-				JNIEnv.mono_unhandled_exception (e);
+				JNIEnv.mono_unhandled_exception?.Invoke (e);
 				return false;
 			}
 			return true;
--- a/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.tt
+++ b/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.tt
@@ -245,7 +245,7 @@ namespace Android.Runtime
 		static bool _unhandled_exception (Exception e)
 		{
 			if (Debugger.IsAttached || !JNIEnv.PropagateExceptions) {
-				JNIEnv.mono_unhandled_exception (e);
+				JNIEnv.mono_unhandled_exception?.Invoke (e);
 				return false;
 			}
 			return true;


dellis1972 (approved):

F-Goncalves:
Any ETA on when this fix will be released?
Or any known workaround to see the underlying exception message?

abhayl21:
I'm getting the same error on VS for Windows v17.2.5
when navigating between pages... it gives me this error.... if i change the navigation sequence this doesn't show up... this has started recently after updating to Xamarin Forms 5



  at Android.Runtime.JNINativeWrapper._unhandled_exception (System.Exception e) [0x0000e] in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:12
  at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V (_JniMarshal_PP_V callback, System.IntPtr jnienv, System.IntPtr klazz) [0x0001c] in /Users/runner/work/1/s/xamarin-android/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs:23
  at (wrapper native-to-managed) Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V(intptr,intptr)

jonathanpeppers:
@abhayl21 if you just hit the play button, does the *real* error show in the console output?

It kind of seems like VS Mac isn't breaking on the right line, but this is somehow working fine in VS Windows?

mahadevans87:
Getting the same error on VS for Windows v17.2.6. I am trying to throw a new Exception("Test"); However, the exception is caught in the unhandled exception function as  - 

at Android.Runtime.JNINativeWrapper._unhandled_exception (System.Exception e) [0x0000e] in ...

MagicAndre1981:
as already said, the fix is part of 17.3 and not 17.2.x

AdamDiament:
> @abhayl21 if you just hit the play button, does the _real_ error show in the console output?
> 
> It kind of seems like VS Mac isn't breaking on the right line, but this is somehow working fine in VS Windows?

@jonathanpeppers speaking for windows, no, I press play and get nothing extra in the output, the app just crashes. I'm on the 17.2.x version. For what it's worth, I upgraded to 17.3.0 preview, and the problem revealed itself. In my case, it was one of these:

```
Xamarin.CommunityToolkit.Exceptions.InvalidCommandParameterException: 'Invalid type for parameter. Expected Type  X but received Y
```

I'm assuming the actual exception type will be different for everyone - but maybe not, just in case.

MagicAndre1981:
> Getting the same error on VS for Windows v17.2.6. 

try the 17.3 Update which is now released:

**Visual Studio for Mac 17.3 is now available**
https://devblogs.microsoft.com/visualstudio/visual-studio-for-mac-17-3-is-now-available/

**Visual Studio 2022 17.3 is now available!**
https://devblogs.microsoft.com/visualstudio/visual-studio-2022-17-3-is-now-available/

PR #7104, opened 2022-06-16
[One .NET] update the .NET 6 packs used in .NET 7

jonathanpeppers:
Context: https://github.com/dotnet/maui/releases/tag/6.0.400
Context: https://github.com/xamarin/xamarin-android/pull/7023

MAUI SR 1 shipped along with a new .NET 6 Android workload. We should
be able to use these packs in .NET 7. This would occur when building
`net6.0-android` apps with a .NET 7 SDK.

Since #7023 is in that release, we can:

* Specify the .NET 6 packs as `sdk` where appropriate.
* Condition `AutoImport.props` to apply for .NET 7-only.
* We also need to remove `$(_DefaultJavaSourceJarPattern)` like we did
  in .NET 6.


jonathanpeppers:
This breaks Xamarin.Legacy.Sdk, need to investigate why.

pjcollins (approved):

PR #5926, opened 2021-05-12
[Xamarin.Android.Build.Tasks] Add Support for AndroidJavaSource to Bindings

dellis1972:
It would be nice to be able to introduce custom Java code and API's to a
binding project. This would allow developers to simplify an API and
still expose that to the C# Developer.

Using the existing `AndroidJavaSource` ItemGroup we can do exactly that.
This ItemGroup now supports the `Bind` metadata which will build and
bind the java code *prior* to the main C# compiled running. Historically
`AndroidJavaSource` was only available on Application Projects. However
this ItemGroup can now be used in any project type. The `Bind` metadata
is `true` by default and will cause the Java code to be compiled into
a .jar file before the C# compiler is run. It will then be bound using
the Java.Interop Java->C# binding system. The final C# apis will be
generated an included in the final C# compilation.

There are some restrictions on the Java code. You can only use Java types
or types which are defined in .jar or .aar files you reference for that
project via `AndroidLibrary` or `EmbeddedJar`.
It is advised that you DO NOT export types which include Java Generics
via this mechanism. Like normal bindings, Java Generics cause significant
problems. Try to stuck to normal types if you can.

PR #7106, opened 2022-06-16
[XABT] Port class-parse/generator response file space fix from 67ce902c.

jpobst:
Fixes: https://github.com/dotnet/maui/issues/8049

There was a bug causing response files to not work when trying to bind Java libraries in a path containing a space, since the CLI argument given to `class-parse`/`generator` was not properly quoted.

This has been fixed in PR https://github.com/xamarin/xamarin-android/pull/5926, however this is a .NET 7 only PR.

Pull the fix out of that PR and port it to the `6.0.4xx` branch so it will go into a .NET 6 servicing release.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/ClassParse.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/ClassParse.cs
@@ -38,7 +38,7 @@ namespace Xamarin.Android.Tasks
 					WriteLine (sw, $"\"{doc}\"");
 			}
 
-			cmd.AppendSwitch ($"@{responseFile}");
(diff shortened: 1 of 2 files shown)


jonathanpeppers (approved):
I backported https://github.com/xamarin/xamarin-android/commit/b21267f4e8b5c89e98b6329399bbb7feb3ea10cb to see if some of these NDK test failures are fixed on this `release/6.0.4xx` branch.

Let me see how the latest run goes, and we might just rebase or `/azp run` this one again.

jonathanpeppers:
Ok, the test failures look better this time. They look unrelated.

PR #7109, opened 2022-06-17
[One .NET] fix missing libaot-* files when building in IDEs

jonathanpeppers:
Fixes: https://github.com/jonathanpeppers/maui-profiling/issues/16#issuecomment-1158183178

Building a .NET 6 Android application in `Release` mode inside Visual
Studio results in `libaot-*.so` files missing from the `.apk`.
However, it works fine at the command-line! This also appears to not
be an issue in Xamarin.Android. I could reproduce this issue in a test.

Reviewing our MSBuild targets, this happens because the list of files
are returned by the `<MonoAOTCompiler/>` task:

    <MonoAOTCompiler ...>
      <Output TaskParameter="CompiledAssemblies" ItemName="_AotCompiledAssemblies" />
      <Output TaskParameter="FileWrites"         ItemName="FileWrites" />
    </MonoAOTCompiler>

The IDE does a `Build`, then `Install` in two steps. `Build` runs the
AOT compiler, but `Install` (runs `SignAndroidPackage`) signs and
creates an `.apk` file.

What's happening is:

* `Build` produces `libaot-*.so` files
* The `_AndroidAot` target is *skipped* during `Install`, and no files
  make it to the `.apk`!

When an MSBuild target is skipped all tasks are skipped, and only
`<ItemGroup>` or `<PropertyGroup>` elements are actually evaluated.

To solve this problem, let's save the output of the
`<MonoAOTCompiler/>` MSBuild task to file. We use the file contents as
the list of files, so it does not matter if MSBuild targets are
skipped or not. We also should use this list of files in `FileWrites`.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
@@ -40,7 +40,20 @@ They run in a context of an inner build with a single $(RuntimeIdentifier).
 
   <Target Name="_AndroidAot"
       Condition=" '$(AotAssemblies)' == 'true' and '$(RuntimeIdentifier)' != '' "
-      DependsOnTargets="_CreatePropertiesCache;_AndroidAotInputs"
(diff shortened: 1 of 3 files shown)


dellis1972 (approved):

jonathanpeppers:
There is a test failure, I need to fix:
```
The target _AndroidAot should have been skipped.
Expected: True
But was:  False
```

PR #7110, opened 2022-06-17
[tests] fix missing .binlog attachments from CodeBehindTests

jonathanpeppers:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6296209&view=ms.vss-test-web.build-test-results-tab&runId=44135466&resultId=100131&paneView=attachments

It looks like we only uploaded `*.log` as attachments in this test.

Let's expand to `*.*log` and pickup `.binlog` files as well.


jonathanpeppers:
This fixed the attachment:
image

PR #7112, opened 2022-06-20
[tests] fix SuccessfulAndroidXApp test

jonathanpeppers:
The `SuccessfulAndroidXApp` test fails to restore with:

    (Restore target) ->
    /Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/NuGet.targets(131,5):
    error : Unable to resolve 'Xamarin.AndroidX.AppCompat (>= 1.4.1.1)' for 'MonoAndroid,Version=v12.0'.

I looked at 1.4.2.1 of the same NuGet, but both versions have
`monoandroid12.0` inside.

I noticed in the `.binlog`:

    NuGet Config files used:
    /Users/runner/.config/NuGet/NuGet.Config

So this test is using the machine-wide `NuGet.config` instead of the
one in the root of the `xamarin-android` repo. I made a change to use
this one, let's see if that solves anything.


jonathanpeppers:
Ok, wow the `NuGet.config` fixed it:

image

dellis1972 (approved):

PR #7118, opened 2022-06-22
Revert "Workaround FirstChanceExceptionEventArgs being trimmed"

akoeplinger:
Reverts xamarin/xamarin-android#6953

The fix for this issue is backported to dotnet/runtime 6.0 now so we can revert the workaround.

### DIFF
--- a/src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Private.CoreLib.xml
+++ b/src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Private.CoreLib.xml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="utf-8" ?>
-<linker>
-	<!-- Workaround for https://github.com/xamarin/xamarin-android/issues/6626, remove once https://github.com/dotnet/runtime/pull/68265 lands -->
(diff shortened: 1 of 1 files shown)


akoeplinger:
@jonpryor ping

jonpryor:
@akoeplinger: sorry for the delay.

Do you have a link handy to the dotnet/runtime commit which contains the fix?

akoeplinger:
@jonpryor https://github.com/dotnet/runtime/pull/68235 for main and https://github.com/dotnet/runtime/pull/68265 for the 6.0 backport

jonathanpeppers (approved):
LGTM, I reran two tests that had some failures.

jonpryor:
Windows build stopped building.  Rebased on main, force-pushed.

jonathanpeppers:
This is good for .NET 7/main. Just FYI, we should be wary backporting this to .NET 6 -- maybe we don't even need to.

We'll need to make sure the dotnet/runtime schedule matches up with ours, if we do backport.

akoeplinger:
Yeah backporting to .NET6 isn't really needed at this point I think since the workaround works and the only impact is the FirstChanceExceptionEventArgs type being preserved unconditionally rather than only when it's used.

PR #7121, opened 2022-06-23
Fix Local DotNet builds

dellis1972:
Commit c21e78ca allowed us to use `dotnet` workloads without having
to call `make pack-dotnet`. We overlooked the copying of the ILLink
PreserveLists xml files into the pack build locations though. As a
result when building an app locally we see the following error

```
error XAGPM7009: System.InvalidOperationException: Unable to find the required Android.Runtime.JNIEnv method tokens
```

This is because the required methods are being linked out by the ILLink
process. This commit makes sure we copy over the PreserveLists xml files
into the packs locations.

This was fine on CI because we always call `make pack-dotnet` as part
of the build process.


jonathanpeppers (approved):

dellis1972:
Test failure seems unrelated. `error ANDZA0000: Unable to open 'obj/Debug/android/bin/com.xamarin.lightlymodifylayout.apk' as zip archive `

Merging :)

PR #7123, opened 2022-06-23
[marshal methods] Part 2 of ?

grendello:
**NOTE: requires https://github.com/xamarin/java.interop/pull/998**
**NOTE: it won't build until :arrow_up: is merged and this PR is updated to pull JI changes in the submodule**

Implement a marshal method candidate classifier which is passed to `JavaCallableWrapperGenerator`
in order to identify methods that can be registered "statically" and for which we can generate
native code instead of using `System.Reflection.Emit`.

The selected methods are marked with the `[UnmanagedCallersOnly]` attribute and their support
code in the form of the connector method (which returns a delegate to the native callback) as
well as its backing field which stores the delegate, are removed from the assemblies.


src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
```diff
@@ -332,17 +359,32 @@ void Run (DirectoryAssemblyResolver res)
...
+				if (Debug) {
+					return;
+				}
+
+				if (!marshalMethodsAssemblyPaths.TryGetValue (name, out HashSet<string> assemblyPaths)) {
+					assemblyPaths = new HashSet<string> ();
+					marshalMethodsAssemblyPaths.Add (name, assemblyPaths);
+				}
+
+				if (!assemblyPaths.Contains (asm.ItemSpec)) {
```
jonpryor, line 375: You don't need `.Contains()` + `.Add()` for a `HashSet<string>`, and the "double-check" is likely *slower* as `.Add()` checks to see if the value is already present *anyway*.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
```diff
@@ -159,6 +164,11 @@ void Run (DirectoryAssemblyResolver res)
 				}
 
 				res.Load (assembly.ItemSpec);
+#if ENABLE_MARSHAL_METHODS
+				if (!Debug) {
```
jonpryor, line 168: Why not skip these `if (!Debug) StoreMarshalAssemblyPath()` checks and just call `StoreMarshalAssemblyPath())`?  `StoreMarshalAssemblyPath()` checks `Debug` anyway, so this is just adding extra levels of complexity & indentation, for no observable benefit.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
```diff
@@ -182,10 +197,6 @@ void Run (DirectoryAssemblyResolver res)
 
 			List<TypeDefinition> allJavaTypes = scanner.GetJavaTypes (allTypemapAssemblies, res);
 
-			// Step 2 - Generate type maps
-			//   Type mappings need to use all the assemblies, always.
-			WriteTypeMappings (allJavaTypes, cache);
```
jonpryor, line 187: Why move this?  Type mappings should be unrelated to the contents of the `.java` files, and `allJavaTypes` appears to be unchanged, so…. What "hidden"/"non-obvious" bits am I not seeing?

src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
```diff
@@ -353,15 +393,19 @@ bool CreateJavaSources (IEnumerable<TypeDefinition> javaTypes, TypeDefinitionCac
...
+						var jti = new JavaCallableWrapperGenerator (t, Log.LogWarning, cache, classifier) {
 							GenerateOnCreateOverrides = generateOnCreateOverrides,
 							ApplicationJavaClass = ApplicationJavaClass,
 							MonoRuntimeInitialization = monoInit,
 						};
 
 						jti.Generate (writer);
 #if ENABLE_MARSHAL_METHODS
-						overriddenMethodDescriptors.AddRange (jti.OverriddenMethodDescriptors);
+						if (!Debug) {
```
jonpryor, line 404: We could remove this `if` + indentation if we just did `if (classifier?.FoundDynamicallyRegisteredMethods)`, thus implicitly relying on the fact that `classifier` is null in Debug builds.

But that means relying on "implicit context".

Not sure which tradeoff is "better": minimizing indentation or keeping things explicit.

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -0,0 +1,134 @@
...
+		ICollection<string> GetAssemblyPaths (AssemblyDefinition asm)
+		{
+			if (!assemblyPaths.TryGetValue (asm.Name.Name, out HashSet<string> paths)) {
+				throw new InvalidOperationException ($"Unable to determine file path for assembly '{asm.Name.Name}'");
+			}
+
+			return paths;
+		}
+
+		CustomAttribute GetUnmanagedCallersOnlyAttribute (AssemblyDefinition targetAssembly, DirectoryAssemblyResolver resolver)
```
jonpryor, line 102: Please rethink the semantics of this method.  As is you're doing *tons* of Cecil type iteration for every assembly int he app, when (current) lines 104-129 really only needs to be done *once*:

```csharp
MethodDefinition  GetUnmanagedCallersOnlyAttributeConstructor (DirectoryAssemblyResolver resolver)
{
    /* lines 103-129, looking up `UnmanagedCallersOnlyAttribute` from `System.Runtime.InteropServices` */
}

CustomAttribute CreateImportedUnmanagedCallersOnlyAttribute (AssemblyDefinition targetAssembly, MethodDefinition ctor)
{
    return new CustomAttribute (targetAssembly.MainModule.ImportReference (ctor));
}
```

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -0,0 +1,134 @@
...
+			this.methods = methods ?? throw new ArgumentNullException (nameof (methods));
+			this.uniqueAssemblies = uniqueAssemblies ?? throw new ArgumentNullException (nameof (uniqueAssemblies));
+			this.assemblyPaths = assemblyPaths ?? throw new ArgumentNullException (nameof (assemblyPaths));
+		}
+
+		public void Rewrite (DirectoryAssemblyResolver resolver)
+		{
+			var unmanagedCallersOnlyAttributes = new Dictionary<AssemblyDefinition, CustomAttribute> ();
+			foreach (AssemblyDefinition asm in uniqueAssemblies) {
+				unmanagedCallersOnlyAttributes.Add (asm, GetUnmanagedCallersOnlyAttribute (asm, resolver));
```
jonpryor, line 28: …then this becomes:

```csharp
MethodDefinition attrCtor = GetUnmanagedCallersOnlyAttributeConstructor (resolver);
var unmanagedCallersOnlyAttributes = new Dictionary<AssemblyDefinition, CustomAttribute> ();
foreach (AssemblyDefinition asm in uniqueAssemblies) {
    unmanagedCallersOnlyAttributes.Add (asm, attrCtor);
}
```

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -0,0 +1,134 @@
...
+
+			Console.WriteLine ("Adding the [UnmanagedCallersOnly] attribute to native callback methods and removing unneeded fields+methods");
+			foreach (MarshalMethodEntry method in methods.Values) {
+				Console.WriteLine ($"\t{method.NativeCallback.FullName} (token: 0x{method.NativeCallback.MetadataToken.RID:x})");
+				method.NativeCallback.CustomAttributes.Add (unmanagedCallersOnlyAttributes [method.NativeCallback.Module.Assembly]);
+				method.Connector.DeclaringType.Methods.Remove (method.Connector);
+				method.CallbackField?.DeclaringType.Fields.Remove (method.CallbackField);
+			}
+
+			Console.WriteLine ();
```
jonpryor, line 39: I think we'll be wanting `Log.LogDebugMessage()`, not `Console.WriteLine()`…

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -0,0 +1,134 @@
...
+					asm.Write (output, writerParams);
+					newAssemblyPaths.Add (output);
+				}
+			}
+
+			// Replace old versions of the assemblies only after we've finished rewriting without issues, otherwise leave the new
+			// versions around.
+			foreach (string path in newAssemblyPaths) {
+				string target = Path.Combine (Path.GetDirectoryName (path), Path.GetFileNameWithoutExtension (path));
+				MoveFile (path, target);
```
jonpryor, line 60: We may want/need to rework the build system so that we write files into new/separate directories instead of modifying files in-place, otherwise we'll provoke the incremental build + file sharing gods.

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsClassifier.cs
```diff
@@ -0,0 +1,388 @@
...
+using Microsoft.Android.Build.Tasks;
+using Microsoft.Build.Utilities;
+using Mono.Cecil;
+
+namespace Xamarin.Android.Tasks
+{
+#if ENABLE_MARSHAL_METHODS
+	public sealed class MarshalMethodEntry
+	{
+		public TypeDefinition TopType             { get; }
```
jonpryor, line 16: `TopType` may be better named `DeclaringType`?  Not sure what "Top" is supposed to mean.

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsClassifier.cs
```diff
@@ -0,0 +1,388 @@
...
+		public MethodDefinition Connector         { get; }
+		public MethodDefinition RegisteredMethod  { get; }
+		public MethodDefinition ImplementedMethod { get; }
+		public FieldDefinition CallbackField      { get; }
+		public string JniTypeName                 { get; }
+		public string JniMethodName               { get; }
+		public string JniMethodSignature          { get; }
+
+		public MarshalMethodEntry (TypeDefinition topType, MethodDefinition nativeCallback, MethodDefinition connector, MethodDefinition
+		                           registeredMethod, MethodDefinition implementedMethod, FieldDefinition callbackField, string jniTypeName,
```
jonpryor, line 27: Indentation convention: an "extra" tab-stop vs. the body of the method, with tabs.  Please do not align with `(`, and start each line with a type:

```csharp
public MarshalMethodEntry (TypeDefinition topType, MethodDefinition nativeCallback, MethodDefinition connector,
[tab][tab]MethodDefinition registeredMethod, MethodDefinition implementedMethod, FieldDefinition callbackField,
[tab][tab]string jniTypeName, string jniName, string jniSignature)
{
[tab]// contents…
}
```

Rationale: if (when?) this method is renamed, all subsequent lines won't need to be altered to re-align with `(`.

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsClassifier.cs
```diff
@@ -0,0 +1,388 @@
...
+		                           string jniName, string jniSignature)
+		{
+			TopType = topType ?? throw new ArgumentNullException (nameof (topType));
+			NativeCallback = nativeCallback ?? throw new ArgumentNullException (nameof (nativeCallback));
+			Connector = connector ?? throw new ArgumentNullException (nameof (connector));
+			RegisteredMethod = registeredMethod ?? throw new ArgumentNullException (nameof (registeredMethod));
+			ImplementedMethod = implementedMethod ?? throw new ArgumentNullException (nameof (implementedMethod));
+			CallbackField = callbackField;
+			JniTypeName = jniTypeName;
```
jonpryor, line 36: Should we also verify that these are not null?

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsClassifier.cs
```diff
@@ -0,0 +1,388 @@
...
+		DirectoryAssemblyResolver resolver;
+		Dictionary<string, IList<MarshalMethodEntry>> marshalMethods;
+		HashSet<AssemblyDefinition> assemblies;
+		TaskLoggingHelper log;
+		bool haveDynamicMethods;
+
+		public IDictionary<string, IList<MarshalMethodEntry>> MarshalMethods => marshalMethods;
+		public ICollection<AssemblyDefinition> Assemblies => assemblies;
+		public bool FoundDynamicallyRegisteredMethods => haveDynamicMethods;
+#endif
```
jonpryor, line 153: Ditto `#endif` comment…

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsClassifier.cs
```diff
@@ -0,0 +1,388 @@
...
+					}
+				}
+
+				return true;
+			}
+		}
+
+		TypeDefinitionCache tdCache;
+		DirectoryAssemblyResolver resolver;
+		Dictionary<string, IList<MarshalMethodEntry>> marshalMethods;
```
jonpryor, line 145: Given that this is a private field, why `IList<MarshalMethodEntry>` and not just `List<MarshalMethodEntry>`?  Interface method invocation is (slightly) slower than class method invocation.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
grendello: In the first case we save one string allocation, in the second instance yes, it's not necessary

src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
grendello: Generating marshal methods also potentially rewrites assemblies, thus changing token ids and type maps must absolutely see the final ones.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
grendello: I much prefer if the code is explicit, makes it easier to understand the flow for someone not familiar with it

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
grendello: The CWLs will be gone eventually, I will remove them in the last PR in the series

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
grendello: I asked @jonathanpeppers about this and he thought it was better to write them in place. We can optimize this in the last PR in the series though

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -0,0 +1,134 @@
...
+				}
+
+				attrConstructor = md;
+			}
+
+			return new CustomAttribute (targetAssembly.MainModule.ImportReference (attrConstructor));
+		}
+	}
+}
+#endif
```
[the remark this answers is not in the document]
grendello: I apply the convention when it's there to stay :) This macro is going to go away in the last PR of the series, it's just a measure to not "poison" current code until marshal methods are finished

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsClassifier.cs
grendello: The field is exported by the `MarshalMethods` property

src/Xamarin.Android.Build.Tasks/Tasks/GenerateJavaStubs.cs
```diff
@@ -332,17 +357,30 @@ void Run (DirectoryAssemblyResolver res)
...
+					marshalMethodsAssemblyPaths.Add (name, assemblyPaths);
+				}
+
+				assemblyPaths.Add (asm.ItemSpec);
+			}
+#endif
 		}
 
-		bool CreateJavaSources (IEnumerable<TypeDefinition> javaTypes, TypeDefinitionCache cache)
+		bool CreateJavaSources (IEnumerable<TypeDefinition> javaTypes, TypeDefinitionCache cache, DirectoryAssemblyResolver resolver, MarshalMethodsClassifier classifier)
```
dellis1972, line 378: The `DirectoryAssemblyResolver` doesn't appear to be used here, is it needed?

grendello: Good catch, it's not anymore (was in an earlier incarnation of the code)

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -0,0 +1,147 @@
...
+			Console.WriteLine ("Method tokens:");
+			foreach (IList<MarshalMethodEntry> methodList in methods.Values) {
+				foreach (MarshalMethodEntry method in methodList) {
+					Console.WriteLine ($"\t{method.NativeCallback.FullName} (token: 0x{method.NativeCallback.MetadataToken.RID:x})");
+				}
+			}
+
+			void MoveFile (string source, string target)
+			{
+				Console.WriteLine ($"Moving '{source}' => '{target}'");
```
dellis1972, line 89: We have a CopyIfChanged methd in `Files` which will only copy a file over if there are changes. It does few other things like making sure the file is writable etc. It might not be as quick, but it does handle a bunch of issue which we have hit before.

dellis1972: That said it won't delete the source file, which I this does.

grendello: Changed to use `CopyIfChanged` and `File.Delete`

jonpryor:
```markdown
[Xamarin.Android.Build.Tasks, monodroid] Marshal Method Classifier

Context: https://github.com/xamarin/xamarin-android/pull/7163

Changes: https://github.com/xamarin/java.interop/compare/c942ab683c12e88e0527ed584a13b411e005ea57...fadbb82c3b8ab7979c19e9f139bdf2589e47549e

  * xamarin/java.interop@fadbb82c: [generator] Add support for @explicitInterface metadata (#1006)
  * xamarin/java.interop@3e4a3c4f: [Java.Interop.Tools.JavaCallableWrappers] JavaCallableMethodClassifier (#998)


xamarin/java.interop@3e4a3c4f reworked how
`Java.Interop.Tools.JavaCallableWrappers.JavaCallableWrapperGenerator`
can be used to find JNI marshal methods.

Update `Xamarin.Android.Build.Tasks.dll` to provide a
`JavaCallableMethodClassifier` to `JavaCallableWrapperGenerator`,
collecting Cecil `MethodDefinition` instances for JNI marshal methods
which can be created at build time via LLVM Marshal Methods.

Methods which cannot be created at build time include methods with
`[Export]`.

Once candidate LLVM marshal methods are found, the marshal method is
updated to have the [`UnmanagedCallersOnlyAttribute` attribute][0],
and related System.Reflection.Emit-based infrastructure such as the
`Get…Handler()` methods and `cb_…` fields are removed.

As with e1af9587, this feature is *not* enabled by default, and
remains a xamarin-android Build time configuration option.

[0]:https://docs.microsoft.com/dotnet/api/system.runtime.interopservices.unmanagedcallersonlyattribute?view=net-6.0
```

PR #7132, opened 2022-06-28
Set TrimMode partial by default

agocke:
Companion to https://github.com/dotnet/linker/pull/2856


src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -64,7 +64,9 @@
     <PublishTrimmed Condition=" '$(PublishTrimmed)' == '' and '$(Configuration)' == 'Release' and '$(AndroidLinkMode)' != 'None' ">true</PublishTrimmed>
     <AndroidLinkMode Condition=" '$(AndroidLinkMode)' == '' and '$(PublishTrimmed)' == 'true' ">SdkOnly</AndroidLinkMode>
     <AndroidLinkMode Condition=" '$(AndroidLinkMode)' == '' ">None</AndroidLinkMode>
-    <TrimmerDefaultAction Condition=" '$(TrimmerDefaultAction)' == '' and '$(AndroidLinkMode)' == 'Full' ">link</TrimmerDefaultAction>
+    <!-- For compat with user code not marked trimmable, only trim opt-in by default. -->
+    <TrimMode Condition=" '$(TrimMode)' == '' and '$(AndroidLinkMode)' == 'Full' ">link</TrimmerDefaultAction>
```
jonathanpeppers, line 68: I think there is a typo:
```suggestion
    <TrimMode Condition=" '$(TrimMode)' == '' and '$(AndroidLinkMode)' == 'Full' ">link</TrimMode>
```

agocke: Nice catch!

agocke:
Note, this trimmode will only be understood once the linker change is merged, so we'll have to wait for that.

agocke:
Worryingly, this passed, so I suspect we're not actually running any trim tests, or that I missed another setting of TrimMode -- I'll double check that.

agocke:
Ah, I see no tests are run here.

jonathanpeppers:
So when the change in [dotnet/linker](https://github.com/dotnet/linker/pull/2856) flows all the way to dotnet/installer, we can try this.

Our Maestro cadence right now is weekly, but I can do it faster if that change lands. Let me know.

agocke:
Yeah, once this goes in I think we'll want to flow through everywhere ASAP to flush out problems. Thanks for your help.

agocke:
SDK change has just moved through been merged, so we probably need to merge it in for this to update.

agocke:
Updated the linker version manually so this should be ready and merged ASAP

agocke:
Note: the SDK change is in p7 so this needs to be in p7 as well

jonpryor:
@agocke : any links to useful documentation about what `$(TrimMode)`=partial means, what it does, how to investigate breakage, etc.?

eng/Versions.props
```diff
@@ -2,7 +2,7 @@
   <!--Package versions-->
   <PropertyGroup>
     <MicrosoftDotnetSdkInternalPackageVersion>7.0.100-preview.7.22354.2</MicrosoftDotnetSdkInternalPackageVersion>
-    <MicrosoftNETILLinkTasksPackageVersion>7.0.100-1.22351.1</MicrosoftNETILLinkTasksPackageVersion>
+    <MicrosoftNETILLinkTasksPackageVersion>7.0.100-1.22362.3</MicrosoftNETILLinkTasksPackageVersion>
```
jonathanpeppers, line 5: Changing this number manually, actually doesn't update the dotnet/linker we are testing against. It only updates the reference assembly this project builds against:

https://github.com/xamarin/xamarin-android/blob/8d22eed397dc56d1ede51010dd192ed65123707f/src/Microsoft.Android.Sdk.ILLink/Microsoft.Android.Sdk.ILLink.csproj#L12

To get a newer dotnet/linker, we bump dotnet/installer.

This one doesn't have your change either (https://github.com/xamarin/xamarin-android/pull/7162), let me try to get newer Maestro bumps in.

jonathanpeppers:
@agocke I think this one would have the changes from dotnet/linker that you need: https://github.com/xamarin/xamarin-android/pull/7170

agocke:
@jonathanpeppers Happy to add a test, but I think android will be broken if we don't merge this soon. The SDK will default to trimming everything unless told otherwise.

jonathanpeppers:
I think we have an existing test for both modes:

https://github.com/xamarin/xamarin-android/blob/6401f2f7da54379b907295b4fe62b46dffea1a54/tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs#L333-L334

`AndroidLinkMode=SdkOnly` should map to `TrimMode=partial`, and `AndroidLinkMode=Full` map to `TrimMode=link`:

https://github.com/xamarin/xamarin-android/blob/d4f94249a1a8c67747f1551627a3796610619c70/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets#L68-L69

This is probably good enough for now, as this test runs for Xamarin.Android and .NET 7+.

jonathanpeppers:
I'm going to merge this into https://github.com/xamarin/xamarin-android/pull/7170, as it is giving some test failures that show `TrimMode=link` is on by default.

PR #7133, opened 2022-06-28
[Xamarin.Android.Build.Tasks] Handle Empty R.txt file.

dellis1972:
Fixes https://i.azdo.io/1439405

The ManagedResourceParser was not handling the case where we do not
get four items per line. As a result it would crash with the following
error.

```
System.IndexOutOfRangeException: Index was outside of bounds of the array.
as Xamarin.And.Tasks.ManagedResourceParser.ProcessRtxtFile (System.String file)
```

If we do not get the expected four items from the line in the file we
should ignore that line. A unit test has been added to check that the
task does not fail when this happens.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
@@ -433,7 +433,7 @@ int xml myxml 0x7f140000
 			CreateResourceDirectory (path);
 			var task = CreateTask (path);
 			task.RTxtFile = Path.Combine (Root, path, "R.txt");
-			File.WriteAllText (task.RTxtFile, string.Empty);
+			File.WriteAllText (task.RTxtFile, Environment.NewLine);
 			Assert.IsTrue (task.Execute (), "Task should have executed successfully.");
 			Assert.IsTrue (File.Exists (task.NetResgenOutputFile), $"{task.NetResgenOutputFile} should have been created.");
 			Directory.Delete (Path.Combine (Root, path), recursive: true);
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
@@ -427,6 +427,19 @@ int xml myxml 0x7f140000
 		}
 
 		[Test]
+		public void GenerateDesignerFileFromEmptyRtxt ()
+		{
+			var path = Path.Combine ("temp", TestName + " Some Space");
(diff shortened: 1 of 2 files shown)


src/Xamarin.Android.Build.Tasks/Utilities/ManagedResourceParser.cs
```diff
@@ -299,6 +299,8 @@ void ProcessRtxtFile (string file)
 			var lines = System.IO.File.ReadLines (file);
 			foreach (var line in lines) {
 				var items = line.Split (new char [] { ' ' }, 4);
+				if (items.Length < 4)
+					continue;
```
jonathanpeppers, line 303: Is it worth logging here? `LogDebugMessage()` with the file path?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Tasks/ManagedResourceParserTests.cs
```diff
@@ -426,6 +426,19 @@ public void GenerateDesignerFileFromRtxt ([Values (false, true)] bool withLibrar
...
 		}
 
+		[Test]
+		public void GenerateDesignerFileFromEmptyRtxt ()
+		{
+			var path = Path.Combine ("temp", TestName + " Some Space");
+			CreateResourceDirectory (path);
+			var task = CreateTask (path);
+			task.RTxtFile = Path.Combine (Root, path, "R.txt");
+			File.WriteAllText (task.RTxtFile, string.Empty);
```
jonpryor, line 436: This test isn't testing what you think it is.  This results in a 0-byte file, and with a 0-byte input, `File.ReadLines()` returns a sequence with 0 elements, meaning we won't actually hit the error condition.

What we *actually* want is a file containing a *blank line*:

```csharp
File.WriteAllText (task.RTxtFile, Environment.NewLine);
```

*Then* `File.ReadLines()` will return a non-empty sequence, and the first (and only) value it returns will be `string.Empty`, and `string.Empty.Split(new char[]{' '}, 4)` returns `new string[]{""}` (a 1 element array), which triggers the error scenario.

src/Xamarin.Android.Build.Tasks/Utilities/ManagedResourceParser.cs
```diff
@@ -299,6 +299,10 @@ void ProcessRtxtFile (string file)
 			var lines = System.IO.File.ReadLines (file);
 			foreach (var line in lines) {
 				var items = line.Split (new char [] { ' ' }, 4);
+				if (items.Length < 4) {
+					Log.LogDebugMessage ($"Ignoring '{line}' from '{file}'. It does not have the correct number of elements.");
```
jonpryor, line 303: Would it be useful to know the line number?  That would be extra work to track, but it *might* be useful to know if this is the 1st line or the 1000th line…?

jonpryor: Could result in a debug message of:

```
{file}:{line_number}: ignoring contents `{line}`, as it does not have the correct number of elements.
```

PR #7135, opened 2022-06-29
Remove deprecated DX and Proguard code

jonathanpeppers:
This removes code related to DX or ProGuard, which already emits an error message if you were trying to use these.

It should drop a lot of parameterized tests, as well as build and package less stuff.


jonathanpeppers:
I think the only failure is `CheckIncludedAssemblies()`:
```
(Restore target) -> 
  /Library/Frameworks/Mono.framework/Versions/6.12.0/lib/mono/msbuild/Current/bin/NuGet.targets(131,5): error : Could not find file '/Users/runner/.nuget/packages/system.runtime.interopservices.windowsruntime/4.0.1/q3bg64o1.8br'. [/Users/runner/work/1/a/TestRelease/06-30_16.23.01/temp/CheckIncludedAssembliesTrue/UnnamedProject.csproj]
```

I think I will probably add `[NonParallelizable]` to this test in a different PR.

PR #7151, opened 2022-07-05
[Xamarin.Android.Build.Tasks] fix @(LibraryProjectZip) and NuGets

jonathanpeppers:
Context: https://github.com/xamarin/GooglePlayServicesComponents/issues/652

In .NET 6, `@(AndroidLibrary)` defaults to `%(Pack)` being true, while
`@(LibraryProjectZip)` mistakenly does not. This would be an issue if
you migrated a project using `@(LibraryProjectZip)` and were not
relying on the default wildcards that would use:

    <AndroidLibrary Include="**/*.aar" />

Let's set this value by default, to ease migration to .NET 6.

I updated a test for this scenario and fixed some assertions that are
working since 2ca2a103 was merged.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.AvailableItems.targets
+++ b/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.AvailableItems.targets
@@ -66,6 +66,9 @@ This item group populates the Build Action drop-down in IDEs.
       <Bind>true</Bind>
       <Pack Condition=" '$(UsingAndroidNETSdk)' == 'true' ">true</Pack>
     </AndroidLibrary>
+    <LibraryProjectZip>
+      <Pack Condition=" '$(UsingAndroidNETSdk)' == 'true' ">true</Pack>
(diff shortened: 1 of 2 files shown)


jpobst (approved):

dellis1972 (approved):

jonpryor:
```markdown
Context: https://github.com/xamarin/GooglePlayServicesComponents/issues/652

In .NET 6, `@(AndroidLibrary.Pack)` defaults to true (97190b6c), while
`%(LibraryProjectZip.Pack)` was never set, and thus was false.  This
would be an issue if you migrated a project using `@(LibraryProjectZip)` and were
not relying on the default wildcards used by:

	<AndroidLibrary Include="**/*.aar" />

The result is that `.aar` files weren't included into the `.nupkg`,
which would result in subsequent build & packaging errors when
using the `.nupkg`, as the `.aar` was missing.

Set `%(LibraryProjectZip.Pack)`=True by default in .NET 6+, to ease
project migration from Classic Xamarin.Android to .NET 6.

I updated a test for this scenario and fixed some assertions that are
working since 2ca2a103 was merged.
```

PR #7156, opened 2022-07-07
[MSBuildDeviceIntegration] Make Debugging Test work under .NET 6+.

dellis1972:
When running the `ApplicationRunsWithDebuggerAndBreaks` unit test under
.NEt 6 we see the following test failure.

```
Error Message:
     Should have hit 2 breakpoints. Only hit 1
  Expected: 2
  But was:  1
```

This is because we are using code like this

```
session.Breakpoints = new BreakpointStore {
   { Path.Combine (Root, appBuilder.ProjectDirectory, "MainActivity.cs"),  19 },
}
```

The line number is hard coded. Now under .NET 6 we seem to be getting a
slightly different file compared to running under Legacy. The usings are
missing. As a result the line numbers have changed. It turns out that
line 19 in this `MainActivity.cs` file is a blank line. Not the
`base.OnCreate (bundle);` which we are expecting. So the breakpoint never
hits.

So lets update these tests to `lookup` the line we want to break on, rather
than hardcoding it. This way if the file does change slightly , as long
as the expected bit of code is present we can find the line number we
want to break on.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -54,7 +54,7 @@ variables:
 - name: RunAllTests
   value: $[or(eq(variables['XA.RunAllTests'], true), eq(variables['IsMonoBranch'], true))]
 - name: DotNetNUnitCategories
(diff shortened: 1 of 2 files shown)


jonathanpeppers (commented):
Would you also want to update this line:

https://github.com/xamarin/xamarin-android/blob/c537dd28c30f482f365ef756214be35aa1553da2/build-tools/automation/azure-pipelines.yaml#L56-L57

Take out where it excludes the `Debugger` test category?

pjcollins (approved):

PR #7158, opened 2022-07-07
[Mono.Android] add RequireViewById<T>() for better NRT support

jonathanpeppers:
Context: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/View.java;l=25322
Context: https://developer.android.com/reference/android/view/View#requireViewById(int)

In porting Xamarin.Android samples to .NET 6, using latest C# 10
features like `Nullable=enable`, you run into a pattern such as:

    var foo = FindViewById<TextView>(Resource.Id.foo);
    ArgumentNullException.ThrowIfNull(foo);
    foo.Text = "bar";

`var` here is a nullable `TextView?`, so you have to null-check. This
is pretty verbose, and actually seems to make NRTs annoying...

API 28 introduced a `requireViewById()`:

https://developer.android.com/reference/android/view/View#requireViewById(int)

Let's introduce a C# version of this, so you can simply do:

    TextView foo = RequireViewById<TextView>(Resource.Id.foo);
    foo.Text = "bar";

Or use `var` if you prefer.

If `foo` was not found, an `ArgumentException` is thrown with a
message containing the target type and hex value of the ID.

I don't think there is a way to easily put the name `foo` in the error
message?

### DIFF
--- a/src/Mono.Android/Android.App/Activity.cs
+++ b/src/Mono.Android/Android.App/Activity.cs
@@ -12,6 +12,18 @@ namespace Android.App {
 			return this.FindViewById (id)!.JavaCast<T> ();
 		}
 
+		// See: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/View.java;l=25322
+		public T RequireViewById<T> (int id)
+			where T : Android.Views.View
+		{
+			var view = FindViewById<T> (id);
+			if (view == null)
+			{
+				throw new ArgumentException ($"ID 0x{id:X} does not reference a View of type '{typeof (T)}' inside this View");
+			}
+			return view;
+		}
+
 		public void StartActivityForResult (Type activityType, int requestCode)
 		{
 			var intent = new Android.Content.Intent (this, activityType);
--- a/src/Mono.Android/Android.App/Activity.cs
+++ b/src/Mono.Android/Android.App/Activity.cs
@@ -18,8 +18,7 @@ namespace Android.App {
 			where T : Android.Views.View
 		{
 			var view = FindViewById<T> (id);
-			if (view == null)
-			{
+			if (view == null) {
 				throw new Java.Lang.IllegalArgumentException ($"ID 0x{id:X} does not reference a View of type '{typeof (T)}' inside this Activity");
 			}
 			return view;
--- a/src/Mono.Android/Android.App/Activity.cs
+++ b/src/Mono.Android/Android.App/Activity.cs
@@ -19,7 +19,7 @@ namespace Android.App {
 		{
 			var view = FindViewById<T> (id);
 			if (view == null) {
-				throw new Java.Lang.IllegalArgumentException ($"ID 0x{id:X} does not reference a View of type '{typeof (T)}' inside this Activity");
+				throw new Java.Lang.IllegalArgumentException ($"Parameter 'id' of value 0x{id:X} does not reference a View of type '{typeof (T)}' inside this Activity");
 			}
 			return view;
 		}
--- a/src/Mono.Android/Android.App/Activity.cs
+++ b/src/Mono.Android/Android.App/Activity.cs
@@ -12,17 +12,19 @@ namespace Android.App {
 			return this.FindViewById (id)!.JavaCast<T> ();
 		}
 
-		// See: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/View.java;l=25322
+#if NET7_0_OR_GREATER || (NET6_0_OR_GREATER && ANDROID_33) || ANDROID_34
+		// See: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/app/Activity.java;l=3430
 		public T RequireViewById<T> (int id)
 			where T : Android.Views.View
 		{
 			var view = FindViewById<T> (id);
 			if (view == null)
 			{
-				throw new ArgumentException ($"ID 0x{id:X} does not reference a View of type '{typeof (T)}' inside this View");
+				throw new Java.Lang.IllegalArgumentException ($"ID 0x{id:X} does not reference a View of type '{typeof (T)}' inside this Activity");
 			}
 			return view;
 		}
+#endif // NET7_0_OR_GREATER || (NET6_0_OR_GREATER && ANDROID_33) || ANDROID_34
 
 		public void StartActivityForResult (Type activityType, int requestCode)
 		{
--- a/src/Mono.Android/Android.Views/View.cs
+++ b/src/Mono.Android/Android.Views/View.cs
@@ -29,6 +29,20 @@ namespace Android.Views {
 			return this.FindViewById (id).JavaCast<T> ();
 		}
 
+#if NET7_0_OR_GREATER || (NET6_0_OR_GREATER && ANDROID_33) || ANDROID_34
+		// See: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/View.java;l=25322
+		public T RequireViewById<T> (int id)
+			where T : Android.Views.View
+		{
+			var view = FindViewById<T> (id);
+			if (view == null)
+			{
+				throw new Java.Lang.IllegalArgumentException ($"ID 0x{id:X} does not reference a View of type '{typeof (T)}' inside this View");
+			}
+			return view;
+		}
+#endif //NET7_0_OR_GREATER || (NET6_0_OR_GREATER && ANDROID_33) || ANDROID_34
+
 		public bool Post (Action action)
 		{
 			return Post (new Java.Lang.Thread.RunnableImplementor (action, true));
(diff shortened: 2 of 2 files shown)


jonpryor:
This PR raises a broader conceptual problem, a discussion that we've "suggested" before but haven't answered.  Guess now is the time?

The assumption / requirement is that if your app builds against e.g. `net6.0-android33.0`, it should *always* built against `net6.0-android33.0`, because all members it references exist in `net6.0-android33.0`.

The problem with this change is that there is no associated "version bump" or `#if` or anything.  Consequently, *if this were merged*, there would be *two* `net6.0-android33.0`s running around:

  * The `net6.0-android33.0` that *doesn't* have `RequireViewById<T>()`, and
  * The `net6.0-android33.0` that *does* have `RequireViewById<T>()`.

Additionally, this impacts *all builds*.  This would add `Activity.RequireViewById<T>()` to Classic `v9.0` (and all others), and if it were used in a NuGet package we could have an eventual state of affairs where e.g. Visual Studio 17.4 is needed to use a NuGet package using v9.0, because *other* v9.0 `Mono.Android.dll` files don't have `Activity.RequireViewById<T>()`.

We have historically been a bit "fast and loose" here -- i.e. not particularly stringent -- in large part because we didn't make that many API additions in the first place.  Thus, we haven't had this discussion in a meaningful manner, and thus we don't know what we can or should do.

Thus, as-is, we can't take this change. 😅

What can/should we do?  A non-exhaustive list of options includes:

  1. Wrap `RequireViewById<T>()` in a `#if ANDROID_34`.  Then it will be auto-included in `net7.0-android34.0` and Classic v14.0 (should it exist).  This unfortunately means we cannot use this member "now", and potentially won't be able to use it until .NET 8.
  2. Wrap `RequireViewById<T>()` in a `#if CONDITION` -- `CONDITION` TBD -- *and also* bump the platform version.  We could then introduce a `net6.0-android33.1` which contains `RequireViewById<T>()`, without altering `net6.0-android33.0`.  Classic would likely still need to wait for the next major API level.

Regardless, we may also need to make the `Mono.Android` version checks *more* stringent, breaking the build if new members are added, to prevent accidental addition of members.  (I'm not sure how easy or tenable this would be for our workflow.)

jonpryor:
Additionally, you mention `View.requireViewById()`.  We have bound this method, and it isn't generic: https://docs.microsoft.com/en-us/dotnet/api/android.views.view.requireviewbyid?view=xamarin-android-sdk-12

If we add an `Activity.RequireViewById<T>()`, we *certainly* should also add a `View.RequireViewById<T>()` "overload" to the existing `View.RequireViewById()` method.

src/Mono.Android/Android.App/Activity.cs
```diff
@@ -12,6 +12,18 @@ partial class Activity {
...
 		}
 
+		// See: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/View.java;l=25322
+		public T RequireViewById<T> (int id)
+			where T : Android.Views.View
+		{
+			var view = FindViewById<T> (id);
+			if (view == null)
+			{
+				throw new ArgumentException ($"ID 0x{id:X} does not reference a View of type '{typeof (T)}' inside this View");
```
jonpryor, line 22: `ArgumentException` should also specify the parameter name, `id`.

jpobst:
We should consider throwing a `Java.Lang.IllegalArgumentException`, as that's what calling the non-generic Java bound method will throw.  It seems a little weird to get a different exception based on it you call the generic version or not.

src/Mono.Android/Android.App/Activity.cs
jonathanpeppers: I changed it to `Java.Lang.IllegalArgumentException` to match the Java/non-generic version. That one doesn't have a second argument, but still seems OK.

src/Mono.Android/Android.Views/View.cs
```diff
@@ -29,6 +29,20 @@ public T FindViewById<T> (int id)
 			return this.FindViewById (id).JavaCast<T> ();
 		}
 
+#if NET7_0_OR_GREATER || (NET6_0_OR_GREATER && ANDROID_33) || ANDROID_34
+		// See: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/View.java;l=25322
+		public T RequireViewById<T> (int id)
+			where T : Android.Views.View
+		{
+			var view = FindViewById<T> (id);
```
jonpryor, line 37: Shouldn't `View.RequireViewById<T>()` call `View.RequireViewById()`?

```csharp
return RequireViewById (id)
    .JavaCast<T>();
```

src/Mono.Android/Android.App/Activity.cs
```diff
@@ -12,6 +12,20 @@ partial class Activity {
...
 		}
 
+#if NET7_0_OR_GREATER || (NET6_0_OR_GREATER && ANDROID_33) || ANDROID_34
+		// See: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/app/Activity.java;l=3430
+		public T RequireViewById<T> (int id)
+			where T : Android.Views.View
+		{
+			var view = FindViewById<T> (id);
+			if (view == null)
+			{
```
jonpryor, line 22: Please place `{` on same line as `if`.

src/Mono.Android/Android.App/Activity.cs
```diff
@@ -12,6 +12,20 @@ partial class Activity {
...
+				throw new Java.Lang.IllegalArgumentException ($"ID 0x{id:X} does not reference a View of type '{typeof (T)}' inside this Activity");
```
jonpryor, line 23: As this exception type *doesn't* have a way to mention the parameter name, should this message instead mention it?

```csharp
"Could not find a View for parameter `id` with value 0x{id:X}!"
```

jonathanpeppers: I used the exact same text on:

https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/app/Activity.java;l=3430

But added the hex value & type.

Maybe I just change `ID` to:

```suggestion
				throw new Java.Lang.IllegalArgumentException ($"Parameter 'id' of value 0x{id:X} does not reference a View of type '{typeof (T)}' inside this Activity");
```

src/Mono.Android/Android.Views/View.cs
jonathanpeppers: If we want to have control over what the error message is, we need to call `FindViewById<T>()` and null check.

jonathanpeppers: `RequireViewById` is also only API 28 and higher. So we're also making a C# version that works all the way back.

jonpryor:
```markdown
Context: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/view/View.java;l=25322
Context: https://developer.android.com/reference/android/view/View#requireViewById(int)

In porting Xamarin.Android samples to .NET 6, using latest C# 10
features like `$(Nullable)=enable`, you run into a pattern such as:

	var foo = FindViewById<TextView>(Resource.Id.foo);
	ArgumentNullException.ThrowIfNull(foo);
	foo.Text = "bar";

`var` here is a nullable `TextView?`, so you have to `null`-check.
This is pretty verbose, and actually seems to make nullable reference
types annoying…

API-28 introduced a [`View.requireViewById(int)` method][0]:

	// Java
	/* partial */ class View {
	    <T> public final T requireViewById (int id);
	}

This was bound as as a non-generic [`View.RequireViewById(int)`][1]
method:

	// C#
	partial class View {
	    public Java.Lang.Object RequireViewById (int id);
	}

"Overload" `View.RequireViewById()` with a generic version, similar
to the existing [`View.FindViewById<T>(int)` overload][2]:

	// C#
	partial class View {
	    public T RequireViewById<T>(int id) => …
	}

Introduce an equivalent `Activity.RequireViewById<T>(int)` method:

	// C#
	partial class Activity {
	    public T RequireViewById<T>(int id) => …
	}

This allows developers to simply:

	TextView foo = RequireViewById<TextView>(Resource.Id.foo);
	foo.Text = "bar";

or use `var` if you prefer.

If the `id` isn't found, a `Java.Lang.IllegalArgumentException` is
thrown with a message containing the `typeof(T)` and hex value of id.

Note: these new `RequireViewById<T>()` methods are only available on:

  * .NET 7 (xamarin-android/main)
  * .NET 6 + API-33 (`net6.0-android33.0`)
  * Future API-34 bindings.

Meaning that "Classic" Xamarin.Android 13.0 will *not* contain these
methods.  It's too late to easily change
`$(TargetFrameworkVersion)`=v13.0, and if we added it we could get
into a state where a future NuGet package used these methods, was
built against Xamarin.Android 13.0, and the new methods could not be
resolved, resulting in build failures.

[0]: https://developer.android.com/reference/android/view/View#requireViewById(int)
[1]: https://docs.microsoft.com/en-us/dotnet/api/android.views.view.requireviewbyid?view=xamarin-android-sdk-12
[2]: https://docs.microsoft.com/en-us/dotnet/api/android.views.view.findviewbyid?view=xamarin-android-sdk-12#android-views-view-findviewbyid-1(system-int32)
```

pjcollins (approved):

PR #7163, opened 2022-07-11
[marshal methods] Part 3 of ?

grendello:
Implement marshal methods LLVM IR executable code generator.

The point of this commit is **only** to generate code and make sure it's valid as far as compiling and linking are concerned.
The code has **not** been tested at run time as not all the infrastructure on the Xamarin.Android side is implemented yet.
This is on purpose, to keep PRs smaller.

The majority of this PR introduces various classes, enums and structures related to code generation.  Support for various LLVM IR instructions
is limited only to those we actually use and only to elements of those constructions that we use. As such, it's not a general purpose code generator which
allows us to make some assumptions and take some shortcuts (without compromising correctness and validity of the generated code)

Portions of the PR (the native type handling system) are to be treated as proof-of-concept as they are not as optimized (design wise) as they should be.
The reason for this limitation is that it requires modifying the previous LLVM IR data generation code and it would contribute to this PR's already substantial
size.  The next PR in the series will take care of that rewrite as well as it will focus on implementing the runtime side of marshal methods, making it
possible to actually run applications which use marshal methods.

What this PR implements is the following:

  * LLVM IR 
    * function and instruction attributes
    * function parameter (declaration/definition) and argument (runtime) handling
    * function variable (including parameters) handling, including unnamed local variables
    * support for native function signatures and pointers to functions
    * The `ret`, `store`, `load`, `icmp`, `br`, `call` and `phi` instructions
  * Marshal method generator
    * managed to JNI signature and symbol name translations


src/Xamarin.Android.Build.Tasks/Utilities/LlvmIrGenerator/FunctionAttributes.cs
```diff
@@ -0,0 +1,793 @@
...
+					RenderParams (sb);
+					sb.Append (')');
+				}
+			} else if (HasValueAsignment) {
+				sb.Append ('=');
+				if (Quoted) {
+					sb.Append ('"');
+				}
+
+				RenderAssignedValue (sb);
```
jonpryor, line 56: Semantically, it feels like `RenderAssignedValue()` should ensure that the assigned value does *not* contain `"`, lest it prematurely end the quoting.  However, none of the overrides do anything to escape or ensure that the value printed doesn't contain `"`; see e.g. `TuneCpuFunctionAttribute`.

Currently, this might not be a problem, but eventually the object graph will be "too big" to easily audit all locations to *know* that `"` won't ever be present in a value.

Should a method be added to deal with escaping?

```csharp
protected void AppendValue (StringBuilder sb, string value)
{
    sb.Append (value.Replace ("\"", "\\\"");
}
```

src/monodroid/jni/monodroid-glue-internal.hh
```diff
@@ -350,7 +350,7 @@ namespace xamarin::android::internal
 		static void monodroid_debugger_unhandled_exception (MonoException *ex);
 
 #if defined (RELEASE) && defined (ANDROID)
-		static void* get_function_pointer (uint32_t mono_image_index, uint32_t class_token, uint32_t method_token) noexcept;
+		static void get_function_pointer (uint32_t mono_image_index, uint32_t class_token, uint32_t method_token, void*& target_ptr) noexcept;
```
jonpryor, line 353: I forget; why make the function pointer an "output" parameter instead of a return value?

src/monodroid/jni/xamarin-android-app-context.cc
```diff
@@ -5,12 +5,12 @@
 
 using namespace xamarin::android::internal;
 
-void* MonodroidRuntime::get_function_pointer (uint32_t mono_image_index, uint32_t class_token, uint32_t method_token) noexcept
+void MonodroidRuntime::get_function_pointer (uint32_t mono_image_index, uint32_t class_index, uint32_t method_token, void *&target_ptr) noexcept
```
jonpryor, line 8: Why rename `class_token` to `class_index`?  `mono_class_get()` uses `type_token` as the parameter name, so a `_token` suffix appears to make more sense?

src/monodroid/jni/monodroid-glue-internal.hh
grendello: It generates faster code

src/monodroid/jni/xamarin-android-app-context.cc
grendello: Because token is no longer used to look up a class. Each generated method passes an index into cache array of class instances:

```
; marshal_methods_class_cache
@marshal_methods_class_cache = global [104 x %struct.MarshalMethodsManagedClass] [
        ; 0
        %struct.MarshalMethodsManagedClass {
                i32 33554780, ; token 0x200015c; class name: Android.Views.View, Mono.Android, Version=0.0.0.0, Culture=neutral, PublicKeyToken=84e04ff9cfb79065
                %struct.MonoClass* null; klass
        }, 
        ; 1
        %struct.MarshalMethodsManagedClass {
                i32 33554448, ; token 0x2000010; class name: AndroidX.Activity.ContextAware.IOnContextAvailableListenerInvoker, Xamarin.AndroidX.Activity, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
                %struct.MonoClass* null; klass
        }, 
        ; 2
        %struct.MarshalMethodsManagedClass {
                i32 33555145, ; token 0x20002c9; class name: Android.Content.IDialogInterfaceOnClickListenerInvoker, Mono.Android, Version=0.0.0.0, Culture=neutral, PublicKeyToken=84e04ff9cfb79065
                %struct.MonoClass* null; klass
        }, 
        ; 3
```

It's faster to look up this way:

```c++
MarshalMethodsManagedClass &klass = marshal_methods_class_cache[class_index];
if (klass.klass == nullptr) {
  klass.klass = mono_class_get (image, klass.token);
}
```

src/Xamarin.Android.Build.Tasks/Utilities/LlvmIrGenerator/FunctionAttributes.cs
grendello: I didn't see any values with quotes in the LLVM IR docs, but I will make that change

jonpryor:
```markdown

Commit e1af9587 mentioned a TODO:

> * Finish the LLVM code generator so that LLVM Marshal Methods are
>   emitted into `libxamarin-app.so`.

Implement LLVM Marshal Methods LLVM-IR executable code generator:

  * LLVM-IR Generation Infrastructure:
    * function and instruction attributes
    * function parameter (declaration/definition) and argument
      (runtime) handling
    * function variable (including parameters) handling, including
      unnamed local variables
    * support for native function signatures and pointers to functions
    * The ret, store, load, icmp, br, call and phi instructions
  * LLVM-IR Marshal Method generator
    * managed to JNI signature and symbol name translations

When `ENABLE_MARSHAL_METHODS` is defined, LLVM-IR code is generated,
compiled, and linked.  Generated code is minimally tested for runtime
behavior, as not all the infrastructure on the Xamarin.Android side
is implemented yet. This is on purpose, to keep commits smaller.

Support for various LLVM IR instructions is limited only to those we
actually use and only to elements of those constructions that we use.
As such, it's not a general purpose code generator which allows us to
make some assumptions and take some shortcuts, without compromising
correctness and validity of the generated code.

The native type handling system portion of this commit is to be
treated as proof-of-concept: is is not as optimized (design wise) as
it should be. The reason for this limitation is that it requires
modifying the previous LLVM IR data generation code and it would
contribute to this commits' already substantial size.

~~ TODO ~~

Update/rewrite infrastructure to focus on implementing the runtime
side of marshal methods, making it possible to actually run
applications which use marshal methods.

Finish prototyping the infrastructure so that a MAUI app can be run
with LLVm Marshal Methods enabled.
```

PR #7168, opened 2022-07-13
Update Apk Size descriptions Script

dellis1972:
Add a shell script for updating these descriptions locally on MacOS/*nix. Since not everyone wants to install PowerShell. 
Update the PowerShell script to use the correct test names, since the Unit tests was moved to a new class.

### DIFF
--- a/Documentation/project-docs/ApkSizeRegressionChecks.md
+++ b/Documentation/project-docs/ApkSizeRegressionChecks.md
@@ -41,7 +41,9 @@ are named like this:
 The new reference files can be obtained from the test results
 archive - artifact of the given CI build (preferred method).
 Or they can be obtained from local build using
-the `build-tools/scripts/UpdateApkSizeReference.ps1` script.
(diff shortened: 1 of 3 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
```diff
@@ -5,49 +5,49 @@
       "Size": 3032
     },
     "assemblies/Java.Interop.dll": {
-      "Size": 59377
+      "Size": 59420
```
jonathanpeppers, line 8: Maybe don't actually update these files here, as they are getting updated in:

https://github.com/xamarin/xamarin-android/pull/7162#issuecomment-1184462655

jonathanpeppers (approved):
Otherwise, the script changes LGTM. 👍

PR #7169, opened 2022-07-14
[Microsoft.Android.Sdk] Prefer API 32 runtime packs

pjcollins:
After backporting stable API 33 bindings, projects that use any of our
`net6.0-android` target framework framework variants (net6.0-android,
net6.0-android32) will attempt to use API 33 runtime packs.

Since the API 33 runtime packs are defined in a different workload that
is not installed by default in net6.0 (android-33 instead of android),
we should prefer API 32 runtime packs when `$(TargetPlatformVersion)` is
less than 33.


src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
```diff
@@ -7,24 +7,22 @@
         "Microsoft.Android.Sdk",
         "Microsoft.Android.Ref.31",
         "Microsoft.Android.Ref.32",
+        "Microsoft.Android.Ref.33",
         "Microsoft.Android.Runtime.32.android-arm",
         "Microsoft.Android.Runtime.32.android-arm64",
         "Microsoft.Android.Runtime.32.android-x86",
         "Microsoft.Android.Runtime.32.android-x64",
```
jonathanpeppers, line 14: Can we also remove these, is there a case where they get used? They can also come down from NuGet.org if they are not in the workload.

pjcollins: Good point, I can't think of anywhere these would be used anymore.

jonathanpeppers (approved):
LGTM if CI is green

PR #7174, opened 2022-07-15
[One .NET] transition to 6.0.400 version band

jonathanpeppers:
Fixes: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1576444

There is a known issue with the VS Mac updater & the 6.0.400 preview
.NET SDK.

The resolution (for now) is to move from:

    Microsoft.NET.Sdk.Android.Manifest-6.0.300

To:

    Microsoft.NET.Sdk.Android.Manifest-6.0.400

Likely further changes would also be needed in the next VS insertion.


pjcollins (approved):

jonathanpeppers:
This looked like the package came out right:

image

Going to merge.

PR #7176, opened 2022-07-15
[One .NET] fix $(DebuggerSupport) for Release builds

jonathanpeppers:
Context: https://github.com/dotnet/sdk/blob/46478aeb4c80b7a52fb76a1affb83ebdb2cca342/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L396-L399
Context: https://github.com/dotnet/runtime/blob/4dd047a78c793ffa88d0c054c41a1200ed8c9901/src/libraries/System.Private.CoreLib/src/ILLink/ILLink.LinkAttributes.Shared.xml#L3

We found that at the time we set `$(DebuggerSupport)`, that `$(AndroidUseDebugRuntime)` is blank.

We can base this value on `$(Optimize)` instead, as it is set very early by the .NET SDK / MSBuild.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
@@ -83,7 +83,7 @@
       Runtime libraries feature switches defaults
       Available feature switches: https://github.com/dotnet/runtime/blob/master/docs/workflow/trimming/feature-switches.md
      -->
-    <DebuggerSupport Condition="'$(DebuggerSupport)' == ''">$(AndroidUseDebugRuntime)</DebuggerSupport>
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
```diff
@@ -5,28 +5,28 @@
...
-      "Size": 6484
+      "Size": 6428
     },
     "assemblies/System.Linq.dll": {
-      "Size": 9980
+      "Size": 9103
     },
     "assemblies/System.Private.CoreLib.dll": {
-      "Size": 532428
+      "Size": 472318
```
jonathanpeppers, line 23: This is the one that this makes a huge difference for.

grendello (approved):

dellis1972 (approved):

PR #7186, opened 2022-07-20
[ci] Rename job names

jonpryor:
Certain tools such as [Component Governance][0] only use the Pipeline
name -- as defined somewhere in Azure DevOps -- and the `displayName`
property on the [`job` definition][1].  The `displayName` property on
the containing [`stage`][2] is *not* displayed.

This can result in "wonderful" alerts which state that the affected
pipeline is e.g. "Xamarin.Android (macOS)", which isn't particularly
helpful when there are *three* jobs with a `displayName` of `macOS`:

  * `designer_integration_mac`
  * `mac_bcl_tests`
  * `mac_build_create_installers`

Fully understanding these alerts is irksome.

Improve this situation by introducing "redundancy" into the job
`displayName` values, and updating them to follow a pattern:

 1. Host Platform
 2. Category
 3. (Optional) Description

Each part is separated from the others by a ` > `.

Examples:

  * Replace `Mac Build` with `macOS > Build`.
  * For the previous `macOS` displayName:
    * The `designer_integration_mac` job is `macOS > Tests > Designer Integration`
    * The `mac_bcl_tests` job is `macOS > Tests > BCL (Emulator)`
    * The `mac_build_create_installers` job is `macOS > Create Installers`
  * Replace `Build and Smoke Test` with `Windows > Build & Smoke Test`

[0]: https://devdiv.visualstudio.com/DevDiv/_componentGovernance/112013?_a=alerts&typeId=6317076&alerts-view-option=active
[1]: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/jobs-job?view=azure-pipelines
[2]: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/stages-stage?view=azure-pipelines

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -81,10 +81,10 @@ variables:
 # Stage and Job "display names" are shortened because they are combined to form the name of the corresponding GitHub check.
 stages:
 - stage: mac_build
-  displayName: macOS Build
+  displayName: Mac
   dependsOn: []
   jobs:
-  # Check - "Xamarin.Android (Mac Build)"
+  # Check - "Xamarin.Android (macOS > Build)"
   - job: mac_build_create_installers
     displayName: macOS > Build
     pool:
@@ -142,10 +142,10 @@ stages:
 
 # This stage ensures Windows specific build steps continue to work, and runs unit tests.
 - stage: win_build_test
-  displayName: Windows Build & Smoke Test
+  displayName: Windows
   dependsOn: []
   jobs:
-  # Check - "Xamarin.Android (Windows Build and Smoke Test)"
+  # Check - "Xamarin.Android (Windows > Build & Smoke Test)"
   - job: win_build_test
     displayName: Windows > Build & Smoke Test
     pool: $(1ESWindowsPool)
@@ -257,9 +257,9 @@ stages:
 
     - template: yaml-templates\fail-on-issue.yaml
 
-# Check - "Xamarin.Android (Linux Build)"
+# Check - "Xamarin.Android (Linux > Build)"
 - stage: linux_build
-  displayName: Linux Build
+  displayName: Linux
   dependsOn: []
   jobs:
   - job: linux_build_create_sdk_pack
@@ -332,10 +332,10 @@ stages:
     - template: yaml-templates/fail-on-issue.yaml
 
 - stage: smoke_tests
-  displayName: macOS > Smoke Tests
+  displayName: Smoke Tests
   dependsOn: mac_build
   jobs:
-  # Check - "Xamarin.Android (Smoke Tests APKs Legacy - macOS)"
+  # Check - "Xamarin.Android (macOS > Tests > APKs Classic)"
   - job: mac_apk_tests_legacy
     displayName: macOS > Tests > APKs Classic
     pool:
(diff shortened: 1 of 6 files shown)


build-tools/automation/azure-pipelines.yaml
```diff
@@ -333,11 +333,11 @@ stages:
 
 - stage: smoke_tests
   displayName: Smoke Tests
-  dependsOn: mac_build
+  dependsOn: macOS > Smoke Tests
```
pjcollins, line 336: I think this change was intended for `displayName` rather than `dependsOn`.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -332,12 +332,12 @@ stages:
     - template: yaml-templates/fail-on-issue.yaml
 
 - stage: smoke_tests
-  displayName: Smoke Tests
+  displayName: macOS > Smoke Tests
```
pjcollins, line 335: In order to reduce the verbosity in the GitHub checks that get reported back to the repo, maybe we should keep the stage names mostly unchanged and as simple as possible, while making job names more verbose.

The rename of the `Smoke Tests` stage to `macOS > Smoke Tests` is a bit odd, as the smoke test stage contains a job that runs on windows. Unless part of the intent behind the rename was to include stage dependencies in the name?

I would suggest something like this instead (leaving stage display names mostly unchanged):

Stage: `Smoke Tests`
Job: `macOS > Tests > APKs Classic`
Resulting GitHub status: `Xamarin.Android-PR (Smoke Tests macOS > Tests > Apks Classic`

We could also continue to use `macOS`, `Windows`, and `Linux` as stage names for the earlier build stages.

build-tools/automation/yaml-templates/run-msbuild-device-tests.yaml
```diff
@@ -11,7 +11,7 @@ parameters:
 
 jobs:
   - job: ${{ parameters.job_name }}
-    displayName: MSBuild With Emulator - macOS-${{ parameters.node_id }} - ${{ parameters.job_suffix }}
+    displayName: macOS > Tests > MSBuild+Emulator ${{ parameters.job_suffix }} \#${{ parameters.node_id }}
```
jonpryor, line 14: :rofl: , the `\` appears in the "alt/hover text" in AzDO.

I'll need to try *quoting* the line to allow `#` in there.

build-tools/automation/azure-pipelines.yaml
jonpryor: I've reverted the `- stage > displayName` changes, so the historical stage names should be used once more.

I also quoted `displayName` values which contain `#`, as the "alt/hover text" for those entries were rendered as `… \# …`, which was _not_ the intent.

pjcollins: Works for me, I think we may still end up with some "clunky" names but that's a fair trade off for better traceability.

pjcollins (approved):

jonpryor:
Just looked at the GitHub "status checks" section of this page, which *does* include the stage displayName.  Consequently, I see values such as:

  * Xamarin.Android-PR (Linux Linux > Build)
  * Xamarin.Android-PR (Mac macOS > Build)
  * Xamarin.Android-PR (Windows Windows > Build & Smoke Test)

It's…kinda funny?  There's no appeasing all the tooling, is there?

Issue #7183, opened 2022-07-20
Rethink .NET Workload Pack Structure


jonpryor:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

.NET 6.0.400 / 32.0.440

### Description

Context? 0ab3db079d43d47ad1bc522e28e65204542fea60

Today, when we build a .NET 6 app, we build the app against the latest runtime pack, no matter what `$(TargetFramework)` was specified.  If `$(TargetFramework)`=net6.0-android (API-31), then the app assembly is built against the API-31 reference pack, *but* the App is *packaged* using the API-33 assemblies.

The result is that in some circumstances, API-33 members may be used by Java Callable Wrappers, even if API-33 isn't installed; even if API-33 `android.jar` *is* installed, it won't be *used* by JCW compilation.

---

The original reason for using the latest runtime pack during app build (TODO: commit?) is because `libmonodroid.so` (via `libmono-android.debug.so`, `libmono-android.release.so`, and `libxamarin-debug-app-helper.so`) is "tied at the hip" to `Xamarin.Android.Build.Tasks.dll` (XABT), as XABT emits LLVM-IR/assembly/etc. which depends upon internals of `libmonodroid.so`.  As the runtime pack contains the `libmonodroid.so` files, it thus felt appropriate to always use the latest runtime pack, as that was the version that would match the SDK pack (which contains XABT).

We are now hitting an unexpected corner case of this arrangement.

What we should instead do is place related components into the same pack.  As `libmonodroid.so` & XABT are tied to each other, `libmonodroid.so` should instead be distributed in the SDK packs, *not* the runtime packs.

### Steps to Reproduce

Create a new Android app:

    dotnet new android -n implicit_api33

Edit `MainActivity.cs` to add:

```csharp
sealed class MyThread : global::Java.Lang.Thread
{
}
```

Build the app.

Expected results: it builds!  (It *does* build with .NET 6.0.300…)

Actual results (with .NET 6.0.400): it *doesn't* build; Java Callable Wrapper compilation fails:

```
JAVAC0000 error: no suitable constructor found for Thread(ThreadGroup,Runnable,String,long.boolean)
```

### Did you find any workaround?

Build the app against API-33, by setting `$(TargetFramework)`=net6.0-android33.0.

jonpryor:
Via:

  * https://discord.com/channels/732297728826277939/732297837953679412/999317617301004419
[...]

PR #7187, opened 2022-07-20
[One .NET] produce 31, 32, 33 packs

jonathanpeppers:
Context: https://github.com/xamarin/xamarin-android/issues/7183

Adding a class like this to a `dotnet new android` project:

    class MyThread : Java.Lang.Thread { }

Causes the build to fail with:

    obj\Debug\net6.0-android\android\src\crc6474063b540de74d1e\RenderingThread.java(76,3): error JAVAC0000:  error: no suitable constructor found for Thread(ThreadGroup,Runnable,String,long,boolean)
      super (p0, p1, p2, p3, p4);
    constructor Thread.Thread() is not applicable
      (actual and formal argument lists differ in length)
    constructor Thread.Thread(Runnable) is not applicable
      (actual and formal argument lists differ in length)
    constructor Thread.Thread(ThreadGroup,Runnable) is not applicable
      (actual and formal argument lists differ in length)
    constructor Thread.Thread(String) is not applicable
      (actual and formal argument lists differ in length)
    constructor Thread.Thread(ThreadGroup,String) is not applicable
      (actual and formal argument lists differ in length)
    constructor Thread.Thread(Runnable,String) is not applicable
      (actual and formal argument lists differ in length)
    constructor Thread.Thread(ThreadGroup,Runnable,String) is not applicable
      (actual and formal argument lists differ in length)
    constructor Thread.Thread(ThreadGroup,Runnable,String,long) is not applicable
      (actual and formal argument lists differ in length)

This is a new `Thread` constructor added in API 33:

https://developer.android.com/reference/java/lang/Thread#Thread(java.lang.ThreadGroup,%20java.lang.Runnable,%20java.lang.String,%20long,%20boolean)

What's happening here is a result of our workload authoring...

* `net6.0-android` defaults to `$(TargetPlatformVersion)` 31

* You get a `Microsoft.Android.Ref.31` reference assembly for
  `Mono.Android.dll`

* However, we mistakenly thought a `Microsoft.Android.Runtime.33`
  "runtime" assembly for `Mono.Android.dll` would work!

And so the `<GenerateJavaStubs/>` MSBuild target runs against an API
33 `Mono.Android.dll`, generating a `Thread` constructor that doesn't
compile.

To fix this:

* `Mono.Android.Apis.projitems` now contains metadata for each API
  level, such as:

```xml
<AndroidApiInfo Include="v11.0">
  <Name>R</Name>
  <Level>30</Level>
  <Id>30</Id>
  <Stable>True</Stable>
  <DotNetSupported>False</DotNetSupported>
</AndroidApiInfo>
<AndroidApiInfo Include="v12.0">
  <Name>S</Name>
  <Level>31</Level>
  <Id>31</Id>
  <Stable>True</Stable>
  <DotNetSupported>True</DotNetSupported>
</AndroidApiInfo>
```

* Build API 31, 32, 33 runtime packs based on
  `%(AndroidApiInfo.DotNetSupported)`. We will ship these going forward.

* The workload now installs 31 packs by default. The API 32 & 33 packs
  can be restored from NuGet, or via the `android-32` or `android-33`
  workloads.

I added a test for this scenario that validates the runtime pack used
during a build.

I will port these changes to xamarin-android/main in a future PR.

### DIFF
--- a/build-tools/create-packs/Directory.Build.targets
+++ b/build-tools/create-packs/Directory.Build.targets
@@ -56,26 +56,28 @@
     <RemoveDir Directories="$(XamarinAndroidSourcePath)bin\Build$(Configuration)\nuget-unsigned" />
   </Target>
 
-  <Target Name="_CreateDefaultRefPack"
-      Condition=" '$(AndroidLatestStableApiLevel)' != '$(AndroidDefaultTargetDotnetApiLevel)' and Exists('$(XAInstallPrefix)xbuild-frameworks\Microsoft.Android\net6.0-android$(AndroidDefaultTargetDotnetApiLevel)\Mono.Android.dll') ">
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=$(AndroidDefaultTargetDotnetApiLevel) &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Ref.proj&quot;" />
-  </Target>
-
-  <Target Name="_CreateAPI32Packs">
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=32 -p:AndroidRID=android-arm -p:AndroidABI=armeabi-v7a-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=32 -p:AndroidRID=android-arm64 -p:AndroidABI=arm64-v8a-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=32 -p:AndroidRID=android-x86 -p:AndroidABI=x86-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=32 -p:AndroidRID=android-x64 -p:AndroidABI=x86_64-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=32 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Ref.proj&quot;" />
-  </Target>
-
   <Target Name="CreateAllPacks"
-      DependsOnTargets="DeleteExtractedWorkloadPacks;_SetGlobalProperties;GetXAVersionInfo;_CleanNuGetDirectory;_CreateAPI32Packs;_CreateDefaultRefPack">
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidRID=android-arm -p:AndroidABI=armeabi-v7a-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidRID=android-arm64 -p:AndroidABI=arm64-v8a-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidRID=android-x86 -p:AndroidABI=x86-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidRID=android-x64 -p:AndroidABI=x86_64-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
-    <Exec Command="dotnet pack @(_GlobalProperties, ' ') &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Ref.proj&quot;" />
+      DependsOnTargets="DeleteExtractedWorkloadPacks;_SetGlobalProperties;GetXAVersionInfo;_CleanNuGetDirectory">
+    <Exec
+        Condition=" '%(AndroidApiInfo.DotNetSupported)' == 'true' "
+        Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=%(AndroidApiInfo.Level) -p:AndroidRID=android-arm -p:AndroidABI=armeabi-v7a-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;"
+    />
+    <Exec
+        Condition=" '%(AndroidApiInfo.DotNetSupported)' == 'true' "
+        Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=%(AndroidApiInfo.Level) -p:AndroidRID=android-arm64 -p:AndroidABI=arm64-v8a-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;"
+    />
+    <Exec
+        Condition=" '%(AndroidApiInfo.DotNetSupported)' == 'true' "
+        Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=%(AndroidApiInfo.Level) -p:AndroidRID=android-x86 -p:AndroidABI=x86-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;"
+    />
+    <Exec
+        Condition=" '%(AndroidApiInfo.DotNetSupported)' == 'true' "
+        Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=%(AndroidApiInfo.Level) -p:AndroidRID=android-x64 -p:AndroidABI=x86_64-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;"
+    />
+    <Exec
+        Condition=" '%(AndroidApiInfo.DotNetSupported)' == 'true' "
+        Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=%(AndroidApiInfo.Level) &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Ref.proj&quot;"
+    />
     <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:HostOS=Linux   &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Sdk.proj&quot;" Condition=" '$(HostOS)' == 'Linux' " />
     <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:HostOS=Darwin  &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Sdk.proj&quot;" Condition=" '$(HostOS)' == 'Darwin' " />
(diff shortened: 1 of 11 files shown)


build-tools/create-packs/Directory.Build.targets
```diff
@@ -61,16 +61,31 @@
...
+        Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidApiLevel=%(AndroidApiInfo.Level) &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Ref.proj&quot;"
+    />
   </Target>
 
   <Target Name="CreateAllPacks"
-      DependsOnTargets="DeleteExtractedWorkloadPacks;_SetGlobalProperties;GetXAVersionInfo;_CleanNuGetDirectory;_CreateAPI32Packs;_CreateDefaultRefPack">
+      DependsOnTargets="DeleteExtractedWorkloadPacks;_SetGlobalProperties;GetXAVersionInfo;_CleanNuGetDirectory;_CreateOtherApiLevels;_CreateDefaultRefPack">
     <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidRID=android-arm -p:AndroidABI=armeabi-v7a-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
     <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidRID=android-arm64 -p:AndroidABI=arm64-v8a-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
     <Exec Command="dotnet pack @(_GlobalProperties, ' ') -p:AndroidRID=android-x86 -p:AndroidABI=x86-net6 &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Runtime.proj&quot;" />
```
pjcollins, line 91: I think these ref/runtime pack creation calls for the default API level are no longer needed with these changes. Maybe we can remove these and the `_CreateOtherApiLevels` target, and move the new Exec calls down here instead?

build-tools/xaprepare/xaprepare/Application/GeneratedMonoAndroidProjitemsFile.cs
```diff
@@ -52,6 +55,7 @@ void WriteGroupApiInfo (StreamWriter sw, AndroidPlatform androidPlatform)
 			sw.WriteLine ($"      <Level>{androidPlatform.ApiLevel}</Level>");
 			sw.WriteLine ($"      <Id>{androidPlatform.PlatformID}</Id>");
 			sw.WriteLine ($"      <Stable>{androidPlatform.Stable}</Stable>");
+			sw.WriteLine ($"      <DotNet>{androidPlatform.ApiLevel >= dotnetApiLevel}</DotNet>");
```
pjcollins, line 58: It might be more clear to rename this, `<DotNetSupported>` or something like that?

build-tools/create-packs/Directory.Build.targets
jonathanpeppers: I removed all the extra targets, so it batches each API level now for the runtime and ref packs.

jonpryor:
Many unit test jobs are failing in `xaprepare`:

```
System.InvalidOperationException: Step Xamarin.Android.Prepare.Step_GenerateFiles failed: Could not find a part of the path '/Users/runner/work/1/s/xamarin-android/bin/Release/lib/xamarin.android/xbuild-frameworks/MonoAndroid/v1.0/Facades'.
 ---> System.IO.DirectoryNotFoundException: Could not find a part of the path '/Users/runner/work/1/s/xamarin-android/bin/Release/lib/xamarin.android/xbuild-frameworks/MonoAndroid/v1.0/Facades'.
```

I'm restarting the failed jobs, but this feels weird.

build-tools/xaprepare/xaprepare/Scenarios/Scenario_AndroidTestDependencies.cs
```diff
@@ -15,6 +15,7 @@ protected override void AddSteps (Context context)
 			Steps.Add (new Step_InstallAdoptOpenJDK8 ());
 			Steps.Add (new Step_InstallMicrosoftOpenJDK11 ());
 			Steps.Add (new Step_Android_SDK_NDK (AndroidToolchainComponentType.CoreDependency));
+			Steps.Add (new Step_GenerateFiles (atBuildStart: false));
```
jonathanpeppers, line 18: @jonpryor yes, I found for this to work, we need to generate `bin\Build$(Configuration)\Mono.Android.Apis.projitems` on test lanes...

Having trouble figuring out what's wrong here...

jonathanpeppers:
There is 1 unrelated `DotNetNew("androidwear")` test failure, I'll fix in a different PR tomorrow.

jonathanpeppers:
This should fix `DotNetNew("androidwear")`: https://github.com/xamarin/xamarin-android/pull/7196

pjcollins (approved):

grendello (approved):

jonpryor:
```markdown
Context: https://github.com/xamarin/xamarin-android/issues/7183
Context: https://bugzilla.xamarin.com/11/1114/bug.html
Context: https://bugzilla.xamarin.com/23/2367/bug.html
Context: https://github.com/xamarin/monodroid/commit/8c03aba6aabeec7fbd5a7425aeff9d73a4a19952
Context: https://github.com/xamarin/monodroid/commit/78ab9b5539d031ff3c334430295b6e2fcb346498
Context: https://github.com/xamarin/monodroid/commit/6679dfc62eae462b5acc9deb095d0fa41786f80b

**Background**: Java Callable Wrapper (JCW) constructors!  Consider the
following C# class:

	public class MyThread1 : Java.Lang.Thread {
	}

This type has one constructor: the implicit default constructor.

How many constructors does the JCW contain?

At least 8:

	// Java
	public /* partial */ class MyThread1 extends java.lang.Thread /* … */ {
	    public MyThread1();
	    public MyThread1(Rnnable target);
	    public MyThread1(ThreadGroup group, Runnable target);
	    public MyThread1(String name);
	    public MyThread1(ThreadGroup group, String name);
	    public MyThread1(Runnable target, String name);
	    public MyThread1(ThreadGroup group, Runnable target, String name);
	    public MyThread1(ThreadGroup group, Runnable target, String name, long stackSize);
	}

Wait, what?  *Why*?  Constructor generation is unexpectedly hard.

Aside: consider the binding for the `Thread(String)` constructor:

	/* partial */ class Thread : Object, IRunnable {
	    [Register (".ctor", "(Ljava/lang/String;)V", "")]
	    public unsafe Thread (string name) : base (IntPtr.Zero, JniHandleOwnership.DoNotTransfer)
	    {
	        const string __id = "(Ljava/lang/String;)V";

	        if (((global::Java.Lang.Object) this).Handle != IntPtr.Zero)
	            return;

	        IntPtr native_name = JNIEnv.NewString ((string?)name);
	        try {
	            __args [0] = new JniArgumentValue (native_name);
	            var __r = _members.InstanceMethods.StartCreateInstance (__id, ((object) this).GetType (), __args);
	            SetHandle (__r.Handle, JniHandleOwnership.TransferLocalRef);
	            _members.InstanceMethods.FinishCreateInstance (__id, this, __args);
	        } finally {
	            JNIEnv.DeleteLocalRef (native_name);
	        }
	    }
	}

The key part are the calls to
[`JniPeerMembers.JniInstanceMethods.StartCreateInstance()`][0] and
[`JniPeerMembers.JniInstanceMethods.FinishCreateInstance()`][1], which:

 1. Gets the JCW corresponding to the *runtime* type `this.GetType()`,
    i.e. `MyThread1`

 2. Looks up the Java-side constructor on the JCW from (1) which has
    the JNI signature in `__id`.

For `MyThread1`, this will lookup the `MyThread1()` default
constructor in the JCW.

Now consider `MyThread2`:

	// C#
	public class MyThread2 : Thread {
	    public MyThread2()
	        : base ("name)
	    {
	    }
	}

Here, the `MyThread2()` default constructor invokes the
`Java.Lang.Thread(string)` constructor, which will attempt to find
the Java Callable Wrapper constructor `MyThread2(String)`!  This in
turn requires that the `MyThread2(String)` constructor exist.

We cannot have a "direct" relationship between the managed type and
the Java Callable Wrapper.  *At minimum*, there needs to be a
`MyThread2(String)` constructor, even though `MyThread2` doesn't have
such a constructor.

The only way to know which constructors need to be generated is by
disassembling the IL to see which constructor is invoked, which was
more work than we wanted to do.

So what we instead do is emit *all possible constructor overloads*.
This ensures that the `MyThread2(String)` JCW constructor will exist
when needed, at the cost of extra unusable constructors also existing.

So far so good; things have worked this way for over 10 years.

**The scenario**:

 1. Create a new default .NET 6 project, using the default
    `$(TargetFramework)` value of `net6.0-android` (API-31):

        dotnet new android

 2. Add a `Java.Lang.Thread` subclass to the new project:

        public class MyThread : Java.Lang.Thread {
        }

 3. Build!

Expected results: it builds.

Actual results: it fails (?!):

	obj\Debug\net6.0-android\android\src\crc6474063b540de74d1e\RenderingThread.java(76,3): error JAVAC0000:  error: no suitable constructor found for Thread(ThreadGroup,Runnable,String,long,boolean)
	  super (p0, p1, p2, p3, p4);
	constructor Thread.Thread() is not applicable
	  (actual and formal argument lists differ in length)
	constructor Thread.Thread(Runnable) is not applicable
	  (actual and formal argument lists differ in length)
	constructor Thread.Thread(ThreadGroup,Runnable) is not applicable
	  (actual and formal argument lists differ in length)
	constructor Thread.Thread(String) is not applicable
	  (actual and formal argument lists differ in length)
	constructor Thread.Thread(ThreadGroup,String) is not applicable
	  (actual and formal argument lists differ in length)
	constructor Thread.Thread(Runnable,String) is not applicable
	  (actual and formal argument lists differ in length)
	constructor Thread.Thread(ThreadGroup,Runnable,String) is not applicable
	  (actual and formal argument lists differ in length)
	constructor Thread.Thread(ThreadGroup,Runnable,String,long) is not applicable
	  (actual and formal argument lists differ in length)

**Discussion**: This build break is caused by the combination of:

 1. Commit df6c716e, which binds API-33 and declares it stable, and
 1. Commit 0ab3db07, wherein we decreed that when building an App
    project, the App is packaged against the *latest runtime pack*. 

(1) adds a new `Java.Lang.Thread` constructor from API-33:
[`Thread(ThreadGrouop, Runnable, String, long, boolean)`][2].

(2) means that the app, which was *built against* API-31, is
*packaged against* API-33.  This in turn means that the JCWs are
generated from an API-33 source, and thus the new API-33 constructor
is emitted as one of the `MyThread` constructors.  This in turn
requires that the app be built against an API-33 `android.jar`, which
isn't used because `$(TargetFramework)` specifies API-31.

**The workaround**: set `$(TargetFramework)`=net6.0-android33.0.

**The fix**: 

Update `Mono.Android.Apis.projitems` to now contains
`%(AndroidApiInfo.DotNetSupported)` item metadata for each API level:

	<AndroidApiInfo Include="v11.0">
	  <Name>R</Name>
	  <Level>30</Level>
	  <Id>30</Id>
	  <Stable>True</Stable>
	  <DotNetSupported>False</DotNetSupported>
	</AndroidApiInfo>
	<AndroidApiInfo Include="v12.0">
	  <Name>S</Name>
	  <Level>31</Level>
	  <Id>31</Id>
	  <Stable>True</Stable>
	  <DotNetSupported>True</DotNetSupported>
	</AndroidApiInfo>

We will now build .NET runtime packs whenever
`%(AndroidApiInfo.DotNetSupported)`=True, specifically for API-31,
API-32, and API-33.  We will ship these going forward.

The workload now installs API-31 runtime packs by default.
The API-32 and API-33 runtime packs can be restored from NuGet, or
via the `android-32` or `android-33` workloads.

I added a test for this scenario that validates the runtime pack used
during a build.

I will port these changes to xamarin-android/main in a future PR.

[0]: https://github.com/xamarin/java.interop/blob/247145e4eed6d7c8dcef947a657f88da6663ad69/src/Java.Interop/Java.Interop/JniPeerMembers.JniInstanceMethods.cs#L129-L144
[1]: https://github.com/xamarin/java.interop/blob/247145e4eed6d7c8dcef947a657f88da6663ad69/src/Java.Interop/Java.Interop/JniPeerMembers.JniInstanceMethods.cs#L162-L175
[2]: https://developer.android.com/reference/java/lang/Thread#Thread(java.lang.ThreadGroup,%20java.lang.Runnable,%20java.lang.String,%20long,%20boolean)
```

PR #7199, opened 2022-07-25
[build] Fix workload install for local dependencies

pjcollins:
The `InstallManifestAndDependencies` target is used by local builds to
install the runtime dependencies and WorkloadManifest files required for
building net*-android projects.  This broke after a recent .NET SDK
update, as the empty workload that we only use locally is not known to
the .NET SDK.  Fix this by adding our empty workload to the list of
included workloads when building locally.

### DIFF
--- a/build-tools/create-packs/Directory.Build.targets
+++ b/build-tools/create-packs/Directory.Build.targets
@@ -254,15 +254,19 @@
 
   <Target Name="InstallManifestAndDependencies">
     <PropertyGroup>
+      <_EmptyDepsWorkloadName>android.deps.workload</_EmptyDepsWorkloadName>
       <_LocalSdkManifestsFolder>$(BuildOutputDirectory)lib\sdk-manifests\$(DotNetSdkManifestsFolder)\</_LocalSdkManifestsFolder>
       <_LocalAndroidManifestFolder>$(_LocalSdkManifestsFolder)microsoft.net.sdk.android\</_LocalAndroidManifestFolder>
-      <_EmptyWorkloadDir>$(_LocalSdkManifestsFolder)android.deps.workload\</_EmptyWorkloadDir>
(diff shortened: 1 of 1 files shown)


build-tools/create-packs/Directory.Build.targets
```diff
@@ -254,15 +254,19 @@
...
-      <_EmptyWorkloadDir>$(_LocalSdkManifestsFolder)android.deps.workload\</_EmptyWorkloadDir>
+      <_EmptyWorkloadDir>$(_LocalSdkManifestsFolder)$(_EmptyDepsWorkloadName)\</_EmptyWorkloadDir>
       <_EmptyWorkloadJsonContent>
 <![CDATA[
 {"version": "0.0.1", "workloads": { "android-deps": { "extends" : [ "microsoft-net-runtime-android", "microsoft-net-runtime-android-aot" ] } } }
 ]]>
       </_EmptyWorkloadJsonContent>
     </PropertyGroup>
+    <ItemGroup>
+      <_WorkloadManifestsTxt Include="$(DotNetPreviewPath)sdk\*\IncludedWorkloadManifests.txt" />
```
jonathanpeppers, line 268: Looking at the code we might also be able to delete this file?

https://github.com/dotnet/sdk/blob/09b31215867d1ffe4955fd5b7cd91eb552d3632c/src/Resolvers/Microsoft.NET.Sdk.WorkloadManifestReader/SdkDirectoryWorkloadManifestProvider.cs#L44-L48

jonathanpeppers (approved):
But adding a line, is probably fin, too. 👍

pjcollins:
Replacing this with https://github.com/xamarin/xamarin-android/pull/7203

build-tools/create-packs/Directory.Build.targets
pjcollins: I tried testing this after removing the `IncludedWorkloadManifests.txt` file and `workload search` and `workload install` failed to recognize any workload IDs. I think a better fix has been added #7203.

PR #7207, opened 2022-07-26
[Xamarin.Android.Build.Tasks] fix @(LibraryProjectZip) and NuGets

jonathanpeppers:
Backport of #7151 

Context: https://github.com/xamarin/GooglePlayServicesComponents/issues/652

In .NET 6, `@(AndroidLibrary.Pack)` defaults to true (97190b6c), while
`%(LibraryProjectZip.Pack)` was never set, and thus was false.  This
would be an issue if you migrated a project using `@(LibraryProjectZip)` and were
not relying on the default wildcards used by:

	<AndroidLibrary Include="**/*.aar" />

The result is that `.aar` files weren't included into the `.nupkg`,
which would result in subsequent build & packaging errors when
using the `.nupkg`, as the `.aar` was missing.

Set `%(LibraryProjectZip.Pack)`=True by default in .NET 6+, to ease
project migration from Classic Xamarin.Android to .NET 6.

I updated a test for this scenario and fixed some assertions that are
working since 2ca2a103 was merged.


jonathanpeppers:
I opened this, so we won't forget to backport this. We can wait a bit to merge.

dellis1972 (approved):

jonathanpeppers:
1 failure we can ignore, we've seen these happening lately:

image

Issue #6961, opened 2022-04-25
.NET 6 app crashes on start in Debug mode


jonathanpeppers:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

.NET 6 / MAUI RC 1

### Description

From @RedChops here: https://github.com/dotnet/maui/issues/6460#issuecomment-1108713166

I can't get any iteration of HttpClient working (managed or native). Native crashes the app immediately on launch, Managed acts incredibly strangely. It appears to crash from a NullReferenceException, but sometimes it crashes silently, causing all code that would otherwise use the result of SendAsync to be skipped.

For example:

1. var result = await client.SendAsync(...);
2. var content = await result.Content.ReadAsStringAsync()
3. [something else]

If I set a breakpoint on line 2, that breakpoint will get skipped and code will continue to execute on line 3, causing both result and content to be null.

I spent a large part of the day uninstalling all VS versions, clearing out the dotnet folder of all SDKs, reinstalling everything, etc, so everything should be clean and at the latest version.

### Steps to Reproduce

I've attached here a small reproduction project. 

[NetworkTest.zip](https://github.com/dotnet/maui/files/8551024/NetworkTest.zip)

### Did you find any workaround?

No

### Relevant log output

The first is [AndroidClientHandler.adb.txt](https://github.com/dotnet/maui/files/8555811/AndroidClientHandler.adb.txt) which is the output from using the Android HttpClient. 

The second is [HttpClientHandler.adb.txt](https://github.com/dotnet/maui/files/8555832/HttpClientHandler.adb.txt). This of course includes the output from the AndroidClientHandler log since it's the same emulator boot.

Both of these logs were generated off that sample project which I created fresh specifically for this bug, so there shouldn't be anything else weird going on there. They also both were run in Debug mode with `<RunAOTCompilation>False</RunAOTCompilation>` set in the csproj.

RedChops:
I can also reproduce this in a pure Xamarin Android project, but it's a little more difficult. The app doesn't crash on the first request, but it will crash on the second. Adb log here: [adb.txt](https://github.com/xamarin/xamarin-android/files/8557300/adb.txt)

The specific workflow here is a request is made to an auth token generator (request succeeds), a link is sent to the email address which contains a deep link to the app. When the deep link is entered, the uri is processed and a second request is made to get user details. This second request fails.

RedChops:
This issue seems to be one expressly of timing.

[...]

PR #7214, opened 2022-07-28
[Mono.Android] Improve the GetHttpMessageHandler function

simonrozsival:
The constructor of `System.Net.Http.HttpClientHandler` calls `AndroidEnvironment.GetHttpMessageHandler()` via reflection to obtain the native HTTP handler (when `$(UseNativeHttpHandler) == true`). The object that is returned depends on the value of the `$(AndroidHttpClientHandlerType)` property.

Incorrect settings of `$(AndroidHttpClientHandlerType)` causes two problems in the .NET 6 builds:
- When the type cannot be resolved or the instance of the type cannot be created, the `GetHttpMessageHandler` function returns `null`. This isn't immediately caught by the `HttpClientHandler` and it leads to a confusing `NullReferenceException` later on.
- When the type is set to anything that derives from `HttpClientHandler` (including `AndroidClientHandler`), the app enters infinite indirect recursion because the `HttpClientHandler`'s constructor calls the `GetHttpMessageHandler` function.

I base this PR on the implementation of Xamarin.iOS where we never return `null` and when the value is invalid, we use the default handler type as a fallback. Whenever the type is invalid (null or derived from `HttpClientHandler`), the `AndroidMessageHandler` class is used instead + a warning is logged.

The docs of `$(AndroidHttpClientHandlerType)` are not applicable for the .NET build as they suggest setting the property to AndroidClientHandler or HttpClientHandler. I think the best options would be to either mark the property as deprecated (and update code to completely ignore it in .NET) or to require the type to extend AndroidMessageHandler. Suggestions are welcome.

Fixes https://github.com/xamarin/xamarin-android/issues/6961


src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
```diff
@@ -344,18 +344,60 @@ static IWebProxy GetDefaultProxy ()
...
+				Logger.Log (LogLevel.Warn, "MonoAndroid", $"The type {handlerType.AssemblyQualifiedName} set as the default HTTP handler is invalid. Use a type that extends System.Net.Http.HttpMessageHandler.");
+				return false;
+			}
+
+			return true;
+		}
+
+		static bool Extends (Type handlerType, string typeName, string assemblyName = "System.Net.Http")
+		{
+			while (handlerType != null) {
```
jonpryor, line 383: This method body seems odd; why not:

```csharp
static bool Extends (Type handlerType, string assemblyQualifiedTypeName)
{
    var baseType = Type.GetType (assemblyQualifiedTypeName, throwOnError:false);
    if (baseType == null)
        return false;
    return baseType.IsAssignableFrom (handlerType);
}
```

…but probably move the `Type.GetType()` into the caller scope…?

My *guess* as to using the `while` loop is to avoid the `Type.GetType()` invocation; is it *that* expensive?  It's not like the `while` loop *prevents* the equivalent of `typeof(System.Net.Http.HttpMessageHandler)` from running, as `handlerType.BaseType` will *require* that it be loaded if it's in the inheritance chain.

If e.g. `System.Net.Http.HttpMessageHandler ` is linked away, then `handlerType` cannot possibly extend it.

src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
```diff
@@ -344,18 +344,60 @@ static IWebProxy GetDefaultProxy ()
...
+			if (!String.IsNullOrEmpty (handlerTypeName))
+				handlerType = Type.GetType (handlerTypeName, throwOnError: false);
+
 			// We don't do any type checking or casting here to avoid dependency on System.Net.Http in Mono.Android.dll
-			return Activator.CreateInstance (handlerType);
+			if (handlerType is null || !IsAcceptableHttpMessageHandlerType (handlerType)) {
+				handlerType = GetFallbackHttpMessageHandlerType ();
+			}
+
+			return Activator.CreateInstance (handlerType)
```
jonpryor, line 359: We should cache `handlerType` in a field, so that on repeated calls to `GetHttpMessageHandler()`, we don't need to redo all this lookup & validation.

jonpryor:
> The docs of `$(AndroidHttpClientHandlerType)` are not applicable for the .NET build as they suggest setting the property to `AndroidClientHandler` or `HttpClientHandler`. I think the best options would be to either mark the property as deprecated (and update code to completely ignore it in .NET) or to require the type to extend `AndroidMessageHandler`. 

I think we should update the docs to note differences between Classic and .NET.

As far as requiring that the type extend `AndroidMessageHandler`, couldn't it just require that the type extend `HttpMessageHandler`?  Why would extending `AndroidMessageHandler` be required?

We should likely *also* check this type during the build process, so that it's a build-time failure instead of a runtime exception.  That can be done separately, by someone else. ;-)

src/Mono.Android/Android.Runtime/AndroidEnvironment.cs
simonrozsival: You're right, that's much cleaner. Thanks for the suggestion.

jonpryor:
Related? https://github.com/xamarin/xamarin-android/issues/6949

simonrozsival:
@jonpryor I'll put some more thought into the docs and update it tomorrow.

The type check doesn't require extending AndroidMessageHandler but HttpMessageHandler. AndroidMessageHandler is used only as the fallback type for incorrect configurations.

#6949 is most likely related

simonrozsival:
I believe the remaining failing tests are unrelated to this PR.

jonpryor:
Commit message for review.

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/6961
Fixes: https://github.com/xamarin/xamarin-android/issues/6949

If a .NET SDK for Android app is built with the
`$(AndroidHttpClientHandlerType)` property set to a type which does
not exist, such as the value `Xamarin.Android.Net.AndroidClientHandler`
used by *Classic* Xamarin.Android apps (!), then the app may crash
during startup:

	I MonoDroid: UNHANDLED EXCEPTION:
	I MonoDroid: Android.Runtime.JavaProxyThrowable: Exception of type 'Android.Runtime.JavaProxyThrowable' was thrown.
	I MonoDroid: 
	I MonoDroid:   --- End of managed Android.Runtime.JavaProxyThrowable stack trace ---
	I MonoDroid: android.runtime.JavaProxyThrowable: System.NullReferenceException: Object reference not set to an instance of an object.
	I MonoDroid:    at System.Net.Http.HttpClientHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
	I MonoDroid:    at System.Net.Http.HttpMessageInvoker.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
	I MonoDroid:    at System.Net.Http.HttpClient.<>n__0(HttpRequestMessage request, CancellationToken cancellationToken)
	I MonoDroid:    at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
	I MonoDroid:    at NetworkTest.MainPage.MakeRequest()
	I MonoDroid:    at NetworkTest.MainPage.OnCounterClicked(Object sender, EventArgs e)
	I MonoDroid:    at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
	I MonoDroid:    at Android.App.SyncContext.<>c__DisplayClass2_0.<Post>b__0()
	I MonoDroid:    at Java.Lang.Thread.RunnableImplementor.Run()
	I MonoDroid:    at Java.Lang.IRunnableInvoker.n_Run(IntPtr jnienv, IntPtr native__this)
	I MonoDroid:    at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V(_JniMarshal_PP_V callback, IntPtr jnienv, IntPtr klazz)
	I MonoDroid:     at mono.java.lang.RunnableImplementor.n_run(Native Method)
	I MonoDroid:     at mono.java.lang.RunnableImplementor.run(RunnableImplementor.java:30)
	I MonoDroid:     at android.os.Handler.handleCallback(Handler.java:938)
	I MonoDroid:     at android.os.Handler.dispatchMessage(Handler.java:99)
	I MonoDroid:     at android.os.Looper.loopOnce(Looper.java:201)
	I MonoDroid:     at android.os.Looper.loop(Looper.java:288)
	I MonoDroid:     at android.app.ActivityThread.main(ActivityThread.java:7870)
	I MonoDroid:     at java.lang.reflect.Method.invoke(Native Method)
	I MonoDroid:     at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	I MonoDroid:     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)
	I MonoDroid: 
	I MonoDroid:   --- End of managed Android.Runtime.JavaProxyThrowable stack trace ---
	I MonoDroid: android.runtime.JavaProxyThrowable: System.NullReferenceException: Object reference not set to an instance of an object.
	I MonoDroid:    at System.Net.Http.HttpClientHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
	I MonoDroid:    at System.Net.Http.HttpMessageInvoker.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
	I MonoDroid:    at System.Net.Http.HttpClient.<>n__0(HttpRequestMessage request, CancellationToken cancellationToken)
	I MonoDroid:    at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)
	I MonoDroid:    at NetworkTest.MainPage.MakeRequest()
	I MonoDroid:    at NetworkTest.MainPage.OnCounterClicked(Object sender, EventArgs e)
	I MonoDroid:    at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_0(Object state)
	I MonoDroid:    at Android.App.SyncContext.<>c__DisplayClass2_0.<Post>b__0()
	I MonoDroid:    at Java.Lang.Thread.RunnableImplementor.Run()
	I MonoDroid:    at Java.Lang.IRunnableInvoker.n_Run(IntPtr jnienv, IntPtr native__this)
	I MonoDroid:    at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PP_V(_JniMarshal_PP_V callback, IntPtr jnienv, IntPtr klazz)
	I MonoDroid:     at mono.java.lang.RunnableImplementor.n_run(Native Method)
	I MonoDroid:     at mono.java.lang.RunnableImplementor.run(RunnableImplementor.java:30)
	I MonoDroid:     at android.os.Handler.handleCallback(Handler.java:938)
	I MonoDroid:     at android.os.Handler.dispatchMessage(Handler.java:99)
	I MonoDroid:     at android.os.Looper.loopOnce(Looper.java:201)
	I MonoDroid:     at android.os.Looper.loop(Looper.java:288)
	I MonoDroid:     at android.app.ActivityThread.main(ActivityThread.java:7870)
	I MonoDroid:     at java.lang.reflect.Method.invoke(Native Method)
	I MonoDroid:     at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:548)
	I MonoDroid:     at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1003)

The cause of the crash is that the
[`System.Net.Http.HttpClientHandler` default constructor][0] calls
[`CreateNativeHandler()`][1], which uses Reflection to call
`AndroidEnvironment.GetHttpMessageHandler()`, in order to obtain the
native HTTP handler (when `$(UseNativeHttpHandler)`=true).

The object that is returned depends on the value of the
`$(AndroidHttpClientHandlerType)` property.

When the type mentioned by `$(AndroidHttpClientHandlerType)` can't be
found, then `GetHttpMessageHandler()` will return `null`.  This isn't
immediately caught by `HttpClientHandler`, leading to a confusing
`NullReferenceException` later on.

When the type mentioned by `$(AndroidHttpClientHandlerType)` is a
type which inherits `HttpClientHandler` -- which includes the Classic
default value of `AndroidClientHandler`! -- then the app enters
infinite indirect recursion because the `HttpClientHandler`'s
constructor calls the `GetHttpMessageHandler` function which calls
the `HttpClientHandler` constructor which calls…

I base this PR on the implementation of Xamarin.iOS where we never
return `null`, and when the value is invalid we use the default handler
type as a fallback.  Whenever the type is invalid (`null` or derived
from `HttpClientHandler`), the `AndroidMessageHandler` class is used
instead and a warning is logged.

The existing docs for `$(AndroidHttpClientHandlerType)` are not
applicable for .NET SDK for Android, as they suggest setting the
property to `AndroidClientHandler` or `HttpClientHandler`.

Update the documentation for `$(AndroidHttpClientHandlerType)` to
explicitly call out .NET 6+ behavior, and that
`Xamarin.Android.Net.AndroidMessageHandler` (1e5bfa3f) or
`System.Net.Http.SocketsHttpHandler, System.Net.Http` should be  used
instead.

TODO: [Emit a build-time warning][2] if
`$(AndroidHttpClientHandlerType)` is set to an invalid value

[0]: https://github.com/dotnet/runtime/blob/7a45201181d37329b7ab0bee541ed7c42b5d94b6/src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.AnyMobile.cs#L41
[1]: https://github.com/dotnet/runtime/blob/7a45201181d37329b7ab0bee541ed7c42b5d94b6/src/libraries/System.Net.Http/src/System/Net/Http/HttpClientHandler.AnyMobile.InvokeNativeHandler.cs#L149-L158
[2]: https://github.com/xamarin/xamarin-android/issues/7326
```

PR #7225, opened 2022-08-01
[ci] Use nuget-msi-convert/job/v3.yml

pjcollins:

Updates the build to the latest MSI generation template.  The `v3`
refactoring and support for workload pack group MSIs.

The generation of workload pack group MSIs is enabled by default.  The
tooling will now generate a single MSI for each workload declared in our
WorkloadManifest.json file.  This should improve MSI based installation
scenarios as it dramatically reduces the number of MSIs that we ship.


pjcollins:
Testing with https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6497953&view=results

PR #7227, opened 2022-08-02
[ci] Add Android Designer test template

pjcollins:
The designer team is no longer using the build and test templates in
[yaml-templates][0], so there is not much value in us maintaining them
externally.

A copy of the test template has been moved into our repo, and the build
and test steps have been updated to reflect the latest changes in the
UITools repo.

[0]: https://github.com/xamarin/yaml-templates/tree/89becd2759512ebd8fe560a53f6916cc684f0b4c/designer


jonathanpeppers (approved):
This test run looks OK:

image

PR #7236, opened 2022-08-05
[build] update to latest JDKs

jonathanpeppers:
Context: https://docs.microsoft.com/java/openjdk/download#openjdk-11016-lts--see-previous-releases
Context: https://adoptium.net/temurin/releases/?version=8

The IDE teams are shipping these updates in Visual Studio shortly.

Update to these builds, so we can test them out.


jonathanpeppers:
I can't believe it, completely green?

grendello (approved):

PR #7241, opened 2022-08-08
[ci] Upload test assemblies after signing

pjcollins:
We've recently seen a handful of build timeouts and/or failures during
signing.  The ESRP signing service seems to be much less performant and
more error prone on macOS.  In these failure cases, rebuild attempts
would also fail when they got to the "upload test assembly" step:

    [error]Artifact test-assemblies already exists for build 6522654.

Improve this by uploading test assemblies and build inventory data at
the end of the build, and by increasing the overall timeout of the job.


jonathanpeppers (approved):
LGTM if the build is green. 👍

Issue #7149, opened 2022-07-05
Hostname not verified when sending request


IgorKravchenko10:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

VS 17.3.0 Preview 2.0, .net 6.0.400

### Description

We started developing MAUI application and stuck with error trying to send http request. We already have a Xamarin Forms application which successfully works with same API.
We get error `Hostname not verified when sending request`.
This issue described [here](https://docs.microsoft.com/en-us/answers/questions/885827/error-hostname-not-verified-when-sending-request-o.html)
Also reported in [maui](https://github.com/dotnet/maui/issues/8518)
MAUI reproduction (not working) [github](https://github.com/IgorKravchenko10/Maui-web-service-bug) [onedrive](https://1drv.ms/u/s!Aun0lW86yXKJijM3oPvspRfY-8nR?e=xKHhcY) [zip](https://github.com/xamarin/xamarin-android/files/9045283/ServiceTest.zip)
Xamarin Forms (working) [github](https://github.com/IgorKravchenko10/XFWebService) [onedrive](https://1drv.ms/u/s!Aun0lW86yXKJijT30Nexve-y0Q7k?e=OC79QY) [zip](https://github.com/xamarin/xamarin-android/files/9045272/XFServiceTest.zip)

### Steps to Reproduce

1. Create a new MAUI application.
2. Make http request to API with self signed certificate.
3. Get error "Hostname not verified when sending request".

### Did you find any workaround?

No

simonrozsival:
@IgorKravchenko10 There's a mismatch between your server's hostname (`tcu.base2base.com.ua`) and your certificate's CN (`DESKTOP-LK5ITUQ`). You can fix this either by getting a certificate specific for your server (you can get free ones from https://letsencrypt.org/) or by setting up your custom hostname verifier for your AndroidMessageHandler (see https://github.com/dotnet/maui/discussions/8131, https://gist.github.com/Eilon/49e3c5216abfa3eba81e453d45cba2d4). Let me know if this helps you or if you have more questions.

simonrozsival:
I think the difference between the Xamarin Forms and MAUI you're seeing is because the `HttpClientHandler` has different behavior on those two platforms by default (managed impl. in Xamarin Forms and native android handler in MAUI). 

[...]

PR #7246, opened 2022-08-10
[Mono.Android] Improve server certificate custom validation callback integration to cover hostname mismatch

simonrozsival:
The recent improvement of AndroidMessageHandler that adds support for the `ServerCertificateCustomValidationCallback` allows customers to trust self-signed certificates. It has one flaw though: it doesn't account for mismatches between the server's hostname and the certificate's hostname. This causes problems for example when our customers are using the `10.0.2.2` IP address to send requests to a local server with self-signed dev certificate in ASP.NET.

This PR addresses the problem and aligns the behavior of AndroidMessageHandler with other platforms, where hostname verification is part of the server certificate custom validation callback.

There are two significant changes in this PR:
- the trust manager that wraps the custom callback now verifies the certificate hostname
- Java's separate hostname verification is skipped

There is only one aspect of this PR that is a little hacky. The `IHostnameVerifier` interface requires an instance of `ISSLSession` which we don't have access to in the trust manager. The default hostname verifier uses just a single getter to access the chain of certificates so it seems OK to me to have a dummy class that implements the interface that holds just the peer certificates. I tried to find access to the session from `HttpsURLConnection` but there doesn't seem to be a way to access the `SSLEngine` and its `HandshakeSesion` property. I'm interested in your feedback.

Improves the original implementation added in #6665 
Closes https://github.com/xamarin/xamarin-android/issues/7149
Other related issues: https://github.com/dotnet/runtime/issues/70434, https://github.com/xamarin/xamarin-android/pull/6665, https://github.com/dotnet/docs-maui/issues/600
Related discussion: https://github.com/dotnet/maui/discussions/8131


grendello (approved):

grendello:
I think the dummy implementation of `ISSLSession` is fine

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
```diff
@@ -157,13 +157,13 @@ public CookieContainer CookieContainer
...
-		X509TrustManagerWithValidationCallback.Helper? _callbackTrustManagerHelper = null;
+		ServerCertificateCustomValidator? _serverCertificateCustomValidator = null;
 
 		public Func<HttpRequestMessage, X509Certificate2?, X509Chain?, SslPolicyErrors, bool>? ServerCertificateCustomValidationCallback
 		{
-			get => _callbackTrustManagerHelper?.Callback;
+			get => _serverCertificateCustomValidator?.Callback;
 			set {
-				_callbackTrustManagerHelper = value != null ? new X509TrustManagerWithValidationCallback.Helper (value) : null;
+				_serverCertificateCustomValidator = value != null ? new ServerCertificateCustomValidator (value) : null;
```
jonpryor, line 166: I'm uncertain about the *structure* of this change.  Every time the `AndroidMessageHandler.ServerCertificateCustomValidationCallback` property is set, we create a new `ServerCertificateCustomValidator` instance, which in turn creates a new `AlwaysAcceptingHostnameVerifier` instance, which is a "bridged" instance (GC overheads).  

Is this necessary?

It feels like we'd be more "GC friendly" if we instead did:

```csharp
if (value == null) {
    _serverCertificateCustomValidator = null;
} else {
    _serverCertificateCustomValidator ??= new ServerCertificateCustomValidator (value);
    _serverCertificateCustomValidator.Callback = value;
}
```

src/Mono.Android/Xamarin.Android.Net/ServerCertificateCustomValidator.cs
```diff
@@ -0,0 +1,185 @@
...
+using Javax.Net.Ssl;
+
+using JavaCertificateException = Java.Security.Cert.CertificateException;
+using JavaX509Certificate = Java.Security.Cert.X509Certificate;
+
+namespace Xamarin.Android.Net
+{
+	internal sealed class ServerCertificateCustomValidator
+	{
+		public IHostnameVerifier HostnameVerifier { get; } = new AlwaysAcceptingHostnameVerifier();
```
jonpryor, line 16: Similarly, this should become an `AlwaysAcceptingHostnameVerifier.Instance` static property; we only ever need *one*, right?

jonpryor: A downside with making a static `AlwaysAcceptingHostnameVerifier.Instance` property is that it would never be disposed, meaning if you had an app that set `AndroidMessageHandler.ServerCertificateCustomValidationCallback` to *something*, then later set it to *null*, the `AlwaysAcceptingHostnameVerifier` would still be kept alive.

Is this a scenario to worry about, though?  If you're setting `AndroidMessageHandler.ServerCertificateCustomValidationCallback`, why would you clear it?  I don't understand the use case.

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
jonpryor: A possible concern here is thread-safety.  I can't find any documented thread-safety requirements on e.g. [`SocketsHttpHandler`](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler?view=net-6.0), though [`HttpMessageHandler`](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpmessagehandler?view=net-6.0) *does* imply that thread-safety is required.

If we thus assume that we need to be thread-safe…. Assigning to `_serverCertificateCustomValidator.Callback` is thread-safe.  What I'm less certain about is the *interaction* between `ServerCertificateCustomValidationCallback` and `SetupSSL()`, as `SetupSSL()` calls `_serverCertificateCustomValidator.ReplaceX509TrustManager()`, so *which* Callback is used could get weird/confusing.

We thus may need to *not* do `_serverCertificateCustomValidator.Callback = value`.

jonpryor: Related meta-question: what are the semantics in a multi-threaded environment when intermixing `.Send()` and assignments to `.ServerCertificateCustomValidationCallback`?  Is this a Don't Do That™ scenario?

tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
```diff
@@ -61,9 +61,11 @@ public async Task ServerCertificateCustomValidationCallback_RejectRequest ()
 
 			try {
 				await client.GetStringAsync ("https://microsoft.com/");
-			} catch {
+			} catch (System.Net.WebException) {
 				// System.Net.WebException is thrown in Debug mode
-				// Java.Security.Cert.CertificateException is thrown in Release mode
+				exceptionWasThrown = true;
+			} catch (Java.IO.IOException) {
+				// Java.IO.IOException is thrown in Release mode
```
jonpryor, line 68: Wait, what?  Why is `Java.IO.IOException` thrown in Release mode?  (Callstack?)  I would hope that Debug & Release configs have consistent behavior when possible, and this comment calls that into question.

tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
```diff
@@ -92,8 +94,57 @@ public async Task ServerCertificateCustomValidationCallback_ApprovesRequestWithI
...
+
+			var handler = new AndroidMessageHandler ();
+			var client = new HttpClient (handler);
+
+			try {
+				await client.GetStringAsync ("https://wrong.host.badssl.com/");
+			} catch (System.Net.WebException) {
+				// System.Net.WebException is thrown in Debug mode
+				exceptionWasThrown = true;
+			} catch (Java.IO.IOException) {
```
jonpryor, line 114: Ditto.

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
simonrozsival: Thanks for the suggestion. I'll update the PR to reduce allocations and make it thread-safe.

Customers can use it anyway they like, but the typical use case should be to set the callback once and then keep it the same for the lifetime of the handler. I don't think it makes too much sense to change the callbacks for different requests.

In general, there are many different properties with public setters that affect the request and that can change independently of calls to `SendAsync` (for example `PreAuthenticate`, `UseProxy`, `Proxy`, `Credentials`, `TrustedCerts`). I think the thread-safety is meant only to parallel calls to `SendAsync` (e.g., redirect state for each parallel request should be independent). IMO this is a Don't Do That™ scenario.

tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
simonrozsival: The Java exception is mapped into WebException when `JNIEnv.ShouldWrapJavaException (ex)` is true. Now that I looked into it again, I realize that I probably mistook Debug vs. Release with legacy vs. .net and the comments are incorrect. I'll update the comments to make sense.

src/Mono.Android/Xamarin.Android.Net/ServerCertificateCustomValidator.cs
simonrozsival: I don't think we need to worry about the scenario where you set the callback to null. On the other hand, it's possible that the whole message handler is collected and the hostname verifier will be kept alive. We could add some instance counting and remove the hostname when the last validator instance is finalized but that seems like an overkill.

The official recomendation is to instantiate HttpClient (and transitively the HTTP message handlers) once and then use it for the lifetime of the application (https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-6.0#instancing). The documentation also suggests creating multiple HttpClients when the app needs to make requests with different setting and I don't think we need to optimize for the case when the `ServerCertificateCustomValidationCallback` property is set multiple times (or set to `null`).

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
grendello: I'm not sure we want to cache the instance of the custom validator. I can imagine a scenario where the validation callback is changed between different URLs (e.g. accessing a local server with a self-signed certificate and the proxying results of the query to a server on the Internet with a valid cert etc) while keeping the same instance of the handler around. With that, I think the most common scenario is where the client handler is a "one short" affair, where the app either makes requests to the same server repeatedly (keeping the client handler around) or where it makes a single request to somewhere. In both scenarios caching won't improve neither memory use nor performance. Caching means we assume `value` passed to the property doesn't change - it feels like an unnecessary restriction.

src/Mono.Android/Xamarin.Android.Net/ServerCertificateCustomValidator.cs
grendello: > Similarly, this should become an `AlwaysAcceptingHostnameVerifier.Instance` static property; we only ever need _one_, right?

Yeah, but it would also add initialization pressure to apps which don't use the validator? Even if they don't access the property, the static constructor would run anyway?

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
```diff
@@ -157,13 +157,19 @@ public CookieContainer CookieContainer
...
+			get => _serverCertificateCustomValidator?.Callback;
 			set {
-				_callbackTrustManagerHelper = value != null ? new X509TrustManagerWithValidationCallback.Helper (value) : null;
+				if (value is null) {
+					_serverCertificateCustomValidator = null;
+				} else if (_serverCertificateCustomValidator is null) {
+					_serverCertificateCustomValidator = new ServerCertificateCustomValidator (value);
+				} else {
+					_serverCertificateCustomValidator.Callback = value;
+				}
```
grendello, line 172: Sorry to be late to the discussion, but I don't think we should do it. Please see my [comment](https://github.com/xamarin/xamarin-android/pull/7246#discussion_r946579603).

grendello (changes requested):

src/Mono.Android/Xamarin.Android.Net/ServerCertificateCustomValidator.cs
simonrozsival: If the app doesn't use the validator, then all of this code is linked away, so I don't think it's a problem.

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
simonrozsival: @grendello I'm not sure if I understand your comment correctly. The custom validator class doesn't have any internal state except for the callback. Replacing the callback on an existing instance and creating a new instance with the new callbacks are functionally identical. The reason why we create an instance of the class, and we don't use just static methods to set up trust managers and the hostname verifier, is to make the code linker friendly.

grendello: @simonrozsival it's possible I'm overthinking this. I'm thinking about a scenario when a single instance of the handler is reused for different servers and uses different callbacks and whether sharing a single instance of the verifier (with a different callback) would/could potentially break something. With callbeck being the only state of the verifier, I guess my worries are unfounded, just can't get rid of that gnawing feeling that something's off :)

grendello (approved):

src/Mono.Android/Xamarin.Android.Net/ServerCertificateCustomValidator.cs
```diff
@@ -0,0 +1,190 @@
...
+				} else {
+					sslPolicyErrors |= SslPolicyErrors.RemoteCertificateNotAvailable;
+				}
+
+				if (!VerifyHostname (javaChain)) {
+					sslPolicyErrors |= SslPolicyErrors.RemoteCertificateNameMismatch;
+				}
+
+				if (!_serverCertificateCustomValidationCallback (_request, certificate, CreateChain (certificates), sslPolicyErrors)) {
+					throw new JavaCertificateException ("The remote certificate was rejected by the provided RemoteCertificateValidationCallback.");
```
jonpryor, line 72: @simonrozsival: Do you know the "context"/calling method stack of how this method is invoked?  Specifically, can this exception type "leak" from/through e.g. `HttpClient.Send()`?

We can't have Java exception types "leak through" .NET methods which document the exception types that can be thrown; see e.g. [the `$(AndroidBoundExceptionType)` docs](https://docs.microsoft.com/en-us/xamarin/android/deploy-test/building-apps/build-properties#androidboundexceptiontype) and 60363ef4.

We probably *don't* want to throw `Java.Security.Cert.CertificateException` here, *unless* Java code is calling this `CheckServerTrusted()` and needs an appropriate Java exception for handling purposes.

tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
```diff
@@ -85,15 +76,51 @@ public async Task ServerCertificateCustomValidationCallback_ApprovesRequestWithI
...
 			Assert.IsTrue (callbackHasBeenCalled, "custom validation callback hasn't been called");
-			Assert.IsNull (exception, $"an exception was thrown: {exception}");
+			Assert.AreEqual (SslPolicyErrors.RemoteCertificateNameMismatch, reportedErrors & SslPolicyErrors.RemoteCertificateNameMismatch);
+		}
+
+		private async Task AssertRejectsRemoteCertificate (Func<Task> makeRequest)
+		{
+			try {
+				await makeRequest();
+			} catch (Java.IO.IOException) {
```
jonpryor, line 118: This is concerning, for the same reasons [as above](https://github.com/xamarin/xamarin-android/pull/7246/files#r950490896): [`HttpClient.GetStringAsync(string)`](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.getstringasync?view=net-6.0#system-net-http-httpclient-getstringasync(system-string)) does not document `Java.IO.IOException` as an acceptable exception type, just: `InvalidOperationException`, `HttpRequestException`, and `TaskCanceledException`.

Consequently, `Java.IO.IOException` *should not be thrown*, ***ever***, from an `HttpClient.GetStringAsync()` invocation.

If a `Java.IO.IOException` can be thrown, *we must fix that*.

tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
```diff
@@ -85,15 +76,51 @@ public async Task ServerCertificateCustomValidationCallback_ApprovesRequestWithI
...
+				// the Java.IO.IOException is thrown when the TLS connection cannot be established
```
jonpryor, line 119: If a TLS connection can't be established, then presumably an `HttpRequestException` should be thrown, with an *inner exception* of `Java.IO.IOException`.

src/Mono.Android/Xamarin.Android.Net/ServerCertificateCustomValidator.cs
simonrozsival: This method is invoked from Java (`SSLContext`) and throwing this specific exception is the way to reject the server's certificate (see the method signature https://developer.android.com/reference/javax/net/ssl/X509TrustManager#checkServerTrusted(java.security.cert.X509Certificate[],%20java.lang.String)). We never call this method from C#.

tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
simonrozsival: I agree that we should actually throw `HttpRequestException` but that will involve modifying a [different part of AndroidMessageHandler](https://github.com/xamarin/xamarin-android/blob/main/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs#L418:L426) and consider how this will affect customers who rely on this behavior in their existing codebases. I believe that's outside of the scope of this PR.

simonrozsival:
@jonpryor I think all the concerts regarding exceptions were addressed. Do you think this PR could now be merged or is there still something you'd like to revisit?

jonpryor:
@simonrozsival: please review this commit message for correctness.

```
[Mono.Android] ServerCertificateCustomValidationCallback & hostname (#7246)

Fixes: https://github.com/xamarin/xamarin-android/issues/7149

Context: 48540d6933e2bbded3db6106d594a3949e9ac020
Context: https://docs.microsoft.com/en-us/answers/questions/885827/error-hostname-not-verified-when-sending-request-o.html
Context: https://github.com/dotnet/docs-maui/issues/600
Context: https://github.com/dotnet/maui/discussions/8131
Context: https://github.com/dotnet/runtime/issues/70434

Commit 48540d69 added support to
`AndroidMessageHandler.ServerCertificateCustomValidationCallback` for
self-signed certificates.  However, it didn't account for mismatches
between the server's hostname and the certificate's hostname.
This can cause issues when e.g. you use 10.0.2.2 to send requests to
a local server with a self-signed dev certificate in ASP.NET:

	var handler     = new AndroidMessageHandler {
	    // Allow any cert; DO NOT USE IN PRODUCTION
	    ServerCertificateCustomValidationCallback   =
	        (httpRequestMessage, cert, cetChain, policyErrors) => true,
	    // …
	};
	var client      = new HttpClient(handler);
	var response    = client.SendAsync(uri);

This could result an exception similar to:

	System.Net.WebException: Hostname EXAMPLE.DOMAIN not verified:
	    certificate: sha1/EXAMPLE_CERT_SHA1
	    DN: CN=EXAMPLE_CN
	    subjectAltNames: [EXAMPLE_CN]
	 ---> Javax.Net.Ssl.SSLPeerUnverifiedException: Hostname EXAMPLE.DOMAIN not verified:
	    certificate: sha1/EXAMPLE_CERT_SHA1
	    DN: CN=EXAMPLE_CN
	    subjectAltNames: [EXAMPLE_CN]
	   at Java.Interop.JniEnvironment.InstanceMethods.CallVoidMethod(JniObjectReference instance, JniMethodInfo method, JniArgumentValue* args)
	   at Java.Interop.JniPeerMembers.JniInstanceMethods.InvokeAbstractVoidMethod(String encodedMember, IJavaPeerable self, JniArgumentValue* parameters)
	   at Javax.Net.Ssl.HttpsURLConnectionInvoker.Connect()
	   at Xamarin.Android.Net.AndroidMessageHandler.<>c__DisplayClass125_0.<ConnectAsync>b__0()
	   at System.Threading.Tasks.Task.InnerInvoke()
	   at System.Threading.Tasks.Task.<>c.<.cctor>b__272_0(Object obj)
	   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
	--- End of managed Javax.Net.Ssl.SSLPeerUnverifiedException stack trace ---
	javax.net.ssl.SSLPeerUnverifiedException: Hostname EXAMPLE.DOMAIN not verified:
	    certificate: sha1/EXAMPLE_CERT_SHA1
	    DN: CN=EXAMPLE_CN
	    subjectAltNames: [EXAMPLE_CN]
	     at com.android.okhttp.internal.io.RealConnection.connectTls(RealConnection.java:205)
	     at com.android.okhttp.internal.io.RealConnection.connectSocket(RealConnection.java:153)
	     at com.android.okhttp.internal.io.RealConnection.connect(RealConnection.java:116)
	     at com.android.okhttp.internal.http.StreamAllocation.findConnection(StreamAllocation.java:186)
	     at com.android.okhttp.internal.http.StreamAllocation.findHealthyConnection(StreamAllocation.java:128)
	     at com.android.okhttp.internal.http.StreamAllocation.newStream(StreamAllocation.java:97)
	     at com.android.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:289)
	     at com.android.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:232)
	     at com.android.okhttp.internal.huc.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:465)
	     at com.android.okhttp.internal.huc.HttpURLConnectionImpl.connect(HttpURLConnectionImpl.java:131)
	     at com.android.okhttp.internal.huc.DelegatingHttpsURLConnection.connect(DelegatingHttpsURLConnection.java:90)
	     at com.android.okhttp.internal.huc.HttpsURLConnectionImpl.connect(HttpsURLConnectionImpl.java:30)
	--- End of inner exception stack trace ---
	   at Xamarin.Android.Net.AndroidMessageHandler.SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
	   at System.Net.Http.HttpClient.<SendAsync>g__Core|83_0(HttpRequestMessage request, HttpCompletionOption completionOption, CancellationTokenSource cts, Boolean disposeCts, CancellationTokenSource pendingRequestsCts, CancellationToken originalCancellationToken)

Address this problem and aligns the behavior of
`AndroidMessageHandler` with other platforms, where hostname
verification is part of the server certificate custom validation
callback.

There are two significant changes:

  - The trust manager that wraps the custom callback now verifies the
    certificate hostname

  - Java's separate hostname verification is skipped

There is only one aspect which is a little hacky: the
[`javax.net.ssl.HostnameVerifier` interface][0] requires an instance
of [`javax.net.ssl.SSLSession`][1], which we don't have access to in
the trust manager.  The default hostname verifier instead uses just a
single getter to access the chain of certificates so it seems OK to me
to have a dummy class that implements the interface that holds just
the peer certificates.

(I tried to find access to the session from `HttpsURLConnection` but
there doesn't seem to be a way to access the `SSLEngine` and its
`HandshakeSesion` property.)

[0]: https://developer.android.com/reference/javax/net/ssl/HostnameVerifier
[1]: https://developer.android.com/reference/javax/net/ssl/SSLSession
```

PR #7261, opened 2022-08-15
[ci] Add support to net6.0 for multi-targeting in VS

pjcollins:
Context: https://github.com/xamarin/yaml-templates/pull/199

Updates the build to use the latest MSI generation template. The v3
refactoring, support for multi-targeting, and support for workload pack
group MSIs.

The build will now produce two different VS Drop artifacts.  The MSI and
VSMAN files generated for SDK packs have been split out into a new
`vsdrop-multitarget-signed` artifact, allowing us to include multiple
versions of the SDK packs in VS.

The `Microsoft.Android.Sdk` pack has been renamed to
`Microsoft.Android.Sdk.net6` to match the pack alias that is referenced
in the .NET 7 manifest.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
@@ -4,7 +4,7 @@
     "android": {
       "description": ".NET SDK Workload for building Android applications.",
       "packs": [
-        "Microsoft.Android.Sdk",
+        "Microsoft.Android.Sdk.net6",
         "Microsoft.Android.Ref.31",
         "Microsoft.Android.Runtime.31.android-arm",
         "Microsoft.Android.Runtime.31.android-arm64",
@@ -41,7 +41,7 @@
     }
   },
   "packs": {
-    "Microsoft.Android.Sdk": {
+    "Microsoft.Android.Sdk.net6": {
       "kind": "sdk",
       "version": "@WORKLOAD_VERSION@",
       "alias-to": {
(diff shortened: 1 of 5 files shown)


src/Xamarin.Android.Build.Tasks/Microsoft.NET.Sdk.Android/WorkloadManifest.in.json
```diff
@@ -41,7 +41,7 @@
     }
   },
   "packs": {
-    "Microsoft.Android.Sdk": {
+    "Microsoft.Android.Sdk.net6": {
```
jonathanpeppers, line 44: I guess renaming this breaks Xamarin.Legacy.Sdk because of this line:

https://github.com/xamarin/Xamarin.Legacy.Sdk/blob/ce48ad6d67a77c7a05bcf31155a033a3539a0ff4/src/Xamarin.Legacy.Sdk/Sdk/Xamarin.Legacy.Android.targets#L28

Is there a way to also keep the old name and alias to `Microsoft.Android.Sdk.net6`?

jonathanpeppers: Ok, I pushed: https://www.nuget.org/packages/Xamarin.Legacy.Sdk/0.2.0-alpha2

I don't know how long it takes to show up on `dotnet-public`:

https://dev.azure.com/dnceng/public/_artifacts/feed/dotnet-public/NuGet/Xamarin.Legacy.Sdk

When it does, we can  update this line:

https://github.com/xamarin/xamarin-android/blob/a689fb9b0e1459c95115b4c9949cb7ec4068f2be/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs#L931

pjcollins: Updated the tests to use the new version, that PR should fix those failures.

jonathanpeppers (approved):
This looks good, I reran a couple test phases, but we can merge if those look OK after.

PR #7262, opened 2022-08-16
Bump LZ4 to v1.9.4

grendello:
Changes: https://github.com/lz4/lz4/releases/tag/v1.9.4
Context: https://github.com/lz4/lz4/commit/22e232dadaa36341afce0d3f114a3846273f0d2f

The most important change from our point of view is:

  * faster decoding speed (~+20%) on aarch64 platforms

The release notes URL above mentions only Apple silicon platforms as
gaining from the improvement, but modern AArch64 chipsets may benefit
as well.


grendello:
The Windows test build failures can be ignored, I think. The code is used at run time on a device and all the emulator tests are green.

Issue #7212, opened 2022-07-27
Add Fragment.OnCreateAnimation(int32, bool, int32) to the list that saves us from a `System.Reflection.Emit` call


PureWeen:
### Android application type

.NET 7.0 MAUI
Android for .NET (net7.0-android)

### Affected platform version

.NET 7

### Description

I'm currently generating the AOT Profiles for non shell apps in .NET MAUI

There's one place where it's currently falling back to `System.Reflection.Emit`

https://github.com/dotnet/maui/blob/0062faeba1dd391b1468798cd29299403b9b67f0/src/Core/src/Platform/Android/Navigation/NavigationViewFragment.cs#L88

```bash
2022-07-25 11:25:06.254 14932-14932/? D/monodroid-assembly: Falling back to System.Reflection.Emit for delegate type '_JniMarshal_PPIZI_L': IntPtr n_OnCreateAnimation_IZI(IntPtr, IntPtr, Int32, Boolean, Int32)
```

We only need these changes for NET7 

### Steps to Reproduce

If you run the following project [ShellVsNonShell.zip](https://github.com/xamarin/xamarin-android/files/9203635/ShellVsNonShell.zip)

with the following logging enabled then you should see it popup in logcat

```bash
adb shell setprop debug.mono.log default,assembly,mono_log_level=debug,mono_log_mask=all
```

You might also see one around `OnPageScrolled` that's ok to ignore. We have a fix in the MAUI repository to remove this code path. 

### Relevant log output

```shell
2022-07-25 11:25:06.254 14932-14932/? D/monodroid-assembly: Falling back to System.Reflection.Emit for delegate type '_JniMarshal_PPIZI_L': IntPtr n_OnCreateAnimation_IZI(IntPtr, IntPtr, Int32, Boolean, Int32)
```
```

jonpryor:
@PureWeen: We can't add *all* possible method signatures (there are hundreds in `Mono.Android.dll` alone), so the question is: where do we draw the line?

The current "line" is "used by our default templates", specifically all `android` and `maui` templates.

It sounds like this signature isn't used by the current default templates.  As such, I am inclined to skip this signature.

In .NET 8, this should be moot, as we should be using LLVM Marshal Methods (e1af9587bb98d4c249bbc392ceccc2b53ffff155), which avoids the need for `System.Reflection.Emit` wrappers at runtime.

PureWeen:
[...]

PR #7267, opened 2022-08-16
[Mono.Android] add "built-in" delegate for MAUI+non-Shell

jonathanpeppers:
Context: https://github.com/dotnet/maui/commit/15df4ed8fc8a8f84b090b1afae08303d847c481f
Fixes: https://github.com/xamarin/xamarin-android/issues/7212

In the app we're recording AOT profiles for .NET MAUI, we added a
`FlyoutPage` and navigated to it. This was good because it added
common scenarios to the profile:

* Navigation from one page to another
* A non-Shell API, `FlyoutPage`
* `FlyoutPage` also implicitly uses `NavigationPage`?

This caused the app to start using System.Reflection.Emit and logging:

```
08-16 09:46:24.687 31517 31517 D monodroid-assembly: Falling back to System.Reflection.Emit for delegate type '_JniMarshal_PPIZI_L': IntPtr n_OnCreateAnimation_IZI(IntPtr, IntPtr, Int32, Boolean, Int32)
```

I added a case for `_JniMarshal_PPIZI_L`, this log message is now gone
when I build and run this app.

I think this should be the last one of these we add, unless one
appears from either:

* `dotnet new maui` or `dotnet new maui-blazor`
* [.NET Podcast app][0]

[0]: https://github.com/microsoft/dotnet-podcasts

### DIFF
--- a/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs
+++ b/src/Mono.Android/Android.Runtime/JNINativeWrapper.g.cs
@@ -367,6 +367,17 @@ namespace Android.Runtime
 			}
 		}
 
+		internal static IntPtr Wrap_JniMarshal_PPIZI_L (this _JniMarshal_PPIZI_L callback, IntPtr jnienv, IntPtr klazz, int p0, bool p1, int p2)
+		{
+			JNIEnv.WaitForBridgeProcessing ();
+			try {
+				return callback (jnienv, klazz, p0, p1, p2);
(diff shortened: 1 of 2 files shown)


PureWeen:
> FlyoutPage also implicitly uses NavigationPage?

For the non-shell path I just made sure to include all non-shell page types (FlyoutPage, TabbedPage, NavigationPage) so that all AOT paths would be covered by whatever combination of these three pages that users will use.

PR #7269, opened 2022-08-16
[tests] ignore InstallAndroidDependenciesTest

jonathanpeppers:
Context: https://github.com/xamarin/android-sdk-installer/pull/685

Our `InstallAndroidDependenciesTest` started failing with:

    Xamarin.Installer.Common.targets(12,3): error : Element 'type-details' with parser TypeDetailsSourceParser has a child element 'extension-level' but this version of the library does not support it (https://dl.google.com/android/repository/repository2-3.xml)

Since the fix is going into main / .NET 7, we think we should just
ignore this test for now in .NET 6.


jonathanpeppers:
Let me review, why this test failed with:
```
_AndroidSdkDirectory was not set to new SDK path.
Expected: True
But was:  False
```

jonathanpeppers:
Looks to be ignored now:

image

jonathanpeppers:
The other test failures, we're fixing in main -- which I'll start backporting as we get those working.

pjcollins (approved):

PR #7285, opened 2022-08-19
Marshal methods part 4 of 5 (runtime)

grendello:
Context: https://github.com/xamarin/java.interop/issues/1027

Implement the necessary runtime elements to enable running of applications with marshal methods.
This commit, if marshal methods are enabled, makes it possible to run both plain XA and MAUI apps.

The `[UnmanagedCallersOnly]` attribute used by marshal methods requires that the invoked method
doesn't have any non-blittable types as either its return type or type of any parameters passed
to the method.  Implement generation of wrapper methods which replace `bool` (the only non-blittable
type currently found in any Mono.Android, AndroidX bindings or in MAUI) with `byte` and convert
the values appropriately before calling the actual target method.  In a "hello world" MAUI test
application there are 133 such methods (out of around 180 total).

Implement code that enables us to show error messages with the proper assembly, class and method
name on failure to look up or obtain pointer to native callback methods.


src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Runtime.InteropServices.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<linker>
+	<!-- Workaround for https://github.com/xamarin/xamarin-android/issues/6626, remove once https://github.com/dotnet/runtime/pull/68265 lands -->
```
jonpryor, line 3: Why are these workarounds still necessary?  dotnet/runtime#68265 was merged 2 months ago; that fix be in .NET 7.

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -473,8 +473,30 @@ static bool CallRegisterMethodByIndex (JniNativeMethodRegistrationArguments argu
 		public override void RegisterNativeMembers (JniType nativeClass, Type type, string? methods) =>
 			RegisterNativeMembers (nativeClass, type, methods.AsSpan ());
 
+#if ENABLE_MARSHAL_METHODS
+		// Temporary hack, see comments in RegisterNativeMembers below
+		static readonly Dictionary<string, List<string>> dynamicRegistrationMethods = new Dictionary<string, List<string>> (StringComparer.Ordinal) {
```
jonpryor, line 478: This would be more efficient if it were a `Dictionary<string, string[]>` instead of a `Dictionary<string, List<string>>`, as the `List<string>` contains a `string[]` anyway…

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -473,8 +473,30 @@ static bool CallRegisterMethodByIndex (JniNativeMethodRegistrationArguments argu
...
+		};
+#endif
+
 		public void RegisterNativeMembers (JniType nativeClass, Type type, ReadOnlySpan<char> methods)
 		{
+#if ENABLE_MARSHAL_METHODS
+			Logger.Log (LogLevel.Info, "monodroid-mm", $"RegisterNativeMembers ('{nativeClass?.Name}', '{type?.FullName}', '{methods.ToString ()}')");
+			Logger.Log (LogLevel.Info, "monodroid-mm", "RegisterNativeMembers called from:");
+			var st = new StackTrace (true);
+			foreach (string l in st.ToString ().Split ("\n")) {
```
jonpryor, line 491: `Logger.Log()` *already* splits on newline: https://github.com/xamarin/xamarin-android/blob/903ba37ce70d2840983774e1d6fb55f8002561e2/src/Mono.Android/Android.Runtime/Logger.cs#L50

There is no need to split on `\n`.

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -530,20 +558,78 @@ public void RegisterNativeMembers (JniType nativeClass, Type type, ReadOnlySpan<
...
+								string dynamicTypeName = kvp.Key;
+
+								foreach (string dynamicCallbackMethodName in kvp.Value) {
+									if (ShouldRegisterDynamically (declaringTypeName, callbackName, dynamicTypeName, dynamicCallbackMethodName)) {
+										createCallback = true;
+										break;
+									}
+								}
+
+								if (createCallback) {
```
jonpryor, line 578: This raises a "meta-question": the implication here is that whether we register via `JNIEnv::RegisterNatives()` or via C `Java_…` function is done on a *per-method* basis.

Is that actually supported on Android?

Rephrased; given the Java type:

```java
package example;

public class C {
    public static native void m_viaRegister();
    public static native void m_viaFunction();
}
```

and then at runtime we have a `Java_example_C_m_1viaFunction()` function in a loaded `.so`, and we also call `JNIEnv:RegisterNatives()` but *only* for `m_viaRegister()`…

Does that work?  The JNI docs are not clear on how these two mechanisms interact.

For that matter, if *both* native mechanisms are provided, which mechanism actually invoked at runtime?  Which has priority?

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
```diff
@@ -530,20 +558,78 @@ public void RegisterNativeMembers (JniType nativeClass, Type type, ReadOnlySpan<
...
-						natives [nativesIndex++] = new JniNativeMethodRegistration (name.ToString (), signature.ToString (), callback);
 					}
 
 					methodsSpan = newLineIndex != -1 ? methodsSpan.Slice (newLineIndex + 1) : default;
 				}
 
-				JniEnvironment.Types.RegisterNatives (nativeClass.PeerReference, natives, natives.Length);
+#if ENABLE_MARSHAL_METHODS
+				if (needToRegisterNatives) {
+					// We need to reallocate as JniEnvironment.Types.RegisterNatives uses a `foreach` loop and will NREX otherwise (since we aren't filling all
```
jonpryor, line 609: We should fix Java.Interop so that null `.Marshaler` values are either allowed, or allowed after `numMethods`.

Rephrased: [`JniEnvironment.Types.RegisterNatives()`](https://github.com/xamarin/java.interop/blob/6d1ae4ee1076341f8af36dbdd306dce0e6266999/src/Java.Interop/Java.Interop/JniEnvironment.Types.cs#L204-L211) should use a `for` loop instead:

```csharp
for (int i = 0; i < numMethods; ++i) {
    var m = methods [i];
    …
}
```

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -34,13 +35,40 @@ public void Rewrite (DirectoryAssemblyResolver resolver)
 				unmanagedCallersOnlyAttributes.Add (asm, CreateImportedUnmanagedCallersOnlyAttribute (asm, unmanagedCallersOnlyAttributeCtor));
 			}
 
+			Console.WriteLine ();
```
jonpryor, line 38: Shouldn't this and other `Console.WriteLine()`s here use `Log.LogDebugMessage()` instead?

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -95,10 +123,161 @@ void MoveFile (string source, string target)
 				Files.CopyIfChanged (source, target);
 				try {
 					File.Delete (source);
-				} catch (Exception ex) {
+				} catch (Exception) {
```
jonpryor, line 126: Instead of removing `ex`, we should add a `Log.LogDebugMessage (ex)` to the `catch` block.

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -95,10 +123,161 @@ void MoveFile (string source, string target)
...
 					log.LogWarning ($"Unable to delete source file '{source}' when moving it to '{target}'");
 				}
 			}
+
+			string ToStringOrNull (object? o)
```
jonpryor, line 131: I totally expected this method to "normalize" the empty string to `null`, not what this method *actually* does.

Also, given that this appears to be for "debug" output, why is it `'null'` (with single-quotes)?  Wouldn't outuput of:

```
method.CallbackField == null
```

make more sense than the current:

```
method.CallbackField == 'null'
```

?

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
```diff
@@ -95,10 +123,161 @@ void MoveFile (string source, string target)
...
+					return "'null'";
+				}
+
+				return o.ToString ();
+			}
+		}
+
+		MethodDefinition GenerateBlittableWrapper (MarshalMethodEntry method, Dictionary<AssemblyDefinition, CustomAttribute> unmanagedCallersOnlyAttributes)
+		{
+			Console.WriteLine ($"\t  Generating blittable wrapper for: {method.NativeCallback.FullName}");
```
jonpryor, line 143: s/Console.WriteLine/Log.LogDebugMessage/g

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsClassifier.cs
```diff
@@ -13,29 +14,44 @@ namespace Xamarin.Android.Tasks
 #if ENABLE_MARSHAL_METHODS
 	public sealed class MarshalMethodEntry
```
jonpryor, line 15: Is there actually need for many of these types to be `public`?  We're not hitting them from a separate unit test assembly, so should they be public at all?

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsClassifier.cs
```diff
@@ -232,6 +283,9 @@ bool IsStandardHandler (TypeDefinition topType, ConnectorInfo connector, MethodD
 				return false;
 			}
 
+			// TODO: if we can't native callback and/or delegate field using `callbackNameCore`, fall back to `jniName` (which is the first argument to the `[Register]`
```
jonpryor, line 286: Is there a missing verb here?  "if we can't [verb?] native callback and/or …"

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsNativeAssemblyGenerator.cs
```diff
@@ -177,40 +205,86 @@ sealed class MarshalMethodsManagedClass
...
+			if (marshalMethods == null || marshalMethods.Count == 0) {
 				return;
 			}
 
 			var seenClasses = new Dictionary<string, int> (StringComparer.Ordinal);
-			methods = new List<MarshalMethodInfo> ();
-			foreach (IList<MarshalMethodEntry> entryList in MarshalMethods.Values) {
+			var allMethods = new List<MarshalMethodInfo> ();
+
+			// It's possible that several otherwise different methods (from different classes, but with the same
```
jonpryor, line 230: Is there an example of this?  Given that the `Java_…` names contain the declaring package name and declaring type name, *how* can two different methods "from different classes" have the same name?

Or is "name" in this context the *method* name, not the `Java_…` name?

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsNativeAssemblyGenerator.cs
```diff
@@ -260,24 +361,52 @@ void ProcessAndAddMethod (MarshalMethodEntry entry, bool useFullNativeSignature,
...
+					case '$':
+						sb.Append ("_00024");
+						break;
+
+					default:
+						if ((int)ch > 127) {
+							sb.Append ("_0");
+							sb.Append (((int)ch).ToString ("x04"));
+						} else {
+							sb.Append (ch);
```
jonpryor, line 405: The problem with this is that it'll "pass through" characters which shouldn't be passed through:

  * https://discord.com/channels/732297728826277939/732297837953679412/1009499855674163231
  * https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.2.2

In particular, Kotlin methods can contain `-`.  (We don't support overriding such methods yet, as we use Java as an intermediary, but if we ever start using Kotlin as an intermediary, we *will* be able to override such methods.)

Thus, what we should instead do is allow *only* `[a-zA-Z0-9]`, and escape *anything else*:

```csharp
if ((ch >= '0' && ch <= '0') ||
        (ch >= 'a' && ch <= 'z') ||
        (ch >= 'A' && ch <= 'Z')) {
   sb.Append (ch);
} else {
    sb.Append ("_0").Append (((int) ch).ToString ("x04"));
}
```

src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Runtime.InteropServices.xml
grendello: The comment is misleading, it applies to a different issue. `System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute` must be preserved here or we get build errors:

```
Xamarin.Android.Common.targets(1463,3): error XAGJS7009: System.InvalidOperationException: Unable to find the System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute type
Xamarin.Android.Common.targets(1463,3): error XAGJS7009:    at Xamarin.Android.Tasks.MarshalMethodsAssemblyRewriter.GetUnmanagedCallersOnlyAttributeConstructor(DirectoryAssemblyResolver resolver)
Xamarin.Android.Common.targets(1463,3): error XAGJS7009:    at Xamarin.Android.Tasks.MarshalMethodsAssemblyRewriter.Rewrite(DirectoryAssemblyResolver resolver)
Xamarin.Android.Common.targets(1463,3): error XAGJS7009:    at Xamarin.Android.Tasks.GenerateJavaStubs.Run(DirectoryAssemblyResolver res)
Xamarin.Android.Common.targets(1463,3): error XAGJS7009:    at Xamarin.Android.Tasks.GenerateJavaStubs.RunTask()
Xamarin.Android.Common.targets(1463,3): error XAGJS7009:    at Microsoft.Android.Build.Tasks.AndroidTask.Execute()
```

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
grendello: Answering that would require a deeper dive into Android JNI sources I suppose...

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
grendello: Not all of them will stay. The next PR in the series cleans that up.

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
grendello: This method will go away in the next PR in the series.

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsAssemblyRewriter.cs
grendello: Most of these will go away in the next PR

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsClassifier.cs
grendello: The next PR in the series will do all kinds of cleanups like this.

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsNativeAssemblyGenerator.cs
grendello: Yep, there are a few examples in the Hello Maui sample app:

```
Overloaded MM: Java_crc64e1fb321c08285b90_CellAdapter_n_1onActionItemClicked
  implemented in: Microsoft.Maui.Controls.Handlers.Compatibility.CellAdapter (System.Boolean Android.Views.ActionMode/ICallback::OnActionItemClicked(Android.Views.ActionMode,Android.Views.IMenuItem))
    mangling 'crc64e1fb321c08285b90/CellAdapter'
    mangling 'n_onActionItemClicked'
  Using FULL signature
    mangling 'Landroid/view/ActionMode;Landroid/view/MenuItem;'
     new native symbol name: Java_crc64e1fb321c08285b90_CellAdapter_n_1onActionItemClicked__Landroid_view_ActionMode_2Landroid_view_MenuItem_2
  implemented in: Microsoft.Maui.Controls.Handlers.Compatibility.CellAdapter (System.Boolean AndroidX.AppCompat.View.ActionMode/ICallback::OnActionItemClicked(AndroidX.AppCompat.View.ActionMode,Android.Views.IMenuItem))
    mangling 'crc64e1fb321c08285b90/CellAdapter'
    mangling 'n_onActionItemClicked'
  Using FULL signature
    mangling 'Landroidx/appcompat/view/ActionMode;Landroid/view/MenuItem;'
     new native symbol name: Java_crc64e1fb321c08285b90_CellAdapter_n_1onActionItemClicked__Landroidx_appcompat_view_ActionMode_2Landroid_view_MenuItem_2

Overloaded MM: Java_crc64e1fb321c08285b90_CellAdapter_n_1onCreateActionMode
  implemented in: Microsoft.Maui.Controls.Handlers.Compatibility.CellAdapter (System.Boolean Android.Views.ActionMode/ICallback::OnCreateActionMode(Android.Views.ActionMode,Android.Views.IMenu))
    mangling 'crc64e1fb321c08285b90/CellAdapter'
    mangling 'n_onCreateActionMode'
  Using FULL signature
    mangling 'Landroid/view/ActionMode;Landroid/view/Menu;'
     new native symbol name: Java_crc64e1fb321c08285b90_CellAdapter_n_1onCreateActionMode__Landroid_view_ActionMode_2Landroid_view_Menu_2
  implemented in: Microsoft.Maui.Controls.Handlers.Compatibility.CellAdapter (System.Boolean AndroidX.AppCompat.View.ActionMode/ICallback::OnCreateActionMode(AndroidX.AppCompat.View.ActionMode,Android.Views.IMenu))
    mangling 'crc64e1fb321c08285b90/CellAdapter'
    mangling 'n_onCreateActionMode'
  Using FULL signature
    mangling 'Landroidx/appcompat/view/ActionMode;Landroid/view/Menu;'
     new native symbol name: Java_crc64e1fb321c08285b90_CellAdapter_n_1onCreateActionMode__Landroidx_appcompat_view_ActionMode_2Landroid_view_Menu_2

Overloaded MM: Java_crc64e1fb321c08285b90_CellAdapter_n_1onDestroyActionMode
  implemented in: Microsoft.Maui.Controls.Handlers.Compatibility.CellAdapter (System.Void Android.Views.ActionMode/ICallback::OnDestroyActionMode(Android.Views.ActionMode))
    mangling 'crc64e1fb321c08285b90/CellAdapter'
    mangling 'n_onDestroyActionMode'
  Using FULL signature
    mangling 'Landroid/view/ActionMode;'
     new native symbol name: Java_crc64e1fb321c08285b90_CellAdapter_n_1onDestroyActionMode__Landroid_view_ActionMode_2
  implemented in: Microsoft.Maui.Controls.Handlers.Compatibility.CellAdapter (System.Void AndroidX.AppCompat.View.ActionMode/ICallback::OnDestroyActionMode(AndroidX.AppCompat.View.ActionMode))
    mangling 'crc64e1fb321c08285b90/CellAdapter'
    mangling 'n_onDestroyActionMode'
  Using FULL signature
    mangling 'Landroidx/appcompat/view/ActionMode;'
     new native symbol name: Java_crc64e1fb321c08285b90_CellAdapter_n_1onDestroyActionMode__Landroidx_appcompat_view_ActionMode_2

Overloaded MM: Java_crc64e1fb321c08285b90_CellAdapter_n_1onPrepareActionMode
  implemented in: Microsoft.Maui.Controls.Handlers.Compatibility.CellAdapter (System.Boolean Android.Views.ActionMode/ICallback::OnPrepareActionMode(Android.Views.ActionMode,Android.Views.IMenu))
    mangling 'crc64e1fb321c08285b90/CellAdapter'
    mangling 'n_onPrepareActionMode'
  Using FULL signature
    mangling 'Landroid/view/ActionMode;Landroid/view/Menu;'
     new native symbol name: Java_crc64e1fb321c08285b90_CellAdapter_n_1onPrepareActionMode__Landroid_view_ActionMode_2Landroid_view_Menu_2
  implemented in: Microsoft.Maui.Controls.Handlers.Compatibility.CellAdapter (System.Boolean AndroidX.AppCompat.View.ActionMode/ICallback::OnPrepareActionMode(AndroidX.AppCompat.View.ActionMode,Android.Views.IMenu))
    mangling 'crc64e1fb321c08285b90/CellAdapter'
    mangling 'n_onPrepareActionMode'
  Using FULL signature
    mangling 'Landroidx/appcompat/view/ActionMode;Landroid/view/Menu;'
     new native symbol name: Java_crc64e1fb321c08285b90_CellAdapter_n_1onPrepareActionMode__Landroidx_appcompat_view_ActionMode_2Landroid_view_Menu_2
```

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsNativeAssemblyGenerator.cs
grendello: I don't know if restricting the allowed character set to `[a-zA-Z0-9]` will yield the same symbol name as computed by JNI. According to [JNI docs](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#resolving_native_method_names), only non-ASCII characters are encoded in this way. I'd prefer to keep the current algorithm until we actually hit the Kotlin `-` problem.

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
jonpryor: I don't think this requires reading Android source.  An on-device test would answer this as well.  (Though that answer could change over time, and thus wouldn't be a "final" answer, but it could still be informative.)

src/Mono.Android/Android.Runtime/AndroidRuntime.cs
jonpryor: Should be fixed by: https://github.com/xamarin/java.interop/pull/1029

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsNativeAssemblyGenerator.cs
jonpryor: Either you updated the comment, or I *horribly* misread this comment; I thought you were saying that the *long*/*full* name had conflicts/collisions.

What you're showing here are *overloads*.

For example, consider `Activity.startActivity()`:

  * [`Activity.startActivity(Intent)`](https://developer.android.com/reference/android/app/Activity#startActivity(android.content.Intent))
  * [`Activity.startActivity(Intent, Bundle)`](https://developer.android.com/reference/android/app/Activity#startActivity(android.content.Intent,%20android.os.Bundle))

You should be able to override *both* of these in C#:

```csharp
class MyActivity : Activity {
    public override unsafe void StartActivity (Android.Content.Intent? intent) => …;
    public override unsafe void StartActivity (Android.Content.Intent? intent, Android.OS.Bundle? options) => …;
}
```

This would also result in needing to use the full `Java_…` name.  (This would likely also make a good unit test in a future commit...)

Updating the comment to use the term "method overload" would be a helpful clarification.  The fact that the methods being overridden came from different classes/interfaces is only slightly relevant; the more important bit is that within the JCW type, these are method *overloads*, a'la:

```java
interface IntFooable {
    void foo(int value);
}
interface StringFooable {
    void foo(String value);
}
class MyFooable implements IntFooable, StringFooable {
    public native void foo(int value) {}
    public native void foo(String value) {}
}
```

Here, you can't just have `Java_example_MyFooable_foo`, because the method is overloaded.  Instead, the full names must be used.

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsNativeAssemblyGenerator.cs
jonpryor: Would make for a fascinating thing to test.  Where's a good Java bytecode assembler?  I need to create a Java method which contains `\n`!

src/Xamarin.Android.Build.Tasks/Utilities/MarshalMethodsNativeAssemblyGenerator.cs
grendello: I modified the comment by adding the examples, but the rest of the wording remained the same. The "simple" overloads you give above are handled in the classifer, the case handled here is when two different managed types will end up having the same Java package name (like the "standard" Android vs AndroidX overloads in the comment) and thus become "native" overloads, requiring native symbol name with full signature.

jonpryor:
```markdown
Context: 903ba37ce70d2840983774e1d6fb55f8002561e2

Context: https://github.com/xamarin/java.interop/issues/1027

Commit 903ba37c mentioned a TODO:
    
> Update/rewrite infrastructure to focus on implementing the runtime
> side of marshal methods, making it possible to actually run
> applications which use marshal methods.

Implement the necessary runtime elements to enable running of
applications with marshal methods.  It is now possible, if
LLVM marshal methods are enabled/`ENABLE_MARSHAL_METHODS` is defined,
to run both plain .NET SDK for Android and MAUI apps.

Update `src/Microsoft.Android.Sdk.ILLink/PreserveLists/System.Runtime.InteropServices.xml`
so that `System.Runtime.InteropServices.UnmanagedCallersOnlyAttribute`
is always preserved, as it is required by LLVM Marshal Methods.

The `[UnmanagedCallersOnly]` attribute used by marshal methods
requires that the invoked method have [blittable types][0] for the
method return type and all parameter types.  Unfortunately, `bool`
is *not* a blittable type.  Implement generation of wrapper methods
which replace `bool` with `byte` and convert the values appropriately
before calling the actual target method.  In a "hello world" MAUI
test app there are 133 such methods (out of around 180 total).

Implement code that enables us to show error messages with the proper
assembly, class and method names on failure to look up or obtain
pointer to native callback methods.

TODO:

  * Process *all* assemblies, including `Mono.Android.dll`, for
    Java Callable Wrapper generation.  This is necessary so that we
    can find and emit LLVM marshal methods for types defined within
    `Mono.Android.dll`.

  * Remove the `ENABLE_MARSHAL_METHODS` define, and enable
    LLVM marshal methods for everyone.

  * Update `<GenerateJavaStubs/>` to rewrite all assemblies for all
    Supported ABIs.  Currently, we don't support `Java.Lang.Object` &
    `Java.Lang.Throwable` subclasses being located in per-ABI
    assemblies.

  * How do `Java_…` native functions interact with
    `JNIEnv::RegisterNatives()`?
    https://github.com/xamarin/xamarin-android/pull/7285#discussion_r951755971

  * *Can* JNI `native` methods contain "non-printable" characters
    such as `\n`, or "non-representable in ELF symbols" characters
    such as `-` (e.g. Kotlin mangled methods)?
    https://github.com/xamarin/xamarin-android/pull/7285#discussion_r951789869

  * Cleanup, cleanup, cleanup

[0]: https://docs.microsoft.com/en-us/dotnet/framework/interop/blittable-and-non-blittable-types
```

PR #7286, opened 2022-08-19
[build] disable classic tests on .NET 6 release branches

jonathanpeppers:
Many of our Windows lanes on `release/6.0.4xx` are failing with:

    Xamarin.Provisioning.ProvisioningException: Failed to provision
    ---> Xamarin.Provisioning.ProvisioningException: Only extension versions greater than that in the instance catalog version (13.0.0.0) can be installed.

This is because the CI machines have VS 17.3 now, and the classic
Xamarin.Android we're building is too old.

For now, let's just disable all these classic tests on this branch. We
can come up with a better idea for this later.

### DIFF
--- a/build-tools/automation/yaml-templates/run-installer.yaml
+++ b/build-tools/automation/yaml-templates/run-installer.yaml
@@ -8,8 +8,7 @@ steps:
     artifactName: $(InstallerArtifactName)
     downloadPath: $(System.DefaultWorkingDirectory)
     patterns: xamarin.android*.pkg
-  # Disabled on .NET 6 release branches
-  condition: false
+  condition: and(succeeded(), eq(variables['agent.os'], 'Darwin'))
 
 - task: DownloadPipelineArtifact@2
   inputs:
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
This gets us past the `provisionator.exe` errors, but I think we still need this to address a few others: https://github.com/xamarin/xamarin-android/pull/7295

build-tools/automation/yaml-templates/run-installer.yaml
```diff
@@ -8,14 +8,16 @@ steps:
     artifactName: $(InstallerArtifactName)
     downloadPath: $(System.DefaultWorkingDirectory)
     patterns: xamarin.android*.pkg
-  condition: and(succeeded(), eq(variables['agent.os'], 'Darwin'))
+  # Disabled on .NET 6 release branches
+  condition: false
```
pjcollins, line 12: I think we could keep this enabled to help make the macOS smoke test jobs more reliable, as they run a combination of classic and .NET tests.

pjcollins (approved):

build-tools/automation/yaml-templates/run-installer.yaml
```diff
@@ -29,14 +31,16 @@ steps:
     }
     Write-Host "##vso[task.setvariable variable=XA.Provisionator.Args]$installer"
   displayName: find installer and set provisionator variable
-  condition: and(succeeded(), ne(variables['agent.os'], 'Linux'))
```
pjcollins, line 32: If we keep the .pkg download we can change this condition to:
```yaml
and(succeeded(), ne(variables['agent.os'], 'Windows_NT'))
```

jonathanpeppers:
This looks a lot greener than it has been, going to merge and then rebase this one: #7288

PR #7305, opened 2022-08-24
Bump NDK to r25b

grendello:
Changes: https://github.com/android/ndk/wiki/Changelog-r25#r25b

There are no interesting changes in this version, from our point
of view.  Bumping merely to track the current supported version
of the NDK.


grendello:
The test failures are unrelated to the PR, merging.

Issue #7194, opened 2022-07-21
Make `$(AndroidLinkResources)` able to be the default for MAUI projects


jonathanpeppers:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

.NET 7

### Description

Context: https://github.com/dotnet/maui/pull/7038

We found some Telerik and/or other vendor libraries would fail due to this code in SkiaSharp:

https://github.com/mono/SkiaSharp/blob/4a687a87b86f127ccc23116ee410dcb5df8fca63/source/SkiaSharp.Views/SkiaSharp.Views.Android/SKCanvasView.cs#L48-L61

So *if* we don't finish #6427 for .NET 7, we could consider:
1. Fix `SKCanvasView` so it doesn't use a `Styleable int[]`.
2. Give an error message at build time if the `AndroidLinkResources` linker step encounters a `int[]`
3. Try turning on `AndroidLinkResources=true` by default for MAUI projects

### Steps to Reproduce

1. Use `SKCanvasView` in a project
2. Build with `-c Release -p:AndroidLinkResources=true`

### Did you find any workaround?

Don't use `$(AndroidLinkResources)`

tipa:
I was hitting the same problem today (not using a 3rd party library) when subclassing a View and using `ObtainStyledAttributes`.
Attached a simple repo project: [AndroidApp2.zip](https://github.com/xamarin/xamarin-android/files/9302722/AndroidApp2.zip)
 
Would appreciate if a fix could be found

jonathanpeppers:
@tipa can you share the error message you got? Probably a crash at runtime?

tipa:
With the demo project attached I first get this exception (only when enabling all exceptions in the VS Exception Settings Window):
`System.BadImageFormatException: 'Could not resolve field token 0x0400000b'
`

When I then click "Continue" in the debugger I get this exception:
`System.NotSupportedException: 'Could not activate JNI Handle 0x7fe1003920 (key_handle 0x73cc609) of Java type 'crc641149b9fe658fbe8e/MyView' as managed type 'AndroidApp2.MyView'.'
`

They are crashes at runtime, yes

PR #7306, opened 2022-08-24
AndroidLinkResources and Styleables

dellis1972:
Fixes https://github.com/xamarin/xamarin-android/issues/7194
Context https://github.com/dotnet/maui/pull/7038

The initial version of `AndroidLinkResources` commited in 9e6ce03c was to broad in its removal of Fields. 
Certain field usage like `Styleable` arrays were not called using the IL`stsfld` op code. As a result they could
not be easily replaced with constant usage. However the linker removed ALL the fields from the `Resource` 
sub classes. This would result in the following error

`System.BadImageFormatException: 'Could not resolve field token 0x0400000b'`.

This was because the `int[]` fields were removed as part of the linking process. 

So to fix this we need to leave the `int[]` fields in the `Resource` subclasses and not remove them. We can 
still remove all the other `int` fields though. We now also need to fix up the `Resource` subclass constructors 
to replace the `int` field access with the constant values like we do for the rest of the app. 
This was not required because because these constructors were removed. But now we have to keep them because
the static array initialisation takes place in these constructors.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/LinkDesignerBase.cs
+++ b/src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/LinkDesignerBase.cs
@@ -7,6 +7,7 @@ using Xamarin.Android.Tasks;
 using System.Collections.Generic;
 using Mono.Cecil.Cil;
 using System.Text.RegularExpressions;
+using Mono.Collections.Generic;
 #if ILLINK
 using Microsoft.Android.Sdk.ILLink;
 #endif
@@ -69,8 +70,17 @@ namespace MonoDroid.Tuner  {
 		protected void ClearDesignerClass (TypeDefinition designer)
 		{
 			LogMessage ($"    TryRemoving {designer.FullName}");
-			designer.NestedTypes.Clear ();
-			designer.Methods.Clear ();
+			// for each of the nested types clear all but the
+			// int[] fields.
+			for (int i = designer.NestedTypes.Count -1; i >= 0; i--) {
+				var nestedType = designer.NestedTypes [i];
+				RemoveFieldsFromType (nestedType, designer.Module);
+				if (nestedType.Fields.Count == 0) {
+					// no fields we do not need this class at all.
+					designer.NestedTypes.RemoveAt (i);
+				}
+			}
+			RemoveUpdateIdValues (designer);
 			designer.Fields.Clear ();
 			designer.Properties.Clear ();
 			designer.CustomAttributes.Clear ();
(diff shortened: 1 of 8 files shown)


samples/HelloWorld/HelloWorld/MainActivity.cs
```diff
@@ -25,7 +25,7 @@ protected override void OnCreate (Bundle savedInstanceState)
 			Button button = FindViewById<Button> (Resource.Id.myButton);
 
 			button.Click += delegate {
-				button.Text = string.Format ("{0} clicks!", count++);
+				button.Text = string.Format ("{0} clicks!", (byte)count++);
```
jonathanpeppers, line 28: Do we need this change? Maybe leftover from something?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/LinkDesignerBase.cs
```diff
@@ -69,8 +70,16 @@ protected bool FindResourceDesigner (AssemblyDefinition assembly, bool mainAppli
...
-			designer.Methods.Clear ();
+			// for each of the nested types clear all but the
+			// int[] fields.
+			foreach (var nestedType in designer.NestedTypes) {
+				RemoveFieldsFromType (nestedType, designer.Module);
+				if (nestedType.Fields.Count == 0) {
+					// no fields we do not need any methods like the static ..cctor
+					nestedType.Methods.Clear ();
+				}
+			}
```
jonathanpeppers, line 81: Can we remove the entire type? something like:
```suggestion
			foreach (var nestedType in designer.NestedTypes.ToArray ()) { // or reverse for-loop to avoid LINQ
				RemoveFieldsFromType (nestedType, designer.Module);
				if (nestedType.Fields.Count == 0) {
					// no fields we do not need this type at all
					designer.NestedTypes.Remove (nestedType);
				}
			}
```

Or would this break something?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/LinkDesignerBase.cs
```diff
@@ -117,6 +126,52 @@ protected void FixType (TypeDefinition type, TypeDefinition localDesigner)
 			}
 		}
 
+		protected void RemoveFieldsFromType (TypeDefinition type, ModuleDefinition module)
+		{
+			List<FieldDefinition> fields = new List<FieldDefinition> ();
+			TypeReference intArray = new ArrayType (module.TypeSystem.Int32);
```
jonathanpeppers, line 132: I don't think this value is used?

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/LinkDesignerBase.cs
```diff
@@ -117,6 +126,52 @@ protected void FixType (TypeDefinition type, TypeDefinition localDesigner)
...
+			foreach (var field in type.Fields) {
+				if (field.FieldType.IsArray) {
+					continue;
+				}
+				fields.Add (field);
+			}
+			foreach (var field in fields) {
+				LogMessage ($"Removing {type.Name}::{field.Name}");
+				type.Fields.Remove (field);
+			}
```
jonathanpeppers, line 142: I think this also would work, and you wouldn't need an intermediate list:
```suggestion
			for (int i = fields.Length - 1; i >= 0; i--) {
				var field = fields [i];
				if (field.FieldType.IsArray) {
					continue;
				}
				LogMessage ($"Removing {type.Name}::{field.Name}");
				type.Fields.RemoveAt (i);
			}
```

jonathanpeppers (approved):

PR #7311, opened 2022-08-24
[ci] Add support to net7.0 for multi-targeting in VS

pjcollins:
Context: https://github.com/xamarin/yaml-templates/pull/199

Updates the build to use the latest MSI generation template. The v3
refactoring, support for multi-targeting, and support for workload pack
group MSIs.

The build will now produce two different VS Drop artifacts.  The MSI and
VSMAN files generated for SDK packs have been split out into a new
`vsdrop-multitarget-signed` artifact, allowing us to include multiple
versions of the SDK packs in VS.

The `Microsoft.Android.Sdk` pack has been renamed to
`Microsoft.Android.Sdk.net7` to allow it to be imported by the .NET 8
workload manifest when the time comes.

### DIFF
--- a/Configuration.props
+++ b/Configuration.props
@@ -47,7 +47,7 @@
     <DebugType Condition=" '$(DebugType)' == '' ">portable</DebugType>
     <Deterministic Condition=" '$(Deterministic)' == '' ">True</Deterministic>
     <LangVersion Condition=" '$(LangVersion)' == '' ">latest</LangVersion>
-    <AndroidNet6Version Condition=" '$(AndroidNet6Version)' == '' ">32.0.461</AndroidNet6Version>
+    <AndroidNet6Version Condition=" '$(AndroidNet6Version)' == '' ">32.0.465</AndroidNet6Version>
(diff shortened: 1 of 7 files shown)


jonathanpeppers (approved):
This looks good, if CI is green. It seems like the main `NuGet.config` might need the `dotnet6` feed, we'll see.

pjcollins:
> It seems like the main `NuGet.config` might need the `dotnet6` feed, we'll see.

I think we can get away with only adding this to certain tests since it is already included in the [workload install command](https://github.com/xamarin/xamarin-android/blob/e4e9153f9ed4c7e90862b8f0fbc207dee526e7e4/build-tools/create-packs/Directory.Build.targets#L126) used for test environment provisioning.

There are some ILLink related failures on the latest run that seem interesting though, rerunning a couple of jobs to see if they pop up again...

pjcollins:
Looks like the test job specific re-runs passed, I'm going to merge so we can get a build and we can look out for any repeat failures in main.

PR #7321, opened 2022-08-30
[ci] Split up signing for files to be notarized

pjcollins:
The signing steps that run on macOS to prepare a PKG for notarization
have been slow and are frequently failing.  Attempt to fix this by
splitting up the files we send to ESRP on macOS into multiple batches to
hopefully put less pressure on the service.

A `signClassicPkgContent` parameter has also been added to allow us to
skip these steps while testing, or when building .NET release branches if
needed.

MicroBuild tooling has also been updated to the latest versions.


pjcollins:
This seems a little better, though we may want to further split up the set of binutils lib files:

image

pjcollins:
Testing this a second time to see if the notarization stage succeeds: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6625844&view=results

pjcollins:
Latest test: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6626622&view=results

PR #7327, opened 2022-09-01
[FabricBot] Apply 'needs-triage' label to new issues.

jpobst:
Today we rely on our templates to apply the `needs-triage` label to new issues.  However, there are several ways issues can be added to our repo that bypasses our templates:

- Using the "blank" template
- Moved from another repository
- Moved from `vsfeedback`

FabricBot can ensure *every* new issue is tagged with `needs-triage`.

Note that the first commit simply allows the FabricBot UI to reformat our existing config file to its desired format.  It does not make any changes.

To see what is actually being added, look at the second commit.

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -1,246 +1,201 @@
-[
-  {
-    "taskType": "trigger",
-    "capabilityId": "IssueResponder",
-    "subCapability": "IssuesOnlyResponder",
-    "version": "1.0",
-    "config": {
-      "conditions": {
-        "operator": "and",
-        "operands": [
+{
+  "version": "1.0",
+  "tasks": [
+    {
+      "taskType": "trigger",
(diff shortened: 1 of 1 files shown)


.github/fabricbot.json
```diff
@@ -546,6 +546,39 @@
...
+          "issues",
+          "project_card"
+        ],
+        "actions": [
+          {
+            "name": "addLabel",
+            "parameters": {
+              "label": "needs-triage"
+            }
+          }
```
jonathanpeppers, line 578: You can't add a label twice, right? So it's fine if issues come in with the label already?

jonathanpeppers (approved):

.github/fabricbot.json
jpobst: I would assume that is true.  You cannot add the label twice in the GitHub UI at least.

PR #7332, opened 2022-09-02
Marshal methods and and dynamic method registration documentation

grendello:
Document the way both the "traditional" (dynamic) method registration and marshal methods registrations work.


grendello:
Hm, the PR will have to wait till the other PRs in the marshal method series are merged. The branch of origin is two levels removed from `main` and thus the PR shows too many changes (it should be just a single file)

PR #7343, opened 2022-09-06
[FabricBot] Add 'possibly-stale' GH label for closing old issues.

jpobst:
Copy https://github.com/dotnet/runtime automation for `backlog-cleanup-candidate` as `possibly-stale`.

When we place the issue label `possibly-stale` on an issue, it will kick off the following workflow:

- A comment will be added to the issue that states:
```
We suspect this issue is stale and no longer relevant. It will be closed if no further activity occurs within 
14 more days. Any new comment (by anyone, not necessarily the author) will undo this process.
```
- If anyone comments on the issue during this timespan, the `possibly-stale` label will be removed and the `need-attention` label will be applied so we can review the comment.
- If no one comments on the issue during this timespan, the issue will be closed with the comment:
```
This issue will now be closed since it had been marked `possibly-stale` but received no further activity in 
the past 14 days. It is still possible to reopen or comment on the issue, but please note that the issue will 
be locked if it remains inactive for another 30 days.
```

### DIFF
--- a/.github/fabricbot.json
+++ b/.github/fabricbot.json
@@ -579,6 +579,185 @@
         ],
         "taskName": "[New Issues] Apply 'needs-triage' to new issues"
       }
+    },
+    {
+      "taskType": "trigger",
+      "capabilityId": "IssueResponder",
+      "subCapability": "IssuesOnlyResponder",
+      "version": "1.0",
+      "config": {
+        "conditions": {
+          "operator": "and",
+          "operands": [
+            {
+              "name": "labelAdded",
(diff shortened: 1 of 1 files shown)


jonathanpeppers (approved):
So the idea is we put `possibly-stale` manually to start this process?

jpobst:
Yes, `dotnet/runtime` also automatically kicks off the process after 5 years of inactivity, but I'm not a huge fan of automatically closing old issues.  So we can start with this and see how it goes.

Issue #7297, opened 2022-08-22
Emit warning for use of `$(AotAssemblies)` in .NET 7


jonathanpeppers:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

.NET 7+

### Description

Context: https://github.com/xamarin/xamarin-android/pull/7295

The way we support `$(AotAssemblies)` requires us to *know* the names of packs in the Mono workload.

Let's drop support for `$(AotAssemblies)` and use the new `$(RunAOTCompilation)` property instead.

We can begin by emitting a warning in .NET 7.

### Steps to Reproduce

1. `dotnet new android`
2. `dotnet build -c Release -p:AotAssemblies=true`
3. Should emit a warning

PR #7344, opened 2022-09-06
[One .NET] add XA1029 for $(AotAssemblies)

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/7297

The way we support `$(AotAssemblies)` requires us to *know* the names
of packs in the Mono workload.

Let's drop support for `$(AotAssemblies)` and use the new
`$(RunAOTCompilation)` property instead. We can begin by emitting a
warning in .NET 7, to potentially remove/error in .NET 8.

### DIFF
--- a/Documentation/building/configuration.md
+++ b/Documentation/building/configuration.md
@@ -87,9 +87,9 @@ Overridable MSBuild properties include:
 
   * `$(AndroidSupportedTargetAotAbis)`: The Android ABIs for which to build the
     Mono AOT compilers. The AOT compilers are required in order to set the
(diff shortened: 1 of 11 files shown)


dellis1972 (changes requested):
There should be some documentation for the AotAssembies property in guides/building-apps/build-properties.md, building/configuration.md and guides/OneDotNet.md which will need updating to use the new property or better still add a deprecated note and link to the new property.

jonathanpeppers:
Thanks, I updated the docs on this.

dellis1972 (approved):

Issue #7315, opened 2022-08-26
Templates need to provide ide.host.json files


jonpryor:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

VS 17.4

### Description

Context: https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-contentview-xaml/.template.config/ide.host.json

The `.template.config/ide.host.json` file is used by Visual Studio to control where in the File > New > New File and File > New > New Project dialogs a given template is shown.

For example, with [File > New > New Project](https://docs.microsoft.com/en-us/visualstudio/ide/create-new-project?view=vs-2022), you can select **All project types** and set the type to `MAUI`.  After doing so, three projects are shown:

  * ".NET MAUI App":
    * https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-mobile/.template.config/ide.host.json
    * https://github.com/dotnet/maui/blob/f2ae851de7c070408c0fc0d85923a6ec075924b7/src/Templates/src/templates/maui-mobile/.template.config/template.json#L8
  * ".NET MAUI Blazor App":
    * https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-blazor/.template.config/ide.host.json
    * https://github.com/dotnet/maui/blob/f2ae851de7c070408c0fc0d85923a6ec075924b7/src/Templates/src/templates/maui-blazor/.template.config/template.json#L8
  * ".NET MAUI Class Library":
    * https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-lib/.template.config/ide.host.json
    * https://github.com/dotnet/maui/blob/f2ae851de7c070408c0fc0d85923a6ec075924b7/src/Templates/src/templates/maui-lib/.template.config/template.json#L8

Similarly, if you have a .NET Maui project, and within the Solution Explorer you right-click the Project > Add > New Item…, in the Add New Item dialog, on the right you can select the node Installed > C# Items > .NET MAUI, which will show on the right e.g.

  * .NET MAUI ContentPage (C#)
    * https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-contentpage-csharp/.template.config/ide.host.json
    * https://github.com/dotnet/maui/blob/f2ae851de7c070408c0fc0d85923a6ec075924b7/src/Templates/src/templates/maui-contentpage-csharp/.template.config/template.json#L8

[...]

Issue #7231, opened 2022-08-04
Android Items Template is missing for .Net Android project


hamiddd1980:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

Android 6, Android 9

### Description

Currently in **.Net Android** project, the template for Android specific items(files) like _Activity,Layout,Menu_ …is missing from menu : **Resource Folder>Add>New Item**.
These items are available in Xamarin.Android , but are not implemented in new **.Net Android**.
Without Android Items Template in **.Net Android**, it’s so difficult to develop Android Apps in VS.
Request:
Please add Android Files Templates (Like Activity,Layout,…) to .Net Android projects like Xamarin.Android.

### Steps to Reproduce

1-Create a new **.Net Android project**
2-Try to add a new **Android item(file)** like Layout,Activity... to the project by right click on project items.

dellis1972:
Hmm, our template pack for .net 6 does include Activity and Layout Item Templates. Namely android-activity and android-layout (see https://github.com/xamarin/xamarin-android/tree/main/src/Microsoft.Android.Templates). 

So I wonder why they do not show up in the IDE. 

@tondat are you aware of any issues with Item Templates not showing up in Visual Studio?

hamiddd1980:
Hi **dellis1972** . is it possible that I add .Net Android templates to Visual Studio manually? (where should I copy those files?)
Thank you.

jonathanpeppers:
You should be able to use `dotnet new android` and then just open the template in Visual Studio.

https://docs.microsoft.com/dotnet/core/tools/dotnet-new

dellis1972:
@jonathanpeppers its the Item Templates they are after, not a new project one. 

I'm not sure how to go about getting those to show up in VS.

jonathanpeppers:
Ah yes, the item template should also work:

https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/OneDotNet.md#dotnet-new

So like:

    dotnet new android-activity --name LoginActivity --namespace MyAndroidApp

hamiddd1980:
but I'm looking for .Net Android specific templates(layout,Activity ...) 
![androidmissing](https://user-images.githubusercontent.com/57061231/183955255-05a66183-8f26-41f5-b3e3-aa2587fcd58a.png)

PR #7349, opened 2022-09-06
[templates] add `ide.host.json` files

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/7315
Fixes: https://github.com/xamarin/xamarin-android/issues/7231

This makes our item templates useable from inside Visual Studio, as it should allow them to show up in a new `Android` category in the "Add New Item" dialog.

### DIFF
--- a/src/Microsoft.Android.Templates/android-activity/.template.config/ide.host.json
+++ b/src/Microsoft.Android.Templates/android-activity/.template.config/ide.host.json
@@ -1,0 +1,6 @@
+{
+    "$schema": "http://json.schemastore.org/ide.host",
+    "appliesTo": "Android",
+    "defaultItemExtension": "cs",
+    "itemHierarchyPaths": [ "Android" ]
+}
--- a/src/Microsoft.Android.Templates/android-layout/.template.config/ide.host.json
+++ b/src/Microsoft.Android.Templates/android-layout/.template.config/ide.host.json
@@ -1,0 +1,6 @@
+{
+    "$schema": "http://json.schemastore.org/ide.host",
+    "appliesTo": "Android",
+    "defaultItemExtension": "xml",
+    "itemHierarchyPaths": [ "Android" ]
+}


jonathanpeppers:
@jonpryor a few questions on this one:

* Do we need to do anything for project templates?
* Do we need to filter on VS 17.3 like MAUI? https://github.com/dotnet/maui/blob/f2ae851de7c070408c0fc0d85923a6ec075924b7/src/Templates/src/templates/maui-mobile/.template.config/ide.host.json#L4-L9

They also require the `Microsoft.VisualStudio.ComponentGroup.Maui.All` component, but that doesn't really make sense? This component contains the Android workload and our templates.

jonathanpeppers:
The answer to all these is "no":

> * Do we need to do anything for project templates?
> * Do we need to filter on VS 17.3 like MAUI? https://github.com/dotnet/maui/blob/f2ae851de7c070408c0fc0d85923a6ec075924b7/src/Templates/src/templates/maui-mobile/.template.config/ide.host.json#L4-L9
> 
> They also require the `Microsoft.VisualStudio.ComponentGroup.Maui.All` component, but that doesn't really make sense? This component contains the Android workload and our templates.

I manually tested by downloading `Microsoft.Android.Templates.33.0.0-ci.pr.gh7349.163.nupkg` and dropped it in:

    C:\Program Files\dotnet\template-packs

Next edit `C:\Program Files\dotnet\sdk-manifests\7.0.100-rc.1\microsoft.net.sdk.android\WorkloadManifest.json` and change the version number:

```json
    "Microsoft.Android.Templates": {
      "kind": "template",
      "version": "33.0.0-ci.pr.gh7349.163"
    }
```

Create a new `net7.0-android` app, and these items show up:

image

Then in a MAUI app:

image

mrward:
Looks mostly fine to me. The layout file seems to be flagged as C#. But that seems to match what is in the template.json file:

https://github.com/xamarin/xamarin-android/blob/main/src/Microsoft.Android.Templates/android-layout/.template.config/template.json#L8

image
image
image
image

jonathanpeppers:
Hmm, even the MAUI "page" templates that are `.xaml` specify C# here:

https://github.com/dotnet/maui/blob/9a323ef559437aebfe27d91db6d7bebd1ba3d82f/src/Templates/src/templates/maui-contentpage-xaml/.template.config/template.json#L11

So we should change this to `XML` instead? Then the text would change here:

image

mrward:
For some reason I added a check to VS Mac that ignored every item template that was not C# language (I think I copied what WebTools was doing but will have to check). So for now I suggest it is left as C#.

hamiddd1980:
Hi.
I have updated my VS to latest version : **Version 17.4.0 Preview 4.0**
but in contrast to **Xamarin.Android**, there are only 2 templates available for .NET Android project.
is **microsoft** team going to add more items to .Net Android Templates?(at least like Xamarin.Android)

xam

PR #7356, opened 2022-09-07
[Xamarin.Android.Build.Tasks] Feature for excluding Kotlin-related files from apps

### LINKED ISSUE #6920: Feature for excluding Kotlin-related files from apps
### Android application type

### Affected platform version

.NET 6

### Description

`dotnet new maui` includes a lot of "weird" Kotlin files like:

image

`DebugProbesKt.bin`, they actually recommend:

```gradle
packagingOptions {
  exclude "DebugProbesKt.bin"
}
```
See: https://github.com/Kotlin/kotlinx.coroutines/issues/2274

The `kotlin` folder above has lot of `*.kotlin_metadata` files that is apparently only used if Kotlin reflection is used at runtime:

https://youtrack.jetbrains.com/issue/KT-9770

## Proposal

Add a new item group to mirror the gradle feature above. I don't think we have an equivalent in Xamarin.Android or .NET 6?

Maybe our MSBuild targets can define something like:

```xml
<ItemGroup>
  <AndroidPackagingOptionsExclude Include="DebugProbesKt.bin" />
  <AndroidPackagingOptionsExclude Include="**/*.kotlin_*" />
</ItemGroup>
```
_NOTE: I think we'd might have to escape the `*` with url encoding._

In the rare case where a customer needs some of these files, they could clear them in their `.csproj` as needed:

```xml
<ItemGroup>
  <AndroidPackagingOptionsExclude Remove="@(AndroidPackagingOptionsExclude)" />
</ItemGroup>
```

### Steps to Reproduce

1. `dotnet new maui`
2. `dotnet build -c Release -f net6.0-android`

Then look at the `.apk` file.

_No response_

### Relevant log output

_No response_

dellis1972:
Fixes #6920

Add a new ItemGroup `AndroidPackagingOptionsExclude` to the common targets. This ItemGroup will contain a set of RegEx compatible patterns which will be used to exclude certain files from the final apk/aab.

The patterns will need to be Url encoded so that MSBuild does not try to interpret them.

The main purpose of this is to allow us to exclude certain Kotlin meta-data files from the final package. These files seem to be excluded by gradle so we should do the same.

### DIFF
--- a/Documentation/guides/building-apps/build-items.md
+++ b/Documentation/guides/building-apps/build-items.md
@@ -222,7 +222,7 @@ This is so MSBuild does not try to interpret them as actual file wildcards.
 NOTE: `*`, `?` and `.` will be replaced in the `BuildApk` task with the
 appropriate RegEx expressions.
 
-Added in Xamarin.Android 13.0.
+Added in Xamarin.Android 13.1 and .NET 7.
 
 ## AndroidResource
 
--- a/Documentation/guides/building-apps/build-items.md
+++ b/Documentation/guides/building-apps/build-items.md
@@ -206,7 +206,7 @@ used to specify the ABI that the library targets. Thus, if you add
 
 ## AndroidPackagingOptionsExclude
 
-A set if RegEx compatible items which will allow for items to be
+A set of file glob compatible items which will allow for items to be
 excluded from the final package. The default values are as follows
 
 ```
--- a/Documentation/guides/building-apps/build-items.md
+++ b/Documentation/guides/building-apps/build-items.md
@@ -204,6 +204,26 @@ used to specify the ABI that the library targets. Thus, if you add
 </ItemGroup>
 ```
 
+## AndroidPackagingOptionsExclude
+
+A set if RegEx compatible items which will allow for items to be
+excluded from the final package. The default values are as follows
+
+```
+<ItemGroup>
+	<AndroidPackagingOptionsExclude Include="DebugProbesKt.bin" />
+	<AndroidPackagingOptionsExclude Include="%2A.kotlin_%2A" />
+</ItemGroup>
+```
+
+NOTE: The Items MUST use URL encoding for characters like `*`. This
+is so MSBuild does not try to interpret them as file wildcards.
+
+NOTE: `*` will be replaced in the `BuildApk` task with the RegEx
+`(.)` to match all characters.
+
+Added in Xamarin.Android XX.X.
+
 ## AndroidResource
 
 All files with an *AndroidResource* build action are compiled into
--- a/Documentation/guides/building-apps/build-items.md
+++ b/Documentation/guides/building-apps/build-items.md
@@ -212,7 +212,7 @@ excluded from the final package. The default values are as follows
 ```
 <ItemGroup>
 	<AndroidPackagingOptionsExclude Include="DebugProbesKt.bin" />
-	<AndroidPackagingOptionsExclude Include="$([MSBuild]::Escape('*.kotlin_*'))" />
+	<AndroidPackagingOptionsExclude Include="$([MSBuild]::Escape('*.kotlin_*')" />
 </ItemGroup>
 ```
 Items can use file blob characters for wildcards such as `*` and `?`.
--- a/Documentation/guides/building-apps/build-items.md
+++ b/Documentation/guides/building-apps/build-items.md
@@ -212,15 +212,15 @@ excluded from the final package. The default values are as follows
 ```
 <ItemGroup>
 	<AndroidPackagingOptionsExclude Include="DebugProbesKt.bin" />
-	<AndroidPackagingOptionsExclude Include="%2A.kotlin_%2A" />
+	<AndroidPackagingOptionsExclude Include="$([MSBuild]::Escape('*.kotlin_*'))" />
 </ItemGroup>
 ```
+Items can use file blob characters for wildcards such as `*` and `?`.
+However these Items MUST use URL encoding or '$([MSBuild]::Escape(''))'.
+This is so MSBuild does not try to interpret them as actual file wildcards.
 
-NOTE: The Items MUST use URL encoding for characters like `*`. This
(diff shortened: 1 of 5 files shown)


dellis1972:
We will need this too https://github.com/xamarin/monodroid/pull/1265

Documentation/guides/building-apps/build-items.md
```diff
@@ -204,6 +204,26 @@ used to specify the ABI that the library targets. Thus, if you add
...
 
+## AndroidPackagingOptionsExclude
+
+A set if RegEx compatible items which will allow for items to be
+excluded from the final package. The default values are as follows
+
+```
+<ItemGroup>
+	<AndroidPackagingOptionsExclude Include="DebugProbesKt.bin" />
+	<AndroidPackagingOptionsExclude Include="%2A.kotlin_%2A" />
```
jonathanpeppers, line 215: Should we update the doc, to recommend customers use `[MSBuild]::Escape()`?

jonathanpeppers (approved):
Otherwise, looks good if CI is green.

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
```diff
@@ -293,6 +304,10 @@ public override bool RunTask ()
 
 			existingEntries.Clear ();
 
+			foreach (var pattern in ExcludeFiles ?? Array.Empty<string> ()) {
+				excludePatterns.Add (new Regex (pattern.Replace ("*", "(.)"), RegexOptions.IgnoreCase | RegexOptions.Compiled));
```
jonpryor, line 308: I don't understand this `.Replace()`?  Why replace `*` with `(.)`?

Documentation/guides/building-apps/build-items.md
```diff
@@ -204,6 +204,26 @@ used to specify the ABI that the library targets. Thus, if you add
 </ItemGroup>
 ```
 
+## AndroidPackagingOptionsExclude
+
+A set if RegEx compatible items which will allow for items to be
```
jonpryor, line 209: I don't think that the description matches the usage.  The example usage is `*.kotlin_*`; that is *not* a regular expression.  That's a *file glob*, though a more "consistent" description may be "search pattern", a'la [`Directory.EnumerateFiles(path, searchPattern, searchOptions)`](https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.enumeratefiles?view=net-6.0#system-io-directory-enumeratefiles(system-string-system-string-system-io-searchoption)).

src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
jonpryor: Realizing that `@(AndroidPackagingOptionsExclude)` contains *file globs*, it makes sense that `*` needs to be replaced, but replacing with `(.)` still doesn't make sense to me.  Wouldn't `.*` make more sense?

On a similar angle, the documentation for "search pattern" (see above `Directory.EnumerateFiles()` doc link) says:

> parameter can contain a combination of valid literal path and wildcard (`*` and `?`) characters, but it doesn't support regular expressions.

This "file glob to Regex" conversion doesn't handle `?`.

We should probably have something like:

```csharp
static Regex GlobToRegex (string glob)
{
    var r = new StringBuilder (glob.Length);
    foreach (var c in glob) {
        switch (c) {
        case '*': r.Append (".*");  break;
        case '?': r.Append (".");   break;
        case '.': r.Append (@"\."); break;
        default:  r.Append (c);
        }
    }
    return new Regex (r.ToString ());
}
```

Documentation/guides/building-apps/build-items.md
```diff
@@ -204,6 +204,26 @@ used to specify the ABI that the library targets. Thus, if you add
...
+</ItemGroup>
+```
+Items can use file blob characters for wildcards such as `*` and `?`.
+However these Items MUST use URL encoding or '$([MSBuild]::Escape(''))'.
+This is so MSBuild does not try to interpret them as actual file wildcards.
+
+NOTE: `*`, `?` and `.` will be replaced in the `BuildApk` task with the
+appropriate RegEx expressions.
+
+Added in Xamarin.Android 13.0.
```
jonathanpeppers, line 225: Should this one say:
```suggestion
Added in Xamarin.Android 13.1 and .NET 7.
```
main is this right now:
https://github.com/xamarin/xamarin-android/blob/619420ae1d77d0ff3ec1b59afd156bfdd63c5c26/Directory.Build.props#L33

jonathanpeppers (approved):

jonpryor:
```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/6920

Certainly NuGet packages pull in Kotlin-related artifacts which
aren't required at runtime.  These artifacts contribute to `.apk`
package size.

The [recommendation for native Android developers][0] is to
explicitly exclude these artifacts by using `packagingOptions`:

	packagingOptions {
	  exclude 'DebugProbesKt.bin'
	}

Xamarin.Android and .NET SDK for Android developers don't use Gradle,
so the above snippet is not useful.

Add support for a new `@(AndroidPackagingOptionsExclude)` item group.
This contains a "search pattern" a'la the `searchPattern` parameter
of [`Directory.EnumerateFiles(path, searchPattern)`][1], in which:

  * `*` matches 0 or more characters
  * `?` matches 1 character.

The default items within `@(AndroidPackagingOptionsExclude)` are:

  * `DebugProbesKt.bin`
  * `*.kotlin_*`

Files which match the search patterns within
`@(AndroidPackagingOptionsExclude)` are *excluded* from `.aab` and
`.apk` files, reducing app size.

[0]: https://github.com/Kotlin/kotlinx.coroutines/issues/2274
[1]: https://docs.microsoft.com/en-us/dotnet/api/system.io.directory.enumeratefiles?view=net-6.0#system-io-directory-enumeratefiles(system-string-system-string)
```

PR #7363, opened 2022-09-08
[tests] export $DOTNET_gcServer as 0

jonathanpeppers:
Context: https://github.com/dotnet/linker/issues/3012#issuecomment-1239958027

See if disabling the "server GC" in favor of the "workstation GC" helps a failing test.

### DIFF
--- a/build-tools/automation/yaml-templates/variables.yaml
+++ b/build-tools/automation/yaml-templates/variables.yaml
@@ -38,5 +38,8 @@ variables:
 # Workaround: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/1585820
 - name: _WriteTelemetryProperties
   value: false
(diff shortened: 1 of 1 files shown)


jonpryor:
On the plus side, `BuildAMassiveApp` passed everywhere.

On the minus side, `BuildAMassiveApp` only fails inconsistently; looking at current [xamarin-android/main builds](https://devdiv.visualstudio.com/DevDiv/_build?definitionId=11410&_a=summary), Windows only failed 4 of the past 15 builds in which the macOS build succeeded.  It only fails ~1/3 of the time, if that. :-/

It will thus be difficult to tell if this *actually* helps.

jonathanpeppers:
It seems like that test passed on 3 runs. Let's maybe try one more?

pjcollins (approved):

PR #7381, opened 2022-09-15
[build] support for multiple API levels in .NET 7

jonathanpeppers:
This ports 2c63936 to xamarin-android/main.

It uses `%(AndroidApiInfo.DotNetSupported)` for deciding what API levels to build. We will need this support when API 34 is available, and we want to bring it to .NET 7.

Bringing the changes here now shouldn't really have any effect, but the changes should still *build*.

There is one test for `net6.0-android31` through `net6.0-android33`. I left that in, so we can eventually move it to `net7.0-android33` and `net7.0-android34`.


jonathanpeppers:
This should no longer be needed.

Issue #7147, opened 2022-07-04
Support "app icons" in the csproj


mattleibow:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

All

### Description

Right now we are almost there with resizetizer (icon generation) for mobile apps. We have tasks that generate all the drawables and add it to the app, but the final step the user has to go edit the manifest and link the drawables/mipmaps to the xml file.

It would be great to be able to get resizetizer to output some build item/property and have the android build insert the values into the manifest.

@jonathanpeppers mentioned that editing the manifest from a build task may break incremental builds and/or fast deploy processes.

### Steps to Reproduce

1. File | New | Maui
2. Replace the icon file with one with a new name (or rename)
3. Fix the csproj with the new name
4. Observe the build log shows the new images (and see in the obj dir)
5. Observe the build fails because the manifest also needs updating

### Did you find any workaround?

Editing the manifest, but this is the issue.

jonathanpeppers:
@mattleibow thinks we can drop the "high priority" label.

If the iOS team, has the time to get this in -- we should do Android at the same time.

PR #7411, opened 2022-09-27
[One .NET] add `$(ApplicationIcon)` property

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/7147

This adds `$(ApplicationIcon)`, so it can be set to a value such as `@mipmap/icon` or `@drawable/icon` instead of putting this value in `AndroidManifest.xml`.

I purposely did not add this to the `android` template, as I don't know what IDE impact it will have. However, MAUI projects already have custom MSBuild targets for `@(MauiIcon)`, so they can start leveraging this property so the value can be set in one place. Currently if you rename `@(MauiIcon)`, you have to also update the
`AndroidManifest.xml` and `Info.plist`.

One "gotcha" is that `$(ApplicationIcon)` is an existing property used for `<Csc/>` for `Win32Icon="$(ApplicationIcon)"`. I cleared the original property when `$(GenerateApplicationManifest)` is `true` to avoid C# compiler errors.

TODO:

* Do we need `android:roundIcon`?
* Update documentation


jonathanpeppers:
@mattleibow #7147 doesn't say what MSBuild properties are needed. Do you need one for `android:roundIcon`?

Documentation/guides/building-apps/build-properties.md
```diff
@@ -1472,6 +1472,19 @@ string.
...
+An MSBuild property that can be used as an alternative to
+`/application/@android:icon` in the [`AndroidManifest.xml`](~/android/platform/android-manifest.md)
+file. To opt into this feature you must also enable
+`<GenerateApplicationManifest>true</GenerateApplicationManifest>`.
+This will be the default going forward in .NET 6.
+
+See the [Android documentation][application-element] for further details
+about the requirements for `/manifest/application/@android:icon`.
+
+Support for this property was added in Xamarin.Android 12.2 and .NET 7.
```
jonpryor, line 1486: 1. Is Classic getting support for this property?
2. If it is, it would be added in 13.2…which may or may not exist?

Probably better to just omit Classic entirely at this point.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/SingleProjectTest.cs
```diff
@@ -52,6 +54,8 @@ public void AndroidManifestProperties (string versionName, string versionCode, s
 				proj.SetProperty ("GenerateApplicationManifest", "true");
 			}
 			proj.SetProperty ("ApplicationId", packageName);
+			proj.SetProperty ("ApplicationIcon", applicationIcon);
```
jonpryor, line 57: The implied semantics here are *weird* to me.

The primary idea is (1) use in MAUI Single Project, and (2) the implied usage would be a `.csproj` with:

```xml
<PropertyGroup>
  <AndroidIcon>@drawable/icon</AndroidIcon>
  <AndroidRoundIcon>@drawable/icon</AndroidRoundIcon>
</PropertyGroup>
```

How is this going to work on iOS?  WinUI?  Tizen?  Shouldn't these properties instead refer to *files*?

```xml
<PropertyGroup>
  <AndroidIcon>Resources\Icon.svg</AndroidIcon>
  <AndroidRoundIcon>Resources\RoundIcon.svg</AndroidRoundIcon>
</PropertyGroup>
```

Documentation/guides/OneDotNetSingleProject.md
```diff
@@ -29,6 +29,11 @@ Xamarin.Android and Xamarin.iOS/Mac SDKs:
 * `$(ApplicationTitle)` maps to `/application/@android:title` or
   `CFBundleDisplayName`
 
+Added in Xamarin.Android 12.2 / .NET 7:
```
jonpryor, line 32: As per another comment, ix-nay on the "Xamarin.Android"-ay.  And it would be 13.2, not 12.2.  Were it to exist.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/SingleProjectTest.cs
dellis1972: This does bring up the qusetion around `@mipmap` vs `@drawable` as both are valid, which one will Maui generate?

jonathanpeppers: The MAUI example above is:
```xml
<MauiIcon Include="Resources\AppIcon\appicon.svg" ForegroundFile="Resources\AppIcon\appiconfg.svg" Color="#512BD4" />
```
The "Resizetizer" MSBuild logic currently creates mipmaps:

image

There is some missing design here though...

Customers in a plain .NET 6 Android project can't set a file path, because an app icon is multiple files:

    Resources/mipmap-hdpi/icon.png
    Resources/mipmap-mdpi/icon.png
    Resources/mipmap-xhdpi/icon.png
    .etc.

So then how useful is `ApplicationIcon=@mipmap/icon`? @mattleibow are the properties above sufficient for what MAUI needs?

dellis1972: Should we be using just the resource name instead? and have a task which will look for the right files in @mipmap then @drawable ? e.g `ApplicationIcon=icon`.  So it will then just look for a resource called `icon.png`...

Maybe get other platforms to do the same?

If we have a `@xxx` at the start, maybe skip the searchng part , at least for android.

jonpryor: This thread doesn't quite answer my question, and I imagine we need @mattleibow's input, so, to rephrase:

Given the [maui-mobile](https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj) template, what changes do we want to have present?  What makes sense in a multi-UI world?

Having `$(ApplicationIcon)=@drawable/icon` does not make sense to me, as I do not understand how that would make sense to iOS or Tizen or WinUI or…

Having `$(ApplicationIcon)` match a `@(MauiIcon)` value *does* make sense to me, as `@(MauiIcon)` applies to all UI, but I am still uncertain about that.  Can there be more than one `@(MauiIcon)`?  Can `%(MauiIcon.Include)` be duplicated?

Perhaps the "better" idea is, instead of an `$(ApplicationIcon)` property, we have `%(MauiIcon.ApplicationIcon)` metadata?

```xml
<ItemGroup>
  <MauiIcon
      Include="Resources\AppIcon\appicon.svg"
      ForegroundFile="Resources\AppIcon\appiconfg.svg"
      Color="#512BD4"
      ApplicationIcon="True"
  />
</ItemGroup>
```

…though that would permit "multiple" `%(MauiIcon.ApplicationIcon)`=True values, which is likely "Bad".

I'm just fundamentally not understanding the high-level usage and expectations.  https://github.com/dotnet/maui/pull/8063 doesn't help me, nor does https://github.com/xamarin/xamarin-android/issues/7147.

Then there's the [`@(MauiIcon)` documentation](https://learn.microsoft.com/en-us/dotnet/maui/user-interface/images/app-icons?tabs=android), which feels like whoever came up with this idea has no idea about MSBuild conventions.  If there can be only one, it *should not* have been an Item!  It should have been a Property!  (Except then you couldn't easily add metadata to the property, so that's weird…)  It should also *warn* if there is more than one and some will be ignored!  Just…bizarre.

Given the existence of `@(MauiIcon)`, does a cross-UI `$(ApplicationIcon)` need to exist?  Or should it instead be `$(AndroidApplicationIcon)`, which is derived from the `@(MauiIcon)` value?

dellis1972: I just had a thought. Perhaps we need the `AndroidApplicationIcon` property (and iOS needs its one to) which we will use if present to set the manifest. 

Then Maui could add a task.property which is cross platform via meta data or a new property `MauiApplicationIcon` which will then generate the right file for the platform and then set the new `AndroidApplicationIcon` in android. It will do the same for other platforms.

PR #7412, opened 2022-09-27
Use LLVM toolchain compressed with UPX

grendello:
On Windows our distribution package is much bigger than on macOS because the LLVM binaries each include the entire LLVM framework, no shared libraries are used.  For that reason, the distribution size on Windows pushes 100MB.  In order to make the footprint smaller, we now compress the Windows binaries with UPX.


jonathanpeppers (approved):
I reran some failed test lanes, because we didn't get a full run of the on-device tests.

PR #7427, opened 2022-09-29
[Mono.Android] Improve generated API docs diff

pjcollins:
Updates the `<UpdateExternalDocumentation/>` target to run the mdoc `export-msxdoc` command for validation, and to include new files in the diff that it produces.

### DIFF
--- a/src/Mono.Android/Mono.Android.targets
+++ b/src/Mono.Android/Mono.Android.targets
@@ -292,7 +292,7 @@
     <DocsApiLevel Condition=" '$(DocsApiLevel)' == '' ">31</DocsApiLevel>
     <DocsPlatformId Condition=" '$(DocsPlatformId)' == '' ">31</DocsPlatformId>
     <DocsFxVersion Condition=" '$(DocsFxVersion)' == '' ">v12.0</DocsFxVersion>
-    <_Binlog>$(MSBuildThisFileDirectory)../../bin/Build$(Configuration)/UpdateApiDocs-$([System.DateTime]::Now.ToString ("yyyyMMddTHHmmss")).binlog</_Binlog>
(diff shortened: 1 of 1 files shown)


pjcollins:
Testing with https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6761593&view=results

src/Mono.Android/Mono.Android.targets
```diff
@@ -365,14 +365,20 @@
...
-    </PropertyGroup>
     <Exec
-        Command="git diff --output=&quot;$(_DiffFile)&quot;"
+        Command="git add ."
+        WorkingDirectory="$(XamarinAndroidSourcePath)external/android-api-docs"
+    />
+    <Exec
+        Command="git diff --cached --output=&quot;$(_LogPrefix).diff&quot;"
         WorkingDirectory="$(XamarinAndroidSourcePath)external/android-api-docs"
     />
```
jonathanpeppers, line 383: In other places we use:
https://github.com/xamarin/xamarin-android/blob/d8f8486718072b040d27a0955a5778e2618116f3/build-tools/scripts/XAVersionInfo.targets#L43-L44

It's probably worth using these variables in case we ever change them repo-wide. Maybe in some cases, someone would override to `/opt/homebrew/bin/git`?

jonathanpeppers: You might also be able to use our `<Git/>` task instead of `<Exec/>`, but only do that if it is easier:
https://github.com/xamarin/xamarin-android/blob/d8f8486718072b040d27a0955a5778e2618116f3/build-tools/xa-prep-tasks/Xamarin.Android.BuildTools.PrepTasks/Git.cs#L15

pjcollins: Thanks, should be easy enough to use `<Git/>` instead as Mono.Android.targets already references a task from `$(PrepTasksAssembly)`.

jonathanpeppers (approved):

PR #7448, opened 2022-10-10
[Xamarin.Android.Build.Tasks] Add XA1031 error for `AndroidHttpClientHandlerType`

### LINKED ISSUE #7326: Emit a warning if AndroidHttpClientHandlerType is invalid
### Android application type

### Affected platform version

.NET 6.0+

### Description

See also: PR #7214.

If the `$(AndroidHttpClientHandlerType)` property is set to an "invalid" value, which can be an "unresolvable" type (`Type.GetType()` returns `null`) *or* a type which has `HttpClientHandler` as a base class, then the build should emit a warning.

### Steps to Reproduce

-

PR #7214 has a runtime workaround.  However, the warning message it produces is a *runtime* warning message.

A *build-time* warning should also be generated.

### Relevant log output

_No response_

dellis1972:
Fixes #7326

We should check the value of the `AndroidHttpClientHandlerType` property during the build process. This property can have a number of different values depending on what version of Xamarin.Android is being used.

Under Classic a valid value is `Xamarin.Android.Net.AndroidClientHandler`, however under .NET 6+ the valid values are

- `Xamarin.Android.Net.AndroidMessageHandler`.
- `System.Net.Http.SocketsHttpHandler, System.Net.Http`.

We should raise an error if the user is using an invalid value.


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/EnvironmentContentTests.cs
```diff
@@ -194,6 +194,20 @@ public void CheckBuildIdIsUnique ([Values ("apk", "aab")] string packageFormat)
...
 		}
 
+		[Test]
+		public void CheckForInvalidHttpClientHandlerType ()
+		{
+			var proj = new XamarinAndroidApplicationProject () {
+				IsRelease = true,
+			};
+			using (var b = CreateApkBuilder ()) {
+				b.ThrowOnBuildFailure = false;
```
jonathanpeppers, line 204: Somehow this flag didn't work?
```
Xamarin.ProjectTools.FailedBuildException : Build failure: UnnamedProject.csprojBuild log recorded at /Users/runner/work/1/a/TestRelease/10-10_10.51.36/temp/CheckHttpClientHandlerType/build.log
```

Documentation/guides/messages/xa1031.md
```diff
@@ -0,0 +1,20 @@
...
+
+```
+The 'AndroidHttpClientHandlerType' has an invalid value of 'xxx' please check your project settings.
+```
+
+## Solution
+
+Open your project settings in Visual Studio and check that the
+'AndroidHttpClientHandlerType' has a valid value. Alternatively
+you can edit your csproj directly and change the 'AndroidHttpClientHandlerType'.
```
jonathanpeppers, line 18: I don't think we have this setting in the VS project options pages for .NET 6 projects. The Mono team wanted to "hide" it so only more advanced users would be using it.

Maybe we just reword this to edit .xml the project file?

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -470,6 +470,11 @@ In this message, the term "binding" means a piece of generated code that makes i
     <value>The 'RunAOTCompilation' MSBuild property is only supported when trimming is enabled. Edit the project file in a text editor to set 'PublishTrimmed' to 'true' for this build configuration.</value>
     <comment>The following are literal names and should not be translated: 'RunAOTCompilation', 'PublishTrimmed'</comment>
   </data>
+  <data name="XA1031" xml:space="preserve">
+    <value>The 'AndroidHttpClientHandlerType' has an invalid value of '{0}' please check your project settings.</value>
```
jonathanpeppers, line 474: Maybe this one should say something like:
```suggestion
    <value>The 'AndroidHttpClientHandlerType' has an invalid value of '{0}' please change the value in your project file to one of X, Y, or remove the property completely.</value>
```

jonathanpeppers (approved):
The actual code changes look good. But maybe we should adjust some of the user-facing messages?

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -488,6 +488,22 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
...
+    <ValidAndroidHttpClientHandlerTypes Condition="'$(UsingAndroidNETSdk)' != 'true'">$(ValidAndroidHttpClientHandlerTypes);System.Net.Http.MonoWebRequestHandler, System.Net.Http</ValidAndroidHttpClientHandlerTypes>
+    <ValidAndroidHttpClientHandlerTypes Condition="'$(UsingAndroidNETSdk)' != 'true'">$(ValidAndroidHttpClientHandlerTypes);System.Net.Http.HttpClientHandler, System.Net.Http</ValidAndroidHttpClientHandlerTypes>
+    <ValidAndroidHttpClientHandlerTypes Condition="'$(UsingAndroidNETSdk)' == 'true'">Xamarin.Android.Net.AndroidMessageHandler</ValidAndroidHttpClientHandlerTypes>
+    <ValidAndroidHttpClientHandlerTypes Condition="'$(UsingAndroidNETSdk)' == 'true'">$(ValidAndroidHttpClientHandlerTypes);System.Net.Http.SocketsHttpHandler, System.Net.Http</ValidAndroidHttpClientHandlerTypes>
+  </PropertyGroup>
+  <AndroidError Code="XA1031"
+      ResourceName="XA1031"
+      FormatArguments="$(AndroidHttpClientHandlerType)"
+      Condition=" !($(ValidAndroidHttpClientHandlerTypes.Contains ('$(AndroidHttpClientHandlerType)')))"
```
jonpryor, line 503: Unfortunately, this check is too strict.  The [`$(AndroidHttpClientHandlerType)` MSBuild property](https://github.com/xamarin/xamarin-android/blob/main/Documentation/guides/building-apps/build-properties.md#androidhttpclienthandlertype) *allows* for arbitrary types:

> Controls the default `System.Net.Http.HttpMessageHandler` implementation which will be used by the `System.Net.Http.HttpClient` default constructor. The value is an assembly-qualified type name of an `HttpMessageHandler` subclass, suitable for use with [System.Type.GetType(string)](https://github.com/xamarin/xamarin-android/blob/main/dotnet/api/system.type.gettype#System_Type_GetType_System_String_)

The *intent* is that if someone *wishes* to provide their own `System.Net.Http.HttpMessageHandler` subclass, *they can do so* by:

 1. Setting `$(AndroidHttpClientHandlerType)` to `Their.HttpMessageHandler.Subclass, Assembly-Name`, and
 2. Setting `$(UseNativeHttpHandler)`=True.

*This* implementation does not allow that; it instead *requires* that *only* `Xamarin.Android.Net.AndroidMessageHandler` *or* `System.Net.Http.SocketsHttpHandler, System.Net.Http` be specified.

What we *actually* need to do is add a new Task/update an existing Task which is able to walk the inheritance chain of the type specified by `$(AndroidHttpClientHandlerType)` and verify that it fulfills our runtime requirements: https://github.com/xamarin/xamarin-android/blob/91669a4e489104668fdd07382e40cf070aa559ac/src/Mono.Android/Android.Runtime/AndroidEnvironment.cs#L356-L372

jonpryor:
WIP commit message:

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/7326

Commit d3e87674 added a requirement that, under .NET 6+, the
`$(AndroidHttpClientHandlerType)` MUST inherit from
[`System.Net.Http.HttpMessageHandler`][0].  This was unfortunately a
change from Classic, which allowed/required that
[`System.Net.Http.HttpClientHandler`][1] be the base class.
Allowing `HttpClientHandler` to be in the inheritance hierarchy for
`$(AndroidHttpClientHandlerType)` would result in a
`NullReferenceException` at runtime under .NET 6.

Commit d3e87674 contained a TODO:

> TODO: [Emit a build-time warning][2] if
> `$(AndroidHttpClientHandlerType)` is set to an invalid value

Implement this TODO: add a build-time check that verifies that
`$(AndroidHttpClientHandlerType)` is a valid type for the target,
e.g. is an `HttpMessageHandler` subclass on .NET 6.

If `$(AndroidHttpClientHandlerType)` is not a valid type, then an
XA1031 error is raised:

	error XA1031: The 'AndroidHttpClientHandlerType' has an invalid value of '{0}' please check your project settings.

[0]: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpmessagehandler?view=net-6.0
[1]: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler?view=net-6.0
[2]: https://github.com/xamarin/xamarin-android/issues/7326
```

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
jonpryor: Error messages should be actionable.  *How* is the value invalid?

Suggested alternative:

```
The 'AndroidHttpClientHandlerType' property value '{0}' must derive from '{1}'.
Please change the value to an assembly-qualifed type name which inherits from '{1}' or remove the property completely.
```

src/Xamarin.Android.Build.Tasks/Tasks/CheckClientHandlerType.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.IO;
+using Microsoft.Android.Build.Tasks;
+using Java.Interop.Tools.Cecil;
+using Mono.Cecil;
+
+namespace Xamarin.Android.Tasks
+{
+	public class CheckClientHandlerType : AndroidTask
```
jonpryor, line 13: My concern with a task such as this is that it adds yet another place where we load and process "lots" of assemblies.

Would this check be better done in an existing task such as ($DEITY forbid) `<GenerateJavaStubs/>` or `<LinkAssemblies/>` or `<LinkAssembliesNoShrink/>`?  *Or*, find a way to *share* the `DirectoryAssemblyResolver` instance *across* these tasks so we reduce how often we need to load them?

dellis1972: I was trying to avoid adding this to `GenerateJavaStubs`. However we do not process `lots` of assemblies, we find the one we want and then walk up the dependency tree. 
Putting it in the Linker Step means it will only show up when packaging... which is not ideal.

jonathanpeppers: Could we narrow the scope of this one? And just error if the value is `System.Net.Http.HttpMessageHandler`?

I think that is the specific setting that will crash your app in .NET 6+.

jonpryor: @jonathanpeppers [asked](https://github.com/xamarin/xamarin-android/pull/7448#discussion_r1001101602):

> Could we narrow the scope of this one?

That's a performant thought.  It also won't crash anymore, as of d3e87674187fca4ce136f5c4b460c488044e1bec.

Part of the hope was that the runtime check in d3e87674187fca4ce136f5c4b460c488044e1bec wouldn't be the *only* way to tell that the property has an invalid value, but maybe I've been overthinking it, and the existing runtime check alongside a "don't use a known bad value that is likely to be present because of project upgrades" is sufficient.

dellis1972: So I did some testing and this task only added 11ms to the build. This is because it selectively targets and loads a specific assembly. I'd rather not add this code to GenerateJavaStubs as it is big enough already.

src/Xamarin.Android.Build.Tasks/Tasks/CheckClientHandlerType.cs
```diff
@@ -0,0 +1,95 @@
...
+			}
+			if (foundAssembly == null) {
+				Log.LogCodedError ("XA1033", Xamarin.Android.Tasks.Properties.Resources.XA1033, assembly);
+				return !Log.HasLoggedErrors;
+			}
+			// find the type.
+			var readerParameters = new ReaderParameters {
+				ReadSymbols = false,
+			};
+			using (var resolver = new DirectoryAssemblyResolver (this.CreateTaskLogger (), loadDebugSymbols: false, loadReaderParameters: readerParameters)) {
```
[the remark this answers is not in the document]
dellis1972: I can take a look. We need to be able to walk up the dependency types. Can metadata do that?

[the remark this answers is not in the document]
jonathanpeppers: If this is looking up a type by name, that should be doable. You can iterate over all the types, and check namespace then type.

Maybe similar to:

https://github.com/jonathanpeppers/Benchmarks/blob/ebdd33b77c502328da5994127b34fc564f4094e8/Benchmarks/Cecil.cs#L105-L107

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
```diff
@@ -95,6 +95,7 @@
     <HttpActivityPropagationSupport Condition="'$(HttpActivityPropagationSupport)' == ''">false</HttpActivityPropagationSupport>
     <InvariantGlobalization Condition="'$(InvariantGlobalization)' == ''">false</InvariantGlobalization>
     <StartupHookSupport Condition="'$(StartupHookSupport)' == ''">false</StartupHookSupport>
+    <UseNativeHttpHandler Condition=" $(AndroidHttpClientHandlerType.Contains ('System.Net.Http.SocketsHttpHandler')) And '$(UseNativeHttpHandler)' == '' ">false</UseNativeHttpHandler>
```
jonpryor, line 98: @dellis1972 : this feels like something that should be called out in the commit message.  What's the rationale here?

src/Xamarin.Android.Build.Tasks/Tasks/CheckClientHandlerType.cs
```diff
@@ -0,0 +1,95 @@
...
+
+				var assemblyDefinition = resolver.GetAssembly (Path.GetFullPath (foundAssembly.ItemSpec));
+				TypeDefinition handlerType = null;
+				foreach (var model in assemblyDefinition.Modules) {
+					handlerType = assemblyDefinition.MainModule.GetType (type);
+					if (handlerType != null)
+						break;
+				}
+				if (handlerType == null) {
+					if (IsLinkerEnabled) {
```
jonpryor, line 68: Why would this be "fine"?  If the type can't be found, things are gonna be weird, even if the linker is enabled.  Why only log a Debug message and return *true*?

dellis1972: I'm gonna have to rework this a bit. 

The problem this was trying to solve was in Release mode if a `HttpClient` is NOT in the code, the entire handler type is removed by the link. So building in Release mode in this instance would give you a build error. 

If the user IS using a `HttpClient` then this does not happen as the type is preserved. 

This does not happen in Debug modes, so you get the error only if the type is wrong. 

I think I'll need to run this task earlier in the process... before the linker in order to get consistent behaviour. I'll test that now.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.DefaultProperties.targets
dellis1972: I think this is left over. But we got info from the runtime team that if a user defines `System.Net.Http.SocketsHttpHandler'` as the handler type they MUST use set the `UseNativeHttpHandler` msbuild property. Otherwise it doesn't work.

This is where the conversaton took place https://discord.com/channels/732297728826277939/732297837953679412/1029777522897993748

jonpryor: Should we remove it ("this is left over") or keep it ("extract rationale from Discord")?

src/Xamarin.Android.Build.Tasks/Tasks/CheckClientHandlerType.cs
jonpryor: Looks like this was invalidated, as `IsLinkerEnabled` no longer exists.  Not sure why GitHub doesn't show this comment as being obsolete.

jonpryor: Doubly weird, clicking "Resolve conversation" does nothing. :-/

PR #7451, opened 2022-10-10
Bump to .NET 8

jonathanpeppers:
Changes: https://github.com/dotnet/installer/compare/11f6b8f7129e34d5317f47ccbc110b2b11c05749...d552037815b75e01be05b6e8d269d54413363e7f
Changes: https://github.com/dotnet/linker/compare/5f9bfd94d9c687207872ae03f751ea19704381c0...493ce626f3278b2d0fd883330bf11a64254981bb
Changes: https://github.com/dotnet/runtime/compare/6d10e4c8bcd9f96ccd73748ff827561afa09af57...b8d49801fe03b96d2fead3d97a11dce1e723dd17

A summary of what it took to get this working:

* Run `darc update-dependencies --id 152596`

* Setup .NET 6,7,8 "multi-targeting"

* Setup dotnet6/dotnet7/dotnet8 NuGet feeds

* `$(DotNetTargetFrameworkVersion)` should be 8.0

* Rename API diff for `net8.0`

* Fix `AutoImport.props` for 8.0

* Update `.apkdesc` files

* Update `CheckIncludedAssemblies` test

* Use .NET Core `3.1.x`

Trying to avoid:

    dotnet-install: Failed to locate the latest version in the channel '3.1.417' with 'preview' quality for 'dotnet-sdk', os: 'win', architecture: 'x64'.

* Install .NET 6 for legacy MSBuild tests

* [Mono.Android.Export] fix `NotSupportedException`

Context: https://github.com/mono/mono/blob/242ac50598e4f596e1a42f9dd377701096ea3a74/mcs/class/referencesource/mscorlib/system/reflection/emit/dynamicmethod.cs#L276-L277
Context: https://github.com/dotnet/runtime/commit/e46523032d72c8b0d2b14ce4b14229a8c268a6e5

    System.NotSupportedException: Wrong MethodAttributes or CallingConventions for DynamicMethod. Only public, static, standard supported

* Use correct Mono pack names for `$(AotAssemblies)`

* Fix `DotNetPublish` test

Was failing with:

    Build should be using C:\a_work\1\s\xamarin-android\bin\Release\dotnet\packs\Microsoft.Android.Ref.33\34.0.0-ci.pr.gh7451.52\ref\net7.0\Mono.Android.dll
Expected: True
But was:  False

* Remove a `net7.0` assertion for build warnings

See: https://github.com/dotnet/runtime/issues/77385

* Pass `-SkipNonVersionedFiles` to `dotnet-install` scripts to avoid:

   dotnet --info
   Exited with code: 137


jonpryor:
`prepare-20221010T181316.msbuild.xat-restore.log` contains:

```
ls-jdks.csproj: error NU1102: Unable to find package Microsoft.NETCore.App.Host.osx-x64 with version (= 3.1.31)
ls-jdks.csproj: error NU1102:   - Found 100 version(s) in dotnet-public [ Nearest version: 5.0.0-preview.1.20120.5 ] 
ls-jdks.csproj: error NU1102:   - Found 1 version(s) in dotnet-eng [ Nearest version: 5.0.0-alpha.1.19618.1 ]
```

Looks like it doesn't like `netcoreapp3.1`? https://github.com/xamarin/xamarin-android-tools/blob/0be567a991396bd26dd2a92ee11d84fd294f57c7/tools/ls-jdks/ls-jdks.csproj#L5

jonathanpeppers:
.NET Core 3.1.31 is probably a servicing release that isn't out yet... The feed for it is probably internal (I'll check), but the easiest fix here is probably to update various projects to at least `net6.0`?

jonathanpeppers:
It builds!

Now builds apps is failing with:

    Microsoft.NET.TargetFrameworkInference.targets(112,5): error NETSDK1139: The target platform identifier android was not recognized.
                         
So likely some workload issue so far.

jonathanpeppers:
The past run was pretty much green, I squashed/rebased this to clean it up.

dellis1972 (approved):
This all looks good if the build is green :) 
Can we run the nightly's as well before we merge?

jonathanpeppers:
Here is a nightly run, if I did it right:

https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6872423&view=results

I did branch `refs/pull/7451/head`, and commit `5861cc818d0c9ab968aac1dfbcf3af2f328d42e5`. I rebased this with main this morning.

grendello (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
```diff
@@ -23,11 +23,11 @@ They run in a context of an inner build with a single $(RuntimeIdentifier).
     https://github.com/dotnet/runtime/blob/69711860262e44458bbe276393ea3eb9f7a2192a/src/mono/nuget/Microsoft.NET.Workload.Mono.Toolchain.Manifest/WorkloadManifest.targets.in#L20-L25
   -->
   <ImportGroup Condition=" '$(MonoAOTCompilerTasksAssemblyPath)' == '' and '$(AotAssemblies)' == 'true' ">
-    <Import Project="Sdk.props" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task.net7" />
-    <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.net7.android-x86" />
-    <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.net7.android-x64" />
-    <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.net7.android-arm" />
-    <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.net7.android-arm64" />
+    <Import Project="Sdk.props" Sdk="Microsoft.NET.Runtime.MonoAOTCompiler.Task" />
+    <Import Project="Sdk.props" Sdk="Microsoft.NETCore.App.Runtime.AOT.Cross.android-x86" />
```
jonpryor, line 27: …wait, what?  The dotnet/runtime team didn't rename these packages to contain `net8`?  What are the long-term implications of this?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
```diff
@@ -801,20 +807,25 @@ public void DotNetBuildXamarinForms ([Values (true, false)] bool useInterpreter)
...
 				XABuildConfig.AndroidDefaultTargetDotnetApiLevel,
 			},
+			new object[] {
+				"net8.0",
+				"android",
+				XABuildConfig.AndroidDefaultTargetDotnetApiLevel,
+			},
 
 			new object[] {
-				"net7.0",
```
jonpryor, line 806: Shouldn't we also assert that net7 builds continue to work?  I'm not sure I understand this change.

jonpryor:
PR text says:

> Setup .NET 6,7,8 "multi-targeting"

I don't understand how .NET 7 is installed?  `use-dot-net.yaml` appears to only provision .NET 6 and "latest" (.NET 8).  How's .NET 7 get into the mix?

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.Aot.targets
jonathanpeppers: I think there are some package names that were updated for 7 and not in main yet.

The .NET 8 version here is a couple weeks old, so if it's not fixed in the next bump I will file an issue to get it fixed.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
jonathanpeppers: Yes, the full assertion list tests 6, 7, and 8:

https://github.com/xamarin/xamarin-android/blob/5861cc818d0c9ab968aac1dfbcf3af2f328d42e5/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs#L799-L832

I made the `net7.0` case the same as the `net6.0` one was.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
```diff
@@ -107,5 +89,5 @@
       "Size": 1904
     }
   },
-  "PackageSize": 2730619
+  "PackageSize": 2590859
```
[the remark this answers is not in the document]
jonathanpeppers: I think this is probably this one: https://github.com/dotnet/runtime/pull/74828

We got this change on .NET 7 release branches already, but not main.

jonathanpeppers:
`use-dot-net.yaml` is only needed for being able to build/restore `xaprepare`, which is now `net7.0`. We also added `net6.0` there to fix Mono MSBuild for some of the classic test lanes.

`xaprepare` provisions .NET 8 for the rest of our build & tests.

> Setup .NET 6,7,8 "multi-targeting"

This is actually talking about workload changes. So you will be able to build `net8.0-android`, `net7.0-android`, and `net6.0-android` apps. We can think about dropping 6 in another PR, but it was easy enough to just make it work for now.

pjcollins (approved):

jonathanpeppers:
Notes on `-SkipNonVersionedFiles`:

> When installing multiple .NET SDKs into a single folder with the `dotnet-install.sh/ps1` script, you can hit an issue where `dotnet` no longer runs:
>```
>dotnet --info
>Exited with code: 137
>```
> The problem is the `dotnet` binary can get overwritten by an older .NET, and be completely broken.\
>
> The solution is:
> * Install the newest .NET first, followed by any older versions
> * Pass `-SkipNonVersionedFiles`, so the `dotnet` binary isn't overwritten
>
> This results with the newest .NET SDK, with a side-by-side older .NET SDK installed as well.

jonpryor:
```markdown
Changes: https://github.com/dotnet/installer/compare/11f6b8f7129e34d5317f47ccbc110b2b11c05749...d552037815b75e01be05b6e8d269d54413363e7f
Changes: https://github.com/dotnet/linker/compare/5f9bfd94d9c687207872ae03f751ea19704381c0...493ce626f3278b2d0fd883330bf11a64254981bb
Changes: https://github.com/dotnet/runtime/compare/6d10e4c8bcd9f96ccd73748ff827561afa09af57...b8d49801fe03b96d2fead3d97a11dce1e723dd17

Context: https://github.com/dotnet/runtime/issues/77273
Context: https://github.com/dotnet/runtime/commit/e46523032d72c8b0d2b14ce4b14229a8c268a6e5
Context: https://github.com/dotnet/runtime/issues/77385

Build and run under .NET 8!

.NET 8 is used by default for the default build.

Running unit tests also requires installing .NET 6 and .NET 7 as well.

Configure dotnet6 + dotnet7 + dotnet8 NuGet feeds.

Run `darc update-dependencies --id 152596` to update the
dotnet/installer version in `eng/Version.Details.xml`.
(ID values come from [Maestro][0].)

Update `AutoImport.props` to only evaluate default android items
in .NET 8.

Update `$(DotNetTargetFrameworkVersion)` to 8.0, and
`$(DotNetStableTargetFramework)` to `net7.0`.  (Technically .NET 7
isn't stable *yet*, but it will be soon enough…)

Update `*.apkdesc` files so that unit tests pass.

Update `PackagingTest.CheckIncludedAssemblies()` test to remove
assemblies which are no longer included in .NET 8-based apps.

Update `use-dot-net.yaml` to take a new `quality` parameter, and
install .NET Core 3.1 quality=GA, not 3.1.417 specifically.
Avoids the error:

	dotnet-install: Failed to locate the latest version in the channel '3.1.417' with 'preview' quality for 'dotnet-sdk', os: 'win', architecture: 'x64'.

Update `src/Mono.Android.Export` to pass
`MethodAttributes.Static | MethodAttributes.Public` to the
`DynamicMethod` constructor, not just `MethodAttributes.Static`.
This avoids a `NotSupportedException` under .NET 8;
see also dotnet/runtime@e4652303 and dotnet/runtime#77273:

	System.NotSupportedException: Wrong MethodAttributes or CallingConventions for DynamicMethod. Only public, static, standard supported
	   at System.Reflection.Emit.DynamicMethod.CheckConsistency(MethodAttributes attributes, CallingConventions callingConvention)
	   at System.Reflection.Emit.DynamicMethod.Init(String name, MethodAttributes attributes, CallingConventions callingConvention, Type returnType, Type[] signature, Type owner, Module m, Boolean skipVisibility, Boolean transparentMethod)
	   at System.Reflection.Emit.DynamicMethod..ctor(String name, MethodAttributes attributes, CallingConventions callingConvention, Type returnType, Type[] parameterTypes, Module m, Boolean skipVisibility)
	   at Java.Interop.DynamicCallbackCodeGenerator.GenerateNativeCallbackDelegate()
	   at Java.Interop.DynamicCallbackCodeGenerator.GetCallback()
	   at Java.Interop.DynamicCallbackCodeGenerator.Create(MethodInfo method)
	   at System.Reflection.MethodInvoker.InterpretedInvoke(Object obj, Span`1 args, BindingFlags invokeAttr)

Update `XASdkTests.DotNetPublish()` for .NET 8.  Fixes:

	Build should be using C:\a_work\1\s\xamarin-android\bin\Release\dotnet\packs\Microsoft.Android.Ref.33\34.0.0-ci.pr.gh7451.52\ref\net7.0\Mono.Android.dll
	Expected: True
	But was:  False

TODO @jonathanpeppers: ^^ failure was for which set of DotNetPublish()
parameters?  Wast it trying to ref a net7 assembly for net8?

Remove an assertion of no build warnings under `net7.0`; this isn't
currently possible, due to dotnet/runtime#77385.

Update the `Microsoft.NETCore.App.Runtime.AOT.Cross` SDK package names
when `$(AotAssemblies)`=True; the names do not (yet?) include the
.NET version in the package names.


When installing multiple .NET SDKs into a single folder with the
`dotnet-install.{.sh,ps1}` script -- needed to run the unit tests --
you can hit an issue where `dotnet` no longer runs:

	dotnet --info
	Exited with code: 137

 The problem is that the `dotnet` binary can get overwritten by an
 older .NET, and be completely broken.

 The solution is:

 1. Install the newest .NET first, followed by any older versions

 2. Use `dotnet-install.ps1 -SkipNonVersionedFiles` or
    `dotnet-install.sh --skip-non-versioned-files` so the `dotnet`
    binary isn't overwritten

This results in the newest .NET SDK, with side-by-side older .NET SDKs
installed as well.

Rename `tests/api-compatibility/acceptable-breakages-vReference-net7.0.txt`
to use `-net8.0`, as the default `$(TargetFramework)` for
`Mono.Android.dll` changed.

[0]: https://maestro-prod.westus2.cloudapp.azure.com/3074/https:%2F%2Fgithub.com%2Fdotnet%2Finstaller/latest/graph
```

PR #7454, opened 2022-10-11
Nullability for AndroidMessageHandler.RequestNeedsAuthorization

mstefarov:
My project uses AndroidMessageHandler.  Currently, idiomatic code like this produces a nullability warning:

```cs
if (handler.RequestNeedsAuthorization)
    handler.PreAuthenticationData = handler.RequestedAuthentication[0];
```

This PR adds a [MemberNotNullWhen attribute](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.membernotnullwhenattribute?view=net-6.0) that allows RequestedAuthentication to be used without a null-check after testing RequestNeedsAuthorization.

### DIFF
--- a/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
+++ b/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
@@ -233,6 +234,7 @@ namespace Xamarin.Android.Net
 		/// If <c>true</c> then the server requested authorization and the application must use information
 		/// found in <see cref="RequestedAuthentication"/> to set the value of <see cref="PreAuthenticationData"/>
 		/// </summary>
+		[MemberNotNullWhen(true, nameof(RequestedAuthentication))]
 		public bool RequestNeedsAuthorization {
 			get { return RequestedAuthentication?.Count > 0; }
 		}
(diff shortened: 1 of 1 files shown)


src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
```diff
@@ -233,6 +234,7 @@ public int MaxAutomaticRedirections
 		/// If <c>true</c> then the server requested authorization and the application must use information
 		/// found in <see cref="RequestedAuthentication"/> to set the value of <see cref="PreAuthenticationData"/>
 		/// </summary>
+		[MemberNotNullWhen(true, nameof(RequestedAuthentication))]
```
jonathanpeppers, line 237: This fails the Xamarin.Android build:
```
src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs(237,4): error CS0246: The type or namespace name 'MemberNotNullWhenAttribute' could not be found (are you missing a using directive or an assembly reference?) [src/Mono.Android/Mono.Android.csproj::TargetFramework=monoandroid10]
src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs(237,4): error CS0246: The type or namespace name 'MemberNotNullWhen' could not be found (are you missing a using directive or an assembly reference?) [src/Mono.Android/Mono.Android.csproj::TargetFramework=monoandroid10]
```

Can you change this to:
```suggestion
#if NETCOREAPP
		[MemberNotNullWhen(true, nameof(RequestedAuthentication))]
#endif
```

Thanks for your contribution!

jonathanpeppers (changes requested):

src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs
mstefarov: Thank you for spotting that! I should've known that monoandroid10 and net7.0 are still built from the same branch.

jonathanpeppers:
Thanks for your contribution!

PR #7460, opened 2022-10-13
[Xamarin.Android.Build.Tasks] Fix up `ToJniName` to use cache.

dellis1972:
The ManifestDocument class has a few instances of `ToJniName` that are not using the `cache` overload. Using the `cache` version should be faster, so lets try to use that :).


jonathanpeppers (approved):
Going to just restart a couple test lanes.

PR #7468, opened 2022-10-17
[c++] runtime sources modernization

grendello:
**NOTE: the PR is on hold. It needs careful review to find what causes the single Designer test failure. It will require re-applying the changes piecemeal and running tests for each of those separately. I will do it incrementally OOB with other work**

Modify our C++ runtime sources to move it a bit more towards the "modern" C++ standards, 
taking advantage of the newer C++ 20 standard features wherever feasible and possible.

Additionally, make a lot of changes driven by the results of running `clang-tidy` over
our sources. Not all of the warnings/suggestions from the utility are correct or make
sense. Sometimes the analyzers are too simplistic - for instance when analyzing the
following fragment of code `clang-tidy` will remark that the variable `n` is not
initialized:

```c++
uint32_t n;

#if defined(ANDROID)
  n = 1234;
#else
  n = 5678;
#endif
```

Another (large) set of changes is making all of our classes static, since we have no
"dynamic" runtime state to speak of and classes serve merely the purpose of encapsulation
and logical partitioning of of code.  Making them static removes the need to instantiate
instances of all of them and should make certain calls to their methods slightly faster.

Tests on `Pixel 7 Pro`, Android 13 show that we are indeed slightly faster for the 
`Displayed` time metrics, however running the tests in series of 10 repeats shows 
pretty unstable results as far as the "winning" configuration is concerned (32-bit vs
64-bit etc) so I suppose it's the usual fickle nature of the `Displayed` measurement
in Android coming into play.  MAUI "hello world" app is between 1.25 and 1.9% faster while
a "plain" Android "hello world" app is between 1.69 and 2.79% faster.

This PR is the first one in a series which will modify the C++ sources with prettier, more
modern, easier maintainable and more performant code as a goal.


dellis1972 (approved):
This looks ok to me. But then my C++ is a bit rusty 🤣

build-tools/xaprepare/xaprepare/Application/GeneratedMonodroidCmakeFiles.cs
```diff
@@ -271,7 +271,7 @@ void WriteShellRuntimeCommand (StreamWriter sw, IEnumerable<string> abis, CmakeB
 				sw.WriteLine ($"{indent}__FORCE=\"${{force}}\"");
 				if (!command.IsHost) {
 					sw.WriteLine ($"{indent}__NATIVE_ABI={abi}");
-					sw.WriteLine ($"{indent}__NATIVE_API_LEVEL=${{{minApiLevel}}}");
```
jonpryor, line 274: …how did this ever work?  Wouldn't it result in e.g. `__NATIVE_API_LEVEL=${21}`, which is non-sensical?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
```diff
@@ -107,5 +107,5 @@
       "Size": 1904
     }
   },
-  "PackageSize": 2730619
+  "PackageSize": 2751099
```
jonpryor, line 110: Seems odd & unexpected that "C++ modernization" would increate `.apk` size by ~20KB.

build-tools/xaprepare/xaprepare/Application/GeneratedMonodroidCmakeFiles.cs
grendello: Yeah, made me wonder too

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
grendello: There's more inlining and more standard library use (header-only classes), which also increases the amount of inlining and generated code.

grendello:
Way, way out of date. We took a different direction.

PR #7472, opened 2022-10-19
[Mono.Android] Generate API docs for API 33

pjcollins:
Updates the default API level used to generate API docs from 31 to 33.
The install path has been updated to use the default path that the
`sdkmanager` tool uses.


pjcollins:
Api docs build: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6845819&view=logs&j=5e189e1d-8dee-542e-977d-a0d63b2bfd80

pjcollins:
Now that a new `xamarin-android-sdk-13` docs moniker has been created, the android-api-docs side of this is building and has only introduced three new warnings which seems acceptable given the scope of change:
https://github.com/xamarin/android-api-docs/pull/42

PR #7475, opened 2022-10-20
[AOT] provide libc and libm stubs for correct linking when NDK is not used

grendello:
Context: 346a93301831047f8d09a477495973249c8acfc7

346a9330 made it possible to build and link AOT shared libraries without having to have Android NDK available.
However, it seems that the produced shared libraries do not depend on `libc` and `libm` standard libraries,
leading to errors at runtime similar to:

```
10-18 23:55:17.844 20389 20389 D monodroid-assembly: monodroid_dlopen: hash match found, DSO name is 'libaot-Mono.Android.dll.so'
10-18 23:55:17.844 20389 20389 I monodroid-assembly: Trying to load shared library '/data/app/~~4kGTzxWG6HKO7nzyg6ryBg==/com.companyname.hellomaui.old-h8Jlutcvf8Dzfpuuq1ouUA==/lib/arm64/libaot-Mono.Android.dll.so'
10-18 23:55:17.844 20389 20389 I monodroid-assembly: Failed to load shared library '/data/app/~~4kGTzxWG6HKO7nzyg6ryBg==/com.companyname.hellomaui.old-h8Jlutcvf8Dzfpuuq1ouUA==/lib/arm64/libaot-Mono.Android.dll.so'. dlopen failed: cannot locate symbol "memset" referenced by "/data/app/~~4kGTzxWG6HKO7nzyg6ryBg==/com.companyname.hellomaui.old-h8Jlutcvf8Dzfpuuq1ouUA==/lib/arm64/libaot-Mono.Android.dll.so"...
```

This is caused by us not passing either `-lc` and `-lm` flags to the linker (or full paths to `libc` and `libm` libraries).
However, without NDK available we do not have neither of these standard libraries to link against.  Fortunately, shared library
builds by default do not try to resolve all the symbols and so we can link AOT libraries against "dummy"/stub versions of
`libc` and `libm`, making sure that the resulting binaries reference them properly:

```shell
$ llvm-readelf -a Mono.Android.dll.so | grep NEEDED
  0x0000000000000001 (NEEDED)    Shared library: [libc.so]
  0x0000000000000001 (NEEDED)    Shared library: [libm.so]
```

This commit adds code to build the `libc` and `libm` stubs, whose paths are then passed during AOT build time to the linker.


dellis1972 (approved):
Looks ok to me. 
Initially I was wondering about the escaping for the ldFlags being removed, but it was moved rather than re-moved :D

grendello:
> Looks ok to me. Initially I was wondering about the escaping for the ldFlags being removed, but it was moved rather than re-moved :D

TBH, I was wondering about the same... :)

grendello:
I'm working on a test for this

src/Xamarin.Android.Build.Tasks/Utilities/ELFHelper.cs
```diff
@@ -28,7 +28,49 @@ public static bool IsEmptyAOTLibrary (TaskLoggingHelper log, string path)
...
+
+		public static bool ReferencesLibrary (string libraryPath, string referencedLibraryName)
+		{
+			if (String.IsNullOrEmpty (libraryPath) || !File.Exists (libraryPath)) {
+				return false;
+			}
+
+			IELF elf = ELFReader.Load (libraryPath);
+
+			// TODO: we need .dynstr
```
jonpryor, line 41: I'm not sure what this TODO is supposed to be saying when the next statement grabs the "need[ed] .dynstr" section.

src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs
```diff
@@ -320,9 +320,13 @@ string GetLdFlags (NdkTools ndk, AndroidTargetArch arch, int level, string toolP
 
 					string libstubsPath = Path.Combine (AndroidBinUtilsDirectory, relPath, "libstubs", ArchToRid (arch)).Replace (" ", "\\ ");
 
+					ldFlags.Append ("\\\"");
 					ldFlags.Append (Path.Combine (libstubsPath, "libc.so"));
-					ldFlags.Append (' ');
+					ldFlags.Append ("\\\"");
+					ldFlags.Append (" \\\"");
```
dellis1972, line 326: Would this code (like the ones in the section above https://github.com/xamarin/xamarin-android/blob/63db768f7cf8925f623a0228f2705a30a6d0783a/src/Xamarin.Android.Build.Tasks/Tasks/GetAotArguments.cs#L293  be better?

```
				libs.Add (Path.Combine (androidLibPath, "libc.so"));
				libs.Add (Path.Combine (androidLibPath, "libm.so"));

				ldFlags.Append ($"\\\"{string.Join ("\\\";\\\"", libs)}\\\"");
```

It looks like we might be able to remove some duplicate code there...

grendello: I tried this, and the AOT still failed, tripping on spaces :(

pjcollins: @grendello I was testing this today and have a diff that appears to be working similar to what @dellis1972 suggested (at least on Windows). I also updated a couple of LLVM/NDK conditions to try to make it a bit easier to follow:

 * https://gist.github.com/pjcollins/067fc494251c7271e4d8d74dfe4afbd3


The only "Failed to load shared library..." [message I am seeing now](https://gist.github.com/pjcollins/93f02c784a9266a0cc36d56f3c4e0cc5#file-gistfile1-txt-L216) mentions `libLLVM.so`, which may still of concern?

grendello: Thanks @pjcollins, updated the PR, let's see how it works :)

pjcollins: I'm also seeing this message on d17-3 classic so this can probably be ignored, at least with respect to the changes in this PR:

> 08-05 18:35:03.581 10257 10257 I monodroid-assembly: Failed to load shared library 'libLLVM.so'. dlopen failed: library "libLLVM.so" not found

src/Xamarin.Android.Build.Tasks/Tasks/Aot.cs
```diff
@@ -170,7 +170,7 @@ IEnumerable<Config> GetAotConfigs (NdkTools ndk)
 						aotOptions.Add ($"ld-name={LdName}");
 					}
 					if (!string.IsNullOrEmpty (LdFlags)) {
-						aotOptions.Add ($"ld-flags={LdFlags}");
+						aotOptions.Add ($"ld-flags=\\\"{LdFlags}\\\"");
```
pjcollins, line 173: I think we may need to revert this change as well

src/Xamarin.Android.Build.Tasks/Tasks/Aot.cs
```diff
@@ -186,6 +183,10 @@ IEnumerable<Config> GetAotConfigs (NdkTools ndk)
 							break;
 					}
 
+					if (!string.IsNullOrEmpty (LdFlags)) {
+						aotOptions.Add ($"ld-flags={LdFlags}");
+					}
+
```
pjcollins, line 189: Do we also want to move the ld-name arg down here? I see there is a comment above about making sure both are always passed last.

pjcollins (approved):
Small question above but otherwise LGTM, the test failures appear to also be occurring on main.

grendello:
The `DotNetInstallAndRun` test failures might have to do with [this issue](https://github.com/dotnet/runtime/issues/77385).

src/Xamarin.Android.Build.Tasks/Tasks/Aot.cs
grendello: Not sure, to be honest. `ld-name` is an argument which takes a simple value which doesn't appear to confuse the AOT parser. I think I'd leave it where it is, especially that everything seems to have worked fine after the `ld-flags` repositioning.

jonpryor:
```markdown
Context: 346a93301831047f8d09a477495973249c8acfc7

346a9330 made it possible to build and link AOT shared libraries
without having to have Android NDK available.  However, it seems
that the produced shared libraries do not depend on the `libc` and
`libm` standard libraries, leading to runtime errors similar to:

	D monodroid-assembly: monodroid_dlopen: hash match found, DSO name is 'libaot-Mono.Android.dll.so'
	I monodroid-assembly: Trying to load shared library '/data/app/~~4kGTzxWG6HKO7nzyg6ryBg==/com.companyname.hellomaui.old-h8Jlutcvf8Dzfpuuq1ouUA==/lib/arm64/libaot-Mono.Android.dll.so'
	I monodroid-assembly: Failed to load shared library '/data/app/~~4kGTzxWG6HKO7nzyg6ryBg==/com.companyname.hellomaui.old-h8Jlutcvf8Dzfpuuq1ouUA==/lib/arm64/libaot-Mono.Android.dll.so'. dlopen failed: cannot locate symbol "memset" referenced by "/data/app/~~4kGTzxWG6HKO7nzyg6ryBg==/com.companyname.hellomaui.old-h8Jlutcvf8Dzfpuuq1ouUA==/lib/arm64/libaot-Mono.Android.dll.so"...

This is caused by us not passing the `-lc` and `-lm` flags to the
linker (or full paths to the `libc.so` and `libm.so` libraries).
However, without the NDK available we do not have either of these
standard libraries to link against.

Fortunately, ELF shared libraries by default will resolve symbols
from anywhere in the process address space, so there is no need to
provide a `libc.so` which contains *all* public symbols.  We can
instead build and use a "dummy"/"stub" version of `libc.so` and
`libm.so`, existing only so that the linker will add the appropriate
`NEEDED` entries and resolve symbols from those libraries.

Update `src/monodroid` to also produce "stub" `libc.so` and `libm.so`
native libraries, and update the `<Aot/>` task to add the equivalent
of `-L path/to/libstubs -lc -lm` to the AOT command-line.

This ensures that the resulting `.so` files now require `libc.so`:

	$ llvm-readelf -a Mono.Android.dll.so | grep NEEDED
	  0x0000000000000001 (NEEDED)    Shared library: [libc.so]
	  0x0000000000000001 (NEEDED)    Shared library: [libm.so]
```

jonpryor:
@dellis1972, @jonathanpeppers : do we have any existing on-device AOT tests?  If so, how hard would it be to update them to verify that the `Failed to load shared library` messages are *not* produced?

dellis1972:
> @dellis1972, @jonathanpeppers : do we have any existing on-device AOT tests? If so, how hard would it be to update them to verify that the `Failed to load shared library` messages are _not_ produced?

We have this one which uses LLVM https://github.com/xamarin/xamarin-android/blob/1526df0115351020902710feeb9903c5387869ea/tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs#L682. 

Plus ANY MSBuildDeviceIntegration test which runs on dotnet and Release Mode is going to be a AOT test. 
So anything in the InstallAndRunTests.cs which uses `IsRelease = true` would do

tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs
```diff
@@ -702,8 +702,30 @@ public void RunWithLLVMEnabled ()
 			else
 				AdbStartActivity ($"{proj.PackageName}/{proj.JavaPackageName}.MainActivity");
 
-			Assert.IsTrue (WaitForActivityToStart (proj.PackageName, "MainActivity",
-				Path.Combine (Root, builder.ProjectDirectory, "startup-logcat.log")));
+			var activityNamespace   = proj.PackageName;
+			var activityName        = "MainActivity";
+			var logcatFilePath      = Path.Combine (Root, builder.ProjectDirectory, "startup-logcat.log");
+			var failedToLoad        = new List<string> ();
+			bool appLaunched        = MonitorAdbLogcat ((line) => {
```
jonpryor, line 709: @grendello, @dellis1972: how does this updated test look for verifying that `.so` files were all loaded successfully?

grendello: It verifies by looking for error messages saying that a library **failed** to load, the next line

jonpryor: @grendello, @dellis1972 : [my attempted unit test change failed](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6994251&view=ms.vss-test-web.build-test-results-tab&runId=61803194&resultId=100030&paneView=debug).  Is that because things aren't working properly, or because the test is looking for the wrong thing?

```
LLVM .so files not loaded:
11-21 23:05:25.104 10192 10192 I monodroid-assembly: Failed to load shared library '/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/UnnamedProject.dll.so'. dlopen failed: library "/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/UnnamedProject.dll.so" not found
11-21 23:05:25.105 10192 10192 I monodroid-assembly: Failed to load shared library 'UnnamedProject.dll.so'. dlopen failed: library "UnnamedProject.dll.so" not found
11-21 23:05:25.106 10192 10192 I monodroid-assembly: Failed to load shared library '/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/Mono.Android.dll.so'. dlopen failed: library "/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/Mono.Android.dll.so" not found
11-21 23:05:25.106 10192 10192 I monodroid-assembly: Failed to load shared library 'Mono.Android.dll.so'. dlopen failed: library "Mono.Android.dll.so" not found
11-21 23:05:25.107 10192 10192 I monodroid-assembly: Failed to load shared library '/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/Java.Interop.dll.so'. dlopen failed: library "/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/Java.Interop.dll.so" not found
11-21 23:05:25.108 10192 10192 I monodroid-assembly: Failed to load shared library 'Java.Interop.dll.so'. dlopen failed: library "Java.Interop.dll.so" not found
11-21 23:05:25.116 10192 10192 I monodroid-assembly: Failed to load shared library '/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/Mono.Android.Runtime.dll.so'. dlopen failed: library "/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/Mono.Android.Runtime.dll.so" not found
11-21 23:05:25.116 10192 10192 I monodroid-assembly: Failed to load shared library 'Mono.Android.Runtime.dll.so'. dlopen failed: library "Mono.Android.Runtime.dll.so" not found
11-21 23:05:25.122 10192 10192 I monodroid-assembly: Failed to load shared library '/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/System.Runtime.dll.so'. dlopen failed: library "/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/System.Runtime.dll.so" not found
11-21 23:05:25.122 10192 10192 I monodroid-assembly: Failed to load shared library 'System.Runtime.dll.so'. dlopen failed: library "System.Runtime.dll.so" not found
11-21 23:05:25.585 10192 10192 I monodroid-assembly: Failed to load shared library '/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/System.Console.dll.so'. dlopen failed: library "/data/app/com.xamarin.runwithinterpreterenabled-EBknjyBcFiejLCycXre3mw==/split_config.x86_64.apk!/lib/x86_64/System.Console.dll.so" not found
11-21 23:05:25.585 10192 10192 I monodroid-assembly: Failed to load shared library 'System.Console.dll.so'. dlopen failed: library "System.Console.dll.so" not found
Expected: 0
But was:  12
```

My guess is that `Failed to load shared library` *always* appears, because not all assemblies are AOT'd. Which seems odd, as `$(EnableLLVM)`=true is set; shouldn't that AOT *all* of them?

Also, those names look weird, e.g. `System.Runtime.dll.so` can't be found.  Shouldn't it be `libaot-System.Runtime.dll.so`?

grendello: Lines ending with `not found` should be ignored, since it's perfectly fine that an AOT library wasn't packaged (e.g. if it was empty) or even created. However, testing for this properly would require finding a library that references one of the symbols in libc/libm and check if that library is loaded. However, I can't find any such library.

@jonathanpeppers originally [found](https://discord.com/channels/732297728826277939/732297837953679412/1032301219391209522) the problem in `libaot-Mono.Android.dll.so`:

```
so I was testing LLVM with the .NET 7 RC, and noticed it was really slow...
Looking at full logcat I get:
10-18 23:55:17.844 20389 20389 D monodroid-assembly: monodroid_dlopen: hash match found, DSO name is 'libaot-Mono.Android.dll.so'
10-18 23:55:17.844 20389 20389 I monodroid-assembly: Trying to load shared library '/data/app/~~4kGTzxWG6HKO7nzyg6ryBg==/com.companyname.hellomaui.old-h8Jlutcvf8Dzfpuuq1ouUA==/lib/arm64/libaot-Mono.Android.dll.so'
10-18 23:55:17.844 20389 20389 I monodroid-assembly: Failed to load shared library '/data/app/~~4kGTzxWG6HKO7nzyg6ryBg==/com.companyname.hellomaui.old-h8Jlutcvf8Dzfpuuq1ouUA==/lib/arm64/libaot-Mono.Android.dll.so'. dlopen failed: cannot locate symbol "memset" referenced by "/data/app/~~4kGTzxWG6HKO7nzyg6ryBg==/com.companyname.hellomaui.old-h8Jlutcvf8Dzfpuuq1ouUA==/lib/arm64/libaot-Mono.Android.dll.so"...

Do you have an idea of what went wrong there?
Basically none of the AOT images are loading
```

And indeed, when you look at symbols referenced by `libaot-Mono.Android.dll.so` from the apk (see the attached zip), you can see both the references **and** a `text` segment:

```
$ llvm-nm -CD libaot-Mono.Android.dll.so | head -10
0000000000048450 W gc.safepoint_poll
                 U memset
00000000000d89d0 D mono_aot_file_info
00000000000a9f80 T p_101_plt_Java_Net_Proxy_get_NoProxy_llvm
00000000000aa00c T p_108_plt_Android_Runtime_JNIEnv_FindClass_string_llvm
00000000000aa020 T p_109_plt_Android_Runtime_JNIEnv_GetMethodID_intptr_string_string_llvm
00000000000aa034 T p_110_plt_Android_Runtime_JNIEnv_DeleteGlobalRef_intptr_llvm
00000000000aa070 T p_113_plt_Android_Runtime_JNIEnv_CallObjectMethod_intptr_intptr_llvm
00000000000aa0d4 T p_118_plt_Java_Net_HttpURLConnection__ctor_intptr_Android_Runtime_JniHandleOwnership_llvm
00000000000aa0e8 T p_119_plt_Android_Runtime_XAPeerMembers__ctor_string_System_Type_llvm
```

However, `Mono.Android.dll.so` produced with XA/main has just the data `mono_aot_file_info` symbol:

```
$ llvm-nm -CD Mono.Android.dll.so | head -10
0000000000062840 D mono_aot_file_info
```

@jonathanpeppers's test was with one of the NET7 RC builds, so perhaps something has changed inbetween? Also with regard to library names?

[libraries.zip](https://github.com/xamarin/xamarin-android/files/10065293/libraries.zip)

jonathanpeppers: Is the test passing because `$(AndroidNdkDirectory)` is set and valid?

When I hit the original problem, I was building on Azure DevOps with an NDK in a non-standard location -- would be the same as not having an NDK installed at all.

I solved the problem by doing:

```yaml
-p:AndroidNdkDirectory=$(ANDROID_NDK_HOME)
```

PR #7492, opened 2022-10-25
[dependabot] Update submodule ignore list

pjcollins:
Fixes the submodule ignore list by using the relative path of the
submodule rather than only the repo name.


jonathanpeppers (approved):
Probably will have to merge/close PRs to actually test this?

PR #7494, opened 2022-10-25
[readme] Be more explicit about support status.

jpobst:
Try to be a bit more explicit about the status of XA and Android for .NET to clear up some confusion we've seen from users.  

Particularly:

- Classic Xamarin.Android is deprecated and has an announced end of support date.
- Android for .NET exists and is supported as a standalone product without MAUI.

We can tweak the exact verbiage as needed, but these are the main points we should try to convey.


jonathanpeppers (commented):
Can we get xamarin-macios to update their README in a similar way?

PR #7524, opened 2022-11-03
[Tests] Replace azurestorage with a github repo.

dellis1972:
Our `https://xamjenkinsartifact.azureedge.net` site for some of the unit test data seems to have disappeared.

So lets use a github repo to store this data instead. The unit tests can download the raw files directly from github. The repo will be public as well so there won't be any need for authorization.

The new repo can currently be found at https://github.com/dellis1972/xamarin-android-unittest-files.


jonathanpeppers (approved):
This looks good if the tests pass.

dellis1972:
This is still in progress so don't merge yet

jonathanpeppers:
I reran some test lanes that timed out.

The tests like these should hopefully be fixed by #7505:

image

jonathanpeppers:
I'm merging this, as it gets things much greener than they were. 💚

PR #7533, opened 2022-11-07
[DevOps] Rename the yaml-tempaltes resources to match the MAUI repo.

mandel-macaque:
In an attempt to unify the build for all the SDKs we need to rename all the resources to use the same names. This change follows the pattern used by tehe MAUI repo.

Partial fix for https://github.com/xamarin/sdk-insertions/issues/39

### DIFF
--- a/build-tools/automation/azure-pipelines-nightly.yaml
+++ b/build-tools/automation/azure-pipelines-nightly.yaml
@@ -11,7 +11,7 @@ pr:
 # External sources, scripts, tests, and yaml template files.
 resources:
   repositories:
-  - repository: yaml
+  - repository: xamarin-templates
     type: github
     name: xamarin/yaml-templates
     ref: refs/heads/main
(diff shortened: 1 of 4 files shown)


pjcollins:
@mandel-macaque Looks like this failed https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=6929520&view=results -- Some templates parameterized the yaml-templates resource name to deal with the fact that various repos used different names, those instances will also need to be updated (e.g. https://github.com/xamarin/xamarin-android/blob/main/build-tools/automation/azure-pipelines.yaml#L1454)

build-tools/automation/azure-pipelines-nightly.yaml
```diff
@@ -11,7 +11,7 @@ pr:
 # External sources, scripts, tests, and yaml template files.
 resources:
   repositories:
-  - repository: yaml
+  - repository: xamarin-templates
```
pjcollins, line 14: If we are working to standardize this name across repos, should we use the repo name `yaml-templates` instead of `xamarin-templates? I wonder if the latter name might get confused with https://github.com/xamarin/xamarin-templates.

pjcollins (approved):

PR #7536, opened 2022-11-07
[build] Make external build targets more consistent

pjcollins:
The external builds of monodroid and java.interop prep have been made
more consistent.  They will both use an `<Exec/>` task to build with the
preview dotnet SDK that is installed into `bin/$(Configuration)/dotnet`.

### DIFF
--- a/build-tools/scripts/DotNet.targets
+++ b/build-tools/scripts/DotNet.targets
@@ -1,17 +1,17 @@
 <Project>
   <PropertyGroup>
     <_Root>$(MSBuildThisFileDirectory)..\..\</_Root>
-    <_BinlogDateTime>$([System.DateTime]::Now.ToString("yyyyMMddTHHmmss"))</_BinlogDateTime>
+    <_BinlogPathPrefix>$(_Root)bin/Build$(Configuration)/msbuild-$([System.DateTime]::Now.ToString("yyyyMMddTHHmmss"))</_BinlogPathPrefix>
   </PropertyGroup>
   <Target Name="BuildExternal">
-    <MSBuild
-        Projects="$(_Root)\external\monodroid\monodroid.proj"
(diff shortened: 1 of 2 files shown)


build-tools/scripts/DotNet.targets
```diff
@@ -4,14 +4,14 @@
...
+        Command="&quot;$(DotNetPreviewTool)&quot; build monodroid.proj -c $(Configuration) -p:DebuggingToolsOutputDirectory=$(MicrosoftAndroidSdkOutDir) -p:CompatTargetsOutputDirectory=$(XAInstallPrefix)xbuild/Novell -v:n"
+        WorkingDirectory="$(_Root)external\monodroid"
     />
   </Target>
   <Target Name="PrepareJavaInterop">
     <Exec
-        Command="&quot;$(DotNetPreviewTool)&quot; build -t:Prepare Java.Interop.sln -c $(Configuration) -p:JdksRoot=$(JavaSdkDirectory) -p:DotnetToolPath=$(DotNetPreviewTool) -bl:$(_Root)bin/Build$(Configuration)/msbuild-$(_BinlogDateTime)-prepare-java-interop.binlog"
+        Command="&quot;$(DotNetPreviewTool)&quot; build -t:Prepare Java.Interop.sln -c $(Configuration) -p:JdksRoot=$(JavaSdkDirectory) -p:DotnetToolPath=$(DotNetPreviewTool) -v:n"
         WorkingDirectory="$(_Root)external\Java.Interop"
     />
```
jonathanpeppers, line 16: It seems like we lose the `.binlog` on these commands now. I don't see how the `make` target will create these, as `<Exec/>` runs `dotnet build` in a new process?

pjcollins: Oh right, we would only get the normal output from exec in the binlog created by the make file. Let me swap this to use the default minimal output verbosity but also produce binlogs again.

jonathanpeppers (approved):

PR #7552, opened 2022-11-11
Add API-33 to the nightly build

### LINKED ISSUE #7490: Add API-33 Emulators to Nightly builds
### Android application type

Classic Xamarin.Android (MonoAndroid12.0, etc.), Android for .NET (net6.0-android, etc.)

### Affected platform version

ALL

### Description

Add API-33 Emulators to Nightly builds so we can test on those platforms.

### Steps to Reproduce

None

None

### Relevant log output

_No response_

dellis1972:
Fixes #7490

Lets ass API 33 to the nightly build tests. Note the `default` image is not available so we have to use the `google_apis` image instead. 

Also it seems that under API 33 dotnet is unable to get the hardware address of the network adapters. As a result the tests which compare these were failing. So for API 33 we ignore the hardware address check, but still continue to check other aspects of the network adapters.

### DIFF
--- a/tests/Mono.Android-Tests/System.Net/NetworkInterfaces.cs
+++ b/tests/Mono.Android-Tests/System.Net/NetworkInterfaces.cs
@@ -85,6 +85,9 @@ namespace System.NetTests
 
 			bool HardwareAddressesAreEqual (byte[] one, byte[] two)
 			{
+				// Under API 33 .Net doesn't return the hardware address. So we need to ignore it
+				if (Android.OS.Build.VERSION.SdkInt == Android.OS.BuildVersionCodes.Tiramisu)
+					return true;
 				if (one == two)
 					return true;
 				if (one == null || two == null)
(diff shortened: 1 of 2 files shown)


pjcollins (approved):

tests/Mono.Android-Tests/System.Net/NetworkInterfaces.cs
```diff
@@ -85,6 +85,9 @@ bool AddressesAreEqual (List <IPAddress> one, List <IPAddress> two)
 
 			bool HardwareAddressesAreEqual (byte[] one, byte[] two)
 			{
+				// Under API 33 .Net doesn't return the hardware address. So we need to ignore it
+				if (Android.OS.Build.VERSION.SdkInt == Android.OS.BuildVersionCodes.Tiramisu)
```
jonpryor, line 89: @dellis1972: Shouldn't this use `>=` so that API-34 hits this codepath?  Or do we *want* it to fail on future API-34?

dellis1972: So we could. but i'd rather change that in API 34 if or when we hit that issue.  This is only in a unit test not the actual System.Net code.

PR #7553, opened 2022-11-11
[ci] Move build stages into yaml templates

pjcollins:
Context: https://github.com/xamarin/sdk-insertions/pull/42
Context: https://devdiv.visualstudio.com/DevDiv/_build?definitionId=17671&_a=summary

As part of efforts to create an integration and release pipeline for all
of MAUI we will want to "templatize" more of our YAML.  Our three build
stages have been moved into new templates that can be referenced by the
MAUI "mega pipeline".


build-tools/automation/yaml-templates/build-macos.yaml
```diff
@@ -0,0 +1,73 @@
+parameters:
+  buildPoolName: $(MacBuildPoolName)
+  buildPoolImage: $(MacBuildPoolImage)
+  dependsOn: []
+  jobName: mac_build_create_installers
+  jobDisplayName: macOS > Build
+  provisionatorChannel: latest
+  repoAlias: self
```
mandel-macaque, line 8: Can we do 'repositoryAlias' ? We have a similar parameter on iOS. And would be nice to be able to pass the hash, I added a template checkout here: https://github.com/xamarin/yaml-templates/blob/main/sdk-unified/steps/checkout/v1.yml

pjcollins: I changed the param name to `repositoryAlias` and updated all the build jobs to use the checkout template.

pjcollins:
> Everything looks "the same". Is there anything here that changed in a significant way?

The only minor change here should be that the `- checkout: self` step has been replaced with https://github.com/xamarin/yaml-templates/blob/main/sdk-unified/steps/checkout/v1.yml, otherwise the rest of the YAML has just been moved around and further parameterized.

PR #7556, opened 2022-11-15
[tests] Switch to `httpbin.org`

dellis1972:
`httpbingo.org` is returning the following error when we try to redirect.

```
Forbidden redirect URL. Be careful with this link.
```

To lets try to use `httpbin.org` directly instead.


pjcollins (approved):

dellis1972:
Hmm getting a timeout now.

```
System.Net.Http.HttpRequestException : Response status code does not indicate success: 504 (Gateway Time-out).
```

But not on both tests...

Issue #7532, opened 2022-11-05
3rd Party library internal invoke native code lead app crash


jingliancui:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

Version 17.3.6 Windows Version

### Description

.Net6 app Invoke 3rd party lib will lead crash.
Native app didn't crash.

Please take a look at the attachment codes and run them.

### Steps to Reproduce

[JPushMAUI.zip](https://github.com/xamarin/xamarin-android/files/9944038/JPushMAUI.zip)
1.Unzip the ziped file.
2.Using Android Studio run the app that placed on the native folder with debug mode.
3.Using Visual Studio run the app that placed on the SampleApp folder with debug mode.
4.Native app normal run after occur `JPushInterface.init(this);` which wrote on `MainActivity.java`
5.SampleApp crash after occur `CN.Jpush.Android.Api.JPushInterface.Init(this);` which wrote on `MainActivity.cs`

Additions info:
Then we should focus on the logcat window on AS and the output window on VS for
`[jiguang]`
At last the crash are occur from the `*.so` exception because output window will show a message error like: fatal signal.

Take a look at csproj file because it need ManifestPlaceholder, But I've double checked in` obj/Debug/net6.0-android/AndroidManifest.xml`, I think all values are correctly set.

### Did you find any workaround?

Sorry, I've no idea about that.

### Relevant log output

```shell
[JIGUANG-JPush] [JPushInterface] action:init - sdkVersion:4.8.4, buildId:225
[JIGUANG-JPush] [AndroidUtil] action:checkValidManifest
[yname.sampleap] Accessing hidden method Landroid/app/ActivityThread;->currentActivityThread()Landroid/app/ActivityThread; (greylist, reflection, allowed)
[yname.sampleap] Accessing hidden field Landroid/app/ActivityThread;->mActivities:Landroid/util/ArrayMap; (greylist, reflection, allowed)
[JIGUANG-JPush] [JPushConstants] found messageReceiverClass :com.companyname.sampleapp.SampleReceiverat com.companyname.sampleapp by getCommonReceiverNames
[JIGUANG-JPush] [AndroidUtil] The fileprovider required for inapp-message is incorrectly configured, please check AndroidManifest.xml
[JIGUANG-JPush] [ActionHelper] doSingleAction: third_init
[...]

PR #7560, opened 2022-11-16
[DotNet] Attach current thread to MonoVM when registering a type

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/7532

Whenever any MonoVM APIs are called, the current thread of execution must be attached to the runtime or we risk native code crashes, for instance:

    Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x40 in tid 1860 (jg_fr_pool_thre), pid 1792 (yname.sampleapp)
    backtrace:
      00 pc 0011588e  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonosgen-2.0.so
      01 pc 001156ae  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonosgen-2.0.so
      02 pc 00115575  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonosgen-2.0.so (mono_threads_enter_gc_unsafe_region_internal+53)
      03 pc 000973c3  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonosgen-2.0.so (mono_runtime_invoke+51) (BuildId: c54662bbf82dbdadd595ca9d3e31dce29735885f)
      04 pc 00027ace  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonodroid.so (xamarin::android::internal::MonodroidRuntime::Java_mono_android_Runtime_register(_JNIEnv*, _jstring*, _jclass*, _jstring*)+286)
      05 pc 000279a0  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonodroid.so (Java_mono_android_Runtime_register+48)

The above trace translates to the following locations in the Mono runtime:

    copy_stack_data_internal
    /__w/1/s/src/mono/mono/utils/mono-threads-coop.c:191
    copy_stack_data
    /__w/1/s/src/mono/mono/utils/mono-threads-coop.c:246

    mono_threads_enter_gc_unsafe_region_unbalanced_with_info
    /__w/1/s/src/mono/mono/utils/mono-threads-coop.c:476

    mono_runtime_invoke
    /__w/1/s/src/mono/mono/metadata/object.c:2442

In this case, a pointer to Mono thread information structure (`MonoThreadInfo*`) is null in `copy_stack_data` which, in turn, causes the segfault when the pointer is dereferenced.

In the case of issue #7532, `Java_mono_android_Runtime_register` is called on behalf of a 3rd party library on a thread that is, most likely, created by that library and thus not (yet) attached to the runtime by the time the registration attempt is made.

Attach thread to the runtime by calling
`mono_jit_thread_attach (nullptr)` in DotNet builds to fix the issue. `nullptr` is used because .NET only has a single AppDomain and there's no need to pass around pointers to it.


lambdageek (approved):

jonpryor:
```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/7532

Whenever any MonoVM APIs are called, the current thread of execution
must be attached to the runtime or we risk native code crashes, e.g.:

	Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x40 in tid 1860 (jg_fr_pool_thre), pid 1792 (yname.sampleapp)
	backtrace:
	  00 pc 0011588e  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonosgen-2.0.so
	  01 pc 001156ae  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonosgen-2.0.so
	  02 pc 00115575  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonosgen-2.0.so (mono_threads_enter_gc_unsafe_region_internal+53)
	  03 pc 000973c3  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonosgen-2.0.so (mono_runtime_invoke+51) (BuildId: c54662bbf82dbdadd595ca9d3e31dce29735885f)
	  04 pc 00027ace  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonodroid.so (xamarin::android::internal::MonodroidRuntime::Java_mono_android_Runtime_register(_JNIEnv*, _jstring*, _jclass*, _jstring*)+286)
	  05 pc 000279a0  /data/app/~~lwgdvtQhIfK0II0Nd0mtag==/com.companyname.sampleapp-8yF54k9YuWCWpsuGRJvVQw==/lib/x86/libmonodroid.so (Java_mono_android_Runtime_register+48)

The above trace translates to the following locations in the Mono runtime:

	copy_stack_data_internal
	/__w/1/s/src/mono/mono/utils/mono-threads-coop.c:191
	copy_stack_data
	/__w/1/s/src/mono/mono/utils/mono-threads-coop.c:246

	mono_threads_enter_gc_unsafe_region_unbalanced_with_info
	/__w/1/s/src/mono/mono/utils/mono-threads-coop.c:476

	mono_runtime_invoke
	/__w/1/s/src/mono/mono/metadata/object.c:2442

In this case, a pointer to Mono thread information structure
(`MonoThreadInfo*`) is null in `copy_stack_data()` which, in turn,
causes the segfault when the pointer is dereferenced.

In the case of issue #7532, `Java_mono_android_Runtime_register()` is
called on behalf of a 3rd party library on a thread that is, most
likely, created by that library and thus not (yet) attached to the
runtime by the time the registration attempt is made.

Attach thread to the runtime by calling
`mono_jit_thread_attach(nullptr)` in .NET builds to fix the issue.
`nullptr` is used because .NET only has a single AppDomain and
there's no need to pass around pointers to it.
```

tests/Mono.Android-Tests/Java.Interop/JnienvTest.cs
```diff
@@ -32,11 +32,22 @@ public void TestMyPaintColor ()
...
 		delegate void CB (IntPtr jnienv, IntPtr java_instance);
 
 		[DllImport ("reuse-threads")]
 		static extern int rt_invoke_callback_on_new_thread (CB cb);
 
+		[Test]
+		public void RegisterTypeOnNewThread ()
+		{
+			Java.Lang.JavaSystem.LoadLibrary ("reuse-threads");
+			int ret = rt_register_type_on_new_thread ("from/NewThread", Application.Context.ClassLoader.Handle);
```
jonpryor, line 47: Shouldn't this be:

```c
rt_register_type_on_new_thread ("from.NewThread", Application.Context.ClassLoader.Handle);
```

because `rt_register_type_on_new_thread()` uses `ClassLoader.loadClass()`?

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +200,46 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+_register_type_from_new_thread (const char *java_type_name, jobject class_loader)
+{
+	JNIEnv *env              = _get_env ("_register_type_from_new_thread");
+	jclass ClassLoader_class = (*env)->FindClass (env, "java/lang/ClassLoader");
+	jmethodID loadClass      = (*env)->GetMethodID (env, ClassLoader_class, "loadClass", "(Ljava/lang/String;)Ljava/lang/Class;");
+	jobject loaded_class     = (*env)->CallObjectMethod (env, class_loader, loadClass, java_type_name);
+
+	if ((*env)->ExceptionOccurred (env) != NULL) {
+		__android_log_print (ANDROID_LOG_INFO, "XA/RuntimeTest", "FAILURE: class '%s' cannot be loaded, Java exception thrown!", java_type_name);
+		return -1;
```
jonpryor, line 213: We should call `(*env)->ExceptionClear(env)` before returning, so that there isn't a pending exception within the unit tests.

jonpryor: You may want to use different negative return values, so that it's more obvious in unit test output which part failed.

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +200,46 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+
+	if (r) {
+		__android_log_print (ANDROID_LOG_INFO, "XA/RuntimeTest", "RegisterOnNewThread: pthread_create() failed! %i: %s", r, strerror (r));
+		return -1;
+	}
+
+	void *tr;
+	pthread_join (t, &tr);
+
+	return 0;
```
jonpryor, line 244: …though the `return` statements within `_register_type_from_new_thread()` are kinda moot, as we drop the return value that `pthread_join()` provides!

Which means that the unit test will always succeed, even if `ClassLoader.loadClass()` fails!  (There just might be some "random unrelated crash elsewhere" as the pending Java exception is actually thrown…

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +200,46 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+}
+
+JNIEXPORT int JNICALL
+rt_register_type_on_new_thread (const char *java_type_name)
+{
+	JNIEnv *env = _get_env ("rt_register_type_on_new_thread");
+	pthread_t t;
+
+	/* 1: Create a thread... */
+	int r = pthread_create (&t, NULL, _register_type_from_new_thread, java_type_name);
```
jonpryor, line 234: Kinda surprised that this compiles without warnings (assuming it *does* compile without warnings), as `pthread_create()` specifies a `void*` return value, and `int` is not implicitly convertible to `void*`.

grendello: The older code there did the same, I didn't check the manpage. I guess it's always worth checking... :)

grendello: hm, no, just checked the manpage:

```
SYNOPSIS
       #include <pthread.h>

       int pthread_create(pthread_t *restrict thread,
                          const pthread_attr_t *restrict attr,
                          void *(*start_routine)(void *),
                          void *restrict arg);

```

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +206,67 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+_register_type_from_new_thread (void *data)
+{
+	RegisterFromThreadContext *context = (RegisterFromThreadContext*)data;
+
+	if (context == NULL) {
+		return -100;
+	}
+
+	JNIEnv *env              = _get_env ("_register_type_from_new_thread");
+	jclass ClassLoader_class = (*env)->FindClass (env, "java/lang/ClassLoader");
```
jonpryor, line 221: you should `(*env)->DeleteLocalRef(env, ClassLoader_class)` in the cleanup of this method, so that we don't overflow the LREF table.

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +206,67 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+	jmethodID loadClass      = (*env)->GetMethodID (env, ClassLoader_class, "loadClass", "(Ljava/lang/String;)Ljava/lang/Class;");
+	jobject loaded_class     = (*env)->CallObjectMethod (env, context->class_loader, loadClass, context->java_type_name);
```
jonpryor, line 223: you should `(*env)->DeleteLocalRef(env, loaded_class)` in the cleanup of this method, so that we don't overflow the LREF table.

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +206,67 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+
+	if ((*env)->ExceptionOccurred (env) != NULL) {
+		(*env)->ExceptionClear (env);
+		__android_log_print (ANDROID_LOG_INFO, "XA/RuntimeTest", "FAILURE: class '%s' cannot be loaded, Java exception thrown!", context->java_type_name);
```
jonpryor, line 227: Should probably add `(*env)->ExceptionDescribe(env)`, so that we know what it is.

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +206,67 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+		return -101;
+	}
+
+	jmethodID Object_ctor    = (*env)->GetMethodID (env, loaded_class, "<init>", "()V");
+	jobject   instance       = (*env)->NewObject (env, loaded_class, Object_ctor);
```
jonpryor, line 232: you should `(*env)->DeleteLocalRef(env, instance)` in the cleanup of this method, so that we don't overflow the LREF table.

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +206,67 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+
+	if ((*env)->ExceptionOccurred (env) != NULL || instance == NULL) {
+		(*env)->ExceptionClear (env);
```
jonpryor, line 235: Ditto calling `(*env)->ExceptionDescribe(env)`.

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +206,115 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+		__android_log_print (ANDROID_LOG_INFO, "XA/RuntimeTest", "FAILURE: unable to get id of method 'loadClass' in the 'java/lang/ClassLoader' class!");
+		goto cleanup;
+	}
+
+	jobject loaded_class = (*env)->CallObjectMethod (env, context->class_loader, loadClass, context->java_type_name);
+
+	if ((*env)->ExceptionOccurred (env) != NULL) {
+		(*env)->ExceptionClear (env);
+		__android_log_print (ANDROID_LOG_INFO, "XA/RuntimeTest", "FAILURE: class '%s' cannot be loaded, Java exception thrown!", context->java_type_name);
+		(*env)->ExceptionDescribe (env);
```
jonpryor, line 253: Calling `ExceptionDescribe()` needs to come *before* calling `ExceptionClear()`, as once you call `ExceptionClear()` there is no longer an exception to describe.

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +206,115 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+		__android_log_print (ANDROID_LOG_INFO, "XA/RuntimeTest", "FAILURE: unable to find the '%s' class constructor!", context->java_type_name);
+		goto cleanup;
+	}
+
+	jobject instance = (*env)->NewObject (env, loaded_class, Object_ctor);
+
+	if ((*env)->ExceptionOccurred (env) != NULL || instance == NULL) {
+		(*env)->ExceptionClear (env);
+		__android_log_print (ANDROID_LOG_INFO, "XA/RuntimeTest", "FAILURE: instance of class '%s' wasn't created!", context->java_type_name);
+		(*env)->ExceptionDescribe (env);
```
jonpryor, line 276: Calling `ExceptionDescribe()` needs to come *before* calling `ExceptionClear()`, as once you call `ExceptionClear()` there is no longer an exception to describe.

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +206,115 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+	}
+
+	JNIEnv *env = _get_env ("_register_type_from_new_thread");
+
+	if ((*env)->PushLocalFrame (env, 3) < 0) {
+		__android_log_print (ANDROID_LOG_INFO, "XA/RuntimeTest", "FAILURE: unable to create a local reference frame!");
+
+		if ((*env)->ExceptionOccurred (env)) {
+			(*env)->ExceptionClear (env);
+			(*env)->ExceptionDescribe (env);
```
jonpryor, line 227: Calling `ExceptionDescribe()` needs to come *before* calling `ExceptionClear()`, as once you call `ExceptionClear()` there is no longer an exception to describe.

tests/Mono.Android-Tests/jni/reuse-threads.c
```diff
@@ -200,3 +206,115 @@ rt_invoke_callback_on_new_thread (CB cb)
...
+	}
+
+	jmethodID loadClass = (*env)->GetMethodID (env, ClassLoader_class, "loadClass", "(Ljava/lang/String;)Ljava/lang/Class;");
+	if (loadClass == NULL) {
+		ret = -103;
+		__android_log_print (ANDROID_LOG_INFO, "XA/RuntimeTest", "FAILURE: unable to get id of method 'loadClass' in the 'java/lang/ClassLoader' class!");
+		goto cleanup;
+	}
+
+	jobject loaded_class = (*env)->CallObjectMethod (env, context->class_loader, loadClass, context->java_type_name);
```
jonpryor, line 248: Calling `ClassLoader.loadClass(String)` requires passing a `java.lang.String` instance, aka `jstring`.  You are instead passing `context->java_type_name`, which is a `const char*`.  JNI is thus interpreting the `const char*` as a `jobject`/`jstring`, and complaining that it's not a valid JNI object reference.  This is correct, because it's *not* a JNI object reference.

Instead, you need:

```c
jstring s = (*env)->NewStringUTF (env, context->java_type_name);
jobject loaded_class = (*env)->CallObjectMethod (env, context->class_loader, loadClass, s);
```

and you'll need to update the value used in the `JNIEnv::PushLocalFrame()` call.

jingliancui:
It look like the pipeline shown red, are there any plan date for preview version or release version? Because this 3rd party library is important in China region. Thanks.

PR #7563, opened 2022-11-17
[ci] Attach logcat files to apk test failures

pjcollins:
Attaching the logcat file to the result when a test fails should improve
our ability to quickly review test failures in the APK test suites.


pjcollins:
Hmm this looks like it will only work when processing the nunit3 test result format, I'll need to look at the differences in that and what we're producing.

PR #2905, opened 2019-03-31
Deploy target

xtmq:
When adb fails it is difficult to understand thats going on especially from IDE where an user usually observes only error messages, something like: 
```
MyAndroidProject
   ...\Xamarin\Android\Xamarin.Android.Common.targets(3281,3): error MSB3073: 
The command ""C:\Program Files (x86)\Android\android-sdk\platform-tools\\adb" -s emulator-5554 install -r "bin\Debug\com.companyname.MyAndroidProject-Signed.apk"" exited with code 1.
```

It is better to have an error message like
```
MyAndroidProject
  ...\Xamarin\Android\Xamarin.Android.Common.targets(3285,9): error : 
The command ""C:\Program Files (x86)\Android\android-sdk\platform-tools\\adb" -s emulator-5554 install -r "bin\Debug\com.companyname.MyAndroidProject-Signed.apk"" exitied with code 1: 
adb: failed to install bin\Debug\com.companyname.MyAndroidProject-Signed.apk: Failure [INSTALL_PARSE_FAILED_UNEXPECTED_EXCEPTION: Failed to parse /data/app/vmdl192544199.tmp/base.apk: AndroidManifest.xml]
```

PR #7566, opened 2022-11-18
[Build] Log better errors when compiling/linking with binutils

grendello:
If an error happens when compiling or linking native assembler files 
generated by Xamarin.Android, we log an error which doesn't contain 
much information:

    Error XA3006 Could not compile native assembly file: compressed_assemblies.x86.ll

The real cause of the issue is logged as well, however not as an error
but rather a "plain" message, which is not immediately accessible
to the user experiencing the issue.

Keep logging the output messages of the assembler and linker programs as
before, but whenever an error occurs, log their entire output (both
stdout and stderr) as an MSBuild error message.


src/Xamarin.Android.Build.Tasks/Tasks/LinkApplicationSharedLibraries.cs
```diff
@@ -97,9 +103,22 @@ void RunLinker (Config config)
 
 				if (proc.ExitCode != 0) {
 					LogCodedError ("XA3007", Properties.Resources.XA3007, Path.GetFileName (config.OutputSharedLibrary));
+					LogErrors ("stdout", stdoutLines);
```
dellis1972, line 106: Not sure what the best thing to do here is. These errors will not be "attached" to the XA3007. they will be reported as different errors per line. 

Should we instead make the `Properties.Resources.XA3007` take an additional argument which is the output of the tool?

grendello: It, somewhat, mimics [this behavior](https://github.com/xamarin/xamarin-android/commit/b002dc38196a0bedb18c82d228056e1ba108459c). If we added the new argument, as you suggest, would it retain the newlines from the outputs?

grendello: @dellis1972 my hope was that the errors will be presented together, so that even though the coded error will be "separate", the following lines would be shown just below it, in the correct context.

jonathanpeppers: @jonpryor do we have an old comment/commit message that talks about multiline errors and how they show up in Visual Studio? I think we might be able to just append all these messages in a single error message?

grendello: > @jonpryor do we have an old comment/commit message that talks about multiline errors and how they show up in Visual Studio? I think we might be able to just append all these messages in a single error message?

If the appended content preserves newlines, then sure, I'll be fine with it. Otherwise I'd rather do what this PR currently does.

dellis1972: If we do log them seperately it would be better if it had the same Error code (XA3007)  So maybe use `Log.LogCodedError` instead.

jonpryor: @jonathanpeppers [asked](https://github.com/xamarin/xamarin-android/pull/7566#discussion_r1026538064):

> do we have an old comment/commit message that talks about multiline errors and how they show up in Visual Studio?

We have this:

https://github.com/xamarin/xamarin-android/pull/2905#issuecomment-479599104

**CreateErrorsItemGroupSingleLine** was the best looking approach, which joins all lines together with `\n` between them.  This appears to be the approach currently taken in this PR.

PR #7621, opened 2022-12-07
[Xamarin.Android.Build.Tasks] avoid `File.Exists()` check

jonathanpeppers:
`dotnet-trace` output of a `dotnet new maui` app:

    dotnet trace collect --format speedscope -- C:\src\xamarin-android\bin\Release\dotnet\dotnet.exe build -bl --no-restore bar.csproj

Shows an interesting case of `File.Exists()`:

    49.88ms xamarin.android.build.tasks!Xamarin.Android.Tasks.FilterAssemblies.RunTask()
     8.09ms System.Private.CoreLib.il!System.IO.File.Exists(class System.String)

The common case is the files always exist. The rare case appears to be 783ac9e9, which would be when something like a `<ProjectReference/>` doesn't exist.

Instead of calling `File.Exists()` on every .NET assembly (likely done multiple times in a build), we can handle `FileNotFoundException` and take appropriate action.

This will save ~8ms on *every* build, and we should get the exact same behavior as before. Potentially larger projects could even save more.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/FilterAssemblies.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/FilterAssemblies.cs
@@ -42,55 +42,60 @@ namespace Xamarin.Android.Tasks
 				if (frameworkReferenceName.StartsWith ("Microsoft.NETCore.", StringComparison.OrdinalIgnoreCase)) {
 					continue; // No need to process BCL assemblies
 				}
-				if (!File.Exists (assemblyItem.ItemSpec)) {
-					Log.LogDebugMessage ($"Skipping non-existent dependency '{assemblyItem.ItemSpec}'.");
-					continue;
-				}
 				if (string.Equals (assemblyItem.GetMetadata ("TargetPlatformIdentifier"), "android", StringComparison.OrdinalIgnoreCase)) {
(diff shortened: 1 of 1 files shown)


dellis1972 (approved):

dellis1972:
@jonathanpeppers  I suspect the `DeduplicateAssemblies` method in `ProcessAssemblies` can use the same treatment. It would also help use remove some Linq usage https://github.com/xamarin/xamarin-android/blob/ee4322054292a8f6e33ca6c53fe9ce609fce3c95/src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs#L121

jonathanpeppers:
Yeah `<ProcessAssemblies/>` might be worth looking at later. It has a faster code path when there is 1 RID, which should be the case for `Debug` builds:

https://github.com/xamarin/xamarin-android/blob/2a16828a6c73d837e81c3a2d0c1ae2840d00873e/src/Xamarin.Android.Build.Tasks/Tasks/ProcessAssemblies.cs#L55-L62

So that might be why I didn't see this one.

jonathanpeppers:
Ok, I need to handle `DirectoryNotFoundException` as well.

PR #7626, opened 2022-12-09
[Xamarin.Android.Build.Tasks] perf improvements for LlvmIrGenerator

jonathanpeppers:
Reviewing `dotnet trace` output for a `dotnet new maui` app:

    dotnet trace collect --format speedscope -- .\bin\Release\dotnet\dotnet.exe build -bl --no-restore foo.csproj

21% of the time in `<GenerateJavaStubs/>` is spent in:

    249.60ms xamarin.android.build.tasks!Xamarin.Android.Tasks.LLVMIR.LlvmIrComposer.Write(value class Xamarin.Android.Tools.Andro...

Drilling down, some of the hot spots are:

    68.64ms xamarin.android.build.tasks!Xamarin.Android.Tasks.LLVMIR.LlvmIrGenerator.WriteString(class System.String,class System.St
    40.39ms xamarin.android.build.tasks!Xamarin.Android.Tasks.LLVMIR.LlvmIrGenerator.WritePointer(class Xamarin.Android.Tasks.LLVMIR

These seem to be places we create intermediate `string` or `byte[]` objects.

The types of things I solved so far:

* I audited every `$""` to see if we could call `TextWriter.Write()` or `StringBuilder.Append()` multiple times to avoid creating a string.

* Use `ArrayPool` in combination with `Encoding.GetBytes()`.

* Use `char` overloads, where we were using a 1-length `string`.

Hopefully, I've changed no behavior at all here, and we've just saved a lot of `string` and `byte[]` allocations.

After these changes go in, we can probably profile and find other areas to improve.

Testing the same `dotnet new maui` app, the overall time is now:

    190.97ms xamarin.android.build.tasks!Xamarin.Android.Tasks.LLVMIR.LlvmIrComposer.Write(value class Xamarin.Android.Tools.AndroidT

I think this will save ~60ms on incremental builds of the .NET MAUI project template. Larger projects with more Java subclasses could save even more.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Utilities/LlvmIrGenerator/LlvmIrGenerator.cs
+++ b/src/Xamarin.Android.Build.Tasks/Utilities/LlvmIrGenerator/LlvmIrGenerator.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.IO;
 using System.Text;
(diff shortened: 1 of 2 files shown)


grendello (approved):
The changes make code slightly less readable, but savings are large enough to justify it.

dellis1972 (approved):
Looks good if its green :D

jonathanpeppers:
> The changes make code slightly less readable, but savings are large enough to justify it.

I like `$""` better too in most of these places...

grendello:
> > The changes make code slightly less readable, but savings are large enough to justify it.
> 
> I like `$""` better too in most of these places...

The main reason I used it was to be able to see what syntax is generated by the line. I wonder, maybe a comment with target syntax above the relevant code would be a good idea?

src/Xamarin.Android.Build.Tasks/Utilities/LlvmIrGenerator/LlvmIrGenerator.cs
```diff
@@ -476,7 +477,12 @@ bool MaybeWritePreAllocatedBuffer<T> (StructureInfo<T> info, StructureMemberInfo
 
 			WriteGlobalSymbolStart (variableName, preAllocatedBufferVariableOptions, output);
 			ulong size = bufferSize * smi.BaseTypeSize;
-			output.WriteLine ($"[{bufferSize} x {irType}] zeroinitializer, align {GetAggregateAlignment ((int)smi.BaseTypeSize, size)}");
+			output.Write ('[');
+			output.Write (bufferSize);
+			output.Write (" x ");
+			output.Write (irType);
+			output.Write ("] zeroinitializer, align ");
+			output.WriteLine (GetAggregateAlignment ((int) smi.BaseTypeSize, size));
```
jonathanpeppers, line 485: For this line, for example, I could go through and add a comment like:
```csharp
// $"[{bufferSize} x {irType}] zeroinitializer, align {GetAggregateAlignment ((int)smi.BaseTypeSize, size)}"
```
I'll go through and do that, but let's see if the PR has any failing tests or anything.

Issue #7622, opened 2022-12-08
Localization via satellite assemblies doesn't work for android when the app built in debug mode.


mattleibow:
---
**Issue moved from dotnet/maui#4470**
- Please respond to @nevse.

---

_From @nevse on Wednesday, February 2, 2022 12:39:22 PM_

### Description

If I run app in debug mode `dotnet build -t:Run -f:net6.0-android` it doesn't show translated message. For `dotnet build -t:Run -f:net6.0-android -c:Release` it does.
Project to reproduce - https://github.com/nevse/Maui-Tests-Localization1
![debugVsRelease](https://user-images.githubusercontent.com/1169410/152153995-2c4d55e6-f1b8-4f0a-ab33-86e8705cfd25.png)

### Steps to Reproduce

1. create app `dotnet new maui -n HumanaizerTests`
2. add references 
```
<ItemGroup>
        <PackageReference Include="Humanizer.Core" Version="2.14.1" />
        <PackageReference Include="Humanizer.Core.de" Version="2.14.1" /> 
</ItemGroup>
```
3. change culture to `de` and use Humanizer to read data localization resources
```
CultureInfo.CurrentCulture = CultureInfo.CurrentUICulture = new CultureInfo("de");

CounterLabel.Text = DateTime.UtcNow.AddHours(-2).Humanize();
```
4. run the app in debug `dotnet build -t:Run -f net6.0-android` and release mode `dotnet build -t:Run -f:net6.0-android -c:Release`

### Version with bug

Preview 12 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 11

mattleibow:
---
**Issue moved from dotnet/maui#4470**
- Please respond to @shimingsg.

---

_From @shimingsg on Thursday, February 17, 2022 10:10:53 AM_

Verified Repro with Android 11, Repro project
[4470.zip](https://github.com/dotnet/maui/files/8130305/4470.zip)

mattleibow:
---
**Issue moved from dotnet/maui#4470**
- Please respond to @AntonKosenkoDX.

---

_From @AntonKosenkoDX on Thursday, May 19, 2022 5:12:25 AM_

Hello, we are developing [DevExpress .NET MAUI controls](https://www.devexpress.com/maui/) and experienced the same issue. This issue has a Normal priority for us but we don't have a workaround. Thank you for your help.

mattleibow:
---
[...]

PR #7627, opened 2022-12-09
Add Nightly Tests for Humanizer 

dellis1972:
Tests https://github.com/xamarin/xamarin-android/issues/7622

Localisation assemblies were not being fast deployed. Commit https://github.com/xamarin/monodroid/commit/1f52d5873960cf72685b9b08dbe262fc1ad4ca13
fixed that. However we need to add a unit test to make sure that these files are deployed and 
actually produce the required results. 

So lets use Humanizer to handle the localisation. This PR adds a test which checks that the 
app on the device outputs the expected text which Humanizer is supposed to generate. 
It also updates existing packaging tests to make sure that the required satellite assemblies are
 packaged.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
@@ -73,6 +73,25 @@ namespace Xamarin.Android.Build.Tests
 			};
 			proj.SetProperty ("AndroidUseAssemblyStore", usesAssemblyStores.ToString ());
 			proj.SetAndroidSupportedAbis ("armeabi-v7a");
+			proj.PackageReferences.Add (new Package {
+				Id = "Humanizer.Core",
+				Version = "2.14.1",
+			});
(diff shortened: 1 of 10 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/PackagingTest.cs
```diff
@@ -74,14 +74,33 @@ public void CheckIncludedAssemblies ([Values (false, true)] bool usesAssemblySto
 			};
 			proj.SetProperty ("AndroidUseAssemblyStore", usesAssemblyStores.ToString ());
 			proj.SetAndroidSupportedAbis ("armeabi-v7a");
+			proj.PackageReferences.Add (new Package {
+				Id = "Humanizer.Core",
+				Version = "*",
```
jonathanpeppers, line 79: We should maybe pick a version for all these for now? I could see Humanizer releasing an update and our tests could randomly start failing. `2.14.1` is on `dotnet-public` right now.

dellis1972:
Will need https://github.com/xamarin/xamarin-android/pull/7925

grendello (approved):

PR #7637, opened 2022-12-14
Add Unit Test for testOnly apps

dellis1972:
Fixes: https://github.com/dotnet/maui/issues/11345

For context users can mark their package as "testOnly" in the `AndroidManifest.xml`. When they do this `adb` also needs an additonal flag to install the package. 

PR https://github.com/xamarin/monodroid/pull/1279 adds the ability to pass this additional flag to `adb`. 
This PR adds the ability to read the flag from the `AndroidManifest.xml`, so it can be used by other tasks. 
A unit test has also been added to the Device tests to make sure we can install `testOnly` packages.


dellis1972:
https://github.com/xamarin/monodroid/pull/1279 is green

dellis1972:
Requires https://github.com/xamarin/monodroid/pull/1279

jonathanpeppers (approved):

PR #7642, opened 2022-12-15
[Xamarin.Android.Build.Tasks] fix cases of missing `@(Reference)`

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/10154

If you have a solution setup with:

* `ApplicationA` project reference ->
* `LibraryB` reference ->
* `LibraryC`

The app will crash at runtime, due to a missing `LibraryC.dll`.

You can solve the problem by changing `@(Reference)` to a `@(ProjectReference)`. However, it appears the same situation works in a .NET 7 self-contained console app:

    dotnet publish --self-contained -r win-x64
    ...
    ls -n .\bin\Debug\net7.0\win-x64\publish\LibraryC.dll
    LibraryC.dll

The underlying issue appears to be:

https://github.com/dotnet/msbuild/blob/a2490dd3f78cce4abc8f9e6f1b5268437332818f/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2322

In the console app, `$(BuildingProject)` is `true` and `$(_FindDependencies)` is empty.

In the android app, `$(BuildingProject)` is `false` and `$(_FindDependencies)` is `false`.

It appears the `BuildOnlySettings` target *should* be running in Android apps when we do an "inner" build per `$(RuntimeIdentifier)`. Simply adding a dependency for the `_ComputeFilesToPublishForRuntimeIdentifiers` MSBuild target solves this issue.

This likely might fix other issues, such as:

https://github.com/dotnet/maui/issues/11364

I added a test for this scenario.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
+++ b/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
@@ -1972,7 +1972,8 @@ because xbuild doesn't support framework reference assemblies.
     DependsOnTargets="_ResolveAssemblies"
 >
   <ItemGroup>
-    <_AndroidResolvedSatellitePaths Include="@(ReferenceSatellitePaths)" />
+    <!-- In .NET 6+, the .NET SDK locates these files -->
+    <_AndroidResolvedSatellitePaths Condition=" '$(UsingAndroidNETSdk)' != 'true' " Include="@(ReferenceSatellitePaths)" />
     <!-- Satellites from the current project, see: https://github.com/microsoft/msbuild/blob/master/src/Tasks/Microsoft.Common.CurrentVersion.targets#L4283-L4299 -->
     <_AndroidResolvedSatellitePaths Include="@(IntermediateSatelliteAssembliesWithTargetPath->'$(OutDir)%(Culture)\$(TargetName).resources.dll')" />
   </ItemGroup>
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
@@ -44,7 +44,7 @@ _ResolveAssemblies MSBuild target.
   </PropertyGroup>
 
   <Target Name="_ComputeFilesToPublishForRuntimeIdentifiers"
-      DependsOnTargets="_FixupIntermediateAssembly;ResolveReferences;ComputeFilesToPublish;$(_RunAotMaybe)"
+      DependsOnTargets="BuildOnlySettings;_FixupIntermediateAssembly;ResolveReferences;ComputeFilesToPublish;$(_RunAotMaybe)"
       Returns="@(ResolvedFileToPublish)">
       <ItemGroup>
         <ResolvedFileToPublish Remove="@(_SourceItemsToCopyToPublishDirectory)" />
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
@@ -1054,6 +1054,67 @@ public abstract class Foo<TVirtualView, TNativeView> : ViewHandler<TVirtualView,
 		}
 
 		[Test]
+		public void ProjectDependencies ([Values(true, false)] bool projectReference)
+		{
+			// Setup dependencies App A -> Lib B -> Lib C
+			var path = Path.Combine ("temp", TestName);
+
+			var libB = new XASdkProject (outputType: "Library") {
+				ProjectName = "LibraryB",
+				IsRelease = true,
+			};
+			libB.Sources.Clear ();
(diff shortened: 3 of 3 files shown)


dellis1972 (approved):

jonathanpeppers:
The `MissingSatelliteAssemblyInLibrary` test fails with:
```
(_BuildApkEmbed target) -> 
error XABBA7009: System.InvalidOperationException: Duplicate assemblies encountered
error XABBA7009:    at Xamarin.Android.Tasks.AssemblyStore.WriteIndex(BinaryWriter blobWriter, StreamWriter manifestWriter, List`1 globalIndex)
error XABBA7009:    at Xamarin.Android.Tasks.AssemblyStore.WriteIndex(BinaryWriter blobWriter, String manifestPath, List`1 globalIndex)
error XABBA7009:    at Xamarin.Android.Tasks.AssemblyStore.WriteIndex(List`1 globalIndex)
error XABBA7009:    at Xamarin.Android.Tasks.AssemblyStoreGenerator.Generate(String outputDirectory)
error XABBA7009:    at Xamarin.Android.Tasks.BuildApk.AddAssemblies(ZipArchiveEx apk, Boolean debug, Boolean compress, IDictionary`2 compressedAssembliesInfo, String assemblyStoreApkName)
error XABBA7009:    at Xamarin.Android.Tasks.BuildApk.ExecuteWithAbi(String[] supportedAbis, String apkInputPath, String apkOutputPath, Boolean debug, Boolean compress, IDictionary`2 compressedAssembliesInfo, String assemblyStoreApkName)
error XABBA7009:    at Xamarin.Android.Tasks.BuildApk.RunTask()
error XABBA7009:    at Microsoft.Android.Build.Tasks.AndroidTask.Execute() in /Users/builder/azdo/_work/1/s/xamarin-android/external/xamarin-android-tools/src/Microsoft.Android.Build.BaseTasks/AndroidTask.cs:line 17
```
So need to investigate that.

jonpryor:
Draft commit message:

```markdown
Fixes: https://github.com/dotnet/maui/issues/10154

Context? https://github.com/dotnet/maui/issues/11364

If you have a solution setup with:

  * `ApplicationA.csproj` has a `@(ProjectReference)` to
    `LibraryB.csproj`.
  * `LibraryB.csproj` which has a `@(Reference)` to `LibraryC.dll`,
    built by-
  * `LibraryC.csproj`

The app will crash at runtime, due to a missing `LibraryC.dll`.

The workaround is for `LibraryB.csproj` to use `@(ProjectReference)`
to `LibraryC.csproj` instead of `@(Reference)` to `LibraryC.dll`.

However, it appears the same situation works in a .NET 7
self-contained console app:

	% dotnet publish --self-contained -r win-x64
	…
	% ls -1 .\bin\Debug\net7.0\win-x64\publish\LibraryC.dll
	LibraryC.dll

The underlying issue appears to be due to [`$(_FindDependencies )`][0]:

	<_FindDependencies Condition="'$(BuildingProject)' != 'true' and '$(_ResolveReferenceDependencies)' != 'true'">false</_FindDependencies>

In the console app, `$(BuildingProject)`=true and
`$(_FindDependencies)` is empty.

In the Android app, `$(BuildingProject)`=false` and
`$(_FindDependencies)` is `false`.

It appears that the `BuildOnlySettings` target *should* be running in
Android apps when we do an "inner" build per `$(RuntimeIdentifier)`.
Simply updating `_ComputeFilesToPublishForRuntimeIdentifiers` so that
the `BuildOnlySettings` target is in `DependsOnTargets` fixes this.
However, this also causes satellite assemblies to now be automatically
found by the .NET SDK.  Update `@(_AndroidResolvedSatellitePaths)` so
that `@(ReferenceSatellitePaths)` is only included on Classic builds,
preventing duplicate entries.

[0]: https://github.com/dotnet/msbuild/blob/a2490dd3f78cce4abc8f9e6f1b5268437332818f/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2322
```

jonpryor:
@jonathanpeppers: given that `BuildOnlySettings` is now built by `_ComputeFilesToPublishForRuntimeIdentifiers `, 
does this suggest that `$(SignAndroidPackageDependsOn)` does *not* need to mention `BuildOnlySettings` anymore? i.e. can we/should we "partially revert" the `Xamarin.Android.Common.targets` change in d298fdaadab1a80bc45f9334d946c0ae8b568c42?

jonathanpeppers:
No, we need `BuildOnlySettings` to run twice. This one is for the "inner" build per runtime identifier.

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -1972,7 +1972,8 @@ because xbuild doesn't support framework reference assemblies.
     DependsOnTargets="_ResolveAssemblies"
 >
   <ItemGroup>
-    <_AndroidResolvedSatellitePaths Include="@(ReferenceSatellitePaths)" />
+    <!-- In .NET 6+, the .NET SDK locates these files -->
```
dellis1972, line 1975: nitpick: Should this be .NET 8+ since we are not releasing this fix for .NEt 6?

jonathanpeppers: I think we should actually send this to 7, could also consider 6?

dellis1972: Can we put stuff in 6 still?

PR #7643, opened 2022-12-15
[Xamarin.Android.Build.Tasks] downgrade d8/r8 `warning` messages to `info`

jonathanpeppers:
Fixes: https://github.com/dotnet/maui/issues/10901

Building a .NET MAUI project such as:

    dotnet new maui
    dotnet build -c Release -f net6.0-android -p:AndroidLinkMode=r8

Results in build warnings from R8 such as:

    R8 : warning : Missing class androidx.window.extensions.WindowExtensions

In 92bc7050, we already had attempted to silence warnings from R8 such as:

    R8 : warning : Resource 'META-INF/MANIFEST.MF' already exists.

At the time, there was no option to hide warnings, now there is!

    --map-diagnostics[:<type>] <from-level> <to-level>
        # Map diagnostics of <type> (default any) reported as
        # <from-level> to <to-level> where <from-level> and
        # <to-level> are one of 'info', 'warning', or 'error' and the
        # optional <type> is either the simple or fully qualified
        # Java type name of a diagnostic. If <type> is unspecified,
        # all diagnostics at <from-level> will be mapped.
        # Note that fatal compiler errors cannot be mapped.

We can pass:

    --map-diagnostics warning info

Which can be done in existing apps via:

    <AndroidD8ExtraArguments>--map-diagnostics warning info</AndroidD8ExtraArguments>
    <AndroidR8ExtraArguments>--map-diagnostics warning info</AndroidR8ExtraArguments>

To solve this problem, let's create a new `$(AndroidD8IgnoreWarnings)` MSBuild property and make use of one we already have `$(AndroidR8IgnoreWarnings)`:

    <ItemGroup>
      <_AndroidD8MapDiagnostics Condition=" '$(AndroidD8IgnoreWarnings)' == 'true' " Include="warning" To="info" />
      <_AndroidR8MapDiagnostics Condition=" '$(AndroidR8IgnoreWarnings)' == 'true' " Include="warning" To="info" />
    </ItemGroup>

We can then use these item groups to pass `--map-diagnostics` as appropriate. Developers can turn off either property to disable this behavior.

We can then remove the weird `Regex` code added in 92bc7050. The existing tests from 92bc7050 should also be sufficient for testing this change.

Now r8 emits:

    Info: Missing class androidx.window.extensions.WindowExtensions (referenced from: androidx.window.extensions.layout.WindowLayoutComponent androidx.window.layout.SafeWindowLayoutComponentProvider$windowLayoutComponent$2.invoke())
    Info: Resource 'META-INF/MANIFEST.MF' already exists.
    
And you don't get an MSBuild warning anymore.

### DIFF
--- a/Documentation/guides/building-apps/build-properties.md
+++ b/Documentation/guides/building-apps/build-properties.md
@@ -1110,6 +1110,14 @@ or use `-p:AndroidCreateProguardMappingFile=False` on the command line.
 
 This property was added in Xamarin.Android 11.2.
 
+## AndroidD8IgnoreWarnings
+
+Specifies `--map-diagnostics warning info` to be passed to `d8`. The
+default value is `True`, but can be set to `False` to enforce more
+strict behavior. See the [D8 and R8 readme][r8-source] for details.
+
+Added in .NET 8.
+
 ## AndroidR8IgnoreWarnings
 
 Specifies
@@ -1118,7 +1126,12 @@ to continue with dex compilation even if certain warnings are
 encountered. The default value is `True`, but can be set to `False` to
 enforce more strict behavior. See the [ProGuard manual](https://www.guardsquare.com/manual/configuration/usage) for details.
 
+Starting in .NET 8, specifies `--map-diagnostics warning info`. See
+the [D8 and R8 readme][r8-source] for details.
+
 Added in Xamarin.Android 10.3.
+
+[r8-source]: https://r8.googlesource.com/r8/
 
 ## AndroidR8JarPath
 
--- a/Documentation/guides/building-apps/build-properties.md
+++ b/Documentation/guides/building-apps/build-properties.md
@@ -1114,7 +1114,7 @@ This property was added in Xamarin.Android 11.2.
 
 Specifies `--map-diagnostics warning info` to be passed to `d8`. The
 default value is `True`, but can be set to `False` to enforce more
-strict behavior. See the [D8 and R8 readme][r8-source] for details.
+strict behavior. See the [D8 and R8 source code][r8-source] for details.
 
 Added in .NET 8.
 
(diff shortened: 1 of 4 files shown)


jonathanpeppers:
The one test failure looks unrelated:
```
dotnet run should succeed
Expected: True
But was:  False
...
[BT Error retrieving device SDK version. Please try again. [/Users/runner/work/1/a/TestRelease/12-15_23.39.44/temp/DotNetInstallAndRunTrueTruenet6.0-android/UnnamedProject.csproj]
```

Documentation/guides/building-apps/build-properties.md
```diff
@@ -1118,8 +1126,13 @@ to continue with dex compilation even if certain warnings are
 encountered. The default value is `True`, but can be set to `False` to
 enforce more strict behavior. See the [ProGuard manual](https://www.guardsquare.com/manual/configuration/usage) for details.
 
+Starting in .NET 8, specifies `--map-diagnostics warning info`. See
+the [D8 and R8 readme][r8-source] for details.
+
 Added in Xamarin.Android 10.3.
 
+[r8-source]: https://r8.googlesource.com/r8/
```
jonpryor, line 1134: Is this the correct link for "details" on `--map-diagnostics`? I do not see anything mentioning `--map-diagnostics`, or even a complete list of supported command-line arguments, at this URL.

jonathanpeppers: No, I can't find anything about it, except reading source code and the `--help` output...

jonpryor: Then a link to the source code with the `--help` output may be better than a link to the "root" of r8.

dellis1972 (changes requested):
Do we have a unit test which convers this to make sure the warnings are mapped to info?

jonathanpeppers:
The test from https://github.com/xamarin/xamarin-android/commit/92bc70508042af779a7cfa1784115304e5a5dc3b, it previously had build warnings like this when enabling multi-dex:

    R8 : warning : Resource 'META-INF/MANIFEST.MF' already exists.

It asserts there are 0 warnings: https://github.com/xamarin/xamarin-android/blob/d4c2a5e2ba148ea1d63fdc7f8c984642e7b5b369/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/BuildTest2.cs#L282

Now it should log a message instead:

    Info: Resource 'META-INF/MANIFEST.MF' already exists.

dellis1972 (approved):

Issue #7639, opened 2022-12-15
Android - APT2000 build error when maui-blazor


skoc10:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

.NET 7.0.100

### Description

```C:\Program Files\dotnet\packs\Microsoft.Android.Sdk.Windows\33.0.4\tools\Xamarin.Android.Aapt2.targets(212,3): error APT2000: The system cannot find the file specified. (2). [D:\ci\Jenkins\workspace\abp-commercial-nightly\volo\abp\templates\app-pro\aspnet-core\src\MyCompanyName.MyProjectName.MauiBlazor\MyCompanyName.MyProjectName.MauiBlazor.csproj::TargetFramework=net7.0-android]```

<img width="1237" alt="image" src="https://user-images.githubusercontent.com/64414348/207810075-46624dfe-402a-440a-a7a3-bc449862a547.png">

I am facing this error when build Maui-Blazor project.
I updated all related SDK but problem continues.
This issue is not related https://github.com/xamarin/xamarin-android/issues/6925.

### Steps to Reproduce

dotnet new maui-blazor
dotnet build

**Comment by dellis1972** [comment]
@skoc10 

As per the other bug can you add the following to your project

```
<AndroidAapt2LinkExtraArgs>$(AndroidAapt2LinkExtraArgs) -v</AndroidAapt2LinkExtraArgs>
```

then build (from clean) on the command line again and post the logs?  We need the binlog file
so you need to run

```
dotnet build MyCompanyName.MyProjectName.MauiBlazor.csproj -bl
```

This will give us more verbose `aapt2` output. It should help narrow down which file is causing the error.

On a side note, the path you have `D:\ci\Jenkins\workspace\abp-commercial-nightly\volo\abp\templates\app-pro\aspnet-core\src\MyCompanyName.MyProjectName.MauiBlazor\MyCompanyName.MyProjectName.MauiBlazor.csproj` might cause some issues on windows. Windows has a max command line length which you might hit when building release builds with aot.

skoc10:
[msbuild.binlog.zip](https://github.com/xamarin/xamarin-android/files/10236795/msbuild.binlog.zip)

dellis1972:
@skoc10 

I'm not seeing anyting obvious in the logs. `aapt2` is raising that error, but its not reporting which file it cannot find. 
I will try to repo here (not luck so far). 

[...]

PR #7644, opened 2022-12-16
Handle `aapt2` error "The system cannot find the file specified." slightly better

dellis1972:
Context #7639

If a project has a long path name this can cause `aapt2` to raise a `The system cannot find the file specified` error. 
This is not very helpful since there is no obvious reason to the user as to WHY the build is failing. So we should do better.
So lets catch that specific error message and report a helpful, actionable message instead. In this case we will advise the
user that the path is probably too long and they should consult the documentation on how to resolve it. 

As for resolutions the documentation includes a number of options so it should cover all use cases. Generally these 
are 

1.  Move the project closer to the root of the drive.
2. Enable long path support in windows.
3. Make changes to the `$(BaseIntermediateOutputPath)` MSbuild property to move only the `obj` folder closer to the root of the drive. 

Most users can get away with 1. But with Maui and blazor based apps this might not solve the problem as the project itself might have a deep folder structure. So options 2. and 3. are available to cover those senarios.

### DIFF
--- a/Documentation/guides/messages/apt2264.md
+++ b/Documentation/guides/messages/apt2264.md
@@ -1,0 +1,53 @@
+---
+title: Xamarin.Android error APT2264
+description: APT2264 error code
+ms.date: 12/16/2022
+---
+# Xamarin.Android error APT2264
+
+## Issue
+
+The tool `aapt2` is unable to resolve one of the files it was passed.
+This is generally caused by the path being longer than the Maximum Path
+length allowed on windows.
+
+## Solution
+
+The best way to avoid this is to ensure that your project is not located
+deep in the folder structure. For example if you create all of your
+projects in folders such as
+
+`C:\Users\shelly\Visual Studio\Android\MyProjects\Com.SomeReallyLongCompanyName.MyBrillantApplication\MyBrilliantApplicaiton.Android\`
+
+you may well encounter problems with not only `aapt2` but also Ahead of Time
+compilation. Keeping your project names and folder structures short and
+concise will help work around these issues. For example instead of the above
+you could use
+
+`C:\Work\Android\MyBrilliantApp`
+
+Which is much shorter and much less likely to encounter path issues.
+
+However this is no always possible. Sometimes a project or a environment requires
+deep folder structures. In this case changing the location of the `$(BaseIntermediateOutputPath)`
+can solve these problems. In order for this to work the setting MUST be changed
+before ANY build or restore occurs. To do this you can make use of the MSBuild
+`Directory.Build.props` support.
+
+Creating a `Directory.Build.props` file in your solution or project directory which
+redefines the `$(BaseIntermediateOutputPath)` to somewhere nearer the root of the drive
+with solve these issues. Adding a file with the following contents will create the `obj`
+directory in a different location of your choosing.
+
+```
+<Project>
+  <PropertyGroup>
+      <BaseIntermediateOutputPath Condition=" '$(OS)' == 'Windows_NT' ">C:\Intermediate\$(ProjectName)</BaseIntermediateOutputPath>
+      <BaseIntermediateOutputPath Condition=" '$(OS)' != 'Windows_NT' ">/tmp/Intermediate/$(ProjectName)</BaseIntermediateOutputPath>
+  </PropertyGroup>
+</Project
+```
+
+Using this technique will reduce the lengths of the paths sent to the various tools like `aapt2`.
+Note this is generally only a Windows issue. So there is no need to override the `$(BaseIntermediateOutputPath)`
+on Mac or Linux based environments. However you might want to override everywhere to be consistent.
--- a/Documentation/guides/messages/apt2264.md
+++ b/Documentation/guides/messages/apt2264.md
@@ -29,10 +29,15 @@ you could use
 Which is much shorter and much less likely to encounter path issues.
 
 However this is no always possible. Sometimes a project or a environment requires
-deep folder structures. In this case changing the location of the `$(BaseIntermediateOutputPath)`
-can solve these problems. In order for this to work the setting MUST be changed
-before ANY build or restore occurs. To do this you can make use of the MSBuild
-`Directory.Build.props` support.
+deep folder structures. In this case enabling long path support in Windows *might*
+be enough to get your project working. Details on how to do this can be found
+[here](https://learn.microsoft.com/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later).
+
+
+If long path support does not work changing the location of the
+`$(BaseIntermediateOutputPath)` can help solve these problems. In order for this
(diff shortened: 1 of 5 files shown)


Documentation/guides/messages/apt2264.md
```diff
@@ -0,0 +1,53 @@
...
+---
+# Xamarin.Android error APT2264
+
+## Issue
+
+The tool `aapt2` is unable to resolve one of the files it was passed.
+This is generally caused by the path being longer than the Maximum Path
+length allowed on windows.
+
+## Solution
```
jonathanpeppers, line 14: Can we suggest enabling long paths as the 1st solution?

https://learn.microsoft.com/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later

jonathanpeppers (approved):
Otherwise, the actual code changes look good. 👍

Documentation/guides/messages/apt2264.md
jonpryor: As per the docs, *two* things must be done: the registry must be updated to set `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem\LongPathsEnabled` -- and I wonder if that could/would break anything -- and ***secondly***:

> * The [application manifest](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests) must also include the `longPathAware` element.
> ```xml
> <application xmlns="urn:schemas-microsoft-com:asm.v3">
>     <windowsSettings xmlns:ws2="http://schemas.microsoft.com/SMI/2016/WindowsSettings">
>         <ws2:longPathAware>true</ws2:longPathAware>
>     </windowsSettings>
> </application>
> ```

We do not currently ship an `aapt2.exe.config` manifest file.

Additionally, it may be moot; the next section in docs states:

> These are the directory management functions that no longer have MAX_PATH restrictions if you opt-in to long path behavior: CreateDirectoryW, CreateDirectoryExW GetCurrentDirectoryW RemoveDirectoryW SetCurrentDirectoryW.
>
> These are the file management functions that no longer have MAX_PATH restrictions if you opt-in to long path behavior: CopyFileW, CopyFile2, CopyFileExW, CreateFileW, CreateFile2, CreateHardLinkW, CreateSymbolicLinkW, DeleteFileW, FindFirstFileW, FindFirstFileExW, FindNextFileW, GetFileAttributesW, GetFileAttributesExW, SetFileAttributesW, GetFullPathNameW, GetLongPathNameW, MoveFileW, MoveFileExW, MoveFileWithProgressW, ReplaceFileW, SearchPathW, FindFirstFileNameW, FindNextFileNameW, FindFirstStreamW, FindNextStreamW, GetCompressedFileSizeW, GetFinalPathNameByHandleW.

Are these used?

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\MSBuild\Xamarin\Android> dumpbin /imports  aapt2.exe  | findstr File
File Type: EXECUTABLE IMAGE
                   8C CreateFileMappingA
                   90 CreateFileMappingW
                  1E9 GetFileAttributesA
                  1EE GetFileAttributesW
                  1F0 GetFileInformationByHandle
                  1F5 GetFileSizeEx
                  1F7 GetFileType
                  216 GetModuleFileNameA
                  27E GetSystemTimeAsFileTime
                  362 MapViewOfFile
                  3CF ReadFile
                  43B SetEndOfFile
                  44D SetFilePointer
                  4BD UnmapViewOfFile
```

Odd is `GetFileAttributesA`, which presumably has `MAX_PATH` limits, but `GetFileAttributesW` is *also* used, so when is `GetFileAttributesA` used vs. `GetFileAttributesW`? ¯\\_(ツ)_/¯.

`GetModuleFileNameA` is also concerning, so it would not surprise me if `aapt2.exe` didn't work with long filenames even if we set the registry entry *and* provided an appropriate `aapt2.exe.config` manifest.

This requires testing, but I suspect that there's no point in suggesting the enabling of long paths, as `aapt2` won't work with them.

grendello: `aapt2` [uses](https://android.googlesource.com/platform/frameworks/base/+/refs/tags/android-13.0.0_r24/tools/aapt2/util/Files.cpp#52) this [function](https://android.googlesource.com/platform/system/libbase/+/refs/tags/android-13.0.0_r24/utf8.cpp#162) to convert UTF-8 file names to Windows Unicode

The android base library also [has wrappers](https://android.googlesource.com/platform/system/libbase/+/refs/tags/android-13.0.0_r24/utf8.cpp#181) around the standard C I/O functions which add support for UTF-8 strings.

jonpryor: This strongly implies that if we provide an appropriate `aapt2.exe.config`, we can get long file name support *if*/when the registry setting is enabled.

jonathanpeppers: Maybe we should just check if you need it? My memory in the past, I just changed the registry key and everything worked.

But I don't remember if I was hitting an issue with `aapt2` or a .NET process (which would probably opt into long paths).

jonpryor: @dellis1972: if this PR were "green" and otherwise ready to merge, I think `aapt2.exe.config` could be done as a separate PR.  Given that this PR *isn't* green and ready to merge, might as well test locally and see if that works?

jonathanpeppers: I think you might have to test an "almost long path", where only aapt2 would get long path errors?

dellis1972: So I did some more testing and it seems that `aapt2` was able to switch to log paths with only the registry setting change (not aapt2.exe.config file). 

I know this because the diagnostic output was putting the `\\?\` stuff at the case of the filenames. So I don't think we need the config file. I will update the docs to mention the registry change though.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -530,6 +530,10 @@ Please change the value to an assembly-qualifed type name which inherits from '{
     <comment>The following are literal names and should not be translated: Java.Interop.DoNotPackageAttribute
 {0} - The assembly name</comment>
   </data>
+  <data name="APT2264" xml:space="preserve">
+    <value>This is probably caused by the project exceeding the Max Path length. See https://learn.microsoft.com/en-us/xamarin/android/errors-and-warnings/apt2264 for details.</value>
```
jonathanpeppers, line 534: I think we are supposed to remove the `en-us` part from MS docs links. Also it seems like they call this `Maximum Path Length Limitation`, maybe we can stick with that name?

```suggestion
    <value>This is probably caused by the project exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
```

Is this page not there yet? Going to be a new page?

dellis1972:
@jonpryor This one is good to go I think.

Issue #7558, opened 2022-11-16
System.Obsolete attribute used instead of ObsoletedOSPlatform attribute


tipa:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

VS 2022 17.4.1, .NET 7

### Description

I am using this property to get the installed app version:
```
public static string VersionString => (OperatingSystem.IsAndroidVersionAtLeast(33) ?
   Application.Context.PackageManager.GetPackageInfo(PackageName, PackageManager.PackageInfoFlags.Of(0)) :
   Application.Context.PackageManager.GetPackageInfo(PackageName, 0)).VersionName;
```

The old ([now deprecated](https://developer.android.com/reference/android/content/pm/PackageManager#getPackageInfo(android.content.pm.VersionedPackage,%20int))) `GetPackageInfo` method still emits this warning, even though `OperatingSystem.IsAndroidVersionAtLeast` is used:
`Warning	CS0618	'PackageManager.GetPackageInfo(string, PackageInfoFlags)' is obsolete: 'deprecated'	
`

This is because of this attribute:
`[System.Obsolete("deprecated")]`

I think this attribute should be used instead:
`[ObsoletedOSPlatform("android30.0")]`

I know this is rather nitpicking, but there might be other usages of the old `Obsolete` attribute that could be replaced with the `ObsoletedOSPlatform` attribute. In the PackageManager class alone there are also these methods currently annotated with the old attribute:
https://developer.android.com/reference/android/content/pm/PackageManager#addPreferredActivity(android.content.IntentFilter,%20int,%20android.content.ComponentName[],%20android.content.ComponentName)
https://developer.android.com/reference/android/content/pm/PackageManager#getPackageGids(java.lang.String,%20int)
https://developer.android.com/reference/android/content/pm/PackageManager#removePackageFromPreferred(java.lang.String)
https://developer.android.com/reference/android/content/pm/PackageManager#addPackageToPreferred(java.lang.String)

### Steps to Reproduce

1. Create new .NET 6 project
2. Add property as shown above
3. Build app / check warnings window

jpobst:
`[ObsoletedOSPlatform]` was added in .NET 7, so if using `net7.0-android` this _should_ be working.  😁 

Fixed in https://github.com/xamarin/xamarin-android/issues/7234

tipa:
I am using `net7.0-android33.0` but still see the warning.
[...]

PR #7645, opened 2022-12-16
[api-merge] Fix for correctly calculating method level 'deprecated-since' attribute.

jpobst:
Context: https://github.com/xamarin/xamarin-android/issues/7558

Currently when `api-merge` loops through type members, its first check is to see if there is any existing member with the same name.  If there isn't, it can short-circuit, add the new member, and move to the next one.  If there is, we compare the existing API level member to the new API level one to see if it has been marked deprecated in this level, and if so we can add the `deprecated-since` attribute.

However, this initial existing member check was solely done on member *name*, not signature.  If the member is a method, a later check finds the exact matching method with the same signature.  We are performing the `deprecated-since` logic against the first method with a matching name rather than the method with the same signature.  Thus we are producing incorrect results.

This commit moves `UpdateDeprecatedSince` to be called after we have the method with the matching signature.

The `acceptable-breakages` changes reflect additional `[Obsolete]` attributes that are now being converted to `[ObsoletedOSPlatform ("android-XX.0")]` attributes.


jonpryor:
Draft commit message:

```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/7558

Context: https://github.com/xamarin/monodroid/commit/16338f99568db892f37f6b5ddf927c139fcd4f50
Context: a073d99a838d15a2e4977fda63ecb1064a214d49
Context: 938c349725b677f1fa291a94f12546ef3d3c5a60

Currently when `api-merge` loops through type members, its first check
is to see if there is any existing member with the same name.  If there
isn't, it can short-circuit, add the new member, and move to the next
one.  If there is, we compare the existing API level member to the new
API level one to see if it has been marked deprecated in this level,
and if so we can add the `deprecated-since` attribute.

However, this initial existing member check was solely done on member
*name*, not signature.  😱

If the member is a method, a later check finds the exact matching
method with the same signature.  We were performing the
`deprecated-since` logic against the first method with a matching name
rather than the method with the same signature.

For example, [`PackageManager.getPackageInfo(String, int)`][0], which
was deprecated in *API-33*.  Yet our binding uses `[Obsolete]`, meaning
it has been deprecated since API-21 or before!

	namespace Android.Content.PM;
	partial class PackageManager {
	    [global::System.Obsolete (@"deprecated")]
	    [Register ("getPackageInfo", "(Ljava/lang/String;I)Landroid/content/pm/PackageInfo;", "GetGetPackageInfo_Ljava_lang_String_IHandler")]
	    public override unsafe Android.Content.PM.PackageInfo? GetPackageInfo (string packageName, [global::Android.Runtime.GeneratedEnum] Android.Content.PM.PackageInfoFlags flags) => …
	}

This was wrong; it should instead have used `[ObsoletedOSPlatform]`:

	namespace Android.Content.PM;
	partial class PackageManager {
	    [global::System.Runtime.Versioning.ObsoletedOSPlatform ("android33.0)]
	    [Register ("getPackageInfo", "(Ljava/lang/String;I)Landroid/content/pm/PackageInfo;", "GetGetPackageInfo_Ljava_lang_String_IHandler")]
	    public override unsafe Android.Content.PM.PackageInfo? GetPackageInfo (string packageName, [global::Android.Runtime.GeneratedEnum] Android.Content.PM.PackageInfoFlags flags) => …
	}


Move the `UpdateDeprecatedSince()` invocations to be called *after*
we have the method with the matching signature.

The `acceptable-breakages` changes reflect additional `[Obsolete]`
attributes that are now being converted to
`[ObsoletedOSPlatform ("android-XX.0")]` attributes.

[0]: https://developer.android.com/reference/android/content/pm/PackageManager#getPackageInfo(java.lang.String,%20int)
```

jonpryor:
@jpobst: one thing that concerns me is that `tests/api-compatibility/acceptable-breakages-vReference-net8.0.txt` *doesn't* have an entry for `PackageManager.GetPackageInfo(string, PackageInfoFlags)`, which was #7558.

Shouldn't `PackageManager.GetPackageInfo()` be in `acceptable-breakages-vReference-net8.0.txt`?

jpobst:
I verified that the package CI is building does include the new attributes:

image

Looking at the reference assembly we use for `api-compat`, the methods are not marked as `[Obsolete]` at all.  Thus, this is simply an attribute addition and it will not show up in `api-compat`:

image

Our reference assembly must not have been updated since these methods were marked as deprecated.

PR #7646, opened 2022-12-16
[build] skip provisioning of Xamarin.Android on .NET lanes

jonathanpeppers:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7102687&view=logs&j=270ba80a-9de5-5c73-abb9-bb787532a07c&s=25cdcba1-3ecb-5490-0002-daf27c697525&t=306f7057-edf0-58b5-5004-8b3b311950d4&l=449

`provisionator` is a source of random failures like:

    [21:41:23 ERR] Error Xamarin.Provisioning.ProvisioningException: Failed to provision C:\a\_work\1\s\xamarin-android\Xamarin.Android.Sdk-13.1.99.88.vsix

In this case, we were downloading a classic Xamarin.Android build for use on a test lane for `One .NET` tests?

To make things better, let's skip this step completely for `One .NET` tests. We shouldn't need to provision classic Xamarin.Android on these test lanes.

This also has the added benefit of faster test runs, just spot-checking the time spent on the above build:

* Saves 1-2 min of time for `provisionator` on macOS lanes.
* Saves up to ~24 min of time for  `provisionator` on Windows lanes.

### DIFF
--- a/build-tools/automation/azure-pipelines.yaml
+++ b/build-tools/automation/azure-pipelines.yaml
@@ -613,6 +613,7 @@ stages:
     steps:
     - template: yaml-templates/setup-test-environment.yaml
       parameters:
+        provisionClassic: false
         provisionatorChannel: ${{ parameters.provisionatorChannel }}
 
     - template: yaml-templates/run-xaprepare.yaml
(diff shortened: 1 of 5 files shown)


jonathanpeppers:
Seems like it is working, One .NET lanes:

image

VS legacy lanes:

image

PR #7648, opened 2022-12-20
[build] pack .NET `.nupkg` files in parallel

jonathanpeppers:
Previously, we were calling `dotnet pack` serially for many projects, such as:

    <Exec Command="&quot;$(DotNetPreviewTool)&quot; pack ... &quot;$(MSBuildThisFileDirectory)Microsoft.Android.Ref.proj&quot;" />

This ends up taking a while:

    Target CreateAllPacks 52.127s
        Task Exec 3.712s
        Task Exec 3.569s
        Task Exec 3.432s
        Task Exec 3.434s
        Task Exec 24.741s
        Task Exec 7.257s
        Task Exec 2.507s
        Task Exec 3.468s

Instead, we can create a `@(_ProjectsToPack)` item group, set `%(AdditionalProperties)`, and call the `<MSBuild/>` task:

    <MSBuild
        Projects="@(_ProjectsToPack)"
        Targets="Pack"
        Properties="@(_GlobalProperties)"
        BuildInParallel="$(BuildInParallel)"
    />

This is faster because:

* We don't shell out to a new `dotnet.exe` each time.

* `<MSBuild/>` task knows how to run in parallel, managing multiple MSBuild nodes.

This also has the benefit of including these `dotnet pack` commands in the `.binlog`. Previously, we would just have console output for them.

After these changes, I instead see:

    Target CreateAllPacks 5.857s
        Task MSBuild 5.857s

From my numbers before, I am unsure why one of the `dotnet pack` commands took ~24 seconds. But I believe this change should reduce the time to the *slowest* individual `dotnet pack` now.

It could save up to ~1 min of time in the `CreateAllPacks` target.

### DIFF
--- a/build-tools/create-packs/Directory.Build.targets
+++ b/build-tools/create-packs/Directory.Build.targets
@@ -46,9 +46,9 @@
 
   <Target Name="_SetGlobalProperties">
     <ItemGroup>
-      <_GlobalProperties Include="-p:Configuration=$(Configuration)" />
-      <_GlobalProperties Include="-p:NuGetLicense=$(NuGetLicense)" />
-      <_GlobalProperties Include="-p:IncludeSymbols=False" />
+      <_GlobalProperties Include="Configuration=$(Configuration)" />
+      <_GlobalProperties Include="NuGetLicense=$(NuGetLicense)" />
+      <_GlobalProperties Include="IncludeSymbols=False" />
     </ItemGroup>
   </Target>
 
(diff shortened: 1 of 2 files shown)


jonathanpeppers:
I might come back to this later, this hits:

    (_GenerateMsxDocXmls target) -> 
         C:\a\_work\1\s\build-tools\installers\create-installers.targets(55,5): error MSB3073: The command " "C:\a\_work\1\s\bin\Release\lib\packs\Microsoft.Android.Sdk.Windows\34.0.0\\tools\mdoc.exe" --debug export-msxdoc -o "C:\a\_work\1\s\bin\Release\lib\xamarin.android\xbuild-frameworks\MonoAndroid\v13.0\Mono.Android.xml" "C:\a\_work\1\s\external\android-api-docs\docs\Mono.Android\en"" exited with code 1. [C:\a\_work\1\s\build-tools\create-packs\Microsoft.Android.Sdk.proj]

Issue #7629, opened 2022-12-10
Wrong exception HttpClient when SSL certificate is expired


rgroenewoudt:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

VS2022 17.4.2

### Description

When doing a HTTP request with expired certificate, using HttpClient in .NET 7 Android app, the wrong exception type returned.
Expected: HttpRequestException. Returned: WebException.

Documentation of HttpClient.SendAsync indicates HttpRequestException or TaskCanceledException should be thrown.

### Steps to Reproduce

```
var httpClient = new HttpClient();
try
{
	var request = new HttpRequestMessage(HttpMethod.Get, "https://expired.badssl.com/");
	var response = await httpClient.SendAsync(request);
}
catch(HttpRequestException ex)
{
	Console.WriteLine("SUCCESS HttpRequestMessage: " + ex);
}
catch(Exception ex)
{
	Console.WriteLine("FAIL " + ex);
}
```
in .NET 7 Android app using native HTTP handler.

No issues when using SocketsHttpHandler.

## Expected behavior

```
System.Net.Http.HttpRequestException: The SSL connection could not be established, see inner exception.
 ---> System.Security.Authentication.AuthenticationException: Authentication failed, see inner exception.
```

## Actual behavior

```
System.Net.WebException: Unacceptable certificate: CN=COMODO RSA Certification Authority, O=COMODO CA Limited, L=Salford, ST=Greater Manchester, C=GB
---> Javax.Net.Ssl.SSLHandshakeException: Unacceptable certificate: CN=COMODO RSA Certification Authority, O=COMODO CA Limited, L=Salford, ST=Greater Manchester, C=GB
```
[AndroidApp7.zip](https://github.com/dotnet/runtime/files/10193918/AndroidApp7.zip)

It would be nice if the returned exception matches SocketHttpHandler by returning HttpRequestException with inner AuthenticationException and then Javax.Net.Ssl.SSLHandshakeException as inner.

Note that WebClient returns the same exception and that should keep working.

### Did you find any workaround?

Simply use broader catch.

PR #7661, opened 2023-01-03
Wrap connection exceptions in HttpRequestException

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/7629

Whenever a Java backend connection fails for any reason, we wrap the 
thrown exception in `HttpRequestException` as described in the 
`HttpClient.SendAsync` documentation.


simonrozsival:
Shouldn't we also fix the other `WebException`s that are thrown from AndroidMessageHandler? (for example https://github.com/xamarin/xamarin-android/blob/main/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs#L428-L436)

grendello:
@simonrozsival I don't want to touch these. We've had a lot of issues filed in the past regarding exception types, I don't want to revisit these unless I absolutely have to... :)

grendello:
@simonrozsival let's leave the legacy Xamarin as it is... :)

PR #7662, opened 2023-01-03
Fix ServerCertificateValidationCallback after redirect

### LINKED ISSUE #7650: AndroidMessageHandler - RemoteCertificateNameMismatch after redirect
### Android application type

Classic Xamarin.Android (MonoAndroid12.0, etc.), Android for .NET (net6.0-android, etc.)

### Affected platform version

Xamarin Android 13, .NET 7.0.100

### Description

`AndroidMessageHandler` incorrectly validates SSL certificate after a redirect to another domain if and only if `ServerCertificateCustomValidationCallback` is set.
Consider the following code:

```c#
// Redirects to facebook.com
const string url = "https://cutt.ly/n0HotDr";
using var handler = new AndroidMessageHandler()
{
    ServerCertificateCustomValidationCallback = (message, certificate, chain, errors) => errors == SslPolicyErrors.None
};
using var client = new HttpClient(handler);
await client.GetAsync(url);
```

When I put a breakpoint inside the lambda above, I can see that I hit it twice:
1. First time I see that the cutt.ly certificate is being validated and the `HttpRequestMessage.RequestUri` points to cutt.ly as well - that's expected.
2. Second time we validate a certificate for a domain we got redirected to (e.g. Facebook), but the `HttpRequestMessage.RequestUri` still points to cutt.ly -- and the `errors` parameter has `SslPolicyErrors.RemoteCertificateNameMismatch` value.

If I change `AndroidMessageHandler` to the managed handler, it works fine. Also, if I continue to use `AndroidMessageHandler`, but simply don't set `ServerCertificateCustomValidationCallback`, it works fine as well.
So it looks like the problem is somewhere inside the custom trust manager the handler uses when the validation callback is not null: https://github.com/xamarin/xamarin-android/blob/2a10e9696cf51a467db58e5c73a4fab764507482/src/Mono.Android/Xamarin.Android.Net/AndroidMessageHandler.cs#L1093 I believe the request URI here is not updated after a redirect, but it should: https://github.com/xamarin/xamarin-android/blob/2a10e9696cf51a467db58e5c73a4fab764507482/src/Mono.Android/Xamarin.Android.Net/ServerCertificateCustomValidator.cs#L85

### Steps to Reproduce

1. Create a Xamarin.Android app a .NET 6/7 Android app
2. Use `AndroidMessageHandler` to send a request that will be redirected to anotehr domain

__Expected__:
No issues
__Actual:__
An exception with the name mismatch

- Using a managed handler (not fully reliable with .NET 6/7 too).
- Not assigning `ServerCertificateCustomValidationCallback`. In my case I only use it in production builds to log the errors if there are any, so I was able to work around this with a `DelegatingHandler` and a few `try/catch`-es. For those who use the callback to apply some additional validation this may be a bigger deal.

### Relevant log output

```shell
Java.Security.Cert.CertificateException: Exception_WasThrown, Java.Security.Cert.CertificateException
   at Xamarin.Android.Net.ServerCertificateCustomValidator.TrustManager.CheckServerTrusted(X509Certificate[] , String )
   at Javax.Net.Ssl.IX509TrustManagerInvoker.n_CheckServerTrusted_arrayLjava_security_cert_X509Certificate_Ljava_lang_String_(IntPtr , IntPtr , IntPtr , IntPtr )
   at Android.Runtime.JNINativeWrapper.Wrap_JniMarshal_PPLL_V(_JniMarshal_PPLL_V , IntPtr , IntPtr , IntPtr , IntPtr )
  --- End of managed Java.Security.Cert.CertificateException stack trace ---
java.security.cert.CertificateException: The remote certificate was rejected by the provided RemoteCertificateValidationCallback.
	at xamarin.android.net.ServerCertificateCustomValidator_TrustManager.n_checkServerTrusted(Native Method)
	at xamarin.android.net.ServerCertificateCustomValidator_TrustManager.checkServerTrusted(ServerCertificateCustomValidator_TrustManager.java:42)
	at com.android.org.conscrypt.Platform.checkServerTrusted(Platform.java:254)
	at com.android.org.conscrypt.ConscryptEngine.verifyCertificateChain(ConscryptEngine.java:1644)
	at com.android.org.conscrypt.NativeCrypto.ENGINE_SSL_read_direct(Native Method)
	at com.android.org.conscrypt.NativeSsl.readDirectByteBuffer(NativeSsl.java:568)
	at com.android.org.conscrypt.ConscryptEngine.readPlaintextDataDirect(ConscryptEngine.java:1095)
	at com.android.org.conscrypt.ConscryptEngine.readPlaintextData(ConscryptEngine.java:1079)
	at com.android.org.conscrypt.ConscryptEngine.unwrap(ConscryptEngine.java:876)
	at com.android.org.conscrypt.ConscryptEngine.unwrap(ConscryptEngine.java:747)
	at com.android.org.conscrypt.ConscryptEngine.unwrap(ConscryptEngine.java:712)
	at com.android.org.conscrypt.ConscryptEngineSocket$SSLInputStream.processDataFromSocket(ConscryptEngineSocket.java:849)
	at com.android.org.conscrypt.ConscryptEngineSocket$SSLInputStream.access$100(ConscryptEngineSocket.java:722)
	at com.android.org.conscrypt.ConscryptEngineSocket.doHandshake(ConscryptEngineSocket.java:238)
	at com.android.org.conscrypt.ConscryptEngineSocket.startHandshake(ConscryptEngineSocket.java:217)
	at com.android.okhttp.internal.io.RealConnection.connectTls(RealConnection.java:196)
	at com.android.okhttp.internal.io.RealConnection.connectSocket(RealConnection.java:153)
	at com.android.okhttp.internal.io.RealConnection.connect(RealConnection.java:116)
	at com.android.okhttp.internal.http.StreamAllocation.findConnection(StreamAllocation.java:186)
	at com.android.okhttp.internal.http.StreamAllocation.findHealthyConnection(StreamAllocation.java:128)
	at com.android.okhttp.internal.http.StreamAllocation.newStream(StreamAllocation.java:97)
	at com.android.okhttp.internal.http.HttpEngine.connect(HttpEngine.java:302)
	at com.android.okhttp.internal.http.HttpEngine.sendRequest(HttpEngine.java:245)
	at com.android.okhttp.internal.huc.HttpURLConnectionImpl.execute(HttpURLConnectionImpl.java:465)
	at com.android.okhttp.internal.huc.HttpURLConnectionImpl.connect(HttpURLConnectionImpl.java:131)
	at com.android.okhttp.internal.huc.DelegatingHttpsURLConnection.connect(DelegatingHttpsURLConnection.java:90)
	at com.android.okhttp.internal.huc.HttpsURLConnectionImpl.connect(HttpsURLConnectionImpl.java:30)
```

simonrozsival:
Fixes #7650 

We check if the server certificate hostname matches the  request's RequestUri. When the request is redirected to a different host, the verification failed because we tested the next server's hostname against the original request URI. This PR keeps the RequestUri in sync with the redirect status to avoid this problem.

### DIFF
--- a/tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
+++ b/tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
@@ -111,6 +111,26 @@ namespace Xamarin.Android.NetTests
 			Assert.AreEqual (SslPolicyErrors.RemoteCertificateNameMismatch, reportedErrors & SslPolicyErrors.RemoteCertificateNameMismatch);
 		}
 
+		[Test]
+		public async Task ServerCertificateCustomValidationCallback_Redirects ()
+		{
+			int callbackCounter = 0;
+
+			var handler = new AndroidMessageHandler {
+				ServerCertificateCustomValidationCallback = (request, cert, chain, errors) => {
+					callbackCounter++;
+					Assert.Contains (request.RequestUri.Host, cert.SubjectName.Name);
+					return errors == SslPolicyErrors.None;
+				}
+			};
+
+			var client = new HttpClient (handler);
+			var result = await client.GetAsync ("https://httpbin.org/redirect-to?url=https://microsoft.com/");
+
+			Assert.AreEqual (3, callbackCounter); // httpbin.org, microsoft.com, www.microsoft.com
+			Assert.IsTrue (result.IsSuccessStatusCode);
+		}
+
 		private async Task AssertRejectsRemoteCertificate (Func<Task> makeRequest)
 		{
 			// there is a difference between the exception that's thrown in the .NET build and the legacy Xamarin
--- a/tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
+++ b/tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
@@ -119,7 +119,6 @@ namespace Xamarin.Android.NetTests
 			var handler = new AndroidMessageHandler {
 				ServerCertificateCustomValidationCallback = (request, cert, chain, errors) => {
 					callbackCounter++;
-					Assert.Contains (request.RequestUri.Host, cert.SubjectName.Name);
 					return errors == SslPolicyErrors.None;
 				}
 			};
--- a/tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
+++ b/tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
@@ -27,9 +27,9 @@ namespace Xamarin.Android.NetTests
 			var handler = new AndroidMessageHandler {
 				ServerCertificateCustomValidationCallback = (request, cert, chain, errors) => {
 					Assert.NotNull (request, "request");
-					Assert.AreEqual ("microsoft.com", request.RequestUri.Host);
+					Assert.AreEqual ("www.microsoft.com", request.RequestUri.Host);
 					Assert.NotNull (cert, "cert");
-					Assert.True (cert!.Subject.Contains ("microsoft.com"), $"Unexpected certificate subject {cert!.Subject}");
+					Assert.True (cert!.Subject.Contains ("www.microsoft.com"), $"Unexpected certificate subject {cert!.Subject}");
 					Assert.True (cert!.Issuer.Contains ("Microsoft"), $"Unexpected certificate issuer {cert!.Issuer}");
 					Assert.NotNull (chain, "chain");
 					Assert.AreEqual (SslPolicyErrors.None, errors);
@@ -40,7 +40,7 @@ namespace Xamarin.Android.NetTests
 			};
 
 			var client = new HttpClient (handler);
-			await client.GetStringAsync ("https://microsoft.com/");
+			await client.GetStringAsync ("https://www.microsoft.com/");
 
 			Assert.IsTrue (callbackHasBeenCalled, "custom validation callback hasn't been called");
 		}
@@ -58,7 +58,7 @@ namespace Xamarin.Android.NetTests
 			};
 			var client = new HttpClient (handler);
 
-			await AssertRejectsRemoteCertificate (() => client.GetStringAsync ("https://microsoft.com/"));
+			await AssertRejectsRemoteCertificate (() => client.GetStringAsync ("https://www.microsoft.com/"));
 
 			Assert.IsTrue (callbackHasBeenCalled, "custom validation callback hasn't been called");
 		}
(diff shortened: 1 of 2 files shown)


grendello (approved):

tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
```diff
@@ -111,6 +111,25 @@ public async Task ServerCertificateCustomValidationCallback_IgnoresCertificateHo
...
+				ServerCertificateCustomValidationCallback = (request, cert, chain, errors) => {
+					callbackCounter++;
+					return errors == SslPolicyErrors.None;
+				}
+			};
+
+			var client = new HttpClient (handler);
+			var result = await client.GetAsync ("https://httpbin.org/redirect-to?url=https://www.microsoft.com/");
+
+			Assert.AreEqual (2, callbackCounter);
```
jonpryor, line 129: @simonrozsival: why is this `2`?  Are there any docs around how many times `ServerCertificateCustomValidationCallback` will be invoked, other than "at least once"?  Why is the callback being called twice?  Is that a bug?  Is that likely to change at any point in the future?

Or *should* it be 2: once for `httpbin.org/redirect-to`, and once again for `www.microsoft.com`, and it's called "once per redirect", i.e. if you "nested" redirects with `https://httpbin.org/redirect-to?url=https://httpbin.org/redirect-to?url=…`, `callbackCounter` should be the number of times we hit `httpbin.org/redirect-to` + 1?

jonpryor:
Draft commit message:

```markdown
[Mono.Android] ServerCertificateValidationCallback() and redirects (#7662)

Fixes: https://github.com/xamarin/xamarin-android/issues/7650

`AndroidMessageHandler` was incorrectly validating SSL certificate
after a redirect to another domain if and only if
`AndroidMessageHandler.ServerCertificateCustomValidationCallback`
was set; consider:

	const string url = "https://httpbin.org/redirect-to?url=https://www.microsoft.com/";
	using var handler = new AndroidMessageHandler() {
	    ServerCertificateCustomValidationCallback = (message, certificate, chain, errors) => errors == SslPolicyErrors.None
	};
	using var client = new HttpClient(handler);
	await client.GetAsync(url);

The expectation is that
`AndroidMessageHandler.ServerCertificateCustomValidationCallback`
will be called *twice*:

 1. Once with `message.RequestUri` set to
    `https://httpbin.org/redirect-to?…`, and
 2. Once again with `message.RequestUri` set to
    `https://www.microsoft.com`.

The problem was that in actuality the 2nd invocation was unchanged
from (1), with `message.RequestUri` set to
`https://httpbin.org/redirect-to…?…`, and `errors` set to
`SslPolicyErrors.RemoteCertificateNameMismatch` (!).

The problem was that `AndroidMessageHandler.DoSendAsync()` didn't
update `HttpRequestMessage.RequestUri` to the redirected URI, causing
the `errors` enum value to contain an error, and the second
`ServerCertificateCustomValidationCallback` invocation to have the
wrong information.

Update `AndroidMessageHandler.DoSendAsync()` so that when dealing
with an HTTP redirect, `message.RequestUri` always contains the
"current" value in the redirect chain.  This keeps `RequestUri` in
sync with the redirect status to avoid this problem.
```

tests/Mono.Android-Tests/Xamarin.Android.Net/AndroidMessageHandlerTests.cs
simonrozsival: The callback is called once for every sub-request that's made (the first request + all redirects). In this case, it's called once for `httpbin.org/...` and once for `www.microsoft.com`. So `callbackCounter` should equal to 1 + number of redirects. The `SocketsHttpHandler` behaves the same way.

PR #7672, opened 2023-01-05
[release/7.0.1xx] [Xamarin.Android.Build.Tasks] fix cases of missing `@(Reference)`

jonathanpeppers:
Backport to .NET 7 (merge conflict)

Fixes: https://github.com/dotnet/maui/issues/10154

Context? https://github.com/dotnet/maui/issues/11364

If you have a solution setup with:

  * `ApplicationA.csproj` has a `@(ProjectReference)` to `LibraryB.csproj`.
  * `LibraryB.csproj` which has a `@(Reference)` to `LibraryC.dll`, built by-
  * `LibraryC.csproj`

The app will crash at runtime, due to a missing `LibraryC.dll`.

The workaround is for `LibraryB.csproj` to use `@(ProjectReference)` to `LibraryC.csproj` instead of `@(Reference)` to `LibraryC.dll`.

However, it appears the same situation works in a .NET 7 self-contained console app:

	% dotnet publish --self-contained -r win-x64
	…
	% ls -1 .\bin\Debug\net7.0\win-x64\publish\LibraryC.dll
	LibraryC.dll

The underlying issue appears to be due to [`$(_FindDependencies)`][0]:

	<_FindDependencies Condition="'$(BuildingProject)' != 'true' and '$(_ResolveReferenceDependencies)' != 'true'">false</_FindDependencies>

In the console app, `$(BuildingProject)`=true and
`$(_FindDependencies)` is empty.

In the Android app, `$(BuildingProject)`=false and `$(_FindDependencies)` is `false`.

It appears that the `BuildOnlySettings` target *should* be running in Android apps when we do an "inner" build per `$(RuntimeIdentifier)`. Simply updating `_ComputeFilesToPublishForRuntimeIdentifiers` so that the `BuildOnlySettings` target is in `DependsOnTargets` fixes this. However, this also causes satellite assemblies to now be automatically found by the .NET SDK.  Update `@(_AndroidResolvedSatellitePaths)` so that `@(ReferenceSatellitePaths)` is only included on Classic builds, preventing duplicate entries.

[0]: https://github.com/dotnet/msbuild/blob/a2490dd3f78cce4abc8f9e6f1b5268437332818f/src/Tasks/Microsoft.Common.CurrentVersion.targets#L2322

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/targets/Microsoft.Android.Sdk.AssemblyResolution.targets
@@ -39,7 +39,7 @@ _ResolveAssemblies MSBuild target.
   </PropertyGroup>
 
   <Target Name="_ComputeFilesToPublishForRuntimeIdentifiers"
-      DependsOnTargets="_FixupIntermediateAssembly;ResolveReferences;ComputeFilesToPublish;_AndroidAot"
+      DependsOnTargets="BuildOnlySettings;_FixupIntermediateAssembly;ResolveReferences;ComputeFilesToPublish;_AndroidAot"
       Returns="@(ResolvedFileToPublish)">
       <ItemGroup>
         <ResolvedFileToPublish Remove="@(_SourceItemsToCopyToPublishDirectory)" />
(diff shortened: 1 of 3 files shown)


dellis1972 (approved):

jonathanpeppers:
Hmm, this test failure is concerning:

    /tr-TR/Localization.resources.dll did not exist in the .override directory.

Maybe there is a fix in xamarin/monodroid we don't have on the .NET 7 branch?

dellis1972:
There is this PR https://github.com/xamarin/monodroid/pull/1276

dellis1972:
But that test is using .resx file so it should be working.

jonathanpeppers:
I'm going to hold off backporting this for now, I'm not sure why it works in main and not here.

PR #7680, opened 2023-01-09
[Xamarin.Android.Build.Tasks] skip XA1034 logic in some cases

jonathanpeppers:
After profiling a build for a `dotnet new maui` app, I noticed:

    193.41ms xamarin.android.build.tasks!Xamarin.Android.Tasks.MonoAndroidHelper.HasResourceDesignerAssemblyReference()

Which equates to:

    Top 10 most expensive tasks
    ...
    AppendCustomMetadataToItemGroup = 204 ms

In dc3ccf2, we added this check to emit `XA1034` when `$(AndroidUseDesignerAssembly)` is false. And so we don't actually need to validate this value when it is `true` (the default).

I moved the logic from the `<AppendCustomMetadataToItemGroup/>` MSBuild task to a new `<CheckForInvalidDesignerConfig/>` task. In most projects, it will be completely skipped:

    Target "_CheckForInvalidDesignerConfig" skipped, due to false condition; ( '$(AndroidUseDesignerAssembly)' != 'True' ) was evaluated as ( 'true' != 'True' ).

After these changes, I get a much better number in projects with default settings:

    AppendCustomMetadataToItemGroup = 2 ms

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/CheckForInvalidDesignerConfig.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/CheckForInvalidDesignerConfig.cs
@@ -1,0 +1,53 @@
+using Microsoft.Android.Build.Tasks;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Reflection.Metadata;
+using System.Reflection.PortableExecutable;
+using System.IO;
+
+namespace Xamarin.Android.Tasks
+{
+	/// <summary>
+	/// Emits XA1034 is an assembly has a reference to _Microsoft.Android.Resource.Designer.dll
+	/// * NOTE: only called when $(AndroidUseDesignerAssembly) is false
+	/// </summary>
+	public class CheckForInvalidDesignerConfig : AndroidTask
+	{
+		public override string TaskPrefix => "CIRF";
+
+		public ITaskItem[] Assemblies { get; set; }
+
+		public override bool RunTask ()
+		{
+			foreach (var assembly in Assemblies) {
+				if (HasResourceDesignerAssemblyReference (assembly)) {
+					Log.LogCodedError ("XA1034", Properties.Resources.XA1034, assembly);
+				}
+			}
+
+			return !Log.HasLoggedErrors;
+		}
+
+		static bool HasResourceDesignerAssemblyReference (ITaskItem assembly)
+		{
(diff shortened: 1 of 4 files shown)


src/Xamarin.Android.Build.Tasks/Utilities/MonoAndroidHelper.cs
```diff
@@ -325,28 +325,6 @@ public static bool HasMonoAndroidReference (MetadataReader reader)
 			return false;
 		}
 
-		public static bool HasResourceDesignerAssemblyReference (ITaskItem assembly)
```
dellis1972, line 328: I would have left these in the MonoAndroidHelper since they might be useful in other parts of the build system at some point.But I guess we could move them back if they are really needed 🤷 😄

src/Xamarin.Android.Build.Tasks/Tasks/CheckForInvalidDesignerConfig.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Reflection.PortableExecutable;
+using System.IO;
+
+namespace Xamarin.Android.Tasks
+{
+	/// <summary>
+	/// Emits XA1034 is an assembly has a reference to _Microsoft.Android.Resource.Designer.dll
+	/// * NOTE: only called when $(AndroidUseDesignerAssembly) is false
+	/// </summary>
+	public class CheckForInvalidDesignerConfig : AndroidTask
```
dellis1972, line 14: Do we have any timings on who quick this tasks is ? Have we not just moved the 200 ms to this task instead? 
I think this is now working on a smaller subset of assemblies , but it would be good to know how quick this new task is.

jonathanpeppers: This task is the exact same speed, but I will add a check to skip BCL assemblies might make it 50% faster.

jonathanpeppers: Ok, the new code is actually faster...

image

It's because the task previously looked at *all* assemblies, and now this only looks at `@(_MonoAndroidReferencePath);@(_MonoAndroidReferenceDependencyPaths)`.

dellis1972 (approved):

PR #7681, opened 2023-01-10
Change Namespace to _Microsoft.Android.Resource.Designer

dellis1972:
We if we used the current `Microsoft.Android.Resource.Designer` namespace we can get into issues with any other library which uses `Microsoft` in the namespace. When building maui we see the following build error

```
src\Essentials\src\AppActions\AppActions.shared.cs(62,28): error CS0234: The type or namespace name 'Content' does not exist in the namespace 'Microsoft.Android' (are you missing an assembly reference?)
```

This is because rather than looking for `Content` in the global `Android` namespace, it looks for it in the `Microsfot.Android` namespace. This can be fixed by the end user by adding `global::` to the type usage, but this is not desireable in all senarios. Given this is an internal class which really should not be used lets change the namespace instead.

In addition we should add a `#pragma` for `IDE0002`[1] so that users are NOT asked to simplify the Resource designer class declaration. We also need to add the `EditorBrowsable` attribute so that the types do not show up in code completion in the IDE.

[1] https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0002


src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixLegacyResourceDesignerStep.cs
```diff
@@ -20,7 +20,7 @@ namespace MonoDroid.Tuner
 	public class FixLegacyResourceDesignerStep : LinkDesignerBase
 	{
 		internal const string DesignerAssemblyName = "_Microsoft.Android.Resource.Designer";
-		internal const string DesignerAssemblyNamespace = "Microsoft.Android.Resource.Designer";
+		internal const string DesignerAssemblyNamespace = "_Microsoft.Android.Resource.Designer";
```
jonathanpeppers, line 23: Were we going to try `[EditorBrowsable (EditorBrowsableState.Never)]` on the `Resource` class, too?

dellis1972: I want to test that if doesn't break anything in a "fake" project first.

src/Xamarin.Android.Build.Tasks/Tasks/GenerateResourceDesignerIntermediateClass.cs
```diff
@@ -19,7 +19,7 @@ public class GenerateResourceDesignerIntermediateClass : AndroidTask
 using System;
 
 namespace %NAMESPACE% {
-	public partial class Resource : %BASECLASS% {
+	public class Resource : %BASECLASS% {
```
jonpryor, line 22: I'm not sure that I like this idea; I *like* `partial` classes…

dellis1972: I was just testing to see if this would break any tests...it didn't 🤣

src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixLegacyResourceDesignerStep.cs
dellis1972: So adding the attribute does hide it from view. 

In this example `Classfoo` derives from `ClassHidden`. And I didn't see `ClassHidden` in the drop  the doing `Classfoo.`. However it is still listed when doing `ClassLibrary1.`. 

Also we get this recommendation. I'm not sure if there is a way to turn that off..

Capture

jonathanpeppers: Are there ways to mark your test class is auto-generated?
```
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by a tool.
// </auto-generated>
//------------------------------------------------------------------------------
```
I don't know if that might suppress some of the lightbulb suggestions...

dellis1972: the generated .cs files we can add a comment, but the assembly itself it all IL, so that probably won't work.

jonathanpeppers (approved):

jonpryor:
```markdown
Context: dc3ccf28cdbe9f8c0a705400b83c11a85c81a980
Context: https://github.com/dotnet/maui/pull/12520#issuecomment-1376431313
Context: https://discord.com/channels/732297728826277939/732297837953679412/1062137297438519296
Context: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0002

Consider this C# fragment:

	namespace Android.Content {
	    public class Intent {        
	    }
	}

	namespace Microsoft.Maui.Example {
	    public class Helper {
	        public static void UseIntent (Android.Content.Intent intent) {}
	    }
	}

Given a C# *namespace_or_type_name* such as `Android.Content.Intent`,
[ECMA 334 §7.8.1][0] will be used to determine what the type refers
to, resolving it to `global::Android.Content.Intent`.

All is good.

However, once dc3ccf28 enters the picture, we now have:

	// via dc3ccf28 & _Microsoft.Android.Resource.Designer.dll
	namespace Microsoft.Android.Resource.Designer {
	    public class Resource {
	    }
	}

	namespace Android.Content {
	    public class Intent {        
	    }
	}

	namespace Microsoft.Maui.Example {
	    public class Helper {
	        // CS0234 on the following line:		
	        public static void UseIntent (Android.Content.Intent intent) {}
	    }
	}

…and it fails to build with [error CS0234][1]:

	error CS0234: The type or namespace name 'Content' does not exist in the namespace 'Microsoft.Android' (are you missing an assembly reference?)

This only happens if the usage of `Android.Content.Intent` happens
*within* a `Microsoft.`-prefixed namespace.  Which is most of MAUI.

The cause of the error is that while attempting to resolve
`Android.Content.Intent`, search starts from the *current* namespace
`Microsoft.Maui.Example`, and the compiler will attempt to resolve
the types:

  * `Microsoft.Maui.Example.Android.Content.Intent`
    (because the usage site is in the `Microsoft.Maui.Example` namespace.)
  * `Microsoft.Maui.Android.Content.Intent`
    (because `Microsoft.Maui` is a parent namespace)
  * `Microsoft.Android.Content.Intent`
    (because `Microsoft` is a parent parent namespace)

These are the *only* types that are resolved, and none of those exist.
This results in the CS0234 error.

There are two workarounds that the "offending" code can employ:

 1. Use `global`:

        namespace Microsoft.Maui.Example {
            public class Helper {
                public static void UseIntent (global::Android.Content.Intent intent) {}
            }
        }

 2. Add a `using Android.Content` and use `Intent`

        using Android.Content;
        namespace Microsoft.Maui.Example {
            public class Helper {
                public static void UseIntent (Intent intent) {}
            }
        }

Both of these require changing things *outside* of xamarin-android.

Rephrasing & simplifying: commit dc3ccf28 constitutes an *API break*,
as code which *previously* compiled *no longer compiles*.

Fix this by updating the `<GenerateResourceDesignerAssembly/>` task
emits types into the `_Microsoft.Android.Resource.Designer` namespace;
note `_` prefix.  No One™ should have their code in a `_Microsoft.*`
namespace, so this shouldn't break anybody.  Additionally, update
`_Microsoft.Android.Resource.Designer.dll` so that
[`EditorBrowsableAttribute`][2] is placed on all the types, a'la:

	namespace _Microsoft.Android.Resource.Designer;
	[EditorBrowsable (EditorBrowsableState.Never)]
	public partial class Resource {
	}

	// For App project builds
	[EditorBrowsable (EditorBrowsableState.Never)]
	internal partial class ResourceConstant {
	}

Finally, update the Source Compatibility types so that the `Resource`
type disables the IDE0002 warning:

	namespace %CompatibilityNamespace% {
	    #pragma warning disable IDE0002
	    public partial class Resource : _Microsoft.Android.Resource.Designer.Resource {
	    }
	    #pragma warning restore IDE0002
	}

[0]: https://github.com/dotnet/csharpstandard/blob/standard-v6/standard/basic-concepts.md#78-namespace-and-type-names
[1]: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-messages/cs0234
[2]: https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.editorbrowsableattribute?view=net-7.0
[3]: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0002
```

PR #7682, opened 2023-01-10
Create .NET version of JcwGen-Tests.

jpobst:


jonpryor:
The addition of `tests/BindingTests/Xamarin.Android.BindingRuntime-Tests/Jars/org.jetbrains.kotlin.kotlin-stdlib.jar` seems odd to me; why don't we instead reference the [`Xamarin.Kotlin.Stdlib` NuGet package](https://www.nuget.org/packages/Xamarin.Kotlin.StdLib)?

tests/BindingTests/Xamarin.Android.BindingRuntime-Tests/Tests/BindingTests.cs
```diff
@@ -0,0 +1,391 @@
...
+				Assert.IsTrue (((CallNonvirtualDerived) b).MethodInvoked);
+			}
+			using (var d2 = new CallNonvirtualDerived2 ()) {
+				d2.Method ();
+				Assert.IsTrue (((CallNonvirtualDerived) d2).MethodInvoked);
+				Assert.IsFalse (((CallNonvirtualBase) d2).MethodInvoked);
+			}
+		}
+
+		//[Test]
```
jonpryor, line 181: Maybe this is a "just me" thing, but if this is a test we want to *eventually* bring back as part of #7794, then I think instead of commenting out the test, we should wrap it in `#if TODO_7794`.  That makes it a bit more obvious "what it's waiting for"/why it isn't being built (yet).

build-tools/automation/azure-pipelines.yaml
```diff
@@ -409,6 +390,27 @@ stages:
...
+        artifactSource: bin/Test$(XA.Build.Configuration)/$(DotNetTargetFramework)-android/Xamarin.Android.BindingRuntime_Tests-Signed.apk
+        artifactFolder: $(DotNetTargetFramework)-Default
+        useDotNet: true
+
+    - template: yaml-templates/apk-instrumentation.yaml
+      parameters:
+        configuration: $(XA.Build.Configuration)
+        testName: Xamarin.Android.BindingRuntime_Tests_FastDev
+        project: tests/BindingTests/Xamarin.Android.BindingRuntime-Tests/Xamarin.Android.BindingRuntime-Tests.csproj
+        testResultsFiles: TestResult-Xamarin.Android.BindingRuntime_Tests_FastDev-$(ApkTestConfiguration).xml
```
pjcollins, line 408: For some reason the `$(ApkTestConfiguration)` is only set on the classic APK test job, I would replace this with `$(XA.Build.Configuration)` instead. I think this is preventing the [test results from being uploaded](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7343675&view=logs&j=5ded96ba-e325-5baf-fde0-f1868aa3be52&t=113564c2-5517-5674-66e2-044addc8c3ea&l=11):

> ##[warning]No test result files matching TestResult-Xamarin.Android.BindingRuntime_Tests-$(ApkTestConfiguration).xml were found.

tests/BindingTests/Xamarin.Android.BindingRuntime-Tests/Tests/BindingTests.cs
pjcollins: Should we do the same for the other tests that are commented out in this file?

jonpryor: i missed the others!

tests/BindingTests/Xamarin.Android.BindingRuntime-Tests/Tests/BindingTests.cs
```diff
@@ -0,0 +1,398 @@
...
+		[Test]
+		public void EnsureTypesAreBound ()
+		{
+#pragma warning disable 0219
+			Com.Xamarin.Android.Bxc9446 ignore  = null;
+#pragma warning restore 0219
+		}
+
+		//[Test]
+		//public void Arrays ()
```
jonpryor, line 76: @jpobst: why is *this* test disabled?  `Com.Xamarin.Android.DataHandler` is in `DataHandler.java`, which is added/copied as part of this PR.

jpobst: It was disabled because Marshal Methods broke it:
https://github.com/xamarin/xamarin-android/issues/7693

As that has now been fixed this test should be enabled.

jonpryor:
```markdown
Context: https://github.com/xamarin/xamarin-android/issues/7794

Create a .NET version of
`tests/CodeGen-Binding/Xamarin.Android.JcwGen-Tests`.  It currently
isn't possible to multi-target our test projects, so instead create a
copy and use `@(AndroidJavaSource)` to simplify the test suite by
eliminating the `Xamarin.Android.McwGen-Tests` project and its custom
MSBuild logic.

This new version uses the more descriptive
`Xamarin.Android.BindingRuntime_Tests` project name.

This ports 34 of the 41 tests in `Xamarin.Android.JcwGen-Tests`.
The remaining tests require additional projects to be ported to .NET
to run successfully or require additional custom MSBuild logic from
`Xamarin.Android.JcwGen-Tests` to be ported.

Additionally, stop running this suite on Classic test agents on `main`.

We are choosing to not prioritize the remaining tests at this point
and committing what we already have.

The unported tests are tracked in #7794.
```

pjcollins (commented):
One last though -- rather than creating a copy of all of the Java sources used by the tests, should we instead link to the files that already exist in the other test directory (for example https://github.com/xamarin/xamarin-android/tree/main/tests/CodeGen-Binding/Xamarin.Android.McwGen-Tests/java/com/xamarin/android)? I'm assuming the java sources are unchanged and could be shared?

If so, should we do the same for the test files as well? We could wrap any test changes or ignored tests in a `#if NET6_0_OR_GREATER` if we wanted to use the same files to help preserve git history.

jpobst:
Honestly, we could just remove the old files and they would show up as renames.  This PR created a copy because it was originally done before we started removing Classic from `main`.

Issue #7684, opened 2023-01-10
[.NET 8] perf regression in `<LinkAssembliesNoShrink/>`


jonathanpeppers:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

.NET 8 / main

### Description

In .NET 7, in an old log I see:

![image](https://user-images.githubusercontent.com/840039/211677412-a672b5fa-7034-4a70-8d3d-8ad3820e6c5c.png)

This was an example of a `.xaml` change in a `dotnet new maui` project.

Now I see:

![image](https://user-images.githubusercontent.com/840039/211677489-b17cc8c3-f4ce-4ce9-8e85-a9b8272813ae.png)

`dotnet trace` shows most of the time is:

![image](https://user-images.githubusercontent.com/840039/211677562-296eed6a-e557-4dc1-880f-08598c8da684.png)

So I think we might need to rework parts of dc3ccf28cdbe9f8c0a705400b83c11a85c81a980, maybe it can skip some things on incremental builds?

Less log messages would also help:

![image](https://user-images.githubusercontent.com/840039/211677697-34f68a48-72da-45a0-aa05-a1e1f7aae0b0.png)

### Steps to Reproduce

1. `dotnet new maui`
2. build
3. change `.xaml`
4. build again

### Did you find any workaround?

Set `$(AndroidUseDesignerAssembly)` to `false`, would skip some of these steps.

### Relevant log output

[msbuild.zip](https://github.com/xamarin/xamarin-android/files/10387362/msbuild.zip)

dellis1972:
This step is converting old Resource.designer.cs code over to the new system. So once Maui has switched over to using the newer system it should be better as it will have less work to do.

jonathanpeppers:
Yeah, we can keep an eye and test again when we have a new MAUI. 

What made me think something else is going on, is this target builds partially:

    Target Name=_LinkAssembliesNoShrink Project=bar.csproj
    Building target "_LinkAssembliesNoShrink" partially, because some output files are out of date with respect to their input files.
    [ResolvedAssemblies: Input=obj\Debug\net8.0-android\android-arm64\bar.dll, Output=obj\Debug\net8.0-android\android-arm64\android\assets\bar.dll] Input file is newer than output file.

So in theory it should only be looking at `bar.dll` in this example?

dellis1972:
It is buiding partially. but the Task in the target is not using any of the "partial" calculated ItemGroups

```
[...]

PR #7686, opened 2023-01-11
[Xamarin.Android.Build.Tasks] lazily populate Resource lookup

jonathanpeppers:
Fixes: https://github.com/xamarin/xamarin-android/issues/7684

Comparing .NET 7 to main, I noticed:

    .NET 7
    Task LinkAssembliesNoShrink 4ms
    main/.NET 8
    Task LinkAssembliesNoShrink 101ms

Under `dotnet trace` a lot of the time was spent in:

    94.74ms MonoDroid.Tuner.FixLegacyResourceDesignerStep.LoadDesigner()

Reviewing the code, I think we can "lazily" call the `LoadDesigner()` method. It creates a `Dictionary` that isn't used until the `FixBody()` method.

I also updated one log message to only log duplicates, as it was logging hundreds of lines:

    if (output.ContainsKey (key)) {
        LogMessage ($"          Found duplicate {key}");
    } else {
        output.Add (key, property.GetMethod);
    }

Which also showed up in `dotnet trace`:

    25.58ms Microsoft.Android.Build.Tasks.MSBuildExtensions.LogDebugMessage()

With these changes, I instead get:

    Task LinkAssembliesNoShrink 5ms

Which is probably ~the same performance as before or plenty good enough!


src/Xamarin.Android.Build.Tasks/Linker/MonoDroid.Tuner/FixLegacyResourceDesignerStep.cs
```diff
@@ -90,6 +77,15 @@ internal override bool ProcessAssemblyDesigner (AssemblyDefinition assembly)
...
 
+			// This is expected for the first call, in <LinkAssembliesNoShrink/>
+			if (!designerLoaded)
+				LoadDesigner ();
+
+			if (designerAssembly == null || designerType == null) {
+				LogMessage ($"   Not using {DesignerAssemblyName}");
+				return false;
+			}
+
```
jonathanpeppers, line 88: This is the main change that helps perf, delay the `LoadDesigner()` call to here. I also had to move the `designerAssembly ` and `designerType ` null checks down for it to actually work...

jonathanpeppers:
I think it's going to work this time, my first attempt triggered the failure:
```
ApplicationRunsWithoutDebugger(False,False,False)
Activity should have started.
Expected: True
But was:  False
```
So we should have decent test coverage on this.

dellis1972:
This might have been an ovesight on my part, but I'm not sure we have a test which specifically tests this senario on device. 
i.e an app which does use the new designer system with a library that does not. We might need to add one or modify an existing one to do that.

Problem is the test would need to be DotNet only since the new system is not supported.

jonathanpeppers:
When I was debugging `ApplicationRunsWithoutDebugger`, I found it was actually testing this scenario (why I got a failure with my first attempt 30501c14b103ac7f645533b050ab60e234fb38f2 before when I broke something):

https://github.com/xamarin/xamarin-android/blob/8b0531465d90f6357bd9271faf7e5d6f7f3e4a46/tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs#L48-L57

Because this app is Xamarin.Forms (which uses `Resource.designer.cs` in a class library), I was getting an `InvalidCastException` on startup when this was wrong. It had an incorrect integer for a resource id, which caused the exception at startup.

dellis1972 (approved):

PR #7694, opened 2023-01-12
[WIP] [Xamarin.Android.Build.Tasks] use hashes as `Inputs` for `_GenerateJavaStubs`

jonathanpeppers:
One of the slower parts of an incremental Android build is:

    _GenerateJavaStubs = 978 ms

Currently, this happens when any .NET assembly changes: C# change, `.xaml` change, etc.

In theory, we could analyze .NET assemblies during the linker and create a "hash" that represents types & members that would invalidate `_GenerateJavaStubs`'s output. We can then save these hashes in a file to be used as `Inputs` instead of the .NET assemblies themselves.

Example of the hash:

    HashAlgorithm hasher = new Microsoft.Android.Build.Tasks.Crc64 ();
    foreach (var type in scanner.GetJavaTypes (assembly)) {
        Hash (hasher, type.FullName);
        foreach (var method in type.Methods) {
            Hash (hasher, method.Name);
        }
    }

The result is:

    Target Name=_GenerateJavaStubs Project=bar.csproj
    Skipping target "_GenerateJavaStubs" because all output files are up-to-date with respect to the input files.

Improving the incremental build time about about ~1 second in a `dotnet new maui` project. `_LinkAssembliesNoShrink` will be slightly slower, but because it builds partially it will only be creating this hash on a single file most of the time. Currently, this seems to be about ~20ms final numbers TBD.

Things left to do:

* Make sure we are hashing the right things that `_GenerateJavaStubs` cares about.

* Fix `Release` builds, they likely aren't setup right.

* Add several tests around this scenario. We need to validate adding new `Java.Lang.Object` subclasses work as expected.

### DIFF
--- a/src/Microsoft.Android.Sdk.ILLink/AssemblyResolverCoda.cs
+++ b/src/Microsoft.Android.Sdk.ILLink/AssemblyResolverCoda.cs
@@ -1,0 +1,12 @@
+using Mono.Cecil;
+using System.IO;
+
+namespace Java.Interop.Tools.Cecil;
+
+public static class AssemblyResolverCoda
+{
+	public static AssemblyDefinition GetAssembly (this IAssemblyResolver resolver, string fileName)
+	{
+		return resolver.Resolve (AssemblyNameReference.Parse (Path.GetFileNameWithoutExtension (fileName)));
+	}
+}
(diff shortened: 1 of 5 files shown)


jonathanpeppers:
Actually, looks like I have a small code change in java.interop. Looks like this won't build yet...

jonathanpeppers:
Due to the various errors trying to use `JavaTypeScanner` in `Microsoft.Android.Sdk.ILLink.csproj`.

I'm going to see if I can do this idea for `Debug` builds only.

jonathanpeppers:
Closing old PRs, will revisit later if needed.

PR #7710, opened 2023-01-17
MAUI project with non-ASCII project name cannot release to my Android phone

dellis1972:
Context https://i.azdo.io/1714603
Context https://issuetracker.google.com/issues/188679588

There is a known issue with `aapt2`, `AndroidAsset` and non-ASCII paths/project names. 
`aapt2` has a bug where it cannot correctly traverse a directory which contains non-ASCII 
characters. We curently have no way to work around this issue. 

So the PR adds some more detail to the specific error message which `aapt2` raises in 
these cases. It suggests the user check their paths to make sure it does not contain 
non-ASCII characters. 

We also add some unit tests to cover this senario.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -542,7 +542,11 @@ Either change the value in the AndroidManifest.xml to match the $(SupportedOSPla
 {0} - The assembly name</comment>
   </data>
   <data name="APT2264" xml:space="preserve">
-    <value>This could be caused by the project having non-ASCII characters in it path or exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
+    <value>This could be caused by the project exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
+    <comment>The following are literal names and should not be translated:</comment>
+  </data>
+  <data name="APT2265" xml:space="preserve">
+    <value>This could be caused by the project having non-ASCII characters in its filename or path. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2265 for details.</value>
     <comment>The following are literal names and should not be translated:</comment>
   </data>
   <data name="XA3001" xml:space="preserve">
--- a/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
+++ b/src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
@@ -542,7 +542,7 @@ Either change the value in the AndroidManifest.xml to match the $(SupportedOSPla
 {0} - The assembly name</comment>
   </data>
   <data name="APT2264" xml:space="preserve">
-    <value>This is probably caused by the project exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
+    <value>This could be caused by the project having non-ASCII characters in it path or exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
     <comment>The following are literal names and should not be translated:</comment>
   </data>
   <data name="XA3001" xml:space="preserve">
--- a/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
@@ -138,22 +138,8 @@ namespace Xamarin.Android.Tasks {
 					LogCodedError ("APT0001", Properties.Resources.APT0001, message.Substring ("unknown option '".Length).TrimEnd ('.', '\''));
 					return false;
 				}
-				if (message.Contains ("in APK") && message.Contains ("is compressed.")) {
-					LogMessage (singleLine, messageImportance);
+				if (LogNotesOrWarnings (message, singleLine, messageImportance))
 					return true;
-				}
-				if (message.Contains ("fakeLogOpen")) {
-					LogMessage (singleLine, messageImportance);
-					return true;
-				}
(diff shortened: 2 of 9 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -110,6 +110,26 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
 			StringAssertEx.DoesNotContainRegex (@$"Using profile data file.*{filename}\.aotprofile", b.LastBuildOutput, "Should not use default AOT profile", RegexOptions.IgnoreCase);
 		}
 
+		[Test]
+		[Category ("SmokeTests")]
```
jpobst, line 114: Going forward, we probably do not need to add any tests to `SmokeTests`, this only affects the tests run on the in-tree Windows build which only runs the bare minimum.

All tests are run against the Mac/Win installers going forward, there are no longer separate Smoke Test jobs.

https://github.com/xamarin/xamarin-android/pull/7872

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -542,7 +542,7 @@ Either change the value in the AndroidManifest.xml to match the $(SupportedOSPla
 {0} - The assembly name</comment>
   </data>
   <data name="APT2264" xml:space="preserve">
-    <value>This is probably caused by the project exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
+    <value>This is probably caused by the project having non-ASCII characters in it path or exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
```
jonathanpeppers, line 545: Should we be slightly less certain in the message?
```suggestion
    <value>This could be caused by the project having non-ASCII characters in it path or exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
```

jonathanpeppers (approved):
The overall code changes look good. 👍

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
```diff
@@ -138,22 +138,8 @@ protected bool LogAapt2EventsFromOutput (string singleLine, MessageImportance me
 					LogCodedError ("APT0001", Properties.Resources.APT0001, message.Substring ("unknown option '".Length).TrimEnd ('.', '\''));
 					return false;
 				}
-				if (message.Contains ("in APK") && message.Contains ("is compressed.")) {
-					LogMessage (singleLine, messageImportance);
+				if (LogNotesOrWarnings (message, singleLine, messageImportance))
```
jonpryor, line 141: Looking at this reminds me of https://github.com/xamarin/xamarin-android-tools/issues/208 and that `<Aapt2/>` should likewise collect *all* `aapt2` output so that everything can be reported when a warning or error occurs.  It doesn't look like it currently "collects and forwards" all output.

src/Xamarin.Android.Build.Tasks/Properties/Resources.resx
```diff
@@ -542,7 +542,7 @@ Either change the value in the AndroidManifest.xml to match the $(SupportedOSPla
 {0} - The assembly name</comment>
   </data>
   <data name="APT2264" xml:space="preserve">
-    <value>This is probably caused by the project exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
+    <value>This could be caused by the project having non-ASCII characters in it path or exceeding the Windows maximum path length limitation. See https://learn.microsoft.com/xamarin/android/errors-and-warnings/apt2264 for details.</value>
```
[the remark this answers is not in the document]
jonpryor: Though a possible counterargument to two separate error codes is "what happens when *both* are present: non-ASCII characters *and* a too-long path!"

I suppose both could be listed…

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
dellis1972: I'd like to deal with this in a seperate PR if I can. It might require a bit more rework than I'd be comfortable with putting in this PR.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -112,6 +112,33 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
...
+			var rootPath = Path.Combine (Root, "temp", TestName);
+			var proj = new XamarinAndroidApplicationProject () {
+				ProjectName = testName,
+				IsRelease = isRelease,
+				AotAssemblies = aot,
+			};
+			proj.SetAndroidSupportedAbis ("armeabi-v7a",  "arm64-v8a", "x86", "x86_64");
+			using (var builder = CreateApkBuilder (Path.Combine (rootPath, proj.ProjectName))){
+				builder.ThrowOnBuildFailure = false;
+				Assert.AreEqual (expectedResult, builder.Build (proj), "Build should have succeeded.");
```
jonpryor, line 138: Should this also verify the build error?

src/Xamarin.Android.Build.Tasks/Tasks/Aapt2.cs
jonpryor: Separate PR is fine.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
dellis1972: done

PR #7721, opened 2023-01-20
[Xamarin.Android.Build.Tasks] Honour the AndroidGenerateResourceDesigner setting

dellis1972:
Currently setting `AndroidGenerateResourceDesigner` to `false` does NOT stop the new resource designer assembly
from being generated. 
We should change this so that library projects which want to use this setting (as per the docs) to disable
the resoruce designer then it should work.  So if the `AndroidGenerateResourceDesigner` is set to fail we will NOT generate
the intermediate `__Microsoft.Resource.Designer.Assembly.cs` (or .fs) file. But we will still generate the assembly. 
This means the `Resource` class will not be part of the built assembly and will not be available. 

A unit test has been updated to test this change.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
@@ -455,7 +455,7 @@ public class JavaSourceTest {
 		}
 
 		[Test]
-		public void GenerateResourceDesigner_false()
+		public void GenerateResourceDesigner_false([Values (false, true)] bool useDesignerAssembly)
 		{
 			var proj = new XASdkProject (outputType: "Library") {
 				Sources = {
@@ -466,7 +466,8 @@ public class JavaSourceTest {
 			};
 			// Turn off Resource.designer.cs and remove usage of it
 			proj.SetProperty ("AndroidGenerateResourceDesigner", "false");
-			proj.SetProperty ("AndroidUseDesignerAssembly", "false");
+			if (!useDesignerAssembly)
+				proj.SetProperty ("AndroidUseDesignerAssembly", "false");
 			proj.MainActivity = proj.DefaultMainActivity
 				.Replace ("Resource.Layout.Main", "0")
 				.Replace ("Resource.Id.myButton", "0");
@@ -474,12 +475,12 @@ public class JavaSourceTest {
 			var dotnet = CreateDotNetBuilder (proj);
 			Assert.IsTrue (dotnet.Build(target: "CoreCompile", parameters: new string[] { "BuildingInsideVisualStudio=true" }), "Designtime build should succeed.");
 			var intermediate = Path.Combine (FullProjectDirectory, proj.IntermediateOutputPath);
-			var resource_designer_cs = Path.Combine (intermediate, "designtime",  "Resource.designer.cs");
+			var resource_designer_cs = GetResourceDesignerPath (dotnet, proj, designtime: true);
 			FileAssert.DoesNotExist (resource_designer_cs);
 
 			Assert.IsTrue (dotnet.Build (), "build should succeed");
 
-			resource_designer_cs = Path.Combine (intermediate, "Resource.designer.cs");
+			resource_designer_cs =  GetResourceDesignerPath (dotnet, proj);
 			FileAssert.DoesNotExist (resource_designer_cs);
 
 			var assemblyPath = Path.Combine (FullProjectDirectory, proj.OutputPath, $"{proj.ProjectName}.dll");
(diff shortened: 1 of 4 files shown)


jonathanpeppers (commented):
So here is a combination I think might break:
1. Class library where I don't want any resource designer, so I set `$(AndroidGenerateResourceDesigner)` to `false`
2. Reference a class library that *does* have a resource designer

So think something might need to happen here (or only apply to application projects?):

https://github.com/xamarin/xamarin-android/blob/baa5a739b957d724df3a33f001e4e1cde914cee5/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Resource.Designer.targets#L143-L147

I think dotnet/maui will have this case.

dellis1972:
> So here is a combination I think might break:
> 
> 1. Class library where I don't want any resource designer, so I set `$(AndroidGenerateResourceDesigner)` to `false`
> 2. Reference a class library that _does_ have a resource designer
> 
> So think something might need to happen here (or only apply to application projects?):
> 
> https://github.com/xamarin/xamarin-android/blob/baa5a739b957d724df3a33f001e4e1cde914cee5/src/Xamarin.Android.Build.Tasks/MSBuild/Xamarin/Android/Xamarin.Android.Resource.Designer.targets#L143-L147
> 
> I think dotnet/maui will have this case.

Hmm. This is a good point. Currently you'd get an error since it won't let you do that. 
That does suggest we can't turn this off....

dellis1972:
I'm gonna have to have a think about this one. Maybe leaving the setting on and just not generating anything in the assembly would be better. The downside of that is you will always get an assembly.

jonathanpeppers:
The other thing I was setting in MAUI, was new public APIs for things like:
```
Microsoft.Maui.Graphics.Skia.Resource
Microsoft.Maui.Graphics.Skia.Resource.Resource() -> void
```
I had to add these to a list to make their API analyzer happy.

So maybe we could avoid generating the `__Microsoft.Android.Resource.Designer.cs` file?

dellis1972:
> So maybe we could avoid generating the `__Microsoft.Android.Resource.Designer.cs` file?

Good idea, I updated this PR to do that and it seems to work. 
Also updated the unit test.

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/XASdkTests.cs
```diff
@@ -466,20 +513,21 @@ public void GenerateResourceDesigner_false()
...
+				proj.SetProperty ("AndroidUseDesignerAssembly", "false");
 			proj.MainActivity = proj.DefaultMainActivity
 				.Replace ("Resource.Layout.Main", "0")
 				.Replace ("Resource.Id.myButton", "0");
 
 			var dotnet = CreateDotNetBuilder (proj);
 			Assert.IsTrue (dotnet.Build(target: "CoreCompile", parameters: new string[] { "BuildingInsideVisualStudio=true" }), "Designtime build should succeed.");
 			var intermediate = Path.Combine (FullProjectDirectory, proj.IntermediateOutputPath);
-			var resource_designer_cs = Path.Combine (intermediate, "designtime",  "Resource.designer.cs");
+			var resource_designer_cs = GetResourceDesignerPath (dotnet, proj, designtime: true);
```
dellis1972, line 525: this test needs fixing as it is now looking for the wrong file

jonathanpeppers (approved):
LGTM if it ends up green now. 👍

jonpryor:
```markdown
[Xamarin.Android.Build.Tasks] fix AndroidGenerateResourceDesigner (#7721)

Context: dc3ccf28cdbe9f8c0a705400b83c11a85c81a980
Context: https://github.com/dotnet/maui/pull/12520

When `_Microsoft.Android.Resource.Designer.dll` is used (dc3ccf28)
and [`$(AndroidGenerateResourceDesigner)`][0]=False, the
`__Microsoft.Android.Resource.Designer.cs` file is still generated,
which is *equivalent* to `Resource.designer.cs`.  This causes a
`Resource` type to be generated.

This showed up when attempting to bump MAUI to use xamarin-android
for .NET 8 (dotnet/maui#12520), which [introduced][1] *lots* of
`*.Resource` types which are *not desirable*.

Fix this so that when `$(AndroidGenerateResourceDesigner)`=False the
`__Microsoft.Android.Resource.Designer.*` files are *not* created.
This prevents `*.Resource` types from being added to the assembly.

Note that `_Microsoft.Android.Resource.Designer.dll` is still created
at build time.

A unit test has been updated to test this change.

[0]: https://learn.microsoft.com/en-us/xamarin/android/deploy-test/building-apps/build-properties#androidgenerateresourcedesigner
[1]: https://github.com/dotnet/maui/pull/12520/commits/2e4981257c1203a9dea935b4b39979859b770c33
```

Issue #4290, opened 2020-02-20
 Is View.SystemUiVisibility bound to View.StatusBarVisibility incorrectly？


suihanhbr:
I found View.SystemUiVisibility return a wrong value which is View.StatusBarVisibility.
There is the document from Android Studio.
![image](https://user-images.githubusercontent.com/17265184/74930784-44003000-5419-11ea-828c-fbc8adea9392.png)
But the content of View.StatusBarVisibility is:
```
public enum StatusBarVisibility
{
	[IntDefinition (null, JniField = "android/view/View.STATUS_BAR_HIDDEN")]
	Hidden = 1,
	[IntDefinition (null, JniField = "android/view/View.STATUS_BAR_VISIBLE")]
	Visible = 0
}
```
I think View.SystemUiFlags is the correct return by View.SystemUiVisibility.

```

[Flags]
public enum SystemUiFlags
{
	[IntDefinition ("Android.Views.View.SystemUiFlagFullscreen", JniField = "android/view/View.SYSTEM_UI_FLAG_FULLSCREEN")]
	Fullscreen = 0x4,
	[IntDefinition ("Android.Views.View.SystemUiFlagHideNavigation", JniField = "android/view/View.SYSTEM_UI_FLAG_HIDE_NAVIGATION")]
	HideNavigation = 0x2,
	[IntDefinition ("Android.Views.View.SystemUiFlagImmersive", JniField = "android/view/View.SYSTEM_UI_FLAG_IMMERSIVE")]
	Immersive = 0x800,
	[IntDefinition ("Android.Views.View.SystemUiFlagImmersiveSticky", JniField = "android/view/View.SYSTEM_UI_FLAG_IMMERSIVE_STICKY")]
	ImmersiveSticky = 0x1000,
	[IntDefinition ("Android.Views.View.SystemUiFlagLayoutFullscreen", JniField = "android/view/View.SYSTEM_UI_FLAG_LAYOUT_FULLSCREEN")]
	LayoutFullscreen = 0x400,
	[IntDefinition ("Android.Views.View.SystemUiFlagLayoutHideNavigation", JniField = "android/view/View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION")]
	LayoutHideNavigation = 0x200,
	[IntDefinition ("Android.Views.View.SystemUiFlagLayoutStable", JniField = "android/view/View.SYSTEM_UI_FLAG_LAYOUT_STABLE")]
[...]

PR #7730, opened 2023-01-23
[Mono.Android] Fix WindowManagerLayoutParams.SystemUiVisibility enumification

jpobst:
Fixes https://github.com/xamarin/xamarin-android/issues/4290

The following API were incorrectly enumified as `StatusBarVisibility` instead of `SystemUiFlags`:
- `android.views.View.[get|set]SystemUiVisibility (int)` ([source](https://developer.android.com/reference/android/view/View#setSystemUiVisibility(int)))
- `android.views.WindowManager.LayoutParams.SystemUiVisibility` ([source](https://developer.android.com/reference/android/view/WindowManager.LayoutParams#systemUiVisibility))

As these are properties, we cannot create an overload.  Instead create a new property called `SystemUiFlags` and obsolete the incorrect property.  Additionally, this whole thing was deprecated in API-30, so make sure we reflect that on the new method.  It's still worth fixing this, as people will be using this for pre-API-30 device support for years to come.

## API Reference Results:
**WindowManagerLayoutParams**
image

**View**
image

### DIFF
--- a/src/Mono.Android/Android.Views/WindowManagerLayoutParams.cs
+++ b/src/Mono.Android/Android.Views/WindowManagerLayoutParams.cs
@@ -4,7 +4,7 @@ namespace Android.Views
 {
 	partial class WindowManagerLayoutParams
 	{
-#if NET
+#if NET && ANDROID_34
 		[global::System.Runtime.Versioning.ObsoletedOSPlatform ("android30.0", "These flags are deprecated. Use WindowInsetsController instead.")]
 		public SystemUiFlags SystemUiFlags {
 			get => (SystemUiFlags) SystemUiVisibility;
--- a/src/Mono.Android/Android.Views/WindowManagerLayoutParams.cs
+++ b/src/Mono.Android/Android.Views/WindowManagerLayoutParams.cs
@@ -1,0 +1,15 @@
+using System;
+
+namespace Android.Views
+{
+	partial class WindowManagerLayoutParams
+	{
+#if NET
+		[global::System.Runtime.Versioning.ObsoletedOSPlatform ("android30.0", "These flags are deprecated. Use WindowInsetsController instead.")]
+		public SystemUiFlags SystemUiFlags {
+			get => (SystemUiFlags) SystemUiVisibility;
+			set => SystemUiVisibility = (Android.Views.StatusBarVisibility) value;
+		}
+#endif
+	}
+}
--- a/src/Mono.Android/Android.Telecom/InCallService.cs
+++ b/src/Mono.Android/Android.Telecom/InCallService.cs
@@ -6,7 +6,7 @@ namespace Android.Telecom
 	public abstract partial class InCallService : Android.App.Service
 	{
 #if ANDROID_23
-		[Obsolete ("Incorrect enum parameter, use the overload that takes a CallAudioRoute paramter instead.")]
+		[Obsolete ("Incorrect enum parameter, use the overload that takes a CallAudioRoute parameter instead.")]
 		[global::System.Runtime.Versioning.SupportedOSPlatformAttribute ("android23.0")]
 		public void SetAudioRoute ([global::Android.Runtime.GeneratedEnum] Android.Telecom.VideoQuality route)
 		{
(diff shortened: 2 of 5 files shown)


src/Mono.Android/Android.Views/WindowManagerLayoutParams.cs
```diff
@@ -0,0 +1,15 @@
+using System;
+
+namespace Android.Views
+{
+	partial class WindowManagerLayoutParams
+	{
+#if NET
```
jonpryor, line 7: Given that we'll eventually do a `net7.0-android34.0` release, requiring cherry-picks to a release/7.0* branch, should this instead be:

```csharp
#if NET && ANDROID_34
```

jpobst: Guarded just in case.  However, I think we should expect `net7.0-android34.0` API may not exactly match `net8.0-android34.0` and we shouldn't backport these fixes.

I don't suspect a user upgrading to .NET 8, using new API, and then wanting to downgrade their app to .NET 7 is a common scenario.

jonpryor: @jpobst: I'm more concerned about the opposite: customer *starts with* net7.0-android34.0 -- probably released in August or September! -- then migrating to net8.0-android34.0 in *November*.

Then there is ABI compatibility: libraries built against net7.0-android34.0 should work as-is in a net8.0-android34.0 app.  Will it?

jpobst: That should not be an issue with this PR, as it only adds API to newer versions, but it is good to keep that in mind in general.

jonpryor:
Re-reading this, and I feel like this is incomplete?

The PR description states:

> `WindowManagerLayoutParams.SystemUiVisibiliy` ([source](https://developer.android.com/reference/android/view/View#setSystemUiVisibility(int))) was incorrectly enumified as `StatusBarVisibility` instead of `SystemUiFlags`.  

However, the `source` link is to `View.setSystemUiVisibility()`, while it mentions `WindowManagerLayoutParams.SystemUiVisibiliy`.  These are not the same.

Similarly, the PR contents adds a `WindowManagerLayoutParams.SystemUiFlags` property, but issue #4290 was about *`View.SystemUiVisibiliy`*.

Shouldn't this change hit *both* `WindowManagerLayoutParams` *and* `View`?

jonpryor:
```markdown
Fixes: https://github.com/xamarin/xamarin-android/issues/4290

The [`View.getSystemUiVisibility()` method][0], bound as the
[`View.SystemUiVisibility` property][1], and the
[`WindowManager.LayoutParams.systemUiVisibility` field][2], bound as
the [`WindowManagerLayoutParams.SystemUiVisibility` property][3],
were incorrectly enumified to use the `StatusBarVisibility` enum when
they should have used the `SystemUiFlags` enum.

We cannot change the existing members, as that would be an ABI and
API break.  Furthermore, as these are properties, we cannot overload
them either.

"Fix" this by:

 1. Marking the existing properties as `[Obsolete]`, and
 2. Introduce *new* `SystemUiFlags` properties.

Additionally, these members were deprecated in API-30, so make sure
we reflect that on the new method.  It's still worth fixing this, as
people will be using this for pre-API-30 device support for years to
come.

The resulting API is:

	partial class View {
	    // Original property
	    [Obsolete ("This property has an incorrect enumeration type. Use the SystemUiFlags property instead.")]
	    public StatusBarVisibility SystemUiVisibility {get; set;}

	    // New hawtness
	    [ObsoletedOSPlatform ("android30.0", "…")]
	    public SystemUiFlags SystemUiFlags {get; set;}
	}
	partial class WindowManagerLayoutParams {
	    // Original property
	    [Obsolete ("This property has an incorrect enumeration type. Use the SystemUiFlags property instead.")]
	    public StatusBarVisibility SystemUiVisibility {get; set;}

	    // New hawtness
	    [ObsoletedOSPlatform ("android30.0", "…")]
	    public SystemUiFlags SystemUiFlags {get; set;}
	}

[0]: https://developer.android.com/reference/android/view/View#getSystemUiVisibility()
[1]: https://learn.microsoft.com/en-us/dotnet/api/android.views.view.systemuivisibility?view=xamarin-android-sdk-12
[2]: https://developer.android.com/reference/android/view/WindowManager.LayoutParams#systemUiVisibility
[3]: https://learn.microsoft.com/en-us/dotnet/api/android.views.windowmanagerlayoutparams.systemuivisibility?view=xamarin-android-sdk-12
```

PR #7731, opened 2023-01-23
[ci] Improvements for emulator test jobs

pjcollins:
The setup execution time for non-device tests has been improved by
moving a handful of adb commands from BaseTest to DeviceTest setup.
These would run for nearly every test run and result in a lot of waiting
due to some RunProcess failures and timeouts.

The output from the shell command we run to check if a device is online
is now cached and only refreshed when required by certain tests/asserts.

The `AssertHasDevices` check has been consolidated into `DeviceTest`
setup, and removed from individual tests in most cases.

Attempts to restart the emulator if a test determines that it is
inaccessible have been fixed, and emulator data will not be reset in
this case.

I played around with a handful of emulator launch settings and found
that removing the no-boot-anim and headless UI options improved the
reliability of nightly LocalizationTests, and resulted in faster boot
times locally. These options are now disabled for those tests.

The `DeploymentTest` class has been replaced by `TimeZoneInfoTests` and
`LocalizationTests`, and the other tests in that class have been moved
to `InstallAndRunTests`. `TimeZoneInfoTests` and `LocalizationTests`
will now validate command line output rather than trying to press a
button and read a UI element. This should make them more reliable and
faster.

The `TimeZoneInfoTests` and `LocalizationTests` suites have been moved
into separate test stages in the nightly test job.

All nightly tests now run against our .NET build/test artifacts rather
than classic XA.

An issue that caused LocalizationTests nodes 11 and 12 to not contain
any tests has been fixed.

An issue that caused LocalizationTests node 1 to also run tests from
node 11 and node 12 has been fixed.

Test result attachments have been fixed for `TimeZoneInfoTests` and
`LocalizationTests`, ensuring that we capture the right logcat and build
files for each test variant.

An issue causing binlog files to be overwritten by tests that use
multiple build targets has been fixed by naming the binlog file after
the build log file.

Console output verbosity has been set to normal for dotnet test
invocations, which should help with debugging.

Test NUnit packages have been updated to their latest versions.


pjcollins:
Testing this with https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7228766&view=results.

pjcollins:
New test run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7248077&view=results

pjcollins:
Latest nightly test run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7369649&view=results

pjcollins:
Latest nightly test run: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7374786&view=results

pjcollins:
I think this is getting pretty close so opening it up for some review. There are still reliability issues, but I think this is move in the right direction.

After moving all of the nightly tests to run against our .NET test assemblies I've been seeing some SSL related issues on API 23 and below which may be real issues:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7369649&view=ms.vss-test-web.build-test-results-tab&runId=69041435&resultId=100139&paneView=debug

tests/MSBuildDeviceIntegration/Tests/XASdkDeployTests.cs
```diff
@@ -71,8 +71,6 @@ public class XASdkDeployTests : DeviceTest
 		[TestCaseSource (nameof (DotNetInstallAndRunSource))]
 		public void DotNetInstallAndRun (bool isRelease, bool xamarinForms, string targetFramework)
 		{
-			AssertHasDevices ();
```
jonathanpeppers, line 74: Some of these if I run them locally and forgot to attach a device, it would give a nice error message. Was there a reason to remove it?

tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
```diff
@@ -248,7 +236,6 @@ public override void OnCreate ()
 					EvaluationOptions = EvaluationOptions.DefaultOptions,
 				};
 				options.EvaluationOptions.UseExternalTypeResolver = true;
-				ClearAdbLogcat ();
```
dellis1972, line 251: Is this needed to clear out any logs from previous runs?

tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
```diff
@@ -248,7 +236,6 @@ public override void OnCreate ()
...
 				b.BuildLogFile = "run.log";
 				Assert.True (b.RunTarget (proj, "_Run", doNotCleanupOnUpdate: true, parameters: new string [] {
```
dellis1972, line 240: Should this be refactored to use the new `RunProjectAndAssert` ?
I know it is setting the debug port as well, but `RunProjectAndAssert` might be able to do both.

tests/MSBuildDeviceIntegration/Tests/MonoAndroidExportTest.cs
```diff
@@ -118,7 +118,6 @@ protected override void OnCreate (Bundle bundle)
 	</application >
 </manifest>";
 				Assert.True (b.Install (proj), "Project should have installed.");
-				ClearAdbLogcat ();
 				b.BuildLogFile = "run.log";
 				Assert.True (b.RunTarget (proj, "StartAndroidActivity", doNotCleanupOnUpdate: true), "Project should have run.");
```
dellis1972, line 122: Should this use `RunProjectAndAssert` ?

dellis1972:
This is all good stuff. 

It might be nice however to keep at least one test which checks a Button Click? (If we removed them all). Since this helps to confirm that our Events works and possiblly that the CodeBehind stuff works.

tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
pjcollins: `ClearAdbLogcat() ` was added to test setup in https://github.com/xamarin/xamarin-android/commit/b485425dc80c16d22a4a4bb72163a3a062f98e93, so I removed a handful of calls to it in this PR that seemed redundant.

tests/MSBuildDeviceIntegration/Tests/XASdkDeployTests.cs
pjcollins: I moved this check into DeviceTest setup so we don't need to define it on every test. We will also attempt to acquire an android target during setup now as well if one is not available: https://github.com/xamarin/xamarin-android/pull/7731/files/0ffe290ec63c4eb15ed1bb7b44811d7ba6007125#diff-b90ec706ed5e04747e6db2983d908cf1dcc5e7ca4fae94575961905975dc886fR97

build-tools/automation/azure-pipelines-nightly.yaml
```diff
@@ -158,70 +160,338 @@ stages:
...
+          testRunTitle: CheckTimeZoneInfoIsCorrectNode17 On Device - macOS
+          testAssembly: $(System.DefaultWorkingDirectory)/bin/Test$(XA.Build.Configuration)/MSBuildDeviceIntegration/$(DotNetStableTargetFramework)/MSBuildDeviceIntegration.dll
+          useDotNet: true
+          dotNetTestExtraArgs: --filter "Name=CheckTimeZoneInfoIsCorrectNode17"
+          testResultsFile: TestResult-CheckTimeZoneInfoIsCorrectNode17-$(XA.Build.Configuration).xml
+      - template: run-nunit-tests.yaml
+        parameters:
+          testRunTitle: CheckTimeZoneInfoIsCorrectNode18 On Device - macOS
+          testAssembly: $(System.DefaultWorkingDirectory)/bin/Test$(XA.Build.Configuration)/MSBuildDeviceIntegration/$(DotNetStableTargetFramework)/MSBuildDeviceIntegration.dll
+          useDotNet: true
```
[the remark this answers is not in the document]
pjcollins: I had started to do this in a separate PR - https://github.com/xamarin/xamarin-android/pull/7826. I think we can also run the desktop test assemblies with a stable .NET version which may improve reliability, even though our tests will still invoke the preview version that we provision.

pjcollins:
> It might be nice however to keep at least one test which checks a Button Click? (If we removed them all). Since this helps to confirm that our Events works and possiblly that the CodeBehind stuff works.

The DeploymentTests class I removed had two other tests that click a button, I've moved those into [InstallAndRunTests](https://github.com/xamarin/xamarin-android/blob/dev/pjc/nightly-locstage/tests/MSBuildDeviceIntegration/Tests/InstallAndRunTests.cs#L840). These will hopefully continue to provide this coverage, and we can add more here in the future as needed.

tests/MSBuildDeviceIntegration/Tests/MonoAndroidExportTest.cs
pjcollins: Yeah I think it is probably best for us to exercise our run target everywhere we can, I will try look for more instances like this.

tests/MSBuildDeviceIntegration/Tests/DebuggingTest.cs
pjcollins: Fixed.

tests/MSBuildDeviceIntegration/Tests/MonoAndroidExportTest.cs
pjcollins: Fixed.

pjcollins:
Previous nightly was close to green with one re-run, https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7396250&view=results.

Queueing one more against this branch to see how it fairs:
https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7411683&view=results

I think this is now in a place where it can be merged and we can continue to iterate on smaller improvements.

jpobst (approved):
:crossed_fingers: for more stable tests!

In future PRs, we can look at `dotnet-test-slicer` and `start-stop-emulator.yaml` as possible improvements.

dellis1972 (approved):

jonathanpeppers (approved):

PR #7732, opened 2023-01-24
Prevent overlapped decompression of embedded assemblies

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/7335

Commit d236af54538ef1fe62d0125798cdde78e46dcd8e introduced embedded assembly compression, 
using the lz4 algorithm, which speeds up startup and reduces final package size.

Assemblies are compressed at build time and, at the same time, pre-allocated buffers for
the **decompressed** data are allocated in `libxamarin-app.so`.  The buffers are then passed
to the LZ4 APIs, all threads using the same output buffer.  The assumption was that we can
do fine without locking as even if overlapped decompression happens, the output data will be
the same and so even if two threads do the same thing at the same time, the data will be valid
at all times, so long as at least one thread completes the decompression.  

This assumption proved to be **largely** true, but it appears that in high concurrency cases it
is possible that the data in decompression buffer differs.  My guess is that LZ4 either uses the 
output buffer as a scratchpad area when decompressing or that it initializes/modifies the buffer
before writing actual data in it.  With overlapped decompression, it may lead to one thread
overwriting valid data previously written by another thread, so that when the latter returns the
buffer it thought to have had valid data may contain certain bytes temporarily overwritten by the
decompression session in the other, still running, thread.  It may happen that MonoVM reads the
corrupted data just when it is still invalid (before the still running decompression session actually
writes the valid data), a classic race condition.

To fix this, the decompression block is now protected with a startup-aware mutex.  Mutex will be held
only after the initial startup phase is completed, so there should not be much loss of startup
performance.

### DIFF
--- a/src/monodroid/jni/embedded-assemblies.cc
+++ b/src/monodroid/jni/embedded-assemblies.cc
@@ -360,9 +360,6 @@ EmbeddedAssemblies::assembly_store_open_from_bundles (dynamic_local_string<SENSI
 		len -= sizeof(SharedConstants::DLL_EXTENSION) - 1;
 	}
 
-	std::string clipped_name;
-	clipped_name.assign (name.get (), len);
-
 	hash_t name_hash = xxhash::hash (name.get (), len);
 	log_debug (LOG_ASSEMBLY, "assembly_store_open_from_bundles: looking for bundled name: '%s' (hash 0x%zx)", name.get (), name_hash);
 
--- a/src/monodroid/jni/embedded-assemblies.cc
+++ b/src/monodroid/jni/embedded-assemblies.cc
@@ -104,6 +104,7 @@ EmbeddedAssemblies::get_assembly_data (uint8_t *data, uint32_t data_size, [[mayb
 			if (cad.loaded) {
 				set_assembly_data_and_size (reinterpret_cast<uint8_t*>(cad.data), cad.uncompressed_file_size, assembly_data, assembly_data_size);
 				return;
+			}
 
 			if (XA_UNLIKELY (cad.data == nullptr)) {
 				log_fatal (LOG_ASSEMBLY, "Invalid compressed assembly descriptor at %u: no data", header->descriptor_index);
--- a/src/monodroid/jni/embedded-assemblies.cc
+++ b/src/monodroid/jni/embedded-assemblies.cc
@@ -35,6 +35,7 @@
 #include "mono-image-loader.hh"
 #include "xamarin-app.hh"
 #include "cpp-util.hh"
+#include "monodroid-glue-internal.hh"
 #include "startup-aware-lock.hh"
 #include "timing-internal.hh"
 #include "search.hh"
@@ -74,6 +75,13 @@ void EmbeddedAssemblies::set_assemblies_prefix (const char *prefix)
 }
 
 force_inline void
+EmbeddedAssemblies::set_assembly_data_and_size (uint8_t* source_assembly_data, uint32_t source_assembly_data_size, uint8_t*& dest_assembly_data, uint32_t& dest_assembly_data_size) noexcept
+{
+	dest_assembly_data = source_assembly_data;
+	dest_assembly_data_size = source_assembly_data_size;
+}
+
+force_inline void
 EmbeddedAssemblies::get_assembly_data (uint8_t *data, uint32_t data_size, [[maybe_unused]] const char *name, uint8_t*& assembly_data, uint32_t& assembly_data_size) noexcept
 {
 #if defined (ANDROID) && defined (HAVE_LZ4) && defined (RELEASE)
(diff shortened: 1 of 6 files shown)


dellis1972 (approved):

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -115,29 +124,23 @@ EmbeddedAssemblies::get_assembly_data (uint8_t *data, uint32_t data_size, [[mayb
 			const char *data_start = reinterpret_cast<const char*>(data + sizeof(CompressedAssemblyHeader));
 			int ret = LZ4_decompress_safe (data_start, reinterpret_cast<char*>(cad.data), static_cast<int>(assembly_data_size), static_cast<int>(cad.uncompressed_file_size));
 
-			if (XA_UNLIKELY (log_timing)) {
-				internal_timing->end_event (decompress_time_index, true /* uses_more_info */);
-				internal_timing->add_more_info (decompress_time_index, name);
```
jonathanpeppers, line 120: Was this timing message not that helpful?

grendello: > Was this timing message not that helpful?

Yeah, I used it initially to see how much decompression itself costs us, but right now it's just wasted CPU cycles.

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -81,31 +89,32 @@ EmbeddedAssemblies::get_assembly_data (uint8_t *data, uint32_t data_size, [[mayb
 	if (header->magic == COMPRESSED_DATA_MAGIC) {
 		if (XA_UNLIKELY (compressed_assemblies.descriptors == nullptr)) {
 			log_fatal (LOG_ASSEMBLY, "Compressed assembly found but no descriptor defined");
-			exit (FATAL_EXIT_MISSING_ASSEMBLY);
+			abort ();
```
jonpryor, line 92: Why `abort()` and not `exit()`?  I forget what `abort()` *does* on Android…

grendello: `exit()` causes the application to restart, `abort()` kills it with a native `SIGABRT` stack trace

jonathanpeppers: There are a lot of places that users complain about a "white screen", and it ended up their app was crashing and restarting.

Should we audit all of the `exit()` calls?

grendello: Yes, we should, I've been meaning to do it for quite a while now. I'll open a separate PR for this.

grendello: PR #7734 opened

grendello:
Let's wait for PR #7734 to be merged first.

src/monodroid/jni/embedded-assemblies.cc
```diff
@@ -104,6 +104,7 @@ EmbeddedAssemblies::get_assembly_data (uint8_t *data, uint32_t data_size, [[mayb
 			if (cad.loaded) {
 				set_assembly_data_and_size (reinterpret_cast<uint8_t*>(cad.data), cad.uncompressed_file_size, assembly_data, assembly_data_size);
 				return;
+			}
```
dellis1972, line 107: glad I'm not the only one who makes these kinds of typo's 🤣 👍

grendello: :D manual merge conflict resolution FTW :)

jonpryor:
```markdown
[monodroid] Prevent overlapped decompression of embedded assemblies (#7732)

Fixes: https://github.com/xamarin/xamarin-android/issues/7335

Commit d236af54 introduced embedded assembly compression, using lz4,
which speeds up startup and reduces final package size.

Assemblies are compressed at build time and, at the same time, pre-
allocated buffers for the **decompressed** data are allocated in
`libxamarin-app.so`.  The buffers are then passed to the LZ4 APIs,
all threads using the same output buffer.  The assumption was that we
an do fine without locking as even if overlapped decompression happens,
the output data will be the same and so even if two threads do the
same thing at the same time, the data will be valid at all times, so
long as at least one thread completes the decompression.  

This assumption proved to be **largely** true, but it appears that in
high concurrency cases it is possible that the data in the
decompression buffer differs.  This can result in app crashes:

	A/libc: Fatal signal 6 (SIGABRT), code -1 (SI_QUEUE) in tid 3092 (.NET ThreadPool), pid 2727 (myapp.name)
	A/DEBUG: pid: 2727, tid: 3092, name: .NET ThreadPool  >>> myapp.name <<<
	A/DEBUG:       #01 pc 0000000000029b1c  /data/app/myapp.name-B9t_3dF9i8mDxJEKodZw5w==/split_config.arm64_v8a.apk!libmono-android.release.so (offset 0x103d000) (xamarin::android::internal::MonodroidRuntime::mono_log_handler(char const*, char const*, char const*, int, void*)+144) (BuildId: 29c5a3805a0bedee1eede9b6668d7c676aa63371)
	A/DEBUG:       #02 pc 00000000002680bc  /data/app/myapp.name-B9t_3dF9i8mDxJEKodZw5w==/split_config.arm64_v8a.apk!libmonosgen-2.0.so (offset 0x109b000) (BuildId: 4a5dd4396e8816b7f69881838bd549285213d53b)
	A/DEBUG:       #03 pc 00000000002681e8  /data/app/myapp.name-B9t_3dF9i8mDxJEKodZw5w==/split_config.arm64_v8a.apk!libmonosgen-2.0.so (offset 0x109b000) (BuildId: 4a5dd4396e8816b7f69881838bd549285213d53b)
	A/DEBUG:       #04 pc 000000000008555c  /data/app/myapp.name-B9t_3dF9i8mDxJEKodZw5w==/split_config.arm64_v8a.apk!libmonosgen-2.0.so (offset 0x109b000) (mono_metadata_string_heap+188) (BuildId: 4a5dd4396e8816b7f69881838bd549285213d53b)
	…

My guess is that LZ4 either uses the output buffer as a scratchpad
area when decompressing or that it initializes/modifies the buffer
before writing actual data in it.  With overlapped decompression, it
may lead to one thread overwriting valid data previously written by
another thread, so that when the latter returns the buffer it thought
to have had valid data may contain certain bytes temporarily
overwritten by the decompression session in the other, still running,
thread.  It may happen that MonoVM reads the corrupted data just when
it is still invalid (before the still running decompression session
actually writes the valid data), a classic race condition.

To fix this, the decompression block is now protected with a startup-
aware mutex.  Mutex will be held only after the initial startup phase
is completed, so there should not be much loss of startup performance.
```

PR #7734, opened 2023-01-24
Replace all `exit()` calls with `abort()` in native code

grendello:
`exit()` causes the application to restart, which can be both annoying and 
confusing when the app seemingly shows something (usually a white screen) 
and yet it doesn't work.  The only way to learn what's going on is to look 
at the logcat output to see the error message and a restart.

`abort()`, however, fully terminates the application and also shows a native 
stack trace, which makes it easy to find the cause of the "crash" (the error 
message will be logged typically just before the native stack trace is printed)

### DIFF
--- a/src/monodroid/jni/debug-app-helper.cc
+++ b/src/monodroid/jni/debug-app-helper.cc
@@ -91,7 +91,7 @@ Java_mono_android_DebugRuntime_init (JNIEnv *env, [[maybe_unused]] jclass klass,
 	void *monosgen = dlopen (monosgen_path, RTLD_LAZY | RTLD_GLOBAL);
 	if (monosgen == nullptr) {
 		log_fatal (LOG_DEFAULT, "Failed to dlopen Mono runtime from %s: %s", monosgen_path, dlerror ());
-		exit (FATAL_EXIT_CANNOT_FIND_LIBMONOSGEN);
+		abort ();
 	}
 }
 
@@ -276,7 +276,7 @@ get_libmonosgen_path ()
 
 	log_fatal (LOG_DEFAULT, "Do you have a shared runtime build of your app with AndroidManifest.xml android:minSdkVersion < 10 while running on a 64-bit Android 5.0 target? This combination is not supported.");
 	log_fatal (LOG_DEFAULT, "Please either set android:minSdkVersion >= 10 or use a build without the shared runtime (like default Release configuration).");
-	exit (FATAL_EXIT_CANNOT_FIND_LIBMONOSGEN);
+	abort ();
 
 	return libmonoso;
 }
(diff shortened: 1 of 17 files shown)


src/monodroid/jni/debug-app-helper.cc
```diff
@@ -91,7 +91,7 @@ Java_mono_android_DebugRuntime_init (JNIEnv *env, [[maybe_unused]] jclass klass,
 	void *monosgen = dlopen (monosgen_path, RTLD_LAZY | RTLD_GLOBAL);
 	if (monosgen == nullptr) {
 		log_fatal (LOG_DEFAULT, "Failed to dlopen Mono runtime from %s: %s", monosgen_path, dlerror ());
-		exit (FATAL_EXIT_CANNOT_FIND_LIBMONOSGEN);
```
jonathanpeppers, line 94: Are all of these `FATAL_EXIT_*` constants somewhere, where we could delete those, too?

jonpryor: the `FATAL_EXIT_*` constants are in xamarin/java.interop, some of which are used there.

grendello:
I'm investigating the size increase of `libmonodroid.so`

src/monodroid/jni/debug.cc
```diff
@@ -474,7 +474,7 @@ Debug::process_cmd (int fd, char *cmd)
 		log_info (LOG_DEFAULT, "Debugger requested an exit, will exit immediately.\n");
 		fflush (stdout);
 		fflush (stderr);
-		exit (0);
+		Helpers::abort_application ();
```
jonpryor, line 477: This gives me pause; the original `exit(0)` is not signaling an *error* condition, so is calling `abort_application()` appropriate?

As this is part of the debugging infrastructure, I think it may be safer to keep this as `exit(0)`, unless you want to do an end-to-end test to exercise this behavior and see what happens…

jonpryor:
Context: https://discord.com/channels/732297728826277939/732297837953679412/1067806732983746621
Context: https://discord.com/channels/732297728826277939/732297837953679412/1067822882274689024

calls to `abort()` are replaced with `Helpers::abort_application()` because the `abort()` calls were contributing to a ~30% increase in the size of `libmonodroid.so`.  Consolidating the `abort()` calls into `abort_application()` removed the size increase.

src/monodroid/jni/debug.cc
grendello: The problem here is that the intention is to terminate the application
but `exit` doesn't do that while `abort` does.

From the pov of the debugger the application dies anyway
but from the pov of the developer, it will immediately restart itself
which is rather weird

jonpryor:
```markdown
Context: https://discord.com/channels/732297728826277939/732297837953679412/1067806732983746621
Context: https://discord.com/channels/732297728826277939/732297837953679412/1067822882274689024

In certain "bad app" scenarios, we would print a message and then
**exit**(3) the app.  The problem with using `exit()` in this manner
is that it *can* result in the app constantly relaunching, "spamming"
the UI.

Improve this by calling **abort**(3) instead.  This prevents Android
from constantly re-launching the app, and also prints useful abort
information such as registers and a native stack trace to `adb logcat`.

Unexpectedly, however, "simply" replacing app calls to `exit()` with
calls to `abort()` resulted in `libmonodroid.so` increasing in size
by ~16%.  For some reason, calling `abort()` caused the compiler to
inline much more code and in a weird manner (it repeated essentially
identical blocks of code from an inlined function, which were
previously folded into a single block when `exit()` was used).

We found that consolidating the `abort()` calls into
`Helpers::abort_application()` removed the size increase.
```

PR #7735, opened 2023-01-24
[Mono.Android] Fix incorrect Context.RegisterReceiver enumification.

### LINKED ISSUE #7503: Argument type of Context.RegisterReceiver() is different from native API
### Android application type

Classic Xamarin.Android (MonoAndroid12.0, etc.)

### Affected platform version

MonoAndroid 13.0, Xamarin.Android 13.1.0.1 , VS Mac 17.4

### Description

I am currently updating the version of the target SDK for my Xamarin.Android project to version 33.

One of the new features of Android 13 is "Safer exporting of context-registered receivers".
https://developer.android.com/about/versions/13/features#runtime-receivers

To make my app compatible with this, I was going to change the RegisterReceiver call point as follows:

```diff
- Context.RegisterReceiver(receiver, filter);
+ Context.RegisterReceiver(receiver, filter, ReceiverFlags.Exported);
```

This implementation conforms to the specification of the following method:                                                                                                                                                                                                                                                                                                                         https://developer.android.com/reference/android/content/Context#registerReceiver(android.content.BroadcastReceiver,%20android.content.IntentFilter,%20java.lang.String,%20android.os.Handler)

However, the above code failed to compile.
After investigating the cause, I found that the third argument of MonoAndroid's RegisterReceiver() overload is not of type Int, but ActivityFlags. This differs from the native API specification.

Please let me know why this is happening and if MonoAndroid plans to fix it.

### Steps to Reproduce

1. Create new Xamarin.Android project that targeting MonoAndroid 13.0
2. Call the following overloads of RegisterReceiver at any position: 
`RegisterReceiver(BroadcastReceiver, IntentFilter, Int)`

 I have confirmed that `ReceiverFlags` can be used with `ContextCompat` API and will consider replacing it with ContextCompat in the near future if this issue is not resolved.   
https://developer.android.com/reference/androidx/core/content/ContextCompat#registerReceiver(android.content.Context,android.content.BroadcastReceiver,android.content.IntentFilter,int)

### Relevant log output

_No response_

jpobst:
Fixes #7503 

`Android.Content.Context.RegisterReceiver(BroadcastReceiver, IntentFilter, int)` (+1 overload) was incorrectly enumified as `ActivityFlags` when it should have been `ReceiverFlags`.

To fix without breaking API:
- Create new overloads methods that chain to the existing incorrectly enumified methods.
- Mark the existing methods as `[Obsolete]` with description.

Additionally, it was noticed that the public API tracking added in https://github.com/xamarin/xamarin-android/pull/8171 did not get flagged as an error, only a warning.  Change to an error to ensure any new public API is accounted for.


jonpryor:
What gives me pause is that the resulting binding makes an `abstract` method `[Obsolete]`, a'la:

```csharp
partial class Context {
    [Obsolete ("This method has an incorrect enumeration type. Use the overload that takes ReceiverFlags instead.")]
    public abstract Android.Content.Intent? RegisterReceiver (Android.Content.BroadcastReceiver? receiver, Android.Content.IntentFilter? filter, [global::Android.Runtime.GeneratedEnum] Android.Content.ActivityFlags flags);
}
```

which I believe means that every override also needs to be `[Obsolete]`, hence the changes in this PR to `ContextWrapper` and `MockContext`.  This feels "weird".

On the plus side, (effectively) nobody should ever override `Context.RegisterReceiver()`, so this should be safe.

jpobst:
> This feels "weird".

Agreed.

> On the plus side, (effectively) nobody should ever override Context.RegisterReceiver(), so this should be safe.

This is the rationale I eventually went with.  ~No one should override this, users should only be calling the system provided version(s) of it.

PR #7741, opened 2023-01-25
Bump to xamarin/monodroid/release/7.0.1xx@9454bde

jonathanpeppers:
Changes: https://github.com/xamarin/monodroid/compare/8f925e3...9454bde

* [tools/msbuild] Fix System App detection
* [tools/msbuild] Deploy *resources.dll from PackageReference

I branched monodroid at the appopriate commit at:

https://github.com/xamarin/monodroid/tree/release/7.0.1xx


jonathanpeppers:
There are some MSBuild perf test failures, but looking at the build logs, I think we can ignore them:

image

It seems like `java -version` and `javac -version` are somewhat slow?

PR #7750, opened 2023-01-27
Address issues found by LGTM

grendello:
Fix, address or disable issues found by LGTM scan.


.lgtm.yml
```diff
@@ -0,0 +1,2 @@
+queries:
+  - exclude: cs/campaign/constantine
```
jonpryor, line 2: @grendello: what's *this* do?

src/monodroid/jni/cpp-util.hh
```diff
@@ -112,7 +112,7 @@ namespace xamarin::android
 		// `sizeof... (Length)` part which subtracts the number of template parameters - the amount of NUL bytes so that
 		// we don't waste space.
 		constexpr size_t total_length = (... + Length) - sizeof... (Length);
-		char_array<total_length + 1> ret;
+		char_array<total_length + 1> ret; // lgtm [cpp/paddingbyteinformationdisclosure] the buffer is filled in the loop below
```
jonpryor, line 115: The docs warn about "uninitialized padding bytes", which I *imagine* could happen if we had:

```cpp
auto a = concat_const ("foo", "bar", "baz");
auto b = concat_const ("qux", "quux");
```

wherein sizeof(a) is 10, which *won't* be stack aligned, necessitating padding between `a` and `b`.

But I also wonder if I'm overthinking that example?

Additionally, the only places I see `concat_const()` being used are for `static constexpr auto` members, which are *not* local variables, so any padding between static members would by definition be zero-initialized…

But if that *weren't* the case, and we *were* using this for local variables…  Could this be a problem?  I'm still dubious, but I'm not completely certain.

jonpryor:
Draft commit message:

```markdown
[lgtm] Fix LGTM-reported issues. (#7750)

Remember CodeQL (acfc1efe)?  CodeQL basically runs [GitHub LGTM][0]
on source code, looking for possible security issues.

Now that CodeQL is running, we can begin addressing reported issues.

Problems found include:

  * HttpClient created with CheckCertificateRevocationList disabled
  * Wrong type of arguments to formatting function
  * Weak cryptography
  * Possible information leakage from uninitialized padding bytes
  * ML Training and Serialization Files Referenced

~~ HttpClient created with CheckCertificateRevocationList disabled ~~

Apparently the `HttpClient` default constructor is "bad"; we should
instead use the [`HttpClient(HttpMessageHandler)` constructor][1],
provide our own `HttpClientHandler`, and ensure that
[`HttpClientHandler.CheckCertificateRevocationList`][2] is True.

~~ Wrong type of arguments to formatting function ~~

Apparently LGTM doesn't realize that in C++ `long int` is synonymous
with `long`, and thus warns that they're not the same. 🤦
Remove a cast to `long int`.

~~ Weak cryptography ~~

This is in `AuthDigestSession.cs`.  Unfortunately, RFC2617 requires
MD5, so we kinda need to use MD5.  Add a `// lgtm [cs/weak-crypto]`
comment to disable the warning.

~~ Possible information leakage from uninitialized padding bytes ~~

This is in `cpp-util.hh`, and it seems that LGTM doesn't appreciate
our use of template metaprogramming to construct a `char_array<Len+1>`
wherein `Len` is computed at compile time with no wasted padding.

~~ ML Training and Serialization Files Referenced ~~

LGTM apparently assumes that mentions of `.pb` are mentions of ML
data training files.  In our case, these were part of error messages
from `aapt2` that we were attempting to translate.

Add a `//lgtm [csharp/responsible-ai/ml-training-and-serialization-files-referenced]`
comment to disable this warning.

Co-authored-by: Alex Hsu <csigs@users.noreply.github.com>

[0]: https://github.com/marketplace/lgtm
[1]: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.-ctor?view=netstandard-2.0#system-net-http-httpclient-ctor(system-net-http-httpmessagehandler)
[2]: https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler.checkcertificaterevocationlist?view=net-7.0
```

.lgtm.yml
grendello: It disables the particular query/check globally, for the entire repo. This one was a campaign asking for contact with regards to certain constructs, not an error or warning.

src/monodroid/jni/cpp-util.hh
grendello: I think this is about the `+ 1` byte. `total_length` is a compile-time constant computed without counting the trailing `\0` bytes of input strings, and so the array length "should" be `total_length` from the POV of the analyzer - the extra byte is seen as uninitialized "padding", while it is a `\0` C char termination byte in reality, set one line below this one. 

I think the analyzer gets confused by the template packs usage in both the declaration and the loop below (or maybe it simply doesn't analyze the loop at all). Whichever is the case, it's not a concern and not a bug, so we can hush it without worries.

PR #7757, opened 2023-01-31
[tests] update `.apkdesc` files for failing tests

jonathanpeppers:
Context: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7264347&view=ms.vss-test-web.build-test-results-tab&runId=67617676&resultId=100033&paneView=debug

A single test was failing on Windows-only:

    BuildReleaseArm64(True)
    apkdiff regression test failed with exit code: 3
    stdErr: Error: apkdiff: PackageSize decrease 8,192 is 3,072 bytes more than the threshold 5,120. apk1 size: 9,521,310 bytes, apk2 size: 9,513,118 bytes.
    Error: apkdiff: Size regression occured, 1 check(s) failed.

Most notably a single file changed:

    14,556 classes.dex

I believe this started failing in 35db5272, due to a combination of changes:

* [dotnet/runtime#77386][0] introduced `libSystem.Security.Cryptography.Native.Android.jar` in the Mono runtime packs.

* We updated `.apkdesc` files *before* including the new `.jar`.

* We added appropriate logic to include the new `.jar`.

The result was we have `.apkdesc` files that don't match the full set of changes in 35db5272. Then in 8872d04 the app size changed beyond the threshold for a test to fail.

[0]: https://github.com/dotnet/runtime/pull/77386

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Resources/Base/BuildReleaseArm64SimpleDotNet.apkdesc
@@ -8,46 +8,46 @@
       "Size": 1024
     },
     "assemblies/Java.Interop.dll": {
-      "Size": 58913
+      "Size": 58914
     },
     "assemblies/Mono.Android.dll": {
-      "Size": 87163
+      "Size": 87157
     },
     "assemblies/Mono.Android.Runtime.dll": {
-      "Size": 5895
+      "Size": 5867
     },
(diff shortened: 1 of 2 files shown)


dellis1972 (approved):

jonathanpeppers:
Need to investigate more:
```
  8,192 Package size difference -0.09% (of 9,521,310)
stdErr: Error: apkdiff: PackageSize decrease 8,192 is 3,072 bytes more than the threshold 5,120. apk1 size: 9,521,310 bytes, apk2 size: 9,513,118 bytes.
Error: apkdiff: Size regression occured, 1 check(s) failed.
Expected: True
But was:  False
```

jonathanpeppers:
I don't think this is needed anymore.

PR #7759, opened 2023-02-01
Maybe fix the `unable to open file as zip archive` error

grendello:
Context: https://gist.github.com/grendello/55b9a74c80c611dd48b726ee6f16d3c9

Sometimes in our tests we see an error from LibZipSharp which claims an file 
cannot be opened as a valid ZIP archive.  The reason for this is ZIP format 
corruption happening sometimes in the `BuildApk` task.

The way the process works is that we first run `aapt2` to produce an APK archive 
named `packaged_resources`, which we then copy to the destination APK (using 
standard `File.Copy`) and open it to append our content.  The first item we 
always append is the `classes.dex` file which is also the **only** corrupted e
ntry in the broken APK files.

ZIP files contain two records describing each entry in the archive:

  1. in the Central Directory (which is located at the end of the file)
  2. local header in the ZIP data stream, offset of whose is contained 
     in the above Central Directory record.

After close examination, it appears that the broken APK files contain `1` above 
for `classes.dex` but not `2`.  Instead of `2` we see a copy of `packaged_resource` 
package's Central Directory record.

The corruption may be caused by invalid stream position after we open the 
`packaged_resources` copy and append `classes.dex` entry to it.

Calling `Flush ()` after adding `classex.dex` **may** fix the problem.

If it doesn't, we should consider copying data from `packaged_resources` to the 
final APK entry by entry by decompressing them from the original archive and 
adding them to the destination one.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/BuildApk.cs
@@ -183,6 +183,7 @@ namespace Xamarin.Android.Tasks
 					string apkName = dex.GetMetadata ("ApkName");
 					string dexPath = string.IsNullOrWhiteSpace (apkName) ? Path.GetFileName (dex.ItemSpec) : apkName;
 					AddFileToArchiveIfNewer (apk, dex.ItemSpec, DalvikPath + dexPath, compressionMethod: dexCompressionMethod);
+					apk.Flush ();
 				}
 
 				if (EmbedAssemblies && !BundleAssemblies) {


dellis1972 (approved):

jonathanpeppers (approved):

dellis1972 (approved):

jonpryor:
Unified commit message:

```markdown
Fixes? https://github.com/xamarin/xamarin-android/issues/6067

Context: https://gist.github.com/grendello/55b9a74c80c611dd48b726ee6f16d3c9
Context: https://github.com/xamarin/LibZipSharp/pull/125

Context: https://github.com/xamarin/xamarin-android/issues/7622

Changes: http://github.com/xamarin/monodroid/compare/50faac94c6a0c27864564829ac83f3988c82f8ef...602aca98245744882a129206b79b5a5e093dae44

  * xamarin/monodroid@602aca982: Bump androidtools for new LibZipSharp (xamarin/monodroid#1287)
  * xamarin/monodroid@1f52d5873: [tools/msbuild] Deploy *resources.dll from PackageReference (xamarin/monodroid#1276)

Changes: https://github.com/xamarin/androidtools/compare/f11d16350b088742f1293a58a3078e99f7deca09...c0bcb66c0beadcb3d1488349ce0fc906f2f3d8e4

  * xamarin/androidtools@c0bcb66: Bump xamarin-android-tools for new LibZipSharp (xamarin/androidtools#377)

Changes: https://github.com/xamarin/xamarin-android-tools/compare/099fd95f5459d9df78af0ad28f46e3016dd7ca1d...dbe42bf1688e79e1ca474dff4a1f8060158cafd9

  * xamarin/xamarin-android-tools@dbe42bf: Bump LibZipSharp version (xamarin/xamarin-android-tools#204)

Sometimes in our tests we see an error from LibZipSharp which claims
a file cannot be opened as a valid ZIP archive:

	error ANDZA0000: Unable to open '…/example.apk' as zip archive

The reason for this is ZIP format corruption happening sometimes in
the `<BuildApk/>` task.

The way the process works is that we first run `aapt2` to produce
an APK archive named `packaged_resources`, which we then copy to the
destination APK (using standard `File.Copy()`) and open it to append
our content.  The first item we always append is the `classes.dex`
file which is also the **only** corrupted entry in the broken APK
files.

ZIP files contain two records describing each entry in the archive:

 1. in the Central Directory (located at the end of the file)

 2. local header in the ZIP data stream, offset of which is contained
    in the above Central Directory record.

After close examination, it appears that the broken APK files contain
(1) for `classes.dex` but not (2).  Instead of (2) we see a *copy*
of `packaged_resource` package's Central Directory record:

@grendello: what does this *mean*?  Is "a copy of `packaged_resources` …"
visible in `zipinfo -vv`?

	% zipinfo -vv example.apk
	# what do i look for here?

We believe that the corruption is caued by an invalid stream position
after we open the copy of `packaged_resources` and append
`classes.dex` to it; 

The corruption may be caused by invalid stream position after we open
the `packaged_resources` copy and append `classes.dex` entry to it;
see xamarin/LibZipSharp#125 for details.

Calling `Flush()` after adding `classex.dex` **may** fix the problem.

If it doesn't, we should consider copying data from
`packaged_resources` to the final APK entry by entry by decompressing
them from the original archive and adding them to the destination one.
```

PR #7770, opened 2023-02-06
[ci] Remove some Classic XA test stages/jobs.

jpobst:
Super easy win to kick off removing Classic from `main`:

Remove the `Legacy Tests` stage and `MSBuild+Emulator Legacy #X` jobs.

These definitely are already running .NET equivalents, and will reduce our CI per-run cost by ~11.5 CPU hours (out of ~39.5 hours total).


jonathanpeppers (approved):

pjcollins (approved):
I took a quick look at the tests we ignore in the .NET runs and don't think this would cause us to lose much coverage. I filed https://github.com/xamarin/xamarin-android/issues/7777 to take a closer look and port any important tests in the near-ish future.

Issue #7764, opened 2023-02-03
.NET 6 + $(BundleAssemblies)=True is broken


jonpryor:
### Android application type

Android for .NET (net6.0-android, etc.)

### Affected platform version

VS 2022 17.2+

### Description

.NET Android dropped support for `$(BundleAssemblies)`=True; see also:

  * https://github.com/xamarin/xamarin-android/commit/d236af54538ef1fe62d0125798cdde78e46dcd8e
  * https://github.com/xamarin/AndroidX/issues/64#issuecomment-609970584

However, this deprecation never made it to the build system.  Now, if you build a .NET 6+ app with `$(BundleAssemblies)` set, the app is broken.

See also: https://discord.com/channels/732297728826277939/732297837953679412/1071145486695792761

The problem is likely due to: https://github.com/xamarin/xamarin-android/blob/47a1df7b6911dc30b52164766984d3ed3c4f004e/src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets#L336

wherein `$(AndroidUseAssemblyStore)`=False when `$(BundleAssemblies)`=True.

*We should not support `$(BundleAssemblies)`=True*.  Rather, the problems are:

  * No warning is emitted to mention that `$(BundleAssemblies)` is not supported, and
  * the resulting `.apk` is broken and doesn't work.

### Steps to Reproduce

```
% dotnet new android -n android-ba
% cd android-ba
% dotnet build -p:BundleAssemblies=true
# run the app, or just look at the .apk contents
% unzip -l bin/Debug/net7.0-android/*-Signed.apk | grep assem
      660  02-03-2023 14:13   assemblies/rc.bin
``` 

Note that there is no `assemblies/assemblies.blob` file (c92702619f5fabcff0ed88e09160baf9edd70f41).

PR #7772, opened 2023-02-06
Remove support for mkbundle

grendello:
Fixes: https://github.com/xamarin/xamarin-android/issues/7764

Remove all usage of the `$(BundleAssemblies)` MSBuild property as well as all 
the code affected by it.  A couple of related properties remain in the `BuildApk` 
and `GeneratePackageManagerJava` tasks, because they are still assigned to by 
the debugger targets from `monodroid`:

  - Stop running bundle test apps on CI
  - Remove bundle unit tests
  - Remove the `CodeGen-MkBundle` project from the tests solution
  - Remove the `Xamarin.Android.MakeBundle-Tests` from the tests solution
  - Remove all native code (and configuration) related to mkbundle
  - Warn about `$(BundleAssemblies)` usage in .NET projects


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -144,9 +144,7 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
 				AotAssemblies = true,
 				PackageName = "com.xamarin.buildaotappwithspecialchars",
 			};
-			if (!Builder.UseDotNet) {
-				proj.BundleAssemblies = true;
-			}
+
```
pjcollins, line 147: I think there is an assert below this that pivots based on BundleAssemblies that may need updating.

pjcollins (commented):
I think we'd also want to remove https://github.com/xamarin/xamarin-android/blob/main/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/MakeBundleTests.cs

src/Xamarin.Android.Build.Tasks/Xamarin.Android.Common.targets
```diff
@@ -204,7 +204,6 @@ Copyright (C) 2011-2012 Xamarin. All rights reserved.
 	<MonoSymbolArchive Condition=" '$(MonoSymbolArchive)' == '' And '$(EmbedAssembliesIntoApk)' == 'True' And '$(DebugSymbols)' == 'True' And '$(Optimize)' == 'True'" >True</MonoSymbolArchive>
 	<MonoSymbolArchive Condition=" '$(MonoSymbolArchive)' == '' ">False</MonoSymbolArchive>
 
-	<BundleAssemblies Condition="'$(BundleAssemblies)' == ''">False</BundleAssemblies>
```
dellis1972, line 207: We might want to leave this one in place for the time being with a non conditional default value of False until we have a monodroid PR in place.
`<BundleAssemblies>False</BundleAssemblies>`

The reason being that MSBuild will error when trying to convert an empty string to a bool. 
Although if all the tests are passing then perhaps this isn't an issue.

grendello: I will restore this line as-is, then. If we force it to `false`, then the warning added elsewhere for this won't work.

build-tools/automation/azure-pipelines.yaml
```diff
@@ -99,209 +99,6 @@ stages:
   displayName: Smoke Tests
   dependsOn: mac_build
   jobs:
-  # Check - "Xamarin.Android (macOS > Tests > APKs Classic)"
-  - job: mac_apk_tests_legacy
```
pjcollins, line 103: I'm not sure that we want to remove this job entirely just yet, I think it would be preferable to port more of these test suites and get them running on the .NET test jobs first?

grendello: Yeah, I was wondering about that, that's why I made it a separate commit to make it easier to revert.

grendello: OTOH, the tests themselves (or YAML templates) aren't removed, so perhaps it's fine to remove the YAML invocation of the tests and restore them once they will have been ported?

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AotTests.cs
```diff
@@ -194,12 +192,7 @@ public void BuildBasicApplicationReleaseProfiledAotWithoutDefaultProfile ()
 						proj.OutputPath, $"{proj.PackageName}-Signed.apk");
 
 					var helper = new ArchiveAssemblyHelper (apk, usesAssemblyBlobs);
-					if (!Builder.UseDotNet) {
-						// BundleAssemblies=true
-						Assert.IsFalse (helper.Exists ("assemblies/UnnamedProject.dll"), $"UnnamedProject.dll should not be in the {proj.PackageName}-Signed.apk");
-					} else {
-						Assert.IsTrue (helper.Exists ("assemblies/UnnamedProject.dll"), $"UnnamedProject.dll should be in the {proj.PackageName}-Signed.apk");
-					}
+					Assert.IsTrue (helper.Exists ("assemblies/UnnamedProject.dll"), $"UnnamedProject.dll should be in the {proj.PackageName}-Signed.apk");
```
jonpryor, line 195: Shouldn't this assert be conditioned on `usesAssemblyBlobs`?

grendello: The helper hides that from the tests.

PR #7780, opened 2023-02-08
Add `_AndroidLibraryProjectImportsCache` to `FileWrites`

dellis1972:
On occasion we see the following error when building a Maui app.

```
MyMauiApp/Platforms/Android/AndroidManifest.xml : error APT2260: resource style/Maui.SplashTheme
(aka com.companyname.mymauiapp:style/Maui.SplashTheme) not found.
```

Looking in detail at te build log we notice that NONE of the resources in the `$(IntermediateOutputPath)\lp` folder 
are being included in the call to `aapt2`. Looking further we see two very important lines

`Skipping target "_ResolveLibraryProjectImports" because all output files are up-to-date with respect to the input files.`

```
Task "ReadLibraryProjectImportsCache" (TaskId:184)
Task Parameter:CacheFile=obj/Debug/net7.0-android/libraryprojectimports.cache (TaskId:184)
Task ReadLibraryProjectImportsCache (TaskId:184)
    CacheFile: obj/Debug/net7.0-android/libraryprojectimports.cache (TaskId:184)
   obj/Debug/net7.0-android/libraryprojectimports.cache does not exist. No Project Library Imports found (TaskId:184)
 Done executing task "ReadLibraryProjectImportsCache". (TaskId:184)
```

So in this case it seems that the `.cache` file which stores the list of resource directories was created, but was then
deleted later on. A search of our build system shows that this case file was never added to the `FileWrites` ItemGroup.
As a result it gets deleted, but the `.stamp` file which controls `_ResolveLibraryProjectImports` is left in place. We end up in a situation where the target which generates the cache will NEVER run again because its stamp file is present. 
Only a full Clean will fix this issue.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/IncrementalBuildTest.cs
@@ -857,6 +857,7 @@ namespace Lib2
 
 				b.BuildLogFile = "build2.log";
 				Assert.IsTrue (b.Build (proj), "second build should have succeeded.");
+				FileAssert.Exists (cacheFile);
 				var actual = ReadCache (cacheFile);
 				CollectionAssert.AreEqual (actual.Jars.Select (j => j.ItemSpec),
 					expected.Jars.Select (j => j.ItemSpec));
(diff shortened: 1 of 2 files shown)


jonathanpeppers (approved):
Is there a way to test? Maybe we could delete `$(_AndroidLibraryProjectImportsCache)` in a test and do an incremental build?

dellis1972:
> Is there a way to test? Maybe we could delete `$(_AndroidLibraryProjectImportsCache)` in a test and do an incremental build?

Not sure that would help in this senario. Deleting the file manually in a test would just cause the issue, once you get into that situation I'm not sure there is a way out of it. We could change the Outputs of `_ResolveLibraryProjectImports` to depend on the cache file rather than a stamp file. That would allow the build to "recover" if the file was deleted.

dellis1972:
This PR is more about stopping it happening in the first place, rather than recover from I guess.

PR #7784, opened 2023-02-10
[ci] Report issues in the API docs build log

pjcollins:
Context: https://github.com/xamarin/java.interop/issues/1071

Adds a step to the API docs build that will fail the job if any
problematic JavadocInfo messages are found in the log.


pjcollins:
This looked like it worked as expected: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7312024&view=logs&j=18eaad8b-760d-59fe-74fb-add94ccecaed&t=34b1759f-be57-527a-7eb1-410f327ad8fa

PR #7786, opened 2023-02-10
[build] Only build the latest API level

pjcollins:
Build timing from main:

    make jenkins  219.78s user 62.79s system 21% cpu 21:56.92 total

Build timing from this branch:

    make jenkins  178.66s user 47.24s system 36% cpu 10:22.81 total

It looks like this will shave off about 10 minutes from the Linux build and 10-15 minutes from the macOS build in CI as well.

Rather than remove the classic installers entirely, this seems like a good incremental step towards obsoletion.

The designer tests have been removed as they rely on API 30 or lower.

The wear test job has been moved to the emulator test stage rather than running in its own stage.

Multiple API levels are still used in the following projects, and could be cleaned up in the future:

 * build-tools/api-merge
 * build-tools/create-android-api


pjcollins:
@jpobst as an FYI there is probably some overlap here and in https://github.com/xamarin/xamarin-android/pull/7796, so we may want to hold this one until 7796 is ready?

jpobst (approved):
:heart_eyes: Love it!

Since this is ready, it's fine if this goes before #7796. I can fix the conflicts if needed.

pjcollins:
Updated commit message:

```markdown

Shaves ~10-15 minutes off of our Unix builds by only attempting to build
Mono.Android.dll for the latest API level.  A local `make jenkins` build
went from ~22 minutes to ~10 minutes with these changes, and CI builds
see a similar time saving.

This seems like a reasonable incremental step towards classic removal,
and allows us to continue to produce classic .vsix and .pkg installers
for the time being.

The designer test stages have been removed as they rely on API 30 or
lower.

The Wear OS test job has been moved into the emulator job stage rather
than running in its own stage.
```

jonathanpeppers (approved):

PR #7795, opened 2023-02-14
[Xamarin.Android.Build.Tasks] Fix Android Version Code for Release builds

dellis1972:
Fixes https://github.com/dotnet/maui/issues/11139

Users trying to use `maui` and the new `ApplicationVersion` in conjunction with `AndroidCreatePackagePerAbi` find that the version is NOT used in the final set of apks.

This is because when we use `AndroidCreatePackagePerAbi` we are totally ignoring the `ApplicationVersion` number. Instead we pick up the one from the `AndroidManifest.xml` `android:versionCode`. For maui users This is not obvious and is counter intuitive.

So lets use the `ApplicationVersion` when using `AndroidCreatePackagePerAbi`. All the old code will remain in place, if the `ApplicationVersion` is not set we will still fall back to `android:versionCode` and if that is not set default to `1`.


dellis1972:
@jonathanpeppers Hmm, this did not seem to break any tests 🤔

jonathanpeppers (commented):
I would say we just make this change and not worry too much.

But we should add a test showing what was broken before?

dellis1972:
> But we should add a test showing what was broken before?
Yup I'll get that sorted.

jonathanpeppers (approved):
I reran some tests to see if they turn green.

PR #7796, opened 2023-02-14
[Mono.Android] Bind Android 14 DP 1.

jpobst:
Context: https://developer.android.com/about/versions/14
Context: https://android-developers.googleblog.com/2023/02/first-developer-preview-android14.html

Android 14 Developer Preview 1 has been released.  The Android 14

  * Feb/Mar: Developer Previews
  * April/May: Unstable Betas
  * June/July: Stable Betas
  * ???: Final

### Acceptable Breakages

- Type `Dalvik.SystemInterop.DexFile` was un-deprecated. A new nested class was added to it. ([Documentation](https://developer.android.com/reference/dalvik/system/DexFile.OptimizationInfo?hl=en))

[0]: https://developer.android.com/about/versions/14/overview


src/Mono.Android.Runtime/Mono.Android.Runtime.csproj
```diff
@@ -31,11 +31,11 @@
 
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
 
-  <!-- Only build the .NET 6+ version of 'Mono.Android.Export.dll' for the latest stable Android version. -->
```
jpobst, line 34: @pjcollins Can you advise here? These lines seem to have been copied from `Mono.Android.Export.csproj`, but I think they are needed in `Export` because it is built for different API levels:

https://github.com/xamarin/xamarin-android/blob/main/build-tools/xaprepare/xaprepare/Application/GeneratedMakeRulesFile.cs#L150-L153

What needs to happen for `Mono.Android.Runtime` which is a "normal" project?

pjcollins: I don't think this needs to change unless it is causing an issue. I believe this was added because `Mono.Android.csproj` references `Mono.Android.Runtime.csproj`, and we don't want/need to rebuild this project every time `Mono.Android.csproj` is built.

jpobst: It did cause an issue when building the the unstable API-34 `Mono.Android.dll`:

```
c:\code\xamarin-android\src\Mono.Android\Android.Runtime\AndroidRuntime.cs(20,41): error CS0103: The name 'RuntimeConstants' does not exist in the current context [c:\code\xamarin-android\src\Mono.Android\Mono.Android.csproj::TargetFramework=net8.0]
c:\code\xamarin-android\src\Mono.Android\Android.Runtime\Logger.cs(47,43): error CS0246: The type or namespace name 'LogLevel' could not be found (are you missing a using directive or an assembly reference?) [c:\code\xamarin-android\src\Mono.Android\Mono.Android.csproj::TargetFramework=net8.0]
c:\code\xamarin-android\src\Mono.Android\Android.Runtime\Logger.cs(49,27): error CS0246: The type or namespace name 'LogLevel' could not be found (are you missing a using directive or an assembly reference?) [c:\code\xamarin-android\src\Mono.Android\Mono.Android.csproj::TargetFramework=net8.0]
c:\code\xamarin-android\src\Mono.Android\Android.Runtime\Logger.cs(8,19): error CS0246: The type or namespace name 'LogCategories' could not be found (are you missing a using directive or an assembly reference?) [c:\code\xamarin-android\src\Mono.Android\Mono.Android.csproj::TargetFramework=net8.0]
c:\code\xamarin-android\src\Mono.Android\Android.Runtime\Logger.cs(63,15): error CS0103: The name 'RuntimeConstants' does not exist in the current context [c:\code\xamarin-android\src\Mono.Android\Mono.Android.csproj::TargetFramework=net8.0]
```

([Source](https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7332897&view=results))

This seemed to be caused by the `Runtime` being missing due to the lack of compilation.

Commenting this optimization out fixes it, but I'm not sure if that's the solution we want.

pjcollins: Oh I see this file was added somewhat recently, in that case I think a better fix would be to use the same conditions defined in [Mono.Android.csproj](https://github.com/xamarin/xamarin-android/blob/main/src/Mono.Android/Mono.Android.csproj#L411-L420):

```xml
<!-- Only build the .NET 6+ version of 'Mono.Android.dll' for the default API level that is supported or higher. -->
  <PropertyGroup Condition=" '$(TargetFramework)' != 'monoandroid10' And '$(AndroidApiLevel)' &lt; '$(AndroidDefaultTargetDotnetApiLevel)' ">
    <BuildDependsOn></BuildDependsOn>
  </PropertyGroup>

  <PropertyGroup Condition=" '$(TargetFramework)' != 'monoandroid10' And '$(AndroidApiLevel)' &gt;= '$(AndroidDefaultTargetDotnetApiLevel)' ">
    <BuildDependsOn>
      $(BuildDependsOn);
      _CopyToPackDirs;
    </BuildDependsOn>
  </PropertyGroup>
```

jpobst: Cool, that seems to work.

jpobst:
@pjcollins We are going to use covariant return types in this binding level, which are not supported by Classic.  With your recent work on building fewer API levels for Classic in `main`, do you know how we could only build API-34 for .NET 8 until we get to the point where we aren't building Classic at all?

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
```diff
@@ -31,6 +31,7 @@ public sealed class CheckApiCompatibility : Task
 			{ "v12.0", "v11.0" },
 			{ "v12.1", "v12.0" },
 			{ "v13.0", "v12.1" },
+			{ "v13.0.99", "v13.0" },
```
jonpryor, line 34: Does this task *work* in a .NET Android world order?

build-tools/automation/azure-pipelines.yaml
```diff
@@ -58,7 +58,7 @@ variables:
 - name: ExcludedNUnitCategories
   value: '& cat != DotNetIgnore & cat != HybridAOT & cat != MkBundle & cat != MonoSymbolicate & cat != PackagesConfig & cat != StaticProject & cat != SystemApplication'
 - name: DefaultTestSdkPlatforms  # Comma-separated SDK Platform(s) to install on test agents (no spaces)
-  value: 33
+  value: 33,UpsideDownCake
```
jonpryor, line 61: UpsideDownCake is required for tests?

src/Mono.Android/Android.Graphics/PathIterator.cs
```diff
@@ -0,0 +1,16 @@
+using System;
+using Android.Runtime;
+
+namespace Android.Graphics
+{	partial class PathIterator
+	{
+
+#if ANDROID_34
+		// This implements an interface method that should be marked as 'default' but isn't.
```
jonpryor, line 9: *Might* be useful to link to: https://developer.android.com/reference/java/util/Iterator#remove()

tests/api-compatibility/acceptable-breakages-v13.0.99.txt
```diff
@@ -0,0 +1,15 @@
...
+MembersMustExist : Member 'public void Android.Views.PixelCopy.Request(Android.Views.SurfaceView, Android.Graphics.Rect, Android.Graphics.Bitmap, Android.Views.PixelCopy.IOnPixelCopyFinishedListener, Android.OS.Handler)' does not exist in the implementation but it does exist in the contract.
+MembersMustExist : Member 'public void Android.Views.PixelCopy.Request(Android.Views.Window, Android.Graphics.Bitmap, Android.Views.PixelCopy.IOnPixelCopyFinishedListener, Android.OS.Handler)' does not exist in the implementation but it does exist in the contract.
+MembersMustExist : Member 'public void Android.Views.PixelCopy.Request(Android.Views.Window, Android.Graphics.Rect, Android.Graphics.Bitmap, Android.Views.PixelCopy.IOnPixelCopyFinishedListener, Android.OS.Handler)' does not exist in the implementation but it does exist in the contract.
+CannotRemoveAttribute : Attribute 'System.ObsoleteAttribute' exists on 'Dalvik.SystemInterop.DexFile' in the contract but not the implementation.
+CannotAddAbstractMembers : Member 'public Java.Nio.ByteBuffer Java.Nio.ByteBuffer.Slice(System.Int32, System.Int32)' is abstract in the implementation but is missing in the contract.
```
jonpryor, line 9: This confuses me; we're using `compatVirtualMethod`, which should emit a *`virtual`* method, not an `abstract` method, so why is this warning emitted at all?

build-tools/automation/azure-pipelines.yaml
jpobst: Yes, there are some tests to ensure users can build with our unstable packages.

tests/api-compatibility/acceptable-breakages-v13.0.99.txt
jpobst: This was a rebase conflict error that I fixed incorrectly.  Updated the `acceptable-breakages` to the only actual entry.

build-tools/Xamarin.Android.Tools.BootstrapTasks/Xamarin.Android.Tools.BootstrapTasks/CheckApiCompatibility.cs
jpobst: Apparently not!  😁 

It appears we only run ApiCompat against the *stable* API platform on .NET.  This is something that will need to be fixed, but will be much easier once Classic has been removed from `main`.  I propose we prioritize this work for later.  I ran it manually locally, so the breakages included in this PR are correct, but CI will not be able to verify them until then.

At the very worst, the PR in ~June that marks API-34 stable will check it with today's ApiCompat.

PR #7800, opened 2023-02-16
Bump 'InstallAndroidDependenciesTest' to use PlatformTools 34.0.0

dellis1972:
We are seeing the `InstallAndroidDependenciesTest` test fail allot recently.
Looking at the logs we see the following warning.

```
warning : Dependency `platform-tools` should have been installed but could not be resolved. You can attempt to install it with: /Users/dean/Documents/Sandbox/Xamarin/WI1714603/bin/TestRelease/temp/InstallAndroidDependenciesTest/android-sdk/cmdline-tools/7.0/lib/sdkmanager-classpath.jar --install "platform-tools"
```

As a result `platform-tools` is not installed. This causes the `ResolveSdks` task to ignore the test sdk directory because `adb` is missing. So lets update the `AndroidSdkPlatformToolsVersion ` to use one that does work.


jonathanpeppers (approved):
The test in question passes now:

image

Only failure I see is one lane the emulator didn't boot.

Issue #7616, opened 2022-12-05
`AndroidNativeLibrary` doesn't trigged `JNI_OnLoad`


thenextman:
### Android application type

Classic Xamarin.Android (MonoAndroid12.0, etc.)

### Affected platform version

VS 2022 17.4, Xamarin.Android 13.1.0.1

### Description

I have a shared library (.so) written in C. The library exports `JNI_OnLoad` and `JNI_OnUnload`, confirmed with `readelf`:

>  478: 00000000002ff2f0   132 FUNC    GLOBAL DEFAULT   13 JNI_OnLoad
>  3111: 00000000002ff380   297 FUNC    GLOBAL DEFAULT   13 JNI_OnUnload

I reference the .so properly in my Xamarin.Android application using `AndroidNativeLibrary`.

However, at runtime `JNI_OnLoad` is never invoked.

### Steps to Reproduce

See above

### Did you find any workaround?

I _can_ invoke `JNI_OnLoad` manually by first using reflection to get the `java_vm` pointer:

```
[DllImport("MyLibrary", CallingConvention = CallingConvention.Cdecl)]
private static extern int JNI_OnLoad(IntPtr vm, IntPtr reserved);

static MyClass()
{
    var field = typeof(Android.Runtime.JNIEnv).GetField("java_vm", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Static);
    JNI_OnLoad((IntPtr)field.GetValue(null), IntPtr.Zero);
}
```

But this feels like an ugly workaround.

### Relevant log output

```shell
N/A
```

grendello:
@thenextman the library needs to be loaded explicitly, most likely, in the similar way [we load](https://github.com/xamarin/xamarin-android/blob/b165bc86c0026b17d0aacb1c1fd81afe9a05c762/src/java-runtime/java/mono/android/MonoPackageManager.java#L113) Xamarin.Android's runtime (`libmonodroid.so`).

thenextman:
@grendello Thanks for the input.

I do see my library (libMyLib.so) packaged inside {app.apk}/lib/{abi} alongside `libmonodroid` et all, as in your example. It exists for all supported ABIs.

However calling `Java.Lang.JavaSystem.LoadLibrary("MyLib");` yields:

> Java.Lang.UnsatisfiedLinkError: dlopen failed: library "libMyLib.so" not found
[...]

PR #7805, opened 2023-02-16
ClassLoader thread init

grendello:


src/monodroid/jni/monodroid-glue-internal.hh
```diff
@@ -385,6 +385,9 @@ namespace xamarin::android::internal
 		timing_period       jit_time;
 		FILE               *jit_log;
 		MonoProfilerHandle  profiler_handle;
+
+		inline static jclass    mono_android_Runtime;
```
jonpryor, line 389: What does it mean to make a static field `inline`?

grendello: It removes the necessity to define the field in a separate source file. Without it you'd have to add somewhere the following code:

```c++
jclass MonodroidRuntime::mono_android_Runtime;
```

Just a convenience really :)

src/Mono.Android/Java.Lang/JavaSystem.cs
```diff
@@ -0,0 +1,23 @@
...
+	{
+		[DllImport (RuntimeConstants.InternalDllName, CallingConvention = CallingConvention.Cdecl)]
+		static extern void monodroid_javasystem_loadLibrary (string libname);
+
+		public static void LoadLibrary (string libname)
+		{
+			monodroid_javasystem_loadLibrary (libname);
+		}
+
+		public static Task LoadLibraryAsync (string libname)
```
jonpryor, line 18: This should be wrapped in `#if ANDROID_34`, so that we don't inadvertently add new public API.

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -2635,6 +2641,27 @@ MonodroidRuntime::get_java_class_name_for_TypeManager (jclass klass)
...
 }
 
+void
+MonodroidRuntime::monodroid_javasystem_loadLibrary (const char *libname) noexcept
+{
+	JNIEnv *env = osBridge.ensure_jnienv ();
+	jmethodID loadLibraryID = env->GetStaticMethodID (mono_android_Runtime, "loadLibrary", "(Ljava/lang/String;)Z");
+	if (loadLibraryID == nullptr) {
+		log_error (LOG_DEFAULT, "Failed to look up the mono.android.Runtime.loadLibrary Java method");
+		return;
```
jonpryor, line 2651: Should this be an "abort the world" scenario?  If we can't find `Runtime.loadLibrary()` something is FUBAR.

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -2635,6 +2641,27 @@ MonodroidRuntime::get_java_class_name_for_TypeManager (jclass klass)
...
+	}
+
+	jstring libname_java = env->NewStringUTF (libname);
+	jboolean success = env->CallStaticBooleanMethod (mono_android_Runtime, loadLibraryID, libname_java);
+	if (success != 0) {
```
jonpryor, line 2656: We should `env->DeleteLocalRef(libname_java)` here, to prevent LREF leaks.  This function is called from Managed code, so we might not "return to Java" anytime soon (if ever), which is the only other time that JNI Local References will be freeed.

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -2635,6 +2641,27 @@ MonodroidRuntime::get_java_class_name_for_TypeManager (jclass klass)
...
+		// Loaded
+		return;
+	}
+
+	log_warn (LOG_DEFAULT, "System.loadLibrary failed, attempting to load the library directly");
```
jonpryor, line 2661: Might be handy if we printed out *what* library failed to load, in case that's needed for future debugging.

tests/LibraryLoadThread/app/MainActivity.cs
```diff
@@ -0,0 +1,36 @@
...
+
+using BCLThread = global::System.Threading.Thread;
+using BCLThreadStart = global::System.Threading.ThreadStart;
+
+namespace LibraryLoadThread;
+
+[Activity (Label = "@string/app_name", MainLauncher = true)]
+public class MainActivity : Activity
+{
+	BCLThread? thread;
```
jonpryor, line 15: Indentation in this file is inconsistent.  Please consistently use tabs.

src/java-runtime/java/mono/android/Runtime.java
```diff
@@ -35,6 +38,23 @@ public static native void initInternal (
...
+		current.setContextClassLoader (classLoader);
+	}
+
+	public static boolean loadLibrary (String libname)
+	{
+		try {
+			System.loadLibrary (libname);
+			return true;
+		} catch (java.lang.UnsatisfiedLinkError ex) {
+			Log.w ("monodroid", "Failed to load shared library with System.loadLibrary", ex);
```
jonpryor, line 54: It may be helpful if this also printed out `libName`, for correlation purposes.

jonpryor:
Context: https://github.com/xamarin/xamarin-android/issues/7616#issuecomment-1344497606

@grendello : does this *fix* #7616?  `tests/LibraryLoadThread` appears to attempt to a similar scenario as #7616, though `libthread-local.so` doesn't export `JNI_OnLoad`.  Perhaps it *should*?

src/Mono.Android/Java.Lang/JavaSystem.cs
grendello: `LoadLibraryAsync` is generated today, without any ifdefs, in `Java.Lang.JavaSystem.cs`:

```csharp
                // Metadata.xml XPath method reference: path="/api/package[@name='java.lang']/class[@name='System']/method[@name='loadLibrary' and count(parameter)=1 and parameter[1][@type='java.lang.String']]"
                [Register ("loadLibrary", "(Ljava/lang/String;)V", "")]
                public static unsafe void LoadLibrary (string libname)
                {
                        const string __id = "loadLibrary.(Ljava/lang/String;)V";
                        IntPtr native_libname = JNIEnv.NewString ((string?)libname);
                        try {
                                __args [0] = new JniArgumentValue (native_libname);
                                _members.StaticMethods.InvokeVoidMethod (__id, __args);
                        } finally {
                                JNIEnv.DeleteLocalRef (native_libname);
                        }
                }

                public static global::System.Threading.Tasks.Task LoadLibraryAsync (string libname)
                {
                        return global::System.Threading.Tasks.Task.Run (() => LoadLibrary (libname!));
                }
```

src/monodroid/jni/monodroid-glue.cc
grendello: I was wondering about that actually. We pass the string to Java in `CallStaticBooleanMethod` - won't it take ownership and release the LREF when the string is garbage collected?

grendello:
@jonpryor it doesn fix #7616 per se, alas. It makes it work because we "intercept" the `System.loadLibrary` call. We do the right thing here with the thread context, but I think there's something not fully right with the way the thread context is retrieved by the Java VM. I think that the thread context is not used in this scenario to retrieve the class loader, but rather the loader taken from `Class`, which we cannot set and which I hoped our `mono.android.Runtime` will have set correctly - meaning the one we retrieve in our package manager will be the same as in our Runtime class:

```java
ClassLoader loader  = context.getClassLoader ();
```

But apparently this is not true, and the context on Runtime is, probably, the boot system context which doesn't know about in-apk locations of shared libraries.  So the real fix would be to set that context on our `Runtime.Class`, but it's not settable, so we'd need to **create** `mono.android.Runtime` in the above `context`, but I don't know how to do that.  Figured that it can be left as material for another PR.

grendello:
@jonpryor also note that the test I added is not used.  This is on purpose, as I want it to test the original issue and fix for it, instead of our work around.  For that we need to figure out the class loader context described in the previous comment first.

src/Mono.Android/Java.Lang/JavaSystem.cs
jonpryor: Doh, my bad!

src/monodroid/jni/monodroid-glue.cc
jonpryor: No: https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html#global_and_local_references

> Local references are valid for the duration of a native method call, and are automatically freed after the native method returns.
> …
> However, there are times when the programmer should explicitly free a local reference. …
>   * A native method creates a large number of local references, although not all of them are used at the same time.

We're (kinda) in that bullet point, but differently so:

 1. Android only allows 512 JNI Local References (per thread) to exist at a time, but more importantly *for us*
 2. JNI Local References are only collected by the JVM when execution returns to Java.

(2) is potentially *rare* for us, particularly here: a `System.Threading.Thread` (POSIX Thread) running managed code which calls *into* Java will *never* "return" to Java.  Consequently, if *we* don't explicitly free them, the number of JNI Local References may constantly increase, resulting in a possible Android JNI Local Reference table overflow + app abort.

tests/LibraryLoadThread/native/library.cc
```diff
@@ -0,0 +1,7 @@
+#include <android/log.h>
+
```
jonpryor, line 2: I think we should also add a `JNI_OnLoad()` export to this file, as part of eventually fixing #7616.  Perhaps it doesn't need to be done before merging, but I also don't want to forget about adding `JNI_OnLoad()` either…

grendello: I agree, however I need time to dive deeper into how all this stuff with class loaders works, especially in the context of `mono.android.Runtime` so I have it on my TODO, won't forget :)

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -2551,6 +2557,29 @@ MonodroidRuntime::get_java_class_name_for_TypeManager (jclass klass)
...
+	jboolean success = env->CallStaticBooleanMethod (mono_android_Runtime, loadLibraryID, libname_java);
+	env->DeleteLocalRef (libname_java);
+
+	if (success != 0) {
+		// Loaded
+		return;
+	}
+
+	log_warn (LOG_DEFAULT, "System.loadLibrary failed for '%s', attempting to load the library directly", libname);
+	monodroidRuntime.monodroid_dlopen (libname, MONO_DL_LOCAL, nullptr, nullptr);
```
jonpryor, line 2580: Can this fail?  (Presumably it can/should, e.g. when someone does `JavaSystem.LoadLibrary("does-not-exist")`.

If it does fail, what should happen?

It feels like `JavaSystem.LoadLibrary()` should re-raise the *original* Java-side exception if this call fails, and to do *that* we'd have to rework things, e.g. have `Runtime.loadLibrary()` return a `java.lang.Throwable` instance and use `JNIEnv::CallStaticObjectMethod()` to invoke it.

What should happen *on success*?  The original report was that `JNI_OnLoad()` wasn't invoked.  Will `monodroid_dlopen()` invoke `JNI_OnLoad()`?  If not, should *we* look for and invoke it?

src/monodroid/jni/monodroid-glue.cc
```diff
@@ -2551,6 +2557,29 @@ MonodroidRuntime::get_java_class_name_for_TypeManager (jclass klass)
...
+{
+	JNIEnv *env = osBridge.ensure_jnienv ();
+	jmethodID loadLibraryID = env->GetStaticMethodID (mono_android_Runtime, "loadLibrary", "(Ljava/lang/String;)Z");
+	if (loadLibraryID == nullptr) {
+		log_fatal (LOG_DEFAULT, "Failed to look up the mono.android.Runtime.loadLibrary Java method");
+		Helpers::abort_application ();
+	}
+
+	jstring libname_java = env->NewStringUTF (libname);
+	jboolean success = env->CallStaticBooleanMethod (mono_android_Runtime, loadLibraryID, libname_java);
```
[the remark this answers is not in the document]
grendello: Yes, it was intentional. I wanted to handle only what we can actually handle here, and what we can act on. The rest I figured should be handled at the call site.

src/monodroid/jni/monodroid-glue.cc
grendello: If that load fails, then the attempt to use the library's interfaces will cause an error later on in the application. I thought that trying to throw a **Java** exception from here and in this situation would be quite complicated and, frankly, not worth the effort.  Also aborting the application for failure to find the library would be, I think, not desired here.  Letting the application fail because of the missing library, I think, is a fair compromise, but of course not ideal.

grendello:
After discussion on Discord, we decided that it's better to close this PR and come back to it once we have more time to dig into Art to see how we can make our thread creation behave the same as Java's.

PR #7807, opened 2023-02-17
Remove support for classic XA from the native code

grendello:
- Remove Designer support
  - Remove Designer test CI jobs
  - Remove support for classic XA from the native runtime
  - Remove support for Designer from the native runtime
  - Stop building classic versions of libmonodroid
  - Stop building host versions of libmonodroid


MagicAndre1981:
so after this gets shipped we can no longer build classic xamarin(.forms) apps? I thought we still have over 1 year of support for classic xamarin

grendello:
@MagicAndre1981 `main` starts to remove classic code, but classic is going to be supported from a branch (`d17-5` I think).

MagicAndre1981:
ok, so 17-5 will be last version that supports classic apps? But if 17.4 is LTS I doubt that 17.5 will be LTS again. Delay it until end of 2023 and merge it here. So 2-3 months later it gets shipped nearly at end of support for the classic apps.

src/sqlite-xamarin/sqlite-xamarin.targets
```diff
@@ -17,7 +17,7 @@
       <_ConfigureSqliteCommands
```
jonathanpeppers, line 17: Can we delete all of `sqlite-xamarin`? (doesn't have to be this PR)

This is for `Mono.Data.Sqlite.dll`? We don't even ship that in .NET 6+?

jonathanpeppers:
> ok, so 17-5 will be last version that supports classic apps? But if 17.4 is LTS I doubt that 17.5 will be LTS again. Delay it until end of 2023 and merge it here. So 2-3 months later it gets shipped nearly at end of support for the classic apps.

Just FYI whatever branch name we happen to use, doesn't mean that is what version of VS is supported. We can name branches anything -- it could actually be renamed to something like `xa` or `classic`, depending on what we decide.

src/sqlite-xamarin/sqlite-xamarin.targets
grendello: Looks like we can do that, but indeed, definitely not in this PR. I'm trying to remove only the bare minimum here, that relates to libmonodroid.so

build-tools/automation/azure-pipelines.yaml
```diff
@@ -94,208 +94,6 @@ stages:
...
-
-    - template: yaml-templates/apk-instrumentation.yaml
-      parameters:
-        configuration: $(ApkTestConfiguration)
-        testName: Xamarin.Android.Locale_Tests-Profiled-Aot
-        project: tests/locales/Xamarin.Android.Locale-Tests/Xamarin.Android.Locale-Tests.csproj
-        testResultsFiles: TestResult-Xamarin.Android.Locale_Tests-$(ApkTestConfiguration)-Profiled-Aot.xml
-        extraBuildArgs: /p:AotAssemblies=True /p:AndroidEnableProfiledAot=true
-        artifactSource: bin/Test$(ApkTestConfiguration)/Xamarin.Android.Locale_Tests-Signed.apk
-        artifactFolder: Profiled-Aot
```
pjcollins, line 173: We may want to hold off on deletion of the classic apk job until we have equivalents for .NET? I know there is a PR for the codegen related projects here https://github.com/xamarin/xamarin-android/pull/7682. Otherwise, we should log an issue for porting the remaining missing tests.

grendello: @pjcollins without removal of these fragments we'd constantly have test errors which would never work after merging of this PR. It's easy to restore these fragments when .NET equivalents of the affected tests are available.

jonpryor:
@MagicAndre1981: see also 618bd4abaefa40e9e8d263d493094af1711dacd0:

> Heads-up: xamarin-android/main is *becoming* .NET Android-only;
> Classic Xamarin.Android is only supported [through 2024-May-1][0];
> all current and future feature work is oriented toward .NET Android.
> Classic Xamarin.Android will only be fully buildable/supportable on
> the [d17-5 release branch][1].
> [0]: https://dotnet.microsoft.com/en-us/platform/support/policy/xamarin
> [1]: https://github.com/xamarin/xamarin-android/commits/d17-5

The d17-5 branch is Xamarin.Android 13.2.x, and Xamarin.Android 13.2.x (and derivatives thereof) will also be shipped in Visual Studio 17.6, 17.7, … until "Classic" Xamarin.Android is removed from Visual Studio entirely.

To answer [your question](https://github.com/xamarin/xamarin-android/pull/7807#issuecomment-1437315039):

> so after this gets shipped we can no longer build classic xamarin(.forms) apps? I thought we still have over 1 year of support for classic xamarin

Classic Xamarin.Android will be supported until May 2024, ~15 months from now.  *This does not require that the `main` branch continue to support "Classic" Xamarin.Android*.  All ongoing *feature* development is *only* for .NET Android, *not* Xamarin.Android.  Removing parts of Classic from the `main` branch makes our maintenance easier, and significantly improves our PR and CI build times. (Something like *half* the tests need to run once we stop caring entirely about Classic Xamarin.Android…)

Regarding [your other question](https://github.com/xamarin/xamarin-android/pull/7807#issuecomment-1437381104):

> 17-5 will be last version that supports classic apps?

The [d17-5 *branch*](https://github.com/xamarin/xamarin-android/commits/d17-5) is the last *branch* that supports Classic apps.

> But if 17.4 is LTS I doubt that 17.5 will be LTS again

17.5 may not be LTS, but you appear to be assuming that 17.5 will be the last release.  *I don't know* if it will be the last release, but if we make some projections from the [Visual Studio Long-Term Servicing Channel (LTSC) support page](https://learn.microsoft.com/en-us/visualstudio/productinfo/vs-servicing#long-term-servicing-channel-ltsc-support), we can infer that Visual Studio 17.6 *may also* be an LTS release, and *if* Visual Studio 17.6 exists, it *will* have a Classic Xamarin.Android SDK…built from our d17-5 branch (or derivative thereof).  If (when?) Visual Studio 17.7 exists, it will have a Classic Xamarin.Android SDK…built from our d17-5 branch (or derivative thereof).  If (when?) Visual Studio 17.8 exists, *it may be an LTS release* ***and*** it will have a Classic Xamarin.Android SDK…built from our d17-5 branch (or derivative thereof).

I hope you see and infer the pattern, and the pattern will continue until *at least* 2024-May-01, when support for Classic Xamarin.Android ends.  (Even then, it just means that *support* ends.  It may very well be possible to install a Classic Xamarin.Android SDK within Visual Studio, if only in an *unsupported* manner.  This is well over a year from now, we have no specifics to share.)

Classic Xamarin.Android is not (yet) being removed from Visual Studio.  When that happens, it will be appropriately announced.

build-tools/automation/azure-pipelines.yaml
pjcollins: We can file an issue to track those updates then if we plan to merge this PR soon.

MagicAndre1981:
but those branches you create for a VS version are practically dead and never really get updated with new fixes. #7644 is something I already do for years but this fix is now missing in that classic branch so others maybe be confused what to do when they hit it

jonathanpeppers:
> but those branches you create for a VS version are practically dead and never really get updated with new fixes. #7644 is something I already do for years but this fix is now missing in that classic branch so others maybe be confused what to do when they hit it

This PR only makes error messages better and doesn't fix anything. It wouldn't meet the bar for servicing anyway.

If there is a fix that isn't available in Xamarin.Android, one way we can have a good signal that it is important -- is if a few customers are commenting/asking for it.

build-tools/automation/azure-pipelines.yaml
jpobst: I filed a tracking issue for this here: https://github.com/xamarin/xamarin-android/issues/7906.

PR #7818, opened 2023-02-21
[ci] Use compliance stage template

pjcollins:
Context: https://github.com/xamarin/yaml-templates/blob/b186f6181c088c34bbf5dd210a343598f3e45489/security/full/v0.yml

Updates our compliance stage to use the universal compliance template,
ensuring that we run all required tasks and their latest versions.

The `policheck-rules-db.mdb` file has been replaced with `.gdnsuppress`
files, which are generated by the compliance build.  These files provide
a more convenient way to exclude specific "invalid" PoliCheck failures.
If a new PoliCheck failure arises in the future that should be excluded,
we can copy the entry for it from the `.gdnsuppress` file produced by
the build and add it to the file we have in our sources.


pjcollins:
Testing end to end here: https://devdiv.visualstudio.com/DevDiv/_build/results?buildId=7380824&view=logs&s=901df423-ce3d-5f89-3c82-d8f726b76ede

build-tools/automation/guardian/source.gdnsuppress
```diff
@@ -0,0 +1,236 @@
...
+      ],
+      "tool": "policheck",
+      "ruleId": "79668",
+      "justification": "Reference to find first set bit function.",
+      "createdDate": "2023-02-22 23:55:29Z",
+      "expirationDate": null,
+      "type": null
+    },
+    "d6b3df0b1d35cb4acec6a954acc145c9ec22041cd463b94ff080682c65a9bd62": {
+      "signature": "d6b3df0b1d35cb4acec6a954acc145c9ec22041cd463b94ff080682c65a9bd62",
```
jonathanpeppers, line 194: Only concern here -- is random security tooling going to complain about these? They *look* like API keys, where I could see future tools complaining about them.

jonathanpeppers (approved):

build-tools/automation/guardian/source.gdnsuppress
pjcollins: I don't think we'll have to worry about this, if some new scan complains about these we would hopefully only see it in a PR build when changing the underlying template.

PR #7824, opened 2023-02-23
Try xamarin/java.interop#1046

jonpryor:
Context: https://github.com/xamarin/java.interop/pull/1046

The assumption is that This Cannot Possibly Impact™ Android.
So… Does It Build™?

…no, it does not build, because xamarin/java.interop#1046 removes
the `net472` build of `jnimarshalmethod-gen.exe`, which causes
packaging to fail.  Noice.

Given that `jnimarshalmethod-gen.exe` only "worked" in Classic,
and that main is (slowly) dropping support for Classic (618bd4ab),
update the repo to stop packaging `jnimarshalmethod-gen.*` and
`Java.Runtime.Environment.*`.  This should fix the packaging errors.


pjcollins (commented):
The installer and build related changes here seem reasonable to me. If we want to add an error to the `_GenerateJniMarshalMethods` task we may want to provide localization for it, unless it is going to be short lived. On the other hand, if we don't plan to use this target again in the future should we remove it and its references instead?

PR #7826, opened 2023-02-24
[ci] Run nunit tests with stable .NET version

pjcollins:
Attempt to improve desktop test execution reliability by running our
test assemblies with a stable version of .NET by default.

The default value of the `useDotNet` yaml template parameters have been
flipped as part of continued efforts to remove classic test runs.

.NET 7.0 SDK provisioning has been updated to install the latest GA
version now that it is stable.


jonathanpeppers (approved):
This looks good if CI ends up green 👍

dellis1972 (approved):

PR #7831, opened 2023-02-27
Replace K4os.Hash.xxHash with System.IO.Hashing

eerhardt:
This allows us to remove an external dependency and use code from dotnet/runtime.


jonathanpeppers (commented):
I think we need to also delete this line:
 https://github.com/xamarin/xamarin-android/blob/977de6f9713ec98ba3f560d2abd091d1ed69aef2/build-tools/installers/create-installers.targets#L325
 
 Do we need to ship `System.IO.Hashing.dll` for .NET framework?

eerhardt:
> Do we need to ship System.IO.Hashing.dll for .NET framework?

We need to ship it for all TFMs. System.IO.Hashing doesn't ship in the .NET Core shared framework.

eerhardt:
> @jonathanpeppers
[Line endings?](https://github.com/xamarin/xamarin-android/pull/7831/commits/ff9de9d9b445338dc03e3f16929996cfb65ae75a)

What's the story with the line endings? I used https://github.dev/xamarin/xamarin-android/pull/7831 to update that file, which is just VS Code. Is the .editorconfig set up correctly?

jonathanpeppers:
I'll check the `.editorconfig`, it had a merge conflict because all the line endings changed.

jonathanpeppers (approved):

jonathanpeppers:
The test failure looks unrelated, going to merge this.

PR #7832, opened 2023-02-27
[ci] Reduce overhead for MSBuildIntegration unit test jobs.

jpobst:
There are 2 opportunities for speeding up the `MSBuildIntegration` unit test stage: removing `SmokeTests` that we also run in a different stage, and opting out of installing dependencies that these jobs do not need.

* Don't run `SmokeTests` in this stage, as they are run in the dedicated `Smoke Tests` stage.
* Refactor `start emulator` and `stop emulator` into reusable templates, and update them to use `dotnet build` instead of `mono`.
* Opt out of installing unneeded dependencies:
  * nunit-console
  * .NET 6
  * apkdiff
  * Unneeded Android SDK Platforms: 19,21,26,32
  * Mono: theoretically this stage isn't using Mono anymore. Regardless, [Mono is now included in the default Mac VM image](https://github.com/actions/runner-images/pull/6799) (it's a newer version (`6.12.0.188`) than we are installing (`6.12.0.145`)), so we don't need this unless we decide in the future we need a newer version than the default.
    * By updating our required Mono version to `6.12.0.188`, `xaprepare` will not need to downgrade it anymore on the current Mac images, saving time for *all* jobs that install Mono.

Current `main`:
image

This PR:
image


build-tools/automation/yaml-templates/shut-down-emulator.yaml
```diff
@@ -0,0 +1,19 @@
...
+  releaseAvdType:        # Device AVD to release, like 'android-wear'
+
+steps:
+# This uses our in-tree 'dotnet' because the system one may not have the 'android' workload installed.
+- template: run-dotnet-preview.yaml
+  parameters:
+    displayName: Shut down emulator
+    condition: always()
+    continueOnError: true
+    project: tests/Mono.Android-Tests/Runtime-Microsoft.Android.Sdk/Mono.Android.NET-Tests.csproj
```
pjcollins, line 15: We should be able to use https://github.com/xamarin/xamarin-android/blob/main/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/Emulator.csproj instead of one of the Android application projects to avoid having to use a custom MSBuild. I didn't see this file when I initially authored this.

build-tools/automation/yaml-templates/start-emulator.yaml
```diff
@@ -0,0 +1,18 @@
...
+# This uses our in-tree 'dotnet' because the system one may not have the 'android' workload installed.
+- template: run-dotnet-preview.yaml
+  parameters:
+    displayName: Start emulator
+    continueOnError: false
+    project: tests/Mono.Android-Tests/Runtime-Microsoft.Android.Sdk/Mono.Android.NET-Tests.csproj
+    ${{ if eq(parameters.installImage, true) }}:
+      arguments: -c $(XA.Build.Configuration) -t:"InstallAvdImage;AcquireAndroidTarget" -p:TestDeviceName=${{ parameters.installDeviceName }} -p:TestAvdApiLevel=${{ parameters.installAvdApiLevel }} -p:TestAvdAbi=${{ parameters.installAvdAbi }} -p:TestAvdType=${{ parameters.installAvdType }} -bl:$(System.DefaultWorkingDirectory)/bin/Test$(XA.Build.Configuration)/install-emulator-${{ parameters.installAvdApiLevel }}.binlog
+    ${{ else }}:
+      arguments: -c $(XA.Build.Configuration) -t:AcquireAndroidTarget -bl:$(System.DefaultWorkingDirectory)/bin/Test$(XA.Build.Configuration)/start-emulator.binlog
```
pjcollins, line 18: Could we use one YAML template for starting and stopping the emulator? I think we could combine these and add an extra parameter that determines whether we want to start or stop it.

pjcollins (approved):

build-tools/automation/yaml-templates/shut-down-emulator.yaml
jpobst: Nice, that is much better!

build-tools/automation/yaml-templates/start-emulator.yaml
jpobst: Updated.

PR #7835, opened 2023-02-28
Add Designer Tests to Nightly

dellis1972:


build-tools/automation/azure-pipelines-nightly.yaml
```diff
@@ -58,6 +68,173 @@ stages:
...
+        if ("$(Build.Reason)" -eq "PullRequest") {
+            $branchName = "$(System.PullRequest.TargetBranch)" -replace $branchPrefix, ""
+        }
+        if (("$branchName" -ne "main") -and ("$branchName" -notlike "d1*")) {
+            $branchName = "main"
+        }
+        Set-Location -Path $(System.DefaultWorkingDirectory)/UITools
+        git checkout $branchName
+        git submodule update -q --init --recursive
+      displayName: Clone and update UITools
```
pjcollins, line 110: This step would have us checkout the d17-5 branch of UITools, but I think we'd want to track their main against our d17-5? I think we could probably remove this check.

pjcollins (commented):
I wonder if we should remove the designer steps from the regular PR/build checks: https://github.com/xamarin/xamarin-android/blob/d17-5/build-tools/automation/azure-pipelines.yaml#L1273 as part of this PR? 

On the other hand, maybe it would be good to run against d17-5 designer source during regular PRs, and source from main during the nightly?

jonpryor:
The (not yet mentioned here) reason for adding Designer tests to the Nightly build is because we now rarely have new commits on the d17-5 branch; the last commit to d17-5 was over a month ago (a200af12c1e846626b8caebd926ac14c185f71ec)!  This in turn means that it has been over a month since the Designer tests ran in an environment which "matters".

This in turn means it is *possible* that UITools/main could have broken the Designer integration tests, and "nobody" would know until the next time we added a commit to d17-5.

This is the scenario we're trying to forestall by adding Designer tests to the nightly builds.

@pjcollins [suggested](https://github.com/xamarin/xamarin-android/pull/7835#pullrequestreview-1320282450):

> I wonder if we should remove the designer steps from the regular PR/build checks…as part of this PR?

I think that if we're doing a PR against d17-5, it should continue to run the Designer tests in the PR, to verify that the PR doesn't break the Designer.  We don't want to have to wait for a subsequent Nightly build to verify that we didn't break them.  d17-5 PR builds are also *very* infrequent, so this shouldn't cause any issues w/ our CI infrastructure.

> On the other hand, maybe it would be good to run against d17-5 designer source during regular PRs, and source from main during the nightly?

[Visual Studio 17.6 P1 has already been released](https://devblogs.microsoft.com/visualstudio/try-visual-studio-2022-v17-6-preview-1/), which ships with Xamarin.Android d17-5.  Thus we need to ensure that we continue to work with whatever UITools branch is in Visual Studio 17.6, which might not have branched yet.

It almost feels like we need to consider running the Designer integration tests against *every* UITools "release" branch (d17-5, main, d17-6+ when they exist…), but that also feels like overkill.  I *hope* that just main will be sufficient.

PR #7837, opened 2023-02-28
[Xamarin.Android.Build.Tasks] guard `AutoImport.props` against empty values

jonathanpeppers:
There is an email thread about ASP.NET projects in VS -- and something is potentially slowing down builds when optional workloads are installed.

One concern is our `AutoImport.props`, take for instance the following `foo.targets` file:

    <Project>
      <ItemGroup>
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.xml" />
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.axml" />
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.png" />
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.jpg" />
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.gif" />
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.webp" />
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\font\*.ttf" />
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\font\*.otf" />
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\font\*.ttc" />
        <AndroidResource Include="$(MonoAndroidResourcePrefix)\raw\*" Exclude="$(MonoAndroidResourcePrefix)\raw\.*" />
        <AndroidAsset Include="$(MonoAndroidAssetsPrefix)\**\*" Exclude="$(MonoAndroidAssetsPrefix)\**\.*\**" />
      </ItemGroup>
      <Target Name="Build" />
    </Project>

In this case `$(MonoAndroidResourcePrefix)` and `$(MonoAndroidAssetsPrefix)` will be blank, and we get:

    msbuild foo.targets -bl
    ...
      MSBUILD : warning MSB5029: The value "\**\.*\**" of the "Exclude" attribute in element <ItemGroup> in file "foo.targets (13,61)" is a wildcard that results in enumerating all files on the drive, which was likely not intended. Check that referenced properties are always defined.
    1 Warning(s)
    0 Error(s)
    Time Elapsed 00:02:45.14

I'm not sure this is causing an actual problem, but I think it's a good idea to add some safety here.

I will make a similar change in Xamarin.Legacy.Sdk.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
+++ b/src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
@@ -22,7 +22,7 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <Using Include="Android.OS.Bundle" Alias="Bundle" Platform="Android" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(EnableDefaultAndroidItems)' == 'true' and $([MSBuild]::VersionEquals($(TargetFrameworkVersion), '8.0')) ">
+  <ItemGroup Condition=" '$(MonoAndroidResourcePrefix)' != '' and '$(EnableDefaultAndroidItems)' == 'true' and $([MSBuild]::VersionEquals($(TargetFrameworkVersion), '8.0')) ">
     <!-- Default Resource file inclusion -->
     <!-- https://developer.android.com/guide/topics/resources/providing-resources -->
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.xml" />
@@ -35,8 +35,14 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\font\*.otf" />
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\font\*.ttc" />
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\raw\*" Exclude="$(MonoAndroidResourcePrefix)\raw\.*" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(MonoAndroidAssetsPrefix)' != '' and '$(EnableDefaultAndroidItems)' == 'true' and $([MSBuild]::VersionEquals($(TargetFrameworkVersion), '8.0')) ">
     <!-- Default Asset file inclusion -->
     <AndroidAsset Include="$(MonoAndroidAssetsPrefix)\**\*" Exclude="$(MonoAndroidAssetsPrefix)\**\.*\**" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(EnableDefaultAndroidItems)' == 'true' and $([MSBuild]::VersionEquals($(TargetFrameworkVersion), '8.0')) ">
     <!-- Default XPath transforms for bindings -->
     <TransformFile Include="Transforms*.xml" />
     <TransformFile Include="Transforms\**\*.xml" />


dellis1972 (approved):

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
```diff
@@ -22,7 +22,7 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <Using Include="Android.OS.Bundle" Alias="Bundle" Platform="Android" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(EnableDefaultAndroidItems)' == 'true' and $([MSBuild]::VersionEquals($(TargetFrameworkVersion), '8.0')) ">
+  <ItemGroup Condition=" '$(MonoAndroidResourcePrefix)' != '' and '$(EnableDefaultAndroidItems)' == 'true' and $([MSBuild]::VersionEquals($(TargetFrameworkVersion), '8.0')) ">
```
jonpryor, line 25: Given that XML attributes *can* contain newlines, at what point -- if an -- should this become a multi-line expression?  It's kinda long right now…

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
```diff
@@ -22,7 +22,7 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
...
     <!-- Default Resource file inclusion -->
     <!-- https://developer.android.com/guide/topics/resources/providing-resources -->
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\*\*.xml" />
```
jonpryor, line 28: Maybe it's too late for this change, but why support `$(MonoAndroidResourcePrefix)` at all anymore?  Consider `@(TransformFile)`,  `@(AndroidLibrary)`, `@(AndroidJavaSource)`, etc., below: *none* of those use `$(MonoAndroidResourcePrefix)`, and thus aren't susceptible to pulling in `C:\*\*`.

*Alternatively*, perhaps we could *require*/force `$(MonoAndroidResourcePrefix)` to *end with* "directory-separator-char", which would allow us to change these includes to e.g.:

```xml
<AndroidResource Include="$(MonoAndroidResourcePrefix)*\*.xml" />
```

which would *also* prevent us from trying to traverse `C:\*\*.xml`.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
```diff
@@ -35,8 +35,14 @@ https://github.com/dotnet/designs/blob/4703666296f5e59964961464c25807c727282cae/
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\font\*.otf" />
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\font\*.ttc" />
     <AndroidResource Include="$(MonoAndroidResourcePrefix)\raw\*" Exclude="$(MonoAndroidResourcePrefix)\raw\.*" />
+  </ItemGroup>
+
+  <ItemGroup Condition=" '$(MonoAndroidAssetsPrefix)' != '' and '$(EnableDefaultAndroidItems)' == 'true' and $([MSBuild]::VersionEquals($(TargetFrameworkVersion), '8.0')) ">
     <!-- Default Asset file inclusion -->
     <AndroidAsset Include="$(MonoAndroidAssetsPrefix)\**\*" Exclude="$(MonoAndroidAssetsPrefix)\**\.*\**" />
```
jonpryor, line 42: [Similar to above](https://github.com/xamarin/xamarin-android/pull/7837/files#r1120937802), but…

Why support `$(MonoAndroidAssetsPrefix)` at all anymore?  Consider `@(TransformFile)`,  `@(AndroidLibrary)`, `@(AndroidJavaSource)`, etc., below: *none* of those use `$(MonoAndroidAssetsPrefix)`, and thus aren't susceptible to pulling in `C:\**\*`.

*Alternatively*, perhaps we could *require*/force `$(MonoAndroidAssetsPrefix)` to *end with* "directory-separator-char", which would allow us to change these includes to e.g.:

```xml
<AndroidResource Include="$(MonoAndroidAssetsPrefix)**\*" />
```

which would *also* prevent us from trying to traverse `C:\**\*`.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
jonathanpeppers: Yeah, putting it on multiple lines would be OK. What should that look like? How is it indented?

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
jonathanpeppers: MAUI uses `$(MonoAndroidResourcePrefix)` so they can redirect it to `Platforms/Android/Resources`:

https://github.com/dotnet/maui/blob/965698b1dd996ed1b062cda3b360cdaa61039ddc/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.SingleProject.targets#L41-L42

But considering they use `EnsureTrailingSlash`:

https://github.com/dotnet/maui/blob/965698b1dd996ed1b062cda3b360cdaa61039ddc/src/Controls/src/Build.Tasks/nuget/buildTransitive/netstandard2.0/Microsoft.Maui.Controls.SingleProject.targets#L11-L12

So, then the only concern is a classic Xamarin.Android app has this in the project file:
```xml
<MonoAndroidResourcePrefix>Resources</MonoAndroidResourcePrefix>
<MonoAndroidAssetsPrefix>Assets</MonoAndroidAssetsPrefix>
```

But they should just delete these lines, and maybe it's fine to require a trailing `/`? It would make the code cleaner.

src/Xamarin.Android.Build.Tasks/Microsoft.Android.Sdk/Sdk/AutoImport.props
jonathanpeppers: After discussion and thinking about `AutoImports.props` MSBuild ordering, we decided to just format my code better...

Spec if needed: https://github.com/dotnet/designs/blob/main/accepted/2020/workloads/workload-resolvers.md#workload-props-files

PR #7850, opened 2023-03-02
[ci] Parallelize and reduce overhead of MSBuild test stage.

jpobst:
Bring the AzDO parallelization from https://github.com/xamarin/xamarin-android/pull/7804 and environment setup improvements from https://github.com/xamarin/xamarin-android/pull/7832 to `Xamarin.Android.Build.Tests.csproj` based test suites.  This includes both the main `MSBuild` stage and the `Smoke Tests` jobs.

Increases parallelization of all jobs as many were approaching ~90 minutes.

As there is no longer a place in the `MSBuild` stage to run `Xamarin.Android.Tools.Aidl-Tests` tests, it was moved to the `macOS > Tests > APKs .NET` stage.  This suite should be fine to only run on Mac and not Windows.  (Note this test assembly was also updated to .NET 7. This required moving it from `Xamarin.Android-Tests.sln` which is currently built with Mono which cannot build .NET 7+ projects. It now is built via `Xamarin.Android.sln`.)

## Ensure test count stayed the same

| Stage | `main` | This PR |
| - | - | - |
| Smoke > Mac | 75 | 75 |
| Smoke > Win | 75 | 75 |
| MSBuild > Mac | 631 | 637 |
| MSBuild > Win | 631 | 637 |

4 excess tests are `DotNetPublish(True,["net8.0", null, 33])`, which gets skipped anyways because the platform is `null`
2 excess tests are `MauiTargetFramework("net8.0",null,33)`, which gets skipped anyways because the platform is `null`

## Results

Before:
image

After:
image

Note runtimes are not very stable. Sometimes you get an agent that is significantly slower than normal. :man_shrugging:


build-tools/automation/azure-pipelines.yaml
```diff
@@ -440,23 +453,25 @@ stages:
...
-      job_name: win_dotnet_tests_0
-      job_suffix: One .NET
-      nunit_categories: '| (TestCategory = SmokeTests $(DotNetNUnitCategories))'
-      target_framework: $(DotNetStableTargetFramework)
-      provisionatorChannel: ${{ parameters.provisionatorChannel }}
+      testOS: Windows
+      jobName: win_smoke_msbuild_tests
+      jobDisplayName: Windows > Tests > MSBuild
+      agentCount: 2
+      testFilter: cat = SmokeTests
```
pjcollins, line 473: Given our environment provisioning overhead, the new splitting approach, and recent reductions to classic tests -- I wonder if it is still beneficial to run MSBuild "SmokeTests" separately. I think we could simplify things by removing these jobs and the extra filtering we have to do, and even potentially save on overall machine time per PR. We could opt to do this in a later PR though.

build-tools/automation/yaml-templates/stage-msbuild-tests.yaml
```diff
@@ -0,0 +1,26 @@
...
+      agentCount: 8
+      testFilter: cat != SmokeTests
+      
+  - template: run-msbuild-tests.yaml
+    parameters:
+      testOS: Windows
+      jobName: win_msbuild_tests
+      jobDisplayName: Windows > Tests > MSBuild
+      agentCount: 4
+      testFilter: cat != SmokeTests
```
pjcollins, line 26: I think we could move this block back into `azure-pipelines.yaml` to help cut down on the number of templates we use that aren't referenced more than once.

build-tools/automation/azure-pipelines.yaml
jpobst: It does seem like the full MSBuild test suite is run on nearly all PR's, so it's rare for us to only run Smoke Tests and thus get any savings from skipping the full run.

I think this is a good exploration for a future PR, as this one is already pretty big.

pjcollins (approved):
One potential change related to stage-msbuild-tests.yaml might be nice, otherwise this looks good to me.

build-tools/automation/yaml-templates/stage-msbuild-tests.yaml
jpobst: My thought was that `azure-pipelines.yaml` is way too big and it takes too long to find the relevant stage yaml.  If each stage was its own template file we could just go directly to the stage we want.

But that's a stylistic opinion and I'm not committed to it if others feel different.  :grin:

pjcollins: I wish yaml had some support around "go to reference" or "go to implementation", I'm often annoyed when working with multiple layers of nested templates having to jump back and forth between multiple page tabs rather than scrolling. That's also just my opinion though and am happy to go either way on this, maybe I can find some VS code extension to make my life easier...

build-tools/automation/azure-pipelines.yaml
```diff
@@ -55,6 +55,8 @@ variables:
   value: $[or(startsWith(variables['Build.SourceBranch'], 'refs/heads/release/'), startsWith(variables['System.PullRequest.TargetBranch'], 'release/'))]
 - name: DotNetNUnitCategories
   value: '& TestCategory != DotNetIgnore & TestCategory != HybridAOT & TestCategory != MkBundle & TestCategory != MonoSymbolicate & TestCategory != PackagesConfig & TestCategory != StaticProject & TestCategory != SystemApplication'
+- name: ExcludedNUnitCategories
+  value: '& cat != DotNetIgnore & cat != HybridAOT & cat != MkBundle & cat != MonoSymbolicate & cat != PackagesConfig & cat != StaticProject & cat != SystemApplication'
```
jonathanpeppers, line 59: This seems duplicate of `DotNetNUnitCategories`, why is the syntax here different?

jonathanpeppers (approved):

build-tools/automation/azure-pipelines.yaml
jpobst: `dotnet test` uses one syntax, `NUnit.Engine` uses a different syntax.

When you use `dotnet test` the adapter does the translation.  Because `dotnet-test-slicer` hits `NUnit.Engine` directly, we need to use NUnit's [test selection language](https://docs.nunit.org/articles/nunit/running-tests/Test-Selection-Language.html).

I put a little note here https://github.com/xamarin/xamarin-android/blob/f0a24bf67b3f0e56ed7a1d24c2e3789f542f0908/build-tools/automation/yaml-templates/run-sliced-nunit-tests.yaml#L3, but it doesn't bubble up to that variable specification.

PR #7853, opened 2023-03-03
New Xamarin.Android app analysis utility

grendello:
Unifies tmt, assembly decompressor and blob reader


grendello:
Moved to https://github.com/dotnet/android/pull/10029

PR #7869, opened 2023-03-09
[tests] Remove `XASdkProject`

pjcollins:
Replaces all instances of `XASdkProject` with `XamarinProject` types to
unify test project generation across all MSBuild suites.


dellis1972 (approved):

jonathanpeppers (approved):
Overall this is probably good to delete double-test code.

@pjcollins do you have a `.csproj` handy from what one of these projects generated?

I remember they seem to generate projects differently, related to `<ItemGroup>` -- so just wondering if something odd might happen there.

pjcollins:
> @pjcollins do you have a `.csproj` handy from what one of these projects generated?
> 
> I remember they seem to generate projects differently, related to `<ItemGroup>` -- so just wondering if something odd might happen there.

I made some changes around this by setting `EnableDefaultItems=true` in a handful of the tests that used XASdkProject, and updated the project writer to also ignore `EmbeddedResource` items in that case.

Here's a couple of projects generated differently in this PR: https://gist.github.com/pjcollins/a8d1902c2a955379858ef8d2947eb56f

There is some extra noise in the `Xamarin*Project` types, but not too much. We could consider setting `EnableDefaultItems=true` as a default, remove some extra properties, and/or remove the OutputPath items in the future.

PR #7873, opened 2023-03-10
[tests] `InstallAndroidDependenciesTest` can use `platform-tools` 34.0.1

jonpryor:
Context: bec42eff7e99a6c93eaa7681d50def138bbf9115
Context: 9752257da933b3709c6de420f0139b623f78fd85
Context: …
Context: https://dl-ssl.google.com/android/repository/repository2-3.xml

Whenever Google updates their `repository2-3.xml`, there is a chance that our `AndroidDependenciesTests.InstallAndroidDependenciesTest()` unit test will fail, as `repository2-3.xml` containsn *only one* platform-tools package version, and when that changes, out test breaks.

Tracking down the cause of this breakage is annoying, usually because when this happens we've forgotten that platform-tools package version changes are the primary reason `InstallAndroidDependencies()` fails. 😅

Update `AndroidDependenciesTests.InstallAndroidDependenciesTest()` to set `$(AndroidSdkPlatformToolsVersion)`=34.0.1, fixing the test, *and also* update the test so that when the wrong
`$(_AndroidSdkDirectory)` value is found, we read the *current version* of the platform-tools package from `repository2-3.xml` and include that information in our assertion message.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Xml;
 using System.Xml.Linq;
 using NUnit.Framework;
 using Xamarin.Android.Tools;
@@ -31,6 +32,7 @@ namespace Xamarin.Android.Build.Tests
 				var proj = new XamarinAndroidApplicationProject {
 					TargetSdkVersion = apiLevel.ToString (),
 				};
+				const string ExpectedPlatformToolsVersion = "34.0.1";
 				using (var b = CreateApkBuilder ()) {
 					b.CleanupAfterSuccessfulBuild = false;
 					string defaultTarget = b.Target;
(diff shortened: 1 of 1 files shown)


src/Xamarin.Android.Build.Tasks/Tests/Xamarin.Android.Build.Tests/AndroidDependenciesTests.cs
```diff
@@ -39,19 +41,50 @@ public void InstallAndroidDependenciesTest ()
...
+						// Is this because the platform-tools version changed (again?!)
+						try {
+							var currentPlatformToolsVersion = GetCurrentPlatformToolsVersion ();
+							if (currentPlatformToolsVersion != ExpectedPlatformToolsVersion) {
+								Assert.Fail ($"_AndroidSdkDirectory not set to new SDK path `{sdkPath}`, *probably* because Google's repository has a newer platform-tools package!  " +
+										$"repository2-3.xml contains platform-tools {currentPlatformToolsVersion}; expected {ExpectedPlatformToolsVersion}!");
+							}
+						}
+						catch (Exception e) {
+							// ignore
```
jonathanpeppers, line 60: I would maybe still `TestContext.WriteLine(e)`?

jonathanpeppers (approved):
This looks good to me if it fixes this test on CI. 👍

PR #7877, opened 2023-03-13
Add UnitTest docs 

dellis1972:


Documentation/workflow/UnitTests.md
```diff
@@ -0,0 +1,267 @@
...
+## Tools
+
+1. [NUnit](https://github.com/nunit/nunit)
+2. [XUnit](https://github.com/xunit/xunit)
+
+## Project Test Types
+
+1. Build Tests
+
+    There are a number of `Build` related tests which test various aspects of the `.NET Android` build system. This includes not only MSBuild Task tests but build integration tests. Since the SDK is consumed from MSBuild, almost all of the unit tests are MSBuild related.
```
jonpryor, line 13: Style-wise, I don't think we need to use "code" formatting for ".NET Android".  It's not code, it's not an identifier, it's the name of our product.

Documentation/workflow/UnitTests.md
```diff
@@ -0,0 +1,267 @@
...
+
+After building the repo you can make use of the `dotnet-local` script
+to run unit tests against the locally build SDK. The `dotnet-local` script is a wrapper around the custom `dotnet` installation which the
+build downloads and installs in `bin/Debug/dotnet` or `bin/Release/dotnet` (depending on your configuration).
+
+### MacOS/Linux
+
+To run ALL the build tests run.
+
+`dotnet-local.sh test bin/TestDebug/net7.0/Xamarin.Android.Build.Tests.dll --filter=Category!=DotNetIgnore`
```
jonpryor, line 42: An example using `--filter "Name~Whatever"` would also be good, a'la `Documentation/workflow/DevelopmentTips.md`.  (Repetition is good, occasionally.)

PR #7887, opened 2023-03-14
[Actions] Do not double quote the commit message.

mandel-macaque:


pjcollins (approved):
It looks like a similar change was already done on the MAUI side and appears to be working: https://github.com/dotnet/maui/actions/runs/4418421657, this seems good to me.

PR #7894, opened 2023-03-17
[vs-workload] Remove @(MultiTargetPackNames)

pjcollins:
Context: https://github.com/xamarin/yaml-templates/pull/251

The VS manifest files generated during MSI conversion will now be split up using the workload pack type rather than a substring match.


jonathanpeppers (approved):
This looks ok, but do we need to test a full build as it would happen on main?

PR #7899, opened 2023-03-20
Fix native code generation when marshal methods are disabled

grendello:
Fix native code generation when marshal methods are disabled

Disabling marshal method generation turns off a lot of native code
generation in the `MarshalMethodsNativeAssemblyGenerator` class, but the
class is also responsible for outputting correctly sized cache area (an
array of X pointers) to be used by the native runtime to cache pointers
to `MonoImage` instances.

Native runtime trusts the managed runtime to generate correct code and,
thus, does not verify the size of generated array.  This trust,
unfortunately, was broken because with marshal methods disabled, the
native code generator output cache array that was 0 entries in size,
thus leading to segfault when attempting to run the application.

Fix the issue and also parametrize one of the on-device tests to be
built twice, with marshal methods explicitly disabled and explicitly
enabled.

### DIFF
--- a/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
@@ -413,7 +413,7 @@ namespace Xamarin.Android.Tasks
 					Log
 				);
 			} else {
-				marshalMethodsAsmGen = new MarshalMethodsNativeAssemblyGenerator (uniqueAssemblyNames);
+				marshalMethodsAsmGen = new MarshalMethodsNativeAssemblyGenerator (assemblyCount, uniqueAssemblyNames);
 			}
 			marshalMethodsAsmGen.Init ();
 
--- a/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidApplicationProject.cs
+++ b/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidApplicationProject.cs
@@ -23,6 +23,8 @@ namespace Xamarin.ProjectTools
 		static readonly string default_main_activity_cs, default_main_activity_fs;
 		static readonly string default_android_manifest;
 
+		public bool? EnableMarshalMethods { get; set; } = null;
+
 		static XamarinAndroidApplicationProject ()
 		{
 			var folder = Builder.UseDotNet ? "DotNet" : "Base";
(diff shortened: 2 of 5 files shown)


jonathanpeppers (approved):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidApplicationProject.cs
```diff
@@ -41,6 +43,10 @@ public XamarinAndroidApplicationProject (string debugConfigurationName = "Debug"
 			: base (debugConfigurationName, releaseConfigurationName)
 		{
 			if (Builder.UseDotNet) {
+				if (EnableMarshalMethods.HasValue) {
```
dellis1972, line 46: This code runs in the constructor. My understanding is that if you set a property using the following style of code. The property set will happen after the constructor code has completed.

```
proj = new XamarinAndroidApplicationProject {
    EnableMarshalMethods = enableMarshalMethods,
};
```

The code after this new code is for setting the defaults which need to be in the csproj. 
You should probably use code like this https://github.com/xamarin/xamarin-android/blob/06ea40b567ac10b17cb4d024f08854c9cc31587f/src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidApplicationProject.cs#L114

dellis1972 (changes requested):

src/Xamarin.Android.Build.Tasks/Tests/Xamarin.ProjectTools/Android/XamarinAndroidApplicationProject.cs
grendello: Good catch sir :)

src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs
```diff
@@ -413,7 +413,7 @@ void AddEnvironment ()
 					Log
 				);
 			} else {
-				marshalMethodsAsmGen = new MarshalMethodsNativeAssemblyGenerator (uniqueAssemblyNames);
+				marshalMethodsAsmGen = new MarshalMethodsNativeAssemblyGenerator (assemblyCount, uniqueAssemblyNames);
```
jonpryor, line 416: Why is `assemblyCount` != `uniqueAssemblyNames.Count`?  In what instances will `assemblyCount` be greater than `uniqueAssemblyNames.Count`?  I guess `UseAssemblyStore`? https://github.com/xamarin/xamarin-android/blob/54a3f6e0bedc9f9b204a535aaf0f664e22285119/src/Xamarin.Android.Build.Tasks/Tasks/GeneratePackageManagerJava.cs#L272-L275

grendello: Yep, it's the assembly stores case

PR #7900, opened 2023-03-20
[build] remove .NET 6 support

jonathanpeppers:
Context: https://dotnet.microsoft.com/platform/support/policy/maui

> A major version of .NET MAUI receives support for a minimum of 6
> months after a successor (the next major release) ships. For
> example, .NET MAUI 6.0 will be supported for 6 months after .NET
> MAUI 7.0 ships. Similarly, .NET MAUI 7.0 will receive support for 6
> months after .NET MAUI 8.0 ships.

By the time .NET 8 GA ships, .NET 6 MAUI projects will not be supported.

Remove .NET 6 support from the `android` workload.

I removed `net6.0-android` parameters from various tests, making sure we still have .NET 7 and .NET 8 tests.

### DIFF
--- a/Configuration.props
+++ b/Configuration.props
@@ -43,7 +43,6 @@
     <Deterministic Condition=" '$(Deterministic)' == '' ">True</Deterministic>
     <LangVersion Condition=" '$(LangVersion)' == '' ">latest</LangVersion>
     <AndroidNet7Version Condition=" '$(AndroidNet7Version)' == '' ">33.0.46</AndroidNet7Version>
-    <AndroidNet6Version Condition=" '$(AndroidNet6Version)' == '' ">32.0.485</AndroidNet6Version>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(HostOS)' == '' ">
     <HostOS Condition="$([MSBuild]::IsOSPlatform('windows'))">Windows</HostOS>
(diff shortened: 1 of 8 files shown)


dellis1972 (approved):
I do wonder if there is a better way to drive what we include in the manifest. Some way that means fewer changes overall. Like a property or item group which xaprepare uses to "include" the required versions which are supported.

jonathanpeppers:
> I do wonder if there is a better way to drive what we include in the manifest. Some way that means fewer changes overall. Like a property or item group which xaprepare uses to "include" the required versions which are supported.

I think when we move to .NET 9, might be a good place to try doing this. At that point, I think we might have the patterns down. Right now things are still getting renamed/changing related to workloads.

pjcollins (approved):

maxkatz6:
Do I understand it correctly, that it's not possible to build a .NET 6 compatible project (single- or multi-target) while on .NET 8 SDK now? Even if the project itself has nothing to do with MAUI and uses native controls or other framework?

jonathanpeppers:
Yeah, this was following the support policy here: https://dotnet.microsoft.com/en-us/platform/support/policy/maui

It is basically the "modern support policy" brought forward from Xamarin.

If you need to build `net6.0-android` projects, you can continue to use the .NET 7 SDK for this.

dellis1972:
@tuyen-vuduc this might be a good stop gap https://github.com/xamarin/Xamarin.Legacy.Sdk

tuyen-vuduc:
@dellis1972 It fails to compile and package the nuget packages even if .NET8 isn't a part of supported frameworks. :( . The check should show as a warning, rather than a error. 

I'll try again to see if I can use global.json to configure to use .NET7 and previous ones.

jonathanpeppers:
Given the support policy: https://dotnet.microsoft.com/platform/support/policy/maui

Why do you need to support `net6.0-android` if we don't? You can continue to do this with the .NET 7 SDK in an unsupported fashion, but can you just target `net7.0-android` instead? It seems like you should file an issue describing your scenario instead of commenting on a pull request. Thanks!

